PR #4557, opened 2024-07-22
UDN: Program flows on default bridge

tssurya:
Credit to @dceara ; base code taken from https://github.com/ovn-org/ovn-kubernetes/pull/4476
Dependent on https://github.com/ovn-org/ovn-kubernetes/pull/4551 (only last 8 commits are valid here rest are from 4551)

Responsible for adding the node side changes needed to complete the last piece of pod2Egress for UDNs.
Currently being invoked only for primary networks when feature gate is ON
UTs and E2E for pod2Egress has been added


martinkennelly:
Waiting on this for EIP e2es because dcearas original PR for egress has a bug when creating and deleting networks. Let me know when it works.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -146,18 +158,56 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, networkID int, node *v1.
...
+	}, nil
 }
 
 // AddNetwork will be responsible to create all plumbings
 // required by this UDN on the gateway side
 func (udng *UserDefinedNetworkGateway) AddNetwork() error {
-	return udng.addUDNManagementPort()
+	err := udng.addUDNManagementPort()
+	if err != nil {
+		return fmt.Errorf("failed to add management port %s; error: %v", udng.GetNetworkName(), err)
```
martinkennelly, line 170: ..for network %s..

go-controller/pkg/node/gateway_udn.go
```diff
@@ -146,18 +158,56 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, networkID int, node *v1.
...
+	}
+	return nil
 }
 
 // DelNetwork will be responsible to remove all plumbings
 // used by this UDN on the gateway side
 func (udng *UserDefinedNetworkGateway) DelNetwork() error {
+	if udng.openflowManager != nil {
+		udng.openflowManager.delNetwork(udng.NetInfo)
+		// TODO (dceara): Do we need to reconcile all flows? Is this heavy?
```
martinkennelly, line 206: did you manage to answer this concern? Is there follow on work needed to optimise this issue?

go-controller/pkg/node/secondary_node_network_controller.go
```diff
@@ -28,15 +28,30 @@ type SecondaryNodeNetworkController struct {
...
 			stopChan:                        make(chan struct{}),
 			wg:                              &sync.WaitGroup{},
 		},
 	}
+	if util.IsNetworkSegmentationSupportEnabled() && snnc.IsPrimaryNetwork() {
+		node, err := snnc.watchFactory.GetNode(snnc.name)
+		if err != nil {
+			return nil, err
+		}
+		networkID, err := snnc.getNetworkID()
```
martinkennelly, line 45: This net ID is set async right - in-order to try to reduce log spam, do you see this being a problem? Should we exit with no error when net ID isnt set yet?

martinkennelly: also enrich the errors in this func. Dont want to see `Failed to create network xyz: object not found`

martinkennelly (commented):
Reviewed upto and include bfdf95e242adbbe0ec719a9b86a3e09da3869698

Very hard for me to review the flows you add but they look ok. Ill leave to Tim that part.

test/e2e/network_segmentation.go
```diff
@@ -364,6 +364,84 @@ var _ = Describe("Network Segmentation", func() {
...
+
+				DeferCleanup(func() {
+					deleteClusterExternalContainer(externalContainerName)
+				})
+			})
+
+			DescribeTable(
+				"can be accessed to from the pods running in the Kubernetes cluster",
+				func(netConfigParams networkAttachmentConfigParams, clientPodConfig podConfiguration) {
+					netConfig := newNetworkAttachmentConfig(netConfigParams)
```
martinkennelly, line 389: skip if not an IPv4 cluster

martinkennelly (commented):
finished

go-controller/pkg/node/openflow_manager.go
```diff
@@ -258,8 +258,11 @@ func bootstrapOVSFlows() error {
 
 	var bridge string
 	var patchPort string
-	// patch-br-int-to-<bridge name>_<node>
-	r := regexp.MustCompile("^patch-(.*)_.*?-to-br-int$")
+	// This needs to work with:
+	// - default network: patch-<bridge name>_<node>-to-br-int
```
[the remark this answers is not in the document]
dceara: Not really, it's `patch-<bridge-name>_<network-name>_<node-name>-to-br-int` for user defined networks.

go-controller/pkg/node/gateway_udn.go
dceara: It really depends on how many UDNs we expect to support.  Short answer: no, not yet. :)

go-controller/pkg/node/gateway_udn.go
```diff
@@ -16,6 +16,14 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
 
+const (
+	// VrfDeviceSuffix vrf device suffix associated with every user defined primary network.
+	VrfDeviceSuffix = "-vrf"
```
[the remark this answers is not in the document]
tssurya: good point! its rebase snafu

go-controller/pkg/node/gateway_localnet.go
```diff
@@ -52,19 +52,29 @@ func newLocalGateway(nodeName string, hostSubnets []*net.IPNet, gwNextHops []net
 
 	if exGwBridge != nil {
 		gw.readyFunc = func() (bool, error) {
-			ready, err := gatewayReady(gwBridge.patchPort)
-			if err != nil {
-				return false, err
+			for _, netConfig := range gwBridge.netConfig {
```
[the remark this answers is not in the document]
tssurya: its a good question, but I think we are OK here because we access this only from 1 spot in code... before making this change it was gwBridge.patchPort which was accessed without a lock...
@dceara do you agree?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1619,13 +1619,14 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, e
...
+	bridge.Lock()
+	defer bridge.Unlock()
 	// Get ofport of patchPort
-	ofportPatch, stderr, err := util.GetOVSOfPort("get", "Interface", bridge.patchPort, "ofport")
-	if err != nil {
-		return fmt.Errorf("failed while waiting on patch port %q to be created by ovn-controller and "+
-			"while getting ofport. stderr: %q, error: %v", bridge.patchPort, stderr, err)
+	for _, netConfig := range bridge.netConfig {
+		if err := netConfig.setBridgeNetworkOfPortsInternal(); err != nil {
+			return err
```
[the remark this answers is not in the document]
tssurya: done! yup included the patch port name which has the network name since we don't have the direct network information here

go-controller/pkg/node/openflow_manager.go
```diff
@@ -223,18 +210,23 @@ func (c *openflowManager) updateBridgeFlowCache(subnets []*net.IPNet, extraIPs [
...
+			return fmt.Errorf("failed to get ofport of %s, stderr: %q: %w", netConfig.patchPort, stderr, err)
 
-	}
-	if ofPortPatch != curOfportPatch {
-		klog.Errorf("Fatal error: patch port %s ofport changed from %s to %s",
-			patchIntf, ofPortPatch, curOfportPatch)
-		os.Exit(1)
+		}
+		if netConfig.ofPortPatch != curOfportPatch {
+			klog.Errorf("Fatal error: patch port %s ofport changed from %s to %s",
```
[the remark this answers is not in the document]
tssurya: ditto here.. given netInfo is not passed here.... from the patchport name we should be able to figure out the network name

go-controller/pkg/node/gateway_udn.go
tssurya: yea I spoke to @trozet about this.. he mentioned he is not concerned around this because flows are atomically added and deleted....

tssurya: if tim agrees then I could remove the comment

go-controller/pkg/ovn/gateway.go
```diff
@@ -1172,9 +1172,8 @@ func (gw *GatewayManager) syncNodeGateway(
...
 
 func physNetName(netInfo util.NetInfo) string {
-	physnetName := types.PhysicalNetworkName
-	if netInfo.IsSecondary() {
-		physnetName = netInfo.GetNetworkName()
+	if netInfo.IsDefault() || netInfo.IsPrimaryNetwork() {
+		return types.PhysicalNetworkName
 	}
-	return physnetName
+	return netInfo.GetNetworkName()
```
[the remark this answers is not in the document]
tssurya: we are not supporting pod2Egress for secondary networks.. so for now support is only being added for primary and for that @dceara mentioned its enough we keep it as physnet... maybe even for 2ndary networks we only actually ever need physnet because our bridge is the same ultimately.. the OVN networks are N mapped to 1 physnet right?

go-controller/pkg/node/gateway_udn_test.go
```diff
@@ -224,4 +337,144 @@ var _ = Describe("UserDefinedNetworkGateway", func() {
...
+			},
+		)
+		fakeClient := &util.OVNNodeClientset{
+			KubeClient: kubeFakeClient,
+		}
+
+		stop := make(chan struct{})
+		wf, err := factory.NewNodeWatchFactory(fakeClient, nodeName)
+		Expect(err).NotTo(HaveOccurred())
+		wg := &sync.WaitGroup{}
```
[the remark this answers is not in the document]
tssurya: this is a good point... I actually just took it from gateway_init_linux_test.go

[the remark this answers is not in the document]
tssurya: need to come back to fix this comment...

test/e2e/network_segmentation.go
tssurya: for now all net-seg lanes are dualstack.. we should know pretty well if we hit issues on v4 :D

go-controller/pkg/ovn/gateway.go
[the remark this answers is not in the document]
maiqueb: no, for secondary (localnet) networks you *can* have configurable bridges. The user is responsible for setting up the underlay network (e.g. create the OVS bridge, attach & configure interface to the bridge, vlan mumbo-jumbo, etc), and then they'll configure the mappings accordingly - something like (assuming kubernetes-nmstate is used): 

```yaml
    ovn:
      bridge-mappings:
      - localnet: secret-net
        bridge: mytinywinybr 
        state: present 
```

go-controller/pkg/node/openflow_manager.go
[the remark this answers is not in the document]
dceara: @maiqueb has a good point here: https://github.com/ovn-org/ovn-kubernetes/commit/bbb476b27cc32c74d2df686f3a7f0bd2d0889b5b#r145160319

The comment below is wrong.  It should be:
```
// but not with:
// - secondary network: patch-<bridge name>_<network name>_<node>-to-br-int
```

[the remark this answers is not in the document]
maiqueb: FWIW localnet networks don't even have the `_<node>` part. They have an `ovn_localnet-to-br-int` suffix instead.

go-controller/pkg/node/gateway_localnet.go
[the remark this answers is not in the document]
dceara: It does look weird to access gwBridge internals without the lock.  OTOH, we just created this bwBridge, 8 lines above, there's nobody else that can access it.

I'm not sure though what would look best from a golang perspective..  Maybe a way to pass a callback to a new `bridgeConfiguration` method that iterates on `bridgeConfiguration.netConfig` and calls the callback for each entry?  A sort of [fold function](https://en.wikipedia.org/wiki/Fold_(higher-order_function) for `bridgeConfiguration`.   That could take the lock internally.

go-controller/pkg/node/openflow_manager.go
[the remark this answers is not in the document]
dceara: > FWIW localnet networks don't even have the `_<node>` part. They have an `ovn_localnet-to-br-int` suffix instead.

Which is right, we don't want to touch localnet networks patch ports.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -42,6 +42,87 @@ type UserDefinedNetworkGateway struct {
...
+	b.Lock()
+	defer b.Unlock()
+	netConfigs := make([]bridgeUDNConfiguration, len(b.netConfig))
+	for _, netConfig := range b.netConfig {
+		netConfigs = append(netConfigs, *netConfig)
+	}
+	return netConfigs, b.uplinkName, b.ofPortPhys
+}
+
+func (b *bridgeConfiguration) addNetworkBridgeConfig(nInfo util.NetInfo, masqCTMark uint) {
```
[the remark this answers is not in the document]
martinkennelly: why modify the networks bridge config then in the func? Wouldnt it be an error if someone tried to add a network with the same name while one already exists?

go-controller/pkg/node/secondary_node_network_controller.go
martinkennelly: done for enriching error msg

go-controller/pkg/node/openflow_manager.go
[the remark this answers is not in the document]
martinkennelly: fixed the secondary networks patch port string

go-controller/pkg/node/gateway_udn.go
martinkennelly: done by Surya already

go-controller/pkg/node/gateway_udn.go
martinkennelly: Removed it in latest commit. Still need Tims ack.

go-controller/pkg/ovn/gateway.go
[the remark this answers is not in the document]
martinkennelly: @maiqueb do youve docs to setup that? Id like to do it myself some day.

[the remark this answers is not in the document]
maiqueb: @martinkennelly 
https://docs.openshift.com/container-platform/4.16/networking/multiple_networks/configuring-additional-network.html#configuring-additional-network_configuration-additional-network-interface

and there are these upstream resources for it:
- https://github.com/maiqueb/fosdem2024-sdn?tab=readme-ov-file#localnet-scenario---accessing-the-physical-network
- https://maiqueb.github.io/blog/2023/11/23/openshift-microsegmentation.html#scenario

martinkennelly:
Looking into e2e test failures now.

martinkennelly:
I am able to reproduce by running `with L2 dualstack primary UDN` then `by one pod with a single IPv4 address over a layer3 network` then seeing error:

`1134 openflow_manager.go:92] Failed to add flows, error: exit status 1, stderr, ovs-ofctl: -:6: 0/0: invalid or unknown port for in_port`

martinkennelly:
Rebased - no changes

martinkennelly:
Needed to rebase because conflict with main.

martinkennelly:
Looks like i failed to find the root cause of the e2e net seg failure - it worked locally (tm). Will try again tomorrow.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -43,6 +43,12 @@ type UserDefinedNetworkGateway struct {
 	// masqCTMark holds the mark value for this network
 	// which is used for egress traffic in shared gateway mode
 	masqCTMark uint
+	// stores a copy of default network's gateway so that
```
trozet, line 46: I dont think this is a copy. It looks like the pointer is passed all the way to this struct.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -140,8 +146,13 @@ func setBridgeNetworkOfPorts(bridge *bridgeConfiguration, netName string) error
 	return netConfig.setBridgeNetworkOfPortsInternal()
 }
 
-func NewUserDefinedNetworkGateway(netInfo util.NetInfo, networkID int, node *v1.Node, nodeLister listers.NodeLister,
-	kubeInterface kube.Interface, vrfManager *vrfmanager.Controller) *UserDefinedNetworkGateway {
+func NewUserDefinedNetworkGateway(netInfo util.NetInfo, networkID int, node *v1.Node,
+	nodeLister listers.NodeLister, kubeInterface kube.Interface, vrfManager *vrfmanager.Controller,
+	defaultNetworkGateway Gateway) (*UserDefinedNetworkGateway, error) {
```
trozet, line 151: isn't it kind of odd we pass an interface here to assert it is a type gateway struct later, and none of the interface methods:
```
type Gateway interface {
	informer.ServiceAndEndpointsEventHandler
	Init(<-chan struct{}, *sync.WaitGroup) error
	Start()
	GetGatewayBridgeIface() string
	SetDefaultGatewayBridgeMAC(addr net.HardwareAddr)
	Reconcile() error
}
```

are needed by UserDefinedNetworkGateway. It seems like all UserDefinedNetworkGateway wants is a gateway struct so it can access openflowmanager. In that case I think it makes more sense that either just a gateway struct is passed to this function, or the interface is properly extended to expose generic methods and UserDefinedNetworkGateway should use the interface.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -167,6 +179,32 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	if err != nil {
 		return fmt.Errorf("could not add VRF %d for network %s, err: %v", vrfTableId, udng.GetNetworkName(), err)
 	}
+	if udng.openflowManager != nil {
```
trozet, line 182: when do we expect this to be nil? if gateway is disabled? If thats the only case then we should check config.Gateway.Mode and error if it is not disabled && we have no openflowManager? Otherwise couldn't we silently skip programming a network?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -167,6 +179,32 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
...
+			return true, nil
+		}
+		postFunc := func() error {
+			if err := udng.Reconcile(); err != nil {
+				return fmt.Errorf("failed to reconcile flows on bridge for network %s; error: %v", udng.GetNetworkName(), err)
+			}
+			return nil
+		}
+		waiter.AddWait(readyFunc, postFunc)
+		if err := waiter.Wait(); err != nil {
```
trozet, line 204: so using wait we could wait up to 300 seconds and block the network manager from creating any other networks (cause threadiness is 1).

go-controller/pkg/node/gateway_udn.go
trozet: It's not heavy for openflow reconciliation. We just update the OF cache and then next time (in 5 seconds or next time its requested) OF flows will be atomically updated. Reconcile however does ask for all services to be requeued for nodePortWatcher to process every key and then also update openflow flows.

If a network gets deleted, we have to update all the flows including service flows, so I dont see anyway around this. I don't think it will be that bad, other than nodePortWatcher may increase its workload if lots of networks are being created/deleted and there are a lot of services in the cluster.

trozet (changes requested):
made it up to Fix existing flows on br-ex to be network aware

Need to look closely at the openflow flows

go-controller/pkg/node/gateway_udn_test.go
```diff
@@ -241,4 +353,148 @@ var _ = Describe("UserDefinedNetworkGateway", func() {
...
+					}
+				}
+			}
+			Expect(udnFlows).To(Equal(0))
+			Expect(len(udnGateway.openflowManager.defaultBridge.netConfig)).To(Equal(1)) // only default network
+
+			Expect(udnGateway.AddNetwork()).To(Succeed())
+			flowMap = udnGateway.gateway.openflowManager.flowCache
+			Expect(len(flowMap["DEFAULT"])).To(Equal(59))                                // 14 UDN Flows are added by default
+			Expect(len(udnGateway.openflowManager.defaultBridge.netConfig)).To(Equal(2)) // default network + UDN network
```
trozet, line 646: nice test

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,84 @@ var _ = Describe("Network Segmentation", func() {
...
+		DescribeTable(
+			"can be accessed to from the pods running in the Kubernetes cluster",
+			func(netConfigParams networkAttachmentConfigParams, clientPodConfig podConfiguration) {
+				netConfig := newNetworkAttachmentConfig(netConfigParams)
+
+				netConfig.namespace = f.Namespace.Name
+				clientPodConfig.namespace = f.Namespace.Name
+
+				By("creating the attachment configuration")
+				_, err := nadClient.NetworkAttachmentDefinitions(f.Namespace.Name).Create(
```
trozet, line 540: can we use the UDN CRD instead now?

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,84 @@ var _ = Describe("Network Segmentation", func() {
...
+
+				By("instantiating the client pod")
+				clientPod, err := cs.CoreV1().Pods(clientPodConfig.namespace).Create(
+					context.Background(),
+					generatePodSpec(clientPodConfig),
+					metav1.CreateOptions{},
+				)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(clientPod).NotTo(BeNil())
+
```
trozet, line 555: check that the Pod is connected to the UDN before you check the network connectivity

test/e2e/network_segmentation.go
```diff
@@ -567,7 +567,7 @@ var _ = Describe("Network Segmentation", func() {
 					return connectToServer(clientPodConfig, externalIpv4, port)
 				}, 2*time.Minute, 6*time.Second).Should(Succeed())
 			},
-			Entry("by one pod with a single IPv4 address over a layer2 network",
+			XEntry("by one pod with a single IPv4 address over a layer2 network",
```
trozet, line 570: maybe just add // TODO or FIXME and make sure we unskip this l2 egress PR @maiqueb @qinqon

trozet (changes requested):
finished review of all commits

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1321,11 +1332,11 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 			fmt.Sprintf("cookie=%s, priority=10, table=1, dl_dst=%s, actions=output:%s",
 				defaultOpenFlowCookie, bridgeMacAddress, ofPortHost))
 	}
-
+	defaultNetConfig := bridge.netConfig[types.DefaultNetworkName]
 	// table 2, dispatch from Host -> OVN
 	dftFlows = append(dftFlows,
 		fmt.Sprintf("cookie=%s, table=2, "+
-			"actions=set_field:%s->eth_dst,output:%s", defaultOpenFlowCookie, bridgeMacAddress, ofPortPatch))
+			"actions=set_field:%s->eth_dst,output:%s", defaultOpenFlowCookie, bridgeMacAddress, defaultNetConfig.ofPortPatch))
```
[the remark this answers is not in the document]
tssurya: Yes! Services flows fix will be done in new PR with proper e2e's

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1321,11 +1332,11 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
```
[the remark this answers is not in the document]
tssurya: Correct table4 and 5 are used for ETP=local.. planned in new PR

go-controller/pkg/node/gateway_localnet.go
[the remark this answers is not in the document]
tssurya: ack makes sense I guess better safe than sorry => +1 to confusing part and I didn't realize nodeIP Manager also edited stuff on bridge configuration..

go-controller/pkg/ovn/gateway.go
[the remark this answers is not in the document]
tssurya: all this need sto be added to ovn-kubernetes.io

go-controller/pkg/node/gateway_udn.go
tssurya: so this is a great point and something I myself found icky. But in DNCC i.e for default network's gateway code this is what we seem to be doing.. we use the Gateway interface and then assign the gateway struct to it, so in DNCC what I have is interface not the struct. That is why I had to resort to this but I could just straight up extract it as the struct.. lmc what I can do

go-controller/pkg/node/gateway_udn.go
tssurya: It was more of a precaution for code moving around places... It will be nil today for localnet topology .. I didn't want a panic in corner cases so checked this.. I can add a else Warning error to this so that we don't silently skip this -> good call out.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1255,31 +1263,46 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
-		// introduced specifically for replies to egress traffic not routed
-		// through the host
-		actions = fmt.Sprintf("check_pkt_larger(%d)->reg0[0],resubmit(,11)", maxPktLength)
-	}
+			if config.IPv4Mode {
+				// table 1, established and related connections in zone 64000 with ct_mark ctMarkOVN go to OVN
+				dftFlows = append(dftFlows,
+					fmt.Sprintf("cookie=%s, priority=100, table=1, ip, ct_state=+trk+est, ct_mark=%s, "+
+						"actions=%s",
+						defaultOpenFlowCookie, ctMarkOVN, actions))
```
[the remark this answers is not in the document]
tssurya: yup the first commit modifies flows to be network aware but still keeps it all at default patch port and the commit following that we modify this

go-controller/pkg/node/gateway_localnet.go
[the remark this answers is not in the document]
tssurya: done! added locks here

go-controller/pkg/node/gateway_udn.go
```diff
@@ -45,6 +45,87 @@ type UserDefinedNetworkGateway struct {
 	masqCTMark uint
 }
 
+// UTILS Needed for UDN (also leveraged for default netInfo) in openflowmanager
+
+func (c *openflowManager) getDefaultBridgePorts() ([]bridgeUDNConfiguration, string, string) {
```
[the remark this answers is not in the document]
tssurya: > also kind of weird that this is called getDefaultBridgePorts, but it returns a slice of bridgeUDNConfiguration. it looks like its only used in checkPorts, which is iterating through the bridgeUDNConfiguration slice to just look for the patch port. Maybe this should just return []string, string, string...where[]string is a slice of patch port names.

haha I started doing this and then realised that in checkPorts we actually need both ofPatchPort value and patchPort value so it has to be a slice of UDNConfiguration in the end; but you are right that its confusing, let me rename the util to `getDefaultBridgeNetworkConfigurations`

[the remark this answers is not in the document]
tssurya: > probably move these ofmanager functions into openflow_manager.go

yup done

go-controller/pkg/node/gateway_udn.go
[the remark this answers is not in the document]
tssurya: > why modify the networks bridge config then in the func? Wouldnt it be an error if someone tried to add a network with the same name while one already exists?

so that's not how we call it today, we are only calling this from addNetwork in openflow manager which calls addNetworkBridgeConfig so there is never a situation where we even consider updates; so we are not going to overwrite anything; I can change the logic to reflect this @dceara LMK if you agree with this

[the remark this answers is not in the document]
tssurya: I added a warning if found instead of updating the value... hopefully we don't expect the ofport to change mid way for a given network patch port.

go-controller/pkg/node/gateway_udn.go
tssurya: true, wrong wordings.. changing this.

go-controller/pkg/node/gateway_udn.go
tssurya: so I tried for extract the `gateway` struct from NCM itself before passing the interface along.. however `gateway` struct is not public outside of the node package only the interface is. That is why I passed the public interface and then extracted the struct from it.

go-controller/pkg/node/gateway_udn.go
tssurya: resolving this comment...we can do scale tests and improve from there if we see slowness.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1368,52 +1379,59 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, e
...
-			// SNATs these into egressIP prior to reaching external bridge.
-			// egressService pods will also undergo this SNAT to nodeIP since these features are tied
-			// together at the OVN policy level on the distributed router.
-			dftFlows = append(dftFlows,
-				fmt.Sprintf("cookie=%s, priority=105, in_port=%s, dl_src=%s, ip, pkt_mark=%s "+
-					"actions=ct(commit, zone=%d, nat(src=%s), exec(set_field:%s->ct_mark)),output:%s",
-					defaultOpenFlowCookie, ofPortPatch, bridgeMacAddress, ovnKubeNodeSNATMark,
-					config.Default.ConntrackZone, physicalIP.IP, ctMarkOVN, ofPortPhys))
+			for _, netConfig := range bridge.netConfig {
+				// table0, packets coming from egressIP pods that have mark 1008 on them
```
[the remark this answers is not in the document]
tssurya: So similar to services being a new PR egressIPs fix will be part o @martinkennelly 's PR, @martinkennelly FYI to take into consideration all the EIP related corner cases and flows.
This is currently not in this PR's scope which only focuses on making the flows network aware

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1368,52 +1379,59 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, e
...
-					"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:%s",
-					defaultOpenFlowCookie, ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
-					ctMarkOVN, ofPortPhys))
+				// table 0, packets coming from pods headed externally. Commit connections with ct_mark ctMarkOVN
+				// so that reverse direction goes back to the pods.
+				dftFlows = append(dftFlows,
+					fmt.Sprintf("cookie=%s, priority=100, in_port=%s, dl_src=%s, ip, "+
+						"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:%s",
+						defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
+						ctMarkOVN, ofPortPhys))
```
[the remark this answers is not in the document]
tssurya: yup

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1518,17 +1542,20 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, e
...
 			}
-			// table 0, drop packets coming from pods headed externally that were not SNATed.
-			dftFlows = append(dftFlows,
-				fmt.Sprintf("cookie=%s, priority=104, in_port=%s, %s, %s_src=%s, actions=drop",
-					defaultOpenFlowCookie, ofPortPatch, ipPrefix, ipPrefix, cidr))
+			for _, netConfig := range bridge.netConfig {
+				// table 0, drop packets coming from pods headed externally that were not SNATed.
+				dftFlows = append(dftFlows,
+					fmt.Sprintf("cookie=%s, priority=104, in_port=%s, %s, %s_src=%s, actions=drop",
+						defaultOpenFlowCookie, netConfig.ofPortPatch, ipPrefix, ipPrefix, cidr))
```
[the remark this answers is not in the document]
tssurya: This will happen in @martinkennelly 's PR so I have moved this flow back to only being for default network currently. @martinkennelly can take care of making it UDN aware in EIP PR.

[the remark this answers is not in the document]
tssurya: great catch tim!

I have moved this back to being:
```
for _, clusterEntry := range config.Default.ClusterSubnets {
			cidr := clusterEntry.CIDR
			ipPrefix := "ip"
			if utilnet.IsIPv6CIDR(cidr) {
				ipPrefix = "ipv6"
			}
			// table 0, drop packets coming from pods headed externally that were not SNATed.
			dftFlows = append(dftFlows,
				fmt.Sprintf("cookie=%s, priority=104, in_port=%s, %s, %s_src=%s, actions=drop",
					defaultOpenFlowCookie, defaultNetConfig.ofPortPatch, ipPrefix, ipPrefix, cidr))
		}
```

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1422,15 +1419,23 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, e
...
-						defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
-						ctMarkOVN, ofPortPhys))
+				if netConfig.masqCTMark == ctMarkOVN {
+					dftFlows = append(dftFlows,
+						fmt.Sprintf("cookie=%s, priority=100, in_port=%s, dl_src=%s, ip, "+
+							"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:%s",
+							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
+							netConfig.masqCTMark, ofPortPhys))
+				} else {
+					dftFlows = append(dftFlows,
```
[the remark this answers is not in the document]
tssurya: yup done

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1422,15 +1419,23 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, e
...
+						fmt.Sprintf("cookie=%s, priority=100, in_port=%s, dl_src=%s, ip, "+
```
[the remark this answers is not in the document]
tssurya: hmmm won't everything coming out of the patch port of that network be SNATed with masquerade IP?
cc @dceara

test/e2e/network_segmentation.go
tssurya: yeah this e2e was written long before we had the CRDs... I don't have time to work on this...as part of this PR
https://redhat-internal.slack.com/archives/C077NGF5HE1/p1723116215316809?thread_ts=1723104646.713079&cid=C077NGF5HE1
we can do it as FUP if its ok with you...
@npinaeva can you help convert this test to use CRD when this PR merges?

tssurya: https://github.com/ovn-org/ovn-kubernetes/issues/4606

tssurya: will be solved via FUP

test/e2e/network_segmentation.go
tssurya: lemme see how the other tests are doing this..

test/e2e/network_segmentation.go
tssurya: ack added a FIX ME

tssurya:
hmmm
the failing e2e's are passing for me locally :o:

```
Network Segmentation pod2Egress on a user defined primary network can be accessed to from the pods running in the Kubernetes cluster by one pod with a single IPv4 address over a layer3 network
/home/surya/go/src/github.com/ovn-org/ovn-kubernetes/test/e2e/network_segmentation.go:580
  STEP: Creating a kubernetes client @ 08/12/24 07:18:27.484
  I0812 07:18:27.484228 578638 util.go:506] >>> kubeConfig: /home/surya/ovn.conf
  STEP: Building a namespace api object, basename network-segmentation @ 08/12/24 07:18:27.484
  I0812 07:18:27.511742 578638 framework.go:275] Skipping waiting for service account
  STEP: creating the attachment configuration @ 08/12/24 07:18:28.073
  STEP: instantiating the client pod @ 08/12/24 07:18:28.083
  W0812 07:18:28.094332  578638 warnings.go:70] would violate PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "agnhost-container" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "agnhost-container" must set securityContext.capabilities.drop=["ALL"]), runAsNonRoot != true (pod or container "agnhost-container" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container "agnhost-container" must set securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
  STEP: asserting the client pod reaches the `Ready` state @ 08/12/24 07:18:28.095
  STEP: asserting the *client* pod can contact the server located outside the cluster @ 08/12/24 07:19:46.157
  I0812 07:19:46.157312 578638 builder.go:121] Running '/usr/bin/kubectl --server=https://127.0.0.1:37885 --kubeconfig=/home/surya/ovn.conf --namespace=network-segmentation-3374 exec client-pod -- curl --connect-timeout 2 172.18.0.5:9000'
  I0812 07:19:46.225049 578638 builder.go:146] stderr: "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100    60  100    60    0     0  18259      0 --:--:-- --:--:-- --:--:-- 20000\n"
  I0812 07:19:46.225088 578638 builder.go:147] stdout: "NOW: 2024-08-12 05:19:46.220428024 +0000 UTC m=+78.181129398"
  STEP: Destroying namespace "network-segmentation-3374" for this suite. @ 08/12/24 07:19:46.613
• [79.137 seconds]
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 1 of 299 Specs in 79.142 seconds
SUCCESS! -- 1 Passed | 0 Failed | 3 Pending | 295 Skipped
PASS | FOCUSED

Ginkgo ran 1 suite in 1m26.36855021s
Test Suite Passed
Detected Programmatic Focus - setting exit status to 197
```

go-controller/pkg/ovn/gateway.go
maiqueb, line 1178: don't you need to adjust the L3 unit tests for this to pass ? 

I've adjusted it in https://github.com/ovn-org/ovn-kubernetes/pull/4561/commits/02dc5b23aaba350a3045865e43129554f50a0072 FWIW.

tssurya:
@maiqueb : you mean https://github.com/ovn-org/ovn-kubernetes/pull/4557/commits/98c482c52a4323368d7d3e16e2e1ef311e9c8016 ? @martinkennelly already fixed it; that's how the unit tests are passing right now :)

tssurya:
I can squash it together in my next push

maiqueb:
> @maiqueb : you mean [98c482c](https://github.com/ovn-org/ovn-kubernetes/commit/98c482c52a4323368d7d3e16e2e1ef311e9c8016) ? @martinkennelly already fixed it; that's how the unit tests are passing right now :)

I'm simply saying the UT fix belongs with the code change.

 > I can squash it together in my next push

Please do.

go-controller/pkg/node/gateway_udn.go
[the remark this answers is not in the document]
dceara: As long as we make sure to delete the netConfig for a given network when we fail to provision the node part of things for that network then we should be fine.  But I'm not sure that's the case.

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
dceara: @tssurya I think so, except when we support ICNI (or whatever it's going to be called - exposing pod IP directly).  But yes, we do SNAT to the masquerade IP in OVN:
https://github.com/ovn-org/ovn-kubernetes/pull/4551/commits/2275260b7265a1674279f84670a27e152f739704#diff-7559bf2cf0fe2601ebf7df9b9d1984826e2dcdd13e489e23d80008280514a571R816

martinkennelly:
Tried testing this PR locally and I found when you create a L3 UDN primary + pod and then deleted both and I found the pod hung in terminating state with error message:
```
missing primary user defined network NAD
```
.. from CNI pkg

martinkennelly:
Looks like race between tearing down the pod and deletion of the NAD. If NAD is deleted first, it looks like we think the pod is configured for a CDN network.
```go

func (p *UserDefinedPrimaryNetwork) ensureActiveNetwork(namespace string) error {
	if p.activeNetwork != nil {
		return nil
	}
	activeNetwork, err := util.GetActiveNetworkForNamespace(namespace, p.nadLister)
	if err != nil {
		return err
	}
	if activeNetwork.IsDefault() {
		return fmt.Errorf("missing primary user defined network NAD")
	}
	p.activeNetwork = activeNetwork
	return nil
}
```
We call ensureActiveNetwork during CmdDel

go-controller/pkg/node/gateway_udn.go
trozet: I guess we can leave it for now and later refactor it to just take an interface and use interface methods rather than type casting it back to a gateway.

go-controller/pkg/node/gateway_udn.go
trozet: I'd feel a little better if we checked the positive cases where we know it will be nil, localnet, gateway disabled, etc rather than just warning if nil. But if it something you would rather fix later thats fine.

go-controller/pkg/node/gateway_udn.go
trozet: maybe? I guess that's something we can tune when we do perf/scale testing. It really depends on how many networks we expect users to create at a time and how long it takes to create those networks. But do we really expect it to take 300 seconds for ovn-controller to create the patch port? I think we used 300 seconds as our waiter for start up when we used centralized OVN and it could take a long time on a large scale cluster for everything to reconcile. This is during runtime when things should be mostly in a steady state, so could we lower the maximum wait timer? I would especially think this would be much faster in IC.

trozet:
Nice find @martinkennelly. We should be able to handle the case where NAD is gone and still delete the port.

tssurya:
Hmm but that's not a bug with this PR right?

The deletion bug is known issue, I think @qinqon was perhaps looking into this before he went on PTO.

Deleting NAD before pods case is not handled.

go-controller/pkg/node/gateway_udn.go
dceara: No we definitely don't expect ovn to take so long to create the patch ports.  In theory it shouldn't be longer than the time it takes to provision any pod port.  So max "a few seconds" at scale?

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
dceara: @trozet There's currently no plan for direct pod access AFAIU (https://github.com/openshift/enhancements/blob/master/enhancements/network/user-defined-network-segmentation.md#multiple-external-gateways-meg) but at the same time, if I'm reading that enhancement correctly, there is a plan to add it.  In which case the SNAT to masquerade IP won't always happen.

I'm OK with changing it to explicitly match the SNAT IP for now for all traffic but we'll have to change it to something else in the future.

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,90 @@ var _ = Describe("Network Segmentation", func() {
...
+				networkAttachmentConfigParams{
+					name:     userDefinedNetworkName,
+					topology: "layer2",
+					cidr:     fmt.Sprintf("%s,%s", userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
+					role:     "primary",
+				},
+				*podConfig("client-pod"),
+			),
+			Entry("by one pod with a single IPv4 address over a layer3 network",
```
maiqueb, line 585:
```suggestion
			Entry("by one pod with dual-stack addresses over a layer3 network",
```

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,90 @@ var _ = Describe("Network Segmentation", func() {
...
+					return connectToServer(clientPodConfig, externalIpv4, port)
+				}, 2*time.Minute, 6*time.Second).Should(Succeed())
+
+				By("asserting the *client* pod can contact the server's v6 IP located outside the cluster")
+				Eventually(func() error {
+					return connectToServer(clientPodConfig, externalIpv6, port)
+				}, 2*time.Minute, 6*time.Second).Should(Succeed())
+			},
+			// FIXME(tssurya): Unskip when L2 egress support is done
+			XEntry("by one pod with a single IPv4 address over a layer2 network",
```
maiqueb, line 576:
```suggestion
			XEntry("by one pod with dual-stack addresses over a layer2 network",
```

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1121,12 +1120,14 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
-			fmt.Sprintf("cookie=%s, priority=500, in_port=%s, ip, ip_dst=%s, ip_src=%s,"+
-				"actions=ct(commit,zone=%d,nat(dst=%s),table=4)",
-				defaultOpenFlowCookie, ofPortPatch, config.Gateway.MasqueradeIPs.V4HostMasqueradeIP.String(), physicalIP.IP,
-				config.Default.HostMasqConntrackZone, physicalIP.IP))
+		for _, netConfig := range bridge.netConfig {
+			// table 0, SVC Hairpin from OVN destined to local host, DNAT and go to table 4
+			dftFlows = append(dftFlows,
+				fmt.Sprintf("cookie=%s, priority=500, in_port=%s, ip, ip_dst=%s, ip_src=%s,"+
+					"actions=ct(commit,zone=%d,nat(dst=%s),table=4)",
+					defaultOpenFlowCookie, netConfig.ofPortPatch, config.Gateway.MasqueradeIPs.V4HostMasqueradeIP.String(), physicalIP.IP,
```
maiqueb, line 1128: we should add a check here where we refuse to proceed if `netConfig.ofPortPatch` is empty.

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,90 @@ var _ = Describe("Network Segmentation", func() {
...
+
+		DescribeTable(
+			"can be accessed to from the pods running in the Kubernetes cluster",
+			func(netConfigParams networkAttachmentConfigParams, clientPodConfig podConfiguration) {
+				netConfig := newNetworkAttachmentConfig(netConfigParams)
+
+				netConfig.namespace = f.Namespace.Name
+				clientPodConfig.namespace = f.Namespace.Name
+
+				By("creating the attachment configuration")
```
maiqueb, line 539: we should list here the generated network name, so it becomes easier to debug when things go wrong.

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,90 @@ var _ = Describe("Network Segmentation", func() {
...
+				By("instantiating the client pod")
+				clientPod, err := cs.CoreV1().Pods(clientPodConfig.namespace).Create(
+					context.Background(),
+					generatePodSpec(clientPodConfig),
+					metav1.CreateOptions{},
+				)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(clientPod).NotTo(BeNil())
+
+				By("asserting the client pod reaches the `Ready` state")
```
maiqueb, line 556: we should print here in which node was the pod scheduled so it becomes easier to debug when things go wrong (e.g. know which is the correct ovnkube-controller to look at).

tssurya:
unrelated UT Flake:
```
2024-08-14T21:54:55.8466926Z 
2024-08-14T21:54:55.8469205Z [91m[1m[Fail] [0m[90mOVN master EgressIP Operations [0m[0mOn node UPDATE [0m[90m[OVN network] should perform proper OVN transactions when pod is created after node egress label switch [0m[91m[1m[It] interconnect disabled [0m
2024-08-14T21:54:55.8472138Z [37m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:927[0m
2024-08-14T21:54:55.8473095Z 
```
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10395359454/job/28787193635?pr=4557

tssurya:
> Hmm but that's not a bug with this PR right?
> 
> The deletion bug is known issue, I think @qinqon was perhaps looking into this before he went on PTO.
> 
> Deleting NAD before pods case is not handled. @kyrtapz and I were talking about this earlier roday

This was discussed between Martin, Patryk, Miguel and myself today and resolved during the meeting

go-controller/pkg/node/gateway_shared_intf.go
tssurya: yeah have taken https://github.com/dceara/ovn-kubernetes/commit/2430febf2ba6141c9442c90fb48e94ed5e32fe6e from @dceara 
@dceara in EMEA hours tomorrow could you please clean this up for it to be prettier?

test/e2e/network_segmentation.go
tssurya: ack added a log to print that

go-controller/pkg/node/gateway_shared_intf.go
dceara: Done, squashed into the 4ec76ceb1 "Add bridgeUDNConfiguration for UDN" commit.

test/e2e/network_segmentation.go
dceara: @maiqueb Do we need a more unified approach (for more than just single test)?  Should we open a follow up issue to track that work?

tssurya: yeah I think all tests need this cleanup

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
dceara: I opened https://github.com/ovn-org/ovn-kubernetes/issues/4629 to track adding explicit drop flows for everything else.

test/e2e/network_segmentation.go
trozet: what did you decide here? We could check the annotations on the pod and verify its default route is via the UDN network?

trozet (changes requested):

go-controller/pkg/node/openflow_manager.go
```diff
@@ -223,18 +222,23 @@ func (c *openflowManager) updateBridgeFlowCache(subnets []*net.IPNet, extraIPs [
...
+				netConfig.patchPort, netConfig.ofPortPatch, curOfportPatch)
+			os.Exit(1)
```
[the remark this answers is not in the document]
dceara: While I agree that you're right, it was like this before this PR.  Let's open a follow up issue to get this fixed?

go-controller/pkg/node/gateway.go
```diff
@@ -404,7 +404,7 @@ func (g *gateway) Reconcile() error {
 	}
 	subnets, err := util.ParseNodeHostSubnetAnnotation(node, types.DefaultNetworkName)
 	if err != nil {
-		return fmt.Errorf("failed to get subnets for node: %s for OpenFlow cache update", node.Name)
+		return fmt.Errorf("failed to get subnets for node: %s for OpenFlow cache update; err: %v", node.Name, err)
```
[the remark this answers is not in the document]
dceara: Ack.

test/e2e/network_segmentation.go
dceara: @tssurya Do you mind taking care of this one please?

go-controller/pkg/node/gateway_udn.go
dceara: OK, I changed the waiter to accept a custom timeout and used 30 seconds here.

go-controller/pkg/node/openflow_manager.go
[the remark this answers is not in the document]
trozet: thats fine: https://github.com/ovn-org/ovn-kubernetes/issues/4630

test/e2e/network_segmentation.go
tssurya: ack I can I don't see other tests doing this but this should be easy.. lemme see if I can push a fix for this

tssurya: This is done!

dceara:
We hit https://github.com/ovn-org/ovn-kubernetes/issues/4432 here https://github.com/ovn-org/ovn-kubernetes/actions/runs/10415227854/job/28845892439?pr=4557
```
 Summarizing 1 Failure:
  [FAIL] External Gateway With Admin Policy Based External Route CRs e2e multiple external gateway validation Should validate ICMP connectivity to multiple external gateways for an ECMP scenario [It] IPV4
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/external_gateways.go:1636
```

Retesting.

test/e2e/network_segmentation.go
maiqueb: Yeah, I guess all tests will benefit from it.

test/e2e/network_segmentation.go
```diff
@@ -508,6 +509,97 @@ var _ = Describe("Network Segmentation", func() {
...
+					}
+					return updatedPod.Status.Phase
+				}, 2*time.Minute, 6*time.Second).Should(Equal(v1.PodRunning))
+				framework.Logf("Client pod was created on node %s", updatedPod.Spec.NodeName)
+
+				By("asserting UDN pod is connected to UDN network")
+				podAnno, err := unmarshalPodAnnotation(updatedPod.Annotations, f.Namespace.Name+"/"+userDefinedNetworkName)
+				Expect(err).NotTo(HaveOccurred())
+				framework.Logf("Client pod's annotation for network %s is %v", userDefinedNetworkName, podAnno)
+				Expect(len(podAnno.Routes)).To(Equal(6)) // 3 v4 routes + 3 v6 routes for UDN
```
maiqueb, line 577: I don't remember writing this expectation.

Isn't this L3 specific ? For sure I would expect to have less routes on L2. 

I would also expect to see even less routes for IPv4 only clusters.

And please use the `HaveLen` matcher instead of expecting the length of an array to be X - the error 
```
  [FAILED] Expected
      <int>: 3
  to equal
      <int>: 6
```
is not informative.

maiqueb: Well, let's postpone this; I can address it in my layer2 topology PR. 

Are you OK with that ?

dceara: I think this route checking was the result of a request from @trozet: https://github.com/ovn-org/ovn-kubernetes/pull/4557#discussion_r1711823486

Sounds good to mee to address it in the L2 PR.

tssurya: that is correct, it came from the fact that we wanted to check routes were there, surely we can make it better in future by grepping for exact routes..

tssurya: and I knew we were only running dualstack upstream which is why I did that, but I should add a check for if IPV6 {} and then do the needed.
but I want to get the ds merge in first so that we can rebase this on master

maiqueb: OK. I agree we should assert the routes are the expected ones on a follow-up.

trozet (approved):

PR #4558, opened 2024-07-22
make openflow manager to be sharable by different network

cathy-zhou:

Fixes #

```release-note

```


tssurya:
@cathy-zhou: see https://github.com/ovn-org/ovn-kubernetes/pull/4557 which will let you do exactly that :) also for other networks..

cathy-zhou:
> @cathy-zhou: see #4557 which will let you do exactly that :) also for other networks..

thanks. let me take a look

maiqueb (commented):
Could you fill the PR template to allow the community to better understand what the PR is about ? 

Also, I think @tssurya could be doing something similar to this in one of her PRs ...

go-controller/pkg/node/openflowmanager/openflow_manager.go
```diff
@@ -0,0 +1,147 @@
...
+		c, found := oc.ofmanagers.LoadOrStore(bridge, &ofmanager{
+			bridge:    bridge,
+			flowCache: make(map[string][]string),
+			flowMutex: sync.Mutex{},
+			flowChan:  make(chan struct{}, 1),
+		})
+		c.flowMutex.Lock()
+		defer c.flowMutex.Unlock()
+		c.flowCache[key] = flows
+		// oc.stopChan is nil in unit testing, so it does not start syncFlow go-routine for unit tests.
```
maiqueb, line 49: why is that ? Shouldn't we want to unit test something that resembles the live system ? 

Meaning, why can't we start the `syncFlow` go-routine for unit tests ?

go-controller/pkg/node/openflowmanager/openflow_manager.go
```diff
@@ -0,0 +1,147 @@
+package openflowmanager
+
+import (
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/syncmap"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
```
maiqueb, line 9: nit: these imports go in the bottom, since they're local to the project.

maiqueb (commented):
I'm just providing a very superficial review - I have a question and a nit.

PR #4565, opened 2024-07-26
Change route table calculation

martinkennelly:
cc @qinqon @maiqueb


go-controller/pkg/util/net.go
```diff
@@ -332,5 +329,5 @@ func IPNetsIPToStringSlice(ips []*net.IPNet) []string {
 // CalculateRouteTableID will calculate route table ID based on the network
 // interface index
 func CalculateRouteTableID(ifIndex int) int {
-	return ifIndex + routingTableIDStart
+	return math.MaxUint32 - ifIndex
```
tssurya, line 332: hmm why is that? what conflicts do you mean?

martinkennelly: Users are more likely to create tables at lower integer ranges i think without reading documentation and conflict with tables we create.

martinkennelly: I havent seen any cases where this has happened yet with EIP but guessing.

qinqon (changes requested):
This will affect non primary UDNs system using egressips since it will migrate the rules tableIDs for something that we don't really not is a problem I would document this propertly, even adding a bullet at troubelshooting guide.

go-controller/pkg/util/net.go
tssurya: hmm I think documenting it should be enough? but given you have done the work here already I don't mind going with this....ultimately I get this feeling someone will say they want to make it configurable LOL!

martinkennelly: I dont mind discarding it if its not worth it. No problem. I think its sensible not to depend on docs alone and this allows us to avoid human readable route tables integers.

martinkennelly:
Planning to move ahead with this once maintainers have time. I moving the base value away from a lower range is the right thing to do. Up for ideas for the range we settle on and documentation.

go-controller/pkg/node/iprulemanager/ip_rule_manager.go
```diff
@@ -85,6 +85,10 @@ func (rm *Controller) Delete(rule netlink.Rule) error {
 	}
 	if reconcileNeeded {
 		return rm.reconcile()
+	} else {
+		// though we don't need to reconcile because we don't manage this rule, we should still ensure its deleted
+		// from the system. No need for full reconcile.
+		_ = netlink.RuleDel(&rule)
```
trozet, line 91: The commit message talks about deleting ip rules that the controller doesn't manage. Wouldn't this be a problem  if the rule is some custom rule created by the admin? Or does the controller only delete rules that have specific attributes so we know RuleManager was the original controller of this rule?

go-controller/pkg/util/net.go
trozet: I agree with starting higher, but I have 2 concerns:
1. It looks like you are starting from the max and decrementing. The maximum is 4294967295. This is a really big number for a user or a developer to just memorize as they are trying to debug and look at routes in their UDN route table. I wonder if there is some kind of compromise where we could start at like 10000 or something?
2. By changing this, does this break upgrade for egress IP that is currently using some other table? Or how does that work? Will there be outage or is it only for assigning new table numbers?

tssurya: > 2. By changing this, does this break upgrade for egress IP that is currently using some other table? Or how does that work? Will there be outage or is it only for assigning new table numbers?

I think no because when upgrading during node reboots we will start fresh. so the new tables would get picked up on their own based on the new ifaceindex

tssurya: > I dont mind discarding it if its not worth it. No problem. I think its sensible not to depend on docs alone and this allows us to avoid human readable route tables integers.

no I don't mean its not worth it.. the change makes sense... just that we need the PRs that are higher priority in before this :D + we still need to document the range we use just to be sure.

go-controller/pkg/node/iprulemanager/ip_rule_manager.go
martinkennelly: The rule manager doesnt have to own the rule to delete it here. Users of rule manager may use it exclusively to manager rules and it may find a stale rule and want to delete it.

martinkennelly:
> It looks like you are starting from the max and decrementing. The maximum is 4294967295. This is a really big number for a user or a developer to just memorize as they are trying to debug and look at routes in their UDN route table. I wonder if there is some kind of compromise where we could start at like 10000 or something?

Sure, changed to 100k

>By changing this, does this break upgrade for egress IP that is currently using some other table? Or how does that work? Will there be outage or is it only for assigning new table numbers?

Not for OCP because nodes become `NotReady` and another egress node maybe selected. If you were simply just upgrading, you may see brief disruption.

go-controller/pkg/util/net.go
trozet: Maybe I didn't phrase my question right or I'm misunderstanding something. We upgrade ovnk-node first (before reboot). So my question is, lets say today we are using VRF 5 (just randomly picking a number) for egress IP. Now OVNK upgrades, will it ocntinue using VRF 5 and there is no outage, or will it now pick a new VRF table and reprogram it due to this commit?

martinkennelly: I think its possible theres disruption when switching to the new routing table.

martinkennelly: but I am not sure and probably should write a test.

martinkennelly:
/hold

I now need to consider migrating existing UDN VRFs to the new routing tables.

PR #4567, opened 2024-07-29
Add support for Pod->KAPI/DNS traffic in user-defined primary networks

kyrtapz:
Depends on https://github.com/ovn-org/ovn-kubernetes/pull/4694.

The PR adds OVS flows that redirect traffic from the masquerade subnet to the specified services to the default gateway router.
For shared gateway mode a static route is added to redirect the service traffic to the management port. Modified the
staticRouteCleanup to ensure that the new routes are not getting removed.

This change does not work on non-IC clusters for l3 networks in shared gateway, tracked here: https://github.com/ovn-org/ovn-kubernetes/issues/4741.

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -12,6 +12,13 @@ import (
 	anpinformerfactory "sigs.k8s.io/network-policy-api/pkg/client/informers/externalversions"
 	anpinformer "sigs.k8s.io/network-policy-api/pkg/client/informers/externalversions/apis/v1alpha1"
 
+	certificatesinformers "k8s.io/client-go/informers/certificates/v1"
+
+	ocpnetworkapiv1alpha1 "github.com/openshift/api/network/v1alpha1"
+	ocpnetworkscheme "github.com/openshift/client-go/network/clientset/versioned/scheme"
+	ocpnetworkinformerfactory "github.com/openshift/client-go/network/informers/externalversions"
+	ocpnetworkinformerv1alpha1 "github.com/openshift/client-go/network/informers/externalversions/network/v1alpha1"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	egressfirewallapi "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressfirewall/v1"
 	egressfirewallscheme "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressfirewall/v1/apis/clientset/versioned/scheme"
@@ -20,12 +27,6 @@ import (
 	egressfirewalllister "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressfirewall/v1/apis/listers/egressfirewall/v1"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
-	certificatesinformers "k8s.io/client-go/informers/certificates/v1"
-
-	ocpnetworkapiv1alpha1 "github.com/openshift/api/network/v1alpha1"
-	ocpnetworkscheme "github.com/openshift/client-go/network/clientset/versioned/scheme"
-	ocpnetworkinformerfactory "github.com/openshift/client-go/network/informers/externalversions"
-	ocpnetworkinformerv1alpha1 "github.com/openshift/client-go/network/informers/externalversions/network/v1alpha1"
 
 	egressipapi "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressip/v1"
 	egressipscheme "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/egressip/v1/apis/clientset/versioned/scheme"
--- a/go-controller/pkg/node/gateway_init_linux_test.go
+++ b/go-controller/pkg/node/gateway_init_linux_test.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/client-go/kubernetes/fake"
 
 	nadfake "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/fake"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	adminpolicybasedrouteclient "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/adminpolicybasedroute/v1/apis/clientset/versioned/fake"
 	udnfakeclient "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/clientset/versioned/fake"
@@ -1275,6 +1276,7 @@ OFPT_GET_CONFIG_REPLY (xid=0x4): frags=normal miss_send_len=0`
 			)
 			expectedTables["nat"]["OVN-KUBE-UDN-MASQUERADE"] = append(expectedTables["nat"]["OVN-KUBE-UDN-MASQUERADE"],
 				"-s 169.254.169.2/29 -j RETURN",     // this guarantees we don't SNAT default network masqueradeIPs
+				"-d 172.16.1.0/24 -j RETURN",        // this guarantees we don't SNAT service traffic
 				"-s 169.254.169.0/29 -j MASQUERADE", // this guarantees we SNAT all UDN MasqueradeIPs traffic leaving the node
 			)
 		}
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -10,6 +10,9 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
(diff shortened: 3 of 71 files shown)


go-controller/pkg/node/gateway_init_linux_test.go
```diff
@@ -1242,6 +1243,7 @@ OFPT_GET_CONFIG_REPLY (xid=0x4): frags=normal miss_send_len=0`
 			)
 			expectedTables["nat"]["OVN-KUBE-UDN-MASQUERADE"] = append(expectedTables["nat"]["OVN-KUBE-UDN-MASQUERADE"],
 				"-s 169.254.169.2/29 -j RETURN",     // this guarantees we don't SNAT default network masqueradeIPs
+				"-d 172.16.1.0/24 -j RETURN",        // this guarantees we don't SNAT service traffic
```
dceara, line 1246: This should be part of the previous commit where we actually add the iptables rule.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -213,7 +216,16 @@ func (c *openflowManager) updateBridgeFlowCache(subnets []*net.IPNet, extraIPs [
 	// CAUTION: when adding new flows where the in_port is ofPortPatch and the out_port is ofPortPhys, ensure
 	// that dl_src is included in match criteria!
 
-	dftFlows, err := flowsForDefaultBridge(c.defaultBridge, extraIPs)
+	var udnAllowedServicesIPs []net.IP
+	var err error
+	if util.IsNetworkSegmentationSupportEnabled() {
+		udnAllowedServicesIPs, err = c.watchFactory.GetUDNAllowedServicesIPs()
```
trozet, line 222: is updateBridgeFlowCache the right place to be doing this? what if the service changes, we have nodeport watcher that handles services and then updates the openflow cache.

// updateBridgeFlowCache generates the "static" per-bridge flows

I only see updateBridgeFlowCache triggered on:

1. node IP change
2. OFM manager creation
3. node MAC changes
4. UDN Add/Del network

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1384,7 +1385,39 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
+
+		// The flows added below have a higher priority than the default network service flows:
+		//   priority=500,ip,in_port=LOCAL,nw_dst=10.96.0.0/16 actions=ct(commit,table=2,zone=64001,nat(src=169.254.0.2))
+		//   priority=500,ip,in_port="patch-breth0_ov",nw_src=10.96.0.0/16,nw_dst=169.254.0.2 actions=ct(table=3,zone=64001,nat)
+		// This ordering ensures that there is no SNAT for UDN originated traffic.
+		// However, it also means that the flows match traffic with the source IP of the default masquerade IP (169.254.0.2) and a
+		// destination of a UDN allowed service. To handle this, the flow commits the connection to conntrack and applies NAT
+		// action in the reply direction to potentially unSNAT from 169.254.0.2.
+
+		// table 0, user-defined network host -> OVN towards default cluster network services
```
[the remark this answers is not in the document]
dceara: In general it's usually a bad idea to only commit some of the traffic.  If the traffic that is not committed to conntrack happens to have a 4-tuple that also matches what we get post-nat for subsequent traffic that we nat and commit, netfilter has no way of figuring that out and won't do any tuple conflict resolution.

go-controller/pkg/factory/factory.go
```diff
@@ -1439,6 +1442,43 @@ func (wf *WatchFactory) GetNADs(namespace string) ([]*nadapi.NetworkAttachmentDe
...
+		if len(svcObj.Spec.ClusterIP) == 0 {
+			klog.Infof("Service %q has no ClusterIP address, ignoring", svc)
+			continue
+		}
+		svcIP := net.ParseIP(svcObj.Spec.ClusterIP)
+		if svcIP == nil {
+			return nil, fmt.Errorf("failed to parse service IP %s for service: %q", svcObj.Spec.ClusterIP, svc)
+		}
+
+		if !util.IsClusterIP(svcIP.String()) {
```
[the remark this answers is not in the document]
kyrtapz: It's not, this is a left over from a  version where I allowed IPs to be provided, will remove it.
It is possible for the user to set the service IP but it has to match the service subnet https://kubernetes.io/docs/concepts/services-networking/service/#choosing-your-own-ip-address.

go-controller/pkg/factory/factory.go
```diff
@@ -1439,6 +1442,43 @@ func (wf *WatchFactory) GetNADs(namespace string) ([]*nadapi.NetworkAttachmentDe
...
 }
 
+// GetUDNAllowedServicesIPs returns the current list of IP addresses for services specified in config.Default.UDNAllowedDefaultServices
+func (wf *WatchFactory) GetUDNAllowedServicesIPs() ([]net.IP, error) {
+	serviceIPs := make([]net.IP, 0, len(config.Default.UDNAllowedDefaultServices.Value()))
+
+	for _, svc := range config.Default.UDNAllowedDefaultServices.Value() {
+		namespace, name, err := cache.SplitMetaNamespaceKey(svc)
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
kyrtapz: Since this is a startup config option I would prefer to do a hard error to avoid hiding the issue, wdyt?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1384,7 +1385,39 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
+		dftFlows = append(dftFlows,
+			fmt.Sprintf("cookie=%s, priority=600, in_port=%s, %s, %s_src=%s, %s_dst=%s,"+
+				"actions=ct(commit,zone=%d,table=2)",
+				defaultOpenFlowCookie, ofPortHost, ipPrefix, ipPrefix, masqueradeSubnet, ipPrefix, svcIP, config.Default.HostMasqConntrackZone))
+
+		// table 0, Reply hairpin traffic to host, coming from OVN
```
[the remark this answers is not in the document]
kyrtapz: I tried to keep the new UDN flow symmetric but I can just modify the existing one, no problem.

go-controller/pkg/factory/factory.go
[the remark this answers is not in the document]
trozet: I think it makes more sense to have this type of validation in config.go

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
trozet: it will help us debug if it is conntrack, i was just trying to do it as a perf optimization. Since we only send CT traffic that matches on 169.254.0.2 I thought it would be fine, but we can commit everything for now.

go-controller/pkg/node/openflow_manager.go
trozet: I guess in service controller you would need to compare if the sync'ed service matches the one of hte UDNAllowedServiceIPs and then, get the network, and from that you can figure out the mgmt ip. That doesn't seem that complicated since UDN is already integrated into he services controller. Or am I missing something?

kyrtapz: I'll look into that and push again.

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
kyrtapz: Thanks! I guess lets keep it for now and I can change the approach if needed in the future.

go-controller/pkg/node/openflow_manager.go
kyrtapz: Done.

go-controller/pkg/util/util.go
```diff
@@ -624,3 +626,8 @@ func GetServiceEndpointSlices(namespace, svcName, network string, endpointSliceL
 	}
 	return GetEndpointSlicesBySelector(namespace, selector, endpointSliceLister)
 }
+
+// IsUDNEnabledService checks whether the provided namespaced name key is a UDN enabled service specified in config.Default.UDNAllowedDefaultServices
+func IsUDNEnabledService(key string) bool {
+	return strings.Contains(config.Default.UDNAllowedDefaultServices.String(), key)
```
trozet, line 632: why is this contains and not equal? configured allowed udn service test/abc could accidentally match service test/ab?

go-controller/pkg/ovn/gateway.go
```diff
@@ -1102,7 +1105,7 @@ func (gw *GatewayManager) delPbrAndNatRules(nodeName string) {
 	gw.removeLRPolicies(nodeName)
 }
 
-func (gw *GatewayManager) staticRouteCleanup(nextHops []net.IP) {
+func (gw *GatewayManager) staticRouteCleanup(nextHops []net.IP, ipPrefix string) {
```
trozet, line 1108: nit: should use net.IPNet

go-controller/pkg/util/util.go
kyrtapz: :facepalm: 
Fixed.

go-controller/pkg/ovn/gateway.go
kyrtapz: Done.

trozet (approved):

PR #4568, opened 2024-07-29
udn, cni: Unconfigure interface

qinqon:
When a pod with a primary UDN is deleted it has to call the unconfigure interface to delete ovs resources.

#### How to verify it
It will have unit test

```release-note
NONE
```

### DIFF
--- a/go-controller/pkg/cni/udn/primary_network.go
+++ b/go-controller/pkg/cni/udn/primary_network.go
@@ -74,32 +74,50 @@ func (p *UserDefinedPrimaryNetwork) WaitForPrimaryAnnotationFn(namespace string,
 		if annotation == nil {
 			return nil, false
 		}
-
-		// CmdAdd from non default network is not realted to primary UDNs
-		if nadName != types.DefaultNetworkName {
-			return annotation, isReady
-		}
-
-		// If these are pods created before primary UDN functionality the
-		// default network without role is the primary network
-		if annotation.Role == "" {
-			annotation.Role = types.NetworkRolePrimary
-		}
-
-		if annotation.Role == types.NetworkRolePrimary {
-			return annotation, isReady
-		}
-
-		if err := p.ensureAnnotation(annotations); err != nil {
-			klog.Errorf("Failed looking for primary network annotation: %v", err)
-			return nil, false
-		}
-		if err := p.ensureActiveNetwork(namespace); err != nil {
-			klog.Errorf("Failed looking for primary network name: %v", err)
+		if err := p.ensure(namespace, annotations, nadName, annotation); err != nil {
+			klog.Errorf("Failed ensuring user defined primary network: %v", err)
 			return nil, false
 		}
 		return annotation, isReady
 	}
+}
+
+func (p *UserDefinedPrimaryNetwork) Ensure(namespace string, annotations map[string]string, nadName string) error {
+	return p.ensure(namespace, annotations, nadName, nil /* parse annotation */)
+}
+
+func (p *UserDefinedPrimaryNetwork) ensure(namespace string, annotations map[string]string, nadName string, annotation *util.PodAnnotation) error {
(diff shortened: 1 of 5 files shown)


go-controller/pkg/cni/udn/primary_network.go
```diff
@@ -104,6 +104,29 @@ func (p *UserDefinedPrimaryNetwork) WaitForPrimaryAnnotationFn(namespace string,
 	}
 }
 
+func (p *UserDefinedPrimaryNetwork) Ensure(namespace string, annotations map[string]string, nadName string) error {
+	// non default network is not related to primary UDNs
+	if nadName != types.DefaultNetworkName {
+		return nil
+	}
```
maiqueb, line 111: nit: shouldn't this belong in the caller ? Or at least in the constructor ? It feels strange to me to have a `UserDefinedPrimaryNetwork` struct instance that is actually not a primary.

go-controller/pkg/cni/udn/primary_network.go
```diff
@@ -104,6 +104,29 @@ func (p *UserDefinedPrimaryNetwork) WaitForPrimaryAnnotationFn(namespace string,
 	}
 }
 
+func (p *UserDefinedPrimaryNetwork) Ensure(namespace string, annotations map[string]string, nadName string) error {
```
maiqueb, line 107: could we re-use this method from the `WaitForPrimaryAnnotationFn` function ? 

Not sure it fits well ...

maiqueb (commented):
Some nits.

go-controller/pkg/cni/udn/primary_network.go
qinqon: We should, there is a common important bits here.

go-controller/pkg/cni/udn/primary_network.go
maiqueb: yeah, but how do I know that if I only look here ? I don't. IIUC, you're actually saying this check is redundant if called only from within a `Found()` block. Right ? 

... then why have it ?

qinqon: So, we need to do this check to break the `GetPodWithAnnotations` as soon as the the default pod network ovn pod annotation is found, and since we are re-using the logic for the unconfigure we do it there too, is not hurting though.

maiqueb (approved):
Thank you.

go-controller/pkg/cni/udn/primary_network.go
```diff
@@ -74,32 +74,50 @@ func (p *UserDefinedPrimaryNetwork) WaitForPrimaryAnnotationFn(namespace string,
...
+	return p.ensure(namespace, annotations, nadName, nil /* parse annotation */)
+}
 
-		if annotation.Role == types.NetworkRolePrimary {
-			return annotation, isReady
-		}
+func (p *UserDefinedPrimaryNetwork) ensure(namespace string, annotations map[string]string, nadName string, annotation *util.PodAnnotation) error {
+	// non default network is not related to primary UDNs
+	if nadName != types.DefaultNetworkName {
```
trozet, line 91: I get confused here with nadName and which nadName we expect we are looking at here. From my understanding (with respect to a UDN primary case), we expect to get an ensure here for the default network. Then later in the code we check if the default network is not primary role, and then continue to verify that a UDN exists with a primary role and correct annotation. It would be beneficial I think to document this function describing the behavior. This can be done in a follow up PR.

trozet (commented):
first commit lgtm. Could use some more documentation if you care to do it while fixing unit tests

trozet (approved):

go-controller/pkg/cni/udn/primary_network.go
qinqon: Comment it out at follow up

PR #4572, opened 2024-07-30
SDN-5139: Direct ingress for default network in LGW mode with route advertisements 

jcaamano:
If the default network is advertised, inhibit:

- pod IP to node IP SNAT on egress in LGW mode
- source IP to managament port IP SNAT on ingress (through secondary nic)

Currently based on top of https://github.com/ovn-org/ovn-kubernetes/pull/4472 and https://github.com/ovn-org/ovn-kubernetes/pull/4533

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -57,7 +57,8 @@ type gateway struct {
 	stopChan     <-chan struct{}
 	wg           *sync.WaitGroup
 
-	isPodNetworkAdvertised bool
+	isPodNetworkAdvertisedLock sync.Mutex
+	isPodNetworkAdvertised     bool
 }
 
 func (g *gateway) AddService(svc *kapi.Service) error {
@@ -468,6 +469,8 @@ func (g *gateway) SetDefaultGatewayBridgeMAC(macAddr net.HardwareAddr) {
 }
 
 func (g *gateway) SetPodNetworkAdvertised(isPodNetworkAdvertised bool) {
+	g.isPodNetworkAdvertisedLock.Lock()
+	defer g.isPodNetworkAdvertisedLock.Unlock()
 	g.isPodNetworkAdvertised = isPodNetworkAdvertised
 }
 
@@ -520,6 +523,8 @@ func (g *gateway) addAllServices() []error {
 }
 
 func (g *gateway) updateSNATRules() error {
+	g.isPodNetworkAdvertisedLock.Lock()
+	defer g.isPodNetworkAdvertisedLock.Unlock()
 	var ipnets []*net.IPNet
 	if g.nodeIPManager.mgmtPortConfig.ipv4 != nil {
 		ipnets = append(ipnets, g.nodeIPManager.mgmtPortConfig.ipv4.ifAddr)
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -485,6 +485,10 @@ func (g *gateway) Reconcile() error {
 	if err := g.openflowManager.updateBridgeFlowCache(subnets, g.nodeIPManager.ListAddresses(), g.isPodNetworkAdvertised); err != nil {
 		return err
 	}
+	err = g.updateSNATRules()
+	if err != nil {
+		return err
+	}
 	// Services create OpenFlow flows as well, need to update them all
 	if g.servicesRetryFramework != nil {
 		if errs := g.addAllServices(); errs != nil {
@@ -513,6 +517,23 @@ func (g *gateway) addAllServices() []error {
 	}
 	g.servicesRetryFramework.RequestRetryObjs()
 	return errs
+}
+
+func (g *gateway) updateSNATRules() error {
+	var ipnets []*net.IPNet
+	if g.nodeIPManager.mgmtPortConfig.ipv4 != nil {
+		ipnets = append(ipnets, g.nodeIPManager.mgmtPortConfig.ipv4.ifAddr)
+	}
+	if g.nodeIPManager.mgmtPortConfig.ipv6 != nil {
+		ipnets = append(ipnets, g.nodeIPManager.mgmtPortConfig.ipv6.ifAddr)
+	}
+	subnets := util.IPsToNetworkIPs(ipnets...)
+
+	if g.isPodNetworkAdvertised || config.Gateway.Mode != config.GatewayModeLocal {
+		return delLocalGatewayPodSubnetNATRules(subnets...)
+	}
+
+	return addLocalGatewayPodSubnetNATRules(subnets...)
 }
 
 type bridgeConfiguration struct {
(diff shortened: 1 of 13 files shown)


go-controller/pkg/node/gateway.go
```diff
@@ -238,9 +241,7 @@ func (g *gateway) AddEgressIP(eip *egressipv1.EgressIP) error {
 		return err
 	}
 	if isSyncRequired {
-		if err = g.Reconcile(); err != nil {
```
trozet, line 241: not sure I understand this change. If this fails, it would retry as part of the retry framework wouldn't it? You are moving it to a separate goroutine and operating outside of retry framework now?

jcaamano: I didn't give this much thought as I took it in from the latest rebase of the EgressIP changes.

I had the chance to operate asynchoronously, pretty much as what `g.openflowManager.requestFlowSync()` is doing below, so I took it. `AddEgressIP` is also making use of other asynchronous downstream components, like address manager.

Please note that this change here did not motivate making `Reconcile` asynchronous, we call `Reconcile` from other places where we had no retry mechanism in place.

go-controller/pkg/node/management-port_linux.go
```diff
@@ -311,7 +320,29 @@ func setupManagementPortNFTables(cfg *managementPortConfig) error {
...
+					"fib saddr type != local",
+					counterIfDebug,
+					"return",
+				),
+			})
+			tx.Add(&knftables.Rule{
+				Chain: nftablesMgmtPortChain,
+				Rule: knftables.Concat(
+					"meta nfproto ipv4",
+					"ip saddr", cfg.ipv4.ifAddr.IP,
```
trozet, line 340: why is this necessary? isn't this the mpx IP?

trozet (commented):
Can you explain further this scenario @jcaamano ?

However, we add an SNAT rule for local IPT traffic flows. Otherwise the
node masquerade address is used as it is set as the source address in
the services route.

go-controller/pkg/node/management-port_linux.go
```diff
@@ -311,7 +320,29 @@ func setupManagementPortNFTables(cfg *managementPortConfig) error {
...
 
+	isPodNetworkAdvertised := cfg.isPodNetworkAdvertised.Load()
+
 	if cfg.ipv4 != nil {
+		if isPodNetworkAdvertised {
+			tx.Add(&knftables.Rule{
+				Chain: nftablesMgmtPortChain,
+				Rule: knftables.Concat(
+					"meta nfproto ipv4",
+					"fib saddr type != local",
```
trozet, line 331: for UDNs do we not have different management port chains per udn or what? @tssurya ? Maybe not yet since host -> UDN pod doesnt work?

go-controller/pkg/node/management-port.go
```diff
@@ -162,3 +156,49 @@ func managementPortReady() (bool, error) {
...
+			case <-p.reconcile:
+				// TODO retry maybe
+				err := p.doReconcile()
+				if err != nil {
+					klog.Errorf("Failed to reconcile management port %s: %v", p.config.ifName, err)
+				}
+			}
+		}
+	}()
+	go wait.Until(
```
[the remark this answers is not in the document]
jcaamano: No specific reason. Will change that.

go-controller/pkg/node/management-port_linux.go
jcaamano: yes, it's done in `getUDNMasqueradeRules`, still relies on iptables, will be handled by @arghosh93 in his work to support UDNs.

jcaamano:
> Can you explain further this scenario @jcaamano ?
> 
> However, we add an SNAT rule for local IPT traffic flows. Otherwise the node masquerade address is used as it is set as the source address in the services route.

So lets say I have a ITP Local node port service
```
NAME         TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)        AGE
nginx        NodePort    10.96.127.6   <none>        80:32734/TCP   8m33
```
with backend pod
```
NAMESPACE            NAME        READY   STATUS    RESTARTS   AGE    IP           NODE 
default              nginx       1/1     Running   0          6m8s   10.244.1.4   ovn-worker
```
If I tried to reach out from that node to the service cluster IP in LGW mode, no route advertisements in place, first it will be marked
```
Chain OVN-KUBE-ITP (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    0     0 MARK       tcp  --  any    any     anywhere             10.96.127.6          tcp dpt:http MARK set 0x1745ec
```
then it will hit the ip rule
```
30:	from all fwmark 0x1745ec lookup 7
```
and routed as per that table
```
10.96.0.0/16 via 10.244.1.1 dev ovn-k8s-mp0 
```
Now you would guess that the source IP address used would be that of `ovn-k8s-mp0`, but that's not really the case, we depend on the SNAT rule:
```
tcp      6 119 TIME_WAIT src=169.254.0.2 dst=10.96.127.6 sport=56306 dport=80 src=10.96.127.6 dst=10.244.1.2 sport=80 dport=56306 [ASSURED] mark=0 secctx=system_u:object_r:unlabeled_t:s0 use=1
tcp      6 119 TIME_WAIT src=10.244.1.2 dst=10.96.127.6 sport=56306 dport=80 src=10.244.1.4 dst=10.244.1.2 sport=80 dport=56306 [ASSURED] mark=2 secctx=system_u:object_r:unlabeled_t:s0 zone=1 use=1
```
Even if you change that routing to 
```
10.96.0.0/16 via 10.244.1.1 dev ovn-k8s-mp0 src 10.244.1.2
```
it doesn't help since network stack is not aware of packet marking when picking the source ip address so at that time is not really aware that ip rule and routing table applies.

If the application happens to bind to the management IP, then we get an unnecessary SNAT
```
tcp      6 117 TIME_WAIT src=10.244.1.2 dst=10.96.127.6 sport=43496 dport=80 src=10.96.127.6 dst=10.244.1.2 sport=80 dport=43496 [ASSURED] mark=0 secctx=system_u:object_r:unlabeled_t:s0 use=1
tcp      6 117 TIME_WAIT src=10.244.1.2 dst=10.96.127.6 sport=43496 dport=80 src=10.244.1.4 dst=10.244.1.2 sport=80 dport=43496 [ASSURED] mark=2 secctx=system_u:object_r:unlabeled_t:s0 zone=1 use=1
``` 

So then, if the pod network is advertised and we just get rid of that SNAT altogether, the source IP used would be `169.254.0.2` and the reply would be an asymmetric path and dropped. We need to keep the SNAT for local addresses.

go-controller/pkg/node/management-port_linux.go
jcaamano: So if the source IP is already mpx IP, then there is no need to SNAT, is it? You can see more detail here:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4572#issuecomment-2551052901

I guess this should not be specific to the advertised if block. Probably misplaced it when I rebased the nft changes.

jcaamano: > I guess this should not be specific to the advertised if block. Probably misplaced it when I rebased the nft changes.

Changed accordingly, let me know if you would rather not.

jcaamano:
Added commit to make gateway reconcile thread safe

trozet (approved):
/lgtm

PR #4577, opened 2024-08-01
OCPBUGS-37433: ignore localnet patch ports while bootstrapping the cluster default network

maiqueb:

This commit changes the openflow manager to ignore localnet patch ports when bootstrapping the default cluster network - these have a different format (they feature the physical network name for which the port was created instead of a physical bridge), which leads to the following error:

```
failed to start node network controller: failed to start default node network
controller: failed to bootstrap OVS flows: failed to get flows on bridge
"vlan2002_ovn_localnet":, stderr: "ovs-ofctl: vlan2002_ovn_localnet is not a
bridge or a socket\n", error: exit status 1
```

Fixes https://issues.redhat.com/browse/OCPBUGS-37433

```release-note

```

### DIFF
--- a/go-controller/pkg/node/openflow_manager.go
+++ b/go-controller/pkg/node/openflow_manager.go
@@ -253,7 +253,7 @@ func checkPorts(patchIntf, ofPortPatch, physIntf, ofPortPhys string) error {
 
 // bootstrapOVSFlows handles ensuring basic, required flows are in place. This is done before OpenFlow manager has
 // been created/started, and only done when there is just a NORMAL flow programmed and OVN/OVS is already setup
-func bootstrapOVSFlows() error {
+func bootstrapOVSFlows(nodeName string) error {
 	// see if patch port exists already
 	var portsOutput string
 	var stderr string
@@ -266,8 +266,11 @@ func bootstrapOVSFlows() error {
 
 	var bridge string
 	var patchPort string
-	// patch-br-int-to-<bridge name>_<node>
-	r := regexp.MustCompile("^patch-(.*)_.*?-to-br-int$")
+	// This needs to work with:
+	// - default network: patch-<bridge name>_<node>-to-br-int
+	// but not with:
+	// - secondary network: patch-<bridge name>_<node>-to-br-int
+	r := regexp.MustCompile(fmt.Sprintf("^patch-([^_]*)_%s-to-br-int$", nodeName))
 	for _, line := range strings.Split(portsOutput, "\n") {
 		matches := r.FindStringSubmatch(line)
 		if len(matches) == 2 {
--- a/go-controller/pkg/node/openflow_manager.go
+++ b/go-controller/pkg/node/openflow_manager.go
@@ -264,21 +264,7 @@ func bootstrapOVSFlows(nodeName string) error {
 		return fmt.Errorf("failed to list ports on existing bridge br-int: %s, %w", stderr, err)
 	}
 
-	var bridge string
-	var patchPort string
-	// This needs to work with:
-	// - default network: patch-<bridge name>_<node>-to-br-int
-	// but not with:
-	// - secondary network: patch-<bridge name>_<node>-to-br-int
-	r := regexp.MustCompile(fmt.Sprintf("^patch-([^_]*)_%s-to-br-int$", nodeName))
-	for _, line := range strings.Split(portsOutput, "\n") {
-		matches := r.FindStringSubmatch(line)
-		if len(matches) == 2 {
-			patchPort = matches[0]
-			bridge = matches[1]
-			break
-		}
-	}
+	bridge, patchPort := localnetPortInfo(nodeName, portsOutput)
 
 	if len(bridge) == 0 {
 		// bridge exists but no patch port was found
@@ -340,4 +326,22 @@ func bootstrapOVSFlows(nodeName string) error {
 	}
 
 	return nil
+}
+
+// localnetPortInfo returns the name of the bridge and the patch port name for the default cluster network
+func localnetPortInfo(nodeName string, portsOutput string) (string, string) {
+	// This needs to work with:
+	// - default network: patch-<bridge name>_<node>-to-br-int
+	// but not with:
+	// - user defined primary network: patch-<bridge name>_<network-name>_<node>-to-br-int
+	// - user defined secondary localnet network: patch-<bridge name>_<network-name>_ovn_localnet_port-to-br-int
+	// TODO: going forward, maybe it would preferable to just read the bridge name from the config.
+	r := regexp.MustCompile(fmt.Sprintf("^patch-([^_]*)_%s-to-br-int$", nodeName))
+	for _, line := range strings.Split(portsOutput, "\n") {
+		matches := r.FindStringSubmatch(line)
+		if len(matches) == 2 {
+			return matches[1], matches[0]
+		}
+	}
+	return "", ""
 }
--- a/go-controller/pkg/node/openflow_manager_test.go
+++ b/go-controller/pkg/node/openflow_manager_test.go
@@ -1,0 +1,79 @@
+package node
+
+import "testing"
+
+func TestOpenFlowManagerDefaultNetOVSBridgeFinder(t *testing.T) {
+	const nodeName = "multi-homing-worker-0.maiqueb.org"
+
+	testCases := []struct {
+		name                  string
+		desc                  string
+		inputPortInfo         string
+		expectedBridgeName    string
+		expectedPatchPortName string
+	}{
+		{
+			name:                  "empty input ports",
+			inputPortInfo:         "",
+			expectedBridgeName:    "",
+			expectedPatchPortName: "",
+		},
+		{
+			name:                  "input ports without patch ports",
+			inputPortInfo:         "port1",
+			expectedBridgeName:    "",
+			expectedPatchPortName: "",
+		},
+		{
+			name: "input ports with a patch port",
+			inputPortInfo: `
+port1
+port2
+patch-br-ex_multi-homing-worker-0.maiqueb.org-to-br-int`,
+			expectedBridgeName:    "br-ex",
+			expectedPatchPortName: "patch-br-ex_multi-homing-worker-0.maiqueb.org-to-br-int",
+		},
+		{
+			name: "input ports with a patch port for a localnet network",
+			inputPortInfo: `
+port1
+port2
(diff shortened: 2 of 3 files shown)


go-controller/pkg/node/openflow_manager.go
```diff
@@ -338,3 +327,25 @@ func bootstrapOVSFlows() error {
...
+	var bridge string
+	var patchPort string
+	// patch-br-int-to-<bridge name>_<node>
+	r := regexp.MustCompile("^patch-(.*)_.*?-to-br-int$")
+	for _, line := range strings.Split(portsOutput, "\n") {
+		matches := r.FindStringSubmatch(line)
+		if len(matches) == 2 {
+			patchPort = matches[0]
+			bridge = matches[1]
+			if strings.Contains(bridge, "ovn_localnet") {
```
maiqueb, line 342: maybe if we change the capture on the regex above to be non-greedy I can match against the `ovn_localnet_port` constant here. 

I didn't want to update it, but it maybe provides better readability. 

Thoughts ?

maiqueb: This is half non-sense. I think it is better now.

go-controller/pkg/node/openflow_manager_test.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+		{
+			name: "input ports with a patch port for the default network and a localnet",
+			inputPortInfo: `
+port1
+port2
+patch-vlan2003_ovn_localnet_port-to-br-int
+patch-br-ex_multi-homing-worker-0.maiqueb.org-to-br-int`,
+			expectedBridgeName:    "br-ex",
+			expectedPatchPortName: "patch-br-ex_multi-homing-worker-0.maiqueb.org-to-br-int",
```
tssurya, line 51: related to this pr but slightly different dimension.. but I was also hitting:
```
F0802 02:59:29.051867   23805 ovnkube.go:136] failed to run ovnkube: failed to start node network controller: failed to start default node network controller: failed to bootstrap OVS flows: failed to get flows on bridge "breth0_l3.network":, stderr: "ovs-ofctl: breth0_l3.network is not a bridge or a socket\n", error: exit status 1
```
in my PR for OPFM, my understanding is @dceara  did https://github.com/ovn-org/ovn-kubernetes/pull/4476/commits/bbb476b27cc32c74d2df686f3a7f0bd2d0889b5b

wondering if that will also be still needed ...I guess so? given we are only thinking about localnet harcoded based parsing...

tssurya: just wondering if there are broader set of "to be excluded ports" and if the reverse is easier i.e allowlisting the known ports for which we need to setup basic flows rather than excluding an unknown list?

tssurya (commented):
logic is fine, thanks for the tests!
a few questions...

go-controller/pkg/node/openflow_manager_test.go
```diff
@@ -0,0 +1,66 @@
...
+port1
+port2
+patch-vlan2003_ovn_localnet_port-to-br-int`,
+			expectedBridgeName:    "",
+			expectedPatchPortName: "",
+		},
+		{
+			name: "input ports with a patch port for the default network and a localnet",
+			inputPortInfo: `
+port1
```
tssurya, line 46: usually is this function also parsing by new line ? is localnetInfo expecting a list of ports or a single port... ?

go-controller/pkg/node/openflow_manager_test.go
maiqueb: well, we can read from the configuration and know which bridge is supposed to be the right one.

People will create all kinds of patch ports in their system ... IMHO we should remove the entire regex shenanigan and instead just read from the config.

go-controller/pkg/node/openflow_manager_test.go
maiqueb: list of ports. It is fed pretty much the output of `ovs-vsctl list interfaces` (just the name).

The command that extracts the port names outputs them separated by `\n`.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -338,3 +327,25 @@ func bootstrapOVSFlows() error {
 
 	return nil
 }
+
+// localnetInfo returns the patch port name and the bridge name for the default network
+func localnetInfo(portsOutput string) (string, string) {
```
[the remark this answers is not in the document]
maiqueb: Can do.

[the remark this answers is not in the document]
maiqueb: Done.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -338,3 +327,25 @@ func bootstrapOVSFlows() error {
 
 	return nil
 }
+
+// localnetPortInfo returns the bridge name and the patch port name for the default network
+func localnetPortInfo(portsOutput string) (string, string) {
```
trozet, line 332: I think @dceara fix is more reliable:
https://github.com/ovn-org/ovn-kubernetes/commit/bbb476b27cc32c74d2df686f3a7f0bd2d0889b5b

As long as we can be sure that only the default patch port will have the node name, it seems safer to filter on. testing with UDN it looks like it should work:

```
[root@ovn-worker ~]# ovs-vsctl list interface | grep to-br-int
options             : {peer=patch-breth0_default.l3.primary_ovn-worker-to-br-int}
options             : {peer=patch-localnet.network_ovn_localnet_port-to-br-int}
name                : patch-localnet.network_ovn_localnet_port-to-br-int
name                : patch-breth0_default.l3.primary_ovn-worker-to-br-int
options             : {peer=patch-breth0_ovn-worker-to-br-int}
name                : patch-breth0_ovn-worker-to-br-int
```

trozet (changes requested):

go-controller/pkg/node/openflow_manager.go
maiqueb: But you can't assure a patch port will feature the bridge name. 

Localnet ports for secondary networks of `localnet` topology do not feature the node name. Check the test I've introduced. For instance, the [issue reported](https://issues.redhat.com/browse/OCPBUGS-37433) features this patch port: `patch-vlan2003_ovn_localnet_port-to-br-int` 

This would work with UDNs for layer3 networks. It will not with localnet.

dceara: So, just to clarify the requirement:
- bootstrapOVSFlows() must be called **only for the  default network patch port**

The original regex was too loose, it would also match on user defined network patch ports.  https://github.com/ovn-org/ovn-kubernetes/commit/bbb476b27cc32c74d2df686f3a7f0bd2d0889b5b tries to fix that by updating the (already existing) regex.

However, I agree with @maiqueb, we should know through configuration what the bridge and node names are so we could replace the full ovs port dump + regex with the equivalent of:
```
ovs-vsctl find interface name=`patch-br-int-to-<bridge name>_<node>`
```
If the above returns a record (it can return max one record) then continue, get flows, ...

dceara: We also wouldn't need this part anymore:
```
	// Get ofport of patchPort
	ofportPatch, stderr, err := util.GetOVSOfPort("get", "Interface", patchPort, "ofport")
	if err != nil {
		return fmt.Errorf("failed while waiting on patch port %q to be created by ovn-controller and "+
			"while getting ofport. stderr: %q, error: %v", patchPort, stderr, err)
	}
```
The first `find` call above would get us all the information about the patch port, including its `ofport`.

dceara:
The test changes look good to me (I'll let other reviewers comment on the first commit).  Thanks @maiqueb!

Regards,
Dumitru

go-controller/pkg/node/openflow_manager.go
```diff
@@ -266,8 +266,11 @@ func bootstrapOVSFlows() error {
 
 	var bridge string
 	var patchPort string
-	// patch-br-int-to-<bridge name>_<node>
-	r := regexp.MustCompile("^patch-(.*)_.*?-to-br-int$")
+	// This needs to work with:
+	// - default network: patch-<bridge name>_<node>-to-br-int
+	// but not with:
+	// - secondary network: patch-<bridge name>_<node>-to-br-int
+	r := regexp.MustCompile(fmt.Sprintf("^patch-([^_]*)_%s-to-br-int$", nodeName))
```
trozet: this regexp isn't bulletproof cause you can have an OVS bridge with _ in it:

```
    Bridge br_blah
        Port br_blah
            Interface br_blah
                type: internal
    ovs_version: "3.2.2"
```

You should be able to use .*, since a kubernetes node name cannot have an underscore in the name.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -341,3 +327,21 @@ func bootstrapOVSFlows(nodeName string) error {
...
 	return nil
 }
+
+// localnetPortInfo returns the name of the bridge and the patch port name for the default cluster network
+func localnetPortInfo(nodeName string, portsOutput string) (string, string) {
+	// This needs to work with:
+	// - default network: patch-<bridge name>_<node>-to-br-int
+	// but not with:
+	// - user defined primary network: patch-<bridge name>_<network-name>_<node>-to-br-int
+	// - user defined secondary localnet network: patch-<bridge name>_<network-name>_ovn_localnet_port-to-br-int
```
trozet, line 337: i see you fixed this in the 2nd commit

trozet (changes requested):

go-controller/pkg/node/openflow_manager.go
maiqueb: if I run that regex, it breaks with our UDN names. 

For instance, we would extract this bridge name - `br-ex_tenant-blue` - from the patch port `patch-br-ex_tenant-blue_multi-homing-worker-0.maiqueb.org-to-br-int`. 

Our UDN names will add another `_` separator between the bridge and the node name name ...

trozet: Sorry I missed that the UDN networks have the node name in them and use underscore to separate the network. In that case we cannot use .*, but we also cannot then use underscore in our bridge names. What you propose is better than what we currently have, so lets merge it. In the future we can see if we can find a better mechanism of determining this port. Maybe add an external_id to the default patch port or something.

maiqueb: 100 %

PR #4584, opened 2024-08-04
hack/lint.sh: Use full image tag

ormergi:

When using podman, `make lint` fails due to the following error:
```
Error: short-name resolution enforced but cannot prompt without a TTY
make: *** [Makefile:93: lint] Error 125
```

Root cause is podman wont add `docker.io` registry name to container image tag if not specified like docker does.

Use full image name including the registry to avid failures on envs that use podman.

Fixes #

```release-note
NONE
```


ormergi:
Rebase

maiqueb (approved):
Thanks.

jcaamano (approved):

PR #4585, opened 2024-08-04
hack/lint.sh: Mount local golangci-lint cache

ormergi:

Currently, `make lint` runs golangci-linter in a container but wont mount golangci-lint local cache.
linter time can be optimized by using the local cache, in tested env lint time droped from 1:11 minutes to 45 seconds.

This PR changes hack/lint.sh to mount golangci-lint local cache to the subject container.

Fixes #

```release-note
NONE
```


maiqueb (approved):
Thank you.

PR #4505, opened 2024-07-05
Bump OVSDBTimeout and make it configurable.

npinaeva:
Make OVSDBTimeout for transactions a part of config.Global instead of const. A reasonable value for this timeout depends on the size of the cluster and making it configurable may help solve crashloop problems. It also speeds up testing as timeout may be set to lower value for unit tests.
`types.OVSDBTimeout` is still used in `libovsdb.newClient` for inactivity and reconnectTimeouts.
How much time a transaction takes depends on the size of this transaction, one of the most complicated examples is address set dbIDs sync, which needs to update address set name and all its references (i.e. acls, lrps, qoses matches). I have set batch size for transaction to 50 https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/ovn/external_ids_syncer/address_set/address_set_sync.go#L76 considering this is small enough, but a customer hit that timeout. 
Then I considered doing something similar to port group code https://github.com/ovn-org/ovn-kubernetes/blob/425a328cd1729d0dc9ce0a52936d47dee1702b4e/go-controller/pkg/ovn/external_ids_syncer/port_group/port_group_sync.go#L42-L76, but it is way too complicated for address sets with 3 different referencing object types, and looks like an overkill just to avoid a valid transaction timing out.
I am not aware of any reasoning for the existing db timeout = 10s, so I am trying to figure out which value makes sense by default. Considering the timeout is only there to avoid rpc call hanging, which should almost never happen, a couple of minutes sounds like a reasonable upper threshold.
Now getting back to 50 address set batch size (as there is not better reference for what should be supported) I measured locally that updating 50 address sets, each referencing 5000 ACLs takes 41s, with 10000 ACLs it takes 80s.
As I have seen clusters using thousands of ACLs, and namespace selector is quite popular (which is converted to the address sets reference in the ACL.match), I assume it is a good enough example to try to support.
That is how I ended up with default value of 100s, to make sure mentioned transactions succeed, but if the rpc call hangs, we can still recover.

PR #4590, opened 2024-08-07
Fix retry handling in case IP pool is empty

oshoval:
When a pod is created, and the IP pool is empty,
the pod will stuck forever in creation state, even if IPs are released.

To simulate the bug:
1. Create a NAD with a pool with just 1 available IP (others excluded; first and last reserved).
2. Create 2 pods that use this NAD:
a. One pod will take the IP.
b. The second pod will remain in "creation" state.
3. Wait around 1 minute and then delete the Running pod.

The bug: The pending pod stays pending forever, even though an IP is available.

Without the fix, update / add Pod retry mechanism isn't triggered at all, because it will be triggered
only if a non nil error is returned.

Note: 
Even with the fix, the retry mechanism is capped in 15 retries [1], which total for around 15m.
After 15m the pod will be pending forever, even if IPs are released, because its annotations
won't be updated anymore.
User will need to manually recreate the pod (unless the pod has a controller of course).

[1] https://github.com/ovn-org/ovn-kubernetes/blob/87fdae7bde5e51cbd6ab50682f9ed6e5e407ca62/go-controller/pkg/retry/obj_retry.go#L258

Added also the missing "return err" of pod delete event just because it was missing as well.
Removed the logs in case of an error, it will be printed anyhow on higher level.

Fixes #

Please find in the comments below manifests that allow simulating the bug easily upon needs.

```release-note
None
```

### DIFF
--- a/go-controller/pkg/clustermanager/network_cluster_controller.go
+++ b/go-controller/pkg/clustermanager/network_cluster_controller.go
@@ -319,6 +319,7 @@ func (h *networkClusterControllerEventHandler) AddResource(obj interface{}, from
 		if err != nil {
 			klog.Infof("Pod add failed for %s/%s, will try again later: %v",
 				pod.Namespace, pod.Name, err)
+			return err
 		}
 	case factory.NodeType:
 		node, ok := obj.(*corev1.Node)
@@ -359,6 +360,7 @@ func (h *networkClusterControllerEventHandler) UpdateResource(oldObj, newObj int
 		if err != nil {
 			klog.Infof("Pod update failed for %s/%s, will try again later: %v",
 				new.Namespace, new.Name, err)
+			return err
 		}
 	case factory.NodeType:
 		node, ok := newObj.(*corev1.Node)
--- a/go-controller/pkg/clustermanager/network_cluster_controller.go
+++ b/go-controller/pkg/clustermanager/network_cluster_controller.go
@@ -317,8 +317,6 @@ func (h *networkClusterControllerEventHandler) AddResource(obj interface{}, from
 		}
 		err := h.ncc.podAllocator.Reconcile(nil, pod)
 		if err != nil {
-			klog.Infof("Pod add failed for %s/%s, will try again later: %v",
-				pod.Namespace, pod.Name, err)
 			return err
 		}
 	case factory.NodeType:
@@ -358,8 +356,6 @@ func (h *networkClusterControllerEventHandler) UpdateResource(oldObj, newObj int
 		}
 		err := h.ncc.podAllocator.Reconcile(old, new)
 		if err != nil {
-			klog.Infof("Pod update failed for %s/%s, will try again later: %v",
-				new.Namespace, new.Name, err)
 			return err
 		}
 	case factory.NodeType:
@@ -392,8 +388,6 @@ func (h *networkClusterControllerEventHandler) DeleteResource(obj, cachedObj int
 		}
 		err := h.ncc.podAllocator.Reconcile(pod, nil)
 		if err != nil {
-			klog.Infof("Pod delete failed for %s/%s, will try again later: %v",
-				pod.Namespace, pod.Name, err)
 			return err
 		}
 	case factory.NodeType:
--- a/go-controller/pkg/clustermanager/network_cluster_controller.go
+++ b/go-controller/pkg/clustermanager/network_cluster_controller.go
@@ -394,6 +394,7 @@ func (h *networkClusterControllerEventHandler) DeleteResource(obj, cachedObj int
 		if err != nil {
 			klog.Infof("Pod delete failed for %s/%s, will try again later: %v",
 				pod.Namespace, pod.Name, err)
+			return err
 		}
 	case factory.NodeType:
 		node, ok := obj.(*corev1.Node)


oshoval:
This allows to simulate the bug
```
---
apiVersion: "k8s.cni.cncf.io/v1"
metadata:
  name: mynet
  namespace: default
spec:
  config: |2
    {
        "cniVersion": "0.3.1",
        "name": "tenantblue",
        "netAttachDefName": "default/mynet",
        "topology": "layer2",
        "subnets": "192.168.200.0/30",
        "excludeSubnets": "192.168.200.1/32"
    }
---
apiVersion: v1
kind: Pod
metadata:
  annotations:
    k8s.v1.cni.cncf.io/networks: '[{"name":"mynet","namespace":"default","interface":"pod8f202bdbf25"}]'
  name: pod1
  namespace: default
spec:
  containers:
  - image: docker.io/fedora:latest 
    name: fedora
    imagePullPolicy: "IfNotPresent"
    command:
    - /bin/bash
    - -c
    - sleep infinity
---
apiVersion: v1
kind: Pod
metadata:
  annotations:
    k8s.v1.cni.cncf.io/networks: '[{"name":"mynet","namespace":"default","interface":"pod8f202bdbf25"}]'
  name: pod2
  namespace: default
spec:
  containers:
  - image: docker.io/fedora:latest 
    name: fedora
    imagePullPolicy: "IfNotPresent"
    command:
    - /bin/bash
    - -c
    - sleep infinity
```

oshoval:
Added missing "return err" also in case of pod delete (not part of the bug, but just saw it symmetric wise)

maiqueb (approved):
Thank you.

maiqueb:
Is it possible to add some unit tests to this ?

oshoval:
> Is it possible to add some unit tests to this ?

Imho it doesnt worth it, it is bit more complex scenario for unit test,
we will have QE at the long run that do test it e2e
(looked on `TestPodAllocator_reconcileForNAD` which is the current place of unit tests)

oshoval:
not sure why build failed this time, just pushed empty changed 
(update also first commit change, just removed unrelevant text about delete pod, because it now has its own commit)

oshoval:
I think the builld failure is not related to this PR, here it looks the same from glance
https://github.com/ovn-org/ovn-kubernetes/pull/4594

oshoval:
no changes push - just to retrigger job one more time

oshoval:
Now it failed due to this iiuc
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10300314435/job/28509457479?pr=4590
2024-08-08T10:41:37.4612347Z     testing.go:1398: race detected during execution of test
SUCCESS! -- 429 Passed | 0 Failed | 1 Pending | 0 Skipped

oshoval:
I will split this PR to 2 PRs, one for fixing the case mentioned on the PR desc
and the 2nd for adding the missing pod delete "return err"
the various flakes / data race started when i added the "return err", maybe it woke up something
so with the split we will be safer, able to focus and also see if it is related

oshoval:
now the failures seems like known flakes

oshoval:
rebased

oshoval:
seems the UT hanged
````
2024-08-08T13:53:01.5120574Z I0808 13:53:01.511449   34830 route_manager.go:110] Route Manager: completed adding route: {Ifindex: 67 Dst: 0.0.0.0/0 Src: <nil> Gw: <nil> Flags: [] Table: 1067 Realm: 0}
2024-08-08T19:43:09.3217245Z ##[error]The operation was canceled.
2024-08-08T19:43:09.3317234Z Post job cleanup.
```

oshoval:
rebased

maiqueb (approved):
Thanks. 

Is there something I'm not seeing here ? At a glance, we should return an error here to force reconciliation to happen later.
@jcaamano / @trozet / @tssurya

jcaamano:
I see this as a global problem though. If there is a reason we shouldn't return err there, is probable that reason is the same for the three cases (create/update/delete). Perhaps chasing the data race tells us more that we are not realizing right know.
I don't realize what reason that might be right now.
Also, if we are returning error we can avoid the log there. It will be logged at a higher level.

I don't see any relation of the data race with this PR, seem more cause by this other PR
https://github.com/ovn-org/ovn-kubernetes/pull/4505
cc @npinaeva

oshoval:
> I see this as a global problem though. If there is a reason we shouldn't return err there, is probable that reason is the same for the three cases (create/update/delete). Perhaps chasing the data race tells us more that we are not realizing right know. I don't realize what reason that might be right now. Also, if we are returning error we can avoid the log there. It will be logged at a higher level.
> 
> I don't see any relation of the data race with this PR, seem more cause by this other PR #4505 cc @npinaeva

Added missing return err on pod delete.
Removed the log.

Rebased

Thanks

jcaamano:
> > I see this as a global problem though. If there is a reason we shouldn't return err there, is probable that reason is the same for the three cases (create/update/delete). Perhaps chasing the data race tells us more that we are not realizing right know. I don't realize what reason that might be right now. Also, if we are returning error we can avoid the log there. It will be logged at a higher level.
> > I don't see any relation of the data race with this PR, seem more cause by this other PR #4505 cc @npinaeva
> 
> Added missing "return err" on pod delete. Removed the log.
> 
> Rebased
> 
> Thanks

Should we do the same for update?

oshoval:
> > > I see this as a global problem though. If there is a reason we shouldn't return err there, is probable that reason is the same for the three cases (create/update/delete). Perhaps chasing the data race tells us more that we are not realizing right know. I don't realize what reason that might be right now. Also, if we are returning error we can avoid the log there. It will be logged at a higher level.
> > > I don't see any relation of the data race with this PR, seem more cause by this other PR #4505 cc @npinaeva
> > 
> > 
> > Added missing "return err" on pod delete. Removed the log.
> > Rebased
> > Thanks
> 
> Should we do the same for update?

already did, originally i had to fix update and add so pods will have a retry chance for both cases

npinaeva:
@jcaamano I think https://github.com/ovn-org/ovn-kubernetes/pull/4505 didn't bring any test failures by itself, but just made existing problems more visible.
Here is an attempt to fix ef race https://github.com/ovn-org/ovn-kubernetes/pull/4505

jcaamano:
> @jcaamano I think #4505 didn't bring any test failures by itself, but just made existing problems more visible. Here is an attempt to fix ef race #4611

Wouldn't we potentially have a race of `PrepareTestConfig` for a test with any other test that might be concurrently executing and reading config values?

oshoval:
the failing test is a reported flake https://github.com/ovn-org/ovn-kubernetes/issues/4432
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10353880486/job/28694475459?pr=4590

https://github.com/ovn-org/ovn-kubernetes/actions/runs/10353880486/job/28694471428?pr=4590 still running so didnt check it

oshoval:
Thanks all

PR #4591, opened 2024-08-07
EIP node controller: use pkt mark for RP filter only for egress inter..

martinkennelly:
…faces

Previous to this change, we enabled `src_valid_mark` for all interfaces when in LGW mode.
This PR sets it only for interfaces that require it (ext bridge + EIP interfaces).

cc @andreaskaris

### DIFF
--- a/go-controller/pkg/node/controllers/egressip/egressip.go
+++ b/go-controller/pkg/node/controllers/egressip/egressip.go
@@ -136,12 +136,13 @@ type Controller struct {
 	iptablesManager *iptables.Controller
 	kube            kube.Interface
 	nodeName        string
+	extBridgeName   string
 	v4              bool
 	v6              bool
 }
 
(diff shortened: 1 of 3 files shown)


martinkennelly:
@andreaskaris Unsure if this actually fixes your issue as ext bridge will still have the sysctl setting enabled..

martinkennelly:
Forced to rebase again - no changes

andreaskaris:
That fix looks good to me, I like that it doesn't override the system defaults any more. However, we'll need confirmation from the entity that reported the issue, as I'm not talking to them directly

martinkennelly:
@andreaskaris do you think its good enough to go ahead with this or do we need to get more info from the reporter?

flavio-fernandes (approved):
Changes lgtm. TY @martinkennelly !

martinkennelly:
close - no traction

PR #4548, opened 2024-07-19
UDN: Add VRF Manager

tssurya:
Depends-On: https://github.com/ovn-org/ovn-kubernetes/pull/4546

Credit to @pperiyasamy ; taken from https://github.com/ovn-org/ovn-kubernetes/pull/4471

#### What this PR does and why is it needed

- [x] Adds a new vrf manager similar to route manager but doesn't use channels and uses events so that we can do stuff with the error returned
- [x] Invokes the vrf manager from the node's NCM so that it can be used to program VRFs for all UDNs
- [x] Given same manager routine will be called from different network controllers; everything done in VRF manager is behind a lock
- [x] Every 2 mins reconcile syncs linux vrfs with what's expected in the cache (need to revisit this along with planning for a repair)
- [x] Also handles deleting the vrf when the network is deleted

#### How to verify it
```
11: mp1-vrf: <NOARP,MASTER,UP,LOWER_UP> mtu 65575 qdisc noqueue state UP mode DEFAULT group default                                                                          
    link/ether d2:80:26:9c:b0:b9 brd ff:ff:ff:ff:ff:ff promiscuity 0  allmulti 0 minmtu 1280 maxmtu 65575                                                                    
    vrf table 1010 addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535 tso_max_size 65536 tso_max_segs 65535 gro_max_size 65536              
13: mp2-vrf: <NOARP,MASTER,UP,LOWER_UP> mtu 65575 qdisc noqueue state UP mode DEFAULT group default                                                                          
    link/ether 5e:50:15:24:e8:3c brd ff:ff:ff:ff:ff:ff promiscuity 0  allmulti 0 minmtu 1280 maxmtu 65575                                                                    
    vrf table 1012 addrgenmode eui64 numtxqueues 1 numrxqueues 1 gso_max_size 65536 gso_max_segs 65535 tso_max_size 65536 tso_max_segs 65535 gro_max_size 65536         
```
where mp1 is L3 and mp2 is L2.
```
10: ovn-k8s-mp1: <BROADCAST,MULTICAST,PROMISC,UP,LOWER_UP> mtu 1300 qdisc noqueue master mp1-vrf state UNKNOWN group default qlen 1000
    link/ether f6:21:8a:9d:87:cf brd ff:ff:ff:ff:ff:ff
    inet 10.128.2.2/24 brd 10.128.2.255 scope global ovn-k8s-mp1
[...]

PR #4593, opened 2024-08-07
EIP Multi NIC: do not config role primary secondary network namespaces

martinkennelly:
Extracted a commit from PR https://github.com/ovn-org/ovn-kubernetes/pull/4530/

Once #4548 merges, I also need to block assignment of EIP IPs to a link that is a slave to UDN VRF.


martinkennelly:
`ovn-ci / e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones)` failed with known flake

martinkennelly:
@tssurya can you add this PR to the review list ?

tssurya:
Yup Wil do can you add the feature labels for udn tp this?

martinkennelly:
@tssurya anyone assigned to review this?

go-controller/pkg/node/controllers/egressip/egressip_test.go
```diff
@@ -1139,6 +1140,49 @@ var _ = ginkgo.Describe("VRF", func() {
...
+				for _, address := range addresses {
+					ipsOnLink = append(ipsOnLink, address.IP.String())
+				}
+				return nil
+			})
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "execution in namespace should succeed")
+			return ipsOnLink
+		}).ShouldNot(gomega.ConsistOf(egressIP1IPV4))
+		gomega.Expect(cleanupControllerFn()).ShouldNot(gomega.HaveOccurred())
+		gomega.Expect(cleanupNodeFn()).ShouldNot(gomega.HaveOccurred())
```
maiqueb, line 1184: maybe these can be moved to a test teardown block in this context (`AfterEach`).

Please also move the test setup section (getting the testConfig, creating the test ns, etc) to a `BeforeEach` block.
It will also make the previous test easier to follow.

go-controller/pkg/node/controllers/egressip/egressip_test.go
```diff
@@ -1139,6 +1140,49 @@ var _ = ginkgo.Describe("VRF", func() {
...
+		ginkgo.By("create VRF and add link")
+		gomega.Expect(createVRFAndEnslaveLink(testNS, dummyLink1Name, vrfName, vrfTable)).Should(gomega.Succeed())
+		ginkgo.By("configure EgressIP to be hosted by enslaved link")
+		egressIPList := []egressipv1.EgressIP{*newEgressIP(egressIP1Name, egressIP1IPV4, node1Name, namespace1Label, egressPodLabel)}
+		pods := []corev1.Pod{newPodWithLabels(namespace1, pod1Name, node1Name, pod1IPv4, egressPodLabel)}
+		namespaces := []corev1.Namespace{newNamespaceWithLabels(namespace1, namespace1Label)}
+		ginkgo.By("start controller")
+		c, _, err := initController(namespaces, pods, egressIPList, nodeConfig, true, false, true)
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+		cleanupControllerFn, err := runController(testNS, c)
```
maiqueb, line 1166: I think all of this could be moved to a `BeforeEach` block.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -822,6 +832,58 @@ func (bnc *BaseNetworkController) GetNetworkRole(pod *kapi.Pod) (string, error)
...
+			return types.NetworkRolePrimary, nil
+		}
+		return types.NetworkRoleSecondary, nil
+	}
+	activeNetwork, err := bnc.getActiveNetworkForNamespace(ns.Name)
+	if err != nil {
+		if util.IsUnknownActiveNetworkError(err) {
+			bnc.recordNamespaceErrorEvent(ns, err)
+		}
+		return "", err
```
maiqueb, line 873: should we enrich the error ? Or is whatever we return from `getActiveNetworkForNamespace` good enough ?

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -822,6 +832,58 @@ func (bnc *BaseNetworkController) GetNetworkRole(pod *kapi.Pod) (string, error)
...
+// (3) "infrastructure-locked" is applicable only to "default" network if
+//
+//	a user defined network is the "primary" network for this namespace. This
+//	signifies the "default" network is only used for probing and
+//	is otherwise locked for all intents and purposes.
+//
+// NOTE: Like in other places, expectation is this function is always called
+// from controller's that have some relation to the given namespace, unrelated
+// networks are treated as secondary networks so caller has to be careful
+func (bnc *BaseNetworkController) GetNetworkRoleForNamespace(ns *kapi.Namespace) (string, error) {
```
maiqueb, line 858: I don't understand where you use this function. Hence, I don't understand why you needed to introduce it.

(I am reviewing commit by commit; maybe this comment doesn't apply if I had reviewed the full PR at once ...)

go-controller/pkg/ovn/egressip.go
```diff
@@ -189,6 +189,18 @@ func (oc *DefaultNetworkController) reconcileEgressIP(old, new *egressipv1.Egres
 				return err
 			}
 			for _, namespace := range namespaces {
+				if util.IsNetworkSegmentationSupportEnabled() {
```
maiqueb, line 192: you also check this in the `GetNetworkRoleForNamespace` function. Isn't it redundant then ?

go-controller/pkg/ovn/egressip.go
```diff
@@ -2307,3 +2387,14 @@ func getPodNamespaceAndNameFromKey(podKey string) (string, string) {
...
+
+// getFirstNonNilNamespace must return a non-nil pointer to a Namespace otherwise panic
+func getFirstNonNilNamespace(n1, n2 *v1.Namespace) *v1.Namespace {
+	if n1 != nil {
+		return n1
+	}
+	if n2 != nil {
+		return n2
+	}
+	panic("getNonNilNamespace(): received two nil Namespace pointers")
```
maiqueb, line 2399: IMHO it would be better if we could avoid the panic, and properly handle the error.

maiqueb (changes requested):
Most of my requests are opinionated.

The one area where I think you could definitely improve the PR is on the introduced unit tests - the test's readability would be improved if you split the added test into a common test setup; test; test teardown pattern by introducing `BeforeEach` / `AfterEach` blocks.

Overall, looks good.

go-controller/pkg/ovn/base_network_controller.go
maiqueb: I've noticed it is used in the next commit. To me, that is a clear sign these 2 commits should be squashed together. 

Not a must - this is an opinionated nit.

martinkennelly:
Not going to use this now.

PR #4596, opened 2024-08-07
OVN-Observability: enable sampling-based observability

npinaeva:
Observability feature uses OVN sampling functionality to generate samples with requested metadata when
specific OVS flows are matched. 
See docs commit for more info.

Depends on https://github.com/ovn-org/ovn-kubernetes/pull/4658


npinaeva:
all known flakes, except 1 new https://github.com/ovn-org/ovn-kubernetes/pull/4597

go-controller/pkg/libovsdb/ops/sample.go
```diff
@@ -0,0 +1,170 @@
...
+		},
+		OnModelUpdates: []interface{}{&sample.Collectors},
+		ErrNotFound:    false,
+		BulkOp:         false,
+	}
+	opModels = append(opModels, opModel)
+	return opModels
+}
+
+func GetSampleID(dbObj hasExternalIDs) uint32 {
```
dceara, line 149: Maybe I'm misunderstanding how this works but shouldn't the sample ID be unique per ACL?  Doesn't this function have a (small) chance of collision?

dist/images/Dockerfile.fedora.dev
```diff
@@ -83,8 +83,10 @@ COPY ovn-kube-util /usr/bin/
 COPY ovndbchecker /usr/bin/
 COPY hybrid-overlay-node /usr/bin
 COPY ovnkube-identity /usr/bin/
+COPY ovnkube-observ /usr/bin
 COPY ovn-k8s-cni-overlay /usr/libexec/cni/ovn-k8s-cni-overlay
 
+
```
dceara, line 89: Nit: unrelated?

docs/observability/ovn-observability.md
```diff
@@ -0,0 +1,99 @@
...
+
+### OVN-Kubernetes Implementation Details
+
+`Sample_collector` and `Sampling_app` are created or cleaned up when the observability is enabled/disabled on startup.
+When one of the supported objects (for example, network policy) is created, ovn-kuberentes generates an nbdb `Sample` for it.
+
+To decode the samples into human-readable information, `go-controller/observability-lib` is used. It finds `Sample`
+by the attached `Sample.Metadata` and then gets corresponding db object based on `Sampling_add.ID` and `Sample.UUID`.
+The message is then constructed using db object `external_ids`.
+
```
dceara, line 82: Maybe we should add a note/known limitation/whatever here saying that sessions established while sampling was enabled for the ACLs they hit will still be sampled after the sampling config is removed (conntrack).  The opposite is also true: if a session was already established before sampling was enabled we won't be getting samples for it afterwards.

dceara (commented):
I only did a first pass review of the PR, looks OK overall, I only had very minor comments. I do plan to try it out in the near future.

dist/images/ovnkube.sh
```diff
@@ -96,6 +96,7 @@ fi
 # OVN_NORTHD_BACKOFF_INTERVAL - ovn northd backoff interval in ms (default 300)
 # OVN_ENABLE_SVC_TEMPLATE_SUPPORT - enable svc template support
 # OVN_ENABLE_DNSNAMERESOLVER - enable dns name resolver support
+# OVNKUBE_OBSERV_ENABLE - enable observability for ovnkube
```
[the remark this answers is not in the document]
npinaeva: we are not really good with the consistency here, https://github.com/ovn-org/ovn-kubernetes/blob/9030c8acbf6b6ad8e1aff1ad271374675e297672/dist/images/ovnkube.sh#L77-L81
I just prefer the same name across all files, so following `kind-helm.sh` together with `kind.sh` naming, so that we don't call the same optinon `OVN_OBSERV_ENABLE` in one file and `OVN_ENABLE_OBSERV` in another.

go-controller/pkg/libovsdb/ops/sample.go
npinaeva: yes, there is always a small chance of collision, but I don't see any other option that would guarantee consistent hashing and no collision.
this function is inspired by https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/util/util.go#L253 that we used before to generate address set and port group names.

dist/images/ovnkube.sh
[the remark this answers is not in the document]
dceara: Ack.

go-controller/pkg/libovsdb/ops/sample.go
dceara: OK, I guess we have to live with that for now.

docs/observability/ovn-observability.md
```diff
@@ -0,0 +1,107 @@
...
+insights of what ovn-kubernetes is doing with a packet and why.
+
+### User-Stories/Use-Cases
+
+- As a user I want to make sure that the network policies/egress firewalls/etc. are correctly enforced in my cluster.
+- As a cluster admin I want to check why some traffic is allowed or dropped.
+
+## How to enable this feature on an OVN-Kubernetes cluster?
+
+To enable this feature, use `--observability` flag with `kind.sh` script or `--enable-observability` flag with `ovnkube` binary.
```
dceara, line 38: I think we need another note mentioning that for now users that want to try this in kind need to bring up a cluster with the `Dockerfile.fedora.dev` dockerfile, that will use latest OVS main.  There's currently no fedora OVS build that supports psample.

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -0,0 +1,290 @@
...
+func getEndian() binary.ByteOrder {
+	// Use network bite order
+	return binary.BigEndian
+}
+
+func getLocalNBClient(ctx context.Context) (client.Client, error) {
+	config := nbConfig{
+		address: "unix:/var/run/ovn/ovnnb_db.sock",
+		scheme:  "unix",
+	}
```
dceara, line 46: This restricts the support to IC (single node per zone) deployments.  I understand it's more work to make this more generic but it should be possible to allow connections to remote NBs.  The rest of the ovnkube components can connect to remote SBs.  Would it make sense to add optional command line arguments to ovnkube-observ to allow the users to pass a custom NB connection string (and potentially more SSL related arguments) in order to support non-IC too?

dceara: I think we should at least document this limitation in `ovn-observability.md`.

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -0,0 +1,290 @@
...
+		return libovsdbOvnNBClient, nil
+	}
+	config = nbConfig{
+		address: "unix:/var/run/ovn-ic/ovnnb_db.sock",
+		scheme:  "unix",
+	}
+	var err2 error
+	libovsdbOvnNBClient, err2 = NewNBClientWithConfig(ctx, config)
+	if err2 != nil {
+		return nil, fmt.Errorf("error creating libovsdb client: [ with /var/run/ovn/ovnnb_db.sock: %w, with /var/run/ovn-ic/ovnnb_db.sock: %w ]", err1, err2)
```
dceara, line 58: Nit: maybe don't hardcode the paths, we have the last one in the `config` variable.  We could, I guess, use two `config` variables.

Also, this kind of gives the impression that we support non-IC, while in reality we don't. :)

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -0,0 +1,290 @@
...
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/observability"
+)
+
+type SampleDecoder struct {
+	nbClient          client.Client
+	ovsdbClient       client.Client
+	cleanupCollectors []int
+}
+
+type nbConfig struct {
```
dceara, line 22: We use this for both NB and OVS DB, I think a better name is: `type dbConfig struct`

go-controller/pkg/network-controller-manager/network_controller_manager.go
```diff
@@ -380,7 +381,19 @@ func (cm *NetworkControllerManager) Start(ctx context.Context) error {
...
+		if err = observabilityManager.Init(); err != nil {
+			return fmt.Errorf("failed to init observability manager: %w", err)
+		}
+	} else {
+		err = observability.Cleanup(cm.nbClient)
+		if err != nil {
+			klog.Warningf("Observability cleanup failed, expected if not all Samples ware deleted yet: %v", err)
+		}
+	}
+	err = cm.initDefaultNetworkController(observabilityManager)
```
dceara, line 396: Shouldn't we pass this same observabilityManager to all secondary network controllers (for user defined networks)?  IIUC user defined networks support network policy, we might as well make them "observable".

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,210 @@
...
+
+// OVN observ app IDs. Make sure to always add new apps in the end.
+const (
+	DropSamplingID = iota + 1
+	ACLNewTrafficSamplingID
+	ACLEstTrafficSamplingID
+)
+
+// temporary const, until we have dynamic config
+const DefaultObservabilityCollectorSetID = 28
```
dceara, line 22: Why 28?  If it's random I would've chosen 42. :)

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	m.sampConfig = libovsdbops.NewSamplingConfig(featuresConfig)
+	return nil
+}
+
+func (m *Manager) setDbCollectors() error {
+	clear(m.dbCollectors)
+	collectors, err := libovsdbops.FindSampleCollectorWithPredicate(m.nbClient, func(collector *nbdb.SampleCollector) bool {
+		return true
+	})
```
dceara, line 88: Shouldn't we add a `libovsdbops.ListSampleCollectors(nbClient)`?  We do it for other tables too.

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,210 @@
...
+		var ok bool
+		if collectorUUID, ok = m.dbCollectors[collectorKey]; !ok {
+			collectorID, err := m.getFreeCollectorID()
+			if err != nil {
+				return sampleFeaturesConfig, err
+			}
+			collector := &nbdb.SampleCollector{
+				ID:          collectorID,
+				SetID:       conf.collectorSetID,
+				Probability: probability,
```
dceara, line 191: `NB.Sample_Collector` has an `external_ids` column.  Does it make sense to store something there?  Maybe something to identify the features that require that probability

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,210 @@
...
+	return m.sampConfig
+}
+
+func (m *Manager) Init() error {
+	if err := m.setSamplingAppIDs(); err != nil {
+		return err
+	}
+	if err := m.setDbCollectors(); err != nil {
+		return err
+	}
```
dceara, line 66: What about potentially stale `NB.Sample_Collector` records, e.g. after upgrade some features have different probabilities?  AFAICT we don't delete any stale records in the DB.  Aside from the fact that those wouldn't be used for anything they also take up available IDs (and OVN currently only allows max 255 of those).

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// Cleanup must be called when observability is no longer needed.
+// It will return an error if some samples still exist in the db.
+// This is expected, and Cleanup may be retried on the next restart.
+func Cleanup(nbClient libovsdbclient.Client) error {
+	// Do the opposite of init
+	err := libovsdbops.DeleteSamplingAppsWithPredicate(nbClient, func(app *nbdb.SamplingApp) bool {
+		return true
+	})
```
dceara, line 106: I had a comment above about "return true" predicates.  But I see now you use it in more places.  Is it an agreed way of doing things?

go-controller/pkg/observability/observability.go
dceara: To clarify: by "identify" I meant "to make it easier for a developer/engineer/etc to identify what features the collector corresponds to directly by looking at the NB record"

go-controller/pkg/observability/observability_test.go
```diff
@@ -0,0 +1,135 @@
...
+	} {
+		msg := "db is empty"
+		if dbSetup != nil {
+			msg = "db is not empty"
+		}
+		When(msg, func() {
+
+			It("should initialize database", func() {
+				startManager(dbSetup)
+				Eventually(nbClient).Should(libovsdbtest.HaveData(initialDB))
```
dceara, line 87: If we add stale Sample_Collector cleanup we should probably test for that too here somewhere.

go-controller/observability-lib/sampledecoder/nb_client.go
```diff
@@ -0,0 +1,110 @@
+package sampledecoder
```
dceara, line 1: This module is more than just a `nb_client`.  It also configures the OVS database.  Should it be called `db_client.go`?

go-controller/observability-lib/sampledecoder/nb_client.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/cenkalti/backoff/v4"
+	"github.com/ovn-org/libovsdb/client"
+	"github.com/ovn-org/libovsdb/model"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/observability-lib/ovsdb"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
+	"k8s.io/klog/v2/textlogger"
+)
+
+const OVSDBTimeout = 100 * time.Second
```
dceara, line 17: 100 seconds seems a bit too long.  For other DB clients, e.g., `libovsdb.newClient()` we use 10 seconds.

Also, why not use the libovsdb.newClient() directly?  It would allow support for remote databases too (see the non-IC comment I made earlier).

go-controller/observability-lib/parse_sample.go
```diff
@@ -0,0 +1,241 @@
...
+			if r.logCookie || r.decoder != nil {
+				if attr.Type == PSAMPLE_ATTR_USER_COOKIE {
+					if uint64(len(attr.Value)) == sampledecoder.CookieSize {
+						c := sampledecoder.Cookie{}
+						err := binary.Read(bytes.NewReader(attr.Value), sampledecoder.SampleEndian, &c)
+						if err != nil {
+							return err
+						}
+						if r.logCookie {
+							r.cookieStr[1] = fmt.Sprintf("obs_domain=%v, obs_point=%v", c.ObsDomainID, c.ObsPointID)
```
dceara, line 203: Nit: Maybe it's because I'm used to 80 column line lengths but I think I'd add `parseSampleGroup()/parseUserCookie()/parseData()` functions.

dceara (dismissed):
I think the only major concern I currently have is the fact that we don't cleanup stale NB.Sample_Collector records.  Aside from that the rest of the code looks mostly OK to me - I left a bunch of other smaller comments.

Thanks!

dceara (changes requested):
See: https://github.com/ovn-org/ovn-kubernetes/pull/4596#pullrequestreview-2243477165

go-controller/pkg/observability/observability.go
amorenoz: I also found this rather strange. It seems a limitation of the `modelClient`.

go-controller/observability-lib/ovsdb/model.go
```diff
@@ -0,0 +1,319 @@
+// Code generated by "libovsdb.modelgen"
+// Edited by npinaeva to only leave required tables.
```
amorenoz, line 2: Editing a generated file can hide problems in the future (e.g: someone updates the schema and forgets to rebase your change and you are suddenly adding _all_ tables to your cache).

You can build your "smaller" db model map in the client code directly since `NewClientDBModel`  is public

dceara: Thinking more about this, I don't expect a lot of records to live in the OVS database so monitoring the whole thing is probably acceptable.  Maybe we can just leave all tables here.

amorenoz: It'll be a waste of resources IMHO. It's not only about memory (we need to put the entire OVSDB in the cache) but also processing time as every change in the OVSDB (e.g: interface statistics, port removal/creation, etc) will cause a notification to be received so the cache is updated. I don't think that is necessary at all when you can just easily move those lines to `nb_client.go`:

```go
	clientDBModel, err :=  model.NewClientDBModel("Open_vSwitch", map[string]model.Model{
		"Bridge":                    &Bridge{},
		"Flow_Sample_Collector_Set": &FlowSampleCollectorSet{},
	})

```

dceara: OK, this is simple indeed.  +1 for this then.

go-controller/observability-lib/sampledecoder/sample_decoder.go
npinaeva: that is a good point, I plan to add remote nbdb option in the future, when we have e2es for this feature to make sure that option works. As of now only local nbdb option will be used and tested downstream, I will add this note to the docs.

go-controller/pkg/network-controller-manager/network_controller_manager.go
npinaeva: I was going to to that in the next release, as I think secondary network observability may need some more data, e.g. to which secondary network given network policy belongs, and also I didn't test this functionality yet.
I think I should just document this limitation for now and make sure I add this feature later.

go-controller/pkg/observability/observability.go
npinaeva: it is random :D and reviewer is always right, so 42 it is :)

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (m *Manager) getFreeCollectorID() (int, error) {
+	for i := 1; i <= maxCollectorID; i++ {
+		if !m.takenCollectorIDs.Has(i) {
+			return i, nil
+		}
+	}
+	return 0, fmt.Errorf("no free collector IDs")
+}
```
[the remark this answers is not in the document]
npinaeva: I was thinking the same, also creating a new collector is a rare operation, so I think performance-wise this is not a problem. 
At the same time, predictable collector IDs make unit testing easier.

go-controller/pkg/observability/observability.go
npinaeva: hmm that sounds potentially useful, let me try that

go-controller/pkg/observability/observability.go
npinaeva: that is not possible yet, as the only supported options are on/off without config change, but I have added some functionality to do that + unit tests (as it is not used by the real code yet)

go-controller/pkg/observability/observability.go
npinaeva: for the previous comment to get all objects, we use `nbCleint.List`, and I updated it.
But to delete, I don't think we have functions like deleteAll, as libovsdb doesn't provide plain Delete function, only conditional delete https://github.com/ovn-org/libovsdb/blob/main/client/api.go#L16

go-controller/observability-lib/sampledecoder/nb_client.go
npinaeva: that was an attempt to decrease the amount of dependencies for this lib as asked by the observ team, since libovsdb.go has quite some imports that were not used at the time.
Now when we know we want to support remote-db connection, I probably will end up using it directly at some point.
About timeout - we updated it recently in ovn-k to default to 100 seconds, but that is mostly for large db transactions, that are not expected with the lib, so I can make it 10 seconds.

go-controller/observability-lib/parse_sample.go
npinaeva: heheh yeah I see what you mean, but there are some message-scoped variables that are set inside potential parsing sub-functions, and I feel like it is easier to follow when it is in the same function body...
But! I can make the lines a bit shorter :D

go-controller/observability-lib/ovsdb/model.go
npinaeva: What I tried to avoid is adding all db files to `ovsdb` folder, while in reality I only need 2 tables.
I think it would be nice to have automation to generate models only for a subset of tables, but libovsdb can't do that yet.
I have added a manually created file `observ_model.go` and removed `model.go`.
The idea is that no one should ever touch that folder until we need new bridge or flow_sample_collector features (potentially never). And hopefully the absence of other model files makes it obvious that we don't use them.
If someone re-generates models from scratch, they will have a bunch of new files, and no changes on the indexed files (until bridge or flow_sample_collector_set changes).

npinaeva:
first push is fixes, second is rebase

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -0,0 +1,290 @@
...
+	// find br-int UUID to attach collector
+	bridges := []*ovsdb.Bridge{}
+	err = d.ovsdbClient.WhereCache(func(item *ovsdb.Bridge) bool {
+		return item.Name == bridgeName
+	}).List(context.Background(), &bridges)
+	if err != nil || len(bridges) != 1 {
+		return fmt.Errorf("failed finding br-int: %w", err)
+	}
+
+	ops, err := d.ovsdbClient.Create(&ovsdb.FlowSampleCollectorSet{
```
dceara, line 258: What happens if the collector already existed (with the right configuration)?  Won't this Create fail with index uniqueness violation because there's already an entry with the same ID and bridge UUID:

https://github.com/openvswitch/ovs/blob/main/vswitchd/vswitch.ovsschema#L577

npinaeva: there is a part before starting with `// find existing collector with the same ID` that is supposed to find any valid existing collectors. It searches by collectorID, so when we come to creating a new collector we know same ID doesn't exist.
But if e.g. someone managed to insert collector with the same ids in the meantime, it will fail and it is the desired outcome.

npinaeva:
Update code and docs to generate different sample IDs based on acl match+action

go-controller/observability-lib/ovsdb/model.go
dceara: OK, this works for me.

go-controller/observability-lib/sampledecoder/sample_decoder.go
dceara: Ah, ok, I totally missed that.

go-controller/observability-lib/parse_sample.go
dceara: Ack, thanks.

go-controller/pkg/observability/observability.go
```diff
@@ -0,0 +1,294 @@
...
+		m.takenCollectorIDs.Insert(collector.ID)
+		// all collectors are unused, until we update existing configs
+		m.unusedCollectors[collectorKey] = collector.ID
+	}
+	return nil
+}
+
+// Stale collectors can't be deleted until all referencing Samples are deleted.
+// Samples will be deleted asynchronously by different controllers on their init with the new Manager.
+// deleteStaleCollectorsWithRetry will retry, considering deletion should eventually succeed when all controllers
```
dceara, line 130: What if deletion never succeeds (e.g., due to a bug somewhere)?  Can we add a hard retry limit?  Would that be useful?

docs/observability/ovn-observability.md
dceara: This actually changed (3 days ago).  There's no an OVS 3.4 build available in fedora:rawhide and it will be installed with the changes from https://github.com/ovn-org/ovn-kubernetes/pull/4658

go-controller/observability-lib/sampledecoder/sample_decoder.go
dceara: Ack, thanks.

go-controller/pkg/observability/observability.go
dceara: Thanks! :)

go-controller/pkg/observability/observability.go
dceara: OK, I think I would've added a libovsdb.DeleteAllSamplingApps() wrapper but if the current code is acceptable that's fine by me too.

go-controller/pkg/observability/observability.go
dceara: Thanks for adding tests!

docs/observability/ovn-observability.md
```diff
@@ -0,0 +1,122 @@
...
+To see the samples, use `ovnkube-observ` binary, use `-h` to see allowed flags.
+
+This feature requires OVS 3.4 and linux kernel 6.11.
+
+As of Aug 2024, the kernel and OVS 3.4 need to be built from the source, therefore to try this feature you need to:
+- rebuild the kernel with the current master branch from [Linus' tree](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git)
+  - to rebuild on fedora: https://docs.fedoraproject.org/en-US/quick-docs/kernel-build-custom/#_building_a_vanilla_upstream_kernel
+- Build an ovn-kubernetes image that uses the latest OVS/OVN code:
+`OVS_BRANCH=main make -C dist/images fedora-dev`
+- Start kind with that image, use `-ov localhost/ovn-kube-fedora-dev:latest` flag with `kind.sh` script.
```
dceara, line 49: Unfortunately things have changed since the last review and now Fedora:rawhide has OVS 3.4 builds available.  When #4658 merges we'll get OSV 3.4 for free.  Let's change this paragraph to only mention the kernel.

go-controller/pkg/observability/observability.go
npinaeva: Considering this will retry every minute, that probably is not too bad to retry indefinitely, but I can add retry limit.
I have seen very long initial syncs up to 20 min for sure (that should be completed before this cleanup can succeed), so setting retry limit to 1 hour to be safe.

npinaeva:
first push - comments, second - rebase.
Dropped dockerfile changes

dceara (approved):
Looks good to me, thanks!

trozet (approved):

PR #4597, opened 2024-08-08
Set gomega poll interval on retry to allow ~5 retries instead of just 1.

npinaeva:
Fixes [e2e (control-plane, HA, shared, ipv4, snatGW, 1br, ic-disabled)](https://github.com/ovn-org/ovn-kubernetes/actions/runs/10290466858/job/28482170103?pr=4596#logs)
Default are set to `5min, 2s` by k8s https://github.com/kubernetes/kubernetes/blob/60c4c2b2521fb454ce69dee737e3eb91a25e0535/test/e2e/framework/timeouts.go#L22-L23
with this comment https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/test_context.go#L503-L506. So when you set timeout to `1s`, poll interval is still `2s`, which gives 1 retry.
There is a chance other tests also don't expect that.

### DIFF
--- a/test/e2e/status_manager.go
+++ b/test/e2e/status_manager.go
@@ -113,8 +113,8 @@ func checkEgressFirewallStatus(namespace string, empty bool, success bool, event
 		return empty && output == "" || success && strings.Contains(output, "EgressFirewall Rules applied")
 	}
 	if eventually {
-		gomega.Eventually(checkStatus, 1*time.Second).Should(gomega.BeTrue())
+		gomega.Eventually(checkStatus, 1*time.Second, 100*time.Millisecond).Should(gomega.BeTrue())
 	} else {
-		gomega.Consistently(checkStatus, 1*time.Second).Should(gomega.BeTrue())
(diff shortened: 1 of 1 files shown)


martinkennelly:
`Eventually checks that an assertion *eventually* passes.  Eventually blocks when called and attempts an assertion periodically until it passes or a timeout occurs.  Both the timeout and polling interval are configurable as optional arguments.
The first optional argument is the timeout (which defaults to 1s), the second is the polling interval (which defaults to 10ms).  Both intervals can be specified as time.Duration, parsable duration strings or floats/integers (in which case they are interpreted as seconds).  In addition an optional context.Context can be passed in - Eventually will keep trying until either the timeout expires or the context is cancelled, whichever comes first.
`

"_the second is the polling interval (which defaults to 10ms)_"

Why you think the polling interval defaults to 1s?

npinaeva:
Great you asked, I added a better description.

martinkennelly (approved):

PR #4599, opened 2024-08-08
Use 169.254.0.0/17 as the default masquerade subnet

kyrtapz:
169.254.0.0/16 is a bad example since there is a well known IP that is matched by it: 169.254.169.254. This can cause routing issues when running on clouds that are using the metadata IP.

/assign @trozet


trozet (approved):

trozet:
note i've seen this flake on several PRs now: External Gateway With Admin Policy Based External Route CRs e2e non-vxlan external gateway through a gateway pod Should validate ICMP connectivity to an external gateway's loopback address via a gateway pod [It] ipv4

https://github.com/ovn-org/ovn-kubernetes/actions/runs/10304447777/job/28524338024?pr=4599

PR #4601, opened 2024-08-09
Udn cni unconfigure follow up

qinqon:
Follow up changes after https://github.com/ovn-org/ovn-kubernetes/pull/4568 got merged:
- Add more cmdDel tests
- Remove the getCNIResult stub
- Use ConfigureInterface stub at cmdDel tests
- Document primary UDN relevant methods.

```release-note
NONE
```


qinqon:
Hey @tssurya those tests need to be totally refactored since how this is unconfigured totally changed, I think we can close this, not sure they will add any value, though.

PR #4603, opened 2024-08-09
UT flake: fix race condition for nad controller tests

pperiyasamy:
Synchronizes access to `nm.networks` map to avoid race condition while running the tests.

Fixes https://github.com/ovn-org/ovn-kubernetes/issues/4602.

### DIFF
--- a/go-controller/pkg/network-attach-def-controller/network_manager.go
+++ b/go-controller/pkg/network-attach-def-controller/network_manager.go
@@ -153,6 +153,8 @@ func (nm *networkManagerImpl) sync(network string) error {
 }
 
 func (nm *networkManagerImpl) syncAll() error {
+	nm.Lock()
+	defer nm.Unlock()
 	// as we sync upon start, consider networks that have not been ensured as
 	// stale and clean them up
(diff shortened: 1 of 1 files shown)


jcaamano (approved):

jcaamano:
one known flake: https://github.com/ovn-org/ovn-kubernetes/issues/4144

another unknown flake?
[FAIL] Network Segmentation a user defined primary network created using NetworkAttachmentDefinitions is isolated from the default network [It] with L2 dualstack primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:264

rerunning

@pperiyasamy can you register an issue for new one if an unknown flake

tssurya:
@jcaamano / @pperiyasamy was this only a unit test bug?

jcaamano:
> @jcaamano / @pperiyasamy was this only a unit test bug? meaning syncAll is not called in multiple threaded fashion in real code right?

nope, but it would be somewhat difficult to trigger: a nad update would have to happen in between the nad controller start and the netwrok manager controller start.

PR #4604, opened 2024-08-09
Add subnet overlap check for POD and join subnets in net-attach-def

arghosh93:

This is to validate whether POD and join subnet mentioned in a
net-attach-def with topology "layer2" and "layer3" does not overlap
with ClusterSubnets, ServiceCIDRs, join subnet and masquerade subnet.
It also considers excluded subnets mentioned in a net-attach-def.

Fixes #

```release-note

```

### DIFF
--- a/go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
+++ b/go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
@@ -70,7 +70,7 @@ var _ = ginkgo.Describe("Cluster manager EndpointSlice mirror controller", func(
 					ObjectMeta: metav1.ObjectMeta{
 						Name:        "test-pod",
 						Namespace:   namespaceT.Name,
-						Annotations: map[string]string{util.OvnPodAnnotationName: `{"default":{"mac_address":"0a:58:0a:f4:02:03","ip_address":"10.244.2.3/24","role":"infrastructure-locked"},"testns/l3-network":{"mac_address":"0a:58:0a:80:02:04","ip_address":"10.128.2.4/24","role":"primary"}}`},
+						Annotations: map[string]string{util.OvnPodAnnotationName: `{"default":{"mac_address":"0a:58:0a:f4:02:03","ip_address":"10.244.2.3/24","role":"infrastructure-locked"},"testns/l3-network":{"mac_address":"0a:58:0a:84:02:04","ip_address":"10.132.2.4/24","role":"primary"}}`},
 					},
 					Status: v1.PodStatus{Phase: v1.PodRunning},
 				}
@@ -121,7 +121,7 @@ var _ = ginkgo.Describe("Cluster manager EndpointSlice mirror controller", func(
 
 				_, err := fakeClient.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespaceT.Name).Create(
 					context.TODO(),
-					testing.GenerateNAD("l3-network", "l3-network", namespaceT.Name, types.Layer3Topology, "10.128.2.0/16/24", types.NetworkRolePrimary),
+					testing.GenerateNAD("l3-network", "l3-network", namespaceT.Name, types.Layer3Topology, "10.132.2.0/16/24", types.NetworkRolePrimary),
 					metav1.CreateOptions{})
 				gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
@@ -162,7 +162,7 @@ var _ = ginkgo.Describe("Cluster manager EndpointSlice mirror controller", func(
 				gomega.Expect(mirroredEndpointSlices.Items[0].Endpoints).To(gomega.HaveLen(1))
 				gomega.Expect(mirroredEndpointSlices.Items[0].Endpoints[0].Addresses).To(gomega.HaveLen(1))
 				// check if the Address is set to the primary IP
-				gomega.Expect(mirroredEndpointSlices.Items[0].Endpoints[0].Addresses[0]).To(gomega.BeEquivalentTo("10.128.2.4"))
+				gomega.Expect(mirroredEndpointSlices.Items[0].Endpoints[0].Addresses[0]).To(gomega.BeEquivalentTo("10.132.2.4"))
 
 				return nil
 			}
@@ -181,7 +181,7 @@ var _ = ginkgo.Describe("Cluster manager EndpointSlice mirror controller", func(
 					ObjectMeta: metav1.ObjectMeta{
 						Name:        "test-pod",
 						Namespace:   namespaceT.Name,
-						Annotations: map[string]string{util.OvnPodAnnotationName: `{"default":{"mac_address":"0a:58:0a:f4:02:03","ip_address":"10.244.2.3/24","role":"primary"},"testns/l3-network":{"mac_address":"0a:58:0a:80:02:04","ip_address":"10.128.2.4/24","role":"secondary}}`},
+						Annotations: map[string]string{util.OvnPodAnnotationName: `{"default":{"mac_address":"0a:58:0a:f4:02:03","ip_address":"10.244.2.3/24","role":"primary"},"testns/l3-network":{"mac_address":"0a:58:0a:84:02:04","ip_address":"10.132.2.4/24","role":"secondary}}`},
 					},
 					Status: v1.PodStatus{Phase: v1.PodRunning},
 				}
@@ -229,7 +229,7 @@ var _ = ginkgo.Describe("Cluster manager EndpointSlice mirror controller", func(
 
 				_, err := fakeClient.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespaceT.Name).Create(
 					context.TODO(),
-					testing.GenerateNAD("l3-network", "l3-network", namespaceT.Name, types.Layer3Topology, "10.128.2.0/16/24", types.NetworkRoleSecondary),
+					testing.GenerateNAD("l3-network", "l3-network", namespaceT.Name, types.Layer3Topology, "10.132.2.0/16/24", types.NetworkRoleSecondary),
 					metav1.CreateOptions{})
 				gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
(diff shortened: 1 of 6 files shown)


go-controller/pkg/util/multi_network.go
```diff
@@ -761,6 +761,61 @@ func ValidateNetConf(nadName string, netconf *ovncnitypes.NetConf) error {
 		return fmt.Errorf("the subnet attribute must be defined for layer2 primary user defined networks")
 	}
 
+	if netconf.Topology != types.LocalnetTopology && netconf.Name != types.DefaultNetworkName {
```
pperiyasamy, line 764: should we also check for `netconf.Role == types.NetworkRolePrimary`  before doing `subnetOverlapCheck` ?

go-controller/pkg/util/multi_network.go
```diff
@@ -761,6 +761,61 @@ func ValidateNetConf(nadName string, netconf *ovncnitypes.NetConf) error {
...
+	if ni.ExcludeSubnets() != nil {
+		for i, configSubnet := range allSubnets.Subnets {
+			if IsContainedInAnyCIDR(configSubnet.Subnet, ni.ExcludeSubnets()...) {
+				allSubnets.Subnets = append(allSubnets.Subnets[:i], allSubnets.Subnets[i+1:]...)
+			}
+		}
+	}
+	err = allSubnets.CheckForOverlaps()
+	if err != nil {
+		return err
```
pperiyasamy, line 816: nit, it may be good to wrap the error with network name ?

go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
```diff
@@ -70,7 +70,7 @@ var _ = ginkgo.Describe("Cluster manager EndpointSlice mirror controller", func(
 					ObjectMeta: metav1.ObjectMeta{
 						Name:        "test-pod",
 						Namespace:   namespaceT.Name,
-						Annotations: map[string]string{util.OvnPodAnnotationName: `{"default":{"mac_address":"0a:58:0a:f4:02:03","ip_address":"10.244.2.3/24","role":"infrastructure-locked"},"testns/l3-network":{"mac_address":"0a:58:0a:80:02:04","ip_address":"10.128.2.4/24","role":"primary"}}`},
+						Annotations: map[string]string{util.OvnPodAnnotationName: `{"default":{"mac_address":"0a:58:0a:f4:02:03","ip_address":"10.244.2.3/24","role":"infrastructure-locked"},"testns/l3-network":{"mac_address":"0a:58:0a:84:02:04","ip_address":"10.132.2.4/24","role":"primary"}}`},
```
pperiyasamy, line 73: any particular reason for changing mac and ip ?

go-controller/pkg/util/multi_network.go
```diff
@@ -761,6 +761,61 @@ func ValidateNetConf(nadName string, netconf *ovncnitypes.NetConf) error {
...
+	}
+	for _, subnet := range ni.Subnets() {
+		allSubnets.Append(config.UserDefinedSubnets, subnet.CIDR)
+	}
+
+	for _, subnet := range ni.JoinSubnets() {
+		allSubnets.Append(config.UserDefinedJoinSubnet, subnet)
+	}
+	if ni.ExcludeSubnets() != nil {
```
pperiyasamy, line 807: what is the reason for having this block ? aren't we just good enough by not adding `ni.ExcludeSubnets()` into allSubnets before doing overlap check ?

go-controller/pkg/util/multi_network_test.go
```diff
@@ -845,6 +845,121 @@ func TestGetPodNADToNetworkMapping(t *testing.T) {
...
+	config.Gateway.V4MasqueradeSubnet = "169.254.169.0/29"
+	config.Gateway.V6MasqueradeSubnet = "fd69::/125"
+	config.Gateway.V4JoinSubnet = "100.64.0.0/16"
+	config.Gateway.V6JoinSubnet = "fd98::/64"
+	type testConfig struct {
+		desc                        string
+		inputNetAttachDefConfigSpec string
+		expectedError               error
+	}
+
```
pperiyasamy, line 864: can have one +ve test with no expectedError

go-controller/pkg/util/multi_network.go
arghosh93: https://github.com/openshift/enhancements/pull/1623/files#diff-76f8949dbcefb73686f569b18267b152f0f1ef3d71ce9888046aa93f97566005R292 makes me think that we need to run subnet overlap check regardless of netconf.Role.

go-controller/pkg/util/multi_network.go
arghosh93: Ack, sure.

go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
arghosh93: yes, otherwise subnet overlap check was complaining about an overlap as subnet("10.128.2.0/16/24") of NAD was overlapping config.Default.ClusterSubnets 10.128.0.0/14.

go-controller/pkg/util/multi_network_test.go
arghosh93: I have kepthttps://github.com/ovn-org/ovn-kubernetes/pull/4604/commits/41e41c69c59b89d41d2cfa9d1f5d1147ec68c0f6#diff-9c47311a9d19959866ca074985bfd6523244f714186bb068c910ac22013fdc28R927 , should i add more? I can do that if required.

go-controller/pkg/util/multi_network.go
arghosh93: https://github.com/openshift/enhancements/pull/1623/files#diff-76f8949dbcefb73686f569b18267b152f0f1ef3d71ce9888046aa93f97566005R319 makes me think that we need to run overlap check for join subnet mentioned at a UDN.

pperiyasamy: ok, if this parameter is meant to exclude matching subnets from overlap check, then i'm fine with it.

go-controller/pkg/util/multi_network_test.go
pperiyasamy: ok, that is enough.

pperiyasamy (dismissed):

go-controller/pkg/util/multi_network.go
arghosh93: Yes, that is my intention here. Thanks!

tssurya:
IIUC this is checking for overlap of ALL UDN's (l3/l3)cluster/join subnet to not be overlapping with default network's cluster,join,masquerade,service right? I think that's what we wanted as well, so this works.

will take a deeper look next week

go-controller/pkg/util/multi_network.go
tssurya, line 764: so this means we will do it for UDPNs and UDSNs as well which is good.

tssurya: my only question would be whether this should happen in validate OR parse portions because it seems like we do parseCIDR here already but then perhaps we do it again at ParseNAD but I guess its fine.

go-controller/pkg/util/multi_network.go
tssurya: @maiqueb / @qinqon when validating exclude subnets.. we don't want to check for overlap with used subnets? I thought we should..

tssurya (dismissed):
This PR is LGTM except for let's fix https://github.com/ovn-org/ovn-kubernetes/pull/4604#discussion_r1727236827

go-controller/pkg/util/multi_network.go
tssurya: ah nvm this exclude subnets is a field where the user provides the values that should be excluded from IPAM and usually its actually contained within the main subnet right?

go-controller/pkg/util/multi_network.go
```diff
@@ -761,6 +761,61 @@ func ValidateNetConf(nadName string, netconf *ovncnitypes.NetConf) error {
 		return fmt.Errorf("the subnet attribute must be defined for layer2 primary user defined networks")
 	}
 
+	if netconf.Topology != types.LocalnetTopology && netconf.Name != types.DefaultNetworkName {
+		if err := subnetOverlapCheck(netconf); err != nil {
+			return err
```
tssurya, line 766: expand on this error

go-controller/pkg/util/multi_network.go
```diff
@@ -761,6 +761,61 @@ func ValidateNetConf(nadName string, netconf *ovncnitypes.NetConf) error {
...
+
+	_, v4MasqueradeCIDR, _ := net.ParseCIDR(config.Gateway.V4MasqueradeSubnet)
+	_, v6MasqueradeCIDR, _ := net.ParseCIDR(config.Gateway.V6MasqueradeSubnet)
+
+	allSubnets.Append(config.ConfigSubnetMasquerade, v4MasqueradeCIDR)
+	allSubnets.Append(config.ConfigSubnetMasquerade, v6MasqueradeCIDR)
+
+	ni, err := NewNetInfo(netconf)
+	if err != nil {
+		return err
```
tssurya, line 798: expand on this error

go-controller/pkg/util/multi_network.go
maiqueb: Yes, the overlap is expected. We want to have a way for users to express they want to ensure some IPs are never addressable within their subnets.

arghosh93: > ah nvm this exclude subnets is a field where the user provides the values that should be excluded from IPAM and usually its actually contained within the main subnet right? so the overlap is expected?

Yes, overlap with either POD subnet or join subnet in a NAD is expected.

go-controller/pkg/util/multi_network.go
arghosh93: Done

go-controller/pkg/util/multi_network_test.go
```diff
@@ -845,6 +845,121 @@ func TestGetPodNADToNetworkMapping(t *testing.T) {
 	}
 }
 
+func TestSubnetOverlapCheck(t *testing.T) {
```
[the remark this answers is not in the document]
arghosh93: Thanks Surya :)

tssurya (approved):

PR #4608, opened 2024-08-11
WIP: Add OwnerReferencesPermissionEnforcement

oshoval:

Allows finding stricer RBAC admission enforcement that are
used on some k8s distributions.

Fixes #

```release-note

```

### DIFF
--- a/contrib/kind.yaml.j2
+++ b/contrib/kind.yaml.j2
@@ -28,6 +28,7 @@ kubeadmConfigPatches:
   apiServer:
     extraArgs:
       "v": "{{ cluster_log_level }}"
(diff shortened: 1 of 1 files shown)


oshoval:
tested locally, the feature does work, it blocks some cases of missing rbacs, hence some tests are failing here
we need to first add the missing rbacs before adding the feature,
meanwhile just using it to test something that happens only on OpenShift related to IPAM feature
once all the lanes finish, will write down which failed and will close the PR

oshoval:
lots of lane fail, not sure why is related and whats not, i cant handle it atm
so would just apply manually when i need to test this specific admission on kind provider for now

PR #4610, opened 2024-08-12
udn, tests: expect cluster global values reset to succeed

maiqueb:

This PR changes the `network_segmentation` unit tests to expect the cluster global values reset to succeed; previously, any hypothetical errors raising during this reset were simply ignored. The next tests would then run using an undesired cluster state, which would lead to unexpected results.

Fixes #

```release-note

```


maiqueb:
Somebody did this at a point in time. Throwing it away.

PR #4612, opened 2024-08-12
Network API: Add CRD for managing cluster scope networks

ormergi:

This PR implements the CRDs for Managing networks  part of the enhancement https://github.com/openshift/enhancements/pull/1623 
Following https://github.com/openshift/enhancements/pull/1638

This PR adds an API (CRD) for managing cluster scope OVN network.

Introducing the ClusterUserDefinedNetwork CRD for representing shared OVN network request on across multiple namespaces and corresponding controller.

The controller create `NetworkAttachmentDefintion` according to the CR in each namespace according to spec.
Having the exact same NAD reside on each desired namespace enable workloads connect to the same network.

<details>
  <summary><b> How to use (click me) </b></summary>

Create cluster-user-defined-network request:
```
apiVersion: k8s.ovn.org/v1
metadata:
  name: net0
spec:
  namespaceSelector:
    matchExpressions:
      - key: "kubernetes.io/metadata.name"
        operator: "In"
        values: 
        - mynamesapce
        - theirnamespace
  network:
    topology: Layer2
    layer2:
      role: Secondary
      subnets: [10.10.0.0/24]
``` 
The controller will create the following NAD in the desired namespaces:
```
metadata:
  name: net0
  namespace: mynamespace
  finalizers:
  - k8s.ovn.org/user-defined-network-protection
  labels:
    k8s.ovn.org/user-defined-network: ""
  ownerReferences:
  - apiVersion: k8s.ovn.org/v1
    blockOwnerDeletion: true
    name: net0
    uid: 97b1da95-d873-44f5-af18-0e6e7a830bd0
spec:
  config:  > 
    '{
        "cniVersion":"1.0.0",
        "type":"ovn-k8s-cni-overlay",
        "name":"cluster.udn.net0",
        "netAttachDefName":"theirnamespace/net0",
        "role":"primary",
        "topology":"layer2",
        "joinSubnets":"100.65.0.0/16,fd99::/64",
        "subnets": "10.10.0.0/24"
      }'
---
metadata:
  name: net0
  namespace: theirnamespace
  finalizers:
  - k8s.ovn.org/user-defined-network-protection
  labels:
    k8s.ovn.org/user-defined-network: ""
  ownerReferences:
  - apiVersion: k8s.ovn.org/v1
    blockOwnerDeletion: true
    name: net0
    uid: 97b1da95-d873-44f5-af18-0e6e7a830bd0
spec:
  config:  > 
    '{
        "cniVersion":"1.0.0",
        "type":"ovn-k8s-cni-overlay",
        "name":"cluster.udn.net0",
        "netAttachDefName":"theirnamespace/net0",
        "role":"primary",
        "topology":"layer2",
        "joinSubnets":"100.65.0.0/16,fd99::/64",
        "subnets": "10.10.0.0/24"
      }'
```

The CRD is updated as follows:
```
apiVersion: k8s.ovn.org/v1
metadata:
  name: net0
spec:
  namespaceSelector:
    matchExpressions:
      - key: "kubernetes.io/metadata.name"
        operator: "In"
        values: 
        - mynamesapce
        - theirnamespace
  network:
    topology: Layer2
    layer2:
      role: Secondary
      subnets: [10.10.0.0/24]
status:
  conditions:
  - message: NetworkAttachmentDefinition has been created in namespaces: [mynamesapce, theirnamespace]
    reason: NetworkAttachmentDefinitionReady
    status: "True"
    type: NetworkReady
```

In case there is already primary network at one of the target namespace, the CRD is updated as follows:
```
apiVersion: k8s.ovn.org/v1
metadata:
  name: net0
spec:
  namespaceSelector:
    matchExpressions:
      - key: "kubernetes.io/metadata.name"
        operator: "In"
        values: 
        - mynamesapce
        - theirnamespace
  template:
    spec:    
      topology: Layer2
      layer2:
        role: Secondary
        subnets: [10.10.0.0/24]
status:
  conditions:
  - type: NetworkReady
    status: "False"    
    reason: SyncError
    message: 'primary network already exist in namespace "theirnamespace": "net0"'
```

</details>

Fixes #

Controller behavior:
* Similar to the UDN controller, each NAD is created with the same `meta.namesapce`, at each desired namespace that aplies to the specified namespace selector.
* The underlying network-name, specified in the NAD netconf, is composed from the CR meta name and a prefix, for example: given CR name is `mynetwork`, the underlying network-name will be: `cluster.udn.mynetwork`
* The controller label the NADs it creates with label `k8s.ovn.org/user-defined-network` to enable easier management (e.g.: listing all Cluster-UDN related NAD) and integration (e.g.: act on Cluster-UDN related NAD changes).
* Finalizes are added to the CR and NAD objects in order enable the controller verifying the network is not being used before deletion. And avoid incomplete state when Cluster-UDN is deleted  before the NAD objects deleted.

This PR also:
- Export code from namespace UDN controller to enable reuse for the Cluster UDN controller.
- Changes namespace UDN controller deletion flow to delete NADs explicitly, in order to provider consistent and predictable behavior. And also better fit the Cluster UDN use-case.

Follow-up implementation phases:
- This CRD status is currently reflect the state of the NAD object, if follow up PR the status will be extended to reflect the state of the underlying OVN network as well.
- In a scenario primary UDN exist at a target namesapce, and a new primary NAD is created, Cluster-UDN CR status should reflect a conflicting primary NAD exist.
- Add CRD and RBAC to CNO.
- https://github.com/ovn-org/ovn-kubernetes/pull/4612#issuecomment-2358158166
- Enable subsystem condition reporting for CUDN types. 
  https://github.com/ovn-org/ovn-kubernetes/pull/4612#issuecomment-2432822314

```release-note
Introduce API for managing cluster scope user-defined networks, targeted for cluster-admins.
Enable cluster-admins create shared user-defined-networks across multiple namespaces.
```


npinaeva (commented):
first look at the crd

go-controller/pkg/crd/clusteruserdefinednetwork/v1/types.go
```diff
@@ -0,0 +1,86 @@
...
+type ClusterUserDefinedNetworkSpec struct {
+	// NamespaceSelector Label selector for which namespace network should be available for.
+	// +kubebuilder:validation:Required
+	// +required
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector"`
+
+	// Template is direct specification of UserDefinedNetwork.
+	// +kubebuilder:validation:Required
+	// +required
+	Template UserDefinedNetworkTemplateSpec `json:"template"`
```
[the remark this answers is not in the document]
ormergi: Having the "template" layer enable extension for other fields then spec, for example: annotation/labels etc..
I followed other k8s object example such as the Demployment,DaemonSet, they have a template and the actual spec inside it.

WDYT?

go-controller/pkg/crd/clusteruserdefinednetwork/v1/types.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// UserDefinedNetworkTemplateSpec UserDefinedNetwork spec template.
+type UserDefinedNetworkTemplateSpec struct {
+	// UserDefinedNetworkSpec contains the UserDefinedNetwork specification.
+	Spec udnv1.UserDefinedNetworkSpec `json:"spec,omitempty"`
```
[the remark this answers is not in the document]
ormergi: You right, the Cluster UDN CRD should have its own spec as it may need to diverge form the UDN CRD spec.
I will change it accordingly.
 
Regarding localnet, yes we did it fit better the cluster UDN CRD as it targeted for cluster admin only.
Although it seems exposing localnet require more thinking as it may not fit with future changed for localnet implementation such as https://github.com/ovn-org/ovn-kubernetes/pull/4320
I would rather add support for localnet on a follow up PR, can we do that?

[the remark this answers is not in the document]
maiqueb: > I would rather add support for localnet on a follow up PR, can we do that?

I think it's fine, since that's not in scope for primary UDN. Whatever's not about primary UDN can be left for another release.

go-controller/pkg/crd/clusteruserdefinednetwork/v1/types.go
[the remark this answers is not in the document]
maiqueb: Given @ormergi 's answer, I'd go with template.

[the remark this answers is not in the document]
npinaeva: so the difference is that usually a template contains the exact spec of the object that will be created, e.g. daemonset and deployment template is a pod spec. When you see cluster UDN template, it gives a feeling that namespaced UDNs will be created from that template, which is not the case. So unless we can put a spec of an existing CRD to the "template" section, I don't think it follows best practices.
Also, namespace and cluster scoped UDNs are not that different in this sense, as both create NADs, so if more NAD-realted functionality is needed, it will likely be done for both namespaced and cluster scoped UDNs.

ormergi:
Rebased

go-controller/pkg/crd/clusteruserdefinednetwork/v1/types.go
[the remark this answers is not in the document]
maiqueb: Could this `template` attribute be a leftover from when we wanted this cluster wide UDN to template individual namespace UDNs ? 

If the motivation for having this attribute is **only** to pass annotations / labels, I think it's not good enough (motivation). We can simply carry-over those from the cluster wide UDN to the NAD, and not have to expose that in the API.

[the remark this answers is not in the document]
npinaeva: after a team meeting discussion, we agreed that template is not required if we don't want to create namespace copies of UDN (which I think we agreed not to do).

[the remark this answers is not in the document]
ormergi: Done

go-controller/pkg/crd/clusteruserdefinednetwork/v1/types.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// NamespaceNetworkStatus describe a user-defined-network status in a namespace who is part of a cluster
+// user-defined-network shared network.
+type NamespaceNetworkStatus struct {
+	// Name is the target namespace name
+	Name string `json:"name"`
+
+	// ActiveSinceTime indicate since when the network is available.
+	ActiveSinceTime metav1.Time `json:"activeSinceTime,omitempty"`
```
npinaeva, line 86: Is there a reason to have a special struct to report per-namespace status instead of using per-namespace condition in the status.conditions? I see you mentioned some discussion https://github.com/openshift/enhancements/pull/1638/files#r1712960977 maybe you can point me to that?

npinaeva (commented):
CRD-related comments

go-controller/pkg/crd/clusteruserdefinednetwork/v1/types.go
ormergi: Currently, only the namespace name is required.
Following feedback on the suggested API and discussion offline it was concluded to avoid using bare strings in order to enable the API evolve and reflect more details.

go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go
```diff
@@ -0,0 +1,64 @@
...
+
+// NamespaceNotifier watches Namespaces objects and notify subscribers upon change.
+// It enqueues the reconciled object keys in the subscribing controllers workqueue.
+type NamespaceNotifier struct {
+	Controller controller.Controller
+
+	subscribers []NamespaceReconciler
+}
+
+func NewNamespaceNotifier(nsInformer corev1informer.NamespaceInformer, subscribers ...NamespaceReconciler) *NamespaceNotifier {
```
npinaeva, line 25: we know there will only be 1 user of this notifier ever, right?
If so, can we just merge it into the main controller, then it will be much easier to link the namespace controller to the real reconcile function without extra interfaces in between?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -169,8 +171,32 @@ func (c *Controller) ReconcileNetAttachDef(key string) error {
 
 // ReconcileNamespace enqueue relevant Cluster UDN CR requests following namespace events.
 func (c *Controller) ReconcileNamespace(key string) error {
-	// TODO: reconcile affected namespaced only.
-	c.cudnController.Reconcile(key)
+	c.namespaceTrackerLock.Lock()
```
npinaeva, line 174: should we use rlock here considering namespace updates may be more often than udn updates?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -335,12 +342,192 @@ func (c *Controller) reconcileCUDN(key string) error {
...
+	selectedNamespaces, err := c.getSelectedNamespaces(cudn.Spec.NamespaceSelector)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get selected namespaces: %w", err)
+	}
+	affectedNamespaces := c.namespaceTracker[cudn.Name]
+
+	if !cudn.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(cudn, template.FinalizerUserDefinedNetwork) {
+			var errs []error
+			for nsToDelete := range affectedNamespaces.Union(selectedNamespaces) {
```
npinaeva, line 388: I don't think we need to account for selectedNamespaces here, since all namespace that had NADs created should be in the affectedNamespaces, and even if new namespaces are selected now, we have nothing to clean up for them?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -335,12 +342,192 @@ func (c *Controller) reconcileCUDN(key string) error {
...
+				if err := c.deleteNAD(cudn, nsToDelete); err != nil {
+					errs = append(errs, fmt.Errorf("failed to delete NetworkAttachmentDefinition: %w", err))
+				}
+				c.namespaceTracker[cudn.Name].Delete(nsToDelete)
```
npinaeva, line 392: should we only do that if NAD deletion succeeded?

npinaeva (commented):
Perfect! Beautiful! A treat for my eyes! 
I know it still needs some work to be done, but I like where it is going :)

ormergi:
[Changes:](https://github.com/ovn-org/ovn-kubernetes/compare/e0ea9d85dc3fd1e18f71d2a8825bba0aec09f3b9..a63542f9c801c655891c00f1903a8a16cdb4bf0e): rebased

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
```diff
@@ -3,41 +3,93 @@ package template
...
+	var spec specGetter
+	switch o := obj.(type) {
+	case *userdefinednetworkv1.UserDefinedNetwork:
+		ownerRef = *metav1.NewControllerRef(obj, userdefinednetworkv1.SchemeGroupVersion.WithKind("UserDefinedNetwork"))
+		spec = &o.Spec
+		networkName = obj.GetNamespace() + "." + obj.GetName()
+	case *cudnv1.ClusterUserDefinedNetwork:
+		ownerRef = *metav1.NewControllerRef(obj, cudnv1.SchemeGroupVersion.WithKind("ClusterUserDefinedNetwork"))
+		spec = &o.Spec.Network
+		const cudnNetworkNamePrefix = "cluster.udn"
```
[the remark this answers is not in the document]
ormergi: Done

go-controller/pkg/clustermanager/userdefinednetwork/controller/validations/nad.go
```diff
@@ -0,0 +1,72 @@
+package validations
```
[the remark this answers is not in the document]
ormergi: Done

go-controller/pkg/clustermanager/userdefinednetwork/controller/validations/nad.go
```diff
@@ -0,0 +1,72 @@
...
+			connectedPods = append(connectedPods, pod.Namespace+"/"+pod.Name)
+		}
+	}
+	if len(connectedPods) > 0 {
+		return fmt.Errorf("network in use by the following pods: %v", connectedPods)
+	}
+	return nil
+}
+
+type specGetter interface {
```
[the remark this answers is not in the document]
ormergi: Done

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -0,0 +1,99 @@
...
+		// any other thread that create NADs.
+		// Since the UserDefinedNetwork controller use single thread (threadiness=1),
+		// and being the only controller that create NADs, this conditions is fulfilled.
+		if validations.PrimaryNetworkSpec(template.GetSpec(obj)) {
+			actualNads, err := c.nadLister.NetworkAttachmentDefinitions(obj.GetNamespace()).List(labels.Everything())
+			if err != nil {
+				return nil, fmt.Errorf("failed to list  NetworkAttachmetDefinition: %w", err)
+			}
+			// This is best effort check no pod use the subject NAD, noting prevent pod creation right after this check.
+			// TODO: introduce webhook to block primary UDN NAD creation in target namespace, when primary UDN exist but not ready.
```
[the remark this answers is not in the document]
ormergi: There is a mistake in this comment, so sorry for the confusion.
 I mean to say noting prevents from primary NAD to be created right after this check and before the controller actually create the NAD.
WDYT?

Commit is fixed, TODO is removed (it shouldn't be tracked in a code comment :sweat_smile: )

go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go
ormergi: I followed the same pattern as with the NAD notifier, because it seem that having them in separate package keep the boilerplate out of the controller business logic in controller.go, and reduce so much noise.
WDYT?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Initially I tried to go on the safe side and include the selected-namespaces, just in case.
Looking at this again it make sense to go over the affected namespaces only.
Thanks for the heads up :slightly_smiling_face:

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Fixed, thanks!

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Thanks for the heads up, done.

ormergi (commented):
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/a63542f9c801c655891c00f1903a8a16cdb4bf0e..c1eb5782a564039d4e070a2fecd933ac7a9a9e4a): fix https://github.com/ovn-org/ovn-kubernetes/pull/4612#pullrequestreview-2289440996 https://github.com/ovn-org/ovn-kubernetes/pull/4612#pullrequestreview-2303229715 comments

In addition:
- Add RBAC permissions commit is squashed, its changes were added to relevant commits.
- Add namespace informer commit is squashed to CUDN informer setup.
- Following changes to CRD files, setup CUDN clientset commit is dropped as it no longer needed.
- Reworked commit messages.
- pod2egress e2e coverage is added.

ormergi:
Changes https://github.com/ovn-org/ovn-kubernetes/compare/c1eb5782a564039d4e070a2fecd933ac7a9a9e4a..8b0802647b6f45d1ba5e99ad90a414b4ee817292 https://github.com/ovn-org/ovn-kubernetes/compare/8b0802647b6f45d1ba5e99ad90a414b4ee817292..c5416baedd26cee73ad036cf33a6d04c241e2207: fixed misplaced changes across commits.

go-controller/pkg/crd/userdefinednetwork/v1/types.go
```diff
@@ -229,3 +284,35 @@ type UserDefinedNetworkStatus struct {
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
+
+// ClusterUserDefinedNetworkStatus contains the observed status of the ClusterUserDefinedNetwork.
+type ClusterUserDefinedNetworkStatus struct {
+	// ActiveNamespaces indicates the network state in namespaces it available at.
+	// +optional
+	ActiveNamespaces []NamespaceNetworkStatus `json:"activeNamespaces,omitempty"`
```
npinaeva, line 292: After some discussion, we came up with a status reporting framework described in https://github.com/ovn-org/ovn-kubernetes/pull/4647
For this CRD it means, we can still use only conditions, but report "ActiveNamespaces" via one of the conditions.
The idea is for every condition type, (that probably will be the same for UDN and CUDN like NetworkAttachmentDefinitionReady and NetworkReady in the future) we can report a list of namespaces that were selected together with "Success" status, but if at least one of the namespaces failed, we could report a generic message saying "Network configuration failed for one or more namespaces, check UDN events for more info.", and then report per-namespace error via UDN events.

npinaeva: also, looking at conditions themselves, I see it was suggested by @trozet that "NetworkCreated == the network topology has been created in OVN. NetworkReady== the NAD has been created." For me it makes more sense vise versa: NetworkCreated == the NAD has been created NetworkReady== the network topology has been created in OVN. wdyt?

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/c5416baedd26cee73ad036cf33a6d04c241e2207..0a4bc126d4914c9fcf0d5b07c129044b31cbd017): fix NAD deletion bug

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Fixed by last push https://github.com/ovn-org/ovn-kubernetes/pull/4612#issuecomment-2353034446

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -358,3 +324,232 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
+	if err != nil {
+		return nil, fmt.Errorf("faild to list namespaces: %w", err)
+	}
+	for _, selectedNs := range selectedNamespacesList {
+		selectedNamespaces.Insert(selectedNs.Name)
+	}
+	return selectedNamespaces, nil
+}
+
+func (c *Controller) updateClusterUDNStatus(cudn *userdefinednetworkv1.ClusterUserDefinedNetwork, nads []netv1.NetworkAttachmentDefinition, syncError error) error {
```
npinaeva, line 457: with cUDN, I have got an idea:
we could just create a special error type for "NetworkAttachmentDefinitionDeleted" reason, and check `nad.DeletionTimestamp` inside `syncUserDefinedNetwork` function instead of doing that in the `updateStatus` function. Then we won't need to pass nads around, but just use errors list, wdyt?

go-controller/pkg/crd/userdefinednetwork/v1/types.go
```diff
@@ -64,6 +64,61 @@ type UserDefinedNetworkSpec struct {
 	Layer2 *Layer2Config `json:"layer2,omitempty"`
 }
 
+// ClusterUserDefinedNetwork describe network request for a shared network across namespaces.
```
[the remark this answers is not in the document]
ormergi: Done

go-controller/pkg/crd/userdefinednetwork/v1/types.go
ormergi: ActiveNamespaces are removed form status, in follow up commit status reporting logic is changed to reflect active-namespaces through condition.

ormergi (commented):
Changes: resolve review comments 

In addition: 
- Enqeue NAD events to the right controller workqueue

npinaeva:
What we can postpone to follow-up:
- Change condition type from "NetworkReady" to "NetworkCreated"
- Do not use NADs to build condition, only use errors
- Report events per namespace error

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
npinaeva: added to follow-up list

ormergi (commented):
Changes: resolve review comments

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/a7c3cf1cd2caa038038de730f3351091defb21bc..5fb6c1f466909f22ea89ef1f1cf42a249a769c3a): 
- Misplaced changes moved to the right commit.
- E2e helper function names
- Make use of ExpectWithOffset

[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/5fb6c1f466909f22ea89ef1f1cf42a249a769c3a..a9572e92fece8e052a4ba6d5bbfefe376f0e9dfa): fix race in namespace notifier tests

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -31,34 +33,46 @@ import (
...
 		nadClient = netv1fakeclientset.NewSimpleClientset()
 		nadInformer = netv1informerfactory.NewSharedInformerFactory(nadClient, 15).K8sCniCncfIo().V1().NetworkAttachmentDefinitions()
 
 		kubeClient = fake.NewSimpleClientset()
 		sharedInformer := informerfactory.NewSharedInformerFactoryWithOptions(kubeClient, 15)
 		podInformer = sharedInformer.Core().V1().Pods()
+		namespaceInformer = sharedInformer.Core().V1().Namespaces()
 	})
 
+	newTestController := func(renderNADStub RenderNetAttachDefManifest) *Controller {
```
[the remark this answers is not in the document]
npinaeva: and we need a test for this, that will create a NAD owned by CUDN in a namespace that is not matched by CUDN anymore, then start controller and make sure that stale NAD is deleted

ormergi (commented):
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/a9572e92fece8e052a4ba6d5bbfefe376f0e9dfa..33a1e4b3116ba6ca1f1f2e60d3098ecb0e345f24): resolve https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1766686027

In addition, add unit test to cover CUDN controller reconcile mutated NADs
This test also covers the ReconcileNetAttachDef implementation, for CUDN NAD case.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -63,61 +83,163 @@ func New(
...
+
+	for cudnName, affectedNamespaces := range c.namespaceTracker {
+		affectedNamespace = affectedNamespaces.Has(key)
+
+		if !affectedNamespace {
+			cudn, err := c.cudnLister.Get(cudnName)
+			if err != nil {
+				return fmt.Errorf("faild to get CUDN %q from cache: %w", cudnName, err)
+			}
+			namespace, err := c.namespaceInformer.Lister().Get(key)
```
npinaeva, line 214: you can do that outside the CUDN loop, then it will only be 1 call

test/e2e/network_segmentation.go
```diff
@@ -905,6 +1219,117 @@ func normalizeConditions(conditions []metav1.Condition) []metav1.Condition {
...
+apiVersion: k8s.ovn.org/v1
+kind: ClusterUserDefinedNetwork
+metadata:
+  name: ` + name + `
+spec:
+  namespaceSelector:
+    matchExpressions:
+    - key: kubernetes.io/metadata.name
+      operator: In
+      values: [ ` + targetNs + ` ]
```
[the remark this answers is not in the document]
ormergi: I tried to switch to having common label for all but it doesnt seem to it easier.

In most tests there  is no need to update both CUDN and namespace to  select new namespace, it require changing CUDN because test use the  `kubernetes.io/metadata.name` label.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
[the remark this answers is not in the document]
ormergi: Some tests inject errors using fake clients PrependReactor, since util.GetOVNClientset does not expose the Fake entitiy its impossible to inject error using it.
Unfortunalty using util.GetOVNClientset in all test would not work

> and we need a test for this, that will create a NAD owned by CUDN in a namespace that is not matched by CUDN anymore, then start controller and make sure that stale NAD is deleted

Done
Covered by `should delete CUDN NAD in namespaces that no longer applies to namespace-selector` test

ormergi:
Changes: resolve https://github.com/ovn-org/ovn-kubernetes/pull/4612#pullrequestreview-2318189330
In addition: 
- Sort the reported namespace name in network-ready condition, to avoid redundant status updated due to inconsistent order in the condition message.
- Remove redundant namespace informer setup, when ovnkube-controller is used.
- Simplify unit-test using helper for creating CR

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Done

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/ab3d62e2380c85b80ef4dd1a9d3b807401fcc82e..68bf0b20728e8901f190e37f47fb9a2eceed83fb): resolve https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1766878624 
- Introduce test method to setup test controller, informer and clientsets
  Make use of util.GetOVNClientset to create clientsets for tests, as in other test suites
  Existing namespace UDN tests setup changed to make use of the new test method
  Test logic did not change, except for one test `when UDN is being deleted, NAD exist, 2 pods using UDN, should delete NAD once no pod uses the network`
  the test UDN CR is created first and then marked for deletion using patch.
  Cluster UDN tests were changed to make use of the test helper as well.
- Cluster UDN CRD to their  namespace UDN CRD controllers tests, each group, to new Context.
- Simplify cluster UDN tests that asserts status for active namespaces.

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/ab3d62e2380c85b80ef4dd1a9d3b807401fcc82e..7e186f0cd7b8cb9a91288d0d77f9b5a5a039311a): resolve https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1766878624 
- Introduce test method to setup test controller, informer and clientsets
  Make use of util.GetOVNClientset to create clientsets for tests, as in other test suites
  Existing namespace UDN tests setup changed to make use of the new test method
  Test logic did not change, except for one test `when UDN is being deleted, NAD exist, 2 pods using UDN, should delete NAD once no pod uses the network`
  the test UDN CR is created first and then marked for deletion using patch.
  Cluster UDN tests were changed to make use of the test helper as well.
- Cluster UDN CRD to their  namespace UDN CRD controllers tests, each group, to new Context.
- Simplify cluster UDN tests that asserts status for active namespaces.
- Add coverage for cluster UDN CRD, stale NADs deletion, following spec change or namespace event.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -17,535 +19,854 @@ import (
...
-		nadClient   *netv1fakeclientset.Clientset
-		udnInformer udninformer.UserDefinedNetworkInformer
-		nadInformer netv1Informer.NetworkAttachmentDefinitionInformer
-		kubeClient  *fake.Clientset
-		podInformer corev1informer.PodInformer
+		cs                *util.OVNClusterManagerClientset
+		udnInformer       udninformer.UserDefinedNetworkInformer
+		nadInformer       netv1Informer.NetworkAttachmentDefinitionInformer
+		podInformer       corev1informer.PodInformer
+		namespaceInformer corev1informer.NamespaceInformer
```
npinaeva, line 50: Using indexers from informers to make changes is a bit strange, you could instead use create/update/delete calls, e.g. for NAD you can do
`cs.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespace).Create(...)` same for all other objects, their clients are already present in the cs.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -17,535 +19,854 @@ import (
...
-		udnClient = udnfakeclient.NewSimpleClientset()
-		udnInformer = udninformerfactory.NewSharedInformerFactory(udnClient, 15).K8s().V1().UserDefinedNetworks()
-		nadClient = netv1fakeclientset.NewSimpleClientset()
-		nadInformer = netv1informerfactory.NewSharedInformerFactory(nadClient, 15).K8sCniCncfIo().V1().NetworkAttachmentDefinitions()
+	newTestController := func(renderNADStub RenderNetAttachDefManifest, objects ...runtime.Object) *Controller {
+		cs = util.GetOVNClientset(objects...).GetClusterManagerClientset()
 
-		kubeClient = fake.NewSimpleClientset()
-		sharedInformer := informerfactory.NewSharedInformerFactoryWithOptions(kubeClient, 15)
+		udnFact := udninformerfactory.NewSharedInformerFactory(cs.UserDefinedNetworkClient, 15)
```
npinaeva, line 56: the idea was to have only 1 newTestController function, and only create 1 watch factory with `factory.NewClusterManagerWatchFactory`. If you don't want to start watch factory every time, you can just make another function like `startNewTestController` that will call `newTestController` and start watch factory

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -17,535 +19,854 @@ import (
...
+						nads, err := cs.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(nsName).List(context.Background(), metav1.ListOptions{})
+						Expect(err).NotTo(HaveOccurred())
+						Expect(nads.Items).To(Equal([]netv1.NetworkAttachmentDefinition{*testClusterUdnNAD(cudnName, nsName)}),
+							"NAD should exist in existing and new labeled namespaces")
+					}
+				})
+
+				It("when CR selector has selection removed, should delete stale NADs in previously matching namespaces", func() {
+					By("remove test label value from namespace-selector")
+					p := fmt.Sprintf(`[{"op": "replace", "path": "./spec/namespaceSelector/matchExpressions/0/values", "value": [""]}]`)
```
npinaeva, line 551: this one is a bit hard to follow, can you just use golang struct and call Update instead of patch?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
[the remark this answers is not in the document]
npinaeva: > We should build namespaceTracker on controller startup with existing objects, so that we can do a cleanup if some NADs are stale now. To do so, you can use controller.StartWithInitialSync function.

this one is still not fixed. Test scenario should cover namespace label change while ovn-k was "off". So on startup you will see namespace and its NAD, but namespace should not match CUDN selector anymore, and its NAD should be removed.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Unfortunately, it doesnt seem to work well when tests relay on watch-factory, on NAD create/update/delete it seem the WF informers take some time to sync and have the actual new object state causing some tests to flake and other to fail.
I managed to work-around it by setting 50ms sleep right after NAD create/update/delete operation, but its nasty.
Using the WF informers directly in tests to simulate create/update/delete NAD does work though.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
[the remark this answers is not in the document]
ormergi: In order to support this scenario its necessary to implement some initial-sync logic for the namespace-tracker, so when the ovn-k starts it will populate the namespace-tacker with all managed NADs and reconcile them.
As a result, it will delete NADs in namespaces that no longer match the spec.

Should we do that in a follow-up PR?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Done

Now there is single function to start the tests controller.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Done

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/7e186f0cd7b8cb9a91288d0d77f9b5a5a039311a..9bb1f4bb14a8d5a2568e27dcf31d687d83fd9258): resolve  https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1781212718 https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1781215759 https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1781222357

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
npinaeva: I think all tests that manually call sync function have setup + sync + check result pattern. In this case, the easiest way to setup all informers with the right objects is to pass all udns/nads/etc to the `newTestController`, which will make sure all objects are propagated during watch factory start. Then, for example, test case "should fail when NAD owner-reference is malformed" may look like
```
                        udn := testUDN()
			mutatedNAD := testNAD()
			mutatedNAD.ObjectMeta.OwnerReferences = []metav1.OwnerReference{{Kind: "DifferentKind"}}

			c := newTestController(noopRenderNadStub(), udn, mutatedNAD)
			_, err := c.syncUserDefinedNetwork(udn)

			Expect(err).To(Equal(errors.New("foreign NetworkAttachmentDefinition with the desired name already exist [test/test]")))
```
which I thin makes it easier to read.
One more thing to make it work, you need to apply similar changes from https://github.com/ovn-org/ovn-kubernetes/commit/434b0590ce8c61ade75edc996b2f7f83d530f840  to the `fake_client.go`, to make sure NADs will be added with `GetOVNClientset`

go-controller/pkg/factory/factory.go
```diff
@@ -859,6 +873,17 @@ func NewClusterManagerWatchFactory(ovnClientset *util.OVNClusterManagerClientset
...
 			return nil, err
 		}
+		wf.informers[ClusterUserDefinedNetworkType], err = newInformer(ClusterUserDefinedNetworkType, wf.udnFactory.K8s().V1().ClusterUserDefinedNetworks().Informer())
+		if err != nil {
+			return nil, err
+		}
+
+		wf.iFactory.InformerFor(&kapi.Namespace{}, func(c kubernetes.Interface, resyncPeriod time.Duration) cache.SharedIndexInformer {
+			return v1coreinformers.NewNamespaceInformer(c, resyncPeriod, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+		})
```
npinaeva, line 883: why do we need this?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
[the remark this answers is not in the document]
npinaeva: I think it is a small enough fix to add here, as the controller doesn't work properly without it anyway

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
npinaeva: I found 8 more patch calls, do you think we can replace all of them with updates?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -17,535 +19,854 @@ import (
...
-	})
+		namespaceInformer = sharedInformer.Core().V1().Namespaces()
 
-	Context("controller", func() {
+		return New(cs.NetworkAttchDefClient, nadInformer,
+			cs.UserDefinedNetworkClient, udnInformer, cudnInformer,
+			renderNADStub, podInformer, namespaceInformer)
+	}
+
+	Context("manager", func() {
```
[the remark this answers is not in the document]
ormergi: I meant to express these tests are of the wrapping controller of  UDN and CUDN controllers.
Make sense?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Done

I left usable of Patch in places the subject object is not available and require fetching.

ormergi:
Changes [[1]](https://github.com/ovn-org/ovn-kubernetes/compare/9bb1f4bb14a8d5a2568e27dcf31d687d83fd9258..b72e0a8aae0906ee75f1e49bd7bc3e2b043e6c10) [[2]](https://github.com/ovn-org/ovn-kubernetes/compare/b72e0a8aae0906ee75f1e49bd7bc3e2b043e6c10..0bbf252b45b0e3a98dd3ea5cac6d214b768b3b57): resolve https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1790014549 https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1790086404

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/0bbf252b45b0e3a98dd3ea5cac6d214b768b3b57..565b840b5610ecb2891925f3ec4d8c729611f4b2): resolve https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1790029373
In addition, fix unit test race https://github.com/ovn-org/ovn-kubernetes/actions/runs/11230700824/job/31218560622?pr=4612

test/e2e/network_segmentation.go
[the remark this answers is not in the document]
npinaeva: I think we should make a copy for "namespace-selector is mutated" test, to use non-namespace-name label and than mutate namespace labels instead of selector and also check that it is handled correctly.
Original problem was that we didn't watch for namespace changes correctly, so to test this case, we need to change namespace labels only and not touch a CUDN itself.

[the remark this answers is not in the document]
ormergi: I think its covered in unit-tests by the test "when existing namespace's matching label removed, should delete stale NADs in previously matching namespaces". 
Does it make senses to add e2e for it as well?

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/565b840b5610ecb2891925f3ec4d8c729611f4b2..d53e7987211db4ee5f5a80b46a5051ccce78f373): fix race in unit tests, seem on https://github.com/ovn-org/ovn-kubernetes/pull/4758 https://github.com/ovn-org/ovn-kubernetes/actions/runs/11260711964/job/31312531425?pr=4758

ormergi:
Cherry-pick #4758, following https://github.com/ovn-org/ovn-kubernetes/pull/4758#issuecomment-2405209734

npinaeva:
this failure https://github.com/ovn-org/ovn-kubernetes/actions/runs/11275797595/job/31359859029?pr=4612 looks suspicious, may need investigation

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/3b49084086b05fa48cbffb65ddb7a92123037cb3..54015d73520b234c6092ad429211a9d3c41a86bf): rebase

ormergi:
> this failure https://github.com/ovn-org/ovn-kubernetes/actions/runs/11275797595/job/31359859029?pr=4612 looks suspicious, may need investigation

Look like network-segmentation no-IC lane pass 
Is this check flaky in IC enable envs?

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/54015d73520b234c6092ad429211a9d3c41a86bf..c587e5a47ef791dc5a045e4b1e9a4c46212531db): rebase on latest master

tssurya:
@ormergi looks like it needs a rebase after the kube rebase landed

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/c587e5a47ef791dc5a045e4b1e9a4c46212531db..20902143e8efc4c7d46063fd517e5ee8c66a07c1):  Rebase on latest master - Adapt code following Kubernetes dependencies & Ginkgo bump

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/20902143e8efc4c7d46063fd517e5ee8c66a07c1..7e7f001705492e8bbfc06d0c62ce90485c6a434e): post `make codegen` changes, following Kubernetes dependencies bump

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/7e7f001705492e8bbfc06d0c62ce90485c6a434e..42e560e13f219120fc16b5343d78f0c8e8fb1c5d): missing bits following kube. deps. versiob bump

ormergi:
CI failures seem note related to this PR:
[e2e (shard-conformance, noHA, local, dualstack, snatGW, 1br, ic-single-node-zones)](https://github.com/ovn-org/ovn-kubernetes/actions/runs/11467071407/job/31910369659?pr=4612#logs)
reached timeout - https://github.com/ovn-org/ovn-kubernetes/issues/4340

[e2e (control-plane, noHA, shared, ipv4, SnatGW, 1br, ic-single-node-zones, disable-forwarding)](https://github.com/ovn-org/ovn-kubernetes/actions/runs/11467071407/job/31910376810?pr=4612#logs)
```
e2e egress IP validation [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts [It] disabling egress nodes impeding GRCP health check
```
https://github.com/ovn-org/ovn-kubernetes/issues/4144

[e2e (kv-live-migration, noHA, local, ipv4, SnatGW, 1br, ic-disabled, 3](https://github.com/ovn-org/ovn-kubernetes/actions/runs/11467071407/job/31910376134?pr=4612#logs)
 Kubevirt Virtual Machines with default pod network when live migration [It] with pre-copy succeeds, should keep connectivity
 https://github.com/ovn-org/ovn-kubernetes/issues/3986

npinaeva:
Let's add to the list of follow ups enabling this https://github.com/ovn-org/ovn-kubernetes/pull/4647 for CUDNs, it should work now since `ParseNetworkName` will looks for exactly 1 '.' in the udn name, and we know cudn will have 2, so it won't be recognized as a UDN.

npinaeva:
great job @ormergi ! thanks for fixing all my nits :)
/lgtm

tssurya:
thank you @ormergi and @npinaeva for the awesome work here!!!

trozet:
@ormergi couple of questions:
In your example you have:
```
apiVersion: k8s.ovn.org/v1
metadata:
  name: net0
```
which yields a NAD:
```
metadata:
  name: net0
  namespace: mynamespace
  finalizers:
  - k8s.ovn.org/user-defined-network-protection
  labels:
    k8s.ovn.org/user-defined-network: ""
  ownerReferences:
  - apiVersion: k8s.ovn.org/v1
    blockOwnerDeletion: true
    name: mynet
    uid: 97b1da95-d873-44f5-af18-0e6e7a830bd0
spec:
  config:  > 
    '{
        "cniVersion":"1.0.0",
        "type":"ovn-k8s-cni-overlay",
        "name":"cluster.udn.net0",
        "netAttachDefName":"theirnamespace/net0",
```

1. Why do we add "cluster.udn" as a prefix to the network name? The namespace scoped CRD will always have the config->name as "<namespace>/<name>", so there's no potential for the namespace CRD to guess the name of (and attach to) the cluster scoped CRD network.

2. If we create a cluster UDN with the same name and matching the namespace as an already created namespaced scope UDN we will have a failure status in the CUDN right? So this way a CUDN and a regular UDN cannot exist for the same namespace with the same name. Is that correct?

ormergi:
>     1. Why do we add "cluster.udn" as a prefix to the network name? The namespace scoped CRD will always have the config->name as "/", so there's no potential for the namespace CRD to guess the name of (and attach to) the cluster scoped CRD network.
> 

Its about the CNI config 'name', it could end up being the same as CUDN one, for exmaple:
UDN CR key is blue/net0, network name will be `blue.net0`.
CUDN CR name is net0 points to namespace blue, network name will be `blue.net0`.

>     2. If we create a cluster UDN with the same name and matching the namespace as an already created namespaced scope UDN we will have a failure status in the CUDN right? So this way a CUDN and a regular UDN cannot exist for the same namespace with the same name. Is that correct?

This is correct, CUDN controller should reflect there is already exist NAD with the same name, in the CUDN CR status.
Although, I think its better to have this prefix anyways as an additional matter to avoid conflicts with existing NADs as its cost is neglectable.
In addition, in order to add support for CUDN CRD controller report subsystem conditions, the controller need to distinct whether the given network is related to UDN or CUDN CR, having the cluster.udn prefix enable that.

In case there are UDN and CUDN NADs with the same name at the same namespace, the controller should be able to distinct between these two types reflect conflicting NAD exist in the right CR status.


For example, in case one of the controller crushed or 


BTW I fixed the example, meta.name was wrong (mynet instead of net0)

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -66,6 +66,9 @@ type Controller struct {
 	namespaceTrackerLock sync.RWMutex
 	// renderNadFn render NAD manifest from given object, enable replacing in tests.
 	renderNadFn RenderNetAttachDefManifest
+	// createPrimaryNetworkLock lock should held when primary network should be created
+	// to avoid having two components trying to create primary network in the same namespace.
```
trozet, line 70: what 2 components are you worried about here? Are you worried someone creates a cUDN and a namespace UDN at same time and in our code we are trying to create both?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -701,6 +701,52 @@ var _ = Describe("User Defined Network Controller", func() {
 					}
 				})
 			})
+
+			It("when started, CR exist, stale NADs exist, should deleted stale NADs", func() {
```
trozet, line 705: where is the code that finds stale nads and cleans them?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -65,63 +78,163 @@ func New(
...
-	// enqueue network-attachment-definitions requests in the controller workqueue
-	c.Controller.Reconcile(key)
+	return c
 }
 
 func (c *Controller) Run() error {
-	klog.Infof("Starting UserDefinedNetworkManager Controllers")
-	if err := controller.Start(c.nadNotifier.Controller, c.Controller); err != nil {
-		return fmt.Errorf("unable to start UserDefinedNetworkManager controller: %v", err)
+	klog.Infof("Starting user-defiend network controllers")
```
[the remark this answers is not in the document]
ormergi: Sorry, fixed!

ormergi (commented):
@trozet thanks for the review, PTAL

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/42e560e13f219120fc16b5343d78f0c8e8fb1c5d..658e3dc527ff3a637292ef1c15855c54cb4a4eae): Fix typos in code and commit messages.

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/42e560e13f219120fc16b5343d78f0c8e8fb1c5d..864db6ab4bd2241e64c94e104770e8055caa726f): Fix typos in code and commit messages.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
```diff
@@ -3,107 +3,135 @@ package template
...
+	}, nil
 }
 
-func ParseNetworkName(networkName string) (udnNamespace, udnName string) {
-	parts := strings.Split(networkName, ".")
-	if len(parts) == 2 {
-		return parts[0], parts[1]
+func validateTopology(spec SpecGetter) error {
+	if spec.GetTopology() == userdefinednetworkv1.NetworkTopologyLayer3 && spec.GetLayer3() == nil ||
+		spec.GetTopology() == userdefinednetworkv1.NetworkTopologyLayer2 && spec.GetLayer2() == nil {
```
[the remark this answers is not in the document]
trozet: right, but we still need localnet for CUDN we just dont want it for namespaced UDN. I guess we can do this in a follow up @maiqueb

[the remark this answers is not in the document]
maiqueb: I'm fine w/ having localnet support in a follow-up.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -265,57 +370,46 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
-	// try to find udn using network name
-	udnNamespace, udnName := template.ParseNetworkName(networkName)
-	if udnName == "" {
-		return nil
-	}
-	udn, err := c.udnLister.UserDefinedNetworks(udnNamespace).Get(udnName)
-	if err != nil {
-		return nil
-	}
+func (c *Controller) cudnNeedUpdate(_ *userdefinednetworkv1.ClusterUserDefinedNetwork, _ *userdefinednetworkv1.ClusterUserDefinedNetwork) bool {
```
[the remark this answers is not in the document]
trozet: well there may be changes that will trigger the update logic that we could filter out here. If there were like labels added to the CUDN, or maybe the status was updated of the CUDN. It would be worth thinking through what we actually care about changing here, and filtering for only that.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -265,57 +370,46 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			return nil
-		}
-		return fmt.Errorf("failed to update UserDefinedNetwork %s/%s status: %w", udnNamespace, udnName, err)
+	updateStatusErr := c.updateClusterUDNStatus(cudnCopy, nads, syncErr)
+
+	var networkInUse *networkInUseError
+	if errors.As(syncErr, &networkInUse) {
+		c.cudnController.ReconcileAfter(key, c.networkInUseRequeueInterval)
```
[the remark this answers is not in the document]
trozet: oh looks like it adds to this delaying queue which is buffered for 1000 entries:

waitingForAddCh: make(chan *waitFor, 1000),

```
func (q *delayingType[T]) AddAfter(item T, duration time.Duration) {
	// don't add if we're already shutting down
	if q.ShuttingDown() {
		return
	}

	q.metrics.retry()

	// immediately add things with no delay
	if duration <= 0 {
		q.Add(item)
		return
	}

	select {
	case <-q.stopCh:
		// unblock if ShutDown() is called
	case q.waitingForAddCh <- &waitFor{data: item, readyAt: q.clock.Now().Add(duration)}:
	}
}
```

so I guess we should good unless 1000 networks are all doing the same thing

@npinaeva fyi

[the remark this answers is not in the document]
npinaeva: `case q.waitingForAddCh <- &waitFor{data: item, readyAt: q.clock.Now().Add(duration)}:`
will wait till there is space in the channel right?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
trozet: My comment was really about NADs that exist at start up that are owned by the CUDN controller, but they no longer have a corresponding CUDN. @npinaeva pointed out that isn't possible due to the finalizer, so I guess it's not something we need to worry about.

dist/templates/rbac-ovnkube-cluster-manager.yaml.j2
```diff
@@ -82,6 +82,9 @@ rules:
           - egressservices/status
           - userdefinednetworks
           - userdefinednetworks/status
+          - clusteruserdefinednetworks
+          - clusteruserdefinednetworks/status
+          - clusteruserdefinednetworks/finalizers
```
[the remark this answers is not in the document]
trozet: yeah i think your commit message just confused me a bit

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
trozet: Couple of questions/comments:
1. Why would we limit this protection to just primary network? Secondary networks could have the same problem.
2. By using this mutex we are serializing  NAD creation across all namespaces. I guess for now this is OK, but in the future if we want to improve perf we may want to use a different mechanism.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -43,51 +48,58 @@ func (c *Controller) updateNAD(obj client.Object, nad *netv1.NetworkAttachmentDe
...
-			return nil, fmt.Errorf("failed to update NetworkAttachmentDefinition: %w", uerr)
-		}
-		klog.Infof("Updated NetworkAttachmentDefinition [%s/%s]", uNAD.Namespace, uNAD.Name)
-		return uNAD, nil
+	if reflect.DeepEqual(nadCopy.Spec.Config, desiredNAD.Spec.Config) {
+		return nadCopy, nil
+	}
+
+	nadCopy.Spec.Config = desiredNAD.Spec.Config
+	updatedNAD, err := c.nadClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(nadCopy.Namespace).Update(context.Background(), nadCopy, metav1.UpdateOptions{})
```
trozet, line 69: so on update we allow to update the spec? I thought there were certain fields that were immutable for UDNs. Are those enforced in the CRD rather than in this code?

dist/templates/rbac-ovnkube-cluster-manager.yaml.j2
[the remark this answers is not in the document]
ormergi: I rephrased that part in the commit message, I hope its better now.

ormergi (commented):
@trozet thanks for the feedback, PTAL

ormergi:
[Changes](https://github.com/ovn-org/ovn-kubernetes/compare/864db6ab4bd2241e64c94e104770e8055caa726f...6e422b37593ddca324d787c5ef3e93d01f5e1b79) : 
- Add missing validation on the CUDN CRD to make the network spec immutable.
- Fix NAD create synchronization to cover secondary network scenario.

PR #4614, opened 2024-08-14
EndpointSlice mirror controller: handle tombstone deletion

kyrtapz:


ricky-rav:
So we're handling the event where the deletion was issued from the API server but the controller missed it by being temporarily disconnected, right? That's what I see in
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/cache/delta_fifo.go#L809-L812

I see the endpointslice mirroring controller does the same https://github.com/kubernetes/kubernetes/blob/69dbf2eee96f1c95c097370ddcb1d5c30f86bec8/pkg/controller/endpointslicemirroring/utils.go#L183

kyrtapz:
The one failure is unrelated and happened on kind setup.

martinkennelly (approved):

PR #4615, opened 2024-08-14
networksegmentation,UDN CRD: Remove localnet

ormergi:

Following #4486, the UserDefinedNetwork CRD is introduced, it enable admins and non admins users create OVN-K networks, including l2, l3, and localnet topologies.

Enable non-admin users should not be able to create localnet networks because unlike l2/l3 they have access to the underlay network.

The only thing prevents non-admin users tap into other localnets is the fact the controller generates the NAD network-name based on the CR request meta.namespace/meta.name.
For example: project admin of namespace "A" create UDN "foo", network-name will be "A.foo".
The cluster-admin has to prepare bridge-mapping with localnet="A.foo" for this to work.

In addition, since only cluster-admin can setup localnet on the nodes, cluster-admin must be involved in setup the localnet networks. Which, for localnet networks, miss the point in having the UDN CRD targeted for non-admin users in the first place.

This PR removes creating localnet topology network using UserDefinedNetwork CRD option.

Fixes #

Localnet should remain exclusive for cluster-admins to manage.

We plan to introduce cluster UDN CRD that is targeted for cluster-admins only, to enable creating cross namesapce networks.
Localnet topology fits better the cluster UDN CRD as it managed by cluster-admin only, same same persona who needs to prepare the nodes localnet bridge-mapping.

```release-note
NONE
```


ormergi:
CI fails due to failing job with known issues  
`ovn-ci / e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones)`  https://github.com/ovn-org/ovn-kubernetes/issues/4432
`ovn-ci / e2e (shard-conformance, HA, local, ipv6, snatGW, 1br, ic-disabled)` https://github.com/ovn-org/ovn-kubernetes/issues/4340

tssurya (approved):
LGTM given @trozet was ok with this change (based on sync meeting discussion last week) and it looks pretty straightforward I can merge specially since @npinaeva has also LGTM-ed this.
I'll leave this comment here for a few hours to see if Tim wants to do the final ack here. If not I can merge it.

PR #4621, opened 2024-08-14
Consistently reorder test/e2e imports

danwinship:
Just a cleanup to `test/e2e`; the `import` ordering was kind of a mess, and very inconsistent.
(`gofmt` requires that the imports section consists of 1 or more newline-separated groups, where each group is individually alphabetized, but it doesn't do any enforcement of what should and shouldn't be in the same group.)

The main codebase is pretty consistent about at least not mixing up stdlib imports and third-party imports, though not totally consistent beyond that. This makes the e2e files use the order (1) stdlib, (2) ginkgo/gomega, (3) ovn-k-internal, (4) everything else; which is sort of the overall consensus ordering.

```release-note
NONE
```

cc @martinkennelly

### DIFF
--- a/test/e2e/acl_logging.go
+++ b/test/e2e/acl_logging.go
@@ -8,14 +8,14 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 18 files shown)


martinkennelly (approved):

martinkennelly:
Rerunning tests because failed to pull metal LB dependency.

tssurya (approved):

PR #4623, opened 2024-08-14
Add some more ExpectNoError checks to the e2e tests

danwinship:
More fixes pulled out of @martinkennelly 's branch to start adding support for running ovn-k e2e downstream in OCP. This just takes a bunch of places where the test cases were doing something and ignoring possible errors, and changes them to assert that an error didn't occur instead.

(In theory, this should not result in us discovering any new bugs in the _existing_ code, since if something actually went wrong, the test would presumably fail later on anyway. It just means that in case of _future_ bugs, the tests should fail closer to the point where something actually went wrong, rather than failing more mysteriously later.)

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -898,7 +898,8 @@ var _ = ginkgo.Describe("e2e control plane", func() {
 				!strings.HasPrefix(pod.Name, "ovnkube-identity") &&
 				!strings.HasPrefix(pod.Name, "ovs-node") {
 				framework.Logf("%q", pod.Namespace)
-				deletePodWithWaitByName(context.TODO(), f.ClientSet, pod.Name, ovnNamespace)
+				err = deletePodWithWaitByName(context.TODO(), f.ClientSet, pod.Name, ovnNamespace)
+				framework.ExpectNoError(err, fmt.Sprintf("failed to delete pod %s", pod.Name))
 				framework.Logf("Deleted control plane pod %q", pod.Name)
 			}
 		}
@@ -931,7 +932,8 @@ var _ = ginkgo.Describe("e2e control plane", func() {
 		for _, pod := range podList.Items {
 			if strings.HasPrefix(pod.Name, controlPlanePodName) && !strings.HasPrefix(pod.Name, "ovs-node") {
 				framework.Logf("%q", pod.Namespace)
-				deletePodWithWaitByName(context.TODO(), f.ClientSet, pod.Name, ovnNamespace)
+				err = deletePodWithWaitByName(context.TODO(), f.ClientSet, pod.Name, ovnNamespace)
+				framework.ExpectNoError(err, fmt.Sprintf("failed to delete pod %s", pod.Name))
 				framework.Logf("Deleted control plane pod %q", pod.Name)
 			}
 		}
@@ -2186,8 +2188,10 @@ var _ = ginkgo.Describe("e2e delete databases", func() {
 		var (
 			command = []string{"/agnhost", "netexec", fmt.Sprintf("--http-port=" + port)}
 		)
-		createGenericPod(f, pod1Name, node1Name, f.Namespace.Name, command)
-		createGenericPod(f, pod2Name, node2Name, f.Namespace.Name, command)
+		_, err := createGenericPod(f, pod1Name, node1Name, f.Namespace.Name, command)
+		framework.ExpectNoError(err, "failed to create pod %s/%s", f.Namespace.Name, pod1Name)
+		_, err = createGenericPod(f, pod2Name, node2Name, f.Namespace.Name, command)
+		framework.ExpectNoError(err, "failed to create pod %s/%s", f.Namespace.Name, pod2Name)
 
 		pod2IP := getPodAddress(pod2Name, f.Namespace.Name)
 
@@ -2340,7 +2344,8 @@ var _ = ginkgo.Describe("e2e delete databases", func() {
 		}
 
 		framework.Logf("wait for all the Deployment to become ready again after pod deletion")
-		e2edeployment.WaitForDeploymentComplete(f.ClientSet, dbDeployment)
+		err = e2edeployment.WaitForDeploymentComplete(f.ClientSet, dbDeployment)
+		framework.ExpectNoError(err, "failed to wait for DB deployment to complete")
 
 		framework.Logf("all the pods finish full restart")
 
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -613,7 +613,7 @@ func restartOVNKubeNodePod(clientset kubernetes.Interface, namespace string, nod
 		return fmt.Errorf("could not find ovnkube-node pod running on node %s", nodeName)
 	}
 	for _, pod := range ovnKubeNodePods.Items {
-		if err := e2epod.DeletePodWithWait(context.TODO(), clientset, &pod); err != nil {
+		if err := deletePodWithWait(context.TODO(), clientset, &pod); err != nil {
 			return fmt.Errorf("could not delete ovnkube-node pod on node %s: %w", nodeName, err)
 		}
 	}
@@ -852,7 +852,7 @@ var _ = ginkgo.Describe("e2e control plane", func() {
 		}
 
 		ginkgo.By("Deleting ovnkube control plane pod " + podName)
-		e2epod.DeletePodWithWaitByName(context.TODO(), f.ClientSet, podName, ovnNamespace)
+		deletePodWithWaitByName(context.TODO(), f.ClientSet, podName, ovnNamespace)
 		framework.Logf("Deleted ovnkube control plane pod %q", podName)
 
 		ginkgo.By("Ensuring there were no connectivity errors")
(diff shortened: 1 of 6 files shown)


martinkennelly (approved):

tssurya (dismissed):
ping me when CI runs, I can merge this LGTM

tssurya:
control-plane's are failing..

test/e2e/e2e.go
```diff
@@ -789,7 +789,8 @@ var _ = ginkgo.Describe("e2e control plane", func() {
 				pod.Name != "etcd-ovn-control-plane" &&
 				!strings.HasPrefix(pod.Name, "ovs-node") {
 				framework.Logf("%q", pod.Namespace)
-				e2epod.DeletePodWithWaitByName(context.TODO(), f.ClientSet, pod.Name, ovnNs)
+				err = e2epod.DeletePodWithWaitByName(context.TODO(), f.ClientSet, pod.Name, ovnNs)
+				framework.ExpectNoError(err, fmt.Sprintf("failed to delete pod %s", pod.Name))
```
danwinship, line 793: LOL, this is causing e2e failures because of a bug in `e2epod.DeletePodWithWaitByName`; it deletes the pod, and then loops until no pod with the given name exists. But if something else recreates the pod immediately, it doesn't notice that the pod that exists isn't the same one as the one it deleted.

(So, without the `ExpectNoError` check, the failure mode would be "it deletes the pod successfully, then gets confused and loops for 5 minutes and then the function returns an error, but we ignore the error and keep going (which is fine, because the pod we meant to delete _was_ deleted)", whereas with this PR, the failure mode is "it deletes the pod successfully, then gets confused and loops for 5 minutes and then the function returns an error, which causes us to fail the test".)

danwinship:
> Filed [kubernetes/kubernetes#126785](https://github.com/kubernetes/kubernetes/issues/126785). Let's see what happens there.

No consensus on a short-term fix, so I'll just fix this locally.

test/e2e/util.go
```diff
@@ -1316,3 +1317,69 @@ func matchIPv4StringFamily(ipStrings []string) (string, error) {
...
+
+// This is an alternative version of e2epod.WaitForPodNotFoundInNamespace(), which takes
+// a resourceVersion as well.
+func waitForPodNotFoundInNamespace(ctx context.Context, c clientset.Interface, podName, ns, resourceVersion string, timeout time.Duration) error {
+        err := framework.Gomega().Eventually(ctx, framework.HandleRetry(func(ctx context.Context) (*v1.Pod, error) {
+                pod, err := c.CoreV1().Pods(ns).Get(ctx, podName, metav1.GetOptions{})
+                if apierrors.IsNotFound(err) {
+                        return nil, nil
+                }
+		if pod != nil && pod.ResourceVersion != resourceVersion {
```
martinkennelly, line 1376: by this stage we have submitted a delete request for the pod - can the resource version for the pod change? If it can, then we may falsely think its deleted here.

danwinship: no, ResourceVersion is immutable; if ResourceVersion has changed, that means the pod was deleted and recreated. (This check is the difference from the original `e2epod.WaitForPodNotFoundInNamespace`.)

martinkennelly: I thought resource version implemented by k8 may increment with each change applied to the pod object. Youre right that if it was deleted and created it would get a different resource version but it can also change with any changes to the pod object.

danwinship: uh... wow, not sure what I was thinking...

danwinship:
@martinkennelly fixed!

martinkennelly (dismissed):

test/e2e/util.go
```diff
@@ -1337,3 +1338,69 @@ func matchIPv4StringFamily(ipStrings []string) (string, error) {
...
+
+// This is a replacement for e2epod.DeletePodWithWaitByName(), which does not handle pods
+// that may be automatically restarted (https://issues.k8s.io/126785)
+func deletePodWithWaitByName(ctx context.Context, c clientset.Interface, podName, podNamespace string) error {
+	pod, err := c.CoreV1().Pods(podNamespace).Get(ctx, podName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil // assume pod was already deleted
+		}
+		return fmt.Errorf("pod Get API error: %w", err)
```
tssurya, line 1379:
```suggestion
		return fmt.Errorf("pod Get API error for %s/%s: %w", err, podNamespace, podName)
```
nit: is it worth it including that info or will the error returned have that info?

test/e2e/util.go
```diff
@@ -1337,3 +1338,69 @@ func matchIPv4StringFamily(ipStrings []string) (string, error) {
...
+// This is a replacement for e2epod.DeletePodWithWait(), which does not handle pods that
+// may be automatically restarted (https://issues.k8s.io/126785)
+func deletePodWithWait(ctx context.Context, c clientset.Interface, pod *v1.Pod) error {
+	if pod == nil {
+		return nil
+	}
+	if pod.UID == "" {
+		// We only recurse into deletePodWithWaitByName when UID is *not* set, to
+		// avoid infinite loops.
+		return deletePodWithWaitByName(ctx, c, pod.Name, pod.Namespace)
```
tssurya, line 1351: this is kinda like providing it a second chance and then bailing if again there is no UID?

test/e2e/util.go
```diff
@@ -1337,3 +1338,69 @@ func matchIPv4StringFamily(ipStrings []string) (string, error) {
...
+	}
+	oldUID := pod.UID
```
tssurya, line 1353: why is this oldUID? its the current pod.UID of the podObj we are about to delete right?

tssurya (commented):
a few doubts otherwise looking good

test/e2e/util.go
danwinship: No, it's because the caller might do something like:

```
pod := &v1.Pod{ ... }
_, err := client.CoreV1().Pods(namespace).Create(...)
...
deletePodWithWait(ctx, client, pod)
```

But since that pod object is just the one they created by hand before, it's not going to have the UID filled in. So in that case we just fall back to doing `deletePodWithWaitByName`, and fetch the pod details from the server.

I was trying to make both APIs still work in all scenarios just like they used to, but maybe it would make more sense to just say "you can only call this with a fully-filled-in Pod struct" and have it return an error if you break that rule?

test/e2e/util.go
danwinship: I don't know... this is mostly a copy of the existing k8s e2e method.
It will end up being clear what pod it's talking about based on where the error gets asserted against anyway.

test/e2e/util.go
```diff
@@ -1337,3 +1338,69 @@ func matchIPv4StringFamily(ipStrings []string) (string, error) {
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil // assume pod was already deleted
+		}
+		return fmt.Errorf("pod Delete API error: %w", err)
+	}
+	framework.Logf("Wait up to %v for pod %q to be fully deleted", e2epod.PodDeleteTimeout, pod.Name)
+	err = waitForPodNotFoundInNamespace(ctx, c, pod.Name, pod.Namespace, oldUID, e2epod.PodDeleteTimeout)
+	if err != nil {
+		return fmt.Errorf("pod %q was not deleted: %w", pod.Name, err)
```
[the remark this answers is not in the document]
danwinship: yeah, again, it will be clear what pod this is about based on what line the error gets asserted on

test/e2e/util.go
```diff
@@ -1337,3 +1338,69 @@ func matchIPv4StringFamily(ipStrings []string) (string, error) {
...
+
+// This is an alternative version of e2epod.WaitForPodNotFoundInNamespace(), which takes
+// a UID as well.
+func waitForPodNotFoundInNamespace(ctx context.Context, c clientset.Interface, podName, ns string, uid types.UID, timeout time.Duration) error {
+        err := framework.Gomega().Eventually(ctx, framework.HandleRetry(func(ctx context.Context) (*v1.Pod, error) {
+                pod, err := c.CoreV1().Pods(ns).Get(ctx, podName, metav1.GetOptions{})
+                if apierrors.IsNotFound(err) {
+                        return nil, nil
+                }
+		if pod != nil && pod.UID != uid {
```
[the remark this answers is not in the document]
danwinship: This doesn't check "uid not set".

There are two reasons to return successfully (`nil, nil`) from the `Eventually`:
- The pod doesn't exist (ie, it was deleted). Lines 1394-1396 handles this.
- The pod does exist, but it has a different UID than the pod we started with (ie, it was deleted and then recreated before we noticed it was deleted). Lines 1397-1399 handle this.

And then line 1400 handles either "the pod still exists" or "we got an error other than NotFound".

test/e2e/util.go
danwinship: hm... yeah, this code looks like I originally wrote something else, and then rewrote it, and there are still bits of the old code leaking through...

test/e2e/util.go
tssurya: aaaaahhhh I see....jeez I kept thinking when could a pod come with no UID unless apiserver was sleeping, i forgot we did stupid things in test

test/e2e/util.go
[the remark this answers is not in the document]
tssurya: > This doesn't check "uid not set".
> 
oops for whatever reason I have read pod.UID == "" in your above util and scrolled down here wow

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13244726208/job/36970382797?pr=4623
is known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4936
```
  [FAIL] e2e egress IP validation Network Segmentation: IPv4 L2 role primary [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts [It] disabling egress nodes impeding GRCP health check
```

tssurya (approved):
LGTM

PR #4624, opened 2024-08-14
Don't use zone 0 for UDN Networks

tssurya:
```
sh-5.2# ovn-nbctl list logical-router GR_l3.network_ovn-control-plane
_uuid               : b2a4db8a-54b2-44c9-89f5-dacf381dca17
copp                : d51232db-7768-4d6c-aab0-9303a7856e8c
enabled             : []
external_ids        : {"k8s.ovn.org/network"=l3-network, "k8s.ovn.org/topology"=layer3, physical_ip="172.18.0.4", physical_ips="172.18.0.4,fc00:f853:ccd:e793::4,169.254.0.11,fd69::b"}
load_balancer       : []
load_balancer_group : []
name                : GR_l3.network_ovn-control-plane
nat                 : []
options             : {always_learn_from_arp_request="false", chassis="53db7ce4-f836-41ca-a176-8e237073f602", dynamic_neigh_routers="true", lb_force_snat_ip=router_ip, mac_binding_age_threshold="300", snat-ct-zone="0"}
policies            : []
ports               : [29589cd0-a17c-4a60-b484-8bac3f9cbaac, df65b326-689a-463c-98e1-9037f54cce05]
static_routes       : [439fb202-8981-4313-b524-cee44c243bbd, 9245d6a3-b918-4699-ac11-43b7cf3132f7, c0ea5743-7f72-4e37-bff8-9c4982fb31da, c52e12b7-d374-4e3b-9222-d1ed02e73820, da67514d-ff82-4627-bd11-dd6b3a704bb5]
```

everytime we create a UDN we were causing disruption to services for default network.
Don't set the UDN GR's snat zone option to 0.

### DIFF
--- a/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
+++ b/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
@@ -453,12 +453,8 @@ func expectedGWEntities(nodeName string, netInfo util.NetInfo, gwConfig util.L3G
 		expectedGatewayChassis(nodeName, netInfo, gwConfig),
 		expectedStaticMACBinding(gwRouterName, nextHopMasqueradeIP()),
 	)
-	for _, entity := range expectedExternalSwitchAndLSPs(netInfo, gwConfig, nodeName) {
-		expectedEntities = append(expectedEntities, entity)
-	}
-	for _, entity := range expectedJoinSwitchAndLSPs(netInfo, nodeName) {
-		expectedEntities = append(expectedEntities, entity)
-	}
+	expectedEntities = append(expectedEntities, expectedExternalSwitchAndLSPs(netInfo, gwConfig, nodeName)...)
+	expectedEntities = append(expectedEntities, expectedJoinSwitchAndLSPs(netInfo, nodeName)...)
 	return expectedEntities
 }
 
@@ -757,7 +753,6 @@ func networkSubnet(netInfo util.NetInfo) string {
 func gwRouterOptions(gwConfig util.L3GatewayConfig) map[string]string {
 	return map[string]string{
 		"lb_force_snat_ip":              "router_ip",
-		"snat-ct-zone":                  "0",
 		"mac_binding_age_threshold":     "300",
 		"chassis":                       gwConfig.ChassisID,
 		"always_learn_from_arp_request": "false",
--- a/go-controller/pkg/ovn/gateway.go
+++ b/go-controller/pkg/ovn/gateway.go
@@ -208,8 +208,12 @@ func (gw *GatewayManager) GatewayInit(
 		"dynamic_neigh_routers":         "true",
 		"chassis":                       l3GatewayConfig.ChassisID,
 		"lb_force_snat_ip":              "router_ip",
-		"snat-ct-zone":                  "0",
 		"mac_binding_age_threshold":     types.GRMACBindingAgeThreshold,
+	}
+	// set the snat-ct-zone only for the default network
+	// for UDN's OVN will pick a random one
+	if gw.netInfo.GetNetworkName() == types.DefaultNetworkName {
+		logicalRouterOptions["snat-ct-zone"] = "0"
 	}
 	logicalRouterExternalIDs := map[string]string{
 		"physical_ip":  physicalIPs[0],


go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
```diff
@@ -453,12 +453,8 @@ func expectedGWEntities(nodeName string, netInfo util.NetInfo, gwConfig util.L3G
 		expectedGatewayChassis(nodeName, netInfo, gwConfig),
 		expectedStaticMACBinding(gwRouterName, nextHopMasqueradeIP()),
 	)
-	for _, entity := range expectedExternalSwitchAndLSPs(netInfo, gwConfig, nodeName) {
-		expectedEntities = append(expectedEntities, entity)
-	}
-	for _, entity := range expectedJoinSwitchAndLSPs(netInfo, nodeName) {
-		expectedEntities = append(expectedEntities, entity)
-	}
+	expectedEntities = append(expectedEntities, expectedExternalSwitchAndLSPs(netInfo, gwConfig, nodeName))
```
tssurya, line 456: this was just some silly lint suggestion that I saw which I included in this PR

dceara (approved):
I didn't test it but it makes sense to me.  If CI is green this should probably be merged.

tssurya:
not sure why this is failing:
```
•! Panic [0.556 seconds]
OVN Multi-Homed pod operations
/home/surya/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go:52
  reconciles a new
  /home/surya/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go:83
    pod on a user defined primary network on an interconnect cluster [It]
    /home/surya/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go:195

    Test Panicked
    Expected model does not have a (named) UUID:
        <[]libovsdb.TestData | len:3, cap:3>: [
            <*nbdb.LogicalSwitch | 0xc001c38460>{
                UUID: "ext-UUID",
                ACLs: nil,
                Copp: nil,
                DNSRecords: nil,
                ExternalIDs: {
                    "k8s.ovn.org/topology": "layer3",
                    "k8s.ovn.org/network": "isolatednet",
                },
                ForwardingGroups: nil,
                LoadBalancer: nil,
                LoadBalancerGroup: nil,
                Name: "ext_isolatednet_test
```

tssurya:
ah missed a spot hehe ; fixed that as well, should be good now

tssurya:
random flake? https://github.com/ovn-org/ovn-kubernetes/actions/runs/10395056569/job/28793634404?pr=4624 let me retry this

dceara (approved):
Looks good to me, thanks!

martinkennelly:
@tssurya can you explain to me why it caused disruption by using 0? If you dont specify 0, what happens? Thanks!

tssurya:
@martinkennelly 0 is the default conntrack zone and default network's GR shares this zone with the host and this is the zone where we SNAT to nodeIP on GR. When a new GR is created with ct-zone set to 0 ovn-controller will flush conntrack  - which means all connections in zone 0 which is probably 80% of traffic will get flushed

For UDN's given we don't SNAT to nodeIP on GR's by design we never intended to set it to 0, when we don't set a value on the GR, ovn-controller will auto-allocate a custom value for SNAT-ing (to masqueradeIPs for that network) which is exactly what we want for UDNs.

However because we were creating a GR set to zone0 for SNAT; every time a new Network was created, we'd create a new GR which would flush connections in zone0 causing trouble.

dceara:
> When a new GR is created with ct-zone set to 0 ovn-controller will flush conntrack - which means all connections in zone 0 which is probably 80% of traffic will get flushed

Slight correction: when ovn-controller starts using any zone it didn't previously use for a new LSP or GR it will flush that zone.

In this case ovn-kubernetes was requesting that OVN uses 0 as SNAT zone for each new UDN gateway router.   That made ovn-controller flush the "newly assigned zone" (well, it was actually newly assigned only for the UDN gateway router).  Which interrupted all existing sessions in the default network.

Aside from that, using the same zone (0) for default and UDN is an issue because IP subnets may overlap between the default and user defined networks.

PR #4625, opened 2024-08-14
corrections to user defined networking 

JacobTanenbaum:
this PR does three things 
1. ensure that user defined primary networks only use ip families the cluster supports
2. ensure that e2e tests select the correct families for primary networks under test
3. increase the testing of NAD and UDN CRDS to CIDR overlap and pod2Egress e2e tests 

Fixes #

```release-note

```

### DIFF
--- a/go-controller/pkg/clustermanager/pod/allocator_test.go
+++ b/go-controller/pkg/clustermanager/pod/allocator_test.go
@@ -569,6 +569,8 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 
 			config.OVNKubernetesFeature.EnableInterconnect = tt.idAllocation
 
+			// config.IPv4Mode needs to be set so that the ipv4 of the userdefined primary networks can match the running cluster
+			config.IPv4Mode = true
 			netInfo, err := util.NewNetInfo(netConf)
 			if err != nil {
 				t.Fatalf("Invalid netConf")
--- a/go-controller/pkg/clustermanager/secondary_network_unit_test.go
+++ b/go-controller/pkg/clustermanager/secondary_network_unit_test.go
@@ -114,6 +114,7 @@ var _ = ginkgo.Describe("Cluster Controller Manager", func() {
 			)
 
 			ginkgo.BeforeEach(func() {
+
 				fakeClient = &util.OVNClusterManagerClientset{
 					KubeClient:            fake.NewSimpleClientset(&v1.NodeList{Items: nodes()}),
 					IPAMClaimsClient:      fakeipamclaimclient.NewSimpleClientset(),
@@ -674,24 +675,6 @@ var _ = ginkgo.Describe("Cluster Controller Manager", func() {
 				netInfo    util.NetInfo
 			)
 
-			ginkgo.BeforeEach(func() {
-				var err error
-				netInfo, err = util.NewNetInfo(
-					&ovncnitypes.NetConf{
-						NetConf:  types.NetConf{Name: "blue"},
-						Role:     ovntypes.NetworkRolePrimary,
-						Subnets:  subnets,
-						Topology: ovntypes.Layer2Topology,
-					})
-				gomega.Expect(err).NotTo(gomega.HaveOccurred())
-
-				fakeClient = &util.OVNClusterManagerClientset{
-					KubeClient:            fake.NewSimpleClientset(&v1.NodeList{Items: nodes()}),
-					IPAMClaimsClient:      fakeipamclaimclient.NewSimpleClientset(),
-					NetworkAttchDefClient: fakenadclient.NewSimpleClientset(),
-				}
-			})
-
 			ginkgo.It("Automatically reserves IPs for the GW (.1) and mgmt port (.2)", func() {
 				app.Action = func(ctx *cli.Context) error {
 					gomega.Expect(
@@ -701,6 +684,20 @@ var _ = ginkgo.Describe("Cluster Controller Manager", func() {
 						)).To(gomega.Succeed())
 
 					var err error
+					netInfo, err = util.NewNetInfo(
+						&ovncnitypes.NetConf{
+							NetConf:  types.NetConf{Name: "blue"},
+							Role:     ovntypes.NetworkRolePrimary,
+							Subnets:  subnets,
+							Topology: ovntypes.Layer2Topology,
+						})
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					fakeClient = &util.OVNClusterManagerClientset{
+						KubeClient:            fake.NewSimpleClientset(&v1.NodeList{Items: nodes()}),
+						IPAMClaimsClient:      fakeipamclaimclient.NewSimpleClientset(),
+						NetworkAttchDefClient: fakenadclient.NewSimpleClientset(),
+					}
 					f, err = factory.NewClusterManagerWatchFactory(fakeClient)
 					gomega.Expect(err).NotTo(gomega.HaveOccurred())
 					gomega.Expect(f.Start()).NotTo(gomega.HaveOccurred())
@@ -740,7 +737,12 @@ var _ = ginkgo.Describe("Cluster Controller Manager", func() {
 					return nil
 				}
 
-				gomega.Expect(app.Run([]string{app.Name})).To(gomega.Succeed())
+				gomega.Expect(app.Run([]string{
+					app.Name,
+					// define the cluster as dualstack so the user defined primary network matches the ip family
+					"--cluster-subnets=10.128.0.0/14,fd00:10:244::/48",
+					"--k8s-service-cidrs=172.16.1.0/24,fd02::/112",
+				})).To(gomega.Succeed())
 			})
 
 		})
--- a/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
@@ -10,6 +10,7 @@ import (
 
 	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	udnv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1"
 )
 
@@ -211,6 +212,9 @@ var _ = Describe("NetAttachDefTemplate", func() {
 				Spec: netv1.NetworkAttachmentDefinitionSpec{Config: expectedNadNetConf},
 			}
 
+			// must be defined so the primary user defined network can match the ip families of the underlying cluster
+			config.IPv4Mode = true
+			config.IPv6Mode = true
 			nad, err := RenderNetAttachDefManifest(testUdn)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(nad.TypeMeta).To(Equal(expectedNAD.TypeMeta))
--- a/go-controller/pkg/util/multi_network.go
+++ b/go-controller/pkg/util/multi_network.go
@@ -700,17 +700,32 @@ func NewNetInfo(netconf *ovncnitypes.NetConf) (NetInfo, error) {
 	if netconf.Name == types.DefaultNetworkName {
 		return &DefaultNetInfo{}, nil
 	}
+	var ni NetInfo
+	var err error
 	switch netconf.Topology {
 	case types.Layer3Topology:
-		return newLayer3NetConfInfo(netconf)
+		ni, err = newLayer3NetConfInfo(netconf)
 	case types.Layer2Topology:
-		return newLayer2NetConfInfo(netconf)
+		ni, err = newLayer2NetConfInfo(netconf)
 	case types.LocalnetTopology:
-		return newLocalnetNetConfInfo(netconf)
+		ni, err = newLocalnetNetConfInfo(netconf)
 	default:
 		// other topology NAD can be supported later
 		return nil, fmt.Errorf("topology %s not supported", netconf.Topology)
 	}
+	if err != nil {
+		return nil, err
+	}
+	if ni.IsPrimaryNetwork() && ni.IsSecondary() {
+		ipv4Mode, ipv6Mode := ni.IPMode()
+		if ipv4Mode && !config.IPv4Mode {
+			return nil, fmt.Errorf("network %s is attempting to use ipv4 subnets but the cluster does not support ipv4", ni.GetNetworkName())
+		}
+		if ipv6Mode && !config.IPv6Mode {
+			return nil, fmt.Errorf("network %s is attempting to use ipv6 subnets but the cluster does not support ipv6", ni.GetNetworkName())
+		}
+	}
+	return ni, nil
 }
 
 // ParseNADInfo parses config in NAD spec and return a NetAttachDefInfo object for secondary networks
--- a/go-controller/pkg/util/multi_network_test.go
+++ b/go-controller/pkg/util/multi_network_test.go
@@ -964,6 +964,90 @@ func TestSubnetOverlapCheck(t *testing.T) {
 	}
 }
 
+func TestNewNetInfo(t *testing.T) {
+	type testConfig struct {
+		desc          string
+		subnets       string
+		ipv4Cluster   bool
+		ipv6Cluster   bool
+		expectedError error
+	}
+
+	tests := []testConfig{
+		{
+			desc:        "ipv4 primary network in ipv4 cluster",
+			subnets:     "192.168.200.0/16",
+			ipv4Cluster: true,
+		},
+		{
+			desc:          "ipv4 primary network in ipv6 cluster",
+			subnets:       "192.168.200.0/16",
+			ipv6Cluster:   true,
+			expectedError: fmt.Errorf("network l3-network is attempting to use ipv4 subnets but the cluster does not support ipv4"),
+		},
+		{
+			desc:        "ipv4 primary network in dualstack cluster",
+			subnets:     "192.168.200.0/16",
+			ipv4Cluster: true,
+			ipv6Cluster: true,
+		},
+		{
+			desc:          "ipv6 primary network in ipv4 cluster",
+			subnets:       "fda6::/48",
+			ipv4Cluster:   true,
+			expectedError: fmt.Errorf("network l3-network is attempting to use ipv6 subnets but the cluster does not support ipv6"),
+		},
+		{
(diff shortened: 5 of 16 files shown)


test/e2e/network_segmentation.go
```diff
@@ -46,6 +46,10 @@ var _ = Describe("Network Segmentation", func() {
 		nadClient, err = nadclient.NewForConfig(f.ClientConfig())
 		Expect(err).NotTo(HaveOccurred())
 	})
+	AfterEach(func() {
```
ormergi, line 49: Why is it necessary to delete the pods although namespace is deleted when test is done?

go-controller/pkg/util/multi_network_test.go
```diff
@@ -21,6 +21,44 @@ import (
 	ovntypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 )
 
+/*
```
ormergi, line 24: Leftover?

go-controller/pkg/util/multi_network.go
```diff
@@ -761,6 +761,21 @@ func ValidateNetConf(nadName string, netconf *ovncnitypes.NetConf) error {
 		return fmt.Errorf("the subnet attribute must be defined for layer2 primary user defined networks")
 	}
 
+	//ensure that the cluster supports ipfamilies that the NAD is using
```
ormergi, line 764: Please note that `parseSubents` is executed as part of `ParseNadInfo` by  `NewNetInfo`.
Maybe it"ll fit better have this check in `NewNetInfo` and save additional parsing.

go-controller/pkg/util/multi_network_test.go
JacobTanenbaum: yes sorry. This has been removed

test/e2e/network_segmentation.go
JacobTanenbaum: The networks are only cleaned up properly if all the pods on a network are removed before the network itself, otherwise artifacts are left in the database. In the future this can be removed when it is decided what pods should do when there network is removed.

ormergi: Make sense, thanks for explaining

test/e2e/network_segmentation.go
```diff
@@ -172,6 +176,8 @@ var _ = Describe("Network Segmentation", func() {
 						}, metav1.CreateOptions{})
 						Expect(err).NotTo(HaveOccurred())
 						defer func() {
+							err := cleanupPods(cs, defaultNetNamespace)
+							Expect(err).NotTo(HaveOccurred())
```
ormergi, line 180: nit: I would move it closer to where pod is create, line 220

JacobTanenbaum: while that would work because defer is a stack I thought it would look cleaner if it was done as part of cleaning up the namespace in general. It does not matter if there are actually pods in the namespace or not since we create it it should try deleting all the pods.

test/e2e/network_segmentation.go
ormergi: Does it make sense to also clean up pod in this test? https://github.com/ovn-org/ovn-kubernetes/blob/6859d3dc4e823de8bce7fe36fceaccf54ed09710/test/e2e/network_segmentation.go#L439

test/e2e/multihoming_utils.go
```diff
@@ -520,3 +520,18 @@ func getNetworkGateway(cli *client.Client, networkName string) (string, error) {
 
 	return "", fmt.Errorf("Gateway not found for network %q", networkName)
 }
+
+// in order to ensure that Network Attached Definitions are cleaned up properly the pods must be removed first
+func cleanupPods(cs clientset.Interface, namespace string) error {
```
tssurya, line 541: you can move this to network_segmentation.go itself since this isn't exactly part of multi homing feature

go-controller/pkg/util/multi_network.go
JacobTanenbaum: moved the code good suggestion

test/e2e/multihoming_utils.go
JacobTanenbaum: moved

test/e2e/network_segmentation.go
JacobTanenbaum: yes, added the cleanup

tssurya:
@JacobTanenbaum we can drop commit1 because we landed https://github.com/ovn-org/ovn-kubernetes/pull/4631

test/e2e/network_segmentation.go
```diff
@@ -932,6 +942,19 @@ spec:
...
+	if isIPv6Supported() && isIPv4Supported() {
+		cidr = `
+    - cidr: 10.20.100.0/16
+    - cidr: 2014:100:200::0/60
+`
+	} else if isIPv6Supported() {
+		cidr = `
+    - cidr: 2014:100:200::0/60
+`
+	}
```
ormergi, line 957: nit: I would export this to a helper function

tssurya:
this needs a rebase now and also we don't need commit1: https://github.com/ovn-org/ovn-kubernetes/pull/4625/commits/8320f4eee3a5620083029d4d87367bec23485d8b

we fixed it via the PR I linked: https://github.com/ovn-org/ovn-kubernetes/pull/4625#issuecomment-2298317820
@JacobTanenbaum not sure if you missed my comment

BTW are you going to be doing the conversion of https://github.com/ovn-org/ovn-kubernetes/blob/3c38f80fb8e5b064e931ca31149e47360fd208da/test/e2e/network_segmentation.go#L686 and https://github.com/ovn-org/ovn-kubernetes/blob/3c38f80fb8e5b064e931ca31149e47360fd208da/test/e2e/network_segmentation.go#L382 to uss CRDs as well in this PR?

test/e2e/network_segmentation.go
```diff
@@ -758,33 +759,42 @@ var _ = Describe("Network Segmentation", func() {
 				podAnno, err := unmarshalPodAnnotation(updatedPod.Annotations, f.Namespace.Name+"/"+userDefinedNetworkName)
 				Expect(err).NotTo(HaveOccurred())
 				framework.Logf("Client pod's annotation for network %s is %v", userDefinedNetworkName, podAnno)
-				Expect(len(podAnno.Routes)).To(Equal(6)) // 3 v4 routes + 3 v6 routes for UDN
+				// number of routes is dependaant on the ip familys supported
+				if isIPv6Supported() && isIPv4Supported() {
+					Expect(len(podAnno.Routes)).To(Equal(6)) // 3 v4 routes + 3 v6 routes for UDN
+				} else {
+					Expect(len(podAnno.Routes)).To(Equal(3)) // either 3 v4 routes or 3 v6 routes for UDN
+				}
```
ormergi, line 767: Please consider exporting this to a function, maybe calling it "assertPodAnnotationRoutes"

test/e2e/network_segmentation.go
```diff
@@ -758,33 +759,42 @@ var _ = Describe("Network Segmentation", func() {
...
-				By("asserting the *client* pod can contact the server's v6 IP located outside the cluster")
-				Eventually(func() error {
-					return connectToServer(clientPodConfig, externalIpv6, port)
-				}, 2*time.Minute, 6*time.Second).Should(Succeed())
+				if isIPv6Supported() {
+					By("asserting the *client* pod can contact the server's v6 IP located outside the cluster")
+					Eventually(func() error {
+						return connectToServer(clientPodConfig, externalIpv6, port)
+					}, 2*time.Minute, 6*time.Second).Should(Succeed())
+				}
```
ormergi, line 781: Please consider export this code to a function, maybe calling it "assertClientExternalConnectivity"

test/e2e/network_segmentation.go
```diff
@@ -691,6 +691,7 @@ var _ = Describe("Network Segmentation", func() {
 		Expect(actualConditions[0].Status).To(Equal(metav1.ConditionFalse))
 		Expect(actualConditions[0].Reason).To(Equal("SyncError"))
 		expectedMessage := fmt.Sprintf("primary network already exist in namespace %q: %q", f.Namespace.Name, primaryNadName)
+		fmt.Printf("actualCondition[0].Message: %s\n", actualConditions[0].Message)
```
ormergi, line 694: Is it a leftover?

test/e2e/multihoming.go
```diff
@@ -91,35 +92,38 @@ var _ = Describe("Multi Homing", func() {
...
+							// ip was found in one of the network ranges
+							foundRange = true
+							break
+						}
+					}
+					Expect(foundRange).To(BeTrue())
+					for _, excludedRange := range strings.Split(netConfig.excludeCIDRs, ",") {
+						Expect(inRange(excludedRange, podIP)).To(
+							MatchError(fmt.Errorf("ip [%s] is NOT in range %s", podIP, excludedRange)))
+					}
```
ormergi, line 115: Asking to undersntad, this code is now necessary because `cidr` could end up having multiple CIDRs?

I think it would make the test easier to follow if this code or some part of it to function.

go-controller/pkg/util/multi_network.go
```diff
@@ -682,17 +682,30 @@ func NewNetInfo(netconf *ovncnitypes.NetConf) (NetInfo, error) {
...
 	}
+	if err != nil {
+		return nil, err
+	}
+	ipv4Mode, ipv6Mode := ni.IPMode()
+	if ipv4Mode && !config.IPv4Mode {
+		return nil, fmt.Errorf("network %s is attempting to use ipv4 subnets, which the cluster does not support", ni.GetNetworkName())
+	}
+	if ipv6Mode && !config.IPv6Mode {
+		return nil, fmt.Errorf("network %s is attempting to use ipv6 subnets, which the cluster does not support", ni.GetNetworkName())
```
ormergi, line 706: nit: how about the following rephrasing "network %q specifies IPv4 subnets but the cluster doesnt support IPv4"

go-controller/pkg/util/multi_network_test.go
```diff
@@ -469,6 +469,7 @@ func TestParseNetconf(t *testing.T) {
 				t.Skip(test.unsupportedReason)
 			}
 			g := gomega.NewWithT(t)
+
```
ormergi, line 472: Is it a leftover?

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
```
ormergi, line 13: nit: I would move this import to its own new line below udnv1 import, because its internal package and not something external or internal API

ormergi (commented):
Thanks for the changes, please see below and inline  comments

Regarding " validate the ipfamily for network attachments " commit:
Could you please explain why is it necessary to import and change "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config" fields?
Please consider adding some unit-test that verifies NewNetInfo raise an error as expected. At least for go-controller/pkg/util/multi_network_test.go

Regarding "working towards making the e2e tests cluster ip family aware " commit: 
Could you please explain in the commit message why this change is necessary and how its being addressed?
In addition I think its worth referring to tests that were changed and test that were removed following this commit changes.

test/e2e/multihoming.go
maiqueb, line 1: I'm not sure these changes make sense in this scope ...

For multi-homing we usually don't care about what the cluster supports, since these are disconnected clusters. 

Meaning, it makes no sense to prevent the user from using IPv6 in their isolated overlays because the cluster in configured in IPv4 only mode.

This comment only applies to **secondary** interfaces (which are tested in this file). Not primary UDNs.

go-controller/pkg/util/multi_network.go
```diff
@@ -682,17 +682,34 @@ func NewNetInfo(netconf *ovncnitypes.NetConf) (NetInfo, error) {
...
 	default:
 		// other topology NAD can be supported later
 		return nil, fmt.Errorf("topology %s not supported", netconf.Topology)
 	}
+	if err != nil {
+		return nil, err
+	}
+
+	// secondary networks can support ipv4 and ipv6 regardless of the underlying cluster
+	if !ni.IsSecondary() {
```
maiqueb, line 703: @JacobTanenbaum I think a primary network also returns true here ...

IIUC, we would want to return true **only** for secondary nets - i.e. `isSecondary && role == primary`. 

... I know, the name for secondary is **very** misleading. I think @trozet meant to push something to address this ?..

test/e2e/network_segmentation.go
JacobTanenbaum: done

test/e2e/network_segmentation.go
JacobTanenbaum: you already did

test/e2e/network_segmentation.go
JacobTanenbaum: done

test/e2e/network_segmentation.go
JacobTanenbaum: yes, I removed it

test/e2e/multihoming.go
JacobTanenbaum: This is no longer needed, I was unaware that secondary networks can have any IP family regardless of the underlying cluster

go-controller/pkg/util/multi_network.go
JacobTanenbaum: done

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
JacobTanenbaum: no longer required

test/e2e/multihoming.go
JacobTanenbaum: removed the changes to multihoming, was unaware that secondary networks can have both IPv4 and IPv6 regardless of the underlying cluster

go-controller/pkg/util/multi_network.go
JacobTanenbaum: added the check for `IsPrimary()` would that work?

go-controller/pkg/util/multi_network_test.go
JacobTanenbaum: yes it was, sorry.

JacobTanenbaum:
test failure appears to be a flake: https://github.com/ovn-org/ovn-kubernetes/issues/4480

go-controller/pkg/clustermanager/pod/allocator_test.go
```diff
@@ -637,6 +639,7 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 
 			err = a.reconcile(old, new, tt.args.release)
 			if len(tt.expectError) > 0 {
+				fmt.Printf("tt.name: %s\n", tt.name)
 				g.Expect(err).To(gomega.MatchError(gomega.ContainSubstring(tt.expectError)))
```
maiqueb, line 643:
```suggestion
g.Expect(err).To(gomega.MatchError(gomega.ContainSubstring(tt.expectError)), fmt.Sprintf("tt.name: %s", tt.name))
```

Not sure it's worth it.

go-controller/pkg/clustermanager/secondary_network_unit_test.go
```diff
@@ -701,6 +684,20 @@ var _ = ginkgo.Describe("Cluster Controller Manager", func() {
...
+							Subnets:  subnets,
+							Topology: ovntypes.Layer2Topology,
+						})
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					fakeClient = &util.OVNClusterManagerClientset{
+						KubeClient:            fake.NewSimpleClientset(&v1.NodeList{Items: nodes()}),
+						IPAMClaimsClient:      fakeipamclaimclient.NewSimpleClientset(),
+						NetworkAttchDefClient: fakenadclient.NewSimpleClientset(),
+					}
```
maiqueb, line 700: I don't follow why we move this here, sorry.

go-controller/pkg/util/multi_network.go
```diff
@@ -682,17 +682,32 @@ func NewNetInfo(netconf *ovncnitypes.NetConf) (NetInfo, error) {
...
-		return newLocalnetNetConfInfo(netconf)
+		ni, err = newLocalnetNetConfInfo(netconf)
 	default:
 		// other topology NAD can be supported later
 		return nil, fmt.Errorf("topology %s not supported", netconf.Topology)
 	}
+	if err != nil {
+		return nil, err
+	}
+	if ni.IsPrimaryNetwork() && ni.IsSecondary() {
```
maiqueb, line 701: OK, so here we are ensuring this is a primary UDN, right ? i.e. a primary network that is *not* the cluster default network.

It makes sense, tks.

test/e2e/multihoming.go
```diff
@@ -863,7 +864,7 @@ var _ = Describe("Multi Homing", func() {
 								[]mnpapi.MultiPolicyType{mnpapi.PolicyTypeIngress, mnpapi.PolicyTypeEgress},
 								nil,
 								[]mnpapi.MultiNetworkPolicyEgressRule{
-									{},
+									mnpapi.MultiNetworkPolicyEgressRule{},
```
maiqueb, line 867: why do we need this ? 

Maybe isolate this in a separate commit providing the motivation in the commit msg ?

maiqueb (commented):
I need to leave now, and I honestly didn't understand what's happening in the last commit, where I think you test both NAD & UserDefinedNetwork CRD options. 

I can approve the first two commits, but I need someone else to look at the top most commit.

maiqueb (changes requested):
It's very hard for me to understand this new table depth of using the NAD / UDN CRD.

But I don't think I see your new tests invoking the function that creates the parametrized network configuration holder (i.e. either the NAD or the UDN CRD), as I do for example for the [isolation tests](https://github.com/ovn-org/ovn-kubernetes/blob/d95389bae4287f3db6f33c8e72198d3930429fd1/test/e2e/network_segmentation.go#L181). 

What I do see them doing (always) is [creating NADs](https://github.com/ovn-org/ovn-kubernetes/blob/d95389bae4287f3db6f33c8e72198d3930429fd1/test/e2e/network_segmentation.go#L412).

... I could be wrong.

JacobTanenbaum:
@maiqueb good spot, the issue is fixed and it should use the networkCreateFn like the other tests do

maiqueb (approved):
Thank you !

Good work. Sorry it took so long for me to properly review.

trozet (approved):

PR #4631, opened 2024-08-16
Fixes deleting CNI DEL with UDN

trozet:
The previous implementation had several issues:
 - If a NAD was removed while pods were active, CNI DEL would fail.
 - If a pod was force removed, CNI DEL would fail.

The above issues were due to:
 - Searching for the NAD to be able to do the CNI DEL
 - Searching for the pod in the informer cache for CNI DEL

This PR simplifies deletion by not relying on the pod object or the NAD
object. When primary CNI deletion happens with this patch, we simply
find any interfaces that belong to this pod's sandbox ID. There is no
performance impact, because clearPodBandwidth was already finding all of
these ports anyway.

CNI deletion happens in reverse order, so we can be sure that secondary
interfaces have already been deleted by the time we search and delete
interfaces for the primary CNI.

Signed-off-by: Tim Rozet <trozet@redhat.com>

### DIFF
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -606,7 +606,7 @@ func (*defaultPodRequestInterfaceOps) ConfigureInterface(pr *PodRequest, getter 
 	if !ifInfo.IsDPUHostMode {
 		err = ConfigureOVS(pr.ctx, pr.PodNamespace, pr.PodName, hostIface.Name, ifInfo, pr.SandboxID, pr.CNIConf.DeviceID, getter)
 		if err != nil {
-			pr.deletePorts(hostIface.Name, pr.PodNamespace, pr.PodName)
+			pr.deletePort(hostIface.Name, pr.PodNamespace, pr.PodName)
 			return nil, err
 		}
 	}
@@ -666,6 +666,7 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 	}
 
 	ifnameSuffix := ""
+	isSecondary := pr.netName != types.DefaultNetworkName
 	// nothing needs to be done for the VFIO case in the container namespace
 	if !pr.IsVFIO {
 		netns, err := ns.GetNS(pr.Netns)
@@ -683,7 +684,6 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 		// 1. For SRIOV case, we'd need to move device from container namespace back to the host namespace
 		// 2. If it is secondary network and not dpu-host mode, then get the container interface index
 		//    so that we know the host-side interface name.
-		isSecondary := pr.netName != types.DefaultNetworkName
 		err = netns.Do(func(_ ns.NetNS) error {
 			// container side interface deletion
 			link, err := util.GetNetLinkOps().LinkByName(pr.IfName)
@@ -728,7 +728,11 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 	if !ifInfo.IsDPUHostMode {
 		var err error
 		// host side interface deletion
-		hostIfName := pr.SandboxID[:(15-len(ifnameSuffix))] + ifnameSuffix
+		var hostIfName string
+		if !util.IsNetworkSegmentationSupportEnabled() || isSecondary {
+			// this is a secondary network (not primary) or segmentation is not enabled
+			hostIfName = pr.SandboxID[:(15-len(ifnameSuffix))] + ifnameSuffix
+		}
 		if pr.CNIConf.DeviceID != "" {
 			hostIfName, err = util.GetFunctionRepresentorName(pr.CNIConf.DeviceID)
 			if err != nil {
@@ -736,10 +740,27 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 					pr.CNIConf.DeviceID, podDesc, err)
 			}
 		}
-		if hostIfName != "" {
-			pr.deletePorts(hostIfName, pr.PodNamespace, pr.PodName)
+		portList, err := ovsFind("interface", "name", "external-ids:sandbox="+pr.SandboxID)
+		if err != nil {
+			return fmt.Errorf("failed to list interfaces in OVS during delete for sandbox: %s, err: %w",
+				pr.SandboxID, err)
 		}
-		err = clearPodBandwidth(pr.SandboxID)
+		// hostIfName is not empty if using device ID, a secondary network, or segmentation not enabled
+		// delete the port in traditional fashion
+		if hostIfName != "" {
+			pr.deletePort(hostIfName, pr.PodNamespace, pr.PodName)
+		} else {
+			// this is a primary interface deletion and segmentation is enabled, delete all ports
+			// delete happens in reverse order for attached networks, so this is the final deletion
+			// In other words we dont have to worry about accidentally deleting a secondary network interface at
+			// this point.
+			if len(portList) > 1 {
+				klog.V(5).Infof("Removing multiple interfaces for primary network segmentation (%+v) %s: %s",
+					*pr, podDesc, strings.Join(portList, ","))
+			}
+			pr.deletePorts(portList, pr.PodNamespace, pr.PodName)
+		}
+		err = clearPodBandwidthForPorts(portList, pr.SandboxID)
 		if err != nil {
 			klog.Errorf("Failed to clearPodBandwidth sandbox %v %s: %v", pr.SandboxID, podDesc, err)
 		}
@@ -775,7 +796,7 @@ func (pr *PodRequest) deletePodConntrack() {
 	}
 }
 
-func (pr *PodRequest) deletePorts(ifaceName, podNamespace, podName string) {
+func (pr *PodRequest) deletePort(ifaceName, podNamespace, podName string) {
 	podDesc := fmt.Sprintf("%s/%s", podNamespace, podName)
 
 	out, err := ovsExec("del-port", "br-int", ifaceName)
@@ -788,5 +809,11 @@ func (pr *PodRequest) deletePorts(ifaceName, podNamespace, podName string) {
 		if err = util.LinkDelete(ifaceName); err != nil {
 			klog.Warningf("Failed to delete pod %q interface %s: %v", podDesc, ifaceName, err)
 		}
+	}
+}
+
+func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string) {
+	for _, iface := range ifaces {
+		pr.deletePort(iface, podNamespace, podName)
 	}
 }
--- a/go-controller/pkg/cni/bandwidth.go
+++ b/go-controller/pkg/cni/bandwidth.go
@@ -13,9 +13,13 @@ func clearPodBandwidth(sandboxID string) error {
 		return err
 	}
 
+	return clearPodBandwidthForPorts(portList, sandboxID)
+}
+
+func clearPodBandwidthForPorts(portList []string, sandboxID string) error {
 	// Clear the QoS for any ports of this sandbox
 	for _, port := range portList {
-		if err = ovsClear("port", port, "qos"); err != nil {
+		if err := ovsClear("port", port, "qos"); err != nil {
 			return err
 		}
 	}
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -256,27 +256,6 @@ func (pr *PodRequest) cmdDel(clientset *ClientSet) (*Response, error) {
 		if err != nil {
 			return nil, err
 		}
-
-		if util.IsNetworkSegmentationSupportEnabled() {
-			pod, err := clientset.getPod(pr.PodNamespace, pr.PodName)
-			if err != nil {
-				return nil, err
-			}
-			primaryUDN := udn.NewPrimaryNetwork(clientset.nadLister)
-			if err := primaryUDN.Ensure(namespace, pod.Annotations, pr.nadName); err != nil {
-				return nil, err
-			}
-			if primaryUDN.Found() {
-				primaryUDNPodRequest := pr.buildPrimaryUDNPodRequest(pod, primaryUDN)
-				primaryUDNPodInfo, err := primaryUDNPodRequest.buildPodInterfaceInfo(pod.Annotations, primaryUDN.Annotation(), primaryUDN.NetworkDevice())
-				if err != nil {
-					return nil, err
-				}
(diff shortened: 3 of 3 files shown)


trozet:
@tssurya FYI

trozet:
@qinqon FYI

tssurya:
shouldn't we do a test for this? we are also doing https://github.com/ovn-org/ovn-kubernetes/pull/4625/commits/a9a34ad511ce4e2d072d431063636ec508f5a7b7 to be safe its in the pipe

trozet:
looks like https://github.com/ovn-org/ovn-kubernetes/pull/3062 flake

trozet:
@martinkennelly ptal

go-controller/pkg/cni/helper_linux.go
```diff
@@ -728,18 +728,39 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
...
+				pr.SandboxID, err)
+		}
+		// hostIfName is not empty if using device ID, a secondary network, or segmentation not enabled
+		// delete the port in traditional fashion
 		if hostIfName != "" {
-			pr.deletePorts(hostIfName, pr.PodNamespace, pr.PodName)
+			pr.deletePort(hostIfName, pr.PodNamespace, pr.PodName)
+		} else {
+			// this is a primary interface deletion and segmentation is enabled, delete all ports
+			// delete happens in reverse order for attached networks, so this is the final deletion
```
tssurya, line 754: I don't follow the ordering logic mentioned here.
If I have a UDN pod, I get a delete pod:
are you saying its guaranteed ordering triggered for CNIAdd is going to be: (a) delete default interface (b) delete primary (c) delete secondary interface and thus CNIDel will be reverse so multus would have removed the secondary interfaces leaving behind only udn and default interfaces here?
I thought you are simply nuking all 3 interfaces together using deletePorts?

can you please add a test for this where we have a pod with bunch of interfaces and ordering is respected?
also a test for delete nad before pod case

maiqueb:
> The previous implementation had several issues:
> 
> * If a NAD was removed while pods were active, CNI DEL would fail.
> * If a pod was force removed, CNI DEL would fail.
> 
> The above issues were due to:
> 
> * Searching for the NAD to be able to do the CNI DEL
> * Searching for the pod in the informer cache for CNI DEL
> 
> This PR simplifies deletion by not relying on the pod object or the NAD object. When primary CNI deletion happens with this patch, we simply find any interfaces that belong to this pod's sandbox ID. There is no performance impact, because clearPodBandwidth was already finding all of these ports anyway.
> 
> CNI deletion happens in reverse order, so we can be sure that secondary interfaces have already been deleted by the time we search and delete interfaces for the primary CNI.

I am far from sure this is true ... CNI happens in reverse order for an attachment, meaning, it would be true for an attachment configured using a CNI chain.

... unfortunately, that statement does not hold for multiple attachments on a pod.

Multus is configured as the only CNI plugin in the cluster; what it'll do when processing a CNI DEL is to get the cluster default network detegate, then get all the delegates listed in the network selection elements, and invoke CNI DELs on each of the delegate plugins.

Trying to say that you'd remove all the interfaces for the secondary networks in the CNI DEL call for the primary UDN.

@dougbtv please keep me honest here.

> 
> Signed-off-by: Tim Rozet [trozet@redhat.com](mailto:trozet@redhat.com)

go-controller/pkg/cni/helper_linux.go
```diff
@@ -666,6 +666,7 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 	}
 
 	ifnameSuffix := ""
+	isSecondary := pr.netName != types.DefaultNetworkName
```
martinkennelly, line 669: will this only be true for ovn-k when we are hot plugging out a link interface? 
From my testing, we always have the default interface (eth0) and that interface/network is the first interface for CNI Del to tear down. We remove all the interfaces for that CNI DEL

go-controller/pkg/cni/helper_linux.go
```diff
@@ -728,18 +728,39 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
...
+			return fmt.Errorf("failed to list interfaces in OVS during delete for sandbox: %s, err: %w",
+				pr.SandboxID, err)
+		}
+		// hostIfName is not empty if using device ID, a secondary network, or segmentation not enabled
+		// delete the port in traditional fashion
 		if hostIfName != "" {
-			pr.deletePorts(hostIfName, pr.PodNamespace, pr.PodName)
+			pr.deletePort(hostIfName, pr.PodNamespace, pr.PodName)
+		} else {
+			// this is a primary interface deletion and segmentation is enabled, delete all ports
```
martinkennelly, line 753: it could be non primary CDN network interface and thats not a primary interface

go-controller/pkg/cni/helper_linux.go
martinkennelly, line 754: delete happens in reverse order? There will be just one CNI Del right.

go-controller/pkg/cni/helper_linux.go
martinkennelly, line 669: go docs for what secondary is

go-controller/pkg/cni/helper_linux.go
```diff
@@ -728,18 +728,39 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 	if !ifInfo.IsDPUHostMode {
 		var err error
 		// host side interface deletion
-		hostIfName := pr.SandboxID[:(15-len(ifnameSuffix))] + ifnameSuffix
+		var hostIfName string
+		if !util.IsNetworkSegmentationSupportEnabled() || isSecondary {
+			// this is a secondary network (not primary) or segmentation is not enabled
```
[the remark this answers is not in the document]
martinkennelly: I cannot think of a scenario where secondary is true when using ovn-k. @trozet can you help me?

go-controller/pkg/cni/helper_linux.go
dougbtv: for what it's worth, from a Multus + net-attach-def standpoint, it happens in reverse: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus.go#L448-L455

so from a multus standpoint this is dependent on the network selection element, e.g. what is the order of `k8s.v1.cni.cncf.io/networks` and the items in that list

dougbtv:
> Multus is configured as the only CNI plugin in the cluster; what it'll do when processing a CNI DEL is to get the cluster default network detegate, then get all the delegates listed in the network selection elements, and invoke CNI DELs on each of the delegate plugins.

> Trying to say that you'd remove all the interfaces for the secondary networks in the CNI DEL call for the primary UDN.

@maiqueb I actually thought the exact same thing until I looked it up when Tim had asked me about it, but... turns out it does actually do the attachments in reverse order.

Given a value of `k8s.v1.cni.cncf.io/networks: red, blue`

You'd essentially have three "delegates": default (is implied), red, blue.

When those are processed, they're deleted in reverse as: blue, red, default.

https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus.go#L448-L455

So, totally see how you got that, I thought the same exact thing until I bumped into this.

maiqueb:
> > Multus is configured as the only CNI plugin in the cluster; what it'll do when processing a CNI DEL is to get the cluster default network detegate, then get all the delegates listed in the network selection elements, and invoke CNI DELs on each of the delegate plugins.
> 
> > Trying to say that you'd remove all the interfaces for the secondary networks in the CNI DEL call for the primary UDN.
> 
> @maiqueb I actually thought the exact same thing until I looked it up when Tim had asked me about it, but... turns out it does actually do the attachments in reverse order.
> 
> Given a value of `k8s.v1.cni.cncf.io/networks: red, blue`
> 
> You'd essentially have three "delegates": default (is implied), red, blue.
> 
> When those are processed, they're deleted in reverse as: blue, red, default.
> 
> https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/pkg/multus/multus.go#L448-L455
> 
> So, totally see how you got that, I thought the same exact thing until I bumped into this.

Hm, OK ... 

Can we rely then on this implementation detail ?... It doesn't sound exactly healthy IMHO.

go-controller/pkg/cni/helper_linux.go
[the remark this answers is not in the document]
trozet: isSecondary means secondary. UnconfigureInterface wont be called for a primary UDN, so the only way the netName is not default is if its a secondary network. Secondary network can be used without NetworkSegmentation enabled by simply using EnableMultiNetwork. Then create a secondary network NAD like:

```
metadata:
  name: l2-network
  namespace: ns1
spec:
  config: |2
    {
            "cniVersion": "0.3.1",
            "name": "l2-network",
            "topology":"layer2",
            "subnets": "10.100.200.0/24",
            "mtu": 1300,
            "netAttachDefName": "ns1/l2-network",
            "excludeSubnets": "10.100.200.0/29"
    }


```

and then attach it to a pod like:

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    k8s.v1.cni.cncf.io/networks: l2-network
  name: tinypod
  namespace: ns1
spec:
  containers:
  - args:
    - pause
    image: registry.k8s.io/e2e-test-images/agnhost:2.36
    imagePullPolicy: IfNotPresent
    name: agnhost-container

```

go-controller/pkg/cni/helper_linux.go
```diff
@@ -728,18 +728,39 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
...
+		if err != nil {
+			return fmt.Errorf("failed to list interfaces in OVS during delete for sandbox: %s, err: %w",
+				pr.SandboxID, err)
+		}
+		// hostIfName is not empty if using device ID, a secondary network, or segmentation not enabled
+		// delete the port in traditional fashion
 		if hostIfName != "" {
-			pr.deletePorts(hostIfName, pr.PodNamespace, pr.PodName)
+			pr.deletePort(hostIfName, pr.PodNamespace, pr.PodName)
+		} else {
```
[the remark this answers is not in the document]
trozet: no, we will hit this if network segmentation is enabled and we are deleting the primary CNI. At that point we have to search for multiple interfaces in case this primary CNI is serving a primary UDN.

dougbtv:
At least in terms of history, I looked at it, and [it changed in 2018](https://github.com/k8snetworkplumbingwg/multus-cni/commit/61d212c59a09a0c90d36b36b8c8d13e3e0f648a9)

However! From a specification standpoint... I can't find a reference to CNI DEL what-so-ever. I think we omitted anything specific to ordering as we thought it would make it hard to specify where order was required or not, but... Yeah, implementation wise, we went with the reverse order.

Usually, the deletion order would seem "somewhat arbitrary" to me, that is, like most terminal/delgate CNI plugins wouldn't care what happens when other plugins have  DEL processed. Reverse order seems intuitive to me that it's like, LIFO, so the primary network is deleted last. But, it's only intuition I don't have a good argument one way or another tbh.

But... If we're finding cases where this does matter and we need to specify it differently, I'm definitely all ears, and it's definitely possible to figure out something that's overall more healthy.

maiqueb:
> At least in terms of history, I looked at it, and [it changed in 2018](https://github.com/k8snetworkplumbingwg/multus-cni/commit/61d212c59a09a0c90d36b36b8c8d13e3e0f648a9)
> 
> However! From a specification standpoint... I can't find a reference to CNI DEL what-so-ever. I think we omitted anything specific to ordering as we thought it would make it hard to specify where order was required or not, but... Yeah, implementation wise, we went with the reverse order.
> 
> Usually, the deletion order would seem "somewhat arbitrary" to me, that is, like most terminal/delgate CNI plugins wouldn't care what happens when other plugins have DEL processed. Reverse order seems intuitive to me that it's like, LIFO, so the primary network is deleted last. But, it's only intuition I don't have a good argument one way or another tbh.
> 
> But... If we're finding cases where this does matter and we need to specify it differently, I'm definitely all ears, and it's definitely possible to figure out something that's overall more healthy.



> At least in terms of history, I looked at it, and [it changed in 2018](https://github.com/k8snetworkplumbingwg/multus-cni/commit/61d212c59a09a0c90d36b36b8c8d13e3e0f648a9)
> 
> However! From a specification standpoint... I can't find a reference to CNI DEL what-so-ever. I think we omitted anything specific to ordering as we thought it would make it hard to specify where order was required or not, but... Yeah, implementation wise, we went with the reverse order.
> 
> Usually, the deletion order would seem "somewhat arbitrary" to me, that is, like most terminal/delgate CNI plugins wouldn't care what happens when other plugins have DEL processed. Reverse order seems intuitive to me that it's like, LIFO, so the primary network is deleted last. But, it's only intuition I don't have a good argument one way or another tbh.
> 
> But... If we're finding cases where this does matter and we need to specify it differently, I'm definitely all ears, and it's definitely possible to figure out something that's overall more healthy.

Do you think a spec update with this would make sense ? I would surely sleep better if we had the guarantees this would never change.

go-controller/pkg/cni/helper_linux.go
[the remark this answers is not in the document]
tssurya: so this shouldn't be called isSecondary; its confusing because in other parts of the code isSecondary means non-default networks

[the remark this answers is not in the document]
tssurya: i'm also fine adding "UnconfigureInterface wont be called for a primary UDN" this as a comment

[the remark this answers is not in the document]
trozet: I plan to fix all of the names of the controllers and remove the "secondary can mean secondary or primary" definitions in our code.

maiqueb (approved):
Thanks. I'd sleep better with some unit tests, but given our tight schedule, those can happen on a follow-up commit.

go-controller/pkg/cni/helper_linux.go
trozet: I meant multus order of operations. Deletion for multiple attachments happens in reverse order. I'll add another PR and clarify some of the comments and add some unit tests.

go-controller/pkg/cni/helper_linux.go
trozet: I guess I should say "primary CNI deletion" which deletes both primary UDN and cluster default interface

go-controller/pkg/cni/helper_linux.go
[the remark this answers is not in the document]
tssurya: hmm anytime net-seg is enabled this will be hit right? doesn't matter if we are deleting primary or secondary here right?

go-controller/pkg/cni/helper_linux.go
trozet: thats true we remove the eth0 and the primary UDN interface, but we dont remove secondary networks that are added using a network attachment. Those come in a different CNI request.

tssurya (commented):
lgtm

PR #4633, opened 2024-08-16
Node IP hander: ignore mgnt port IPs

martinkennelly:
Looking for better ideas to filter UDN mgnt port IPs but this is the simplest I can think of right now.

### DIFF
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -69,10 +69,10 @@ func newAddressManagerInternal(nodeName string, k kube.Interface, config *manage
 
 // updates the address manager with a new IP
 // returns true if there was an update
-func (c *addressManager) addAddr(ipnet net.IPNet) bool {
+func (c *addressManager) addAddr(ipnet net.IPNet, linkIndex int) bool {
 	c.Lock()
 	defer c.Unlock()
-	if !c.cidrs.Has(ipnet.String()) && c.isValidNodeIP(ipnet.IP) {
+	if !c.cidrs.Has(ipnet.String()) && c.isValidNodeIP(ipnet.IP, linkIndex) {
 		klog.Infof("Adding IP: %s, to node IP manager", ipnet)
 		c.cidrs.Insert(ipnet.String())
 		return true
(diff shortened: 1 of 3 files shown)


go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -388,6 +388,14 @@ func (c *addressManager) isValidNodeIP(addr net.IP) bool {
 			return false
 		}
 	}
+	// check CDN + UDN management ports
```
kyrtapz, line 391: We should only check this if udn is enabled?

kyrtapz (commented):
Looks good!
Unless we track ALL UDN networks in a central place I don't see a better way to ignore UDN mgnt IPs. We could potentially revisit this if we decide to do a NAD/UDN cache in the feature for the performance implications of parsing the NADs across ovn-k.

martinkennelly:
git diff of changes made:

```git
diff --git a/go-controller/pkg/node/node_ip_handler_linux.go b/go-controller/pkg/node/node_ip_handler_linux.go
index b0febb8f6..2508d14ce 100644
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -388,12 +388,14 @@ func (c *addressManager) isValidNodeIP(addr net.IP, linkIndex int) bool {
                        return false
                }
        }
-       // check CDN + UDN management ports
-       if mpLink, err := util.GetNetLinkOps().LinkByIndex(linkIndex); err != nil {
-               klog.Errorf("Unable to determine if link is an OVN management port for address %s and link index %d: %v", addr.String(), linkIndex, err)
-       } else {
-               if strings.HasPrefix(mpLink.Attrs().Name, types.K8sMgmtIntfNamePrefix) {
-                       return false
+       if util.IsNetworkSegmentationSupportEnabled() {
+               // check CDN + UDN management ports
+               if mpLink, err := util.GetNetLinkOps().LinkByIndex(linkIndex); err != nil {
+                       klog.Errorf("Unable to determine if link is an OVN management port for address %s and link index %d: %v", addr.String(), linkIndex, err)
+               } else {
+                       if strings.HasPrefix(mpLink.Attrs().Name, types.K8sMgmtIntfNamePrefix) {
+                               return false
+                       }
                }
        }
 
diff --git a/go-controller/pkg/node/node_ip_handler_linux_test.go b/go-controller/pkg/node/node_ip_handler_linux_test.go
index d9952058c..164fb5ecb 100644
--- a/go-controller/pkg/node/node_ip_handler_linux_test.go
+++ b/go-controller/pkg/node/node_ip_handler_linux_test.go
@@ -307,6 +307,8 @@ var _ = Describe("Node IP Handler tests", func() {
                })
 
                ovntest.OnSupportedPlatformsIt("doesn't allow OVN management port IPs", func() {
+                       config.OVNKubernetesFeature.EnableMultiNetwork = true
+                       config.OVNKubernetesFeature.EnableNetworkSegmentation = true
                        Expect(tc.ns.Do(func(netNS ns.NetNS) error {
                                mpLink := ovntest.AddLink(fmt.Sprintf("%s1234", ovntypes.K8sMgmtIntfNamePrefix))
                                return netlink.AddrAdd(mpLink, &netlink.Addr{LinkIndex: mpLink.Attrs().Index, Scope: unix.RT_SCOPE_UNIVERSE,

 

```

tssurya (approved):
I agree we could do better solutions for this but because there isn't a 1 place get all networks info available for us I don't see a simpler way out... for now let's get this in because its critical we don't reconcile ALL services everytime a UDN mgmtIP is added/deleted.

PR #4641, opened 2024-08-20
Make Route Manager API blocking

trozet:
This commit changes the behavior of route add and delete operations to be blocking and synchronous. If the route add/del operation fails an error is now returned to the caller, rather than sending the route to a channel to be later sync'ed by the route manager loop.

<


trozet:
@martinkennelly @tssurya fyi

martinkennelly (dismissed):

pperiyasamy (approved):

tssurya (approved):

PR #4647, opened 2024-08-22
Improve error reporting for UDN

npinaeva:
Cluster manager controllers may have errors related to the UDN network. For example, if Network's cluster subnet doesn't have enough node subnets for all existing nodes. This will cause pod creation error for the nodes, that were not allocated a subnet for a given UDN, and finding the real cause may be cumbersome.

I propose the following framework to report errors that are related to other objects (e.g. nodes in this case).
We use one Success condition, which is true when all related objects were handled fine (all nodes in this case), and is false if at least one error was. The error message in this case should report one failed node to simplify debugging in case events will disappear, but the condition will not be updated until reported node is fixed changed to limit the amount of api calls.
To provide per-related-object info (per-node error in this case) we can use events, that will be related to the UDN, but will contain the related object type+name. If a related object (node) is listed in the UDN condition as failed, the latest event will reflect error message.

Here is a part of ` k describe userdefinednetwork` (##<string> are my comments, not a real output)
```
Status:
  Conditions:
    Last Transition Time:  2024-09-13T13:18:05Z
    Message:               NetworkAttachmentDefinition has been created
    Reason:                NetworkAttachmentDefinitionReady
    Status:                True
    Type:                  NetworkReady
    Last Transition Time:  2024-09-13T13:18:05Z
## generic message with only 1 node name
    Message:             Network allocation failed for at least one node: udn-worker2, check UDN events for more info.
    Reason:                InternalError
    Status:                False
    Type:                  NetworkAllocationSucceeded
Events:
  Type     Reason                   Age   From          Message
  ----     ------                   ----  ----          -------
## node-specific event
  Warning  NetworkAllocationFailed  4s    controlplane  Error occurred for node udn-worker2: Node add failed: error allocating network for node udn-worker2: no subnets available
```
To make this work:
1. add UDN controller interface to allow other subsystems reporting their conditions
2. use NetworkStatusReporter interface to report errors from the `network_cluster_controller` for node handling failures.

### DIFF
--- a/go-controller/pkg/clustermanager/network_cluster_controller.go
+++ b/go-controller/pkg/clustermanager/network_cluster_controller.go
@@ -8,9 +8,8 @@ import (
 	"reflect"
 	"sync"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cache "k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/retry"
@@ -31,6 +30,8 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
+
+type NetworkStatusReporter func(networkName string, fieldManager string, condition *metav1.Condition, events ...*util.EventDetails) error
 
 // networkClusterController is the cluster controller for the networks. An
 // instance of this struct is expected to be created for each network. A network
@@ -67,10 +68,21 @@ type networkClusterController struct {
 	// event recorder used to post events to k8s
 	recorder record.EventRecorder
 
+	statusReporter NetworkStatusReporter
+
+	// nodeName: errMessage
+	nodeErrors     map[string]string
+	nodeErrorsLock sync.Mutex
+	// Error condition only reports one of the failed nodes.
+	// To avoid changing that error report with every update, we store reported error node.
+	reportedErrorNode string
+
 	util.NetInfo
 }
 
-func newNetworkClusterController(networkIDAllocator idallocator.NamedAllocator, netInfo util.NetInfo, ovnClient *util.OVNClusterManagerClientset, wf *factory.WatchFactory, recorder record.EventRecorder, nadController *networkAttachDefController.NetAttachDefinitionController) *networkClusterController {
+func newNetworkClusterController(networkIDAllocator idallocator.NamedAllocator, netInfo util.NetInfo,
+	ovnClient *util.OVNClusterManagerClientset, wf *factory.WatchFactory, recorder record.EventRecorder,
+	nadController *networkAttachDefController.NetAttachDefinitionController, errorReporter NetworkStatusReporter) *networkClusterController {
 	kube := &kube.KubeOVN{
 		Kube: kube.Kube{
 			KClient: ovnClient.KubeClient,
@@ -89,6 +101,9 @@ func newNetworkClusterController(networkIDAllocator idallocator.NamedAllocator, 
 		networkIDAllocator: networkIDAllocator,
 		recorder:           recorder,
 		nadController:      nadController,
+		statusReporter:     errorReporter,
+		nodeErrors:         make(map[string]string),
+		nodeErrorsLock:     sync.Mutex{},
 	}
 
 	return ncc
@@ -108,7 +123,7 @@ func newDefaultNetworkClusterController(netInfo util.NetInfo, ovnClient *util.OV
 	}
 
 	namedIDAllocator := networkIDAllocator.ForName(types.DefaultNetworkName)
-	return newNetworkClusterController(namedIDAllocator, netInfo, ovnClient, wf, recorder, nil)
+	return newNetworkClusterController(namedIDAllocator, netInfo, ovnClient, wf, recorder, nil, nil)
 }
 
 func (ncc *networkClusterController) hasPodAllocation() bool {
@@ -147,6 +162,11 @@ func (ncc *networkClusterController) allowPersistentIPs() bool {
 }
 
 func (ncc *networkClusterController) init() error {
+	// report no errors on restart, then propagate any new errors by the started handlers
+	if err := ncc.resetStatus(); err != nil {
+		return fmt.Errorf("failed to reset network status: %w", err)
+	}
+
 	networkID, err := ncc.networkIDAllocator.AllocateID()
 	if err != nil {
 		return err
@@ -200,6 +220,96 @@ func (ncc *networkClusterController) init() error {
 	}
 
 	return nil
+}
+
+// updateNetworkStatus allows to report a status for networkClusterController's network via a UDN status condition
+// of type "NetworkAllocationSucceeded", if the network was created by UDN.
+// When at least one node reports an error, condition will be set to false and an event with node-specific error will be
+// generated.
+// Call this function after every node event handling, set handlerErr to nil to report no error.
+// There are potential optimization to when an error should be reported, see https://github.com/ovn-org/ovn-kubernetes/pull/4647#discussion_r1763352619.
+func (ncc *networkClusterController) updateNetworkStatus(nodeName string, handlerErr error) error {
+	if ncc.statusReporter == nil {
+		return nil
+	}
+	errorMsg := ""
+	if handlerErr != nil {
+		errorMsg = handlerErr.Error()
+	}
+
+	ncc.nodeErrorsLock.Lock()
+	defer ncc.nodeErrorsLock.Unlock()
+	if ncc.nodeErrors[nodeName] == errorMsg {
+		// error message didn't change for that node, no need to update
+		return nil
+	}
+
+	reportedErrorNode := ncc.reportedErrorNode
+	if ncc.reportedErrorNode == "" && errorMsg != "" {
+		reportedErrorNode = nodeName
+	}
+	if ncc.reportedErrorNode == nodeName && errorMsg == "" {
+		// error for this node is fixed, report next error node
+		reportedErrorNode = ""
+		for errorNode := range ncc.nodeErrors {
+			if errorNode != nodeName {
+				reportedErrorNode = errorNode
+				break
+			}
+		}
+	}
+
+	var condition *metav1.Condition
+	if reportedErrorNode != ncc.reportedErrorNode {
+		// We know condition only changes if ncc.reportedErrorNode value changes.
+		// Otherwise, condition will stay nil and the error message will be reflected in an event.
+		condition = getNetworkAllocationUDNCondition(reportedErrorNode)
+	}
+	events := make([]*util.EventDetails, 0, 1)
+	if errorMsg != "" {
+		events = append(events, &util.EventDetails{
+			EventType: util.EventTypeWarning,
+			Reason:    "NetworkAllocationFailed",
+			Note:      fmt.Sprintf("Error occurred for node %s: %s", nodeName, errorMsg),
+		})
+	}
+
+	netName := ncc.NetInfo.GetNetworkName()
+	if err := ncc.statusReporter(netName, "NetworkClusterController", condition, events...); err != nil {
+		return fmt.Errorf("failed to report network status: %w", err)
+	}
+	ncc.nodeErrors[nodeName] = errorMsg
+	ncc.reportedErrorNode = reportedErrorNode
+
+	return nil
+}
+
+// resetStatus should be called on startup before any handler is started to avoid status race.
+func (ncc *networkClusterController) resetStatus() error {
+	if ncc.statusReporter == nil {
+		return nil
+	}
+	netName := ncc.NetInfo.GetNetworkName()
+	return ncc.statusReporter(netName, "NetworkClusterController", getNetworkAllocationUDNCondition(""))
+}
+
+// We only report one failed node in condition to avoid too long messages and too many condition updates.
+// The node to be reported is passed as errorNode, if empty, all nodes are considered to be succeeded.
+func getNetworkAllocationUDNCondition(errorNode string) *metav1.Condition {
+	condition := &metav1.Condition{
+		Type:               "NetworkAllocationSucceeded",
+		LastTransitionTime: metav1.Now(),
+	}
+	if errorNode == "" {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = "NetworkAllocationSucceeded"
+		condition.Message = "Network allocation succeeded for all synced nodes."
+	} else {
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = "InternalError"
+		condition.Message = fmt.Sprintf("Network allocation failed for at least one node: %v, check UDN events for more info.", errorNode)
+	}
+	return condition
 }
 
 // Start the network cluster controller. Depending on the cluster configuration
@@ -322,12 +432,17 @@ func (h *networkClusterControllerEventHandler) AddResource(obj interface{}, from
 		if !ok {
 			return fmt.Errorf("could not cast %T object to *corev1.Node", obj)
 		}
-		if err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node); err != nil {
-			klog.Infof("Node add failed for %s, will try again later: %v",
-				node.Name, err)
-			return err
+		err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node)
+		if err == nil {
+			h.clearInitialNodeNetworkUnavailableCondition(node)
 		}
-		h.clearInitialNodeNetworkUnavailableCondition(node)
+		statusErr := h.ncc.updateNetworkStatus(node.Name, err)
+		joinedErr := errors.Join(err, statusErr)
+		if joinedErr != nil {
+			klog.Infof("Node add failed for %s, will try again later: %v",
+				node.Name, joinedErr)
+			return joinedErr
+		}
 	case factory.IPAMClaimsType:
 		return nil
 	default:
@@ -361,12 +476,17 @@ func (h *networkClusterControllerEventHandler) UpdateResource(oldObj, newObj int
 		if !ok {
 			return fmt.Errorf("could not cast %T object to *corev1.Node", newObj)
 		}
-		if err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node); err != nil {
+		err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node)
+		if err == nil {
+			h.clearInitialNodeNetworkUnavailableCondition(node)
+		}
+		statusErr := h.ncc.updateNetworkStatus(node.Name, err)
+		joinedErr := errors.Join(err, statusErr)
+		if joinedErr != nil {
 			klog.Infof("Node update failed for %s, will try again later: %v",
 				node.Name, err)
 			return err
 		}
-		h.clearInitialNodeNetworkUnavailableCondition(node)
 	case factory.IPAMClaimsType:
 		return nil
 	default:
@@ -393,7 +513,9 @@ func (h *networkClusterControllerEventHandler) DeleteResource(obj, cachedObj int
 		if !ok {
 			return fmt.Errorf("could not cast obj of type %T to *knet.Node", obj)
 		}
-		return h.ncc.nodeAllocator.HandleDeleteNode(node)
+		err := h.ncc.nodeAllocator.HandleDeleteNode(node)
+		statusErr := h.ncc.updateNetworkStatus(node.Name, err)
+		return errors.Join(err, statusErr)
 	case factory.IPAMClaimsType:
 		ipamClaim, ok := obj.(*ipamclaimsapi.IPAMClaim)
 		if !ok {
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -14,6 +14,8 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	corev1informer "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/reference"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -26,6 +28,7 @@ import (
 	userdefinednetworkv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1"
 	udnapplyconfkv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/applyconfiguration/userdefinednetwork/v1"
 	userdefinednetworkclientset "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/clientset/versioned"
+	userdefinednetworkscheme "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/clientset/versioned/scheme"
 	userdefinednetworkinformer "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/informers/externalversions/userdefinednetwork/v1"
 	userdefinednetworklister "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/listers/userdefinednetwork/v1"
 
@@ -54,6 +57,7 @@ type Controller struct {
 	podInformer corev1informer.PodInformer
 
 	networkInUseRequeueInterval time.Duration
+	eventRecorder               record.EventRecorder
 }
 
 const defaultNetworkInUseCheckInterval = 1 * time.Minute
@@ -65,6 +69,7 @@ func New(
 	udnInformer userdefinednetworkinformer.UserDefinedNetworkInformer,
 	renderNadFn RenderNetAttachDefManifest,
 	podInformer corev1informer.PodInformer,
+	eventRecorder record.EventRecorder,
 ) *Controller {
 	udnLister := udnInformer.Lister()
 	c := &Controller{
@@ -75,6 +80,7 @@ func New(
 		renderNadFn:                 renderNadFn,
 		podInformer:                 podInformer,
 		networkInUseRequeueInterval: defaultNetworkInUseCheckInterval,
+		eventRecorder:               eventRecorder,
 	}
 	cfg := &controller.ControllerConfig[userdefinednetworkv1.UserDefinedNetwork]{
 		RateLimiter:    workqueue.DefaultControllerRateLimiter(),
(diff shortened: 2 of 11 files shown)


go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -358,3 +358,60 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
+		condition := metav1.Condition{
+			Type:               conditionType,
+			Status:             metav1.ConditionTrue,
+			Reason:             "SubsystemError",
+			Message:            errorMessage,
+			LastTransitionTime: metav1.Now(),
+		}
+		udnStatus = udnStatus.WithConditions(condition)
+		for _, statusCondition := range udn.Status.Conditions {
+			if statusCondition.Type == condition.Type && statusCondition.Message == condition.Message {
```
jcaamano, line 388: I think the comparison `statusCondition.Message == condition.Message` should be a concern of the caller.

test/e2e/network_segmentation.go
```diff
@@ -558,95 +556,98 @@ var _ = Describe("Network Segmentation", func() {
...
-		Context("pod connected to UserDefinedNetwork", func() {
-			const testPodName = "test-pod-udn"
-
-			var (
-				udnInUseDeleteTimeout = 65 * time.Second
-				deleteNetworkTimeout  = 5 * time.Second
-				deleteNetworkInterval = 1 * time.Second
-			)
-
+		Context("with L2 secondary network", func() {
```
jcaamano, line 560: nit: `for L2 secondary network`?

test/scripts/e2e-cp.sh
```diff
@@ -124,7 +124,7 @@ fi
 
 # Only run Node IP/MAC address migration tests if they are explicitly requested
 IP_MIGRATION_TESTS="Node IP and MAC address migration"
-if [ "${WHAT}" != "${IP_MIGRATION_TESTS}" ]; then
+if [[ "${WHAT}" != "${IP_MIGRATION_TESTS}"* ]]; then
```
jcaamano, line 127: Why do we need these changes?

dist/templates/k8s.ovn.org_userdefinednetworks.yaml.j2
```diff
@@ -287,6 +287,9 @@ spec:
                   - type
                   type: object
                 type: array
+                x-kubernetes-list-map-keys:
+                - type
+                x-kubernetes-list-type: map
```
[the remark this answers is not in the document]
npinaeva: yep, this part was actually merged separately for that reason https://github.com/ovn-org/ovn-kubernetes/pull/4664

go-controller/pkg/crd/userdefinednetwork/v1/types.go
```diff
@@ -225,5 +225,7 @@ type UserDefinedNetworkList struct {
 
 // UserDefinedNetworkStatus contains the observed status of the UserDefinedNetwork.
 type UserDefinedNetworkStatus struct {
+	// +listType=map
+	// +listMapKey=type
```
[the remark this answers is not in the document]
npinaeva: so strategic merge is not supported for CRDs, which is why I didn't add related tags
some detail from gateway api https://github.com/kubernetes-sigs/gateway-api/issues/342

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -358,3 +358,60 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
+	udnNamespace, udnName := template.ParseNetworkName(networkName)
+	if udnName == "" {
+		return nil
+	}
+	udn, err := c.udnClient.K8sV1().UserDefinedNetworks(udnNamespace).Get(context.Background(), udnName, metav1.GetOptions{})
+	if err != nil {
+		return nil
+	}
+	udnStatus := udnapplyconfkv1.UserDefinedNetworkStatus()
+	conditionType := fmt.Sprintf("SubsystemError.%s", fieldManager)
```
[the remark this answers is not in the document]
npinaeva: Currently we report zone status with conditions with type `Ready-In-Zone+<zoneName>` and I think this is the only condition we have for now.
It looks like we have 2 condition types for now: readiness and errors. Readiness is reported always and may contain error message if status if false. We can use those for per-zone readiness reports and for subsystem reports (if needed). Error condition is only reported when an error occurred and is not added otherwise. I think Error conditions make sense in case we have many subsystems reporting errors in a status and we don't want to overload the status with "no error" messages.
If that makes sense, every subsystem should only report one type of condition: readiness or error. In that case
`<fieldmanager>.k8s.ovn.org/<type>` works. But if we ever get another condition type or want the same fieldManager to report 2 types, we may need to switch to `type=<subsystem>.k8s.ovn.org/<type>` + `fieldmanager = <subsystem.type>` in case subsystem wants to update those conditions separately with different filed managers. Probably too much for now, I can go with single-type assumption

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -358,3 +358,60 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
+		for _, statusCondition := range udn.Status.Conditions {
+			if statusCondition.Type == condition.Type && statusCondition.Message == condition.Message {
+				return nil
+			}
+		}
+	} else {
+		conditionFound := false
+		for _, statusCondition := range udn.Status.Conditions {
+			if statusCondition.Type == conditionType {
+				conditionFound = true
```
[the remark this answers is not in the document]
npinaeva: nice!

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -204,6 +220,47 @@ func (ncc *networkClusterController) init() error {
...
+	} else {
+		ncc.nodeErrors[nodeName] = errorMsg
+	}
+	// To provide per-node error, the following message format is used:
+	// <nodeName1>: <error message1>;
+	// <nodeName2>: <error message2>
+	//
+	// If no errors are reported, condition will not be present in the status.
+	joinedMsg := ""
+	for nodeName, errMsg := range ncc.nodeErrors {
```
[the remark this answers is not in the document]
npinaeva: hehe I thought about this! if the message for the node didn't change, we won't even get here (returns earlier). If the message has changed, we need to update status anyway.

test/scripts/e2e-cp.sh
npinaeva: that allows you running tests with WHAT="full test spec". e.g. now if you call it with
WHAT="Network Segmentation a user defined primary network created using NetworkAttachmentDefinitions"
it won't get executed, because it searched for an exact "Network Segmentation" match.
I use it often locally to run specific tests and I have already done this for some tests https://github.com/ovn-org/ovn-kubernetes/blob/master/test/scripts/e2e-cp.sh#L136-L142

npinaeva:
first push - changes, second - master rebase

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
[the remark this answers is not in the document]
jcaamano: Not sure I get you. If we want the same fieldManager to report 2 types, whats wrong with this scheme:
```
fieldmanager1.k8s.ovn.org/type1
fieldmanager1.k8s.ovn.org/type2
```
I am assuming `subsystem == field manager`

Is the main cluster manager UDN reconciler not reporting any condition already?

jcaamano:
I have a major concern with this approach. By reporting even the most transient errors as status conditions, it just looks like we are using them as an extra logging mechanism that could potentially stress out the api server and that looks in any case like a misuse. Conditions should be used for meaningful states that resource goes through and errors should be reasons for not reaching those states rather that states themselves.

npinaeva:
> I have a major concern with this approach. By reporting even the most transient errors as status conditions, it just looks like we are using them as an extra logging mechanism that could potentially stress out the api server and that looks in any case like a misuse. Conditions should be used for meaningful states that resource goes through and errors should be reasons for not reaching those states rather that states themselves.

We actually can convert this error condition to a state condition, but I don't see why it should stress api server (e.g. any more than per-zone status reports). We could make this condition report cluster-level config for network (as ip allocation is a required first step for UDN setup anyway). Or maybe you have other ideas?

jcaamano:
> > I have a major concern with this approach. By reporting even the most transient errors as status conditions, it just looks like we are using them as an extra logging mechanism that could potentially stress out the api server and that looks in any case like a misuse. Conditions should be used for meaningful states that resource goes through and errors should be reasons for not reaching those states rather that states themselves.
> 
> We actually can convert this error condition to a state condition, but I don't see why it should stress api server (e.g. any more than per-zone status reports). We could make this condition report cluster-level config for network (as ip allocation is a required first step for UDN setup anyway). Or maybe you have other ideas?

I have!

Okay, let's say we work with condition `Ready`. We set it to `False`. For reason, we can use `Degraded` or `NodeDegraded` or `NodesDegraded`. 

Now, most of the time we should not update the condition unless the status or the reason changes.

For status `True` it would be nice to update the timestamp (and potentially the observed generation) on any configuration change but that might be difficult to realize for subcontrollers.

For status `False`, we should minimize any updates on error message changes with no reason changes. For the most part the error message should be a human readable and helpful explanation of the reason that is otherwise generic. In this specific case we can include **one** node name that is causing the condition and keep it updated if that node is fixed and then another one causes the same condition.

The message could say for example  "The network is degraded in one or more nodes, including <node>. Check the node events for more details.".

And thus, on the node side, we can record events for a particular node with more details. The events subsystem should be more prepared to take the load.

We should not make status changes or record events for temporary errors, just for permanent or semi-permanent: when we are done retrying, on a humanly meaningful back-off or even just on our longer back-off).

I am not particularly proud on how we did per-zone status reports, those were a bit rushed and also we didn't have a particular resource to map a zone to (although the idea of multiplexing events to all nodes of a zone wouldn't be terrible, like not good, not terrible).

npinaeva:
Added condition for network allocation, and events reporting per node.

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -204,6 +222,63 @@ func (ncc *networkClusterController) init() error {
...
+	} else {
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = "InternalError"
+		condition.Message = "Network allocation failed for one or more nodes, check UDN events for more info."
+		for nodeName, errMsg := range ncc.nodeErrors {
+			events = append(events, &util.EventDetails{
+				EventType: util.EventTypeWarning,
+				Reason:    "NetworkAllocationFailed",
+				Note:      fmt.Sprintf("Error occurred for node %s: %s", nodeName, errMsg),
+			})
```
jcaamano, line 273: Why would we accumulate all the node errors and for each error, send a batch of events with all of the errors?

So for example node A fail, and we sent an event for node A. Then node B fails, and we send an event for node A and B, so now we sent two events for node A.

vs

When node A fail, and we sent an event for node A. Then node B fails, and we send an event for node B only.

test/scripts/e2e-cp.sh
jcaamano: hmm looks we have started miss-using $WHAT. Yours is the expected use case, it was not to focus on a set of well known tests.

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -204,6 +222,63 @@ func (ncc *networkClusterController) init() error {
...
+		}
+	}
+	err := ncc.statusReporter(netName, "NetworkClusterController", &condition, events...)
```
jcaamano, line 276: Just wanna bring up that you are holding a lock while I/O is being done, in case it can potentially become an issue and it could be avoided.

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -327,8 +402,10 @@ func (h *networkClusterControllerEventHandler) AddResource(obj interface{}, from
 		if err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node); err != nil {
 			klog.Infof("Node add failed for %s, will try again later: %v",
 				node.Name, err)
+			h.ncc.updateNetworkStatus(node.Name, fmt.Sprintf("Node add failed: %v", err))
```
jcaamano, line 405: The idea we would just do this when we are done with retries or at least on a long backoff.

I don't know if it can happen in this specific case but in general it might be expected that we fail until something else happens in the system and we might end up reporting spurious errors that while fast enough to go unnoticed by humans, might trigger machines.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: I still think we shouldn't be comparing because it is racey.

But don't you want to record the events even if condition didn't change?

jcaamano: You could use `FindStatusCondition` here as well

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: yeah I like that idea, but I think it only works for controllers setting their retry queues themselves, and for thos eusing obj_retry it is difficult to get that info.
On the bright side: new events will only be recorded if an error message changes, and retry_obj only retries every 30 sec, so at least we should not overload api server with events.

jcaamano: > for thos eusing obj_retry it is difficult to get that info

So the retry framework does provide `RecordErrorEvent` callback that can be implemented. If that doesn't suit our purpose we could just add a different callback that does.

> retry_obj only retries every 30 sec

we should be prepared for this to changed in the future

You can convince me though to maybe tackle this one later.

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -204,6 +220,79 @@ func (ncc *networkClusterController) init() error {
...
+		events = append(events, &util.EventDetails{
+			EventType: util.EventTypeWarning,
+			Reason:    "NetworkAllocationFailed",
+			Note:      fmt.Sprintf("Error occurred for node %s: %s", nodeName, errorMsg),
+		})
+	}
+	if errsBefore == 0 && len(ncc.nodeErrors) == 0 || errsBefore > 0 && len(ncc.nodeErrors) > 0 {
+		// condition didn't change, no need to update
+		condition = nil
+	}
```
jcaamano, line 259: hmmm

Not sure how I feel about reporting events without updating the condition. I feel that if the events are related to the condition, it should be updated.

Are you sure about it?

go-controller/pkg/clustermanager/network_cluster_controller.go
jcaamano: I guess it is not only about the retry period, but also about just everything potentially failing on the first attempt.

go-controller/pkg/clustermanager/network_cluster_controller.go
jcaamano: We could potentially handoff the I/O itself to another thread. But might not be an issue.

jcaamano (commented):
minor comment that might just require you to reflect upon them but no actual changes

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: I think you may want to skip reporting an error immediately for 2 reasons:
1. you expect next retry to succeed or the error message to change
2. you are trying to reduce the error reporting load on the kapi server.
Otherwise I think reporting an error immediately is the best choice. 
So in this case, waiting for 30 seconds to see the error reported may be too much, e.g. if the error is something easy to fix, like too small node subnet in the UDN spec. Also I don't expect this specific case to overload the apiserver. BUT if we ever change the retry logic, this definitely should be reconsidered.
But if we have some exponential backoff, and will have 5 retries in a seconds or something around that, then it makes sense to report error on like 5th retry, since it follows both reasons. wdyt?

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: I think it depends on the condition: if you have some special condition reason for a given error, then it should change with the error, but if the reason is generic like "InternalError", then I don't think updating it makes sense, since the condition itself didn't actually change.

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: If you think it has a high chance to fail on the first try but succeed on the second, I could add a logic to report error starting with the second retry only.
But here I am also thinking about the condition itself: if I don't report an error, and I can't report success, should I have that condition as Unknown until next retry? Kinda knowing at the same time there is a failure, and if a pod in that namespace failed to start it is because of that failure. That could work for fast retries, but not sure if 30s is a good time to keep unknown status

jcaamano: OK so we agree that we need to find a deadline when it is reasonable to send the status report. And for our retry framework to send that status report once on the retry closest to that deadline or soemthing similar.

Let's say we think today that deadline is 30s. Oh! It matches our retry framework deadline! What a coincidence! 
 :P 

That would mean we would report once on the second attempt as you suggest.

However I don't know if I would like to hack this now or follow up with a proper implementation. Will let you decide.

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: yeah that is exactly what I thought and I removed the extra logic from the UDN interface skipping condition updates.
But getting back to the your example, I don't know if there is a way to solve this, as you kinda want to not report same error if nothing changed in the object spec, but do report an error if something did to say to the user that it didn't help.. also doesn't feel quite right, but if I was about to make it work by comparing obj spec, I would try to add this logic to the events generation not to the condition update.

npinaeva: you example made me think about 2 things:
1. events may get lost of expire, but then you are only left with message of "some node error" without even knowing which node you should fix
2. when there are multiple failed nodes, and you have fixed one of them, how do you know it actually worked? One way could be to report success events when an error is fixed or you could list all failed node names in the condition itself, and you don't need a success event.

Getting into account that reporting failed nodes in a condition message will solve both problems, I think I will add it.

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: currently under the lock i/o is the slowest operation (everything else is fairly quick), so I think making another system for i/o lock only will only complicate the code, but won't help performance.

go-controller/pkg/clustermanager/network_cluster_controller.go
jcaamano: > as you kinda want to not report same error if nothing changed in the object spec

This would be the normal behavior of a controller for a resource. But this is not a normal controller for a resource. We consider the resource reconciles in a distributed manner and that's why we have error reports from different controller for that same resource.

This means that changing something in a different place than the resource spec itself may fix the error that we are reporting on that resource (or not). So that means we may want to report even if nothing changed on the resource itself.

> I don't know if there is a way to solve this

This would be my take: we need to be smart on when we send the status report from the source but not try to be very smart from the middleware to cache and filter out on no state change.

However, while we report for each an every error from the source (source I mean caller of `updateNetworkStatus`) then better to be caching in the middleware (meaning the implementation of `updateNetworkStatus`

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -202,6 +219,80 @@ func (ncc *networkClusterController) init() error {
...
+		LastTransitionTime: metav1.Now(),
+	}
+	if len(errorNodes) == 0 {
+		condition.Status = metav1.ConditionTrue
+		condition.Reason = "NetworkAllocationSucceeded"
+		condition.Message = "Network allocation succeeded for all synced nodes."
+	} else {
+		condition.Status = metav1.ConditionFalse
+		condition.Reason = "InternalError"
+		condition.Message = fmt.Sprintf("Network allocation failed for nodes: %v, check UDN events for more info.", errorNodes)
```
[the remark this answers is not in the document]
npinaeva: hmmm I think this is doable: if we think it is fine for a user to fix nodes in the order we report them in status.
If they e.g. see an event for another node that also failed, and they fix it, still no way to know if it worked or not.. probably not too bad?

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -325,8 +416,10 @@ func (h *networkClusterControllerEventHandler) AddResource(obj interface{}, from
 		if err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node); err != nil {
 			klog.Infof("Node add failed for %s, will try again later: %v",
 				node.Name, err)
+			h.ncc.updateNetworkStatus(node.Name, fmt.Sprintf("Node add failed: %v", err))
 			return err
 		}
+		h.ncc.updateNetworkStatus(node.Name, "")
```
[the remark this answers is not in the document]
npinaeva: I think usually we don't bother with status update retries, considering it is not critical. But I can add a mechanism to track if last status report failed and update it on the next call

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -346,3 +352,49 @@ func updateCondition(conditions []metav1.Condition, cond *metav1.Condition) ([]m
...
+
+	udnRef, err := reference.GetReference(userdefinednetworkscheme.Scheme, udn)
+	if err != nil {
+		return fmt.Errorf("failed to get object reference for UserDefinedNetwork %s/%s: %w", udnNamespace, udnName, err)
+	}
+	for _, event := range events {
+		c.eventRecorder.Event(udnRef, event.EventType, event.Reason, event.Note)
+	}
+
+	if condition != nil {
```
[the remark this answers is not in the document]
npinaeva: sometimes events may be reported without condition (currently, if we only report one node name in condition, but another node error message changed, we want to only send a new event)

go-controller/pkg/clustermanager/network_cluster_controller.go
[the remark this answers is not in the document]
jcaamano: Yeah, not too bad, fine compromise in my view.

go-controller/pkg/clustermanager/network_cluster_controller.go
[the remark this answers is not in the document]
jcaamano: I don't think it's worth it specific code, also thinking how it depends on the a next call to actually happen which might not necessarily happen for some time. I would either retry or ignore if that's what we think we should do.

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -322,12 +428,17 @@ func (h *networkClusterControllerEventHandler) AddResource(obj interface{}, from
 		if !ok {
 			return fmt.Errorf("could not cast %T object to *corev1.Node", obj)
 		}
-		if err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node); err != nil {
+		err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node)
+		statusErr := h.ncc.updateNetworkStatus(node.Name, err)
+		joinedErr := errors.Join(err, statusErr)
+		if err == nil {
+			h.clearInitialNodeNetworkUnavailableCondition(node)
+		}
```
jcaamano, line 436: nit: move this two lines up

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -202,6 +224,90 @@ func (ncc *networkClusterController) init() error {
...
+	}
+
+	reportedErrorNode := ncc.reportedErrorNode
+	if ncc.reportedErrorNode == "" && errorMsg != "" {
+		reportedErrorNode = nodeName
+	}
+	if ncc.reportedErrorNode == nodeName && errorMsg == "" {
+		// error for this node is fixed, report next error node
+		// zero value (aka empty string) will be returned if no more nodes have errors
+		reportedErrorNode, _ = sets.New(maps.Keys(ncc.nodeErrors)...).Delete(nodeName).PopAny()
```
jcaamano, line 255: nit
```
for reportedErrorNode := range ncc.nodeErrors {
    if reportedErrorNode != nodeName {
        break
    }
}
```

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -361,12 +472,17 @@ func (h *networkClusterControllerEventHandler) UpdateResource(oldObj, newObj int
 		if !ok {
 			return fmt.Errorf("could not cast %T object to *corev1.Node", newObj)
 		}
-		if err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node); err != nil {
+		err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(node)
+		statusErr := h.ncc.updateNetworkStatus(node.Name, err)
+		joinedErr := errors.Join(err, statusErr)
+		if err == nil {
```
jcaamano, line 478: nit: move this two lines up

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -67,10 +70,21 @@ type networkClusterController struct {
 	// event recorder used to post events to k8s
 	recorder record.EventRecorder
 
+	statusReporter NetworkStatusReporter
+
+	// nodeName: errMessage
+	nodeErrors     map[string]string
```
jcaamano, line 76: Would you see udefulness in reflecting in a TODO comment some of the conversation that we had about caching the errors vs being dumber here but smarter on when this errors get reported trhough a msarter retry mechanism?

jcaamano (commented):
looks good to me, minor nits

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: this will leave you with `reportedErrorNode = nodeName` if it is the only node in the errors map?
I can convert this to the loop though

go-controller/pkg/clustermanager/network_cluster_controller.go
npinaeva: yes! added a link to our conversation here to the "updateNetworkStatus" function comment

jcaamano (approved):

jcaamano (approved):

jcaamano:
Ok, new failures that I haven't seen before
```
  [FAIL] Network Segmentation: Network Policies on a user defined primary network pods within namespace should be isolated when deny policy is present [It] in L2 dualstack primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:1310
  [FAIL] Network Segmentation: Network Policies on a user defined primary network pods within namespace should be isolated when deny policy is present [It] in L3 dualstack primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:1310
  [FAIL] Network Segmentation: Network Policies on a user defined primary network allow ingress traffic to one pod from a particular namespace [It] in L2 primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:1310
  [FAIL] Network Segmentation: Network Policies on a user defined primary network allow ingress traffic to one pod from a particular namespace [It] in L3 primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:1310
```
from https://github.com/ovn-org/ovn-kubernetes/actions/runs/11384905304/job/31675052357?pr=4647

```
  [FAIL] e2e ingress traffic validation Validating ingress traffic [BeforeEach] Should be allowed by nodeport services
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:445
  [FAIL] e2e ingress traffic validation Validating ingress traffic [BeforeEach] Should be allowed by nodeport services after upgrade to DualStack
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:445
  [FAIL] e2e ingress traffic validation Validating ingress traffic [BeforeEach] Should be allowed to node local cluster-networked endpoints by nodeport services with externalTrafficPolicy=local
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:445
```
from https://github.com/ovn-org/ovn-kubernetes/actions/runs/11384905304/job/31675044516?pr=4647

npinaeva:
Network Segmentation one is indeed suspicious, looking for help in slack. Also found that with an error from that test, we have a panic in shutdown, filed a fix https://github.com/ovn-org/ovn-kubernetes/pull/4782

"e2e ingress traffic validation" failed with "docker: Error response from daemon: Head "https://registry.k8s.io/v2/e2e-test-images/agnhost/manifests/2.26": net/http: TLS handshake timeout.". At some point agnhost image became available and the following tests were fine

jcaamano:
> "e2e ingress traffic validation" failed with "docker: Error response from daemon: Head "https://registry.k8s.io/v2/e2e-test-images/agnhost/manifests/2.26": net/http: TLS handshake timeout.". At some point agnhost image became available and the following tests were fine

Could mean we need to retry for longer if there was no specific cause for the delay of that container?

jcaamano:
Issues opened for the flakes
https://github.com/ovn-org/ovn-kubernetes/issues/4783
https://github.com/ovn-org/ovn-kubernetes/issues/4784

PR #4648, opened 2024-08-22
UDN: SGW: L2/L3 Add support for external -> service traffic.

dceara:
Adds support for accessing services (nodePort/LB) from outside (or other nodes).

Fixes #

#### How to verify it

Unit tests have been adapted where needed.

E2e tests for UDN + nodePort have been enabled for shared gateway.

```release-note
NONE
```

### DIFF
--- a/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
+++ b/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
@@ -662,9 +662,11 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
 		}
 	}
 
-	if zic.IsSecondary() {
+	if zic.IsSecondary() && !(util.IsNetworkSegmentationSupportEnabled() && zic.IsPrimaryNetwork()) {
 		// Secondary network cluster router doesn't connect to a join switch
 		// or to a Gateway router.
+		//
+		// Except for UDN primary L3 networks.
 		return nil
 	}
 
--- a/test/e2e/network_segmentation_services.go
+++ b/test/e2e/network_segmentation_services.go
@@ -256,19 +256,33 @@ var _ = Describe("Network Segmentation: services", func() {
 // TODO Once https://github.com/ovn-org/ovn-kubernetes/pull/4567 merges, use the vendored *TestJig.Run(), which tests
 // the reachability of a service through its name and through its cluster IP. For now only test the cluster IP.
 
-const OVNNodeHostCIDRs = "k8s.ovn.org/host-cidrs"
+const OvnNodeIfAddr = "k8s.ovn.org/node-primary-ifaddr"
 
-// ParseNodeHostCIDRsDropNetMask returns the parsed host IP addresses found on a node's host CIDR annotation. Removes the mask.
-func ParseNodeHostCIDRsDropNetMask(node *kapi.Node) (sets.Set[string], error) {
-	addrAnnotation, ok := node.Annotations[OVNNodeHostCIDRs]
+type primaryIfAddrAnnotation struct {
+	IPv4 string `json:"ipv4,omitempty"`
+	IPv6 string `json:"ipv6,omitempty"`
+}
+
+// ParseNodeHostIPDropNetMask returns the parsed host IP addresses found on a node's host CIDR annotation. Removes the mask.
+func ParseNodeHostIPDropNetMask(node *kapi.Node) (sets.Set[string], error) {
+	nodeIfAddrAnnotation, ok := node.Annotations[OvnNodeIfAddr]
 	if !ok {
-		return nil, newAnnotationNotSetError("%s annotation not found for node %q", OVNNodeHostCIDRs, node.Name)
+		return nil, newAnnotationNotSetError("%s annotation not found for node %q", OvnNodeIfAddr, node.Name)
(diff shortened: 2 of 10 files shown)


tssurya:
I am going to leave a comment here so that we have history this point was discussed and why this design was chosen so that in 3 months from now this is not questioned (which is what has been happening often for me with UDN).

1. This PR is going for using the same default network's nodeport watcher to add flows for UDNs as well. So the default node network controller will now start programming flows for user defined network owned services.
2. This is anti-pattern to everything we have done so far => all other bits of UDN are done from the given UDN's network controller including the default per network openflows we program => reconcile/entry point is always from that network's controller
3. Why was this anti-pattern chosen? For scale reasons => we don't want to add a per network service watcher (which is what we do on the ovnkube-controller side which actually can also be avoided if we think about it; maybe we should in the future just treat all networks from same controller; the per network controller actually need not be done to be honest because its all going into the same NBDB)
4. Con of this approach: Developers are going to find this super hard to maintain because one set of things are done from DNCC and other is done from SNCC and I wonder how EIPs is planning to do this. Basically I want to call out the inconsistency part here, but it is what it is.

The point of this comment is not to tell one is better than the other; its to simply capture what was discussed and why this was done so that a few months from now we can look back as to why we did this.

dceara:
@trozet This is ready for review now.

test/e2e/network_segmentation_services.go
```diff
@@ -256,19 +256,33 @@ var _ = Describe("Network Segmentation: services", func() {
 // TODO Once https://github.com/ovn-org/ovn-kubernetes/pull/4567 merges, use the vendored *TestJig.Run(), which tests
 // the reachability of a service through its name and through its cluster IP. For now only test the cluster IP.
 
-const OVNNodeHostCIDRs = "k8s.ovn.org/host-cidrs"
+const OvnNodeIfAddr = "k8s.ovn.org/node-primary-ifaddr"
```
trozet, line 265: for "udn, e2e: Don't use secondary node IPs for node port test." it would be good to have a commit message explaining why this change is needed

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -662,9 +662,11 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
 		}
 	}
 
-	if zic.IsSecondary() {
+	if zic.IsSecondary() && !(util.IsNetworkSegmentationSupportEnabled() && zic.IsPrimaryNetwork()) {
```
trozet, line 665: note to self, we dont need to conditionalize this on topology being layer3, cause (zic *ZoneInterconnectHandler) AddRemoteZoneNode only gets called for default controller and layer 3 secondary net

test/e2e/network_segmentation_services.go
```diff
@@ -156,9 +156,11 @@ var _ = Describe("Network Segmentation: services", func() {
 				By("Connect to the UDN service cluster IP from the UDN client pod on the same node")
 				checkConnectionToClusterIPs(f, udnClientPod, udnService, udnServerPod.Name)
 				checkConnectionToNodePort(f, udnClientPod, udnService, &nodes.Items[0], "endpoint node", udnServerPod.Name)
-				// TODO uncomment below as soon as ovnkube-node supports UDN
-				// checkConnectionToNodePort(f, clientPod2, udnService, &nodes.Items[1], "client node", udnServerPod.Name)
-				// checkConnectionToNodePort(f, clientPod2, udnService, &nodes.Items[2], "other node", udnServerPod.Name)
+				// FIXME(dceara): Remove this check when Local Gateway external->service support is implemented.
+				if !IsGatewayModeLocal() {
+					checkConnectionToNodePort(f, udnClientPod, udnService, &nodes.Items[1], "client node", udnServerPod.Name)
```
trozet, line 161: nit: we should probably also add a couple more ingress tests. I'm thinking external IP, and maybe LB IP. Perhaps with ETP local. Don't want to overkill this since we plan on migrating to k8s tests, but would be nice until then to have a little more coverage. Can be done in follow up.

trozet (approved):

tssurya:
I have a few questions:

1. I don't see any new unit tests added for the new sevices flows on UDN, am I missing them?
2. E2E tests for external client to load balancer/nodeport services seem missing?

Will these be taken care of in another PR? Shouldn't we be adding external->services e2e test? Ricky's tests are only for UDN pod-> other nodeport which I understand is being treat as external->services tests but IMHO we should use external docker clients and add some good real tests

Also be using metalLB and write LB tests

dceara:
> I have a few questions:
> 
>     1. I don't see any new unit tests added for the new sevices flows on UDN, am I missing them?
> 

No I didn't add those, let's do it as a follow up:
https://github.com/ovn-org/ovn-kubernetes/issues/4698

>     2. E2E tests for external client to load balancer/nodeport services seem missing?
> 

Same here:
https://github.com/ovn-org/ovn-kubernetes/issues/4699

> 
> Will these be taken care of in another PR? Shouldn't we be adding external->services e2e test? Ricky's tests are only for UDN pod-> other nodeport which I understand is being treat as external->services tests but IMHO we should use external docker clients and add some good real tests
> 
> Also be using metalLB and write LB tests

Thanks!

PR #4651, opened 2024-08-22
udn: set persistentIPs for UDN ifaces

maiqueb:

This PR looks for a special annotation on pods, where KubeVirt (or an agent on behalf of KubeVirt) will indicate what is the name of the `IPAMClaim` used for the primary UDN interface.

An alternative would be to use an hard-coded value, but then we wouldn't be able to distinguish between a pod and a VM, and we do not want to abuse reading the KubeVirt label.

Fixes #

Pending adding the CI requirements, ensuring kubevirt e2e tests are run.

Configure persistent IPs on UDN networks when requested

```release-note
Configure persistent IPs on UDN networks when requested
```

### DIFF
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -603,3 +603,11 @@ spec:
 }'
 EOF
 }
+
+deploy_passt_binary() {
+  echo "Installing passt-binding-cni-ds ..."
+  local manifest="https://raw.githubusercontent.com/kubevirt/ipam-extensions/main/passt/passt-binding-cni-ds.yaml"
+  run_kubectl apply -f "$manifest"
+
+  run_kubectl rollout status -n kube-system daemonset/passt-binding-cni --timeout 2m
+}
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -349,6 +349,11 @@ install_kubevirt() {
             kubectl logs --all-containers=true -n kubevirt $p || true
         done
     fi
+
+    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/developerConfiguration","value":{"featureGates":[]}},{"op":"add","path":"/spec/configuration/developerConfiguration/featureGates/-","value":"NetworkBindingPlugins"}]'
+    
+    local passt_binding_image="quay.io/kubevirt/network-passt-binding:${kubevirt_version}"
+    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/network","value":{}},{"op":"add","path":"/spec/configuration/network/binding","value":{"passt":{"computeResourceOverhead":{"requests":{"memory":"500Mi"}},"migration":{"method":"link-refresh"},"networkAttachmentDefinition":"default/primary-udn-kubevirt-binding","sidecarImage":"'"${passt_binding_image}"'"}}}]'
     
     if [ ! -d "./bin" ]
     then
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -582,3 +582,24 @@ kubectl_wait_dnsnameresolver_pods() {
   echo "Waiting for pods in dnsnameresolver-operator namespace to become ready (timeout ${timeout})..."
   kubectl wait -n dnsnameresolver-operator --for=condition=ready pods --all --timeout=${timeout}s
 }
+
+deploy_kubevirt_binding() {
+  cat <<EOF | run_kubectl apply -f -
+---
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: primary-udn-kubevirt-binding
+  namespace: default
+spec:
+  config: '{
+  "cniVersion": "1.0.0",
+  "name": "primary-udn-kubevirt-binding",
+  "plugins": [
+    {
+      "type": "network-passt-binding"
+    }
+  ]
+}'
+EOF
+}
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -379,12 +379,14 @@ install_kubevirt() {
     chmod +x ./bin/virtctl
 }
 
-install_kubevirt_ipam_controller() {
+install_cert_manager() {
   local cert_manager_version="v1.14.4"
   echo "Installing cert-manager ..."
   manifest="https://github.com/cert-manager/cert-manager/releases/download/${cert_manager_version}/cert-manager.yaml"
   run_kubectl apply -f "$manifest"
+}
 
+install_kubevirt_ipam_controller() {
   echo "Installing KubeVirt IPAM controller manager ..."
   manifest="https://raw.githubusercontent.com/kubevirt/ipam-extensions/main/dist/install.yaml"
   run_kubectl apply -f "$manifest"
--- a/go-controller/pkg/util/multi_network.go
+++ b/go-controller/pkg/util/multi_network.go
@@ -954,8 +954,18 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, ac
 		Name:      activeNetworkNADKey[1],
 	}
 
+	if nInfo.IsPrimaryNetwork() &&
+		nInfo.TopologyType() == types.Layer2Topology &&
+		nInfo.AllowsPersistentIPs() {
+		ipamClaimName, wasPersistentIPRequested := pod.Annotations[OvnUDNIPAMClaimName]
+		if wasPersistentIPRequested {
+			networkSelections[activeNetworkNADs[0]].IPAMClaimReference = ipamClaimName
+		}
+	}
+
 	return true, networkSelections, nil
 }
+
 func IsMultiNetworkPoliciesSupportEnabled() bool {
 	return config.OVNKubernetesFeature.EnableMultiNetwork && config.OVNKubernetesFeature.EnableMultiNetworkPolicy
 }
--- a/go-controller/pkg/util/multi_network.go
+++ b/go-controller/pkg/util/multi_network.go
@@ -954,9 +954,7 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, ac
 		Name:      activeNetworkNADKey[1],
 	}
 
-	if nInfo.IsPrimaryNetwork() &&
-		nInfo.TopologyType() == types.Layer2Topology &&
-		nInfo.AllowsPersistentIPs() {
+	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) {
 		ipamClaimName, wasPersistentIPRequested := pod.Annotations[OvnUDNIPAMClaimName]
 		if wasPersistentIPRequested {
 			networkSelections[activeNetworkNADs[0]].IPAMClaimReference = ipamClaimName
@@ -981,4 +979,18 @@ func DoesNetworkRequireIPAM(netInfo NetInfo) bool {
 func DoesNetworkRequireTunnelIDs(netInfo NetInfo) bool {
 	// Layer2Topology with IC require that we allocate tunnel IDs for each pod
 	return netInfo.TopologyType() == types.Layer2Topology && config.OVNKubernetesFeature.EnableInterconnect
+}
+
+func AllowsPersistentIPs(netInfo NetInfo) bool {
+	switch {
+	case netInfo.IsPrimaryNetwork():
+		return netInfo.TopologyType() == types.Layer2Topology && netInfo.AllowsPersistentIPs()
+
+	case netInfo.IsSecondary():
+		return (netInfo.TopologyType() == types.Layer2Topology || netInfo.TopologyType() == types.LocalnetTopology) &&
+			netInfo.AllowsPersistentIPs()
+
+	default:
+		return false
+	}
 }
--- a/go-controller/pkg/util/pod_annotation.go
+++ b/go-controller/pkg/util/pod_annotation.go
@@ -51,6 +51,9 @@ const (
 	OvnPodAnnotationName = "k8s.ovn.org/pod-networks"
 	// DefNetworkAnnotation is the pod annotation for the cluster-wide default network
 	DefNetworkAnnotation = "v1.multus-cni.io/default-network"
+	// OvnUDNIPAMClaimName is used for workload owners to instruct OVN-K which
+	// IPAMClaim will hold the allocation for the workload
+	OvnUDNIPAMClaimName = "k8s.ovn.org/primary-udn-ipamclaim"
 )
 
 var ErrNoPodIPFound = errors.New("no pod IPs found")
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -14,6 +14,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/diagnostics"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/kubevirt"
 
@@ -32,6 +33,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	testutils "k8s.io/kubernetes/test/utils"
 	utilnet "k8s.io/utils/net"
 	"k8s.io/utils/pointer"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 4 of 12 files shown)


go-controller/pkg/util/pod_annotation.go
```diff
@@ -51,6 +51,9 @@ const (
 	OvnPodAnnotationName = "k8s.ovn.org/pod-networks"
 	// DefNetworkAnnotation is the pod annotation for the cluster-wide default network
 	DefNetworkAnnotation = "v1.multus-cni.io/default-network"
+	// OvnUDNIPAMClaimName is used for workload owners to instruct OVN-K which
+	// IPAMClaim will hold the allocation for the workload
+	OvnUDNIPAMClaimName = "k8s.ovn.org/ovn-udn-ipamclaim-reference"
```
maiqueb, line 56: @trozet / @tssurya 

let's discuss the annotation value.

go-controller/pkg/util/multi_network.go
```diff
@@ -939,6 +939,13 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, ac
 		Name:      activeNetworkNADKey[1],
 	}
 
+	if nInfo.IsPrimaryNetwork() && nInfo.TopologyType() == types.Layer2Topology && nInfo.AllowsPersistentIPs() {
+		ipamClaimName, wasPersistentIPRequested := pod.Annotations[OvnUDNIPAMClaimName]
+		if wasPersistentIPRequested {
+			networkSelections[activeNetworkNADs[0]].IPAMClaimReference = ipamClaimName
+		}
```
jcaamano, line 946: nit: wouldn't just this be equivalent:
```
networkSelections[activeNetworkNADs[0]].IPAMClaimReference = pod.Annotations[OvnUDNIPAMClaimName]
```

go-controller/pkg/util/pod_annotation.go
jcaamano: This could just be `k8s.ovn.org/ipamclaim-reference`

go-controller/pkg/util/multi_network.go
```diff
@@ -939,6 +939,13 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, ac
 		Name:      activeNetworkNADKey[1],
 	}
 
+	if nInfo.IsPrimaryNetwork() && nInfo.TopologyType() == types.Layer2Topology && nInfo.AllowsPersistentIPs() {
```
jcaamano, line 942: it would probably be useful to have a free `AllowsPersistentIPs(netInfo)` method here that summarizes which network allow persistent IPs

test/e2e/kubevirt.go
```diff
@@ -313,16 +315,31 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
 			}
 		}
 
-		httpServerTestPodsMultusNetworkIPs = func(netName string) map[string][]string {
+		checkPodRunningReady = func() func(Gomega, *corev1.Pod) {
+			return func(g Gomega, pod *corev1.Pod) {
+				GinkgoHelper()
```
jcaamano, line 320: hmm I don't think you need `GinkgoHelper` for functions that receive a gomega instance from gomega

jcaamano (commented):
Will we have integration docs for this new annotation?

contrib/kind-common
```diff
@@ -581,3 +581,24 @@ kubectl_wait_dnsnameresolver_pods() {
   echo "Waiting for pods in dnsnameresolver-operator namespace to become ready (timeout ${timeout})..."
   kubectl wait -n dnsnameresolver-operator --for=condition=ready pods --all --timeout=${timeout}s
 }
+
+deploy_primary_udn_nad() {
```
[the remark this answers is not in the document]
maiqueb: what do you suggest ? passt_binding ?

go-controller/pkg/util/multi_network.go
maiqueb: maybe. I try to repeat what I think are best practices without thinking too much. 
Always checking if something was found seems good. 

In this particular case I think you're right.

go-controller/pkg/util/multi_network.go
maiqueb: OK, I'll add it.

test/e2e/kubevirt.go
maiqueb: @qinqon ^ ??

go-controller/pkg/util/pod_annotation.go
maiqueb: really ? I think it should somehow indicate this is meant to tell ovn-k of what is the claim to be used for the primary UDN. 

I agree dropping the `ovn` should happen ... awfully redundant.

maiqueb:
> Will we have integration docs for this new annotation?

Yes. But since we have no documentation about primary UDN, it feels very weird to add it. 

Once that exists, I will explain the annotation.

test/e2e/kubevirt.go
qinqon: ALl right, I will remove it after CI finish so we have results

contrib/kind-common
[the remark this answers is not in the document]
jcaamano: I don't know. I don't know what network is this. Is there a diagram somewhere?

go-controller/pkg/util/pod_annotation.go
maiqueb: Done.

maiqueb: @qinqon we'll need to update the client for the e2e tests to pass ...

go-controller/pkg/util/multi_network.go
maiqueb: Done.

jcaamano:
One other question:

Would it have been theoretically possible to use the ipam claim reference in a network selection element for the default network (instead of this new annotation) leveraging `v1.multus-cni.io/default-network` annotation. I know that is out of spec but we are out of spec here anyway and we wouldn't need to define new things.

contrib/kind-common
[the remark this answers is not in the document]
maiqueb: This is how we install the plugin-based kubevirt binding that will be selected by users that want to connect to primary UDNs from KubeVirt. 

No docs right now.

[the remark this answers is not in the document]
jcaamano: past_binding looks fine to me.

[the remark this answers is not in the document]
maiqueb: went for kubevirt binding, since we'll end up replacing this binding for another - i.e. that's just an implementation detail.

What matters (IMO) is we indicate what this is for - the kubevirt integration.

maiqueb:
> One other question:
> 
> Would it have been theoretically possible to use the ipam claim reference in a network selection element for the default network (instead of this new annotation) leveraging `v1.multus-cni.io/default-network` annotation. I know that is out of spec but we are out of spec here anyway and we wouldn't need to define new things.

Not sure ... how would that play along for the rest of the synthetic network selection element we are using to attach to UDN ? For UDN pods, we'd use the synthetic network-selection-element, for UDN VMs we'd use this explicitly defined network-selection-element ?... We'd merge the synthetic NSE w/ the existing one ?... 

Maybe it's me, but it sounds more complex.

jcaamano:
> > One other question:
> > Would it have been theoretically possible to use the ipam claim reference in a network selection element for the default network (instead of this new annotation) leveraging `v1.multus-cni.io/default-network` annotation. I know that is out of spec but we are out of spec here anyway and we wouldn't need to define new things.
> 
> Not sure ... how would that play along for the rest of the synthetic network selection element we are using to attach to UDN ? For UDN pods, we'd use the synthetic network-selection-element, for UDN VMs we'd use this explicitly defined network-selection-element ?... We'd merge the synthetic NSE w/ the existing one ?...
> 
> Maybe it's me, but it sounds more complex.

I am not sure what's the synthetic network selection element used for, but that looks like an implementation detail.

I would value more having a consistent API for our users.

maiqueb:
> > > One other question:
> > > Would it have been theoretically possible to use the ipam claim reference in a network selection element for the default network (instead of this new annotation) leveraging `v1.multus-cni.io/default-network` annotation. I know that is out of spec but we are out of spec here anyway and we wouldn't need to define new things.
> > 
> > 
> > Not sure ... how would that play along for the rest of the synthetic network selection element we are using to attach to UDN ? For UDN pods, we'd use the synthetic network-selection-element, for UDN VMs we'd use this explicitly defined network-selection-element ?... We'd merge the synthetic NSE w/ the existing one ?...
> > Maybe it's me, but it sounds more complex.
> 
> I am not sure what's the synthetic network selection element used for, but that looks like an implementation detail.
> 
> I would value more having a consistent API for our users.

It's how we're requesting the UDN interface attachment without persistent IPs. 

Meaning, the API is not consistent now. I don't think that can converge in this PR.

jcaamano:
> It's how we're requesting the UDN interface attachment without persistent IPs.

But you mean completely internal to OVN-K, we are not using it as an API to interface with any other external component. Or what does synthetic refer to in this context?

> Meaning, the API is not consistent now. I don't think that can converge in this PR.

The IPAM claim reference API could be consistent because it would be the same for secondary and primary networks.

Not for this PR, but maybe worth exploring before the final integration.

maiqueb:
> > It's how we're requesting the UDN interface attachment without persistent IPs.
> 
> But you mean completely internal to OVN-K, we are not using it as an API to interface with any other external component. Or what does synthetic refer to in this context?
> 
> > Meaning, the API is not consistent now. I don't think that can converge in this PR.
> 
> The IPAM claim reference API could be consistent because it would be the same for secondary and primary networks.
> 
> Not for this PR, but maybe worth exploring before the final integration.

This would maybe be an interesting idea to explore in the [enhancement proposal](https://github.com/openshift/enhancements/pull/1673).

test/e2e/kubevirt.go
maiqueb: Done; removed it.

jcaamano:
> Yes, but, I think that by making that API consistent, you are then making the UDN API inconsistent - if persistent IPs are used, we require a network selection element; if not, we'd just annotate the namespace.

The network selection element would not replace the namespace annotation, that is just a shortcut to default a different network in your namespace (so more akin to a shortcut for `v1.multus-cni.io/default-network`). But if you want to change the details of that network on a pod-per-pod basis, then you would have to annotate the pod with the NSE for the default network as you do today, for example, to use sr-iov on the default network with a specific pod.

maiqueb:
> The network selection element would not replace the namespace annotation, that is just a shortcut to default a different network in your namespace (so more akin to a shortcut for v1.multus-cni.io/default-network). But if you want to change the details of that network on a pod-per-pod basis, then you would have to annotate the pod with the NSE for the default network as you do today, for example, to use sr-iov on the default network with a specific pod.

Right now, it is impossible to refer to a primary UDN via network-selection-element. We disallow that explicitly in the code.

I think I start to see your point. I will consider this option in the [enhancement](https://github.com/openshift/enhancements/pull/1673). Thanks for raising it.

go-controller/pkg/util/multi_network.go
```diff
@@ -957,3 +965,7 @@ func DoesNetworkRequireTunnelIDs(netInfo NetInfo) bool {
 	// Layer2Topology with IC require that we allocate tunnel IDs for each pod
 	return netInfo.TopologyType() == types.Layer2Topology && config.OVNKubernetesFeature.EnableInterconnect
 }
+
+func AllowsPersistentIPs(netInfo NetInfo) bool {
+	return netInfo.IsPrimaryNetwork() && netInfo.TopologyType() == types.Layer2Topology && netInfo.AllowsPersistentIPs()
```
jcaamano, line 970: Shouldn't this cover secondary networks as well?

maiqueb: I'd rather not; I would prefer to have an helper tailored for this use: primary layer2 UDN with persistent IPs. 

I can make the name more appropriate though.

jcaamano: It doesn't make sense then because what I wanted help with was: ` summarizes which networks allow persistent IPs`

maiqueb: So let's put it back the way it was originally, because it is different in primary UDNs and in secondary networks (in primary UDNs it is only for layer2; in secondary networks it allows for localnet networks as well).

My bad in blindly going forward with this.

maiqueb: Done.

jcaamano: > So let's put it back the way it was originally, because it is different in primary UDNs and in secondary networks (in primary UDNs it is only for layer2; in secondary networks it allows for localnet networks as well).
> 
> My bad in blindly going forward with this.

That's actually the nuance that I wanted to put on the table and summarize in a single place. Wouldn't this work, for example:
```
func AllowsPersistentIPs(netInfo NetInfo) bool {
  switch {
      case netInfo.IsPrimaryNetwork():
          return netInfo.TopologyType() == types.Layer2Topology && netInfo.AllowsPersistentIPs()
  
      case netInfo.IsSecondaryNetwork():
          return (netInfo.TopologyType() == types.Layer2Topology || netInfo.TopologyType() == types.LocalnetTopology) && netInfo.AllowsPersistentIPs()
  
      default:
          return false
  }
}
```

maiqueb: OK, I'll do this.

maiqueb: Done; the other usages are just too weirdly scoped for me to change in the context of this PR.

jcaamano: what about cluster manager? Aren't layer 2 networks reconciling ipam claims from there? This might have changed and I missed it.

maiqueb: Correct; done.

qinqon:
@maiqueb @jcaamano the kubevirt failure is a new flake related to health checks after exposing a services to a VM 
```
[FAILED] worker1: Expose tcpServer as a service
  Unexpected error:
      <*fmt.wrapError | 0xc0005c4320>: 
      failed DialTCP: dial tcp 172.18.0.5:32102: connect: connection refused
      {
          msg: "failed DialTCP: dial tcp 172.18.0.5:32102: connect: connection refused",
          err: <*net.OpError | 0xc00057c910>{
              Op: "dial",
              Net: "tcp",
              Source: nil,
              Addr: <*net.TCPAddr | 0xc0007b58f0>{
                  IP: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 172, 18, 0, 5],
                  Port: 32102,
                  Zone: "",
              },
              Err: <*os.SyscallError | 0xc0005c4300>{
                  Syscall: "connect",
                  Err: <syscall.Errno>0x6f,
              },
          },
      }
```

go-controller/pkg/util/multi_network.go
```diff
@@ -939,8 +939,16 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, ac
 		Name:      activeNetworkNADKey[1],
 	}
 
+	if AllowsPersistentIPs(nInfo) {
```
jcaamano, line 942: Do we need to check if this is a primary network here?

jcaamano:
> @maiqueb @jcaamano the kubevirt failure is a new flake related to health checks after exposing a services to a VM, is not realted to this PR since the "post-copy" test is doing the same and is passing.

Do we need to file an issue for this one?

go-controller/pkg/util/multi_network.go
maiqueb: Why ? Isn't that checked inside the allows persistent IPs helper you've requested ?

maiqueb:
> ovn-ci / e2e (kv-live-migration, noHA, shared, dualstack, SnatGW, 1br, ic-single-node-zones, 3

https://github.com/ovn-org/ovn-kubernetes/issues/4709

maiqueb:
The upgrade test is flaking on https://github.com/ovn-org/ovn-kubernetes/issues/4481; updated that issue.

go-controller/pkg/util/multi_network.go
jcaamano: Right, but the annotation itself is just for primary networks and not secondary ones, right? So if we just use `AllowsPersistentIPs` we would be allowing the use of the annotation for secondary networks as well?

maiqueb: Thus, IMHO, coalescing all the info about when persistent ips in a single place is counter-productive. 

... since we'll have to perform redundant checks elsewhere to ensure we're not covering more than what we want.

But I'll blindly throw it in. Done.

jcaamano: The way we have the code now, which could be better but I think is an improvement over what we had, highlights that the persistent IPs functionality is the same for all networks and that the only difference is the API. I don't really understand how you see a problem with the code but at the same time be fine with the API differences.

maiqueb: The UDN API differences were proposed, discussed, and accepted on an enhancement.

I personally think they do make sense, and simplify the users tasks. Hence I'm OK with it.

Still, I think this is tangential to this particular discussion.

qinqon:
/lgtm

The kubevirt issues were related to https://github.com/ovn-org/ovn-kubernetes/pull/4651

maiqueb:
Fixed the errors in the kubevirt lane, introduced after rebasing #4625 (since we were testing a dual stack UDN in an IPv4 cluster) in [this forced push](https://github.com/ovn-org/ovn-kubernetes/compare/cf318cebb7c27e7231439f8f7189d8053be28d38..46dc6a0240d0ab7a4cd99f4ba2a08c80a50efd4c).

maiqueb:
> /lgtm
> 
> The kubevirt issues were related to #4625

Thanks for the help !!! :rocket:

jcaamano (approved):

PR #4652, opened 2024-08-26
Serializes Network Manager Start up

trozet:
NAD controller adds keys to Network Manager to process. Upon Network Manager start up, it does not add any handler for Kubernetes and therefore does not block until it processes all current objects. For certain controllers that integrate with UDN, they contain caches, which depend on being initialized by networks having been created at start up. If those controllers were to start without a fully initialized cache, they may end up temporarily deleting configuration that could cause disruption.

An example of this is the ip rule manager. If the ip rule manager reconciliation loop runs, it will try to remove any stale network rules. If the cache is not fully updated by the time the loop runs, it would accidentally remove previously configured rules, causing disruption until the network add is processed.

This change makes network manager start up block until all initial networks have been sync'ed.

### DIFF
--- a/go-controller/pkg/network-attach-def-controller/network_manager.go
+++ b/go-controller/pkg/network-attach-def-controller/network_manager.go
@@ -45,7 +45,7 @@ func newNetworkManager(name string, ncm NetworkControllerManager) networkManager
 	// added to the queue for processing
 	config := &controller.ReconcilerConfig{
 		RateLimiter: workqueue.DefaultControllerRateLimiter(),
-		Reconcile:   nc.sync,
+		Reconcile:   nc.syncLocked,
 		Threadiness: 1,
 	}
 	nc.controller = controller.NewReconciler(
@@ -97,10 +97,14 @@ func (nm *networkManagerImpl) DeleteNetwork(network string) {
 	nm.controller.Reconcile(network)
 }
 
-func (nm *networkManagerImpl) sync(network string) error {
+func (nm *networkManagerImpl) syncLocked(network string) error {
 	nm.Lock()
 	defer nm.Unlock()
+	return nm.sync(network)
+}
 
+// sync must be called with nm mutex locked
+func (nm *networkManagerImpl) sync(network string) error {
 	startTime := time.Now()
 	klog.V(5).Infof("%s: sync network %s", nm.name, network)
 	defer func() {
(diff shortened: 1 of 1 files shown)


trozet:
@tssurya fyi

go-controller/pkg/network-attach-def-controller/network_manager.go
```diff
@@ -158,8 +162,26 @@ func (nm *networkManagerImpl) syncAll() error {
...
+	}
+	if err := nm.ncm.CleanupDeletedNetworks(validNetworks...); err != nil {
+		return err
 	}
-	return nm.ncm.CleanupDeletedNetworks(validNetworks...)
+
+	// sync all known networks. There is no informer for networks. Keys are added by NAD controller.
+	// Therefore, there is no serialization for initial adds, since addHandler is not called and therefore
+	// networkManager has a non-blocking start up. Leverage syncAll to block NetworkManager start up to
+	// simulate addHandler behavior, waiting for all networks to be added.
```
jcaamano, line 177: I don't think this comment really explains why this is needed. I think the key here is that existing networks need to be started synchronously, maybe this is what you mean with serialization? (adds are serialized because we only have a single thread in the controller). 

Perhaps this other section of your explanation is the better perspective
```
For certain controllers that integrate with UDN, they contain caches, which depend on being initialized by networks having been created at start up. If those controllers were to start without a fully initialized cache, they may end up temporarily deleting configuration that could cause disruption.
```

You talk about IP rule manager. I guess that is shared among networks and if it starts up without being aware of them it removes configuration it thinks as stale. I guess this works but all these things that will be shared by multiple networks would really need to be managed at a higher level, like maybe by network manager itself?

Anyway, I would probably edit the this comment to be something similar to
```
Certain downstream controllers that handle configuration for multiple networks depend on being aware of all the existing networks on initialization. To achieve that, we need to start existing networks synchronously. Otherwise these controllers might incorrectly incorrectly asses certain configuration as stale"   
```

jcaamano (approved):

go-controller/pkg/network-attach-def-controller/network_manager.go
trozet: imo with almost all controllers today we think about them as being fully initialized and handled all the adds before we move onto starting the next controller. The order matters for us with node handler, pod handler, etc. I think we should keep that consistent even with our network controller, unless there is some good justification. Otherwise I worry we will just confuse ourselves even more.

PR #4655, opened 2024-08-27
Integrate passt components to kubevirt lane

RamLavi:
This PR integrates passt components to kubevirt lane. 
This is going to be useful for primary-UDN development with kubevirt.

Also, in order to allow working with dev ipam-controller, adding a new flag that will opt-out installing the latest ipam-controller but keep the cert-manager install.

Fixes #

```release-note
NONE
```

### DIFF
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -348,6 +348,11 @@ install_kubevirt() {
             kubectl logs --all-containers=true -n kubevirt $p || true
         done
     fi
+
+    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/developerConfiguration","value":{"featureGates":[]}},{"op":"add","path":"/spec/configuration/developerConfiguration/featureGates/-","value":"NetworkBindingPlugins"}]'
+    
+    local passt_binding_image="quay.io/kubevirt/network-passt-binding:${kubevirt_version}"
+    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/network","value":{}},{"op":"add","path":"/spec/configuration/network/binding","value":{"passt":{"computeResourceOverhead":{"requests":{"memory":"500Mi"}},"migration":{"method":"link-refresh"},"networkAttachmentDefinition":"default/primary-udn-kubevirt-binding","sidecarImage":"'"${passt_binding_image}"'"}}}]'
     
     if [ ! -d "./bin" ]
     then
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -378,12 +378,14 @@ install_kubevirt() {
     chmod +x ./bin/virtctl
 }
 
-install_kubevirt_ipam_controller() {
+install_cert_manager() {
   local cert_manager_version="v1.14.4"
   echo "Installing cert-manager ..."
   manifest="https://github.com/cert-manager/cert-manager/releases/download/${cert_manager_version}/cert-manager.yaml"
   run_kubectl apply -f "$manifest"
+}
 
+install_kubevirt_ipam_controller() {
   echo "Installing KubeVirt IPAM controller manager ..."
   manifest="https://raw.githubusercontent.com/kubevirt/ipam-extensions/main/dist/install.yaml"
   run_kubectl apply -f "$manifest"
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -602,3 +602,11 @@ spec:
 }'
 EOF
 }
+
+deploy_passt_binary() {
+  echo "Installing passt-binding-cni-ds ..."
+  local manifest="https://raw.githubusercontent.com/kubevirt/ipam-extensions/main/passt/passt-binding-cni-ds.yaml"
+  run_kubectl apply -f "$manifest"
+
+  run_kubectl rollout status -n kube-system daemonset/passt-binding-cni --timeout 2m
+}
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -581,3 +581,24 @@ kubectl_wait_dnsnameresolver_pods() {
   echo "Waiting for pods in dnsnameresolver-operator namespace to become ready (timeout ${timeout})..."
   kubectl wait -n dnsnameresolver-operator --for=condition=ready pods --all --timeout=${timeout}s
 }
+
+deploy_primary_udn_nad() {
+  cat <<EOF | run_kubectl apply -f -
+---
+apiVersion: "k8s.cni.cncf.io/v1"
+kind: NetworkAttachmentDefinition
+metadata:
+  name: primary-udn-kubevirt-binding
+  namespace: default
+spec:
+  config: '{
+  "cniVersion": "1.0.0",
+  "name": "primary-udn-kubevirt-binding",
+  "plugins": [
+    {
+      "type": "network-passt-binding"
+    }
+  ]
+}'
+EOF
+}
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -163,6 +163,8 @@ parse_args() {
                                                 ;;
             -ikv | --install-kubevirt)          KIND_INSTALL_KUBEVIRT=true
                                                 ;;
+            -nokvipam | --opt-out-kv-ipam)      KIND_OPT_OUT_KUBEVIRT_IPAM=true
+                                                ;;
             -ha | --ha-enabled )                OVN_HA=true
                                                 ;;
             -me | --multicast-enabled)          OVN_MULTICAST_ENABLE=true
@@ -350,6 +352,7 @@ print_params() {
      echo "KIND_INSTALL_METALLB = $KIND_INSTALL_METALLB"
      echo "KIND_INSTALL_PLUGINS = $KIND_INSTALL_PLUGINS"
      echo "KIND_INSTALL_KUBEVIRT = $KIND_INSTALL_KUBEVIRT"
+     echo "KIND_OPT_OUT_KUBEVIRT_IPAM = $KIND_OPT_OUT_KUBEVIRT_IPAM"
      echo "OVN_HA = $OVN_HA"
      echo "RUN_IN_CONTAINER = $RUN_IN_CONTAINER"
      echo "KIND_CLUSTER_NAME = $KIND_CLUSTER_NAME"
@@ -503,6 +506,7 @@ set_default_params() {
   KIND_INSTALL_METALLB=${KIND_INSTALL_METALLB:-false}
   KIND_INSTALL_PLUGINS=${KIND_INSTALL_PLUGINS:-false}
   KIND_INSTALL_KUBEVIRT=${KIND_INSTALL_KUBEVIRT:-false}
+  KIND_OPT_OUT_KUBEVIRT_IPAM=${KIND_OPT_OUT_KUBEVIRT_IPAM:-false}
   OVN_HA=${OVN_HA:-false}
   KIND_LOCAL_REGISTRY=${KIND_LOCAL_REGISTRY:-false}
   KIND_LOCAL_REGISTRY_NAME=${KIND_LOCAL_REGISTRY_NAME:-kind-registry}
@@ -1167,5 +1171,7 @@ if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   deploy_passt_binary
 
   install_cert_manager
-  install_kubevirt_ipam_controller
+  if [ "$KIND_OPT_OUT_KUBEVIRT_IPAM" != true ]; then
+    install_kubevirt_ipam_controller
+  fi
 fi
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -1164,5 +1164,7 @@ fi
 if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   install_kubevirt
   deploy_primary_udn_nad
+  deploy_passt_binary
+
   install_kubevirt_ipam_controller
 fi
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -1166,5 +1166,6 @@ if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   deploy_primary_udn_nad
   deploy_passt_binary
 
+  install_cert_manager
   install_kubevirt_ipam_controller
 fi
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -1163,5 +1163,6 @@ if [ "$KIND_INSTALL_PLUGINS" == true ]; then
 fi
 if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   install_kubevirt
+  deploy_primary_udn_nad
   install_kubevirt_ipam_controller
 fi


RamLavi:
locally checked with this command:
```
pushd contrib/; sudo ./kind.sh --multi-network-enable -ic -nse -ikv; sudo cp /root/ovn.conf ~/.kube/config; cp -f ~/.kube/config ~/ovn.conf; popd
```

RamLavi:
Also checked east-west connectivity between 2 VMs.

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/f7e9a5944be562ca634ae641d2b8aebf0e94b7f3..be14514556f5bf00f54b23ce17058c188fa2e252): Rebase

oshoval:
Please consider to create a gist with a script which does the same thing, and that we can run manually,
it can assist on non kind clusters to do the same

RamLavi:
> Please consider to create a gist with a script which does the same thing, and that we can run manually, it can assist on non kind clusters to do the same

https://gist.github.com/RamLavi/a2edac1ae4b42b7251b1354ab9dda433

maiqueb (changes requested):
This works as expected. 

Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ? 

As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).

Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.

Thank you.

maiqueb:
> locally checked with this command:
> 
> ```
> pushd contrib/; sudo ./kind.sh --multi-network-enable -ic -nse -ikv; sudo cp /root/ovn.conf ~/.kube/config; cp -f ~/.kube/config ~/ovn.conf; popd
> ```

FWIW, it seems you don't use podman. Thus, you don't need to sudo.

RamLavi:
> This works as expected.
> 
> Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ?
> 
> As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).
> 
> Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.
> 
> Thank you.

do you mean add another flag like the one that adds kubevirt? 
Do we need to add it to kind-helm script as well (what is that script about?) ?

maiqueb:
> > This works as expected.
> > Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ?
> > As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).
> > Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.
> > Thank you.
> 
> do you mean add another flag like the one that adds kubevirt? Do we need to add it to kind-helm script as well (what is that script about?) ?

That script provides an OVN-K helm installer. 

Does kind-helm take into account the `install-kubevirt` flag ? If not, you can ignore that script for now.

If it does, we got to ask; @flavio-fernandes o/ can we skip adding this flag for now for helm ?

RamLavi:
> > > This works as expected.
> > > Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ?
> > > As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).
> > > Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.
> > > Thank you.
> > 
> > 
> > do you mean add another flag like the one that adds kubevirt? Do we need to add it to kind-helm script as well (what is that script about?) ?
> 
> That script provides an OVN-K helm installer.
> 
> Does kind-helm take into account the `install-kubevirt` flag ? If not, you can ignore that script for now.
> 
> If it does, we got to ask; @flavio-fernandes o/ can we skip adding this flag for now for helm ?

It [does](https://github.com/ovn-org/ovn-kubernetes/blob/master/contrib/kind-helm.sh#L470), but without ipam for some reason

RamLavi:
> > This works as expected.
> > Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ?
> > As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).
> > Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.
> > Thank you.
> 
> do you mean add another flag like the one that adds kubevirt? Do we need to add it to kind-helm script as well (what is that script about?) ?

how about instead of the kubevirt flag installing ipam without anyone asking:
```
if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
  install_kubevirt
  install_kubevirt_ipam_controller
fi
```
we separate it to a separate flag?
```
if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
  install_kubevirt
fi
if [ "$KIND_INSTALL_IPAM_CTRL" == true ]; then
  install_kubevirt_ipam_controller
fi
```

maiqueb:
> > > This works as expected.
> > > Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ?
> > > As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).
> > > Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.
> > > Thank you.
> > 
> > 
> > do you mean add another flag like the one that adds kubevirt? Do we need to add it to kind-helm script as well (what is that script about?) ?
> 
> how about instead of the kubevirt flag installing ipam without anyone asking:
> 
> ```
> if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
>   install_kubevirt
>   install_kubevirt_ipam_controller
> fi
> ```
> 
> we separate it to a separate flag?
> 
> ```
> if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
>   install_kubevirt
> fi
> if [ "$KIND_INSTALL_IPAM_CTRL" == true ]; then
>   install_kubevirt_ipam_controller
> fi
> ```

That's what I was proposing above: "could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed". 

I think the right way is actually what @oshoval proposed out of band: the flag would allow the dev to opt out having that installed. Thus, if the flag is passed, you do **not** install the ipam-extensions. 
Hence the suggested name `prevent-kubevirt-ipam-extensions-installation`. I'm sure you'll find a better name for it.

RamLavi:
> > > > This works as expected.
> > > > Still, to improve the dev experience, could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed ?
> > > > As I see it, that flag only makes sense when the install kubevirt knob is configured, and it allows the developer to install the ipam-extensions manually (i.e. whatever version they want, instead of the latest upstream).
> > > > Maybe name it `prevent-kubevirt-ipam-extensions-installation` or something like that.
> > > > Thank you.
> > > 
> > > 
> > > do you mean add another flag like the one that adds kubevirt? Do we need to add it to kind-helm script as well (what is that script about?) ?
> > 
> > 
> > how about instead of the kubevirt flag installing ipam without anyone asking:
> > ```
> > if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
> >   install_kubevirt
> >   install_kubevirt_ipam_controller
> > fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > we separate it to a separate flag?
> > ```
> > if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
> >   install_kubevirt
> > fi
> > if [ "$KIND_INSTALL_IPAM_CTRL" == true ]; then
> >   install_kubevirt_ipam_controller
> > fi
> > ```
> 
> That's what I was proposing above: "could you add an additional knob to prevent the `kubevirt_ipam_controller` from being installed".
> 
> I think the right way is actually what @oshoval proposed out of band: the flag would allow the dev to opt out having that installed. Thus, if the flag is passed, you do **not** install the ipam-extensions. Hence the suggested name `prevent-kubevirt-ipam-extensions-installation`. I'm sure you'll find a better name for it.

We all agree we want to be able to deploy IPAM separately..
What I'm trying to say - IMO it makes more sense to decouple kubevirt and IPAM controller, and make 2 "install kv", "install ipam" flags, instead of a "install kv, but really install kv+ipam" AND a "you know what, install only kv" flag..

oshoval:
> We all agree we want to be able to deploy IPAM separately.. What I'm trying to say - IMO it makes more sense to decouple kubevirt and IPAM controller, and make 2 "install kv", "install ipam" flags, instead of a "install kv, but really install kv+ipam" AND a "you know what, install only kv" flag..

imo keep it simple, dont separate, one time i tried and saw it is a bit mess (dont remember all the details)
and wasnt worth it
we can always have a follow-up, in this scope i would suggest to focus please

RamLavi:
> > We all agree we want to be able to deploy IPAM separately.. What I'm trying to say - IMO it makes more sense to decouple kubevirt and IPAM controller, and make 2 "install kv", "install ipam" flags, instead of a "install kv, but really install kv+ipam" AND a "you know what, install only kv" flag..
> 
> imo keep it simple, dont separate, one time i tried and saw it is a bit mess (dont remember all the details) and wasnt worth it we can always have a follow-up, in this scope i would suggest to focus please

 IMO separating => keeping it simple. 
 Adding an ignore flag is convoluted. Can you explain why you think otherwise?

oshoval:
separating PRs and keeping them focused is also a pro
i dont remember the details, and for me it seems not high priority to separate it
but better to be focused on the target
i am just trying to help and keep it focused

RamLavi:
> separating PRs and keeping them focused is also a pro i dont remember the details, and for me it seems not high priority to separate it but better to be focused on the target i am just trying to help and keep it focused

We're having a conversation about what is best, we have the same interests here.
@maiqueb I suggest we merge the content of this PR and I'll issue a PR right after that and try to decouple using separate flags. Shouldn't be that hard. if I fail as @oshoval suggests - I will promptly revert to a ignore flag. 
what say you?

oshoval:
another reason is to have less params
default once kubevirt is enable is to install all needed
and just opt out in case ipam is not required

otherwise one always need to say install kubevirt + install ipam
and then ipam cant be default true, unless you set kubevirt and then it make less sense to have them seperated

it is doable but i found it unnecessary so i skipped

maiqueb:
My thought process is when we have everything in place, it doesn't make much sense to have kubevirt without the ipam-extensions. Hence they're now one flag. 

We want now to have a separate dev flow; for that, it makes sense to separate the flags. Either an opt out one, or an opt in one. 

I really don't care much how it is added - given the nature of the other flags, it probably makes more sense for it to be opt in.

@RamLavi do it in this PR please. Let's try to disrupt our ovn-k colleagues as little as possible, and allow them to only bother with this once.

RamLavi:
> another reason is to have less params default once kubevirt is enable is to install all needed and just opt out in case ipam is not required
> 
> otherwise one always need to say install kubevirt + install ipam and then ipam cant be default true, unless you set kubevirt and then it make less sense to have them seperated
> 
> it is doable but i found it unnecessary so i skipped

I appreciate the thought you gave to this. 
IMO ipam != kv, it seems too convoluted to add an ignore flag to something that's deployed implicitly. 

BTW same logic can go for the passt components: Right now we always want to deploy them with kubevirt so it makes sense to add them implicitly via the "intall-kv" flag. but if we ever want to opt them out, it would make much more sense to  give them an opt-in flag.
In terms of number of flags - the sum is the same, we just swapped one opt-out flag for an opt-in one. 
You would need to explicitly ask for it, but that's what was requested.

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/pull/4655/commits/0f35fc3abf9da02c7b2aa098ac682ad1c9810189): Added commit to decouple ipam from kubevirt to allow deploying it out of band for dev purposes.

contrib/kind.sh
```diff
@@ -1163,5 +1167,9 @@ if [ "$KIND_INSTALL_PLUGINS" == true ]; then
 fi
 if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   install_kubevirt
+  deploy_primary_udn_nad
+  deploy_passt_binary
+fi
+if [ "$KIND_INSTALL_IPAM_CONTROLLER" == true ]; then
```
oshoval, line 1173: a by product is that whoever not installs ipam wont get cert-manager and would need to install it itself

i.e ipam-extensions

unless the cert-manager would moved as part of kubevirt

RamLavi: yes. The way I see it, only ipam-ext needs cert manager - so it's OK to only deploy cert-mgr if ipam-ext is requested.
If someone wants to deploy ipam-ext out of band (like @maiqueb) then they will have to also install cert manager themsleves.
The important thing is that e2e tests use the official ipam (with the official supported cert-manager for ipam).

oshoval: it will break on ipam-extensions
this is an action item that need to be solved please, this what i meant

RamLavi: ah I got you then. If I understand what your mean - then it's not related to this PR right?

well we can decide to include that cert-manager manifest on the ipam-repo, then you could align them here, possibly diverging them.

oshoval: It is part of the task you are on, i dont see why it should be delegated to someone else

contrib/kind.sh
```diff
@@ -163,6 +163,8 @@ parse_args() {
                                                 ;;
             -ikv | --install-kubevirt)          KIND_INSTALL_KUBEVIRT=true
                                                 ;;
+            -ipam | --install-ipam-controller) KIND_INSTALL_IPAM_CONTROLLER=true
```
maiqueb, line 166: we can't have this flag name. 

We have IPAM already; it is misleading. What we want to have is IPAM for virt. Maybe that name is good enough.

contrib/kind.sh
maiqueb: Maybe that's the correct thing to do - have ipam-extensions deploy its dependencies.

contrib/kind.sh
oshoval: nit: please also fix ident (at least looks like that)

RamLavi: suggestions for new name? `ipam-ext`? `ipam-kv`?

maiqueb: kubevirt-ipam ?

contrib/kind.sh
RamLavi: I think it's more complicated than that. cert-manager by design is not always used. Adding it to the flow may have consequences on other components, so this needs to thought through.

oshoval: While theoretically true, the way the repos work atm (OVN K8s, CNAO, HCO, ipam-ext), on upstream and on OpenShift, should be suitable for deploying cert-manager on ipam-extensions and thats it.
Each component handles it different, and as far as i know it should be fine.

You might want to check deeper, or we can just do a follow-up upon needs.
Note please it might be better to do it on cluster-sync rather than cluster-up, so it will also work for external providers.

contrib/kind.sh
RamLavi: short flag is `-kvipam` 
long flag is `--install-kubevirt-ipam`

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/cbd8535a76e9954701124ed7c9385ae8dca74115..9997e659619b69a4175c87d5bee268dc7da6ea82): Change flags and const names to be more clear

contrib/kind.sh
oshoval: please add the above info to the PR desc
it is helpful that PR desc has a TL;DR implicit user guide style

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/9997e659619b69a4175c87d5bee268dc7da6ea82..bd1c2231f2bda37dc1f2e44b4d41c315aaf72f99): Due to @maiqueb request to have cert-manager always installed when kubevirt is requested, it makes more sense to change the new flag behavior to only opt-out ipam-extentions in case requested.

contrib/kind.sh
```diff
@@ -1163,5 +1167,11 @@ if [ "$KIND_INSTALL_PLUGINS" == true ]; then
 fi
 if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   install_kubevirt
-  install_kubevirt_ipam_controller
+  deploy_primary_udn_nad
+  deploy_passt_binary
+
+  install_cert_manager
+  if [ ! "$KIND_OPT_OUT_KUBEVIRT_IPAM" == true ]; then
```
oshoval, line 1174: use `!= true` instead `! x == true` please

oshoval, line 1174: worth to add please `KIND_OPT_OUT_KUBEVIRT_IPAM == true` to ipam-extensions repo, it will improve the flow there
no need to deploy HEAD and then cluster-sync the PR itself on PRs as done now
but just cluster-sync the PR, once this flag is used there
it improves stability and saves time

contrib/kind.sh
```diff
@@ -1163,5 +1167,11 @@ if [ "$KIND_INSTALL_PLUGINS" == true ]; then
 fi
 if [ "$KIND_INSTALL_KUBEVIRT" == true ]; then
   install_kubevirt
-  install_kubevirt_ipam_controller
+  deploy_primary_udn_nad
+  deploy_passt_binary
+
+  install_cert_manager
```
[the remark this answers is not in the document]
RamLavi: Right now IPAM repo does not deploy cert-manager at all - leaving that to the user consuming it. 
That makes some sense, as not all users need to manually deploy it (i.e. they get it "for free" from their cluster)

I suggest we keep that behavior as we want to give @maiqueb something fast and easy for him to develop on. Adding a script to ipam-ext so that users can deploy it from the ipam-ext repo can be a follow up, indeed, but let's not condition it for this PR.

contrib/kind.sh
RamLavi: DONE

RamLavi: answered [above](https://github.com/ovn-org/ovn-kubernetes/pull/4655/commits/bd1c2231f2bda37dc1f2e44b4d41c315aaf72f99#r1736279474).

contrib/kind.sh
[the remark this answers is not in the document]
oshoval: > Right now IPAM repo does not deploy cert-manager at all - leaving that to the user consuming it. That makes some sense, as not all users need to manually deploy it (i.e. they get it "for free" from their cluster)

Thats not correct, it doesnt deploy it because it vendor OVN k8s and use kind.sh
with DEPLOY_KUBEVIRT, which deploys kubevirt+ipam+cert

Bundling the cert-manger with the IPAM, will make OVN-k8s cleaner,
and it will need to just deploy cert-manager once it opt-out IPAM
in the future it will even allow to opt out Kubevirt and install custom one.

For users that don't opt-out same default behavior is kept, so there is no reason to not do the change in this PR, right ? (i.e just move cert-manager next to ipam)
This way we can do it in steps, where each step is self contained and clean.

contrib/kind.sh
oshoval: This is orthogonal of https://github.com/ovn-org/ovn-kubernetes/pull/4655#discussion_r1736279474
if thats what you mean, and can be done on a follow-up PR under ipam-ext
(the link you posted doesnt work)

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/bd1c2231f2bda37dc1f2e44b4d41c315aaf72f99..589acc8e6990e58bef0f9c4489e8e5fa064e384f): fix condition logic format

contrib/kind.sh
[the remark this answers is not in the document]
RamLavi: Way I see it, we have 2 options here:
1. will consist of the following:
- move cert-manager to ipam-ext repo (will have a separate Makefile target). 
- remove cert-manager install from all the places it is currently deployed manually (CNAO when bumped, OVNK, etc..).

OR as suggested currently on this PR
2. ipam-ext doesn't deploy cert-manager. ovn-k will always deploy cert-manager. 

@oshoval am I right?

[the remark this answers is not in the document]
maiqueb: I would go with option 2 for now, and if needed, change it later on.

ipam-extensions shouldn't have an opinion of what certificate tool to use.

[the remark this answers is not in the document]
RamLavi: @maiqueb this PR currently does exactly that.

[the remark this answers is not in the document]
oshoval: Option [1] is not correct, only repo that depends on cert-manager that is deployed by OVN K8s kind.sh
is ipam-extensions, so we should not touch any other beside that.
Option [3] is what i suggested [above](https://github.com/ovn-org/ovn-kubernetes/pull/4655#discussion_r1736292990), it is better than [2] but will require a little follow-up
Miguel is the Captain here.

For completeness note please i also suggested [here](https://github.com/ovn-org/ovn-kubernetes/pull/4655#discussion_r1735841627) a solution for the OPT-IN in case desired in the future.

contrib/kind.sh
oshoval: still missing please i think, even that marked with emoji

maiqueb (dismissed):
I think this is good enough for what we need.

maiqueb:
@RamLavi could you rebase ? Maybe the unit test failure will go away.

Or is there something wrong with this PR ?

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/589acc8e6990e58bef0f9c4489e8e5fa064e384f..f94fb01a5a59a81f5d4e6a60ee426a39d9987bdb): Rebase

RamLavi:
> @RamLavi could you rebase ? Maybe the unit test failure will go away.

DONE

> 
> Or is there something wrong with this PR ?

Shouldn't be. The failed tests seem unrelated.

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/f94fb01a5a59a81f5d4e6a60ee426a39d9987bdb..62dedb7a541c78ace484b094597da50e50ca45f4): Rebase

oshoval:
not related to this PR but we should make the git actions jobs rebase themselves (doable)

contrib/kind-common
```diff
@@ -576,3 +582,32 @@ kubectl_wait_dnsnameresolver_pods() {
   echo "Waiting for pods in dnsnameresolver-operator namespace to become ready (timeout ${timeout})..."
   kubectl wait -n dnsnameresolver-operator --for=condition=ready pods --all --timeout=${timeout}s
 }
+
+deploy_primary_udn_nad() {
+  cat <<EOF | run_kubectl create -f -
```
oshoval, line 587: please make sure create (or any of the changes) doesnt fail on redeploy
(this is create that might fail if already exists iiuc)
`./kind.sh -mne --deploy` which reuse the kind cluster but redeploy everything else

contrib/kind-common
```diff
@@ -348,6 +348,10 @@ install_kubevirt() {
             kubectl logs --all-containers=true -n kubevirt $p || true
         done
     fi
+
+    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/developerConfiguration","value":{"featureGates":[]}},{"op":"add","path":"/spec/configuration/developerConfiguration/featureGates/-","value":"NetworkBindingPlugins"}]'
+    local passt_binding_image="quay.io/kubevirt/network-passt-binding:${kubevirt_version}"
```
oshoval, line 353: nit: worth to add please a new line above this line (since other changes are likely to be done anyhow https://github.com/ovn-org/ovn-kubernetes/pull/4655/files#r1741485104

contrib/kind-common
RamLavi: that could be it. I couldn't figure out why we start failing on other unrelated lanes, but if kind does this kind of thing then it could explain it. Let's change to apply and see.

oshoval: `--deploy` is not used by anyone beside local
apply does fix it (tried locally)
I am not sure about the other lanes at all, but i bet it wont be because of this change specifically (apply / create)

contrib/kind-common
RamLavi: DONE

contrib/kind-common
RamLavi: This script contains other create commands, so I'm not sure. 
But it doesn't hurt to try.

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/62dedb7a541c78ace484b094597da50e50ca45f4..0a046b17648a18b84862a7869b3fb4fdd5a03cf9): changed following @oshoval review.

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/caf5a16fa1db186709567395b2e602b711865ef0..b81a7283c186b208b80015d8e77e051186d4ec65): Rebase

oshoval:
@qinqon 
since you raised we would need to add UDN e2e tests on ipam-ext repo as well
maybe we should consider adding the NAD creation / kubevirt patch / Passt DaemonSet creation
commands as a script on ipam-ext, instead on this PR, and then both ipam-ext can OVN-k8s can use it from ipam-ext repo ?
else it would be duplication of bits
It will also allow external users to perform those bits on custom clusters.

RamLavi:
[Change](https://github.com/ovn-org/ovn-kubernetes/compare/b81a7283c186b208b80015d8e77e051186d4ec65..b6cc2c5071de318303017233e99b9468744e2b77): Rabase

oshoval:
can be closed as was merged as part of https://github.com/ovn-org/ovn-kubernetes/pull/4651 right ?

maiqueb:
Superseded by #4651

PR #4658, opened 2024-08-27
Update Dockerfile.fedora to use pre-released 24.09 ovn rpm.

npinaeva:
Depends on https://github.com/openshift/ovn-kubernetes/pull/2269 to be merged

### DIFF
--- a/dist/images/Dockerfile.fedora
+++ b/dist/images/Dockerfile.fedora
@@ -9,13 +9,13 @@
 # are built locally and included in the image (instead of the rpm)
 #
 
-FROM fedora:39
+FROM fedora:rawhide
 
 USER root
 
 ENV PYTHONDONTWRITEBYTECODE yes
 
-ARG ovnver=ovn-24.03.2-19.fc39
+ARG ovnver=ovn-24.03.90-6.fc42
 # Automatically populated when using docker buildx
 ARG TARGETPLATFORM
 ARG BUILDPLATFORM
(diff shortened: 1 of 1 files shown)


tssurya:
I don't think this PR needs to wait for ds merge to happen only the piece that uses new OVN this needs to be blocked which in this case is the UDN piece and observability piece right?

npinaeva:
> I don't think this PR needs to wait for ds merge to happen only the piece that uses new OVN this needs to be blocked which in this case is the UDN piece and observability piece right?

that is right, the only difference is that it seems easier to link dependency on the upstream PRs using newest OVN to this PR than to the openshift PR. If a different order will save someone time, we can do it vise versa.

tssurya:
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10581432241/job/29320023378?pr=4658
```
Summarizing 1 Failure:
  [FAIL] Services when a nodePort service targeting a pod with hostNetwork:false is created when tests are run towards the agnhost echo service [It] queries to the nodePort service shall work for UDP
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/service.go:484
```
this is not a known flake ^

its happening in lanes with enable-dns thing set..

npinaeva:
not sure if we need a separate bug for UDP, but in this flake for TCP https://github.com/ovn-org/ovn-kubernetes/issues/3932, I have reported a similar failure https://github.com/ovn-org/ovn-kubernetes/issues/3932#issuecomment-2275213318 here https://github.com/ovn-org/ovn-kubernetes/actions/runs/10290466858/job/28482169763?pr=4596#logs

dist/images/Dockerfile.fedora
```diff
@@ -24,14 +24,14 @@ RUN echo "Running on $BUILDPLATFORM, building for $TARGETPLATFORM"
...
 RUN INSTALL_PKGS=" \
-	python3-pip python3-pyyaml bind-utils procps-ng openssl numactl-libs firewalld-filesystem \
-	libpcap hostname kubernetes-client util-linux \
-        ovn ovn-central ovn-host python3-openvswitch tcpdump openvswitch-test python3-pyOpenSSL \
-	iptables iproute iputils strace socat koji \
-        libreswan openvswitch-ipsec \
-        " && \
-	dnf install --best --refresh -y --setopt=tsflags=nodocs $INSTALL_PKGS && \
-	dnf clean all && rm -rf /var/cache/dnf/*
+    python3-pip python3-pyyaml bind-utils procps-ng openssl numactl-libs firewalld-filesystem \
```
dceara, line 27: This part is not really needed.  No need to skip the openvswitch-ipsec package (I'm not sure if we use it anywhere - probably not in CI - but who knows).

dceara (changes requested):
The bump is fine, just a small nit about the unnecessary package list change.

dist/images/Dockerfile.fedora
npinaeva: I will leave formatting changes though, these tabs look bad in my editor :)

dceara (approved):
LGTM, thanks!

trozet (approved):

PR #4662, opened 2024-08-29
Make GetActiveNetworkForNamespace use a controller

trozet:
Leverages the NAD Controller to also track primary networks. This should be faster than listing all NADs in a namespace and iterating through them to find the primary network for a namespace.

### DIFF
--- a/go-controller/pkg/clustermanager/pod/allocator.go
+++ b/go-controller/pkg/clustermanager/pod/allocator.go
@@ -14,11 +14,11 @@ import (
 
 	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
-	nadlister "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/id"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/ip/subnet"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/pod"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	nad "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/network-attach-def-controller"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/persistentips"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
@@ -41,7 +41,7 @@ type PodAllocator struct {
 
 	ipamClaimsReconciler persistentips.PersistentAllocations
 
-	nadLister nadlister.NetworkAttachmentDefinitionLister
+	nadController nad.NADController
 
 	// event recorder used to post events to k8s
 	recorder record.EventRecorder
@@ -58,7 +58,7 @@ func NewPodAllocator(
 	podAnnotationAllocator *pod.PodAnnotationAllocator,
 	ipAllocator subnet.Allocator,
 	claimsReconciler persistentips.PersistentAllocations,
-	nadLister nadlister.NetworkAttachmentDefinitionLister,
+	nadController nad.NADController,
 	recorder record.EventRecorder,
 ) *PodAllocator {
 	podAllocator := &PodAllocator{
@@ -66,7 +66,7 @@ func NewPodAllocator(
 		releasedPods:           map[string]sets.Set[string]{},
 		releasedPodsMutex:      sync.Mutex{},
 		podAnnotationAllocator: podAnnotationAllocator,
-		nadLister:              nadLister,
+		nadController:          nadController,
 		recorder:               recorder,
 	}
 
@@ -107,11 +107,11 @@ func (a *PodAllocator) Init() error {
 }
 
 // getActiveNetworkForNamespace returns the active network for the given pod's namespace
-// and is a wrapper around util.GetActiveNetworkForNamespace
+// and is a wrapper around GetActiveNetworkForNamespace
 func (a *PodAllocator) getActiveNetworkForPod(pod *corev1.Pod) (util.NetInfo, error) {
-	activeNetwork, err := util.GetActiveNetworkForNamespace(pod.Namespace, a.nadLister)
+	activeNetwork, err := a.nadController.GetActiveNetworkForNamespace(pod.Namespace)
 	if err != nil {
-		if util.IsUnknownActiveNetworkError(err) {
+		if util.IsUnprocessedActiveNetworkError(err) {
 			a.recordPodErrorEvent(pod, err)
 		}
 		return nil, err
--- a/go-controller/pkg/cni/cniserver.go
+++ b/go-controller/pkg/cni/cniserver.go
@@ -18,6 +18,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
+	nad "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/network-attach-def-controller"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
@@ -48,10 +49,12 @@ import (
 // started.
 
 // NewCNIServer creates and returns a new Server object which will listen on a socket in the given path
-func NewCNIServer(factory factory.NodeWatchFactory, kclient kubernetes.Interface) (*Server, error) {
+func NewCNIServer(factory factory.NodeWatchFactory, kclient kubernetes.Interface,
+	nadController *nad.NetAttachDefinitionController) (*Server, error) {
 	if config.OvnKubeNode.Mode == types.NodeModeDPU {
 		return nil, fmt.Errorf("unsupported ovnkube-node mode for CNI server: %s", config.OvnKubeNode.Mode)
 	}
+
 	router := mux.NewRouter()
 
 	s := &Server{
@@ -72,7 +75,7 @@ func NewCNIServer(factory factory.NodeWatchFactory, kclient kubernetes.Interface
 	}
 
 	if util.IsNetworkSegmentationSupportEnabled() {
-		s.clientSet.nadLister = factory.NADInformer().Lister()
+		s.nadController = nadController
 	}
 
 	if len(config.Kubernetes.CAData) > 0 {
@@ -218,7 +221,7 @@ func (s *Server) handleCNIRequest(r *http.Request) ([]byte, error) {
 	}
 	defer req.cancel()
 
-	result, err := s.handlePodRequestFunc(req, s.clientSet, s.kubeAuth)
+	result, err := s.handlePodRequestFunc(req, s.clientSet, s.kubeAuth, s.nadController)
 	if err != nil {
 		// Prefix error with request information for easier debugging
 		return nil, fmt.Errorf("%s %v", req, err)
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -212,18 +212,6 @@ func NewMasterWatchFactory(ovnClientset *util.OVNMasterClientset) (*WatchFactory
 		}
 	}
 
-	if util.IsNetworkSegmentationSupportEnabled() {
-		if err := userdefinednetworkapi.AddToScheme(userdefinednetworkscheme.Scheme); err != nil {
-			return nil, err
-		}
-
-		wf.udnFactory = userdefinednetworkapiinformerfactory.NewSharedInformerFactory(ovnClientset.UserDefinedNetworkClient, resyncInterval)
-		wf.informers[UserDefinedNetworkType], err = newInformer(UserDefinedNetworkType, wf.udnFactory.K8s().V1().UserDefinedNetworks().Informer())
-		if err != nil {
-			return nil, err
-		}
-	}
-
 	return wf, nil
 }
 
@@ -403,6 +391,14 @@ func NewOVNKubeControllerWatchFactory(ovnClientset *util.OVNKubeControllerClient
 			if err != nil {
 				return nil, err
 			}
+		}
+	}
+
+	if util.IsNetworkSegmentationSupportEnabled() {
+		wf.udnFactory = userdefinednetworkapiinformerfactory.NewSharedInformerFactory(ovnClientset.UserDefinedNetworkClient, resyncInterval)
+		wf.informers[UserDefinedNetworkType], err = newInformer(UserDefinedNetworkType, wf.udnFactory.K8s().V1().UserDefinedNetworks().Informer())
+		if err != nil {
+			return nil, err
 		}
 	}
 
@@ -704,6 +700,14 @@ func NewNodeWatchFactory(ovnClientset *util.OVNNodeClientset, nodeName string) (
 	if config.OVNKubernetesFeature.EnableMultiNetwork || config.OVNKubernetesFeature.EnableNetworkSegmentation {
 		wf.nadFactory = nadinformerfactory.NewSharedInformerFactory(ovnClientset.NetworkAttchDefClient, resyncInterval)
 		wf.informers[NetworkAttachmentDefinitionType], err = newInformer(NetworkAttachmentDefinitionType, wf.nadFactory.K8sCniCncfIo().V1().NetworkAttachmentDefinitions().Informer())
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if util.IsNetworkSegmentationSupportEnabled() {
+		wf.udnFactory = userdefinednetworkapiinformerfactory.NewSharedInformerFactory(ovnClientset.UserDefinedNetworkClient, resyncInterval)
+		wf.informers[UserDefinedNetworkType], err = newInformer(UserDefinedNetworkType, wf.udnFactory.K8s().V1().UserDefinedNetworks().Informer())
 		if err != nil {
 			return nil, err
 		}
--- a/go-controller/pkg/libovsdb/ops/switch.go
+++ b/go-controller/pkg/libovsdb/ops/switch.go
@@ -312,6 +312,9 @@ func createOrUpdateLogicalSwitchPortsOps(nbClient libovsdbclient.Client, ops []l
 	m := newModelClient(nbClient)
 	ops, err := m.CreateOrUpdateOps(ops, opModels...)
 	sw.Ports = originalPorts
+	if err != nil && errors.Is(err, libovsdbclient.ErrNotFound) && !createSwitch {
+		err = fmt.Errorf("could not find switch: %q, %w", sw.Name, err)
+	}
 	return ops, err
 }
 
--- a/go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
+++ b/go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -48,13 +49,20 @@ type watchFactory interface {
 	NADInformer() nadinformers.NetworkAttachmentDefinitionInformer
 }
 
+type NADController interface {
+	Start() error
+	Stop()
+	GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error)
+}
+
 // NetAttachDefinitionController handles namespaced scoped NAD events and
 // manages cluster scoped networks defined in those NADs. NADs are mostly
-// referred from pods to give them access to the network. Different NADs can
+// referenced from pods to give them access to the network. Different NADs can
 // define the same network as long as those definitions are actually equal.
 // Unexpected situations are handled on best effort basis but improper NAD
-// adminstration can lead to undefined behavior in referred from running pods.
+// administration can lead to undefined behavior if referenced by running pods.
 type NetAttachDefinitionController struct {
+	sync.RWMutex
 	name               string
 	netAttachDefLister nadlisters.NetworkAttachmentDefinitionLister
 	controller         controller.Controller
@@ -62,10 +70,11 @@ type NetAttachDefinitionController struct {
 	// networkManager is used to manage the network controllers
 	networkManager networkManager
 
-	networks map[string]util.NetInfo
-
 	// nads to network mapping
 	nads map[string]string
+
+	// primaryNADs holds a mapping of namespace to primary NAD names
+	primaryNADs map[string]string
 }
 
 func NewNetAttachDefinitionController(
@@ -77,17 +86,16 @@ func NewNetAttachDefinitionController(
 	nadController := &NetAttachDefinitionController{
 		name:           fmt.Sprintf("[%s NAD controller]", name),
 		recorder:       recorder,
-		networkManager: newNetworkManager(name, ncm),
-		networks:       map[string]util.NetInfo{},
 		nads:           map[string]string{},
+		primaryNADs:    map[string]string{},
+		networkManager: newNetworkManager(name, ncm),
 	}
 
 	config := &controller.ControllerConfig[nettypes.NetworkAttachmentDefinition]{
 		RateLimiter:    workqueue.DefaultControllerRateLimiter(),
 		Reconcile:      nadController.sync,
 		ObjNeedsUpdate: nadNeedsUpdate,
-		// this controller is not thread safe
-		Threadiness: 1,
+		Threadiness:    1,
 	}
 
 	nadInformer := wf.NADInformer()
@@ -183,6 +191,11 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 	var nadNetwork, oldNetwork, ensureNetwork util.NetInfo
 	var err error
 
+	namespace, _, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return fmt.Errorf("%s: failed splitting key %s: %v", nadController.name, key, err)
+	}
+
 	if nad != nil {
 		nadNetwork, err = util.ParseNADInfo(nad)
 		if err != nil {
@@ -196,6 +209,14 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 		nadNetworkName = nadNetwork.GetNetworkName()
 	}
 
+	nadController.Lock()
+	defer nadController.Unlock()
+	// We can only have one primary NAD per namespace
+	primaryNAD := nadController.primaryNADs[namespace]
+	if nadNetwork != nil && nadNetwork.IsPrimaryNetwork() && primaryNAD != "" && primaryNAD != key {
+		return fmt.Errorf("%s: NAD %s is primary for the namespace, NAD %s can't be primary", nadController.name, primaryNAD, key)
+	}
+
 	// As multiple NADs may define networks with the same name, these networks
 	// should also have the same config to be considered compatible. If an
 	// incompatible network change happens on NAD update, we can:
@@ -205,16 +226,14 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 
 	// the NAD refers to a different network than before
 	if nadNetworkName != nadController.nads[key] {
-		oldNetwork = nadController.networks[nadController.nads[key]]
+		oldNetwork = nadController.networkManager.getNetwork(nadController.nads[key])
 	}
-
-	currentNetwork := nadController.networks[nadNetworkName]
+	currentNetwork := nadController.networkManager.getNetwork(nadNetworkName)
 
 	switch {
 	case currentNetwork == nil:
 		// the NAD refers to a new network, ensure it
 		ensureNetwork = nadNetwork
-		nadController.networks[nadNetworkName] = ensureNetwork
 	case currentNetwork.Equals(nadNetwork):
 		// the NAD refers to an existing compatible network, ensure that
 		// existing network holds a reference to this NAD
@@ -243,7 +262,6 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 		oldNetwork.DeleteNADs(key)
 		if len(oldNetwork.GetNADs()) == 0 {
 			nadController.networkManager.DeleteNetwork(oldNetworkName)
-			delete(nadController.networks, oldNetworkName)
 		} else {
 			nadController.networkManager.EnsureNetwork(oldNetwork)
 		}
@@ -252,6 +270,9 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 	// this was a nad delete
 	if ensureNetwork == nil {
 		delete(nadController.nads, key)
+		if nadController.primaryNADs[namespace] == key {
+			delete(nadController.primaryNADs, namespace)
+		}
 		return err
 	}
 
@@ -260,11 +281,22 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 		return nil
 	}
 
-	// ensure the network associated with the NAD
+	// ensure the network is associated with the NAD
 	ensureNetwork.AddNADs(key)
 	nadController.nads[key] = ensureNetwork.GetNetworkName()
+	// track primary NAD
+	switch {
+	case ensureNetwork.IsPrimaryNetwork():
+		nadController.primaryNADs[namespace] = key
+	default:
+		if nadController.primaryNADs[namespace] == key {
+			delete(nadController.primaryNADs, namespace)
+		}
+	}
+
+	// reconcile the network
 	nadController.networkManager.EnsureNetwork(ensureNetwork)
-	return err
+	return nil
 }
 
 func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
@@ -279,4 +311,29 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
 	}
 
 	return !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec)
+}
+
+func (nadController *NetAttachDefinitionController) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	if !util.IsNetworkSegmentationSupportEnabled() {
+		return &util.DefaultNetInfo{}, nil
+	}
+	nadController.RLock()
+	defer nadController.RUnlock()
+	primaryNAD := nadController.primaryNADs[namespace]
+	if primaryNAD != "" {
+		// we have a primary NAD, get the network
+		netName := nadController.nads[primaryNAD]
+		if netName == "" {
+			// this should never happen where we have a nad keyed in the primaryNADs
+			// map, but it doesn't exist in the nads map
+			panic("NAD Controller broken consistency between primary NADs and cached NADs")
+		}
+		network := nadController.networkManager.getNetwork(netName)
+		n := util.CopyNetInfo(network)
+		// update the returned netInfo copy to only have the primary NAD for this namespace
+		n.SetNADs(primaryNAD)
+		return n, nil
+	}
+
+	return &util.DefaultNetInfo{}, nil
 }
--- a/go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
+++ b/go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
@@ -21,6 +21,8 @@ import (
 	nadinformers "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/informers/externalversions/k8s.cni.cncf.io/v1"
 	nadlisters "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controller"
+	userdefinednetworkinformer "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/informers/externalversions/userdefinednetwork/v1"
+	userdefinednetworklister "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/listers/userdefinednetwork/v1"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
 
@@ -47,6 +49,7 @@ type NetworkControllerManager interface {
 
 type watchFactory interface {
 	NADInformer() nadinformers.NetworkAttachmentDefinitionInformer
+	UserDefinedNetworkInformer() userdefinednetworkinformer.UserDefinedNetworkInformer
 }
 
 type NADController interface {
@@ -65,6 +68,7 @@ type NetAttachDefinitionController struct {
 	sync.RWMutex
 	name               string
 	netAttachDefLister nadlisters.NetworkAttachmentDefinitionLister
+	udnLister          userdefinednetworklister.UserDefinedNetworkLister
 	controller         controller.Controller
 	recorder           record.EventRecorder
 	// networkManager is used to manage the network controllers
@@ -103,6 +107,13 @@ func NewNetAttachDefinitionController(
 		nadController.netAttachDefLister = nadInformer.Lister()
 		config.Informer = nadInformer.Informer()
 		config.Lister = nadController.netAttachDefLister.List
+	}
+	if util.IsNetworkSegmentationSupportEnabled() {
+		udnInformer := wf.UserDefinedNetworkInformer()
+
+		if udnInformer != nil {
+			nadController.udnLister = udnInformer.Lister()
+		}
 	}
 
 	nadController.controller = controller.NewController(
@@ -333,6 +344,17 @@ func (nadController *NetAttachDefinitionController) GetActiveNetworkForNamespace
 		// update the returned netInfo copy to only have the primary NAD for this namespace
 		n.SetNADs(primaryNAD)
 		return n, nil
+	}
+
+	// no primary network found, make sure we just haven't processed it yet and no UDN exists
+	udns, err := nadController.udnLister.UserDefinedNetworks(namespace).List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error getting user defined networks: %w", err)
+	}
+	for _, udn := range udns {
+		if util.IsPrimaryNetwork(udn.Spec) {
+			return nil, util.NewUnprocessedActiveNetworkError(namespace, udn.Name)
+		}
 	}
 
 	return &util.DefaultNetInfo{}, nil
--- a/go-controller/pkg/network-attach-def-controller/network_attach_def_controller_test.go
+++ b/go-controller/pkg/network-attach-def-controller/network_attach_def_controller_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"sync"
 	"testing"
 
@@ -31,6 +32,7 @@ type testNetworkController struct {
 func (tnc *testNetworkController) Start(context.Context) error {
 	tnc.tncm.Lock()
 	defer tnc.tncm.Unlock()
+	fmt.Printf("starting network: %s\n", testNetworkKey(tnc))
 	tnc.tncm.started = append(tnc.tncm.started, testNetworkKey(tnc))
 	return nil
 }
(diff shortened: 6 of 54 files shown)


trozet:
I should probably add some unit tests for network attachment definition controller, but can at least run this through scale and see if its an improvement. 

@tssurya @jcaamano @dceara fyi

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +331,28 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
 	return !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec)
 }
+
+func (nadController *NetAttachDefinitionController) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	if !util.IsNetworkSegmentationSupportEnabled() {
+		return &util.DefaultNetInfo{}, nil
+	}
+	nadController.primaryNetworks.LockKey(namespace)
+	defer nadController.primaryNetworks.UnlockKey(namespace)
+	primaryNetworks, _ := nadController.primaryNetworks.Load(namespace)
```
jcaamano, line 354: What happens if the mapping of `network name -> NetInfo` of `nadController.primaryNetworks` disagrees with that on `nadController.networks`? No consequence?
Have you considered to keep `nadController.primaryNetworks` just as a map of `namespace -> network names` and then rely and keep in sync `nadController.networks` to get the actual `NetInfo`?

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +331,28 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
+	if len(primaryNetworks) > 1 {
+		return nil, util.NewUnknownActiveNetworkError(namespace)
```
jcaamano, line 356: Should the UDN CRD prevent this from happening? If so, but NADs are being used directly with no UDN CRD, should we really care about NADs misuse? 
Also, if this happens, why wouldn't we just fail the reconciliation?

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +331,28 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
+	}
+
+	return &util.DefaultNetInfo{}, nil
+
+}
+
+// SetPrimaryNetworksForTest is used for testing purposes only
+func (nadController *NetAttachDefinitionController) SetPrimaryNetworksForTest(primaryNetworks *syncmap.SyncMap[map[string]util.NetInfo]) {
+	nadController.primaryNetworks = primaryNetworks
+}
```
jcaamano, line 371: Why not have an interface for `GetActiveNetworkForNamespace` and then have a test specific implementation?

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -117,9 +125,11 @@ func (nadController *NetAttachDefinitionController) Start() error {
 		return err
 	}
 
-	err = nadController.networkManager.Start()
-	if err != nil {
-		return err
+	if nadController.networkManager != nil {
```
jcaamano, line 128: In which cases do we have a nil `networkManager`?  Is this test specific?

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -260,10 +308,26 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
...
 
-	// ensure the network associated with the NAD
 	ensureNetwork.AddNADs(key)
+
+	if ensureNetwork.IsPrimaryNetwork() {
+		nadController.primaryNetworks.LockKey(namespace)
+		var primaryNetInfos map[string]util.NetInfo
+		var loaded bool
+		if primaryNetInfos, loaded = nadController.primaryNetworks.Load(namespace); loaded {
+			primaryNetInfos[ensureNetwork.GetNetworkName()] = util.CopyNetInfo(ensureNetwork)
```
jcaamano, line 318: In regards to this CopyNetInfo, I just checked that the NetInfo NAD methods are synchronized:

https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/util/multi_network.go#L364-L397

and the rest is immutable. So I am not sure what kind of race you got there.

jcaamano (commented):
Just comments on the NAD controller first.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +331,28 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
 
 	return !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec)
 }
+
+func (nadController *NetAttachDefinitionController) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
```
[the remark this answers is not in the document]
trozet: We don't support a user changing the active network unless all objects in the namespace are deleted.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: wont that break the suggestion you gave me to use a nil struct method call?

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +331,28 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
+		return nil, util.NewUnknownActiveNetworkError(namespace)
+	} else if len(primaryNetworks) == 1 {
+		for _, netInfo := range primaryNetworks {
+			// only 1 value
+			return netInfo, nil
+		}
+	}
+
+	return &util.DefaultNetInfo{}, nil
+
```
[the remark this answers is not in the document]
trozet: done

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -247,6 +257,33 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
...
+				} else if len(primaryNetworks) > 1 {
+					// multiple primary networks for this namespace, just remove the one
+					delete(primaryNetworks, oldNetworkName)
+					nadController.primaryNetworks.Store(namespace, primaryNetworks)
+				} else {
+					// only 1 network, delete the entire namespace from being tracked
+					nadController.primaryNetworks.Delete(namespace)
+				}
+			}
+			nadController.primaryNetworks.UnlockKey(namespace)
```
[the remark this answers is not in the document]
trozet: done

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: so I tried this and its going to make things more complicated. This is because nadController.networks has no protected parallel access. In fact, nadController itself cannot be run with more than one thread right now. Therefore I would need to introduce either a syncmap for nadController.networks or add a mutex to the nadController itself.

Let's say I just took the simple route and turned the primaryNetworks into a regular map, and used a nadController mutex to lock everything. Now threads all over OVNK trying to get the active network are going to be fighting for the global lock and this is going to cause performance issues. So imo it makes more sense to optimize locking.

To optimize locking I need nadController.networks to have a syncmap lockable key. However, now to get the network from primaryNetworks I need to lock both syncmaps (primaryNetworks and nadController.networks) keys. So we could get into a deadlock scenario with order of locking each of these maps. I'd rather take the tradeoff of duplicate netinfos stored rather than risk complicated dead locks by having maps depend on each other.

Perhaps it would make more sense to change the nadController.networks to contain keys for primary and namespace. For now im just going to leave it like it is if you dont object.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: yeah it shouldnt happen if the user only uses UDN CRs. This check was added by @qinqon I think before we had the UDN CRD. My objective in this PR was to just keep the current get method as is and migrate it to use the controller, so I didn't change this logic. I don't think it hurts to keep it, but either way should probably be addressed in a separate PR.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: this was originally added so that I could start a nadcontroller without a network manager for CNI. Now that the nadcontroller from the node is extended to CNI, I can probably remove it.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: I wish I had saved it. I also noticed that and was really confused at first how the race was happening. Then I figured it out and now I've forgotten it. I'll try to remember/reproduce it.

trozet: So I managed to reproduce something. I'm guessing this is it but I cant be sure:

https://gist.github.com/trozet/1ebc50a7c4919f778c5c08c227f60c39

This looks like it is racing because the test is doing a reflect deepEqual which doesnt care about the netinfo mutex while the network manager controller is trying to update NADs and has the mutex.

Either way I think it is a bad idea to be propagating pointers between controller caches. I'd like to keep the copying of the netinfo in place unless you have an objection.

qinqon:
Hey @trozet maybe labeling the NAD when it's created by UserDefinedNetwork controller with role=primary alow for  fast listing instead of iterating ?.

trozet:
> Hey @trozet maybe labeling the NAD when it's created by UserDefinedNetwork controller with role=primary alow for fast listing instead of iterating ?.

not sure what iterating you are referring to

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: Hmm okay, I agree that it might not be the best idea to be propagating pointers between controller caches. 

But I am not sure this specific copy will solve the issue. It looks like it is the test which should copy before doing deep equal.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -247,6 +258,11 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 		} else {
 			nadController.networkManager.EnsureNetwork(oldNetwork)
 		}
+		// if oldNetwork is primary, and we are not going to update the active network, delete it
+		// as long as no other NADs reference it
```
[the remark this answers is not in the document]
jcaamano: Although it seems all of the code is protected against receiving NAD events out of order somehow, since we can have multiple primary network per namespace in the map, so also probably worth commenting this detail.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: Ok, it's just that it complicates things a bit not just here but also `unsetActiveNetworkForNamespace`.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: Well the advantage is not delivering test code to production. Bit by bit might not be much but if we keep doing it we bloat unnecessarily. It's true that sometimes it can't be avoided or the effort to do so would be unreasonable. 

`FakeNetworkController` might be another example? If that's just used in tests, shouldn't it go in `pkg/testing` somewhere?

Another potential advantage is not having people calling other current or future methods of nad controller. This is minor, I wasn't really thinking about this one.

Anyway, I don't wanna push in our urgent circumstances.

go-controller/pkg/util/multi_network.go
```diff
@@ -943,7 +943,7 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, ac
 	// Add the active network to the NSE map if it is configured
 	activeNetworkNADs := activeNetwork.GetNADs()
 	if len(activeNetworkNADs) < 1 {
-		return false, nil, fmt.Errorf("missing NADs at active network '%s' for namesapce '%s'", activeNetwork.GetNetworkName(), pod.Namespace)
+		return false, nil, fmt.Errorf("missing NADs at active network '%s' for namespace '%s'", activeNetwork.GetNetworkName(), pod.Namespace)
```
jcaamano, line 946: nit: you can just use `%q` instead of `'%s'`

go-controller/pkg/cni/cniserver.go
```diff
@@ -72,6 +75,9 @@ func NewCNIServer(factory factory.NodeWatchFactory, kclient kubernetes.Interface
 	}
 
 	if util.IsNetworkSegmentationSupportEnabled() {
+		if !config.UnprivilegedMode {
+			s.nadController = nadController
+		}
```
jcaamano, line 80: Not sure what this is really doing. My ignorance but a comment would be welcome.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +301,90 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
+	var loaded bool
+	if primaryNetInfos, loaded = nadController.primaryNetworks.Load(namespace); loaded {
+		primaryNetInfos[ensureNetwork.GetNetworkName()] = ensureNetwork
+	} else {
+		primaryNetInfos = map[string]util.NetInfo{ensureNetwork.GetNetworkName(): ensureNetwork}
+	}
+	nadController.primaryNetworks.Store(namespace, primaryNetInfos)
+}
+
+func (nadController *NetAttachDefinitionController) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
```
jcaamano, line 345: So the `NetInfo` that we return here, wouldn't it just have all the NADs for the network instead of just eh NADs in the namespace? That's what the old `GetActiveNetworkForNamespace` was doing.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: I don't think that having a mutex that just guards an unimpeded  access to `nadController.networks` and `nadController.primaryNetworks` is going to hurt performance.

I think internally in order to lock a specific key for the sync map, you first need need to lock the map itself, lock the key, then unlock the map. So you are incurring in the same kind of penalty there, if any, with that map lock. 

The sync map just makes sense when you need the key locked for a meaningful amount of time I would say.

If I cannot convince you, then its fine to leave things as they are.

jcaamano:
Full review, so no more comments.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -75,13 +79,15 @@ func NewNetAttachDefinitionController(
...
-		networks:       map[string]util.NetInfo{},
-		nads:           map[string]string{},
+		name:            fmt.Sprintf("[%s NAD controller]", name),
+		recorder:        recorder,
+		networks:        map[string]util.NetInfo{},
+		nads:            map[string]string{},
+		primaryNetworks: syncmap.NewSyncMap[map[string]util.NetInfo](),
 	}
 
+	nadController.networkManager = newNetworkManager(name, ncm)
```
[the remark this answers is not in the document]
trozet: yeah leftover, i'll fix it

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -183,6 +189,11 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 	var nadNetwork, oldNetwork, ensureNetwork util.NetInfo
 	var err error
 
+	namespace, _, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return fmt.Errorf("%s: failed splitting key %s: %v", nadController.name, key, err)
```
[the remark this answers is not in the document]
trozet: that's true in a normal case where an event occurs, and would have already been caught and skipped in that case (before this syncNAD function is called):

```
func (nadController *NetAttachDefinitionController) sync(key string) error {
	startTime := time.Now()
	klog.V(5).Infof("%s: sync NAD %s", nadController.name, key)
	defer func() {
		klog.V(4).Infof("%s: finished syncing NAD %s, took %v", nadController.name, key, time.Since(startTime))
	}()

	namespace, name, err := cache.SplitMetaNamespaceKey(key)
	if err != nil {
		klog.Errorf("%s: failed splitting key %s: %v", nadController.name, key, err)
		return nil
	}

	nad, err := nadController.netAttachDefLister.NetworkAttachmentDefinitions(namespace).Get(name)
	if err != nil && !apierrors.IsNotFound(err) {
		return err
	}

	return nadController.syncNAD(key, nad)
}
```

The other place syncNAD is called is in syncAll, which also already splits and checks it:

```
func (nadController *NetAttachDefinitionController) syncAll() (err error) {
	existingNADs, err := nadController.netAttachDefLister.List(labels.Everything())
	if err != nil {
		return fmt.Errorf("%s: failed to list NADs: %w", nadController.name, err)
	}

	// create all networks with their updated list of NADs and only then start
	// the corresponding controllers so as to avoid to the extent possible the
	// errors and retries that would result if the controller attempted to
	// process pods attached with NADs we wouldn't otherwise know about yet
	for _, nad := range existingNADs {
		key, err := cache.MetaNamespaceKeyFunc(nad)
		if err != nil {
			klog.Errorf("%s: failed to sync %v: %v", nadController.name, nad, err)
			continue
		}
		err = nadController.syncNAD(key, nad)
		if err != nil {
			return fmt.Errorf("%s: failed to sync %s: %v", nadController.name, key, err)
		}
	}

	return nil
}
```

I think its fine to just leave it, but if you want I can change the arguments of the function to pass down the namespace if you want.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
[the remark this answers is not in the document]
trozet: I think my comment is outdated and should just say:
// if oldNetwork is primary, then delete it as long as no other NADs reference it

I'm not sure what your comment is describing. There can be multiple primary NADs per namespace (although there shouldn't be), and we have to handle this type of scenario. We only protect against multiple primary NADs in the case where a user only uses UDN CRD.

go-controller/pkg/util/multi_network.go
trozet: i just fixed the spelling error in namespace, but i'll also fix the quotes

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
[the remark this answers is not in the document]
jcaamano: no, it's fine

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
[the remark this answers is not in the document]
jcaamano: The specifically confusing part here is that the `no other NADs reference it` comment makes it look like the `unsetActiveNetworkForNamespace` call belongs in the `if len(oldNetwork.GetNADs()) == 0` block above, 

I know that's not true but the comment doesn't help.

go-controller/pkg/cni/cniserver.go
trozet: so in CNI we have 2 modes: unprivileged and privileged mode. In unprivileged mode, ovnkube-node doesn't have sys admin ability to be able to wire the interfaces. I can't remember if its also missing other capabilities. In this mode, the CNI shim (executed by kubelet) wires the interfaces vs. privileged mode, where cni shim sends a request to ovnkube cni server and it performs the wiring.

For UDN so far, we have only supported privileged mode:
https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/cni/cni.go#L165

Technically I don't really need to check here for unprivileged mode...I was just trying to not add a nadController unless I needed to (in order to not waste resources on the system), but now that it is inherited from the node itself there is no cost. I'll just remove this conditional.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: That's a good point. I'm returning all NADs associated with this network, rather than just the one scoped to the namespace (note there can be only one returned, cause otherwise we return the error for multiple primary NADs). I'll fix this.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: I see your point, but consider some other things. Right now nadController does very little work when it processes a NAD so it wont hold the lock for very long, especially since its single threaded. It takes nm.lock and then enqueues a key. However, network manager also takes this nm lock when it syncs an object. This can take a long time when its trying to configure everything in OVN.

So now imagine many UDNs, namespaces, pods/services, network policy being created at the same time. pod handlers have 15 threads. We could now have 10s of threads all trying to GetActiveNetworkForNamespace for all different namespaces. Now, if nm is syncing a namespace z, and all these other threads are trying to GetActiveNetworkForNamespace (namespaces A-Y), they are all blocked.

So the NAD controller/NM thread has single handedly stalled other core controllers in ovnk.

This is a similar problem to what we used to have with the infamous  namespaceInfo mutex. We had a ton of other controllers trying to get information about specific namespaces, but the global lock was blocking all of them. That's when I turned to try to implement lockable keys, and Nadia improved this by making a new syncmap type.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: the problem is the FakeNetworkController needs to be used by other packages testing code, and in golang you cant import test code to other test code. I guess I can use build tags and create a common package that is ignored when building ovnk, like using: // +build test

Let me know if you want me to try that.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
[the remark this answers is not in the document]
trozet: I'll just remove the comment then? There are comments in the unset function itself.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: But that's precisely why we have `pkg/testing`, to share testing code across tests. You just import that in your tests. Like what we do with mock packages and the such.

BTW there is already a `testNetworkControllerManager` in `network_attach_def_controller_test.go`. You might find convenient to just move that one to `pkg/testing` but maybe not, your choice.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: I think we wouldn't need for this new lock to guard all the processing of nadController. Just literally the access to the maps and that's it, because NetInfo has its own locks. And that's why I am thinking this critical region is comparable with the critical region you would already have with SyncMap.

BTW, you have a point on the `nm.lock`. Starting a controller can take time (it shouldn't, but unfortunately it might). However, the need of that lock is also only to protect access to `nm.networks` so it could guard only just that. Potential improvement.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: BTW, my second question:  why wouldn't we just fail the reconciliation in this case? Otherwise we would be passing a configuration to network manager that is invalid.

jcaamano:
I have one extra comment, sorry.

I think it would be useful to have a method `NetInfo.GetNads(namespace string) []string` that guarantees it only returns a single NAD for primary networks, for the network controllers to use.

It would be directly usable there, because we already have a NetInfo, to be aware of what is the NAD serving in that namespace. Otherwise we would have to use `GetActiveNetworkForNamespace` just to get a NetInfo with the filtered NADs, which means we would need to be passing around NAD controller to more places, including the family of utility functions `PodNadNames`, `GetPodIPsOfNetwork`, ...

The alternative implementation for those methods would be to parse the pod annotation and find the matching NAD from `NetInfo.GetNads()` but would a pity if we already have that info in memory.

jcaamano:
> I have one extra comment, sorry.
> 
> I think it would be useful to have a method `NetInfo.GetNads(namespace string) []string` that guarantees it only returns a single NAD for primary networks, for the network controllers to use.
> 
> It would be directly usable there, because we already have a NetInfo, to be aware of what is the NAD serving in that namespace. Otherwise we would have to use `GetActiveNetworkForNamespace` just to get a NetInfo with the filtered NADs, which means we would need to be passing around NAD controller to more places, including the family of utility functions `PodNadNames`, `GetPodIPsOfNetwork`, ...
> 
> The alternative implementation for those methods would be to parse the pod annotation and find the matching NAD from `NetInfo.GetNads()` but would a pity if we already have that info in memory.

Wait, there is an alternative implementation which is to parse the namespace from the nad namespaced names from `GetNads()` and match the that to the pod namespace. I still think `NetInfo.GetNads(namespace string) []string` woudl be useful but not as important. However, all this depends on some consistence I asked for in https://github.com/ovn-org/ovn-kubernetes/pull/4662#discussion_r1758575311

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: will do that

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: agreed. I'll add this

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: After our discussion I am posting a new version where we reduce the network tracking to just NetworkManager and remove it from NAD Controller, as well as modify locking to reduce contention on GetActiveNetworkForNamespace.

trozet:
flake?
```

24-09-19T02:03:14.1070652Z [91m[1m[Fail] [0m[90mHybrid SDN Master Operations [0m[91m[1m[It] handles a HO node is switched to a OVN node [0m
2024-09-19T02:03:14.1072499Z [37m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:1387[0m

2024-09-19T01:58:18.6956889Z [91m[1m• Failure [2.579 seconds][0m
2024-09-19T01:58:18.6957661Z Hybrid SDN Master Operations
2024-09-19T01:58:18.6958864Z [90m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:134[0m
2024-09-19T01:58:18.6960288Z   [91m[1mhandles a HO node is switched to a OVN node [It][0m
2024-09-19T01:58:18.6961821Z   [90m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:1218[0m
2024-09-19T01:58:18.6962761Z 
2024-09-19T01:58:18.6963039Z   [91mTimed out after 2.001s.
2024-09-19T01:58:18.6963554Z   Expected
2024-09-19T01:58:18.6964212Z       <[]*nbdb.LogicalRouterStaticRoute | len:2, cap:2>: [
2024-09-19T01:58:18.6964925Z           {
2024-09-19T01:58:18.6965906Z               UUID: "8a84a82f-eaf2-4a41-8ffe-3af1e693fef0",
2024-09-19T01:58:18.6966747Z               BFD: nil,
2024-09-19T01:58:18.6967295Z               ExternalIDs: {
2024-09-19T01:58:18.6968519Z                   "name": "hybrid-subnet-node1:node-windows",
2024-09-19T01:58:18.6969211Z               },
2024-09-19T01:58:18.6969788Z               IPPrefix: "10.1.3.0/24",
2024-09-19T01:58:18.6970484Z               Nexthop: "10.1.1.3",
2024-09-19T01:58:18.6971114Z               Options: nil,
2024-09-19T01:58:18.6971719Z               OutputPort: nil,
2024-09-19T01:58:18.6972307Z               Policy: nil,
2024-09-19T01:58:18.6972901Z               RouteTable: "",
2024-09-19T01:58:18.6973402Z           },
2024-09-19T01:58:18.6973786Z           {
2024-09-19T01:58:18.6974602Z               UUID: "4a94b693-d276-40b6-b645-403474ae8b4d",
2024-09-19T01:58:18.6975312Z               BFD: nil,
2024-09-19T01:58:18.6975871Z               ExternalIDs: {
2024-09-19T01:58:18.6976766Z                   "name": "hybrid-subnet-node1-gr",
2024-09-19T01:58:18.6977423Z               },
2024-09-19T01:58:18.6978155Z               IPPrefix: "10.1.3.0/24",
2024-09-19T01:58:18.6978874Z               Nexthop: "100.64.0.1",
2024-09-19T01:58:18.6979669Z               Options: nil,
2024-09-19T01:58:18.6980281Z               OutputPort: nil,
2024-09-19T01:58:18.6980947Z               Policy: nil,
2024-09-19T01:58:18.6981536Z               RouteTable: "",
2024-09-19T01:58:18.6982058Z           },
2024-09-19T01:58:18.6982435Z       ]
2024-09-19T01:58:18.6982916Z   to have length 0[0m
2024-09-19T01:58:18.6983203Z 
2024-09-19T01:58:18.6984071Z   /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:1387

2024-09-19T01:58:16.8105818Z E0919 01:58:16.724456   35874 obj_retry.go:671] Failed to update *v1.Node, old=node-windows, new=node-windows, error: [macAddress annotation not found for node node-windows; error: could not find "k8s.ovn.org/node-mgmt-port-mac-addresses" annotation, failed to set up hybrid overlay logical switch port for node-windows: cannot set up hybrid overlay ports, distributed router ip is nil, k8s.ovn.org/l3-gateway-config annotation not found for node "node-windows"]
```

https://github.com/ovn-org/ovn-kubernetes/actions/runs/10932941999/job/30350584438?pr=4662

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -46,26 +49,36 @@ type NetworkControllerManager interface {
...
+	GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error)
 }
 
 // NetAttachDefinitionController handles namespaced scoped NAD events and
 // manages cluster scoped networks defined in those NADs. NADs are mostly
 // referred from pods to give them access to the network. Different NADs can
 // define the same network as long as those definitions are actually equal.
 // Unexpected situations are handled on best effort basis but improper NAD
-// adminstration can lead to undefined behavior in referred from running pods.
+// administration can lead to undefined behavior in referred from running pods.
```
jcaamano, line 66: nit: s/in/if

:angel:

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +323,50 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
+			if err != nil {
+				klog.Errorf("Cannot parse NAD reference: %q on network info: %q", nadName, netName)
+				continue
+			}
+			if nadNs == namespace {
+				filteredNADs = append(filteredNADs, nadName)
+			}
+		}
+		n := util.CopyNetInfo(network)
+		n.SetNADs(filteredNADs...)
```
jcaamano, line 356: If we already had the NADs per namespace in NetInfo, which is an information we already come by naturally in sync method, we wouldn't have to do this search every time GetActiveNetworkForNamespace is called

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -280,3 +323,50 @@ func nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
...
+	// no primary network found, make sure we just haven't processed it yet and no UDN exists
+	udns, err := nadController.udnLister.UserDefinedNetworks(namespace).List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("error getting user defined networks: %w", err)
+	}
+	for _, udn := range udns {
+		if util.IsPrimaryNetwork(udn.Spec) {
+			return nil, util.NewUnprocessedActiveNetworkError(namespace, udn.Name)
+		}
+	}
```
jcaamano, line 369: food for thought: we can have a controller listening to UDNs so that we can have a map of primary UDNs to make this also faster. Because, aren't we going to end up doing this search always for the default network?

go-controller/pkg/network-attach-def-controller/network_manager.go
```diff
@@ -32,6 +32,8 @@ type networkManager interface {
 
 	// Stop the controller
 	Stop()
+
+	getNetwork(string) util.NetInfo
```
jcaamano, line 36: nit: I would just do exported `GetNetwork` like the others

go-controller/pkg/network-attach-def-controller/network_manager.go
```diff
@@ -77,42 +79,82 @@ func (nm *networkManagerImpl) Start() error {
 
 func (nm *networkManagerImpl) Stop() {
 	controller.Stop(nm.controller)
-
-	for _, networkControllerState := range nm.networkControllers {
+	for _, networkControllerState := range nm.getAllNetworkStates() {
```
jcaamano, line 82: hmm 

technically we don't need `getAllNetworkStates` because the controller should be fully stopped so no need for thread safety?

 Does no harm but it did confuse me enough to give it a thought

go-controller/pkg/util/util.go
```diff
@@ -387,50 +384,27 @@ func IsUnknownActiveNetworkError(err error) bool {
...
-	}
-	numberOfPrimaryNetworks := 0
-	var primaryNetwork NetInfo
-	for _, nad := range namespaceNADs {
-		netInfo, err := ParseNADInfo(nad)
-		if err != nil {
-			klog.Warningf("Skipping nad '%s/%s' as active network after failing parsing it with %v", nad.Namespace, nad.Name, err)
-			continue
-		}
+func NewUnknownActiveNetworkError(namespace string) *UnknownActiveNetworkError {
```
jcaamano, line 387: Are we using this one?

go-controller/pkg/clustermanager/pod/allocator.go
```diff
@@ -107,9 +107,9 @@ func (a *PodAllocator) Init() error {
 }
 
 // getActiveNetworkForNamespace returns the active network for the given pod's namespace
-// and is a wrapper around util.GetActiveNetworkForNamespace
+// and is a wrapper around GetActiveNetworkForNamespace
 func (a *PodAllocator) getActiveNetworkForPod(pod *corev1.Pod) (util.NetInfo, error) {
-	activeNetwork, err := util.GetActiveNetworkForNamespace(pod.Namespace, a.nadLister)
+	activeNetwork, err := a.nadController.GetActiveNetworkForNamespace(pod.Namespace)
```
jcaamano, line 112: I am going to flag some use cases where, if we had per-namespace NAD information in NetInfo, we probably wouldn't need access to the nad controller to query this information. This is one of them.

We would lose the protection of actually querying UDNs, true.

Just for awareness.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -789,14 +790,9 @@ func (bnc *BaseNetworkController) isLocalZoneNode(node *kapi.Node) bool {
...
-// and is a wrapper around util.GetActiveNetworkForNamespace
-func (cnci *CommonNetworkControllerInfo) getActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
-	var nadLister nadlister.NetworkAttachmentDefinitionLister
-	if util.IsNetworkSegmentationSupportEnabled() {
-		nadLister = cnci.watchFactory.NADInformer().Lister()
-	}
-	return util.GetActiveNetworkForNamespace(namespace, nadLister)
+// and is a wrapper around GetActiveNetworkForNamespace
+func (bnc *BaseNetworkController) getActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	return bnc.nadController.GetActiveNetworkForNamespace(namespace)
```
jcaamano, line 795: I am going to flag some use cases where, if we had per-namespace NAD information in NetInfo, we probably wouldn't need access to the nad controller to query this information. This is one of them.

We would lose the protection of actually querying UDNs, true.

Just for awareness.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller_test.go
```diff
@@ -399,27 +459,37 @@ func TestNetAttachDefinitionController(t *testing.T) {
...
-
-					// test that the controller have the expected config and NADs
-					g.Expect(tncm.controllers).To(gomega.HaveKey(testNetworkKey))
-					g.Expect(tncm.controllers[testNetworkKey].Equals(netInfo)).To(gomega.BeTrue(),
-						fmt.Sprintf("matching network config for network %s", name))
-					g.Expect(tncm.controllers[testNetworkKey].GetNADs()).To(gomega.ConsistOf(expected.nads),
-						fmt.Sprintf("matching NADs for network %s", name))
+					func() {
+						tncm.Lock()
+						defer tncm.Unlock()
```
jcaamano, line 471: Just wondering why you can't hold this lock throughout `meetsExpectations`

jcaamano (commented):
minor comments

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
jcaamano: I am also wondering...should we disallow multiple NADs per namespace on a primary network?

So there could be multiple NADs per namespace for a primary network as long as only one of them is flagged as primary. That would be an interesting use case if, let's say, selected pods want to customize how they attach to the primary network, for example use sr-iov. The one flagged as primary would be used by default. The others would be used through the network selection element annotation.

We currently disallow the network selection element annotation for primary networks so there's actually no use for that. However we can't prevent the mere existence of multiple NADs per namespace for a primary network. So we either have a way to tell around which one of them is really primary or we don't make anyone aware of those NADs since they have no use for us.

jcaamano: To disallow, we would just have to check against the current network instead of the nad network:
```
if currentNetwork != nil && currentNetwork.IsPrimaryNetwork() && primaryNAD != "" && primaryNAD != key
```

Then it his method you would just have to do
```
n.SetNADs(primaryNAD)
```

go-controller/pkg/network-attach-def-controller/network_attach_def_controller_test.go
trozet: it caused deadlock, because normal locking order is:
1. lock NAD controller
2. lock NM
3. tncm lock

When I introduced GetActiveNetworkNamespace to the test, the lock order would be:
1. lock tncm
2. lock NAD controller
3. lock NM

That can cause deadlock

go-controller/pkg/cni/cniserver.go
```diff
@@ -72,6 +75,7 @@ func NewCNIServer(factory factory.NodeWatchFactory, kclient kubernetes.Interface
 	}
 
 	if util.IsNetworkSegmentationSupportEnabled() {
+		s.nadController = nadController
```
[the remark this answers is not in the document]
trozet: i dont think so. I'll see if I can remove it.

go-controller/pkg/factory/factory.go
```diff
@@ -406,6 +394,14 @@ func NewOVNKubeControllerWatchFactory(ovnClientset *util.OVNKubeControllerClient
 		}
 	}
 
+	if util.IsNetworkSegmentationSupportEnabled() {
```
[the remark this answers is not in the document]
trozet: because master watch factory calls 
wf, err := NewOVNKubeControllerWatchFactory(ovnClientset.GetOVNKubeControllerClientset())

and we create it there now

go-controller/pkg/libovsdb/ops/switch.go
```diff
@@ -312,6 +312,9 @@ func createOrUpdateLogicalSwitchPortsOps(nbClient libovsdbclient.Client, ops []l
 	m := newModelClient(nbClient)
 	ops, err := m.CreateOrUpdateOps(ops, opModels...)
 	sw.Ports = originalPorts
+	if err != nil && errors.Is(err, libovsdbclient.ErrNotFound) && !createSwitch {
+		err = fmt.Errorf("could not find switch: %q, %w", sw.Name, err)
+	}
```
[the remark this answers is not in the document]
trozet: I was running into some bugs while running the tests and the addLogicalPort would fail and just say "object not found" i tracked down that it was because of the switch missing, so i made a quality of life improvement to make the error more descriptive

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: but we have a way to really tell which is the active one, since we have the primaryNADs map? Why don't i just use that?

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: agree. that could be a follow up PR. I'll see if its something easy i could add to this PR.

go-controller/pkg/network-attach-def-controller/network_manager.go
trozet: i did this intentionally. exposing it outside of the package would allow another package to get the direct pointer to the value in the cache. As you said yesterday nadController and NM are pals, but we cant trust consumers outside of the package.

go-controller/pkg/network-attach-def-controller/network_manager.go
trozet: yes, just to be safe and consistent

go-controller/pkg/util/util.go
trozet: I guess not anymore after we guarantee the controller only renders a single primary network.

go-controller/pkg/network-attach-def-controller/network_manager.go
jcaamano: The interface itself is not exported so it can't be used outside of the package. In these cases I usually export methods to tell me its being used within the package widely. But fair enough, I see how that's a personal perspective while yours is different and valid.

go-controller/pkg/libovsdb/ops/switch.go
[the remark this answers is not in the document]
trozet: you can see i have trust issues :)

go-controller/pkg/cni/cniserver.go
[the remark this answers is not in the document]
trozet: removed

go-controller/pkg/util/util.go
trozet: removed

go-controller/pkg/clustermanager/pod/allocator.go
trozet: it looks like this was intentional for some reason:
```
func (a *PodAllocator) GetNetworkRole(pod *corev1.Pod) (string, error) {
		// if user defined network segmentation is not enabled
		// then we know pod's primary network is "default" and
		// pod's secondary network is NOT its primary network
		if a.netInfo.IsDefault() {
			return types.NetworkRolePrimary, nil
		}
		return types.NetworkRoleSecondary, nil
	}
	activeNetwork, err := a.getActiveNetworkForPod(pod)
	if err != nil {
		return "", err
	}
	if activeNetwork.GetNetworkName() == a.netInfo.GetNetworkName() {
		return types.NetworkRolePrimary, nil
	}
	if a.netInfo.IsDefault() {
		// if default network was not the primary network,
		// then when UDN is turned on, default network is the
		// infrastructure-locked network forthis pod
		return types.NetworkRoleInfrastructure, nil
	}
	return types.NetworkRoleSecondary, nil
}
```

It is comparing the network name of the network found for the pod and the netinfo held in the allocator. @qinqon maybe you wrote the original code and can explain why it is like this? We are wondering why you cant just use the NetInfo on the allocator itself to know whether or not this is a primary or secondary role?

go-controller/pkg/ovn/base_network_controller.go
trozet: I think this is a result of me changing this function to be a receiver for bnc rather than the common network controller (that doesn't have netInfo). However, it looks like this function is used to determine pod wiring for the default controller, so I think makes sense to have the UDN protection here.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
trozet: lets do this as follow up i dont want to delay this PR any longer

trozet:
rebased

go-controller/pkg/clustermanager/pod/allocator.go
qinqon: @trozet `activeNetwork` is the pod's namespace primary network so this checks if the pod allocator network is the pod's namespace primary network (default or UDN) and terturn Primary according to that.

qinqon:
> > Hey @trozet maybe labeling the NAD when it's created by UserDefinedNetwork controller with role=primary alow for fast listing instead of iterating ?.
> 
> not sure what iterating you are referring to

In case this PR is about improving performance when looking for the namespace's primary UDN, we can label de NAD to improve listing performance.

go-controller/pkg/clustermanager/pod/allocator.go
jcaamano: Makes sense.

However this can be equally achieved if we could query the controller NetInfo for the namespaces served by the network and checking if the one we are concerned about is among them.

Anyway, I will follow up with this on a different PR maybe.

jcaamano (approved):

go-controller/pkg/clustermanager/pod/allocator.go
qinqon: Make sense too, thanks @jcaamano

PR #4669, opened 2024-08-29
UDN: Fix per pod SNATing

tssurya:
```
sh-5.2# ovn-nbctl lr-nat-list GR_l3.network_ovn-worker2
TYPE             GATEWAY_PORT          EXTERNAL_IP        EXTERNAL_PORT    LOGICAL_IP          EXTERNAL_MAC         LOGICAL_PORT
snat                                   169.254.0.11                        10.128.2.5
sh-5.2# ovn-nbctl lr-nat-list GR_l3.network_ovn-worker2
sh-5.2# ovn-nbctl lr-nat-list GR_l3.network_ovn-worker2
sh-5.2#
```
let's see how this looks in CI...

TODO:
- [x] We have e2e coverage for this
- [ ] Need to add UT which is tracked here: https://github.com/ovn-org/ovn-kubernetes/issues/4684

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_secondary.go
+++ b/go-controller/pkg/ovn/base_network_controller_secondary.go
@@ -15,6 +15,7 @@ import (
 	"github.com/ovn-org/libovsdb/ovsdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/generator/udn"
 	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
@@ -338,6 +339,18 @@ func (bsnc *BaseSecondaryNetworkController) addLogicalPortToNetworkForNAD(pod *k
 		ops = append(ops, addOps...)
 	}
 
+	if util.IsNetworkSegmentationSupportEnabled() && bsnc.IsPrimaryNetwork() && config.Gateway.DisableSNATMultipleGWs {
+		// we need to add per-pod SNATs for UDN networks
+		snatOps, err := bsnc.addPerPodSNATOps(pod, podAnnotation.IPs)
+		if err != nil {
+			return fmt.Errorf("failed to construct SNAT for pod %s/%s which is part of network %s, err: %v",
+				pod.Namespace, pod.Name, bsnc.GetNetworkName(), err)
+		}
+		if snatOps != nil {
+			ops = append(ops, snatOps...)
+		}
+	}
+
 	recordOps, txOkCallBack, _, err := bsnc.AddConfigDurationRecord("pod", pod.Namespace, pod.Name)
 	if err != nil {
 		klog.Errorf("Config duration recorder: %v", err)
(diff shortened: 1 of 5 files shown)


go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -338,6 +339,32 @@ func (bsnc *BaseSecondaryNetworkController) addLogicalPortToNetworkForNAD(pod *k
 		ops = append(ops, addOps...)
 	}
 
+	if util.IsNetworkSegmentationSupportEnabled() && bsnc.IsPrimaryNetwork() && config.Gateway.DisableSNATMultipleGWs {
```
maiqueb, line 342: shouldn't this be part of the gateway code ?

tssurya: nope gateway code doesn't have podIPs.. this has to be part of pod code unfortunately.

maiqueb: but can't we move this func there and wrap it in a method ? It would do exactly this, and receive the pod IPs from the controller. 

not a must.

maiqueb (commented):
does this really belong in `go-controller/pkg/ovn/base_network_controller_secondary.go` ? Localnet doesn't need this at all for instance. 

If we could move this to the GW we could invoke it from the required network controllers. 

Food for thought. Maybe it doesn't make sense. Or worth the (I think minimal) trouble.

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya: pod LSP add + snat add should be in same transact...

tssurya:
OK cool so L3/L2 pod2Egress is passing ; need to fix L2 overlay tests next

go-controller/pkg/ovn/base_network_controller_secondary.go
maiqueb: > pod LSP add + snat add should be in same transact.

Have the method return the ops.

I mostly would rather not have all these details shoved in the code's reader face whenever they read the add logical port flow.

tssurya: yea please note this is not the final version see my PR message "need to clean up some things " .. for sure this will a util.. I will see if it belongs in gatewayconfig or not..but this is not the final version of this PR.

tssurya:
there are some issues with l2 pod2pod that I am looking into
```
0xffff89b831f21900      0     [<empty>(0)] kfree_skb_reason(SKB_DROP_REASON_NEIGH_FAILED) 10.100.200.6:0->10.100.200.9:0(icmp)                                               
```

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya: hmm I am trying to move this to gatewayconfig let's take L2 or L3 for example, I think you meant:
```
type SecondaryL2GatewayConfig struct {
	config      *util.L3GatewayConfig
	hostSubnets []*net.IPNet
	gwLRPIPs    []*net.IPNet
	externalIPs []net.IP
}

func (oc *SecondaryLayer2NetworkController) nodeGatewayConfig(node *corev1.Node) (*SecondaryL2GatewayConfig, error) {
	l3GatewayConfig, err := util.ParseNodeL3GatewayAnnotation(node)
```
to add it as a util to this config and also L3 right? let's talk monday in our sync to see if I can figure this out with you

but I have pushed a new version where I have cleaned up the addLSP bits and broken the logic into smaller piece dunno if that is better or not, but happy to hear your thoughts..

tssurya:
> does this really belong in `go-controller/pkg/ovn/base_network_controller_secondary.go` ? Localnet doesn't need this at all for instance.

great catch for localnet.. it was done to avoid code duplication because its needed for both l3 and l2. Let me see if the new clean up I did makes this better for you.

> 
> If we could move this to the GW we could invoke it from the required network controllers.
> 

right so about this, let's talk monday because I am not entirely sure I understand how you intended the controllers to pass each pod's IP into this.. maybe you just mean like a util in gatewayconfig? but I don't see where this config is invoked from.

tssurya:
waiting for CI to run till I push the version that will be ready for review

tssurya:
```
[ReportAfterSuite] PASSED [0.029 seconds]
Ran 27 of 310 Specs in 946.969 seconds
SUCCESS! -- 27 Passed | 0 Failed | 2 Pending | 281 Skipped
```
OK now we should be good

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: the code similar to addLogicalPort code in default controller: https://github.com/ovn-org/ovn-kubernetes/blob/master/go-controller/pkg/ovn/pods.go#L299

It seems OK to me.

trozet (approved):

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya: yeah will talk to @maiqueb on monday to see what he meant.

maiqueb (approved):
This is OK for now; let's have a follow-up to ensure the code is moved away from the base network controller, since localnet has no need of any of this.

PR #4671, opened 2024-08-29
udn, cni: report UDN iface info on network-status

maiqueb:

Before this commit, this was the network status we reported for a UDN pod:
```
kubectl get pods pod -ojsonpath="{@.metadata.annotations.k8s\.v1\.cni\.cncf\.io\/network-status}" | jq
[
  {
    "name": "ovn-kubernetes",
    "interface": "eth0",
    "ips": [
      "10.244.1.9",
      "fd00:10:244:2::9"
    ],
    "mac": "0a:58:0a:f4:01:09",
    "default": true,
    "dns": {}
  }
]
```

With it, we now report:
```
[
  {
    "name": "ovn-kubernetes",
    "interface": "ovn-udn1",
    "ips": [
      "10.128.0.4"
    ],
    "mac": "0a:58:0a:80:00:04",
    "default": true,
    "dns": {}
  },
  {
    "name": "ovn-kubernetes",
    "interface": "eth0",
    "ips": [
      "10.244.0.4"
    ],
    "mac": "0a:58:0a:f4:00:04",
    "dns": {}
  }
]
```

This way, applications complying to the k8snetworkplumbingwg de-facto standard can be aware of the UDN interface information.

Fixes #

#### How to verify it

Run the unit tests. I've added one that asserts this behavior.

Report the primary UDN information in the multi-network de-facto standard `network-status` annotation.

```release-note
Report the primary UDN information in the multi-network de-facto standard `network-status` annotation.
```

### DIFF
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -174,9 +174,24 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(kubeAuth *KubeAPIAuth, clientse
 			if err != nil {
 				return nil, err
 			}
-			if _, err := getCNIResultFn(primaryUDNPodRequest, clientset, primaryUDNPodInfo); err != nil {
+			primaryUDNResult, err := getCNIResultFn(primaryUDNPodRequest, clientset, primaryUDNPodInfo)
+			if err != nil {
 				return nil, err
 			}
+
+			defaulClusterNetworkResult := response.Result
+			originalNumberOfIfaces := len(primaryUDNResult.Interfaces)
+			response.Result.Interfaces = append(primaryUDNResult.Interfaces, response.Result.Interfaces...)
+			for i := range response.Result.IPs {
+				ifaceIPConfig := defaulClusterNetworkResult.IPs[i].Copy()
+
+				if ifaceIPConfig.Interface != nil {
+					ifaceIPConfig.Interface = current.Int(*ifaceIPConfig.Interface + originalNumberOfIfaces)
+				}
+				response.Result.IPs = append(primaryUDNResult.IPs, ifaceIPConfig)
+			}
+
+			response.Result.Routes = append(primaryUDNResult.Routes, response.Result.Routes...)
 		}
 	} else {
 		response.PodIFInfo = podInterfaceInfo
--- a/go-controller/pkg/cni/cni_test.go
+++ b/go-controller/pkg/cni/cni_test.go
@@ -2,6 +2,8 @@ package cni
 
 import (
 	"context"
+	"fmt"
+	"net"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -9,14 +11,17 @@ import (
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	current "github.com/containernetworking/cni/pkg/types/100"
-	v1nadmocks "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing/mocks/github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes/fake"
+
+	nadv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	v1nadmocks "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing/mocks/github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
 	v1mocks "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing/mocks/k8s.io/client-go/listers/core/v1"
 	ovntypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 )
@@ -157,6 +162,148 @@ var _ = Describe("Network Segmentation", func() {
 			})
 		})
 
+		Context("pod with a user defined primary network", func() {
+			const (
+				dummyMACHostSide = "07:06:05:04:03:02"
+				nadName          = "tenantred"
+				namespace        = "foo-ns"
+			)
+
+			dummyGetCNIResult := func(request *PodRequest, getter PodInfoGetter, podInterfaceInfo *PodInterfaceInfo) (*current.Result, error) {
+				var gatewayIP net.IP
+				if len(podInterfaceInfo.Gateways) > 0 {
+					gatewayIP = podInterfaceInfo.Gateways[0]
+				}
+				var ips []*current.IPConfig
+				ifaceIdx := 1 // host side of the veth is 0; pod side of the veth is 1
+				for _, ip := range podInterfaceInfo.IPs {
+					ips = append(ips, &current.IPConfig{Address: *ip, Gateway: gatewayIP, Interface: &ifaceIdx})
+				}
+				ifaceName := "eth0"
+				if request.netName != "default" {
+					ifaceName = "ovn-udn1"
+				}
+
+				interfaces := []*current.Interface{
+					{
+						Name: "host_" + ifaceName,
+						Mac:  dummyMACHostSide,
+					},
+					{
+						Name:    ifaceName,
+						Mac:     podInterfaceInfo.MAC.String(),
+						Sandbox: "bobloblaw",
+					},
+				}
+				return &current.Result{
+					CNIVersion: "0.3.1",
+					Interfaces: interfaces,
+					IPs:        ips,
+				}, nil
+			}
+
+			BeforeEach(func() {
+				pod = &v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      pr.PodName,
+						Namespace: pr.PodNamespace,
+						Annotations: map[string]string{
+							"k8s.ovn.org/pod-networks": `{"default":{"ip_address":"100.10.10.3/24","mac_address":"0a:58:fd:98:00:01", "role":"infrastructure-locked"}, "meganet":{"ip_addresses":["10.10.10.30/24","fd10::3/64"],"mac_address":"02:03:04:05:06:07", "role":"primary"}}`,
+						},
+					},
+				}
+				nadNamespaceLister := &v1nadmocks.NetworkAttachmentDefinitionNamespaceLister{}
+				nadNamespaceLister.On("List", labels.Everything()).Return([]*nadv1.NetworkAttachmentDefinition{{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      nadName,
+						Namespace: namespace,
+					},
+					Spec: nadv1.NetworkAttachmentDefinitionSpec{
+						Config: dummyPrimaryUDNConfig(namespace, nadName),
+					},
+				}}, nil)
+				nadLister.On("NetworkAttachmentDefinitions", "foo-ns").Return(nadNamespaceLister)
+
+				getCNIResultStub = dummyGetCNIResult
+			})
+
+			It("should return the information of both the default net and the primary UDN in the result", func() {
+				podNamespaceLister.On("Get", pr.PodName).Return(pod, nil)
+				response, err := pr.cmdAddWithGetCNIResultFunc(kubeAuth, clientSet, getCNIResultStub)
+				Expect(err).NotTo(HaveOccurred())
+				// for every interface added, we return 2 interfaces; the host side of the
+				// veth, then the pod side of the veth.
+				// thus, the UDN interface idx will be 3:
+				// idx: iface
+				//   0: host side primary UDN
+				//   1: pod side primary UDN
+				//   2: host side default network
+				//   3: pod side default network
+				podDefaultClusterNetIfaceIDX := 3
+				podUDNIfaceIDX := 1
+				Expect(response.Result).To(Equal(
+					&current.Result{
+						CNIVersion: "0.3.1",
+						Interfaces: []*current.Interface{
+							{
+								Name: "host_ovn-udn1",
+								Mac:  dummyMACHostSide,
+							},
+							{
+								Name:    "ovn-udn1",
+								Mac:     "02:03:04:05:06:07",
+								Sandbox: "bobloblaw",
+							}, {
+								Name: "host_eth0",
+								Mac:  dummyMACHostSide,
+							},
+							{
+								Name:    "eth0",
+								Mac:     "0a:58:fd:98:00:01",
+								Sandbox: "bobloblaw",
+							},
+						},
+						IPs: []*current.IPConfig{
+							{
+								Address: net.IPNet{
+									IP:   net.ParseIP("10.10.10.30"),
+									Mask: net.CIDRMask(24, 32),
+								},
+								Interface: &podUDNIfaceIDX,
+							},
+							{
+								Address: net.IPNet{
+									IP:   net.ParseIP("fd10::3"),
+									Mask: net.CIDRMask(64, 128),
+								},
+								Interface: &podUDNIfaceIDX,
+							},
+							{
+								Address: net.IPNet{
+									IP:   net.ParseIP("100.10.10.3"),
+									Mask: net.CIDRMask(24, 32),
+								},
+								Interface: &podDefaultClusterNetIfaceIDX,
+							},
+						},
+					},
+				))
+			})
+		})
 	})
 
 })
+
+func dummyPrimaryUDNConfig(ns, nadName string) string {
+	namespacedName := fmt.Sprintf("%s/%s", ns, nadName)
+	return fmt.Sprintf(`
+    {
+            "name": "tenantred",
+            "type": "ovn-k8s-cni-overlay",
+            "topology": "layer2",
+            "subnets": "10.10.0.0/16,fd10::0/64",
+            "netAttachDefName": %q,
+            "role": "primary"
+    }
+`, namespacedName)
+}
--- a/go-controller/pkg/cni/cni_test.go
+++ b/go-controller/pkg/cni/cni_test.go
@@ -59,6 +59,8 @@ var _ = Describe("Network Segmentation", func() {
 
 	BeforeEach(func() {
 
+		config.IPv4Mode = true
+		config.IPv6Mode = true
 		enableMultiNetwork = config.OVNKubernetesFeature.EnableMultiNetwork
 		enableNetworkSegmentation = config.OVNKubernetesFeature.EnableNetworkSegmentation
 
--- a/go-controller/pkg/cni/cni_test.go
+++ b/go-controller/pkg/cni/cni_test.go
@@ -131,25 +131,30 @@ var _ = Describe("Network Segmentation", func() {
 		BeforeEach(func() {
 			config.OVNKubernetesFeature.EnableMultiNetwork = true
 			config.OVNKubernetesFeature.EnableNetworkSegmentation = true
-			pod = &v1.Pod{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      pr.PodName,
-					Namespace: pr.PodNamespace,
-					Annotations: map[string]string{
-						"k8s.ovn.org/pod-networks": `{"default":{"ip_address":"100.10.10.3/24","mac_address":"0a:58:fd:98:00:01", "role":"primary"}}`,
+		})
+
+		Context("pod with default primary network", func() {
+			BeforeEach(func() {
+				pod = &v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      pr.PodName,
+						Namespace: pr.PodNamespace,
+						Annotations: map[string]string{
+							"k8s.ovn.org/pod-networks": `{"default":{"ip_address":"100.10.10.3/24","mac_address":"0a:58:fd:98:00:01", "role":"primary"}}`,
+						},
 					},
-				},
-			}
-		})
-		It("should not fail at cmdAdd", func() {
-			podNamespaceLister.On("Get", pr.PodName).Return(pod, nil)
-			Expect(pr.cmdAddWithGetCNIResultFunc(kubeAuth, clientSet, getCNIResultStub)).NotTo(BeNil())
-			Expect(obtainedPodIterfaceInfos).ToNot(BeEmpty())
-		})
-		It("should not fail at cmdDel", func() {
-			podNamespaceLister.On("Get", pr.PodName).Return(pod, nil)
-			Expect(pr.cmdDel(clientSet)).NotTo(BeNil())
-			Expect(prInterfaceOpsStub.unconfiguredInterfaces).To(HaveLen(2))
+				}
+			})
+			It("should not fail at cmdAdd", func() {
+				podNamespaceLister.On("Get", pr.PodName).Return(pod, nil)
+				Expect(pr.cmdAddWithGetCNIResultFunc(kubeAuth, clientSet, getCNIResultStub)).NotTo(BeNil())
+				Expect(obtainedPodIterfaceInfos).ToNot(BeEmpty())
+			})
+			It("should not fail at cmdDel", func() {
+				podNamespaceLister.On("Get", pr.PodName).Return(pod, nil)
+				Expect(pr.cmdDel(clientSet)).NotTo(BeNil())
+				Expect(prInterfaceOpsStub.unconfiguredInterfaces).To(HaveLen(2))
+			})
 		})
 
 	})


go-controller/pkg/cni/cni.go
```diff
@@ -174,9 +174,29 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(kubeAuth *KubeAPIAuth, clientse
 			if err != nil {
 				return nil, err
 			}
-			if _, err := getCNIResultFn(primaryUDNPodRequest, clientset, primaryUDNPodInfo); err != nil {
+			primaryNetRes, err := getCNIResultFn(primaryUDNPodRequest, clientset, primaryUDNPodInfo)
```
jcaamano, line 177: nit: `primaryUDNRes` for consistency

go-controller/pkg/cni/cni.go
```diff
@@ -174,9 +174,29 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(kubeAuth *KubeAPIAuth, clientse
...
+			response.Result.Interfaces = append(response.Result.Interfaces, primaryNetRes.Interfaces...)
+			for i := range primaryNetRes.IPs {
+				ifaceIPConfig := primaryNetRes.IPs[i].Copy()
+
+				// we add a single ovn-udn1 primary interface, and must account
+				// for the host veths, which we also return.
+				// The order matters, hence the + 1 below. Check how we return
+				// the interfaces in
+				// https://github.com/ovn-org/ovn-kubernetes/blob/a1e88dcd0b79b6c5eb033daea178708993f8664e/go-controller/pkg/cni/helper_linux.go#L650
+				newIDX := originalNumberOfIfaces + 1
```
jcaamano, line 192: Shouldn't this be offset based? Like:
```
newIDX := ifaceIPConfig.Interface + originalNumberOfIfaces
```

Have you tested when you get multiple IPs for the same interface, for example, dual-stack?

maiqueb: > Have you tested when you get multiple IPs for the same interface, for example, dual-stack?

Yes, I did. Works as expected. If you want, I can update the yamls on the PR description to use a dual-stack network instead.

For this NAD:
```yaml
metadata:
  name: access-tenant-blue
  namespace: default
spec:
  config: |2

    {
            "cniVersion": "0.3.0",
            "name": "tenantblue",
            "topology":"layer2",
            "subnets": "10.128.0.0/16,fd10:1234::0/64",
            "mtu": 1300,
            "netAttachDefName": "default/access-tenant-blue",
            "role": "primary"
    }
```

It returns this annotation:
```sh
kubectl get pods pod -ojsonpath="{@.metadata.annotations.k8s\.v1\.cni\.cncf\.io\/network-status}" | jq
[
  {
    "name": "ovn-kubernetes",
    "interface": "eth0",
    "ips": [
      "10.244.2.3",
      "fd00:10:244:3::3"
    ],
    "mac": "0a:58:0a:f4:02:03",
    "default": true,
    "dns": {}
  },
  {
    "name": "ovn-kubernetes",
    "interface": "ovn-udn1",
    "ips": [
      "10.128.0.4",
      "fd10:1234::4"
    ],
    "mac": "0a:58:0a:80:00:04",
    "default": true,
    "dns": {}
  }
]
```

The pod's IPs:
```sh
kubectl exec pod -- ip a show ovn-udn1
3: ovn-udn1@if12: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1300 qdisc noqueue state UP group default 
    link/ether 0a:58:0a:80:00:04 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 10.128.0.4/16 brd 10.128.255.255 scope global ovn-udn1
    inet6 fd10:1234::4/64 scope global 
    inet6 fe80::858:aff:fe80:4/64 scope link 
```

> Shouldn't this be offset based?

Good point ! It reads a lot better IMHO. Let me just check it works as expected, and I'll re-push.

go-controller/pkg/cni/cni.go
maiqueb: Done.

go-controller/pkg/cni/cni.go
maiqueb: Done.

maiqueb:
closed and re-opened to get CI to run again ...

oshoval:
> closed and re-opened to get CI to run again ...

i think there is "run / run failed" on actions tab

oshoval:
Thanks
I ran it locally and i see eth0 "twice", not as the PR desc says, is it fine ?

```
k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "ovn-kubernetes",
          "interface": "eth0",
          "ips": [
              "10.244.1.8"
          ],
          "mac": "0a:58:0a:f4:01:08",
          "default": true,
          "dns": {}
      },{
          "name": "ovn-kubernetes",
          "interface": "ovn-udn1",
          "ips": [
              "10.128.0.4"
          ],
          "mac": "0a:58:0a:80:00:04",
          "default": true,
          "dns": {}
      },{
          "name": "default/primary-udn-kubevirt-binding",
          "interface": "eth0",
          "mac": "0a:58:0a:f4:01:08",
          "dns": {}
      }]
```

qinqon:
> Thanks I ran it locally and i see eth0 "twice" as interface, not as the PR desc says, is it fine ? I guess it is related to the NAD we are adding? (running with the [PR](https://github.com/ovn-org/ovn-kubernetes/pull/4655) that adds this NAD as well)
> 
> ```
> $ oc get net-attach-def -A
> NAMESPACE   NAME                           AGE
> blue-ns     access-tenant-blue             40m
> default     primary-udn-kubevirt-binding   43m
> ```
> 
> ```
> k8s.v1.cni.cncf.io/network-status: |-
>       [{
>           "name": "ovn-kubernetes",
>           "interface": "eth0",
>           "ips": [
>               "10.244.1.8"
>           ],
>           "mac": "0a:58:0a:f4:01:08",
>           "default": true,
>           "dns": {}
>       },{
>           "name": "ovn-kubernetes",
>           "interface": "ovn-udn1",
>           "ips": [
>               "10.128.0.4"
>           ],
>           "mac": "0a:58:0a:80:00:04",
>           "default": true,
>           "dns": {}
>       },{
>           "name": "default/primary-udn-kubevirt-binding",
>           "interface": "eth0",
>           "mac": "0a:58:0a:f4:01:08",
>           "dns": {}
>       }]
> ```

Yes, this is expected, kubevirt add a NSE to virt-launcher to do the passt network  namespace things it needs.

maiqueb:
> @maiqueb having a pair of entries with `"default": true` could be an issue ?

Not sure. That's what the library is doing: https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/master/pkg/utils/net-attach-def.go#L128

I can take this with the plumbing group today, since I'm already meeting them.

go-controller/pkg/cni/cni.go
```diff
@@ -174,9 +174,26 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(kubeAuth *KubeAPIAuth, clientse
...
 			}
+
+			originalNumberOfIfaces := len(response.Result.Interfaces)
+			response.Result.Interfaces = append(response.Result.Interfaces, primaryUDNResult.Interfaces...)
+			for i := range primaryUDNResult.IPs {
+				ifaceIPConfig := primaryUDNResult.IPs[i].Copy()
+
+				if ifaceIPConfig.Interface != nil {
+					newIDX := *ifaceIPConfig.Interface + originalNumberOfIfaces
+					ifaceIPConfig.Interface = &newIDX
```
jcaamano, line 189: heads up: you can use 
```
ifaceIPConfig.Interface = current.Int(*ifaceIPConfig.Interface + originalNumberOfIfaces)
```

go-controller/pkg/cni/cni.go
```diff
@@ -174,9 +174,26 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(kubeAuth *KubeAPIAuth, clientse
...
+				}
+				response.Result.IPs = append(response.Result.IPs, ifaceIPConfig)
+			}
+
+			if len(primaryUDNResult.Routes) >= 0 {
```
jcaamano, line 194: I don't think the len check is actually needed

go-controller/pkg/cni/cni_test.go
```diff
@@ -131,27 +136,168 @@ var _ = Describe("Network Segmentation", func() {
...
+					Spec: nadv1.NetworkAttachmentDefinitionSpec{
+						Config: dummyPrimaryUDNConfig(namespace, nadName),
+					},
+				}}, nil)
+				nadLister.On("NetworkAttachmentDefinitions", "foo-ns").Return(nadNamespaceLister)
+
+				getCNIResultStub = dummyGetCNIResult
+			})
+
+			It("should return the information of both the default net and the primary UDN in the result", func() {
```
jcaamano, line 230: could we have a dual stack test?

maiqueb: If you prefer, I'll just turn this single test into dual stack. How about that ?

jcaamano:
> ```
> [
>   {
>     "name": "ovn-kubernetes",
>     "interface": "eth0",
>     "ips": [
>       "10.244.1.6",
>       "fd00:10:244:2::6"
>     ],
>     "mac": "0a:58:0a:f4:01:06",
>     "default": true,
>     "dns": {}
>   },
>   {
>     "name": "ovn-kubernetes",
>     "interface": "ovn-udn1",
>     "ips": [
>       "10.128.0.3"
>     ],
>     "mac": "0a:58:0a:80:00:03",
>     "default": true,
>     "dns": {}
>   }
> ]
> ```

I am curious, what does `"name": "ovn-kubernetes"` represent here? The plugin type? The network? If the network, does it make sense that it is the same value for both? Could that be problematic?

jcaamano:
> Yes, this is expected, kubevirt add a NSE to virt-launcher to do the passt network namespace things it needs.

What is `NSE`? How does
```
{
          "name": "default/primary-udn-kubevirt-binding",
          "interface": "eth0",
          "mac": "0a:58:0a:f4:01:08",
          "dns": {}
}
```
end up there?

maiqueb:
> > Yes, this is expected, kubevirt add a NSE to virt-launcher to do the passt network namespace things it needs.
> 
> What is `NSE`? How does
> 
> ```
> {
>           "name": "default/primary-udn-kubevirt-binding",
>           "interface": "eth0",
>           "mac": "0a:58:0a:f4:01:08",
>           "dns": {}
> }
> ```
> 
> end up there?

It doesn't. This is @oshoval messing around with secondary networks. NSE is the de-facto standard network selection element.

go-controller/pkg/cni/cni_test.go
jcaamano: I am fine with that.

maiqueb: Done.

go-controller/pkg/cni/cni.go
maiqueb: Done.

go-controller/pkg/cni/cni.go
maiqueb: I don't follow this code; won't we find a nil pointer if `ifaceIPConfig.Interface` is nil ?

That's what I'm trying to avoid. I am probably missing something.

maiqueb:
> > ```
> > [
> >   {
> >     "name": "ovn-kubernetes",
> >     "interface": "eth0",
> >     "ips": [
> >       "10.244.1.6",
> >       "fd00:10:244:2::6"
> >     ],
> >     "mac": "0a:58:0a:f4:01:06",
> >     "default": true,
> >     "dns": {}
> >   },
> >   {
> >     "name": "ovn-kubernetes",
> >     "interface": "ovn-udn1",
> >     "ips": [
> >       "10.128.0.3"
> >     ],
> >     "mac": "0a:58:0a:80:00:03",
> >     "default": true,
> >     "dns": {}
> >   }
> > ]
> > ```
> 
> I am curious, what does `"name": "ovn-kubernetes"` represent here? The plugin type? The network? If the network, does it make sense that it is the same value for both? Could that be problematic?

It's the network name, as can be seen in https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/ec02fb320d0b50eb4e24164bb8d6593402717118/pkg/utils/net-attach-def.go#L128

Since this is invoked by multus (as a result of a single CNI ADD) this has the same name.
@jcaamano

jcaamano:
> It's the network name, as can be seen in https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/ec02fb320d0b50eb4e24164bb8d6593402717118/pkg/utils/net-attach-def.go#L128

This is what the spec says about it:
```
5.3.1 "name"
This required key’s value (type string) shall contain either a NetworkAttachmentDefinition object name from the pod’s Network Attachment Selection Annotation, or the name of the cluster-wide default network. The "name" should contain a namespace reference as defined in Section 4.1.1. [ref 2018-02-01 meeting @22:30]
```

Strictly speaking we are in conformance and the fact that we are using a NAD internally becomes an implementation detail.

But just from that same perspective, that this is an implementation detail, why would we want to include it in the status? What's the use case? How is it going to be used?  I am not sure applications complying to the k8snetworkplumbingwg de-facto standard will know how to interpret the presence of two interfaces for the default network (other than ignore them). It can be useful for OVN-K aware applications, but I guess we need to doc it somewhere then.

Not against, but this can be less useful than we think it is and can in turn break applications that (incorrectly) expect a single interface on the default network.

maiqueb:
> > It's the network name, as can be seen in https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/ec02fb320d0b50eb4e24164bb8d6593402717118/pkg/utils/net-attach-def.go#L128
> 
> This is what the spec says about it:
> 
> ```
> 5.3.1 "name"
> This required key’s value (type string) shall contain either a NetworkAttachmentDefinition object name from the pod’s Network Attachment Selection Annotation, or the name of the cluster-wide default network. The "name" should contain a namespace reference as defined in Section 4.1.1. [ref 2018-02-01 meeting @22:30]
> ```
> 
> Strictly speaking we are in conformance and the fact that we are using a NAD internally becomes an implementation detail.
> 
> But just from that same perspective, that this is an implementation detail, why would we want to include it in the status? What's the use case? How is it going to be used? I am not sure applications complying to the k8snetworkplumbingwg de-facto standard will know how to interpret the presence of two interfaces for the default network (other than ignore them). It can be useful for OVN-K aware applications, but I guess we need to doc it somewhere then.

We just discussed this is the plumbingwg; we will report a single interface w/ the `default` attribute enabled. 

This will be used - for instance - by KubeVirt to know from which interface to report the IP address. So, for some applications, this is very helpful - without, we wouldn't know where to find the required info.

> 
> Not against, but this can be less useful than we think it is and can in turn break applications that (incorrectly) expect a single interface on the default network.

The applications that break because they expect to find a single interface reported for the default network can use this attribute to know which of the interfaces to choose.

maiqueb:
@jcaamano opened [this issue](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/issues/69) upstream to track the multiple `default: true` interfaces.

maiqueb:
> > We just discussed this is the plumbingwg; we will report a single interface w/ the `default` attribute enabled; this is a bug that needs to be addressed.
> 
> So how would multus (?) know which of the two interfaces should be set as default?

Multus doesn't care. It plumbs the network, and that's it. 

> 
> > This will be used - for instance - by KubeVirt to know from which interface to report the IP address. So, for some applications, this is very helpful - without, we wouldn't know where to find the required info.
> > ...
> > The applications that break because they expect to find a single interface reported for the default network can use this attribute to know which of the interfaces to choose.
> 
> Having the `default` singleton issue resolved, I guess the only other weird thing is reporting one interface being on the `default` network when it is actually not; or seen from other perspective, changing what `default` actually means? One application my interpret default as "this is the network used by kubelet, or to reach the api server, or to reach DNS" which does not match the interpretation that kubevirt will make of `default` if it is to be set to the UDN interface.
> 
> Dunno, maybe this is not an issue from the plumbingwg spec perspective and probably I am not the most indicated person to say it is. I just wanna make sure there is awareness.

There is awareness, and we've thoroughly discussed this on the plumbing meeting yesterday. The end decision was this would be up to the implementation to figure out, as you can read in the [minutes](https://docs.google.com/document/d/1oE93V3SgOGWJ4O1zeD1UmpeToa0ZiiO6LqRAmZBPFWM/edit):
> Should be implementation specific

Tagging @dougbtv for confirmation.

jcaamano:
> Multus doesn't care. It plumbs the network, and that's it.
> ...
> The end decision was this would be up to the implementation to figure out, as you can read in the [minutes]

I guess I am referring to this proposal on the spec
```
An implementation may attempt to determine the default by using the routes element in CNI looking for a dst of 0.0.0.0/0 and check the gateway against the IPs to see if the subnet matches, and if found, mark that interface as the default:true, otherwise, it may use the first interface in the "interfaces" list from the CNI result. (In CNI 1.y..z with an interface index in the routes…)
```
from
https://docs.google.com/document/d/1DUTV-o6E6zlRTKZkxeDhAeyGrmq03qPgPbU580Rm7-g/edit

That means we would need to take care of the order in this PR (because we are not returning routes)?

> > Having the `default` singleton issue resolved, I guess the only other weird thing is reporting one interface being on the `default` network when it is actually not; or seen from other perspective, changing what `default` actually means? One application my interpret default as "this is the network used by kubelet, or to reach the api server, or to reach DNS" which does not match the interpretation that kubevirt will make of `default` if it is to be set to the UDN interface.

So this is not a concern?

maiqueb:
> > Multus doesn't care. It plumbs the network, and that's it.
> > ...
> > The end decision was this would be up to the implementation to figure out, as you can read in the [minutes]
> 
> I guess I am referring to this proposal on the spec
> 
> ```
> An implementation may attempt to determine the default by using the routes element in CNI looking for a dst of 0.0.0.0/0 and check the gateway against the IPs to see if the subnet matches, and if found, mark that interface as the default:true, otherwise, it may use the first interface in the "interfaces" list from the CNI result. (In CNI 1.y..z with an interface index in the routes…)
> ```
> 
> from https://docs.google.com/document/d/1DUTV-o6E6zlRTKZkxeDhAeyGrmq03qPgPbU580Rm7-g/edit
> 
> That means we would need to take care of the order in this PR (because we are not returning routes)?

IMHO the correct thing to do would be to return the routes. At least the default ones for the default VRF.

> 
> > > Having the `default` singleton issue resolved, I guess the only other weird thing is reporting one interface being on the `default` network when it is actually not; or seen from other perspective, changing what `default` actually means? One application my interpret default as "this is the network used by kubelet, or to reach the api server, or to reach DNS" which does not match the interpretation that kubevirt will make of `default` if it is to be set to the UDN interface.
> 
> So this is not a concern?

I don't see why the consuming application would care about the name of the network. Not saying there isn't; just saying I don't see one.

The only thing KubeVirt cares about is which to which IPs will services latch on to or, at most, where is the default route located.

jcaamano:
> IMHO the correct thing to do would be to return the routes. At least the default ones for the default VRF.
> 
> The only thing KubeVirt cares about is which to which IPs will services latch on to or, at most, where is the default route located.

The default route is not necessarily on the default network. It could be on a secondary network.

maiqueb:
> > IMHO the correct thing to do would be to return the routes. At least the default ones for the default VRF.
> > The only thing KubeVirt cares about is which to which IPs will services latch on to or, at most, where is the default route located.
> 
> The default route is not necessarily on the default network. It could be on a secondary network.

Yes, that will surely **not** be output on the result of the CNI ADD issued for the default network.

jcaamano:
> Yes, but that will surely **not** be output on the result of the CNI ADD issued for the default network.

So then the order becomes relevant?

maiqueb:
> > Yes, but that will surely **not** be output on the result of the CNI ADD issued for the default network.
> 
> So then the order becomes relevant?

No. Successive CNI calls (for the secondary networks will be invoked with defaultNetwork = false, meaning they **won't** have `default` enabled.
- create network statuses: https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/ec02fb320d0b50eb4e24164bb8d6593402717118/pkg/utils/net-attach-def.go#L128
- multus invocation  https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/multus/multus.go#L744
- multus compute deletegate default network CNI ADD (for which UDN returns 2 interfaces): https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/k8sclient/k8sclient.go#L619
- multus compute deletegate secondary network: https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/k8sclient/k8sclient.go#L452

The `masterPlugin` flag will only be enabled for the CNI ADD of the default cluster network. 

The idea if to only check the default route thing for the CNI ADD of the default cluster network (where it matters).

Plus, you can't have UDN + a secondary network with the route on. That's not supported.

jcaamano:
> No. Successive CNI calls (for the secondary networks will be invoked with defaultNetwork = false, meaning they **won't** have `default` enabled.

I meant all this time the order in which we return the interfaces in the result of our single CNI Add call, which is what I assume the excerpt of the proposed change of the spec that I pasted above is referring to.

maiqueb:
> > No. Successive CNI calls (for the secondary networks will be invoked with defaultNetwork = false, meaning they **won't** have `default` enabled.
> 
> I meant all this time the order in which we return the interfaces in the result of our single CNI Add call, which is what I assume the excerpt of the proposed change of the spec that I pasted above is referring to.

It is, but isn't returning the routes and allowing it to figure out what's the default preferable ? 

One of those will **always** have the default route.

If a secondary network has the default route, multus will [remove it](https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/multus/multus.go#L709) from the interface with the default route, and [set it in the secondary](https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/multus/multus.go#L721). That's how it works today (and I don't think it can change).

jcaamano:
> One of those will **always** have the default route.

I am not sure about that:
https://github.com/ovn-org/ovn-kubernetes/blob/a551930f51beee9f4619e91a400c9d9af339ead9/go-controller/pkg/util/pod_annotation.go#L542

maiqueb:
> > > No. Successive CNI calls (for the secondary networks will be invoked with defaultNetwork = false, meaning they **won't** have `default` enabled.
> > 
> > 
> > I meant all this time the order in which we return the interfaces in the result of our single CNI Add call, which is what I assume the excerpt of the proposed change of the spec that I pasted above is referring to.
> 
> It is, but isn't returning the routes and allowing it to figure out what's the default preferable ?
> 
> One of those will **always** have the default route.
> 
> If a secondary network has the default route, multus will [remove it](https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/multus/multus.go#L709) from the interface with the default route, and [set it in the secondary](https://github.com/k8snetworkplumbingwg/multus-cni/blob/31e77aafab63a808766e1a22f0cd4dbcf9fd8f22/pkg/multus/multus.go#L721). That's how it works today (and I don't think it can change).

Having said that, I am OK with using the ordinal approach, and reverse the order of the array.

maiqueb:
> > One of those will **always** have the default route.
> 
> I am not sure about that:
> 
> https://github.com/ovn-org/ovn-kubernetes/blob/a551930f51beee9f4619e91a400c9d9af339ead9/go-controller/pkg/util/pod_annotation.go#L542

Ah, so we don't rely on multus performing this change then ? That's a surprise. 

OK, that leaves the ordinal approach. 

Would that be OK for you then ?

maiqueb:
We'll also need https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/pull/71 ... My bad.

With it, provisioning the following manifest will produce a pod we'll get the following network-status:
```sh
cat ~/ovn-k-manifests/udn/01-net-config-egress-l2.yaml
---
metadata:
  name: access-tenant-blue
  namespace: default
spec:
  config: |2

    {
            "cniVersion": "0.3.0",
            "name": "tenantblue",
            "topology":"layer2",
            "subnets": "10.128.0.0/16",
            "mtu": 1300,
            "netAttachDefName": "default/access-tenant-blue",
            "role": "primary"
    }
---
apiVersion: v1
kind: Pod
metadata:
  name: pod
  namespace: default
spec:
  containers:
  - args:
    - pause
    image: registry.k8s.io/e2e-test-images/agnhost:2.45
    imagePullPolicy: IfNotPresent
    name: agnhost-container
  nodeName: ovn-worker
  restartPolicy: Always
  securityContext: {}
  serviceAccount: default
  serviceAccountName: default

kubectl get pods pod -ojsonpath="{@.metadata.annotations.k8s\.v1\.cni\.cncf\.io\/network-status}" | jq
[
  {
    "name": "ovn-kubernetes",
    "interface": "ovn-udn1",
    "ips": [
      "10.128.0.4"
    ],
    "mac": "0a:58:0a:80:00:04",
    "default": true,
    "dns": {}
  },
  {
    "name": "ovn-kubernetes",
    "interface": "eth0",
    "ips": [
      "10.244.1.9"
    ],
    "mac": "0a:58:0a:f4:01:09",
    "dns": {}
  }
]
```

jcaamano:
Closing in favor of #4725

PR #4672, opened 2024-08-29
Fixes remote node checks to be network aware

trozet:
Everytime a UDN was created, we were adding the all remote nodes for every network all over again, including the default network. This makes the checks on the annotations network aware.


trozet:
@dceara FYI

dceara (approved):
Makes sense to me.

We should probably have a look at `func (h *zoneClusterControllerEventHandler) AreResourcesEqual(obj1, obj2 interface{}) (bool, error) {` too (as you mentioned in our offline chat) but that can be a separate PR.

The CI failures seem to be known flakes to me.

PR #4674, opened 2024-08-30
Add static route to the hairpin masquerade IPs to pod

pliurh:
When users attach pod to a secondary network and override the default route pod. It will cause the assymetric routing for service haripin traffic.

We add static routes to ensure the traffic to the hairpin masquerade IP always goes to OVN.

https://issues.redhat.com/browse/OCPBUGS-39009

#### How to verify it
With this patch, in the pod routing table a static route to hairpin masquerade IPs will be added.

```
# IPv4
root@ovn-control-plane:/# nsenter --net=/proc/3607/ns/net ip route
default via 10.244.2.1 dev eth0 
10.96.0.0/16 via 10.244.2.1 dev eth0 
10.244.0.0/16 via 10.244.2.1 dev eth0 
10.244.2.0/24 dev eth0 proto kernel scope link src 10.244.2.5 
100.64.0.0/16 via 10.244.2.1 dev eth0 
169.254.0.5 via 10.244.2.1 dev eth0 

#IPv6
root@ovn-control-plane:/# nsenter --net=/proc/3607/ns/net ip -6 route
fd00:10:96::/112 via fd00:10:244:3::1 dev eth0 metric 1024 pref medium
fd00:10:244:3::/64 dev eth0 proto kernel metric 256 pref medium
fd00:10:244::/48 via fd00:10:244:3::1 dev eth0 metric 1024 pref medium
fd69::5 via fd00:10:244:3::1 dev eth0 metric 1024 pref medium
fd98::/64 via fd00:10:244:3::1 dev eth0 metric 1024 pref medium
fe80::/64 dev eth0 proto kernel metric 256 pref medium
default via fd00:10:244:3::1 dev eth0 metric 1024 pref medium
```

No

```release-note

```

### DIFF
--- a/go-controller/pkg/util/pod_annotation.go
+++ b/go-controller/pkg/util/pod_annotation.go
@@ -467,6 +467,20 @@ func serviceCIDRToRoute(isIPv6 bool, gatewayIP net.IP) []PodRoute {
 	return podRoutes
 }
 
+func hairpinMasqueradeIPToRoute(isIPv6 bool, gatewayIP net.IP) PodRoute {
+	ip := config.Gateway.MasqueradeIPs.V4OVNServiceHairpinMasqueradeIP
+	if isIPv6 {
+		ip = config.Gateway.MasqueradeIPs.V6OVNServiceHairpinMasqueradeIP
+	}
+	return PodRoute{
+		Dest: &net.IPNet{
+			IP:   ip,
+			Mask: GetIPFullMask(ip),
+		},
+		NextHop: gatewayIP,
+	}
+}
+
 // addRoutesGatewayIP updates the provided pod annotation for the provided pod
 // with the gateways derived from the allocated IPs
 func AddRoutesGatewayIP(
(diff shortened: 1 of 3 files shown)


tssurya:
I've opened an issue to track this for UDN: https://github.com/ovn-org/ovn-kubernetes/issues/4665 which can be done as a FUP

go-controller/pkg/util/pod_annotation.go
```diff
@@ -467,6 +467,20 @@ func serviceCIDRToRoute(isIPv6 bool, gatewayIP net.IP) []PodRoute {
 	return podRoutes
 }
 
+func hairpinMasqueradeIPToRoute(isIPv6 bool, gatewayIP net.IP) PodRoute {
+	ip := config.Gateway.MasqueradeIPs.V4OVNServiceHairpinMasqueradeIP.To4()
```
jcaamano, line 471: Could this be normalized on the test side using `To16()` there?

tssurya:
So for UDPNs: the expectation is that pod's default gateway/route should be towards the UDPN.
When users do this it will break UDPN. Maybe this overriding default pod route via multus when using UDNPNs shouldn't be supported?
makes less sense to create a primary network only to route all your pod traffic via secondary network..

jcaamano:
> So for UDPNs: the expectation is that pod's default gateway/route should be towards the UDPN. When users do this it will break UDPN. Maybe this overriding default pod route via multus when using UDNPNs shouldn't be supported? makes less sense to create a primary network only to route all your pod traffic via secondary network..but I might be missing something

I guess it makes the same sense as doing it for the default network, doesn't it?
Anyway, if there are no service split between default & primary network, then this PR as is would do it?

jcaamano (approved):

PR #4677, opened 2024-08-30
Emit Event if NAD cannot be parsed

martinkennelly:
Emit a K8 event within the namespace of the NAD if the NAD cannot be parsed.


martinkennelly:
Only do this for CM to avoid redundant repeating of events hence why recorder is nil for some instances.

martinkennelly:
Ive come to the conclusion that I cannot emit an event per pod that is both selected by EIP and also is attached to primary UDN network. I think docs + this is sufficient.

trozet (approved):

PR #4680, opened 2024-08-30
Do not update EncapIP if it is configured

cathy-zhou:
If EncapIP is configurated, it means it is different from the node's primary address. Do not update EncapIP when node's primary address changes.

Fixes #

```release-note

```


cathy-zhou:
@girishmg can you please take a look. it is required by dpu setup.

girishmg (approved):

PR #4653, opened 2024-08-27
UDN: Basic layer2 service support

kyrtapz:
Adds support for Layer2 services for cluster networked endpoints. 
Based on https://github.com/ovn-org/ovn-kubernetes/pull/4567.

PR #4690, opened 2024-09-02
Support Network Policy for UDN namespaces

pperiyasamy:

This PR has required changes to support Network Policy on User Defined Network.

- [X] NetPol Support on L3 UDN network.
- [X] NetPol support on on L2 UDN network.

TODOs:

- [ ] Run NetPol upstream conformance tests over UDN network.
- [ ] Add hairpin allow ACLs for primary UDNs network policies (https://issues.redhat.com/browse/SDN-5290).

Fixes #

```release-note

```


npinaeva:
there are some netpol-related test failures. Also, could you please add a couple of e2es for the new netpol behaviour?

pperiyasamy:
hit this ci flake https://github.com/ovn-org/ovn-kubernetes/issues/4299 in this run:
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10716745664/job/29716177061?pr=4690

go-controller/pkg/util/pod_annotation.go
```diff
@@ -301,12 +302,30 @@ func GetPodCIDRsWithFullMask(pod *v1.Pod, nInfo NetInfo) ([]*net.IPNet, error) {
...
+			} else if err != nil {
+				return nil, err
+			}
+			ips := make([]net.IP, 0, len(userNet.IPs))
+			for _, netIP := range userNet.IPs {
+				ips = append(ips, netIP.IP)
+			}
+			return ips, nil
+		}
+	}
```
jcaamano, line 328: Please note that `nInfo.IsSecondary` return `true` for primary networks as well.

However, how is this implementation any different functionally than `SecondaryNetworkPodIPs`?

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -696,12 +696,26 @@ func (bnc *BaseNetworkController) doesNetworkRequireIPAM() bool {
...
 }
 
-func (bnc *BaseNetworkController) getPodNADNames(pod *kapi.Pod) []string {
+func (bnc *BaseNetworkController) getPodNADNames(pod *kapi.Pod) ([]string, error) {
 	if !bnc.IsSecondary() {
-		return []string{types.DefaultNetworkName}
+		return []string{types.DefaultNetworkName}, nil
+	}
+	if bnc.IsPrimaryNetwork() {
+		namespacePrimaryNetwork, err := bnc.getActiveNetworkForNamespace(pod.Namespace)
```
jcaamano, line 704: Why do we need to do this?

Isn't this network controller the one for the active network for the namespace?

In which cases do we expect `namespacePrimaryNetwork` to be different than `bnc.NetInfo`?

jcaamano: Comments always help though

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -200,6 +200,81 @@ func NewNetworkPolicy(policy *knet.NetworkPolicy) *networkPolicy {
...
+	}
+
+	if bnc.NetInfo.IsSecondary() {
+		return nil
+	}
+
+	// add default hairpin allow acl
+	err = bnc.addHairpinAllowACL()
+	if err != nil {
+		return fmt.Errorf("failed to create allow hairping acl: %w", err)
```
jcaamano, line 230: nit: `hairpin`

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -200,6 +200,81 @@ func NewNetworkPolicy(policy *knet.NetworkPolicy) *networkPolicy {
...
+	if err != nil {
+		return err
+	}
+
+	if bnc.NetInfo.IsSecondary() {
+		return nil
+	}
+
+	// add default hairpin allow acl
+	err = bnc.addHairpinAllowACL()
```
jcaamano, line 228: Why don't we need the hairpin allow ACLs for primary networks? Add comment.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -68,6 +69,18 @@ func (bsnc *BaseSecondaryNetworkController) AddSecondaryNetworkResourceCommon(ob
...
+		if !ok {
+			return fmt.Errorf("could not cast %T object to *knet.NetworkPolicy", obj)
+		}
+
+		if err := bsnc.addNetworkPolicy(np); err != nil {
+			klog.Infof("Network Policy add failed for %s/%s, will try again later: %v",
+				np.Namespace, np.Name, err)
+			return err
+		}
+
```
jcaamano, line 83: I guess these could live in the base event handler if they are going to be handled equally for all network types.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -727,6 +767,27 @@ func (bsnc *BaseSecondaryNetworkController) WatchMultiNetworkPolicy() error {
 		return nil
 	}
 
+	if util.IsNetworkSegmentationSupportEnabled() && bsnc.IsPrimaryNetwork() {
```
jcaamano, line 770: I guess througout the PR, why do we check `IsNetworkSegmentationSupportEnabled` along with `IsPrimaryNetwork`?

When don't do the equivalent with secondary networks checking `MultiNetworkEnabled` along with `IsSecondary()`

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -727,6 +767,27 @@ func (bsnc *BaseSecondaryNetworkController) WatchMultiNetworkPolicy() error {
...
+		return nil
+	}
+
+	if bsnc.policyHandler != nil {
+		return nil
+	}
+	handler, err := bsnc.retryNetworkPolicies.WatchResource()
```
jcaamano, line 777: Don't we need a different handler for NetPol vs MultiNetPol?

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -727,6 +767,27 @@ func (bsnc *BaseSecondaryNetworkController) WatchMultiNetworkPolicy() error {
...
+	if err != nil {
+		bsnc.policyHandler = handler
+	}
+	return err
+}
+
+// WatchNetworkPolicy starts the watching of networkpolicy resource and calls
+// back the appropriate handler logic
+func (bsnc *BaseSecondaryNetworkController) WatchNetworkPolicy() error {
```
jcaamano, line 786: I guess this can go to the base_network_controller

go-controller/pkg/ovn/base_network_controller_secondary.go
jcaamano: Also, multinetpol should only apply to non primary networks?

go-controller/pkg/ovn/base_secondary_layer2_network_controller.go
```diff
@@ -34,7 +34,9 @@ func (oc *BaseSecondaryLayer2NetworkController) stop() {
 	if oc.ipamClaimsHandler != nil {
 		oc.watchFactory.RemoveIPAMClaimsHandler(oc.ipamClaimsHandler)
 	}
-	if oc.policyHandler != nil {
+	if oc.policyHandler != nil && util.IsNetworkSegmentationSupportEnabled() && oc.IsPrimaryNetwork() {
```
jcaamano, line 37: nit: just checking `oc.policyHandler != nil` should be enough. As already mentioned, you need a different handler for multinetpols

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -62,6 +63,10 @@ func (h *secondaryLayer2NetworkControllerEventHandler) GetResourceFromInformerCa
 // RecordAddEvent records the add event on this given object.
 func (h *secondaryLayer2NetworkControllerEventHandler) RecordAddEvent(obj interface{}) {
 	switch h.objType {
+	case factory.PolicyType:
```
jcaamano, line 66: all this could go into base_event_handler

jcaamano: Also, a bit below, why for RecordUpdateEvent we are calling `h.baseHandler.recordAddEvent` instead of `h.baseHandler.recordUpdateEvent`? 

PTAL @tssurya for awareness

go-controller/pkg/ovn/base_network_controller_secondary.go
jcaamano: I see that you initialize them different. However, I would still use separate fields for clarity.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
tssurya: yeah sounds like a copy paste snafu nice catch, should be add, update, delete respectively. I'll take care of a cleanup PR for that

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: added comment to explain this.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: added comment about why hairpin allow ACLs disabled now for UDN. will revisit once we have cluster port group exists per network. currently this is not impacting unit tests and e2e's.

go-controller/pkg/ovn/base_network_controller_secondary.go
pperiyasamy: all the secondary networks event handling done here, similarly for default network, it's being handled in default network controller.  do you mean to say PolicyType handling can be moved into BaseNetworkController as policy event handling is common for both ?

go-controller/pkg/ovn/base_network_controller_secondary.go
pperiyasamy: yes, done.

go-controller/pkg/util/pod_annotation.go
pperiyasamy: > Please note that `nInfo.IsSecondary` return `true` for primary networks as well.

yes @jcaamano , thanks for the catch, it was actually causing failures for upstream netpol tests. fixed it now.

> 
> However, how is this implementation any different functionally than `SecondaryNetworkPodIPs`?

The retrieval of IPs from `SecondaryNetworkPodIPs` is retrieving nad name based on `k8s.v1.cni.cncf.io/networks` annotation which doesn't work for primary networks. hence separate handling is needed for primary networks.

go-controller/pkg/ovn/base_network_controller_namespace.go
```diff
@@ -73,8 +73,10 @@ func getNamespaceAddrSetDbIDs(namespaceName, controller string) *libovsdbops.DbO
 func (bnc *BaseNetworkController) WatchNamespaces() error {
 	if bnc.IsSecondary() {
 		// For secondary networks, we don't have to watch namespace events if
-		// multi-network policy support is not enabled.
-		if !util.IsMultiNetworkPoliciesSupportEnabled() {
+		// multi-network policy support is not enabled and if it's not used a
+		// primary network for the namespaced.
+		if !util.IsMultiNetworkPoliciesSupportEnabled() &&
+			(!util.IsNetworkSegmentationSupportEnabled() || !bnc.IsPrimaryNetwork()) {
```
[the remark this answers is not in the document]
pperiyasamy: with two handlers, this check is not needed anymore.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: done.

go-controller/pkg/ovn/base_network_controller_secondary.go
pperiyasamy: yes, i think multinetpol is applicable only for non primary networks.


> I guess througout the PR, why do we check `IsNetworkSegmentationSupportEnabled` along with `IsPrimaryNetwork`?
> 
This is a common practice being used all over the places. isn't this kind of gating check ? 
If this is not needed at secondary network controller level, can remove this.

go-controller/pkg/ovn/base_secondary_layer2_network_controller.go
pperiyasamy: yes, done.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
pperiyasamy: @tssurya I've addressed it with this PR, please take a look.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -300,14 +301,37 @@ func GetPodCIDRsWithFullMask(pod *v1.Pod, nInfo NetInfo) ([]*net.IPNet, error) {
...
-		return SecondaryNetworkPodIPs(pod, nInfo)
+	if !nInfo.IsSecondary() || PodWantsHostNetwork(pod) {
+		return DefaultNetworkPodIPs(pod)
+	}
+	if nInfo.IsPrimaryNetwork() {
+		nadList := nInfo.GetNADs()
+		for _, nad := range nadList {
+			userNet, err := UnmarshalPodAnnotation(pod.Annotations, nad)
+			if err != nil && strings.Contains(err.Error(), "no ovn pod annotation for network") {
+				// NAD belongs to same network, but serving different namespace. so skip processing it.
```
jcaamano, line 320: Some things here:

- Un-marshaling the annotation for each NAD is inefficient
- Why would be a pod be annotated with a NAD of a different namespace?
- Use a well defined error. We should we using `newAnnotationNotSetError`, we are probably not doing it for historic reasons but I don't see why.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -696,12 +700,28 @@ func (bnc *BaseNetworkController) doesNetworkRequireIPAM() bool {
...
 }
 
-func (bnc *BaseNetworkController) getPodNADNames(pod *kapi.Pod) []string {
+func (bnc *BaseNetworkController) getPodNADNames(pod *kapi.Pod) ([]string, error) {
 	if !bnc.IsSecondary() {
-		return []string{types.DefaultNetworkName}
+		return []string{types.DefaultNetworkName}, nil
+	}
+	if bnc.IsPrimaryNetwork() {
+		// We couldn't rely on bnc.NetInfo because this may span across different namespaces.
```
jcaamano, line 708: You could also comment here that `PodNadNames` is based on the network selection element so it wouldn't work. All this implementation could be in `util.PodNadNames` I think

test/scripts/e2e-cp.sh
```diff
@@ -183,3 +185,32 @@ go test -test.timeout 180m -v . \
...
+
+  export NUM_WORKER_NODES=3
+  if [ "$SINGLE_NODE_CLUSTER" == true ]; then
+    export NUM_WORKER_NODES=1
+  fi
+
+  # NetPol with named port is not supported in ovn-kubernetes, so skip those tests.
+  ginkgo --nodes=${NUM_NODES} \
+	  --focus="Netpol" \
+	  --skip="named port.+\[Feature:NetworkPolicy\]" \
```
jcaamano, line 205: Why do we need this when we already have 
https://github.com/ovn-org/ovn-kubernetes/blob/master/test/scripts/e2e-kind.sh#L28-L29

go-controller/pkg/ovn/base_network_controller_secondary.go
jcaamano: So if segmentation is not enabled, why woudl we have a network controller for a primary network? It shouldn't exist in the first place, already checked by the NAD controller. No harm I guess though.

go-controller/pkg/util/pod_annotation.go
jcaamano: Can you add a comment?

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: added comment and i tried to move this logic into util.PodNadNames, but it requires nadLister nadlister.NetworkAttachmentDefinitionLister, so needed intrusive changes across all the calling methods. do you still this to be part of this PR or can it be done separately ?

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -720,19 +760,34 @@ func (bsnc *BaseSecondaryNetworkController) deleteNamespace4SecondaryNetwork(ns
...
+	if !util.IsNetworkSegmentationSupportEnabled() || !bsnc.IsPrimaryNetwork() {
+		return nil
+	}
+	if bsnc.netPolicyHandler != nil {
+		return nil
+	}
+	handler, err := bsnc.retryNetworkPolicies.WatchResource()
```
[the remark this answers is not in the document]
pperiyasamy: oh! no problem, didn't notice it, using different variables now.

test/scripts/e2e-cp.sh
pperiyasamy: the e2e-kind.sh script does not run for network segmentation lane instead e2e-cp.sh script is [invoked](https://github.com/ovn-org/ovn-kubernetes/blob/master/.github/workflows/test.yml#L551) for it. hence another skip for named port netpol tests here.

go-controller/pkg/util/pod_annotation.go
pperiyasamy: > * Un-marshaling the annotation for each NAD is inefficient

do we have any other option other than parsing netwoks annotation ?

> * Why would be a pod be annotated with a NAD of a different namespace?

This function may be invoked with netInfo object which is global object for multiple namespaces, hence needed logic to fetch right NAD name for the namespace.

> * Use a well defined error. We should we using `newAnnotationNotSetError`, we are probably not doing it for historic reasons but I don't see why.

yes, using `newAnnotationNotSetError` for this purpose now, didn't see any issue during the tests.

go-controller/pkg/ovn/base_secondary_layer2_network_controller.go
```diff
@@ -106,6 +109,11 @@ func (oc *BaseSecondaryLayer2NetworkController) run() error {
 		return err
 	}
 
+	// WatchNetworkPolicy depends on WatchPods and WatchNamespaces
+	if err := oc.WatchNetworkPolicy(); err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
pperiyasamy: sure, done.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -720,19 +760,34 @@ func (bsnc *BaseSecondaryNetworkController) deleteNamespace4SecondaryNetwork(ns
...
+	if err == nil {
```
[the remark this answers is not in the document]
pperiyasamy: done.

go-controller/pkg/ovn/base_network_controller_namespace.go
```diff
@@ -71,12 +71,16 @@ func getNamespaceAddrSetDbIDs(namespaceName, controller string) *libovsdbops.DbO
...
 // back the appropriate handler logic
 func (bnc *BaseNetworkController) WatchNamespaces() error {
-	if bnc.IsSecondary() {
+	if bnc.IsPrimaryNetwork() && !util.IsNetworkSegmentationSupportEnabled() {
+		// For primary user defined networks, we don't have to watch namespace events if
+		// network segmentation support is not enabled.
+		return nil
+	}
+
+	if bnc.IsSecondary() && !bnc.IsPrimaryNetwork() && !util.IsMultiNetworkPoliciesSupportEnabled() {
```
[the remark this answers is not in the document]
pperiyasamy: oh! yes, multinetpolicy is also enabled when multinetwork is set. 
removed  the check `!bnc.IsPrimaryNetwork() `now.

go-controller/pkg/ovn/base_network_controller_secondary.go
pperiyasamy: you are right, removed `IsNetworkSegmentationSupportEnabled` check whereever it's possible.

go-controller/pkg/util/pod_annotation.go
pperiyasamy: done.

go-controller/pkg/util/pod_annotation.go
jcaamano: > > * Un-marshaling the annotation for each NAD is inefficient
> 
> do we have any other option other than parsing netwoks annotation ?

You can parse once getting a map of NADs to IPs. Instead of parsing the full annotation for each NAD.

> > * Why would be a pod be annotated with a NAD of a different namespace?
> 
> This function may be invoked with netInfo object which is global object for multiple namespaces, hence needed logic to fetch right NAD name for the namespace.

So why not use `GetActiveNetworkForNamespace` which will give you the NADs for the namespace? This is related to the previous comment, you are parsing the annotation for each NAD when many of them might not even be relevant.

> 
> > * Use a well defined error. We should we using `newAnnotationNotSetError`, we are probably not doing it for historic reasons but I don't see why.
> 
> yes, using `newAnnotationNotSetError` for this purpose now, didn't see any issue during the tests.

Did you check that this is ok for previously existing uses of `UnmarshalPodAnnotation`? While it should be fine, we need to be careful.

jcaamano: > So why not use `GetActiveNetworkForNamespace` which will give you the NADs for the namespace? This is related to the previous comment, you are parsing the annotation for each NAD when many of them might not even be relevant.

Meaning, GetActiveNetworkForNamespace should give you a netinfo with a single NAD because there can only be one primary NAD per namespace, so the issue wouldn't be as important.

go-controller/pkg/ovn/base_network_controller.go
jcaamano: But the lister is already being passed to the global function `util.GetActiveNetworkForNamespace`

test/scripts/e2e-cp.sh
jcaamano: But is that is not running conformance tests? I am not aware that we have any `named port.+\[Feature:NetworkPolicy\]` in our control plane tests but I might be missing something.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -380,7 +357,32 @@ func SecondaryNetworkPodIPs(pod *v1.Pod, networkInfo NetInfo) ([]net.IP, error)
 	return ips, nil
 }
 
+func GetPrimaryNetworkNADNameForPodFromNetInfo(pod *v1.Pod, netinfo NetInfo) (string, error) {
+	podNetworks, err := UnmarshalPodAnnotationAllNetworks(pod.Annotations)
+	if err != nil {
+		return "", err
+	}
+	for _, nadName := range netinfo.GetNADs() {
```
jcaamano, line 365: So `nadName` is a namespaced name: `<namespace>/<name>`. 

So you can just parse the namespace and match it against the pod namespace.

I don't think we need to check the role. If we have the pod annotated with a non-primary NAD that we have as primary in NetInfo, we have bigger problems.

jcaamano: I have asked @trozet to add a `GetNADs(namespace string)` method to avoid having to parse the namespaced name however.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -356,7 +358,26 @@ func SecondaryNetworkPodIPs(pod *v1.Pod, networkInfo NetInfo) ([]net.IP, error)
 	return ips, nil
 }
 
+func GetPrimaryNetworkNADNameForNamespaceFromNetInfo(namespace string, netinfo NetInfo) string {
+	for _, nadName := range netinfo.GetNADs() {
+		if !strings.HasPrefix(nadName, namespace) {
```
jcaamano, line 363: `!strings.HasPrefix(nadName, namespace+'/')`

Being strict, we should be using `SplitMetaNamespaceKey` however.

pperiyasamy: ok, using `SplitMetaNamespaceKey`  function to check namespace now.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -899,6 +905,65 @@ func (bnc *BaseNetworkController) findMigratablePodIPsForSubnets(subnets []*net.
...
+		if err := bnc.addNetworkPolicy(newNp); err != nil {
+			klog.Infof("NetworkPolicy add failed for %s/%s, will try again later: %v",
+				newNp.Namespace, newNp.Name, err)
+			return err
+		}
+	default:
+		return fmt.Errorf("object type %s not supported", objType)
+	}
+	return nil
+}
```
jcaamano, line 952: I don't think update for network policies is going to be called, from 
https://github.com/jcaamano/ovn-kubernetes/blob/14afcbf9621264f6d1de32107bfdcf77f655bbbc/go-controller/pkg/ovn/base_event_handler.go#L25-L26

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -899,6 +905,65 @@ func (bnc *BaseNetworkController) findMigratablePodIPsForSubnets(subnets []*net.
...
+			return fmt.Errorf("could not cast %T object to *knet.NetworkPolicy", obj)
+		}
+
+		if err := bnc.addNetworkPolicy(np); err != nil {
+			klog.Infof("Network Policy add failed for %s/%s, will try again later: %v",
+				np.Namespace, np.Name, err)
+			return err
+		}
+	default:
+		return fmt.Errorf("no add function for object type %s", objType)
```
jcaamano, line 922: Technically I would just log here because the retries would be useless

test/scripts/e2e-cp.sh
jcaamano: We discussed not adding this in this PR for the time being.

go-controller/pkg/ovn/base_network_controller_policy.go
jcaamano: https://issues.redhat.com/browse/SDN-5290

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: right, removed the method as it's no longer needed.

jcaamano:
Potential flake
 [FAIL] Network Segmentation: services on a user defined primary network should be reachable through their cluster IP and node port [It] L3 dualstack primary UDN, cluster-networked pods, NodePort service
 https://github.com/ovn-org/ovn-kubernetes/actions/runs/10900030212/job/30253567040?pr=4690

jcaamano:
> Potential flake [FAIL] Network Segmentation: services on a user defined primary network should be reachable through their cluster IP and node port [It] L3 dualstack primary UDN, cluster-networked pods, NodePort service https://github.com/ovn-org/ovn-kubernetes/actions/runs/10900030212/job/30253567040?pr=4690

@pperiyasamy does not look like a flake

trozet (approved):
minor comment nits, lgtm

test/e2e/network_segmentation_policy.go
```diff
@@ -0,0 +1,344 @@
...
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				ginkgo.By("asserting the *client* pod can not contact the server pod exposed endpoint")
+				gomega.Eventually(func() error {
+					return reachToServerPodFromClient(cs, serverPodConfig, clientPodConfig, serverIP, port)
+				}, 1*time.Minute, 6*time.Second).ShouldNot(gomega.Succeed())
+
+			},
+			ginkgo.Entry(
+				"in L2 dualstack primary UDN",
```
trozet, line 125: the title of this PR says netpol for L3, but here are L2 test cases and changes to layer 2 controller. So is L2 also supported and the title just needs to be updated?

pperiyasamy: yes @trozet , updated title and description accordingly now.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -395,6 +379,20 @@ func PodNadNames(pod *v1.Pod, netinfo NetInfo) ([]string, error) {
...
+func GetPrimaryNetworkNADNamesForNamespaceFromNetInfo(namespace string, netinfo NetInfo) ([]string, error) {
+	for _, nadName := range netinfo.GetNADs() {
+		ns, _, err := cache.SplitMetaNamespaceKey(nadName)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing nad name %s from network %s: %v", nadName, netinfo.GetNetworkName(), err)
+		}
+		if ns != namespace {
+			continue
+		}
+		return []string{nadName}, nil
```
jcaamano, line 391: This is based on two assumptions:

- There is a single NAD per namespace for primary networks. However, there is nothing currently that prevents this from happening.
- We are aware of the aware of the primary network of a pod. However this might be reacey.

The first issue should be handled in some way by the NAD controller. It might do that through `GetActiveNetworkForNamespace`. For the second issue, `GetActiveNetworkForNamespace` offers a bit more of protection since it will check against the UDN CRDs as well.

I don't want to hold this because of that. But I am entertaining the idea that we could just implement `GetActiveNetworkForNamespace` in NetInfo that would just end up calling the nad controller `GetActiveNetworkForNamespace` which is feasible since the nad controller is the one building those NetInfo after all.

test/e2e/network_segmentation_policy.go
```diff
@@ -0,0 +1,352 @@
...
+						},
+					}, metav1.CreateOptions{})
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+					defer func() {
+						gomega.Expect(cs.CoreV1().Namespaces().Delete(
+							context.Background(),
+							namespace,
+							metav1.DeleteOptions{},
+						)).To(gomega.Succeed())
+					}()
```
jcaamano, line 205: Don't use defer to clean up please, it hinders troubleshooting. Use `f.AddNamespacesToDelete`; or manually in `AfterEach` depending on how the framework is configured and if the test case failed or not.

npinaeva:
Is it intended to configure netpol for both default and primary UDN at the same time?

pperiyasamy:
> Is it intended to configure netpol for both default and primary UDN at the same time?

yes @npinaeva , The netpol for default and primary UDN networks would coexist with this PR as well. The existing shard-conformance CI lane tests netpol over default network and we had few e2e's to test NetPol over UDN network. 
We also ran upstream NetPol tests over UDN L3 network on this PR and it worked fine. We're waiting for https://github.com/ovn-org/ovn-kubernetes/pull/4653 to run upstream tests over L2 network. Does it suffice ? or do you want to cover something more ?

test/e2e/network_segmentation_policy.go
pperiyasamy: yes @jcaamano , modified the test to use `f.AddNamespacesToDelete`.

go-controller/pkg/util/pod_annotation.go
pperiyasamy: I like the idea of having `GetActiveNetworkForNamespace` inside netInfo.

npinaeva:
> yes @npinaeva , The netpol for default and primary UDN networks would coexist with this PR as well. The existing shard-conformance CI lane tests netpol over default network and we had few e2e's to test NetPol over UDN network. We also ran upstream NetPol tests over UDN L3 network on this PR and it worked fine. We're waiting for #4653 to run upstream tests over L2 network. Does it suffice ? or do you want to cover something more ?

So when primary UDN exists, we will configure network policy both for UDN and for default network, which means 2 times more port groups, acls, address sets, which will not be used (as default network traffic has its own ACLs that deny almost everything already). So UDN netpol will work, but not sure if it is the most optimal solution.

jcaamano:
> So when primary UDN exists, we will configure network policy both for UDN and for default network, which means 2 times more port groups, acls, address sets, which will not be used (as default network traffic has its own ACLs that deny almost everything already). So UDN netpol will work, but not sure if it is the most optimal solution.

@npinaeva has a point.

So when handling a policy we would have to `GetActiveNetworkForNamespace(netpolNamespace)` and check that it matches our network. Seems straight forward enough.

pperiyasamy:
> @npinaeva has a point.
> 
> So when handling a policy we would have to `GetActiveNetworkForNamespace(netpolNamespace)` and check that it matches our network. Seems straight forward enough.

Good find ! addressed this with commit [b5d3dc4](https://github.com/ovn-org/ovn-kubernetes/pull/4690/commits/b5d3dc4ced6219d35c85e339e12ac3c7e9327ebc).

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -727,6 +727,26 @@ func (h *defaultNetworkControllerEventHandler) AddResource(obj interface{}, from
...
+			return fmt.Errorf("could not cast %T object to *knet.NetworkPolicy", obj)
+		}
+
+		netinfo, err := h.oc.getActiveNetworkForNamespace(np.Namespace)
+		if err != nil {
+			return fmt.Errorf("could not get active network for namespace %s: %v", np.Namespace, err)
+		}
+		if !netinfo.IsDefault() {
+			return nil
+		}
```
jcaamano, line 742: Why don't you do this in `base_network_controller.go`?

You would just have to check `bnc.GetNetwokrName() != netinfo.GetNetworkName()`

How would you otherwise prevent the netpol that is processed by the default network controller to be processed as well on primary network controllers?

pperiyasamy: ok @jcaamano , having this in  `base_network_controller.go`  check is a good idea. done.
one question though: secondary network controllers never handle default network event ? correct ?

jcaamano: No sure what you mean. Non primary but still secondary networks don't handle events for `factory.PolicyType` because the retry framework should not be setup for them.

jcaamano (approved):

tssurya:
wohoo :) thanks @jcaamano / @pperiyasamy for working on this and getting it to a good shape

PR #4696, opened 2024-09-04
Add UDN pod annotation to open default network ports.

npinaeva:
An example annotation is
```
    k8s.ovn.org/open-default-ports: |
      - protocol: tcp
        port: 80
      - protocol: udp
        port: 53
```

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -176,6 +176,24 @@ func (oc *DefaultNetworkController) ensureLocalZonePod(oldPod, pod *kapi.Pod, ad
 		}
 	}
 
+	// update open ports for UDN pods on pod update.
+	if util.IsNetworkSegmentationSupportEnabled() && !util.PodWantsHostNetwork(pod) && !addPort &&
+		pod != nil && oldPod != nil &&
+		pod.Annotations[util.UDNOpenPortsAnnotationName] != oldPod.Annotations[util.UDNOpenPortsAnnotationName] {
+		networkRole, err := oc.GetNetworkRole(pod)
+		if err != nil {
+			return err
+		}
+		if networkRole != ovntypes.NetworkRolePrimary {
+			// only update for non-default network pods
+			portName := oc.GetLogicalPortName(pod, oc.GetNetworkName())
+			err := oc.setUDNPodOpenPorts(pod.Namespace+"/"+pod.Name, pod.Annotations, portName)
+			if err != nil {
+				return fmt.Errorf("failed to update UDN pod  %s/%s open ports: %w", pod.Namespace, pod.Name, err)
+			}
+		}
+	}
+
 	if kubevirt.IsPodLiveMigratable(pod) {
 		return kubevirt.EnsureLocalZonePodAddressesToNodeRoute(oc.watchFactory, oc.nbClient, oc.lsManager, pod, ovntypes.DefaultNetworkName)
 	}
--- a/go-controller/pkg/ovn/udn_isolation.go
+++ b/go-controller/pkg/ovn/udn_isolation.go
@@ -4,16 +4,19 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 
 	libovsdbclient "github.com/ovn-org/libovsdb/client"
+	libovsdb "github.com/ovn-org/libovsdb/ovsdb"
 	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
 	libovsdbutil "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/util"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+	utilerrors "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/errors"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


go-controller/pkg/util/pod_annotation.go
```diff
@@ -597,3 +608,35 @@ func AddRoutesGatewayIP(
...
+	if !ok {
+		return result, nil
+	}
+	for _, portDef := range strings.Split(ports, ",") {
+		parsedPort := strings.Split(portDef, ":")
+		if len(parsedPort) != 2 {
+			return nil, fmt.Errorf("failed to parse port definition %q", portDef)
+		}
+		protocol := strings.ToLower(parsedPort[0])
+		if protocol != "tcp" && protocol != "udp" && protocol != "sctp" {
```
trozet, line 626: what if a user wants to specify a protocol other than these 3? Let's say a user just wants to be able to ping a pod with ICMP. Should we allow a user to just match on port if protocol is empty? or does ovn support matching on ICMP as well?

trozet (commented):
@martinkennelly PTAL

go-controller/pkg/ovn/ovn.go
```diff
@@ -176,6 +176,23 @@ func (oc *DefaultNetworkController) ensureLocalZonePod(oldPod, pod *kapi.Pod, ad
 		}
 	}
 
+	// update open ports for UDN pods on pod update.
+	if !util.PodWantsHostNetwork(pod) && !addPort && pod != nil && oldPod != nil &&
+		!reflect.DeepEqual(pod.Annotations, oldPod.Annotations) {
```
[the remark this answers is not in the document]
npinaeva: yes, we can actually replace deepequal with that check

go-controller/pkg/util/pod_annotation.go
npinaeva: We can add icmp match, but I am not sure why anyone may need to just ping a pod, even for some liveness check I would expect something more than just a ping.
I didn't get the point about matching on port without protocol - how is that expected to work?

go-controller/pkg/ovn/ovn.go
```diff
@@ -176,6 +176,23 @@ func (oc *DefaultNetworkController) ensureLocalZonePod(oldPod, pod *kapi.Pod, ad
 		}
 	}
 
+	// update open ports for UDN pods on pod update.
+	if !util.PodWantsHostNetwork(pod) && !addPort && pod != nil && oldPod != nil &&
```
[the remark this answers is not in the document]
npinaeva: nice catch!

go-controller/pkg/util/pod_annotation.go
```diff
@@ -51,6 +53,9 @@ const (
 	OvnPodAnnotationName = "k8s.ovn.org/pod-networks"
 	// DefNetworkAnnotation is the pod annotation for the cluster-wide default network
 	DefNetworkAnnotation = "v1.multus-cni.io/default-network"
+	// UDNOpenPortsAnnotationName is the pod annotation to open default network pods on UDN pods.
+	// Annotation content has format: "tcp:80,udp:53"
```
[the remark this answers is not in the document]
trozet: I think there are 2 advantages:
1. The user is used to writing yaml lists like this and I think it is easier to read.
2. The input is now encoded yaml, so rather than having to add a for loop logic to try to split on csv and read values, you can just unmarshal it to a native data type: var data []map[string]string.

go-controller/pkg/util/pod_annotation.go
trozet: we have had some users in the past that just want to use icmp to make sure their pod is alive. I guess we cannot match on just port without specifying protocol. Let's see if we can include ICMP and if there are any other protocols missing that OVN supports matching on.

npinaeva:
added some unit tests

tssurya:
oops needs rebase,,,
@martinkennelly can you do another round?

go-controller/pkg/ovn/udn_isolation.go
```diff
@@ -136,3 +141,96 @@ func (oc *DefaultNetworkController) getUDNACLDbIDs(name string, aclDir libovsdbu
...
+			return ops, parseErr, fmt.Errorf("failed to add open ports ACLs to portGroup %s: %v", udnPGName, err)
+		}
+	}
+	return ops, parseErr, nil
+}
+
+func (oc *DefaultNetworkController) getUDNOpenPortDbIDs(podName string, aclDir libovsdbutil.ACLDirection) *libovsdbops.DbObjectIDs {
+	return libovsdbops.NewDbObjectIDs(libovsdbops.ACLUDN, oc.controllerName,
+		map[libovsdbops.ExternalIDKey]string{
+			libovsdbops.ObjectNameKey:      OpenPortACLPrefix + podName,
```
martinkennelly, line 233: should we not have namespace also in the name key? Pod name isnt unique throughout a cluster.

martinkennelly (commented):
otherwise lgtm

go-controller/pkg/ovn/udn_isolation.go
npinaeva: yeah bad variable name, in reality namespaced name is used

martinkennelly (approved):

martinkennelly:
just rebase? @npinaeva

go-controller/pkg/util/pod_annotation.go
```diff
@@ -615,3 +625,34 @@ func AddRoutesGatewayIP(
...
+func UnmarshalUDNOpenPortsAnnotation(annotations map[string]string) ([]*OpenPort, error) {
+	result := []*OpenPort{}
+	ports, ok := annotations[UDNOpenPortsAnnotationName]
+	if !ok {
+		return result, nil
+	}
+	if err := yaml.Unmarshal([]byte(ports), &result); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal UDN open ports annotation %s: %v", ports, err)
+	}
+	allowedProtocols := sets.New("tcp", "udp", "sctp", "icmp")
```
trozet, line 640: does this also cover ipv6?

npinaeva: yes, added comment https://github.com/ovn-org/ovn-kubernetes/compare/f2fba8ee2a4fd35af8e21e6f62665cc721f1ab9d..005a822c86dcc3966c3f6396578682511b12b288

trozet (approved):

PR #4697, opened 2024-09-04
UDN LGW: ensure masq chain exists before adding rules

martinkennelly:
cc @tssurya 

Fixes https://github.com/ovn-org/ovn-kubernetes/issues/4695

### DIFF
--- a/go-controller/pkg/node/gateway_iptables.go
+++ b/go-controller/pkg/node/gateway_iptables.go
@@ -79,6 +79,18 @@ func deleteIptRules(rules []nodeipt.Rule) error {
 	return nodeipt.DelRules(rules)
 }
 
+// ensureChain ensures that a chain exists within a table
+func ensureChain(table, chain string) error {
+	for _, proto := range clusterIPTablesProtocols() {
+		ipt, err := util.GetIPTablesHelper(proto)
+		if err != nil {
+			return fmt.Errorf("failed to get IPTables helper to add UDN chain: %v", err)
+		}
+		addChaintoTable(ipt, table, chain)
+	}
+	return nil
+}
+
 func getGatewayInitRules(chain string, proto iptables.Protocol) []nodeipt.Rule {
 	iptRules := []nodeipt.Rule{}
 	if chain == egressservice.Chain {
--- a/go-controller/pkg/node/gateway_localnet.go
+++ b/go-controller/pkg/node/gateway_localnet.go
@@ -26,6 +26,11 @@ func newLocalGateway(nodeName string, hostSubnets []*net.IPNet, gwNextHops []net
 	klog.Info("Creating new local gateway")
 	gw := &gateway{}
 
+	if util.IsNetworkSegmentationSupportEnabled() {
+		if err := ensureChain("nat", iptableUDNMasqueradeChain); err != nil {
+			return nil, fmt.Errorf("failed to ensure chain %s in NAT table: %w", iptableUDNMasqueradeChain, err)
+		}
+	}
 	for _, hostSubnet := range hostSubnets {
 		// local gateway mode uses mp0 as default path for all ingress traffic into OVN
 		var nextHop *net.IPNet


go-controller/pkg/node/gateway_iptables.go
```diff
@@ -79,6 +79,18 @@ func deleteIptRules(rules []nodeipt.Rule) error {
 	return nodeipt.DelRules(rules)
 }
 
+// ensureChain ensures that a chain exists within a table
+func ensureChain(table, chain string) error {
```
tssurya, line 83: I think we already have a util for this somewhere no?
I thought insertIpt does chain creation automatically... how is this crash transient and goes away eventually?

martinkennelly: I couldnt find any.

If its a rule for a chain then itll auto create the chain, but if its a jump, it wont create the chain you jump to.
I assume it hit a rule during init startup that then created the chain but the error inserting the jump rule happened earlier and following reboot of the container, the new chain is there.

JacobTanenbaum: could you use the IPTablesHelper function `NewChain()` similar to https://github.com/ovn-org/ovn-kubernetes/blob/244efcf34e99780cab97793392d76e6e2ed77807/go-controller/pkg/node/iptables/iptables.go#L108-L111

go-controller/pkg/node/gateway_iptables.go
```diff
@@ -79,6 +79,18 @@ func deleteIptRules(rules []nodeipt.Rule) error {
...
+	for _, proto := range clusterIPTablesProtocols() {
+		ipt, err := util.GetIPTablesHelper(proto)
+		if err != nil {
+			return fmt.Errorf("failed to get IPTables helper to add UDN chain: %v", err)
+		}
+		addChaintoTable(ipt, table, chain)
```
tssurya, line 89: ah! I was talking about addChainToTable...OK looks like there isn't a wrapper for that and hence this happens..

tssurya (approved):

go-controller/pkg/node/gateway_iptables.go
tssurya: > 
> If its a rule for a chain then itll auto create the chain, but if its a jump, it wont create the chain you jump to. I assume it hit a rule during init startup that then created the chain but the error inserting the jump rule happened earlier and following reboot of the container, the new chain is there.

yea that's it. thanks martin for rca

tssurya: @JacobTanenbaum yea

```
func addChaintoTable(ipt util.IPTablesHelper, tableName, chain string) {
	if err := ipt.NewChain(tableName, chain); err != nil {
		klog.V(5).Infof("Chain: \"%s\" in table: \"%s\" already exists, skipping creation: %v", chain, tableName, err)
	}
}
```
he is indirectly using that

go-controller/pkg/node/gateway_iptables.go
kyrtapz: I know it is unrelated to this change but `addChaintoTable` ignores all errors, not only the ones where the chain already exists. I feel like it would be best to introduce a new method to the helper, something like:
```
// EnsureChain creates a new chain if it doesn't already exist
func (ipt *IPTables) EnsureChain(table, chain string) error {
	exists, err := ipt.ChainExists(table, chain)
	if err != nil || exists {
		return err
	}
	return ipt.NewChain(table, chain)
}
```
This doesn't have to happen in this PR.

kyrtapz (approved):

go-controller/pkg/node/gateway_localnet.go
```diff
@@ -26,6 +26,11 @@ func newLocalGateway(nodeName string, hostSubnets []*net.IPNet, gwNextHops []net
 	klog.Info("Creating new local gateway")
 	gw := &gateway{}
 
+	if util.IsNetworkSegmentationSupportEnabled() {
+		if err := ensureChain("nat", iptableUDNMasqueradeChain); err != nil {
```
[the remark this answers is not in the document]
martinkennelly: i think the purpose of that func was generate the rules and apply them, hence i put the creation of a chain outside it.

tssurya (approved):

PR #4707, opened 2024-09-09
Add unit tests for UDN while DS is true

### LINKED ISSUE #4684: Add unit tests for UDNs and DS=true
### What would you like to see added/updated/deleted in our unit tests?

All UDN topology UTs are now running with DS=false; will be good to run some with DS=true

### Why is this needed?

For enhanced coverage

arghosh93:
This commit is to add some unit tests to make sure proper NAT entries are being created i NBDB while DisableSNATMultipleGWs is set to true.

Fixes #4684 

```release-note

```


trozet (approved):

tssurya:
thanks for adding tests!

PR #4712, opened 2024-09-10
Adds e2e test: conntrack flush after ovnkube delete + bumps OVN with fix

trozet:
Test opens a TCP connection that simulates a GCP LB environment where the packet is redirected via iptables to a local server on a node. Note, in GCP the LB does not DNAT the VIP, so the packet arrives to the node with the GCP VIP on it. In OCP, we then redirect that packet to the local kapi server running on the node.

Once the test opens the TCP connection, it leaves it open for 2 minutes while ovnkube-node is then deleted. Post ovn-controller starting it should not flush the conntrack in zone 0, and the test ensures that the conntrack entry still exists.

Recent OVN regression that prompted this E2E: https://issues.redhat.com/browse/FDP-773

I tested this with OVN 24.03 and it passed. Testing with 24.09 will fail due to the above regression. We should wait until an OVN version is ready with the fix for FDP-773 and include it in this PR for OVN bump.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -46,9 +46,69 @@ const (
 	agnhostImage         = "registry.k8s.io/e2e-test-images/agnhost:2.26"
 	iperf3Image          = "quay.io/sronanrh/iperf"
 	ovnNs                = "ovn-kubernetes" //OVN kubernetes namespace
+	redirectIP           = "123.123.123.123"
+	redirectPort         = "13337"
+	exContainerName      = "tcp-continuous-client"
 )
 
 type podCondition = func(pod *v1.Pod) (bool, error)
+
+// setupHostRedirectPod
+func setupHostRedirectPod(f *framework.Framework, node *v1.Node, exContainerName string, isIPv6 bool) error {
+	_, _ = createClusterExternalContainer(exContainerName, externalContainerImage, []string{"-itd", "--privileged", "--network", externalContainerNetwork}, []string{})
+	nodeV4, nodeV6 := getContainerAddressesForNetwork(node.Name, externalContainerNetwork)
+	mask := 32
+	ipCmd := []string{"ip"}
+	nodeIP := nodeV4
+	if isIPv6 {
+		mask = 128
+		ipCmd = []string{"ip", "-6"}
+		nodeIP = nodeV6
+	}
+	cmd := []string{"docker", "exec", exContainerName}
+	cmd = append(cmd, ipCmd...)
+	cmd = append(cmd, "route", "add", fmt.Sprintf("%s/%d", redirectIP, mask), "via", nodeIP)
+	_, err := runCommand(cmd...)
+	if err != nil {
+		return err
+	}
+
+	// setup redirect iptables rule in node
+	ipTablesArgs := []string{"PREROUTING", "-t", "nat", "--dst", redirectIP, "-j", "REDIRECT"}
+	updateIPTablesRulesForNode("insert", node.Name, ipTablesArgs, isIPv6)
+
+	command := []string{
+		"bash", "-c",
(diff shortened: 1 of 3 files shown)


trozet:
flake: 
```
2024-09-10T20:12:59.8376788Z 
2024-09-10T20:12:59.8377677Z [91m[1m[Fail] [0m[90mHybrid SDN Master Operations [0m[91m[1m[It] handles a HO node is switched to a OVN node [0m
2024-09-10T20:12:59.8379428Z [37m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:1387[0m
```
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10799611587/job/29955734477?pr=4712

tssurya:
The HO has a flake issue... I've seen it before...

dceara:
@trozet I didn't really review the code but what I think we shouldn't forget to do is backport this PR to branch `origin/release-1.0`.  That's what ovn-org/ovn uses in CI and it would allow us to catch breakages early.

dceara:
We should bump ovn to [ovn-24.03.90-7.fc42](https://koji.fedoraproject.org/koji/buildinfo?buildID=2544099).  This includes the fix for the ct zone flush.

trozet:
looks like everything is good except one job it flaked:
[38;5;9m[FAILED] failed to run "docker exec tcp-continuous-client ip route add 123.123.123.123/32 via 172.18.0.4": exit status 2 (Error: Nexthop has invalid gateway

I need to account for the 2 gateway bridge scenario. Will post a fix.

dceara (approved):
FWIW, this looks ok to me.

test/e2e/util.go
```diff
@@ -1045,31 +1046,37 @@ func allowOrDropNodeInputTrafficOnPort(op, nodeName, protocol, port string) {
...
+	if ipv6 {
+		iptables = "ip6tables"
+	}
+
+	args = []string{"exec", ovnKubePodName, "-c", getNodeContainerName(), "--", iptables, "--check"}
+	_, err := e2ekubectl.RunKubectl(ovnNamespace, append(args, ipTablesArgs...)...)
+	// errors known to be equivalent to not found
+	notFound1 := "No chain/target/match by that name"
+	notFound2 := "does a matching rule exist in that chain?"
+	notFound := err != nil && (strings.Contains(err.Error(), notFound1) || strings.Contains(err.Error(), notFound2))
```
tssurya, line 1066: we might want to improve this in future... are we sure we only expect these two?

dceara:
@tssurya can we please cherry pick the first commit to `release-1.0` branch too?  That ensures ovn-org/ovn CI runs the test too.  Or do you prefer if I open a PR with the cherry pick?

tssurya:
@dceara yeah go for it! Please. Cherry pick to 1.0. I can merge it

test/e2e/e2e.go
```diff
@@ -700,6 +769,26 @@ var _ = ginkgo.Describe("e2e control plane", func() {
...
+
+		// start TCP client
+		go func() {
+			defer ginkgo.GinkgoRecover()
+			_, _ = runCommand(containerRuntime, "exec", exContainerName, "nc", "--idle-timeout", "120s", redirectIP, redirectPort)
+		}()
+
+		ginkgo.By("Checking that TCP redirect connection entry in conntrack before ovnkube-node restart")
+		gomega.Eventually(func() int {
+			return pokeConntrackEntries(nodeName, redirectIP, "tcp", nil)
```
flavio-fernandes, line 790: NOTE: this function "pokeConntrackEntries" assumes that ovs is being ran as a pod, which is a problem for clusters that have ovs as part od systemctl in the node.

PR #4735, opened 2024-09-18
Fix arp proxy configuration for UDNs

jcaamano:


jcaamano:
@qinqon PTAL

qinqon:
@jcaamano I am not sure about this, it only make sense for L3 live migration, for L2 is not needed, I don't expect implementing live migration at UDN L3 anytime soon, but who knows...

jcaamano:
@qinqon PTAL

dist/images/.gitignore
```diff
@@ -3,6 +3,7 @@ ovn-kube-util
 ovnkube
 ovnkube-trace
 ovnkube-identity
+ovnkube-observ
```
qinqon, line 6: This is not related.

qinqon (changes requested):

dist/images/.gitignore
jcaamano: what do you want me to do?

qinqon: I mean, I am ok adding those to this PR since the PR itself is super small, if you think is not worth it to do the .gitignore at different PR I am fine, we can go with it.

qinqon:
> failing everywhere now
> 
> [FAIL] Services of type NodePort [AfterEach] should handle IP fragments

To extend on that, the error is 
```
[38;5;9m[FAILED] failed to remove ip address fc00:f853:ccd:e793:1111::0 from node ovn-control-plane, err: "failed to run \"docker exec ovn-control-plane ip addr delete fc00:f853:ccd:e793:1111::0/128 dev breth0\": exit status 2 (Error: ipv6: address not found.\n)"[0m
```

Don't get the relation, also those lanes are all about default network wich is not changed by this PR.

qinqon:
Ahh looks like this is the fix:
- https://github.com/ovn-org/ovn-kubernetes/pull/4740

jcaamano:
kubevirt is also failing now

trozet (approved):

PR #4740, opened 2024-09-19
Services NodePort E2Es: Reset shared variables before each test

kyrtapz:
Reset shared variables before each test to avoid cleanup issues.
Fixes: https://github.com/ovn-org/ovn-kubernetes/issues/4733.

### DIFF
--- a/test/e2e/service.go
+++ b/test/e2e/service.go
@@ -708,11 +708,8 @@ var _ = ginkgo.Describe("Services", func() {
 		var nodes *v1.NodeList
 		var err error
 		nodeIPs := make(map[string]map[int]string)
-		var egressPod *v1.Pod
 		var egressNode string
-		targetSecondaryNode := node{
-			name: "egressSecondaryTargetNode-allowed",
-		}
+		var targetSecondaryNode node
 
 		const (
 			endpointHTTPPort    = 80
@@ -721,6 +718,14 @@ var _ = ginkgo.Describe("Services", func() {
 			clusterUDPPort      = 91
 			clientContainerName = "npclient"
 		)
+
+		ginkgo.BeforeEach(func() {
+			nodeIPs = make(map[string]map[int]string)
+			egressNode = ""
+			targetSecondaryNode = node{
+				name: "egressSecondaryTargetNode-allowed",
+			}
+		})
 
 		ginkgo.AfterEach(func() {
 			ginkgo.By("Cleaning up external container")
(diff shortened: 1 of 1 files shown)


test/e2e/service.go
```diff
@@ -740,6 +740,7 @@ var _ = ginkgo.Describe("Services", func() {
 					}
 				}
 			}
+			nodeIPs = make(map[string]map[int]string)
```
dceara, line 743: Shouldn't we do this also just before the `framework.Failf("failed to remove ip address %s from node %s, err: %q", ip, nodeName, err)` call above?

dceara:
The commit log probably needs a
```
Fixes: https://github.com/ovn-org/ovn-kubernetes/issues/4733
```

dceara:
Also, @tssurya can we please also cherry pick this to the `release-1.0` branch after it gets merged to master?  We hit this in ovn-orv/ovn CI too.

test/e2e/service.go
dceara: I meant when we fail.  `framework.Failf(..)` will break out of the loop, right?  But if we don't reset the map there we might cause other tests in the suite (that are executed after the currently failing one) to fail too.

npinaeva: came here to post the same fix :)
I think the most robust way would be to move 
```
nodeIPs = make(map[string]map[int]string)
// and maybe this one for company
targetSecondaryNode = node{
			name: "egressSecondaryTargetNode-allowed",
		}
```
to BeforeEach?

kyrtapz: Good point, I will wait for the control-plane tests to pass and will change this to:
```
ginkgo.BeforeEach(func() {
	nodeIPs = make(map[string]map[int]string)
})
```
I think it makes more sense logically anyway.

kyrtapz: Actually, I will reset all shared vars to avoid any future issues.

kyrtapz: All control-plane lanes are green, re-pushing.

dceara (approved):
Looks good to me, thanks!

jcaamano (approved):

tssurya:
@kyrtapz can you please open a backport to release-1.0 as well?

kyrtapz:
> @kyrtapz can you please open a backport to release-1.0 as well?

https://github.com/ovn-org/ovn-kubernetes/pull/4742

PR #4743, opened 2024-09-20
[DNM] Run test without the fix

npinaeva:


npinaeva:
previous failed run https://github.com/ovn-org/ovn-kubernetes/actions/runs/10957665776/job/30428062312?pr=4743
fixed test names to distinguish new tests

PR #4744, opened 2024-09-20
Fix egress gateway pod cleanup for remote zone pods.

npinaeva:
Cleanup gateway pod for remote zone.
For local zone pods, deleteLogicalPort cleans this up, but before IC
this function was called for all non-host-network pods, hence this
logic. After IC, deleteLogicalPort won't be called for all remote zone
pods, so condition is not needed.

See test result without the fix here https://github.com/ovn-org/ovn-kubernetes/pull/4743

### DIFF
--- a/test/e2e/external_gateways.go
+++ b/test/e2e/external_gateways.go
@@ -730,7 +730,7 @@ var _ = ginkgo.Describe("External Gateway", func() {
 				servingNamespace = ns.Name
 
 				addressesv4, addressesv6 = setupGatewayContainersForConntrackTest(f, nodes, gwContainer1, gwContainer2, srcPodName)
-				sleepCommand = []string{"bash", "-c", "sleep 20000"}
+				sleepCommand = []string{"bash", "-c", "trap : TERM INT; sleep infinity & wait"}
 				_, err = createGenericPod(f, gatewayPodName1, nodes.Items[0].Name, servingNamespace, sleepCommand)
 				framework.ExpectNoError(err, "Create and annotate the external gw pods to manage the src app pod namespace, failed: %v", err)
 				_, err = createGenericPod(f, gatewayPodName2, nodes.Items[1].Name, servingNamespace, sleepCommand)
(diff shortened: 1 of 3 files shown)


test/e2e/external_gateways.go
```diff
@@ -802,7 +802,7 @@ var _ = ginkgo.Describe("External Gateway", func() {
 				ginkgo.Entry("IPV6 udp", &addressesv6, "udp"),
 				ginkgo.Entry("IPV6 tcp", &addressesv6, "tcp"))
 
-			ginkgo.DescribeTable("ExternalGWPod annotation: Should validate conntrack entry deletion for TCP/UDP traffic via multiple external gateways a.k.a ECMP routes", func(addresses *gatewayTestIPs, protocol string) {
+			ginkgo.DescribeTable("ExternalGWPod annotation: Should validate conntrack entry deletion for TCP/UDP traffic via multiple external gateways a.k.a ECMP routes", func(addresses *gatewayTestIPs, protocol string, deletePod bool) {
```
jordigilh, line 805: Wouldn't it be better to split this into its own table/use case? one for annotation deletion and another one for pod deletion. The flow is mostly the same, but the breakdown can help to write a more specific description of what the test does in case the test fails (it failed while deleting annotation or it failed while deleting the pod). WDYT?

npinaeva:
previous green run https://github.com/ovn-org/ovn-kubernetes/actions/runs/10957241679/job/30426092464?pr=4744

test/e2e/external_gateways.go
npinaeva: You can distinguish the tests even now:
1. test names will be different 
2. ginkgo.By will log different steps
I also find it much easier to understand the difference between test sequences when it shows which step exactly differs, e.g. vs having a full copy of this test with 3 different lines

jordigilh: Honestly it's a personal preference between readability vs reusability. I tend to prefer readability in tests even if that means replicating code that can't be extracted as functions (perhaps in this case it could be, but that's another story).

On the other hand, `true` or `false` is not clear what they mean as arguments without understanding the logic inside the test. I'd recommend to use constants whose names define their behavior rather than booleans as arguments for what they mean, e.g. `BY_DELETE_ANNOTATION`, `BY_DELETE_GW_POD`.

It's a cosmetic comment since this logic will go away, eventually I hope 😄 .

jordigilh:
@npinaeva shouldn't the title be updated to include the bug reference [OCPBUGS-38753](https://issues.redhat.com/browse/OCPBUGS-38753)?

npinaeva:
> @npinaeva shouldn't the title be updated to include the bug reference [OCPBUGS-38753](https://issues.redhat.com/browse/OCPBUGS-38753)?

We usually do that for d/s PRs, but I added a link to this one to the bug itself

test/scripts/e2e-cp.sh
```diff
@@ -37,7 +37,7 @@ SKIPPED_TESTS=""
 if [ "$KIND_IPV4_SUPPORT" == true ]; then
     if  [ "$KIND_IPV6_SUPPORT" == true ]; then
 	# No support for these features in dual-stack yet
-	SKIPPED_TESTS="hybrid.overlay|external.gateway"
+	SKIPPED_TESTS="hybrid.overlay"
```
trozet, line 40: yay

trozet (approved):

PR #4745, opened 2024-09-20
Verify udn-allowed-default-services config field

kyrtapz:
Ensure that the values specified in `udn-allowed-default-services` are valid namespaced service names.
Stop using `cli.StringSlice` to allow for setting the field through a config file.

### DIFF
--- a/go-controller/pkg/config/config.go
+++ b/go-controller/pkg/config/config.go
@@ -13,14 +13,14 @@ import (
 	"strings"
 	"time"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-
 	"github.com/urfave/cli/v2"
 	gcfg "gopkg.in/gcfg.v1"
 	lumberjack "gopkg.in/natefinch/lumberjack.v2"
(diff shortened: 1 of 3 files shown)


trozet (approved):

go-controller/pkg/config/config.go
```diff
@@ -2124,13 +2128,39 @@ func completeDefaultConfig(allSubnets *ConfigSubnets) error {
 		allSubnets.Append(ConfigSubnetCluster, subnet.CIDR)
 	}
 
+	Default.UDNAllowedDefaultServices, err = parseServicesNamespacedNames(Default.RawUDNAllowedDefaultServices)
```
ormergi, line 2131: In a scenario of parse fails, `config.Default` will be `nil`.
Other component may read  `config`, right? 
Is so do you think we should change it to not override `Default` in case of an error?

kyrtapz: `Default.UDNAllowedDefaultServices` will be `nil` in case the parse fails but this is fine for a slice.
Additionally this is done at startup so a failure to parse `RawUDNAllowedDefaultServices` will cause ovn-k to exit early.

PR #4746, opened 2024-09-20
SDN-4909: upload logs to testspace

jluhrsen:
- adds composable action to handle kind and test artifacts and upload to gh artifacts as well as junit data to testspace
- adds junit reporting for all test executions
- removes schedule condition for upgrade and dualstack jobs so that they will run with the cron schedule as well

Fixes #

```release-note

```


jluhrsen:
/hold

something not right. will fix it

PR #4747, opened 2024-09-22
ut,multi net: Fix assertion in case of unexpected error

oshoval:
In case there is an unexpected error, for a test that should pass, 
there will be a panic because `MatchError` expects non nil value.
```
panic: reflect: call of reflect.Value.Type on zero Value [recovered]
	panic: reflect: call of reflect.Value.Type on zero Value
```

Fix it to be explicit and also print the error to ease debug.

Fixes #

```release-note

```

### DIFF
--- a/go-controller/pkg/util/multi_network_test.go
+++ b/go-controller/pkg/util/multi_network_test.go
@@ -1049,6 +1049,7 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
 			)
 
 			if err != nil {
+				g.Expect(test.expectedError).NotTo(gomega.BeNil(), err.Error())
 				g.Expect(err).To(gomega.MatchError(test.expectedError))
 			}
 			g.Expect(isAttachmentRequested).To(gomega.Equal(test.expectedIsAttachmentRequested))


maiqueb (approved):
Tks, good catch.

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1044,6 +1044,7 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
 			)
 
 			if err != nil {
+				g.Expect(test.expectedError).ToNot(gomega.BeNil(), err.Error())
```
maiqueb, line 1047: nit:

```suggestion
				g.Expect(test.expectedError).NotTo(gomega.BeNil(), err.Error())
```

oshoval: done

maiqueb (approved):
tks

oshoval:
rebased

oshoval:
@jcaamano 
can you please take a look ?

oshoval:
@tssurya 
can you please take a look on this one liner?
thanks

oshoval:
fixed lint problem

oshoval:
reverted fixing the lint, because it isn't correct
we don't want to check it has occurred, but just that `test.expectedError` is not nil, in case an error occurred

PR #4749, opened 2024-09-25
Update ginkgo to version v2

JacobTanenbaum:
ginkgo has updated to v2 three years ago and we should update our test framework. I think the two most important features of ginkgo/v2 we could make use of is 
    1. nested tables to help create more succinct testing of ovn-kubernetes 
    2. the [FlakeAttempts()](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#flakeattempts-decorator) decorator, which could help diagnose flaky tests and at minimum allow us to unblock development when a flaky test is identified  

Fixes #

```release-note

```


JacobTanenbaum:
this PR should only effect the unit testing and not e2e, rerunning the e2e jobs

maiqueb (dismissed):
Got up to the vendoring code ...

I'm trusting you used a tool to update vendor, because I didn't check file by file the vendored changes are good ...

go-controller/hack/test-go.sh
```diff
@@ -61,7 +61,7 @@ function testrun {
     fi
     if grep -q "ginkgo" ."${path}"/*_test.go; then
 	    prefix=$(echo "${path}" | cut -c 2- | sed 's,/,_,g')
-        ginkgoargs="-ginkgo.v ${ginkgo_focus} -ginkgo.reportFile ${TEST_REPORT_DIR}/junit-${prefix}.xml"
+        ginkgoargs="-ginkgo.v ${ginkgo_focus} --ginkgo.junit-report ${TEST_REPORT_DIR}/junit-${prefix}.xml"
```
jcaamano, line 64: I would just use single dash here as with the other ginkgo parameter

jcaamano (commented):
minor comment and pending rebase to resolve conflicts

go-controller/hack/test-go.sh
JacobTanenbaum: done

jcaamano (approved):

PR #4750, opened 2024-09-26
Kubevirt layer2 UDPN DHCP/NDP support

qinqon:
Virtual machines from kubevirt using primary UDN with layer2 topology need to configure the DHCP/NDP flows to receive the ipv4 and ipv6 address and routing.

This PR configure LSPs with DHCP options and the LRP with dhcpv6_stateful

There is a limitation related to how IPv6 LLAs are generated at the guest, if they are not generated from mac address the PortSecurity configuration will discard them (for example when receiving the ND RS). This is the same limitation as openstack https://opendev.org/openstack/neutron/src/branch/master/doc/source/admin/config-ipv6.rst#configuring-interfaces-of-the-guest

This can be impleneted in a follow up PR adding the full link local range cidr to the PortSecurity.

Also some containers like Alpine triggers a ND RS so it ends up with two default gws for ipv6 
```bash
default via 2014:100:200::1 dev ovn-udn1 metric 1024 pref medium
default via fe80::858:64ff:fe41:3 dev ovn-udn1 proto ra metric 1024 expires 65438sec hoplimit 255 pref medium
```
Since they resolve to the same mac address it should not affect the traffic but e2e test were adjusted.

Also the no default gw routes are not configured as stateless routes at DHCPv4 or RAs for IPV6 since steering traffic using default gw is enough for primary UDN ingress, take into account that if ingress is implemented at secondary L2 UDNs this kind of steering may be needed at the VM, depending on how the virt network binding is implemented.

#### How to verify it
It contains e2e to verify addresses, dns, hostname and routing.

Install DHCP/NDP flows for kubevirt primary UDN layer2 scenarios.

```release-note
Add DHCP/NDP flows for L2 primary UDNs with VMs.
```


tssurya:
@kyrtapz PTAL this is in the integration piece

tssurya:
@maiqueb will also review this

qinqon:
flakiness ? 

```
2024-10-02T13:21:11.9112914Z 
2024-10-02T13:21:11.9113461Z [91m[1mSummarizing 1 Failure:[0m
2024-10-02T13:21:11.9113878Z 
2024-10-02T13:21:11.9114822Z [91m[1m[Fail] [0m[90mHybrid SDN Master Operations [0m[91m[1m[It] handles a HO node is switched to a OVN node [0m
2024-10-02T13:21:11.9116603Z [37m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:1387[0m
2024-10-02T13:21:11.9117521Z 
2024-10-02T13:21:11.9117854Z [1m[91mRan 443 of 444 Specs in 620.447 seconds[0m
2024-10-02T13:21:11.9118661Z [1m[91mFAIL![0m -- [32m[1m442 Passed[0m | [91m[1m1 Failed[0m | [33m[1m1 Pending[0m | [36m[1m0 Skipped[0m
2024-10-02T13:21:11.9119671Z --- FAIL: TestClusterNode (620.69s)
2024-10-02T13:21:11.9120005Z FAIL
```

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -22,17 +22,62 @@ const (
...
+}
+
 func EnsureDHCPOptionsForMigratablePod(controllerName string, nbClient libovsdbclient.Client, watchFactory *factory.WatchFactory, pod *corev1.Pod, ips []*net.IPNet, lsp *nbdb.LogicalSwitchPort) error {
+	dnsServerIPv4, dnsServerIPv6, err := retrieveDNSServiceClusterIPs(watchFactory)
+	if err != nil {
+		return fmt.Errorf("failed retrieving dns service cluster ip: %v", err)
+	}
+
+	return EnsureDHCPOptionsForLSP(controllerName, nbClient, pod, ips, lsp,
+		WithIPv4Router(ARPProxyIPv4),
```
maiqueb, line 69: this is only for hypershift right ? For primary UDN, we would put here the GW, wouldn't we ?

qinqon:
This breaks isolation somehow 
```
[FAILED] Unexpected error:
      <exec.CodeExitError>: 
      error running /usr/local/bin/kubectl --server=https://127.0.0.1:40083 --kubeconfig=/home/ellorent/ovn.conf --namespace=network-segmentation-2872 exec udn-pod -- ip -6 route del default via 2014:100:200::1 dev ovn-udn1 metric 1024 pref medium
      default via fe80::858:64ff:fe41:3 dev ovn-udn1 proto ra metric 1024 expires 65494sec hoplimit 255 pref medium:
      Command stdout:
      
      stderr:
      Error: argument "medium
      default" is wrong: "pref" value is invalid
      
      command terminated with exit code 255
      
      error:
      exit status 255
      {
          Err: <*errors.errorString | 0xc00119d6d0>{
              s: "error running /usr/local/bin/kubectl --server=https://127.0.0.1:40083 --kubeconfig=/home/ellorent/ovn.conf --namespace=network-segmentation-2872 exec udn-pod -- ip -6 route del default via 2014:100:200::1 dev ovn-udn1 metric 1024 pref medium\ndefault via fe80::858:64ff:fe41:3 dev ovn-udn1 proto ra metric 1024 expires 65494sec hoplimit 255 pref medium:\nCommand stdout:\n\nstderr:\nError: argument \"medium\ndefault\" is wrong: \"pref\" value is invalid\n\ncommand terminated with exit code 255\n\nerror:\nexit status 255",
          },
          Code: 255,
      }
  occurred

```

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -104,17 +145,15 @@ func ComposeDHCPv4Options(cidr, dnsServer, controllerName string, vmKey ktypes.N
...
 }
 
-func ComposeDHCPv6Options(cidr, dnsServer, controllerName string, vmKey ktypes.NamespacedName) *nbdb.DHCPOptions {
+func ComposeDHCPv6Options(cidr, controllerName string, vmKey ktypes.NamespacedName) *nbdb.DHCPOptions {
 	serverMAC := util.IPAddrToHWAddr(net.ParseIP(ARPProxyIPv6)).String()
 	dhcpOptions := &nbdb.DHCPOptions{
 		Cidr: cidr,
 		Options: map[string]string{
 			"server_id": serverMAC,
+			"fqdn":      fmt.Sprintf("%q", vmKey.Name),
```
maiqueb, line 154: your commit msg omits you're also setting the hostname for ipv6. 

Maybe consider exporting this to a separate commit ? Also, shouldn't it be an option ? Or are we after ipv4 parity (which always sets the hostname).

go-controller/pkg/kubevirt/dhcp_test.go
```diff
@@ -46,101 +31,175 @@ var _ = Describe("Kubevirt", func() {
...
+						"dns_server": "192.167.23.44",
+					},
+				},
+				V6: &nbdb.DHCPOptions{
+					Cidr: "2002:0:0:1234::/64",
+					ExternalIDs: map[string]string{
+						"k8s.ovn.org/owner-controller": "defaultController",
+						"k8s.ovn.org/owner-type":       "VirtualMachine",
+						"k8s.ovn.org/name":             "namespace1/foo1",
+						"k8s.ovn.org/cidr":             "2002.0.0.1234../64",
```
maiqueb, line 176: this CIDR is fishy ... 

Are replacing the colons for `.` ?

go-controller/pkg/kubevirt/dhcp_test.go
```diff
@@ -46,101 +31,175 @@ var _ = Describe("Kubevirt", func() {
...
+						"server_id":  "0a:58:6d:6d:c1:50",
+						"fqdn":       `"foo1"`,
+						"dns_server": "2001:1:2:3:4:5:6:7",
+					},
+				},
 			},
 		}),
-		Entry("Dual stack and k8s dns with ipv4 only", dhcpTest{
+
+		Entry("Dual stack and dns with ipv4 only", dhcpTest{
```
maiqueb, line 142: Is this test about a dual-stack cluster with only IPv4 DNS ?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -9,8 +9,6 @@ import (
 	ktypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 
-	kubevirtv1 "kubevirt.io/api/core/v1"
-
```
maiqueb, line 13: please put this back the way it was

go-controller/pkg/kubevirt/pod.go
```diff
@@ -343,3 +342,10 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
 	// to a node
 	return hostSubnets, !util.IsContainedInAnyCIDR(annotation.IPs[0], hostSubnets...), nil
 }
+
+// ExtractVMNameFromPod retunes namespace and name of vm backed up but the pod
```
maiqueb, line 346: wrong comment

go-controller/pkg/kubevirt/pod.go
```diff
@@ -343,3 +342,10 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
...
+// for regular pods return nil
+func VirtualMachineOwnsPod(pod *corev1.Pod) bool {
```
maiqueb, line 348: I think we should prefix the fn name with `is` or `does`, to hint the fn returns a bool.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -836,6 +841,36 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+func (bsnc *BaseSecondaryNetworkController) ensureDHCP(pod *corev1.Pod, podAnnotation *util.PodAnnotation, lsp *nbdb.LogicalSwitchPort) error {
+
+	// Configure DHCP only for kubevirt VMs layer2 primary udn with subnets
+	if !kubevirt.VirtualMachineOwnsPod(pod) ||
+		!util.IsNetworkSegmentationSupportEnabled() ||
+		!bsnc.IsPrimaryNetwork() ||
+		bsnc.TopologyType() != types.Layer2Topology ||
+		len(bsnc.Subnets()) == 0 {
+		return nil
+	}
```
maiqueb, line 853: I think this belongs in the caller. For a regular pod, I don't have to ensure DHCP flows at all. 

If you want this in the helper, I think we should rename the method. 

Also, please consider adding a helper for this fn. IIUC all it needs is the `netInfo`, along with the pod.

go-controller/pkg/ovn/base_network_controller_secondary.go
maiqueb, line 1: I assume we don't have unit tests for this file ...

qinqon:
Looks like alpine containers triggers ND RS so we get a duplicated default gw 
```
default via 2014:100:200::1 dev ovn-udn1 metric 1024 pref medium
default via fe80::858:64ff:fe41:3 dev ovn-udn1 proto ra metric 1024 expires 65438sec hoplimit 255 pref medium
```

test/e2e/kubevirt.go
```diff
@@ -5,15 +5,19 @@ import (
...
+	"net/netip"
 	"os"
 	"os/exec"
 	"strings"
 	"sync"
 	"time"
 
+	iputils "github.com/containernetworking/plugins/pkg/ip"
 	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
```
maiqueb, line 17: this is wrong, you're already importing it in the line below.

test/e2e/kubevirt.go
```diff
@@ -303,8 +312,8 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
 		checkPodRunningReady = func() func(Gomega, *corev1.Pod) {
 			return func(g Gomega, pod *corev1.Pod) {
 				ok, err := testutils.PodRunningReady(pod)
-				Expect(err).ToNot(HaveOccurred())
-				Expect(ok).To(BeTrue())
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(ok).To(BeTrue())
```
maiqueb, line 316: these changes aren't needed.

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1429,145 @@ passwd:
...
+				networkAttachmentConfigParams{
+					namespace: namespace,
+					name:      "net1",
+					topology:  "layer2",
+					cidr:      correctCIDRFamily(cidrIPv4, cidrIPv6),
+					role:      "primary",
+				})
+			By("Creating NetworkAttachmentDefinition")
+			nad := generateNAD(netConfig)
+			Expect(crClient.Create(context.Background(), nad)).To(Succeed())
```
maiqueb, line 1463: nit: why don't you inline these calls ?

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1429,145 @@ passwd:
...
+						Image: "quay.io/nmstate/c10s-nmstate-dev:latest",
+						SecurityContext: &corev1.SecurityContext{
+							Privileged: ptr.To(true),
+							Capabilities: &corev1.Capabilities{
+								Add: []corev1.Capability{"NET_ADMIN"},
+							},
+						},
+					}},
+				},
+			}
```
maiqueb, line 1485: nit: please consider moving this into an helper.

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1429,145 @@ passwd:
...
+			waitForPodsCondition([]*corev1.Pod{kubevirtPod}, func(g gomega.Gomega, pod *corev1.Pod) {
+				ok, err := testutils.PodRunningReady(pod)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(ok).To(BeTrue())
+
+				primaryUDNNetworkStatuses, err := podNetworkStatus(pod, func(networkStatus nadapi.NetworkStatus) bool {
+					return networkStatus.Name == "ovn-kubernetes" && networkStatus.Default && networkStatus.Interface == "ovn-udn1"
+				})
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(primaryUDNNetworkStatuses).To(HaveLen(1))
```
maiqueb, line 1499: for this check to make sense, imho, you should not filter above by interface name, just by the `Default` flag.

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1429,145 @@ passwd:
...
+				return err
+			}).
+				WithTimeout(5 * time.Second).
+				WithPolling(time.Second).
+				Should(Succeed())
+
+			By("Reconfigure primary UDN interface to use dhcp/nd for ipv4 and ipv6")
+			_, err := virtLauncherCommand(`
+nmcli c mod ovn-udn1 ipv4.addresses "" ipv6.addresses "" ipv4.gateway "" ipv6.gateway "" ipv6.method auto ipv4.method auto ipv6.addr-gen-mode eui64
+nmcli d reapply ovn-udn1`)
```
maiqueb, line 1515: maybe bury these details in a properly named helper ?

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1429,145 @@ passwd:
...
+			Expect(err).ToNot(HaveOccurred())
+
+		})
+		It("should configuration IPv4 and IPv6 using DHCP and NDP", func() {
```
maiqueb, line 1519:
```suggestion
		It("should configure IPv4 and IPv6 using DHCP and NDP", func() {
```

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1429,145 @@ passwd:
...
+			if isIPv4Supported() {
+				expectedIP, err := util.MatchIPStringFamily(false /*ipv4*/, primaryUDNNetworkStatus.IPs)
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedDNS, err := util.MatchIPStringFamily(false /*ipv4*/, dnsService.Spec.ClusterIPs)
+				Expect(err).ToNot(HaveOccurred())
+
+				cidrIP, _, err := net.ParseCIDR(cidrIPv4)
+				Expect(err).ToNot(HaveOccurred())
+				expectedGatway := iputils.NextIP(cidrIP).String()
```
maiqueb, line 1534: please export to an helper. Or just use https://github.com/ovn-org/ovn-kubernetes/blob/ef04a67650d3a7048de3525512ba514cdfb31ff0/go-controller/pkg/util/net.go#L38

maiqueb (changes requested):
Thank you. The e2e test you're adding looks really good and useful.

go-controller/pkg/ovn/gateway.go
```diff
@@ -386,6 +386,11 @@ func (gw *GatewayManager) GatewayInit(
 			types.NetworkExternalID:  gw.netInfo.GetNetworkName(),
 			types.TopologyExternalID: gw.netInfo.TopologyType(),
 		}
+		if gw.netInfo.IsPrimaryNetwork() && gw.netInfo.TopologyType() == types.Layer2Topology {
```
maiqueb, line 389: and this should only be used for KubeVirt VMs AFAICT.

go-controller/pkg/kubevirt/dhcp.go
qinqon: > this is only for hypershift right ? For primary UDN, we would put here the GW, wouldn't we ?

Yep, this is just a wrapper for `EnsureDHCPOptionsForMigratablePod` that is only called for the hypershift case.

For layer2 primary UDN is called directly at the secondaries controller.

go-controller/pkg/kubevirt/dhcp_test.go
qinqon: Yep, If I remember correctly having ":" at external_ids was problematic.

test/e2e/kubevirt.go
qinqon: They are, in fact is a bug we have at the test suite, when we use this functionat at Eventually it will not retry if it fails (it did happend to me).

maiqueb: I honestly don't understand how these functions that receive gomega work / what problem that solves / why isn't this code simpler. 

Using expect inside an eventually is simply something we should stay away from. Honestly, all expects inside functions IMHO.

qinqon: It's all related to this 

```
Eventually also supports a special case polling function that takes a single Gomega argument and returns no values. Eventually assumes such a function is making assertions and is designed to work with the Succeed matcher to validate that all assertions have passed.

For example:

Eventually(func(g Gomega) {
   model, err := client.Find(1138)
   g.Expect(err).NotTo(HaveOccurred())
   g.Expect(model.Reticulate()).To(Succeed())
   g.Expect(model.IsReticulated()).To(BeTrue())
   g.Expect(model.Save()).To(Succeed())
}).Should(Succeed())
will rerun the function until all assertions pass.
```

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -836,6 +841,36 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+	if err != nil {
+		return err
+	}
+
+	ipv4Gateway, _ := util.MatchFirstIPFamily(false /*ipv4*/, podAnnotation.Gateways)
+	if ipv4Gateway != nil {
+		opts = append(opts, kubevirt.WithIPv4Router(ipv4Gateway.String()))
+	}
+
+	//TODO classless routes
```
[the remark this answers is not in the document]
qinqon: I needed to ask maintainers about it, since current routes are redundant, the default route is enough, but looks like after services is implemented we may need them.

About implementation for IPv4 we have 

```
options : classless_static_route: optional string
              The DHCPv4 option code for this option is 121.

              This option can contain one or more static routes, each of
              which consists of a destination descriptor and the IP
              address of the router that should be used to reach that
              destination. Please see RFC 3442 for more details.

              Example: {30.0.0.0/24,10.0.0.10, 0.0.0.0/0,10.0.0.1}
```

I have to check what we do for IPv6.

go-controller/pkg/kubevirt/dhcp.go
qinqon: > your commit msg omits you're also setting the hostname for ipv6.

Sure let me add it to the commit message

> 
> Maybe consider exporting this to a separate commit ? Also, shouldn't it be an option ? Or are we after ipv4 parity (which always sets the hostname).

Idea is to have IPv4 parity, so singlestack IPv6 works as expected, there is no special option because hypershift and primary UDN should do the same here, don't want to add complexity if not needed.

go-controller/pkg/kubevirt/dhcp_test.go
qinqon: Yes, renaming the test to  `Dual stack with single IPv4 dns server`

go-controller/pkg/kubevirt/pod.go
qinqon: Renaming to `IsPodOwnByVirtualMachine`

go-controller/pkg/ovn/base_network_controller_secondary.go
qinqon: we do, I can try to put some unit test if you want.

test/e2e/kubevirt.go
qinqon: If something breaks in multus and we go back having eth0 as default or multiple defaults we may have a false possitive here, that's why the extra check is added here, at the end this is about primary UDN.

go-controller/pkg/ovn/gateway.go
```diff
@@ -386,6 +386,11 @@ func (gw *GatewayManager) GatewayInit(
...
+			logicalRouterPort.Ipv6RaConfigs = map[string]string{
+				"address_mode": "dhcpv6_stateful",
```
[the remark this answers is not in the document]
qinqon: Is guest going to send RS after flush ? if this is the case we don't need to send periodically.

go-controller/pkg/ovn/gateway.go
```diff
@@ -386,6 +386,11 @@ func (gw *GatewayManager) GatewayInit(
 			types.NetworkExternalID:  gw.netInfo.GetNetworkName(),
 			types.TopologyExternalID: gw.netInfo.TopologyType(),
 		}
+		if gw.netInfo.IsPrimaryNetwork() && gw.netInfo.TopologyType() == types.Layer2Topology {
+			logicalRouterPort.Ipv6RaConfigs = map[string]string{
```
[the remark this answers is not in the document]
qinqon: We are also missing it for for IPv4 with DHCP_Options

go-controller/pkg/ovn/base_network_controller_secondary.go
maiqueb: Sure, let's at least try :)

test/e2e/kubevirt.go
maiqueb: I can't prevent you from using a gomega feature.

I prefer the simple alternative of having the fn return an error, and that it eventually / consistently succeeds - but that's opinionated.

Thanks for explaining to me this new feature.

test/e2e/kubevirt.go
maiqueb: AFAIU, for your argument to make sense, two things need to break in multus / status library:
- multiple default interfaces reported as default
- multiple interfaces reported with the same interface name

I'm simply saying the second criteria is safe to assume, since it would also mean the CNIs are returning the same interface twice in the CNI result.

go-controller/pkg/ovn/gateway.go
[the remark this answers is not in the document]
maiqueb: > Is guest going to send RS after flush ?

No.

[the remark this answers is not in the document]
qinqon: Just as side not, we don't need to configure min_interval since ovn calculate it as a third of max_interval.

qinqon:
e2e failing on connectivity when mtu is changed, the nad has 1300 but primary UDN link is 1500

```yaml
apiVersion: v1
items:
- apiVersion: k8s.cni.cncf.io/v1
  metadata:
    creationTimestamp: "2024-10-04T07:40:38Z"
    generation: 1
    name: gryffindor
    namespace: network-segmentation-6335
    resourceVersion: "4780"
    uid: f130145d-1df1-4ee3-bda9-ce98e3abec3f
  spec:
    config: "\n{\n        \"cniVersion\": \"0.3.0\",\n        \"name\": \"5nnjq_gryffindor\",\n
      \       \"type\": \"ovn-k8s-cni-overlay\",\n        \"topology\":\"layer2\",\n
      \       \"subnets\": \"10.128.0.0/16,2014:100:200::0/60\",\n        \"excludeSubnets\":
      \"\",\n        \"mtu\": 1300,\n        \"netAttachDefName\": \"network-segmentation-6335/gryffindor\",\n
      \       \"vlanID\": 0,\n        \"allowPersistentIPs\": false,\n        \"role\":
      \"primary\",\n\t\t\"mtu\": 0\n}\n"
kind: List
metadata:
  resourceVersion: ""

```

```bash
3: ovn-udn1@if35: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
    link/ether 0a:58:0a:80:00:03 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

maiqueb:
> on alpine image at e2e failing on connectivity when mtu is changed, the nad has 1300 but primary UDN link is 1500
> 
> ```yaml
> apiVersion: v1
> items:
> - apiVersion: k8s.cni.cncf.io/v1
>   kind: NetworkAttachmentDefinition
>   metadata:
>     creationTimestamp: "2024-10-04T07:40:38Z"
>     generation: 1
>     name: gryffindor
>     namespace: network-segmentation-6335
>     resourceVersion: "4780"
>     uid: f130145d-1df1-4ee3-bda9-ce98e3abec3f
>   spec:
>     config: "\n{\n        \"cniVersion\": \"0.3.0\",\n        \"name\": \"5nnjq_gryffindor\",\n
>       \       \"type\": \"ovn-k8s-cni-overlay\",\n        \"topology\":\"layer2\",\n
>       \       \"subnets\": \"10.128.0.0/16,2014:100:200::0/60\",\n        \"excludeSubnets\":
>       \"\",\n        \"mtu\": 1300,\n        \"netAttachDefName\": \"network-segmentation-6335/gryffindor\",\n
>       \       \"vlanID\": 0,\n        \"allowPersistentIPs\": false,\n        \"role\":
>       \"primary\",\n\t\t\"mtu\": 0\n}\n"
> kind: List
> metadata:
>   resourceVersion: ""
> ```
> 
> ```shell
> 3: ovn-udn1@if35: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
>     link/ether 0a:58:0a:80:00:03 brd ff:ff:ff:ff:ff:ff link-netnsid 0
> ```
> 
> UPDATE: the nad has "mtu" field twice, that's a deffect on the PR's e2e changes.

I don't remember the outcome of what we discussed regarding MTU on primary UDN. 

Did we want it to respect the NAD ?... We did right ? If omitted, it would default to the cluster default network configuration, which takes into account the physical infra.

qinqon:
> > on alpine image at e2e failing on connectivity when mtu is changed, the nad has 1300 but primary UDN link is 1500
> > ```yaml
> > apiVersion: v1
> > items:
> > - apiVersion: k8s.cni.cncf.io/v1
> >   kind: NetworkAttachmentDefinition
> >   metadata:
> >     creationTimestamp: "2024-10-04T07:40:38Z"
> >     generation: 1
> >     name: gryffindor
> >     namespace: network-segmentation-6335
> >     resourceVersion: "4780"
> >     uid: f130145d-1df1-4ee3-bda9-ce98e3abec3f
> >   spec:
> >     config: "\n{\n        \"cniVersion\": \"0.3.0\",\n        \"name\": \"5nnjq_gryffindor\",\n
> >       \       \"type\": \"ovn-k8s-cni-overlay\",\n        \"topology\":\"layer2\",\n
> >       \       \"subnets\": \"10.128.0.0/16,2014:100:200::0/60\",\n        \"excludeSubnets\":
> >       \"\",\n        \"mtu\": 1300,\n        \"netAttachDefName\": \"network-segmentation-6335/gryffindor\",\n
> >       \       \"vlanID\": 0,\n        \"allowPersistentIPs\": false,\n        \"role\":
> >       \"primary\",\n\t\t\"mtu\": 0\n}\n"
> > kind: List
> > metadata:
> >   resourceVersion: ""
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```shell
> > 3: ovn-udn1@if35: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default 
> >     link/ether 0a:58:0a:80:00:03 brd ff:ff:ff:ff:ff:ff link-netnsid 0
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > UPDATE: the nad has "mtu" field twice, that's a deffect on the PR's e2e changes.
> 
> I don't remember the outcome of what we discussed regarding MTU on primary UDN.
> 
> Did we want it to respect the NAD ?... We did right ? If omitted, it would default to the cluster default network configuration, which takes into account the physical infra.

Well we should do at VMs whatever is done for pod and since we are respecting NAD's mtu we should configure it too with DHCPv4 and RAs.

go-controller/pkg/ovn/base_network_controller_secondary.go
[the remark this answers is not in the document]
qinqon: @maiqueb after some discussion with maintainers is ok to no implement classless routes, we may need to re-visit this for L2 secondary interfaces if ingress is implemented there using k8s services.

test/e2e/kubevirt.go
qinqon: Ok, will filter just by default flag.

tssurya:
@maiqueb and @qinqon this PR is also using the LSP up/down logic for live migration aspects right? I don't see us mentioning this in the PR description, this seems totally around DHCP options 
But last I checked with @maiqueb we are doing it here right?

@kyrtapz could you start reviews for this one please - whenever you get some time

maiqueb:
> @maiqueb and @qinqon this PR is also using the LSP up/down logic for live migration aspects right? I don't see us mentioning this in the PR description, this seems totally around DHCP options But last I checked with @maiqueb we are doing it here right?

No @tssurya , that'll happen in a separate PR. 

> 
> @kyrtapz could you start reviews for this one please - whenever you get some time

kyrtapz (commented):
The changes look good to me, just a few NITs.
Thanks for the awesome E2Es!

During an offline discussion we agreed to add some docs for the LLA limitation we have so we don't forget it in a week.

As for the two default routes, I think we can set the RA route priority to low so the one we set in the CNI takes priority:
```
       ipv6_ra_configs : router_preference: optional string
              Default Router Preference (PRF) indicates whether to
              prefer this router over other default routers (RFC 4191).
              Possible values are:
              •      HIGH: mapped to 0x01 in RA PRF field
              •      MEDIUM: mapped to 0x00 in RA PRF field
              •      LOW: mapped to 0x11 in RA PRF field
```

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -836,6 +843,37 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+	opts = append(opts, kubevirt.WithIPv4DNServer(ipv4DNSerer), kubevirt.WithIPv6DNServer(ipv6DNServer))
+
+	return kubevirt.EnsureDHCPOptionsForLSP(bsnc.controllerName, bsnc.nbClient, pod, podAnnotation.IPs, lsp, opts...)
+}
+
 func getMasqueradeManagementIPSNATMatch(dstMac string) string {
 	return fmt.Sprintf("eth.dst == %s", dstMac)
 }
+
+func (bsnc *BaseSecondaryNetworkController) isPodVirtualMachineWithPrimaryUDN(pod *corev1.Pod) bool {
```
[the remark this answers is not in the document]
qinqon: Let's do that, I am ok with that too.

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -22,17 +22,71 @@ const (
...
+		if configs.V4 == nil {
+			return
+		}
+		configs.V4.Options["dns_server"] = dnsServer
+	}
+}
+
+func WithIPv6DNServer(dnsServer string) func(*dhcpConfigs) {
+	return func(configs *dhcpConfigs) {
+		if dnsServer == "" {
```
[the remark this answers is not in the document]
qinqon: is very common to have dual stack but IPv4 DNS server since the DNS protocol is able to answer names for IPv6 over ipv4, A record versus AAA record.

don't know if we want to be prepare for single stack IPV6 though.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -343,3 +343,9 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
 	// to a node
 	return hostSubnets, !util.IsContainedInAnyCIDR(annotation.IPs[0], hostSubnets...), nil
 }
+
+// IsPodOwnByVirtualMachine returns true if the pod is own by a
+// kubevirt virtual machine, false otherwise.
+func IsPodOwnByVirtualMachine(pod *corev1.Pod) bool {
```
maiqueb, line 349:
```suggestion
// IsPodOwnedByVirtualMachine returns true if the pod is owned by a
// kubevirt virtual machine, false otherwise.
func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
```

maiqueb (changes requested):
Getting there; most I have are opinionated nits.

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1426,131 @@ passwd:
 			}),
 		)
 	})
+	Context("with kubevirt VM using layer2 UDPN", func() {
+		var (
+			cidrIPv4                = "10.128.0.0/24"
+			cidrIPv6                = "2010:100:200::/60"
+			primaryUDNNetworkStatus nadapi.NetworkStatus
+			virtLauncherCommand     = func(command string) (string, error) {
+				stdout, stderr, err := ExecShellInPodWithFullOutput(fr, namespace, "virt-launcher-vm1", command)
```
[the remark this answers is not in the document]
qinqon: Don't want to polute the test description to it since we don't care, let me move it to a const.

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1426,131 @@ passwd:
...
+
+		})
+		It("should configure IPv4 and IPv6 using DHCP and NDP", func() {
+
+			dnsService, err := fr.ClientSet.CoreV1().Services(config.Kubernetes.DNSServiceNamespace).
+				Get(context.Background(), config.Kubernetes.DNSServiceName, metav1.GetOptions{})
+			Expect(err).ToNot(HaveOccurred())
+
+			if isIPv4Supported() {
+				expectedIP, err := util.MatchIPStringFamily(false /*ipv4*/, primaryUDNNetworkStatus.IPs)
```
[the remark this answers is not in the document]
qinqon: What do you think of doing 
```golang
util.MatchIPStringFamily(isIPv6Supported(), primaryUDNNetworkStatus.IPs)
```

test/e2e/multihoming_utils.go
```diff
@@ -83,6 +84,9 @@ func uniqueNadName(originalNetName string) string {
 }
 
 func generateNAD(config networkAttachmentConfig) *nadapi.NetworkAttachmentDefinition {
+	if config.mtu == 0 {
+		config.mtu = 1300
```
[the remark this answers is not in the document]
qinqon: It do not change the behaviour, there isn't any test passing the mtu except this new one since it's new field so for all the test is going to be 1300 as expected and we are also adding the value that you can now configure mtu at tests.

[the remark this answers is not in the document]
maiqueb: Yes, but now you explicitly set the MTU to 1300 in *all* tests. Why can't you explicitly set the MTU in the *only* test that requires it ? 

Let's say for whatever reason the platform changes and the default provided by ovn-k stops being 1300. Then your code has changed the behavior of the tests.

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1426,131 @@ passwd:
...
+					WithArguments("connection", "DHCP6.OPTION").
+					WithTimeout(10 * time.Second).
+					WithPolling(time.Second).
+					Should(ContainElements(
+						"fqdn_fqdn = vm1",
+						fmt.Sprintf("ip6_address = %s", expectedIP),
+					))
+				Expect(primaryUDNValueForConnection("IP6.ADDRESS")).To(SatisfyAll(HaveLen(2), ContainElements(expectedIP+"/128")))
+				Expect(primaryUDNValueForConnection("IP6.GATEWAY")).To(ConsistOf(WithTransform(func(ipv6 string) bool {
+					return netip.MustParseAddr(ipv6).IsLinkLocalUnicast()
```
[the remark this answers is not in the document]
qinqon: It's quite involve since it means we have to read ovn-nbctl to get LRP's mac and calculate the LinkLocal address, don't think is worth it, we are not going to test OVN IPv6 capabilities.

Maybe follow up

test/e2e/kubevirt.go
```diff
@@ -1420,4 +1426,131 @@ passwd:
...
+				Expect(err).ToNot(HaveOccurred())
+				Eventually(primaryUDNValueFor).
+					WithArguments("connection", "DHCP6.OPTION").
+					WithTimeout(10 * time.Second).
+					WithPolling(time.Second).
+					Should(ContainElements(
+						"fqdn_fqdn = vm1",
+						fmt.Sprintf("ip6_address = %s", expectedIP),
+					))
+				Expect(primaryUDNValueForConnection("IP6.ADDRESS")).To(SatisfyAll(HaveLen(2), ContainElements(expectedIP+"/128")))
```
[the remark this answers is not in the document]
qinqon: We don't care.

test/e2e/kubevirt.go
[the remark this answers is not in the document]
maiqueb: I had thought it would be possible to extract it from either the node's annotations, or compute it from the config. 

Yeah, an hypothetical follow-up is good enough for me.

test/e2e/multihoming_utils.go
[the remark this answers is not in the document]
qinqon: > Yes, but now you explicitly set the MTU to 1300 in _all_ tests. Why can't you explicitly set the MTU in the _only_ test that requires it ?
> 

It was 1300 explicity before, so I think it's the same
```diff
-        "mtu": 1300,
```

> Let's say for whatever reason the platform changes and the default provided by ovn-k stops being 1300. Then your code has changed the behavior of the tests.

As said the original version was explicitly seting it to 1300 is not something at ovn-k is at this very function.

test/e2e/kubevirt.go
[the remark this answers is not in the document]
maiqueb: it's better, still I think you could still wrap it over. It would look better.

test/e2e/multihoming_utils.go
```diff
@@ -92,11 +96,11 @@ func generateNAD(config networkAttachmentConfig) *nadapi.NetworkAttachmentDefini
         "topology":%q,
         "subnets": %q,
         "excludeSubnets": %q,
-        "mtu": 1300,
         "netAttachDefName": %q,
         "vlanID": %d,
         "allowPersistentIPs": %t,
-        "role": %q
+        "role": %q,
+        "mtu": %d
```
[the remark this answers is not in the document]
qinqon: This is what the change here is doing.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -826,6 +833,37 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+	ipv4Gateway, _ := util.MatchFirstIPFamily(false /*ipv4*/, podAnnotation.Gateways)
+	if ipv4Gateway != nil {
+		opts = append(opts, kubevirt.WithIPv4Router(ipv4Gateway.String()))
+	}
+
+	if bsnc.MTU() > 0 {
+		opts = append(opts, kubevirt.WithIPv4MTU(bsnc.MTU()))
+	}
+
+	opts = append(opts, kubevirt.WithIPv4DNSServer(ipv4DNSServer), kubevirt.WithIPv6DNSServer(ipv6DNSServer))
```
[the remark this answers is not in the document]
qinqon: Is as good as this, the only thing I see this way is that all the options are included with "append" so it's more clear, anyhow I don't think is worth it to change it.

go-controller/pkg/ovn/gateway.go
```diff
@@ -386,6 +386,17 @@ func (gw *GatewayManager) GatewayInit(
 			types.NetworkExternalID:  gw.netInfo.GetNetworkName(),
 			types.TopologyExternalID: gw.netInfo.TopologyType(),
 		}
+		if gw.netInfo.IsPrimaryNetwork() && gw.netInfo.TopologyType() == types.Layer2Topology {
```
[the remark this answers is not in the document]
qinqon: Discussin for follow up.

test/e2e/network_segmentation.go
```diff
@@ -354,19 +354,24 @@ var _ = Describe("Network Segmentation", func() {
...
-							Expect(err).NotTo(HaveOccurred())
+
+							// Remove last new line to propertly split them using new line
+							defRoutes = strings.TrimSuffix(defRoutes, "\n")
+
+							framework.Logf("Found default routes %v, deleting", defRoutes)
+							for _ = range strings.Split(defRoutes, "\n") {
+								cmd := append(ipCommand, "route", "del")
+								_, err = e2ekubectl.RunKubectl(udnPod.Namespace,
+									append(cmd, "default")...)
```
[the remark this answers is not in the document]
qinqon: > OK, so IIUC, here we just delete the N default routes on the pod; I was confused since you aren't using the route data to delete it.
> 
> While it's exactly the same, maybe it's more readable ?... Not a must, it's pretty much opinionated.

It simplify the test since we don't have to retrieve the route we just care for number of routes.

> 
> Finally, maybe encapsulating this in an helper would be good.

I don't want to refactor this test since it's not new, I am just fixing it.

test/e2e/multihoming_utils.go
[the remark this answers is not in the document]
maiqueb: no, it's not. You delete the existing line (95), and add a new one in a different position (103).

I'm saying to replace the line 95 with the contents you now have on line 103. It renders quite differently on the diff - even if the contents are the same.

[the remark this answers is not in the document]
qinqon: oook, doing that, thanks.

maiqueb (approved):
Thank you for your patience. 

Nice virt-aware features !!! :rocket:

qinqon:
I see some unexpected test failing at net-seg lane sometimes, don't know if it's related yet.

kyrtapz (dismissed):
Thank you @qinqon!

maiqueb (approved):
Putting my approval back.

qinqon:
known flake:
- https://github.com/ovn-org/ovn-kubernetes/issues/4144

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -88,14 +140,12 @@ func retrieveDNSServiceClusterIPs(k8scli *factory.WatchFactory) (string, string,
 	return clusterIPv4, clusterIPv6, nil
 }
 
-func ComposeDHCPv4Options(cidr, dnsServer, controllerName string, vmKey ktypes.NamespacedName) *nbdb.DHCPOptions {
+func ComposeDHCPv4Options(cidr, controllerName string, vmKey ktypes.NamespacedName) *nbdb.DHCPOptions {
 	serverMAC := util.IPAddrToHWAddr(net.ParseIP(ARPProxyIPv4)).String()
 	dhcpOptions := &nbdb.DHCPOptions{
 		Cidr: cidr,
 		Options: map[string]string{
 			"lease_time": fmt.Sprintf("%d", dhcpLeaseTime),
```
tssurya, line 148: not entirely related to this PR, but is there any reason for us to make the lease expire every hour in OVN-Kubernetes for VM Pods? I would assume the podIP, gateway, dnsserver all stay the same for the life of the VM pod no?

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -104,17 +154,15 @@ func ComposeDHCPv4Options(cidr, dnsServer, controllerName string, vmKey ktypes.N
...
 }
 
-func ComposeDHCPv6Options(cidr, dnsServer, controllerName string, vmKey ktypes.NamespacedName) *nbdb.DHCPOptions {
+func ComposeDHCPv6Options(cidr, controllerName string, vmKey ktypes.NamespacedName) *nbdb.DHCPOptions {
 	serverMAC := util.IPAddrToHWAddr(net.ParseIP(ARPProxyIPv6)).String()
 	dhcpOptions := &nbdb.DHCPOptions{
 		Cidr: cidr,
 		Options: map[string]string{
 			"server_id": serverMAC,
+			"fqdn":      fmt.Sprintf("%q", vmKey.Name),
```
tssurya, line 163: nit: a comment on the fact that fqdn is the v6 equivalent of hostname in v4

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -22,17 +22,71 @@ const (
...
+}
+
 func EnsureDHCPOptionsForMigratablePod(controllerName string, nbClient libovsdbclient.Client, watchFactory *factory.WatchFactory, pod *corev1.Pod, ips []*net.IPNet, lsp *nbdb.LogicalSwitchPort) error {
+	dnsServerIPv4, dnsServerIPv6, err := retrieveDNSServiceClusterIPs(watchFactory)
+	if err != nil {
+		return fmt.Errorf("failed retrieving dns service cluster ip: %v", err)
+	}
+
+	return EnsureDHCPOptionsForLSP(controllerName, nbClient, pod, ips, lsp,
+		WithIPv4Router(ARPProxyIPv4),
```
tssurya, line 78: why is it that we don't set MTU option here - its not needed for l3 live migration scenario ?

go-controller/pkg/ovn/gateway.go
tssurya, line 389: @kyrtapz / @maiqueb / @qinqon : Should we do this only when at least 1 VM pod is present in this L2 UDN network? Meaning do an "update to LRP" as day2... I am not sure if we should always be enabling this for all users cause we only set DHCP options for LSPs of VM pods, so unsure if sending periodic RAs for non-VM pods is needed? Or is there an existing virt feature gate we can leverage?

go-controller/pkg/ovn/gateway.go
```diff
@@ -390,6 +390,9 @@ func (gw *GatewayManager) GatewayInit(
 			logicalRouterPort.Ipv6RaConfigs = map[string]string{
 				"address_mode": "dhcpv6_stateful",
 			}
+			if gw.netInfo.MTU() > 0 {
+				logicalRouterPort.Ipv6RaConfigs["mtu"] = fmt.Sprintf("%d", gw.netInfo.MTU())
```
tssurya, line 398: hmm if user is not providing the value the default value should be provided right? so if MTU is 0 we should use config.Default.MTU?
self-answered:
netconf := &ovncnitypes.NetConf{MTU: Default.MTU}
it seems to be taken care of in the config section already..

go-controller/pkg/ovn/gateway.go
```diff
@@ -388,7 +388,10 @@ func (gw *GatewayManager) GatewayInit(
 		}
 		if gw.netInfo.IsPrimaryNetwork() && gw.netInfo.TopologyType() == types.Layer2Topology {
 			logicalRouterPort.Ipv6RaConfigs = map[string]string{
-				"address_mode": "dhcpv6_stateful",
+				"address_mode":  "dhcpv6_stateful",
+				"send_periodic": "true",
+				"max_interval":  "900", // 15 minutes
+				"min_interval":  "300", // 5 minutes
```
tssurya: what's the point of setting both options here? min and max? it seems 5mins is going to win here correct?

go-controller/pkg/ovn/gateway.go
```diff
@@ -388,10 +388,11 @@ func (gw *GatewayManager) GatewayInit(
...
 			logicalRouterPort.Ipv6RaConfigs = map[string]string{
-				"address_mode":  "dhcpv6_stateful",
-				"send_periodic": "true",
-				"max_interval":  "900", // 15 minutes
-				"min_interval":  "300", // 5 minutes
+				"address_mode":      "dhcpv6_stateful",
+				"send_periodic":     "true",
+				"max_interval":      "900", // 15 minutes
+				"min_interval":      "300", // 5 minutes
+				"router_preference": "LOW",
```
tssurya, line 395: nit; please add comment for why "LOW" is being set
this is for when the pod ends up with a LLA of join subnet right?
maybe if we go with what we discussed yesterday around using podsubnet IPs on GR for it will make things easier cause then the LLA sent back will be the same

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -837,6 +844,37 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+
+	return kubevirt.EnsureDHCPOptionsForLSP(bsnc.controllerName, bsnc.nbClient, pod, podAnnotation.IPs, lsp, opts...)
```
tssurya, line 866: ok I have 1 more stupid question. with regards to RAs in v6. I can see on the LSP we set ARPProxyIPv6's MAC which is a hardcoded value for VMs as the server_id. Same for v4. will the server_id be the same for L2 UDN pods and we don't care right cause its a dummy thing at this point since OVN is our DHCP server of sorts?

one more thing: for L3 live migration case we seem to set the ARPProxyIPv4 as the default gateway for the pod but in UDPN we will set the GR's LRP addr i.e .1 of pod subnet correct?

tssurya (commented):
I enjoyed reviewing this PR cause I learnt a lot, most of my questions were self-answered, but I have one main point around RA config enablement in LRP (whether we should do it always or only do it for virt scenario; do we need a knob control?) and regarding the timers set for RA, otherwise this lgtm. Thanks to everyone for teaching me new things.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -837,6 +844,37 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
 	return fmt.Sprintf("eth.dst == %s", dstMac)
 }
+
+func (bsnc *BaseSecondaryNetworkController) requireDHCP(pod *corev1.Pod) bool {
+	// Configure DHCP only for kubevirt VMs layer2 primary udn with subnets
+	return kubevirt.IsPodOwnedByVirtualMachine(pod) &&
+		util.IsNetworkSegmentationSupportEnabled() &&
+		bsnc.IsPrimaryNetwork() &&
+		bsnc.TopologyType() == types.Layer2Topology &&
+		len(bsnc.Subnets()) > 0
```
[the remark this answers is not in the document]
qinqon: Right, will remove it, we are not going to be here if there is no subnets.

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -22,17 +22,71 @@ const (
...
 )
 
+type DHCPConfigsOpt = func(*dhcpConfigs)
+
 type dhcpConfigs struct {
 	V4 *nbdb.DHCPOptions
 	V6 *nbdb.DHCPOptions
 }
 
+func WithIPv4Router(router string) func(*dhcpConfigs) {
```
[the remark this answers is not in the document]
qinqon: Is not an option, they decided to not offer ANY route with DHCPv6, we can only offer them with RAs.

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -22,17 +22,71 @@ const (
...
+		configs.V4.Options["router"] = router
+	}
+}
+
+func WithIPv4MTU(mtu int) func(*dhcpConfigs) {
+	return func(configs *dhcpConfigs) {
+		if configs.V4 == nil {
+			return
+		}
+		configs.V4.Options["mtu"] = fmt.Sprintf("%d", mtu)
```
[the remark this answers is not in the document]
qinqon: At IPv6 we are specifying it at RAs, although I think is possible using DHCPv6 too, I see MTU as a NAD configuration so it make sense to do it over RAs.

go-controller/pkg/kubevirt/dhcp.go
qinqon: The option is mandatory, so for hypershift we configured it with an hour, no special reason though, I prefer not to change this if we don't have customer's complains yet.

go-controller/pkg/ovn/base_network_controller_secondary.go
qinqon: > ok I have 1 more stupid question. with regards to RAs in v6. I can see on the LSP we set ARPProxyIPv6's MAC which is a hardcoded value for VMs as the server_id. Same for v4. will the server_id be the same for L2 UDN pods and we don't care right cause its a dummy thing at this point since OVN is our DHCP server of sorts?
> 

We would care only if there are multiple DHCP server, but I don't say users configuring VMs with primary UDN and also configuring a DHCP server.

> one more thing: for L3 live migration case we seem to set the ARPProxyIPv4 as the default gateway for the pod but in UDPN we will set the GR's LRP addr i.e .1 of pod subnet correct? - we hard code a value because we don't want migration to have gateway change?

Yep, that's exactly it, at L3 it was quite bad since not only MAC but gateway IP is different per node, and at ipv4 you cannot the path from a multipath default gw sending a message.

go-controller/pkg/ovn/gateway.go
qinqon: I don't see much use of L2 for pods, not sure if it's worth it, I mean it will affect pods within layer2, I think this can be fixed later on, if users has issue with this at pods. It will mean that there are l2+pods users.

go-controller/pkg/ovn/gateway.go
qinqon: Don't know, this is OVN implementation, maybe they use a random number between the two, is very common to do some kind of jitter on repeated messages, so systems do get hammered.

tssurya: @qinqon what I meant was do we need to set both just because both are available? it seems either or right?

tssurya:
@qinqon : one more question: why do we set RA periodically? why don't we make it on-demand, so when VM asks for it GR can respond..

qinqon:
> @qinqon : one more question: why do we set RA periodically? why don't we make it on-demand, so when VM asks for it GR can respond..

There is no such things at IPv6, RS is only send at link up, after that the router is resposible of configuring the lifetime (after which the host will flush the route) and send it periodically.

After users flush the route they don't send RAs.

go-controller/pkg/ovn/gateway.go
tssurya: There's some random jitter ovn-controller adds:
```
static long long int
ipv6_ra_calc_next_announce(time_t min_interval, time_t max_interval)
{
    long long int min_interval_ms = min_interval * 1000LL;
    long long int max_interval_ms = max_interval * 1000LL;

    return time_msec() + min_interval_ms +
        random_range(max_interval_ms - min_interval_ms);
}
```
thanks @qinqon saw the response :)

tssurya (approved):
LGTM

My questions were answered
Once https://github.com/ovn-org/ovn-kubernetes/pull/4777 we can do a FUP to block RAs from remote LRPs.

go-controller/pkg/ovn/gateway.go
maiqueb: I think long-term, what would make sense it to expose this functionality using yet another knob in the network configuration ...

We should probably refactor this.

PR #4751, opened 2024-09-30
Add hybird overlay pod IPs to the namespace address_set

pliurh:
The IPs of the pods that are on hybrid overlay nodes shall also be added to the namespace address_set.

Fixes https://issues.redhat.com//browse/OCPBUGS-42244, https://issues.redhat.com/browse/OCPBUGS-42605

#### How to verify it

1. Create a hybrid overlay enabled cluster.
2. Create pods on hybrid overlay nodes.
3. check if the pod IPs are added to the namespace address_set.

```release-note

```

### DIFF
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -124,10 +124,10 @@ func (oc *DefaultNetworkController) ensurePod(oldPod, pod *kapi.Pod, addPort boo
 		return nil
 	}
 
-	// skip the pods on no host subnet nodes
+	// Add podIPs on no host subnet Nodes to the namespace address_set
 	switchName := pod.Spec.NodeName
 	if oc.lsManager.IsNonHostSubnetSwitch(switchName) {
-		return nil
+		return oc.ensureRemotePodIP(oldPod, pod, addPort)
 	}
 
 	if oc.isPodScheduledinLocalZone(pod) {
@@ -201,11 +201,7 @@ func (oc *DefaultNetworkController) ensureLocalZonePod(oldPod, pod *kapi.Pod, ad
 	return nil
 }
 
-// ensureRemoteZonePod tries to set up remote zone pod bits required to interconnect it.
-//   - Adds the remote pod ips to the pod namespace address set for network policy and egress gw
-//
-// It returns nil on success and error on failure; failure indicates the pod set up should be retried later.
-func (oc *DefaultNetworkController) ensureRemoteZonePod(oldPod, pod *kapi.Pod, addPort bool) error {
+func (oc *DefaultNetworkController) ensureRemotePodIP(oldPod, pod *kapi.Pod, addPort bool) error {
 	if (addPort || (oldPod != nil && len(pod.Status.PodIPs) != len(oldPod.Status.PodIPs))) && !util.PodWantsHostNetwork(pod) {
 		podIfAddrs, err := util.GetPodCIDRsWithFullMask(pod, oc.NetInfo)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


go-controller/pkg/ovn/ovn.go
```diff
@@ -139,6 +140,23 @@ func (oc *DefaultNetworkController) ensurePod(oldPod, pod *kapi.Pod, addPort boo
 	return oc.ensureRemoteZonePod(oldPod, pod, addPort)
 }
 
+// ensureHybridOverlayPod tries to set up a pod in hybrid overlay nodes. It returns nil on success and error on failure; failure
+// indicates the pod set up should be retried later.
+func (oc *DefaultNetworkController) ensureHybridOverlayPod(oldPod, pod *kapi.Pod, addPort bool) error {
```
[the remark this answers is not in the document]
pliurh: I replaced `ensureHybridOverlayPod` by `ensureHybridOverlayPod`. Could you take another look?

go-controller/pkg/ovn/ovn.go
```diff
@@ -124,18 +124,13 @@ func (oc *DefaultNetworkController) ensurePod(oldPod, pod *kapi.Pod, addPort boo
...
-		return nil
-	}
-
 	if oc.isPodScheduledinLocalZone(pod) {
 		klog.V(5).Infof("Ensuring zone local for Pod %s/%s in node %s", pod.Namespace, pod.Name, pod.Spec.NodeName)
 		return oc.ensureLocalZonePod(oldPod, pod, addPort)
 	}
 
-	klog.V(5).Infof("Ensuring zone remote for Pod %s/%s in node %s", pod.Namespace, pod.Name, pod.Spec.NodeName)
+	// Pods in no host subnet nodes can be seen as remote pods.
```
npinaeva, line 132: ensureRemoteZonePod does more than just add ips to the namespace addr set.
Also, this will add pod IPs for all no hostsubnet nodes, not only during live migration. Not sure what are possible side-effects for that.

go-controller/pkg/ovn/namespace.go
```diff
@@ -320,13 +321,25 @@ func (oc *DefaultNetworkController) getAllHostNamespaceAddresses() []net.IP {
...
 		for _, node := range existingNodes {
+			var hostNetworkIPs []net.IP
 			if config.HybridOverlay.Enabled && util.NoHostSubnet(node) {
-				// skip hybrid overlay nodes
-				continue
-			}
-			hostNetworkIPs, err := oc.getHostNamespaceAddressesForNode(node)
-			if err != nil {
-				klog.Errorf("Error parsing annotation for node %s: %v", node.Name, err)
+				// during SDN live migration, add the SDN node GW IP to the host network address set.
```
npinaeva, line 326: this change is very openshift-specific, not sure if it belongs upstream...

pliurh: Can we have d/s only patch in openshift repo? If yes, I can propose this PR to d/s.

go-controller/pkg/ovn/ovn.go
pliurh: Yes, ensureRemoteZonePod does more things. But it relies on some k8s specific pod annotations. In our case, those pod annotations don't exist for pods running on the non-ovn node. So I think we should be fine. But if you prefer, I can revise the code.

npinaeva:
For the second commit "[Add hybird overlay pod IPs to the namespace address_set](https://github.com/ovn-org/ovn-kubernetes/pull/4751/commits/52f83dde4a0c752719f61f678de87b3ac195be3c)" it makes sense, since `PodSelectorAddressSet` (I believe as inherited from the previous netpol implementation) already adds HO pod IPs to the peers set. 
If we think no one relies on `namespaceSelector` being treated differently from `podSelector` in netpol peers for HO pods, then it is safe and more consistent from the netpol side.
Another note is that egressQoS also uses namespace addr set, so we need to assess whether anyone may be affected by this change (using HO nodes and QoS at the same time).

npinaeva:
@trozet need some info from you to address my comments ^

go-controller/pkg/ovn/namespace.go
trozet: I think this should be a downstream specific patch, and the IP should only be added during SDN migration and removed after. Otherwise we would always do this for hybrid overlay and it might cause inconsistent behavior that would be hard in the future to root cause.

trozet (changes requested):
The 2nd commit seems fine to me and can go in upstream.

pliurh:
@trozet @npinaeva I removed the commit for d/s. PTAL

trozet (approved):

PR #4755, opened 2024-10-03
e2e flake fix: kubevirt

flavio-fernandes:
This PR addresses a flake in KubeVirt E2E tests, improving
the connectivity tests.

It also introduces the use of an environment check for
KUBEVIRT_SKIP_MIGRATE_POST_COPY to avoid failing the
kubevirtv1.MigrationPostCopy mode, which is currently
used in the GitHub workflow.

~~Lastly, it makes it possible to skip tests in deployments
where OVS is installed on the node instead of running as
a pod.~~

Fixes: https://github.com/ovn-org/ovn-kubernetes/issues/4709

### DIFF
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -126,7 +126,7 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
 		sendEcho = func(conn *net.TCPConn) error {
 			strEcho := "Halo"
 
-			if err := conn.SetDeadline(time.Now().Add(2 * time.Second)); err != nil {
+			if err := conn.SetDeadline(time.Now().Add(5 * time.Second)); err != nil {
 				return fmt.Errorf("failed configuring connection deadline: %w", err)
 			}
 			_, err := conn.Write([]byte(strEcho))
@@ -207,20 +207,18 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
 			return endpoints, nil
 		}
 
-		/*
-
-			reconnect = func(conns []*net.TCPConn) error {
-				for i, conn := range conns {
-					conn.Close()
-					conn, err := dial(conn.RemoteAddr().String())
-					if err != nil {
-						return err
-					}
-					conns[i] = conn
+		reconnect = func(conns []*net.TCPConn) error {
+			for i, conn := range conns {
+				conn.Close()
+				conn, err := dial(conn.RemoteAddr().String())
+				if err != nil {
+					return err
 				}
-				return nil
+				conns[i] = conn
 			}
-		*/
+			return nil
+		}
+
 		composeService = func(name, vmName string, port int32) *corev1.Service {
 			ipFamilyPolicy := corev1.IPFamilyPolicyPreferDualStack
 			return &corev1.Service{
@@ -345,10 +343,17 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
 			}
 			err := crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)
 			Expect(err).ToNot(HaveOccurred())
-			polling := 15 * time.Second
-			timeout := time.Minute
+			polling := 6 * time.Second
+			timeout := 2 * time.Minute
 			step := by(vmName, stage+": Check tcp connection is not broken")
-			Eventually(func() error { return sendEchos(endpoints) }).
+			Eventually(func() error {
+				err = sendEchos(endpoints)
+				if err != nil {
+					by(vmName, fmt.Sprintf("%s: Check tcp connection failed: %s", stage, err))
+					_ = reconnect(endpoints)
+				}
+				return err
+			}).
 				WithPolling(polling).
 				WithTimeout(timeout).
 				Should(Succeed(), step)
@@ -878,10 +883,11 @@ passwd:
 			}()
 
 			By("Wait some time for service to settle")
-			time.Sleep(2 * time.Second)
-
-			endpoints, err := dialServiceNodePort(svc)
-			Expect(err).ToNot(HaveOccurred(), step)
+			endpoints := []*net.TCPConn{}
+			Eventually(func() error {
+				endpoints, err = dialServiceNodePort(svc)
+				return err
+			}).WithPolling(time.Second).WithTimeout(20*time.Second).Should(Succeed(), "Should dial service port once service settled")
 
 			checkConnectivityAndNetworkPolicies(vm.Name, endpoints, "before live migration")
 			// Do just one migration that will fail
@@ -1034,6 +1040,9 @@ passwd:
 			if td.mode == kubevirtv1.MigrationPostCopy && os.Getenv("GITHUB_ACTIONS") == "true" {
 				Skip("Post copy live migration not working at github actions")
 			}
+			if td.mode == kubevirtv1.MigrationPostCopy && os.Getenv("KUBEVIRT_SKIP_MIGRATE_POST_COPY") == "true" {
+				Skip("Post copy live migration explicitly skipped")
+			}
 			var (
 				err error
 			)
@@ -1094,13 +1103,15 @@ passwd:
 				By("annotating the VMI with `fail fast`")
 				vmKey := types.NamespacedName{Namespace: namespace, Name: "worker1"}
 				var vmi kubevirtv1.VirtualMachineInstance
+
 				Eventually(func() error {
-					return crClient.Get(context.TODO(), vmKey, &vmi)
+					err = crClient.Get(context.TODO(), vmKey, &vmi)
+					if err == nil {
+						vmi.ObjectMeta.Annotations[kubevirtv1.FuncTestLauncherFailFastAnnotation] = "true"
+						err = crClient.Update(context.TODO(), &vmi)
+					}
+					return err
 				}).WithPolling(time.Second).WithTimeout(time.Minute).Should(Succeed())
-
-				vmi.ObjectMeta.Annotations[kubevirtv1.FuncTestLauncherFailFastAnnotation] = "true"
-
-				Expect(crClient.Update(context.TODO(), &vmi)).To(Succeed())
 			}
 
 			for _, vm := range vms {


flavio-fernandes:
@maiqueb PTAL

flavio-fernandes:
Known Flake: https://github.com/ovn-org/ovn-kubernetes/issues/4432
`FLAKE: External Gateway With Admin Policy Based External Route CRs e2e multiple external gateway validation Should validate ICMP connectivity to multiple external gateways for an ECMP scenario IPV4`

test/e2e/localnet-underlay.go
```diff
@@ -52,7 +52,7 @@ func teardownUnderlay(ovsPods []v1.Pod) error {
 	return nil
 }
 
-func ovsPods(clientSet clientset.Interface) []v1.Pod {
+func ovsPods(clientSet clientset.Interface) (error, []v1.Pod) {
```
maiqueb, line 55: oh, no. 

Please re-use the common pattern of having the error as the second argument.
Everything in gomega is built to use it - for instance, if the signature is `func ovsPods(clientSet clientset.Interface) ([]v1.Pod, error)`  and the function fails when invoking, it will be caught by `Expect(ovsPods(...)).NotTo(BeEmpty())`.

test/e2e/kubevirt.go
```diff
@@ -1267,8 +1278,11 @@ passwd:
 
 			if td.topology == "localnet" {
 				By("setting up the localnet underlay")
-				nodes := ovsPods(clientSet)
-				Expect(nodes).NotTo(BeEmpty())
+				err, nodes := ovsPods(clientSet)
+				Expect(err).To(Succeed())
+				if len(nodes) == 0 {
+					Skip("No ovs pods: topology localnet test skipped")
+				}
```
maiqueb, line 1285: I actually do wonder if we want to skip these tests ... 

Are you doing this so you can e2e test against deployments running ovs in the nodes ? 
If so, I'd say the long term solution would be to tap into the node based OVS instead ...

Can be done in a follow up, of course.

test/e2e/kubevirt.go
```diff
@@ -872,10 +877,11 @@ passwd:
 			}()
 
 			By("Wait some time for service to settle")
-			time.Sleep(2 * time.Second)
```
maiqueb, line 875: you have my +1 just for removing this sleep :grinning:

maiqueb (changes requested):
The only deal-breaker I see is the ovsPods return argument order.

Thanks for this PR.
@qinqon could you take a look as well ?

test/e2e/kubevirt.go
```diff
@@ -1267,8 +1278,11 @@ passwd:
 
 			if td.topology == "localnet" {
 				By("setting up the localnet underlay")
-				nodes := ovsPods(clientSet)
-				Expect(nodes).NotTo(BeEmpty())
+				err, nodes := ovsPods(clientSet)
+				Expect(err).To(Succeed())
```
qinqon, line 1282: Nit: when checking the error instead of the function directlly it reads better like "this error should not ocurr" doing 
```suggestion
				Expect(err).ToNot(HaveOcurred)
```

qinqon (changes requested):
Thanks @flavio-fernandes ! 

Just questions, suggestions and nits, nothing big

test/e2e/kubevirt.go
```diff
@@ -872,10 +877,11 @@ passwd:
...
 			By("Wait some time for service to settle")
-			time.Sleep(2 * time.Second)
-
-			endpoints, err := dialServiceNodePort(svc)
-			Expect(err).ToNot(HaveOccurred(), step)
+			endpoints := []*net.TCPConn{}
+			Eventually(func() bool {
+				endpoints, err = dialServiceNodePort(svc)
+				return err == nil
+			}).WithPolling(time.Second).WithTimeout(20*time.Second).Should(BeTrue(), "Should dial service port once service settled")
```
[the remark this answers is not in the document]
maiqueb: Good catch, I missed that. I would indeed prefer if the fn being retried returned the error directly, instead of a bool.

qinqon:
@flavio-fernandes also some of the flakiness are real errors that should be fixed with:
- https://github.com/ovn-org/ovn-kubernetes/pull/4416

test/e2e/kubevirt.go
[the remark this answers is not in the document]
qinqon: Also I found the other they that we have a way at Ginkgo to pass function arguments at Eventually so we don't need to create a function closure

```golang
			Eventually(dialServiceNodePort).WithArguments(svc).
			WithPolling(time.Second).WithTimeout(20*time.Second).Should(BeTrue(), "Should dial service port once service settled")
```

test/e2e/localnet-underlay.go
flavio-fernandes: ha, absolutely. my bad in not paying attention....

test/e2e/kubevirt.go
flavio-fernandes: yeah, that is it. I would like to address that as a follow up pr. but the long term fix is really to modify the functions to make the ovs changes in the node.

test/e2e/kubevirt.go
```diff
@@ -1028,6 +1034,9 @@ passwd:
 			if td.mode == kubevirtv1.MigrationPostCopy && os.Getenv("GITHUB_ACTIONS") == "true" {
 				Skip("Post copy live migration not working at github actions")
 			}
+			if td.mode == kubevirtv1.MigrationPostCopy && os.Getenv("KUBEVIRT_SKIP_MIGRATE_POST_COPY") == "true" {
```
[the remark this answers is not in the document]
flavio-fernandes: yeah. my change is here was to mirror that check against "GITHUB_ACTIONS" env. I still would like the entries with the other td.mode to run and am not sure how to accomplish that using ginko skip.

test/e2e/kubevirt.go
```diff
@@ -1088,13 +1097,15 @@ passwd:
...
+					if err == nil {
+						vmi.ObjectMeta.Annotations[kubevirtv1.FuncTestLauncherFailFastAnnotation] = "true"
+						err = crClient.Update(context.TODO(), &vmi)
+					}
+					return err
 				}).WithPolling(time.Second).WithTimeout(time.Minute).Should(Succeed())
-
-				vmi.ObjectMeta.Annotations[kubevirtv1.FuncTestLauncherFailFastAnnotation] = "true"
-
-				Expect(crClient.Update(context.TODO(), &vmi)).To(Succeed())
```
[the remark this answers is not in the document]
flavio-fernandes: oh, that is very cool: https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict  yes, that would a great use case for it here.

test/e2e/kubevirt.go
```diff
@@ -339,10 +337,17 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
...
-			timeout := time.Minute
+			polling := 6 * time.Second
+			timeout := 2 * time.Minute
 			step := by(vmName, stage+": Check tcp connection is not broken")
-			Eventually(func() error { return sendEchos(endpoints) }).
+			Eventually(func() error {
+				err = sendEchos(endpoints)
+				if err != nil {
+					by(vmName, fmt.Sprintf("%s: Check tcp connection failed: %s", stage, err))
+					_ = reconnect(endpoints)
```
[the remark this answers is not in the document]
flavio-fernandes: In my experiment, I was speculating that the endpoints were stale and by doing the reconnect we ensure that when sendEchos failed, we retry with a new connection. Which will definitely fail if th reconnect fail, so I did not care much if that failed.

test/e2e/kubevirt.go
```diff
@@ -1088,13 +1097,15 @@ passwd:
...
 				vmKey := types.NamespacedName{Namespace: namespace, Name: "worker1"}
 				var vmi kubevirtv1.VirtualMachineInstance
-				Eventually(func() error {
-					return crClient.Get(context.TODO(), vmKey, &vmi)
-				}).WithPolling(time.Second).WithTimeout(time.Minute).Should(Succeed())
-
-				vmi.ObjectMeta.Annotations[kubevirtv1.FuncTestLauncherFailFastAnnotation] = "true"
 
-				Expect(crClient.Update(context.TODO(), &vmi)).To(Succeed())
+				Expect(retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
flavio-fernandes, line 1101: This did not work with `retry.RetryOnConflict(` HUE HUE...

test/e2e/kubevirt.go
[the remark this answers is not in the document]
flavio-fernandes: oh no, @qinqon the retryOnConflict was not good enough to handle this particular case! Sad panda!
I will restore it to the original change to make it happy again. I tried... :) Thanks for showing this to me, tho.

https://github.com/ovn-org/ovn-kubernetes/actions/runs/11185035426/job/31098147703?pr=4755
```
2024-10-04T19:11:50.0897708Z [0mKubevirt Virtual Machines [38;5;243mwith default pod network [0mwhen live migration [38;5;9m[1m[It] with pre-copy fails, should keep connectivity[0m
2024-10-04T19:11:50.0899676Z [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:1133[0m
2024-10-04T19:11:50.0900653Z 
2024-10-04T19:11:50.0901101Z   [38;5;9m[FAILED] Expected success, but got an error:
2024-10-04T19:11:50.0901611Z       <*errors.StatusError | 0xc000fee5a0>: 
2024-10-04T19:11:50.0902264Z       virtualmachineinstances.kubevirt.io "worker1" not found
2024-10-04T19:11:50.0902780Z       {
2024-10-04T19:11:50.0903027Z           ErrStatus: 
2024-10-04T19:11:50.0903790Z               code: 404
2024-10-04T19:11:50.0904455Z               details:
2024-10-04T19:11:50.0905012Z                 group: kubevirt.io
2024-10-04T19:11:50.0905730Z                 kind: virtualmachineinstances
2024-10-04T19:11:50.0906496Z                 name: worker1
2024-10-04T19:11:50.0907590Z               message: virtualmachineinstances.kubevirt.io "worker1" not found
2024-10-04T19:11:50.0908573Z               metadata: {}
2024-10-04T19:11:50.0908933Z               reason: NotFound
2024-10-04T19:11:50.0909412Z               status: Failure,
2024-10-04T19:11:50.0909967Z       }[0m
2024-10-04T19:11:50.0911182Z   [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:1108[0m [38;5;243m@ 10/04/24 19:11:48.618[0m
```

This is what I did here:
https://github.com/ovn-org/ovn-kubernetes/pull/4755/commits/1ce0db03166f9c8b171864f116f8be255e236d87#r1788274608
```
				Expect(retry.RetryOnConflict(retry.DefaultRetry, func() error {
					err = crClient.Get(context.TODO(), vmKey, &vmi)
					if err == nil {
						vmi.ObjectMeta.Annotations[kubevirtv1.FuncTestLauncherFailFastAnnotation] = "true"
						err = crClient.Update(context.TODO(), &vmi)
					}
					return err
				})).To(Succeed())
```

test/e2e/kubevirt.go
flavio-fernandes: Fixed

trozet:
"Lastly, it makes it possible to skip tests in deployments
where OVS is installed on the node instead of running as
a pod."

@flavio-fernandes why would we want to skip it if OVS is running on the node? Isn't the test case still applicable?

flavio-fernandes:
> "Lastly, it makes it possible to skip tests in deployments where OVS is installed on the node instead of running as a pod."
> 
> @flavio-fernandes why would we want to skip it if OVS is running on the node? Isn't the test case still applicable?

Hi @trozet!

The issue I have is that:
- I would like to run the e2e tests against deployments where OVS is running on the host, not in a pod, and
- The tests, as written now, are not smart enough to make the necessary changes in OVS in that scenario.

However, there are other tests in this e2e suite that do not need to make these changes and work just fine in such a cluster. The long-term solution is to improve this test, but I believe that doing so is beyond the scope of this PR.

Hope this makes sense.

trozet:
> * The tests, as written now, are not smart enough to make the necessary changes in OVS in that scenario.
> 
> However, there are other tests in this e2e suite that do not need to make these changes and work just fine in such a cluster. The long-term solution is to improve this test, but I believe that doing so is beyond the scope of this PR.

@flavio-fernandes I would argue skipping the tests because of the OVS stuff doesn't seem to belong in this PR that is fixing the kubevirt flake either. I think the OVS issue should be easy to fix. If the search for OVS pods yields 0 pods, then assume OVS must be on the host. Then get the ovnkube-node pods. Their containers have access to the host and should be able to delete  bridges or do whatever else needs to be done in OVS. I suggest probably splitting out the OVS stuff into a separate PR.

flavio-fernandes:
> > * The tests, as written now, are not smart enough to make the necessary changes in OVS in that scenario.
> > 
> > However, there are other tests in this e2e suite that do not need to make these changes and work just fine in such a cluster. The long-term solution is to improve this test, but I believe that doing so is beyond the scope of this PR.
> 
> @flavio-fernandes I would argue skipping the tests because of the OVS stuff doesn't seem to belong in this PR that is fixing the kubevirt flake either. I think the OVS issue should be easy to fix. If the search for OVS pods yields 0 pods, then assume OVS must be on the host. Then get the ovnkube-node pods. Their containers have access to the host and should be able to delete bridges or do whatever else needs to be done in OVS. I suggest probably splitting out the OVS stuff into a separate PR.

Sounds reasonable @trozet ! Let me fix this PR to be about the flake, which is what I really intended it to be.

flavio-fernandes:
@trozet PTAL

trozet (approved):

PR #4757, opened 2024-10-04
interconenct: Ask ovn-northd to bind the remote ports.

numansiddique:
Presently for each remote node, ovnkube after creating the remote port in the OVN Northbound db transit switch, binds the port to the remote chassis in the Southbound database.  In order to do this, we have to wait for the remote logical ports (for each of the remote zone nodes) to appear in the OVN Southbound database. This can lead to race conditions.  Instead ask ovn-northd to bind to the remote chassis.  OVN after this commit [1] already supports this.  We just need to set the requested-chassis option in the logical port.

[1] - https://github.com/ovn-org/ovn/commit/170d3e5f19f0fcd760db068cda29fc1b7f1d7ef5

Jira: https://issues.redhat.com/browse/OCPBUGS-42616

Fixes #

```release-note

```

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_secondary.go
+++ b/go-controller/pkg/ovn/base_network_controller_secondary.go
@@ -371,16 +371,6 @@ func (bsnc *BaseSecondaryNetworkController) addLogicalPortToNetworkForNAD(pod *k
 		_ = bsnc.logicalPortCache.add(pod, switchName, nadName, lsp.UUID, podAnnotation.MAC, podAnnotation.IPs)
 	}
 
-	// we need to create the binding ourselves for the remote ports we create on
-	// layer2 topologies with interconnect
-	isRemotePort := !isLocalPod && bsnc.isLayer2Interconnect()
-	if isRemotePort {
-		err := bsnc.zoneICHandler.BindTransitRemotePort(pod.Spec.NodeName, lsp.Name)
-		if err != nil {
-			return fmt.Errorf("failed to bind remote transit port: %w", err)
-		}
-	}
-
 	if isLocalPod {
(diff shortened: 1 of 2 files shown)


trozet (approved):

trozet:
thanks @numansiddique !

hzhou8:
Thanks @numansiddique . Just a quick question, why would we even need to set the requested-chassis option for the remote port? BTW, I have no access to the Jira: https://issues.redhat.com/browse/OCPBUGS-42616

numansiddique:
@hzhou8 In the case ovn OVN-IC,  ovn-ic service of an AZ  (which is it managing,  creates remote ports for other AZs and it binds those remote ports right ?  This is what ovnk was doing for OVN-IC mode (where each k8s node is an AZ and all other nodes are remote zone nodes) before this patch.  i.e  ovnkube running on a node 'A'  creates a remote port  in the OVN Northbound db for all other  k8s nodes in the Transit switch and then it sets the chassis column of the corresponding SB port binding.  

A while back we added the support in ovn-northd to set the chassis column of the Port binding for the remote logical ports if requested-chassis is set.  This commit makes use of this OVN native feature and sets the requested-chassis option in the logical port (of type remote).

In a high scaled deployment we observed that ovnk is missing out setting the chassis column for a couple of remote ports in the SB.  Hence this PR.  We could have debugged that and figured out the root cause.  But instead we are making use of the native OVN feature.

Regarding the jira, unfortunately I don't have the permissions to set the visibility.  

Thanks

hzhou8:
@numansiddique Thanks for explaining. Sorry I misunderstood it earlier. I thought it was setting requested-chassis for security purpose like we used to do, but it is for setting the chassis column, which makes sense :)

PR #4758, opened 2024-10-07
UDN controller: Initialize namespace-tracker on startup

ormergi:

Initialize the namespace-tracker when controller starts
as part of Run().

Populate the namespace-tracker according to the cluster
state.
When the controller actually start it will reconcile CUDN
and NAD objects.

This change enable the controller to cleanup stale NADs
when it starts, recover from a crush and reconciling the
cluster state following changes when controller was down.

Fixes #

Resolve threads on parent PR https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1781228553 https://github.com/ovn-org/ovn-kubernetes/pull/4612#discussion_r1781228553

This PR rebased on top of #4612, **only last commit is relevant**.

```release-note

```


go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -142,6 +143,48 @@ func (c *Controller) Run() error {
...
+	indexedNADs := map[string]netv1.NetworkAttachmentDefinition{}
+	for _, nad := range nads {
+		if nad != nil {
+			indexedNADs[nad.Namespace+"/"+nad.Name] = *nad.DeepCopy()
+		}
+	}
+
+	for _, cudn := range cudns {
+		cudnCopy := cudn.DeepCopy()
+		for nadKey, nad := range indexedNADs {
```
npinaeva, line 172: I think you may not need indexedNADs, and just iterate over `nads` from the `List` call

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -142,6 +143,48 @@ func (c *Controller) Run() error {
...
+	}
+	indexedNADs := map[string]netv1.NetworkAttachmentDefinition{}
+	for _, nad := range nads {
+		if nad != nil {
+			indexedNADs[nad.Namespace+"/"+nad.Name] = *nad.DeepCopy()
+		}
+	}
+
+	for _, cudn := range cudns {
+		cudnCopy := cudn.DeepCopy()
```
npinaeva, line 171: I don't think you need to deepcopy here, as we are not changing cudns in this function?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -142,6 +143,48 @@ func (c *Controller) Run() error {
...
+		for nadKey, nad := range indexedNADs {
+			if !metav1.IsControlledBy(&nad, cudnCopy) {
+				continue
+			}
+			if _, exists := c.namespaceTracker[cudnCopy.Name]; !exists {
+				c.namespaceTracker[cudnCopy.Name] = sets.New[string]()
+			}
```
npinaeva, line 178: just to be consistent with the logic from cudn reconcile, we can add an empty set for every cudn in the outer loop

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -697,6 +698,61 @@ var _ = Describe("User Defined Network Controller", func() {
...
+				connectedNsNames := []string{"green", "yellow"}
+				connectedLabel := map[string]string{"test.io": "connected"}
+				for _, nsName := range connectedNsNames {
+					ns := testNamespace(nsName)
+					ns.SetLabels(connectedLabel)
+					testObjs = append(testObjs, ns)
+				}
+				rand.Shuffle(len(testObjs), func(i, j int) {
+					testObjs[i], testObjs[j] = testObjs[j], testObjs[i]
+				})
```
npinaeva, line 720: interesting, what do we achieve with that?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -697,6 +698,61 @@ var _ = Describe("User Defined Network Controller", func() {
...
+				}}
+				testObjs = append(testObjs, cudn)
+
+				c := newTestController(renderNadStub(testClusterUdnNAD(cudn.Name, "")), testObjs...)
+
+				for _, nsName := range append(staleNADsNsNames, connectedNsNames...) {
+					nad := testClusterUdnNAD("test", nsName)
+					nad, err := cs.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(nsName).Create(context.Background(), nad, metav1.CreateOptions{})
+					Expect(err).NotTo(HaveOccurred())
+				}
```
npinaeva, line 734: why not add these to the testObjs?

npinaeva (commented):
nice!

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Having a map enable optimize this loop in a way that it wont process a NAD that already been tracked.
When a NAD is tracked it is removed from the map in line 181.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Its a left over from testing trying to see order doesnt matter and things work as it should, I will removed it sorry for the noise.

ormergi:
Rebase

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Done

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
ormergi: Done

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Done

ormergi (commented):
Changes: resolve

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
npinaeva: oh I completely missed that part, I don't think that it is a good idea to modify the map you are iterating over inside the loop.

npinaeva: somewhat of a grey area https://groups.google.com/g/golang-nuts/c/eJqwQLONhLs but probably should work

ormergi: Added comment to clearfy when we remove iterated entries.

npinaeva:
Looks good now, let's add it to the https://github.com/ovn-org/ovn-kubernetes/pull/4612

ormergi:
> Looks good now, let's add it to the #4612

Done

trozet:
so should we close this in favor of 4612?

ormergi:
Yep, close in favor of #4612 

/close

Issue #4765, opened 2024-10-09
UT: ensure UDNGateway is invoked for Primary UDNs when feature gate is ON is Panicking


tssurya:
### Which jobs are flaking?

unit tests

### Which tests are flaking?

ensure UDNGateway is invoked for Primary UDNs when feature gate is ON

### Since when has it been flaking?

like today?

tssurya:
```
SecondaryNodeNetworkController
/home/surya/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/secondary_node_network_controller_test.go:34
  ensure UDNGateway is invoked for Primary UDNs when feature gate is ON [It]
  /home/surya/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/secondary_node_network_controller_test.go:121

  Test Panicked
  runtime error: invalid memory address or nil pointer dereference
  /usr/local/go/src/runtime/panic.go:261

  Full Stack Trace
  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util.RunOVSVsctl({0xc000882fe0, 0xc, 0xc0008830a0?})
  	/home/surya/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/ovs.go:318 +0x121
  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node.(*UserDefinedNetworkGateway).addUDNManagementPort(0xc000324930)
  	/home/surya/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/gateway_udn.go:328 +0x28f
  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node.(*UserDefinedNetworkGateway).AddNetwork(0xc000324930)
  	/home/surya/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/gateway_udn.go:225 +0x2a
  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node.(*SecondaryNodeNetworkController).Start(0xc0004b0180, {0x273f6e0?, 0x3ac5620?})
  	/home/surya/go/src/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/secondary_node_network_controller.go:77 +0x1b4
  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node.init.func17.5()
[...]

PR #4713, opened 2024-09-11
Fix sudo perms when running tests

tssurya:
#### What this PR does and why is it needed
`
mount --make-rshared /run/user/1000/netns failed: "operation not permitted"
`

#### Which issue(s) this PR fixes

Fixes #

#### Special notes for reviewers
Rather than making all the tests there go into sudo mode I moved out the test that requires the testNS framework into its own package so that we can run the other tests properly downstream.

#### Does this PR introduce a user-facing change?

```release-note

```

PR #4767, opened 2024-10-09
Fix panic in UT

tssurya:
This was introduced in https://github.com/ovn-org/ovn-kubernetes/pull/4713 but seriously I don't know why UT CI did not fail in that PR?

Fixes: #4765

### DIFF
--- a/go-controller/pkg/node/secondary_node_network_controller_test.go
+++ b/go-controller/pkg/node/secondary_node_network_controller_test.go
@@ -35,6 +35,8 @@ var _ = Describe("SecondaryNodeNetworkController", func() {
 	var (
 		nad = ovntest.GenerateNAD("bluenet", "rednad", "greenamespace",
 			types.Layer3Topology, "100.128.0.0/16", types.NetworkRolePrimary)
+		fexec      *ovntest.FakeExec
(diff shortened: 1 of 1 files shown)


pperiyasamy:
The run https://github.com/ovn-org/ovn-kubernetes/actions/runs/11251572547/job/31282909395?pr=4762 hits this as well.

tssurya:
let's see if CI passes now

kyrtapz:
> This was introduced in #4713 but seriously I don't know why UT CI did not fail in that PR?

Maybe that's because we are not resetting the runner and one describe block sets it so depending on which goes first it works or crashes?
I think we are missing this in both blocks:
```
	AfterEach(func() {
		util.ResetRunner()
	})
```

kyrtapz (approved):

tssurya:
> Maybe that's because we are not resetting the runner and one describe block sets it so depending on which goes first it works or crashes? I think we are missing this in both blocks:
> 
> ```
> 	AfterEach(func() {
> 		util.ResetRunner()
> 	})
> ```

yea perhaps though it is funny is started perma failing since morning he he he
but good catch here, I have repushed with this

tssurya:
This PR only touches unit tests, has no changes to e2e's has an lgtm, merging it

PR #4771, opened 2024-10-11
Improves pod deletion with user defined networks

trozet:
Delete path was printing removing the pod for every network controller like:

I1009 23:16:54.375327    8134 base_network_controller_secondary.go:287] [e2e-test-endpointslices-mirror-e2e-hg7st/backend-0] addLogicalPort for NAD e2e-test-endpointslices-mirror-e2e-hg7st/gryffindor took 3.202984ms, libovsdb time 2.456846ms
I1009 23:16:54.377764    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-network-segmentation-e2e-b66l5.gryffindor
I1009 23:16:54.378006    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-endpointslices-mirror-e2e-89gg6.gryffindor
I1009 23:16:54.378095    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network cw92p_gryffindor
I1009 23:16:54.378168    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-network-segmentation-e2e-cq4pk.test-net
I1009 23:16:54.378238    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-endpointslices-mirror-e2e-d8xg4.gryffindor
I1009 23:16:54.378313    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network 26x4f_gryffindor
I1009 23:16:54.378384    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-endpointslices-mirror-e2e-hg7st.gryffindor
I1009 23:16:54.382338    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network zw5ld_blue
I1009 23:16:54.382415    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-network-segmentation-e2e-nkltz-red.red
I1009 23:16:54.382479    8134 base_network_controller_secondary.go:416] Deleting pod: e2e-test-endpointslices-mirror-e2e-hg7st/backend-0 for network e2e-test-network-segmentation-e2e-nkltz-blue.blue

This changes delete to return early if this pod is not configured by this network controller.

Also stops getting the network mapping multiple times in a single delete.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_secondary.go
+++ b/go-controller/pkg/ovn/base_network_controller_secondary.go
@@ -458,27 +458,11 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 	}
 
 	podDesc := pod.Namespace + "/" + pod.Name
-	klog.Infof("Deleting pod: %s for network %s", podDesc, bsnc.GetNetworkName())
 
 	// there is only a logical port for local pods or remote pods of layer2
 	// networks on interconnect, so only delete in these cases
 	isLocalPod := bsnc.isPodScheduledinLocalZone(pod)
 	hasLogicalPort := isLocalPod || bsnc.isLayer2Interconnect()
-
-	// otherwise just delete pod IPs from the namespace address set
-	if !hasLogicalPort {
-		if bsnc.doesNetworkRequireIPAM() &&
-			(util.IsMultiNetworkPoliciesSupportEnabled() || (util.IsNetworkSegmentationSupportEnabled() && bsnc.IsPrimaryNetwork())) {
-			return bsnc.removeRemoteZonePodFromNamespaceAddressSet(pod)
-		}
-
-		// except for localnet networks, continue the delete flow in case a node just
-		// became remote where we might still need to cleanup. On L3 networks
-		// the node switch is removed so there is no need to do this.
-		if bsnc.TopologyType() != types.LocalnetTopology {
-			return nil
-		}
-	}
 
 	// for a specific NAD belongs to this network, Pod's logical port might already be created half-way
 	// without its lpInfo cache being created; need to deleted resources created for that NAD as well.
@@ -492,19 +476,31 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 		portInfoMap = map[string]*lpInfo{}
 	}
 
-	activeNetwork, err := bsnc.getActiveNetworkForNamespace(pod.Namespace)
-	if err != nil {
-		return fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
-	}
+	var alreadyProcessed bool
 	for nadName := range podNetworks {
 		if !bsnc.HasNAD(nadName) {
 			continue
 		}
 
-		_, networkMap, err := util.GetPodNADToNetworkMappingWithActiveNetwork(pod, bsnc.NetInfo, activeNetwork)
-		if err != nil {
-			bsnc.recordPodErrorEvent(pod, err)
-			return err
+		// pod has a network managed by this controller
+		klog.Infof("Deleting pod: %s for network %s, NAD: %s", podDesc, bsnc.GetNetworkName(), nadName)
+
+		// handle remote pod clean up but only do this one time
+		if !hasLogicalPort && !alreadyProcessed {
+			if bsnc.doesNetworkRequireIPAM() &&
+				// address set is for network policy only. So either multi network policy is enabled or network
+				// segmentation, and it is a primary UDN (regular netpol)
+				(util.IsMultiNetworkPoliciesSupportEnabled() || (util.IsNetworkSegmentationSupportEnabled() && bsnc.IsPrimaryNetwork())) {
+				return bsnc.removeRemoteZonePodFromNamespaceAddressSet(pod)
+			}
+
+			// except for localnet networks, continue the delete flow in case a node just
+			// became remote where we might still need to cleanup. On L3 networks
+			// the node switch is removed so there is no need to do this.
+			if bsnc.TopologyType() != types.LocalnetTopology {
+				return nil
+			}
+			alreadyProcessed = true
 		}
 
 		if kubevirt.IsPodAllowedForMigration(pod, bsnc.NetInfo) {
@@ -538,32 +534,18 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 			continue
 		}
 
-		network := networkMap[nadName]
-
-		hasPersistentIPs := bsnc.allowPersistentIPs()
-		hasIPAMClaim := network != nil && network.IPAMClaimReference != ""
-		if hasIPAMClaim && !hasPersistentIPs {
-			klog.Errorf(
-				"Pod %s/%s referencing an IPAMClaim on network %q which does not honor it",
-				pod.GetNamespace(),
-				pod.GetName(),
-				bsnc.NetInfo.GetNetworkName(),
-			)
-			hasIPAMClaim = false
-		}
-		if hasIPAMClaim {
-			ipamClaim, err := bsnc.ipamClaimsReconciler.FindIPAMClaim(network.IPAMClaimReference, network.Namespace)
-			hasIPAMClaim = ipamClaim != nil && len(ipamClaim.Status.IPs) > 0
-			if apierrors.IsNotFound(err) {
-				klog.Errorf("Failed to retrieve IPAMClaim %q but will release IPs: %v", network.IPAMClaimReference, err)
-			} else if err != nil {
-				return fmt.Errorf("failed to get IPAMClaim %s/%s: %w", network.Namespace, network.IPAMClaimReference, err)
+		// if we allow for persistent IPs, then we need to check if this pod has an IPAM Claim
+		if bsnc.allowPersistentIPs() {
+			hasIPAMClaim, err := bsnc.hasIPAMClaim(pod, nadName)
+			if err != nil {
+				return fmt.Errorf("unable to determine if pod %s has IPAM Claim: %w", podDesc, err)
+			}
+			// if there is an IPAM claim, don't release the pod IPs
+			if hasIPAMClaim {
+				continue
 			}
 		}
 
-		if hasIPAMClaim {
-			continue
-		}
 		// Releasing IPs needs to happen last so that we can deterministically know that if delete failed that
 		// the IP of the pod needs to be released. Otherwise we could have a completed pod failed to be removed
 		// and we dont know if the IP was released or not, and subsequently could accidentally release the IP
@@ -578,6 +560,45 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 
 	}
 	return nil
+}
+
+// hasIPAMClaim determines whether a pod's IPAM is being handled by IPAMClaim CR.
+// pod passed should already be validated as having a network connection to nadName
+func (bsnc *BaseSecondaryNetworkController) hasIPAMClaim(pod *kapi.Pod, nadName string) (bool, error) {
+	// FIXME(trozet): relying on NAD/UDN existing can lead to a race condition on deletion and
+	// we should figure out a way to determine if the pod has an IPAMClaim without depending on this
+	activeNetwork, err := bsnc.getActiveNetworkForNamespace(pod.Namespace)
+	if err != nil {
+		return false, fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
+	}
+
+	on, networkMap, err := util.GetPodNADToNetworkMappingWithActiveNetwork(pod, bsnc.NetInfo, activeNetwork)
+	if err != nil {
+		bsnc.recordPodErrorEvent(pod, err)
+		return false, fmt.Errorf("error getting network-attachment for pod %s/%s network %s: %w",
+			pod.Namespace, pod.Name, bsnc.GetNetworkName(), err)
+	}
+
+	if !on {
+		return false, nil
+	}
+
+	network := networkMap[nadName]
+	hasIPAMClaim := network != nil && network.IPAMClaimReference != ""
+	if !hasIPAMClaim {
+		return false, nil
+	}
+
+	ipamClaim, err := bsnc.ipamClaimsReconciler.FindIPAMClaim(network.IPAMClaimReference, network.Namespace)
+	if apierrors.IsNotFound(err) {
+		klog.Errorf("Failed to retrieve IPAMClaim %q but will release IPs: %v", network.IPAMClaimReference, err)
+		return false, nil
+	} else if err != nil {
+		return false, fmt.Errorf("failed to get IPAMClaim %s/%s: %w", network.Namespace, network.IPAMClaimReference, err)
+	}
+
+	hasIPAMClaim = ipamClaim != nil && len(ipamClaim.Status.IPs) > 0
+	return hasIPAMClaim, nil
 }
 
 // delPerPodSNAT will delete the SNAT towards masqueradeIP for this given pod
--- a/go-controller/pkg/ovn/base_network_controller_secondary.go
+++ b/go-controller/pkg/ovn/base_network_controller_secondary.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"strings"
 	"time"
 
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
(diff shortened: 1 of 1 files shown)


tssurya:
I have never seen tools lane fail: https://github.com/ovn-org/ovn-kubernetes/actions/runs/11298323245/job/31427992514#step:12:320 gonna re-run to see if it passes

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -412,6 +412,27 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 		return nil
 	}
 
+	activeNetwork, err := bsnc.getActiveNetworkForNamespace(pod.Namespace)
+	if err != nil {
+		return fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
```
tssurya, line 417: nit, any value in adding the pod Name here as well to understand which pod's namespace couldn't get pulled?

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -412,6 +412,27 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+		klog.Errorf("Error getting network-attachment for pod %s/%s network %s: %v",
+			pod.Namespace, pod.Name, bsnc.GetNetworkName(), err)
+		return nil
+	}
+
+	if !on {
+		// the pod is not attached to this specific network
+		klog.V(5).Infof("Pod %s/%s is not attached on this network controller %s",
+			pod.Namespace, pod.Name, bsnc.GetNetworkName())
+		return nil
```
tssurya, line 433: so for L3/localnet I don't see a problem.. for L2 we usually create a remote port for the remote pod, so if we move it here will we not call `pInfo, err := bsnc.deletePodLogicalPort(pod, portInfoMap[nadName], nadName)` in that case? or are the remote ports for pods in l2 removed elsewhere...

tssurya (changes requested):
1 QQ on l2 plus we need unit tests for pod deletes on UDNs (I can't seem to find any maybe I didn't search well)- sorry I know we don't have any today and its not supposed to be part of this PR but while we are here can we add the missing gap coverage so that we know L2's remote port is removed?

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: it will already get printed cause the calling function prints the pod key:

```
				if err := r.ResourceHandler.DeleteResource(entry.oldObj, entry.config); err != nil {
					entry.timeStamp = time.Now()
					entry.failedAttempts++
					if entry.failedAttempts >= MaxFailedAttempts {
						klog.Errorf("Retry delete failed final attempt for %s %s: error: %v", r.ResourceHandler.ObjType, objKey, err)
					} else {
						klog.Infof("Retry delete failed for %s %s, will try again later: %v",
							r.ResourceHandler.ObjType, objKey, err)
					}
```

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: this code will remove that port as well I think, but I dont see how that is related to the change here. The remote pod would still be on this network right?

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya: makes sense:
```
func (bsnc *BaseSecondaryNetworkController) DeleteSecondaryNetworkResourceCommon(objType reflect.Type, obj, cachedObj interface{}) error {
	switch objType {
	case factory.PodType:
		var portInfoMap map[string]*lpInfo
		pod := obj.(*kapi.Pod)

		if cachedObj != nil {
			portInfoMap = cachedObj.(map[string]*lpInfo)
		}
		return bsnc.removePodForSecondaryNetwork(pod, portInfoMap)
```

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya: my bad. I was totally confused between local/remote pods and
same/notsame networks.

this fix is trying to skip processing pods that don't belong to this network.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -412,6 +412,27 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+		return fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
+	}
+
+	on, networkMap, err := util.GetPodNADToNetworkMappingWithActiveNetwork(pod, bsnc.NetInfo, activeNetwork)
+	if err != nil {
+		bsnc.recordPodErrorEvent(pod, err)
+		// configuration error, no need to retry, do not return error
+		klog.Errorf("Error getting network-attachment for pod %s/%s network %s: %v",
+			pod.Namespace, pod.Name, bsnc.GetNetworkName(), err)
+		return nil
```
tssurya, line 426: return nil was intentional? we used to retry the deletion if we failed to get this

trozet:
@maiqueb PTAL

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -446,21 +431,29 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 		portInfoMap = map[string]*lpInfo{}
 	}
 
-	activeNetwork, err := bsnc.getActiveNetworkForNamespace(pod.Namespace)
-	if err != nil {
-		return fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
-	}
+	var alreadyProcessed bool
 	for nadName := range podNetworks {
 		if !bsnc.HasNAD(nadName) {
```
tssurya, line 436: self-note this is what helps here, if this pod is not part of this network then we don't process it, so that's good
@jcaamano or @maiqueb do we know why we added the whole "Active network"/"GetPodNADToNetworkMappingWithActiveNetwork" logic here without resusing the logic we had already for secondary networks to determing if this pod is part of this network controller or not..?

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -529,6 +508,31 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 	return nil
 }
 
+// hasIPAMClaim determines whether a pod's IPAM is being handled by IPAMClaim CR.
+// pod passed should already be validated as having a network connection to nadName
```
tssurya, line 512: I love the function split here, its nice its split outside so the other function became simpler.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -446,21 +431,29 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+			}
 
+			// except for localnet networks, continue the delete flow in case a node just
+			// became remote where we might still need to cleanup. On L3 networks
+			// the node switch is removed so there is no need to do this.
+			if bsnc.TopologyType() != types.LocalnetTopology {
+				return nil
+			}
+		}
+		alreadyProcessed = true
```
tssurya, line 456: you have this bool so that we don't try to remove the IPs from the address set each time for a given pod as it loops through all networks and instead only do it once huh? should you be putting this inside the "if" block (just to couple that logic together)

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -446,21 +431,29 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+		if !hasLogicalPort && !alreadyProcessed {
+			if bsnc.doesNetworkRequireIPAM() && util.IsMultiNetworkPoliciesSupportEnabled() {
+				return bsnc.removeRemoteZonePodFromNamespaceAddressSet(pod)
+			}
 
+			// except for localnet networks, continue the delete flow in case a node just
+			// became remote where we might still need to cleanup. On L3 networks
+			// the node switch is removed so there is no need to do this.
+			if bsnc.TopologyType() != types.LocalnetTopology {
+				return nil
```
tssurya, line 453: @trozet I learnt recently that its possible to have localnet networks with 0 subnets, in that case will we have a list of podNetworks coming from the podAnnotation for the localnet? same for L2... so if we don't have a IPAM on the pod for those networks then I think we might not be able to clear the address-set ips for secondary networks but then I guess in that case we don't also support multi-nework policies there right?

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -529,6 +508,31 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 	return nil
 }
 
+// hasIPAMClaim determines whether a pod's IPAM is being handled by IPAMClaim CR.
+// pod passed should already be validated as having a network connection to nadName
+func (bsnc *BaseSecondaryNetworkController) hasIPAMClaim(pod *kapi.Pod) (bool, error) {
+	if !bsnc.IsPrimaryNetwork() || !bsnc.AllowsPersistentIPs() {
```
tssurya, line 514: we don't need to do this for secondary networks? i thought we support ipam claim for secondary as well? - but it certainly seems like we were doing the earlier logic only for active networks of the pod...

tssurya (commented):
overall I think it makes sense to me

@maiqueb would be good if you can check the ipamclaim pieces since i am not too familar with that

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: since its not possible for a pod to have a logical port but then be connected to another network and not have a logical port I don't think its functionally wrong, but I can move it into the statement to make it consistent code-wise

go-controller/pkg/ovn/base_network_controller_secondary.go
maiqueb: I'm also confused here .. 

> we don't need to do this for secondary networks? i thought we support ipam claim for secondary as well? - but it certainly seems like we were doing pthe logic only for active networks of the pod in the older code as well..

For secondary networks, the `IPAMClaim` will be pointed by the `IPAMClaimReference` [attribute](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/7d2def1df606637b28a3d8cde6793025f3bd053e/pkg/apis/k8s.cni.cncf.io/v1/types.go#L170) defined in the network selection element, as indicated in the multi-network de-facto standard.

For sure this code is required *also* for secondary networks.

maiqueb (changes requested):
The IPAMClaims are supported also for secondary networks of layer2 and localnet topologies.

Unless I'm missing something, I don't see how the current proposal can work.

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: So in the current code:
```
hasIPAMClaim := network != nil && network.IPAMClaimReference != ""
```

network.IPAMClaimReference is set in 
```
_, networkMap, err := util.GetPodNADToNetworkMappingWithActiveNetwork(pod, bsnc.NetInfo, activeNetwork)
```

In there we get GetPodNADToNetworkMapping, which returns the network selection elements for the pod, keyed by NAD, then if the active network is not default, we add that network to the map and if the pod has the annotation:

```
OvnUDNIPAMClaimName = "k8s.ovn.org/primary-udn-ipamclaim"
```

we add its ipamclaim reference to the network.

I missed that the secondary ipam claim reference was stored in the network selection element and thought you were looking for primary only. I will modify my code to account for this!

maiqueb: @trozet let me know once you have re-worked it.

trozet:
last push includes the fixes @tssurya @maiqueb . I'll now push again with just rebasing.

maiqueb (dismissed):
Thanks. I think this will output less junk to the logs, and help by not having to re-read the network mapping on a delete operation.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -542,6 +524,58 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+		ipamClaimName, wasPersistentIPRequested = pod.Annotations[util.OvnUDNIPAMClaimName]
+	} else {
+		// secondary network the IPAM claim reference is on the network selection element
+		nadKeys := strings.Split(nadNamespacedName, "/")
+		if len(nadKeys) != 2 {
+			return false, fmt.Errorf("invalid NAD name %s", nadNamespacedName)
+		}
+		nadNamespace := nadKeys[0]
+		nadName := nadKeys[1]
+		allNetworks, err := util.GetK8sPodAllNetworkSelections(pod)
```
tssurya, line 547: @maiqueb I think we should do a FUP for adding tests for IPAM claims deletion, the fact that we were not handling all the networks and we didn't catch it looks like its not being tested? wanna open an issue?

tssurya (commented):
a few questions for the first half
and asked Miguel to also review the ipamclaim stuff again

go-controller/pkg/ovn/base_network_controller_secondary.go
maiqueb: What do you mean we weren't handling all the networks ? 

And we have tests deleting the allocations (... on the cluster manager side ...): 
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d1ebf1c6e335cf095458a3beafd1a601d3eb5a71/go-controller/pkg/clustermanager/pod/allocator_test.go#L452
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d1ebf1c6e335cf095458a3beafd1a601d3eb5a71/go-controller/pkg/clustermanager/pod/allocator_test.go#L475

This actually leads me to question: isn't this part of the code only used on non-IC ?...

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -542,6 +524,58 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+		if err != nil {
+			return false, err
+		}
+		for _, network := range allNetworks {
+			if network.Namespace == nadNamespace && network.Name == nadName {
```
maiqueb, line 552: I kind of miss the point of this comparison ... 

The `nadNamespacedName` we are passing via input parameter already comes from the pod's network annotations: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/38c2eca51200fe5f7b6a70f7a6604430046006fd/go-controller/pkg/ovn/base_network_controller_secondary.go#L448

I'm trying to make sure iterating the pod's networks is not redundant.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -542,6 +524,58 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
 	return nil
 }
 
+// hasIPAMClaim determines whether a pod's IPAM is being handled by IPAMClaim CR.
+// pod passed should already be validated as having a network connection to nadName
+func (bsnc *BaseSecondaryNetworkController) hasIPAMClaim(pod *kapi.Pod, nadNamespacedName string) (bool, error) {
```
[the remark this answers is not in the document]
maiqueb: yes. `nadName` is in `<ns>/<metadata.name>` format for **all** secondary networks.

Keep in mind it'll be `default` for the cluster default network! But I don't think this **cannot** explode since we only invoke this method for network managed by the `BaseSecondaryNetworkController`. Which I assume can't manage the default network NAD.

maiqueb (commented):
@tssurya FYI

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: I think by "all" she means primary and secondary networks with network selection elements.

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: what do you propose instead @maiqueb ? We get a list of networks and we need to find which one matches, I don't see a better way. Although I realize i am not breaking the for loop once we find the matching nadName. I'll update that.

maiqueb: I though we could get the network-selection-element from the caller, but what we are [iterating](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/38c2eca51200fe5f7b6a70f7a6604430046006fd/go-controller/pkg/ovn/base_network_controller_secondary.go#L448) in the caller method is the podAnnotation, not the network selection elements.

Yeah, you need to do this.

maiqueb (approved):
Thank you.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -542,6 +524,58 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+		return false, nil
+	}
+
+	var ipamClaimName string
+	var wasPersistentIPRequested bool
+	if bsnc.IsPrimaryNetwork() {
+		// primary network ipam reference claim is on the annotation
+		ipamClaimName, wasPersistentIPRequested = pod.Annotations[util.OvnUDNIPAMClaimName]
+	} else {
+		// secondary network the IPAM claim reference is on the network selection element
```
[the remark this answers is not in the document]
maiqueb: LGTM.

trozet:
@maiqueb just rebased and had to resolve one conflict here:

https://github.com/ovn-kubernetes/ovn-kubernetes/compare/33ec28ddcf63cd330ef215a41cfa5fdf3c10069c..00f1533be76304dd8d6dbf51b82b8609271931ab#diff-7ade52078b2eac50eeb5f96fb67df39d804cf8b48592206691f748334f785b58R507

Spoke with @RamLavi and I think I resolved it correctly. @RamLavi if you want to double check thats cool too.

maiqueb (approved):

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -459,21 +444,32 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
+				return bsnc.removeRemoteZonePodFromNamespaceAddressSet(pod)
+			}
+
+			// except for localnet networks, continue the delete flow in case a node just
+			// became remote where we might still need to cleanup. On L3 networks
+			// the node switch is removed so there is no need to do this.
+			if bsnc.TopologyType() != types.LocalnetTopology {
+				return nil
+			}
+			alreadyProcessed = true
```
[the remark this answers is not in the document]
trozet: I think I missed that we return after calling bsnc.removeRemoteZonePodFromNamespaceAddressSet(pod). We also return in the current code:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/base_network_controller_secondary.go#L472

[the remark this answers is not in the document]
tssurya: yea this is for localnet pods alone, @trozet and I spoke offline on slack, this is harmless

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -459,21 +444,32 @@ func (bsnc *BaseSecondaryNetworkController) removePodForSecondaryNetwork(pod *ka
...
-		_, networkMap, err := util.GetPodNADToNetworkMappingWithActiveNetwork(pod, bsnc.NetInfo, activeNetwork)
-		if err != nil {
-			bsnc.recordPodErrorEvent(pod, err)
-			return err
-		}
+		// pod has a network managed by this controller
+		klog.Infof("Deleting pod: %s for network %s, NAD: %s", podDesc, bsnc.GetNetworkName(), nadName)
 
+		// handle remote pod clean up but only do this one time
+		if !hasLogicalPort && !alreadyProcessed {
```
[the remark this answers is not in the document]
trozet: handled by the pInfo, err := bsnc.deletePodLogicalPort(pod, portInfoMap[nadName], nadName) function:

```
	// if the ip is in use by another pod we should not try to remove it from the address set
	if shouldRelease {
		if ops, err = bnc.deletePodFromNamespace(pod.Namespace,
			podIfAddrs, portUUID); err != nil {
			return nil, fmt.Errorf("unable to delete pod %s from namespace: %w", podDesc, err)
		}
		allOps = append(allOps, ops...)
	}
```

tssurya (approved):

tssurya:
```
Summarizing 1 Failure:
  [FAIL] Network Segmentation: services on a user defined primary network should be reachable through their cluster IP and node port [It] L3 primary UDN, cluster-networked pods, NodePort service
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_services.go:422
```
is known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4829

PR #4772, opened 2024-10-12
Wip shared portgroup

cathy-zhou:

Fixes #

```release-note

```


cathy-zhou:
@npinaeva please take a look

for upgrade (specifically, the deletion of stale network policy based port gorup), I am current delete those during the next reboot. I can also change it to delete it after the new shared port group is created for the specific network policy, but that will be done each time during restart even when the upgrade is already done. Let me know if you have any preference.

npinaeva:
@cathy-zhou I remember about this PR, but unfortunately I am swamped with the UDN/conferences/etc. I hope to have some time to review it in December, but can't promise :(

girishmg:
This whole PR needs to be re-worked and there is no review context in this PR that we would loose by closing this PR, so closing this PR.

PR #4773, opened 2024-10-13
UDN: Patch Kubevirt CR to support managedTap binding

oshoval:

Fixes #

```release-note

```


maiqueb (changes requested):
you only need one in upstream OVN-K, right ?

oshoval:
True, i just wanted to do it smoother, i.e that as long as IPAM is not updated, it wont break IPAM / CNAO etc

maiqueb:
> True, i just wanted to do it smoother, i.e that as long as IPAM repo is not updated, it wont break it (we can make the change there first if you prefer and drop in this PR)

Whatever is both conceptually correct and simpler for you.

oshoval:
Removed the passt here - going to test it on the dummy PR

also added the PR for IPAM that works on CI as well

qinqon:
Going to create a new PR doing the opposite, so the userDefinedPrimaryNetworkBinding will point to managedTap, we can test later on passt by installing HCO using the json capababilities so we patch the binding and change from managedTap to passt, no need for new binding.

oshoval:
injected to use nightly
lets see all pass (and then we can remove it)

oshoval:
either i did something wrong or the managedTap is broken 
will check
https://github.com/ovn-org/ovn-kubernetes/actions/runs/11540164957/job/32121009180?pr=4773

oshoval:
added new required FG, lets see it works

oshoval:
Hi Miguel, please disregard the re-review request, we need anyhow first to discuss offline about generic name and such

contrib/kind-common
```diff
@@ -328,6 +328,8 @@ install_kubevirt() {
     # nightly tag - install specific nightly (i.e 20240910)
     KUBEVIRT_VERSION=${KUBEVIRT_VERSION:-"stable"}
 
+    KUBEVIRT_VERSION=nightly
```
oshoval, line 331: will be dropped

oshoval:
failures are not related, the UDN ones passed

contrib/kind-common
```diff
@@ -356,11 +358,8 @@ install_kubevirt() {
         done
     fi
 
-    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/developerConfiguration","value":{"featureGates":[]}},{"op":"add","path":"/spec/configuration/developerConfiguration/featureGates/-","value":"NetworkBindingPlugins"}]'
-    
-    local kubevirt_stable_release_url=$(get_kubevirt_release_url "stable")
-    local passt_binding_image="quay.io/kubevirt/network-passt-binding:${kubevirt_stable_release_url##*/}"
-    kubectl -n kubevirt patch kubevirt kubevirt --type=json --patch '[{"op":"add","path":"/spec/configuration/network","value":{}},{"op":"add","path":"/spec/configuration/network/binding","value":{"passt":{"computeResourceOverhead":{"requests":{"memory":"500Mi"}},"migration":{"method":"link-refresh"},"networkAttachmentDefinition":"default/primary-udn-kubevirt-binding","sidecarImage":"'"${passt_binding_image}"'"}}}]'
```
maiqueb, line 363: why don't we want both of them ? and let the user pick which binding to use ?

test/e2e/kubevirt.go
```diff
@@ -1317,11 +1317,7 @@ passwd:
 
 			step := by(vmi.Name, "Login to virtual machine for the first time")
 			Eventually(func() error {
-				if td.role != "primary" {
-					return kubevirt.LoginToFedora(vmi, "core", "fedora")
-				} else {
-					return kubevirt.LoginToFedoraWithHostname(vmi, "core", "fedora", "localhost")
-				}
```
maiqueb, line 1324: why did this binding type changed the prompt ?

maiqueb (commented):
Some questions.

contrib/kind-common
oshoval: just because you asked so i dropped
i can return it if desired as we spoke

test/e2e/kubevirt.go
oshoval: since with the new one the DHCP populate now the right hostname, no need this special treatment that was to workaround this when passt is used
Quique is familiar with it best, and said it is good btw

maiqueb: to me, it seems like we first need to decide if we want to keep both bindings around or not. 

like, if we keep passt around, we can't have this change.

oshoval: those are the tests, in tests as long as we dont test both passt and managedTap we can have just one
it is orthogonal to have both on the CR and then the user can still choose

oshoval:
Updated as we spoke offline
CR support both
e2e tests here runs mangedTap 

Note please the 3rd commit, if we dont want nightly because it might break OVN if kubevirt breaks,
we can pin specific nightly meanwhile `20241029` or ask for a stable but it might take time and we would be block until that if we do.

contrib/kind-common
```diff
@@ -326,7 +326,7 @@ install_kubevirt() {
     # vX.Y.Z - install specific stable (i.e v1.3.1)
     # nightly - install newest nightly
     # nightly tag - install specific nightly (i.e 20240910)
-    KUBEVIRT_VERSION=${KUBEVIRT_VERSION:-"stable"}
+    KUBEVIRT_VERSION=${KUBEVIRT_VERSION:-"nightly"}
```
qinqon, line 329: Should we wait for a  kubevirt release ?

test/e2e/kubevirt.go
qinqon: just to add some context, passt do not support the DHCPv4 option hostname (12), so VM ends up with localhost, also is good that we test this kind of things since at managedTap we pass that with the OVN DHCP.

qinqon (approved):
Looks good as far as CI is happy.

contrib/kind-common
oshoval: https://github.com/kubevirt/sig-release/blob/main/releases/v1.4/schedule.md
lets wait, rc should be any day

oshoval: disregard, stable.txt wont have it, checking about alternatives (i.e new file on that path etc)

maiqueb (changes requested):
Let's pin a KubeVirt 1.4 RC version (which is not available yet ... should be imminent).

contrib/kind-common
oshoval: yep

test/e2e/kubevirt.go
oshoval: sgtm

contrib/kind-common
oshoval: there already

oshoval:
Updated to use `v1.4.0-rc.0`

oshoval:
rebased

qinqon (approved):
/lgtm

maiqueb (approved):
Tks.

@tssurya could you take a look ? We are registering the new binding (that allows live migration) in KubeVirt.

We are keeping the other around because we have people working on improving it (the long term plan is to use it).

contrib/kind-common
```diff
@@ -326,7 +326,7 @@ install_kubevirt() {
     # vX.Y.Z - install specific stable (i.e v1.3.1)
     # nightly - install newest nightly
     # nightly tag - install specific nightly (i.e 20240910)
-    KUBEVIRT_VERSION=${KUBEVIRT_VERSION:-"stable"}
+    KUBEVIRT_VERSION=${KUBEVIRT_VERSION:-"v1.4.0-rc.0"}
```
oshoval, line 329: /hold

1.4.0 is out, so i will first change to use stable as always

PR #4776, opened 2024-10-14
Skip processing unrelated namespace peers for Network Policy

pperiyasamy:
When network policy object contains peers from namespace which is not on subject's user defined network, then it doesn't make sense to add that peer namespace into OVN's network policy ACL, because those namespaces are isolated already. Hence this PR avoids processing such peers in network policy namespace peer handler.


go-controller/pkg/ovn/base_network_controller_multipolicy.go
```diff
@@ -11,7 +11,10 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const PolicyForAnnotation = "k8s.v1.cni.cncf.io/policy-for"
+const (
+	PolicyForAnnotation = "k8s.v1.cni.cncf.io/policy-for"
+	MultiNetPolicyLabel = "multinet-policy"
```
tssurya, line 16: is this a new label we expect users to add? 
I am totally spaced out on this design/discussion part totally

npinaeva:
The only case where this optimization works, is network policy in primary UDN namespace A, having a peer that allows a primary UDN B namespace. Considering this is an allow rule, but this traffic won't work anyway, I don't think this is an important enough case to optimize for.

PR #4778, opened 2024-10-15
Fix incorrect removal of VRF device

pperiyasamy:
This PR fixes repair method of VRF manager which is buggy and it incorrectly removes valid VRF device associated with a user defined network while running reconcile loop.

### DIFF
--- a/go-controller/pkg/node/vrfmanager/vrf_manager.go
+++ b/go-controller/pkg/node/vrfmanager/vrf_manager.go
@@ -18,6 +18,11 @@ import (
 	"k8s.io/klog/v2"
 )
 
+// reconcile period for vrf manager, this would kick in for every 60 seconds if there is no
+// explicit link update events. In the event of link update, reconcile period is automatically
+// extended by another 60 seconds.
+var reconcilePeriod = 60 * time.Second
+
 type vrf struct {
 	name  string
 	table uint32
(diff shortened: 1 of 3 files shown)


kyrtapz (dismissed):
The change itself looks good to me but it feels like we are missing either a unit test or an E2E to make sure that the repair behaves as expected.

pperiyasamy:
> but it feels like we are missing either a unit test

yes @kyrtapz , updated the unit test to cover this.

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -298,8 +298,8 @@ func (vrfm *Controller) repair(validVRFs sets.Set[string]) error {
 			if err != nil {
 				klog.Errorf("VRF Manager: error deleting stale VRF device %s, err: %v", name, err)
 			}
+			delete(vrfm.vrfs, link.Attrs().Index)
```
tssurya, line 301: nit: yeah the other option was to continue if it was validVRFs.Has(name); neater for reading loop

go-controller/pkg/node/vrfmanager/vrf_manager_test.go
```diff
@@ -116,6 +116,9 @@ var _ = ginkgo.Describe("VRF manager", func() {
 			enslaveLinkMock1.On("Type").Return("dummy")
 			err = c.reconcile()
 			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			// Invoke reconcile again to ensure both vrf links in sync.
+			err = c.reconcile()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
```
tssurya, line 121: hmm I see you are testing error didn't happen... but shouldn't we check the vrf was not removed? I don't see the ensure vrf links in sync is happening...? what am I missing..

tssurya: spoke with Peri, he will add another test to check reconcile+ vrf deletion's better

pperiyasamy: yes @tssurya , added a test now. PTAL.

kyrtapz (approved):

go-controller/pkg/node/vrfmanager/vrf_manager_test.go
```diff
@@ -130,3 +141,98 @@ var _ = ginkgo.Describe("VRF manager", func() {
...
+			wg3.Wait()
+
+			// Invoke reconcile method explicitly few times to ensure it's always working fine.
+			err = c.reconcile()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			err = checkforVrfLinkExistence()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			err = c.reconcile()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			err = checkforVrfLinkExistence()
```
tssurya, line 231: awesome!!

tssurya (approved):

PR #4781, opened 2024-10-17
node_annotations: Make GetNodeHostAddrs() return stable results.

dceara:
The host address annotations were parsed as a set of strings and sets are unordered.  In consequence GetNodeHostAddrs() might return differently ordered host addresses when invoked on the same annotation.

In some cases this can cause DeepEqual() comparisons of objects that contain the parsed annotation results to fail even though the annotation itself didn't change.

One such example is with NodeTracker.updateNodeInfo() which compares two nodeInfo structures that contain parsed lists of host addresses.  The unstable parsing was causing unexpected full service resyncs when nothing relevant had actually changed on the node annotation.

This becomes more visible when testing with a larger number of UDNs because the NodeTracker objects are created per UDN.

Without this fix, on a 20 node kind cluster when creating 100 L3 user defined networks with one pod each, pprof CPU traces show Controller.syncService() -> Controller.configureUDNEnabledServiceRoute() taking up ~35% of the CPU time.  Some of the services were resynced thousands of times on some of the UDNs.

With the fix, no service resync happens on already configured UDNs in this case.
#### How to verify it

To manually test this bring up a reasonably sized kind cluster (e.g., 20 nodes) and create a large number of L3 UDNs with one pod in each (e.g., 100 namespaces).  Make sure that resync-ing services doesn't happen on existing nodes when either new nodes are added or new UDNs are created.

#### Details to documentation updates

N/A

N/A

```release-note
NONE
```

### DIFF
--- a/go-controller/pkg/util/node_annotations.go
+++ b/go-controller/pkg/util/node_annotations.go
@@ -1072,7 +1072,7 @@ func GetNodeHostAddrs(node *kapi.Node) ([]string, error) {
 	if err != nil && !IsAnnotationNotSetError(err) {
 		return nil, fmt.Errorf("failed to get node host CIDRs for %s: %s", node.Name, err.Error())
 	}
-	return hostAddresses.UnsortedList(), nil
+	return sets.List(hostAddresses), nil
 }
 
 func ParseNodeHostCIDRsExcludeOVNNetworks(node *kapi.Node) ([]string, error) {


dceara:
CC: @tssurya @kyrtapz

kyrtapz (approved):
The change looks good to me.
We might consider using a set only when it's needed, breaking it out from `ParseNodeHostCIDRsDropNetMask`, just an idea for a potential followup.

go-controller/pkg/util/node_annotations.go
```diff
@@ -1072,7 +1072,7 @@ func GetNodeHostAddrs(node *kapi.Node) ([]string, error) {
 	if err != nil && !IsAnnotationNotSetError(err) {
 		return nil, fmt.Errorf("failed to get node host CIDRs for %s: %s", node.Name, err.Error())
 	}
-	return hostAddresses.UnsortedList(), nil
+	return sets.List(hostAddresses), nil
```
tssurya, line 1075: I think it makes total sense here to sort it so that we have a consistency in deepequal (I have faced similar issues in other parts of the code where I had to do this)

tssurya (approved):

PR #4782, opened 2024-10-18
controller: make stop idempotent.

npinaeva:
Usually controller should only be stopped once, but when it is a part of another controller that tries to cleanup, that may be complicated.
Currently `nodeNetworkControllerManager` on `Start` failure will call its own `Stop` method.
https://github.com/ovn-org/ovn-kubernetes/blob/715d5252f6c8ea62885d7da2c5c67970dedd65c3/go-controller/pkg/network-controller-manager/node_network_controller_manager.go#L173-L177
If it was `nadController` that failed to start https://github.com/ovn-org/ovn-kubernetes/blob/715d5252f6c8ea62885d7da2c5c67970dedd65c3/go-controller/pkg/network-controller-manager/node_network_controller_manager.go#L195 it doesn't not expect to be stopped (it actually already cleaned up its resources on failed start), but ncm `Stop` funciton doesn't track which part failed to start, and will try to `Stop` all controllers https://github.com/ovn-org/ovn-kubernetes/blob/715d5252f6c8ea62885d7da2c5c67970dedd65c3/go-controller/pkg/network-controller-manager/node_network_controller_manager.go#L244-L247. For `nadController` it will be a second cleanup.

### DIFF
--- a/go-controller/pkg/controller/controller.go
+++ b/go-controller/pkg/controller/controller.go
@@ -139,7 +139,12 @@ func (c *controller[T]) startWorkers() error {
 }
 
 func (c *controller[T]) stop() {
+	// we assign stopChan to nil to signal that controller was already stopped.
+	if c.stopChan == nil {
+		return
+	}
 	close(c.stopChan)
+	c.stopChan = nil
 	c.cleanup()
 	c.wg.Wait()
 }
--- a/go-controller/pkg/controller/controller_test.go
+++ b/go-controller/pkg/controller/controller_test.go
@@ -91,6 +91,11 @@ var _ = Describe("Level-driven controller", func() {
 		Stop(controller)
 	})
 
+	It("has idempotent Stop", func() {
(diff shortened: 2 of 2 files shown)


go-controller/pkg/controller/controller.go
```diff
@@ -139,6 +141,10 @@ func (c *controller[T]) startWorkers() error {
 }
 
 func (c *controller[T]) stop() {
+	if c.stopped {
+		return
+	}
+	c.stopped = true
```
jcaamano, line 147: could we just set the `stopChan` to nil on stop and test for it?

npinaeva: that feels like a bit less obvious way to say that controller is already stopped, but I can do that

jcaamano: we save memory :stuck_out_tongue: 
we can add a comment if we feel it would be needed

jcaamano (approved):

Issue #4783, opened 2024-10-18
Fatal error: patch port <port name> ofport changed from 32 to 


npinaeva:
### Which jobs are flaking?

https://github.com/ovn-org/ovn-kubernetes/actions/runs/11384905304/job/31675052357?pr=4647

### Which tests are flaking?

any test may flake because of it

### Since when has it been flaking?

18 Oct 2024 first seen

### Reason for failure (if possible)

Fatal error: patch port patch-breth0_network.segmentation.4885.blue.blue_ovn-worker2-to-br-int ofport changed from 32 to

tssurya:
maybe this is same as https://github.com/ovn-org/ovn-kubernetes/issues/4769 ? hard to tell but let's keep an eye out

PR #4786, opened 2024-10-18
node: Relax check_ports() for UDN patch ports.

dceara:
The UDN deletion process is asynchronous between ovn-kubernetes and ovn-controller.

Considering the following sequence of events:
1. UDN created
2. ovnkube's network controller creates the OVN external switch for the UDN (with a localnet port)
3. ovn-controller creates the patch port for the UDN localnet port
4. ovnkube's node controller updates netConfigs include the UDN network
5. ovnkube's node controller determines the ofPort of the newly created patch port
6. UDN deleted
7. ovnkube's network controller deletes the OVN external switch
8. ovn-controller removes the patch port
9. ovnkube's node controller finally gets the update about UDN being deleted and removes the corresponding netConfig.

Steps "8" and "9" are asynchronous and can happen in any order.  If "8" happens before "9" and if in between "8" and "9"
openflowManager.checkPorts() is called then the patch port was already deleted so ovnkube reports a fatal error and exits.

For the default network that behavior is probably OK because the patch port will never be deleted under normal circumstances.  However, for user defined networks, we eventually converge to a valid topology as soon as both "8" and "9" complete.

Relax the checkPorts() behavior for UDNs to allow for asynchronous deletion.

This was reported by Nadia as it was hit in one unrelated PR CI run: https://github.com/ovn-org/ovn-kubernetes/actions/runs/11384905304/job/31675052357?pr=4647

Fixes: #4783

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1443,7 +1443,7 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 	// already been SNATed to the UDN's masq IP.
 	if config.IPv4Mode {
 		for _, netConfig := range bridge.patchedNetConfigs() {
-			if netConfig.masqCTMark == ctMarkOVN {
+			if netConfig.isDefaultNetwork() {
 				continue
 			}
 			dftFlows = append(dftFlows,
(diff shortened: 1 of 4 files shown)


dceara:
CC: @npinaeva

dceara:
CC: @tssurya

trozet (approved):

PR #4423, opened 2024-06-05
Remove FakeAddressSet from network policy testing

JacobTanenbaum:
in an effort to remove the FakeAddressSet code from testing this commit remove FakeAddressSet from network policy testing

#### Which issue(s) this PR fixes

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #4789, opened 2024-10-21
Ensure UDN namespace is configured when MultiNetPolicy is disabled

pperiyasamy:
The namespace is currently not getting configured in ovnk and ovn for user defined network when multi network policy is disabled for the cluster. But multinet policy is not a requirement for UDN network policies, hence fixing it so that namespaces get configured for those namespaces on secondary controller and network policies can be created for the same.

### DIFF
--- a/go-controller/pkg/ovn/multihoming_test.go
+++ b/go-controller/pkg/ovn/multihoming_test.go
@@ -401,23 +401,25 @@ func enableICFeatureConfig() *config.OVNKubernetesFeatureConfig {
 	return featConfig
 }
 
-func icClusterTestConfiguration() testConfiguration {
-	return testConfiguration{
+type testConfigOpt = func(*testConfiguration)
+
+func icClusterTestConfiguration(opts ...testConfigOpt) testConfiguration {
+	config := testConfiguration{
 		configToOverride:   enableICFeatureConfig(),
 		expectationOptions: []option{withInterconnectCluster()},
 	}
-}
-
-func nonICClusterTestConfiguration() testConfiguration {
-	return testConfiguration{}
-}
-
-func icClusterWithDisableSNATTestConfiguration() testConfiguration {
-	return testConfiguration{
-		configToOverride:   enableICFeatureConfig(),
-		expectationOptions: []option{withInterconnectCluster()},
-		gatewayConfig:      &config.GatewayConfig{DisableSNATMultipleGWs: true},
+	for _, opt := range opts {
+		opt(&config)
 	}
+	return config
+}
+
+func nonICClusterTestConfiguration(opts ...testConfigOpt) testConfiguration {
+	config := testConfiguration{}
+	for _, opt := range opts {
+		opt(&config)
+	}
+	return config
 }
 
 func newMultiHomedPod(testPod testPod, multiHomingConfigs ...secondaryNetInfo) *v1.Pod {
--- a/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
+++ b/go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
@@ -23,16 +23,16 @@ import (
 
 	ovncnitypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
 	libovsdbutil "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/util"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
 	networkAttachDefController "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/network-attach-def-controller"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing"
-	ovntest "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing"
 	libovsdbtest "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing/libovsdb"
 	fakenad "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing/nad"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	ovntypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+	networkingv1 "k8s.io/api/networking/v1"
 )
 
 type secondaryNetInfo struct {
@@ -49,6 +49,8 @@ const (
 	nadName              = "blue-net"
 	ns                   = "namespace1"
 	secondaryNetworkName = "isolatednet"
+	denyPolicyName       = "deny-all-policy"
+	denyPG               = "deny-port-group"
 )
 
 type testConfiguration struct {
@@ -112,11 +114,9 @@ var _ = Describe("OVN Multi-Homed pod operations", func() {
 				)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(netInfo.setupOVNDependencies(&initialDB)).To(Succeed())
-
 				if netInfo.isPrimary {
 					networkConfig, err := util.NewNetInfo(netInfo.netconf())
 					Expect(err).NotTo(HaveOccurred())
-
 					initialDB.NBData = append(
 						initialDB.NBData,
 						&nbdb.LogicalSwitch{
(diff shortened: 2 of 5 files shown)


kyrtapz (approved):

kyrtapz:
We should likely add a unit test for this :)

pperiyasamy:
> We should likely add a unit test for this :)

yes @kyrtapz , added a test now, but couldn't check on address set because fake address set doesn't have support for ovsdb operation (at [here](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/address_set/fake_address_set.go#L483)), so used namespaced port group check for this purpose because that's also added in [ensureNamespaceLockedCommon](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/base_network_controller_namespace.go#L273-L280) method.

kyrtapz (dismissed):

tssurya:
we should move away from fake address set as much as possible - jake is trying to clean up all that to remove that totally - https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4423 I hope you didn't add new instances

go-controller/pkg/ovn/multihoming_test.go
```diff
@@ -386,13 +386,26 @@ func enableICFeatureConfig() *config.OVNKubernetesFeatureConfig {
 	return featConfig
 }
 
+func enableEgressFirewallConfig() *config.OVNKubernetesFeatureConfig {
```
tssurya, line 389: @maiqueb I see such utils are added everywhere across unit test files.. this is just pure duplication over and over just to enable some feature flags, I prefer being explicit instead of having a bunch of nonIC/IC function utils some in multihoming, some in secondary-unit tests etc etc to hide that aspect and I see all these icEnable/disable and same things repeated - please check other legacy unit tests to see how we did it there -> also minimalConfig means nothing from the name itself
this is anti-pattern to what we have in other files, we usually just use the config knobs, booleans are simple enough to be read clearly as is @pperiyasamy can we not have this many inner constructs just to enable 1 flag? what I mean is enableEFW is calling enableIC ?? not intuitive at all which calls minimal config? not intuitive at all - like I need to read all these utils to get what's happening.. can we not just enable it within the IT blocks and simply reset all the configs when we are done in the beforeeach or something?

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
```diff
@@ -369,6 +385,10 @@ var _ = Describe("OVN Multi-Homed pod operations", func() {
 			dummyPrimaryLayer3UserDefinedNetwork("192.168.0.0/16", "192.168.1.0/24"),
 			icClusterWithDisableSNATTestConfiguration(),
 		),
+		Entry("pod on a user defined primary network on an IC cluster with EgressFirewall enabled",
+			dummyPrimaryLayer3UserDefinedNetwork("192.168.0.0/16", "192.168.1.0/24"),
+			icClusterEnableEgressFirewallTestConfiguration(),
```
tssurya, line 390: Isn't this better to be {EFW:true, IC: true, whatever else is in minimal config to be true?} here and in other It functions so that readers can understand it in one go what is being performed here? the abstraction seems overengineerd for enabling the booleans but I am happy to stand corrected if its only me who feels this eay @pperiyasamy what do you think?

tssurya (commented):
main logic LGTM
I left some comments around the config utils that we seem to be spraying across all unit test files PTAL

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
pperiyasamy: yes @tssurya , i agree with your concern, this would bloat config util with too many methods for different configs. doing it now opts argument on the base method. hope you're fine with it now.

pperiyasamy:
> we should move away from fake address set as much as possible - jake is trying to clean up all that to remove that totally - #4423 I hope you didn't add new instances

This is because of using FakeAddressSet from ovn test utility https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/ovn_test.go#L98. I think this must be fixed separately if we want to avoid fake address sets.

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
```diff
@@ -239,6 +246,13 @@ var _ = Describe("OVN Multi-Homed pod operations", func() {
 			icClusterWithDisableSNATTestConfiguration(),
 			config.GatewayModeShared,
 		),
+		Entry("pod on a user defined primary network on an IC cluster with EgressFirewall enabled",
+			dummyPrimaryLayer3UserDefinedNetwork("192.168.0.0/16", "192.168.1.0/24"),
+			icClusterTestConfiguration(func(config *testConfiguration) {
+				config.configToOverride.EnableEgressFirewall = true
```
tssurya, line 252: ++ I like this better, I wanted to say can you do it for the other It entries as well but that's a cleanup and I cannot ask you to do it cause its not related to your PR. but if you ever get time, would love this change in favor or removing the other existing utils.

tssurya (commented):
looks almost good, but one more question
why are we testing EnableEFW when the bug was around NetPol on UDNs ? :D is it because you have no way to enable netpol here and so instead of creating a networkPolicy in UDN you went with the easy way to enable firewall feature to see if namespaces address-sets are configured properly?

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
pperiyasamy: yes @tssurya , updated  existing utils.

pperiyasamy:
> why are we testing EnableEFW when the bug was around NetPol on UDNs ? :D is it because you have no way to enable netpol here and so instead of creating a networkPolicy in UDN you went with the easy way to enable firewall feature to see if namespaces address-sets are configured properly?

The `EnableEFW` test was added only to cover code changes done in the commit [66844b7](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4789/commits/66844b77fa133db0512d203ea695c306b115e363). but of course, you're right. we must validate networkPolicy as well. added another commit [e5c4c68](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4789/commits/e5c4c683f52cedac797cc2b3696655a3f10c797e) to address this.

tssurya:
> > why are we testing EnableEFW when the bug was around NetPol on UDNs ? :D is it because you have no way to enable netpol here and so instead of creating a networkPolicy in UDN you went with the easy way to enable firewall feature to see if namespaces address-sets are configured properly?
> 
> The `EnableEFW` test was added only to cover code changes done in the commit [66844b7](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4789/commits/66844b77fa133db0512d203ea695c306b115e363). but of course, you're right. we must validate networkPolicy as well. added another commit [e5c4c68](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4789/commits/e5c4c683f52cedac797cc2b3696655a3f10c797e) to address this.

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4789/commits/66844b77fa133db0512d203ea695c306b115e363 is not rendering for me @pperiyasamy can you relink which commit you meant?

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
```diff
@@ -1004,3 +1067,84 @@ func buildNamespacedPortGroup(namespace, controller string) *nbdb.PortGroup {
 	pg.UUID = pg.Name + "-UUID"
 	return pg
 }
+
+func createHairpinAllowACLs(controllerName string) (*nbdb.ACL, *nbdb.ACL) {
```
tssurya, line 1071: nit: the utils here, are we sure we cannot reuse the ones in policy_test.go ? some of these below seem to exist there as well right? trying to make sure there is no code duplication here by adding utils in each unit test file if there is a chance of reuse

tssurya (approved):
last test looks good to me, just a tiny bit concerned with all the utils we seem to add per test file instead of seeing if there is potential to reuse then from existing netpol test files

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
pperiyasamy: oh! yes @tssurya , i didn't notice `policy_test.go` before, my bad. reusuing most of its functions for multi homing test now.

go-controller/pkg/ovn/secondary_layer3_network_controller_test.go
```diff
@@ -168,6 +184,9 @@ var _ = Describe("OVN Multi-Homed pod operations", func() {
 
 				Expect(fakeOvn.controller.WatchNamespaces()).NotTo(HaveOccurred())
 				Expect(fakeOvn.controller.WatchPods()).NotTo(HaveOccurred())
+				if netInfo.isPrimary {
```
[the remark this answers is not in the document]
pperiyasamy: yes, `EnableMultiNetworkPolicy` config option is set to false by default from config util functions, so not setting it explicitly from any of the tests.

PR #4790, opened 2024-10-21
Fix tombstone handling for various resources

trozet:


go-controller/pkg/ovn/controller/admin_network_policy/admin_network_policy_controller.go
```diff
@@ -538,9 +538,22 @@ func (c *Controller) onANPPodUpdate(oldObj, newObj interface{}) {
 
 // onANPPodDelete queues the pod for processing.
 func (c *Controller) onANPPodDelete(obj interface{}) {
-	key, err := cache.MetaNamespaceKeyFunc(obj)
+	pod, ok := obj.(*v1.Pod)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
```
tssurya, line 543: so this happens when somehow we missed the deletion event due to disconnection from apiserver perhaps for whatever reason and the object gets placed into the cache and its possible this object is stale but still there is a chance we could trigger deletion based on that huh?

go-controller/pkg/controller/controller.go
```diff
@@ -201,7 +201,22 @@ func (c *controller[T]) onUpdate(oldObjInterface, newObjInterface interface{}) {
 }
 
 func (c *controller[T]) onDelete(objInterface interface{}) {
-	key, err := cache.MetaNamespaceKeyFunc(objInterface)
+	obj, ok := objInterface.(*T)
+	if !ok {
+		tombstone, ok := objInterface.(cache.DeletedFinalStateUnknown)
```
tssurya, line 206: I think you can use:
```
       // get the key for our object, which may have been deleted already
	if key, err = cache.DeletionHandlingMetaNamespaceKeyFunc(obj); err != nil {
		utilruntime.HandleError(blah blah err)
		return
	}
```
this straight? though it seems tiny bit different since it doesn't do the tombstone error reporting but it seems to be the client-go equivalent of what should be called for deletion?

tssurya (changes requested):
some more spots to cleanup:

1. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/clustermanager/egressservice/egressservice_cluster_node.go#L116
2. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/clustermanager/egressservice/egressservice_cluster_service.go#L87
3. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/clustermanager/egressservice/egressservice_cluster.go#L422
4. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/node/controllers/egressservice/egressservice_node.go#L165
5. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/egressqos.go#L338
6. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/admin_network_policy/admin_network_policy_controller.go#L387
7. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/admin_network_policy/admin_network_policy_controller.go#L435
8. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/admin_network_policy/admin_network_policy_controller.go#L481
9. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/egressservice/egressservice_zone_node.go#L60
10. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/egressservice/egressservice_zone_service.go#L101
11. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/egressservice/egressservice_zone.go#L603
12. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/ovn/controller/services/services_controller.go#L636
13. https://github.com/ovn-org/ovn-kubernetes/blob/8926513336b5fbcb3f039994e82ef3d16524eb4d/go-controller/pkg/retry/obj_retry.go#L708 trickier I guess?

PR #4793, opened 2024-10-22
abproute: Do not fail when a pod has bad gw config

zeeke:
Setting the `k8s.ovn.org/routing-network` pod annotation to an empty string on a pod that is not host-networked is the wrong configuration, and it has no effect on the OVN configuration. However, if an ovnkube-node pod is restarted, it is not able to go up and running as the above mentioned configuration triggers an error.

Make the `abprout.Repair()` procedure resilient to bad user configuration, in order to improved the stability of the cluster.

Setting the `k8s.ovn.org/routing-network` pod annotation to an empty string on a pod that is not host-networked is the wrong configuration, and it has no effect on the OVN configuration. However, if an ovnkube-node pod is restarted, it is not able to go up and running as the above mentioned configuration triggers an error.

Make the `abprout.Repair()` procedure resilient to bad user configuration, in order to improved the stability of the cluster.

Fixes #
https://issues.redhat.com/browse/OCPBUGS-37685

I'm not sure if there is a cleaner way to unit test this case. @jordigilh can you please review?

#### How to verify it

Steps to reproduce the problem are listed here
https://gist.github.com/zeeke/378842c9b64f02895ce0175172238c0c

```release-note

```


zeeke:
@jordigilh Any chance to get a review here?

jordigilh:
@zeeke as far as I know, the APB annotations are deprecated and replaced by the [APB External Route CRD](https://ovn-kubernetes.io/api-reference/admin-epbr-api-spec/). Is it possible for you to [replace](https://docs.redhat.com/en/documentation/openshift_container_platform/4.17/html/network_apis/adminpolicybasedexternalroute-k8s-ovn-org-v1) the annotation with the CR and try again?

zeeke:
Yes, using the ABP resources and dropping the annotations would solve the problem. 

BTW, the annotations still have an effect as of now, and the above configuration might break a node. If you prefer not fixing a deprecated use case, I'm fine with it.

PR #4794, opened 2024-10-22
Kv lsp enabled flag

qinqon:

Fixes #

```release-note

```

### DIFF
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -349,3 +349,72 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
 func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
 	return ExtractVMNameFromPod(pod) != nil
 }
+
+func podDuringMigration(vmPods []*corev1.Pod) bool {
+	return len(vmPods) > 0
+}
+
+func isSourcePodAlreadyStale(pod *corev1.Pod, vmPods []*corev1.Pod) bool {
+	if util.PodCompleted(pod) {
+		return true
+	}
+	for _, vmPod := range vmPods {
+		if vmPod.CreationTimestamp.After(pod.CreationTimestamp.Time) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func getTargetPod(vmPods []*corev1.Pod) *corev1.Pod {
+	targetPodNominee := vmPods[0]
+	for i := 1; i < len(vmPods); i++ {
+		if vmPods[i].CreationTimestamp.After(targetPodNominee.CreationTimestamp.Time) {
+			targetPodNominee = vmPods[i]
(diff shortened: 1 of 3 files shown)


go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1178,6 +1187,6 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
 	return bnc.releasedPodsBeforeStartup[nad].Has(uid)
 }
 
-func isAllowedForMigration(isSecondary, isPrimaryNetwork, isl2Topology bool) bool {
-	return isSecondary && !isPrimaryNetwork && isl2Topology
+func (bnc *BaseNetworkController) isAllowedForMigration() bool {
```
maiqueb, line 1190: I'd have this as a function instead. There's no need for the layer3 controller to have this method for instance.

Also, should't localnet be also a "customer" of this feature ?... The exact same mechanism should be re-used there to improve the migration downtime, right ?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -402,19 +376,56 @@ func filterNotComplete(vmPods []*corev1.Pod) []*corev1.Pod {
...
+
+	// Sort vmPods by creation time
+	sort.Slice(vmPods, func(i, j int) bool {
+		// i less than j
+		return vmPods[j].CreationTimestamp.After(vmPods[i].CreationTimestamp.Time)
+	})
+
+	status := LiveMigrationStatus{
+		SourcePod: vmPods[0],
+		TargetPod: vmPods[1],
```
maiqueb, line 423: could we trim down the objects ? IIUC we just require the labels on the pods. Right ?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -402,19 +376,56 @@ func filterNotComplete(vmPods []*corev1.Pod) []*corev1.Pod {
...
-		return false, nil
+	// no migration
+	if len(vmPods) < 2 {
+		return nil, nil
 	}
 
-	if isSourcePodAlreadyStale(pod, vmPods) && isTargetPodReady(getTargetPod(vmPods)) {
-		return true, nil
+	if len(vmPods) > 2 {
+		return nil, fmt.Errorf("unexpected live migration state at pods: %+v", vmPods)
```
maiqueb, line 412: too much info IMHO

go-controller/pkg/ovn/base_network_controller_pods.go
qinqon: Having a function that is going to receive NetInfo is kind of the same also if will be in the same context as layer3, also I am total agains a free function receiving random booleans.

maiqueb: I don't understand what you mean, but FWIW, I also don't want a function that receives random booleans.

PR #4796, opened 2024-10-22
Suppress remote node annotation missing

trozet:
When we go to add a remote node, its likely it is missing one or more anotations on the first couple events. Suppress these errors as we do for other remote node cases.


go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -617,7 +626,11 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
 			var err1 error
 			nodeGRPIPs, err1 = util.ParseNodeGatewayRouterLRPAddrs(node)
 			if err1 != nil {
-				return fmt.Errorf("failed to parse node %s Gateway router LRP Addrs annotation %w", node.Name, err1)
+				err1 = fmt.Errorf("failed to parse node %s Gateway router LRP Addrs annotation %w", node.Name, err1)
```
tssurya, line 629: hmm why are we reconstructing the err again and parsing that.. ? for more info?

trozet: yeah i just left the message intact and only changed suppression

tssurya: done!

PR #4799, opened 2024-10-23
UDN host isolation

npinaeva:
pulls commits from https://github.com/ovn-org/ovn-kubernetes/pull/3709 and https://github.com/ovn-org/ovn-kubernetes/pull/4612
Adds host isolation except for kubelet, open-ports handling for hostnetwork pods.

To enable this feature on a kind cluster, just comment out https://github.com/ovn-org/ovn-kubernetes/pull/4799/commits/71ad044a128ec5c548465963de50581adec89d29#diff-ea48eab674fe05348e580125fccb19b775e0fd6f6a2d28a2dab5fb382590a16aR1641
I have run new e2es on the fedora rawhide VM, which has the fix

https://github.com/ovn-org/ovn-kubernetes/pull/4800 is supposed to show host-isolation test parts failing on a kind with feature disabled


go-controller/pkg/node/udn_isolation.go
```diff
@@ -0,0 +1,459 @@
...
+		return fmt.Errorf("could not update nftables rule for management port: %v", err)
+	}
+	return nil
+}
+
+// runKubeletRestartTracker listens to systemd events to re-apply the UDN host isolation rules after kubelet restart.
+// cgroupv2 match doesn't actually match cgroup paths, but rather resolves them to numeric cgroup IDs when such
+// rules are loaded into kernel, and does not automatically update them in any way afterwards.
+// From the patch https://patchwork.ozlabs.org/project/netfilter-devel/patch/1479114761-19534-1-git-send-email-pablo@netfilter.org/#1511797:
+// If the cgroup is gone, the filtering policy would not match anymore. You only have to subscribe to events
```
trozet, line 291: nice comments

go-controller/pkg/node/udn_isolation.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	return nil
+}
+
+func podNeedsUpdate(oldObj, newObj *v1.Pod) bool {
+	if oldObj == nil || newObj == nil {
+		return true
+	}
+	// react to pod IP changes
+	return !reflect.DeepEqual(oldObj.Status, newObj.Status) || !reflect.DeepEqual(oldObj.Annotations, newObj.Annotations)
```
trozet, line 407: nit: might be a little better to only scope the deep equal on the specific k8s.ovn.org pod annotation

go-controller/pkg/node/udn_isolation.go
```diff
@@ -0,0 +1,459 @@
...
+	if err != nil {
+		// update event should come later with ips
+		klog.V(5).Infof("Failed to get default network pod IPs for pod %s in namespace %s: %v", name, namespace, err)
+		return nil
+	}
+	return m.updateUDNPodIPs(key, podIPs)
+}
+
+func (m *UDNHostIsolationManager) isPodPrimaryUDN(pod *v1.Pod) (bool, error) {
+	activeNetwork, err := m.nadController.GetActiveNetworkForNamespace(pod.Namespace)
```
trozet: since you have the pod, you could look at the annotations to avoid having to use the nad controller for network with role: primary. I think there is already a util method somewhere to parse this for you.

go-controller/pkg/node/udn_isolation.go
```diff
@@ -368,73 +368,106 @@ func (m *UDNHostIsolationManager) updateUDNPodIPs(namespacedName string, podIPs
...
-type nftPodIPSet struct {
+// nftPodElementsSet is a helper struct to manage an nftables set with pod-owned elements.
+// Can be used to store pod IPs, or more complex elements.
+type nftPodElementsSet struct {
 	setName string
-	// podName: ips
-	podsIPs map[string]sets.Set[string]
+	// podName: set elements
+	podElements map[string]sets.Set[string]
+	// podIPs may be reused as soon as the pod reaches Terminating state, and delete event may come later.
```
trozet, line 541: I think this statement is a little backwards. The delete event will come immediately, and OVNK will release the IP without waiting for termination grace period (this is probably a bug). Since the pod informer is serialized you should get the delete pod, before another new pod event can take its IP. I'm not sure what your exact scenario is, but to protect against races with this stuff we use the logical port cache in the default network controller and have an expiry timer to indicate when a pod has been deleted.

go-controller/pkg/node/udn_isolation.go
```diff
@@ -368,11 +493,45 @@ func (m *UDNHostIsolationManager) updateUDNPodIPs(namespacedName string, podIPs
...
+	openPortsv4 = sets.New[string]()
+	openPortsv6 = sets.New[string]()
+
+	for _, openPort := range openPorts {
+		if openPort.Protocol == "icmp" {
+			icmpv4 = newV4IPs
+			icmpv6 = newV6IPs
+		} else {
+			for podIPv4 := range newV4IPs {
+				openPortsv4.Insert(joinNFTSlice([]string{podIPv4, openPort.Protocol, fmt.Sprintf("%d", *openPort.Port)}))
```
trozet, line 525: ok so our openPorts set contains both the protocol and the port

go-controller/pkg/node/udn_isolation.go
```diff
@@ -0,0 +1,459 @@
...
+}
+
+func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
+	tx := m.nft.NewTransaction()
+	tx.Flush(&knftables.Chain{
+		Name: nodenft.UDNIsolationChain,
+	})
+	m.addRules(tx)
+
+	err := m.nft.Run(context.TODO(), tx)
```
martinkennelly, line 280: just flying by but can this block forever since TODO has no limit?

go-controller/pkg/node/udn_isolation.go
```diff
@@ -0,0 +1,459 @@
...
+			case err := <-errChan:
+				klog.Errorf("Systemd listener error: %v", err)
+			}
+		}
+	}()
+	return nil
+}
+
+func (m *UDNHostIsolationManager) podInitialSync() error {
+	pods, err := m.podLister.List(labels.Everything())
```
martinkennelly, line 353: do we need to know about all pods in the cluster?

go-controller/pkg/node/udn_isolation.go
npinaeva: how do I know from the annotation if pod's primary network is UDN? Can you help me find that util function because I can't? :(

go-controller/pkg/node/udn_isolation.go
npinaeva: yeah I think theoretically it can, but in reality it never happens.
cc @danwinship for more thoughts on this

go-controller/pkg/node/udn_isolation.go
npinaeva: yeah I think so, if you mean compared to per-node filtering.
when the host tries to reach a UDN pod on default network (including pods on other nodes), it will go through mp0. UDN ACLs allow all traffic coming from the management port to allow kubelet traffic, because you can't distinguish host processes in the OVN network. So if you want to avoid node reaching pods on other nodes, you have to drop that traffic on the source node, hence the need to watch for all pods.

go-controller/pkg/node/udn_isolation.go
trozet:
```
    k8s.ovn.org/pod-networks: '{"default":{"ip_addresses":["10.244.5.4/24"],"mac_address":"0a:58:0a:f4:05:04","routes":[{"dest":"10.244.0.0/16","nextHop":"10.244.5.1"},{"dest":"100.64.0.0/16","nextHop":"10.244.5.1"}],"ip_address":"10.244.5.4/24","role":"infrastructure-locked"},"test-namespace-9/user-defined-network-9":{"ip_addresses":["10.20.7.3/24"],"mac_address":"0a:58:0a:14:07:03","gateway_ips":["10.20.7.1"],"routes":[{"dest":"10.20.0.0/16","nextHop":"10.20.7.1"},{"dest":"10.96.0.0/16","nextHop":"10.20.7.1"},{"dest":"100.65.0.0/16","nextHop":"10.20.7.1"}],"ip_address":"10.20.7.3/24","gateway_ip":"10.20.7.1","role":"primary"}}'
```

you see the role there. That tells you if it is a UDN or not:

```
	// defined internally by ovnkube to recognize "default"
	// network's role as a "infrastructure-locked" network
	// when user defined network is the primary network for
	// the pod which makes "default" network niether primary
	// nor secondary
	NetworkRoleInfrastructure = "infrastructure-locked"
```

You can unmarshal the pod annotations for the default network and then check role:
```
podAnnotation, err := util.UnmarshalPodAnnotation(pod.Annotations, types.DefaultNetworkName)
if annotation.Role ==NetworkRoleInfrastructure { return true
} else {
activeNetwork, err := m.nadController.GetActiveNetworkForNamespace(pod.Namespace)
}

go-controller/pkg/node/udn_isolation.go
trozet: oh you are right, my bad

go-controller/pkg/node/udn_isolation.go
npinaeva: Oh I thought using `GetActiveNetworkForNamespace` is a new more efficient way of doing that because it doesn't require unmarshalling? Are you saying if you have pod object, you should prefer unmarshalling network annotations to get network info?

npinaeva: UnmarshalPodAnnotation seems to be more reliable, as it dosn't depend on NAD presence (which can actually be deleted). I have skipped `GetActiveNetworkForNamespace` fallback and just returned error as no annotations means I can't extract ips anyways

trozet (approved):

PR #4801, opened 2024-10-23
Remove PreStart from NetworkController interface

jcaamano:
As it is unused in any network controller except the node default network controller.

The future plan is to strip PreStart from the node default network controller as well and move that somewhere else.

The ultimate reason is we want to configure the management ports (SNAT) correctly depending on whether the network is advertised or not:
https://github.com/ovn-org/ovn-kubernetes/pull/4572
On that PR, commit `Don't SNAT to mgmt port if pod network advertised`:
https://github.com/ovn-org/ovn-kubernetes/pull/4572/commits/ec47746eac307caebd0172b02835f2cec610107c

For this we will depend on the NAD controller, even for the default network:
https://github.com/ovn-org/ovn-kubernetes/pull/4533
On that PR, commit `Provide NAD controller access to the default network controller`:
https://github.com/ovn-org/ovn-kubernetes/pull/4533/commits/0792356cc66c843f9cf7d91a7c862f7fbef9af73

So we need to start the NAD controller before the management port is initialized. So we need a way to make the gateway functionality available to network controllers independently of whether we have applied its configuration or not. So, basically, we need to be able to reconcile gateway configuration asynchronously.

### DIFF
--- a/go-controller/pkg/network-controller-manager/node_network_controller_manager.go
+++ b/go-controller/pkg/network-controller-manager/node_network_controller_manager.go
@@ -36,7 +36,7 @@ type nodeNetworkControllerManager struct {
 	wg            *sync.WaitGroup
 	recorder      record.EventRecorder
 
-	defaultNodeNetworkController nad.BaseNetworkController
+	defaultNodeNetworkController *node.DefaultNodeNetworkController
 
 	// net-attach-def controller handle net-attach-def and create/delete secondary controllers
 	// nil in dpu-host mode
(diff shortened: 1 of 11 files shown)


jcaamano:
@ricky-rav @trozet PTAL

go-controller/pkg/network-controller-manager/node_network_controller_manager.go
```diff
@@ -36,7 +36,7 @@ type nodeNetworkControllerManager struct {
 	wg            *sync.WaitGroup
 	recorder      record.EventRecorder
 
-	defaultNodeNetworkController nad.BaseNetworkController
```
ricky-rav, line 39: This is the bit that I didn't question in my PR and which made me add `preStart` to the interface. My bad! Thanks for improving this!

ricky-rav (approved):
/lgtm

trozet (approved):

go-controller/pkg/network-controller-manager/node_network_controller_manager.go
tssurya: same here nice we changed this.. when I added the dereference I wanted to come back to this to find out why DNNC had nad.BNC type...thanks for changing this

PR #4802, opened 2024-10-23
Set shard-conformance timeout lower tham github timeout

jcaamano:
Otherwise github terminates the job on timeout, doesn't give a chance to ginkgo to print status and we don't know why the timeout happened.

### DIFF
--- a/test/scripts/e2e-kind.sh
+++ b/test/scripts/e2e-kind.sh
@@ -170,10 +170,17 @@ export NUM_WORKER_NODES=3
 if [ "$SINGLE_NODE_CLUSTER" == true ]; then
 	export NUM_WORKER_NODES=1
 fi
+
+# Until we know how to make github actions gracefully terminate our tests, this
+# timeout needs to be lower than github's timeout. Otherwise github terminates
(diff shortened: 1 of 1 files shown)


test/scripts/e2e-kind.sh
```diff
@@ -170,10 +170,17 @@ export NUM_WORKER_NODES=3
 if [ "$SINGLE_NODE_CLUSTER" == true ]; then
 	export NUM_WORKER_NODES=1
 fi
+
+# Until we know how to make github actions gracefully terminate our tests, this
+# timeout needs to be lower than github's timeout. Otherwise github terminates
```
tssurya, line 175: is github's timeout: https://github.com/ovn-org/ovn-kubernetes/blob/8551af5a88ee86169b433d9b2edf304440a02438/.github/workflows/test.yml#L530 120mins? hence 100mins selected to be lower?

tssurya (approved):

tssurya:
all shard-conformances have passed, the failure in external-gateways is unrelated.. merging this so that in the future we get a better picture of what's failing

PR #4803, opened 2024-10-24
Remove 5hours of sleeping time from e2e

tssurya:
The past-me thought this was a good idea?? :) not really looks like I sneaked in a 5hour sleep at some point and never removed it...


tssurya:
thanks to @dceara for catching this

dceara (approved):
Looks good to me, thanks!

trozet (approved):

PR #3374, opened 2023-01-24
Control forwarding on OVNK managed interfaces

tssurya:
This PR does two things:

1) We add the following rules to our setup: 
```
	"-d 172.16.1.0/24 -j ACCEPT",
	"-s 172.16.1.0/24 -j ACCEPT",
	"-o breth0 -j DROP",
	"-i breth0 -j DROP",
```
(for primary and secondary bridges) if the DisableForwardingOnInterfaces=true

Problem: When packets are destined towards externalIPs/loadbalancerVIPs against unexposed ports, the packets follow this path: external->br-ex->host->br-ex->external and this causes a loop with the user's gw router. As per CNI plugin requirements, we shouldn't be interfering with this undefined traffic and sending it out if we don't know where to route this since its afterall service traffic. Just like we drop serviceCIDR traffic against unexposed ports, let's add drop rules for this kind of traffic coming from breth0 and getting routed back to breth0. Except the lgw service traffic flows nothing else should be using this path. We allow service CIDR traffic to be forwarded across all interfaces. NOTE: Same as https://github.com/ovn-org/ovn-kubernetes/pull/3346 but for externally exposed services.

Apart from the above problem, hosts are acting as routers between multiple interfaces which is undesired. (See @trozet 's comments below), this PR aims to fix that problem as well. Since we can't break backwards compatibility we also add a new knob to gatewayConfig which can be used to disable forwarding. It's an opt-in behaviour so users can stay on the default mode for backwards compatibility

2) We explicitly enable forwarding on mp0 interface irrespective of the OS settings since we own it.

Signed-off-by: Surya Seetharaman <suryaseetharaman.9@gmail.com>

PR #4805, opened 2024-10-25
Enable IP forwarding on the UDN management interface

kyrtapz:
Enable IP forwarding on the UDN management interface for IPv4 enabled networks. IPv6 forwarding is
globally enabled.
Disable IP forwarding in the SGW UDN lane to catch any potential issues caused by the restricted forwarding.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -433,7 +433,7 @@ jobs:
           - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv4",      "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-disabled", "num-workers": "3"}
           - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "num-workers": "3"}
           - {"target": "control-plane", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4", "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "forwarding": "disable-forwarding"}
-          - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
+          - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "forwarding": "disable-forwarding"}
           - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "local", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW", "second-bridge": "1br", "ic": "ic-disabled"}
           - {"target": "tools", "ha": "noHA", "gateway-mode": "local", "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -11,6 +11,8 @@ import (
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 
+	"github.com/vishvananda/netlink"
+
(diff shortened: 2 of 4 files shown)


trozet (commented):
If disable forwarding is enabled, we also add these iptables rules:

```
// initExternalBridgeForwardingRules sets up iptables rules for br-* interface svc traffic forwarding
// -A FORWARD -s 10.96.0.0/16 -j ACCEPT
// -A FORWARD -d 10.96.0.0/16 -j ACCEPT
// -A FORWARD -s 169.254.169.1 -j ACCEPT
// -A FORWARD -d 169.254.169.1 -j ACCEPT
func initExternalBridgeServiceForwardingRules(cidrs []*net.IPNet) error {
	return insertIptRules(getGatewayForwardRules(cidrs))
}
```

That includes the masquerade IP of OVN.  I cant remember why:

https://github.com/ovn-org/ovn-kubernetes/pull/3374

trozet:
> If disable forwarding is enabled, we also add these iptables rules:
> 
> ```
> // initExternalBridgeForwardingRules sets up iptables rules for br-* interface svc traffic forwarding
> // -A FORWARD -s 10.96.0.0/16 -j ACCEPT
> // -A FORWARD -d 10.96.0.0/16 -j ACCEPT
> // -A FORWARD -s 169.254.169.1 -j ACCEPT
> // -A FORWARD -d 169.254.169.1 -j ACCEPT
> func initExternalBridgeServiceForwardingRules(cidrs []*net.IPNet) error {
> 	return insertIptRules(getGatewayForwardRules(cidrs))
> }
> ```
> 
> That includes the masquerade IP of OVN. I cant remember why:
> 
> #3374

@tssurya pointed me to https://github.com/ovn-org/ovn-kubernetes/pull/3374#discussion_r1153327332

I think my comment there was wrong...or at least we need to walk-through this path again. So .1 is the OVN masq ip. It is only used OVN is going to the host. One possible path is nodePort, and assume hairpin back to host networked pod:
1. external ---> host ---> DNAT (cluster IP) --> forwarding rules --->br-ex --> ovn
2. ovn ---> DNAT (pod ip) ---> SNAT( 169.254.169.1) ----> br-ex --->host--->host networked pod

In this case there is no forwarding in 2, because the destination is the localhost. On the reply when the host networked pod sends back to OVN 169.254.169.1, the host is directly attached to the 169.254.169.0/24 subnet via breth0 (in this case 169.254.0.0/17):

```
5: breth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN group default qlen 1000
    link/ether 02:42:ac:12:00:03 brd ff:ff:ff:ff:ff:ff
    inet 172.18.0.3/16 brd 172.18.255.255 scope global breth0
    inet 169.254.0.2/17 brd 169.254.127.255 scope global breth0

```

Therefore what I said was incorrect, there wont be any route lookup for masq .4, because the host is already attached to this subnet. It would normally ARP for .1, but in this case we install static mac entries:
```
root@ovn-worker:/# ip neighbor show 169.254.0.1
169.254.0.1 dev breth0 lladdr 0a:58:a9:fe:00:01 PERMANENT
```
So there is no forwarding happening here as this was a locally generated packet. I don't see a case where 169.254.169.1 is forwarded anywhere. The only way I could think of this happening is if something send by OVN to the host is then routed into mp0 or an mpx, which I dont thinks is possible.

It is possible for mpx masq ip (UDN source IP) to enter into the host and need to be forwarded to 169.254.0.4. But we are already covered there with the rules for services:

```
> // -A FORWARD -s 10.96.0.0/16 -j ACCEPT
> // -A FORWARD -d 10.96.0.0/16 -j ACCEPT
```

So long story short, I think this PR is OK and we don't need to add masq ips for UDNs. We can probably just leave the ones that are there for default if we are scared, but we could also remove them. I would like @kyrtapz or @tssurya to double check my logic here or tell me if I'm missing some other use case.

tssurya:
The past me and tim were not stupid it seems like :) 

@kyrtapz and I looked into this

There is the LGW service scenario (the e2e's andreas added for LGW services)
src = external client or ovnkube-control-plane node
dst = ovn-worker2:nodePort
backend = ovn-worker host-networked pod

Here we go from external to ovn-worker2's br-ex -> host -> DNAT to clsuterIP -> route to br-ex again -> takes it into OVN -> DNATs to ovn-worker IP -> comes back into br-ex (SNATs here to .1 IP) -> sends it to the host -> HERE HOST SHOULD BE ACTING AS THE ROUTER so forwarding will be hit with srcIP == .1 because the src is not local and dst here is also not local. This aligns with what @danwinship correctly pointed out here: https://github.com/ovn-org/ovn-kubernetes/pull/3374#issuecomment-1404103125

@kyrtapz confirmed this by seeing the counter of the FORWARDING rules go up as he ran this test locally.

But for UDN's given this scenario is not going to be a thing anyways - we don't even have host-net pod support yet maybe for now we are good here.

tssurya (approved):

trozet:
> The past me and tim were not stupid it seems like :)

maybe I'm just getting dumber :)

The example you showed is indeed forwarding with .1

As for should we support this with UDN...I guess to do this we would need to need to go via geneve to the mpx port of the udn. That is the only way to preserve the VRF and get to the port where the host networked pod is attached to. So:

external ----->nodeport worker 2-----DNAT 169.254.0.x ----> SNAT join IP =====geneve====worker 1 ----mpx -----host networked pod with a socket on mpx vrf

Maybe something to build on after I finish: https://github.com/ovn-org/ovn-kubernetes/pull/4069

PR #4811, opened 2024-10-28
Adds startup probe to UDN e2e

trozet:
It was seen in downstream that this container can take longer than a second for its http server to start sometimes:

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2314/pull-ci-openshift-ovn-kubernetes-master-e2e-gcp-ovn-techpreview/1849904878739001344

Therefore add a startup probe to give it a few seconds to fully start, then run liveness probe. The test doesn't check for at least 8 seconds (claimed in the comments) that the container never restarted, so this should be fine to finish start up probe and get a few iterations of liveness probe in.

### DIFF
--- a/test/e2e/network_segmentation.go
+++ b/test/e2e/network_segmentation.go
@@ -261,6 +261,17 @@ var _ = Describe("Network Segmentation", func() {
 								PeriodSeconds:       1,
 								FailureThreshold:    1,
 							}
+							pod.Spec.Containers[0].StartupProbe = &v1.Probe{
+								ProbeHandler: v1.ProbeHandler{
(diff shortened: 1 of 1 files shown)


tssurya:
ugh why is tools lane flaking..

kyrtapz (approved):
LGTM
A pod will not become ready until the startup probe and the readiness probe succeeds so we are good on that front.
This change allows for up to 2 startup checks to fail but I think that is reasonable to allow some more time to start.

tssurya (approved):

PR #4813, opened 2024-10-31
OCPBUGS-43454: Ignore cluster manager `topology not managed` errors for localnet  with no subnets

maiqueb:

This PR ignores `topology net managed` errors on the initial all networks sync of the cluster manager controllers.

Only the `secondaryNetworkClusterManager` - which is created by the cluster controller manager, and executed in the control plane pods - actually returns this type of error.

Fixes #

#### How to verify it

1. Configure the bridge mappings for the localnet network
2. Create a localnet network without subnets.
3. Attach a pod to the localnet network.
4. Restart the ovnkube-control-plane pods

To configure the mappings for the localnet network you can (upstream):
```sh
for pod in $(kubectl get pods -novn-kubernetes -l "app=ovs-node" --no-headers -ocustom-columns=":metadata.name"); do 
    kubectl -n ovn-kubernetes exec $pod -- ovs-vsctl set open . external-ids:ovn-bridge-mappings="physnet:breth0,tenantblue:breth0"
done
```

The following manifests can be used to reproduce this upstream:
```yaml
---
metadata:
  name: tenantblue
spec:
    config: '{
        "cniVersion": "0.4.0",
        "name": "tenantblue",
        "netAttachDefName": "default/tenantblue",
        "topology": "localnet",
        "logFile": "/var/log/ovn-kubernetes/ovn-k8s-cni-overlay.log",
        "logLevel": "5",
        "logfile-maxsize": 100,
        "logfile-maxbackups": 5,
        "logfile-maxage": 5
    }'
---
apiVersion: v1                
kind: Pod                        
metadata:
  namespace: default
  name: pod1
  annotations:
    k8s.v1.cni.cncf.io/networks: '[
      {
        "name": "tenantblue","ips": ["192.168.200.20/24"]
      }                                          
    ]'                         
spec:                                 
  containers:                                                        
  - name: centos        
    image: docker.io/centos/tools:latest
    command:                    
    - /sbin/init           
```

Without this PR, the control-plane pods will crash loop.

Prevent a localnet network without subnets (typical virt use case) from crash-looping the control plane pods when they restart (for instance, during an upgrade).

```release-note
Prevent a localnet network without subnets from crash-looping the control plane pods upon restart (e.g. during an upgrade)
```

### DIFF
--- a/go-controller/pkg/network-attach-def-controller/network_manager.go
+++ b/go-controller/pkg/network-attach-def-controller/network_manager.go
@@ -2,6 +2,7 @@ package networkAttachDefController
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -214,7 +215,13 @@ func (nm *networkManagerImpl) syncAll() error {
 	start := time.Now()
 	klog.Infof("%s: syncing all networks", nm.name)
 	for _, network := range validNetworks {
-		if err := nm.sync(network.GetNetworkName()); err != nil {
+		if err := nm.sync(network.GetNetworkName()); errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
+			klog.V(5).Infof(
+				"ignoring network %q since %q does not manage it",
+				network.GetNetworkName(),
+				nm.name,
+			)
+		} else if err != nil {
 			return fmt.Errorf("failed to sync network %s: %w", network.GetNetworkName(), err)
 		}
 	}
--- a/go-controller/pkg/clustermanager/secondary_network_cluster_manager.go
+++ b/go-controller/pkg/clustermanager/secondary_network_cluster_manager.go
@@ -68,7 +68,7 @@ func (sncm *secondaryNetworkClusterManager) SetNetworkStatusReporter(errorReport
 	sncm.errorReporter = errorReporter
 }
 
-// Start the secondary layer3 controller, handles all events and creates all
+// Start the secondary network controller, handles all events and creates all
 // needed logical entities
 func (sncm *secondaryNetworkClusterManager) Start() error {
 	klog.Infof("Starting secondary network cluster manager")


go-controller/pkg/network-attach-def-controller/network_manager.go
```diff
@@ -214,7 +215,13 @@ func (nm *networkManagerImpl) syncAll() error {
 	start := time.Now()
 	klog.Infof("%s: syncing all networks", nm.name)
 	for _, network := range validNetworks {
-		if err := nm.sync(network.GetNetworkName()); err != nil {
+		if err := nm.sync(network.GetNetworkName()); errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
```
tssurya, line 218: do we have a unit test for localnet with no subnets?

tssurya:
do we have a unit test for localnet with no subnets?

go-controller/pkg/network-attach-def-controller/network_manager.go
tssurya, line 218: wow learnt something new here, so from code perspective, its a localnet network of type ovn-k8s-cni-overlay with  no subnets meaning its pure l2 usecase ? OR something else does IPAM for those VM pods and they talk to the VLAN outside?

so this means we don't create any NAD controller for such networks? hence ErrNetworkControllerTopologyNotManaged ?

maiqueb: No, we have one, but running in the ovnkube-node.

maiqueb:
> do we have a unit test for localnet with no subnets?

I can add a test case that ensure this error is thrown for this network, but I'm not sure if I can add more than that.

go-controller/pkg/network-attach-def-controller/network_manager.go
tssurya: Hmm why do we create one in the node side? Was it for dpu use case?

tssurya: what does that controller do on the node side? is it for the DPU case?

jcaamano (approved):

PR #4818, opened 2024-11-05
Observability: change decode funcs to provide structured data

### LINKED ISSUE #4817: Observability needs to provide structured data

Currently in `observability-lib`, the sample decoder only provides a human-readable string representation of the network events (such as: "Allowed by network policy foo, direction Ingress")
If consumers need to do more than just displaying this string, they should receive structured data instead ; and the string representation can be derived from this data.

### Why is this needed?

Here's 2 use cases that we have with network observability:

1. If consumers want to produce labelled metrics out of the decoded events, without having to parse the string to extract meaningful data such as the network policy name or direction. E.g. a prometheus metric such as `network_policy_allowed` with labels `name`, `pod`, `namespace`.
2. Another use case is when consumer needs to take specific action depending on the type of event, for instance, a "dropped by policy" type of event correlated with packet counters means that those packets are dropped. In Network Observability, dropped packets are represented in a specific way.

An alternative to structured data could be to have the consumers parse the string a extract data, but:
- this has a performance cost, not acceptable in the case of network observability
- it would be error prone e.g. in case the string representations change over time

jotak:

This change allows consumers to get structured data instead of pre-built strings. A first use-case (that we have with Network Observability) is if consumers want to produce metrics out of the decoded events, without having to parse the string to extract meaningful data such as the network policy name or direction.
Another use case if when consumer needs to take specific action depending on the type of event, for instance, a "dropped by policy" type of event correlated with packet counters means that those packets are dropped.

If the consumers need to do more than just display strings, they need to get structured data.

The previous string representation is still accessible by calling `String()` method from the resulting objects.

Fixes #4817 

```release-note

```


jotak:
cc @npinaeva

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -133,74 +135,60 @@ func (d *SampleDecoder) DecodeCookieIDs(obsDomainID, obsPointID uint32) (string,
...
+		event.Namespace = o.ExternalIDs[libovsdbops.ObjectNameKey.String()]
+		event.Direction = string(libovsdbutil.ACLEgress)
+	case libovsdbops.UDNIsolationOwnerType:
+		event.Name = o.ExternalIDs[libovsdbops.ObjectNameKey.String()]
 	}
+	return &event
 }
 
-func (d *SampleDecoder) DecodeCookieBytes(cookie []byte) (string, error) {
+func (d *SampleDecoder) DecodeCookieBytes(cookie []byte) (model.NetworkEvent, error) {
```
jotak, line 191: I'm purposely doing a breaking change here (since it's not GA feature yet, it's still time to break, right?)
@npinaeva can you point me to other consumers than netobserv where I should adapt code?

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -133,74 +135,60 @@ func (d *SampleDecoder) DecodeCookieIDs(obsDomainID, obsPointID uint32) (string,
...
-		return ""
+func newACLEvent(o *nbdb.ACL) *model.ACLEvent {
+	actor := o.ExternalIDs[libovsdbops.OwnerTypeKey.String()]
+	event := model.ACLEvent{
+		Action: o.Action,
+		Actor:  actor,
+	}
+	switch actor {
+	case libovsdbops.AdminNetworkPolicyOwnerType, libovsdbops.BaselineAdminNetworkPolicyOwnerType, libovsdbops.NetworkPolicyOwnerType:
+		event.Name = o.ExternalIDs[libovsdbops.ObjectNameKey.String()]
```
npinaeva, line 175: this part for network policy is confusing, because it has both namespace and name

npinaeva (commented):
you also may want to update https://github.com/ovn-kubernetes/ovn-kubernetes/blob/192138e10acfdebde9c0a568b0be89bebef54b6a/go-controller/observability-lib/parse_sample.go#L208 to call `.String()` method on returned event

go-controller/observability-lib/sampledecoder/sample_decoder.go
jotak: do you mean, AdminNetworkPolicyOwnerType / BaselineAdminNetworkPolicyOwnerType / NetworkPolicyOwnerType ?
Previously each of them only mentioned name and direction in their associated message, that's why I did not mention a namespace. You're saying there should be a namespace there too ?

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -133,90 +134,76 @@ func (d *SampleDecoder) DecodeCookieIDs(obsDomainID, obsPointID uint32) (string,
...
+	case libovsdbops.MulticastNamespaceOwnerType, libovsdbops.NetpolNamespaceOwnerType:
+		event.Namespace = o.ExternalIDs[libovsdbops.ObjectNameKey.String()]
+		event.Direction = o.ExternalIDs[libovsdbops.PolicyDirectionKey.String()]
+	case libovsdbops.MulticastClusterOwnerType:
+		event.Direction = o.ExternalIDs[libovsdbops.PolicyDirectionKey.String()]
+	case libovsdbops.EgressFirewallOwnerType:
+		event.Namespace = o.ExternalIDs[libovsdbops.ObjectNameKey.String()]
+		event.Direction = "Egress"
+	case libovsdbops.UDNIsolationOwnerType:
+		event.Name = o.ExternalIDs[libovsdbops.ObjectNameKey.String()]
```
msherif1234, line 185: this function doesn't handle `NetpolNodeOwnerType` I think u can statically populate name, action and direction

jotak: yes it does, in the first case: 
```golang
case libovsdbops.AdminNetworkPolicyOwnerType, libovsdbops.BaselineAdminNetworkPolicyOwnerType, libovsdbops.NetworkPolicyOwnerType
```

jotak: In fact `NetpolNodeOwnerType` should work and has Action and Actor already populated (see lines 169-170), it just that it hasn't got other extra variables. I can add direction indeed.

go-controller/observability-lib/sampledecoder/sample_decoder.go
jotak: better like that: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4818/commits/bf1db3464aaca129bc926d88dc1cd876689a3315 ?

npinaeva: it is called "name" in the externalIDs, but for netpol it contains namespace+name. Since externalIDs don't distinguish namespace vs name, it kinda makes sense, but if your struct has separate namespace and name, then it probably should be parsed

go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -133,90 +135,88 @@ func (d *SampleDecoder) DecodeCookieIDs(obsDomainID, obsPointID uint32) (string,
...
+		Actor:  actor,
+	}
+	switch actor {
+	case libovsdbops.NetworkPolicyOwnerType:
+		nsname := strings.SplitN(o.ExternalIDs[libovsdbops.ObjectNameKey.String()], ":", 2)
+		if len(nsname) == 2 {
+			event.Namespace = nsname[0]
+			event.Name = nsname[1]
+		} else {
+			// do not split (which use cases?)
```
npinaeva, line 180: this should never happen, you can return an error is it does

jotak: done in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4818/commits/b8ab657385ff1c6c21c80ed998cbb6cdaa59af3b

PR #4823, opened 2024-11-06
ipam: Fix init flow in case there are sticky ips in the system

oshoval:
## 📑 Description

In case the `ovnkube-control-plane` pod is restarted,
(for example, node restart, upgrade, or any pod restart reason),
the ipam claim Sync function creates an array of all the ipams
belonging to a specific network, and used named allocator `AllocateIPs`
method to allocate them, so the allocator will reflect the current
cluster state.

The problem is `AllocateIPs` allows to allocate only one IP per given subnet,
so once it is used for all the IPs in the network at once, it will fail,
causing the control plane to have a crash loop.

Fix it by calling `AllocateIPs` per each claim on its own.
The claims were already created correctly, so each claim on its
own is safe to call `AllocateIPs`.

Add unit test to show the relevant assertion.

Fixes #

Seen by deleting the pod while there were 2 ipam claims with IPs, from the same NAD
in the system.

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/allocator/ip/subnet/allocator_test.go
+++ b/go-controller/pkg/allocator/ip/subnet/allocator_test.go
@@ -5,12 +5,14 @@ import (
 
 	ipam "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/ip"
 	ovntest "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 )
 
 var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
+	const subnetName = "subnet1"
 	var (
 		allocator Allocator
 	)
@@ -21,7 +23,6 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 
 	ginkgo.Context("when adding subnets", func() {
 		ginkgo.It("creates each IPAM and reserves IPs correctly", func() {
-			subnetName := "subnet1"
 			subnets := []string{
 				"10.1.1.0/24",
 				"2000::/64",
@@ -40,7 +41,6 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 		})
 
 		ginkgo.It("handles updates to the subnets correctly", func() {
-			subnetName := "subnet1"
 			subnets := []string{
 				"10.1.1.0/24",
 				"2000::/64",
@@ -69,7 +69,6 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 		})
 
 		ginkgo.It("excludes subnets correctly", func() {
-			subnetName := "subnet1"
 			subnets := []string{
 				"10.1.1.0/24",
 			}
@@ -93,7 +92,6 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 
 	ginkgo.Context("when allocating IP addresses", func() {
 		ginkgo.It("IPAM for each subnet allocates IPs contiguously", func() {
-			subnetName := "subnet1"
 			subnets := []string{
 				"10.1.1.0/24",
 				"2000::/64",
@@ -116,7 +114,6 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 		})
 
 		ginkgo.It("IPAM allocates, releases, and reallocates IPs correctly", func() {
-			subnetName := "subnet1"
 			subnets := []string{
 				"10.1.1.0/24",
 			}
@@ -141,8 +138,19 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 			}
 		})
 
+		ginkgo.It("fails to allocate multiple IPs from the same subnet", func() {
+			subnets := []string{"10.1.1.0/24", "2000::/64"}
+
+			gomega.Expect(allocator.AddOrUpdateSubnet(subnetName, ovntest.MustParseIPNets(subnets...))).To(gomega.Succeed())
+
+			ips, err := util.ParseIPNets([]string{"10.1.1.1/24", "10.1.1.2/24"})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			gomega.Expect(allocator.AllocateIPs(subnetName, ips)).To(gomega.MatchError(
+				"failed to allocate IP 10.1.1.2 for subnet1: attempted to reserve multiple IPs in the same IPAM instance",
+			))
+		})
+
 		ginkgo.It("releases IPs for other subnets when any other subnet allocation fails", func() {
-			subnetName := "subnet1"
 			subnets := []string{
 				"10.1.1.0/24",
 				"10.1.2.0/29",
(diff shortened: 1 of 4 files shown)


go-controller/pkg/allocator/ip/subnet/allocator.go
```diff
@@ -159,6 +161,41 @@ func (allocator *allocator) AllocateUntilFull(name string) error {
 	return nil
 }
 
+// ReserveIPs reserves IPs that are already exist in the cluster. S
+// Should be used for Sync only.
+// We ignore the `IsErrAllocated`, because in case of an error it can reach this function again in reconcile.
+// We break upon any other error, but we don't release IPS.
+// Claims have the IPs, so in case claim is deleted, IP will be released.
+// This way we give another change in the reconcile in case of an error.
```
oshoval, line 169: We can take another approach which is all or nothing in the sync, fail in any error, and revert by releasing all what was allocated
since we are under a lock it should be fine
question is what is easier to maintain and simpler

oshoval:
Added unit test that shows the current `AllocateIPs` behavior when trying to allocate several IPs from the same subnet
This behavior is the problematic one for the current Sync logic

go-controller/pkg/allocator/ip/subnet/allocator_test.go
```diff
@@ -141,6 +142,24 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
...
+			subnets := []string{"10.1.1.0/24"}
+
+			gomega.Expect(allocator.AddOrUpdateSubnet(subnetName, ovntest.MustParseIPNets(subnets...))).To(gomega.Succeed())
+
+			var ips []*net.IPNet
+			for _, cidr := range []string{"10.1.1.1/24", "10.1.1.2/24"} {
+				ip, ipNet, _ := net.ParseCIDR(cidr)
+				ipNet.IP = ip
+				ips = append(ips, ipNet)
+			}
```
maiqueb, line 156: let's move this into an helper ?

go-controller/pkg/allocator/ip/subnet/allocator_test.go
```diff
@@ -141,6 +142,24 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 			}
 		})
 
+		ginkgo.It("fails to allocate multiple IPs from the same subnet", func() {
+			subnetName := "subnet1"
```
maiqueb, line 146: const ?

maiqueb (dismissed):
Looks good.

maiqueb:
Maybe we could go a step further and add more unit test that asserts that you can indeed invoke sync for multiple claims:
- ipam claim reconciler level: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/cc3c784bda1fe46d8e84221aec4d96b906d582ed/go-controller/pkg/persistentips/allocator_test.go#L110
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e492123c87e051cdfa1ec996efebad29de8739e3/go-controller/pkg/clustermanager/secondary_network_unit_test.go#L469 (I think this would reproduce if you simply just pre-provision yet another claim in the claim fake client).

@oshoval

oshoval:
Addressed comments
Thanks

go-controller/pkg/persistentips/allocator.go
```diff
@@ -169,11 +169,11 @@ func (icr *IPAMClaimReconciler) Sync(objs []interface{}, ipAllocator IPAllocator
...
 		}
-		ips = append(ips, ipnets...)
-	}
-	if len(ips) > 0 {
-		if err := ipAllocator.AllocateIPs(ips); err != nil && !ipam.IsErrAllocated(err) {
-			return fmt.Errorf("failed allocating persistent ips: %w", err)
+
+		if len(ipnets) != 0 {
+			if err := ipAllocator.AllocateIPs(ipnets); err != nil && !ipam.IsErrAllocated(err) {
+				return fmt.Errorf("failed allocating persistent ips: %w", err)
```
qinqon, line 175: Let's refernece the function so we differenciate with normal allocations.
```suggestion
				return fmt.Errorf("failed syncing persistent ips: %w", err)
```

go-controller/pkg/allocator/ip/subnet/allocator_test.go
```diff
@@ -141,8 +138,18 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 			}
 		})
 
+		ginkgo.It("fails to allocate multiple IPs from the same subnet", func() {
+			subnets := []string{"10.1.1.0/24"}
```
qinqon, line 142: Let's test it with dual stack to be sure ips at different families don't trigger the guard 
```suggestion
			subnets := []string{"10.1.1.0/24",
				"2000::/64"}
```

go-controller/pkg/allocator/ip/subnet/allocator_test.go
```diff
@@ -215,3 +221,13 @@ func TestSubnetIPAllocator(t *testing.T) {
...
+
+func parseCIDRs(cidrs []string) []*net.IPNet {
+	var ips []*net.IPNet
+	for _, cidr := range cidrs {
+		ip, ipNet, _ := net.ParseCIDR(cidr)
+		ipNet.IP = ip
+		ips = append(ips, ipNet)
+	}
+	return ips
+}
```
qinqon, line 233: There is a `util.ParseIPNets` for this

qinqon (changes requested):
I don't see a Sync test, just allocators.

go-controller/pkg/clustermanager/secondary_network_unit_test.go
maiqueb, line 1: Yup, I confirm this test, which can be seen as an integration test with the controllers, fails without the code changes you've introduced. 

Nice !

go-controller/pkg/persistentips/allocator.go
maiqueb: I don't follow. AFAIU, this log is not repeated anywhere. 

Hence, differentiating it is a no-brainer.

oshoval: indeed it exists just here

qinqon: It's just that this is not "allocating" since there were already assigned, this is syncing them, to make even more clear even if this is the only place.

go-controller/pkg/clustermanager/secondary_network_unit_test.go
```diff
@@ -387,6 +389,7 @@ var _ = ginkgo.Describe("Cluster Controller Manager", func() {
 							KubeClient: fake.NewSimpleClientset(),
 							IPAMClaimsClient: fakeipamclaimclient.NewSimpleClientset(
 								ipamClaimWithIPAddr(claimName, namespace, networkName, subnetIP),
+								ipamClaimWithIPAddr(claimName2, namespace, networkName, subnetIP2),
```
qinqon, line 392: Ahh this is the sync test.... then we good.

oshoval:
addressed comments
thanks

go-controller/pkg/persistentips/allocator.go
maiqueb: We swallow that already allocated error ...

Not that I care enough to fight for preserving the log.

oshoval: changed in the last push to syncing

qinqon (approved):
/lgtm

oshoval:
rebased

trozet (approved):
we should change this AllocateIPs function to make it obvious to developers that this really is Allocate one IP per subnet. @maiqueb agreed to handle that in a follow up (thanks Miguel!)

PR #4825, opened 2024-11-08
Fixes unexpected mp0 route removal during start up

trozet:
Routes via mp0 were being deleted on every ovnkube-node restart: [root@ovn-worker ~]# ip monitor route
Deleted 192.72.3.0/24 dev ovn-k8s-mp0 proto kernel scope link src 192.72.3.2 Deleted broadcast 192.72.3.255 dev ovn-k8s-mp0 table local proto kernel scope link src 192.72.3.2 Deleted local 192.72.3.2 dev ovn-k8s-mp0 table local proto kernel scope host src 192.72.3.2 local 192.72.3.2 dev ovn-k8s-mp0 table local proto kernel scope host src 192.72.3.2 broadcast 192.72.3.255 dev ovn-k8s-mp0 table local proto kernel scope link src 192.72.3.2

This causes traffic outage during upgrade, as well as other unwanted side effects when pod-destined traffic is routed via default gateway route in the host. This is especially disruptive in local gateway mode.

This patch removes the teardown, and then makes the synchronization of addresses and routes more robust, so that we can safely handle changes to MTU or mp0 addresses.

### DIFF
--- a/go-controller/pkg/node/management-port_linux.go
+++ b/go-controller/pkg/node/management-port_linux.go
@@ -160,48 +160,27 @@ func tearDownInterfaceIPConfig(link netlink.Link, ipt4, ipt6 util.IPTablesHelper
 	return nil
 }
 
-func tearDownManagementPortConfig(mpcfg *managementPortConfig) error {
-	// for the initial setup we need to start from the clean slate, so flush
-	// all (non-LL) addresses on this link, routes through this link, and
-	// finally any IPtable rules for this link.
-	var ipt4, ipt6 util.IPTablesHelper
-
-	if mpcfg.ipv4 != nil {
-		ipt4 = mpcfg.ipv4.ipt
-	}
-	if mpcfg.ipv6 != nil {
-		ipt6 = mpcfg.ipv6.ipt
-	}
-	return tearDownInterfaceIPConfig(mpcfg.link, ipt4, ipt6)
-}
-
 func setupManagementPortIPFamilyConfig(routeManager *routemanager.Controller, mpcfg *managementPortConfig, cfg *managementPortIPFamilyConfig) ([]string, error) {
 	var warnings []string
 	var err error
 	var exists bool
 
-	if exists, err = util.LinkAddrExist(mpcfg.link, cfg.ifAddr); err == nil && !exists {
-		// we should log this so that one can debug as to why addresses are
-		// disappearing
-		warnings = append(warnings, fmt.Sprintf("missing IP address %s on the interface %s, adding it...",
-			cfg.ifAddr, mpcfg.ifName))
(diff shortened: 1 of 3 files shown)


go-controller/pkg/node/management-port_linux.go
```diff
@@ -160,48 +160,34 @@ func tearDownInterfaceIPConfig(link netlink.Link, ipt4, ipt6 util.IPTablesHelper
...
 		if err != nil {
 			return warnings, err
 		}
-		if exists {
-			continue
+		// Check MTU manually since netlink doesn't support it as an RT_FILTER
+		if route != nil {
+			if route.MTU == config.Default.RoutableMTU {
+				continue
+			}
```
jcaamano, line 185: Don't we need to add the route to the route manager anyway? Isn't it in charge of keeping routes in sync?

go-controller/pkg/util/net_linux.go
```diff
@@ -416,9 +477,9 @@ func LinkRouteGetFilteredRoute(routeFilter *netlink.Route, filterMask uint64) (*
 }
 
 // LinkRouteExists checks for existence of routes for the given subnet through gwIPStr
-func LinkRouteExists(link netlink.Link, gwIP net.IP, subnet *net.IPNet) (bool, error) {
+func LinkRouteExists(link netlink.Link, gwIP net.IP, subnet *net.IPNet) (*netlink.Route, error) {
```
jcaamano, line 480: Maybe rename to `LinkRouteGetByDstAndGw`

go-controller/pkg/util/net_linux.go
```diff
@@ -249,6 +249,67 @@ func LinkAddrFlush(link netlink.Link) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to list addresses for the link %s: %v",
+			link.Attrs().Name, err)
+	}
+
+	// desired addresses - true if already exist
+	matched := map[*net.IPNet]bool{}
+	for _, desiredAddr := range addresses {
+		matched[desiredAddr] = false
+	}
```
jcaamano, line 276: nits: 

you can build this map in the iteration you are already doing above.

but actually it might just be cleaner to build a `map[string]*util.IPNet` of desired addresses. And then handle the difference between desired and actual comparing two extra `k8s.io/apimachinery/pkg/util/sets`.

go-controller/pkg/node/management-port_linux.go
trozet: good point...I was investigating this on an older version of code. In <=4.13 we will need to do something like this, but for later releases we can just rely on routeManager

go-controller/pkg/util/net_linux.go
trozet: i dont think its worth my time. this code only executes once on start up

jcaamano (approved):

PR #4831, opened 2024-11-11
kind, ic, single node zone: Improve redeploy speed

oshoval:
## 📑 Description

By using `maxUnavailable`, we can make the redeploy faster.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -901,6 +901,7 @@ install_ovn_single_node_zones() {
   fi
   run_kubectl apply -f ovnkube-control-plane.yaml
   run_kubectl apply -f ovnkube-single-node-zone.yaml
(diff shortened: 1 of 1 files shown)


contrib/kind.sh
```diff
@@ -901,6 +901,7 @@ install_ovn_single_node_zones() {
   fi
   run_kubectl apply -f ovnkube-control-plane.yaml
   run_kubectl apply -f ovnkube-single-node-zone.yaml
+  kubectl patch ds -n ovn-kubernetes ovnkube-node --type='json' -p='[{"op": "add", "path": "/spec/updateStrategy/rollingUpdate", "value": {"maxUnavailable": "100%"}}]'
```
oshoval, line 904: can use here run_kubectl just to be aligend, but we don't need retry basically

maiqueb (approved):

PR #4832, opened 2024-11-12
node network controller: Add CUDN check to GetActiveNetworkForNamespace

ormergi:
## 📑 Description

Following #4612 this PR add support for the NAD controller to take in account CUDN objects when looking for active namespace in a namespace, similar to the UDN case introduced by  https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4662.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/factory/mocks/NodeWatchFactory.go
+++ b/go-controller/pkg/factory/mocks/NodeWatchFactory.go
@@ -200,6 +200,26 @@ func (_m *NodeWatchFactory) AddServiceHandler(handlerFuncs cache.ResourceEventHa
 	return r0, r1
 }
 
+// ClusterUserDefinedNetworkInformer provides a mock function with given fields:
+func (_m *NodeWatchFactory) ClusterUserDefinedNetworkInformer() userdefinednetworkv1.ClusterUserDefinedNetworkInformer {
+	ret := _m.Called()
+
+	if len(ret) == 0 {
+		panic("no return value specified for ClusterUserDefinedNetworkInformer")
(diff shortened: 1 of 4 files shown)


npinaeva:
This follows current logic, but looks fairly expensive to me to list all UDNs and CUDNs for every default network pod.
@trozet do you have a plan to optimize this?

ormergi:
> This follows current logic, but looks fairly expensive to me to list all UDNs and CUDNs for every default network pod. @trozet do you have a plan to optimize this?

Asking to understand, is it that bad even when informers are used?

trozet:
> > This follows current logic, but looks fairly expensive to me to list all UDNs and CUDNs for every default network pod. @trozet do you have a plan to optimize this?

That only happens if it isnt in the nadcontroller cache. It's a fallback mechanism. When we implement a safeguard so that the namespace has a magical annotation to indicate if UDN is enabled for this namespace, then we can short circuit if the annotation is missing, then we know its default network. In other cases where UDN is desired, but the nad controller cache is somehow missing the info, we fallback to this check. This shouldn't happen in a typical case, but it can happen during object delete path and NAD is already removed. For this case I have been commenting in PRs not not to use getActiveNetworkForNamespace if possible.

> 
> Asking to understand, is it that bad even with informers?

Yeah even though its fast with informers, its still iterating over tons of objects. Since this code is hit so many times it adds up, so we want it to be as fast as possible.

npinaeva (approved):

PR #4438, opened 2024-06-13
EgressIP support on user defined networks

pperiyasamy:
This commit adds an enhancement proposal to support EgressIP for the user defined segmentation networks which serves as a primary network for the pods.

PR #4833, opened 2024-11-13
L2 UDN: EgressIP hosted by primary interface (`breth0`)

martinkennelly:
Please read enhancement before review: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4438
Depends on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4530

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -1181,6 +1181,18 @@ func (e *EgressIPController) SyncLocalNodeZonesCache() error {
 	return nil
 }
 
+// getALocalZoneNodeName fetches the first local OVN zone Node. Support for multiple Nodes per OVN zone is not supported
+// and neither is changing a Nodes OVN zone. This function supports said assumptions.
+func (e *EgressIPController) getALocalZoneNodeName() (string, error) {
+	nodeNames := e.nodeZoneState.GetKeys()
+	for _, nodeName := range nodeNames {
+		if isLocal, ok := e.nodeZoneState.Load(nodeName); ok && isLocal {
+			return nodeName, nil
+		}
+	}
+	return "", fmt.Errorf("failed to find a local OVN zone Node")
+}
+
 func (e *EgressIPController) syncStaleAddressSetIPs(egressIPCache egressIPCache) error {
 	for _, networkPodCache := range egressIPCache.egressIPNameToPods {
 		for networkName, podCache := range networkPodCache {
@@ -1209,12 +1221,12 @@ func (e *EgressIPController) syncStaleAddressSetIPs(egressIPCache egressIPCache)
 }
 
 // syncStaleGWMarkRules removes stale or invalid LRP that packet mark. They are attached to egress nodes gateway router.
-// It adds expected LRPs that packet mark.
+// It adds expected LRPs that packet mark. This is valid only for L3 networks.
 func (e *EgressIPController) syncStaleGWMarkRules(egressIPCache egressIPCache) error {
 	// Delete all stale LRPs then add missing LRPs
 	// This func assumes one node per zone. It determines if an LRP is a valid local LRP. It doesn't determine if the
 	// LRP is attached to the correct GW router
-	if !util.IsNetworkSegmentationSupportEnabled() || !config.OVNKubernetesFeature.EnableInterconnect {
+	if !isEgressIPForUDNSupported() {
 		return nil
 	}
 	for _, networkPodCache := range egressIPCache.egressIPNameToPods {
@@ -1426,9 +1438,18 @@ func (e *EgressIPController) syncPodAssignmentCache(egressIPCache egressIPCache)
 			if ni == nil {
 				return fmt.Errorf("failed to get active network for network name %q", networkName)
 			}
-			reRoutePolicies, err := libovsdbops.FindALogicalRouterPoliciesWithPredicate(e.nbClient, ni.GetNetworkScopedClusterRouterName(), p1)
+			routerName := ni.GetNetworkScopedClusterRouterName()
+			if ni.TopologyType() == types.Layer2Topology {
+				// no support for multiple Nodes per OVN zone, therefore pick the first local zone node
+				localNodeName, err := e.getALocalZoneNodeName()
+				if err != nil {
+					return err
+				}
+				routerName = ni.GetNetworkScopedGWRouterName(localNodeName)
+			}
+			reRoutePolicies, err := libovsdbops.FindALogicalRouterPoliciesWithPredicate(e.nbClient, routerName, p1)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to retrieve a logical router polices attached to router %s: %w", routerName, err)
 			}
 			// not scoped by network since all NATs selected by the follow predicate select only CDN NATs
 			p2 := func(item *nbdb.NAT) bool {
@@ -1730,7 +1751,17 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
 			continue
 		}
 		redirectCache[ni.GetNetworkName()] = map[string]redirectIPs{}
-		cache.networkToRouter[ni.GetNetworkName()] = ni.GetNetworkScopedClusterRouterName()
+		var localNodeName string
+		if localZoneNodes.Len() > 0 {
+			localNodeName = localZoneNodes.UnsortedList()[0]
+		}
+		routerName, err := getTopologyScopedRouterName(ni, localNodeName)
+		if err != nil {
+			klog.Errorf("Failed to get network topology scoped router name for network %s attached to namespace %s, stale objects may remain: %v",
+				ni.GetNetworkName(), namespace.Name, err)
+			continue
+		}
+		cache.networkToRouter[ni.GetNetworkName()] = routerName
 		for _, node := range nodes {
 			r := redirectIPs{}
 			mgmtPort := &nbdb.LogicalSwitchPort{Name: ni.GetNetworkScopedK8sMgmtIntfName(node.Name)}
@@ -1763,14 +1794,14 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
 
 			if localZoneNodes.Has(node.Name) {
 				if e.v4 {
-					if gatewayRouterIP, err := e.getGatewayRouterJoinIP(ni, node.Name, false); err != nil {
+					if gatewayRouterIP, err := e.getGatewayNextHop(ni, node.Name, false); err != nil {
 						klog.V(5).Infof("Unable to retrieve gateway IP for node: %s, protocol is IPv4: err: %v", node.Name, err)
 					} else {
 						r.v4Gateway = gatewayRouterIP.String()
 					}
 				}
 				if e.v6 {
-					if gatewayRouterIP, err := e.getGatewayRouterJoinIP(ni, node.Name, true); err != nil {
+					if gatewayRouterIP, err := e.getGatewayNextHop(ni, node.Name, true); err != nil {
 						klog.V(5).Infof("Unable to retrieve gateway IP for node: %s, protocol is IPv6: err: %v", node.Name, err)
 					} else {
 						r.v6Gateway = gatewayRouterIP.String()
@@ -1950,12 +1981,11 @@ func (e *EgressIPController) addEgressNode(node *corev1.Node) error {
 			// on IC if the node is gone, then the ovn_cluster_router is also gone along with all
 			// the routes on it.
 			processNetworkFn := func(ni util.NetInfo) error {
-				clusterSubnetsNetEntry := ni.Subnets()
-				if len(clusterSubnetsNetEntry) == 0 {
+				if ni.TopologyType() == types.Layer2Topology || len(ni.Subnets()) == 0 {
 					return nil
 				}
 				if err := libovsdbutil.CreateDefaultRouteToExternal(e.nbClient, ni.GetNetworkScopedClusterRouterName(),
-					ni.GetNetworkScopedGWRouterName(node.Name), clusterSubnetsNetEntry); err != nil {
+					ni.GetNetworkScopedGWRouterName(node.Name), ni.Subnets()); err != nil {
 					return fmt.Errorf("failed to create route to external for network %s: %v", ni.GetNetworkName(), err)
 				}
 				return nil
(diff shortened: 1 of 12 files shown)


martinkennelly:
Rebasing again main

go-controller/pkg/ovn/egressip.go
```diff
@@ -2442,13 +2485,78 @@ func (e *EgressIPController) deleteExternalGWPodSNATOps(ni util.NetInfo, ops []o
...
+		return e.getGatewayRouterJoinIP(ni, nodeName, wantsIPv6)
+	}
+	if ni.TopologyType() == types.Layer2Topology {
+		// For L2, if the pod selected by an EgressIP and also that EgressIP is managed by this controller, then the next hop must be
+		// retrieved from the node annotation. If a node is managed by another controller, we must get the remote Node gateway router to switch port IP
+		// from the node annotation.
+		// FIXME: remove gathering the required information from a Node annotations as this approach does not scale
+		node, err := e.watchFactory.GetNode(nodeName)
+		if err != nil {
+			return nil, fmt.Errorf("faild to retrive Node %s: %w", nodeName, err)
```
trozet, line 2500: s/faild/failed

go-controller/pkg/ovn/egressip.go
```diff
@@ -2442,13 +2485,78 @@ func (e *EgressIPController) deleteExternalGWPodSNATOps(ni util.NetInfo, ops []o
...
+				if util.IsAnnotationNotSetError(err) {
+					// remote node may not have the annotation yet, suppress it
+					return nil, types.NewSuppressedError(err)
+				}
+				return nil, fmt.Errorf("failed to get the node %s L3 gateway annotation: %w", node.Name, err)
+			}
+			if len(nextHopIPs.NextHops) == 0 {
+				return nil, fmt.Errorf("annotation on the node %s had empty next hop IPs", node.Name)
+			}
+			return util.MatchFirstIPFamily(wantsIPv6, nextHopIPs.NextHops)
```
trozet, line 2519: I don't understand this. The next hop here will be the default gateway IP out of the cluster. IIUC this function is used to find the next hop to put as the reroute IP in the LRP. Why would you ever put the next hop outside of the cluster for that? Wouldn't you just not render the LRP at all and let the normal IP routing do its job?

go-controller/pkg/ovn/egressip.go
```diff
@@ -2442,13 +2485,78 @@ func (e *EgressIPController) deleteExternalGWPodSNATOps(ni util.NetInfo, ops []o
...
+	}
+	return nil, fmt.Errorf("unsupported network %s", ni.TopologyType())
+}
+
 func (e *EgressIPController) getGatewayRouterJoinIP(ni util.NetInfo, node string, wantsIPv6 bool) (net.IP, error) {
-	gatewayIPs, err := libovsdbutil.GetLRPAddrs(e.nbClient, types.GWRouterToJoinSwitchPrefix+ni.GetNetworkScopedGWRouterName(node))
+	return e.getRouterPortIP(types.GWRouterToJoinSwitchPrefix+ni.GetNetworkScopedGWRouterName(node), wantsIPv6)
+}
+
+func (e *EgressIPController) getGatewayRouterSwitchIP(ni util.NetInfo, node *kapi.Node, wantsIPv6 bool) (net.IP, error) {
```
trozet, line 2531: awkward name with "RouterSwitch" maybe getGatewayRouterJoinSubnetIP

trozet: why do you have this instad of just using func (e *EgressIPController) getGatewayRouterJoinIP ? or why does that function exist?

there's also util functions available:

```
func ParseNodeGatewayRouterJoinNetwork(node *kapi.Node, netName string) (primaryIfAddrAnnotation, error) {
	var val primaryIfAddrAnnotation
	joinSubnetMap, err := parseJoinSubnetAnnotation(node.Annotations, OVNNodeGRLRPAddrs)
	if err != nil {
		return val, fmt.Errorf("unable to parse annotation %s on node %s: err %w",
			OVNNodeGRLRPAddrs, node.Name, err)
	}
	val, ok := joinSubnetMap[netName]
	if !ok {
		return val, newAnnotationNotSetError("unable to fetch annotation value on node %s for network %s",
			node.Name, netName)
	}
	return val, nil
}

// ParseNodeGatewayRouterJoinIPv4 returns the IPv4 address for the node's gateway router port
// stored in the 'OVNNodeGRLRPAddrs' annotation
func ParseNodeGatewayRouterJoinIPv4(node *kapi.Node, netName string) (net.IP, error) {
	primaryIfAddr, err := ParseNodeGatewayRouterJoinNetwork(node, netName)
	if err != nil {
		return nil, err
	}
	if primaryIfAddr.IPv4 == "" {
		return nil, fmt.Errorf("failed to find an IPv4 address for gateway route interface in node: %s, net: %s, "+
			"annotation values: %+v", node, netName, primaryIfAddr)
	}

	ip, _, err := net.ParseCIDR(primaryIfAddr.IPv4)
	if err != nil {
		return nil, fmt.Errorf("failed to parse gateway router IPv4 address %s, err: %w", primaryIfAddr.IPv4, err)
	}
	return ip, nil
}
```

go-controller/pkg/ovn/egressip.go
```diff
@@ -3517,3 +3659,13 @@ func getEIPIPFamily(isIPv6 bool) egressIPFamilyValue {
 func addPktMarkToLRPOptions(options map[string]string, mark string) {
 	options["pkt_mark"] = mark
 }
+
+func getLayerScopedRouterName(ni util.NetInfo, nodeName string) (string, error) {
```
trozet, line 3663: maybe call this getTopologyScopedRouterName

go-controller/pkg/ovn/egressip.go
```diff
@@ -2521,23 +2629,23 @@ func (e *EgressIPController) getNextHop(ni util.NetInfo, egressNodeName, egressI
...
-					nextHopIP = mgmtPortIP.To16().String()
+					nextHop = mgmtPortIP.To16().String()
 				} else {
-					nextHopIP = mgmtPortIP.To4().String()
+					nextHop = mgmtPortIP.To4().String()
 				}
 			}
 		}
-	} else if config.OVNKubernetesFeature.EnableInterconnect {
+	} else if config.OVNKubernetesFeature.EnableInterconnect && ni.TopologyType() == types.Layer3Topology {
```
trozet, line 2638: So this logic is saying if the egress node is remote, then use the transit switch IP as the next hop, but only for Layer 3. Why wouldn't this also be true for Layer 2?

trozet (changes requested):

go-controller/pkg/ovn/egressip.go
martinkennelly: Not sure I get what you mean - there is no transit switch for a L2 network.
I added the extra constraint( ` && ni.TopologyType() == types.Layer3Topology {`) to increase code readability. Its a redundant constraint if we only process L3s or L2s networks.

go-controller/pkg/ovn/egressip.go
martinkennelly: > why do you have this instad of just using func (e *EgressIPController) getGatewayRouterJoinIP ? or why does that function exist?

I removed that func. It didnt increase code readability imo. 
I am using `getRouterPortIP`  directly without wrapping it with `getGatewayRouterJoinIP`.

I am using the utils now. I didn't know they existed.  I created one for IPv6.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1981,8 +2011,12 @@ func (e *EgressIPController) initClusterEgressPolicies(nodes []interface{}) erro
 	if err != nil {
 		return fmt.Errorf("failed to get default network: %v", err)
 	}
+	localNodeName, err := e.getALocalZoneNodeName()
+	if err != nil {
+		klog.Warningf(err.Error())
```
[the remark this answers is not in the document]
martinkennelly: Its intentional because we may not require this info if we do not process L2 networks and if we need to process a L2 network later on in this function, an error is returned.

go-controller/pkg/ovn/egressip.go
martinkennelly: > I don't understand this.

That makes two of us :)
I think I wrote this when there was no static routes associated with the L2 gw router. Not sure.
Ill rework this.

go-controller/pkg/ovn/egressip.go
```diff
@@ -2503,15 +2597,15 @@ func (e *EgressIPController) getNextHop(ni util.NetInfo, egressNodeName, egressI
 	isEgressIPv6 := utilnet.IsIPv6String(egressIP)
 	// NOTE: No need to check if status.node exists or not in the cache, we are calling this function only if it
 	// is present in the nodeZoneState cache. Since we call it with lock on cache, we are safe here.
-	if isLocalZoneEgressNode {
+	if isLocalZoneEgressNode || ni.TopologyType() == types.Layer2Topology {
 		if isOVNNetwork {
```
trozet, line 2601: can you add to doc text on this function what isOVNNetwork means? Everytime I see this variable i forget what it is

trozet: I think it means is the egress IP served by an OVS bridge (sgw mode) vs is it on interface not part of an OVS/OVN topology

go-controller/pkg/ovn/egressip.go
```diff
@@ -2503,15 +2597,15 @@ func (e *EgressIPController) getNextHop(ni util.NetInfo, egressNodeName, egressI
 	isEgressIPv6 := utilnet.IsIPv6String(egressIP)
 	// NOTE: No need to check if status.node exists or not in the cache, we are calling this function only if it
 	// is present in the nodeZoneState cache. Since we call it with lock on cache, we are safe here.
-	if isLocalZoneEgressNode {
+	if isLocalZoneEgressNode || ni.TopologyType() == types.Layer2Topology {
```
trozet, line 2600: i think we need a bunch more comments in this function to the different scenarios, because I think this is complex to understand without it being written down. So, this function is trying to determine next hop. Feel free to comment on the below scenarios.

1. If the local node is egress and the interface is not on the ovn network, we send to mp0 as next hop. This should be true for any topology type.
2. If the local node is egress and the interface is ovn networked, then just allow normal routing out of our node via the GR to breth0. This is the same for all topologies.
3. If the egress node is remote, and the topology is layer 3, then on ovn_cluster_router we reroute to the transit switch IP of the remote node as the next hop. This is true only for layer 3.
4. If the egress node is remote, and the topology is layer 2, then on the GR router we reroute to the remote node's GR join IP as the next hop.

agree? Is your code following this logic?

go-controller/pkg/ovn/egressip.go
trozet: true, no transit switch ip for l2 network. I tried to clarify in my earlier comment in this file what I'm trying to say.

go-controller/pkg/ovn/egressip.go
martinkennelly: > If the local node is egress and the interface is not on the ovn network, we send to mp0 as next hop. This should be true for any topology type.

Agree

> If the local node is egress and the interface is ovn networked, then just allow normal routing out of our node via the GR to breth0. This is the same for all topologies.

Lets say a pod is local and  egress node is also local, therefore for an LRP attached to the pods gateway router, How is ECMP accomplished? 

> If the egress node is remote, and the topology is layer 3, then on ovn_cluster_router we reroute to the transit switch IP of the remote node as the next hop. This is true only for layer 3.

agree

> If the egress node is remote, and the topology is layer 2, then on the GR router we reroute to the remote node's GR join IP as the next hop.

Its GR switch to router port - theres no join switch.

trozet: > > If the local node is egress and the interface is ovn networked, then just allow normal routing out of our node via the GR to breth0. This is the same for all topologies.
> 
> Lets say a pod is local and egress node is also local, therefore for an LRP attached to the pods gateway router, How is ECMP accomplished?
> 

For layer 3 we need the LRP on ovn_cluster_router with the next hops thats true. For layer 2 as you pointed out this becomes a problem because we are already on the GR, so adding an LRP with multiple nexthops means we have to include the default gateway next hop. Basically cloning routes that already exist in the routing table. It would be nice to have a feature where the nexthops could be (assuming 100.64.0.5 is another egress node) "local_routing_table,100.64.0.5". OVN would interpret local_routing_table to be the value it decided in the previous lr_route_lookup stage in the pipeline and ecmp between those 2 things. @dceara FYI. Since we dont have a feature like that, and with BGP routes can change dynamically in the GR, we would need to constantly update this LRP whenever routes for this pod update. This might be hard to do. In the meantime we can have this is as a limitation in layer 2 egress IP (file a jira), and not do the LRP with multiple next hops when the selected egress node is local.

> > If the egress node is remote, and the topology is layer 2, then on the GR router we reroute to the remote node's GR join IP as the next hop.
> 
> Its GR router to switch port - theres no join switch for L2.

There's no join switch true, but we still use a 2nd subnet (called the join subnet) for handing out unique IP addresses to the GRs and rerouting to them.

martinkennelly: >  Basically cloning routes that already exist in the routing table. It would be nice to have a feature where the nexthops could be (assuming 100.64.0.5 is another egress node) "local_routing_table,100.64.0.5".

Yes!! I was thinking something similiar because we  if there are multiple default gateway next hops, we cannot just add them into a reroute because it will no longer be ECMP.

martinkennelly:
Follow on work is required:
* to update the EIP docs for L2.
* fix fragmentation and unskip the test.
* Determine how we will integrate with BGP and remove the constraint of having a next hop outside the cluster for UDN L2

go-controller/pkg/ovn/egressip.go
trozet: just documenting (since my future self will forget anyway)...for now we decided leave the LRP with the next hop as the default gw. In the future with BGP we may need to dynamically update these LRPs @jcaamano . Another alternative may be what @dceara suggested with moving to the transit router.

dceara: Leaving this here too for more context: https://issues.redhat.com/browse/FDP-872

The suggested transit router topology change for L2 UDNs is:
```
  AZ1       AZ2
  ---       ---

  GR1       GR2
   |         |
   +-- TDR --+   
       |
       TS
       |
POD1 --+--- POD2
```

trozet (approved):

PR #4835, opened 2024-11-13
Import learnt BGP routes into OVN

jcaamano:
This commit adds a new controller to import BGP learnt routes into OVN.

The controller runs in ovnkube-controller so it only supports IC
architecture where ovnkube-controller has kernel access on each node.

Networks should register to this controller to have routes imported for
them. Routes are imported into the network's gateway router. Multipath
routes are supported.

The controller subscribes for netlink route events. When a route is
updated, the corresponding network is queued to be sync'ed. A network is
also sync'ed when registered to the controller.

Synchronizations are delayed by a small amount of time to prevent a
series of consecutive route updates so synchornize the same network
twice. Synchronizations apply the difference between current and desired
state.

The controller subscribes to netlink link events to learn the routing
table associated to a network vrf. The network is inferred from the vrf
device name. When learning the routing table, the corresponding network
is queued to be sync'ed.

Currently based on top of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4691

### DIFF
--- a/go-controller/pkg/libovsdb/ops/router.go
+++ b/go-controller/pkg/libovsdb/ops/router.go
@@ -602,6 +602,30 @@ func FindLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Client, 
 	return found, err
 }
 
+// GetRouterLogicalRouterStaticRoutesWithPredicate looks up logical router
+// static routes associated to the provided logical router from the cache based
+// on a given predicate
+func GetRouterLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Client, router *nbdb.LogicalRouter, p logicalRouterStaticRoutePredicate) ([]*nbdb.LogicalRouterStaticRoute, error) {
+	r, err := GetLogicalRouter(nbClient, router)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get router: %s, error: %w", router.Name, err)
+	}
+
+	lrsrs := []*nbdb.LogicalRouterStaticRoute{}
+	for _, uuid := range r.StaticRoutes {
+		lrsr := &nbdb.LogicalRouterStaticRoute{UUID: uuid}
+		err := nbClient.Get(context.Background(), lrsr)
+		if err != nil {
+			return nil, err
+		}
+		if p(lrsr) {
+			lrsrs = append(lrsrs, lrsr)
+		}
+	}
+
+	return lrsrs, nil
+}
+
 // CreateOrUpdateLogicalRouterStaticRoutesWithPredicateOps looks up a logical
 // router static route from the cache based on a given predicate. If it does not
 // exist, it creates the provided logical router static route. If it does, it
@@ -664,25 +688,12 @@ func CreateOrReplaceLogicalRouterStaticRouteWithPredicate(nbClient libovsdbclien
 	lrsr *nbdb.LogicalRouterStaticRoute, p logicalRouterStaticRoutePredicate, fields ...interface{}) error {
 
 	lr := &nbdb.LogicalRouter{Name: routerName}
-	router, err := GetLogicalRouter(nbClient, lr)
-	if err != nil {
-		return fmt.Errorf("unable to get logical router %s: %w", routerName, err)
-	}
-	newPredicate := func(item *nbdb.LogicalRouterStaticRoute) bool {
-		for _, routeUUID := range router.StaticRoutes {
-			if routeUUID == item.UUID && p(item) {
-				return true
-			}
-		}
-		return false
-	}
-	routes, err := FindLogicalRouterStaticRoutesWithPredicate(nbClient, newPredicate)
+	routes, err := GetRouterLogicalRouterStaticRoutesWithPredicate(nbClient, lr, p)
 	if err != nil {
 		return fmt.Errorf("unable to get logical router static routes with predicate on router %s: %w", routerName, err)
 	}
 
 	var ops []libovsdb.Operation
-	m := newModelClient(nbClient)
 
 	if len(routes) > 0 {
 		lrsr.UUID = routes[0].UUID
@@ -691,27 +702,7 @@ func CreateOrReplaceLogicalRouterStaticRouteWithPredicate(nbClient libovsdbclien
 	if len(routes) > 1 {
 		// should only be a single route remove all except the first
 		routes = routes[1:]
-		opModels := make([]operationModel, 0, len(routes)+1)
-		router.StaticRoutes = []string{}
-		for _, route := range routes {
-			route := route
-			router.StaticRoutes = append(router.StaticRoutes, route.UUID)
-			opModel := operationModel{
-				Model:       route,
-				ErrNotFound: false,
-				BulkOp:      false,
-			}
-			opModels = append(opModels, opModel)
-		}
-		opModel := operationModel{
-			Model:            router,
-			OnModelMutations: []interface{}{&router.StaticRoutes},
-			ErrNotFound:      true,
-			BulkOp:           false,
-		}
-		opModels = append(opModels, opModel)
-
-		ops, err = m.DeleteOps(nil, opModels...)
+		ops, err = DeleteLogicalRouterStaticRoutesOps(nbClient, ops, routerName, routes...)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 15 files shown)


go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -0,0 +1,560 @@
...
+		}
+		routeEventCh := subscribeNetlinkRouteEvents(c.netlink, stop, onError)
+		t := setTimer(nil, subscribePeriod)
+		for {
+			select {
+			case <-stop:
+				return
+			case <-t.C:
+				if routeEventCh != nil {
+					continue
```
trozet, line 211: so timer goes off and we are still subscribed, then we dont restart the timer? What happens if right after this we are unsubscribed and dont get any more events, now we never resubscribe?

go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -0,0 +1,560 @@
...
+}
+
+func (c *controller) Stop() {
+	controllerutil.Stop(c.reconciler)
+	c.ctx.Cancel()
+	c.log.Info("Controller stopped")
+}
+
+func (c *controller) subscribe(stop <-chan struct{}) {
+	setTimer := func(t *time.Timer, d time.Duration) *time.Timer {
```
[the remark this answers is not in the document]
trozet: how is this better than the methods we use in other controllers wehre we just use a ticker and every so often we check to make sure we are subscribed?

go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -0,0 +1,560 @@
...
+		t.Reset(d)
+		return t
+	}
+
+	go func() {
+		onError := func(err error) {
+			c.log.Error(err, "Error on netlink route event subscription")
+		}
+		routeEventCh := subscribeNetlinkRouteEvents(c.netlink, stop, onError)
+		t := setTimer(nil, subscribePeriod)
```
trozet, line 204: if we stop we need to in this goroutine we need to stop the timer to avoid leak. In go 1.23 we dont have to but we are still on 1.22 I think:

https://pkg.go.dev/time#NewTimer

go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -0,0 +1,560 @@
...
+		c.tables[table] = id
+	default:
+		c.log.Info("Unexpected VRF update event type", "type", update.IfInfomsg.Type)
+		return
+	}
+	if info != nil {
+		info.table = table
+	}
+
+	reconcile := info != nil && table != 0 && !current
```
trozet, line 327: nit: s/reconcile/needsReconcile

trozet: it looks like you are using table 0 to represent an invalid or not found table. Table 0 is a legit table in linux (although it is reserved for unicast addressing). I think it would be better to make a constant called "noTable" or something and use -1.

trozet (changes requested):

go-controller/pkg/libovsdb/ops/router.go
```diff
@@ -602,6 +602,30 @@ func FindLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Client,
 	return found, err
 }
 
+// GetRouterLogicalRouterStaticRoutesWithPredicate looks up logical router
+// static routes associated to the provided logical router from the cache based
+// on a given predicate
+func GetRouterLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Client, router *nbdb.LogicalRouter, p logicalRouterStaticRoutePredicate) ([]*nbdb.LogicalRouterStaticRoute, error) {
```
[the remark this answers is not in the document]
jcaamano: `FindLogicalRouterStaticRoutesWithPredicate` looks up the entire table. `GetRouterLogicalRouterStaticRoutesWithPredicate` looks up just among the routes of a specific router, which is what we want since we operate in the context of a specific network.

go-controller/pkg/libovsdb/ops/router.go
```diff
@@ -768,36 +792,42 @@ func DeleteLogicalRouterStaticRoutesWithPredicateOps(nbClient libovsdbclient.Cli
...
+// returns the ops to remove them from the provided logical router
+func DeleteLogicalRouterStaticRoutesOps(nbClient libovsdbclient.Client, ops []libovsdb.Operation, routerName string, lrsrs ...*nbdb.LogicalRouterStaticRoute) ([]libovsdb.Operation, error) {
 	router := &nbdb.LogicalRouter{
 		Name:         routerName,
 		StaticRoutes: make([]string, 0, len(lrsrs)),
 	}
 
 	opModels := make([]operationModel, 0, len(lrsrs)+1)
 	for _, lrsr := range lrsrs {
-		lrsr := lrsr
```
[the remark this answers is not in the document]
jcaamano: I can leave it but we aren't copying that pointer anywhere anymore so it is not needed.

go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -0,0 +1,560 @@
...
+	networkID := network.GetNetworkID()
+	if c.networkIDs[networkID] != "" {
+		return fmt.Errorf("already tracking network %q with ID %d",
+			c.networkIDs[networkID],
+			networkID,
+		)
+	}
+
+	name := network.GetNetworkName()
+	if c.networks[name] != nil {
```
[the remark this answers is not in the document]
jcaamano: Not possible. Check not needed.

This is remnant code from when the network ID was not being set by NetworkManager.

go-controller/pkg/ovn/routeimport/route_import.go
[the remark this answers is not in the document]
jcaamano: hmm I didn't have those other controllers in mind when I did this. If I had to think about differences though:

- Not having a ticker running if we don't actually need one
- Be in control of the delay with which resubscribe which helps because I will have a full sync after that.

go-controller/pkg/ovn/routeimport/route_import.go
jcaamano: If we are unsubscribed, the channel is closed and then we set the timer again.

go-controller/pkg/ovn/routeimport/route_import.go
jcaamano: Technically not a leak per se, it's just that the timer is not garbage collected until it expires. This becomes a problem if you are spawning a lot of different timers with long timeouts or in shorts periods of time (using constructs as `time.After` or the like). Here we are just reusing the same timers all over again, so its only two.

Anyway, aside from the clarification, you are right.

go-controller/pkg/ovn/routeimport/route_import.go
jcaamano: Will do. Just thought that 0 is just as invalid as -1 for us since we are not operating with that table and makes things easier since I don't have to deal with initialization.

go-controller/pkg/ovn/routeimport/route_import.go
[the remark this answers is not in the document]
jcaamano: I prefer to leave as is though, just in case there is problems elsewhere.

go-controller/pkg/ovn/routeimport/route_import.go
trozet: the original controller we started with for subscribing to netlink was address manager:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/node_ip_handler_linux.go#L173

I remember there were a bunch of problems with it and several fixes went in. The other controllers like link manager:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/linkmanager/link_network_manager.go

rule manager, route manager all follow this pattern. Probably would have been better at some point to make it a netlinkController generic type that all these things could just share. Anyway the channel getting stuck thing...that may have been in an underlying library and I'm misremembering, but I do for sure remember a case where we were in a state where we did not resubscribe.

My main point here is I would just leverage the other patterns that already exist, and hopefully someone consolidates them all to a single shareable controller later. If you think the pattern that already exists is flawed or that somehow yours is better then we can potentially switch those later to use this pattern. I just dont want a bunch of different ways to do the same thing, especially when I think your code is watching links, and we already have a link manager.

jcaamano: All those instances you mention are re-subscribing when the channel is closed, they are not working around any stuck channel issue really, in that respect they are doing the same thing that I am doing here.

So the difference is the use of a ticker vs the use of a timer, or in other words, when we a re-resubscribing: immediately when the subscription is lost, then again with a varying delay depending on ticker state; vs with a controlled delay always, respectively. I feel more comfortable with my approach but I will change to the other one .

VRF Manager is also watching links :stuck_out_tongue:

go-controller/pkg/libovsdb/ops/router.go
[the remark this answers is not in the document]
jcaamano: I see.  I guess since the client doesn't give us the tools to guarantee this consistency we have to find other ways. The existing approach might be better but it is not really consistent either since the snapshot of the router is taken at a different point in time than that of the routes.

I think in general we achieve consistency by defining an owner of the data and hoping nobody else changes it. Which is what I am doing. It would be better done with ObjectIDs but I haven't had the time yet.

I am not really sure about the other use cases but I had already changed the use of `FindLogicalRouterStaticRoutesWithPredicate` that you pointed me to to `GetRouterLogicalRouterStaticRoutesWithPredicate`. But I am not sure that is what you mean since it is just a private use and you talk about `single common method`. Let me know.

trozet (approved):

PR #4841, opened 2024-11-15
Silence errors caused by reconciling a NAD whose handler is a CNI plugin other than OVN-Kubernetes

maiqueb:
## 📑 Description

This PR silences errors caused by reconciling a NAD whose handler is a CNI plugin other than OVN-Kubernetes.

These errors are confusing to users, and might look like this is actually the cause of other errors.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
+++ b/go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
@@ -20,6 +20,7 @@ import (
 	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	nadinformers "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/informers/externalversions/k8s.cni.cncf.io/v1"
 	nadlisters "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controller"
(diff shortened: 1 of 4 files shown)


go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -211,6 +212,9 @@ func (nadController *NetAttachDefinitionController) syncNAD(key string, nad *net
 	if nad != nil {
 		nadNetwork, err = util.ParseNADInfo(nad)
 		if err != nil {
+			if errors.Is(err, config.ErrorAttachDefNotOvnManaged) {
+				return nil
+			}
```
maiqueb, line 217: @ricky-rav I think we should explicitly swallow the error at this layer rather than not return it from `util.ParseNADInfo`.

`util.ParseNADInfo` has the following signature:
```go
func ParseNADInfo(netattachdef *nettypes.NetworkAttachmentDefinition) (NetInfo, error)
```
Meaning, it receives the NAD, and returns an OVN-K `NetInfo` struct. When fed a NAD for a different plugin, it won't be able to return an OVN-K netinfo struct, and should return an error instead. 

It is up to the caller to know if that error warrants breaking the entire flow, or can be tolerated. In this case, it can be tolerated - and even ignored. We should break our flow because we see a macvlan NAD ... 

The alternative would be to only parse NADs whose handler is OVN-Kubernetes, but that involves a performance penalty since we'd have to marshall the NAD.spec.config into json, and look at the plugin type ... Or we'd need to start labeling the OVN-K NADs with a label, something which is easily circumvented.

ricky-rav: I like this approach :)

ricky-rav (approved):
LGTM
Thanks, Miguel!

Issue #4432, opened 2024-06-11
FLAKE: External Gateway With Admin Policy Based External Route CRs e2e multiple external gateway validation Should validate ICMP connectivity to multiple external gateways for an ECMP scenario IPV4


jcaamano:
### Which jobs are flaking?

[e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones)](https://github.com/ovn-org/ovn-kubernetes/actions/runs/9461898340/job/26064881251#logs)

### Which tests are flaking?

 [FAIL] External Gateway With Admin Policy Based External Route CRs e2e multiple external gateway validation Should validate ICMP connectivity to multiple external gateways for an ECMP scenario [It] IPV4

### Since when has it been flaking?

First seen now since there were no previous flakes reported

flavio-fernandes:
Also seen here: https://github.com/ovn-org/ovn-kubernetes/actions/runs/9567591018/job/26383336628?pr=4452

flavio-fernandes:
Also seen here: https://github.com/ovn-org/ovn-kubernetes/pull/4544

oshoval:
https://github.com/ovn-org/ovn-kubernetes/actions/runs/10301096279/job/28513027290?pr=4590

ormergi:
Seen here as well #4615

flavio-fernandes:
Seen here as well: https://github.com/ovn-org/ovn-kubernetes/actions/runs/10456435157/job/28955029631?pr=4544

maiqueb:
Seen [here](https://github.com/ovn-org/ovn-kubernetes/actions/runs/10827586108/job/30042447178?pr=4651).

flavio-fernandes:
Also seen here: https://github.com/ovn-org/ovn-kubernetes/actions/runs/11169273625/job/31050754172?pr=4755

flavio-fernandes:
Also seen here: https://github.com/ovn-org/ovn-kubernetes/actions/runs/11336443454/job/31571019568?pr=4760

qinqon:
https://github.com/ovn-org/ovn-kubernetes/actions/runs/11613999486/job/32342825247?pr=4702

ormergi:
Seen here https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/11929777670/job/33250603324#step:12:6882

martinkennelly:
[...]

Issue #4387, opened 2024-05-23
UT Flake: `handles a HO node is switched to a OVN node` is flaking


tssurya:
### Which jobs are flaking?

Unit Test

### Which tests are flaking?

- handles a HO node is switched to a OVN node

### Since when has it been flaking?

This started happening from this week

### Reason for failure (if possible)

The test is failing because we are having a race in how we compare the libovsdb objects:
```
I0521 23:40:23.312702   25592 reflector.go:295] Stopping reflector *v1.Node (0s) from k8s.io/client-go/informers/factory.go:159
I0521 23:40:23.312726   25592 reflector.go:295] Stopping reflector *v1.EndpointSlice (0s) from k8s.io/client-go/informers/factory.go:159
[91m[1m• Failure [2.584 seconds][0m
Hybrid SDN Master Operations
[90m/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:132[0m
I0521 23:40:23.312733   25592 watch.go:183] Stopping fake watcher.
  [91m[1mhandles a HO node is switched to a OVN node [It][0m
I0521 23:40:23.312862   25592 reflector.go:295] Stopping reflector *v1.Service (0s) from k8s.io/client-go/informers/factory.go:159
  [90m/go/src/github.com/openshift/ovn-kubernetes/go-controller/pkg/ovn/hybrid_test.go:1212[0m

I0521 23:40:23.312942   25592 reflector.go:295] Stopping reflector *v1.Namespace (0s) from k8s.io/client-go/informers/factory.go:159
I0521 23:40:23.312925   25592 handler.go:217] Removed *v1.Node event handler 3
  [91mTimed out after 2.000s.
  Expected
      <[]*nbdb.LogicalRouterStaticRoute | len:2, cap:2>: [
          {
              UUID: "7bc8e72a-5f2c-42e3-bb3f-28548e62048f",
              BFD: nil,
              ExternalIDs: {
                  "name": "hybrid-subnet-node1:node-windows",
              },
              IPPrefix: "10.1.3.0/24",
              Nexthop: "10.1.1.3",
              Options: nil,
              OutputPort: nil,
              Policy: nil,
              RouteTable: "",
          },
          {
              UUID: "05656f3b-320d-467c-9b86-909b01a8eb0f",
              BFD: nil,
              ExternalIDs: {
                  "name": "hybrid-subnet-node1-gr",
              },
[...]

Issue #3947, opened 2023-10-04
flakey test: Services should be able to switch session affinity for NodePort service


jcaamano:
It eventually passes after a long time but makes the suite to timeout.

Observed:
https://github.com/ovn-org/ovn-kubernetes/actions/runs/6398516819/job/17369527216
https://github.com/ovn-org/ovn-kubernetes/actions/runs/6382269653/job/17321633419

ormergi:
Also seen here https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/11892934771/job/33138386988?pr=4842
The test passed on second attempt

flavio-fernandes:
Also seen here https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12040294930/job/33571185708?pr=4849

oshoval:
here as well
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12113907226/job/33770425460?pr=4852

martinkennelly:
Also seen here https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12275067320/job/34302468635?pr=4530

martinkennelly:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12431347167/job/34709656534?pr=4833

From PR https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4833

I see it passes eventually but flaked.

martinkennelly:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12733411760/job/35519932462?pr=4877

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4877

npinaeva:
The test was fixed in https://github.com/kubernetes/kubernetes/pull/129049 and now as we have done k8s 1.33 bump https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5319 it should be in.

PR #4842, opened 2024-11-17
e2e, CUDN: Improve stability

ormergi:
## 📑 Description

This PR improve stability of CUDN e2e tests with the following changes:
- On CUDN deletion, wait for the object to be gone, and avoid pollution the next test with CUDN object.
- In the CUDN primary network already exist test, when CUDN is created wait for status to populate.
- Use random names for CUDN objects to enable running related tests in parallel.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/test/e2e/network_segmentation.go
+++ b/test/e2e/network_segmentation.go
@@ -21,6 +21,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/rand"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubectl/pkg/util/podutils"
 	"k8s.io/kubernetes/test/e2e/framework"
@@ -64,11 +65,11 @@ var _ = Describe("Network Segmentation", func() {
 	Context("a user defined primary network", func() {
 
 		DescribeTableSubtree("created using",
-			func(createNetworkFn func(c networkAttachmentConfigParams) error) {
+			func(createNetworkFn func(c *networkAttachmentConfigParams) error) {
 
 				DescribeTable(
 					"creates a networkStatus Annotation with UDN interface",
-					func(netConfig networkAttachmentConfigParams) {
+					func(netConfig *networkAttachmentConfigParams) {
 						By("creating the network")
 						netConfig.namespace = f.Namespace.Name
 						Expect(createNetworkFn(netConfig)).To(Succeed())
@@ -104,7 +105,7 @@ var _ = Describe("Network Segmentation", func() {
 						}
 					},
 					Entry("L2 primary UDN",
-						networkAttachmentConfigParams{
+						&networkAttachmentConfigParams{
 							name:     nadName,
 							topology: "layer2",
 							cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
@@ -112,7 +113,7 @@ var _ = Describe("Network Segmentation", func() {
 						},
 					),
 					Entry("L3 primary UDN",
-						networkAttachmentConfigParams{
+						&networkAttachmentConfigParams{
 							name:     nadName,
 							topology: "layer3",
 							cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
@@ -124,7 +125,7 @@ var _ = Describe("Network Segmentation", func() {
 				DescribeTable(
 					"can perform east/west traffic between nodes",
 					func(
-						netConfig networkAttachmentConfigParams,
+						netConfig *networkAttachmentConfigParams,
 						clientPodConfig podConfiguration,
 						serverPodConfig podConfiguration,
 					) {
@@ -166,7 +167,7 @@ var _ = Describe("Network Segmentation", func() {
 					},
 					Entry(
 						"two pods connected over a L2 primary UDN",
-						networkAttachmentConfigParams{
+						&networkAttachmentConfigParams{
 							name:     nadName,
 							topology: "layer2",
 							cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
@@ -186,7 +187,7 @@ var _ = Describe("Network Segmentation", func() {
 					),
 					Entry(
 						"two pods connected over a L3 primary UDN",
-						networkAttachmentConfigParams{
+						&networkAttachmentConfigParams{
 							name:     nadName,
 							topology: "layer3",
 							cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
@@ -209,7 +210,7 @@ var _ = Describe("Network Segmentation", func() {
 				DescribeTable(
 					"is isolated from the default network",
 					func(
-						netConfigParams networkAttachmentConfigParams,
+						netConfigParams *networkAttachmentConfigParams,
 						udnPodConfig podConfiguration,
 					) {
 						if !isInterconnectEnabled() {
@@ -454,7 +455,7 @@ var _ = Describe("Network Segmentation", func() {
 					},
 					Entry(
 						"with L2 primary UDN",
-						networkAttachmentConfigParams{
+						&networkAttachmentConfigParams{
 							name:     nadName,
 							topology: "layer2",
 							cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
@@ -470,7 +471,7 @@ var _ = Describe("Network Segmentation", func() {
 					),
 					Entry(
 						"with L3 primary UDN",
-						networkAttachmentConfigParams{
+						&networkAttachmentConfigParams{
 							name:     nadName,
 							topology: "layer3",
 							cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
@@ -501,8 +502,7 @@ var _ = Describe("Network Segmentation", func() {
 						namespaceRed := f.Namespace.Name + "-" + red
 						namespaceBlue := f.Namespace.Name + "-" + blue
 
-						netConfig := networkAttachmentConfigParams{
-
+						netConfig := &networkAttachmentConfigParams{
 							topology: topology,
 							cidr:     correctCIDRFamily(userDefinedv4Subnet, userDefinedv6Subnet),
 							role:     "primary",
@@ -524,8 +524,15 @@ var _ = Describe("Network Segmentation", func() {
 							By("creating the network " + network + " in namespace " + namespace)
 							netConfig.namespace = namespace
 							netConfig.name = network
+
 							Expect(createNetworkFn(netConfig)).To(Succeed())
+							// update the name because createNetworkFn may mutate the netConfig.name
+							// for cluster scope objects (i.g.: CUDN cases) to enable parallel testing.
+							networkNamespaceMap[namespace] = netConfig.name
 						}
+						red = networkNamespaceMap[namespaceRed]
+						blue = networkNamespaceMap[namespaceBlue]
+
 						pods := []*v1.Pod{}
 						redIPs := []string{}
 						blueIPs := []string{}
@@ -648,27 +655,29 @@ var _ = Describe("Network Segmentation", func() {
 					),
 				)
 			},
-			Entry("NetworkAttachmentDefinitions", func(c networkAttachmentConfigParams) error {
-				netConfig := newNetworkAttachmentConfig(c)
+			Entry("NetworkAttachmentDefinitions", func(c *networkAttachmentConfigParams) error {
+				netConfig := newNetworkAttachmentConfig(*c)
 				nad := generateNAD(netConfig)
 				_, err := nadClient.NetworkAttachmentDefinitions(c.namespace).Create(context.Background(), nad, metav1.CreateOptions{})
 				return err
 			}),
-			Entry("UserDefinedNetwork", func(c networkAttachmentConfigParams) error {
-				udnManifest := generateUserDefinedNetworkManifest(&c)
+			Entry("UserDefinedNetwork", func(c *networkAttachmentConfigParams) error {
+				udnManifest := generateUserDefinedNetworkManifest(c)
 				cleanup, err := createManifest(c.namespace, udnManifest)
 				DeferCleanup(cleanup)
 				Expect(waitForUserDefinedNetworkReady(c.namespace, c.name, 5*time.Second)).To(Succeed())
 				return err
 			}),
-			Entry("ClusterUserDefinedNetwork", func(c networkAttachmentConfigParams) error {
-				cudnManifest := generateClusterUserDefinedNetworkManifest(&c)
+			Entry("ClusterUserDefinedNetwork", func(c *networkAttachmentConfigParams) error {
+				cudnName := randomNetworkMetaName()
+				c.name = cudnName
+				cudnManifest := generateClusterUserDefinedNetworkManifest(c)
 				cleanup, err := createManifest("", cudnManifest)
 				DeferCleanup(func() {
 					cleanup()
 					By("delete pods in test namespace to unblock CUDN CR & associate NAD deletion")
 					Expect(cs.CoreV1().Pods(c.namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{})).To(Succeed())
-					_, err := e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", c.name, "--wait", fmt.Sprintf("--timeout=%ds", 120))
+					_, err := e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", cudnName, "--wait", fmt.Sprintf("--timeout=%ds", 120))
 					Expect(err).NotTo(HaveOccurred())
 				})
 				Expect(waitForClusterUserDefinedNetworkReady(c.name, 5*time.Second)).To(Succeed())
@@ -955,13 +964,10 @@ spec:
 	})
 
 	Context("ClusterUserDefinedNetwork CRD Controller", func() {
-		const (
-			testClusterUdnName                = "test-cluster-net"
-			clusterUserDefinedNetworkResource = "clusteruserdefinednetwork"
-		)
-		var (
-			testTenantNamespaces []string
-		)
+		const clusterUserDefinedNetworkResource = "clusteruserdefinednetwork"
+
+		var testTenantNamespaces []string
+
 		BeforeEach(func() {
 			testTenantNamespaces = []string{
 				f.Namespace.Name + "blue",
@@ -979,7 +985,10 @@ spec:
 			}
 		})
 
+		var testClusterUdnName string
+
 		BeforeEach(func() {
+			testClusterUdnName = randomNetworkMetaName()
 			By("create test CR")
 			cleanup, err := createManifest("", newClusterUDNManifest(testClusterUdnName, testTenantNamespaces...))
 			DeferCleanup(func() error {
@@ -1198,7 +1207,7 @@ spec:
 		Expect(err).NotTo(HaveOccurred())
 
 		By("create primary Cluster UDN CR")
-		const cudnName = "primary-net"
+		cudnName := randomNetworkMetaName()
 		cleanup, err := createManifest(f.Namespace.Name, newPrimaryClusterUDNManifest(cudnName, testTenantNamespaces...))
 		DeferCleanup(func() {
 			cleanup()
@@ -1239,11 +1248,11 @@ spec:
 			})
 		})
 		DescribeTableSubtree("created using",
-			func(createNetworkFn func(c networkAttachmentConfigParams) error) {
+			func(createNetworkFn func(c *networkAttachmentConfigParams) error) {
 
 				DescribeTable(
 					"can be accessed to from the pods running in the Kubernetes cluster",
-					func(netConfigParams networkAttachmentConfigParams, clientPodConfig podConfiguration) {
+					func(netConfigParams *networkAttachmentConfigParams, clientPodConfig podConfiguration) {
 						if netConfigParams.topology == "layer2" && !isInterconnectEnabled() {
 							const upstreamIssue = "https://github.com/ovn-org/ovn-kubernetes/issues/4642"
 							e2eskipper.Skipf(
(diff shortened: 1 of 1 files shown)


ormergi:
Failure in `ovn-ci / Upgrade OVN from Master to PR branch based image (local) (pull_request) ` is know and tracked in https://github.com/ovn-kubernetes/ovn-kubernetes/issues/3947

ormergi:
Failure in `ovn-ci / Build-PR (pull_request) ` is due to unit test failure `Hybrid SDN Master Operations [It] handles a HO node is switched to a OVN node` 
tracked by https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4387

ormergi:
The job `ovn-ci / e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones) (pull_re..` failed due to flaky test `External Gateway With Admin Policy Based External Route CRs e2e multiple external gateway validation Should validate ICMP connectivity to multiple external gateways for an ECMP scenario IPV4 `
Tracked by https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4432

test/e2e/network_segmentation.go
```diff
@@ -104,15 +105,15 @@ var _ = Describe("Network Segmentation", func() {
 					},
 					Entry("L2 primary UDN",
 						networkAttachmentConfigParams{
-							name:     nadName,
+							name:     randomNetworkMetaName(),
```
maiqueb, line 108: this change is misleading.

In the NAD, the parameter that matters is actually the `networkName`, which is already auto-generated when omitted (it usually is):
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/multihoming_utils.go#L76

The name of the NAD is used to attach the workload to a network, whereas the name of the `network` - which is the ID of the network - is used to define the network. The latter is what must be unique across the namespaces, and it already is.

I think you generate the network names from the metadata name of the cUDN CR. You'll need randomize, sure. But not for NADs - that's already done.

test/e2e/network_segmentation.go
```diff
@@ -1067,16 +1067,19 @@ spec:
...
+			Expect(err).NotTo(HaveOccurred())
+			var actualConditions []metav1.Condition
+			Expect(json.Unmarshal([]byte(conditionsJSON), &actualConditions)).To(Succeed())
+			return normalizeConditions(actualConditions)
+		}, 5*time.Second, 1*time.Second).Should(Equal([]metav1.Condition{{
+			Type:    "NetworkReady",
+			Status:  metav1.ConditionFalse,
+			Reason:  "NetworkAttachmentDefinitionSyncError",
+			Message: expectedMessage,
+		}}))
```
maiqueb, line 1094: if you're going to use expectations within eventually / consistently blocks, please use `GinkgoHelper` and pass gomega instance as the first function attribute.

Copying from ginkgo's documentation:
> Eventually can also accept functions that take a Gomega parameter. These functions are then passed a local Gomega that can be used to make assertions inside the function as it is polled. Eventually will retry the function if an assertion fails. This would allow us to replace:

```suggestion
		Eventually(func(g Gomega) []metav1.Condition {
			GinkgoHelper()
			conditionsJSON, err := e2ekubectl.RunKubectl(f.Namespace.Name, "get", "clusteruserdefinednetwork", cudnName, "-o", "jsonpath={.status.conditions}")
			g.Expect(err).NotTo(HaveOccurred())
			var actualConditions []metav1.Condition
			g.Expect(json.Unmarshal([]byte(conditionsJSON), &actualConditions)).To(Succeed())
			return normalizeConditions(actualConditions)
		}, 5*time.Second, 1*time.Second).Should(Equal([]metav1.Condition{{
			Type:    "NetworkReady",
			Status:  metav1.ConditionFalse,
			Reason:  "NetworkAttachmentDefinitionSyncError",
			Message: expectedMessage,
		}}))
```

test/e2e/network_segmentation.go
```diff
@@ -612,8 +612,11 @@ var _ = Describe("Network Segmentation", func() {
 					By("delete pods in test namespace to unblock CUDN CR & associate NAD deletion")
 					_, err := e2ekubectl.RunKubectl(c.namespace, "delete", "pod", "--all")
 					Expect(err).NotTo(HaveOccurred())
-					_, err = e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", c.name)
-					Expect(err).NotTo(HaveOccurred())
+					_, _ = e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", c.name)
+					Eventually(func() bool {
+						_, err := e2ekubectl.RunKubectl("", "get", "clusteruserdefinednetwork", c.name)
+						return strings.Contains(err.Error(), fmt.Sprintf("clusteruserdefinednetworks.k8s.ovn.org %q not found", c.name))
+					}, 1*time.Minute, 3*time.Second).Should(BeTrueBecause("ClusterUserDefinedNetwork %q should be gone", c.name))
```
maiqueb, line 620: should we use the `MatchError` matcher instead ? It works with [string matchers internally](https://github.com/onsi/gomega/blob/f1ff459a1aa4fb0000c65d4051b09d669a09d346/matchers.go#L127).

Meaning, I **think** the following would work:

```suggestion
					Eventually(func() error {
						_, err := e2ekubectl.RunKubectl("", "get", "clusteruserdefinednetwork", c.name)
						return err 
					}, 1*time.Minute, 3*time.Second).Should(MatchError(ContainSubstring(fmt.Sprintf("clusteruserdefinednetworks.k8s.ovn.org %q not found", c.name))))
```

maiqueb (changes requested):
Thank you.

The NAD naming changes is what I am most worried about.

The rest are just readability nits - I favor using `MatchError` whenever possible, rather than returning booleans and comparing within the `Eventually` block's body.

test/e2e/network_segmentation.go
ormergi: I had to make meta.name random because CUDN type is cluster scope, some tests fail every now and then on creating CUDN because there is already existing one with the same name.

The affected tests wrapped with `DescribeTableSubtree` to enable testing using CUDN type (similar to UDN type).
I had to change it for all cases (NAD and UDN types) due to the structure we have, its not optimal but I think its still better then duplicating test tables.

Make sense?

maiqueb: No. Or I don't follow why you **have** to impact NADs - AFAIU, you don't. NADs are OK. NADs already generate their names. 

Your cUDN building function just needs to build UDNs from the generic data in which the network name is already generated - thus unique.

I think that's more aligned with the current status-quo.

ormergi: I had to impact NAD and UDN types due to the structure we have today in tests.
That is the fact we wrap all tests with `DescribeTableSubtree` to test the exact same scenarios using different CRD types 
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e38f234624752f15755036cc2ca7975887786b51/test/e2e/network_segmentation.go#L1169-L1195

FWIU it works for NADs because there is some code that checks if no meta.name was specified it"ll generate one, its implicit and not trivial (at least for me).
I tried to avoid that for UDN and CUDN types to keep things simple.

test/e2e/network_segmentation.go
ormergi: For fetching the object (line 1072) it make sense because there could be some slow down with KAPI.
But if unmarshaling (line 1075) fails we have nothing to do, and I dont mind if the test will blow up. WDYT?

test/e2e/network_segmentation.go
maiqueb: keeping things simple - to me - is to follow the established pattern, and adjust to the existing status-quo. Meaning, re-use the established pattern for NADs, and extend your code to generate a name based on the networkName parameter.

How is that non-trivial ? And how is that complex ? 
At most you can say it is not explicit - but I don't see why it has to be: it's an implementation detail. The important fact here is the network name is the ID of the network, and it must be generated. 

The fact the CRD you've implemented generates the `networkName` of the underlying NAD from the metadata.name of the CRD is IMHO an implementation detail.

test/e2e/network_segmentation.go
maiqueb: How can unmarshalling something that comes from the k8s API fail ? You're asserting that for completeness AFAIU.

And still, the argument still applies: it a next reconcile cycle, it could have been "fixed".

ormergi: Done

test/e2e/network_segmentation.go
ormergi: I got it wrong, it seem for NADs the networkName is generated if not specified I though it was the metadata.name, sorry for the noise.

maiqueb: But you're affecting an object you have no need - at all - to affect. I still don't understand why you can't just carry the generated [attribute](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9cf5a7707ca278b96fed631a8780fbf3ff7e32bb/test/e2e/multihoming_utils.go#L60) from the `networkAttachmentConfigParams` and use that as metadata.name of the CRD, in the CRD's constructor functions.

test/e2e/network_segmentation.go
ormergi: Done

test/e2e/network_segmentation.go
ormergi: I tried it, somthing similar where only CUDN cases are affected, generating random meta.name only at the test setup for CUDN cases [[1]](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9cf5a7707ca278b96fed631a8780fbf3ff7e32bb/test/e2e/network_segmentation.go#L608-L621)
And having the `networkAttachmentConfigParams` cray that name into the test - [[2]](https://github.com/ormergi/ovn-kubernetes/commit/4e2ac804f0535c65cbe2c0f571e54bfae3ef88da#diff-2be1e01ba9cf5c059a4f4f0cd2b974861043381664e7102e3783f48842891d43R469-R477) (its a different branch)

It works, but require additional smelly change  in the "isolates overlapping CIDRs" tests table [[2]](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9cf5a7707ca278b96fed631a8780fbf3ff7e32bb/test/e2e/network_segmentation.go#L433-L594) test table.

The "isolates overlapping CIDRs" works differently, it wont take meta.name like other tests (from the Entry part), 
it hard code names for object it creates; two namespaces, is each namespace, network object (NAD/UDN/CUDN) & pods.
The problem is that the objects names depends on each other, and the test assume objects has certain names.
For CUDN cases to work, its require to override the names used for the objects it creates (namepaces, pods and NAD/UDN/CUDN) [[1]](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9cf5a7707ca278b96fed631a8780fbf3ff7e32bb/test/e2e/network_segmentation.go#L608-L621).

IMO the current solution is easier to understand thus I prefer it over the above alternative, that is due to how the  "isolates overlapping CIDRs" structured today.
WDYT?

Refactoring the "isolates overlapping CIDRs" requires some additional work and time I prefer to keep outside this PR socpe.

maiqueb: I essentially think the tests - and the helpers, as they exist today - are meant to be used in a certain way.

The fact we've chosen in the past to embed everything in tables (especially once we've nested the NAD/CRD table in ...) only made this contract more important: *everything * should adhere to these principles, otherwise, other areas in the tests will be impacted.

Sadly, part of the tests is not using the generated names. I understand that. Probably there's a reason, maybe there isn't one - but, the ugly truth is it has impacts elsewhere. **Everywhere** actually because we have nested the tables. Which, was a good choice. 

As you said, the proper solution is refactoring the "isolates overlapping CIDRs" tests, breaking the dependency it has on object's `metadata.name`; I'm aware that takes more work.

What you propose here is an immediate patch that works, but diverges us further from the was the helpers are supposed to be used. It's something that will impact every test, when it shouldn't, and it definitely feels you're hitting the code with an hammer to make it bend to your use case. I'm afraid in the long term, this will hurt us more.

I'd still vote for refactoring that test, but I won't block this PR over this.

ormergi: Thanks for the feedback.

> Sadly, part of the tests is not using the generated names. I understand that. Probably there's a reason, maybe there isn't one - but, the ugly truth is it has impacts elsewhere. Everywhere actually because we have nested the tables. Which, was a good choice.

All tests use hard-coded metadata.names, it works because each test run in its own namespace, no need to generate names.
Including cases that cover NAD usage, the do generate the network-name, if not given.
On CUDN usage cases it wont go for reasons mentioned above (its cluster scope object, must have unique name).

> What you propose here is an immediate patch that works, but diverges us further from the was the helpers are supposed to be used. It's something that will impact every test, when it shouldn't, and it definitely feels you're hitting the code with an hammer to make it bend to your use case. I'm afraid in the long term, this will hurt us more.

If you think the other alternative you suggest is better - we can go with that (patching the used names inside "isolates overlapping CIDRs"), eventually it doesnt impact the entire nested table.
Otherwise, I will appreciate your approval on this PR :)

maiqueb: I don't know how else to indicate that I think that's better. I've been arguing for that since the beginning. 

I will not approve, because I think this is wrong. I have explained what I think is correct and why in the previous comments. 

I also don't plan to block the PR; while I think this is wrong, if you can find someone to review and approve as is, be my guest.

npinaeva: If I am following this discussion correctly, I think all we need to do is keep NAD name as it was, but change https://github.com/ovn-kubernetes/ovn-kubernetes/blob/be2dfda65ad8ecec188f842233108768e1ab4085/test/e2e/network_segmentation.go#L1407-L1413 to generate unique CUDN.name instead of just using `params.name` 
@ormergi will that solve the problem?

ormergi: Unfortunately it wont solve it for all tests, the "isolates overlapping CIDRs" cases using CUDN will break, 
because there are dependencies on the CUDN object (e.g.: test expects CUDN name to be A, but the name is now B). 

I will change it to what suggested above: generate random meta.name for CUDNs here [[1]](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6b8ef77825556517e2116d9ac3e756867f815a48/test/e2e/network_segmentation.go#L608), and have the `networkAttachmentConfigParams` cray that name into the tests - [[2]](https://github.com/ormergi/ovn-kubernetes/commit/4e2ac804f0535c65cbe2c0f571e54bfae3ef88da#diff-2be1e01ba9cf5c059a4f4f0cd2b974861043381664e7102e3783f48842891d43R613-R615) (its a different branch)

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/6b8ef77825556517e2116d9ac3e756867f815a48..d26710e62847c4e6aad5825a6706624e1f917e81): change how CUDN random name are generated following discussion on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4842#discussion_r1866845321


@maiqueb @npinaeva please take another look :pray:

test/e2e/network_segmentation.go
ormergi: Done

maiqueb (changes requested):
Thank you, it looks much better now (IMHO ...).

Just a couple questions - I think you can still use a `MatchError` instead of comparing to a bool ...

maiqueb (dismissed):
Thank you for your patience.

test/e2e/network_segmentation.go
```diff
@@ -844,12 +851,19 @@ spec:
...
+			testClusterUdnName = randomNetworkMetaName()
 			By("create test CR")
 			cleanup, err := createManifest("", newClusterUDNManifest(testClusterUdnName, testTenantNamespaces...))
 			DeferCleanup(func() error {
 				cleanup()
 				_, _ = e2ekubectl.RunKubectl("", "delete", clusterUserDefinedNetworkResource, testClusterUdnName)
+				Eventually(func() bool {
+					_, err := e2ekubectl.RunKubectl("", "get", clusterUserDefinedNetworkResource, testClusterUdnName)
+					return strings.Contains(err.Error(), fmt.Sprintf("clusteruserdefinednetworks.k8s.ovn.org %q not found", testClusterUdnName))
+				}, 1*time.Minute, 3*time.Second).Should(BeTrueBecause("ClusterUserDefinedNetwork %q should be gone", testClusterUdnName))
```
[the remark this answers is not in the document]
ormergi: Done

test/e2e/network_segmentation.go
```diff
@@ -1059,24 +1073,31 @@ spec:
...
-		const cudnName = "primary-net"
+		cudnName := randomNetworkMetaName()
 		cleanup, err := createManifest(f.Namespace.Name, newPrimaryClusterUDNManifest(cudnName, testTenantNamespaces...))
 		DeferCleanup(func() error {
 			cleanup()
 			_, _ = e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", cudnName)
+			Eventually(func() bool {
+				_, err := e2ekubectl.RunKubectl("", "get", "clusteruserdefinednetwork", cudnName)
+				return strings.Contains(err.Error(), fmt.Sprintf("clusteruserdefinednetworks.k8s.ovn.org %q not found", cudnName))
+			}, 1*time.Minute, 3*time.Second).Should(BeTrueBecause("ClusterUserDefinedNetwork %q should be gone", cudnName))
```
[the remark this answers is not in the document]
ormergi: Done

test/e2e/network_segmentation.go
```diff
@@ -1059,24 +1073,31 @@ spec:
...
+			g.Expect(err).NotTo(HaveOccurred())
+			var actualConditions []metav1.Condition
+			g.Expect(json.Unmarshal([]byte(conditionsJSON), &actualConditions)).To(Succeed())
+			return normalizeConditions(actualConditions)
+		}, 5*time.Second, 1*time.Second).Should(Equal([]metav1.Condition{{
+			Type:    "NetworkReady",
+			Status:  metav1.ConditionFalse,
+			Reason:  "NetworkAttachmentDefinitionSyncError",
+			Message: expectedMessage,
+		}}))
```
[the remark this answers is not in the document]
ormergi: Done

We will need to do similar changes in a follow-up PR to other tests, given we are going to have more conditions in the future.

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/6587672cb0d15373ee58544a0c9ccb235594a070..7dd528658cfdb8dd57d06bbdb84c48c105be3ff9): resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4842#discussion_r1878480378
@maiqueb  I resolved the last nit, PTAL

maiqueb (dismissed):
Tks.

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/7dd528658cfdb8dd57d06bbdb84c48c105be3ff9..9c33946601bf76532a371c612c88e07082a8a65e): resolve comment from the D/S PR https://github.com/openshift/origin/pull/29274#discussion_r1888867716

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/7dd528658cfdb8dd57d06bbdb84c48c105be3ff9..9c33946601bf76532a371c612c88e07082a8a65e): resolve comment from the D/S PR https://github.com/openshift/origin/pull/29274#discussion_r1888867716 
@maiqueb PTAL

test/e2e/network_segmentation.go
```diff
@@ -612,8 +612,11 @@ var _ = Describe("Network Segmentation", func() {
 					By("delete pods in test namespace to unblock CUDN CR & associate NAD deletion")
 					_, err := e2ekubectl.RunKubectl(c.namespace, "delete", "pod", "--all")
 					Expect(err).NotTo(HaveOccurred())
-					_, err = e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", c.name)
-					Expect(err).NotTo(HaveOccurred())
+					_, _ = e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", c.name)
+					Eventually(func() error {
+						_, err := e2ekubectl.RunKubectl("", "get", "clusteruserdefinednetwork", c.name)
```
tssurya, line 624: hmm is it taking that long for the CUDNs to be deleted?
maybe `--wait` is better here? which blocks till deletion is complete? you can also use the --timeout=120s parameter with it

tssurya (commented):
commit1 lgtm
commit2: PTAL my suggestion
commit3: why did we change things to a pointer?

maiqueb (dismissed):
I'm OK with this iteration.

test/e2e/network_segmentation.go
ormergi: It shouldn't take long, but since we dont wait on pod deletion and CUDN blocks deletion until no consuming pod exist anyways, we have this long timeout because pods are involved.

I changes it to specify --wait --timeout=120s instead of the eventuality and pod deletion to be done using go-client, PTAL

test/e2e/network_segmentation.go
```diff
@@ -591,28 +597,28 @@ var _ = Describe("Network Segmentation", func() {
...
+				c.name = randomNetworkMetaName()
+				cudnManifest := generateClusterUserDefinedNetworkManifest(c)
 				cleanup, err := createManifest("", cudnManifest)
 				DeferCleanup(func() {
 					cleanup()
 					By("delete pods in test namespace to unblock CUDN CR & associate NAD deletion")
-					_, err := e2ekubectl.RunKubectl(c.namespace, "delete", "pod", "--all")
-					Expect(err).NotTo(HaveOccurred())
-					_, err = e2ekubectl.RunKubectl("", "delete", "clusteruserdefinednetwork", c.name)
+					Expect(cs.CoreV1().Pods(c.namespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{})).To(Succeed())
```
maiqueb, line 620: does `DeleteCollection` block ? Asking because I've noticed you've stopped using the eventually for deleting the cUDN.

ormergi: DeleteCollection shouldn't block, the next CUDN deletion command does block because it now specify `--wait`.
So we basically relay on the CUDN wait, which I think make sense because CUDN deletion blocks until connected pods are gone.

maiqueb (dismissed):

tssurya (dismissed):
thanks lgtm

tssurya:
@ormergi : all the net-seg lanes are red, ptal

ormergi:
[Rebased](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/cde7070224d2f847f8e475f34cbc7f0f5fed3543..4182bd3f9c47008c76f917c75103c5b4709f92e7)

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/4182bd3f9c47008c76f917c75103c5b4709f92e7..a059f124424e78c34d6e9994bd2703575c2953cf): Fix CUDN clean-up in "isolates overlapping CIDRs" tests.

ormergi:
[Changes:](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/a059f124424e78c34d6e9994bd2703575c2953cf..b7a0531455a1891aa09ca041293b04fba1df9945) revert using `--wait --timeout=..` to delete CUDN in "ClusterUserDefinedNetwork CRD Controller" tests, due to incompatibility with some tests.
Using Eventually poling CUDN object until is good middle-ground providing good results and stabilize the tests.

ormergi:
@tssurya PTAL

tssurya:
> [Changes:](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/a059f124424e78c34d6e9994bd2703575c2953cf..b7a0531455a1891aa09ca041293b04fba1df9945) revert using `--wait --timeout=..` to delete CUDN in "ClusterUserDefinedNetwork CRD Controller" tests, due to incompatibility with some tests. Using Eventually poling CUDN object until is good middle-ground providing good results and stabilize the tests.

what was the incompatibility issue in some tests... isn't the wait doing what we wanted it to do?

tssurya (approved):

ormergi:
> what was the incompatibility issue in some tests... isn't the wait doing what we wanted it to do? looks like its OK in two places but bad in 1 place? I won't block this PR anymore but just for my own understanding.

Sorry for being ambuguied lemme elaborate: it did work for some tests in "ClusterUserDefinedNetwork CRD Controller", tests that involve deleting the CUDN object.
Once these tests reached defer-cleanup the CUDN was already gone, causing the deletion command fail with "not exist" errors.

Issue #4829, opened 2024-11-10
ClusterUserDefinedNetwork isolation tests are flaking


tssurya:
### Which jobs are flaking?

network-segmentation lane

### Which tests are flaking?

Summarizing 2 Failures:
  [FAIL] Network Segmentation a user defined primary network created using ClusterUserDefinedNetwork is isolated from the default network [It] with L2 primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:357
  [FAIL] Network Segmentation a user defined primary network created using ClusterUserDefinedNetwork is isolated from the default network [It] with L3 primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:357

### Since when has it been flaking?

Since CUDN got added

### Anything else we need to know?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/11756667770/job/32753396792?pr=4827

ormergi:
It seem that the tests fails on checking pod connected to UDN is able to reach KAPI
[ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:357](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/network_segmentation.go#L345-L357)
This issue wont seem to reproduce on local env, I tried to run it continually using Ginkgo unti-it-fails option.
I couldn't find something related to the fact the test use the CUDN CRD yet.

trozet:
Also seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/3709#issuecomment-2474936262

flavio-fernandes:
Also seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12206166602/job/34056107375?pr=4873

qinqon:
Here fails at local gateway net seg service lane:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4774 -> https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12243700143/job/34155127680?pr=4774

oshoval:
Worth to add please
"should be reachable through their cluster IP and node port [It] L3 primary UDN, cluster-networked pods, NodePort service"
so the search will find this issue (maybe it can find it already due to this comment, will try)

oshoval:
Yep this why i added this paragraph, thanks

[...]

PR #4846, opened 2024-11-21
OCPBUGS-38151: NAD controller: emit err/event for NADs we cannot sync at startup

martinkennelly:
.. instead of blocking ovnkube controller startup. This can occur when one NAD has an invalid config and this leads to DOS for other services ovnkube controller offers because it cannot sync the nad controller at startup.

The sync can fail for reasons that retry will not fix:
* `namespace, _, err := cache.SplitMetaNamespaceKey(key)`
* `if nadNetwork != nil && nadNetwork.IsPrimaryNetwork() && primaryNAD != "" && primaryNAD != key {`
* `	err = fmt.Errorf("%s: NAD %s CNI config does not match that of network %s", nadController.name, key, nadNetworkName)`

Retry wont fix this and for NAD controllers startup `syncAll()`, a config error can cause ovnkube controller not to start. I think its best to log and emit event.

### DIFF
--- a/go-controller/pkg/networkmanager/nad_controller.go
+++ b/go-controller/pkg/networkmanager/nad_controller.go
@@ -1,6 +1,7 @@
 package networkmanager
 
 import (
+	"errors"
 	"fmt"
 	"reflect"
 	"sync"
@@ -21,13 +22,13 @@ import (
 	nettypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	nadinformers "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/informers/externalversions/k8s.cni.cncf.io/v1"
 	nadlisters "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controller"
 	rainformers "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/routeadvertisements/v1/apis/informers/externalversions/routeadvertisements/v1"
 	userdefinednetworkinformer "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/informers/externalversions/userdefinednetwork/v1"
 	userdefinednetworklister "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/listers/userdefinednetwork/v1"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/errors"
 	utiludn "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/udn"
 )
 
@@ -160,8 +161,12 @@ func (c *nadController) syncAll() (err error) {
 			continue
 		}
 		err = c.syncNAD(key, nad)
-		if err != nil {
-			return fmt.Errorf("%s: failed to sync %s: %v", c.name, key, err)
+		if err != nil && !errors.Is(err, config.ErrorAttachDefNotOvnManaged) {
+			if c.recorder != nil {
+				c.recorder.Eventf(&corev1.ObjectReference{Kind: nad.Kind, Namespace: nad.Namespace, Name: nad.Name}, corev1.EventTypeWarning,
+					"FailedSync", "Failed to sync network %s config: %v", key, err)
+			}
+			klog.Warningf("%s: unable to sync %s: %v", c.name, key, err)
 		}
 	}
 
--- a/go-controller/pkg/networkmanager/nad_controller_test.go
+++ b/go-controller/pkg/networkmanager/nad_controller_test.go
@@ -549,10 +549,13 @@ func TestSyncAll(t *testing.T) {
 			Name: "network_A",
 			Type: "ovn-k8s-cni-overlay",
 		},
-		Role: types.NetworkRolePrimary,
-		MTU:  1400,
+		Subnets: "10.20.0.0/16/24",
+		Role:    types.NetworkRolePrimary,
+		MTU:     1400,
 	}
 	network_A_Copy := *network_A
+	network_A_Copy_InvalidSubnet := *network_A
+	network_A_Copy_InvalidSubnet.Subnets = "20.20.0.0/16/24" // network subnet was previously defined as "10.20.0.0/16/24"
 	network_B := &ovncnitypes.NetConf{
 		Topology: types.LocalnetTopology,
 		NetConf: cnitypes.NetConf{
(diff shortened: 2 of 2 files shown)


go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -170,7 +170,11 @@ func (nadController *NetAttachDefinitionController) syncAll() (err error) {
 		}
 		err = nadController.syncNAD(key, nad)
 		if err != nil {
-			return fmt.Errorf("%s: failed to sync %s: %v", nadController.name, key, err)
+			if nadController.recorder != nil {
+				nadController.recorder.Eventf(&corev1.ObjectReference{Kind: nad.Kind, Namespace: nad.Namespace, Name: nad.Name}, corev1.EventTypeWarning,
+					"FailedSync", "Failed to sync network %s config: %v", key, err.Error())
```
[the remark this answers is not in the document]
martinkennelly: yep.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -170,7 +170,11 @@ func (nadController *NetAttachDefinitionController) syncAll() (err error) {
...
+			}
+			klog.Errorf("%s: failed to sync %s: %v", nadController.name, key, err)
```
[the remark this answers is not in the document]
martinkennelly: but this is an actual error though that the user should be notified about during NAD sync at startup. I dont agree about increasing the verbosity. Maybe I am missing something.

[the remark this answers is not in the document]
martinkennelly: missed that - thanks - updating to account for this

maiqueb (dismissed):
Thank you !

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -169,8 +170,12 @@ func (nadController *NetAttachDefinitionController) syncAll() (err error) {
 			continue
 		}
 		err = nadController.syncNAD(key, nad)
-		if err != nil {
-			return fmt.Errorf("%s: failed to sync %s: %v", nadController.name, key, err)
+		if err != nil && !errors.Is(err, config.ErrorAttachDefNotOvnManaged) {
```
tssurya, line 173: do we need a test? :D

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
```diff
@@ -169,8 +170,12 @@ func (nadController *NetAttachDefinitionController) syncAll() (err error) {
...
+			if nadController.recorder != nil {
+				nadController.recorder.Eventf(&corev1.ObjectReference{Kind: nad.Kind, Namespace: nad.Namespace, Name: nad.Name}, corev1.EventTypeWarning,
+					"FailedSync", "Failed to sync network %s config: %v", key, err)
+			}
+			klog.Errorf("%s: failed to sync %s: %v", nadController.name, key, err)
```
tssurya, line 178: Is this an error or warning? like ovnk shouldn't be managing these anyways right? its a warning to the user they did something wrong? or a proper error happened for ovnk?

martinkennelly: > its a warning to the user they did something wrong?

Correct. Today, thats all that can cause this, so I agree, Ill change it to Warning. 
Warning event is the highest "error" event level.

go-controller/pkg/network-attach-def-controller/network_attach_def_controller.go
martinkennelly: added "sync" unit test

martinkennelly:
@tssurya PTAL when youre back.

martinkennelly:
E2Es caught flakes notes here: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4829

martinkennelly:
Rebase only

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -160,8 +161,12 @@ func (c *nadController) syncAll() (err error) {
...
 		}
 		err = c.syncNAD(key, nad)
-		if err != nil {
-			return fmt.Errorf("%s: failed to sync %s: %v", c.name, key, err)
+		if err != nil && !errors.Is(err, config.ErrorAttachDefNotOvnManaged) {
+			if c.recorder != nil {
+				c.recorder.Eventf(&corev1.ObjectReference{Kind: nad.Kind, Namespace: nad.Namespace, Name: nad.Name}, corev1.EventTypeWarning,
+					"FailedSync", "Failed to sync network %s config: %v", key, err)
+			}
+			klog.Warningf("%s: unable to sync %s: %v", c.name, key, err)
```
tssurya, line 169: nit:
```suggestion
			klog.Warningf(" nadController %s unable to sync NAD %s: %v", c.name, key, err)
```

go-controller/pkg/networkmanager/nad_controller_test.go
```diff
@@ -588,6 +591,19 @@ func TestSyncAll(t *testing.T) {
...
+		{
+			name: "invalid network reference with erroneous subnet entry doesn't block sync",
+			testNADs: []TestNAD{
+				{
+					name:    "test/nad1",
+					netconf: network_A,
+				},
+				{
+					name:    "test2/nad2",
+					netconf: &network_A_Copy_InvalidSubnet,
```
tssurya, line 603: ah the fields need to be exactly the same here cause its two different NADs referecing same network..?

tssurya (commented):
I think overall I agree here with not blocking sync on these unresolvable errors which can cause crashing I think?

```
	// this was a nad delete
	if ensureNetwork == nil {
		delete(c.nads, key)
		if c.primaryNADs[namespace] == key {
			delete(c.primaryNADs, namespace)
		}
		return err
	}
```
what about this error? is there a reason to retry for that?

go-controller/pkg/networkmanager/nad_controller.go
martinkennelly: This is what is written to `c.name`:
`name:              fmt.Sprintf("[%s NAD controller]", name),`

go-controller/pkg/networkmanager/nad_controller_test.go
martinkennelly: ya, same network name but different network subnets.

maiqueb:
> Looks like Jaime refactor the code for this PR and the problem is now fixed here:
> 
> ```
> nadNetwork, err = util.ParseNADInfo(nad)
> 		if err != nil {
> 			if c.recorder != nil {
> 				c.recorder.Eventf(&corev1.ObjectReference{Kind: nad.Kind, Namespace: nad.Namespace, Name: nad.Name}, corev1.EventTypeWarning,
> 					"InvalidConfig", "Failed to parse network config: %v", err.Error())
> 			}
> 			klog.Errorf("%s: failed parsing NAD %s: %v", c.name, key, err)
> 			return nil
> 		}
> ```
> 
> Jaime handled the error internally within `syncNAD` I see there are also additional cases within that func that may trigger a pointless retry. Ill update the PR but the original bug that spawned this PR is now fixed in main :D

Nice. 

Regarding the pointless retry, I've pushed [this](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4816/files#diff-dd1451a36b97147d7381e265cbf45bbfbcbfe483750e426322013433a5523422) a while back ... 
What are your thoughts ? The logic could be re-used.

tssurya:
> Jaime handled the error internally within `syncNAD` I see there are also additional cases within that func that may trigger a pointless retry. Ill update the PR but the original bug that spawned this PR is now fixed in main :D

do we still need this PR then?

Issue #4474, opened 2024-06-26
Flake:  [FAIL] External Gateway With Admin Policy Based External Route CRs e2e non-vxlan external gateway through a gateway pod Should validate ICMP connectivity to an external gateway's loopback address via a gateway pod [It] ipv4


jcaamano:
### Which jobs are flaking?

ovn-ci / e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones):
https://github.com/ovn-org/ovn-kubernetes/actions/runs/9667215971/job/26703057307?pr=4461

### Which tests are flaking?

 [FAIL] External Gateway With Admin Policy Based External Route CRs e2e non-vxlan external gateway through a gateway pod Should validate ICMP connectivity to an external gateway's loopback address via a gateway pod [It] ipv4

### Since when has it been flaking?

unknown

flavio-fernandes:
Also seen here: https://github.com/ovn-org/ovn-kubernetes/actions/runs/9769556574/job/26996867102?pr=4494

qinqon:
another https://github.com/ovn-org/ovn-kubernetes/actions/runs/9903086035/job/27358690904?pr=4473

kyrtapz:
another one: https://github.com/ovn-org/ovn-kubernetes/actions/runs/9957563177/job/27511947616?pr=4499

npinaeva:
[e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones)](https://github.com/ovn-org/ovn-kubernetes/actions/runs/10290466858/job/28482174179?pr=4596#logs)

trozet:
https://github.com/ovn-org/ovn-kubernetes/pull/4599#issuecomment-2276381435

martinkennelly:
[e2e (external-gateway, noHA, shared, ipv4, noSnatGW, 2br, ic-single-node-zones)](https://github.com/ovn-org/ovn-kubernetes/commit/2211a2db6eea81fcda1b792f60bd214d37679c1a/checks/28565400249/logs)

PR: https://github.com/ovn-org/ovn-kubernetes/pull/4593

flavio-fernandes:
Seen here https://github.com/ovn-org/ovn-kubernetes/actions/runs/10379907269/job/28759220303?pr=4544
Same signature as the previous replies.

flavio-fernandes:
[...]

Issue #4220, opened 2024-03-13
Flake: `AdminNetworkPolicyIngressUDP` -> `Should_support_an_'deny-ingress'_policy_for_UDP_protocol;_ensure_rule_ordering_is_respected`


martinkennelly:
```
2024-03-12T15:26:10.8723587Z === RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressUDP
2024-03-12T15:26:10.8728131Z     suite.go:187: Applying base/admin_network_policy/core-ingress-udp-rules.yaml
2024-03-12T15:26:10.8839261Z     apply.go:124: Creating ingress-udp AdminNetworkPolicy
2024-03-12T15:26:10.8893134Z === RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressUDP/Should_support_an_'allow-ingress'_policy_for_UDP_protocol;_ensure_rule_ordering_is_respected
2024-03-12T15:26:11.0462320Z === RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressUDP/Should_support_an_'allow-ingress'_policy_for_UDP_protocol_at_the_specified_port
2024-03-12T15:26:14.1921073Z === RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressUDP/Should_support_an_'deny-ingress'_policy_for_UDP_protocol;_ensure_rule_ordering_is_respected
2024-03-12T15:26:14.2764159Z     admin-network-policy-core-ingress-udp-rules.go:118: FAILED Command was [/agnhost connect --timeout=3s --protocol=udp 10.244.0.5:53]
2024-03-12T15:26:14.2767812Z     admin-network-policy-core-ingress-udp-rules.go:118: Expected connection to fail from network-policy-conformance-ravenclaw/luna-lovegood-0 to 10.244.0.5, but instead it successfully connected.
2024-03-12T15:26:14.2770006Z     admin-network-policy-core-ingress-udp-rules.go:120: 
2024-03-12T15:26:14.2772654Z         	Error Trace:	/home/runner/go/pkg/mod/sigs.k8s.io/network-policy-api@v0.1.2/conformance/tests/admin-network-policy-core-ingress-udp-rules.go:120
2024-03-12T15:26:14.2774002Z         	Error:      	Not equal: 
2024-03-12T15:26:14.2774436Z         	            	expected: true
[...]

PR #4847, opened 2024-11-25
kubevirt, l2, udpn: send unsolicited router advertisement after live migration to reconcile ipv6 default gw

qinqon:
## 📑 Description
At current layer2 topology each node has gateway router ipv6 link local ip and mac addresses, this end up with a multipath ipv6 defaujl gateways at VMs.

This change send a router advertisment with lifetime 0 to remove the ipv6 default gw path using the node where the VM was running before migration and send another one with lifetime max with the gw path where the VM is currently running after migration (after live migration the VM is not going to send a router solicitation so the router advertisment should be generated from ovn-k).

Fixes https://issues.redhat.com/browse/SDN-5424

Note that another [PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852) that uses nftables to block RAs from external nodes is in progress, this PR should work in conjuntion with that.

Also note that proper solution is to use the future [OVN transit router](https://issues.redhat.com/browse/FDP-872) topology where the gw will have the same ip and mac addresses independently of the node where the VM is running, when that lands at OVN the changes introduced by this PR will no longer be needed.

The helper that send the RAs is being placed at `pkg/util/ndp/ra.go` instead of `pkg/util/ra.go` so the PR do not breaks windows compilation, the hybrid overlay binary depends on `pkg/util`.

Also instead of using the high level NDP library `mdlayher/ndp`  we use `github.com/google/gopacket` (already included on the project), with the mdlayher/ndp the src(ip, mac) from the RA is the one from the interface we are sending the message from, in our case the mgmt one, but instead we need them to be the gateway routers join IP LLA/MAC (this is how the VM's recognice what route path are we talking about).

Related-to:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852

**NOTE**: For this to work the VM guest need to fully support IPv6 standard, specially understanding router advertisments with lifetime=0, for example fedora35 do not.

## ✅ Checks

- [x] My code requires tests

## How to verify it
The PR has e2e check to ipv6 default gw path after live migration.

### DIFF
--- a/go-controller/go.mod
+++ b/go-controller/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/k8snetworkplumbingwg/sriovnet v1.2.1-0.20230427090635-4929697df2dc
 	github.com/mdlayher/arp v0.0.0-20220512170110-6706a2966875
 	github.com/mdlayher/ndp v1.0.1
+	github.com/mdlayher/socket v0.2.1
 	github.com/metallb/frr-k8s v0.0.15
 	github.com/miekg/dns v1.1.31
 	github.com/mitchellh/copystructure v1.2.0
@@ -103,7 +104,6 @@ require (
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mdlayher/ethernet v0.0.0-20220221185849-529eae5b6118 // indirect
 	github.com/mdlayher/packet v1.0.0 // indirect
-	github.com/mdlayher/socket v0.2.1 // indirect
 	github.com/mitchellh/reflectwalk v1.0.2 // indirect
 	github.com/moby/spdystream v0.5.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -22,7 +22,28 @@ import (
 	logicalswitchmanager "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/logical_switch_manager"
 	ovntypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/ndp"
 )
+
+// DefaultGatewayReconciler is responsible for reconciling the default gateway
+// configuration of a virtual machine's network interface after a live migration.
+// It supports both IPv4 and IPv6 configurations.
+type DefaultGatewayReconciler struct {
+	watchFactory  *factory.WatchFactory
+	netInfo       util.NetInfo
+	interfaceName string
+}
+
+// NewDefaultGatewayReconciler creates a new instance of DefaultGatewayReconciler.
+// It takes a WatchFactory for managing resource watches, a NetInfo object for network information,
+// and the name of the network interface to send ARPs or RAs as parameters.
+func NewDefaultGatewayReconciler(watchFactory *factory.WatchFactory, netInfo util.NetInfo, interfaceName string) *DefaultGatewayReconciler {
+	return &DefaultGatewayReconciler{
+		watchFactory:  watchFactory,
+		netInfo:       netInfo,
+		interfaceName: interfaceName,
+	}
+}
 
 // IsPodLiveMigratable will return true if the pod belongs
 // to kubevirt and should use the live migration features
@@ -482,17 +503,20 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod) 
 	return &status, nil
 }
 
-func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFactory, netInfo util.NetInfo, liveMigrationStatus *LiveMigrationStatus, interfaceName string) error {
+// ReconcileIPv4AfterLiveMigration will send a GARP after live migration
+// to update the default gw mac address to the node where the VM is running
+// now.
+func (r *DefaultGatewayReconciler) ReconcileIPv4AfterLiveMigration(liveMigrationStatus *LiveMigrationStatus) error {
 	if liveMigrationStatus.State != LiveMigrationTargetDomainReady {
 		return nil
 	}
 
-	targetNode, err := watchFactory.GetNode(liveMigrationStatus.TargetPod.Spec.NodeName)
+	targetNode, err := r.watchFactory.GetNode(liveMigrationStatus.TargetPod.Spec.NodeName)
 	if err != nil {
 		return err
 	}
 
-	lrpJoinAddress, err := util.ParseNodeGatewayRouterJoinNetwork(targetNode, netInfo.GetNetworkName())
+	lrpJoinAddress, err := util.ParseNodeGatewayRouterJoinNetwork(targetNode, r.netInfo.GetNetworkName())
 	if err != nil {
 		return err
 	}
@@ -503,16 +527,104 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
 	}
 
 	lrpMAC := util.IPAddrToHWAddr(lrpJoinIPv4)
-	for _, subnet := range netInfo.Subnets() {
+	for _, subnet := range r.netInfo.Subnets() {
 		gwIP := util.GetNodeGatewayIfAddr(subnet.CIDR).IP.To4()
 		if gwIP == nil {
 			continue
 		}
 		garp := util.GARP{IP: gwIP, MAC: &lrpMAC}
-		if err := util.BroadcastGARP(interfaceName, garp); err != nil {
+		if err := util.BroadcastGARP(r.interfaceName, garp); err != nil {
 			return err
 		}
 	}
-
 	return nil
+}
+
+// ReconcileIPv6AfterLiveMigration will do two things at VM's:
+// - Remove ipv6 default gw path from VM's node before live migration
+// - Add ipv6 default gw path from VM's node after live migration
+// This is done by sending a pair of unsolicited RA's one with lifetime=0
+// (to remove the gateway path) another with lifetime=max to add the new
+// default gateway path
+func (r *DefaultGatewayReconciler) ReconcileIPv6AfterLiveMigration(liveMigration *LiveMigrationStatus) error {
+	if !liveMigration.IsTargetDomainReady() {
+		return nil
+	}
+	nodes, err := r.watchFactory.GetNodes()
+	if err != nil {
+		return err
+	}
+
+	targetPod := liveMigration.TargetPod
+	if len(r.netInfo.GetNADs()) != 1 {
+		return fmt.Errorf("expected only one nad for network %q, got %d", r.netInfo.GetNetworkName(), len(r.netInfo.GetNADs()))
+	}
+
+	targetPodAnnotation, err := util.UnmarshalPodAnnotation(targetPod.Annotations, r.netInfo.GetNADs()[0])
+	if err != nil {
+		return ovntypes.NewSuppressedError(fmt.Errorf("failed parsing ovn pod annotation for pod '%s/%s' and network %q: %w", targetPod.Namespace, targetPod.Name, r.netInfo.GetNetworkName(), err))
+	}
+
+	destinationIP, err := util.MatchFirstIPNetFamily(true /* ipv6 */, targetPodAnnotation.IPs)
+	if err != nil {
+		return err
+	}
+	destinationMAC := targetPodAnnotation.MAC
+
+	ras := make([]ndp.RouterAdvertisement, 0, len(nodes))
+	for _, node := range nodes {
+		if node.Name == liveMigration.TargetPod.Spec.NodeName {
+			// skip the target node since this is the proper gateway
+			continue
+		}
+		nodeJoinAddrs, err := util.ParseNodeGatewayRouterJoinAddrs(node, r.netInfo.GetNetworkName())
+		if err != nil {
+			return ovntypes.NewSuppressedError(fmt.Errorf("failed parsing join addresss from node %q and network %q to reconcile ipv6 gateway: %w", node.Name, r.netInfo.GetNetworkName(), err))
+		}
+		// During upgrades, nftables blocks Router Advertisements (RAs) from other nodes.
+		// However, Virtual Machines (VMs) may still retain old default gateway paths.
+		// To address this, we create a new Router Advertisement with a lifetime of 0
+		// to signal the removal of the old default gateway.
+		// NOTE: This is a workaround for the issue and may not be needed in the future, after
+		//       upgrading to a version that supports the new behavior.
+		ras = append(ras, newRouterAdvertisementFromJoinIPAndLifetime(nodeJoinAddrs[0].IP, destinationMAC, destinationIP.IP, 0))
+	}
+	targetNode, err := r.watchFactory.GetNode(liveMigration.TargetPod.Spec.NodeName)
+	if err != nil {
+		return fmt.Errorf("failed fetching node %q to reconcile ipv6 gateway: %w", liveMigration.TargetPod.Spec.NodeName, err)
+	}
+	targetNodeJoinAddrs, err := util.ParseNodeGatewayRouterJoinAddrs(targetNode, r.netInfo.GetNetworkName())
+	if err != nil {
+		return ovntypes.NewSuppressedError(fmt.Errorf("failed parsing join addresss from live migration target node %q and network %q to reconcile ipv6 gateway: %w", targetNode.Name, r.netInfo.GetNetworkName(), err))
+	}
+	ras = append(ras, newRouterAdvertisementFromJoinIPAndLifetime(targetNodeJoinAddrs[0].IP, destinationMAC, destinationIP.IP, 65535))
+	return ndp.SendRouterAdvertisements(r.interfaceName, ras...)
+}
+
+// newRouterAdvertisementFromJoinIPAndLifetime creates a new Router Advertisement (RA) message
+// using the provided join IP address, destination MAC, destination IP, and lifetime.
+//
+// This function performs the following:
+// - Derives the source MAC address from the given IP using util.IPAddrToHWAddr.
+// - Calculates the link-local address (LLA) from the source MAC using util.HWAddrToIPv6LLA.
+// - Configures the destination IP and MAC address to use the provided values.
+// - Sets the RA message's lifetime to the specified value.
+//
+// Parameters:
+// - ip: The join IP address used to derive the source MAC and LLA.
+// - destinationMAC: The MAC address to which the RA message will be sent.
+// - destinationIP: The IP address to which the RA message will be sent.
+// - lifetime: The lifetime value for the RA message, in seconds.
+//
+// Returns:
+// - An ndp.RouterAdvertisement object configured with the calculated source MAC, LLA, and the provided destination MAC, IP, and lifetime.
+func newRouterAdvertisementFromJoinIPAndLifetime(ip net.IP, destinationMAC net.HardwareAddr, destinationIP net.IP, lifetime uint16) ndp.RouterAdvertisement {
+	sourceMAC := util.IPAddrToHWAddr(ip)
+	return ndp.RouterAdvertisement{
+		SourceMAC:      sourceMAC,
+		SourceIP:       util.HWAddrToIPv6LLA(sourceMAC),
+		DestinationMAC: destinationMAC,
+		DestinationIP:  destinationIP,
+		Lifetime:       lifetime,
+	}
 }
--- a/go-controller/pkg/ovn/secondary_layer2_network_controller.go
+++ b/go-controller/pkg/ovn/secondary_layer2_network_controller.go
@@ -915,7 +915,7 @@ func (oc *SecondaryLayer2NetworkController) reconcileLiveMigrationTargetZone(kub
 	mgmtInterfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID()))
 
 	if hasIPv4Subnet, _ := oc.IPMode(); hasIPv4Subnet {
-		if err := kubevirt.ReconcileIPv4DefaultGatewayAfterLiveMigration(oc.watchFactory, oc.GetNetInfo(), kubevirtLiveMigrationStatus, mgmtInterfaceName); err != nil {
+		if err := kubevirt.NewDefaultGatewayReconciler(oc.watchFactory, oc.GetNetInfo(), mgmtInterfaceName).ReconcileIPv4AfterLiveMigration(kubevirtLiveMigrationStatus); err != nil {
 			return err
 		}
 	}
--- a/go-controller/pkg/ovn/secondary_layer2_network_controller.go
+++ b/go-controller/pkg/ovn/secondary_layer2_network_controller.go
@@ -293,6 +293,9 @@ type SecondaryLayer2NetworkController struct {
 
 	// EgressIP controller utilized only to initialize a network with OVN polices to support EgressIP functionality.
 	eIPController *EgressIPController
+
+	// reconcile the virtual machine default gateway sending GARPs and RAs
+	defaultGatewayReconciler *kubevirt.DefaultGatewayReconciler
 }
 
 // NewSecondaryLayer2NetworkController create a new OVN controller for the given secondary layer2 nad
@@ -363,6 +366,7 @@ func NewSecondaryLayer2NetworkController(
 		if err != nil {
 			return nil, fmt.Errorf("unable to create new service controller while creating new layer2 network controller: %w", err)
 		}
+		oc.defaultGatewayReconciler = kubevirt.NewDefaultGatewayReconciler(oc.watchFactory, oc.GetNetInfo(), util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID())))
 	}
 
 	if oc.allocatesPodAnnotation() {
@@ -908,15 +912,20 @@ func (oc *SecondaryLayer2NetworkController) updateLocalPodEvent(pod *corev1.Pod)
 }
 
 func (oc *SecondaryLayer2NetworkController) reconcileLiveMigrationTargetZone(kubevirtLiveMigrationStatus *kubevirt.LiveMigrationStatus) error {
-	// Only primary networks has a gateway to reconcile
-	if !oc.IsPrimaryNetwork() {
+	if oc.defaultGatewayReconciler == nil {
 		return nil
 	}
-	mgmtInterfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID()))
-
-	if hasIPv4Subnet, _ := oc.IPMode(); hasIPv4Subnet {
-		if err := kubevirt.NewDefaultGatewayReconciler(oc.watchFactory, oc.GetNetInfo(), mgmtInterfaceName).ReconcileIPv4AfterLiveMigration(kubevirtLiveMigrationStatus); err != nil {
-			return err
+	hasIPv4Subnet, hasIPv6Subnet := oc.IPMode()
+	if hasIPv4Subnet {
+		if err := oc.defaultGatewayReconciler.ReconcileIPv4AfterLiveMigration(kubevirtLiveMigrationStatus); err != nil {
+			return fmt.Errorf("failed reconciling IPv4 default gw after live migration at target pod '%s/%s': %w",
+				kubevirtLiveMigrationStatus.TargetPod.Namespace, kubevirtLiveMigrationStatus.TargetPod.Name, err)
+		}
+	}
+	if hasIPv6Subnet {
+		if err := oc.defaultGatewayReconciler.ReconcileIPv6AfterLiveMigration(kubevirtLiveMigrationStatus); err != nil {
+			return fmt.Errorf("failed reconciling IPv6 default gw after live migration at target pod '%s/%s': %w",
+				kubevirtLiveMigrationStatus.TargetPod.Namespace, kubevirtLiveMigrationStatus.TargetPod.Name, err)
 		}
 	}
 	return nil
--- a/go-controller/pkg/util/ndp/ra.go
+++ b/go-controller/pkg/util/ndp/ra.go
@@ -1,0 +1,125 @@
+package ndp
+
+import (
+	"fmt"
+	"net"
+	"syscall"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/mdlayher/socket"
+	"golang.org/x/sys/unix"
+)
+
+// RouterAdvertisement with mac, ips and lifetime field to send
+type RouterAdvertisement struct {
+	SourceMAC, DestinationMAC net.HardwareAddr
+	SourceIP, DestinationIP   net.IP
+	Lifetime                  uint16
+}
+
+// SendRouterAdvertisements sends one or more Router Advertisements (RAs) on the specified network interface.
+// This function requires raw socket capabilities because the source MAC and IP addresses in the RAs
+// are not the ones from the interface used to send the packets.
+//
+// Parameters:
+// - interfaceName: The name of the network interface to send the RAs on.
+// - ras: A variadic list of RouterAdvertisement objects containing the details of each RA to be sent.
+//
+// Returns:
+// - error: An error object if an error occurs, otherwise nil.
+//
+// The function performs the following steps:
+// 1. Retrieves the network interface by name.
+// 2. Creates a raw socket for sending packets.
+// 3. Serializes each Router Advertisement into a byte slice.
+// 4. Sends the serialized RAs using the raw socket.
+func SendRouterAdvertisements(interfaceName string, ras ...RouterAdvertisement) error {
+	iface, err := net.InterfaceByName(interfaceName)
+	if err != nil {
+		return fmt.Errorf("failed to find interface %s: %w", interfaceName, err)
+	}
+	c, err := socket.Socket(syscall.AF_PACKET, syscall.SOCK_RAW, syscall.ETH_P_ALL, "ra", nil)
+	if err != nil {
+		return fmt.Errorf("failed to create raw socket to send unsolicited RAs: %w", err)
+	}
+	defer c.Close()
+
+	serializedRAs := [][]byte{}
+	for _, ra := range ras {
+		serializeBuffer := gopacket.NewSerializeBuffer()
+
+		// Create the Ethernet layer with destination and source MAC addresses.
+		ethernetLayer := layers.Ethernet{
+			DstMAC:       ra.DestinationMAC,
+			SrcMAC:       ra.SourceMAC,
+			EthernetType: layers.EthernetTypeIPv6,
+		}
+
+		// Create the IPv6 layer with source and destination IP addresses.
+		ip6Layer := layers.IPv6{
+			Version:    6,
+			NextHeader: layers.IPProtocolICMPv6,
+			HopLimit:   255,
+			SrcIP:      ra.SourceIP,
+			DstIP:      ra.DestinationIP,
+		}
+
+		// Create the ICMPv6 layer for the Router Advertisement.
+		icmp6Layer := layers.ICMPv6{
+			TypeCode: layers.CreateICMPv6TypeCode(layers.ICMPv6TypeRouterAdvertisement, 0),
+		}
+		if err := icmp6Layer.SetNetworkLayerForChecksum(&ip6Layer); err != nil {
+			return err
+		}
+
+		// https://datatracker.ietf.org/doc/html/rfc4861#section-4.2
+		// Managed address configuration flag.
+		managedAddressFlag := uint8(0x80)
+
+		// https://datatracker.ietf.org/doc/html/rfc4191#section-2.2
+		// Prf (Default Router Preference)
+		//   2-bit signed integer.  Indicates whether to prefer this
+		//    router over other default routers.  If the Router Lifetime
+		//    is zero, the preference value MUST be set to (00) by the
+		//    sender and MUST be ignored by the receiver.  If the Reserved
+		//    (10) value is received, the receiver MUST treat the value as
+		//    if it were (00).
+		defaultRoutePreferenceFlag := uint8(0x08)
+		if ra.Lifetime == 0 {
+			defaultRoutePreferenceFlag = uint8(0x00)
+		}
+
+		// Create the ICMPv6 Router Advertisement layer.
+		raLayer := layers.ICMPv6RouterAdvertisement{
+			HopLimit:       255,
+			Flags:          managedAddressFlag | defaultRoutePreferenceFlag,
+			RouterLifetime: ra.Lifetime,
+			ReachableTime:  0,
+			RetransTimer:   0,
+			Options: layers.ICMPv6Options{{
+				Type: layers.ICMPv6OptSourceAddress,
+				Data: ra.SourceMAC,
+			}},
+		}
+
+		// Serialize the layers into a byte slice.
+		if err := gopacket.SerializeLayers(serializeBuffer, gopacket.SerializeOptions{ComputeChecksums: true, FixLengths: true},
+			&ethernetLayer,
+			&ip6Layer,
+			&icmp6Layer,
+			&raLayer,
+		); err != nil {
+			return err
+		}
+		serializedRAs = append(serializedRAs, serializeBuffer.Bytes())
+	}
+
+	// Send each serialized Router Advertisement using the raw socket.
+	for _, serializedRA := range serializedRAs {
+		if err := c.Sendto(serializedRA, &unix.SockaddrLinklayer{Ifindex: iface.Index}, 0); err != nil {
+			return err
+		}
+	}
+	return nil
+}
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -1555,7 +1555,7 @@ runcmd:
 				l[RequiredUDNNamespaceLabel] = ""
 			}
 			ns, err := fr.CreateNamespace(context.TODO(), fr.BaseName, l)
-			Expect(err).ToNot(HaveOccurred())
+			Expect(err).NotTo(HaveOccurred())
 			fr.Namespace = ns
 			namespace = fr.Namespace.Name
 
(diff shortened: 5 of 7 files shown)


qinqon:
Know flake: 
- https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4936

test/e2e/kubevirt.go
```diff
@@ -1665,22 +1665,46 @@ runcmd:
...
+
+					expectedGatewayMAC, err := kubevirt.GenerateGatewayMAC(targetNode, netConfig.networkName)
+					Expect(err).ToNot(HaveOccurred(), step)
+
+					Expect(err).ToNot(HaveOccurred(), step)
+					Eventually(kubevirt.RetrieveCachedGatewayMAC).
+						WithArguments(vmi, "enp1s0", cidrIPv4).
+						WithTimeout(10*time.Second).
+						WithPolling(time.Second).
+						Should(Equal(expectedGatewayMAC), step)
```
oshoval, line 1684: please consider wrapping in some helper functions to make it shorter and ease readability

(same for the snippet below for isIPv6Supported)

oshoval (commented):
Nice, thanks

go-controller/pkg/util/ndp/ra.go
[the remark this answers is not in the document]
qinqon: That was my first approach, but I needed more low level mechanism to "cook" the src ip and mac since we don't want to send the RA with the management port src/ip.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -513,6 +514,48 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
 	}
-
 	return nil
 }
+
+func ReconcileIPv6DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFactory, netInfo util.NetInfo, liveMigration *LiveMigrationStatus, interfaceName string) error {
+	if liveMigration.State != LiveMigrationTargetDomainReady {
+		return nil
+	}
+	targetPodAnnotation, err := util.UnmarshalPodAnnotation(liveMigration.TargetPod.Annotations, netInfo.GetNADs()[0])
```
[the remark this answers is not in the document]
qinqon: Hummm, maybe  you right let me investigate the scenario where we have a primary UDN shared between namesapces, at the end he controller is per network not per NAD, and when primary UDN is shared between namespaces we have different NAD with the same network so...

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -896,5 +896,10 @@ func (oc *SecondaryLayer2NetworkController) reconcileLiveMigrationTargetZone(kub
 			return err
 		}
 	}
+	if hasIPv6Subnet, _ := oc.IPMode(); hasIPv6Subnet {
```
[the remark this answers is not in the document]
qinqon: We don't support layer2 topology for non IC, does not make sense to add a check here, if a lot of "checks" are already missing.

Also at non IC these RAs do not hurt.

test/e2e/kubevirt.go
qinqon: I don't see this code being called more than once.

oshoval:
Thanks, lgtm (limited of course to what i know)

qinqon:
After latest change to remove `GetNADs[0]` CI is unstable, the fix was looking for the active netinfo at the namespace and use it for GARP and unsolicited RAs

oshoval:
but isn't `targetPodAnnotation, err := util.UnmarshalPodAnnotation(liveMigration.TargetPod.Annotations, netInfo.GetNADs()[0])` problematic as well to keep ?
maybe need to iterate on them manually and find the required one ?

qinqon:
> but isn't `targetPodAnnotation, err := util.UnmarshalPodAnnotation(liveMigration.TargetPod.Annotations, netInfo.GetNADs()[0])` problematic as well to keep ? maybe need to iterate on them manually and find the required one ?

It's something else, I am manually generating RAs with lifetime 0 and looks like they don't remove the routes

qinqon:
> Thanks, so netInfo.GetNADs()[0]) should be changed to the fix ?

We needed the NAD to retrieve the mac address, but I think we can just broadcast the RA so no need for nad, it will simplify the code.

maiqueb:
I wonder about this explanation (maybe I'm missing something ...):
> so the issue is that fedora35 is quite old and don't fully implement IPv6 standard ignoring RA's lifetime=0

How did it pass before ?... Were you using a different image ?

test/e2e/kubevirt.go
```diff
@@ -1665,22 +1686,46 @@ runcmd:
...
+				}
+				if isIPv6Supported() {
+					step = by(vm.Name, fmt.Sprintf("Checking IPv6 gateway after %s %s", td.resource.description, td.test.description))
+					obtainedIPv6GatwayPaths, err := kubevirt.RetrieveIPv6GatwayPaths(vmi)
+					Expect(err).ToNot(HaveOccurred())
```
maiqueb, line 1710: nit: 
```suggestion
					Expect(err).NotTo(HaveOccurred())
```

test/e2e/kubevirt.go
```diff
@@ -1665,22 +1686,46 @@ runcmd:
...
+
+					sourceNode, err := fr.ClientSet.CoreV1().Nodes().Get(context.Background(), vmi.Status.MigrationState.SourceNode, metav1.GetOptions{})
+					Expect(err).ToNot(HaveOccurred(), step)
+
+					sourceNodeIPv6GatewayPath, err := kubevirt.GenerateGatwayIPv6RouterLLA(sourceNode, netConfig.networkName)
```
maiqueb, line 1715:
```suggestion
					sourceNodeIPv6GatewayPath, err := kubevirt.GenerateGatewayIPv6RouterLLA(sourceNode, netConfig.networkName)
```

test/e2e/kubevirt/net.go
```diff
@@ -62,3 +88,11 @@ func GenerateGatewayMAC(node *corev1.Node, networkName string) (string, error) {
 
 	return util.IPAddrToHWAddr(lrpJoinIP).String(), nil
 }
+
+func GenerateGatwayIPv6RouterLLA(node *corev1.Node, networkName string) (string, error) {
```
maiqueb, line 92:
```suggestion
func GenerateGatewayIPv6RouterLLA(node *corev1.Node, networkName string) (string, error) {
```

test/e2e/kubevirt/net.go
```diff
@@ -62,3 +88,11 @@ func GenerateGatewayMAC(node *corev1.Node, networkName string) (string, error) {
...
+	lrpAddresses, err := util.ParseNodeGatewayRouterJoinAddrs(node, networkName)
+	if err != nil {
+		return "", err
+	}
+	return util.HWAddrToIPv6LLA(util.IPAddrToHWAddr(lrpAddresses[0].IP)).String(), nil
```
maiqueb, line 97: IMHO you need to assert the nodes have join addresses. Either in the caller, or here, within the helper.

And please remind me: the join IP addresses will **always** be the first entry in the list, right ?

test/e2e/kubevirt/net.go
```diff
@@ -39,6 +41,30 @@ func RetrieveCachedGatewayMAC(vmi *kubevirtv1.VirtualMachineInstance, dev, cidr
 	return outputSplit[4], nil
 }
 
+func RetrieveIPv6GatwayPaths(vmi *v1.VirtualMachineInstance) ([]string, error) {
```
maiqueb, line 44:
```suggestion
func RetrieveIPv6GatewayPaths(vmi *v1.VirtualMachineInstance) ([]string, error) {
```

maiqueb: And actually, what do you think about naming this `RetrieveIPv6Gateways` ? The `path` suffix confuses me.

maiqueb (changes requested):
Really nice PR. 

I've caught something which I think is wrong (bad copy paste) and have a bunch of opinionated suggestions. 

Let me hear your thoughts !

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,78 @@
...
+		ip6Layer := layers.IPv6{
+			Version:    6,
+			NextHeader: layers.IPProtocolICMPv6,
+			HopLimit:   255,
+			SrcIP:      ra.SourceIP,
+			DstIP:      ra.DestinationIP,
+		}
+		icmp6Layer := layers.ICMPv6{
+			TypeCode: layers.CreateICMPv6TypeCode(layers.ICMPv6TypeRouterAdvertisement, 0),
+		}
```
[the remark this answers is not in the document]
qinqon: I prefer to have it here with the nice struct field names and all, no need for indirections if we are using this only once.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -513,6 +514,42 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
 			return err
 		}
 	}
-
 	return nil
 }
+
+func ReconcileIPv6DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFactory, netInfo util.NetInfo, liveMigration *LiveMigrationStatus, interfaceName string) error {
```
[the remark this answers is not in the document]
qinqon: This is going to be gone after transit router change, let's just do a free function here, so it's easier to remove later on.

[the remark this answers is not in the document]
maiqueb: Yeah, but we have literally no idea on when that's landing. I'm not saying this is a must, I'm just saying "we'll change this in the future" is not an amazing argument. Everything's subject to change.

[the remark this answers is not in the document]
qinqon: Hummm, ok let me do the struct with the ipv4 and ipv6 functions (at the end it looks good).

go-controller/pkg/kubevirt/pod.go
```diff
@@ -513,6 +514,42 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+}
+
+func newRouterAdvertisementFromJoinIPAndLifetime(ip net.IP, lifetime uint16) ndp.RouterAdvertisment {
+	ra := ndp.RouterAdvertisment{
+		SourceMAC:      util.IPAddrToHWAddr(ip),
+		DestinationMAC: net.HardwareAddr{0x33, 0x33, 0, 0, 0, 0x1},
+		DestinationIP:  net.IPv6linklocalallnodes,
+		Lifetime:       lifetime,
+	}
+	ra.SourceIP = util.HWAddrToIPv6LLA(ra.SourceMAC)
```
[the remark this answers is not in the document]
qinqon: fair enough

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -887,11 +887,15 @@ func (oc *SecondaryLayer2NetworkController) reconcileLiveMigrationTargetZone(kub
 		return nil
 	}
 	mgmtInterfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID()))
-
 	if hasIPv4Subnet, _ := oc.IPMode(); hasIPv4Subnet {
 		if err := kubevirt.ReconcileIPv4DefaultGatewayAfterLiveMigration(oc.watchFactory, oc.GetNetInfo(), kubevirtLiveMigrationStatus, mgmtInterfaceName); err != nil {
 			return err
 		}
 	}
+	if hasIPv6Subnet, _ := oc.IPMode(); hasIPv6Subnet {
```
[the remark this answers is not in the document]
qinqon: wow good catch thanks !

test/e2e/kubevirt/net.go
qinqon: > IMHO you need to assert the nodes have join addresses. Either in the caller, or here, within the helper.

ok. 


And please remind me: the join IP addresses will **always** be the first entry in the list, right ?

This is the node annotation for join addreses it just has join addreses this is not the LRP network addreses.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -292,6 +292,9 @@ type SecondaryLayer2NetworkController struct {
 
 	// EgressIP controller utilized only to initialize a network with OVN polices to support EgressIP functionality.
 	eIPController *EgressIPController
+
+	// reconcile the virtual machine default gateway sending GARPs and RAs
+	defaultGatewayReconciler *kubevirt.DefaultGatewayReconciler
```
maiqueb, line 297: nit: maybe defaultGateway is a bit redundant ? Isn't gateway enough ?

maiqueb (dismissed):
Thank you !

qinqon:
known flake:
- https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4936

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,79 @@
...
+		if err := icmp6Layer.SetNetworkLayerForChecksum(&ip6Layer); err != nil {
+			return err
+		}
+		managedAddressFlag := uint8(0x80)
+		defaultRoutePreferenceFlag := uint8(0x08)
```
kyrtapz, line 53: From the RFC:
https://datatracker.ietf.org/doc/html/rfc4191#section-2.2
```
   Prf (Default Router Preference)
            2-bit signed integer.  Indicates whether to prefer this
            router over other default routers.  If the Router Lifetime
            is zero, the preference value MUST be set to (00) by the
            sender and MUST be ignored by the receiver.  If the Reserved
            (10) value is received, the receiver MUST treat the value as
            if it were (00).
```

So we are sending incorrect route preference flags for the RAs with 0 lifetime.

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,79 @@
...
+			}},
+		}
+		serializeBuffer := gopacket.NewSerializeBuffer()
+		if err := gopacket.SerializeLayers(serializeBuffer, gopacket.SerializeOptions{ComputeChecksums: true, FixLengths: true},
+			&ethernetLayer,
+			&ip6Layer,
+			&icmp6Layer,
+			&raLayer,
+		); err != nil {
+			return err
```
kyrtapz, line 72: Would it be possible for us to Serialize all packets first and send them all after that?
If a failure occurs in consecutive RAs and we exit the function, we might clean up routes without sending the subsequent ones.

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,79 @@
...
+			SrcIP:      ra.SourceIP,
+			DstIP:      ra.DestinationIP,
+		}
+		icmp6Layer := layers.ICMPv6{
+			TypeCode: layers.CreateICMPv6TypeCode(layers.ICMPv6TypeRouterAdvertisement, 0),
+		}
+		if err := icmp6Layer.SetNetworkLayerForChecksum(&ip6Layer); err != nil {
+			return err
+		}
+		managedAddressFlag := uint8(0x80)
```
kyrtapz, line 52: Please add a comment explaining the used flags.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +504,52 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
-		if err := util.BroadcastGARP(interfaceName, garp); err != nil {
+		if err := util.BroadcastGARP(r.interfaceName, garp); err != nil {
 			return err
 		}
 	}
-
 	return nil
 }
+
+func (r *DefaultGatewayReconciler) ReconcileIPv6AfterLiveMigration(liveMigration *LiveMigrationStatus) error {
```
kyrtapz, line 520: I think this would be a good place to provide a comment around the approach you took to clean up the extra routes. Basically explaining what and why we send the RAs after migration.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +504,52 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+	if err != nil {
+		return err
+	}
+	return ndp.SendRouterAdvertisments(r.interfaceName,
+		newRouterAdvertisementFromJoinIPAndLifetime(sourceNodeJoinAddrs[0].IP, 0),
+		newRouterAdvertisementFromJoinIPAndLifetime(targetNodeJoinAddrs[0].IP, 65535),
+	)
+}
+
+func newRouterAdvertisementFromJoinIPAndLifetime(ip net.IP, lifetime uint16) ndp.RouterAdvertisment {
```
kyrtapz, line 546: Could you please add a docblock to this function explaining the purpose of using multicast?
I know the function is private but I am sure I will forget the details tomorrow :dizzy:

go-controller/pkg/kubevirt/types.go
```diff
@@ -12,3 +14,17 @@ const (
 	NamespaceExternalIDsKey      = "k8s.ovn.org/namespace"
 	VirtualMachineExternalIDsKey = "k8s.ovn.org/vm"
 )
+
+type DefaultGatewayReconciler struct {
```
kyrtapz, line 18: NIT: missing docblock

test/e2e/kubevirt.go
```diff
@@ -1665,22 +1665,46 @@ runcmd:
...
-					WithPolling(time.Second).
-					Should(Equal(expectedGatewayMAC), step)
+					Expect(err).NotTo(HaveOccurred(), step)
+					Eventually(kubevirt.RetrieveCachedGatewayMAC).
+						WithArguments(vmi, "enp1s0", cidrIPv4).
+						WithTimeout(10*time.Second).
+						WithPolling(time.Second).
+						Should(Equal(expectedGatewayMAC), step)
+				}
+				if isIPv6Supported() {
```
kyrtapz, line 1707: Question: Do we run this test in CI on dualstack?

test/e2e/kubevirt.go
```diff
@@ -1665,22 +1665,46 @@ runcmd:
...
+					step = by(vm.Name, fmt.Sprintf("Checking IPv6 gateway after %s %s", td.resource.description, td.test.description))
+					obtainedIPv6Gateways, err := kubevirt.RetrieveIPv6Gateways(vmi)
```
kyrtapz, line 1709: Should we also verify the count of V6 gateways?

test/e2e/kubevirt/types.go
```diff
@@ -1,7 +1,8 @@
 package kubevirt
 
 const (
-	FedoraCoreOSContainerDiskImage = "quay.io/kubevirtci/fedora-coreos-kubevirt:v20230905-be4fa50"
-	FedoraContainerDiskImage       = "quay.io/kubevirtci/fedora-with-test-tooling:v20241128-4d4c8fe"
-	FakeLauncherImage              = "quay.io/nmstate/c10s-nmstate-dev:latest"
+	FedoraCoreOSContainerDiskImage          = "quay.io/kubevirtci/fedora-coreos-kubevirt:v20230905-be4fa50"
+	FedoraWithTestToolingContainerDiskImage = "quay.io/kubevirtci/fedora-with-test-tooling:v20241128-4d4c8fe"
+	FedoraContainerDiskImage                = "quay.io/containerdisks/fedora:39"
```
kyrtapz, line 6: Please add an explanation for moving to F39 in the commit message.
Is there any bug tracking the RA with lifetime 0 issue in Fedora? That would be useful as well.

kyrtapz (commented):
Thanks @qinqon!
I really appreciate the commit breakdown :)

test/e2e/kubevirt/types.go
maiqueb: I mean ... we were running fc35 - which is not supported since 2022-12-13 per the [fedora docs](https://docs.fedoraproject.org/en-US/releases/eol/).

Actually, even fc39 is currently unsupported. Couldn't we use 40 instead ?

qinqon: > I mean ... we were running fc35 - which is not supported since 2022-12-13 per the [fedora docs](https://docs.fedoraproject.org/en-US/releases/eol/).
> 
> Actually, even fc39 is currently unsupported. Couldn't we use 40 instead ?

If we use fedora40 then we need either https://github.com/kubevirt/kubevirt/pull/13458 (not present at v1.4.0 kubevirt) or disable the NetworkManager acd mechanism

kyrtapz: > I mean ... we were running fc35 - which is not supported since 2022-12-13 per the [fedora docs](https://docs.fedoraproject.org/en-US/releases/eol/).

Sure that's a valid reason, would be good to have it in the commit message.
I don't think that was the case here though? It would be great to have a tracker for the issue @qinqon mentioned.

qinqon: > > I mean ... we were running fc35 - which is not supported since 2022-12-13 per the [fedora docs](https://docs.fedoraproject.org/en-US/releases/eol/).
> 
> Sure that's a valid reason, would be good to have it in the commit message. I don't think that was the case here though? It would be great to have a tracker for the issue @qinqon mentioned.

Also I was thinking that would be good to backport https://github.com/kubevirt/kubevirt/pull/13458 to openshift virt 4.18 since this is more or less GA for primary UDNs ?

qinqon: Ahh we have kubevirt/kubevirt 1.5.0 hours ago, that should contain the ignore arp fix, let me try fedora 40, but still we need to backport the fix to 1.4 so it lands on openshift virt 4.18

qinqon: > > I mean ... we were running fc35 - which is not supported since 2022-12-13 per the [fedora docs](https://docs.fedoraproject.org/en-US/releases/eol/).
> 
> Sure that's a valid reason, would be good to have it in the commit message. I don't think that was the case here though? It would be great to have a tracker for the issue @qinqon mentioned.

@kyrtapz I don't find the bug since it's quite old stuff, is this mandatory ?

go-controller/pkg/util/ndp/ra.go
qinqon: Good catch! thanks

maiqueb: actually, could this be the reason why it doesn't work on fc35 ? Just wondering, I got no proof.

go-controller/pkg/util/ndp/ra.go
qinqon: Make sense, this has to be as much as possible all or nothing.

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,79 @@
...
+	"syscall"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/mdlayher/socket"
+	"golang.org/x/sys/unix"
+)
+
+// RouterAdvertisment with mac, ips and lifetime field to send
+type RouterAdvertisment struct {
```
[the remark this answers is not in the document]
qinqon: I am not using that library for this, since I need some more low level stuff, so do not make sense to use it just for their structs, also that would be an implementation details that we want to hide, like for example now we can change the library we use to implement this without affecting the caller code.

test/e2e/kubevirt.go
qinqon: Yep, kubevirt interconnect CI job is dual stack

test/e2e/kubevirt.go
qinqon: Didn't want to do it since that would change after merging 
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852
we just care for old node gateway path gone and new node gateway path configured.

go-controller/pkg/util/ndp/ra.go
qinqon: Humm good point, let me check.

go-controller/pkg/util/ndp/ra.go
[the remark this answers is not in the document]
kyrtapz: I am ok with keeping it as is if that is what you prefer but could you add doc lines for fields explaining their use? Especially because some of those are not attributes of the route advertisement.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+// mac/ip from the join ip address and adding the Lifetime is configuring the
+// destination ip/mac with ipv6 multicast all nodes, this way we don't need to
+// discover/pass the VM's network lla/mac, this RA is kind of idempotent so if
+// other VM's receive this it's ok, since it match with the network configuration
+func newRouterAdvertisementFromJoinIPAndLifetime(ip net.IP, lifetime uint16) ndp.RouterAdvertisement {
+	sourceMAC := util.IPAddrToHWAddr(ip)
+	return ndp.RouterAdvertisement{
+		SourceMAC:      sourceMAC,
+		SourceIP:       util.HWAddrToIPv6LLA(sourceMAC),
+		DestinationMAC: net.HardwareAddr{0x33, 0x33, 0, 0, 0, 0x1},
```
tssurya, line 573: to whom does this hardcoded MAC belong to ? can you add a comment here for that

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+// newRouterAdvertisementFromJoinIPAndLifetime apart from calculating the LLA and mac for source
+// mac/ip from the join ip address and adding the Lifetime is configuring the
+// destination ip/mac with ipv6 multicast all nodes, this way we don't need to
+// discover/pass the VM's network lla/mac, this RA is kind of idempotent so if
+// other VM's receive this it's ok, since it match with the network configuration
+func newRouterAdvertisementFromJoinIPAndLifetime(ip net.IP, lifetime uint16) ndp.RouterAdvertisement {
+	sourceMAC := util.IPAddrToHWAddr(ip)
+	return ndp.RouterAdvertisement{
+		SourceMAC:      sourceMAC,
+		SourceIP:       util.HWAddrToIPv6LLA(sourceMAC),
```
tssurya, line 572: so this is the LLA IP of the GR's? not the joinIP?

go-controller/pkg/kubevirt/types.go
```diff
@@ -12,3 +14,19 @@ const (
 	NamespaceExternalIDsKey      = "k8s.ovn.org/namespace"
 	VirtualMachineExternalIDsKey = "k8s.ovn.org/vm"
 )
+
+// DefaultGatewayReconciler implements a pair of Reconcile functions for
+// ipv4 and ipv6 to update VM's default gw config after live migration
+type DefaultGatewayReconciler struct {
```
tssurya, line 20: nit: hmm why is this in types.go file? the functions are living in pod.go? can we just co-locate them?
types.go from what I can tell is just constants?

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -882,14 +886,17 @@ func (oc *SecondaryLayer2NetworkController) updateLocalPodEvent(pod *corev1.Pod)
...
 		return nil
 	}
-	mgmtInterfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID()))
-
-	if hasIPv4Subnet, _ := oc.IPMode(); hasIPv4Subnet {
-		if err := kubevirt.ReconcileIPv4DefaultGatewayAfterLiveMigration(oc.watchFactory, oc.GetNetInfo(), kubevirtLiveMigrationStatus, mgmtInterfaceName); err != nil {
+	hasIPv4Subnet, hasIPv6Subnet := oc.IPMode()
+	if hasIPv4Subnet {
+		if err := oc.defaultGatewayReconciler.ReconcileIPv4AfterLiveMigration(kubevirtLiveMigrationStatus); err != nil {
+			return err
```
tssurya, line 895: can we wrap these into custom errors that can be easily spotted for debugging purposes?

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -361,6 +364,7 @@ func NewSecondaryLayer2NetworkController(
 		if err != nil {
 			return nil, fmt.Errorf("unable to create new service controller while creating new layer2 network controller: %w", err)
 		}
+		oc.defaultGatewayReconciler = kubevirt.NewDefaultGatewayReconciler(oc.watchFactory, oc.GetNetInfo(), util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID())))
```
tssurya, line 367: so this reconciler is always created but only used from live migration path so application only to live migrated vms right?

go-controller/pkg/util/ndp/ra.go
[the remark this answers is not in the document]
tssurya: let's add this info into the PR describe on why we didn't pick existing golang ndp wrapper pkg and did our own?
I am not sure I follow the management port src/ip aspect you are mentioning

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+	Lifetime                  uint16
+}
+
+// SendRouterAdvertisements will send variadic RAs.
+// Requires raw socket capabilities, since the RAs src MAC
+// and IP are not the ones on the LRP.
+func SendRouterAdvertisements(interfaceName string, ras ...RouterAdvertisement) error {
+	iface, err := net.InterfaceByName(interfaceName)
+	if err != nil {
+		return err
```
tssurya, line 26: will this error include the interfaceName?

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	c, err := socket.Socket(syscall.AF_PACKET, syscall.SOCK_RAW, syscall.ETH_P_ALL, "ra", nil)
+	if err != nil {
+		return err
```
tssurya, line 30: change it to 
return fmt.Errorf("failed to create raw socket for RA: %w", err) ?

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	defer c.Close()
```
tssurya, line 32: shouldn't the defer be as close as possible to the create, or is it guaranteed sock is always nil if err is not nil?

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+// RouterAdvertisement with mac, ips and lifetime field to send
+type RouterAdvertisement struct {
+	SourceMAC, DestinationMAC net.HardwareAddr
+	SourceIP, DestinationIP   net.IP
+	Lifetime                  uint16
+}
+
+// SendRouterAdvertisements will send variadic RAs.
+// Requires raw socket capabilities, since the RAs src MAC
+// and IP are not the ones on the LRP.
```
tssurya, line 22: logical router policy?
no
logical router port?
yes

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+	iface, err := net.InterfaceByName(interfaceName)
+	if err != nil {
+		return err
+	}
+	c, err := socket.Socket(syscall.AF_PACKET, syscall.SOCK_RAW, syscall.ETH_P_ALL, "ra", nil)
+	if err != nil {
+		return err
+	}
+	defer c.Close()
+	serializedRAs := [][]byte{}
```
tssurya, line 33: this logic and below - was it inspired from the ndp library? I dunno enough to  check its correctness

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+		// https://datatracker.ietf.org/doc/html/rfc4861
+		// 1-bit "Managed address configuration" flag.  When
+		// set, it indicates that addresses are available via
+		// Dynamic Host Configuration Protocol [DHCPv6].
+		managedAddressFlag := uint8(0x80)
+
+		//  https://datatracker.ietf.org/doc/html/rfc4191#section-2.2
+		//	Prf (Default Router Preference):
+		//		2-bit signed integer.  Indicates whether to prefer this
+		//		router over other default routers.  If the Router Lifetime
```
tssurya, line 64: thanks for the comments ++

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+	targetNode, err := r.watchFactory.GetNode(liveMigration.TargetPod.Spec.NodeName)
+	if err != nil {
+		return err
+	}
+	targetNodeJoinAddrs, err := util.ParseNodeGatewayRouterJoinAddrs(targetNode, r.netInfo.GetNetworkName())
+	if err != nil {
+		return err
+	}
+	ras = append(ras, newRouterAdvertisementFromJoinIPAndLifetime(targetNodeJoinAddrs[0].IP, 65535))
+	return ndp.SendRouterAdvertisements(r.interfaceName, ras...)
```
tssurya, line 560: I see some scope for unit testing some of these bits but I won't hold the PR for that given you do have e2e's

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+			&ethernetLayer,
+			&ip6Layer,
+			&icmp6Layer,
+			&raLayer,
+		); err != nil {
+			return err
+		}
+		serializedRAs = append(serializedRAs, serializeBuffer.Bytes())
+	}
+	for _, serializedRA := range serializedRAs {
```
tssurya, line 95: This file I heavily trust your skills haha
I wish we had a library help for sending out these RAs but ..

go-controller/pkg/util/ndp/ra.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+// RouterAdvertisement with mac, ips and lifetime field to send
+type RouterAdvertisement struct {
+	SourceMAC, DestinationMAC net.HardwareAddr
+	SourceIP, DestinationIP   net.IP
+	Lifetime                  uint16
+}
+
+// SendRouterAdvertisements will send variadic RAs.
```
tssurya, line 20: can you please add more than this as in it will open a new socket on the interface we pass and send the RAs :D

test/e2e/kubevirt.go
```diff
@@ -1705,22 +1726,38 @@ runcmd:
...
+					targetNode, err := fr.ClientSet.CoreV1().Nodes().Get(context.Background(), vmi.Status.MigrationState.TargetNode, metav1.GetOptions{})
+					Expect(err).NotTo(HaveOccurred(), step)
+
+					targetNodeIPv6GatewayPath, err := kubevirt.GenerateGatewayIPv6RouterLLA(targetNode, netConfig.networkName)
+					Expect(err).NotTo(HaveOccurred())
+					Eventually(kubevirt.RetrieveIPv6Gateways).
+						WithArguments(vmi).
+						WithTimeout(5*time.Second).
+						WithPolling(time.Second).
+						Should(Equal([]string{targetNodeIPv6GatewayPath}), "should reconcile ipv6 gateway nexthop after live migration")
```
tssurya, line 1759: thanks so this is the check that makes sure we only have 1 lifetime=max RA from the local node correct?

go-controller/go.mod
```diff
@@ -27,6 +27,7 @@ require (
 	github.com/k8snetworkplumbingwg/sriovnet v1.2.1-0.20230427090635-4929697df2dc
 	github.com/mdlayher/arp v0.0.0-20220512170110-6706a2966875
 	github.com/mdlayher/ndp v1.0.1
+	github.com/mdlayher/socket v0.2.1
```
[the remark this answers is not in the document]
qinqon: To be able to "cook" RAs, meaning sending RAs with changed src ip/mac we need raw sockets, this library is a raw sockets implementation from the same author.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+		if node.Name == liveMigration.TargetPod.Spec.NodeName {
+			// skip the target node since this is the proper gateway
+			continue
+		}
+		nodeJoinAddrs, err := util.ParseNodeGatewayRouterJoinAddrs(node, r.netInfo.GetNetworkName())
+		if err != nil {
+			return err
+		}
+		// although nftables is blocking the RA's from the other nodes during upgrades before this fix
+		// the VM's will still have the old default gw paths, so we need to remove it
```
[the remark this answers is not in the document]
qinqon: After the upgrade from these changes, sending to all the nodes is not really needed, but don't know if it's worth it to discover if we really need to send to all of them or not.

go-controller/go.mod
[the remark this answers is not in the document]
tssurya: no I meant why did it jump positions?

go-controller/pkg/kubevirt/pod.go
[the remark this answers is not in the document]
tssurya: so I think if you update your comment with that info its better because currently the comment states this is a must needed thing to be done

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+	}
+	ras := make([]ndp.RouterAdvertisement, 0, len(nodes))
+	for _, node := range nodes {
+		if node.Name == liveMigration.TargetPod.Spec.NodeName {
+			// skip the target node since this is the proper gateway
+			continue
+		}
+		nodeJoinAddrs, err := util.ParseNodeGatewayRouterJoinAddrs(node, r.netInfo.GetNetworkName())
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
qinqon: wrapping and using NewSuppressError to skip event, but we cannot skip the IsNotSet error since this has to be retried.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+		}
+		// although nftables is blocking the RA's from the other nodes during upgrades before this fix
+		// the VM's will still have the old default gw paths, so we need to remove it
+		ras = append(ras, newRouterAdvertisementFromJoinIPAndLifetime(nodeJoinAddrs[0].IP, 0))
+	}
+	targetNode, err := r.watchFactory.GetNode(liveMigration.TargetPod.Spec.NodeName)
```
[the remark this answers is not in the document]
qinqon: I prefer to fetch it again so if the target node we can easily throw and error, if we do this stuff in the loop then we have to check if we did found it after the loop, also I don't think adding the target node logic into the loop is a big performance gain.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
 		}
 	}
-
 	return nil
 }
+
+// ReconcileIPv6AfterLiveMigration will do two things at VM's:
+// - Remove ipv6 default gw path from VM's node before live migration
+// - Add ipv6 default gw path from VM's node after live migration
+// This is done by sending a pair of unsolicited RA's one with lifetime=0
```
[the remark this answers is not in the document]
qinqon: Fixing, this is from previous version where we were jus sending RA with lifetime=0 at the live migration source node

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+// (to remove the gateway path) another with lifetime=max to add the new
+// default gateway path
+func (r *DefaultGatewayReconciler) ReconcileIPv6AfterLiveMigration(liveMigration *LiveMigrationStatus) error {
+	if liveMigration.State != LiveMigrationTargetDomainReady {
+		return nil
```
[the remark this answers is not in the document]
qinqon: When target domain is ready, the target pod is annotation with a special kubevirt annotation, this will trigger reconcile and this code will be execute again.

go-controller/pkg/kubevirt/pod.go
qinqon: Right, my though is that for this "temporal" code e2e is enough.

go-controller/go.mod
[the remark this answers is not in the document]
maiqueb: Previously it was a dependency of some package we were importing, and now we're directly using it. Hence it got moved from "indirect" dependency, to a "direct" dependency.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+	}
+	nodes, err := r.watchFactory.GetNodes()
+	if err != nil {
+		return err
+	}
+	ras := make([]ndp.RouterAdvertisement, 0, len(nodes))
+	for _, node := range nodes {
```
maiqueb, line 538: while I don't see anything wrong w/ this approach, I think the code would be more readable if it looked something like (using some sort of pseudo-code):
```golang
targetNode := migrationTargetNode(liveMigration)
nonTargetNodes := nonTargetNodes(liveMigration)

ras := []ndp.RouterAdvertisement{configNewGateway(targetNode), forgetOldGateways(nonTargetNodes)...}
...
```

Or something along those lines. This is highly opinionated, feel free to disregard.

go-controller/pkg/kubevirt/pod.go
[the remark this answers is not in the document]
tssurya: ack

go-controller/pkg/kubevirt/pod.go
```diff
@@ -503,16 +507,71 @@ func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFa
...
+	}
+	ras = append(ras, newRouterAdvertisementFromJoinIPAndLifetime(targetNodeJoinAddrs[0].IP, 65535))
+	return ndp.SendRouterAdvertisements(r.interfaceName, ras...)
+}
+
+// newRouterAdvertisementFromJoinIPAndLifetime apart from calculating the LLA and mac for source
+// mac/ip from the join ip address and adding the Lifetime is configuring the
+// destination ip/mac with ipv6 multicast all nodes, this way we don't need to
+// discover/pass the VM's network lla/mac, this RA is kind of idempotent so if
+// other VM's receive this it's ok, since it match with the network configuration
```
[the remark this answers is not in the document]
qinqon: > I don't understand this last bit
> 
> so what you mean is actually we only want to send these for the VM being migrated but ofc all hosts on the network will receive this RA and that is still OK? this is making me think the following:
> 

As far as I remember on IC the RAs where not traversing the zones, so only the VMs running at the node will receive the unsolicited RA.

> * Imagine I have podA(vmA) on nodeA and podB(vmB) on nodeB
> * I live migrated both pods so now podA(vmA) shows up on nodeB and podB(vmB) shows up on nodeA
> * As a result here when the GRA sends out the RA with lifetime=0 won't it hurt the podB which shouldn't be receiving this but it will?

podB are not going to receive the RAs with lifetime=0 that should not since RAs are not traversing the zones.

go-controller/pkg/kubevirt/pod.go
qinqon: Yep, as by the IPv6 RFCs the RA's should use the LLAs as src ip instead of global IPs.

go-controller/pkg/kubevirt/pod.go
qinqon: I will add a comment about ipv6 mac multicast address

go-controller/pkg/ovn/secondary_layer2_network_controller.go
qinqon: Yep, we are guarding it later on with `IsPodAllowedForMigration`

go-controller/go.mod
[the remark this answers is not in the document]
qinqon: Yep, that's it, now the dependency is not transitive so it pops up at "direct" one.

go-controller/pkg/util/ndp/ra.go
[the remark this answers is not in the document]
qinqon: > let's add this info into the PR describe on why we didn't pick existing golang ndp wrapper pkg and did our own?  

I will add a note the PR description, just as a note we didn't create our own, we used the ones from `github.com/google/gopacket` (already included on the project) that allow us to do this kind of things.

> I am not sure I follow the management port src/ip aspect you are mentioning

with the `mdlayher/ndp` the src(ip, mac) from the RA is the one from the interface we are sending the message from, in our case the mgmt one, but instead we need them to be the gateway routers join IP LLA/MAC (this is how the VM's recognice what route path are we talking about).

go-controller/pkg/util/ndp/ra.go
qinqon: rewording, sinc eit does not make any sense.

go-controller/pkg/util/ndp/ra.go
qinqon: From what I see in example from the socket project they close after handling error, so I suppose this is fine 

https://github.com/mdlayher/socket/blob/main/internal/sockettest/sockettest.go#L32

go-controller/pkg/util/ndp/ra.go
qinqon: So I came up with thise using the ipv6toolkit to generate the unsolicited RAs that make the default gw multipath be reconciled, then used tcpdump to inspect them, then replicate whatever is needed with this library comparing tcpdumps.

The only validation we have that is ok is the e2e, ipv6 default gw is reconciled so we good.

go-controller/pkg/util/ndp/ra.go
qinqon: well now we have.

qinqon:
Addressing @tssurya comments at [force-push](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/17ab8ba5a4b517d2db7f6935277160bbe730125d..572278b3e237c9ce04592a07e8abd50562a96b74)

go-controller/pkg/kubevirt/pod.go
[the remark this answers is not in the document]
qinqon: I want to verify again that this unsolicited RAs are not traversing IC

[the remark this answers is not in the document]
qinqon: @tssurya good catch, I will narrow down dst ip/mac to VM's one.

We are suppose to block them at other zones with the nftables and all, but during upgrades who knows what will happend let's limit it.

go-controller/go.mod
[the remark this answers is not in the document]
tssurya: ahh sorry didn't notice that, thanks

go-controller/pkg/util/ndp/ra.go
tssurya: hmm why did you remove this on the latest push?

qinqon:
Last [force-push](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/572278b3e237c9ce04592a07e8abd50562a96b74..f6cef3296a3d376e7f5bf0e970f8a9a3582737cc):
- Fix lint issues
- Send RAs only to the VM instead of multicast

go-controller/pkg/kubevirt/pod.go
qinqon: Don't know if it's worth it for temporal code like this.

qinqon:
@tssurya this is ready for re review

go-controller/pkg/util/ndp/ra.go
qinqon: humm did I ? damn it

tssurya (approved):
lgtm
thank you @qinqon for the changes this is ready for merge my comments were addressed except https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4847/files#r2041530592 but that's fine its a nit
the other nit is you suppressed too much on the error from annotation parsing - what i mean was "annotation not found" suppression but again its a nit doesn't matter you can do a FUP if you push another PR
Once CI is green will merge this one

tssurya:
All 3 lanes that flaked in CI are known issues

1. https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14472019313/job/40590614857?pr=4847 is https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4474
2. https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14472019313/job/40590614820?pr=4847 is https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4220
3. https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14472019313/job/40590614816?pr=4847 is related to https://github.com/actions/runner-images/issues/11985

PR #4851, opened 2024-11-26
Add egress IP support for route advertisements

jcaamano:
Make cluster manager, zone and node EIP controllers aware of route
advertisements.

An egress IP can be either local to a node (the node is directly
connected to the egress IP network through the primary or a secondary
interface) or advertised through a node.

Cluster manager EIP controller checks if an egress IP that is not local
to any node can be instead advertised on one and if so, assigns it to
that node. For an egress IP to be advertised on a node, a route
advertisement configured to advertise egress IPs must select a network
that is serving any of the namespaces selected by the egress IP. If
different networks serve the namespaces selected by an egress IP, all
those networks must be configured to advertise egress IPs on at least
one common node for the egress IP to be assigned. 
To be able to reconcile updates correctly, the controller listens to 
network and namespace events.

Zone and node controllers require minor changes to merely treat
advertised egress IPs as if they were local to the primary network
interface.

Egress IP utility methods are reorganized in util/egressip/util.go

Add support to advertise EIPs for UDNs in cluster manager
RouteAdvertisements controller. Advertised EIPs are those that are
served on the same namespaces as networks configured to advertise EIPs.
To handle changes, the controller needs to track updates on the served
namespaces both by the EIPs and the networks.

Also updated the controller to track EIPs that are served through
secondary interfaces as those will not be advertised.

There is an existing limitation which is lack of support of EIP
advertisements on VFR-Lite scenario.


trozet:
"If different networks serve the namespaces selected by an egress IP, all
those networks must be configured to advertise egress IPs on at least
one common node for the egress IP to be assigned."

@jcaamano this sentence is a little confusing to me. I don't get why you are talking about "the egress IP to be assigned". I would think Egress IP assignment happens regardless of this PR in the same way it did before. Furthermore, using nodeSelector with egressIP advertisements seems tricky. It seems like it could over complicate things where there is a mismatch between nodes selected, and nodes that are labeled as egress-assignable. Then when an egress IP moves to a node not selected, suddenly we lose advertisement for that egressIP. Maybe we need to hav explicit documentation that your nodeSelector should include all nodes that are labeled as egress-assignable.

jcaamano:
I am closing this PR to keep history. I will be opening a new one.

PR #4852, opened 2024-11-26
UDN: Filter RA according local GW Router LLA only to avoid multi path

oshoval:
## 📑 Description

IPv6 Router Advertisements from non-local sources can interfere with primary UDN interfaces
in multi-gateway setups, resulting in multiple default gateways.

To resolve this, ingress filtering was added using nftables to block unwanted RAs,
allowing only RAs from the node's local gateway and RAs with Lifetime 0,
which are needed to invalidate entries during migration.

## ✅ Checks

- [x] My code requires tests

## How to verify it
The PR include tests

### DIFF
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -20,6 +20,7 @@ import (
 	"github.com/vishvananda/netlink"
 
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/knftables"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
@@ -250,6 +251,13 @@ func setupInterface(netns ns.NetNS, containerID, ifName string, ifInfo *PodInter
 		// to generate the unique host interface name, postfix it with the podInterface index for non-default network
 		if ifInfo.NetName != types.DefaultNetworkName {
 			ifnameSuffix = fmt.Sprintf("_%d", containerVeth.Index)
+		}
+
+		// If we have the ipv6 gateway LLA then this is a primary layer2 UDN
+		if len(ifInfo.GatewayIPv6LLA) > 0 {
+			if err = setupIngressFilter(ifName, ifInfo.GatewayIPv6LLA.String()); err != nil {
+				return err
+			}
 		}
 
 		return nil
@@ -597,6 +605,7 @@ func (*defaultPodRequestInterfaceOps) ConfigureInterface(pr *PodRequest, getter 
 			return nil, fmt.Errorf("unexpected configuration, pod request on dpu host. " +
 				"device ID must be provided")
 		}
+
 		// General case
 		hostIface, contIface, err = setupInterface(netns, pr.SandboxID, pr.IfName, ifInfo)
 	}
@@ -817,4 +826,69 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
 	for _, iface := range ifaces {
 		pr.deletePort(iface, podNamespace, podName)
 	}
+}
+
+// setupIngressFilter sets up an ingress filter using nftables to block
+// unwanted ICMPv6 Router Advertisement (RA) packets on a specific device.
+// It creates a new nftables table, chain, and rule to drop RA packets
+// that do not match the specified gateway link-local address (gwLLA) and
+// have a Router Advertisement (RA) lifetime not equal to 0.
+//
+// The nftables rule created by this function looks like:
+// `icmpv6 type nd-router-advert ip6 saddr != <gwLLA> @th,48,16 != 0 drop`
+//
+// Parameters:
+//   - ifaceName: The name of the network interface where the ingress filter
+//     should be applied.
+//   - gwLLA: The gateway link-local address to match against in the RA packets.
+//
+// Returns:
+// - error: An error if the nftables setup fails, otherwise nil.
+func setupIngressFilter(ifaceName, gwLLA string) error {
+	const (
+		tableName   = "ingress_filter" // Name of the nftables table to be created
+		rasLifetime = "@th,48,16"      // Offset for RA lifetime field in ICMPv6 packets
+	)
+
+	// Initialize a new nftables table for the ingress filter
+	nft, err := knftables.New(knftables.NetDevFamily, tableName)
+	if err != nil {
+		return fmt.Errorf("failed to initialize table: %w", err)
+	}
+
+	// Delegate the setup of the filter with the specified table and lifetime matcher
+	return setupIngressFilterWithTableAndLifetimeMatcher(nft, ifaceName, gwLLA, rasLifetime)
+}
+
+func setupIngressFilterWithTableAndLifetimeMatcher(nft knftables.Interface, ifaceName, gwLLA, rasLifetime string) error {
+	const (
+		chainName = "input"
+	)
+
+	tx := nft.NewTransaction()
+
+	tx.Add(&knftables.Table{})
+
+	tx.Add(&knftables.Chain{
+		Name:     chainName,
+		Type:     knftables.PtrTo(knftables.FilterType),
+		Hook:     knftables.PtrTo(knftables.IngressHook),
+		Priority: knftables.PtrTo(knftables.FilterPriority),
+		Device:   knftables.PtrTo(ifaceName),
+	})
+
+	tx.Add(&knftables.Rule{
+		Chain: chainName,
+		Rule: knftables.Concat(
+			"icmpv6", "type", "nd-router-advert", "ip6", "saddr", "!=", gwLLA, rasLifetime, "!=", "0", "drop",
+		),
+	})
+
+	// Execute the transaction
+	if err := nft.Run(context.Background(), tx); err != nil {
+		return fmt.Errorf("could not update netdev nftables: %v", err)
+	}
+
+	return nil
+
 }
--- a/go-controller/pkg/cni/helper_linux_test.go
+++ b/go-controller/pkg/cni/helper_linux_test.go
@@ -21,6 +21,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes/fake"
 	kexec "k8s.io/utils/exec"
+	"sigs.k8s.io/knftables"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/mocks"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/types"
@@ -1649,6 +1650,21 @@ func TestConfigureOVS_getPfEncapIpWithError(t *testing.T) {
 			mockLink.AssertExpectations(t)
 		})
 	}
+}
+
+func TestSetupIngressFilter(t *testing.T) {
+	nft := knftables.NewFake(knftables.NetDevFamily, "ingress_filter")
+
+	gwLLA := "fe80::1"
+	err := setupIngressFilterWithTableAndLifetimeMatcher(nft, "test-iface-1", gwLLA, "mark")
+	require.NoError(t, err)
+	output := nft.Dump()
+
+	expectedDump := `add table netdev ingress_filter
+add chain netdev ingress_filter input { type filter hook ingress device "test-iface-1" priority 0 ; }
+add rule netdev ingress_filter input icmpv6 type nd-router-advert ip6 saddr != fe80::1 mark != 0 drop`
+
+	assert.Equal(t, expectedDump, strings.TrimRight(output, "\n"), "The nftables dump output does not match the expected output")
 }
 
 // TODO(leih): Below functions are copied from pkg/node/base_node_network_controller_dpu_test.go.
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -1683,6 +1683,20 @@ runcmd:
 			nodeIPs := e2enode.CollectAddresses(nodes, v1.NodeInternalIP)
 
 			if td.role == "primary" {
+				if isIPv6Supported() && isInterconnectEnabled() {
+					step = by(vmi.Name, fmt.Sprintf("Checking IPv6 gateway before %s %s", td.resource.description, td.test.description))
+
+					nodeRunningVMI, err := fr.ClientSet.CoreV1().Nodes().Get(context.Background(), vmi.Status.NodeName, metav1.GetOptions{})
+					Expect(err).NotTo(HaveOccurred(), step)
+
+					expectedIPv6GatewayPath, err := kubevirt.GenerateGatewayIPv6RouterLLA(nodeRunningVMI, netConfig.networkName)
+					Expect(err).NotTo(HaveOccurred())
+					Eventually(kubevirt.RetrieveIPv6Gateways).
+						WithArguments(vmi).
+						WithTimeout(5*time.Second).
+						WithPolling(time.Second).
+						Should(Equal([]string{expectedIPv6GatewayPath}), "should filter remote ipv6 gateway nexthop")
+				}
 				step = by(vmi.Name, fmt.Sprintf("Check north/south traffic before %s %s", td.resource.description, td.test.description))
 				startNorthSouthIngressIperfTraffic(externalContainerName, nodeIPs, svc.Spec.Ports[0].NodePort, step)
 				checkNorthSouthIngressIperfTraffic(externalContainerName, nodeIPs, svc.Spec.Ports[0].NodePort, step)


go-controller/pkg/cni/helper_linux.go
```diff
@@ -817,3 +831,71 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+	})
+
+	tx.Add(&knftables.Rule{
+		Chain: chainName,
+		Rule: knftables.Concat(
+			"icmpv6", "type", "nd-router-advert", "ip6", "saddr", "!=", linkLocalIP, "@th,48,16", "!=", "0", "drop",
+		),
+	})
+
+	err = nft.Run(context.TODO(), tx)
```
maiqueb, line 894:
```suggestion
	err = nft.Run(context.Background(), tx)
```

go-controller/pkg/cni/helper_linux.go
```diff
@@ -817,3 +831,71 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+		Type:     knftables.PtrTo(knftables.FilterType),
+		Hook:     knftables.PtrTo(knftables.IngressHook),
+		Priority: knftables.PtrTo(knftables.FilterPriority),
+		Device:   knftables.PtrTo(deviceName),
+	})
+
+	tx.Add(&knftables.Rule{
+		Chain: chainName,
+		Rule: knftables.Concat(
+			"icmpv6", "type", "nd-router-advert", "ip6", "saddr", "!=", linkLocalIP, "@th,48,16", "!=", "0", "drop",
```
maiqueb, line 890: let's define `@th,48,16` as a properly named constant, so nobody's surprised by this option.

go-controller/pkg/cni/helper_linux.go
```diff
@@ -817,3 +831,71 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+
+	tx.Add(&knftables.Chain{
+		Name:     chainName,
+		Type:     knftables.PtrTo(knftables.FilterType),
+		Hook:     knftables.PtrTo(knftables.IngressHook),
+		Priority: knftables.PtrTo(knftables.FilterPriority),
+		Device:   knftables.PtrTo(deviceName),
+	})
+
+	tx.Add(&knftables.Rule{
```
maiqueb, line 887: please encapsulate this in an helper.

go-controller/pkg/cni/helper_linux.go
```diff
@@ -252,6 +253,19 @@ func setupInterface(netns ns.NetNS, containerID, ifName string, ifInfo *PodInter
 			ifnameSuffix = fmt.Sprintf("_%d", containerVeth.Index)
 		}
 
+		if ifName == "ovn-udn1" {
```
maiqueb, line 256: uf I don't like how this is done ... let's see if we can come up with a better way. 

For now, invest in the unit tests instead. Let's just keep thinking.

maiqueb (commented):
Very early feedback :)

go-controller/pkg/cni/helper_linux.go
oshoval: not sure i understand why, 
`setupIngressFilter` (line 865) is a function that encapsulate the nft creation
(init table, create tx, create table. chain, rule and apply)
isnt it enough ?

maiqueb: maybe; I just want to have a place to encapsulate the transaction, so you can unit test it's correctness.

go-controller/pkg/cni/helper_linux.go
oshoval: note that if we use bridge, we can create the table if it doesnt exists already, and then it doesnt need the ifname 

(we can even create bridge table on the OVN side where LLC exists if desired, in case it simplify things, if we will have the right network NS there)

maiqueb: > note that if we use bridge, we can create the table if it doesnt exists already, and then it doesnt need the ifname

That's a good argument for using bridge instead ...

> we can even create bridge table on the OVN side where LLC exists if desired, in case it simplify things, if we will have the right network NS there

Tapping into net namespaces is very expensive IIRC. We should try to do everything that requires accessing the net ns from a single place. 

I would prefer if you pushed the link local address to the CNI layer via network selection elements CNI args for IPv6 networks.

maiqueb: > ah oops, its not correct, we still need to create it only if we have some indication that we use primary UDN
which is what this condition also doing (please correct me if i am wrong)

That's what I really dislike about this comparison. You use the name of the interface to infer if primary UDN is used. 

Try to use something like https://github.com/ovn-kubernetes/ovn-kubernetes/blob/30585c17728872c1a1a6d199c3130998ac0f269c/go-controller/pkg/cni/cni.go#L174 (or pass along that value)

oshoval: @maiqueb @qinqon 
please let me know if you want CNI args instead so i can focus on one of the directions

maiqueb: I think if you use CNI args you'll do less in the CNI sandbox, which is better. You'll just consume the value previously made available by the ovnkube controller.

But how will you then know if you're in primary UDN ? Would it be the presence of the LRP LLA in the CNI args ?

qinqon: @maiqueb so the idea is to not spread code since this is going to be deleted after transit routers from OVN, so I think is better if we keep everything under CNI, so we don't have to change expectations from CNI args or change other packages.

go-controller/pkg/cni/helper_linux.go
oshoval: np will change, but note that all over we are using nft.Run with context.TODO

maiqueb: it's a nit, it's pretty much the same thing. 

Still, TODO means you mean to change it in the future for something else. Do you ? 
And FWIW, I actually think we should put a context w/ a timeout of a couple of seconds. Let's say 5. Arbitrary number.

Thoughts ?

go-controller/pkg/cni/helper_linux.go
maiqueb: @qinqon I understand the motive, but I don't buy the argument. 

Wouldn't we just revert the commit / whole PR once that lands ? (really sorry Or ...)

oshoval: Since we are going to revert it anyhow once the real fix is out, and since we also had an agreement about the current design, please consider to just keep going with the current one, time is also precious.
Moreover I also think that changing CNI args is more intrusive a little fwiw

go-controller/pkg/cni/helper_linux.go
oshoval: I don't think we should add timeout, while all the others don't have, and since we don't know about any problems with it as it is now.
Just if we will see any of those is doing problems we can add

go-controller/pkg/cni/helper_linux.go
maiqueb: isn't it easier to unit test the solution with CNI_ARGS ?

oshoval: possible it is, but not sure by heart yet
generally speaking, unit tests will be enough ? (i am against over testing personally)

oshoval:
Cherry picked https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4866
adapted the code to it, but please dont review yet, not ready

oshoval:
Rebased over https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878
so depend on that

go-controller/pkg/cni/helper_linux.go
oshoval: as concluded the annotation on the pod was added on the pending PR, and there will be a bool to add it only if primary exists

go-controller/pkg/cni/helper_linux.go
oshoval: used fake and added a unit test
did trick there with "mark"  because fake lib doesnt support the `@th`
`Error:      	Expected nil, but got: &errors.errorString{s:"could not update netdev nftables: no such set \"th,48,16\""}`

oshoval:
addressed comments
will update PR desc / commit soon

oshoval:
```
2024-12-31T15:07:37.7765736Z   [38;5;243m------------------------------[0m
2024-12-31T15:07:37.7767087Z   [38;5;204mProgress Report for Ginkgo Process #[1m3[0m
2024-12-31T15:07:37.7767822Z   [1m[38;5;214mA running node failed to exit in time[0m
2024-12-31T15:07:37.7769126Z   Ginkgo is moving on but a node has timed out and failed to exit before its grace period elapsed.  The node has now leaked and is running in the background.
2024-12-31T15:07:37.7770210Z   Here's a current progress report:
2024-12-31T15:07:37.7771846Z     [0m[sig-network] Services[0m [1m[38;5;214mshould be able to switch session affinity for NodePort service [LinuxOnly] [Conformance][0m (Spec Runtime: 1h39m1.898s)
2024-12-31T15:07:37.7773475Z       [38;5;243mk8s.io/kubernetes/test/e2e/network/service.go:2216[0m
2024-12-31T15:07:37.7774358Z       In [1m[38;5;214m[It][0m (Node Runtime: 1h39m1.544s)
2024-12-31T15:07:37.7777259Z         [38;5;243mk8s.io/kubernetes/test/e2e/network/service.go:2216[0m
2024-12-31T15:07:37.7779138Z         At [1m[38;5;214m[By Step] creating replication controller affinity-nodeport-transition in namespace services-6967[0m (Step Runtime: 1h39m1.342s)
2024-12-31T15:07:37.7780860Z           [38;5;243mk8s.io/kubernetes/test/e2e/framework/rc/rc_utils.go:88[0m
2024-12-31T15:07:37.7781250Z 
2024-12-31T15:07:37.7781561Z         [38;5;243mBegin Captured GinkgoWriter Output >>[0m
2024-12-31T15:07:37.7781931Z           [38;5;243m...[0m
2024-12-31T15:07:37.7783480Z           I1231 13:28:47.439542 48325 builder.go:121] Running '/usr/local/bin/kubectl --server=https://127.0.0.1:45891 --kubeconfig=/home/runner/ovn.conf --namespace=services-6967 exec execpod-affinityx2pgg -- /bin/sh -x -c echo hostName | nc -v -t -w 2 10.96.85.135 80'
2024-12-31T15:07:37.7785084Z           I1231 13:28:47.943621 48325 builder.go:146] stderr: "+ echo hostName\n+ nc -v -t -w 2 10.96.85.135 80\nConnection to 10.96.85.135 80 port [tcp/http] succeeded!\n"
2024-12-31T15:07:37.7786319Z           I1231 13:28:47.943662 48325 builder.go:147] stdout: "HTTP/1.1 400 Bad Request\r\nContent-Type: text/plain; charset=utf-8\r\nConnection: close\r\n\r\n400 Bad Request"
2024-12-31T15:07:37.7789790Z           I1231 13:28:47.943773 48325 builder.go:121] Running '/usr/local/bin/kubectl --server=https://127.0.0.1:45891 --kubeconfig=/home/runner/ovn.conf --namespace=services-6967 exec execpod-affinityx2pgg -- /bin/sh -x -c echo hostName | nc -v -t -w 2 172.18.0.3 32448'
2024-12-31T15:07:37.7794690Z           I1231 13:28:49.403405 48325 builder.go:146] stderr: "+ nc -v -t -w 2 172.18.0.3 32448\n+ echo hostName\nConnection to 172.18.0.3 32448 port [tcp/*] succeeded!\n"
2024-12-31T15:07:37.7798289Z           I1231 13:28:49.403441 48325 builder.go:147] stdout: "HTTP/1.1 400 Bad Request\r\nContent-Type: text/plain; charset=utf-8\r\nConnection: close\r\n\r\n400 Bad Request"
2024-12-31T15:07:37.7805852Z           I1231 13:28:49.403542 48325 builder.go:121] Running '/usr/local/bin/kubectl --server=https://127.0.0.1:45891 --kubeconfig=/home/runner/ovn.conf --namespace=services-6967 exec execpod-affinityx2pgg -- /bin/sh -x -c echo hostName | nc -v -t -w 2 172.18.0.4 32448'
2024-12-31T15:07:37.7808116Z           I1231 13:28:51.004756 48325 builder.go:146] stderr: "+ nc -v -t -w 2 172.18.0.4 32448\n+ echo hostName\nConnection to 172.18.0.4 32448 port [tcp/*] succeeded!\n"
2024-12-31T15:07:37.7809921Z           I1231 13:28:51.004807 48325 builder.go:147] stdout: "HTTP/1.1 400 Bad Request\r\nContent-Type: text/plain; charset=utf-8\r\nConnection: close\r\n\r\n400 Bad Request"
2024-12-31T15:07:37.7812177Z           I1231 13:28:51.033934 48325 builder.go:121] Running '/usr/local/bin/kubectl --server=https://127.0.0.1:45891 --kubeconfig=/home/runner/ovn.conf --namespace=services-6967 exec execpod-affinityx2pgg -- /bin/sh -x -c for i in $(seq 0 15); do echo; curl -q -s --connect-timeout 2 http://172.18.0.4:32448/ ; done'
2024-12-31T15:07:37.7814617Z         [38;5;243m<< End Captured GinkgoWriter Output[0m
2024-12-31T15:07:37.7814904Z 
2024-12-31T15:07:37.7815099Z         [1m[4mSpec Goroutine[0m
2024-12-31T15:07:37.7815527Z         [38;5;214mgoroutine 10762 [select, 99 minutes][0m
2024-12-31T15:07:37.7822234Z           [38;5;243mk8s.io/kubernetes/test/e2e/framework/kubectl.KubectlBuilder.ExecWithFullOutput({0xc001315200?, 0x0?})[0m
2024-12-31T15:07:37.7823530Z             [38;5;243mk8s.io/kubernetes/test/e2e/framework/kubectl/builder.go:129[0m
2024-12-31T15:07:37.7824786Z           [38;5;243mk8s.io/kubernetes/test/e2e/framework/kubectl.KubectlBuilder.Exec(...)[0m
2024-12-31T15:07:37.7826399Z             [38;5;243mk8s.io/kubernetes/test/e2e/framework/kubectl/builder.go:111[0m
2024-12-31T15:07:37.7828284Z           [38;5;243mk8s.io/kubernetes/test/e2e/framework/kubectl.RunKubectl({0xc001c87e03?, 0xc005689ac0?}, {0xc00008da88?, 0xc00008dae0?, 0xc00008dae8?})[0m
2024-12-31T15:07:37.7829307Z             [38;5;243mk8s.io/kubernetes/test/e2e/framework/kubectl/builder.go:158[0m
2024-12-31T15:07:37.7830924Z           [38;5;243mk8s.io/kubernetes/test/e2e/framework/pod/output.RunHostCmd(...)[0m
2024-12-31T15:07:37.7832344Z             [38;5;243mk8s.io/kubernetes/test/e2e/framework/pod/output/output.go:82[0m
2024-12-31T15:07:37.7833583Z         [38;5;214m[1m> k8s.io/kubernetes/test/e2e/network.affinityCheckFromPod.func1()[0m
2024-12-31T15:07:37.7834856Z             [38;5;214m[1mk8s.io/kubernetes/test/e2e/network/service.go:148[0m
2024-12-31T15:07:37.7835921Z         [38;5;214m[1m> k8s.io/kubernetes/test/e2e/network.checkAffinity.func1()[0m
2024-12-31T15:07:37.7837121Z             [38;5;214m[1mk8s.io/kubernetes/test/e2e/network/service.go:196[0m
2024-12-31T15:07:37.7838684Z           [38;5;243mk8s.io/apimachinery/pkg/util/wait.PollImmediate.ConditionFunc.WithContext.func1({0x0?, 0x0?})[0m
2024-12-31T15:07:37.7839639Z             [38;5;243mk8s.io/apimachinery/pkg/util/wait/wait.go:109[0m
2024-12-31T15:07:37.7840952Z           [38;5;243mk8s.io/apimachinery/pkg/util/wait.runConditionWithCrashProtectionWithContext({0x539ec50?, 0x74045e0?}, 0x4d88900?)[0m
2024-12-31T15:07:37.7841969Z             [38;5;243mk8s.io/apimachinery/pkg/util/wait/wait.go:154[0m
2024-12-31T15:07:37.7843138Z           [38;5;243mk8s.io/apimachinery/pkg/util/wait.poll({0x539ec50, 0x74045e0}, 0xa0?, 0xc005393c58, 0xc005393ca8)[0m
2024-12-31T15:07:37.7844394Z             [38;5;243mk8s.io/apimachinery/pkg/util/wait/poll.go:245[0m
2024-12-31T15:07:37.7845607Z           [38;5;243mk8s.io/apimachinery/pkg/util/wait.PollImmediateWithContext({0x539ec50?, 0x74045e0?}, 0x5340058?, 0x4565820?, 0xc004f7ab70?)[0m
2024-12-31T15:07:37.7846650Z             [38;5;243mk8s.io/apimachinery/pkg/util/wait/poll.go:175[0m
2024-12-31T15:07:37.7847571Z           [38;5;243mk8s.io/apimachinery/pkg/util/wait.PollImmediate(0xc001b2c908?, 0xc001c877f0?, 0xa?)[0m
2024-12-31T15:07:37.7848426Z             [38;5;243mk8s.io/apimachinery/pkg/util/wait/poll.go:157[0m
2024-12-31T15:07:37.7849739Z         [38;5;214m[1m> k8s.io/kubernetes/test/e2e/network.checkAffinity({0x7fdb10bd4ca8?, 0xc0015ca4e0?}, {0x53f04e8?, 0xc0056d56c0?}, 0xc001b2c908?, {0xc001c877f0?, 0x4d13bed?}, 0x10?, 0x0)[0m
2024-12-31T15:07:37.7850979Z             [38;5;214m[1mk8s.io/kubernetes/test/e2e/network/service.go:195[0m
2024-12-31T15:07:37.7852660Z         [38;5;214m[1m> k8s.io/kubernetes/test/e2e/network.execAffinityTestForNonLBServiceWithOptionalTransition({0x7fdb10bd4ca8, 0xc0015ca4e0}, 0x187bb7f?, {0x53f04e8, 0xc0056d56c0}, 0xc004c44788, 0x1)[0m
2024-12-31T15:07:37.7854451Z             [38;5;214m[1mk8s.io/kubernetes/test/e2e/network/service.go:4078[0m
2024-12-31T15:07:37.7855805Z         [38;5;214m[1m> k8s.io/kubernetes/test/e2e/network.execAffinityTestForNonLBServiceWithTransition(...)[0m
2024-12-31T15:07:37.7856843Z             [38;5;214m[1mk8s.io/kubernetes/test/e2e/network/service.go:4019[0m
2024-12-31T15:07:37.7857718Z         [38;5;214m[1m> k8s.io/kubernetes/test/e2e/network.init.func23.30({0x7fdb10bd4ca8, 0xc0015ca4e0})[0m
2024-12-31T15:07:37.7858610Z             [38;5;214m[1mk8s.io/kubernetes/test/e2e/network/service.go:2219[0m
2024-12-31T15:07:37.7859652Z           [38;5;243mgithub.com/onsi/ginkgo/v2/internal.extractBodyFunction.func2({0x53a8af8?, 0xc0015ca4e0?})[0m
2024-12-31T15:07:37.7861133Z             [38;5;243mgithub.com/onsi/ginkgo/v2@v2.19.0/internal/node.go:465[0m
2024-12-31T15:07:37.7862019Z           [38;5;243mgithub.com/onsi/ginkgo/v2/internal.(*Suite).runNode.func3()[0m
2024-12-31T15:07:37.7862967Z             [38;5;243mgithub.com/onsi/ginkgo/v2@v2.19.0/internal/suite.go:894[0m
2024-12-31T15:07:37.7864270Z           [38;5;243mgithub.com/onsi/ginkgo/v2/internal.(*Suite).runNode in goroutine 6722[0m
2024-12-31T15:07:37.7865626Z             [38;5;243mgithub.com/onsi/ginkgo/v2@v2.19.0/internal/suite.go:881[0m
2024-12-31T15:07:37.7866560Z   [38;5;243m------------------------------[0m
2024-12-31T15:07:55.3500120Z [38;5;10m↺ [FLAKEY TEST - TOOK 2 ATTEMPTS TO PASS] [5959.522 seconds][0m
2024-12-31T15:07:55.3501946Z [0m[sig-network] Services [38;5;10m[1mshould be able to switch session affinity for NodePort service [LinuxOnly] [Conformance][0m [38;5;204m[sig-network, Conformance][0m
2024-12-31T15:07:55.3503567Z [38;5;243mk8s.io/kubernetes/test/e2e/network/service.go:2216[0m
2024-12-31T15:07:55.3504340Z [38;5;243m------------------------------[0m
2024-12-31T15:07:55.6598973Z 
```

Not related to this PR, but worth to limit the time it waits, otherwise we can see this test took 1.6h and the lane got timeout

oshoval:
Updated PR desc / commit message

go-controller/pkg/cni/helper_linux.go
```diff
@@ -818,3 +843,48 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+	if err != nil {
+		return "", err
+	}
+
+	podNetworks, err := util.UnmarshalPodAnnotationAllNetworks(pod.Annotations)
+	if err != nil {
+		return "", err
+	}
+
+	return podNetworks[nadName].IPv6LLAGateway, err
```
maiqueb, line 858: are we sure we'll always find `podNetworks[nadName]` ?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -818,3 +843,48 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+}
+
+func setupIngressFilter(nft knftables.Interface, gwLLA string, lifetime string) error {
+	chainName := "input"
+	deviceName := "ovn-udn1"
```
maiqueb, line 863: nit: maybe consts ? :)

go-controller/pkg/cni/helper_linux.go
```diff
@@ -818,3 +843,48 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+
+	tx.Add(&knftables.Rule{
+		Chain: chainName,
+		Rule: knftables.Concat(
+			"icmpv6", "type", "nd-router-advert", "ip6", "saddr", "!=", gwLLA, lifetime, "!=", "0", "drop",
+		),
+	})
+
+	err := nft.Run(context.Background(), tx)
+	if err != nil {
```
maiqueb, line 885: nit: consider inlining

maiqueb (approved):
Looks good to me. 

Only minor nits.

go-controller/pkg/cni/helper_linux.go
oshoval: On non interconnect we won't have it due to the [bug](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876), Once that bug is fixed, we will be able to assert here in case desired.
The previous PR adds it on IC [only](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878/files#diff-3d3fe81a31268662d021e60eb37f4a1f7e96ddb7e61de7f49acbbd2ae82e81caR595).

Anyhow, if it is not in the map (and assuming we are on IC), it is an higher level bug because it shouldn't happen,
in this PR we just don't config the nftables in such [case](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852/files#diff-01a2ef868473804cd3d9b894b093e5be0735cfb21bf27f375ece3a33cc07edc9R256)

(p.s we just added a variable to the existing annotation)

oshoval:
rebased to solve conflict

oshoval:
addressed comments, thanks

maiqueb (changes requested):
Thanks. Nearly there, from my perspective.

go-controller/pkg/cni/types.go
```diff
@@ -166,6 +166,9 @@ type PodRequest struct {
 
 	// the DeviceInfo struct
 	deviceInfo nadapi.DeviceInfo
+
+	// indicates if interface is primaryUDN
```
martinkennelly, line 170: Format ?

go-controller/pkg/util/pod_annotation.go
```diff
@@ -191,6 +198,11 @@ func MarshalPodAnnotation(annotations map[string]string, podInfo *PodAnnotation,
 			NextHop: nh,
 		})
 	}
+
+	if podInfo.IPv6LLAGateway != nil {
```
martinkennelly, line 202: Validate its a valid ip before setting it ?

go-controller/pkg/cni/types.go
oshoval: do you mean to lower case IsPrimaryUDN ?

go-controller/pkg/util/pod_annotation.go
oshoval: no need imho, all validations are already done at the setter
see please https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878

oshoval:
will solve rebase conflicts once https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878 is merged

oshoval:
rebased on top of master branch + on top https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878

oshoval:
@ricky-rav @kyrtapz 
Once you want / can, please check the last commit of this PR (the others are cherry-pick of 
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878 which is still in review)
Unless you prefer that the above PR is merged first
Thanks

oshoval:
rebased on master
rebased on top of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878

go-controller/pkg/cni/helper_linux.go
oshoval: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852/commits/75c5988dc18db6b577ff900a96c12f5454a8dc1a#r1977501949

tssurya:
> ## 📑 Description
> Depends on #4878 Please see only last commit
> 
> IPv6 Router Advertisements from non-local sources can interfere with primary UDN interfaces in multi-gateway setups, resulting in multiple default gateways.
> 
> To resolve this, ingress filtering was added using nftables to block unwanted RAs, allowing only RAs from the node's local gateway and RAs with Lifetime 0, which are needed to invalidate entries during migration.

Ingress filtering was added where? on each pod's network namespace?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -252,6 +253,21 @@ func setupInterface(netns ns.NetNS, containerID, ifName string, ifInfo *PodInter
 			ifnameSuffix = fmt.Sprintf("_%d", containerVeth.Index)
 		}
 
+		if gwLLA != "" {
```
tssurya, line 256: This field won't be set for non-vms right? that's how you guarantee this has no effect for normal pods?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -597,8 +613,17 @@ func (*defaultPodRequestInterfaceOps) ConfigureInterface(pr *PodRequest, getter
 			return nil, fmt.Errorf("unexpected configuration, pod request on dpu host. " +
 				"device ID must be provided")
 		}
+
+		var gwLLA string
+		if pr.IsPrimaryUDN {
+			gwLLA, err = getGWLinkLocalAddress(pr.PodNamespace, pr.PodName, pr.nadName, getter)
```
tssurya, line 619: this is doing an extra unmarshalling of annotation for all primary UDN Pods? aren't there any scale implications from doing so?
can we restrict this to only VM pods please?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -818,3 +843,49 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+
+	podNetworks, err := util.UnmarshalPodAnnotationAllNetworks(pod.Annotations)
+	if err != nil {
+		return "", err
+	}
+
+	return podNetworks[nadName].GatewayIPv6LLA, err
+}
+
+func setupIngressFilter(nft knftables.Interface, gwLLA string, lifetime string) error {
```
tssurya, line 861: add doc string describing how your ingress filter logic looks like

go-controller/pkg/cni/types.go
```diff
@@ -166,6 +166,9 @@ type PodRequest struct {
...
+	IsPrimaryUDN bool
```
tssurya, line 171: @qinqon is this boolean really needed? isn't there another we used to infer if its a pudn pod or not? isn't that how we plug two interfaces at the same time?

go-controller/pkg/cni/helper_linux_test.go
```diff
@@ -1651,6 +1652,21 @@ func TestConfigureOVS_getPfEncapIpWithError(t *testing.T) {
 	}
 }
 
+func TestSetupIngressFilter(t *testing.T) {
```
tssurya, line 1655: thanks for the test!

tssurya (commented):
Is there any test that will actually show me this PR works?
Meaning a test that shows multi-path routes inside the pod and how that goes away with this and @qinqon 's PR together?
@qinqon 's PR is deliberately doing RAs for older node to lifetime=0 -> which seems to be doing it from all the nodes
and this PR is blocking all of that from within the pod's net namespace?
I see a UT with the rules which is awesome, but E2E? Is it easy to show the side effect somehow

go-controller/pkg/cni/helper_linux.go
qinqon: A layer2 pod would want to use RAs to configure itself, on that case this still applies

go-controller/pkg/cni/helper_linux.go
qinqon: We don't need that since PodAnnotation is already at PodInterfaceInfo, I am going to refactor this.

go-controller/pkg/cni/helper_linux_test.go
qinqon: I think is better to have an e2e, I don't see this adding much value.

maiqueb: I don't see them as mutually exclusive tbh. Why don't you add and e2e test and keep this one around ?

qinqon: I see the unit test quite intrussive since we need to pass the the testing function the table and the lifetime field matcher, at the end to just check the nftables rule is as expected, I know we can make it better by having two functions, on is the one we are going to unit test and the other is the production code one that do not need to pass the table and the matcher.

qinqon:
[force-push](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/73c277c2d6495c3a7a13f4304ad9528404010801..233beb5dc3058004a9fa96a42485f21f924ceb77):
- Add e2e
- Remove unit test
- Make code simpler by just checking GatewayIPv6 LLA from PodInterfaceInfo.PodAnnotation

go-controller/pkg/cni/helper_linux_test.go
maiqueb: Knowing you're generating the nftable rule you expect is valuable enough IMHO. We can discuss the function's signature though. 

I won't pretend I understood what you meant with "we can make it better by having two functions".

qinqon: Ok, I will put the unit test back with some refactoring so production code do not have to pass the table and the matcher.

maiqueb: Thank you, it does make sense to me.

go-controller/pkg/cni/helper_linux_test.go
```diff
@@ -1651,6 +1652,21 @@ func TestConfigureOVS_getPfEncapIpWithError(t *testing.T) {
...
+	nft := knftables.NewFake(knftables.NetDevFamily, "ingress_filter")
+
+	gwLLA := "fe80::1"
```
maiqueb, line 1658: nit: make this a const

go-controller/pkg/cni/helper_linux_test.go
```diff
@@ -1651,6 +1652,21 @@ func TestConfigureOVS_getPfEncapIpWithError(t *testing.T) {
...
+	err := setupIngressFilterWithTableAndLifetimeMatcher(nft, gwLLA, "mark")
+	require.NoError(t, err)
```
maiqueb, line 1660: nit: inline these 2 calls

go-controller/pkg/cni/helper_linux_test.go
```diff
@@ -1651,6 +1652,21 @@ func TestConfigureOVS_getPfEncapIpWithError(t *testing.T) {
...
+	output := nft.Dump()
+
+	expectedDump := `add table netdev ingress_filter
```
maiqueb, line 1663: nit

```suggestion
	expectedNFTRule := `add table netdev ingress_filter
```

go-controller/pkg/cni/helper_linux.go
```diff
@@ -818,3 +827,64 @@ func (pr *PodRequest) deletePorts(ifaces []string, podNamespace, podName string)
...
+	if err != nil {
+		return fmt.Errorf("failed to initialize table: %w", err)
+	}
+	return setupIngressFilterWithTableAndLifetimeMatcher(nft, gwLLA, rasLifetime)
+}
+
+func setupIngressFilterWithTableAndLifetimeMatcher(nft knftables.Interface, gwLLA, rasLifetime string) error {
+	const (
+		chainName  = "input"
+		deviceName = "ovn-udn1"
```
kyrtapz, line 861: NIT: This should be a constant.
Currently it is specified here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/1a5a6c27856d59fc31ac9b5c391b8d0db055701d/go-controller/pkg/cni/udn/primary_network.go#L29 but could be extracted so it can be reused. 
This can be a followup.

kyrtapz (dismissed):
LGTM

go-controller/pkg/cni/helper_linux_test.go
```diff
@@ -1651,6 +1652,21 @@ func TestConfigureOVS_getPfEncapIpWithError(t *testing.T) {
...
+add chain netdev ingress_filter input { type filter hook ingress device "ovn-udn1" priority 0 ; }
+add rule netdev ingress_filter input icmpv6 type nd-router-advert ip6 saddr != fe80::1 mark != 0 drop`
+
+	assert.Equal(t, expectedDump, strings.TrimRight(output, "\n"), "The nftables dump output does not match the expected output")
```
maiqueb, line 1667:
```suggestion
	assert.Equal(t, expectedNFTRule, strings.TrimRight(output, "\n"), "The generated nftables rules do not match the expectation")
```

go-controller/pkg/cni/helper_linux.go
```diff
@@ -252,6 +253,13 @@ func setupInterface(netns ns.NetNS, containerID, ifName string, ifInfo *PodInter
 			ifnameSuffix = fmt.Sprintf("_%d", containerVeth.Index)
 		}
 
+		// If we have the ipv6 gateway LLA then this is a primary layer2 UDN
+		if len(ifInfo.GatewayIPv6LLA) > 0 {
+			if err = setupIngressFilter(ifInfo.GatewayIPv6LLA.String()); err != nil {
+				return err
```
maiqueb, line 259: maybe wrap this error ? indicating we could not install the RA filtering rules ?

test/e2e/kubevirt.go
```diff
@@ -1662,6 +1662,20 @@ runcmd:
...
+					Expect(err).NotTo(HaveOccurred(), step)
+
+					expectedIPv6GatewayPath, err := kubevirt.GenerateGatewayIPv6RouterLLA(nodeRunningVMI, netConfig.networkName)
+					Expect(err).NotTo(HaveOccurred())
+					Eventually(kubevirt.RetrieveIPv6Gateways).
+						WithArguments(vmi).
+						WithTimeout(5*time.Second).
+						WithPolling(time.Second).
+						Should(Equal([]string{expectedIPv6GatewayPath}), "should filter remote ipv6 gateway nexthop")
+				}
```
maiqueb, line 1678: nit: maybe wrap this in an helper ?

test/e2e/kubevirt.go
```diff
@@ -1705,22 +1719,24 @@ runcmd:
...
 
-				Expect(err).ToNot(HaveOccurred(), step)
-				Eventually(kubevirt.RetrieveCachedGatewayMAC).
-					WithArguments(vmi, "enp1s0", cidrIPv4).
-					WithTimeout(10*time.Second).
-					WithPolling(time.Second).
-					Should(Equal(expectedGatewayMAC), step)
+					Expect(err).NotTo(HaveOccurred(), step)
+					Eventually(kubevirt.RetrieveCachedGatewayMAC).
+						WithArguments(vmi, "enp1s0", cidrIPv4).
```
maiqueb, line 1735: I worry about this hard-coded iface name.

This is prone to change. Can we retrieve this somehow, or directly influence the interface naming within the VM ? Using deterministic interface names.

go-controller/pkg/cni/helper_linux.go
qinqon: well this is called from `setupInterface` so clearly we should pass tha interface to this filter function, fixing...

test/e2e/kubevirt.go
qinqon: This is not worth it an indirection, is quite declartive, there is no logic here.

test/e2e/kubevirt.go
qinqon: This is at the original code, also we don't need this changes here since they are related to ipv4, I will remove the changes and leave like that, later on we can fix this at a proper PR.

maiqueb (dismissed):
Thank you.

go-controller/pkg/cni/helper_linux.go
```diff
@@ -252,6 +253,13 @@ func setupInterface(netns ns.NetNS, containerID, ifName string, ifInfo *PodInter
 			ifnameSuffix = fmt.Sprintf("_%d", containerVeth.Index)
 		}
 
+		// If we have the ipv6 gateway LLA then this is a primary layer2 UDN
```
tssurya, line 256: ++ oh yes looks much better now

test/e2e/kubevirt.go
```diff
@@ -1706,16 +1720,16 @@ runcmd:
 			}
 
 			if td.role == "primary" && td.test.description == liveMigrate.description && isIPv4Supported() && isInterconnectEnabled() {
-				step = by(vm.Name, fmt.Sprintf("Checking IPv4 gateway cached mac after %s %s", td.resource.description, td.test.description))
+				step = by(vmi.Name, fmt.Sprintf("Checking IPv4 gateway cached mac after %s %s", td.resource.description, td.test.description))
 				Expect(crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)).To(Succeed())
 
 				targetNode, err := fr.ClientSet.CoreV1().Nodes().Get(context.Background(), vmi.Status.MigrationState.TargetNode, metav1.GetOptions{})
-				Expect(err).ToNot(HaveOccurred(), step)
```
tssurya, line 1713: how were these things even compiling? ToNot is not a thing right?

tssurya (dismissed):
lgtm

test/e2e/kubevirt.go
maiqueb: unfortunately, it is ...

tssurya:
oops this needs a rebase :( I can't do anything

test/e2e/kubevirt.go
tssurya: oh wow haha

tssurya:
upgrade jobs are failing because of https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5183

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14479206134/job/40613810397?pr=4852
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14479206134/job/40613810404?pr=4852

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14479206134/job/40613810606?pr=4852 isknown issue with ubuntu kernel bug

tssurya (approved):
lgtm
thank you @oshoval

go-controller/pkg/cni/helper_linux.go
oshoval: [PR](https://github.com/kubernetes-sigs/knftables/pull/26) was opened to fix the knftables [issue](https://github.com/kubernetes-sigs/knftables/issues/22)

PR #4856, opened 2024-11-29
DNM: Localnet ux e2e test idempotent vlan creation create via bash

maiqueb:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/test/e2e/localnet-underlay.go
+++ b/test/e2e/localnet-underlay.go
@@ -2,11 +2,14 @@ package e2e
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


maiqueb:
The code introduced in #4807 requires NET_ADMIN / root since it creates a vlan / adds an ip address to it. 

Thus, the "create" operations are not suited to be executed with golang code. 

Let's try to use netlink to READ, and bash commands ran w/ sudo for creating ... 

A follow up PR can improve on this, making the tests themselves run with sudo.

maiqueb:
Superseded by #4807

PR #4859, opened 2024-12-03
[WIP] Refactor legacy informer/handler framework

trozet:
The legacy event driven informer/handler framework has flaws with performance. The regular "informer" object is not multi-threaded and handles events without a queue. The federated "queued informer" sends events to queues where worker threads pull items off and process them. In either informer version, the processing of objects is paused when a new handler is added to the informer. When a new handler is added, all existing objects in the cluster are processed as ADDs. This is extremely slow with UDN and adding new handlers on the fly as UDNs are created. Additionally, the previous model for queued informer did not actually stop the shared informer from enqueuing objects directly. It only paused the workers, which indirectly would eventually stop the shared informer from enqueuing after the channel buffer (size 10) was full.

To address this, the following has been changed:
 1. Classic informer type object is removed, all legacy informers now use a queued informer. These queued informers only use 1 worker thread.
 2. The queued informer now only enqueues events on initial add. The queue size has been adjusted to 10k to account for buffering initial add objects. Can tweak this later.
 3. Locking has been modified. The informer lock is now used to control event flow into the queues. In order to enqueue events from the informer cache, a read lock must be obtained. When a handler is added the informer lock is temporarily taken, to stop the informer cache so that initial objects can be enqueued. Additionally a new handlerMutex is introduced to control adding/removing handlers on an queued informer object.
 4. Since now all objects are enqueued when a handler is added and all handlers will iterate over these already processed objects, a map has been added to the handler. This processedInitialAdd map allows the handler to determine if it has already processed the object or not. In the case of existing handlers, they will ignore adds for objects they already procesed. For newly added handlers, they will ignore updates and deletes received before they have first added the object.

### DIFF
--- a/go-controller/pkg/factory/handler.go
+++ b/go-controller/pkg/factory/handler.go
@@ -48,6 +48,19 @@ type Handler struct {
 	// example: a handler with priority 0 will process the received event first
 	// before a handler with priority 1.
 	priority int
+
+	// processedInitialAdd used to determine if this handler has ever processed this object
+	// before attempting to handle any deletes/updates
+	processedInitialAdd *sync.Map
+}
+
+func (h *Handler) NumberOfProcessedItems() int {
+	count := 0
+	h.processedInitialAdd.Range(func(key, value interface{}) bool {
+		count++
+		return true
+	})
+	return count
 }
 
 func (h *Handler) OnAdd(obj interface{}, isInInitialList bool) {
@@ -96,6 +109,7 @@ type queueMapEntry struct {
 }
 
 type informer struct {
+	// informer mutex is used to control the flow of enqueuing events
 	sync.RWMutex
 	oType reflect.Type
 	inf   cache.SharedIndexInformer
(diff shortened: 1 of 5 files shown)


trozet:
unit tests are broken

testing downstream https://github.com/openshift/ovn-kubernetes/pull/2372

go-controller/pkg/factory/handler.go
```diff
@@ -585,28 +560,27 @@ func newQueuedInformer(oType reflect.Type, sharedInformer cache.SharedIndexInfor
...
 		// channel array.
 		for _, obj := range items {
-			addsMap.enqueueEvent(nil, obj, i.oType, false, func(e *event) {
+			objMeta, err := getObjectMeta(i.oType, obj)
+			if err != nil {
+				panic(err)
+			}
+			namespacedName := ktypes.NamespacedName{Namespace: objMeta.GetNamespace(), Name: objMeta.GetName()}
+			// we enqueue events on initial add directly, since on initial add we are holding informer lock
+			i.queueMap.enqueueEvent(nil, obj, i.oType, false, func(e *event) {
```
jcaamano, line 572: We will see how much we relied on the order in which we process events on startup (i.e. nodes before pods)

go-controller/pkg/factory/handler.go
```diff
@@ -233,7 +222,16 @@ func newQueueMap(numEventQueues uint32, wg *sync.WaitGroup, stopChan chan struct
...
+		// TODO(trozet): tweak this later
+		// we are constrained by how fast the worker threads can pull work off the queue before we can add more items
+		// into the queue. This becomes a sensitive area when we want to enqueue all initial objects. If the capacity is
+		// too low, we are worker thread bound to be able to add more objects for initial add. If the number is too high,
+		// we are consuming too much memory.
+		//
+		// If the overall workers are too slow to keep up with the rate of adds, it will still take just as long to process
+		// all the work. However, the add handler with initial objects will finish faster with a larger queue, and then
+		// other handlers can be added without waiting.
+		qm.queues[j] = make(chan *event, 10000)
```
jcaamano, line 234: A clear sign that we should avoid using this framework in cluster manager. Hoping that at node we filter out uninteresting objects fast enough.

go-controller/pkg/factory/handler.go
```diff
@@ -48,6 +48,19 @@ type Handler struct {
 	// example: a handler with priority 0 will process the received event first
 	// before a handler with priority 1.
 	priority int
+
+	// processedInitialAdd used to determine if this handler has ever processed this object
+	// before attempting to handle any deletes/updates
+	processedInitialAdd *sync.Map
```
jcaamano, line 54: Let's see what is the price to pay in memory consumption here. Why a pointer though?

PR #4861, opened 2024-12-03
[WIP] Performance improvements

kyrtapz:
Based on ideas from https://github.com/trozet/ovn-kubernetes/tree/split_udn_node


go-controller/pkg/retry/obj_retry.go
```diff
@@ -492,6 +493,9 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
 		labelSelectorForFilteredHandler, // filter out objects not matching these labels
 		cache.ResourceEventHandlerFuncs{
 			AddFunc: func(obj interface{}) {
+				if !r.ResourceHandler.FilterResource(obj) {
```
trozet, line 496: what is the meaning of "Filter Resource"? I would think filtering means to "filter it out" and not process it, but these checks seem to have the inverse behavior. If we do not filter it, then we return?

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -199,6 +199,40 @@ func (oc *BaseSecondaryNetworkController) Reconcile(netInfo util.NetInfo) error
...
+		return true
+	}
+	switch objType {
+	case factory.NamespaceType:
+		ns, ok := obj.(*kapi.Namespace)
+		if !ok {
+			klog.Errorf("Failed to cast the provided object to a namespace")
+			return false
+		}
+		namespaceNet, err := oc.networkManager.GetActiveNetworkForNamespace(ns.Name)
```
trozet, line 213: rather than rely on active network for namespace, you can just do oc.GetNamespaces() and check if the namespace is in the list.

trozet (changes requested):

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -199,6 +199,40 @@ func (oc *BaseSecondaryNetworkController) Reconcile(netInfo util.NetInfo) error
 	return util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
 }
 
+func (oc *BaseSecondaryNetworkController) FilterResource(objType reflect.Type, obj interface{}) bool {
+	if !oc.IsPrimaryNetwork() {
+		return true
+	}
```
trozet, line 205: why this check?

go-controller/pkg/retry/obj_retry.go
kyrtapz: I agree this is confusing, I have changed it to `FilterOutResource` that returns true to exclude an object.

go-controller/pkg/ovn/base_network_controller.go
kyrtapz: Good idea, this also addresses https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4861#discussion_r1911318631 as it should work for all secondary networks.

go-controller/pkg/ovn/base_network_controller.go
kyrtapz: I removed it but I had to modify the multinet policy unit tests, @pperiyasamy @npinaeva ptal

pperiyasamy: yes @kyrtapz , that was a bug with unit test, both namespaces must be having same NAD config referring to same network to get multinet policy to be really effective for the network.

go-controller/pkg/ovn/ovn.go
```diff
@@ -406,35 +406,46 @@ func (oc *DefaultNetworkController) syncNodeGateway(node *kapi.Node, hostSubnets
...
-	macAddress, _ := util.ParseNodeManagementPortMACAddresses(node, netName)
-	return !bytes.Equal(oldMacAddress, macAddress)
+	var oldMacAddress, newMacAddress map[string]json.RawMessage
+
+	if err := json.Unmarshal([]byte(oldNode.Annotations[util.OvnNodeManagementPortMacAddresses]), &oldMacAddress); err != nil {
+		klog.Errorf("Failed to unmarshal old node %s annotation: %v", oldNode.Name, err)
+		return true
+	}
+	if err := json.Unmarshal([]byte(node.Annotations[util.OvnNodeManagementPortMacAddresses]), &newMacAddress); err != nil {
+		klog.Errorf("Failed to unmarshal new node %s annotation: %v", node.Name, err)
```
jcaamano, line 427: Not sure how useful it is going to be to return true when the new value fails to parse. Assuming that processing the corresponding event will also fail and we get a barrage of useless retries. Same comment for other instances. For your consideration, not asking fro change.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -210,6 +210,42 @@ func (oc *BaseSecondaryNetworkController) Reconcile(netInfo util.NetInfo) error
 	return nil
 }
 
+func (oc *BaseSecondaryNetworkController) FilterOutResource(objType reflect.Type, obj interface{}) bool {
+	// Secondary networks that are not primary can potentially handle namespaces that are not a part of the networks
+	// NAD namespaces
+	if oc.IsPrimaryNetwork() {
+		return false
+	}
```
jcaamano, line 218: Since this is bound to be used in more places, what about encapsulating this in a new utility method `CanServeNamespace(network NetInfo, ns string) bool` in `multinetwork.go`:

- If the network is non-default primary it only returns true if the provided namespace is among the NAD namespaces. Otherwise false.
- If the network is not primary just return true

go-controller/pkg/retry/obj_retry.go
```diff
@@ -551,6 +555,9 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
 				})
 			},
 			UpdateFunc: func(old, newer interface{}) {
+				if r.ResourceHandler.FilterOutResource(newer) {
```
jcaamano, line 558: I know this works for the current intention of `FilterOutResource`, but the general logic I would use here is
```
if r.ResourceHandler.FilterOutResource(old) && r.ResourceHandler.FilterOutResource(newer) {
    return
}
```

go-controller/pkg/retry/obj_retry.go
```diff
@@ -710,6 +717,9 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
 				})
 			},
 			DeleteFunc: func(obj interface{}) {
+				if r.ResourceHandler.FilterOutResource(obj) {
```
jcaamano, line 720: Just giving visibility that this can be an object of type `DeletedFinalStateUnknown`. I guess we are not handling this correctly in general with the retry framework so not asking anything for this PR, but if we are not sure of the state of the object when it was deleted, we should try to process it anyway?

kyrtapz: Good point, FilterOutResource should not filter those out and the controller should get a chance to handle it.

go-controller/pkg/ovn/ovn.go
kyrtapz: We were ignoring errors previously but I agree that it doesn't make sense to return true if we failed to parse.

go-controller/pkg/ovn/ovn.go
```diff
@@ -406,35 +406,46 @@ func (oc *DefaultNetworkController) syncNodeGateway(node *kapi.Node, hostSubnets
...
 // macAddressChanged() compares old annotations to new and returns true if something has changed.
 func macAddressChanged(oldNode, node *kapi.Node, netName string) bool {
-	oldMacAddress, _ := util.ParseNodeManagementPortMACAddresses(oldNode, netName)
-	macAddress, _ := util.ParseNodeManagementPortMACAddresses(node, netName)
-	return !bytes.Equal(oldMacAddress, macAddress)
+	var oldMacAddress, newMacAddress map[string]json.RawMessage
+
+	if err := json.Unmarshal([]byte(oldNode.Annotations[util.OvnNodeManagementPortMacAddresses]), &oldMacAddress); err != nil {
+		klog.Errorf("Failed to unmarshal old node %s annotation: %v", oldNode.Name, err)
+		return false
```
jcaamano, line 424: Being nitpicky, it might make sense to return true when failing to parse the old value. My previous comment was specific to the new value. Won't hold the PR on this however.

jcaamano (approved):

trozet (approved):

PR #4864, opened 2024-12-04
gateway: Access default net config under bridge lock.

dceara:
The netConfig map is accessed from multiple threads (e.g., for UDNs).

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -298,7 +298,7 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *kapi.Service, netInf
 				masqueradeSubnet = config.Gateway.V6MasqueradeSubnet
 			}
 			// table 2, user-defined network host -> OVN towards default cluster network services
-			defaultNetConfig := npw.ofm.defaultBridge.netConfig[types.DefaultNetworkName]
+			defaultNetConfig := npw.ofm.defaultBridge.getActiveNetworkBridgeConfig(types.DefaultNetworkName)
 
 			npw.ofm.updateFlowCacheEntry(key, []string{fmt.Sprintf("cookie=%s, priority=300, table=2, %s, %s_src=%s, %s_dst=%s, "+
 				"actions=set_field:%s->eth_dst,output:%s",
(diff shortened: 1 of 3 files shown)


dceara:
CC: @tssurya @kyrtapz

kyrtapz (approved):

go-controller/pkg/node/gateway_udn.go
```diff
@@ -121,10 +121,14 @@ func (b *bridgeConfiguration) delNetworkBridgeConfig(nInfo util.NetInfo) {
 // NOTE: if the network configuration can't be found or if the network is not patched by OVN
 // yet this returns nil.
 func (b *bridgeConfiguration) getActiveNetworkBridgeConfig(nInfo util.NetInfo) *bridgeUDNConfiguration {
+	return b.getActiveNetworkBridgeConfigByName(nInfo.GetNetworkName())
+}
+
+func (b *bridgeConfiguration) getActiveNetworkBridgeConfigByName(network string) *bridgeUDNConfiguration {
```
tssurya, line 127: nit: I would prefer not to add another util and instead just change the one other spot in the code :
```
func (c *openflowManager) getActiveNetwork(nInfo util.NetInfo) *bridgeUDNConfiguration {
	return c.defaultBridge.getActiveNetworkBridgeConfig(nInfo)
}
```
to pass the name of the network? looks like we don't really need to pass the whole netInfo anyways?

nit2: if you prefer to keep the ByName util over the other one change the network param to be networkName

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -298,7 +298,7 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *kapi.Service, netInf
 				masqueradeSubnet = config.Gateway.V6MasqueradeSubnet
 			}
 			// table 2, user-defined network host -> OVN towards default cluster network services
-			defaultNetConfig := npw.ofm.defaultBridge.netConfig[types.DefaultNetworkName]
+			defaultNetConfig := npw.ofm.defaultBridge.getActiveNetworkBridgeConfigByName(types.DefaultNetworkName)
```
tssurya, line 301: good catch :)

go-controller/pkg/node/gateway_udn.go
dceara: Makes sense, I changed it to take the networkName instead.

PR #4866, opened 2024-12-05
l2, udpn: Add ipv6 lla gateway field to the ovn pod annotation

qinqon:
## 📑 Description
For primary UDN layer2 the proper IPv6 gateway that should be configured
with received router advertisements is the one from the local GR IPv6
LLA, this change add field to the ovn pod annotations with that
information so it can be consumed by follow up PRs that will ensure
remove GRs are ignore as layer2 ipv6 gateways.

This is related to the following PRs that remove multipath ipv6 gateway for l2 primary UDNs:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4847

## ✅ Checks

- [x] My code requires tests

## How to verify it
It include unit tests

### DIFF
--- a/go-controller/pkg/allocator/pod/pod_annotation.go
+++ b/go-controller/pkg/allocator/pod/pod_annotation.go
@@ -54,6 +54,7 @@ func NewPodAnnotationAllocator(
 // false.
 func (allocator *PodAnnotationAllocator) AllocatePodAnnotation(
 	ipAllocator subnet.NamedAllocator,
+	node *v1.Node,
 	pod *v1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	reallocateIP bool,
@@ -67,6 +68,7 @@ func (allocator *PodAnnotationAllocator) AllocatePodAnnotation(
 		allocator.kube,
 		ipAllocator,
 		allocator.netInfo,
+		node,
 		pod,
 		network,
 		allocator.ipamClaimsReconciler,
@@ -80,6 +82,7 @@ func allocatePodAnnotation(
 	kube kube.Interface,
 	ipAllocator subnet.NamedAllocator,
 	netInfo util.NetInfo,
+	node *v1.Node,
 	pod *v1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	claimsReconciler persistentips.PersistentAllocations,
@@ -91,13 +94,13 @@ func allocatePodAnnotation(
 
 	// no id allocation
 	var idAllocator id.NamedAllocator
-
 	allocateToPodWithRollback := func(pod *v1.Pod) (*v1.Pod, func(), error) {
 		var rollback func()
 		pod, podAnnotation, rollback, err = allocatePodAnnotationWithRollback(
 			ipAllocator,
 			idAllocator,
 			netInfo,
+			node,
 			pod,
 			network,
 			claimsReconciler,
@@ -170,12 +173,18 @@ func allocatePodAnnotationWithTunnelID(
 	podAnnotation *util.PodAnnotation,
 	err error) {
 
+	node, err := kube.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return nil, nil, err
+	}
+
 	allocateToPodWithRollback := func(pod *v1.Pod) (*v1.Pod, func(), error) {
 		var rollback func()
 		pod, podAnnotation, rollback, err = allocatePodAnnotationWithRollback(
 			ipAllocator,
 			idAllocator,
 			netInfo,
+			node,
 			pod,
 			network,
 			claimsReconciler,
@@ -220,6 +229,7 @@ func allocatePodAnnotationWithRollback(
 	ipAllocator subnet.NamedAllocator,
 	idAllocator id.NamedAllocator,
 	netInfo util.NetInfo,
+	node *v1.Node,
 	pod *v1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	claimsReconciler persistentips.PersistentAllocations,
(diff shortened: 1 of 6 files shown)


go-controller/pkg/util/pod_annotation.go
```diff
@@ -555,6 +575,11 @@ func AddRoutesGatewayIP(
 					podAnnotation.Gateways = append(podAnnotation.Gateways, gatewayIPnet.IP)
 				}
 			}
+			joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
```
oshoval, line 578: this will fail on non IC, because it doesnt have the primary network in the map from what i saw
so i had to split the function and allow it to be missing 

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852/files#diff-01a2ef868473804cd3d9b894b093e5be0735cfb21bf27f375ece3a33cc07edc9R848-R858

oshoval: we should not fail for non IC if the IP is not there, because there is no multi path problem on non IC anyhow
of course that if it is non IC bug and it will be added then nothing need to be done
but the tests will fail now iiuc, if indeed non IC reach here (this what happened to me)

qinqon:
known flake:
- https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4387

go-controller/pkg/util/pod_annotation.go
maiqueb: I'd say we're pretty far behind on non-IC ...

Please open an upstream issue about this missing node annotation.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -555,6 +575,11 @@ func AddRoutesGatewayIP(
...
+			if err != nil {
+				return err
+			}
+			podAnnotation.IPv6LLAGateway = HWAddrToIPv6LLA(IPAddrToHWAddr(joinAddrs[0].IP))
```
oshoval, line 582: dont you need to add check like this ?
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852/files#diff-01a2ef868473804cd3d9b894b093e5be0735cfb21bf27f375ece3a33cc07edc9R859

(so ipv4 single stack wont have the new annotation even that it is derived from ipv4), else how will CNI know it shouldnt create the tables ?
(unless i am missing something, i am just learning the ropes here)

qinqon: Right, we should not added for single stack ipv4.

go-controller/pkg/util/pod_annotation.go
oshoval: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876

oshoval:
Given this NAD

```
metadata:                                                                       
  namespace: poc1
  name: l2
spec:                                                                           
  config: |2                                                                    
    {                                                                           
            "cniVersion": "0.3.0",                                              
            "name": "l2-network",                                               
            "topology":"layer2",                                                
            "subnets": "10.100.200.0/24,2014:100:200::0/60",                                       
            "netAttachDefName": "poc1/l2",                               
            "allowPersistentIPs": true,
            "role": "primary"
    }
```

The key is `namespace + / + name` (`poc1/l2` in our case)
it means that we would need on the CNI to know those (might be known via `PodRequest.nadName`, assuming it will always have the full namespaced name, even if namespace is default) 
but it might be better to just use the underline network name `l2-network` in this case ?
Because it is per real network, not per NAD basically
 
```
    k8s.ovn.org/pod-networks: '{"default":{"ip_addresses":["10.244.2.11/24","fd00:10:244:3::b/64"],"mac_address":"0a:58:0a:f4:02:0b","routes":[{"dest":"10.244.0.0/16","nextHop":"10.244.2.1"},{"dest":"100.64.0.0/16","nextHop":"10.244.2.1"},{"dest":"fd00:10:244::/48","nextHop":"fd00:10:244:3::1"},{"dest":"fd98::/64","nextHop":"fd00:10:244:3::1"}],"role":"infrastructure-locked"},"poc1/l2":{"ip_addresses":["10.100.200.3/24","2014:100:200::3/60"],"mac_address":"0a:58:0a:64:c8:03","gateway_ips":["10.100.200.1","2014:100:200::1"],"routes":[{"dest":"10.96.0.0/16","nextHop":"10.100.200.1"},{"dest":"100.65.0.0/16","nextHop":"10.100.200.1"},{"dest":"fd00:10:96::/112","nextHop":"2014:100:200::1"},{"dest":"fd99::/64","nextHop":"2014:100:200::1"}],"ipv6_lla_gateway_ip":"fe80::858:64ff:fe41:4","tunnel_id":6,"role":"primary"}}'
```

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -91,13 +91,17 @@ func allocatePodAnnotation(
 
 	// no id allocation
 	var idAllocator id.NamedAllocator
-
+	node, err := kube.GetNode(pod.Spec.NodeName)
```
kyrtapz, line 94: `kube.GetNode(pod.Spec.NodeName)` Will perform a direct API call, can't this be passed as a parameter instead to avoid calling the API on every `allocatePodAnnotation`?

kyrtapz (commented):
The general approach looks good to me,  I have one concern with getting the node.

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: Let me try

qinqon:
Closing, @oshoval is going to continue this with
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878

PR #4868, opened 2024-12-05
correctly remove stale EgressService rules at startup

danwinship:
Discovered while porting to nftables; there was a bug in the `FakeIPTables` implementation used for unit tests (when you called `List()` it would mutate its own internal state rather than only mutating its return value), and the EgressService controller startup cleanup code only worked correctly in the presence of that bug. (It passed the entire rule returned from `List()`, including the "`-A OVN-KUBE-EGRESS-SVC`", as the argument to `Delete()`, which ended up working with `FakeIPTables` because it had accidentally modified its internal state to include that string as part of the rule, but with real iptables it would have ended up trying to do `iptables -D OVN-KUBE-EGRESS-SVC -A OVN-KUBE-EGRESS-SVC -s 10.128.0.3 ...`.)

But this code only runs when ovnkube-node starts up on a node that already has EgressService rules, so it doesn't get tested in the e2e tests, so no one noticed.

I'm going to replace this with nftables code soon but I figured I'd submit this separately since it's a bugfix.

### DIFF
--- a/go-controller/pkg/node/controllers/egressservice/egressservice_node.go
+++ b/go-controller/pkg/node/controllers/egressservice/egressservice_node.go
@@ -613,6 +613,11 @@ func (c *Controller) repairIPTables(v4EpsToServices, v6EpsToServices map[string]
 		errorList := []error{}
 		for _, rule := range rulesToDel {
 			args := strings.Fields(rule)
+			if len(args) < 2 {
+				continue
+			}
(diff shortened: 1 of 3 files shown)


trozet (approved):
Thanks @danwinship! I had seen this before but never tracked it down.

trozet:
@danwinship can you sign the commit to make dco happy please?

PR #4872, opened 2024-12-06
Fix the port value used for marking externally exposed services

kyrtapz:
Fix the port value used for marking externally exposed services.
Additionally modified the E2Es to use different svcPort and targetPort to avoid this issues being missed in the future.


test/e2e/network_segmentation_services.go
```diff
@@ -31,7 +31,7 @@ var _ = Describe("Network Segmentation: services", func() {
 	Context("on a user defined primary network", func() {
 		const (
 			nadName                      = "tenant-red"
-			servicePort                  = 80
+			servicePort                  = 88
```
tssurya, line 34: hehehe

go-controller/pkg/node/gateway_nftables.go
```diff
@@ -98,7 +98,7 @@ func getUDNExternalIPsMarkNFTRules(svcPort kapi.ServicePort, externalIPs []strin
 		nftRules = append(nftRules,
 			&knftables.Element{
 				Map:   mapName,
-				Key:   []string{externalIP, strings.ToLower(string(svcPort.Protocol)), fmt.Sprintf("%v", svcPort.TargetPort.IntValue())},
+				Key:   []string{externalIP, strings.ToLower(string(svcPort.Protocol)), fmt.Sprintf("%v", svcPort.Port)},
```
tssurya, line 101: ouchy...

tssurya (approved):

go-controller/pkg/node/gateway_nftables.go
kyrtapz: yeah, embarrassing :facepalm:

kyrtapz:
Fixed E2E bugs.

tssurya:
why is license compliance failing on PRs.. hmm

PR #4873, opened 2024-12-06
e2e: Improve IGMP validation test with retries

flavio-fernandes:
Fix a flake where IGMP queries do not occur while iperf runs bound to a multicast address. Instead, use smaller intervals, allowing the test to complete in less than the fixed 30 seconds.

### DIFF
--- a/test/e2e/multicast.go
+++ b/test/e2e/multicast.go
@@ -188,7 +188,7 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
 	createGenericPod(f, multicastListenerPod, serverNodeInfo.name, f.Namespace.Name, tcpDumpCommand)
 
 	// Wait for tcpdump on listener pod to be ready
-	err := wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
+	err := wait.PollUntilContextTimeout(context.Background(), retryInterval, retryTimeout, true /*immediate*/, func(context.Context) (bool, error) {
 		kubectlOut, err := e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", "ls")
 		if err != nil {
 			framework.Failf("failed to retrieve multicast IGMP query: " + err.Error())
@@ -202,20 +202,12 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
 		framework.Failf("failed to retrieve multicast IGMP query: " + err.Error())
 	}
 
-	// The multicast listener pod join multicast group (-B 224.1.1.1), UDP (-u), during (-t 30) seconds, report every (-i 5) seconds
-	ginkgo.By("multicast listener pod join multicast group")
-	e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", fmt.Sprintf("iperf -s -B %s -u -t 30 -i 5", mcastGroup))
-
-	ginkgo.By(fmt.Sprintf("verifying that the IGMP query has been received"))
-	kubectlOut, err := e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", fmt.Sprintf("cat %s | grep igmp", tcpdumpFileName))
-	if err != nil {
-		framework.Failf("failed to retrieve multicast IGMP query: " + err.Error())
-	}
-	framework.Logf("output:")
-	framework.Logf(kubectlOut)
-	if kubectlOut == "" {
-		framework.Failf("failed to retrieve multicast IGMP query: igmp messages on the tcpdump logfile not found")
-	}
+	ginkgo.By("verifying that the IGMP query has been received")
+	// ovn-k leaves both mcast_idle_timeout and mcast_query_interval to default which means that
+	// ovn-controller will send IGMP queries every 150 seconds.
+	gomega.Eventually(func() (string, error) {
+		return e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", fmt.Sprintf("grep igmp %s", tcpdumpFileName))
+	}, 2*150*time.Second, 5*time.Second).Should(gomega.ContainSubstring("igmp"), "Failed to retrieve multicast IGMP query within the expected time")
 }
 
 func enableMulticastForNamespace(fr *framework.Framework) {


flavio-fernandes:
@qinqon @dceara PTAL

test/e2e/multicast.go
```diff
@@ -202,19 +202,36 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
...
+		kubectlOut, err := e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", fmt.Sprintf("grep igmp %s", tcpdumpFileName))
+		if err != nil || kubectlOut == "" {
+			// No igmp queries found. Retry up to maxDuration
+			continue
+		}
+
+		framework.Logf("output:")
+		framework.Logf(kubectlOut)
+		break
 	}
```
qinqon, line 235: Why this is not an eventually ?

test/e2e/multicast.go
```diff
@@ -202,19 +202,36 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
 		framework.Failf("failed to retrieve multicast IGMP query: " + err.Error())
 	}
 
-	// The multicast listener pod join multicast group (-B 224.1.1.1), UDP (-u), during (-t 30) seconds, report every (-i 5) seconds
-	ginkgo.By("multicast listener pod join multicast group")
-	e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", fmt.Sprintf("iperf -s -B %s -u -t 30 -i 5", mcastGroup))
+	ginkgo.By("multicast listener pod joins multicast group and verifies IGMP query")
+	maxDuration := 2 * time.Minute
+	start := time.Now()
```
qinqon, line 207: If you use Eventually you don't need these.

test/e2e/multicast.go
```diff
@@ -188,7 +188,7 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
 	createGenericPod(f, multicastListenerPod, serverNodeInfo.name, f.Namespace.Name, tcpDumpCommand)
 
 	// Wait for tcpdump on listener pod to be ready
-	err := wait.PollImmediate(retryInterval, retryTimeout, func() (bool, error) {
+	err := wait.PollUntilContextTimeout(context.Background(), retryInterval, retryTimeout, true, func(context.Context) (bool, error) {
```
qinqon, line 191: Could you add a comment to the boolean ? `true /*immediate*/`

Although maybe is better to use ginkg.Eventually here since it integrate with test framework.

qinqon (changes requested):

test/e2e/multicast.go
```diff
@@ -202,19 +202,36 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
...
-	if kubectlOut == "" {
-		framework.Failf("failed to retrieve multicast IGMP query: igmp messages on the tcpdump logfile not found")
+	for {
+		elapsed := time.Since(start)
+		if elapsed > maxDuration {
+			framework.Failf("Failed to retrieve multicast IGMP query within %v", maxDuration)
+			break
+		}
+
+		ginkgo.By("Running iperf to join the multicast group")
```
dceara, line 216: @flavio-fernandes This part of the test wasn't actually needed at all.  The test is about checking whether ovn-kubernetes (through ovn-controller) generates IGMP Queries periodically.  That should happen regardless of pods joining specific groups or not.

Which also makes me wonder, where was this flake seen, do you have a link to the logs?  I didn't see it in upstream CI but I also didn't follow all runs closely.

dceara (changes requested):

test/e2e/multicast.go
flavio-fernandes: hi @dceara ! Not sure I see the failure upstream anymore. Will check. I mostly see them downstream now.
Valid point about the iperf step. I think it was added to add a 30 seconds delay, which (in most cases?) is enough to get the periodic igmp queries. On the same token, are you sure that having an app binding to the multicast address not a trigger for igmp messages to happen sooner? I suspected it, but did not dig deep enough to confirm it.

test/e2e/multicast.go
flavio-fernandes: Valid point, @qinqon. Let me do it using `eventually`.

test/e2e/multicast.go
flavio-fernandes: @qinqon : assuming you were the original implementer of this test, would you know? ^^

test/e2e/multicast.go
flavio-fernandes: My intent here is just to stop using the deprecated api. Hope you are okay with me adding `true /*immediate*/` to minimize my change on this pr.

test/e2e/multicast.go
flavio-fernandes: Rewrote this logic to just wait for the igmp packet

dceara: @flavio-fernandes ovn-controller IGMP queries are periodic and are flooded to all logical ports of the switch, regardless of who joined multicast groups or not.

test/e2e/multicast.go
```diff
@@ -202,20 +202,19 @@ func testMulticastIGMPQuery(f *framework.Framework, clientNodeInfo, serverNodeIn
...
-	framework.Logf(kubectlOut)
-	if kubectlOut == "" {
-		framework.Failf("failed to retrieve multicast IGMP query: igmp messages on the tcpdump logfile not found")
+		framework.Failf("failed to join multicast group: %v", err)
 	}
+
+	ginkgo.By("verifying that the IGMP query has been received")
+	gomega.Eventually(func() (string, error) {
+		return e2ekubectl.RunKubectl(f.Namespace.Name, "exec", multicastListenerPod, "--", "/bin/bash", "-c", fmt.Sprintf("grep igmp %s", tcpdumpFileName))
+	}, 2*time.Minute, 5*time.Second).Should(gomega.ContainSubstring("igmp"), "Failed to retrieve multicast IGMP query within the expected time")
```
dceara, line 217: @flavio-fernandes I checked the IGMP configuration now and ovn-k leaves both `mcast_idle_timeout` and `mcast_query_interval` to default which means that ovn-controller will send IGMP queries every 150 seconds.  I think you an just drop the "iperf to join the multicast group" part, and check up to 300s to see if an IGMP query was captured by tcpdump.

flavio-fernandes: ack. Let's leave timeout at 300, just to account for a very slow environment. TYSM @dceara !

dceara:
@flavio-fernandes looks like this hit the already known https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4387 flake in CI.  It's probably a good idea to retest.

dceara (approved):
Looks good to me, thanks!

flavio-fernandes:
`UT Flake: handles a HO node is switched to a OVN node is flaking`
Flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4387

flavio-fernandes:
Flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4474

girishmg (approved):

Issue #4876, opened 2024-12-08
[Non IC] Nodes on Non IC mode miss entry for primary UDN networks


oshoval:
### What happened?

Create a primary UDN NAD, spin a VM using that
Node has this annotation 
`k8s.ovn.org/node-gateway-router-lrp-ifaddrs: '{"default":{"ipv4":"100.64.0.4/16","ipv6":"fd98::4/64"}}'`
It doesn't have expected `l2-network` entry

### What did you expect to happen?

Expected to have it, in order to fix multi-path issue [2].
Since all the LRPs attached to the logical switch will answer the router solicitation with a corresponding router advertisement, the VM will have multi default gateways.
Having the gateway LLA allows to create a nft rule that will drop all RA that were not originated from local GW.

On IC mode we have this as expected
```
    k8s.ovn.org/node-gateway-router-lrp-ifaddrs: '{"default":{"ipv4":"100.64.0.3/16","ipv6":"fd98::3/64"},"l2-network":{"ipv4":"100.65.0.3/16","ipv6":"fd99::3/64"}}'
```

[2] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4866
and https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852 (both WIP)

### How can we reproduce it (as minimally and precisely as possible)?

Create a NAD (might be enough, else also spin a VM that use this NAD)

[1]
```
apiVersion: k8s.cni.cncf.io/v1                                                  
kind: NetworkAttachmentDefinition                                               
metadata:                                                                       
  namespace: poc1
  name: l2
spec:                                                                           
  config: |2                                                                    
    {                                                                           
            "cniVersion": "0.3.0",                                              
            "name": "l2-network",                                               
            "type": "ovn-k8s-cni-overlay",                                      
            "topology":"layer2",                                                
            "subnets": "10.100.200.0/24,2014:100:200::0/60",                                       
            "netAttachDefName": "poc1/l2",                               
            "allowPersistentIPs": true,
            "role": "primary"
    }
```

### OVN-Kubernetes version

<details>

```console
$ ovnkube --version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### OVN version

<details>

```console
$ oc rsh -n ovn-kubernetes ovnkube-node-xxxxx (pick any ovnkube-node pod on your cluster)
$ rpm -q ovn
# paste output here
```

[...]

Issue #4721, opened 2024-09-11
UT: should re-balance EgressIPs when their node is removed


tssurya:
### Which jobs are flaking?

unit test

### Which tests are flaking?

```
2024-09-11T10:33:33.3229158Z 
2024-09-11T10:33:33.3230602Z [91m[1m[Fail] [0m[90mOVN master EgressIP Operations [0m[0mWatchEgressNodes running with WatchEgressIP [0m[91m[1m[It] 2024-09-11T10:33:33.3229158Z 
2024-09-11T10:33:33.3230602Z [91m[1m[Fail] [0m[90mOVN master EgressIP Operations [0m[0mWatchEgressNodes running with WatchEgressIP [0m[91m[1m[It] should re-balance EgressIPs when their node is removed [0m
2024-09-11T10:33:33.3232456Z [37m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:11879[0m
2024-09-11T10:33:33.3232996Z 
 [0m
2024-09-11T10:33:33.3232456Z [37m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:11879[0m
2024-09-11T10:33:33.3232996Z 
```

### Since when has it been flaking?

this week

tssurya:
hit this here as well: https://github.com/ovn-org/ovn-kubernetes/pull/4554#issuecomment-2341345581

tssurya:
and https://github.com/ovn-org/ovn-kubernetes/pull/4718#issuecomment-2344026976

ricky-rav:
I've just seen it as well: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13062093159/job/36447238443?pr=5018

trozet:
I see in the failure case is that in the deleteEgressIPStatusSetup function, the code below is not executed:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/egressip.go#L2900

[...]

PR #4878, opened 2024-12-09
l2, udpn: Add ipv6 lla gateway field to the ovn pod annotation

oshoval:
## 📑 Description

For primary UDN layer2 the proper IPv6 gateway that should be configured
with received router advertisements is the one from the local GR IPv6
LLA, this change add field to the ovn pod annotations with that
information so it can be consumed by follow up PRs that will ensure
remove GRs are ignore as layer2 ipv6 gateways.

This is related to the following PRs that remove multipath ipv6 gateway for l2 primary UDNs:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4847

Based on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4866, just finalizing it.
Thanks @qinqon

Fixes #

## ✅ Checks

- [x] My code requires tests

### DIFF
--- a/go-controller/pkg/allocator/pod/pod_annotation.go
+++ b/go-controller/pkg/allocator/pod/pod_annotation.go
@@ -54,6 +54,7 @@ func NewPodAnnotationAllocator(
 // false.
 func (allocator *PodAnnotationAllocator) AllocatePodAnnotation(
 	ipAllocator subnet.NamedAllocator,
+	node *corev1.Node,
 	pod *corev1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	reallocateIP bool,
@@ -67,6 +68,7 @@ func (allocator *PodAnnotationAllocator) AllocatePodAnnotation(
 		allocator.kube,
 		ipAllocator,
 		allocator.netInfo,
+		node,
 		pod,
 		network,
 		allocator.ipamClaimsReconciler,
@@ -80,6 +82,7 @@ func allocatePodAnnotation(
 	kube kube.Interface,
 	ipAllocator subnet.NamedAllocator,
 	netInfo util.NetInfo,
+	node *corev1.Node,
 	pod *corev1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	claimsReconciler persistentips.PersistentAllocations,
@@ -98,6 +101,7 @@ func allocatePodAnnotation(
 			ipAllocator,
 			idAllocator,
 			netInfo,
+			node,
 			pod,
 			network,
 			claimsReconciler,
@@ -133,6 +137,7 @@ func allocatePodAnnotation(
 func (allocator *PodAnnotationAllocator) AllocatePodAnnotationWithTunnelID(
 	ipAllocator subnet.NamedAllocator,
 	idAllocator id.NamedAllocator,
+	node *corev1.Node,
 	pod *corev1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	reallocateIP bool,
@@ -147,6 +152,7 @@ func (allocator *PodAnnotationAllocator) AllocatePodAnnotationWithTunnelID(
 		ipAllocator,
 		idAllocator,
 		allocator.netInfo,
+		node,
 		pod,
 		network,
 		allocator.ipamClaimsReconciler,
@@ -161,6 +167,7 @@ func allocatePodAnnotationWithTunnelID(
 	ipAllocator subnet.NamedAllocator,
 	idAllocator id.NamedAllocator,
 	netInfo util.NetInfo,
+	node *corev1.Node,
 	pod *corev1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	claimsReconciler persistentips.PersistentAllocations,
@@ -176,6 +183,7 @@ func allocatePodAnnotationWithTunnelID(
 			ipAllocator,
 			idAllocator,
 			netInfo,
+			node,
 			pod,
 			network,
 			claimsReconciler,
@@ -220,6 +228,7 @@ func allocatePodAnnotationWithRollback(
 	ipAllocator subnet.NamedAllocator,
 	idAllocator id.NamedAllocator,
 	netInfo util.NetInfo,
+	node *corev1.Node,
 	pod *corev1.Pod,
 	network *nadapi.NetworkSelectionElement,
 	claimsReconciler persistentips.PersistentAllocations,
@@ -397,7 +406,7 @@ func allocatePodAnnotationWithRollback(
 		}
 
 		// handle routes & gateways
-		err = util.AddRoutesGatewayIP(netInfo, pod, tentative, network)
+		err = util.AddRoutesGatewayIP(netInfo, node, pod, tentative, network)
 		if err != nil {
 			return
 		}
--- a/go-controller/pkg/allocator/pod/pod_annotation_test.go
+++ b/go-controller/pkg/allocator/pod/pod_annotation_test.go
@@ -124,6 +124,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 		wantReleaseID             bool
 		wantRelasedIDOnRollback   bool
 		wantErr                   bool
+		multiNetworkDisabled      bool
 	}{
 		{
 			// on secondary L2 networks with no IPAM, we expect to generate a
@@ -146,7 +147,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			},
 		},
 		{
-			// on secondary L2 network with no IPAM, honor static IP requests
+			// with multiNetwork disabled, on secondary L2 network with no IPAM, honor static IP requests
 			// present in the network selection annotation
 			name: "expect requested static IP, no gateway, no IPAM",
 			args: args{
@@ -163,7 +164,8 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				MAC:  util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.4/24")[0].IP),
 				Role: types.NetworkRolePrimary,
 			},
-			role: types.NetworkRolePrimary,
+			role:                 types.NetworkRolePrimary,
+			multiNetworkDisabled: true,
 		},
 		{
 			// on secondary L2 network with no IPAM, honor static IP and gateway
@@ -662,6 +664,12 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			network.Name = "network"
 			network.Namespace = "namespace"
 
+			config.OVNKubernetesFeature.EnableInterconnect = tt.idAllocation
+			config.OVNKubernetesFeature.EnableMultiNetwork = !tt.multiNetworkDisabled
+			config.OVNKubernetesFeature.EnableNetworkSegmentation = true
+			config.IPv4Mode = true
+			config.IPv6Mode = true
+
 			var netInfo util.NetInfo
 			netInfo = &util.DefaultNetInfo{}
 			nadName := types.DefaultNetworkName
@@ -686,7 +694,6 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				}
 			}
 
-			config.OVNKubernetesFeature.EnableInterconnect = tt.idAllocation
 
 			pod := &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
--- a/go-controller/pkg/allocator/pod/pod_annotation_test.go
+++ b/go-controller/pkg/allocator/pod/pod_annotation_test.go
@@ -6,6 +6,8 @@ import (
 	"net"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
 	nadapi "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
@@ -124,6 +126,8 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 		wantReleaseID             bool
 		wantRelasedIDOnRollback   bool
 		wantErr                   bool
+		isSingleStackIPv4         bool
+		isSingleStackIPv6         bool
 		multiNetworkDisabled      bool
 	}{
 		{
@@ -232,6 +236,113 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				},
 				Role: types.NetworkRolePrimary,
 			},
+			wantReleasedIPsOnRollback: ovntest.MustParseIPNets("192.168.0.3/24"),
+			role:                      types.NetworkRolePrimary,
+		},
+		{
+			name:                   "expect new IP and ipv6 gateway LLA for primary udn layer2 with dual stack",
+			ipam:                   true,
+			idAllocation:           true,
+			persistentIPAllocation: true,
+			args: args{
+				ipAllocator: &ipAllocatorStub{
+					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24", "2010:100:200::3/60"),
+				},
+				idAllocator: &idAllocatorStub{
+					nextID: 100,
+				},
+			},
+			wantUpdatedPod: true,
+			wantPodAnnotation: &util.PodAnnotation{
+				IPs:            ovntest.MustParseIPNets("192.168.0.3/24", "2010:100:200::3/60"),
+				MAC:            util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.3/24")[0].IP),
+				Gateways:       []net.IP{ovntest.MustParseIP("192.168.0.1").To4(), ovntest.MustParseIP("2010:100:200::1")},
+				GatewayIPv6LLA: util.HWAddrToIPv6LLA(util.IPAddrToHWAddr(ovntest.MustParseIP("100.65.0.4"))),
+				Routes: []util.PodRoute{
+					{
+						Dest: &net.IPNet{
+							IP:   ovntest.MustParseIP("100.65.0.0").To4(),
+							Mask: net.CIDRMask(16, 32),
+						},
+						NextHop: ovntest.MustParseIP("192.168.0.1").To4(),
+					},
+					{
+						Dest:    ovntest.MustParseIPNet("fd99::/64"),
+						NextHop: ovntest.MustParseIP("2010:100:200::1"),
+					},
+				},
+				Role:     types.NetworkRolePrimary,
+				TunnelID: 100,
+			},
+			wantRelasedIDOnRollback:   true,
+			wantReleasedIPsOnRollback: ovntest.MustParseIPNets("192.168.0.3/24", "2010:100:200::3/60"),
+			role:                      types.NetworkRolePrimary,
+		},
+		{
+			name:                   "expect new IP and ipv6 gateway LLA for primary udn layer2 with single stack IPv6",
+			isSingleStackIPv6:      true,
+			ipam:                   true,
+			idAllocation:           true,
+			persistentIPAllocation: true,
+			args: args{
+				ipAllocator: &ipAllocatorStub{
+					netxtIPs: ovntest.MustParseIPNets("2010:100:200::3/60"),
+				},
+				idAllocator: &idAllocatorStub{
+					nextID: 100,
+				},
+			},
+			wantUpdatedPod: true,
+			wantPodAnnotation: &util.PodAnnotation{
+				IPs:            ovntest.MustParseIPNets("2010:100:200::3/60"),
+				MAC:            util.IPAddrToHWAddr(ovntest.MustParseIPNets("2010:100:200::3/60")[0].IP),
+				Gateways:       []net.IP{ovntest.MustParseIP("2010:100:200::1")},
+				GatewayIPv6LLA: util.HWAddrToIPv6LLA(util.IPAddrToHWAddr(ovntest.MustParseIP("fd99::4"))),
+				Routes: []util.PodRoute{
+					{
+						Dest:    ovntest.MustParseIPNet("fd99::/64"),
+						NextHop: ovntest.MustParseIP("2010:100:200::1"),
+					},
+				},
+				Role:     types.NetworkRolePrimary,
+				TunnelID: 100,
+			},
+			wantRelasedIDOnRollback:   true,
+			wantReleasedIPsOnRollback: ovntest.MustParseIPNets("2010:100:200::3/60"),
+			role:                      types.NetworkRolePrimary,
+		},
+		{
+			name:                   "expect new IP but no ipv6 gateway LLA for primary udn layer2 with single stack IPv4",
+			isSingleStackIPv4:      true,
+			ipam:                   true,
+			idAllocation:           true,
+			persistentIPAllocation: true,
+			args: args{
+				ipAllocator: &ipAllocatorStub{
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+				},
+				idAllocator: &idAllocatorStub{
+					nextID: 100,
+				},
+			},
+			wantUpdatedPod: true,
+			wantPodAnnotation: &util.PodAnnotation{
+				IPs:      ovntest.MustParseIPNets("192.168.0.3/24"),
+				MAC:      util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.3/24")[0].IP),
+				Gateways: []net.IP{ovntest.MustParseIP("192.168.0.1").To4()},
+				Routes: []util.PodRoute{
+					{
+						Dest: &net.IPNet{
+							IP:   ovntest.MustParseIP("100.65.0.0").To4(),
+							Mask: net.CIDRMask(16, 32),
+						},
+						NextHop: ovntest.MustParseIP("192.168.0.1").To4(),
+					},
+				},
+				Role:     types.NetworkRolePrimary,
+				TunnelID: 100,
+			},
+			wantRelasedIDOnRollback:   true,
 			wantReleasedIPsOnRollback: ovntest.MustParseIPNets("192.168.0.3/24"),
 			role:                      types.NetworkRolePrimary,
 		},
@@ -668,8 +779,13 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			config.OVNKubernetesFeature.EnableMultiNetwork = !tt.multiNetworkDisabled
 			config.OVNKubernetesFeature.EnableNetworkSegmentation = true
 			config.IPv4Mode = true
+			if tt.isSingleStackIPv6 {
+				config.IPv4Mode = false
+			}
 			config.IPv6Mode = true
-
+			if tt.isSingleStackIPv4 {
+				config.IPv6Mode = false
+			}
 			var netInfo util.NetInfo
 			netInfo = &util.DefaultNetInfo{}
 			nadName := types.DefaultNetworkName
@@ -677,7 +793,12 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				nadName = util.GetNADName(network.Namespace, network.Name)
 				var subnets string
 				if tt.ipam {
-					subnets = "192.168.0.0/24"
+					subnets = "192.168.0.0/24,2001:db8::/64"
+					if tt.isSingleStackIPv4 {
+						subnets = "192.168.0.0/24"
+					} else if tt.isSingleStackIPv6 {
+						subnets = "2001:db8::/64"
+					}
 				}
 				netInfo, err = util.NewNetInfo(&ovncnitypes.NetConf{
 					Topology: types.Layer2Topology,
@@ -694,6 +815,20 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				}
 			}
 
+			ifaddrs := `{"network":{"ipv4":"100.65.0.4/16","ipv6":"fd99::4/64"}}`
+			if tt.isSingleStackIPv4 {
+				ifaddrs = `{"network":{"ipv4":"100.65.0.4/16"}}`
+			} else if tt.isSingleStackIPv6 {
+				ifaddrs = `{"network":{"ipv6":"fd99::4/64"}}`
+			}
+
+			node := &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"k8s.ovn.org/node-gateway-router-lrp-ifaddrs": ifaddrs,
+					},
+				},
+			}
 
 			pod := &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
@@ -729,6 +864,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				tt.args.ipAllocator,
 				tt.args.idAllocator,
 				netInfo,
+				node,
 				pod,
 				network,
 				claimsReconciler,
@@ -774,7 +910,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				g.Expect(podAnnotation.MAC[0]&2).To(gomega.BeEquivalentTo(2), "Expected local MAC")
 				return
 			}
-			g.Expect(podAnnotation).To(gomega.Equal(tt.wantPodAnnotation))
+			g.Expect(podAnnotation).To(gomega.Equal(tt.wantPodAnnotation), "diff: %s", cmp.Diff(tt.wantPodAnnotation, podAnnotation))
 
 			if tt.wantUpdatedPod {
 				g.Expect(pod).NotTo(gomega.BeNil(), "Expected an updated pod")
--- a/go-controller/pkg/allocator/pod/pod_annotation_test.go
+++ b/go-controller/pkg/allocator/pod/pod_annotation_test.go
@@ -6,9 +6,8 @@ import (
 	"net"
 	"testing"
 
-	"github.com/google/go-cmp/cmp"
-
 	cnitypes "github.com/containernetworking/cni/pkg/types"
+	"github.com/google/go-cmp/cmp"
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
 	nadapi "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/onsi/gomega"
@@ -28,7 +27,7 @@ import (
 )
 
 type ipAllocatorStub struct {
-	netxtIPs         []*net.IPNet
+	nextIPs          []*net.IPNet
 	allocateIPsError error
 	releasedIPs      []*net.IPNet
 }
@@ -38,7 +37,7 @@ func (a *ipAllocatorStub) AllocateIPs([]*net.IPNet) error {
 }
 
 func (a *ipAllocatorStub) AllocateNextIPs() ([]*net.IPNet, error) {
-	return a.netxtIPs, nil
+	return a.nextIPs, nil
 }
 
 func (a *ipAllocatorStub) ReleaseIPs(ips []*net.IPNet) error {
@@ -159,7 +158,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					IPRequest: []string{"192.168.0.4/24"},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
@@ -181,7 +180,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					GatewayRequest: ovntest.MustParseIPs("192.168.0.1"),
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
@@ -213,7 +212,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			ipam: true,
 			args: args{
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
@@ -246,7 +245,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			persistentIPAllocation: true,
 			args: args{
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24", "2010:100:200::3/60"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24", "2010:100:200::3/60"),
 				},
 				idAllocator: &idAllocatorStub{
 					nextID: 100,
@@ -286,7 +285,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			persistentIPAllocation: true,
 			args: args{
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("2010:100:200::3/60"),
+					nextIPs: ovntest.MustParseIPNets("2010:100:200::3/60"),
 				},
 				idAllocator: &idAllocatorStub{
 					nextID: 100,
@@ -410,7 +409,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					IPRequest: []string{"192.168.0.4/24"},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
@@ -447,7 +446,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					IPRequest: []string{"192.168.0.4/24"},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs:         ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs:          ovntest.MustParseIPNets("192.168.0.3/24"),
 					allocateIPsError: ipam.ErrAllocated,
 				},
 			},
@@ -484,7 +483,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					IPRequest: []string{"192.168.0.4/24"},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs:         ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs:          ovntest.MustParseIPNets("192.168.0.3/24"),
 					allocateIPsError: errors.New("Allocate IPs failed"),
 				},
 			},
@@ -520,7 +519,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					IPRequest: []string{"ivalid"},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantErr: true,
@@ -534,7 +533,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					MacRequest: "ivalid",
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantErr:         true,
@@ -550,7 +549,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					MacRequest: requestedMAC,
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
@@ -586,7 +585,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					MacRequest: "invalid",
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			invalidNetworkAnnotation: true,
@@ -612,7 +611,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
@@ -640,7 +639,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod:            true,
@@ -667,7 +666,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 					Status: ipamclaimsapi.IPAMClaimStatus{},
 				},
 				ipAllocator: &ipAllocatorStub{
-					netxtIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
+					nextIPs: ovntest.MustParseIPNets("192.168.0.3/24"),
 				},
 			},
 			wantUpdatedPod: true,
--- a/go-controller/pkg/ovn/base_network_controller_pods.go
+++ b/go-controller/pkg/ovn/base_network_controller_pods.go
@@ -909,7 +909,13 @@ func (bnc *BaseNetworkController) allocatePodAnnotation(pod *corev1.Pod, existin
 		return nil, false, fmt.Errorf("cannot retrieve subnet for assigning gateway routes for pod %s, switch: %s",
 			podDesc, switchName)
 	}
-	err = util.AddRoutesGatewayIP(bnc.GetNetInfo(), pod, podAnnotation, network)
+
+	node, err := bnc.watchFactory.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return nil, false, err
+	}
+
+	err = util.AddRoutesGatewayIP(bnc.GetNetInfo(), node, pod, podAnnotation, network)
 	if err != nil {
 		return nil, false, err
 	}
@@ -973,9 +979,14 @@ func (bnc *BaseNetworkController) allocatePodAnnotationForSecondaryNetwork(pod *
 	if bnc.doesNetworkRequireIPAM() {
 		ipAllocator = bnc.lsManager.ForSwitch(switchName)
 	}
-
+	node, err := bnc.watchFactory.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to get pod %s/%s/%s node %q: %w",
+			nadName, pod.Namespace, pod.Name, pod.Spec.NodeName, err)
+	}
 	updatedPod, podAnnotation, err := bnc.podAnnotationAllocator.AllocatePodAnnotation(
 		ipAllocator,
+		node,
 		pod,
 		network,
 		reallocate,
--- a/go-controller/pkg/util/pod_annotation.go
+++ b/go-controller/pkg/util/pod_annotation.go
@@ -75,6 +75,12 @@ type PodAnnotation struct {
 	// Gateways are the pod's gateway IP addresses; note that there may be
 	// fewer Gateways than IPs.
 	Gateways []net.IP
+
+	// GatewayIPv6LLA is the IPv6 Link Local Address for the pod's gateway, that is the address
+	// that will be set as gateway with router advertisements
+	// generated from the gateway router from the node where the pod is running.
+	GatewayIPv6LLA net.IP
+
 	// Routes are additional routes to add to the pod's network namespace
 	Routes []PodRoute
 
@@ -117,8 +123,9 @@ type podAnnotation struct {
 	Gateways []string   `json:"gateway_ips,omitempty"`
 	Routes   []podRoute `json:"routes,omitempty"`
 
-	IP      string `json:"ip_address,omitempty"`
-	Gateway string `json:"gateway_ip,omitempty"`
+	IP             string `json:"ip_address,omitempty"`
+	Gateway        string `json:"gateway_ip,omitempty"`
+	GatewayIPv6LLA string `json:"ipv6_lla_gateway_ip,omitempty"`
 
 	TunnelID int    `json:"tunnel_id,omitempty"`
 	Role     string `json:"role,omitempty"`
@@ -192,6 +199,11 @@ func MarshalPodAnnotation(annotations map[string]string, podInfo *PodAnnotation,
 			NextHop: nh,
 		})
 	}
+
+	if podInfo.GatewayIPv6LLA != nil {
+		pa.GatewayIPv6LLA = podInfo.GatewayIPv6LLA.String()
+	}
+
 	podNetworks[nadName] = pa
 	bytes, err := json.Marshal(podNetworks)
 	if err != nil {
@@ -280,6 +292,14 @@ func UnmarshalPodAnnotation(annotations map[string]string, nadName string) (*Pod
 			}
 		}
 		podAnnotation.Routes = append(podAnnotation.Routes, route)
+	}
+
+	if a.GatewayIPv6LLA != "" {
+		llaGW := net.ParseIP(a.GatewayIPv6LLA)
+		if !isIPv6LLA(llaGW) {
+			return nil, fmt.Errorf("failed to parse pod ipv6 lla gateway, or non ipv6 lla %q", a.GatewayIPv6LLA)
+		}
+		podAnnotation.GatewayIPv6LLA = llaGW
 	}
 
 	return podAnnotation, nil
@@ -693,4 +713,9 @@ func UnmarshalUDNOpenPortsAnnotation(annotations map[string]string) ([]*OpenPort
 		}
 	}
 	return result, nil
+}
+
+// Ensure the IP is a valid IPv6 LLA
+func isIPv6LLA(ip net.IP) bool {
+	return utilnet.IsIPv6(ip) && ip.IsLinkLocalUnicast()
 }
--- a/go-controller/pkg/util/pod_annotation.go
+++ b/go-controller/pkg/util/pod_annotation.go
@@ -556,6 +556,7 @@ func hairpinMasqueradeIPToRoute(isIPv6 bool, gatewayIP net.IP) PodRoute {
 // with the gateways derived from the allocated IPs
 func AddRoutesGatewayIP(
 	netinfo NetInfo,
+	node *corev1.Node,
 	pod *corev1.Pod,
 	podAnnotation *PodAnnotation,
 	network *nadapi.NetworkSelectionElement) error {
@@ -589,6 +590,19 @@ func AddRoutesGatewayIP(
 				podAnnotation.Routes = append(podAnnotation.Routes, joinSubnetToRoute(netinfo, isIPv6, gatewayIPnet.IP))
 				if network != nil && len(network.GatewayRequest) == 0 { // if specific default route for pod was not requested then add gatewayIP
 					podAnnotation.Gateways = append(podAnnotation.Gateways, gatewayIPnet.IP)
+				}
+			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						if IsAnnotationNotSetError(err) {
+							return types.NewSuppressedError(err)
+						}
+						return fmt.Errorf("failed parsing node gateway router join addresses, network %q, %w", netinfo.GetNetworkName(), err)
+					}
+					podAnnotation.GatewayIPv6LLA = HWAddrToIPv6LLA(IPAddrToHWAddr(joinAddrs[0].IP))
 				}
 			}
 			return nil
--- a/go-controller/pkg/util/pod_annotation_unit_test.go
+++ b/go-controller/pkg/util/pod_annotation_unit_test.go
@@ -112,6 +112,13 @@ func TestMarshalPodAnnotation(t *testing.T) {
 			},
 			expectedOutput: map[string]string{"k8s.ovn.org/pod-networks": `{"default":{"ip_addresses":null,"mac_address":"","routes":[{"dest":"192.168.1.0/24","nextHop":""}]}}`},
 		},
+		{
+			desc: "ipv6 LLA gateway ip is set",
+			inpPodAnnot: PodAnnotation{
+				GatewayIPv6LLA: ovntest.MustParseIP("fe80::"),
+			},
+			expectedOutput: map[string]string{"k8s.ovn.org/pod-networks": `{"default":{"ip_addresses":null,"mac_address":"","ipv6_lla_gateway_ip":"fe80::"}}`},
+		},
 	}
 
 	for i, tc := range tests {
@@ -137,6 +144,7 @@ func TestUnmarshalPodAnnotation(t *testing.T) {
 		inpAnnotMap map[string]string
 		errAssert   bool
 		errMatch    error
+		nadName     string
 	}{
 		{
 			desc:        "verify `OVN pod annotation not found` error thrown",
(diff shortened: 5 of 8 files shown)


oshoval:
previous error was fixed, now it is a known flake
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4721

Note: maybe would be best to allow running the lanes (or at least some of them), during development, even if UT is failing (using some flag or a chatop directly to build and trigger them)

oshoval:
as far as i can tell the failures aren't related to the PR

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12250911507/job/34175635923?pr=4878
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12250911507/job/34175638991?pr=4878
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12250911507/job/34175644667?pr=4878

but best to rerun / have another opinion please

oshoval:
@qinqon @maiqueb 
basically ready for review please, rerunning tests, just to make sure the above aren't related

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -170,12 +173,18 @@ func allocatePodAnnotationWithTunnelID(
 	podAnnotation *util.PodAnnotation,
 	err error) {
 
+	node, err := kube.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return nil, nil, err
+	}
```
qinqon, line 179: We should not add an api call here as stated on the comment from original PR.

qinqon (changes requested):

oshoval:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12250911507/job/34175644667?pr=4878
Might be real failure, because i saw it fail twice on this PR and the follow-up

go-controller/pkg/allocator/pod/pod_annotation.go
oshoval: ah missed it, so to do as you done with the factory ? (assuming it is reachable)

qinqon: Yep, like we pass the Node directly from controllers and we red the Node there.

qinqon:
> https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12250911507/job/34175644667?pr=4878 Might be real failure, because i saw it fail twice on this PR and the follow-up (#4852)
> 
> ```
> Summarizing 1 Failure:
>   [FAIL] Network Segmentation: services on a user defined primary network should be reachable through their cluster IP and node port [It] L3 primary UDN, cluster-networked pods, NodePort service
>   /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_services.go:422
> ```

There is a flaky issue for this one.

oshoval:
> > https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12250911507/job/34175644667?pr=4878 Might be real failure, because i saw it fail twice on this PR and the follow-up (#4852)
> > ```
> > Summarizing 1 Failure:
> >   [FAIL] Network Segmentation: services on a user defined primary network should be reachable through their cluster IP and node port [It] L3 primary UDN, cluster-networked pods, NodePort service
> >   /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_services.go:422
> > ```
> 
> There is a flaky issue for this one.

Thanks, i didn't find an issue searching "L3 primary UDN, cluster-networked pods, NodePort service", might worth to open one ?
It fails in pretty high rate, imho would be nice to consider Quarantine or escalation sometimes for those ?

oshoval:
Thanks as we spoke this is the flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4829

go-controller/pkg/allocator/pod/pod_annotation.go
oshoval: Done, ptal

oshoval:
rebased

qinqon (approved):
LGTM

oshoval:
Thanks
Removed non required leftover mock

go-controller/pkg/util/pod_annotation_unit_test.go
```diff
@@ -211,6 +218,10 @@ func TestUnmarshalPodAnnotation(t *testing.T) {
 			desc:        "verify successful unmarshal of pod annotation when *only* the MAC address is present",
 			inpAnnotMap: map[string]string{"k8s.ovn.org/pod-networks": `{"default":{"mac_address":"0a:58:fd:98:00:01"}}`},
 		},
+		{
+			desc:        "verify successful unmarshal of pod annotation with ipv6 lla gateway ip",
+			inpAnnotMap: map[string]string{"k8s.ovn.org/pod-networks": `{"default":{"mac_address":"0a:58:fd:98:00:01","ipv6_lla_gateway_ip":"fe80::"}}`},
```
maiqueb, line 223: could we maybe align the MAC address w/ the LLA address ? Keep in mind we only support this feature for MAC generated LLAs.

I think it'll make the test "look" better, and allow for future reference.

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -161,11 +163,9 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			},
 			wantUpdatedPod: true,
 			wantPodAnnotation: &util.PodAnnotation{
-				IPs:  ovntest.MustParseIPNets("192.168.0.4/24"),
-				MAC:  util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.4/24")[0].IP),
-				Role: types.NetworkRolePrimary,
+				IPs: ovntest.MustParseIPNets("192.168.0.4/24"),
+				MAC: util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.4/24")[0].IP),
 			},
-			role: types.NetworkRolePrimary,
```
maiqueb, line 168: why did you remove the primary role from this test ?

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -769,7 +822,7 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				g.Expect(podAnnotation.MAC[0]&2).To(gomega.BeEquivalentTo(2), "Expected local MAC")
 				return
 			}
-			g.Expect(podAnnotation).To(gomega.Equal(tt.wantPodAnnotation))
+			g.Expect(podAnnotation).To(gomega.Equal(tt.wantPodAnnotation), "diff: %s", cmp.Diff(tt.wantPodAnnotation, podAnnotation))
```
maiqueb, line 825: is this really needed ? doesn't gomega print out the diffs between the expected and real results ?

maiqueb (commented):
Thank you.

go-controller/pkg/allocator/pod/pod_annotation_test.go
qinqon: The test is "expect requested static IP, no gateway, no IPAM" we don't have primaries without gateway, the primary role here feels like copy paste and make testing this feature more difficult.

kyrtapz (approved):

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -235,6 +235,46 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+				idAllocator: &idAllocatorStub{
+					nextID: 100,
+				},
+			},
+			wantUpdatedPod: true,
+			wantPodAnnotation: &util.PodAnnotation{
+				IPs:            ovntest.MustParseIPNets("192.168.0.3/24", "2010:100:200::3/60"),
+				MAC:            util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.3/24")[0].IP),
+				Gateways:       []net.IP{ovntest.MustParseIP("192.168.0.1").To4(), ovntest.MustParseIP("2010:100:200::1")},
+				IPv6LLAGateway: util.HWAddrToIPv6LLA(util.IPAddrToHWAddr(ovntest.MustParseIP("100.65.0.4"))),
```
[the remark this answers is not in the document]
qinqon: This is how it is for the LRP mac address generation it just take [0] that is ipv4 for dual stack.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -281,6 +293,13 @@ func UnmarshalPodAnnotation(annotations map[string]string, nadName string) (*Pod
 		podAnnotation.Routes = append(podAnnotation.Routes, route)
 	}
 
+	if a.IPv6LLAGateway != "" {
+		podAnnotation.IPv6LLAGateway = net.ParseIP(a.IPv6LLAGateway)
```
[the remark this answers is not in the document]
qinqon: > We can check if the IP is V6 and LLA too.

You mean like extra check to be sure we don't do something wrong ?

go-controller/pkg/allocator/pod/pod_annotation_test.go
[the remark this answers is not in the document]
kyrtapz: Ah, thanks for the clarification. Would it be worth it to add a test for single stack ipv6?

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
kyrtapz: yes, I want to make sure we are honoring the variable name.

go-controller/pkg/allocator/pod/pod_annotation_test.go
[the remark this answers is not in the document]
qinqon: Maybe the proper thing is to read the mac address from the LRP directly instead of re-calculate, but it means we have to traverse more things between all the functions and affect the unit tests and the like, but looks like the proper thing to do.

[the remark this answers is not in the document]
oshoval: imho if possible please lets stick to the current (wrt to the 2 points)
we are going to drop it once the proper fix is out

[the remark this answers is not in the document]
qinqon: fair enough, @oshoval Let's add a test for single stack ipv6.

maiqueb:
@oshoval needs to be rebased.

go-controller/pkg/allocator/pod/pod_annotation_test.go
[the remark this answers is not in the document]
maiqueb: @oshoval seems you also need to add a unit test here.

oshoval:
rebased only

go-controller/pkg/allocator/pod/pod_annotation_test.go
oshoval: you are correct, however it adds the first compact diff form below, if you want i will remove it

```
W1222 14:44:10.132065   24043 pod_annotation.go:368] failed to ensure requested or annotated IPs [192.168.0.4/24] for default/namespace/pod: Allocate IPs failed
E1222 14:44:10.132381   24043 pod_annotation.go:316] Pod namespace/pod referencing an IPAMClaim on network "network" which does not honor it
--- FAIL: Test_allocatePodAnnotationWithRollback (0.00s)
    --- FAIL: Test_allocatePodAnnotationWithRollback/expect_new_IP_and_ipv6_gateway_LLA_for_primary_udn_layer2_with_dual_stack (0.00s)
        pod_annotation_test.go:825: 
            diff:   &util.PodAnnotation{
              	IPs: []*net.IPNet{
              		&{IP: s"192.168.0.3", Mask: s"ffffff00"},
              		&{
            - 			IP:   s"200:100:200::3",
            + 			IP:   s"2010:100:200::3",
              			Mask: s"fffffffffffffff00000000000000000",
              		},
              	},
              	MAC:      s"0a:58:c0:a8:00:03",
              	Gateways: {s"192.168.0.1", s"2010:100:200::1"},
              	... // 4 identical fields
              }
            
            Expected
                <*util.PodAnnotation | 0xc00041b0e0>: {
                    IPs: [
                        {
                            IP: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 192, 168, 0, 3],
                            Mask: [255, 255, 255, 0],
                        },
                        {
                            IP: [32, 16, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3],
                            Mask: [255, 255, 255, 255, 255, 255, 255, 240, 0, 0, 0, 0, 0, 0, 0, 0],
                        },
                    ],
                    MAC: [10, 88, 192, 168, 0, 3],
                    Gateways: [
                        [192, 168, 0, 1],
                        [32, 16, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1],
                    ],
                    IPv6LLAGateway: [254, 128, 0, 0, 0, 0, 0, 0, 8, 88, 100, 255, 254, 65, 0, 4],
                    Routes: [
                        {
                            Dest: {
                                IP: [100, 65, 0, 0],
                                Mask: [255, 255, 0, 0],
                            },
                            NextHop: [192, 168, 0, 1],
                        },
                        {
                            Dest: {
                                IP: [253, 153, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
                                Mask: [255, 255, 255, 255, 255, 255, 255, 255, 0, 0, 0, 0, 0, 0, 0, 0],
                            },
                            NextHop: [32, 16, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1],
                        },
                    ],
                    TunnelID: 100,
                    Role: "primary",
                }
            to equal
                <*util.PodAnnotation | 0xc0001d7b90>: {
                    IPs: [
                        {
                            IP: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 255, 255, 192, 168, 0, 3],
                            Mask: [255, 255, 255, 0],
                        },
                        {
                            IP: [2, 0, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3],
                            Mask: [255, 255, 255, 255, 255, 255, 255, 240, 0, 0, 0, 0, 0, 0, 0, 0],
                        },
                    ],
                    MAC: [10, 88, 192, 168, 0, 3],
                    Gateways: [
                        [192, 168, 0, 1],
                        [32, 16, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1],
                    ],
                    IPv6LLAGateway: [254, 128, 0, 0, 0, 0, 0, 0, 8, 88, 100, 255, 254, 65, 0, 4],
                    Routes: [
                        {
                            Dest: {
                                IP: [100, 65, 0, 0],
                                Mask: [255, 255, 0, 0],
                            },
                            NextHop: [192, 168, 0, 1],
                        },
                        {
                            Dest: {
                                IP: [253, 153, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
                                Mask: [255, 255, 255, 255, 255, 255, 255, 255, 0, 0, 0, 0, 0, 0, 0, 0],
                            },
                            NextHop: [32, 16, 1, 0, 2, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1],
                        },
                    ],
                    TunnelID: 100,
                    Role: "primary",
                }
FAIL
exit status 1
FAIL	github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/pod	0.015s
```

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -898,7 +898,13 @@ func (bnc *BaseNetworkController) allocatePodAnnotation(pod *kapi.Pod, existingL
...
 			podDesc, switchName)
 	}
-	err = util.AddRoutesGatewayIP(bnc.GetNetInfo(), pod, podAnnotation, network)
+
+	node, err := bnc.watchFactory.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return nil, false, err
+	}
+
+	err = util.AddRoutesGatewayIP(bnc.GetNetInfo(), node, pod, podAnnotation, network)
```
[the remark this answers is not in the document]
oshoval: keep me honest here please, this is non IC flow
so atm it is not used, only because https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876
but the condition that guard about it appears in the function which is used for both IC and non IC 
within `AddRoutesGatewayIP` so it is good to have it as it is now imho

oshoval:
Thanks
Addressed / answered comments

oshoval:
failures seems unrelated / known flaky

go-controller/pkg/util/pod_annotation.go
```diff
@@ -693,3 +712,8 @@ func UnmarshalUDNOpenPortsAnnotation(annotations map[string]string) ([]*OpenPort
 	}
 	return result, nil
 }
+
+// Ensure the IP is valid IPv6 and falls within the `fe80::/10` LLA range
+func isIPv6LLA(ip net.IP) bool {
+	return ip != nil && ip.To4() == nil && ip[0] == 0xfe && ip[1]&0xc0 == 0x80
```
qinqon, line 729: There is wais at golang "netip" to check if it's link local 

https://pkg.go.dev/net/netip#Addr.IsLinkLocalUnicast

We should try to use those first before going into hex territory.

so something like 
 ```golang
if utilnet.IsIPv6(addr.IP) && addr.IP.IsLinkLocalUnicast()
```

**Review by qinqon — CHANGES_REQUESTED** [review]

File: go-controller/pkg/util/pod_annotation.go (hunk 8 above)
Comment (oshoval) [reply go-controller/pkg/util/pod_annotation.go]: done thx

**Review by qinqon — APPROVED** [review]
LGTM

**Comment by oshoval** [comment]
Thanks
@maiqueb you might also want to take another look please ?

File: go-controller/pkg/util/pod_annotation.go [hunk 9]
```diff
@@ -570,6 +590,16 @@ func AddRoutesGatewayIP(
...
 				}
 			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						return err
+					}
+					podAnnotation.IPv6LLAGateway = HWAddrToIPv6LLA(IPAddrToHWAddr(joinAddrs[0].IP))
```
maiqueb, line 600: shouldn't we check the `joinAddrs` are not empty ? 
Do you have a test for that scenario ?

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: I mostly worry that we're not doing this anywhere else ... Are we ?

go-controller/pkg/util/pod_annotation.go
oshoval: maybe already verified here  https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6c366bb49d6fa57c80d6b39834952ce1d4d9dbb6/go-controller/pkg/ovn/secondary_layer2_network_controller.go#L622?

there are few unit tests about `ParseNodeGatewayRouterJoinAddrs`
i.e https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6c366bb49d6fa57c80d6b39834952ce1d4d9dbb6/go-controller/pkg/clustermanager/clustermanager_test.go#L918
if related

Imho even if we should at something in the end (which might not be the case, if the first link is good enough), it should be minimal, and not over testing,

go-controller/pkg/allocator/pod/pod_annotation_test.go
oshoval: OVN doesnt, OVN/vendor does fwiw
imho it is nit, and good to have
up to you

go-controller/pkg/util/pod_annotation.go
maiqueb: I don't see the relationship between your code and `addPortForRemoteNodeGR` (the fn you've linked).

IMHO it would have been OK if the parsing function checked it, but, we see it doesn't.

oshoval: would it be fine to just add an assert here ?
UT sounds redundant

oshoval: added a condition, please let me know if you request beside

go-controller/pkg/util/pod_annotation.go
```diff
@@ -570,6 +590,19 @@ func AddRoutesGatewayIP(
...
 			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						return err
+					}
+					if len(joinAddrs) == 0 {
+						return nil
```
maiqueb, line 601: I think we should return error.

If we don't, the GW will not be properly oconfigured, thus giving the user something that is not what they've requested.

go-controller/pkg/util/pod_annotation_unit_test.go
```diff
@@ -211,6 +218,20 @@ func TestUnmarshalPodAnnotation(t *testing.T) {
 			desc:        "verify successful unmarshal of pod annotation when *only* the MAC address is present",
 			inpAnnotMap: map[string]string{"k8s.ovn.org/pod-networks": `{"default":{"mac_address":"0a:58:fd:98:00:01"}}`},
 		},
+		{
```
maiqueb, line 221: please add a test without join ips.

maiqueb (changes requested):

go-controller/pkg/util/pod_annotation_unit_test.go
oshoval: this is the `UnmarshalPodAnnotation` one, it just have `k8s.ovn.org/pod-networks` and already have a test with just mac 
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878/files/88e1380294ad4b97a33e38e98291d10b29a36ad2#diff-8cc89a16276f003dd2f648bf85b5e23228e14807462ca836329d5159afcbe2d1L211

maybe you meant to add a test on `go-controller/pkg/allocator/pod/pod_annotation_test.go` for the new condition ?

maiqueb: My bad; yes, add it where you're saying.

It's a bit more complex, but you got it.

oshoval: I tried, i really think it is overkill for `go-controller/pkg/allocator/pod/pod_annotation_test.go`
it is too far and small / deep for such test
moreover no where in the code there is such case where `OVNNodeGRLRPAddrs` annotation has a network without ips (function is guarded that network do exists)
Please reconsider that if i may ask, another option is to wrap 
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878/files#diff-3d3fe81a31268662d021e60eb37f4a1f7e96ddb7e61de7f49acbbd2ae82e81caR596-R602
in a function and test it, but even that imo is too much

oshoval: if i set
```
ifaddrs = `{"network":{}}`
```
Then even without the new guard
we fail on 
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6c366bb49d6fa57c80d6b39834952ce1d4d9dbb6/go-controller/pkg/util/node_annotations.go#L700

is it fine to remove the guard / test then or you still want it one of them / both ?

go-controller/pkg/util/pod_annotation.go
oshoval: done, yep also raised it here https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878#discussion_r1895789390

oshoval:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12480204864/job/34831434078?pr=4878
didnt run fully, need to rerun maybe

go-controller/pkg/util/pod_annotation_unit_test.go
maiqueb: OK.

Please consider adding a code comment in the production code indicating that it is impossible to have that key in the annotations without IPs in it.

maiqueb:
> https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12480204864/job/34831434078?pr=4878 didnt run fully, need to rerun maybe

Re-ran this job ... Let's see what happens now.

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
maiqueb: Thank you @oshoval .

This is a nit, you can choose to ignore; I think it would be preferable if you'd asserted if the IP is an IPv6 LLA *before* assigning it to the podAnnotation. i.e. something like:

```go
llaGW := net.ParseIP(a.IPv6LLAGateway)
if !isIPv6LLA(llaGW) {
    return nil, fmt.Errorf("failed to parse pod ipv6 lla gateway, or non ipv6 lla %q", llaGW)
}
podAnnotation.IPv6LLAGateway = llaGW
```

go-controller/pkg/util/pod_annotation.go
```diff
@@ -570,6 +590,19 @@ func AddRoutesGatewayIP(
...
 			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						return err
+					}
+					if len(joinAddrs) == 0 {
+						return fmt.Errorf("failed to parse node GW router, joinAddrs is empty")
```
maiqueb, line 601:
```suggestion
						return fmt.Errorf("failed to compute the IPv6 GW LLA; node joinAddrs is empty")
```

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: I'm good with it if you are. 

... it's up to the maintainers, not me :)

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -161,11 +164,9 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			},
 			wantUpdatedPod: true,
 			wantPodAnnotation: &util.PodAnnotation{
-				IPs:  ovntest.MustParseIPNets("192.168.0.4/24"),
-				MAC:  util.IPAddrToHWAddr(ovntest.MustParseIPNets("192.168.0.4/24")[0].IP),
-				Role: types.NetworkRolePrimary,
```
maiqueb, line 166: why did you remove the `Primary` thing ? was it to sync w/ the test's description ? 

If so, please, isolate it in a separate commit. It's a bit distracting here.

go-controller/pkg/allocator/pod/pod_annotation_test.go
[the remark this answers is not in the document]
maiqueb: Yeah, I agree w/ keeping it simple; let's stick to the current approach.

maiqueb (dismissed):
I think this is good enough state to be reviewed by a maintainer. 

Thank you @oshoval .

I've left some nits you might want to consider.

go-controller/pkg/util/pod_annotation_unit_test.go
oshoval: imho it isn't required

oshoval:
rebased only

maiqueb:
@kyrtapz could you spare a round of reviews in this PR ?...

go-controller/pkg/allocator/pod/pod_annotation_test.go
oshoval: nice catch, please see the first commit, i returned the primary, but added disableMultiNetwork for this test
since adding EnableMultiNetwork affected this test (it does add routes / gateway in case multi networks + primary is enabled)

oshoval:
Thanks for the review
addressed comments

qinqon:
It has the changes I ask so LGTM

qinqon (approved):

maiqueb (approved):
Tks.

oshoval:
Thanks

@kyrtapz @tssurya can you please take a look ?

oshoval:
rebased to solve conflict

maiqueb (dismissed):
/lgtm

go-controller/pkg/allocator/pod/pod_annotation_test.go
ricky-rav: Oh my, the original diff is horrible! Thanks, Or!

go-controller/pkg/util/pod_annotation.go
```diff
@@ -590,6 +591,19 @@ func AddRoutesGatewayIP(
...
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						return err
+					}
+					if len(joinAddrs) == 0 {
+						return fmt.Errorf("failed to compute the IPv6 GW LLA; node joinAddrs is empty")
+					}
+					podAnnotation.IPv6LLAGateway = HWAddrToIPv6LLA(IPAddrToHWAddr(joinAddrs[0].IP))
```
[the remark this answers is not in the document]
oshoval: I think we don't support yet with it single stack IPv6 ? @qinqon knows best
based on his code here
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4795/commits/be476f4a22c311eb3b04009f51d55d92cc996ec2#diff-96c63415f0ac6412104d38c4645d5ddfca4cacfd0b4efe568579c324da764763R308

oshoval:
Thanks, addressed comments, going now to rebase in a separate push

oshoval:
rebased
ptal

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
qinqon: We do support ipv6 here, and the code here is fine since at singlestack ipv6 joinAddrs[0] is just the single ipv6 address so we good.

qinqon:
@oshoval I am rebasing the unsolicited RAs on top of this, I am not sure is really worth it, for RAs we also need to compute the source and target gateway routes mac address and I suspect we don't want to add `gateway_mac` annotation here, also using the migration source pod (that can be removed) as source of truth can be problematic.

ricky-rav (approved):

go-controller/pkg/ovn/base_network_controller_pods.go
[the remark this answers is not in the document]
kyrtapz: Hmm, isn't `allocatePodAnnotation` only called for default network though?
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/ovn/base_network_controller_pods.go#L560-L564

kyrtapz (commented):
LGTM
one remaining question: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878/commits/20389f9269eaa4519da23a180975d759e81cff46#r1983368025

go-controller/pkg/ovn/base_network_controller_pods.go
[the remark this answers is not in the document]
oshoval: interesting, @qinqon @maiqueb can you assist here please ?

qinqon (approved):
/lgtm

go-controller/pkg/ovn/base_network_controller_pods.go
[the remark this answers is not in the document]
qinqon: As far as I remember `IsSecondary` is true for primary UDNs too, so `allocatePodAnnotation` will stil be called for default pod network.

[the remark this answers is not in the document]
maiqueb: let's change as little things that used everywhere as possible. I.e. update only `allocatePodAnnotationForSecondaryNetwork`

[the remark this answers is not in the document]
kyrtapz: That's what I was thinking, +1.

oshoval:
just rebased atm

go-controller/pkg/ovn/base_network_controller_pods.go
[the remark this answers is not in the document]
oshoval: Looked deeper 
i think it is good as is,

The function that we changed is 
`func (allocator *PodAnnotationAllocator) AllocatePodAnnotation` [1]
not `func (bnc *BaseNetworkController) allocatePodAnnotation` [2] that appears here
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878#discussion_r1983368025

and [1] is called from `allocatePodAnnotationForSecondaryNetwork` that we all agree is relevant
as seen here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d66a8074f539d2c8125cb732f0046a9da15b1eea/go-controller/pkg/ovn/base_network_controller_pods.go#L977

oshoval:
tried to answer the request, please let me know what you think

one additional fix is to limit adding the annotation only when using Layer2, but it might be not straight forward and not worth, please let me know if you want me to dive deep into that

go-controller/pkg/ovn/base_network_controller_pods.go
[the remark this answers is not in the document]
oshoval: while you are right that i changed also `func (bnc *BaseNetworkController) allocatePodAnnotation`
since it calls `AddRoutesGatewayIP` [1]
which is common for both IC and non IC, and has the important logic [2] for those two,
if we remove the node from there, it means that we will need to check if node != nil, and then once non IC flow is fixed, to also remember to fix that temporary guard, i think it isn't nice, unless you think otherwise still please,
in such case please suggest what you would like to see and i will change

[1] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878/files#diff-3d3fe81a31268662d021e60eb37f4a1f7e96ddb7e61de7f49acbbd2ae82e81caR557

[2] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878/files#diff-3d3fe81a31268662d021e60eb37f4a1f7e96ddb7e61de7f49acbbd2ae82e81caR595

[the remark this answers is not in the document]
qinqon: @kyrtapz @maiqueb so passing "nil" instead of node from allocatePodAnnotation to AllocatePodAnnotation does not feel right, maybe we can keep this as it is so we can move on.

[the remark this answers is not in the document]
kyrtapz: Fine by me, I don't want to hold this PR over this.

oshoval:
@tssurya 
can you please review ?
as far as i can tell, the comments were addressed / not relevant (Quique at least agrees with me on that)

Thanks

tssurya:
before I start reviewing this @oshoval are all the red CI flakes known ? could you please provide a comment linking it to the flake issue?

oshoval:
one of them seems enviromental issue, that started since the last rebase, the others seems known, but i will close and reopen so it will retest, and then i will enclose lanes
i already checked in the past that they were all good, and since then didnt add changes beside rebase as far as i rememebr

oshoval:
@tssurya fyi please

the 2 failures (one per lane) both are known flakes that flakes heavily
https://github.com/ovn-kubernetes/ovn-kubernetes/issues?q=is%3Aissue%20state%3Aopen%20disabling%20egress%20nodes%20impeding%20GRCP%20health%20check

on one of the lanes 2 tests flaked, i guess we dont need to check deep about them, also the nature of this PR, is not intrusive that much fwiw

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -26,7 +27,7 @@ import (
 )
 
 type ipAllocatorStub struct {
-	netxtIPs         []*net.IPNet
+	nextIPs          []*net.IPNet
```
tssurya, line 30: lol nice catch

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -233,6 +237,78 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+						NextHop: ovntest.MustParseIP("2010:100:200::1"),
+					},
+				},
+				Role:     types.NetworkRolePrimary,
+				TunnelID: 100,
+			},
+			wantRelasedIDOnRollback:   true,
+			wantReleasedIPsOnRollback: ovntest.MustParseIPNets("2010:100:200::3/60"),
+			role:                      types.NetworkRolePrimary,
+		},
```
tssurya, line 311: i assume single single stack v4 not having it is tested by an existing test?

tssurya (commented):
can I see a oc get yaml of a pod with this annotation for the relevant use case and a check for what happens in a case where its not a L2 UDN pod and if its looking OK?

What about upgrades? what if a v6 UDN L2 pod gets upgraded? will the annotation on the pod be updated to include this field? I don't think we touch IPAM annotations in that case since it exists right? was this case tested?

also on the overall design - if the goal is to easily just have HWToIP based on JOINIP which is already on the node annotation.. why can't the code in the future where-ever this is used from be using that same logic? why do we need to store this on the pod's annotation specifically? I don't see this aspect of the design mentioned anywhere on the PR description.. I'll have to assume you need it on the CNI binary somewhere where you don't have access to annotation info.. -> if that is the case why was this not derived from the joinIP routes we add on the pod as next hop?

also if I have 100 L2 PUDN pods on network blue, I assume this value will be same for all pods on that blue network?

go-controller/pkg/util/pod_annotation.go
```diff
@@ -571,6 +592,19 @@ func AddRoutesGatewayIP(
...
 			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						return err
+					}
+					if len(joinAddrs) == 0 {
+						return fmt.Errorf("failed to compute the IPv6 GW LLA; node joinAddrs is empty")
```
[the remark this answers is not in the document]
oshoval: this guard is redundant, it is just for completeness, we wont reach here if joinAddrs is empty

we will fail earlier at
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6c366bb49d6fa57c80d6b39834952ce1d4d9dbb6/go-controller/pkg/util/node_annotations.go#L700

so we don't need to test it and such

go-controller/pkg/allocator/pod/pod_annotation_test.go
oshoval: it fails earlier, much before our code
at https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6c366bb49d6fa57c80d6b39834952ce1d4d9dbb6/go-controller/pkg/util/node_annotations.go#L700

so i didnt add tests for that, it is either tested already or not reaching the new code (since it fails before that)

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
tssurya: not sure I understand the answer...let me try to see if I understood you correctly:
my question was if its a dualstack pod, if the v4 bits are good but not the v6 will we fail here?

your answer is that during parse annotations itself we'd fail and so this check is redundant? If its redundant code please remove it.. its causing more confusion than anything else. and if you want to keep it please include a comment stating why and how we won't ever hit this condition and yet you have it here.

go-controller/pkg/allocator/pod/pod_annotation_test.go
tssurya: no I mean I see tests for v6 and dualstack.. so do you mean to say tests for v4 exist already -> a test that checks the code you are adding is not taking effect for v4...

tssurya: I'd say similarly a test for a role:secondary pod to ensure nothing is changing there

go-controller/pkg/util/pod_annotation.go
```diff
@@ -571,6 +592,19 @@ func AddRoutesGatewayIP(
 					podAnnotation.Gateways = append(podAnnotation.Gateways, gatewayIPnet.IP)
 				}
 			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						return err
```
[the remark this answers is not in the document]
oshoval: will wrap

iiuc this runs under the retry framework, it is bit fundamental information relatively that would exists on the node, beside rare corner cases, that will simply reconcile (same for the other errors there more or less),
we cant supress it, but need to fail and we do expect that the annotation will be there at the end
so i would suggest to keep as is

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
oshoval: > not sure I understand the answer...let me try to see if I understood you correctly: my question was if its a dualstack pod, if the v4 bits are good but not the v6 will we fail here?

Yes if we are dualstack, and the parsing failed, we fail here and it fine (referring to question here)
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878#discussion_r1993238529
`ParseNodeGatewayRouterJoinAddrs` does few checks, for example extract the annotations, parse the cidrs etc
a failure is a failure, we don't look on the resolution here and accept some and fail some,
there is no need for such resolution

> your answer is that during parse annotations itself we'd fail and so this check is redundant? If its redundant code please remove it.. its causing more confusion than anything else. and if you want to keep it please include a comment stating why and how we won't ever hit this condition and yet you have it here.

agree, will remove

go-controller/pkg/util/pod_annotation_unit_test.go
```diff
@@ -213,6 +220,20 @@ func TestUnmarshalPodAnnotation(t *testing.T) {
...
+			desc:        "verify successful unmarshal of pod annotation with ipv6 lla gateway ip",
+			inpAnnotMap: map[string]string{"k8s.ovn.org/pod-networks": `{"default":{"mac_address":"0a:58:fd:98:00:01","ipv6_lla_gateway_ip":"fe80::858:fdff:fe98:1"}}`},
```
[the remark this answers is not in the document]
oshoval: yep only for l2, changing to reflect that instead the default

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -973,9 +979,14 @@ func (bnc *BaseNetworkController) allocatePodAnnotationForSecondaryNetwork(pod *
...
 	}
-
+	node, err := bnc.watchFactory.GetNode(pod.Spec.NodeName)
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to get pod %s/%s/%s node %q: %w",
+			nadName, pod.Namespace, pod.Name, pod.Spec.NodeName, err)
+	}
 	updatedPod, podAnnotation, err := bnc.podAnnotationAllocator.AllocatePodAnnotation(
 		ipAllocator,
+		node,
```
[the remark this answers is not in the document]
oshoval: it is used, if we follow the call stack here, we will reach `AddRoutesGatewayIP` which calls `ParseNodeGatewayRouterJoinAddrs` that use it

oshoval:
Addressed / answered comments beside the link below
will now answer the questions on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4878#pullrequestreview-2681322675

will update once done, thanks

go-controller/pkg/allocator/pod/pod_annotation_test.go
oshoval: right added

oshoval:
> can I see a oc get yaml of a pod with this annotation for the relevant use case and a check for what happens in a case where its not a L2 P-UDN pod and if its looking OK?

TODO

> What about upgrades? what if a v6 UDN L2 pod gets upgraded? will the annotation on the pod be updated to include this field? I don't think we touch IPAM annotations in that case since it exists right? was this case tested?

TODO 

> also on the overall design - if the goal is to easily just have HWToIP based on JOINIP which is already on the node annotation.. why can't the code in the future where-ever this is used from be using that same logic? why do we need to store this on the pod's annotation specifically? I don't see this aspect of the design mentioned anywhere on the PR description.. I'll have to assume you need it on the CNI binary somewhere where you don't have access to annotation info.. -> if that is the case why was this not derived from the joinIP routes we add on the pod as next hop?

TODO - Will add in a new comment so it will be nicer

> also if I have 100 L2 PUDN pods on network blue, I assume this value will be same for all pods on that blue network?

It is derived from the node, so if the pods are on other nodes, each pod will have the one acccording its own node
If it helps, last commit on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4852 shows what we are doing with that
based on that we create the nftables that block RA from non local node, so we won't have multi path gateways

oshoval:
[Fixed](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/1e648ea43c78253bf676fe7c000d822c999b921c..30c1069b9fac4fbf663c734fce17e590ab2b7277) UT

oshoval:
Notes about the design in this PR.

Motivation: Fix multi path issue, see [1]
At current layer2 topology each node has gateway router ipv6 link local ip and mac addresses, this end up with a multipath ipv6 default gateways at VMs.

Originally, we tried to configure OVNs ACLs to drop RA from non local gateways, but since ND_RA is always accepted [2] we had to fall back to using nft tables.

The proposed work-around:
We will create Nftables rules on the pod network namespace.
For that, we require the gateway local link address, so we can exclude it from the rule (i.e. we want to accept those RAs, drop all others).

We considered the following two ways to get the local LRP LLA: (pros and cons inline).

1. Allow the CNI side to read the node, by adding node getter to the clientSet.
Using the node annotations we can compute the local gateway LLA.
The advantage is the logic is concentrated in one place, but it requires reading nodes on the CNI side (currently, we only read pods).
It affects both CNI and CNI shim, due to the common “getter” interface.

2. Derive the gateway LLA on the OVN side, and pass it via CNI_ARGS,
This way the CNI is kept cleaner. We would assume the presence of this attribute on CNI args to mean we need to configure these rules (which are only meant for primary UDNs with IPv6 subnets).

3. After discussion about the above, Patryk suggested instead to add the pod annotation we introduced on this PR.

Please keep in mind that https://issues.redhat.com/browse/FDP-872 addresses this issue in the proper layer, and once it is available, we can drop the work-around we introduce here.

[1]
```
# ip -6 route (in a Kubevirt VM)
::1 dev lo proto kernel metric 256 pref medium
2014:100:200::5 dev eth0 proto kernel metric 100 pref medium
2014:100:200::/60 dev eth0 proto ra metric 100 pref medium
fd99::/64 dev eth0 proto ra metric 100 pref medium
fe80::/64 dev eth0 proto kernel metric 100 pref medium
default proto ra metric 100 pref low
	nexthop via fe80::858:64ff:fe41:3 dev eth0 weight 1 
	nexthop via fe80::858:64ff:fe41:4 dev eth0 weight 1 
	nexthop via fe80::858:64ff:fe41:2 dev eth0 weight 1
```

[2] https://github.com/ovn-org/ovn/blob/af3a76020187b0cd78d6c0bdaf5107b2323c3e15/northd/northd.c#L7303C29-L7303C40

maiqueb:
> > can I see a oc get yaml of a pod with this annotation for the relevant use case and a check for what happens in a case where its not a L2 P-UDN pod and if its looking OK?
> 
> TODO
> 
> > What about upgrades? what if a v6 UDN L2 pod gets upgraded? will the annotation on the pod be updated to include this field? I don't think we touch IPAM annotations in that case since it exists right? was this case tested?
> 
> TODO

It would surely be nice if we did - but isn't the answer if we reconcile all pods on ovnkube-control-plane restart ? I assume we do.  We just need to ensure that we still add this attribute if the annotation already exists. And I think we recompute the entire annotation on the following path:
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/532c7d0344256fa793dc38fe292336bdd9d8d00c/go-controller/pkg/ovn/base_network_controller_secondary.go#L66
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/532c7d0344256fa793dc38fe292336bdd9d8d00c/go-controller/pkg/ovn/base_network_controller_secondary.go#L292
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/532c7d0344256fa793dc38fe292336bdd9d8d00c/go-controller/pkg/ovn/base_network_controller_secondary.go#L334
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/ovn/base_network_controller_pods.go#L561

@oshoval could you confirm this actually happens ? 

> 
> > also on the overall design - if the goal is to easily just have HWToIP based on JOINIP which is already on the node annotation.. why can't the code in the future where-ever this is used from be using that same logic? why do we need to store this on the pod's annotation specifically? I don't see this aspect of the design mentioned anywhere on the PR description.. I'll have to assume you need it on the CNI binary somewhere where you don't have access to annotation info.. -> if that is the case why was this not derived from the joinIP routes we add on the pod as next hop?
> 
> Added below the design options that we considered
> 
> > also if I have 100 L2 PUDN pods on network blue, I assume this value will be same for all pods on that blue network?
> 
> It is derived from the node, so if the pods are on other nodes, each pod will have the one acccording its own node If it helps, last commit on #4852 shows what we are doing with that based on that we create the nftables that block RA from non local node, so we won't have multi path gateways

maiqueb:
> > > can I see a oc get yaml of a pod with this annotation for the relevant use case and a check for what happens in a case where its not a L2 P-UDN pod and if its looking OK?
> > 
> > 
> > TODO
> > > What about upgrades? what if a v6 UDN L2 pod gets upgraded? will the annotation on the pod be updated to include this field? I don't think we touch IPAM annotations in that case since it exists right? was this case tested?
> > 
> > 
> > TODO
> 
> It would surely be nice if we did - but isn't the answer if we reconcile all pods on ovnkube-control-plane restart ? I assume we do. We just need to ensure that we still add this attribute if the annotation already exists. And I think we recompute the entire annotation on the following path:
> 
>     * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/532c7d0344256fa793dc38fe292336bdd9d8d00c/go-controller/pkg/ovn/base_network_controller_secondary.go#L66
> 
>     * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/532c7d0344256fa793dc38fe292336bdd9d8d00c/go-controller/pkg/ovn/base_network_controller_secondary.go#L292
> 
>     * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/532c7d0344256fa793dc38fe292336bdd9d8d00c/go-controller/pkg/ovn/base_network_controller_secondary.go#L334
> 
>     * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/ovn/base_network_controller_pods.go#L561
> 
> 
> @oshoval could you confirm this actually happens ?

I think you can check it by:
- checkout ovn-k master; deploy OVN-K ensuring the proper flags are passed. Ensure IPv6 is setup on the cluster.
-  create a pod attached to a primary UDN which has an IPv6 subnet.
- ensure it doesn't have the annotation.
- redeploy the components, using the --deploy flag. This emulates an upgrade.
- check if the pod got an updated annotation. If it doesn't, you'll need to trace the flow. check the code path I've pasted in the comment above. I think that's it, but I'm not 100% sure, I might have tracked the non-IC flow by mistake.

oshoval:
as far as i see, we don't reconcile all pods upon OVN upgrade
for example we don't reach `AddRoutesGatewayIP` which update the annotation upon upgrade, only on create
please let me know about it, how you think we should continue, 
if it can be a follow-up effort upon needs, will be great

oshoval:
rebased

oshoval:
about
```
can I see a oc get yaml of a pod with this annotation for the relevant use case and a check for what happens in a case where its not a L2 P-UDN pod and if its looking OK?
```

with the new field (at the bottom, truncated some)
```
{
  "default": {
   ...
  },
  "poc1/l2": {
    "ip_addresses": [
      "10.100.200.3/24",
      "2014:100:200::3/60"
    ],
    "mac_address": "0a:58:0a:64:c8:03",
    "gateway_ips": [
      "10.100.200.1",
      "2014:100:200::1"
    ],
    "routes": [
  ...
    ],
    "ipv6_lla_gateway_ip": "fe80::858:64ff:fe41:2",
    "tunnel_id": 7,
    "role": "primary"
  }
}
```

if nad doesnt have ipv6, pod starts fine, just without the field
```
{
  "default": {
    "ip_addresses": [
      "10.244.1.9/24",
      "fd00:10:244:2::9/64"
    ],
    "mac_address": "0a:58:0a:f4:01:09",
    "routes": [
      {
        "dest": "10.244.0.0/16",
        "nextHop": "10.244.1.1"
      },
      {
        "dest": "100.64.0.0/16",
        "nextHop": "10.244.1.1"
      },
      {
        "dest": "fd00:10:244::/48",
        "nextHop": "fd00:10:244:2::1"
      },
      {
        "dest": "fd98::/64",
        "nextHop": "fd00:10:244:2::1"
      }
    ],
    "role": "infrastructure-locked"
  },
  "poc1/l2": {
    "ip_addresses": [
      "10.100.200.3/24"
    ],
    "mac_address": "0a:58:0a:64:c8:03",
    "gateway_ips": [
      "10.100.200.1"
    ],
    "routes": [
      {
        "dest": "10.96.0.0/16",
        "nextHop": "10.100.200.1"
      },
      {
        "dest": "100.65.0.0/16",
        "nextHop": "10.100.200.1"
      }
    ],
    "ip_address": "10.100.200.3/24",
    "gateway_ip": "10.100.200.1",
    "tunnel_id": 5,
    "role": "primary"
  }
}
```

maiqueb:
> for example we don't reach `AddRoutesGatewayIP` which update the annotation upon upgrade, only on create
> please let me know about it, how you think we s

No matter if we tackle this as a follow up or not, you should start investigating how we can reconcile this on a restart. 

Can you debug the code on OVN restart to figure out how can that be achieved ? 
@oshoval

maiqueb:
> > > > What about upgrades? what if a v6 UDN L2 pod gets upgraded? will the annotation on the pod be updated to include this field? I don't think we touch IPAM annotations in that case since it exists right? was this case tested?
> 
> > > for example we don't reach `AddRoutesGatewayIP` which update the annotation upon upgrade, only on create
> > > please let me know about it, how you think we ...
> > 
> > 
> > No matter if we tackle this as a follow up or not, you should start investigating how we can reconcile this on a restart.
> > Can you debug the code on OVN restart to figure out how can that be achieved ? @oshoval
> 
> we have [this](https://github.com/oshoval/ovn-kubernetes/blob/master/go-controller/pkg/ovn/secondary_layer2_network_controller.go#L232) which will sync the pods, in case it can also update them (like Update func does maybe) Note please that the annotation doesnt affect migration, because Quique did ad hoc solution for migration, as he needed more parameters

So the way I see it is we *need* to reconcile on control-plane upgrades. Meaning, if we want to allow existing VMs to be able to successfully migrate to another node (without breaking the established connections) we need to update their annotation to feature this information.

Hence, we need to reconcile the annotations on control-plane restart.
@qinqon please coordinate with @oshoval . Maybe he needs help (we'd really appreciate to know / understand on what), maybe he needs to offload this task to you.

qinqon:
> > > > > What about upgrades? what if a v6 UDN L2 pod gets upgraded? will the annotation on the pod be updated to include this field? I don't think we touch IPAM annotations in that case since it exists right? was this case tested?
> > 
> > 
> > > > for example we don't reach `AddRoutesGatewayIP` which update the annotation upon upgrade, only on create
> > > > please let me know about it, how you think we ...
> > > 
> > > 
> > > No matter if we tackle this as a follow up or not, you should start investigating how we can reconcile this on a restart.
> > > Can you debug the code on OVN restart to figure out how can that be achieved ? @oshoval
> > 
> > 
> > we have [this](https://github.com/oshoval/ovn-kubernetes/blob/master/go-controller/pkg/ovn/secondary_layer2_network_controller.go#L232) which will sync the pods, in case it can also update them (like Update func does maybe) Note please that the annotation doesnt affect migration, because Quique did ad hoc solution for migration, as he needed more parameters
> 
> So the way I see it is we _need_ to reconcile on control-plane upgrades. Meaning, if we want to allow existing VMs to be able to successfully migrate to another node (without breaking the established connections) we need to update their annotation to feature this information.
> 
> Hence, we need to reconcile the annotations on control-plane restart. @qinqon please coordinate with @oshoval . Maybe he needs help (we'd really appreciate to know / understand on what), maybe he needs to offload this task to you.

Sure, whatever needed.

oshoval:
Handling Upgrade flow initial design discussion

1. In sync function [1], need to calculate the ipv6 lla gateway
And use it to create the nftables (see 2). No need to patch the pod iiuc, because upon migration it will be calculated again.
2. Using the lla addr, we need to create the Nftables
For that, we need to enter the virt-launcher pod from the secondary l2 network controller, is it doable ? the controller is on the host network so it can enter ns of any ?
Note that it is expansive, and maybe a bit intrusive by design ?
Meaning that until now the controller didn’t do such things, it adds complexity instead leaving the controller light.
Also need to make sure that on IC mode, each pod mentioned on [1] runs on the right node, or at least syncs all pods, and then we need to filter the local ones only
3. We might need to also invalidate the multi path entries that already exist, since we create the Nftables on the fly, it means that entries are there already and need to be clean, this also makes the flow a bit heavier.

Should we consider a better design, if we think deeper ?
Lets first consider those points high level imo and then we can advance this way or another
Thanks

[1] https://github.com/oshoval/ovn-kubernetes/blob/master/go-controller/pkg/ovn/secondary_layer2_network_controller.go#L232

qinqon:
> Handling Upgrade flow initial design discussion
> 
> 1. In sync function [1], need to calculate the ipv6 lla gateway
>    And use it to create the nftables (see 2). No need to patch the pod iiuc, because upon migration it will be calculated again.
> 2. Using the lla addr, we need to create the Nftables
>    For that, we need to enter the virt-launcher pod from the secondary l2 network controller, is it doable ? the controller is on the host network so it can enter ns of any ?
>    Note that it is expansive, and maybe a bit intrusive by design ?
>    Meaning that until now the controller didn’t do such things, it adds complexity instead leaving the controller light.
>    Also need to make sure that on IC mode, each pod mentioned on [1] runs on the right node, or at least syncs all pods, and then we need to filter the local ones only
> 3. We might need to also invalidate the multi path entries that already exist, since we create the Nftables on the fly, it means that entries are there already and need to be clean, this also makes the flow a bit heavier.
> 
> Should we consider a better overall design, if we think deeper ? Lets first consider please those points high level imo and then we can advance this way or another Thanks
> 
> [1] https://github.com/oshoval/ovn-kubernetes/blob/master/go-controller/pkg/ovn/secondary_layer2_network_controller.go#L232

So since we don't support IPv6 for VMs before this fixes we should not care about upgrades, there are not going to be VMs at primary UDN with IPv6 before this.

And the scenario where we are with latest ovn-kubernetes and we want to migrate a VM from ipv4 to dualstack or singlestack ipv6 means at least vm hard restart or even VM reconstruction, it also involve re-create UDNs/modify  UDNs since it means changing `Subnet` field.

oshoval:
As discussed offline,
we don't need to fix the virt-launcher before the migration wrt having the annotation / nftables  / non multi path.

After Upgrade, in order to consume the fix for "Established TCP connections over IPv6 on east/west traffic *breaks* when on VM live-migration"
a user needs to live migrate, which will also will create the required nftables and send RA = 0 to clean the non required ones, hence migration works fine, and nothing need to be done wrt upgrade.

One point that was found, is that when sending RA = 0 to invalidate the outdated entries, we need to send it about all non local node, so it will clean all the non required entries, in case the VM was created before having the fix.

Thanks

oshoval:
@tssurya can you please take another look ?
addressed the questions / comments

failures arent related, need to retest

thanks

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
tssurya: @oshoval 

> but need to fail and we do expect that the annotation will be there at the end (i wouldnt worry from event burst and such)

well this will fail on downstream CI, we have a cap on event bursts.

tssurya:
> as far as i see, we don't reconcile all pods upon OVN upgrade (i.e virt-launcher) for example we don't reach `AddRoutesGatewayIP` which update the annotation upon upgrade, only on create please let me know about it, how you think we should continue, if it can be a follow-up effort upon needs, will be great

hmm it can't be a followup.. I think we need the fields to be consistent on all v6 l2 pods right? We can't have some without and some with. Because the code where you will then use this annotation will need those extra checks? -> if a pod doesn't have that field in its annotation do nothing? How do we then explain "we have fixed this problem" :D just recreate your pods? :)
I was hoping during reboot somehow they will get added but I agree with your theory that existing annotations don't get updated, they only can done during creation. Can we please test this?

tssurya:
> As discussed offline, we don't need to fix the virt-launcher before the migration wrt having the annotation / nftables / non multi path.
> 
> After Upgrade, in order to consume the fix for "Established TCP connections over IPv6 on east/west traffic _breaks_ when on VM live-migration" a user needs to live migrate, which will also will create the required annotation, nftables and send RA = 0 to clean the non required ones (all beside new local node address), hence migration works fine, and nothing need to be done wrt upgrade.
> 
> One point that was found, is that when sending RA = 0 to invalidate the outdated entries, we need to send it about all non local node, so it will clean all the non required entries, in case the VM was created before having the fix.
> 
> Thanks

migration and upgrades is confusing me. let's talk about this in the meeting

go-controller/pkg/util/pod_annotation.go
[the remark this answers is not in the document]
oshoval: can you please point on an example how to adapt the code / or to elaborate how you suggest ?
not sure how to retry but skip,
otoh we must make sure that a pod that runs in the very beginning, wouldn't skip getting the annotation, else it would end with multi path

oshoval:
Hi @tssurya 
Did the change you asked, thanks (it is the 2 last pushes, which are small)
all tests passed beside https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14337838503/job/40191623537?pr=4878
which failed due to kind setup, so not related, i think the PR is safe please to review
Thanks

tssurya:
> @tssurya @oshoval so during upgrades VMs get automatically live migrated meaning that the target virt-launcher pod will have the proper lla annotation and the nftables configured with it.

@qinqon did you test it? did you see if happening during upgrades? if so can you show me the link to the test run?

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -124,6 +125,9 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 		wantReleaseID             bool
 		wantRelasedIDOnRollback   bool
 		wantErr                   bool
+		isSingleStackIPv4         bool
```
tssurya, line 128: nit: to add a comment for what these two extra bolleans are doing -> from what I can tell forcing the single stack ness in tests which by default is dualstack?
instead using these two toggles to tell what the stack should be might have been less confusing:
i.e if dualstack then set both to true, if single stack v4 only set 1 of them to true and same for v6 -> then use those values to determine the config.ipv4 and config.ipv6
but i won't hold for this.. its just confusing to maintain in future can be a FUP.

go-controller/pkg/util/pod_annotation.go
```diff
@@ -571,6 +592,19 @@ func AddRoutesGatewayIP(
...
 			}
+			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					if err != nil {
+						if IsAnnotationNotSetError(err) {
+							return types.NewSuppressedError(err)
+						}
+						return fmt.Errorf("failed parsing node gateway router join addresses, %w", err)
```
tssurya, line 603:
```suggestion
						return fmt.Errorf("failed parsing node gateway router join addresses, %w", err)
```
for which network would be helpful message to include in this error...

go-controller/pkg/util/pod_annotation.go
```diff
@@ -571,6 +592,19 @@ func AddRoutesGatewayIP(
...
+					}
+					podAnnotation.GatewayIPv6LLA = HWAddrToIPv6LLA(IPAddrToHWAddr(joinAddrs[0].IP))
```
tssurya, line 605: joinAddrs[0] has the IPV4 joinIP? which we convert to MAC and then convert to LLA ok gotcha

qinqon:
> > @tssurya @oshoval so during upgrades VMs get automatically live migrated meaning that the target virt-launcher pod will have the proper lla annotation and the nftables configured with it.
> 
> @qinqon did you test it? did you see if happening during upgrades? if so can you show me the link to the test run? until yesterday @oshoval told us he tested annotation bits - they are not going to change during upgrades/migrations.. has this changed? so if the vm has migrated then the annotation get's updated -> even if IPAM is persistent and we won't update IPAM annotation?

@tssurya from the openshift virt team team 

```
OCP upgrade typically involves updates to RHCOS RPMs, and that requires reboot. So it is safe to assume that during every OCP z-stream release, nodes would go down one by one, forcing live migration to the next node. After the upgrade, all the VMs will be migrated at least once. There is no such thing as moving them back (only by accident, if they moved away to another old node, and then move again to the node that was just upgraded)
```

Although in upstream we will need a forced live migration

tssurya (commented):
two nits which can be FUPs - although that error info container network will be really using if we have 100's of networks, but otherwise my comments were addressed
thank you @oshoval 

waiting for @qinqon to reply to my above comment,

go-controller/pkg/allocator/pod/pod_annotation_test.go
oshoval: Prefer follow-up please

go-controller/pkg/util/pod_annotation.go
oshoval: done

tssurya (approved):

oshoval:
the other failures is a known flake

tssurya:
> #4936

reopened this flake

tssurya:
upgrades weren't tested since pre-merge is hard for virt.. post-merge testing will be done to confirm things.

PR #4882, opened 2024-12-10
Calculates mgmt port MAC rather than storing

trozet:
Today ovnkube-controller relies on ovnkube-node to annotate the mac address so that it can properly program the OVN mgmt port. In practice with UDN, this means cluster manager creates the node and allocates some subnet info to the node, both the node and the ovnkube-controller side get updates, but ovnkube-controller fails to program the first time because it is waiting for the node side to configure the mac address.

This patch changes the behavior of ovn-kubernetes to calculate the MAC address for the management port from the mgmt port IP address of the first subnet for the network. For backwards compatibility (on default network), it will first attempt to read the node annotation, and if it no mgmt MAC exists, it will derive it from the mgmt IP of the subnet.

Note the DPU code already uses the mgmt IP to derive the MAC.

### DIFF
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"strings"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -290,7 +289,7 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	// port is created first and its MAC address configured. The IP(s) on that link are added after enslaving to a VRF device (addUDNManagementPortIPs)
 	// because IPv6 addresses are removed by the kernel (if not link local) when enslaved to a VRF device.
 	// Add the routes after setting the IP(s) to ensure that the default subnet route towards the mgmt network exists.
-	mplink, macAddress, err := udng.addUDNManagementPort()
+	mplink, err := udng.addUDNManagementPort()
 	if err != nil {
 		return fmt.Errorf("could not create management port netdevice for network %s: %w", udng.GetNetworkName(), err)
 	}
@@ -308,9 +307,6 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	}
 	if err = udng.vrfManager.AddVRFRoutes(vrfDeviceName, routes); err != nil {
 		return fmt.Errorf("could not add VRF %s routes for network %s, err: %v", vrfDeviceName, udng.GetNetworkName(), err)
-	}
-	if err := util.UpdateNodeManagementPortMACAddressesWithRetry(udng.node, udng.nodeLister, udng.kubeInterface, macAddress, udng.GetNetworkName()); err != nil {
-		return fmt.Errorf("unable to update mac address annotation for node %s, for network %s, err: %w", udng.node.Name, udng.GetNetworkName(), err)
 	}
 	// create the iprules for this network
 	udnReplyIPRules, err := udng.constructUDNVRFIPRules(vrfTableId)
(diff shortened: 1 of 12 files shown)


trozet:
some unrelated egress firewall flake:

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12261395552/job/34208326354?pr=4882

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -627,10 +627,21 @@ func (bnc *BaseNetworkController) deleteNamespaceLocked(ns string) (*namespaceIn
...
 
 func (bnc *BaseNetworkController) syncNodeManagementPort(node *kapi.Node, switchName, routerName string, hostSubnets []*net.IPNet) ([]net.IP, error) {
-	macAddress, err := util.ParseNodeManagementPortMACAddresses(node, bnc.GetNetworkName())
-	if err != nil {
+	// get mac address from node only for legacy reasons, if it doesn't exist, then calculate it from subnets
+	var macAddress net.HardwareAddr
+	var err error
+	// find suitable MAC address
+	// check node annotation first, to ensure we are not picking a new MAC when one was already configured
+	if macAddress, err = util.ParseNodeManagementPortMACAddresses(node, bnc.GetNetworkName()); err != nil && !util.IsAnnotationNotSetError(err) {
```
kyrtapz, line 635: When configuring the management port in UDN we always use the IP address as base, so there is no point of looking for annotation for UDN networks?
Do you plan to handle UDNs that have the annotation?

trozet: >When configuring the management port in UDN we always use the IP address as base, so there is no point of looking for annotation for UDN networks?

Right.

>Do you plan to handle UDNs that have the annotation?

Originally I had this same check for if the node had the UDN mac annotation we would use it. This was for handling upgrade where UDNs were already programmed. After thinking about it some, I realized UDN is not GA so we don't need to handle this case. We can make sure this gets backported to 4.17z and pretend UDN never used that annotation :)

go-controller/pkg/node/gateway_udn.go
```diff
@@ -401,39 +397,35 @@ func (udng *UserDefinedNetworkGateway) DelNetwork() error {
...
-	}
-	// persist the MAC address so that upon node reboot we get back the same mac address.
-	_, stderr, err = util.RunOVSVsctl("set", "interface", interfaceName,
-		fmt.Sprintf("mac=%s", strings.ReplaceAll(macAddress.String(), ":", "\\:")))
-	if err != nil {
-		return nil, nil, fmt.Errorf("failed to persist MAC address %q for %q while plumbing network %s: stderr:%s (%v)",
-			macAddress.String(), interfaceName, udng.GetNetworkName(), stderr, err)
-	}
-
-	// STEP3
```
kyrtapz, line 433: NIT: There is a STEP4 comment below

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -818,11 +818,17 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
 		return nil, fmt.Errorf("failed to get networkID for network %q: %v", bsnc.GetNetworkName(), err)
 	}
+	// legacy lookup for mac
 	dstMac, err := util.ParseNodeManagementPortMACAddresses(node, bsnc.GetNetworkName())
-	if err != nil {
+	if err != nil && !util.IsAnnotationNotSetError(err) {
 		return nil, fmt.Errorf("failed to parse mac address annotation for network %q on node %q, err: %w",
 			bsnc.GetNetworkName(), node.Name, err)
 	}
+	if len(dstMac) == 0 && len(localPodSubnets) > 0 {
```
kyrtapz, line 827: NIT: localPodSubnets length is already checked at the beginning of the functon

go-controller/pkg/ovn/base_network_controller.go
trozet: That shouldn't happen, but I can make a guard checking if this controller is default or not for the annotation check. Does that work for you?

kyrtapz: I am ok to keep it as is if this is not a possible issue.

kyrtapz (dismissed):

kyrtapz (approved):

go-controller/pkg/node/gateway_udn.go
trozet: fixed

trozet:
I missed updating some UDN unit tests which were depending on the annotation. Updating those now in next push.

kyrtapz (approved):

PR #4884, opened 2024-12-10
UDN: Rename NetworkReady condition to NetworkCreated

ormergi:
## 📑 Description

As of today the `NetworkReady` condition indicated a NAD has been created.
And not necessarily that the underlying network is ready to work with,
because it require other internal components to act (e.g.: set ovs ports,
ovn flows, etc..).

Rename the `NetworkReady` condition type to `NetworkCreated` so it describe
better what it indicates.

This change enable introducing alternative `NetworkReady` condition that
provider actual indication a UDN network is ready, and that other
internal component acted successfully.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
@@ -84,9 +84,9 @@ var _ = Describe("User Defined Network Controller", func() {
 					Expect(err).NotTo(HaveOccurred())
 					return normalizeConditions(udn.Status.Conditions)
 				}).Should(Equal([]metav1.Condition{{
-					Type:    "NetworkReady",
+					Type:    "NetworkCreated",
 					Status:  "True",
-					Reason:  "NetworkAttachmentDefinitionReady",
+					Reason:  "NetworkAttachmentDefinitionCreated",
 					Message: "NetworkAttachmentDefinition has been created",
 				}}))
 
@@ -107,7 +107,7 @@ var _ = Describe("User Defined Network Controller", func() {
 					Expect(err).NotTo(HaveOccurred())
 					return normalizeConditions(udn.Status.Conditions)
 				}).Should(Equal([]metav1.Condition{{
-					Type:    "NetworkReady",
+					Type:    "NetworkCreated",
 					Status:  "False",
 					Reason:  "SyncError",
 					Message: "failed to generate NetworkAttachmentDefinition: " + renderErr.Error(),
@@ -132,7 +132,7 @@ var _ = Describe("User Defined Network Controller", func() {
 					Expect(err).NotTo(HaveOccurred())
 					return normalizeConditions(udn.Status.Conditions)
 				}).Should(Equal([]metav1.Condition{{
-					Type:    "NetworkReady",
+					Type:    "NetworkCreated",
 					Status:  "False",
 					Reason:  "SyncError",
 					Message: "failed to create NetworkAttachmentDefinition: create NAD error",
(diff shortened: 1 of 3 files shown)


go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -84,9 +84,9 @@ var _ = Describe("User Defined Network Controller", func() {
 					Expect(err).NotTo(HaveOccurred())
 					return normalizeConditions(udn.Status.Conditions)
 				}).Should(Equal([]metav1.Condition{{
-					Type:    "NetworkReady",
+					Type:    "NetworkCreated",
```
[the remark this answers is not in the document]
ormergi: Did you mean for a const in controller_test.go, used in unit-tests only?

Please note that  since the controller.go  and controller_test.go defined under the same package (`controller`), 
if the const is added to controller_test.go nothing prevents the next dev use it in controller.go implicitly. 

Having the condition type hardcoded in tests has a benefit, changing it will not go silently as tests will break, and the reviewer will have clear indication a condition type was changed.
If the code and unit-tests are separated to different package (controller_test.go under `controller_test`) we could have an additional const in controller_test.go for convenience.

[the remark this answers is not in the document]
ormergi: It make sense, but note that if we use this const in tests, and it value changes, unit-tests will not fail and we lose this indication as tests will pass.
Having it hardcoded in tests, ensure CI will be red on such changes.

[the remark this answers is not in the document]
npinaeva: you can leave unit tests with string if you find it better, but I would add a constant at least in the controller code itself, and ensure udn and cudn use the same condition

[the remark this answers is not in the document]
ormergi: Done, PTAL

npinaeva (approved):

test/e2e/network_segmentation.go
```diff
@@ -1533,12 +1532,12 @@ func generateLayer3Subnets(cidrs string) []string {
 }
 
 func waitForUserDefinedNetworkReady(namespace, name string, timeout time.Duration) error {
```
tssurya, line 1534: nit: also change the name of the utils to reflect its networkcreated?

test/e2e/network_segmentation.go
```diff
@@ -1533,12 +1532,12 @@ func generateLayer3Subnets(cidrs string) []string {
...
-	_, err := e2ekubectl.RunKubectl(namespace, "wait", "userdefinednetwork", name, "--for", "condition=NetworkReady=True", "--timeout", timeout.String())
+	_, err := e2ekubectl.RunKubectl(namespace, "wait", "userdefinednetwork", name, "--for", "condition=NetworkCreated=True", "--timeout", timeout.String())
 	return err
 }
 
 func waitForClusterUserDefinedNetworkReady(name string, timeout time.Duration) error {
```
tssurya, line 1539: same here

trozet:
dont we need downstream backwards compatibility tests already merged before we merge this or we cant do a downstream merge?

test/e2e/network_segmentation.go
ormergi: The intent is to have func name that describe CUDN is ready to work with regardless of which condition being asserted, make sense?

ormergi:
[Rebase](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/7d615a26f9065902618a6754bf2fb16d8c0ff020..97d9b67ab59632e3bcebc66049b400c2a096e370)

test/e2e/network_segmentation.go
tssurya: no it doesn't make much sense if we are going to have a networkReady status in future which we plan to have, then we need to know its not networkready but more of networkcreated right? or are you saying networkready func here will check created and ready in the future?

tssurya: I'd rather have the name of the function as networkcreated for now given ready is not even in the picture to not mislead as the network is read when someone reads the test code specially in origin, but again I don't really want to go back/forth on this `nit` and block this PR, so upto you whether you want to fix this or not.

tssurya (approved):
/lgtm

PR #4886, opened 2024-12-11
unit tests: speed up by small sleep before cache sync wait

poroh:
## 📑 Description

Introducing small delay before waiting for cache sync prevents 100ms wait on each
informer creation. There are up to 12 informers in factory that adds up to 1s wait on
each factory creation. 

On my machine introducing small delay reduces tests time from 15m44.683s to 13m56.218s.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -1659,6 +1659,8 @@ type waitForCacheSyncer interface {
 }
 
 func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) map[reflect.Type]bool {
+	// Give some small time for sync. It helps significantly reduce unit tests time
+	time.Sleep(5 * time.Millisecond)
 	return factory.WaitForCacheSync(util.GetChildStopChanWithTimeout(stopCh, types.InformerSyncTimeout))
 }
 


go-controller/pkg/factory/factory.go
```diff
@@ -1659,6 +1659,8 @@ type waitForCacheSyncer interface {
 }
 
 func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) map[reflect.Type]bool {
+	// Give some small for sync. It helps significantly reduce unit tests time
+	time.Sleep(5 * time.Millisecond)
```
npinaeva, line 1663: this sleep works for unit tests, but do you have some data from the real cluster? could you run a kind cluster with some extra logging around cache sync to check how it is affected?

poroh: This function is only called on start of the factory. So effect on real production cluster should not be anyhow affected by this change. If I'm wrong then we have another level of issue on startup.

Let me know if you have any specific concerns related to this change.

go-controller/Makefile
```diff
@@ -63,7 +63,7 @@ check test:
 ifeq ($(CONTAINER_RUNNABLE), 0)
 	$(CONTAINER_RUNTIME) run -it --rm --security-opt label=disable ${C_ARGS} -v $(shell dirname $(PWD)):/go/src/github.com/ovn-org/ovn-kubernetes -w /go/src/github.com/ovn-org/ovn-kubernetes/go-controller -e COVERALLS=${COVERALLS} -e GINKGO_FOCUS="${GINKGO_FOCUS}" $(GO_DOCKER_IMG) sh -c "RACE=1 DOCKER_TEST=1 COVERALLS=${COVERALLS} PKGS="${PKGS}" hack/test-go.sh focus \"${GINKGO_FOCUS}\" "
 else
-	RACE=1 hack/test-go.sh
+	RACE=1 hack/test-go.sh focus "${GINKGO_FOCUS}"
```
girishmg, line 66: looks like this fix is not relevant to what you have in the commit?

Can you please add this change in a separate commit capturing what this does and its use?

go-controller/pkg/factory/factory.go
```diff
@@ -1659,6 +1659,8 @@ type waitForCacheSyncer interface {
 }
 
 func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) map[reflect.Type]bool {
+	// Give some small for sync. It helps significantly reduce unit tests time
```
girishmg, line 1662: can you fix the comment here please?

Perhaps you meant... `Give some small time for sync`

go-controller/pkg/factory/factory.go
girishmg: I agree with @poroh here. Worst case it will add 5ms for every informer we are going to start when hybrid-overlay is enabled.

if this small change can cut down our unit tests run time, then that is a good win IMHO. Especially when we start this factory in 100s of tests.

npinaeva: let's at least document in the PR description that it may add 5 ms delay to each informer startup depending on the speed of its sync.

girishmg: @poroh can you please capture that in your commit description? thanks.

girishmg (approved):

PR #4890, opened 2024-12-11
Small iptables/nftables fixes

danwinship:
Three fixes:
- ~`Improve egress service repair unit test`: adds a third Service to the test and adjusts the initial incorrect state to make use of it, because there wouldn't be any way to make the initial state have multiple incorrect rules _for the same endpoint_ in the upcoming nftables version.~ (EDIT: This got merged as part of #4893)
- `Fix some nftables checking in the unit tests`: in half of the tests we were doing `err = nodenft.MatchNFTRules(...)` but then not checking `err` afterward. (This fix didn't turn up any new bugs; presumably if the generated rules were wrong before, the e2es would have failed and we would have noticed that way.) (EDIT: most of this got fixed as part of #4939.)
- `Slightly fix UDN local gateway masquerade rules (and unit tests)`: don't depend on undocumented behavior of the iptables binary, and tweak the unit test config so that the generated output actually demonstrates that the code is using the smaller subnet in one rule and the larger subnet in the other (rather than using a `/29` for both of them).

### DIFF
--- a/go-controller/pkg/node/gateway_init_linux_test.go
+++ b/go-controller/pkg/node/gateway_init_linux_test.go
@@ -942,6 +942,8 @@ func localGatewayInterfaceTest(app *cli.App, testNS ns.NetNS,
 	const mtu string = "1234"
 	const clusterCIDR string = "10.1.0.0/16"
 	config.Gateway.DisableForwarding = true
+	// Make this larger-than-default, so it makes sense for the UDN case
+	config.Gateway.V4MasqueradeSubnet = "169.254.169.0/24"
 
 	if len(eth0GWIP) > 0 {
 		// And a default route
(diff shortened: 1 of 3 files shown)


danwinship:
The only non-unit-test change here is to change the UDN masquerade rule from

    -A OVN-KUBE-UDN-MASQUERADE -s 169.254.169.2/29 -j RETURN

to

    -A OVN-KUBE-UDN-MASQUERADE -s 169.254.169.0/29 -j RETURN

So the non-network-segmentation e2e failures must be flakes. That iptables change should be a no-op in terms of actual functionality so I think the network-segmentation failure is a flake too.

tssurya (approved):
I reviewed the last 2 commits, thanks @danwinship they look correct to me
@ricky-rav I think you volunteered for this PR in team meeting? I will leave the 1st commit to you and once you are LGTMed I can merge this

tssurya:
the net-seg flake is a known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4909

ricky-rav (approved):
/LGTM

PR #4892, opened 2024-12-11
Refactor nftables add/delete/sync handling

danwinship:
Right now the nftables node code only supports syncing set/map elements, but it will eventually need to support syncing rules as well. (There is no way to implement the equivalent of `generateIPTRulesForLoadBalancersWithoutNodePorts` without having either a single chain containing a rule per service, or else a chain per service where each chain contains a single rule.) This moves us toward being able to support that, while also fixing things so the generic code supports the UDN map deletion case rather than that needing to be its own separate hack.

(As the existing code notes, if we had "nft destroy", we could do that a lot more simply, but that requires both a very new `nft` and a very new kernel (6.3+) so we probably don't want to try to require that.)

cc @kyrtapz since you touched this code recently for the UDN stuff


danwinship:
e2e failures are in code paths not touched by this change...

go-controller/pkg/node/nftables/util.go
```diff
@@ -5,39 +5,147 @@ package nftables
...
-		// they weren't already in the set/map, we won't get an error on delete.
-		tx.Add(elem)
-		tx.Delete(elem)
+		if len(elem.Key) != len(key) {
+			// All elements have the same key length, so if one fails, they all fail.
+			return nil
+		}
+		for i := range elem.Key {
+			if elem.Key[i] != key[i] {
+				continue elemLoop
```
martinkennelly, line 104: i am not following this. Why do you jump to back to the start again?

martinkennelly (commented):
looks good - just one question about the findElem func

go-controller/pkg/node/nftables/util.go
danwinship: "if this element doesn't have the correct key value, then continue on to the next element."

lines 102-107 could be rewritten as

    if reflect.DeepEqual(elem.Key, key) {
            return elem
    }

(except `reflect.DeepEqual` is inefficient)

martinkennelly: makes sense - brain fart remembering how `range` worked with this `continue` to `elemLoop`

martinkennelly (approved):

danwinship:
(rebased around the import renaming from the linter PR)

danwinship:
rebased around various changes, especially #5250

tssurya:
@martinkennelly this one too do you want to do a second stab?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1413,15 +1403,9 @@ func (npwipt *nodePortWatcherIptables) SyncServices(services []interface{}) erro
...
-		types.NFTMgmtPortNoSNATNodePorts,
-		types.NFTMgmtPortNoSNATServicesV4,
-		types.NFTMgmtPortNoSNATServicesV6,
-	}
-	for _, set := range nftableManagementPortSets {
-		if err = recreateNFTSet(set, keepNFTElems); err != nil {
-			errors = append(errors, err)
-		}
+	nftContainers = append(nftContainers, getGatewayNFTContainerObjects()...)
+	if err = nodenft.SyncObjects(context.TODO(), nftContainers, keepNFTObjects); err != nil {
```
martinkennelly, line 1372: unrelated to this change, but should we open an issue for correcting `context` to nodePortWatcher* ?

martinkennelly:
E2E failure is unrelated - we failed to find the `frr` docker created network.

martinkennelly:
@tssurya looks good

go-controller/pkg/node/gateway_shared_intf.go
danwinship: :man_shrugging:

danwinship:
ok, going to resubmit in smaller pieces...

PR #4893, opened 2024-12-11
Migrate EgressService rules from iptables to nftables

danwinship:
Switch Egress Services over to nftables. This code is mostly separate from all the other iptables code. (Actually... I'm pretty sure the existing reference to `egresservice.Chain` in `gateway_shared_intf.go:SyncServices` was incorrect, and would just cause it to flush all rules in `egressservice.Chain`. That only runs at startup, and I guess the ordering must have been such that it ran before the egressservice startup repair code, which I guess explains why #4868 wasn't a problem...)

(includes one commit from #4890, which can merge either here or there...)

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -17,7 +17,6 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/kube"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/controllers/egressservice"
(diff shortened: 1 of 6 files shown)


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1043,7 +1042,7 @@ func (npw *nodePortWatcher) SyncServices(services []interface{}) error {
 	// sync netfilter rules once only for Full mode
 	if !npw.dpuMode {
 		// (NOTE: Order is important, add jump to iptableETPChain before jump to NP/EIP chains)
-		for _, chain := range []string{iptableITPChain, egressservice.Chain, iptableNodePortChain, iptableExternalIPChain, iptableETPChain} {
+		for _, chain := range []string{iptableITPChain, iptableNodePortChain, iptableExternalIPChain, iptableETPChain} {
```
jcaamano, line 1045: We weren't actually creating any egress service rules here, were we? At least I couldn't trace any.

jcaamano (dismissed):

PR #4897, opened 2024-12-12
Introduce TEST_OVSDB_TXN_TIMEOUT environment variable.

yoks:
Setting it  reduces  test exections time where TNX_TIMEOUT is expected. For ci environments it can be set at higher value, for local development it can be safely lowered to 500ms.

## 📑 Description

Expands on work by @npinaeva #4505, i experimented with different values and found what i can safely set to `500ms` in local dev. This speed whole make check suite by ~2:30 (from 19:57.47 to 17:26.73). This is most noticable on suites in `ovn` folder.

This change is aimed to help hunt flakes in `egressip` tests as well as improve CI time.

PR also sets env variable for CI in workflow file to 1s. But this part is experimental and can be tuned.
 

Fixes #

## ✅ Checks

- [ ] My code requires tests

## How to verify it

Exectute
`time make check`
`TEST_OVSDB_TXN_TIMEOUT=500ms time make check`
Both should pass (except for known flakes). Test launch with reduced timeout should take less time.

### DIFF
--- a/go-controller/pkg/config/config.go
+++ b/go-controller/pkg/config/config.go
@@ -653,7 +653,13 @@ func PrepareTestConfig() error {
 	ClusterManager = savedClusterManager
 	Kubernetes.DisableRequestedChassis = false
 	EnableMulticast = false
-	Default.OVSDBTxnTimeout = 5 * time.Second
+
+	// OVSDBTimeout can be overriden for test, reduce to reduce test exection time
+	// increase to reduce flakiness, if unset or parsing failed, defaults to 5s
+	Default.OVSDBTxnTimeout, _ = time.ParseDuration(os.Getenv("TEST_OVSDB_TXN_TIMEOUT"))
+	if Default.OVSDBTxnTimeout <= 0 {
+		Default.OVSDBTxnTimeout = 5 * time.Second
+	}
 
 	if err := completeConfig(); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


go-controller/pkg/config/config.go
```diff
@@ -653,7 +653,13 @@ func PrepareTestConfig() error {
 	ClusterManager = savedClusterManager
 	Kubernetes.DisableRequestedChassis = false
 	EnableMulticast = false
-	Default.OVSDBTxnTimeout = 5 * time.Second
+
+	// OVSDBTimeout can be overriden for test, reduce to reduce test exection time
+	// increase to reduce flakiness, if unset or parsing failed, defaults to 5s
+	Default.OVSDBTxnTimeout, _ = time.ParseDuration(os.Getenv("TEST_OVSDB_TXN_TIMEOUT"))
+	if Default.OVSDBTxnTimeout <= 0 {
+		Default.OVSDBTxnTimeout = 5 * time.Second
```
npinaeva, line 661: why don't you just set it to 1s here without extra env variables?

yoks: My first reaction was just set it, right. But this timeout is used in different tests, and it used to verify asynchronous behavior. I thought it can be tuned for runners as tuned are, usually, not as powerful as local dev env.

But new env variable for sure creates configuration overhead for tests. But as it only used it tests so I thought it would be ok. 

So if you feel we need just lower the value I will revert env variable changes and will just reduce value to sensible default (1s seems to work on GitHub runner).

yoks:
Env variable scrapped, change will be made in #4910

PR #4902, opened 2024-12-12
UDN Gateway: Ignore EndpointSlices without a service label

kyrtapz:
Instead of failing ignore EndpointSlices without a service label set.
These custom, user created slices should be ignored.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1084,7 +1084,7 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) e
 	}
 
 	svcNamespacedName, err := util.ServiceFromEndpointSlice(epSlice, netInfo)
-	if err != nil {
+	if err != nil || svcNamespacedName == nil {
 		return err
 	}
 
@@ -1118,7 +1118,7 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) e
 	// Here we make sure the correct rules are programmed whenever an AddEndpointSlice event is
 	// received, only alter flows if we need to, i.e if cache wasn't set or if it was and
 	// hasLocalHostNetworkEp or localEndpoints state (for LB svc where NPs=0) changed, to prevent flow churn
-	out, exists := npw.getAndSetServiceInfo(svcNamespacedName, svc, hasLocalHostNetworkEp, localEndpoints)
+	out, exists := npw.getAndSetServiceInfo(*svcNamespacedName, svc, hasLocalHostNetworkEp, localEndpoints)
 	if !exists {
 		klog.V(5).Infof("Endpointslice %s ADD event in namespace %s is creating rules", epSlice.Name, epSlice.Namespace)
 		return addServiceRules(svc, netInfo, sets.List(localEndpoints), hasLocalHostNetworkEp, npw)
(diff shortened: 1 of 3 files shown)


kyrtapz:
cc @tssurya

kyrtapz:
The LGW service flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4829
I will need to look into that separately

go-controller/pkg/util/util.go
```diff
@@ -594,21 +594,24 @@ func IsUDNEnabledService(key string) bool {
...
 	}
-	svcName := eps.Labels[labelKey]
+	svcName, found := eps.Labels[labelKey]
+	if !found {
+		return nil, nil
+	}
 
 	if svcName == "" {
-		return k8stypes.NamespacedName{}, fmt.Errorf("endpointslice %s/%s: empty value for label %s in network %s",
+		return nil, fmt.Errorf("endpointslice %s/%s: empty value for label %s in network %s",
```
tssurya, line 612:
```suggestion
		return nil, fmt.Errorf("endpointslice %s/%s has empty svcName for label %s in network %s",
```

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1154,7 +1154,7 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 	klog.V(5).Infof("Deleting endpointslice %s in namespace %s", epSlice.Name, epSlice.Namespace)
 	// remove rules for endpoints and add back normal ones
 	namespacedName, err := util.ServiceFromEndpointSlice(epSlice, netInfo)
-	if err != nil {
+	if err != nil || namespacedName == nil {
```
tssurya, line 1157: Is having an empty serviceName on label key an error for us? Or we should ignore the processing for such cases with a warning..and return nil ? meaning the upstream test that causes this I wonder what happens there if we start returning an error

go-controller/pkg/util/util.go
```diff
@@ -594,21 +594,24 @@ func IsUDNEnabledService(key string) bool {
 }
 
 // ServiceFromEndpointSlice returns the namespaced name of the service that corresponds to the given endpointSlice
-// in the given network
-func ServiceFromEndpointSlice(eps *discovery.EndpointSlice, netInfo NetInfo) (k8stypes.NamespacedName, error) {
+// in the given network. If the service label is missing the returned namespaced name and the error are nil.
+func ServiceFromEndpointSlice(eps *discovery.EndpointSlice, netInfo NetInfo) (*k8stypes.NamespacedName, error) {
 	labelKey := discovery.LabelServiceName
 	if netInfo.IsPrimaryNetwork() {
 		if eps.Labels[types.LabelUserDefinedEndpointSliceNetwork] != netInfo.GetNetworkName() {
```
tssurya, line 601: will this UDN network label always exist? I guess we don't care if its empty cause then it won't match ? IsEndpointSliceForNetwork util seems more accurate in the sense it handles the empty being assuming its default network...

go-controller/pkg/util/util_unit_test.go
```diff
@@ -250,3 +258,125 @@ func TestGetNetworkScopedK8sMgmtHostIntfName(t *testing.T) {
...
+		},
+		{
+			name: "Primary network with no service label set",
+			args: args{
+				eps: &discovery.EndpointSlice{
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "test-namespace",
+						Name:      "test-eps",
+						Labels: map[string]string{
+							types.LabelUserDefinedEndpointSliceNetwork: "primary-network",
```
tssurya, line 332: ++

tssurya (commented):
nice tests..

go-controller/pkg/util/util.go
tssurya: I see you do:
```
if util.IsNetworkSegmentationSupportEnabled() && !util.IsEndpointSliceForNetwork(newEpSlice, netInfo) {
		return nil
	}
```
before calling this util but yea maybe the util itself also needs to be foolproof to avoid putting burden on the caller?

kyrtapz: Since `ServiceFromEndpointSlice` takes the network as a parameter I think it should restrict the processing to this network and not return services that do not belong to it.

go-controller/pkg/node/gateway_shared_intf.go
tssurya: what I meant was shouldn't we return nil here if namesacedName is nil? why do we return error?

tssurya: oh shit :) bad semantics,, from the actual func we return nil and so here err=nil silly me

tssurya (approved):

PR #4905, opened 2024-12-13
Fixes pkt larger dispatch flows for UDNs

trozet:
Looks like this path was missed and causes egress IP with UDNs on local gateway mode to not function correctly. Reply traffic was being sent to default cluster network patch port rather than the correct UDN patch port.

@martinkennelly FYI


trozet:
fixes lgw UDN egress IP issue seen in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4530

martinkennelly:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4829 

^ that flake is getting persistent.

trozet:
closed in favor of #4907

PR #4912, opened 2024-12-13
Enforce udn ns label

trozet:
Require namespace label for primary UDN
k8s.ovn.org/primary-user-defined-network is now required to be labeled on a
namespace at namespace creation time in order to use a primary UDN. 
The label may not be updated and may only be added at creation time.

The following conditions are true:

1. If namespace is missing the label, and a pod is created, it attaches
   to default network.
2. If the namespace is missing the label, and a primary UDN or CUDN is
   created that matches that namespace, the UDN/CUDN will report error
   status and the NAD will not be generated.
3. If the namespace is missing the label, and a primary UDN/CUDN exists,
   a pod in the namespace will be created and attached to default
   network.
4. If the namespace has the label, and a primary UDN/CUDN does not exist
   a pod in the namespace will fail creation until the UDN/CUDN is
   created.

Also includes some fixes to unit tests that were brought to light by
this PR. For example, the layer 2 multi-network tests were adding
invalid annotations for node-subnets, etc.

Related OCP jira:
https://issues.redhat.com/browse/OCPBUGS-42609


trozet:
Note: e2es will fail because they have not been updated to add the annotation, but that's fine for now to see the guard work as we expect

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -15,10 +15,23 @@ import (
...
+	// check if required UDN label is on namespace
+	ns, err := c.namespaceInformer.Lister().Get(namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get namespace %q: %w", namespace, err)
+	}
+
+	if _, exists := ns.Labels[types.RequiredUDNNamespaceLabel]; !exists {
+		// UDN required set on namespace, but no NAD/UDN exists, invalid state
+		return nil, util.NewInvalidActiveNetworkError(namespace)
+	}
```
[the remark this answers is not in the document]
trozet: sorry the comment is wrong. the code is really saying there is no label on the namespace, so we refuse to render the NAD.

test/e2e/network_segmentation.go
```diff
@@ -32,9 +32,12 @@ import (
 )
 
 const openDefaultPortsAnnotation = "k8s.ovn.org/open-default-ports"
+const RequiredUDNNamespaceLabel = "k8s.ovn.org/user-defined-network"
 
 var _ = Describe("Network Segmentation", func() {
 	f := wrappedTestFramework("network-segmentation")
+	// disable automatic namespace creation, we need to add the required UDN annotation
```
trozet, line 39: nit: should say label

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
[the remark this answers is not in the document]
maiqueb: Ah, right ! That I understand. 

Makes sense.

go-controller/pkg/util/util.go
```diff
@@ -390,6 +390,20 @@ func NewUnprocessedActiveNetworkError(namespace, udnName string) *UnprocessedAct
 	return &UnprocessedActiveNetworkError{namespace: namespace, udnName: udnName}
 }
 
+type InvalidActiveNetworkError struct {
```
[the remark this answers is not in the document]
trozet: I think at the time I wrote the original code too we were trying to minimize the race between UDN CRDs and when the thing was actually available in ovnk. So Active also implied that the nad controller had seen it and started the secondary controller.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -15,10 +15,23 @@ import (
...
 )
 
 func (c *Controller) updateNAD(obj client.Object, namespace string) (*netv1.NetworkAttachmentDefinition, error) {
+	// check if required UDN label is on namespace
+	ns, err := c.namespaceInformer.Lister().Get(namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get namespace %q: %w", namespace, err)
+	}
+
+	if _, exists := ns.Labels[types.RequiredUDNNamespaceLabel]; !exists {
```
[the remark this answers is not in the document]
trozet: good catch, will fix

go-controller/pkg/types/const.go
```diff
@@ -209,6 +209,8 @@ const (
 	LoadBalancerOwnerExternalID = OvnK8sPrefix + "/" + "owner"
 	// key for UDN enabled services routes
 	UDNEnabledServiceExternalID = OvnK8sPrefix + "/" + "udn-enabled-default-service"
+	// required namespace label for enabling UDNs
+	RequiredUDNNamespaceLabel = "k8s.ovn.org/user-defined-network"
```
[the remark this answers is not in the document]
trozet: what do you suggest? @kyrtapz wdyt?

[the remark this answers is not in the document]
kyrtapz: I am fine with adding the primary for clarity, maybe `k8s.ovn.org/user-defined-primary-network`?

go-controller/pkg/util/util.go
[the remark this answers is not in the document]
trozet: changed it

go-controller/pkg/types/const.go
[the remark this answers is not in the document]
ormergi: nit: I would call it `primary-user-defined-network` for better consistently (we usually put primary at the start in code and docs).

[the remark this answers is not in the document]
maiqueb: FWIW, I am also more "used to" seeing primary UDN ...

test/e2e/network_segmentation.go
```diff
@@ -32,9 +32,12 @@ import (
 )
 
 const openDefaultPortsAnnotation = "k8s.ovn.org/open-default-ports"
+const RequiredUDNNamespaceLabel = "k8s.ovn.org/user-defined-primary-network"
 
 var _ = Describe("Network Segmentation", func() {
 	f := wrappedTestFramework("network-segmentation")
+	// disable automatic namespace creation, we need to add the required UDN label
+	f.SkipNamespaceCreation = true
```
ormergi, line 40: Commenting here but it related to other places as well, can we do the same as in egressip, label the namespace we got from the infra?

test/e2e/network_segmentation.go
```diff
@@ -223,7 +232,8 @@ var _ = Describe("Network Segmentation", func() {
 						defaultNetNamespace := f.Namespace.Name + "-default"
 						_, err := cs.CoreV1().Namespaces().Create(context.Background(), &v1.Namespace{
 							ObjectMeta: metav1.ObjectMeta{
-								Name: defaultNetNamespace,
+								Name:   defaultNetNamespace,
+								Labels: map[string]string{RequiredUDNNamespaceLabel: ""},
```
ormergi, line 236: Commenting here but it related to other places as well.
Please consider introducing a helper to create the test namespace, it will reduce the noise in all places that create the namespace, and make it easier to introduce additional changes to such namespaces.
``` 
_, err := cs.CoreV1().Namespaces().Create(context.Background(), newTestNamespace(defaultNetNamespace))...
...
func newTestNamesapce(name string) *Namespace {
  return  &v1.Namespace{ ObjectMeta: metav1.ObjectMeta{
    Name: name,
    Labels: map[string]string{RequiredUDNNamespaceLabel: ""},
}}
```

test/e2e/network_segmentation.go
```diff
@@ -1031,7 +1045,11 @@ spec:
 			assertClusterUDNStatusReportsActiveNamespaces(testClusterUdnName, testTenantNamespaces...)
 
 			By("create the new target namespace")
-			_, err = cs.CoreV1().Namespaces().Create(context.Background(), &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNewNs}}, metav1.CreateOptions{})
+			_, err = cs.CoreV1().Namespaces().Create(context.Background(), &v1.Namespace{
```
ormergi, line 1048: It seem this test create CUDN with role=secondary at the before-each, does it fail w/o this change?

test/e2e/network_segmentation.go
```diff
@@ -1054,7 +1072,11 @@ spec:
 				testNewNs := f.Namespace.Name + "green"
 
 				By("create new namespace")
-				_, err := cs.CoreV1().Namespaces().Create(context.Background(), &v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: testNewNs}}, metav1.CreateOptions{})
+				_, err := cs.CoreV1().Namespaces().Create(context.Background(), &v1.Namespace{
```
ormergi, line 1075: It seem this test create CUDN with role=secondary at the before-each, does it fail w/o this change?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -1334,6 +1386,18 @@ func testNAD() *netv1.NetworkAttachmentDefinition {
 	}
 }
 
+func invalidTestNamespace(name string) *corev1.Namespace {
```
ormergi, line 1389: super nit: I would define this function to use  `testNamespace` instead of the other way around.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -76,7 +77,7 @@ var _ = Describe("User Defined Network Controller", func() {
 			It("should create NAD successfully", func() {
 				udn := testUDN()
 				expectedNAD := testNAD()
-				c = newTestController(renderNadStub(expectedNAD), udn)
+				c = newTestController(renderNadStub(expectedNAD), udn, testNamespace("test"))
```
ormergi, line 80: Question: are these tests fail if we dont pass the test-namespace object?

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -15,10 +15,25 @@ import (
 	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/clustermanager/userdefinednetwork/template"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 	utiludn "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/udn"
 )
 
 func (c *Controller) updateNAD(obj client.Object, namespace string) (*netv1.NetworkAttachmentDefinition, error) {
+	if utiludn.IsPrimaryNetwork(template.GetSpec(obj)) {
```
ormergi, line 24: Does it make sense to move the label check under the existing primary-network spec check? (line 54)

ormergi (commented):
Thanks of the PR, Please see my inline comments

go-controller/pkg/types/const.go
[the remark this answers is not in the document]
trozet: yeah will change it

test/e2e/network_segmentation.go
ormergi: Not a must, can be done in follow-up PR

test/e2e/network_segmentation.go
```diff
@@ -59,6 +62,12 @@ var _ = Describe("Network Segmentation", func() {
 		var err error
 		nadClient, err = nadclient.NewForConfig(f.ClientConfig())
 		Expect(err).NotTo(HaveOccurred())
+		namespace, err := f.CreateNamespace(context.TODO(), f.BaseName, map[string]string{
+			"e2e-framework":           f.BaseName,
+			RequiredUDNNamespaceLabel: "",
+		})
+		f.Namespace = namespace
```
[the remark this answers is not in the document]
trozet: yeah it does:
```
func (f *Framework) CreateNamespace(ctx context.Context, baseName string, labels map[string]string) (*v1.Namespace, error) {
	createTestingNS := TestContext.CreateTestingNS
	if createTestingNS == nil {
		createTestingNS = CreateTestingNS
	}

	if labels == nil {
		labels = make(map[string]string)
	} else {
		labelsCopy := make(map[string]string)
		for k, v := range labels {
			labelsCopy[k] = v
		}
		labels = labelsCopy
	}

	labels[admissionapi.EnforceLevelLabel] = firstNonEmptyPSaLevelOrRestricted(f.NamespacePodSecurityEnforceLevel, f.NamespacePodSecurityLevel)
	labels[admissionapi.WarnLevelLabel] = firstNonEmptyPSaLevelOrRestricted(f.NamespacePodSecurityWarnLevel, f.NamespacePodSecurityLevel)
	labels[admissionapi.AuditLevelLabel] = firstNonEmptyPSaLevelOrRestricted(f.NamespacePodSecurityAuditLevel, f.NamespacePodSecurityLevel)

	ns, err := createTestingNS(ctx, baseName, f.ClientSet, labels)
	// check ns instead of err to see if it's nil as we may
	// fail to create serviceAccount in it.
	f.AddNamespacesToDelete(ns)  <--------############### here
```

ormergi:
I have some suggestion around tests, but they can be done on a follow-up PR, sorry for the noise.

I had a question about the label check on the UDN controllers but I think I got it.

LGTM.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
maiqueb: I'm *assuming* we just want to bail out from this asap ...

maiqueb (approved):
From my perspective, this is good. 

Thanks for the prompt fix to the nasty race we had.

test/e2e/network_segmentation.go
trozet: cant, because the label must exist when the namespace is created. The validating admission policy wont let you add it later.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
trozet: in almost all cases yes cause it needs to check the namespace, kidn of makes sense for the test with NAD to also include a namespace though

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -158,3 +158,35 @@ rules:
     - apiGroups: [""]
       resources: ["configmaps"]
       verbs: ["create", "patch", "update"]
+
+---
+apiVersion: admissionregistration.k8s.io/v1
```
kyrtapz, line 163: @trozet just as a reminder, this will require a separate PR downstream(CNO).

trozet: i also need to get the e2es backward compatible downstream before we merge this

go-controller/pkg/ovn/egressip.go
```diff
@@ -865,12 +866,21 @@ func (e *EgressIPController) deleteEgressIPAssignments(name string, statusesToRe
...
 			podNamespace, podName := getPodNamespaceAndNameFromKey(podKey)
-			ni, err := e.networkManager.GetActiveNetworkForNamespace(podNamespace)
-			if err != nil {
-				return fmt.Errorf("failed to get active network for namespace %s", podNamespace)
-			}
 			cachedNetwork := e.getNetworkFromPodAssignment(podKey)
-			err = e.nodeZoneState.DoWithLock(statusToRemove.Node, func(key string) error {
+			var ni util.NetInfo
+			if _, err := e.watchFactory.GetNamespace(podNamespace); err == nil {
+				ni, err = e.networkManager.GetActiveNetworkForNamespace(podNamespace)
```
trozet, line 872: @martinkennelly can we just drop trying to get active network for namespace all together and just rely on the cache?

go-controller/pkg/ovn/egressip.go
```diff
@@ -865,12 +866,21 @@ func (e *EgressIPController) deleteEgressIPAssignments(name string, statusesToRe
...
+				if err != nil {
+					return fmt.Errorf("failed to get active network for namespace %s", podNamespace)
```
martinkennelly, line 874: include the error

trozet:
hit https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4921

martinkennelly (approved):

go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
tssurya, line 1: what if user creates:
1. A CUDN first
2. then the namespaces
3. namespace is labelled
4. then pod is created
I assume pod will be attached to the UDN

what if user creates:
1. A CUDN first
2. then the namespaces
3. namespace is NOT labelled
4. then pod is created
I assume pod will be attached to the default network -> Is this the case3 here? '

Overall my head knows the logic is "User needs to create the UDN/CUDN CR + label their namespaces" for this to work, so its making sense when only one trigger exists and other doesn't (cases 2 and 4) we strongly fail and report error to the user. What is confusing is case3 - I would expect it to behave like:
what if user creates:
1. A CUDN first
2. then the namespaces
3. namespace is NOT labelled
4. then pod is created
we should fail here telling the user they have not labelled their namespace right?  I find it weird we attach to the default network, its breaking the semantics a bit?? or am I overthinking this. Like it makes it look like user is specifically asking for the UDN/CUDN that's the whole point of creating the CRs and then they forgot to label the namespace and now we have made their workloads attach to the default network... the only way out is to delete and recreate this right? Isn't that back to the initial race problem where don't want to do that?

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -1353,14 +1470,9 @@ func testNADWithDeletionTimestamp(ts time.Time) *netv1.NetworkAttachmentDefiniti
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
-			Labels: map[string]string{
-				"kubernetes.io/metadata.name": name,
-			},
-		},
-	}
+	ns := invalidTestNamespace(name)
+	ns.ObjectMeta.Labels[ovntypes.RequiredUDNNamespaceLabel] = ""
+	return ns
```
tssurya, line 1475: nit: I don't see a test for case3 where pod gets attached to default network.. maybe its in other files.. but good to track that behaviour somewhere...

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -413,11 +413,23 @@ func (c *nadController) GetActiveNetworkForNamespace(namespace string) (util.Net
 	if !util.IsNetworkSegmentationSupportEnabled() {
 		return &util.DefaultNetInfo{}, nil
 	}
+
+	// check if required UDN label is on namespace
+	ns, err := c.namespaceLister.Get(namespace)
```
tssurya, line 418: nit: I think this snippet is coming here a second time after I saw it in cluster-manager, time for a new util called "ValidUDNNamespace" or something which get's the namespace and checks for the label, might be useful for the future

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -413,11 +413,23 @@ func (c *nadController) GetActiveNetworkForNamespace(namespace string) (util.Net
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get namespace %q: %w", namespace, err)
+	}
+	if _, exists := ns.Labels[types.RequiredUDNNamespaceLabel]; !exists {
+		// UDN required label not set on namespace, assume default network
+		return &util.DefaultNetInfo{}, nil
```
tssurya, line 424: is it always correct to assume not having the label means default net? because semantically speaking if label is missing and CRs are present there are cases where we block pod creation and don't make the pod part of default network right? do we need an unknown state or at least in how we use GetActiveNamespace we are OK ?
I guess it could be that I am confusing what it means for "a pod to be part of UDN namespace" v/s "when a namespace is considered for udn or not"

tssurya: what I mean is not having a NAD + not having a UDN + not having CUDN + not having an annotation is what should be considered as "defaultnet" right?

go-controller/pkg/ovn/secondary_layer2_network_controller_test.go
```diff
@@ -405,15 +407,12 @@ var _ = Describe("OVN Multi-Homed pod operations for layer2 network", func() {
...
 
-				Expect(fakeOvn.controller.WatchNamespaces()).To(Succeed())
-				Expect(fakeOvn.controller.WatchPods()).To(Succeed())
+				Expect(fakeOvn.networkManager.Start()).To(Succeed())
 				secondaryNetController, ok := fakeOvn.secondaryControllers[secondaryNetworkName]
 				Expect(ok).To(BeTrue())
 
 				secondaryNetController.bnc.ovnClusterLRPToJoinIfAddrs = dummyJoinIPs()
 				podInfo.populateSecondaryNetworkLogicalSwitchCache(fakeOvn, secondaryNetController)
-				Expect(secondaryNetController.bnc.WatchNodes()).To(Succeed())
```
tssurya, line 415: lol watching nodes and pods was not wanted at all? how is the delete pod working on L417 immediately after this?

go-controller/pkg/ovn/secondary_layer2_network_controller_test.go
```diff
@@ -405,15 +407,12 @@ var _ = Describe("OVN Multi-Homed pod operations for layer2 network", func() {
 					Expect(ok).To(BeFalse())
 				}
 
-				Expect(fakeOvn.controller.WatchNamespaces()).To(Succeed())
-				Expect(fakeOvn.controller.WatchPods()).To(Succeed())
+				Expect(fakeOvn.networkManager.Start()).To(Succeed())
```
tssurya, line 410: defer a stop of the networkManager to prevent data races?

tssurya (commented):
commit1 reviewed

dist/templates/rbac-ovnkube-master.yaml.j2
tssurya, line 1: also update of the label is not allowed right? basically once the namespace is created they cannot do anything with this label correct?

can k8s.ovn.org/primary-user-defined-network become k8s.ovn.org/primary-user-defined-network-eeeee ?

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -156,3 +156,35 @@ rules:
...
+  matchConstraints:
+    resourceRules:
+      - apiGroups:   [""]
+        apiVersions: ["v1"]
+        operations:  ["UPDATE"]
+        resources:   ["namespaces"]
+  failurePolicy: Fail
+  validations:
+    - expression: "('k8s.ovn.org/primary-user-defined-network' in oldObject.metadata.labels) == ('k8s.ovn.org/primary-user-defined-network' in object.metadata.labels)"
+      message: "The 'k8s.ovn.org/primary-user-defined-network' label cannot be added/removed after the namespace was created"
```
tssurya, line 175: nit:
message: "The 'k8s.ovn.org/primary-user-defined-network' label cannot be added/updated/removed after the namespace was created"

dist/templates/rbac-ovnkube-master.yaml.j2
tssurya, line 1: commit2 reviewed, are there any tests to cover this RBAC change? a simple test would be nice to have

test/e2e/e2e.go
```diff
@@ -469,8 +469,15 @@ func deleteClusterExternalContainer(containerName string) {
 	}, 5).Should(gomega.HaveLen(0))
 }
 
-func updateNamespace(f *framework.Framework, namespace *v1.Namespace) {
```
tssurya, line 472: `updateNamespace` was used only from EIP tests from what I can tell,
can we leave the older func as is for nonUDN tests if needed ?
we can add your new util and invoke that only from UDN tests/
I assume not all EIP tests need to use the UDN label....

test/e2e/network_segmentation.go
```diff
@@ -33,9 +33,12 @@ import (
 )
 
 const openDefaultPortsAnnotation = "k8s.ovn.org/open-default-ports"
+const RequiredUDNNamespaceLabel = "k8s.ovn.org/primary-user-defined-network"
 
 var _ = Describe("Network Segmentation", func() {
 	f := wrappedTestFramework("network-segmentation")
+	// disable automatic namespace creation, we need to add the required UDN label
+	f.SkipNamespaceCreation = true
```
tssurya, line 41: ohh nice
self-reminder: this is going to screw up my admin network policy e2e tests which come from upstream k8s-sig
@trozet given we can't have external entities create namespaces anymore, is there a way for us to actually port the upstream k8s services and netpol tests like we planned to run them all on UDNs? I assume not...

test/e2e/network_segmentation_endpointslices_mirror.go
```diff
@@ -174,17 +179,26 @@ var _ = Describe("Network Segmentation EndpointSlices mirroring", func() {
 					func(
 						netConfig networkAttachmentConfigParams,
 					) {
+						By("creating default net namespace")
+						defaultNetNamespace := &v1.Namespace{
```
tssurya, line 183: nit: maybe this can be a util now.. :D "createDefaultNetNamespace"

tssurya (commented):
commit3 lgtm looks like you had to fix almost every UDN tests wow, nice job - in the end this PR just became a lot of work huh

test/e2e/network_segmentation.go
```diff
@@ -736,13 +736,13 @@ var _ = Describe("Network Segmentation", func() {
 				ginkgo.Entry("with primary layer3 UDN", networkAttachmentConfigParams{
 					name:     nadName,
 					topology: "layer3",
-					cidr:     fmt.Sprintf("%s,%s", userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
+					cidr:     correctCIDRFamily(userDefinedNetworkIPv4Subnet, userDefinedNetworkIPv6Subnet),
```
tssurya, line 739: ++

go-controller/pkg/ovn/egressip.go
```diff
@@ -865,27 +866,27 @@ func (e *EgressIPController) deleteEgressIPAssignments(name string, statusesToRe
 				continue
 			}
 			podNamespace, podName := getPodNamespaceAndNameFromKey(podKey)
-			ni, err := e.networkManager.GetActiveNetworkForNamespace(podNamespace)
-			if err != nil {
-				return fmt.Errorf("failed to get active network for namespace %s", podNamespace)
-			}
 			cachedNetwork := e.getNetworkFromPodAssignment(podKey)
-			err = e.nodeZoneState.DoWithLock(statusToRemove.Node, func(key string) error {
+			if cachedNetwork == nil {
```
tssurya, line 870: hmm for deletion of EIP status we rely on the NAD being present in that namespace that determines the active network? that does sound wrong cause the NAD / namespace could just be gone i guess in order of events we don't control...
the fix looks correct to me but the fact that not a single unit test had to be changed worries me :D 
@martinkennelly can we track this somewhere in TODOs as we need to add a test for the sequence of events of deletion if a NAD doesn't exist and ensure nothing breaks... 

meanwhile @trozet why was GetActiveNetworkForNamespace not working in this specific instances? I assume this is some CI issue you happened to fix in this PR right?

go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
trozet: > what if user creates:
> 
> 1. A CUDN first
> 2. then the namespaces
> 3. namespace is labelled
> 4. then pod is created
>    I assume pod will be attached to the UDN

Cannot label the namespace after it is created. The admission controller stops that. CUDN will not render a NAD for this namespace because it has no label, causing an error in the CUDN. Pods created for this namespace will be attached to default cluster network.

> what if user creates:
> 
> 1. A CUDN first
> 2. then the namespaces
> 3. namespace is NOT labelled
> 4. then pod is created
>    I assume pod will be attached to the default network -> Is this the case3 here? '
> 

Same result as the first example.

> Overall my head knows the logic is "User needs to create the UDN/CUDN CR + label their namespaces" for this to work, so its making sense when only one trigger exists and other doesn't (cases 2 and 4) we strongly fail and report error to the user. What is confusing is case3 - I would expect it to behave like: what if user creates:
> 
> 1. A CUDN first
> 2. then the namespaces
> 3. namespace is NOT labelled
> 4. then pod is created
>    we should fail here telling the user they have not labelled their namespace right?  I find it weird we attach to the default network, its breaking the semantics a bit?? or am I overthinking this. Like it makes it look like user is specifically asking for the UDN/CUDN that's the whole point of creating the CRs and then they forgot to label the namespace and now we have made their workloads attach to the default network... the only way out is to delete and recreate this right? Isn't that back to the initial race problem where don't want to do that?

The failure is tracked in the CUDN CR. It does not block the pods from coming up and attaching to the default network. That was what we agreed upon was the "right" behavior. Since whomever created the namespace didn't put the required label on it, we assume that means it wants DNC. I guess our position is the admin should have enforced the label on the NS that it selected with CUDN.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -96,10 +97,53 @@ var _ = Describe("User Defined Network Controller", func() {
...
+					Status:  "False",
+					Reason:  "SyncError",
+					Message: "invalid primary network state for namespace \"test\": a valid primary user defined network or network attachment definition custom resource, and required namespace label \"k8s.ovn.org/primary-user-defined-network\" must both be present",
+				}}))
+
+				_, err := cs.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(udn.Namespace).Get(context.Background(), udn.Name, metav1.GetOptions{})
+				Expect(kerrors.IsNotFound(err)).To(BeTrue())
+			})
+
+			It("should NOT fail when required namespace label is missing for secondary network", func() {
```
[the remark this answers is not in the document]
trozet: secondary UDN doesn't have this problem because secondary network is specified during pod creation. So the pod explicitly tells us what secondary network it wants to attach to. This is different from primary UDN where this information is not specified by the pod, and is decided based on a combination of the required label, and the NAD existing.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -324,14 +368,14 @@ var _ = Describe("User Defined Network Controller", func() {
 			It("when UDN is being deleted, NAD exist, 2 pods using UDN, should delete NAD once no pod uses the network", func() {
 				var err error
 				nad := testNAD()
-				udn := testUDN()
+				udn := testPrimaryUDN()
```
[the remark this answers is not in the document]
trozet: yeah i was just making it explicit so easier to read

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
trozet: these tests are just in clustermanager, so there is no test for pods here. I'll add a test in the  ovn/secondary_layer2/3 files to cover a UDN exists, but namespace not labeled, and see pod is attached to default network.

go-controller/pkg/networkmanager/nad_controller.go
trozet: I think it's ok for now. The logic is pretty short and the decision made on the logic is different in each case.

go-controller/pkg/networkmanager/nad_controller.go
trozet: "is it always correct to assume not having the label means default net?"

that's what we decided to go with. It means default net for the pod, but the UDN will report error.

go-controller/pkg/ovn/secondary_layer2_network_controller_test.go
trozet: good catch :)

go-controller/pkg/ovn/secondary_layer2_network_controller_test.go
trozet: I don't really remember what happened here, but I think it is because now that I start networkManager, I thought it would take care of starting the watchers and programming everything for the secondary network. I see:

```
I0110 12:14:50.729229 1319489 factory.go:453] Starting watch factory
I0110 12:14:50.745434 1319489 controller.go:132] Adding controller [zone-nad-controller NAD controller] event handlers
I0110 12:14:50.745460 1319489 shared_informer.go:313] Waiting for caches to sync for [zone-nad-controller NAD controller]
I0110 12:14:50.745463 1319489 shared_informer.go:320] Caches are synced for [zone-nad-controller NAD controller]
I0110 12:14:50.745512 1319489 controller.go:156] Starting controller [zone-nad-controller NAD controller] with 1 workers
I0110 12:14:50.745520 1319489 network_controller.go:244] [zone-nad-controller network controller]: syncing all networks
I0110 12:14:50.745525 1319489 network_controller.go:255] [zone-nad-controller network controller]: finished syncing all networks. Time taken: 5.129µs
I0110 12:14:50.745529 1319489 controller.go:156] Starting controller [zone-nad-controller network controller] with 1 workers
I0110 12:14:50.745532 1319489 nad_controller.go:161] [zone-nad-controller NAD controller]: started
I0110 12:14:50.745543 1319489 nad_controller.go:166] [zone-nad-controller NAD controller]: shutting down
```

I don't see the secondary network controller starting though. I would have thought during initial sync it would have seen the NAD and blocked until it started that secondary network controller. @jcaamano any ideas?

jcaamano: The networkManager in `fakeOvn` uses a `FakeControllerManager` that is just a stub.

All along the tests have been just calling the `Watch` methods instead of starting controllers.

go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
tssurya: hmm I see will ping you a bit on Monday to talk about this

go-controller/pkg/networkmanager/nad_controller.go
tssurya: hmm I see will ping you a bit on Monday to talk about this

go-controller/pkg/clustermanager/endpointslicemirror/endpointslice_mirror_controller_test.go
tssurya: > The failure is tracked in the CUDN CR. It does not block the pods from coming up and attaching to the default network. That was what we agreed upon was the "right" behavior. Since whomever created the namespace didn't put the required label on it, we assume that means it wants DNC. I guess our position is the admin should have enforced the label on the NS that it selected with CUDN.

I think that's the confusing part, the admin also created the CR right which indicates they wanted the UDN/CUDN and they forgot to label the namespace..
what I mean is if we decided the right behaviour upon missing annotation and presence of CR is to attach the pod to the default network why is the reverse not true? where annotation is present but missing CR :) ?

dist/templates/rbac-ovnkube-master.yaml.j2
kyrtapz: > also update of the label is not allowed right? basically once the namespace is created they cannot do anything with this label correct?

This check only ensures that the existence of the label doesn't change. We are not using the label values and we are not checking for them here. We don't care about the label value.

> can k8s.ovn.org/primary-user-defined-network become k8s.ovn.org/primary-user-defined-network-eeeee ?

no, this would mean that the `k8s.ovn.org/primary-user-defined-network` was removed which is not allowed.

go-controller/pkg/ovn/secondary_layer2_network_controller_test.go
trozet: Thanks @jcaamano. We should probably fix this in the future. That way we can have more robust testing with network manager integrated in to our tests and make sure things start and stop correctly. I was confused by the renaming so im putting in a note here with my definitions now so I can refer to it later:

NetworkManager == old nad controller
ControllerManager == called within network manager to launch "secondary" network controllers

I'll fix the test case now.

dist/templates/rbac-ovnkube-master.yaml.j2
trozet: hmm i dont have one, let me see if i can make one easily

go-controller/pkg/ovn/egressip.go
trozet: well the tests do fail now, thast how i caught it :) It's because the old getActiveNetworkForNamespace behavior would assume default network if a NAD didn't exist. Now it wont do that anymore if the namespace has the label. So an error was getting thrown here during the test and the status was not being removed in the e2es, causing them to fail. I'm not sure if the false positive was just enough to clean up stuff correctly for egress IP. Will defer to @martinkennelly for that.

trozet:
> https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12693489980/job/35419153714?pr=4912
> 
> ```
> Summarizing 2 Failures:
>   [FAIL] Network Segmentation a user defined primary network with multicast feature enabled for namespace should be able to send multicast UDP traffic between nodes [It] with primary layer3 UDN
>   /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:734
>   [FAIL] Network Segmentation a user defined primary network with multicast feature enabled for namespace should be able to send multicast UDP traffic between nodes [It] with primary layer2 UDN
>   /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:734
> 
> Ran 79 of 428 Specs in 3145.252 seconds
> FAIL! -- 77 Passed | 2 Failed | 1 Flaked | 2 Pending | 347 Skipped
> ```
> 
> multicast tests are failing here

yeah I see them failing for ipv6, in the latest run and this time we have logs and e2e dbs. I fixed these tests to use the correctCIDRFamily and now they seem to fail. I looked at the dbs and I dont see anything obvious. All the 3 servers were created and I see them attached to the UDN, but traffic seems to fail. @dceara do you mind taking a look into this please?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12756338404/job/35555521085?pr=4912

trozet:
hmm some new flake that I haven't seen before:

2025-01-14T21:27:42.0290463Z   [38;5;9m[FAIL][0m [0mOVN master EgressIP Operations cluster default network [38;5;243mOn node DELETE [0m[secondary host network] should perform proper OVN transactions when namespace and pod is created after node egress label switch [38;5;9m[1m[It] interconnect enabled; node1 in global and node2 in remote zones[0m
2025-01-14T21:27:42.0293085Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:3296[0m

looks legit as the next hop for the stale egress node was not removed:

```
2025-01-14T21:26:07.5804576Z   the missing elements were
2025-01-14T21:26:07.5805165Z       <[]*libovsdb.testDataMatcher | len:2, cap:2>: [
2025-01-14T21:26:07.5805641Z           {
2025-01-14T21:26:07.5806393Z               expected: <*nbdb.LogicalRouterPolicy | 0xc02c2a3180>{
2025-01-14T21:26:07.5807258Z                   UUID: "reroute-UUID",
2025-01-14T21:26:07.5807891Z                   Action: "reroute",
2025-01-14T21:26:07.5808495Z                   BFDSessions: nil,
2025-01-14T21:26:07.5809076Z                   ExternalIDs: {
2025-01-14T21:26:07.5809752Z                       "ip-family": "ip4",
2025-01-14T21:26:07.5810481Z                       "network": "default",
2025-01-14T21:26:07.5811681Z                       "k8s.ovn.org/name": "egressip_egressip-namespace/egress-pod",
2025-01-14T21:26:07.5813730Z                       "k8s.ovn.org/id": "default-network-controller:EgressIP:100:egressip_egressip-namespace/egress-pod:ip4:default",
2025-01-14T21:26:07.5815499Z                       "k8s.ovn.org/owner-controller": "default-network-controller",
2025-01-14T21:26:07.5816549Z                       "k8s.ovn.org/owner-type": "EgressIP",
2025-01-14T21:26:07.5817282Z                       "priority": "100",
2025-01-14T21:26:07.5817783Z                   },
2025-01-14T21:26:07.5818432Z                   Match: "ip4.src == 10.128.0.15",
2025-01-14T21:26:07.5819044Z                   Nexthop: nil,
2025-01-14T21:26:07.5819700Z                   Nexthops: ["100.88.0.3"],
2025-01-14T21:26:07.5820280Z                   Options: nil,
2025-01-14T21:26:07.5820842Z                   Priority: 100,
2025-01-14T21:26:07.5821268Z               },
2025-01-14T21:26:07.5821733Z               ignoreUUID: false,
2025-01-14T21:26:07.5822137Z           },
2025-01-14T21:26:07.5822450Z           {
2025-01-14T21:26:07.5823148Z               expected: <*nbdb.LogicalRouter | 0xc02c2ea9c0>{
2025-01-14T21:26:07.5824411Z                   UUID: "fca81e0b-99d1-4181-bcac-31a0e5f5f893 [ovn_cluster_router-UUID]",
2025-01-14T21:26:07.5825170Z                   Copp: nil,
2025-01-14T21:26:07.5825705Z                   Enabled: nil,
2025-01-14T21:26:07.5826274Z                   ExternalIDs: nil,
2025-01-14T21:26:07.5826887Z                   LoadBalancer: nil,
2025-01-14T21:26:07.5827554Z                   LoadBalancerGroup: nil,
2025-01-14T21:26:07.5828293Z                   Name: "ovn_cluster_router",
2025-01-14T21:26:07.5828856Z                   Nat: nil,
2025-01-14T21:26:07.5829386Z                   Options: nil,
2025-01-14T21:26:07.5830035Z                   Policies: [
2025-01-14T21:26:07.5830632Z                       "reroute-UUID",
2025-01-14T21:26:07.5831878Z                       "9a60bd0d-1012-4865-b98c-1e1723f22335 [default-no-reroute-UUID]",
2025-01-14T21:26:07.5833289Z                       "68da3de2-a809-425b-81df-2ad6158121b1 [no-reroute-service-UUID]",
2025-01-14T21:26:07.5834646Z                       "661e7434-239d-4862-a161-2c2de5e0efb3 [no-reroute-node-UUID]",
2025-01-14T21:26:07.5836401Z                       "35a3fb4e-fe39-4d88-986e-ce543597eee3 [default-no-reroute-reply-traffic]",
2025-01-14T21:26:07.5837134Z                   ],
2025-01-14T21:26:07.5837609Z                   Ports: nil,
2025-01-14T21:26:07.5838203Z                   StaticRoutes: nil,
2025-01-14T21:26:07.5838648Z               },
2025-01-14T21:26:07.5839123Z               ignoreUUID: false,
2025-01-14T21:26:07.5839518Z           },
2025-01-14T21:26:07.5839821Z       ]
2025-01-14T21:26:07.5840143Z   the extra elements were
2025-01-14T21:26:07.5840615Z       <[]interface {} | len:2, cap:2>: [
2025-01-14T21:26:07.5841295Z           <*nbdb.LogicalRouterPolicy | 0xc02c2a3680>{
2025-01-14T21:26:07.5842110Z               UUID: "e2efbc3b-602e-40a6-b0b9-46924b5951b9",
2025-01-14T21:26:07.5842734Z               Action: "reroute",
2025-01-14T21:26:07.5843262Z               BFDSessions: nil,
2025-01-14T21:26:07.5843765Z               ExternalIDs: {
2025-01-14T21:26:07.5844528Z                   "k8s.ovn.org/owner-type": "EgressIP",
2025-01-14T21:26:07.5845354Z                   "network": "default",
2025-01-14T21:26:07.5846424Z                   "k8s.ovn.org/name": "egressip_egressip-namespace/egress-pod",
2025-01-14T21:26:07.5847184Z                   "priority": "100",
2025-01-14T21:26:07.5847801Z                   "ip-family": "ip4",
2025-01-14T21:26:07.5849608Z                   "k8s.ovn.org/id": "default-network-controller:EgressIP:100:egressip_egressip-namespace/egress-pod:ip4:default",
2025-01-14T21:26:07.5851117Z                   "k8s.ovn.org/owner-controller": "default-network-controller",
2025-01-14T21:26:07.5851739Z               },
2025-01-14T21:26:07.5852304Z               Match: "ip4.src == 10.128.0.15",
2025-01-14T21:26:07.5852857Z               Nexthop: nil,
2025-01-14T21:26:07.5853507Z               Nexthops: ["100.88.0.3", "10.128.0.2"],
2025-01-14T21:26:07.5854068Z               Options: nil,
2025-01-14T21:26:07.5854558Z               Priority: 100,
```

I don't think it is related to this PR, but @martinkennelly heads up

trozet:
Filed https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4965 for the dualstack failure. @dceara has seen this before as well.

tssurya (commented):
I have not reviewed all the new commits, will pause till we are sure we are getting this in

tssurya:
tools seems to be failing:
```
ovn-trace from pod to IP indicates success from pod-test-6d94fd4497-g59sm to 8.8.8.8
ovs-appctl ofproto/trace pod to IP indicates success from pod-test-6d94fd4497-g59sm to 8.8.8.8
ovn-detrace pod to external IP indicates success from pod-test-6d94fd4497-g59sm to 8.8.8.8
Running: ./scripts/../../go-controller/_output/go/bin/ovnkube-trace -src-namespace ovn-kubetrace-pod-test -src pod-test-6d94fd4497-g59sm -dst-ip 8.8.8.8 -udp -loglevel 0
ovn-trace from pod to IP indicates success from pod-test-6d94fd4497-g59sm to 8.8.8.8
ovs-appctl ofproto/trace pod to IP indicates success from pod-test-6d94fd4497-g59sm to 8.8.8.8
ovn-detrace pod to external IP indicates success from pod-test-6d94fd4497-g59sm to 8.8.8.8
Run ovnkube-trace from all egressip-pod-test pods to 8.8.8.8
Running: ./scripts/../../go-controller/_output/go/bin/ovnkube-trace -src-namespace egressip-kubetrace-pod-test -src egressip-pod-test-f96c669c5-7hht7 -dst-ip 8.8.8.8 -tcp -loglevel 0
ovn-trace from pod to IP indicates success from egressip-pod-test-f96c669c5-7hht7 to 8.8.8.8
ovs-appctl ofproto/trace pod to IP indicates failure from egressip-pod-test-f96c669c5-7hht7 to 8.8.8.8
make: *** [Makefile:47: tools] Error 255
make: Leaving directory '/home/runner/work/ovn-kubernetes/ovn-kubernetes/test'
Error: Process completed with exit code 2.
```

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -414,6 +414,16 @@ func (oc *SecondaryLayer2NetworkController) Cleanup() error {
 		}
 		return true
 	})
+
+	// remove load balancer groups
```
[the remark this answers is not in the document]
tssurya: Tim explained it in slack..

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -414,6 +414,16 @@ func (oc *SecondaryLayer2NetworkController) Cleanup() error {
...
+	lbGroups := make([]*nbdb.LoadBalancerGroup, 0, 3)
+	for _, lbGroupUUID := range []string{oc.switchLoadBalancerGroupUUID, oc.clusterLoadBalancerGroupUUID, oc.routerLoadBalancerGroupUUID} {
+		lbGroups = append(lbGroups, &nbdb.LoadBalancerGroup{UUID: lbGroupUUID})
+	}
+	if err := libovsdbops.DeleteLoadBalancerGroups(oc.nbClient, lbGroups); err != nil {
+		klog.Errorf("Failed to delete load balancer groups on network: %q, error: %v", oc.GetNetworkName(), err)
```
tssurya, line 424: best effort with no retries is intentional?

test/e2e/e2e.go
```diff
@@ -44,10 +44,13 @@ const (
 	retryTimeout         = 40 * time.Second // polling timeout
 	rolloutTimeout       = 10 * time.Minute
 	agnhostImage         = "registry.k8s.io/e2e-test-images/agnhost:2.26"
+	agnhostImageNew      = "registry.k8s.io/e2e-test-images/agnhost:2.53"
```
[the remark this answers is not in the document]
tssurya: Tim said when he updated it to 53 EIP tests started failing, in lieu of time he added the 53 version on top of 26.

tssurya (approved):
/lgtm

tssurya:
```

Summarizing 1 Failure:
  [FAIL] Network Segmentation ClusterUserDefinedNetwork CRD Controller pod connected to ClusterUserDefinedNetwork [BeforeEach] CR & managed NADs cannot be deleted when being used
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:2007

Ran 56 of 431 Specs in 1801.979 seconds
```
known flake, merging this PR

EdDev:
I am feedbacking too late here, but AFAIU this seems not to be in sync with the original UDN requirements.

Per the design user-stories [1]:
> As a user, I want to be able to request a unique, primary network for my namespace without having to get administrator permission.

A project-admin should be able to enable a UDN network without a cluster-admin assistance.
But per the change in this PR, only the cluster-admin can now create the namespace that has UDN abilities.
It seems to be problematic for projects (e.g. virtualization) that depend on this.

I am not 100% clear if this change is indeed limiting the user story above and/or if it was considered.

[1] https://github.com/openshift/enhancements/blob/4065a3c36352bec19fd144daa526e4c24fbd1aed/enhancements/network/user-defined-network-segmentation.md?plain=1#L77-L78

PR #4914, opened 2024-12-15
WIP check main branch state

oshoval:
want to see if we have same storage removal problem as here
https://github.com/kubevirt/ipam-extensions/pull/78#issuecomment-2543893334

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

PR #4916, opened 2024-12-16
Statically set GH runner image to ubuntu 22.04 for tests

martinkennelly:
'ubuntu-latest' now references ubuntu 24.04.
Previously, it referenced 22.04.
Potential short term measure to unblock GH CI.
Failures for ubuntu 24.04 can be seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4914

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
   # separate job for parallelism
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     - name: Check out code
       uses: actions/checkout@v4
@@ -57,7 +57,7 @@ jobs:
 
   build-master:
     name: Build-master
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
     # Create a cache for the built master image
     - name: Restore master image cache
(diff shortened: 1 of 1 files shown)


martinkennelly:
cc @ormergi if you need a commit to unblock the CI

maiqueb (approved):
Thank you.

.github/workflows/test.yml
```diff
@@ -36,7 +36,7 @@ jobs:
   # separate job for parallelism
   lint:
     name: Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
tssurya, line 39: why exactly is CI failing on 24.04?
I can't tell the reason from the failed PR as in not sure why the new ubuntu is causing failures..
Is there an upstream github issue somewhere - I assume this is not unique to our project?

tssurya (commented):
I can merge this once I have a reason on why the new ubuntu is breaking CI

tssurya (approved):
going to merge this because this is blocking all PRs/CI - but still need to know if this is the fix and if the problem is with the new ubuntu OR our steps need to be adopted for the new ubuntu - i/e shouldn't we remove only the ones we have installed?
I would still like to know why ubuntu version causes free disk space step to fail

martinkennelly:
Issue created to revert this: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4919

PR #4929, opened 2024-12-18
Add BGP support for UDN

arghosh93:
## 📑 Description
This commit adds capability to advertise an user defined L3
    primary network over providers network. This commit takes care
    of implementation required while using either shared gateway
    mode or local gateway mode.
    This commit introduces various Reconcile function for different
    controllers so that node and ovn resources gets reconciled upon
    advertising an user defined network.
    As part of those Reconcile functions, following resources gets
    added or removed to allow traffic from outside network when the
    network is advertised:
    - OpenFlow flows. For shared gateway mode, one priority 16 flow
    gets added to forward traffic to patch port. For local gateway mode,
    one priority 15 flow gets added to forward traffic to LOCAL port.
    - SNAT rule from ovn_cluster_router specific to and UDN gets removed
    when that network gets advertised.
    - iptables return rule matching destination same as UDN subnet gets
    added to OVN-KUBE-UDN-MASQUERADE chain when the network is advertised
    and gateway mode is local.
    - IP rule gets modified to match UDN subnet instead of the UDN
    specific masquerade subnet.
    - One IP route gets added to the UDN specific vrf table when gateway
    mode is local. This is to route traffic coming for UDN network and to
    forward to UDN specific management port.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -33,7 +33,7 @@ type Gateway interface {
 	Start()
 	GetGatewayBridgeIface() string
 	SetDefaultGatewayBridgeMAC(addr net.HardwareAddr)
-	SetPodNetworkAdvertised(bool)
+	SetDefaultPodNetworkAdvertised(bool)
 	Reconcile() error
 }
 
@@ -57,9 +57,6 @@ type gateway struct {
 	watchFactory *factory.WatchFactory // used for retry
 	stopChan     <-chan struct{}
 	wg           *sync.WaitGroup
-
-	isPodNetworkAdvertisedLock sync.Mutex
-	isPodNetworkAdvertised     bool
 }
 
 func (g *gateway) AddService(svc *corev1.Service) error {
@@ -352,15 +349,16 @@ func setupUDPAggregationUplink(ifname string) error {
 	return nil
 }
 
-func gatewayInitInternal(nodeName, gwIntf, egressGatewayIntf string, gwNextHops []net.IP, gwIPs []*net.IPNet, nodeAnnotator kube.Annotator) (
+func gatewayInitInternal(nodeName, gwIntf, egressGatewayIntf string, gwNextHops []net.IP, nodeSubnets, gwIPs []*net.IPNet,
+	advertised bool, nodeAnnotator kube.Annotator) (
 	*bridgeConfiguration, *bridgeConfiguration, error) {
-	gatewayBridge, err := bridgeForInterface(gwIntf, nodeName, types.PhysicalNetworkName, gwIPs)
+	gatewayBridge, err := bridgeForInterface(gwIntf, nodeName, types.PhysicalNetworkName, nodeSubnets, gwIPs, advertised)
 	if err != nil {
 		return nil, nil, fmt.Errorf("bridge for interface failed for %s: %w", gwIntf, err)
 	}
 	var egressGWBridge *bridgeConfiguration
 	if egressGatewayIntf != "" {
-		egressGWBridge, err = bridgeForInterface(egressGatewayIntf, nodeName, types.PhysicalNetworkExGwName, nil)
+		egressGWBridge, err = bridgeForInterface(egressGatewayIntf, nodeName, types.PhysicalNetworkExGwName, nodeSubnets, nil, false)
 		if err != nil {
 			return nil, nil, fmt.Errorf("bridge for interface failed for %s: %w", egressGatewayIntf, err)
 		}
@@ -469,27 +467,21 @@ func (g *gateway) SetDefaultGatewayBridgeMAC(macAddr net.HardwareAddr) {
 	klog.Infof("Default gateway bridge MAC address updated to %s", macAddr)
 }
 
-func (g *gateway) SetPodNetworkAdvertised(isPodNetworkAdvertised bool) {
-	g.isPodNetworkAdvertisedLock.Lock()
-	defer g.isPodNetworkAdvertisedLock.Unlock()
-	g.isPodNetworkAdvertised = isPodNetworkAdvertised
+func (g *gateway) SetDefaultPodNetworkAdvertised(isPodNetworkAdvertised bool) {
+	g.openflowManager.defaultBridge.netConfig[types.DefaultNetworkName].advertised.Store(isPodNetworkAdvertised)
+}
+
+func (g *gateway) GetDefaultPodNetworkAdvertised() bool {
+	return g.openflowManager.defaultBridge.netConfig[types.DefaultNetworkName].advertised.Load()
 }
 
 // Reconcile handles triggering updates to different components of a gateway, like OFM, Services
 func (g *gateway) Reconcile() error {
 	klog.Info("Reconciling gateway with updates")
-	node, err := g.watchFactory.GetNode(g.nodeIPManager.nodeName)
-	if err != nil {
+	if err := g.openflowManager.updateBridgeFlowCache(g.nodeIPManager.ListAddresses()); err != nil {
 		return err
 	}
-	subnets, err := util.ParseNodeHostSubnetAnnotation(node, types.DefaultNetworkName)
-	if err != nil {
-		return fmt.Errorf("failed to get subnets for node: %s for OpenFlow cache update; err: %w", node.Name, err)
-	}
-	if err := g.openflowManager.updateBridgeFlowCache(subnets, g.nodeIPManager.ListAddresses(), g.isPodNetworkAdvertised); err != nil {
-		return err
-	}
-	err = g.updateSNATRules()
+	err := g.updateSNATRules()
 	if err != nil {
 		return err
 	}
@@ -524,8 +516,6 @@ func (g *gateway) addAllServices() []error {
 }
 
 func (g *gateway) updateSNATRules() error {
-	g.isPodNetworkAdvertisedLock.Lock()
-	defer g.isPodNetworkAdvertisedLock.Unlock()
 	var ipnets []*net.IPNet
 	if g.nodeIPManager.mgmtPortConfig.ipv4 != nil {
 		ipnets = append(ipnets, g.nodeIPManager.mgmtPortConfig.ipv4.ifAddr)
@@ -535,7 +525,7 @@ func (g *gateway) updateSNATRules() error {
 	}
 	subnets := util.IPsToNetworkIPs(ipnets...)
 
-	if g.isPodNetworkAdvertised || config.Gateway.Mode != config.GatewayModeLocal {
+	if g.GetDefaultPodNetworkAdvertised() || config.Gateway.Mode != config.GatewayModeLocal {
 		return delLocalGatewayPodSubnetNATRules(subnets...)
 	}
 
@@ -586,9 +576,12 @@ func (b *bridgeConfiguration) updateInterfaceIPAddresses(node *corev1.Node) ([]*
 	return ifAddrs, nil
 }
 
-func bridgeForInterface(intfName, nodeName, physicalNetworkName string, gwIPs []*net.IPNet) (*bridgeConfiguration, error) {
+func bridgeForInterface(intfName, nodeName, physicalNetworkName string, nodeSubnets, gwIPs []*net.IPNet,
+	advertised bool) (*bridgeConfiguration, error) {
 	defaultNetConfig := &bridgeUDNConfiguration{
-		masqCTMark: ctMarkOVN,
+		masqCTMark:  ctMarkOVN,
+		subnets:     config.Default.ClusterSubnets,
+		nodeSubnets: nodeSubnets,
 	}
 	res := bridgeConfiguration{
 		nodeName: nodeName,
@@ -597,6 +590,7 @@ func bridgeForInterface(intfName, nodeName, physicalNetworkName string, gwIPs []
 		},
 		eipMarkIPs: newMarkIPsCache(),
 	}
+	res.netConfig[types.DefaultNetworkName].advertised.Store(advertised)
 	gwIntf := intfName
 
 	if bridgeName, _, err := util.RunOVSVsctl("port-to-br", intfName); err == nil {
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1762,7 +1762,7 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 	return dftFlows, nil
 }
 
-func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetworkAdvertised bool) ([]string, error) {
+func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
 	// CAUTION: when adding new flows where the in_port is ofPortPatch and the out_port is ofPortPhys, ensure
 	// that dl_src is included in match criteria!
 	ofPortPhys := bridge.ofPortPhys
@@ -2000,7 +2000,7 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
 				fmt.Sprintf("cookie=%s, priority=104, in_port=%s, %s, %s_src=%s, actions=drop",
 					defaultOpenFlowCookie, defaultNetConfig.ofPortPatch, ipv, ipv, cidr))
 		}
-		for _, subnet := range subnets {
+		for _, subnet := range defaultNetConfig.nodeSubnets {
 			ipv := getIPv(subnet)
 			if ofPortPhys != "" {
 				// table 0, commit connections from local pods.
@@ -2015,10 +2015,15 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
 	}
 
 	if ofPortPhys != "" {
-		if config.Gateway.DisableSNATMultipleGWs || isPodNetworkAdvertised {
-			// table 1, traffic to pod subnet go directly to OVN
-			output := defaultNetConfig.ofPortPatch
-			if isPodNetworkAdvertised && config.Gateway.Mode == config.GatewayModeLocal {
+		for _, netConfig := range bridge.patchedNetConfigs() {
+			isNetworkAdvertised := netConfig.advertised.Load()
+			// disableSNATMultipleGWs only applies to default network
+			disableSNATMultipleGWs := netConfig.isDefaultNetwork() && config.Gateway.DisableSNATMultipleGWs
+			if !disableSNATMultipleGWs && !isNetworkAdvertised {
+				continue
+			}
+			output := netConfig.ofPortPatch
+			if isNetworkAdvertised && config.Gateway.Mode == config.GatewayModeLocal {
 				// except if advertised through BGP, go to kernel
 				// TODO: MEG enabled pods should still go through the patch port
 				// but holding this until
@@ -2026,7 +2031,7 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
 				// are assuming MEG & BGP are not used together
 				output = ovsLocalPort
 			}
-			for _, clusterEntry := range config.Default.ClusterSubnets {
+			for _, clusterEntry := range netConfig.subnets {
 				cidr := clusterEntry.CIDR
 				ipv := getIPv(cidr)
 				dftFlows = append(dftFlows,
@@ -2034,9 +2039,9 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
 						"actions=output:%s",
 						defaultOpenFlowCookie, ipv, ipv, cidr, output))
 			}
-			if output == defaultNetConfig.ofPortPatch {
+			if output == netConfig.ofPortPatch {
 				// except node management traffic
-				for _, subnet := range subnets {
+				for _, subnet := range netConfig.nodeSubnets {
 					mgmtIP := util.GetNodeManagementIfAddr(subnet)
 					ipv := getIPv(mgmtIP)
 					dftFlows = append(dftFlows,
@@ -2229,8 +2234,9 @@ func newGateway(
 		}
 	}
 
+	advertised := util.IsPodNetworkAdvertisedAtNode(networkManager.GetNetwork(types.DefaultNetworkName), nodeName)
 	gwBridge, exGwBridge, err := gatewayInitInternal(
-		nodeName, gwIntf, egressGWIntf, gwNextHops, gwIPs, nodeAnnotator)
+		nodeName, gwIntf, egressGWIntf, gwNextHops, subnets, gwIPs, advertised, nodeAnnotator)
 	if err != nil {
 		return nil, err
 	}
@@ -2315,7 +2321,7 @@ func newGateway(
 			}
 		}
 
-		gw.openflowManager, err = newGatewayOpenFlowManager(gwBridge, exGwBridge, subnets, nodeIPs)
+		gw.openflowManager, err = newGatewayOpenFlowManager(gwBridge, exGwBridge, nodeIPs)
 		if err != nil {
 			return err
 		}
@@ -2323,7 +2329,7 @@ func newGateway(
 		// resync flows on IP change
 		gw.nodeIPManager.OnChanged = func() {
 			klog.V(5).Info("Node addresses changed, re-syncing bridge flows")
-			if err := gw.openflowManager.updateBridgeFlowCache(subnets, gw.nodeIPManager.ListAddresses(), gw.isPodNetworkAdvertised); err != nil {
+			if err := gw.openflowManager.updateBridgeFlowCache(gw.nodeIPManager.ListAddresses()); err != nil {
 				// very unlikely - somehow node has lost its IP address
 				klog.Errorf("Failed to re-generate gateway flows after address change: %v", err)
 			}
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -4,12 +4,15 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync/atomic"
 	"time"
 
 	"github.com/vishvananda/netlink"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 	"k8s.io/utils/ptr"
@@ -76,24 +79,32 @@ type UserDefinedNetworkGateway struct {
 	// iprules manager that creates and manages iprules for
 	// all UDNs. Must be accessed with a lock
 	ruleManager *iprulemanager.Controller
+
+	// reconcile channel to signal reconciliation of the gateway on network
+	// configuration changes
+	reconcile chan struct{}
 }
 
 // UTILS Needed for UDN (also leveraged for default netInfo) in bridgeConfiguration
 
 // getBridgePortConfigurations returns a slice of Network port configurations along with the
 // uplinkName and physical port's ofport value
-func (b *bridgeConfiguration) getBridgePortConfigurations() ([]bridgeUDNConfiguration, string, string) {
+func (b *bridgeConfiguration) getBridgePortConfigurations() ([]*bridgeUDNConfiguration, string, string) {
 	b.Lock()
 	defer b.Unlock()
-	netConfigs := make([]bridgeUDNConfiguration, len(b.netConfig))
+	var netConfigs []*bridgeUDNConfiguration
 	for _, netConfig := range b.netConfig {
-		netConfigs = append(netConfigs, *netConfig)
+		netConfigs = append(netConfigs, netConfig.shallowCopy())
 	}
 	return netConfigs, b.uplinkName, b.ofPortPhys
 }
 
 // addNetworkBridgeConfig adds the patchport and ctMark value for the provided netInfo into the bridge configuration cache
-func (b *bridgeConfiguration) addNetworkBridgeConfig(nInfo util.NetInfo, masqCTMark, pktMark uint, v6MasqIPs, v4MasqIPs *udn.MasqueradeIPs) {
+func (b *bridgeConfiguration) addNetworkBridgeConfig(
+	nInfo util.NetInfo,
+	nodeSubnets []*net.IPNet,
+	masqCTMark, pktMark uint,
+	v6MasqIPs, v4MasqIPs *udn.MasqueradeIPs) error {
 	b.Lock()
 	defer b.Unlock()
 
@@ -103,18 +114,22 @@ func (b *bridgeConfiguration) addNetworkBridgeConfig(nInfo util.NetInfo, masqCTM
 	_, found := b.netConfig[netName]
 	if !found {
 		netConfig := &bridgeUDNConfiguration{
-			patchPort:  patchPort,
-			masqCTMark: fmt.Sprintf("0x%x", masqCTMark),
-			pktMark:    fmt.Sprintf("0x%x", pktMark),
-			v4MasqIPs:  v4MasqIPs,
-			v6MasqIPs:  v6MasqIPs,
+			patchPort:   patchPort,
+			masqCTMark:  fmt.Sprintf("0x%x", masqCTMark),
+			pktMark:     fmt.Sprintf("0x%x", pktMark),
+			v4MasqIPs:   v4MasqIPs,
+			v6MasqIPs:   v6MasqIPs,
+			subnets:     nInfo.Subnets(),
+			nodeSubnets: nodeSubnets,
 		}
+		netConfig.advertised.Store(util.IsPodNetworkAdvertisedAtNode(nInfo, b.nodeName))
 
 		b.netConfig[netName] = netConfig
 	} else {
 		klog.Warningf("Trying to update bridge config for network %s which already"+
 			"exists in cache...networks are not mutable...ignoring update", nInfo.GetNetworkName())
 	}
+	return nil
 }
 
 // delNetworkBridgeConfig deletes the provided netInfo from the bridge configuration cache
@@ -125,7 +140,7 @@ func (b *bridgeConfiguration) delNetworkBridgeConfig(nInfo util.NetInfo) {
 	delete(b.netConfig, nInfo.GetNetworkName())
 }
 
-// getActiveNetworkBridgeConfig returns a copy of the network configuration corresponding to the
+// getActiveNetworkBridgeConfig returns a shallow copy of the network configuration corresponding to the
 // provided netInfo.
 //
 // NOTE: if the network configuration can't be found or if the network is not patched by OVN
@@ -135,8 +150,7 @@ func (b *bridgeConfiguration) getActiveNetworkBridgeConfig(networkName string) *
 	defer b.Unlock()
 
 	if netConfig, found := b.netConfig[networkName]; found && netConfig.ofPortPatch != "" {
-		result := *netConfig
-		return &result
+		return netConfig.shallowCopy()
 	}
 	return nil
 }
@@ -163,6 +177,24 @@ type bridgeUDNConfiguration struct {
 	pktMark     string
 	v4MasqIPs   *udn.MasqueradeIPs
 	v6MasqIPs   *udn.MasqueradeIPs
+	subnets     []config.CIDRNetworkEntry
+	nodeSubnets []*net.IPNet
+	advertised  atomic.Bool
+}
+
+func (netConfig *bridgeUDNConfiguration) shallowCopy() *bridgeUDNConfiguration {
+	copy := &bridgeUDNConfiguration{
+		patchPort:   netConfig.patchPort,
+		ofPortPatch: netConfig.ofPortPatch,
+		masqCTMark:  netConfig.masqCTMark,
+		pktMark:     netConfig.pktMark,
+		v4MasqIPs:   netConfig.v4MasqIPs,
+		v6MasqIPs:   netConfig.v6MasqIPs,
+		subnets:     netConfig.subnets,
+		nodeSubnets: netConfig.nodeSubnets,
+	}
+	netConfig.advertised.Store(netConfig.advertised.Load())
+	return copy
 }
 
 func (netConfig *bridgeUDNConfiguration) isDefaultNetwork() bool {
@@ -232,6 +264,7 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, networkID int, node *cor
 		v6MasqIPs:     v6MasqIPs,
 		gateway:       gw,
 		ruleManager:   ruleManager,
+		reconcile:     make(chan struct{}, 1),
 	}, nil
 }
 
@@ -309,14 +342,9 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 		return fmt.Errorf("could not add VRF %s routes for network %s, err: %v", vrfDeviceName, udng.GetNetworkName(), err)
 	}
 	// create the iprules for this network
-	udnReplyIPRules, err := udng.constructUDNVRFIPRules(vrfTableId)
+	err = udng.updateUDNVRFIPRule()
 	if err != nil {
-		return fmt.Errorf("unable to get iprules for network %s, err: %v", udng.GetNetworkName(), err)
-	}
-	for _, rule := range udnReplyIPRules {
-		if err = udng.ruleManager.AddWithMetadata(rule, udng.GetNetworkRuleMetadata()); err != nil {
-			return fmt.Errorf("unable to create iprule %v for network %s, err: %v", rule, udng.GetNetworkName(), err)
-		}
+		return fmt.Errorf("failed to update IP rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 	// add loose mode for rp filter on management port
 	mgmtPortName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(udng.networkID))
@@ -324,7 +352,13 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 		return fmt.Errorf("could not set loose mode for reverse path filtering on management port %s: %v", mgmtPortName, err)
 	}
 	if udng.openflowManager != nil {
-		udng.openflowManager.addNetwork(udng.NetInfo, udng.masqCTMark, udng.pktMark, udng.v6MasqIPs, udng.v4MasqIPs)
+		nodeSubnets, err := udng.getLocalSubnets()
+		if err != nil {
+			return fmt.Errorf("failed to get node subnets for network %s: %w", udng.GetNetworkName(), err)
+		}
+		if err = udng.openflowManager.addNetwork(udng.NetInfo, nodeSubnets, udng.masqCTMark, udng.pktMark, udng.v6MasqIPs, udng.v4MasqIPs); err != nil {
+			return fmt.Errorf("could not add network %s: %v", udng.GetNetworkName(), err)
+		}
 
 		waiter := newStartupWaiterWithTimeout(waitForPatchPortTimeout)
 		readyFunc := func() (bool, error) {
@@ -341,7 +375,7 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 			return true, nil
 		}
 		postFunc := func() error {
-			if err := udng.Reconcile(); err != nil {
+			if err := udng.gateway.Reconcile(); err != nil {
 				return fmt.Errorf("failed to reconcile flows on bridge for network %s; error: %v", udng.GetNetworkName(), err)
 			}
 			return nil
@@ -357,6 +391,10 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	if err := udng.addMarkChain(); err != nil {
 		return fmt.Errorf("failed to add the service masquerade chain: %w", err)
 	}
+
+	// run gateway reconciliation loop on network configuration changes
+	udng.run()
+
 	return nil
 }
 
@@ -367,6 +405,8 @@ func (udng *UserDefinedNetworkGateway) GetNetworkRuleMetadata() string {
 // DelNetwork will be responsible to remove all plumbings
 // used by this UDN on the gateway side
 func (udng *UserDefinedNetworkGateway) DelNetwork() error {
+	close(udng.reconcile)
+
 	vrfDeviceName := util.GetNetworkVRFName(udng.NetInfo)
 	// delete the iprules for this network
 	if err := udng.ruleManager.DeleteWithMetadata(udng.GetNetworkRuleMetadata()); err != nil {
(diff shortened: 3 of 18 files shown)


go-controller/pkg/node/secondary_node_network_controller.go
```diff
@@ -122,11 +122,108 @@ func (oc *SecondaryNodeNetworkController) getNetworkID() (int, error) {
...
+	subnets, err := util.ParseNodeHostSubnetAnnotation(node, types.DefaultNetworkName)
+	if err != nil {
+		return fmt.Errorf("failed to get subnets for node: %s for OpenFlow cache update; err: %w", node.Name, err)
+	}
+	if err := oc.gateway.openflowManager.updateBridgeFlowCache(subnets, oc.gateway.nodeIPManager.ListAddresses(),
+		oc.gateway.isPodNetworkAdvertised, oc.gateway.isUDNNetworkAdvertised); err != nil {
+		return err
+	}
 	return nil
 }
```
jcaamano, line 229: I would move these two methods to `UserDefinedNetworkGateway` as it is responsible for what it is being done (and undone) here.

I would also have them under a general `UserDefinedNetworkGateway.Reconcile` method just as we have `UserDefinedNetworkGateway.gateway.Reconcile` which you can reuse from the former to sync flows, and you can also reuse `UserDefinedNetworkGateway.gateway.Reconcile`

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1986,35 +1987,53 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
...
-					mgmtIP := util.GetNodeManagementIfAddr(subnet)
-					ipv := getIPv(mgmtIP)
+			for netName, netConfig := range bridge.netConfig {
+				if isPodNetworkAdvertised && netName != types.DefaultNetworkName {
+					continue
+				} else if isUDNNetworkAdvertised && netName == types.DefaultNetworkName {
+					continue
+				}
+				output := netConfig.ofPortPatch
+				if (isPodNetworkAdvertised || isUDNNetworkAdvertised) && config.Gateway.Mode == config.GatewayModeLocal {
```
jcaamano, line 1999: How do you know to which network does the `isUDNNetworkAdvertised` apply to? Have you tested with multiple UDNs where only a subset of them is advertised? It looks like just as `bridge.netConfig` is a map with different networks, we should also have a map to store `isUDNNetworkAdvertised` per network. I guess we can use `bridge.netConfig` as well for this.

go-controller/pkg/node/secondary_node_network_controller.go
```diff
@@ -122,11 +122,108 @@ func (oc *SecondaryNodeNetworkController) getNetworkID() (int, error) {
...
+// updateUDNFlow adds below OpenFlows based on the gateway mode and whether the network
+// is advertised or not:
+// table=1, n_packets=0, n_bytes=0, priority=16,ip,nw_dst=128.192.0.2 actions=LOCAL (Both gateway modes)
+// table=1, n_packets=0, n_bytes=0, priority=15,ip,nw_dst=128.192.0.0/14 actions=output:3 (shared gateway mode)
+func (oc *SecondaryNodeNetworkController) updateUDNFlow() error {
+	node, err := oc.gateway.watchFactory.GetNode(oc.gateway.nodeIPManager.nodeName)
+	if err != nil {
+		return fmt.Errorf("unable to get node %s: %s", oc.gateway.nodeIPManager.nodeName, err)
+	}
+	subnets, err := util.ParseNodeHostSubnetAnnotation(node, types.DefaultNetworkName)
```
jcaamano, line 220: I guess we can avoid passing this subnets if they are already provided through the `bridgeConfiguration`

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1986,35 +1987,53 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
...
+				}
+				if output == netConfig.ofPortPatch {
+					// except node management traffic
+					nodeSubnets := subnets
+					if netName != types.DefaultNetworkName {
+						node, err := c.watchFactory.GetNode(bridge.nodeName)
+						if err != nil {
+							return nil, fmt.Errorf("unable to get node %s: %s", bridge.nodeName, err)
+						}
+						nodeSubnets, err = util.ParseNodeHostSubnetAnnotation(node, netName)
```
jcaamano, line 2023: This parsing is very expensive and we will parse the annotation fully for each of all the networks each time a network config changes. 

We better do it just once so it should probably be set on `bridge.netConfig` from `UserDefinedNetworkGateway`

go-controller/pkg/node/secondary_node_network_controller.go
```diff
@@ -122,11 +122,108 @@ func (oc *SecondaryNodeNetworkController) getNetworkID() (int, error) {
...
+	mplink, err := util.LinkByName(interfaceName)
+	if err != nil {
+		return fmt.Errorf("unable to get link for %s, error: %v", interfaceName, err)
+	}
+	vrfTableId := util.CalculateRouteTableID(mplink.Attrs().Index)
+
+	if err = oc.gateway.ruleManager.DeleteWithMetadata(oc.gateway.GetNetworkRuleMetadata()); err != nil {
+		return fmt.Errorf("unable to delete iprule for network %s, err: %v", oc.gateway.GetNetworkName(), err)
+	}
+	udnReplyIPRules, err := oc.gateway.constructUDNVRFIPRules(vrfTableId)
```
jcaamano, line 199: Will these replace the old rules os will there be leftovers?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -659,23 +665,36 @@ func (udng *UserDefinedNetworkGateway) getV6MasqueradeIP() (*net.IPNet, error) {
 // 2000:   from all to 169.254.0.12 lookup 1007
 // 2000:   from all fwmark 0x1002 lookup 1009
 // 2000:   from all to 169.254.0.14 lookup 1009
+// If isPodNetworkAdvertised is set to true then we add following IP rules
+// for 10.132.0.0/14 UDN subnet
+// 2000:	from all fwmark 0x1001 lookup 1009
+// 2000:	from all to 10.132.0.0/14 lookup 1009
```
jcaamano, line 671: `add` is confusing here. It can be understood as added to the rules above. I would explain it better.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -193,7 +196,55 @@ type BaseSecondaryNetworkController struct {
...
+		if err != nil {
+			klog.Errorf("Failed to retry node %s for network %s: %v", node.Name, oc.GetNetworkName(), err)
+		}
+	}
+	if len(retryNodes) > 0 {
+		oc.retryNodes.RequestRetryObjs()
+	}
+
+	return nil
 }
```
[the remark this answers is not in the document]
arghosh93: As per discussion DefaultNetworkController.Reconcile and SecondaryLayer3NetworkController.Reconcile both calls BaseNetworkController.reconcile to complete the reconciliation.

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -848,6 +920,11 @@ func (oc *SecondaryLayer3NetworkController) addNode(node *kapi.Node) ([]*net.IPN
 			return nil, err
 		}
 	}
+	if util.IsPodNetworkAdvertisedAtNode(oc, node.Name) {
+		if err := oc.deleteUDNNodeSubnetEgressSNAT(hostSubnets, node); err != nil {
+			return nil, err
+		}
+	}
 	return hostSubnets, nil
```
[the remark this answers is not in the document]
arghosh93: Ack

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -294,11 +298,37 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
 	return vrfm.sync(vrfDev)
 }
 
+// DeleteVRFRoutes delets routes from the specified VRF
+func (vrfm *Controller) DeleteVRFRoutes(name string, routes []netlink.Route) error {
```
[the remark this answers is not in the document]
arghosh93: Removed it. Will push new changes soon.

go-controller/pkg/node/secondary_node_network_controller.go
```diff
@@ -122,11 +122,108 @@ func (oc *SecondaryNodeNetworkController) getNetworkID() (int, error) {
...
+	} else {
+		for _, udnSubnet := range oc.Subnets() {
+			if err := deleteIptRules(getUDNSubnetReturnRule(udnSubnet.CIDR,
+				getIPTablesProtocol(udnSubnet.CIDR.IP.String()))); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
arghosh93: Yaa, I agree. I have tested without this RETURN rule and it works as expected.

go-controller/pkg/node/secondary_node_network_controller.go
arghosh93: Ack

go-controller/pkg/node/gateway_shared_intf.go
arghosh93: yes, makes sense. I have made required changes. Will push soon.

go-controller/pkg/node/secondary_node_network_controller.go
arghosh93: It will replace old IP rules.

go-controller/pkg/node/gateway_udn.go
arghosh93: Ack

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -584,6 +584,38 @@ func (oc *SecondaryLayer3NetworkController) Run() error {
...
+		retryNodes = append(retryNodes, node)
+		retryNodeNames = append(retryNodeNames, node.Name)
+		return true
+	})
+	if err != nil {
+		return fmt.Errorf("failed to reconcile network %s: %w", oc.GetNetworkName(), err)
+	}
+
+	for _, node := range retryNodes {
+		oc.addNodeFailed.Store(node.Name, true)
```
jcaamano, line 613: Why don't you use `gatewaysFailed`? Then you don't have to pass in `retryNodes` and then the whole reconcile implementation si the same for all network types?

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -182,6 +182,34 @@ type BaseNetworkController struct {
...
+
+	err := util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	if err != nil {
+		klog.Errorf("Failed to reconcile network %s: %v", oc.GetNetworkName(), err)
+	}
+
+	if reconcileRoutes {
+		err = oc.routeImportManager.ReconcileNetwork(oc.GetNetworkName())
+		if err != nil {
+			return err
```
jcaamano, line 196: Technically we should just klog an error here, since `ReconcileNetInfo` above is a point of no return (I had a comment on the original code, perhaps you can add it back) as once it is called. `NeedsReconciliation` will return `false` anyway.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -463,6 +463,10 @@ func (oc *SecondaryLayer2NetworkController) Stop() {
 	oc.BaseSecondaryLayer2NetworkController.stop()
 }
 
+func (oc *SecondaryLayer2NetworkController) Reconcile(netInfo util.NetInfo) error {
+	return nil
```
jcaamano, line 467: You should `return util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)`

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -308,6 +308,10 @@ func (oc *SecondaryLocalnetNetworkController) Stop() {
 	oc.BaseSecondaryLayer2NetworkController.stop()
 }
 
+func (oc *SecondaryLocalnetNetworkController) Reconcile(netInfo util.NetInfo) error {
+	return nil
```
jcaamano, line 312: You should `return util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)`

go-controller/pkg/node/secondary_node_network_controller.go
```diff
@@ -122,11 +122,37 @@ func (oc *SecondaryNodeNetworkController) getNetworkID() (int, error) {
...
 	if err != nil {
 		klog.Errorf("Failed to reconcile network %s: %v", oc.GetNetworkName(), err)
 	}
+
+	if reconcilePodNetwork {
+		isUDNNetworkAdvertisedAtNode := oc.isUDNNetworkAdvertisedAtNode()
+		if oc.gateway != nil {
+			oc.gateway.SetUDNNetworkAdvertised(oc.GetNetworkName(), isUDNNetworkAdvertisedAtNode)
+			if err := oc.gateway.Reconcile(); err != nil {
+				return err
```
jcaamano, line 148: Technically we should just klog an error here, since ReconcileNetInfo above is a point of no return (I had a comment on the original code, perhaps you can add it back) as once it is called. NeedsReconciliation will return false anyway.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -108,13 +126,16 @@ func (b *bridgeConfiguration) addNetworkBridgeConfig(nInfo util.NetInfo, masqCTM
 			pktMark:    fmt.Sprintf("0x%x", pktMark),
 			v4MasqIPs:  v4MasqIPs,
 			v6MasqIPs:  v6MasqIPs,
+			subnets:    nInfo.Subnets(),
+			nodeSubnet: nodeSubnets,
```
jcaamano, line 130: nit: s/nodeSubnet/nodeSubnets

go-controller/pkg/node/gateway_udn.go
```diff
@@ -659,23 +685,35 @@ func (udng *UserDefinedNetworkGateway) getV6MasqueradeIP() (*net.IPNet, error) {
 // 2000:   from all to 169.254.0.12 lookup 1007
 // 2000:   from all fwmark 0x1002 lookup 1009
 // 2000:   from all to 169.254.0.14 lookup 1009
+// If isPodNetworkAdvertised is set to true then we update IP rules as below
```
jcaamano, line 688: nit: maybe s/update/replace ?

jcaamano: Are you sure it is actually replacing the rules?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +767,83 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+	defer udng.isUDNNetworkAdvertisedLock.Unlock()
+	udng.isUDNNetworkAdvertised = isUDNNetworkAdvertised
+	if udng.openflowManager != nil {
+		for netName, netConfig := range udng.openflowManager.defaultBridge.netConfig {
+			if netName == networkName {
+				netConfig.isUDNNetworkAdvertised = isUDNNetworkAdvertised
+			} else if netName == types.DefaultNetworkName {
+				netConfig.isUDNNetworkAdvertised = udng.isPodNetworkAdvertised
+			}
+		}
```
jcaamano, line 782: You shouldn't be handling the default network here

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +767,83 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+// table=1, n_packets=0, n_bytes=0, priority=15,ip,nw_dst=128.192.0.0/14 actions=output:3 (shared gateway mode)
+func (udng *UserDefinedNetworkGateway) updateUDNFlow() error {
+	node, err := udng.watchFactory.GetNode(udng.nodeIPManager.nodeName)
+	if err != nil {
+		return fmt.Errorf("unable to get node %s: %s", udng.nodeIPManager.nodeName, err)
+	}
+	subnets, err := util.ParseNodeHostSubnetAnnotation(node, types.DefaultNetworkName)
+	if err != nil {
+		return fmt.Errorf("failed to get subnets for node: %s for OpenFlow cache update; err: %w", node.Name, err)
+	}
```
jcaamano, line 843: We should not need to pass the subnets if they are already stored in `openflowManager.defaultBridge.netConfig`

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +767,83 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+	udnReplyIPRules, err := udng.constructUDNVRFIPRules(vrfTableId)
+	if err != nil {
+		return fmt.Errorf("unable to get iprules for network %s, err: %v", udng.GetNetworkName(), err)
+	}
+
+	for _, rule := range udnReplyIPRules {
+		if err = udng.ruleManager.AddWithMetadata(rule, udng.GetNetworkRuleMetadata()); err != nil {
+			return fmt.Errorf("unable to create iprule %v for network %s, err: %v", rule, udng.GetNetworkName(), err)
+		}
```
jcaamano, line 826: I think that the correct approach is add the new rules first, and then delete the old ones. We don't want a moment in time when there is just no rules.

go-controller/pkg/node/gateway_udn.go
jcaamano: Ok, I see that you are deleting in `updateUDNVRFIPRule`.

So I would jsut preface the previous example with 
`If the network is not advertise, we set rules...`

And this example with
`If the network is advertised we set rules...`

go-controller/pkg/node/gateway_udn.go
```diff
@@ -76,6 +78,11 @@ type UserDefinedNetworkGateway struct {
 	// iprules manager that creates and manages iprules for
 	// all UDNs. Must be accessed with a lock
 	ruleManager *iprulemanager.Controller
+
+	// isUDNNetworkAdvertised is a place holder to indicate whether the
+	// network is advertised or not
+	isUDNNetworkAdvertised     bool
+	isUDNNetworkAdvertisedLock sync.Mutex
```
jcaamano, line 85: I don't think you need this, the information is already available in `util.NetInfo` above.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +767,83 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
 	}
 	return nil
 }
+
+func (udng *UserDefinedNetworkGateway) SetUDNNetworkAdvertised(networkName string, isUDNNetworkAdvertised bool) {
```
jcaamano, line 771: I don't think you need this method. Just do this on `Reconcile`. You will have the information updated iin NetInfo.

go-controller/pkg/node/gateway.go
```diff
@@ -472,6 +472,13 @@ func (g *gateway) SetPodNetworkAdvertised(isPodNetworkAdvertised bool) {
 	g.isPodNetworkAdvertisedLock.Lock()
 	defer g.isPodNetworkAdvertisedLock.Unlock()
 	g.isPodNetworkAdvertised = isPodNetworkAdvertised
+	if g.openflowManager != nil {
+		for netName, netConfig := range g.openflowManager.defaultBridge.netConfig {
+			if netName == types.DefaultNetworkName {
+				netConfig.isUDNNetworkAdvertised = isPodNetworkAdvertised
+			}
+		}
```
jcaamano, line 480: This is a map that can be accessed by Key, why are you iterating?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +767,83 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+	udng.isUDNNetworkAdvertisedLock.Lock()
+	defer udng.isUDNNetworkAdvertisedLock.Unlock()
+	udng.isUDNNetworkAdvertised = isUDNNetworkAdvertised
+	if udng.openflowManager != nil {
+		for netName, netConfig := range udng.openflowManager.defaultBridge.netConfig {
```
jcaamano, line 776: This is a map that can be accessed by Key, why are you iterating?

go-controller/pkg/node/openflow_manager.go
```diff
@@ -219,7 +229,7 @@ func (c *openflowManager) updateBridgeFlowCache(subnets []*net.IPNet, extraIPs [
 	if err != nil {
 		return err
 	}
-	dftCommonFlows, err := commonFlows(subnets, c.defaultBridge, isPodNetworkAdvertised)
+	dftCommonFlows, err := commonFlows(subnets, c.defaultBridge, isPodNetworkAdvertised, isUDNNetworkAdvertised)
```
jcaamano, line 232: We don't need to be passing `isPodNetworkAdvertised, isUDNNetworkAdvertised` around anymore, all that information is now readily available in `openflowManager.defaultBridge.netConfig[<network name>]`

go-controller/pkg/node/gateway_udn.go
```diff
@@ -157,12 +178,15 @@ func (b *bridgeConfiguration) patchedNetConfigs() []*bridgeUDNConfiguration {
...
-	v6MasqIPs   *udn.MasqueradeIPs
+	patchPort              string
+	ofPortPatch            string
+	masqCTMark             string
+	pktMark                string
+	v4MasqIPs              *udn.MasqueradeIPs
+	v6MasqIPs              *udn.MasqueradeIPs
+	subnets                []config.CIDRNetworkEntry
+	nodeSubnet             []*net.IPNet
+	isUDNNetworkAdvertised bool
```
jcaamano, line 189: nit: s/isUDNNetworkAdvertised/isNetworkAdvertised

go-controller/pkg/node/gateway.go
```diff
@@ -588,6 +595,7 @@ func (b *bridgeConfiguration) updateInterfaceIPAddresses(node *kapi.Node) ([]*ne
 func bridgeForInterface(intfName, nodeName, physicalNetworkName string, gwIPs []*net.IPNet) (*bridgeConfiguration, error) {
 	defaultNetConfig := &bridgeUDNConfiguration{
 		masqCTMark: ctMarkOVN,
+		subnets:    config.Default.ClusterSubnets,
```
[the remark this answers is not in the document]
jcaamano: Also isUDNNetworkAdvertised

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2286,15 +2297,16 @@ func newGateway(
...
 		if err != nil {
 			return err
 		}
 
 		// resync flows on IP change
 		gw.nodeIPManager.OnChanged = func() {
 			klog.V(5).Info("Node addresses changed, re-syncing bridge flows")
-			if err := gw.openflowManager.updateBridgeFlowCache(subnets, gw.nodeIPManager.ListAddresses(), gw.isPodNetworkAdvertised); err != nil {
+			if err := gw.openflowManager.updateBridgeFlowCache(subnets, gw.nodeIPManager.ListAddresses(),
+				gw.isPodNetworkAdvertised, false); err != nil {
```
jcaamano, line 2309: So it make sense to remove `gw.isPodNetworkAdvertised` from here and pass it instead in `gatewayInitInternal.bridgeForInterface` instead to initialize the proper value for the `bridgeUDNConfiguration` there

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1986,35 +1987,45 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
 	}
 
 	if ofPortPhys != "" {
-		if config.Gateway.DisableSNATMultipleGWs || isPodNetworkAdvertised {
+		if config.Gateway.DisableSNATMultipleGWs || isPodNetworkAdvertised || isUDNNetworkAdvertised {
```
jcaamano, line 1990: We shouldn't use these `PodNetworkAdvertised`,`isUDNNetworkAdvertised` any more. That information comes now from `bridge.netConfig`. Yoyu shoudl just iterate that and based the config from information obtained there.

go-controller/pkg/node/gateway.go
```diff
@@ -472,6 +472,13 @@ func (g *gateway) SetPodNetworkAdvertised(isPodNetworkAdvertised bool) {
 	g.isPodNetworkAdvertisedLock.Lock()
 	defer g.isPodNetworkAdvertisedLock.Unlock()
 	g.isPodNetworkAdvertised = isPodNetworkAdvertised
```
jcaamano, line 474: We can probably get rid of these fields and just use `g.openflowManager.defaultBridge.netConfig`

go-controller/pkg/node/gateway_udn.go
jcaamano: Also, we probably need this to be an atomic, since `isUDNNetworkAdvertised` can change at any time and you will get a race

go-controller/pkg/node/gateway.go
arghosh93: yaa, its a mistake. Will rectify in next push.

go-controller/pkg/node/gateway_udn.go
arghosh93: yaa, its a mistake. Will rectify in next push.

go-controller/pkg/ovn/secondary_layer3_network_controller.go
arghosh93: I am moving both addition and deletion of SNAT rule from addNode to a section which gets executed when gwFailed is set. I hope I am not breaking anything existing.

go-controller/pkg/node/secondary_node_network_controller.go
jcaamano: Yes, below the point of no return which is `ReconcileNetInfo` call.

go-controller/pkg/ovn/secondary_layer3_network_controller.go
jcaamano: No, please don't do that. I just had the question and I can take an answer.

jcaamano: alternatively you can pass a callback

```
return oc.BaseNetworkController.reconcile(
    netInfo,
    func(node string) {oc.addNodeFailed.Store(node, true)}
)
```

so on base network controller
```
func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, prepareNodeReconcile func(string)) error {
	var err error
	var retryNodes []*kapi.Node
	var retryNodeNames []string
	oc.localZoneNodes.Range(func(key, value any) bool {
	...
	...
	for _, node := range retryNodes {
	        prepareNodeReconcile(node)
		err = oc.retryNodes.AddRetryObjWithAddNoBackoff(node)
		if err != nil {
			klog.Errorf("Failed to retry node %s for network %s: %v", node.Name, oc.GetNetworkName(), err)
		}
	}
	if len(retryNodes) > 0 {
		oc.retryNodes.RequestRetryObjs()
	}

	return nil
}
```

arghosh93: Very nice idea. still there would be multiple nodes which would need reconciliation.

jcaamano: all the nodes are being reconciled in the code excerpt above

go-controller/pkg/node/gateway_udn.go
```diff
@@ -654,27 +694,68 @@ func (udng *UserDefinedNetworkGateway) getV6MasqueradeIP() (*net.IPNet, error) {
...
+		masqIPRule, err := udng.constructUDNVRFIPRulesMasqueradeIP(vrfTableId)
+		if err != nil {
+			return nil, nil, err
+		}
+		addIPRules = append(addIPRules, masqIPRule...)
+	}
+	return addIPRules, delIPRules, nil
+}
+
+func (udng *UserDefinedNetworkGateway) constructUDNVRFIPRulesMasqueradeIP(vrfTableId int) ([]netlink.Rule, error) {
```
jcaamano, line 738: I feel like this could be way cleaner. 

I suggest we merge `constructUDNVRFIPRules` and `constructUDNVRFIPRulesMasqueradeIP`  methods into one.

then have as needed
```
addIPRules = append(addIPRules, generateIPRuleForPacketMark(...
```
then as needed
```
masqIPRules := append(masqIPRules, generateIPRuleForMasqIP(..
subnetIPRules := append(subnetIPRules, generateIPRuleForUDNSubnet(...
```
and then at the end
```
switch {
case !isNetworkAdvertised:
    addIPRules = append(addIPRules, masqIPRules...)
    delIPRules = append(delIPRules, masqIPRules...)
default:
    addIPRules = append(addIPRules, subnetIPRules...)
    delIPRules = append(delIPRules, masqIPRules...)
}
```
I guess we decided to leave the masqIPRules around in any case. But I say for now lets remove them.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +810,84 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+// updateUDNVRFIPRule updates IP rules for a network depending on whether the
+// network is advertised or not
+func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRule() error {
+	interfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(udng.networkID))
+	mplink, err := util.LinkByName(interfaceName)
+	if err != nil {
+		return fmt.Errorf("unable to get link for %s, error: %v", interfaceName, err)
+	}
+	vrfTableId := util.CalculateRouteTableID(mplink.Attrs().Index)
+	udnReplyIPRules, deleteIPRule, err := udng.constructUDNVRFIPRules(vrfTableId)
```
jcaamano, line 877: nit: 
s/udnReplyIPRules/addIPRules
s/deleteIPRule/deleteIPRules

go-controller/pkg/node/openflow_manager.go
```diff
@@ -27,24 +28,30 @@ type openflowManager struct {
 	exGWFlowCache map[string][]string
 	exGWFlowMutex sync.Mutex
 	// channel to indicate we need to update flows immediately
-	flowChan chan struct{}
+	flowChan     chan struct{}
+	watchFactory factory.NodeWatchFactory
```
jcaamano, line 32: I don't think we need the watchFactory for anything so you can just remove it

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -850,10 +879,14 @@ func (oc *SecondaryLayer3NetworkController) addNode(node *kapi.Node) ([]*net.IPN
...
 	}
-	if util.IsNetworkSegmentationSupportEnabled() && oc.IsPrimaryNetwork() {
+	if util.IsNetworkSegmentationSupportEnabled() && oc.IsPrimaryNetwork() && !util.IsPodNetworkAdvertisedAtNode(oc, node.Name) {
 		if err := oc.addUDNNodeSubnetEgressSNAT(hostSubnets, node); err != nil {
 			return nil, err
 		}
+	} else if util.IsPodNetworkAdvertisedAtNode(oc, node.Name) {
+		if err := oc.deleteUDNNodeSubnetEgressSNAT(hostSubnets, node); err != nil {
+			return nil, err
+		}
```
jcaamano, line 889: better:
```
if util.IsNetworkSegmentationSupportEnabled() && oc.IsPrimaryNetwork() {
   if !util.IsPodNetworkAdvertisedAtNode(oc, node.Name) {
       ....
   } else {
       ....
   }
}
```

jcaamano (commented):
minor comments

we need to find out why the network-segmentation lanes are failing.

jcaamano (approved):

jcaamano:
holding for this week

arghosh93:
multi-homing lane is failing after my recent rebase on master.
`Summarizing 1 Failure:
  [FAIL] Multi Homing multiple pods connected to the same OVN-K secondary network [It] eventually configures pods that were added to an already existing network before the nad`
This e2e job was added as part of a recent commit[1] and I also did a change[2] to mitigate conflict. I will check this next week.
[1] - https://github.com/ovn-kubernetes/ovn-kubernetes/commit/73208692d382a8568ee39030273cc8a08f2ef357
[2] - https://github.com/ovn-kubernetes/ovn-kubernetes/blob/0dec7a4502786dcaf6bfd829607e7d9be249c77a/go-controller/pkg/ovn/base_network_controller.go#L236-L240

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -489,6 +490,20 @@ func (oc *SecondaryLayer2NetworkController) Stop() {
...
+	reconcilePendingPods := !oc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+	err := util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	if err != nil {
+		klog.Errorf("Failed to reconcile network information for network %s: %v", oc.GetNetworkName(), err)
+	}
+	if reconcilePendingPods {
+		if err := ovnretry.RequeuePendingPods(oc.kube, oc.GetNetInfo(), oc.retryPods); err != nil {
+			klog.Errorf("Failed to requeue pending pods for network %s: %v", oc.GetNetworkName(), err)
+		}
+	}
```
jcaamano, line 503: Why are you doing this?

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -312,6 +313,20 @@ func (oc *SecondaryLocalnetNetworkController) Stop() {
...
+	reconcilePendingPods := !oc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+	err := util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	if err != nil {
+		klog.Errorf("Failed to reconcile network information for network %s: %v", oc.GetNetworkName(), err)
+	}
+	if reconcilePendingPods {
+		if err := ovnretry.RequeuePendingPods(oc.kube, oc.GetNetInfo(), oc.retryPods); err != nil {
+			klog.Errorf("Failed to requeue pending pods for network %s: %v", oc.GetNetworkName(), err)
+		}
+	}
```
jcaamano, line 326: Why are you doing this?

jcaamano (dismissed):

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +798,84 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+
+func (udng *UserDefinedNetworkGateway) Reconcile() {
+	select {
+	case udng.reconcile <- struct{}{}:
+	default:
+	}
+}
+
+func (udng *UserDefinedNetworkGateway) doReconcile() error {
+	klog.Info("Reconciling UDN gateway with updates")
```
trozet, line 836: list which UDN? network name

trozet (changes requested):
only like halfway done. Will finish on Monday

go-controller/pkg/node/gateway.go
```diff
@@ -469,26 +467,16 @@ func (g *gateway) SetDefaultGatewayBridgeMAC(macAddr net.HardwareAddr) {
 }
 
 func (g *gateway) SetPodNetworkAdvertised(isPodNetworkAdvertised bool) {
```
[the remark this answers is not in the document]
jcaamano: Might be more confusing since this `gateway` only exports API to serve the default network, whereas `UserDefinedNetworkGateway` is the one that exports functionality for UDN networks.

go-controller/pkg/node/gateway.go
```diff
@@ -534,7 +520,8 @@ func (g *gateway) updateSNATRules() error {
 	}
 	subnets := util.IPsToNetworkIPs(ipnets...)
 
-	if g.isPodNetworkAdvertised || config.Gateway.Mode != config.GatewayModeLocal {
+	isNetworkAdvertised := g.openflowManager.defaultBridge.netConfig[types.DefaultNetworkName].advertised.Load()
```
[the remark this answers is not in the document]
jcaamano: I would do `gateway.getPodNetworkAdvertised` (unexported). Anyway, note that this is hardcoded for the default network which is guaranteed to exist.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2015,28 +2015,31 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
 	}
 
 	if ofPortPhys != "" {
-		if config.Gateway.DisableSNATMultipleGWs || isPodNetworkAdvertised {
-			// table 1, traffic to pod subnet go directly to OVN
-			output := defaultNetConfig.ofPortPatch
-			if isPodNetworkAdvertised && config.Gateway.Mode == config.GatewayModeLocal {
+		for _, netConfig := range bridge.patchedNetConfigs() {
+			isNetworkAdvertised := netConfig.advertised.Load()
+			if !config.Gateway.DisableSNATMultipleGWs && !isNetworkAdvertised {
```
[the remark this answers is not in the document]
jcaamano: hmmm so if the trick of this is `netConfig is not default`, wouldn't that check have to be done as part of UDN changes? Are we saying `disableSNATMultipleGWs` does not apply to L3 non default networks?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2015,28 +2015,31 @@ func commonFlows(subnets []*net.IPNet, bridge *bridgeConfiguration, isPodNetwork
...
+			if isNetworkAdvertised && config.Gateway.Mode == config.GatewayModeLocal {
 				// except if advertised through BGP, go to kernel
 				// TODO: MEG enabled pods should still go through the patch port
 				// but holding this until
 				// https://issues.redhat.com/browse/FDP-646 is fixed, for now we
 				// are assuming MEG & BGP are not used together
 				output = ovsLocalPort
 			}
-			for _, clusterEntry := range config.Default.ClusterSubnets {
+			for _, clusterEntry := range netConfig.subnets {
```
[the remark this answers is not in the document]
jcaamano: It is the entire subnet, and I guess that is what we want if we were using the entire subnet for the default network as well.

go-controller/pkg/node/gateway.go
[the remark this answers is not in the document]
trozet: Can we still name it SetDefaultPodNetworkAdvertised? Perhaps also change "gateway" to be defaultGateway or something. The latter could be done in another PR. It's not obvious that gateway is only default network.

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
trozet: So disableSNATMultipleGWs was not originally intended to apply to UDN. It was for MEG to signal that if a pod has multiple external gateways, do not SNAT the egress traffic. That being said, I think it is good how it is evolving to cover the UDN case with BGP if pod ips are advertised. So I think disableSNATMultipleGWs *DOES* apply to UDN.

But like I said in my comment if netConfig is not default (so its UDN) && !isNetworkAdvertised && disableSNATMultipleGWs then we should not program the flows for direct pod ingress

go-controller/pkg/node/gateway_udn.go
arghosh93: Ack

go-controller/pkg/node/gateway_udn.go
```diff
@@ -717,3 +798,84 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
...
+}
+
+func (udng *UserDefinedNetworkGateway) Reconcile() {
+	select {
+	case udng.reconcile <- struct{}{}:
+	default:
+	}
+}
+
+func (udng *UserDefinedNetworkGateway) doReconcile() error {
```
trozet, line 835: how come this reconcile was implemented asynchronously? It looks like the calls inside of it are not expensive or blocked for long periods of time. The side effect of this being async is on node start up, the syncall method for networks will not be blocked on the reconcile.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -183,6 +183,65 @@ type BaseNetworkController struct {
...
+	if err != nil {
+		return fmt.Errorf("failed to reconcile network %s: %w", oc.GetNetworkName(), err)
+	}
+	reconcileRoutes := oc.routeImportManager != nil && oc.routeImportManager.NeedsReconciliation(netInfo)
+	reconcilePendingPods := !oc.IsDefault() && !oc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+
+	// set the new NetInfo, point of no return
+	err = util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	if err != nil {
+		klog.Errorf("Failed to reconcile network information for network %s: %v", oc.GetNetworkName(), err)
```
trozet, line 215: should you still continue past this point? If so, why?

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -855,8 +884,14 @@ func (oc *SecondaryLayer3NetworkController) addNode(node *kapi.Node) ([]*net.IPN
...
 	}
 	if util.IsNetworkSegmentationSupportEnabled() && oc.IsPrimaryNetwork() {
-		if err := oc.addUDNNodeSubnetEgressSNAT(hostSubnets, node); err != nil {
-			return nil, err
+		if !util.IsPodNetworkAdvertisedAtNode(oc, node.Name) {
+			if err := oc.addUDNNodeSubnetEgressSNAT(hostSubnets, node); err != nil {
+				return nil, err
+			}
+		} else {
+			if err := oc.deleteUDNNodeSubnetEgressSNAT(hostSubnets, node); err != nil {
```
trozet, line 892: i guess you need this for update. The function is a little confusing cause it's called addNode, we should probably rename it to addOrUpdateNode

go-controller/pkg/node/gateway_udn.go
jcaamano: This is not used on startup. Everything that needs to happen on startup happens synchronously in `AddNetwork`.

go-controller/pkg/node/gateway.go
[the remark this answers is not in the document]
jcaamano: I am ok with `SetDefaultPodNetworkAdvertised`

go-controller/pkg/ovn/secondary_layer2_network_controller.go
arghosh93: This was discussed and resolved.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
arghosh93: This was discussed and resolved.

go-controller/pkg/node/gateway.go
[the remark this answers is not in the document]
arghosh93: does below function definition seems fair? I am naming it GetDefaultPodNetworkAdvertised to keep it consistent with SetDefaultPodNetworkAdvertised.
`func (g *gateway) GetDefaultPodNetworkAdvertised() bool {
	return g.openflowManager.defaultBridge.netConfig[types.DefaultNetworkName].advertised.Load()
}`

go-controller/pkg/ovn/secondary_layer3_network_controller.go
arghosh93: I have not yet change the function name to keep it consistent with all other network controller's(ex: DefaultNetworkController) addNode function name. Moreover addNode gets called by addUpdateLocalNodeEvent(). Should I still go ahead and rename the function name?

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
trozet: yeah sorry my comment was wrong. We should *NOT* consider DisableSNATMultipleGWs for UDN right now

go-controller/pkg/node/gateway_udn.go
jcaamano: It would be

func (c *networkController) Start() -> c.syncAll -> err := c.syncNetwork(network.GetNetworkName()) -> err = err = c.ensureNetwork(want) -> err := networkController.Start() -> oc.gateway.AddNetwork()

go-controller/pkg/ovn/secondary_layer3_network_controller.go
trozet: you can just leave it for now. I'll refactor it later.

go-controller/pkg/node/gateway_udn.go
jcaamano: Yeah, I think we should. In general terms Start should just get things started and have minimum surface for failure. However that requires not having dependencies to do things synchronously. Too much for this PR.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -827,6 +834,8 @@ func (oc *SecondaryLayer2NetworkController) gatewayManagerForNode(nodeName strin
 		return oc.newGatewayManager(nodeName)
 	} else {
 		gwManager, isGWManagerType := obj.(*GatewayManager)
+		// reinstate GatewayManager's netInfo which might have changed during reconciliation
+		gwManager.netInfo = oc.GetNetInfo()
```
jcaamano, line 838: You don't need to do this. `gwManager.netInfo` is already `oc.GetNetInfo()` which is an interface/pointer that is never replaced. It's just updated internally.

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -1090,6 +1128,8 @@ func (oc *SecondaryLayer3NetworkController) gatewayManagerForNode(nodeName strin
 		return oc.newGatewayManager(nodeName)
 	} else {
 		gwManager, isGWManagerType := obj.(*GatewayManager)
+		// reinstate GatewayManager's netInfo which might have changed during reconciliation
+		gwManager.netInfo = oc.GetNetInfo()
```
jcaamano, line 1132: You don't need to do this. gwManager.netInfo is already oc.GetNetInfo() which is an interface/pointer that is never replaced. It's just updated internally.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
arghosh93: Ack

go-controller/pkg/ovn/secondary_layer3_network_controller.go
arghosh93: Ack

trozet (approved):

PR #4930, opened 2024-12-18
udn host isolation: fix initialSync.

npinaeva:
Handle host-network pods as default network.
Don't return per-pod errors on startup.


npinaeva:
new flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4932

trozet (approved):

PR #4933, opened 2024-12-19
(C)UDN CRD: update CEL and add IPAM section

npinaeva:
based on https://github.com/openshift/enhancements/pull/1725
Whoever reviews that, I beg you to run some manual tests by applying CRs with various wrong and right configs checking the CEL rules work as expected. To do that, run a kind cluster with these changes, create namespace `demo`, then apply and delete various CR versions, e.g.
```
apiVersion: k8s.ovn.org/v1
kind: UserDefinedNetwork
metadata:
  name: db-network
  namespace: demo
spec:
  topology: Layer2
  layer2:
    role: Primary
    mtu: 1400
    ipam: 
      mode: Disabled
      lifecycle: Persistent
```
should return
```
The UserDefinedNetwork "db-network" is invalid: 
* spec.layer2: Invalid value: "object": Disabled ipam.mode is only supported for Secondary network
* spec.layer2.ipam: Invalid value: "object": lifecycle Persistent is only supported when ipam.mode is Enabled
```


npinaeva:
cc @maiqueb @ormergi this is the IPAM changes PR, but it has a surprise part with some CEL tweaks :)

maiqueb:
> based on [openshift/enhancements#1725](https://github.com/openshift/enhancements/pull/1725) Whoever reviews that, I beg you to run some manual tests by applying CRs with various wrong and right configs checking the CEL rules work as expected. 

It *was* a trap !!! 

> To do that, run a kind cluster with these changes, create namespace `demo`, then apply and delete various CR versions, e.g.
> 
> ```
> apiVersion: k8s.ovn.org/v1
> kind: UserDefinedNetwork
> metadata:
>   name: db-network
>   namespace: demo
> spec:
>   topology: Layer2
>   layer2:
>     role: Primary
>     mtu: 1400
>     ipam: 
>       mode: Disabled
>       lifecycle: Persistent
> ```
> 
> should return
> 
> ```
> The UserDefinedNetwork "db-network" is invalid: 
> * spec.layer2: Invalid value: "object": Disabled ipam.mode is only supported for Secondary network
> * spec.layer2.ipam: Invalid value: "object": lifecycle Persistent is only supported when ipam.mode is Enabled
> ```

Could you help me out by pasting the most complex UDN you can come up w/ ? JoinSubnets and everything ...

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
```diff
@@ -128,16 +128,44 @@ type Layer2Config struct {
 	// +optional
 	JoinSubnets DualStackCIDRs `json:"joinSubnets,omitempty"`
 
-	// IPAMLifecycle controls IP addresses management lifecycle.
+	// IPAM section contains IPAM-related configuration for the network.
+	// +optional
+	IPAM *IPAMSpec `json:"ipam,omitempty"`
```
maiqueb, line 133: nit: the spec suffix kind of makes me hope for a status sub-resource.

Maybe we could call this `IPAMConfig` ? 

Totally opinionated, and you can disregard it.

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
```diff
@@ -128,16 +128,44 @@ type Layer2Config struct {
...
+	// `Enabled` means OVN-Kubernetes will apply IP configuration to the SDN infrastructure and it will also assign IPs
+	// from the selected subnet to the individual pods.
+	// `Disabled` means OVN-Kubernetes will only assign MAC addresses and provide layer 2 communication, letting users
+	// configure IP addresses for the pods.
+	// `Disabled` is only available for Secondary networks.
+	// By disabling IPAM, any Kubernetes features that rely on selecting pods by IP will no longer function
+	// (such as network policy, services, etc). Additionally, IP port security will also be disabled for interfaces attached to this network.
+	// Defaults to `Enabled`.
+	// +optional
+	Mode IPAMMode `json:"mode,omitempty"`
```
maiqueb, line 149: pardon my ignorance, but couldn't we set a default using kubebuilder ?

maiqueb (commented):
I just have some borderline idiotic questions. 

Keep in me I'm really not that used to creating Kubernetes APIs ... The few I've created are extremely simple and predate CEL ... 

Your friendly neighborhood code reviewer (me) will get back to this and try to mess up your rules once he can find some time.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
```diff
@@ -134,7 +134,7 @@ func renderCNINetworkConfig(networkName, nadName string, spec SpecGetter) (map[s
 		cfg := spec.GetLayer2()
 		netConfSpec.Role = strings.ToLower(string(cfg.Role))
 		netConfSpec.MTU = int(cfg.MTU)
-		netConfSpec.AllowPersistentIPs = cfg.IPAMLifecycle == userdefinednetworkv1.IPAMLifecyclePersistent
+		netConfSpec.AllowPersistentIPs = cfg.IPAM != nil && cfg.IPAM.Lifecycle == userdefinednetworkv1.IPAMLifecyclePersistent
```
ormergi, line 137: Shouldn't it check that IPAM Mode is true?
Do we need complimentary validation on the controller side, similar to other fields?

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
```diff
@@ -273,10 +273,12 @@ var _ = Describe("NetAttachDefTemplate", func() {
...
 				Topology: udnv1.NetworkTopologyLayer2,
 				Layer2: &udnv1.Layer2Config{
-					Role:          udnv1.NetworkRolePrimary,
-					Subnets:       udnv1.DualStackCIDRs{"192.168.100.0/24", "2001:dbb::/64"},
-					MTU:           1500,
-					IPAMLifecycle: udnv1.IPAMLifecyclePersistent,
+					Role:    udnv1.NetworkRolePrimary,
+					Subnets: udnv1.DualStackCIDRs{"192.168.100.0/24", "2001:dbb::/64"},
+					MTU:     1500,
+					IPAM: &udnv1.IPAMSpec{
```
ormergi, line 279: Shouldn't spec specify both `IPAM.Mode=Enabled` & `Lifecycle=Persistent`?
Isn't it the intention of the API change?

Should we have additional negative tests, for covering case when IPAM.Mode=Disable & Lifecycle=Persistent?

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
```diff
@@ -128,16 +128,44 @@ type Layer2Config struct {
...
+	// Mode controls how much of the IP configuration will be managed by OVN.
+	// `Enabled` means OVN-Kubernetes will apply IP configuration to the SDN infrastructure and it will also assign IPs
+	// from the selected subnet to the individual pods.
+	// `Disabled` means OVN-Kubernetes will only assign MAC addresses and provide layer 2 communication, letting users
+	// configure IP addresses for the pods.
+	// `Disabled` is only available for Secondary networks.
+	// By disabling IPAM, any Kubernetes features that rely on selecting pods by IP will no longer function
+	// (such as network policy, services, etc). Additionally, IP port security will also be disabled for interfaces attached to this network.
+	// Defaults to `Enabled`.
+	// +optional
```
ormergi, line 148: This field shouldn't be mandatory?

ormergi (commented):
Hi, please see my inline comments

I still need to go over the new CEL rules changes.
(We really need these  CEL validation tests :grimacing:)

dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
```diff
@@ -113,6 +139,7 @@ spec:
                           It is not recommended to set this field without explicit need and understanding of the OVN network topology.
                           When omitted, the platform will choose a reasonable default which is subject to change over time.
                         items:
+                          maxLength: 43
```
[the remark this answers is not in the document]
npinaeva: also, checked with https://github.com/kubernetes-sigs/network-policy-api/blob/main/apis/v1alpha1/shared_types.go#L151

dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
```diff
@@ -114,9 +114,17 @@ spec:
...
+                          x-kubernetes-validations:
+                          - message: CIDR is invalid
+                            rule: isCIDR(self)
                         maxItems: 2
                         minItems: 1
                         type: array
+                        x-kubernetes-validations:
+                        - message: When 2 CIDRs are set, they must be from different
+                            IP families
+                          rule: size(self) != 2 || !isCIDR(self[0]) || !isCIDR(self[1])
```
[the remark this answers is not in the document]
npinaeva: yeah, that is a strange logic of CEL. If the condition of CEL is not true, an error will be returned, but we don't want to fail all conditions e.g. if the CIDR is wrong. So isCIDR will be validated separately by the CIDR struct CEL ans say "wrong CIDR format", but if ^ this CEL fails it will say "When 2 CIDRs are set, they must be from different  IP families", together with the wrong cidr message, which makes it more confusing.
So we only want to fail a given CEL when a specific problem happens (considering other failures are covered by other CELs), in this case we kinda "return true" is the CIDR is wrong, because we just can't check which ip families exist AND we know worng cidr is covered by another CEL

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
npinaeva: With current CEL there is no way to have persistentIPs enabled with disabled IPAM, but I guess I can add the same check in the code

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
npinaeva: Enables is the default value, so you don't have to set it, but if you do put Disabled there, Persistent lifecycle won't be allowed anymore. I have added a validation test to NAD-template

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
npinaeva: good point, +1 easy change, -1 confused dev

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
```diff
@@ -128,16 +128,44 @@ type Layer2Config struct {
...
-	// IPAMLifecycle controls IP addresses management lifecycle.
+	// IPAM section contains IPAM-related configuration for the network.
+	// +optional
+	IPAM *IPAMSpec `json:"ipam,omitempty"`
+}
+
+// +kubebuilder:validation:XValidation:rule="!has(self.lifecycle) || self.lifecycle != 'Persistent' || !has(self.mode) || self.mode == 'Enabled'", message="lifecycle Persistent is only supported when ipam.mode is Enabled"
+// +kubebuilder:validation:MinProperties=1
+type IPAMSpec struct {
+	// Mode controls how much of the IP configuration will be managed by OVN.
```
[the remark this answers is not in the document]
npinaeva: the phrasing is copied from the enhancement, so I would like to leave it as is unless some serious problems are outlined.
About the newline - I have never payed enough attention to it, but how did you decide where they should be present?

npinaeva:
> It _was_ a trap !!!

I promise it wasn't supposed to be :D you can easily skip the CEL-changing commit especially if you are not an expert in CEL and just say it in the comment.

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
```diff
@@ -92,11 +86,10 @@ type Layer3Subnet struct {
 	HostSubnet int32 `json:"hostSubnet,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="self.role != 'Primary' || has(self.subnets) && size(self.subnets) > 0", message="Subnets is required for Primary Layer2 topology"
+// +kubebuilder:validation:XValidation:rule="self.role != 'Primary' || has(self.subnets)", message="Subnets is required for Primary Layer2 topology"
```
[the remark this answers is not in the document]
npinaeva: correct, but we do check min length here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6c366bb49d6fa57c80d6b39834952ce1d4d9dbb6/go-controller/pkg/crd/userdefinednetwork/v1/shared.go#L166
you can try applying zero-length list and see what happens :)

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
maiqueb: you cannot change the default value without breaking the API afaiu.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
ormergi: I thought the idea was to have explicit IPAM mode whether it enable or disabled.
But it seem one would not have to set mode=enabled and just define subnet, similar to what we have today.
Did I got it right?

go-controller/pkg/crd/userdefinednetwork/v1/shared.go
npinaeva: you kinda can if you can figure out a "smooth upgrade" https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#defaulting

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
npinaeva: yes, and another "new" effect is that you can't set an empty subnet anymore without explicitly setting disabled IPAM mode

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
```diff
@@ -183,6 +183,19 @@ var _ = Describe("NetAttachDefTemplate", func() {
 			&udnv1.UserDefinedNetwork{Spec: udnv1.UserDefinedNetworkSpec{
 				Topology: udnv1.NetworkTopologyLayer3, Layer2: &udnv1.Layer2Config{}}},
 		),
+		Entry("UDN, invalid IPAM lifecycle",
```
maiqueb, line 186: maybe be a bit more clear here: AFAIU, the IPAM lifecycle is invalid *because* IPAM is disabled.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
maiqueb, line 1: maybe we should add some MTU related unit tests ? i.e. MTU = 1000 on an IPv6 network, expect ka-boom.

I'd also appreciate seeing a unit test where IPAM is disabled, but we define subnets / excludedCIDRs.

maiqueb (changes requested):
Requesting a couple more unit-tests.

Other than that, LGTM ! 

Real nice work.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
npinaeva: MTU validation (I think) is only done at the CEL level, and not in the code. Unfortunately we don't yet have CEL-related unit tests, which is what this pr is mostly about. We only have unit tests for generic NAD-related checks, so I don't think I can add an MTU-validation test right now. The new unit test was added for the only code change from https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4933/files#diff-9e3b739334a5197a7018eaec856b6705c2378e0095dd1d9d28376c12c9f766e1R140

maiqueb: Hm. The MTU part makes sense! 

But would it make sense to add a unit test where IPAM is disabled, but we define subnets / excludedCIDRs ? I'm fine if you prefer this on a follow-up (not saying it has to be you to push it ...).

npinaeva: I originally didn't want to duplicate CEL logic in the code, but maybe that is fine. I will add that validation and unit tests to the code then

maiqueb: hey, if that's the cost to pay, let's *not* do it. 

Like, I thought it would be simpler to check it ... It the tests force us to write more code, we're not in a good place.

npinaeva: I am still not 100% convinced, but I think it is reasonable to duplicate some of the CEL validations in the code. A "bad" object should never reach the code, but if you build it manually e.g. for some unit-testing, it may be useful to catch wrong configs I think

maiqueb: If you're willing, I won't stand in your way. 

Let's have that extra coverage then. 

Thanks for the discussion.

maiqueb (dismissed):
/lgtm

Nice ! 

Thanks for adding that test :)

npinaeva:
re-opened because of the unit test flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4968

tssurya:
@trozet : I think this is ready for your review now..PTAL when you have a chance (I know you are busy with other things sorry for the reping)

trozet (approved):

trozet:
@npinaeva did you check to see if this will break downstream origin tests for downstream merge? I don't see any e2e tests changed, so I'm guessing not. Should we have e2e tests for this?

npinaeva:
no changes in e2e only means that we don't have tests for the persisten IPs and disabled IPAM (probably would be nice to have those, but I would say such e2es are more related to the corresponding features than to the API itself). Otherwise the functionality hasn't changed, most changes are done on the CEL side, for which we unfortunately don't have any tests yet.
There is a chance we have some different origin tests, that could brake. need to check that part

PR #4940, opened 2024-12-23
Add GetInterfaceUDNs to convert interface name to (C)UDN ns+name.

npinaeva:
Waiting for feedback from @msherif1234


go-controller/observability-lib/sampledecoder/sample_decoder.go
```diff
@@ -291,3 +291,49 @@ func (d *SampleDecoder) DeleteCollector(collectorID int) error {
...
+// UDN namespace+name are joined by "/", CUDN will just have a name.
+func (d *SampleDecoder) GetInterfaceUDNs() (map[string]string, error) {
+	res := map[string]string{}
+	ifaces := []*ovsdb.Interface{}
+	err := d.ovsdbClient.List(context.Background(), &ifaces)
+	if err != nil {
+		return nil, fmt.Errorf("failed listing interfaces: %w", err)
+	}
+	for _, iface := range ifaces {
+		if iface.ExternalIDs["iface-id-ver"] == "" {
```
dceara, line 328: While this is probably correct, the "most correct" check would have to validate iface-id as well.

dceara (commented):
This looks correct to me (with a minor nit).

Thanks,
Dumitru

go-controller/observability-lib/sampledecoder/sample_decoder.go
npinaeva: ah you mean check both `iface-id-ver` and `iface-id`?

dceara: Yes, that's what I meant.

npinaeva: i prefer both, because `iface-id` also exists on non-pod interfaces, let me update

dceara (approved):
LGTM, thanks!

msherif1234:
I tested the API using netobserv stack and I can get the udb tags for pod's host side end as expected
image (3)

npinaeva:
Failure is network-segmentation test and not related to the changes in this PR

trozet (approved):

PR #4946, opened 2025-01-07
RouteAdvertisements: fail if DisableMP is unset

jcaamano:
If MultiProtocol is enabled (default) then a BGP session carries prefixes of both IPv4 and IPv6 families. Our problem is that with an IPv4 session, FRR can incorrectly pick the masquerade IPv6 address (instead of the real address) as next hop for IPv6 prefixes and that won't work. Note that with a dedicated IPv6 session that can't happen since FRR will use the same address that was used to stablish the session. Let's enforce the use of DisableMP for now.

### DIFF
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -506,13 +506,16 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements) 
 		matchedNetworks := sets.New[string]()
 		for _, frrConfig := range frrConfigs {
 			// generate FRRConfiguration for each source FRRConfiguration/node combination
-			new := c.generateFRRConfiguration(
+			new, err := c.generateFRRConfiguration(
 				ra,
 				frrConfig,
 				nodeName,
 				selectedNetworks,
 				matchedNetworks,
 			)
+			if err != nil {
+				return nil, nil, err
+			}
 			if new == nil {
 				// if we got nil, we didn't match any VRF
 				return nil, nil, fmt.Errorf("%w: FRRConfiguration %q selected for node %q has no VRF matching the RouteAdvertisements target VRF or any selected network",
@@ -538,7 +541,7 @@ func (c *Controller) generateFRRConfiguration(
 	nodeName string,
 	selectedNetworks *selectedNetworks,
 	matchedNetworks sets.Set[string],
-) *frrtypes.FRRConfiguration {
+) (*frrtypes.FRRConfiguration, error) {
 	routers := []frrtypes.Router{}
 	advertisements := sets.New(ra.Spec.Advertisements...)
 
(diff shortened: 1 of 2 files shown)


go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -596,16 +599,30 @@ func (c *Controller) generateFRRConfiguration(
...
-			advertisePrefixes := advertisePrefixes
-			receivePrefixes := receivePrefixes
+			// If MultiProtocol is disabled (not by default) then a BGP session
+			// carries prefixes of both IPv4 and IPv6 families. Our problem is
+			// that with an IPv4 session, FRR can incorrectly pick the
+			// masquerade IPv6 address (instead of the real address) as next hop
+			// for IPv6 prefixes and that won't work. Note that with a dedicated
+			// IPv6 session that can't happen since FRR will use the same
+			// address that was used to stablish the session. Let's disallow the
+			// use of DisableMP for now.
```
trozet, line 609: this doesn't seem right to me. I think ipv6 is carried separately always over its own session. The only exception is vpnv4 and vpvn6 routes will be sent via ipv4 with MP support.

jcaamano: @fedepaol can you help here? Did I get the opposite meaning of DisableMP?

fedepaol: > @fedepaol can you help here? Did I get the opposite meaning of DisableMP?

TL;DR: yes. Disable MP means "don't advertise v6 over v4", because frr doesn't have a proper way to advertise the next hop for v6 in case of a session established over a v4 bgp connection.

This behavior (and the consequent parameter) didn't start with the right foot, and despite it being a breaking change, it might make sense to make it the default in frr-k8s.

In retrospective, even calling it MP is wrong, as it has a much broader meaning than just v6 over v4 :man_shrugging:

jcaamano: Changed to be the other way around then. A bit wicked that the default value is what does not work but so be it.

trozet: I don't understand why we would disable MP. We need it for evpn support. Isn't the answer to just use a routemap to force the nexthop to be set if a user is advertising v6 over a v4 connection?

jcaamano: I don't know how a route map would fix the issue.

This other thing might
```
set ipv6 next-hop global <whatever>
```
but I haven't tried it and I am not sure when that will happen.

If you are willing to make this a blocker bug for BGP, then fine. Otherwise I think we need this in until we come around to something better.

jcaamano: Without FRR-K8s being involved in the solution, I am not sure we can gracefully generate the config from ovnk without abusing raw config.

fedepaol: > I don't understand why we would disable MP. We need it for evpn support. Isn't the answer to just use a routemap to force the nexthop to be set if a user is advertising v6 over a v4 connection?

again, it's bad wording as I wrote above. MP is enabled by default. What this means is, don't advertise v6 over v4 (and viceversa).
As you wrote, there can be ways to make it work. It's just not implemented now.
In metallb and frr-k8s, so far it's be fine to say "if you want dual stack, you need to have two separate sessions, one for v4 and the other for v6".

trozet (approved):
This is a bit confusing with the semantics of what disableMP means, but for now it makes sense to map route advertisement to the ip family of the neighbor.

PR #4947, opened 2025-01-08
Fix WatchNamespaces for UDNs

kyrtapz:
When multinetwork policies support is disabled WatchNamespaces should run to completion for primary UDNs.
This was not happening because a primary UDN is also a secondary network.

cc: @dceara

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_namespace.go
+++ b/go-controller/pkg/ovn/base_network_controller_namespace.go
@@ -68,18 +68,21 @@ func getNamespaceAddrSetDbIDs(namespaceName, controller string) *libovsdbops.DbO
 	})
 }
 
+func (bnc *BaseNetworkController) shouldWatchNamespaces() bool {
+	// Watch namespaces only if one of the following conditions is met:
+	// - The network is the default network.
+	// - The network is primary, and network segmentation is enabled.
+	// - The network is secondary, and multi NetworkPolicies are enabled.
+	return bnc.IsDefault() ||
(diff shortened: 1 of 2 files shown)


tssurya:
cc @pperiyasamy

go-controller/pkg/ovn/base_network_controller_namespace.go
```diff
@@ -77,8 +77,9 @@ func (bnc *BaseNetworkController) WatchNamespaces() error {
 		return nil
 	}
 
-	if bnc.IsSecondary() && !util.IsMultiNetworkPoliciesSupportEnabled() {
-		// For secondary networks, we don't have to watch namespace events if
+	if (!bnc.IsPrimaryNetwork() && util.IsNetworkSegmentationSupportEnabled()) &&
```
tssurya, line 80: this match got complicated lol specially with the first match on L74 cause we seem to need to do the opposite of that here...

dceara (approved):
LGTM, thanks!

go-controller/pkg/ovn/base_network_controller_namespace.go
kyrtapz: I've modified the code to make it more readable.

trozet (approved):

go-controller/pkg/ovn/base_network_controller_namespace.go
tssurya: lgtm

PR #4953, opened 2025-01-10
Helm static adminnetwork policies crd

flavio-fernandes:
Helm: Enable Admin Network Policy by Default

Instead of pulling the CRDs for the Admin Network Policy from
the script, include them as part of helm/ovn-kubernetes/crds. The
net effect is the same, as we already had the links hardcoded to
apply them:

```
- run_kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/network-policy-api/v0.1.5/config/crd/experimental/policy.networking.k8s.io_adminnetworkpolicies.yaml
- run_kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/network-policy-api/v0.1.5/config/crd/experimental/policy.networking.k8s.io_baselineadminnetworkpolicies.yaml
```


tssurya:
We actually want to use links and not maintain copies for ANP. When API changes its hard to keep track and update the cr, the api changes often because its alpha, lets see the of apply neatness and not install crs by copying them.

girishmg:
Helm packages all-things-ovn-k8s-CNI -- CRDs, Namespaces, roles, rolebindings, and so on.

The order in which it applies these objects are:  (see: https://helm.sh/docs/intro/using_helm/)
1. CRDs
2. Namespace
3. <bunch-of-other-things>
4. Clusterrole*
5. Role*

So, if we refer to ANP and BANP objects in (4) and (5) above, then Helm install will fail if ANP and BANP is not applied yet.

So, we will need to include everything we reference to in one packaging.

It is better to include a CRD at specific version  along with the code that implements that CRD at that version. If I separate it out, then the code and the API will diverge. Also, the helm installation will not work.

Let us help our end users in using our CNI by packing everything in Helm and have `helm install` as one command to experience our CNI.

PR #4955, opened 2025-01-12
Fix Kind IPV6 deplyoment with RUN_IN_CONTAINER

yboaron:
## 📑 Description
Currently, IPV6 deployment of Kind with RUN_IN_CONTAINER flag [1] fails, the root cause seems to be setting of IPV4 address of master node in the server field in Kubeconfig [2] instead of the IPV6 address.

This PR sets the server IP address in Kubeconfig considering IP family type.

[1] KIND_IPV4_SUPPORT=false KIND_IPV6_SUPPORT=true RUN_IN_CONTAINER=true ./contrib/kind.sh 
[2]
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/contrib/kind.sh#L1078

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -1072,11 +1072,16 @@ docker_create_second_interface() {
 }
 
 # run_script_in_container should be used when kind.sh is run nested in a container
-# and makes sure the control-plane node is rechable by substituting 127.0.0.1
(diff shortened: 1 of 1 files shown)


pperiyasamy (approved):

yboaron:
cc @flavio-fernandes  ^^

PR #4834, opened 2024-11-13
kubevirt, e2e, udn: Add north south traffic and tcp connection checks for e/w and n/s.

qinqon:
## 📑 Description
This PR add e2e test to check tcp connection are not broken during live migration for UDN and north/south traffic for primary UDN.

TODO:
- [x] Don't use --one-off for iperf is not needed and complicate the tests.
- [x] Localnet e/w tcp connection get broken
   - Localnet + IPAM live migration is broken until we don't fix LSP enabled for localnet, let's mark the test.
- [x] non IC + local gw + layer2 + egress not working.
   - We should skip. 
- [x] n/s tcp connection broken at live migration with ipv6
   - Fixed with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4774  
- [x] n/s tcp connection check

## Additional Information for reviewers
Depends-on:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4773
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4774

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

PR #4960, opened 2025-01-14
kubevirt, pudn: Keep SNAT after live migration

qinqon:
## 📑 Description
After kubevirt live migration the migration source pod is deleted and this end up deleting the per pod snat that allow north south traffic on that node also the migrated VM has still the default gw mac address pointing to the previous node and not to the node where the VM is running, that get updated after ip neigh reconciliation and that depends on the guest. This end up with VMs not being able to do north south traffic after live migration.

This change keep the SNAT at all he nodes where the VM is being migrated from so north south traffic works even if the VM default gw mac is pointing to previous nodes.

This will cover only IC since we don't support live migration with primary UDNs layer2 for non IC.

## ✅ Checks

- [ ] My code requires tests

## How to verify it
TODO:
- [ ] unit test 

This should be tested with at snat per pod IC cluster and e2e at PR:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4834

### DIFF
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -479,3 +479,17 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod) 
 	}
 	return &status, nil
 }
+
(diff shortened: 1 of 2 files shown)


qinqon:
snat per pod is going to be changed to snat per subnet, this change is not longer needed.

PR #4963, opened 2025-01-15
cudn,crd: Block namespace-selector with zero-rules

ormergi:
## 📑 Description

Nothing prevents from creating one to create primary CUDN CR selecting all namespaces, unintentionally by specifying empty namespace-selector.
Such CUDN instance could affect all pod in the cluster including the system components (e.g.:kube-system pods, cluster API server, etcd).

Reverting this state is almost impossible w/o causing disruption to cluster components and workloads, because CUDN deletion will be blocked until all connected pods are deleted (pod created after the CUDN).

To avoid such scenarios,validate CUDN namespace-selector, rejecting CUDNs with empty namespace-selectors, at the controller and API level.

Having a CUDN that affects all namespaces in the cluster is still possible by labeling all namespaces with a common label, and specify that label in the CUDN namespace-selector.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -617,6 +617,9 @@ func (c *Controller) getSelectedNamespaces(sel metav1.LabelSelector) (sets.Set[s
 	if err != nil {
 		return nil, fmt.Errorf("failed to create label-selector: %w", err)
 	}
+	if labelSelector.Empty() {
+		return nil, fmt.Errorf("namespace selector is invalid: selecting all namespaces is not allowed")
+	}
 	selectedNamespacesList, err := c.namespaceInformer.Lister().List(labelSelector)
 	if err != nil {
 		return nil, fmt.Errorf("failed to list namespaces: %w", err)
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
@@ -740,6 +740,60 @@ var _ = Describe("User Defined Network Controller", func() {
 					Expect(kerrors.IsNotFound(err)).To(Equal(true))
 				}
 			})
+
+			DescribeTable("should report error in status, when namespace-selector",
+				func(selector metav1.LabelSelector, expectedMessage string) {
+					cudn := &udnv1.ClusterUserDefinedNetwork{
+						ObjectMeta: metav1.ObjectMeta{Name: "test-cudn"},
+						Spec:       udnv1.ClusterUserDefinedNetworkSpec{NamespaceSelector: selector},
+					}
+					c = newTestController(template.RenderNetAttachDefManifest, cudn)
+					Expect(c.Run()).To(Succeed())
+
+					Eventually(func() []metav1.Condition {
+						cudn, err := cs.UserDefinedNetworkClient.K8sV1().ClusterUserDefinedNetworks().Get(context.Background(), cudn.Name, metav1.GetOptions{})
+						Expect(err).NotTo(HaveOccurred())
+						return normalizeConditions(cudn.Status.Conditions)
+					}).Should(Equal([]metav1.Condition{{
+						Type:    "NetworkCreated",
+						Status:  "False",
+						Reason:  "NetworkAttachmentDefinitionSyncError",
+						Message: expectedMessage,
+					}}))
+				},
+				Entry("is empty",
+					metav1.LabelSelector{},
+					"failed to get selected namespaces: namespace selector is invalid: selecting all namespaces is not allowed",
+				),
+				Entry("has nil match-labels",
+					metav1.LabelSelector{MatchLabels: nil},
+					"failed to get selected namespaces: namespace selector is invalid: selecting all namespaces is not allowed",
+				),
+				Entry("has empty match-label",
+					metav1.LabelSelector{MatchLabels: map[string]string{}},
+					"failed to get selected namespaces: namespace selector is invalid: selecting all namespaces is not allowed",
+				),
+				Entry("has single match-label rule with empty key ",
+					metav1.LabelSelector{MatchLabels: map[string]string{"": ""}},
+					"failed to get selected namespaces: failed to create label-selector: key: Invalid value: \"\": name part must be non-empty; name part must consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]')",
+				),
+				Entry("has nil match expressions",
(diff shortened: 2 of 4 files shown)


maiqueb:
We've discussed out-of-band to provide *somewhere* a place for the admin to configure a list of namespaces that couldn't be referenced here. The use case would be for the `openshift-*` namespaces to be ignored.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
```diff
@@ -740,6 +740,60 @@ var _ = Describe("User Defined Network Controller", func() {
...
 			})
+
+			DescribeTable("should report error in status, when namespace-selector",
+				func(selector metav1.LabelSelector, expectedMessage string) {
+					cudn := &udnv1.ClusterUserDefinedNetwork{
+						ObjectMeta: metav1.ObjectMeta{Name: "test-cudn"},
+						Spec:       udnv1.ClusterUserDefinedNetworkSpec{NamespaceSelector: selector},
+					}
+					c = newTestController(template.RenderNetAttachDefManifest, cudn)
+					Expect(c.Run()).To(Succeed())
```
maiqueb, line 751: nit: couldn't you inline these ? Are you actually using `c` anywhere else ? Just curious, you can leave it as is.

maiqueb (approved):
Looks good.

Just left a dumb nit you can ignore.

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
ormergi: Unfortunately we cannot inline it because there is some AfterEach code that shutdown the tested controller instance.
Without this cleanup unit-tests would fail due to race errors (e.g.: multiple places touches OVN config).

maiqueb: Right, thanks for explaining; I didn't check the entire code.

go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
```diff
@@ -26,6 +26,7 @@ type ClusterUserDefinedNetworkSpec struct {
 	// NamespaceSelector Label selector for which namespace network should be available for.
 	// +kubebuilder:validation:Required
 	// +required
+	// +kubebuilder:validation:XValidation:rule="self.matchLabels.size() > 0 || self.matchExpressions.size() > 0", message="namespace selector must have at least one rule"
```
npinaeva, line 29: it may be better to add has() check, like
`has(self.matchLabels) && self.matchLabels.size() > 0 || has(self.matchExpressions) && self.matchExpressions.size() > 0` so that the error message is more clean.
For example, if you apply the following selector
```
namespaceSelector:
    matchExpressions: []
```
you will get this error
`spec.namespaceSelector: Invalid value: "object": no such key: matchLabels evaluating rule: namespace selector must have at least one rule` which looks like matchLabels is required, but it is not the case, with the suggested rule you will get
`spec.namespaceSelector: Invalid value: "object": namespace selector must have at least one rule`

npinaeva (commented):
a couple nits, otherwise looks good

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -612,6 +612,9 @@ func (c *Controller) syncClusterUDN(cudn *userdefinednetworkv1.ClusterUserDefine
 // getSelectedNamespaces list all selected namespaces according to given selector and create
 // a set of the selected namespaces keys.
 func (c *Controller) getSelectedNamespaces(sel metav1.LabelSelector) (sets.Set[string], error) {
+	if len(sel.MatchExpressions)+len(sel.MatchLabels) == 0 {
```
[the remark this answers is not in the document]
ormergi: At first I wasn't sure if Empty() would do, checking it again seem that it does. Thanks!

Done

go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
ormergi: Done

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/636d2478ef20cba12949d9e9d31a7f8bb0203ae8..f9df3e2721cb1f62dd31fa8ee247f4e38af71f87): address https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4963#pullrequestreview-2566671986

@npinaeva thanks for the feedback, please have a look :pray:

ormergi:
Build job failed due to some know flake reported here https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4092#issuecomment-2624067325

ricky-rav (approved):
/LGTM

ormergi:
@trozet could you please have a look? :pray:

go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
```diff
@@ -26,6 +26,7 @@ type ClusterUserDefinedNetworkSpec struct {
 	// NamespaceSelector Label selector for which namespace network should be available for.
 	// +kubebuilder:validation:Required
 	// +required
+	// +kubebuilder:validation:XValidation:rule="has(self.matchLabels) && self.matchLabels.size() > 0 || has(self.matchExpressions) && self.matchExpressions.size() > 0", message="namespace selector must have at least one rule"
```
tssurya, line 29: coming here with fresh mind but looking at the message "namespace selector must have at least one rule" means you have to select something for sure..

but it doesn't reflect the problem which is if I do have 1 rule and that rule is MatchLabels: {} then I think what you are trying to fix is "selecting all namespaces is not allowed like in your tests right?"

for user facing side is that more clearer a message than what you have here "have more than 1 rule?"

tssurya (commented):
almost lgtm, just one clarification around the actual user facing message but otherwise I am good...

for me "must have 1 or more rules" is confusing... because my 1 rule might be to select all namespaces which is not allowed..?

tssurya:
/hold
this is a breaking change actually given we have shipped the API right?

tssurya (changes requested):
Reasons for not merging this API,
@ormergi and @trozet have discussed this PR in the past I am going to try to summarize that here before I close this PR:

1. For P-UDNs, even if admin creates a CUDN matching all namespaces using empty selector, since we need namespaces to have the well-known label the infrastructure namespaces won't be selected anyways
2. For S-UDNs, even if admin creates a CUDN matching all namespaces using empty selector, since we need pods to have the well-known label the infrastructure pods won't be selected anyways.

So why do we still need this change ?

ormergi:
Thanks for the feedback!
At the time of posting this the impact of having the CR created affecting all namespaces was greater and much harder to revert.

Given there are some measures that prevent primary CUDN create in all namespaces, and secondary CUDN from consumed by pods automatically, the impact is lower and it should be easier to revert.

PR #4964, opened 2025-01-15
kubevirt, l2, pudn: reconcile ipv4 gateway mac after live migration

qinqon:
## 📑 Description
After a L2 primary UDN virtual machine is live migrated its ipv4 gateway mac cache is still pointing to the previous node gateway mac from the gateway router LRP.

This PR send a GARP after live migration to update the VM's gateway mac to the new node GR LRP mac address.

This created some utils under pkg/kubevirt to send the GARP and reconcile the gateway and wired up at the layer2 controller so only this controller has code for it.

Closes https://issues.redhat.com/browse/OCPBUGS-48337

## ✅ Checks

- [x] My code requires tests

## How to verify it
It includes tests.

### DIFF
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -479,3 +479,38 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod) 
 	}
 	return &status, nil
 }
+
+func ReconcileIPv4DefaultGatewayAfterLiveMigration(watchFactory *factory.WatchFactory, netInfo util.NetInfo, liveMigrationStatus *LiveMigrationStatus, interfaceName string) error {
+	if liveMigrationStatus.State != LiveMigrationTargetDomainReady {
+		return nil
+	}
+
+	targetNode, err := watchFactory.GetNode(liveMigrationStatus.TargetPod.Spec.NodeName)
+	if err != nil {
+		return err
+	}
+
+	lrpJoinAddress, err := util.ParseNodeGatewayRouterJoinNetwork(targetNode, netInfo.GetNetworkName())
+	if err != nil {
+		return err
+	}
+
+	lrpJoinIPv4, _, err := net.ParseCIDR(lrpJoinAddress.IPv4)
+	if err != nil {
+		return err
+	}
+
+	lrpMAC := util.IPAddrToHWAddr(lrpJoinIPv4)
+	for _, subnet := range netInfo.Subnets() {
+		gwIP := util.GetNodeGatewayIfAddr(subnet.CIDR).IP.To4()
+		if gwIP == nil {
+			continue
+		}
+		garp := util.GARP{IP: gwIP, MAC: &lrpMAC}
+		if err := util.BroadcastGARP(interfaceName, garp); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/go-controller/pkg/ovn/secondary_layer2_network_controller.go
+++ b/go-controller/pkg/ovn/secondary_layer2_network_controller.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/generator/udn"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/kubevirt"
 	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
@@ -184,6 +185,17 @@ func (h *secondaryLayer2NetworkControllerEventHandler) UpdateResource(oldObj, ne
 			_, syncZoneIC := h.oc.syncZoneICFailed.Load(newNode.Name)
 			return h.oc.addUpdateRemoteNodeEvent(newNode, syncZoneIC)
 		}
+	case factory.PodType:
+		newPod := newObj.(*corev1.Pod)
+		oldPod := oldObj.(*corev1.Pod)
+		if err := h.oc.ensurePodForSecondaryNetwork(newPod, shouldAddPort(oldPod, newPod, inRetryCache)); err != nil {
+			return err
+		}
+
+		if h.oc.isPodScheduledinLocalZone(newPod) {
+			return h.oc.updateLocalPodEvent(newPod)
+		}
+		return nil
 	default:
 		return h.oc.UpdateSecondaryNetworkResourceCommon(h.objType, oldObj, newObj, inRetryCache)
 	}
@@ -821,6 +833,36 @@ func (oc *SecondaryLayer2NetworkController) StartServiceController(wg *sync.Wait
 	err := oc.svcController.Run(5, oc.stopChan, wg, runRepair, useLBGroups, false)
 	if err != nil {
 		return fmt.Errorf("error running OVN Kubernetes Services controller for network %s: %v", oc.GetNetworkName(), err)
+	}
+	return nil
+}
+
+func (oc *SecondaryLayer2NetworkController) updateLocalPodEvent(pod *corev1.Pod) error {
+	if kubevirt.IsPodAllowedForMigration(pod, oc.GetNetInfo()) {
+		kubevirtLiveMigrationStatus, err := kubevirt.DiscoverLiveMigrationStatus(oc.watchFactory, pod)
+		if err != nil {
+			return err
+		}
+		if kubevirtLiveMigrationStatus != nil && kubevirtLiveMigrationStatus.TargetPod.Name == pod.Name {
+			if err := oc.reconcileLiveMigrationTargetZone(kubevirtLiveMigrationStatus); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (oc *SecondaryLayer2NetworkController) reconcileLiveMigrationTargetZone(kubevirtLiveMigrationStatus *kubevirt.LiveMigrationStatus) error {
+	// Only primary networks has a gateway to reconcile
+	if !oc.IsPrimaryNetwork() {
+		return nil
+	}
+	mgmtInterfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(oc.GetNetworkID()))
+
+	if hasIPv4Subnet, _ := oc.IPMode(); hasIPv4Subnet {
+		if err := kubevirt.ReconcileIPv4DefaultGatewayAfterLiveMigration(oc.watchFactory, oc.GetNetInfo(), kubevirtLiveMigrationStatus, mgmtInterfaceName); err != nil {
+			return err
+		}
 	}
 	return nil
 }
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -1515,7 +1515,7 @@ runcmd:
 				l[RequiredUDNNamespaceLabel] = ""
 			}
 			ns, err := fr.CreateNamespace(context.TODO(), fr.BaseName, l)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
 			fr.Namespace = ns
 			namespace = fr.Namespace.Name
 
@@ -1532,7 +1532,7 @@ runcmd:
 			if td.topology == "localnet" {
 				By("setting up the localnet underlay")
 				nodes := ovsPods(clientSet)
-				Expect(nodes).NotTo(BeEmpty())
+				Expect(nodes).ToNot(BeEmpty())
 				defer func() {
 					By("tearing down the localnet underlay")
 					Expect(teardownUnderlay(nodes)).To(Succeed())
@@ -1661,6 +1661,24 @@ runcmd:
 				step = by(vmi.Name, fmt.Sprintf("Check north/south traffic after %s %s", td.resource.description, td.test.description))
 				checkNorthSouthIngressIperfTraffic(externalContainerName, nodeIPs, svc.Spec.Ports[0].NodePort, step)
 				checkNorthSouthEgressICMPTraffic(vmi, []string{externalContainerIPV4Address, externalContainerIPV6Address}, step)
+			}
+
+			if td.role == "primary" && td.test.description == liveMigrate.description && isIPv4Supported() && isInterconnectEnabled() {
+				step = by(vm.Name, fmt.Sprintf("Checking IPv4 gateway cached mac after %s %s", td.resource.description, td.test.description))
+				Expect(crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)).To(Succeed())
+
+				targetNode, err := fr.ClientSet.CoreV1().Nodes().Get(context.Background(), vmi.Status.MigrationState.TargetNode, metav1.GetOptions{})
+				Expect(err).ToNot(HaveOccurred(), step)
+
+				expectedGatewayMAC, err := kubevirt.GenerateGatewayMAC(targetNode, netConfig.networkName)
+				Expect(err).ToNot(HaveOccurred(), step)
+
+				Expect(err).ToNot(HaveOccurred(), step)
+				Eventually(kubevirt.RetrieveCachedGatewayMAC).
+					WithArguments(vmi, "enp1s0", cidrIPv4).
+					WithTimeout(10*time.Second).
+					WithPolling(time.Second).
+					Should(Equal(expectedGatewayMAC), step)
 			}
 		},
 			func(td testData) string {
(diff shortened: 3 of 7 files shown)


maiqueb (approved):
I mostly have some opinionated nits you can choose to ignore.

Thanks for this PR. Nice work ! Very readable, and the commit split you've provided makes it very easy to review.

go-controller/pkg/kubevirt/arp.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"net"
+	"net/netip"
+
+	"github.com/mdlayher/arp"
+)
+
+func sendGARP(interfaceName string, ip net.IP, mac net.HardwareAddr) error {
```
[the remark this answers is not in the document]
qinqon: `NewPacket` is implementation details, we should not expose this at sendGARP, the caller just need to know that sendGARP will send the ip->mac mapping as GARP to broadcast to reconcile this information at neigh cache.

I can add a comment to NewPacket call though.

go-controller/pkg/kubevirt/arp.go
```diff
@@ -0,0 +1,35 @@
...
+	addr := netip.AddrFrom4([4]byte(ip))
+	p, err := arp.NewPacket(arp.OperationReply, mac, addr, net.HardwareAddr{0, 0, 0, 0, 0, 0}, addr)
```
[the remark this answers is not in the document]
qinqon: make sense, will add some docs.

go-controller/pkg/kubevirt/arp.go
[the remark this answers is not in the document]
maiqueb: Fair enough I guess. I was just clueless when I read the code as to what `NewPacket` received.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -479,3 +479,41 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod)
...
+	}
+
+	lrpJoinAddress, err := util.ParseNodeGatewayRouterJoinNetwork(targetNode, netInfo.GetNetworkName())
+	if err != nil {
+		return err
+	}
+
+	if lrpJoinAddress.IPv4 == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
qinqon: The original logic to compose LRP macs is 
```golang
gwLRPMAC := util.IPAddrToHWAddr(gwLRPIPs[0])
```
So it's quite different, there is no place on the code that compose the LRP mac using the node annotations, so moving this to helper is not worth it.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -824,5 +825,20 @@ func (oc *SecondaryLayer2NetworkController) StartServiceController(wg *sync.Wait
 }
 
 func (oc *SecondaryLayer2NetworkController) updateLocalPodEvent(pod *corev1.Pod) error {
+	if kubevirt.IsPodAllowedForMigration(pod, oc.GetNetInfo()) {
+		kubevirtLiveMigrationStatus, err := kubevirt.DiscoverLiveMigrationStatus(oc.watchFactory, pod)
+		if err != nil {
+			return nil
```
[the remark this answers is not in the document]
qinqon: good catch, weird that the linter is not catching it.

go-controller/pkg/kubevirt/pod.go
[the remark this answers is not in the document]
qinqon: ack

test/e2e/kubevirt.go
```diff
@@ -1371,6 +1371,25 @@ passwd:
 
 			step = by(vmi.Name, fmt.Sprintf("Check east/west traffic after %s %s", td.resource.description, td.test.description))
 			checkEastWestTraffic(vmi, testPodsIPs, step)
+
+			if td.role == "primary" && td.test.description == liveMigrate.description && isIPv4Supported() && isInterconnectEnabled() {
```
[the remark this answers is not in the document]
qinqon: I am not sure `td.test == liveMigrate` will work to detect that this test if for live migration.

test/e2e/kubevirt/net.go
```diff
@@ -2,10 +2,63 @@ package kubevirt
...
+	config.IPv4Mode = true
+	lrpJoinAddress, err := util.ParseNodeGatewayRouterJoinNetwork(node, networkName)
+	if err != nil {
+		return "", err
+	}
+
+	lrpJoinIPString := lrpJoinAddress.IPv4
+	if lrpJoinIPString == "" {
+		lrpJoinIPString = lrpJoinAddress.IPv6
+	}
```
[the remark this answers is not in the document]
qinqon: Is for completeness and will be useful for when we also fix this same issue for ipv6 (or changes and RAs changes).

test/e2e/kubevirt/net.go
```diff
@@ -2,10 +2,63 @@ package kubevirt
...
+
+	output, err := RunCommand(vmi, fmt.Sprintf("ip neigh get %s dev %s", gatewayIP, dev), 2*time.Second)
+	if err != nil {
+		return "", fmt.Errorf("%s: %v", output, err)
+	}
+	outputSplit := strings.Split(output, " ")
+	if len(outputSplit) < 5 {
+		return "", fmt.Errorf("unexpected 'ip neigh' output %q", output)
+	}
+	return strings.Split(output, " ")[4], nil
```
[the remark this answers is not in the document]
qinqon: Right, good catch.

test/e2e/kubevirt.go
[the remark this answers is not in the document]
maiqueb: Ah, I think I get it now.

OK, let's keep it as is. In hindsight, I simply assumed what the `description` parameter meant.

test/e2e/kubevirt/net.go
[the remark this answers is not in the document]
maiqueb: I'd prefer if you added it only when it made sense. 

I found this confusing, but it's not harmful. If you prefer, just leave it.

go-controller/pkg/kubevirt/arp.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/mdlayher/arp"
+)
+
+func sendGARP(interfaceName string, ip net.IP, mac net.HardwareAddr) error {
+	srcIP := netip.AddrFrom4([4]byte(ip))
+
+	// At at GARP the source and target IP should be the same and point to the
+	// the IP we want to reconcile -> https://wiki.wireshark.org/Gratuitous_ARP
+	p, err := arp.NewPacket(arp.OperationReply, mac /* srcHw */, srcIP, net.HardwareAddr{0, 0, 0, 0, 0, 0}, srcIP)
```
jcaamano, line 16: Is there any specific reason we are using the Reply form of the GARP? If so, might be worth mentioning.

go-controller/pkg/kubevirt/arp.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"net"
+	"net/netip"
+
+	"github.com/mdlayher/arp"
+)
+
+func sendGARP(interfaceName string, ip net.IP, mac net.HardwareAddr) error {
```
jcaamano, line 11: Seems like this should be implemented at the same level as the existing `gratuitousArpOverIfaceByName` in a way that the differences between them, if needed, are highlighted and easier to understand.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -180,6 +180,18 @@ func (h *secondaryLayer2NetworkControllerEventHandler) UpdateResource(oldObj, ne
...
+			return err
+		}
+		newPod, ok := newObj.(*corev1.Pod)
+		if !ok {
+			return fmt.Errorf("could not cast %T object to Pod", newObj)
+		}
+		if h.oc.isPodScheduledinLocalZone(newPod) {
+			return h.oc.updateLocalPodEvent(newPod)
+		}
+		return nil
```
jcaamano, line 195: This is not going to be very maintainable because we can't just assume whatever its done in `UpdateSecondaryNetworkResourceCommon` is compatible and has precedence order to what will be done in `updateLocalPodEvent`.

I would find it to be a more standard approach to not call `UpdateSecondaryNetworkResourceCommon` here, and just call whatever base private method you need from `updateLocalPodEvent`.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -850,5 +850,20 @@ func (oc *SecondaryLayer2NetworkController) updateLocalPodEvent(pod *corev1.Pod)
 }
 
 func (oc *SecondaryLayer2NetworkController) reconcileLiveMigrationTargetZone(kubevirtLiveMigrationStatus *kubevirt.LiveMigrationStatus) error {
+	// Only primary networks has a gateway to reconcile
+	if !oc.IsPrimaryNetwork() {
+		return nil
+	}
+	networkID, err := oc.getNetworkID()
```
jcaamano, line 857: you can use `GetNetworkID` which won't do parsing and will just get the cached value on NetInfo itself

jcaamano, line 857: you can use `GetNetworkID` which won't do parsing and will just get teh cached value on NetInfo itself

jcaamano, line 857: you can use `GetNetworkID` which won't do parsing and will just get the cached value on NetInfo itself

go-controller/pkg/kubevirt/pod.go
```diff
@@ -479,3 +479,37 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod)
...
+	if err != nil {
+		return err
+	}
+
+	lrpMAC := util.IPAddrToHWAddr(lrpJoinIPv4)
+	for _, subnet := range netInfo.Subnets() {
+		gwIP := util.GetNodeGatewayIfAddr(subnet.CIDR).IP.To4()
+		if gwIP == nil {
+			continue
+		}
```
jcaamano, line 508: When does this happen? When the subnet is IPv6? What happens if the subnet is IPv6?

jcaamano (changes requested):

go-controller/pkg/kubevirt/arp.go
qinqon: Make sense, let me try it out, I didn't want to affect the non kubevirt code.

go-controller/pkg/kubevirt/pod.go
qinqon: If subnet is IPv6 we don't have to reconcile the gateway with this mechanism, we have other PRs for IPv6 doing similar stuff, but they have less priority than this.

jcaamano: But that doesn't answer my question

qinqon: This happends after VM live migration success, the subne is ipv6 if cluster is dual stack or single ipv6 and the NAD has ipv6 subnet at the NAD, if the subnet is ipv6 the gateway is configureding using router advetisments from gateway router LRP,

go-controller/pkg/kubevirt/arp.go
qinqon: The GARP with `OperationRequest` is ignored by the VM guest and their ip neigh is not updated, I was surprised by this though.

qinqon: @jcaamano from what I read here https://wiki.wireshark.org/Gratuitous_ARP, looks like we should send both request and reply, since it depend on the system to process them or not.

```
Note that some devices will respond to the gratuitous request and some will respond to the gratuitous reply. If one is trying to write software for moving IP addresses around that works with all routers, switches and IP stacks, it is best to send both the request and the reply. These are documented by [RFC 2002](https://tools.ietf.org/html/rfc2002) and [RFC 826](https://tools.ietf.org/html/rfc826). Software implementing the gratuitious ARP function can be found [in the Linux-HA source tree](http://hg.linux-ha.org/lha-2.1/file/1d5b54f0a2e0/heartbeat/libnet_util/send_arp.c). A request may be preceded by a probe to avoid polluting the address space. For an ARP Probe the Sender IP address field is 0.0.0.0. ARP probes were not considered by the original ARP RFC.
```

What do you think about doing so ?

go-controller/pkg/ovn/secondary_layer2_network_controller.go
qinqon: Problem is that we need to do the GARP after `UpdateSecondaryNetworkResourceCommon` only when pods are updated, an alternative is to do it at "default" switch case checking if we are at pods after calling `UpdateSecondaryNetworkResourceCommon`, that would be better ?

jcaamano: I am proposing not calling `UpdateSecondaryNetworkResourceCommon` and just doing whatever it was being done in `UpdateSecondaryNetworkResourceCommon` in the `updateLocalPodEvent` you are adding. I don't understand what you are saying otherwise.

go-controller/pkg/kubevirt/pod.go
jcaamano: I can't understand what you are saying.
Are you trying to say that `gwIP == nil` when `util.GetNodeGatewayIfAddr(subnet.CIDR).IP` is an IPv6 address? And thus we skip it because this only intends to handle IPv4?

go-controller/pkg/kubevirt/arp.go
jcaamano: I guess we need to do so. Otherwise the method should be really specific on its doc and name about what it is doing or/and who is supposed to be targeting.

go-controller/pkg/kubevirt/pod.go
qinqon: The golang net.IP.To4 returns nil if it's not a IPv4 -> https://pkg.go.dev/net#IP.To4

So the idea is to skip this if it's not IPv4.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
qinqon: So moving UpdateSecondaryNetworkResourceCommon call to updateLocalPodEvent ?

jcaamano: No, the implementation, not the call. The caller should be aware that whatever is being done on the base is compatible to what is being done on the concrete. `UpdateSecondaryNetworkResourceCommon` is not really telling you what its doing. So you need to split it into in smaller pieces that are more specific and call them from updateLocalPodEvent.

qinqon: I see it does the following related to pods 
```golang
case factory.PodType:                                                       
        oldPod := oldObj.(*corev1.Pod)                                          
        newPod := newObj.(*corev1.Pod)                                          
                                                                                
        return bsnc.ensurePodForSecondaryNetwork(newPod, inRetryCache || util.PodScheduled(oldPod) != util.PodScheduled(newPod))
```

But this is called independently if the pod runnins locally or not, so we cannot call it at updateLocalEvent, maybe we can call this at the PodType case and call updateLocalEvent after it ? 

```golang
               newPod, ok := newObj.(*corev1.Pod)
		if !ok {
			return fmt.Errorf("could not cast %T object to Pod", newObj)
		}
               oldPod, ok := oldObj.(*corev1.Pod)
		if !ok {
			return fmt.Errorf("could not cast %T object to Pod", oldObj)
		}
               if err := h.oc.ensurePodForSecondaryNetwork(h.objType, oldObj, newObj, inRetryCache); err != nil {
			return err
		}
		
		if h.oc.isPodScheduledinLocalZone(newPod) {
			return h.oc.updateLocalPodEvent(newPod)
		}
```

go-controller/pkg/kubevirt/arp.go
qinqon: Ok will send a pair of them with request and reply, so we simplfy the interface.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
qinqon: @jcaamano pushed this kind of change, Can you take a look to see if it make sense or if you see it a different solution ?

tssurya:
I know nothing about this PR, but given the time constraints and Tim's approve I am merging this upon request.

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/ea4c09b3437a77a56a898e6fca648d60c7d68ca7..0a79dabfdffbde9975bc6bee3d86fe849a078362 shows 0 net changes after tim approved, looks like quique only rebased on master

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13142923864/job/36675765352

```
Summarizing 1 Failure:
  [FAIL] e2e egress IP validation Network Segmentation: IPv6 L2 role primary [It] Should re-assign egress IPs when node readiness / reachability goes down/up
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/egressip.go:1672
```
err wait is this a known flake? oops

PR #4975, opened 2025-01-17
Fix pods cleanup in ClusterUserDefinedNetwork test

kyrtapz:
netConfig pointer was being modified in a loop so pods were removed from only one namespace.


tssurya (approved):
/lgmt

Issue #4966, opened 2025-01-15
Network Segmentation ClusterUserDefinedNetwork CRD Controller pod connected to ClusterUserDefinedNetwork > [BeforeEach] CR & managed NADs cannot be deleted when being used


tssurya:
### Which jobs are flaking?

NetworkSegmentation lanes

### Which tests are flaking?

Network Segmentation ClusterUserDefinedNetwork CRD Controller pod connected to ClusterUserDefinedNetwork > [BeforeEach] CR & managed NADs cannot be deleted when being used

### Since when has it been flaking?

since yesterday?

### Reason for failure (if possible)

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12770279634/job/35607834353?pr=4877

trozet:
Hit: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12780024327/job/35626401384?pr=4912

seen this several times in last few days

ormergi:
Looking at https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12770279634/job/35607834353?pr=4877 
The test failed at the before each, on waiting for the tested pod to become ready (reach Running phase).
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12770279634/job/35607834353?pr=4877#step:14:4711
The pod wasn't ready at the expected time, 2 minutes.

flavio-fernandes:
Also seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12790625455/job/35658539911?pr=4967

tssurya:
also seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12808065451/job/35711279149?pr=4970

npinaeva:
Closing since it hasn't been updated for a while

Issue #4936, opened 2024-12-20
EIP test `disabling egress nodes impeding GRCP health check` is flaking a lot


martinkennelly:
### Which jobs are flaking?

Ive only seen on on `network segmentation` jobs so far.

### Which tests are flaking?

Network Segmentation: X Y role primary : Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts  disabling egress nodes impeding GRCP health check

### Since when has it been flaking?

Unsure if this was always a flake and its not being highted more because we are running this test a lot more for each PR.

### Reason for failure (if possible)

So far ive seen step:
`4. Check connectivity from both to an external "node" and verify that the IPs are both of the above `

Failed with message and timeout curl exit code:
`2024-12-20T13:37:43.2748685Z   I1220 13:37:43.274273 83719 builder.go:121] Running '/usr/local/bin/kubectl --server=https://127.0.0.1:36687 --kubeconfig=/home/runner/ovn.conf --namespace=egressip-6642 exec e2e-egressip-pod-1 -- curl --connect-timeout 2 172.18.0.6:80'
2024-12-20T13:37:45.4178776Z   I1220 13:37:45.417611 83719 builder.go:135] rc: 28`

### Anything else we need to know?

It passes on subsequent runs.

martinkennelly:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4833

oshoval:
seen here
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12852849945/job/35835711443?pr=4878

https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4144 might be duplicate

tssurya:
also here:
Summarizing 2 Failures:
  [FAIL] e2e egress IP validation Network Segmentation: IPv4 L2 role primary [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts [It] disabling egress nodes impeding GRCP health check
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/egressip.go:479
  [FAIL] e2e egress IP validation Network Segmentation: IPv6 L2 role primary [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts [It] disabling egress nodes impeding GRCP health check
[...]

PR #4979, opened 2025-01-18
Fix missing error wrap on IC remote node add

trozet:
Without this the error for missing annotations will not be suppressed and then false positive errors will be reported.


tssurya (approved):
/lgtm

tssurya (commented):
looks like this code also has been around for a long time... wondering how come these are only surfacing up now..

tssurya:
known flakes:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4966

PR #4980, opened 2025-01-19
localnet: Change default MTU to 1500

RamLavi:
## 📑 Description
In localnet there is no tunneling; thus, there is no point in accounting for tunnel overhead when setting the MTU.
This PR changes the localnet MTU default to kernel default (1500).

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/config/cni.go
+++ b/go-controller/pkg/config/cni.go
@@ -100,7 +100,7 @@ func ParseNetConf(bytes []byte) (*ovncnitypes.NetConf, error) {
 }
 
 func parseNetConfSingle(bytes []byte) (*ovncnitypes.NetConf, error) {
-	netconf := &ovncnitypes.NetConf{MTU: Default.MTU}
+	netconf := &ovncnitypes.NetConf{}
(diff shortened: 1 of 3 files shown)


RamLavi:
@maiqueb do we need to consider this upgrade scenario:
1. user creates NAD without MTU (i.e. default is old 1400)
2. user upgrades OCP to a version that will include this change.
3. user updates NAD with something unrelated (say VLANID) - and thus the MTU would internally change to 1500 on new workloads, even though the user didn't intentionally want it to.

maiqueb:
> @maiqueb do we need to consider this upgrade scenario:
> 
> 1. user creates NAD without MTU (i.e. default is old 1400)
> 2. user upgrades OCP to a version that will include this change.
> 3. user updates NAD with something unrelated (say VLANID) - and thus the MTU would internally change to 1500 on new workloads, even though the user didn't intentionally want it to.

We must preserve the old value, no matter what.
We just want this to apply to newly created networks.

RamLavi:
We have reached an agreement to achieve this with the managed approach of localnet CUDN.
closing this PR

PR #4981, opened 2025-01-19
Fix IPV6 single stack support for udn manifest

tssurya:
Looks like we forgot a spot?
@pperiyasamy PTAL


tssurya:
hmm looks like for secondary networks it doesn't matter or what? this is strange

pperiyasamy:
> hmm looks like for secondary networks it doesn't matter or what? this is strange

yes @tssurya , i ignored the change in my PR since CI was failing, I assumed it's ok to have secondary IPv4 network coexist with IPv6 primary network.

PR #4982, opened 2025-01-20
Use accelerated device as Gateway interface

hareeshpc:
For host networking, external bridge acts as the input/output port with Node IP configured on the bridge itself as a local port.

When hardware acceleration capable devices, like ConnectX or Bluefield2 cards are used, pods can use hardware accelerated Virtual Functions (VFs) or SubFunctions(SFs) as interfaces, and fully offload all kubernetes traffic flows.
But for host networking pods or when the host itself is the traffic endpoint, not all kubernetes flows are accelerated since current CT infrastructure cannot offload CT flows where external bridge is the in/out port.

To allow accelerated traffic flows for host networking, this patch allows specifying a gateway accelerated interface via the `--gateway-accelerated-interface` flag. This can either be a switchdev VF or SF, connected to the external bridge and holding the Node IP.

```
                        ┌──────────┐
                        │  br-ext  │
                  ┌─────┴──┐       │    ┌──────────┐
                  │  eth0  │       │    │  br-int  │
                  └─────┬──┘       │    │          │
                        │          X────X          │
   ┌────────┐     ┌─────┴──┐       │    │          │
   │ eth0v0 ├─────┤ eth0_0 │       │    │          │
   └────────┘     └─────┬──┘       │    └──────────┘
     NODE_IP            │          │
                        └──────────┘
```
where, eth0v0 and eth0_0 are, for ex., VF and VF representor of eth0 uplink. Note that used netdevice must be excluded from device plugin pools, so it won't be used for workload pods.

This flag should be used mutually exclusive to the existing `--gateway-interface` flag.

## 📑 Description
Allows accelerating host networking kubernetes flows by specifying a accelerated netdev as the gateway interface.

Note: It is possible to combine this functionality into the existing `--gateway-interface` flag and do auto detect if it is a accelerated device or not and configure accordingly. Though that would have been simpler from a user perspective, new flag was added from a feature gate perspective as well as to avoid any potential risks introduced by this feature. 
Open for suggestions in this regard.

## ✅ Checks

- [ ] My code requires tests

## How to verify it
On a compute node:

Create switchdev VF or SF from the uplink device. Connect the representor to the external bridge.
Assign/Move the Node IP to this device (VF/SF).
Deploy a cluster, with OVN_GATEWAY_OPTS using this value of VF or SF netdevice. For the scheme above it is `--gateway-accelerated-interface=eth0v0`.
All openflows for the NodeIP should use the openflow port number of this port.
Deploy some workload pod to see if cluster work as expected.

### DIFF
--- a/docs/design/gatway-accelerated-interface-configuration.md
+++ b/docs/design/gatway-accelerated-interface-configuration.md
@@ -1,0 +1,84 @@
+# Gateway Accelerated Interface Configuration
+
+## Description
+
+To provide hardware acceleration for traffic, both IN and OUT ports need to be a hardware 
+accelerated netdevice backed by the Network Interface Card hardware itself.
+In case of external traffic, when one such port is the external OVS bridge, which for example has the gateway IP, 
+such traffic (like host networking traffic) would not be accelerated.
+Using Switchdev VirtualFunction (VF) or SubFunction (SF) as a gateway interface allows to accelerate these too.
+
+
+## How it works?
+
+Instead of using the gateway interface as the external bridge itself, use a  switchdev VF or SF instead. 
+This is depicted as following:
+
+```
+                         +----------+
+                         |  br-ext  |
+                   +--------+       |
+                   | UPLINK |       |
+                   +--------+       |  patch  +----------+
+                         |          x---------x  br-int  |
+    +--------+     +--------+       |  port   +----------+
+    | NETDEV +-----+   REP  |       |
+    +--------+     +--------+       |
+                         +----------+
+```
+
+Where `UPLINK` is a port on an offloading capable network interface hardware, `NETDEV` is a switchdev function 
+of this port and `REP` is a representor netdevice of the switchdev function. 
+Node/Host IP assigned to `NETDEV` which make OVS to chose `REP` port for external flows instead of the bridge.
+
+
+## How to use?
+
+Gateway accelerated interface can be used in two steps:
+
+a) Creating and configuring the device. 
+See figure above.
+An `UPLINK` device is connected to the OVS external bridge. 
+An existing VF or SF `NETDEV` from the `UPLINK` is first selected as the the Gateway Interface. Its associated 
+representor `REP` is plugged into the OVS external bridge (br-ext). The gateway IP is assigned to this interface 
+instead of the OVS external bridge (br-ext). 
+
+b) Specify `NETDEV` as a gateway interface explicitly via `OVN_GATEWAY_OPTS` environment variable for
+  ovnkube-node container. Example:
+
+```yaml
+            - name: OVN_GATEWAY_OPTS
+              value: "--gateway-accelerated-interface=<<NETDEV>>"
+```
+
+Note that this is mutually exclusive to the `--gateway-interface` flag for GATEWAY_OPTIONS.
+
+c) Set the external-id on the bridge to detect the uplink device correctly. This is useful for instances where,
+the name of the bridge (eg: br-ext) does not use the uplink device (eg: p0) in its name. The uplink can also 
+be a bond device. 
+```bash
+ovs-vsctl br-set-external-id br-ext bridge-uplink p0
+```
+This gives more flexibility in detecting the uplink device in cases where the auto detection fails (like in case of 
+bonded uplinks etc.)
+
+## Verification
+
+Openflow rules added to the external bridge will use this port as the IN/OUT port instead.
+
+Example flows when pf0vf1 is the netdev and pf0vf1_r is the representor
+```bash
+ cookie=0xdeff105, duration=505314.637s, table=0, n_packets=0, n_bytes=0, priority=500,ip,in_port="pf0vf1_r",nw_dst=169.254.0.1 actions=ct(table=5,zone=64002,nat)
+ cookie=0xdeff105, duration=505314.637s, table=0, n_packets=655, n_bytes=129843, priority=500,ip,in_port="pf0vf1_r",nw_dst=10.96.0.0/16 actions=ct(commit,table=2,zone=64001,nat(src=169.254.0.2))
+ cookie=0xdeff105, duration=505314.637s, table=0, n_packets=359877855, n_bytes=531033264511, priority=205,udp,in_port=p0,dl_dst=42:0b:9a:f1:83:b2,tp_dst=6081 actions=output:"pf0vf1_r"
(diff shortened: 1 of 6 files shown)


hareeshpc:
@wizhaoredhat : can you please take a look and provide comments.

docs/design/gatway-accelerated-interface-configuration.md
```diff
@@ -0,0 +1,75 @@
...
+An existing VF or SF `NETDEV` from the `UPLINK` is first selected as the the Gateway Interface. Its associated 
+representor `REP` is plugged into the OVS external bridge (br-ext). The gateway IP is assigned to this interface 
+instead of the OVS external bridge (br-ext). 
+
+b) Specify `NETDEV` as a gateway interface explicitly via `OVN_GATEWAY_OPTS` environment variable for
+  ovnkube-node container. Example:
+
+```yaml
+            - name: OVN_GATEWAY_OPTS
+              value: "--gateway-accelerated-interface=<<NETDEV>>"
```
wizhaoredhat, line 51: Shouldn't we have an API similar to the management port using VF/VFRep to accelerate certain flows?

Not just a Netdev but a Device Plugin API (where it uses SR-IOV). The ovn container would get the VF resource from kubelet an using the ENV would take that interface.

hareeshpc: Hello @wizhaoredhat 
Thanks for taking a look.
Yes, using device plugin API approach with management port for gateway interface is possible. I just wanted to keep this patch fairly simple for reviewing and do that in a separate PR. 
Also I was not sure whether the resource pool approach with device plugin was compatible with NetworkManager etc, which could be controlling this interface and its IP etc. 
Hence I thought this approach was most flexible for a start.  Let me know your thoughts.

wizhaoredhat: I think you can create a separate commit for using device plugin API.

wizhaoredhat: I still prefer to see using the device plugin, because it is the correct API to target instead of interface names to VFs/SFs.

cathy-zhou:
@winsopc please take a look.

docs/design/gatway-accelerated-interface-configuration.md
hareeshpc: Hello @wizhaoredhat 
> I still prefer to see using the device plugin, because it is the correct API to target instead of interface names to VFs/SFs.
Will do.

cathy-zhou:
a general question, do you need to update the yaml template files and helm for this support?

go-controller/pkg/node/gateway.go
```diff
@@ -596,15 +616,48 @@ func bridgeForInterface(intfName, nodeName, physicalNetworkName string, gwIPs []
...
 	}
-	gwIntf := intfName
 
-	if bridgeName, _, err := util.RunOVSVsctl("port-to-br", intfName); err == nil {
+	if config.Gateway.GatewayAcceleratedInterface != "" {
+		// Try to get representor and uplink name for the specified gateway device.
+		// If function succeeds, then it is either a valdi switchdev VF or SF, and we can use this accelerated device
+		// for node IP, Host Ofport for Openflow etc.
+		// If failed - error for improper configuration option
+		intfRep, uplink, err = getRepresentorAndUplinkName(config.Gateway.GatewayAcceleratedInterface)
```
cathy-zhou, line 625: it does not seem to have downstream fix for "VF gateway support need support brbond0 with bond0 as uplink"?

hareeshpc: It does. This function is being used to check only that the configuration is a valid switchdev device.
line 643, uses the right device for the uplinkName.

go-controller/pkg/node/gateway.go
```diff
@@ -670,6 +726,42 @@ func bridgeForInterface(intfName, nodeName, physicalNetworkName string, gwIPs []
...
+
+func getRepresentorAndUplinkName(intfName string) (string, string, error) {
+	var rep, uplink string
+	var index int
+	deviceID, err := util.GetDeviceIDFromNetdevice(intfName)
+	if err != nil {
+		return "", "", err
+	}
+
+	if util.IsPCIDeviceName(deviceID) { // PCI device
```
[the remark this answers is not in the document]
hareeshpc: yes, but GetFunctionRepresentorName() did not return the uplink device.
Now, since it is no longer used I can refactor for that.

go-controller/pkg/node/gateway.go
```diff
@@ -596,15 +616,48 @@ func bridgeForInterface(intfName, nodeName, physicalNetworkName string, gwIPs []
...
+		uplinkName, err := util.GetNicName(bridgeName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find nic name for bridge %s: %w", bridgeName, err)
+		}
+		res.bridgeName = bridgeName
+		res.uplinkName = uplinkName
+		res.gwIfaceRep = intfRep
+		res.gwIface = gWAcclInterface
+		res.macAddress = link.Attrs().HardwareAddr
+		res.interfaceID = bridgeName + "_" + nodeName
```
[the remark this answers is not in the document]
hareeshpc: good catch. Will remove.

hareeshpc:
> a general question, do you need to update the yaml template files and helm for this support?

No, since this is a gateway option, the existing yaml templates and helm supports it. I documented the new field  and how to use it in both config and the included docs.

go-controller/pkg/node/gateway.go
winsopc: our downstream create bridge with bonding use brbond0, brbond1.  maybe you need update the document about the how to set the br-ext with bonding uplink.  User need set bridge-uplink external id if the name is br-ext, right?

hareeshpc: right. I will update the document so that this is clear.

cathy-zhou:
LGTM

docs/design/gatway-accelerated-interface-configuration.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+Note that this is mutually exclusive to the `--gateway-interface` flag for GATEWAY_OPTIONS.
+
+c) Optionally set the external-id on the bridge to detect the uplink device.
+```bash
```
winsopc, line 57: if uplink is bond0, and bridge name is not brbond0, we must set this option.  so, it's not optional for this condition.

winsopc:
LGTM

girishmg (approved):

PR #4984, opened 2025-01-20
Refactor and fix UDN/CUDN readiness checks

kyrtapz:
This should fix a flake where the cluster UDN status was expected to be updated straight after creating a namespace:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/acc803c79e4517975f70f7e315a3f09002c739b2/test/e2e/network_segmentation.go#L1138-L1151

The PR includes an alternative solution to: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4967.

Refactored the code to use a dynamic client instead of parsing raw kubectl output. I opted not to use udn/cudn client as this makes the tests more generic and portable.

### DIFF
--- a/test/e2e/network_segmentation.go
+++ b/test/e2e/network_segmentation.go
@@ -6,21 +6,24 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"reflect"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


kyrtapz:
@ormergi @tssurya ptal

tssurya (approved):
/lgtm
thanks Pat!

Issue #4965, opened 2025-01-15
Conformance local dualstack job timing out after 1hr 40 min


trozet:
### Which jobs are flaking?

[e2e (shard-conformance, noHA, local, dualstack, snatGW, 1br, ic-single-node-zones)](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12780024327/job/35626396421?pr=4912#logs)

### Which tests are flaking?

[sig-network] Services[0m [1m[38;5;214mshould be able to switch session affinity for NodePort service [LinuxOnly] [Conformance]

### Since when has it been flaking?

Since January.

### Reason for failure (if possible)

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12780024327/job/35626396421?pr=4912

Goroutine is hanging on a select while executing a kubectl command:

```
2025-01-15T04:18:41.3619915Z   [1m[38;5;214mA running node failed to exit in time[0m
2025-01-15T04:18:41.3620700Z   Ginkgo is moving on but a node has timed out and failed to exit before its grace period elapsed.  The node has now leaked and is running in the background.
2025-01-15T04:18:41.3621413Z   Here's a current progress report:
2025-01-15T04:18:41.3622415Z     [0m[sig-network] Services[0m [1m[38;5;214mshould be able to switch session affinity for NodePort service [LinuxOnly] [Conformance][0m (Spec Runtime: 1h32m50.197s)
2025-01-15T04:18:41.3623747Z       [38;5;243mk8s.io/kubernetes/test/e2e/network/service.go:2216[0m
2025-01-15T04:18:41.3624272Z       In [1m[38;5;214m[It][0m (Node Runtime: 1h32m50.154s)
2025-01-15T04:18:41.3624953Z         [38;5;243mk8s.io/kubernetes/test/e2e/network/service.go:2216[0m
2025-01-15T04:18:41.3626214Z         At [1m[38;5;214m[By Step] creating replication controller affinity-nodeport-transition in namespace services-8083[0m (Step Runtime: 1h32m50.144s)
2025-01-15T04:18:41.3627917Z           [38;5;243mk8s.io/kubernetes/test/e2e/framework/rc/rc_utils.go:88[0m
2025-01-15T04:18:41.3628363Z 
2025-01-15T04:18:41.3628886Z         [38;5;243mBegin Captured GinkgoWriter Output >>[0m
[...]

PR #4985, opened 2025-01-20
Skip session affinity conformance test

trozet:
Test has been flaking and timing out jobs. Already fixed in k8s upstream. See #4965


dceara (approved):
Looks correct to me.

PR #4987, opened 2025-01-21
EIP node: skip OVN managed interfaces that don't serve as secondary

jcaamano:
On ovnkube-node rollout, the node EIP controller was removing routes from route tables associated to UDN VRFs potentially causing either connectivity issues for the UDN or even preventing the UDN controller to start. The node EIP controller should only be handling routes of routing tables associated to secondary interfaces.


martinkennelly (approved):

martinkennelly:
This makes me question our UDN  + BM tests because this would only be seen on BM.
I guess the issue may be transient because after repair node completes, route manager would just restore the routes that it deleted.

martinkennelly:
Ready to go here @jcaamano 

Do you need a maintainer to merge or is my lgtm enough ?

PR #4989, opened 2025-01-21
Ignore removed namespaces in clustermanager

kyrtapz:
When a namespace is removed it shouldn't be considered as an error during reconciliation.

cc: @ormergi


ormergi (approved):
LGTM

ormergi:
Thanks for fixing this, I planned refactoring that area in order to cover it in unit-tests, I will do that on a follow-up PR :)

trozet (approved):
thanks @kyrtapz

PR #4995, opened 2025-01-22
Unit tests: set klog level explicitly

trozet:
We use PrepareTestConfig to set the logging level to 5, however that was not taking effect for klog all the time. This results in some unit tests only printing libovsdb logs, and regular klog level lines. We miss a ton of debug logs required to investigate unit tests failures.

### DIFF
--- a/go-controller/pkg/config/config.go
+++ b/go-controller/pkg/config/config.go
@@ -662,6 +662,12 @@ func PrepareTestConfig() error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


npinaeva (approved):
that is definitely better than before, but tests that don't call `PrepareTestConfig` still exist. Maybe the solution is to just add this call to all tests over time..

tssurya (approved):
/lgtm

waiting on CI

PR #4996, opened 2025-01-22
Fix race between NAD and NetPol deletions

pperiyasamy:
It's noticed NAD is deleted even before handling NetPol delete event which causes NetPol deletion never happened and error is always thrown upon retries. This skips such primary network error and proceeds with deleting netpol object.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller.go
+++ b/go-controller/pkg/ovn/base_network_controller.go
@@ -1005,10 +1005,13 @@ func (bnc *BaseNetworkController) DeleteResourceCommon(objType reflect.Type, obj
 			return fmt.Errorf("could not cast obj of type %T to *knet.NetworkPolicy", obj)
 		}
 		netinfo, err := bnc.networkManager.GetActiveNetworkForNamespace(knp.Namespace)
-		if err != nil {
-			return fmt.Errorf("could not get active network for namespace %s: %v", knp.Namespace, err)
+		// The InvalidPrimaryNetworkError error is thrown when UDN is not found because
+		// it has been already deleted, so just proceed with deleting NetworkPolicy in
+		// such a scenario as well.
+		if err != nil && !util.IsInvalidPrimaryNetworkError(err) {
+			return fmt.Errorf("could not get active network for namespace %s: %w", knp.Namespace, err)
 		}
-		if bnc.GetNetworkName() != netinfo.GetNetworkName() {
+		if err == nil && bnc.GetNetworkName() != netinfo.GetNetworkName() {
 			return nil
 		}
 		return bnc.deleteNetworkPolicy(knp)
--- a/go-controller/pkg/util/util.go
+++ b/go-controller/pkg/util/util.go
@@ -402,6 +402,11 @@ func NewInvalidPrimaryNetworkError(namespace string) *InvalidPrimaryNetworkError
 	return &InvalidPrimaryNetworkError{namespace: namespace}
 }
 
+func IsInvalidPrimaryNetworkError(err error) bool {
+	var invalidPrimaryNetworkError *InvalidPrimaryNetworkError
+	return errors.As(err, &invalidPrimaryNetworkError)
+}
+
 func GetUserDefinedNetworkRole(isPrimary bool) string {
 	networkRole := types.NetworkRoleSecondary
 	if isPrimary {


go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -1005,10 +1005,10 @@ func (bnc *BaseNetworkController) DeleteResourceCommon(objType reflect.Type, obj
 			return fmt.Errorf("could not cast obj of type %T to *knet.NetworkPolicy", obj)
 		}
 		netinfo, err := bnc.networkManager.GetActiveNetworkForNamespace(knp.Namespace)
-		if err != nil {
+		if err != nil && !errors.Is(err, util.NewInvalidPrimaryNetworkError(knp.Namespace)) {
```
kyrtapz, line 1008: NIT: Can you please add a comment explaining why we accept this error here?

pperiyasamy: done.

kyrtapz (dismissed):

trozet (approved):

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -1005,10 +1005,13 @@ func (bnc *BaseNetworkController) DeleteResourceCommon(objType reflect.Type, obj
 			return fmt.Errorf("could not cast obj of type %T to *knet.NetworkPolicy", obj)
 		}
 		netinfo, err := bnc.networkManager.GetActiveNetworkForNamespace(knp.Namespace)
-		if err != nil {
+		// The InvalidPrimaryNetworkError error is thrown when UDN is not found because
+		// it has been already deleted, so just proceed with deleting NetworkPolicy in
+		// such a scenario as well.
+		if err != nil && !errors.Is(err, util.NewInvalidPrimaryNetworkError(knp.Namespace)) {
```
trozet, line 1011: Actually can you change this please to use a util function ( like we do for the other error type):

```
func IsInvalidPrimaryNetworkError(err error) bool {
	var invalidPrimaryNetworkError *InvalidPrimaryNetworkError
	return errors.As(err, &invalidPrimaryNetworkError)
}
```

trozet (changes requested):

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -1005,10 +1005,13 @@ func (bnc *BaseNetworkController) DeleteResourceCommon(objType reflect.Type, obj
...
 			return fmt.Errorf("could not get active network for namespace %s: %v", knp.Namespace, err)
```
trozet, line 1012: also maybe wrap this error with %w

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -1005,10 +1005,13 @@ func (bnc *BaseNetworkController) DeleteResourceCommon(objType reflect.Type, obj
 			return fmt.Errorf("could not cast obj of type %T to *knet.NetworkPolicy", obj)
 		}
 		netinfo, err := bnc.networkManager.GetActiveNetworkForNamespace(knp.Namespace)
-		if err != nil {
-			return fmt.Errorf("could not get active network for namespace %s: %v", knp.Namespace, err)
+		// The InvalidPrimaryNetworkError error is thrown when UDN is not found because
+		// it has been already deleted, so just proceed with deleting NetworkPolicy in
+		// such a scenario as well.
+		if err != nil && !util.IsInvalidPrimaryNetworkError(err) {
+			return fmt.Errorf("could not get active network for namespace %s: %w", knp.Namespace, err)
```
tssurya, line 1012: nit do we also need the networkpolicy name for easier debugging ? we could have more than 1 netpol in this namespace and we won't know which one's deletion failed?

tssurya (commented):
@pperiyasamy and @kyrtapz so if its an invalid netinfo that we get i.e namespace is not labelled so we don't get back any netinfo then what's the expectation here for handling deletion? we should be deleting using default netinfo right? 

what about the situation when there is label but no CRD i.e will getactivenetworkfornamespace return a default netinfo?

go-controller/pkg/ovn/base_network_controller.go
kyrtapz: I think it should be fine as is, the network is an attribute of the whole namespace so having a network policy name here is not strictly necessary

kyrtapz:
> @pperiyasamy and @kyrtapz so if its an invalid netinfo that we get i.e namespace is not labelled so we don't get back any netinfo then what's the expectation here for handling deletion? we should be deleting using default netinfo right?
Actual deletion doesn't rely on the netinfo, it is retrieved to avoid unnecessary no-op operations

> what about the situation when there is label but no CRD i.e will getactivenetworkfornamespace return a default netinfo?
it will return an error, this is what this PR fixes actually. in a labeled namespace without a primary nad GetActiveNetworkForNamespace will always error out

go-controller/pkg/ovn/base_network_controller.go
tssurya: `I think it should be fine as is, the network is an attribute of the whole namespace so having a network policy name here is not strictly necessary`

hmm what I mean is how do you which network policy's deletion here got screwed up?

tssurya:
> It will return an error, this is what this PR fixes actually. in a labeled namespace without a primary nad GetActiveNetworkForNamespace will always error out

the DNC will do this I assume?

tssurya (commented):
@pperiyasamy given our time constraints  I am merging this, could you please do a FUP for UT around delete a netpol with no NAD ? OR the one where CR is missing and label is present?

trozet:
> e label is missing on a namespace do we assume namespace part of DNC for netpol as well? that means in this specific case we will have ACL set up for ne

@tssurya all controllers be using getActiveNetworkForNamespace. It should return a consistent value to all controllers. There is protection built into the function:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/715687ec9c85747c0680af82a0e95a7d0ea30788/go-controller/pkg/networkmanager/nad_controller.go#L423

If the label is missing it means default network. All controllers will then program netpol, svc, etc for default network. If the label exists, then it knows it must be either a UDN or error. CDN is not possible. If there is a controller not using this function then we need to fix it.

PR #4999, opened 2025-01-23
UDN isolation: trigger update on open-default-ports annotation change.

npinaeva:
tested on OCP cluster

### DIFF
--- a/go-controller/pkg/node/udn_isolation.go
+++ b/go-controller/pkg/node/udn_isolation.go
@@ -408,7 +408,8 @@ func podNeedsUpdate(oldObj, newObj *v1.Pod) bool {
 	}
 	// react to pod IP changes
 	return !reflect.DeepEqual(oldObj.Status, newObj.Status) ||
-		oldObj.Annotations[util.OvnPodAnnotationName] != newObj.Annotations[util.OvnPodAnnotationName]
+		oldObj.Annotations[util.OvnPodAnnotationName] != newObj.Annotations[util.OvnPodAnnotationName] ||
+		oldObj.Annotations[util.UDNOpenPortsAnnotationName] != newObj.Annotations[util.UDNOpenPortsAnnotationName]
 }
 
 func (m *UDNHostIsolationManager) reconcilePod(key string) error {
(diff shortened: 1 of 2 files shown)


go-controller/pkg/node/udn_isolation.go
```diff
@@ -408,7 +408,8 @@ func podNeedsUpdate(oldObj, newObj *v1.Pod) bool {
 	}
 	// react to pod IP changes
 	return !reflect.DeepEqual(oldObj.Status, newObj.Status) ||
-		oldObj.Annotations[util.OvnPodAnnotationName] != newObj.Annotations[util.OvnPodAnnotationName]
+		oldObj.Annotations[util.OvnPodAnnotationName] != newObj.Annotations[util.OvnPodAnnotationName] ||
+		oldObj.Annotations[util.UDNOpenPortsAnnotationName] != newObj.Annotations[util.UDNOpenPortsAnnotationName]
```
tssurya, line 412: @pliurh and @martinkennelly I think this fix explains a lot why it sometimes took longer for the run to complete :P 
@npinaeva were we never handling this pod annotation update? I don't think that can be the case cause otherwise that test would be perma-failing.. is it more than there are two paths for the update somewhere and its based on the even ordering during test execution?

npinaeva: the only effect of this change is that open-default-ports annotation will be handled now when host-isolation is enabled
afaik it is not enabled in any of the currently running tests, so not sure how it is related to the failures

tssurya: hmm we have seen instances where 10seconds poll timeout was not sufficient sometimes - I was wondering if there is another path or event miss or something that could happen
but if this change is only relevant to host-isolation enablement then discard what i said

meanwhile could you please explain why this effects only host-isolation enablement piece?

trozet (approved):

go-controller/pkg/node/udn_isolation.go
npinaeva: yep, let me add update case

trozet (approved):

PR #5000, opened 2025-01-23
localnet: Make sure bridge mapping is valid

oshoval:
## 📑 Description
For localnet, the ovs external_ids should have entry for
the network (either `name`, or `physicalNetworkName` if used).
Validate it.

Fixes #

This PR focus on IC only.
See https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5055 for more details.

## ✅ Checks

- [x] My code requires tests

### DIFF
--- a/go-controller/pkg/ovn/address_set/fake_address_set.go
+++ b/go-controller/pkg/ovn/address_set/fake_address_set.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/onsi/gomega"
 
@@ -247,6 +248,13 @@ func (f *FakeAddressSetFactory) EventuallyExpectAddressSetWithAddresses(dbIDsOrN
 	gomega.Eventually(func(g gomega.Gomega) {
 		f.expectAddressSetWithAddresses(g, dbIDs, addresses)
 	}).Should(gomega.Succeed())
+}
+
+func (f *FakeAddressSetFactory) EventuallyExpectAddressSetWithAddressesFailure(dbIDsOrNsName any, addresses []string) {
+	dbIDs := f.getDbIDsFromNsNameOrDbIDs(dbIDsOrNsName)
+	gomega.ConsistentlyWithOffset(1, func(g gomega.Gomega) {
+		f.expectAddressSetWithAddresses(g, dbIDs, addresses)
+	}, 2*time.Second).ShouldNot(gomega.Succeed())
 }
 
 // ExpectEmptyAddressSet ensures the address set owned by dbIDsOrNsName exists with no Addresses
--- a/go-controller/pkg/ovn/multipolicy_test.go
+++ b/go-controller/pkg/ovn/multipolicy_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 
 	cnitypes "github.com/containernetworking/cni/pkg/types"
@@ -539,9 +540,11 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 		})
 
 		ginkgo.DescribeTable("correctly adds and deletes pod IPs from secondary network namespace address set",
-			func(topology string, remote bool) {
+			func(topology string, remote bool, ovnMapping string, expectedError string) {
 				app.Action = func(*cli.Context) error {
 					var err error
+					os.Setenv("K8S_NODE", "node1")
+					ginkgo.DeferCleanup(os.Unsetenv, "K8S_NODE")
 
 					subnets := "10.1.0.0/16"
 					nodeSubnet := ""
@@ -581,6 +584,12 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 					namespace1 := *newNamespace(namespaceName1)
 
 					config.EnableMulticast = false
+
+					if ovnMapping != "" {
+						fakeOvn.SetExternalIDs(map[string]string{"ovn-bridge-mappings": ovnMapping})
+						ginkgo.DeferCleanup(fakeOvn.ClearExternalIDs)
+					}
+
 					startOvn(initialDB, watchNodes, []corev1.Node{node}, []corev1.Namespace{namespace1}, nil, nil,
 						[]nettypes.NetworkAttachmentDefinition{*nad}, []testPod{}, map[string]string{labelName: labelVal})
 
@@ -615,8 +624,13 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 						gomega.Expect(err).NotTo(gomega.HaveOccurred())
 					}
 
-					ocInfo.asf.EventuallyExpectAddressSetWithAddresses(namespaceName1, []string{"10.1.1.1"})
-
+					if expectedError == "" {
+						ocInfo.asf.EventuallyExpectAddressSetWithAddresses(namespaceName1, []string{"10.1.1.1"})
+					} else {
+						ocInfo.asf.EventuallyExpectAddressSetWithAddressesFailure(namespaceName1, []string{"10.1.1.1"})
+						gomega.Expect(fakeOvn.fakeRecorder.Events).To(gomega.HaveLen(1))
+						gomega.Expect(<-fakeOvn.fakeRecorder.Events).To(gomega.Equal(`Warning ErrorReconcilingPod failed to find bridge mapping for network: "network1" on node "node1"; Current ovn-bridge-mappings: ""`))
+					}
 					// Delete the pod
 					ginkgo.By("Deleting the pod")
 					err = fakeOvn.fakeClient.KubeClient.CoreV1().Pods(nPodTest.namespace).Delete(context.TODO(), nPodTest.podName, metav1.DeleteOptions{})
@@ -629,12 +643,52 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 				err := app.Run([]string{app.Name})
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 			},
-			ginkgo.Entry("on local zone for layer3 topology", ovntypes.Layer3Topology, false),
-			ginkgo.Entry("on remote zone for layer3 topology", ovntypes.Layer3Topology, true),
-			ginkgo.Entry("on local zone for layer2 topology", ovntypes.Layer2Topology, false),
-			ginkgo.Entry("on remote zone for layer2 topology", ovntypes.Layer2Topology, true),
-			ginkgo.Entry("on local zone for localnet topology", ovntypes.LocalnetTopology, false),
-			ginkgo.Entry("on remote zone for localnet topology", ovntypes.LocalnetTopology, true),
+			NewTestCase(ovntypes.Layer3Topology).Build(),
+			NewTestCase(ovntypes.Layer3Topology).WithRemote().Build(),
+			NewTestCase(ovntypes.Layer2Topology).Build(),
+			NewTestCase(ovntypes.Layer2Topology).WithRemote().Build(),
+			NewTestCase(ovntypes.LocalnetTopology).Build(),
+			NewTestCase(ovntypes.LocalnetTopology).WithRemote().WithOVNMapping("physnet:breth0,network1:ovsbr1").Build(),
+			NewTestCase(ovntypes.LocalnetTopology).WithRemote().ExpectFailure(
+				`Warning ErrorReconcilingPod failed to find bridge mapping for network: "network1", node "node1", ovn-bridge-mappings: ""`,
+			).Build(),
 		)
 	})
 })
+
+type TestCaseBuilder struct {
+	topology      string
+	remote        bool
+	expectedError string
+	mapping       string
+}
+
+func NewTestCase(topology string) *TestCaseBuilder {
+	return &TestCaseBuilder{topology: topology}
+}
+
+func (b *TestCaseBuilder) WithRemote() *TestCaseBuilder {
+	b.remote = true
+	return b
+}
+
+func (b *TestCaseBuilder) ExpectFailure(msg string) *TestCaseBuilder {
+	b.expectedError = msg
+	return b
+}
+
+func (b *TestCaseBuilder) Build() ginkgo.TableEntry {
+	zone := "local"
+	if b.remote {
+		zone = "remote"
+	}
+
+	return ginkgo.Entry(
+		fmt.Sprintf("on %s zone for %s topology", zone, b.topology), b.topology, b.remote, b.mapping, b.expectedError,
+	)
+}
+
+func (b *TestCaseBuilder) WithOVNMapping(mapping string) *TestCaseBuilder {
+	b.mapping = mapping
+	return b
+}
--- a/go-controller/pkg/ovn/ovn_test.go
+++ b/go-controller/pkg/ovn/ovn_test.go
@@ -52,6 +52,7 @@ import (
 	testnm "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/testing/networkmanager"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	util "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/vswitchd"
 )
 
 const (
@@ -90,6 +91,7 @@ type FakeOVN struct {
 	networkManager networkmanager.Controller
 	eIPController  *EgressIPController
 	portCache      *PortCache
+	externalIDs    map[string]string
 
 	// information map of all secondary network controllers
 	secondaryControllers       map[string]secondaryControllerInfo
@@ -563,7 +565,11 @@ func (o *FakeOVN) NewSecondaryNetworkController(netattachdef *nettypes.NetworkAt
 			secondaryController = &l2Controller.BaseSecondaryNetworkController
 			o.fullSecondaryL2Controllers[netName] = l2Controller
 		case types.LocalnetTopology:
-			localnetController := NewSecondaryLocalnetNetworkController(cnci, nInfo, o.networkManager.Interface())
+			ovsClient, err := o.setupOVSWithExternalIDs()
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			localnetController, err := newSecondaryLocalnetNetworkController(cnci, nInfo, o.networkManager.Interface(), ovsClient)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 			if o.asf != nil { // use fake asf only when enabled
 				localnetController.addressSetFactory = asf
 			}
@@ -591,6 +597,17 @@ func (o *FakeOVN) NewSecondaryNetworkController(netattachdef *nettypes.NetworkAt
 	return nil
 }
 
+func (o *FakeOVN) SetExternalIDs(externalIDs map[string]string) {
+	o.externalIDs = make(map[string]string, len(externalIDs))
+	for k, v := range externalIDs {
+		o.externalIDs[k] = v
+	}
+}
+
+func (o *FakeOVN) ClearExternalIDs() {
+	o.externalIDs = nil
+}
+
 func (o *FakeOVN) patchEgressIPObj(nodeName, egressIPName, egressIP string) {
 	// NOTE: Cluster manager is the one who patches the egressIP object.
 	// For the sake of unit testing egressip zone controller we need to patch egressIP object manually
@@ -611,4 +628,16 @@ func nadGVR() metav1.GroupVersionResource {
 		Version:  "v1",
 		Resource: "network-attachment-definitions",
 	}
+}
+
+func (o *FakeOVN) setupOVSWithExternalIDs() (libovsdbclient.Client, error) {
+	dbSetup := libovsdbtest.TestSetup{
+		OVSData: []libovsdbtest.TestData{
+			&vswitchd.OpenvSwitch{
+				ExternalIDs: o.externalIDs,
+			},
+		},
+	}
+	ovsClient, _, err := libovsdbtest.NewOVSTestHarness(dbSetup)
+	return ovsClient, err
 }
--- a/go-controller/pkg/ovn/secondary_localnet_network_controller.go
+++ b/go-controller/pkg/ovn/secondary_localnet_network_controller.go
@@ -3,7 +3,9 @@ package ovn
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
+	"strings"
 	"sync"
 	"time"
 
@@ -12,9 +14,14 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
+	"github.com/ovn-org/libovsdb/client"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/pod"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb"
 	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
+	ovsops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops/ovs"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
@@ -105,6 +112,11 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) AddResource(obj interfa
 			return fmt.Errorf("could not cast %T object to Node", obj)
 		}
 		return h.oc.addUpdateNodeEvent(node)
+	case factory.PodType:
+		if err := h.oc.checkBridgeMapping(obj); err != nil {
+			return fmt.Errorf("failed pod AddResource %w", err)
+		}
+		fallthrough
 	default:
 		return h.oc.AddSecondaryNetworkResourceCommon(h.objType, obj)
 	}
@@ -122,6 +134,11 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) UpdateResource(oldObj, 
 			return fmt.Errorf("could not cast %T object to Node", newObj)
 		}
 		return h.oc.addUpdateNodeEvent(node)
+	case factory.PodType:
+		if err := h.oc.checkBridgeMapping(newObj); err != nil {
+			return fmt.Errorf("failed pod UpdateResource %w", err)
+		}
+		fallthrough
 	default:
 		return h.oc.UpdateSecondaryNetworkResourceCommon(h.objType, oldObj, newObj, inRetryCache)
 	}
@@ -186,6 +203,9 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) IsObjectInTerminalState
 // for a secondary localnet network
 type SecondaryLocalnetNetworkController struct {
 	BaseSecondaryLayer2NetworkController
+	ovsClient          client.Client
+	ovsClientCancel    context.CancelFunc
+	controllerNodeName string
 }
 
 // NewSecondaryLocalnetNetworkController create a new OVN controller for the given secondary localnet NAD
@@ -193,14 +213,33 @@ func NewSecondaryLocalnetNetworkController(
 	cnci *CommonNetworkControllerInfo,
 	netInfo util.NetInfo,
 	networkManager networkmanager.Interface,
-) *SecondaryLocalnetNetworkController {
+) (*SecondaryLocalnetNetworkController, error) {
+	return newSecondaryLocalnetNetworkController(cnci, netInfo, networkManager, nil)
+}
+
+func newSecondaryLocalnetNetworkController(
+	cnci *CommonNetworkControllerInfo,
+	netInfo util.NetInfo,
+	networkManager networkmanager.Interface,
+	ovsClient client.Client,
+) (*SecondaryLocalnetNetworkController, error) {
 
 	stopChan := make(chan struct{})
+
+	ctx, cancel := context.WithCancel(context.Background())
+	if ovsClient == nil {
+		var err error
+		ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
+		if err != nil {
+			cancel()
+			return nil, err
+		}
+	}
 
 	ipv4Mode, ipv6Mode := netInfo.IPMode()
 	addressSetFactory := addressset.NewOvnAddressSetFactory(cnci.nbClient, ipv4Mode, ipv6Mode)
 	oc := &SecondaryLocalnetNetworkController{
-		BaseSecondaryLayer2NetworkController{
+		BaseSecondaryLayer2NetworkController: BaseSecondaryLayer2NetworkController{
 			BaseSecondaryNetworkController: BaseSecondaryNetworkController{
 				BaseNetworkController: BaseNetworkController{
 					CommonNetworkControllerInfo: *cnci,
@@ -222,6 +261,9 @@ func NewSecondaryLocalnetNetworkController(
 				},
 			},
 		},
+		ovsClient:          ovsClient,
+		ovsClientCancel:    cancel,
+		controllerNodeName: os.Getenv("K8S_NODE"),
 	}
 
 	if oc.allocatesPodAnnotation() {
@@ -247,7 +289,7 @@ func NewSecondaryLocalnetNetworkController(
 	oc.multicastSupport = false
 
 	oc.initRetryFramework()
-	return oc
+	return oc, nil
 }
 
 // Start starts the secondary localnet controller, handles all events and creates all needed logical entities
@@ -309,6 +351,7 @@ func (oc *SecondaryLocalnetNetworkController) init() error {
 
 func (oc *SecondaryLocalnetNetworkController) Stop() {
 	klog.Infof("Stoping controller for secondary network %s", oc.GetNetworkName())
+	oc.ovsClientCancel()
 	oc.BaseSecondaryLayer2NetworkController.stop()
 }
 
@@ -360,11 +403,82 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
 }
 
 func (oc *SecondaryLocalnetNetworkController) localnetPortNetworkNameOptions() map[string]string {
-	localnetLSPOptions := map[string]string{
-		"network_name": oc.GetNetworkName(),
+	return map[string]string{
+		"network_name": oc.findNetworkName(),
 	}
+}
+
+func (oc *SecondaryLocalnetNetworkController) checkBridgeMapping(obj interface{}) error {
+	if !config.OVNKubernetesFeature.EnableInterconnect {
+		return nil
+	}
+
+	pod, isPod := obj.(*corev1.Pod)
+	if !isPod {
+		return fmt.Errorf("failed casting %T object to *knet.Pod", obj)
+	}
+
+	if !util.PodScheduled(pod) {
+		return nil
+	}
+
+	if oc.controllerNodeName != pod.Spec.NodeName {
+		return nil
+	}
+
+	isNADKnown, err := oc.isPodAttachedToKnownNAD(pod)
+	if err != nil {
+		return err
+	}
+	if !isNADKnown {
+		return nil
+	}
+
+	return oc.networkMappedToBridge(pod)
+}
+
+func (oc *SecondaryLocalnetNetworkController) isPodAttachedToKnownNAD(pod *corev1.Pod) (bool, error) {
+	networks, err := util.GetK8sPodAllNetworkSelections(pod)
+	if err != nil {
+		return false, err
+	}
+
+	nInfo := oc.GetNetInfo()
+	for _, network := range networks {
+		nadName := util.GetNADName(network.Namespace, network.Name)
+		if nInfo.HasNAD(nadName) {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func (oc *SecondaryLocalnetNetworkController) networkMappedToBridge(pod *corev1.Pod) error {
+	openvSwitch, err := ovsops.GetOpenvSwitch(oc.ovsClient)
+	if err != nil {
+		return err
+	}
+
+	ovnBridgeMappings := openvSwitch.ExternalIDs["ovn-bridge-mappings"]
+
+	networkName := oc.findNetworkName()
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		networkBridgeAssociation := strings.Split(bridgeMapping, ":")
+		if len(networkBridgeAssociation) == 2 && networkBridgeAssociation[0] == networkName {
+			return nil
+		}
+	}
+
+	err = fmt.Errorf("failed to find bridge mapping for network: %q on node %q; Current ovn-bridge-mappings: %q", networkName, oc.controllerNodeName, ovnBridgeMappings)
+	oc.recordPodErrorEvent(pod, err)
+	return err
+}
+
+func (oc *SecondaryLocalnetNetworkController) findNetworkName() string {
 	if oc.PhysicalNetworkName() != "" {
-		localnetLSPOptions["network_name"] = oc.PhysicalNetworkName()
+		return oc.PhysicalNetworkName()
 	}
-	return localnetLSPOptions
+	return oc.GetNetworkName()
 }
--- a/test/e2e/multihoming.go
+++ b/test/e2e/multihoming.go
@@ -67,6 +67,12 @@ var _ = Describe("Multi Homing", func() {
 			netConfig.namespace = f.Namespace.Name
 			podConfig.namespace = f.Namespace.Name
 
+			if netConfig.topology == "localnet" {
+				By("applying ovs bridge mapping")
+				Expect(setBridgeMappings(cs, defaultNetworkBridgeMapping(), bridgeMapping(netConfig.networkName, bridgeName))).NotTo(HaveOccurred())
+				ginkgo.DeferCleanup(setBridgeMappings, cs, defaultNetworkBridgeMapping())
+			}
+
 			By("creating the attachment configuration")
 			_, err := nadClient.NetworkAttachmentDefinitions(netConfig.namespace).Create(
 				context.Background(),
@@ -1811,4 +1817,19 @@ func createMultiNetworkPolicy(mnpClient mnpclient.K8sCniCncfIoV1beta1Interface, 
 		metav1.CreateOptions{},
 	)
 	return err
+}
+
+func setBridgeMappings(cs clientset.Interface, mappings ...BridgeMapping) error {
+	pods := ovsPods(cs)
+	if len(pods) == 0 {
+		return fmt.Errorf("pods list is empty")
+	}
+
+	for _, pods := range pods {
+		if err := configureBridgeMappings(pods.Name, mappings...); err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/go-controller/pkg/controllermanager/controller_manager.go
+++ b/go-controller/pkg/controllermanager/controller_manager.go
@@ -79,7 +79,7 @@ func (cm *ControllerManager) NewNetworkController(nInfo util.NetInfo) (networkma
 	case ovntypes.Layer2Topology:
 		return ovn.NewSecondaryLayer2NetworkController(cnci, nInfo, cm.networkManager.Interface(), cm.eIPController, cm.portCache)
 	case ovntypes.LocalnetTopology:
-		return ovn.NewSecondaryLocalnetNetworkController(cnci, nInfo, cm.networkManager.Interface()), nil
+		return ovn.NewSecondaryLocalnetNetworkController(cnci, nInfo, cm.networkManager.Interface())
 	}
 	return nil, fmt.Errorf("topology type %s not supported", topoType)
 }
@@ -99,7 +99,7 @@ func (cm *ControllerManager) newDummyNetworkController(topoType, netName string)
 	case ovntypes.Layer2Topology:
 		return ovn.NewSecondaryLayer2NetworkController(cnci, netInfo, cm.networkManager.Interface(), cm.eIPController, cm.portCache)
 	case ovntypes.LocalnetTopology:
-		return ovn.NewSecondaryLocalnetNetworkController(cnci, netInfo, cm.networkManager.Interface()), nil
+		return ovn.NewSecondaryLocalnetNetworkController(cnci, netInfo, cm.networkManager.Interface())
 	}
 	return nil, fmt.Errorf("topology type %s not supported", topoType)
 }


oshoval:
@maiqueb 
do we need to verify the mapping only when `physicalNetworkName` is used?
or always (for localnet only), even for the `name field when `physicalNetworkName` na

maiqueb:
> @maiqueb do we need to verify the mapping only when `physicalNetworkName` is used? or also for the `name` field (localnet only) when`physicalNetworkName` is na

When it isn't, you'll have to verify the network `name` attribute.

maiqueb:
/cc
Adding this to my queue. 

Thanks for work; I'll start reviewing ASAP - probably next week.

oshoval:
some unit tests are correctly broken
so i need to fix them first, and then see if i need to add more

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1177,3 +1190,21 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
...
+
+	bridgeMappings := strings.Split(external_ids, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		m := strings.Split(bridgeMapping, ":")
+		if m[0] == physicalNetworkName {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("failed to find bridge mapping for physicalNetworkName: %s, external_ids %s", physicalNetworkName, external_ids)
```
maiqueb, line 1209:
```suggestion
	return fmt.Errorf("failed to find bridge mapping for physicalNetworkName: %q, external_ids %s", physicalNetworkName, external_ids)
```

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1177,3 +1190,21 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
 	}
 	return bnc.releasedPodsBeforeStartup[nad].Has(uid)
 }
+
+func checkBridgeMapping(physicalNetworkName string) error {
+	external_ids, stderr, err := util.RunOVSVsctl("--if-exists", "get", "Open_vSwitch", ".",
```
maiqueb, line 1195: let's rename the variable; something like `ovsTableMappings` or something like that - you're grabbing a subset of the external ids, not all of them.

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1177,3 +1190,21 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
...
+		"external_ids:ovn-bridge-mappings")
+	if err != nil {
+		return fmt.Errorf("failed to get ovn-bridge-mappings stderr:%s (%v)", stderr, err)
```
maiqueb, line 1198:
```suggestion
		return fmt.Errorf("failed to get ovn-bridge-mappings from the OVS global table. stderr: %q (%v)", stderr, err)
```

go-controller/pkg/ovn/base_network_controller_pods.go
maiqueb: And I wonder if we should print the name of the node ... I guess not.

maiqueb (changes requested):
Veeeeeeeeeeeeeeeeeeeeery early feedback.

Let's see how this looks regarding the unit-tests :)

I think we should have a different design here - someway to enable this to be mocked in the unit tests. 

But, the production code looks solid.

oshoval:
Thanks

1st push - Fixed the current unit tests (in a separate commit, easier to review / change upon needs)
This might be enough as unit tests, as it does check the changed localnet flow, with mock as you suggested

2nd push - Rebased

Will now look on the comments

go-controller/pkg/ovn/base_network_controller_pods.go
oshoval: dont think so, after all it already on the pod that runs on that node

go-controller/pkg/ovn/base_network_controller_pods.go
oshoval: %q doesnt fit for stderr i think ?
for example if it has \n, it will print them escaped iiuc,
in any case, since it is not a name, but a string, %q shouldn't be used imho

oshoval:
addressed / answered comments
thanks

oshoval:
tests are failing due to this PR, and also might do some refactors, so not ready yet

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13012320869/job/36294831718?pr=5000

go-controller/pkg/ovn/admin_network_policy_test.go
```diff
@@ -299,7 +299,7 @@ var _ = ginkgo.Describe("OVN ANP Operations", func() {
 		app.Name = "test"
 		app.Flags = config.Flags
 
-		fakeOVN = NewFakeOVN(false)
+		fakeOVN = NewFakeOVN(false, nil)
```
maiqueb, line 302: How about we create a *new* function for this, instead of force adding a parameter to the current one, which forces you to change a lot of files ?

oshoval: yep thats the plan

oshoval:
1st push
used factory
rearranged commits
fix e2e tests

2nd push - rebase

oshoval:
@qinqon @RamLavi 
can you please take a look ?
thanks

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -454,6 +454,19 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(pod *kapi.Pod, nadName
 		return nil, nil, nil, false, fmt.Errorf("[%s] failed geting expected switch name when adding logical switch port: %v", podDesc, err)
 	}
 
+	if bnc.TopologyType() == ovntypes.LocalnetTopology {
```
qinqon, line 457: Can we move this to the localnet controller ?

go-controller/pkg/ovn/ovn_test.go
```diff
@@ -91,32 +91,39 @@ type FakeOVN struct {
...
-		egressQoSWg:  &sync.WaitGroup{},
-		egressSVCWg:  &sync.WaitGroup{},
-		anpWg:        &sync.WaitGroup{},
-
+		asf:                        asf,
+		fakeRecorder:               record.NewFakeRecorder(10),
+		egressQoSWg:                &sync.WaitGroup{},
+		egressSVCWg:                &sync.WaitGroup{},
+		anpWg:                      &sync.WaitGroup{},
+		fexec:                      fexec,
```
RamLavi, line 116: Looks like every FakeOVN object uses the Start method, so why not have NewFakeOVN directly set:
```
fexec: ovntest.NewFakeExec(),
```
?

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1177,3 +1190,21 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
...
+
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		m := strings.Split(bridgeMapping, ":")
+		if m[0] == physicalNetworkName {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("failed to find bridge mapping for physicalNetworkName: %q, ovn-bridge-mappings external_ids %s", physicalNetworkName, ovnBridgeMappings)
```
RamLavi, line 1209: half asking: bridge mapping is node specific, shouldn't we add on what node is the bridge-mapping absent?

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -454,6 +454,19 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(pod *kapi.Pod, nadName
 		return nil, nil, nil, false, fmt.Errorf("[%s] failed geting expected switch name when adding logical switch port: %v", podDesc, err)
 	}
 
+	if bnc.TopologyType() == ovntypes.LocalnetTopology {
+		netInfo := bnc.GetNetInfo()
+		networkName := netInfo.GetNetworkName()
+		if physicalNetworkName := netInfo.PhysicalNetworkName(); physicalNetworkName != "" {
+			networkName = physicalNetworkName
+		}
```
qinqon, line 462: We can call `PyhsicalNetworkName` twice to make it more readable, also maybe this little thing merits a helper.

```golang
if netInfo.PhysicalNetworkName(); != "" {
			networkName = netInfo.PhysicalNetworkName()
		}
```

qinqon (changes requested):
Also change the last commit message and instead of `external_ids` talk about the bridge mapping `external_ids:ovn-bridge-mappings`

test/e2e/multihoming.go
```diff
@@ -67,6 +67,18 @@ var _ = Describe("Multi Homing", func() {
 			netConfig.namespace = f.Namespace.Name
 			podConfig.namespace = f.Namespace.Name
 
+			By("applying ovs bridge mapping")
```
RamLavi, line 70: bridge-mapping is only relevant for localnet, so could we do it only on these cases? 
Also, could we DeferClean it after?

go-controller/pkg/ovn/base_network_controller_pods.go
oshoval: will do the change, but i dont think it merits a helper
we have already helper for `checkBridgeMapping` few lines below

RamLavi (commented):
Hey Or, I Asked some questions on the PR

test/e2e/multihoming.go
oshoval: since we just add a mapping to the default one, it is fine to do it always + to not clean it

go-controller/pkg/ovn/base_network_controller_pods.go
oshoval: the error will be presented on a pod, and the pod is node specific already

go-controller/pkg/ovn/ovn_test.go
oshoval: you need to have access to fexec outside this function,
unless we export Fexec and then it is possible, but it means it will always export it even when not needed
so tried to do it as close as to current (i.e only when needed)

go-controller/pkg/ovn/base_network_controller_pods.go
oshoval: looking on `go-controller/pkg/ovn/secondary_localnet_network_controller.go` 
i dont see where it fit,
if you still think it should be there, can you suggest where please?

go-controller/pkg/ovn/base_network_controller_pods.go
qinqon: Fair enough, just simplify the if statement and we good.

oshoval: did, thx

go-controller/pkg/ovn/base_network_controller_pods.go
qinqon: So since this is a check that should prevent localnet network infra to be created we can put it directly at the localne event handler add, update and sync functions at https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/secondary_localnet_network_controller.go#L96-L124

Just wrap the check with all the logic at a helper with the event handler as receiver and call it from the three places, before calling common network functions.

oshoval: this operation is expansive, so i dont think we need to have it on sync
also having it on 3 places makes it harder to maintain + the current file already has switch for the type
lets please see what @maiqueb  think

maiqueb: I think @qinqon 's right. 

While this is less code, it's polluting the [base_network_controller_pods.go](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/files#diff-941b484ab7f41b84a6ba9c2dee91f3454401447db0ca70cb401a8cc582cc072e) with something localnet specific ...

Let's at least try to be good citizens. If it proves to be too complex, let's leave it as is now. 
But you have more than enough time to provide nice code.

go-controller/pkg/ovn/base_network_controller_pods.go
maiqueb: right, this is the msg that will be listed on pod's describe operation, right ? 

If so, I agree: no point logging the node name there. 

... it would be totally different if this were a log on ovnkube-controller ...

test/e2e/multihoming.go
maiqueb: yeah man, let's only do this once needed, an strive to clean it up afterwards. 

You have enough time to provide amazing code. Let's take that opportunity. No pressure.

maiqueb (changes requested):
Looking good, but can look better !

Good work so far; let's make it even better.

go-controller/pkg/ovn/ovn_test.go
RamLavi: yeah, I just figured everyone who initializes fakeOVN also used Start (also checked on 4 filed..) so if it was possible to always use that it would've made the code much simpler, even at the cost of making fexec public.
But if you say you checked it then it must be more complicated than I assumed so nevermind.

go-controller/pkg/ovn/base_network_controller_pods.go
RamLavi: yeah this is going to be on the pod events so it's easy to know what is the node.
But keeping in mind the story of "user created a bridge-mapping on desired node, user created vm, user trying to figure out why the vm is not running" it would help to shed a light in the event that the pod got to the wrong place. 
But it's not a must indeed - up to you..

maiqueb: ... I mean ... it doesn't cost much to add the node name tbh ...

go-controller/pkg/ovn/base_network_controller_pods.go
oshoval: can this be done please on a follow-up ?
it will make it easier to be focused

qinqon: @oshoval this do not merit a follow up, just make a helper function with localnet controller as receiver put the logic there, then you call this a proper places at localnet controller, we have already a lot of spaghetti code at base controller.

oshoval:
Addressed comments beside [1] (will now continue that)
no need to review yet
[1] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/#discussion_r1933845665

go-controller/pkg/ovn/ovn_test.go
oshoval: adapted

oshoval:
Addressed comments as we spoke offline (2 last pushes)
Didn't include sync because of the mentioned scenario.
Return nil in case the pod isn't scheduled yet (update will add the nodeName once scheduler did the pod binding
and then we can assert only on the right node controller).
Improved fetching the ovs mapping.

Do we also need to verify that the bridge the mapping points to exists?
The mapping might be stale.

Not related to this PR, can be done on follow-up if desired:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/eb3eeeba5d0cf08bfa15963fd434975b5d854f3d/go-controller/pkg/ovn/secondary_localnet_network_controller.go#L207
everywhere that we use `GetNetworkName` we should use the new oc.getNetworkName function,
because if phys network exists, it is the real network.
(might be also outside of this file)

oshoval:
Fixed mocking, lets see all passing first

oshoval:
Possible idea - in case a known flaky unit test fails, do run rest of the e2e lanes (else sometimes we do need to see their result, but need to rerun few times because just one flaky unit test)
We can just mark the flaky ones with some tag, and the job will analyze the results,
It can maybe even run the e2e but have one additional lane of e2e that just populate the real retcode of the unit test upon failure, hence blocking the PR if desired

wdyt?
If you like the idea we can discuss it deeper on an issue / slack etc

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +390,44 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	}
+
+	if !util.PodScheduled(pod) {
+		return nil
+	}
+
+	openvSwitch, err := ovsops.GetOpenvSwitch(oc.ovsClient)
+	if err != nil {
+		return err
+	}
```
maiqueb, line 411: Do we need to invoke this everytime we run `checkBridgeMapping` ? 
Couldn't we encapsulate this in some struct and just read the external IDs from it ? 

Just asking, I haven't checked the actual implementation of  `ovsops.GetOpenvSwitch` .

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +390,44 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	ovnBridgeMappings := openvSwitch.ExternalIDs["ovn-bridge-mappings"]
+
+	networkName := oc.getNetworkName()
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		m := strings.Split(bridgeMapping, ":")
+		if m[0] == networkName {
+			return nil
+		}
 	}
```
maiqueb, line 421: maybe encapsulate this in an helper func ?

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +390,44 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	if err != nil {
+		return err
+	}
+	ovnBridgeMappings := openvSwitch.ExternalIDs["ovn-bridge-mappings"]
+
+	networkName := oc.getNetworkName()
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		m := strings.Split(bridgeMapping, ":")
+		if m[0] == networkName {
```
maiqueb, line 418: should we also assert len(m) == 2 ? 

And let's try to use a better variable name. `networkToOVSBridgeAssociation`  ?... That's probably too verbose ..

test/e2e/multihoming.go
```diff
@@ -67,6 +67,13 @@ var _ = Describe("Multi Homing", func() {
 			netConfig.namespace = f.Namespace.Name
 			podConfig.namespace = f.Namespace.Name
 
+			if netConfig.topology == "localnet" {
+				By("applying ovs bridge mapping")
+				err := setBridgeMappings(cs, defaultNetworkBridgeMapping(), bridgeMapping(netConfig.networkName, bridgeName))
+				Expect(err).NotTo(HaveOccurred())
```
maiqueb, line 73: nit: couldn't you in-line it ?

test/e2e/multihoming.go
```diff
@@ -1812,3 +1819,19 @@ func createMultiNetworkPolicy(mnpClient mnpclient.K8sCniCncfIoV1beta1Interface,
...
+
+func setBridgeMappings(cs clientset.Interface, mappings ...BridgeMapping) error {
+	pods := ovsPods(cs)
+	if len(pods) == 0 {
+		return fmt.Errorf("pods list is empty")
+	}
+
+	for _, pods := range pods {
+		err := configureBridgeMappings(pods.Name, mappings...)
+		if err != nil {
```
maiqueb, line 1831: nit: couldn't you inline it ?

maiqueb (changes requested):
Nice ! 

We're nearly there. I have some questions, and suggestions, but most are opinionated nits.

Good work !

oshoval:
Thanks for the review

ovn-ci / Upgrade OVN from Master to PR branch based image (local) (pull_request)
job didnt run well, i think it isn't related to this PR, but for the record, once we have the next run we can compare
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13107856585/job/36566525622?pr=5000

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -105,6 +109,11 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) AddResource(obj interfa
 			return fmt.Errorf("could not cast %T object to Node", obj)
 		}
 		return h.oc.addUpdateNodeEvent(node)
+	case factory.PodType:
+		if err := h.oc.checkBridgeMapping(obj); err != nil {
+			return fmt.Errorf("failed AddResource %w", err)
```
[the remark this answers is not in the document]
oshoval: will add, might be redundant, because it wraps the error `checkBridgeMapping` returns, which is pod scoped already

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -194,13 +210,31 @@ func NewSecondaryLocalnetNetworkController(
...
+) *SecondaryLocalnetNetworkController {
 
 	stopChan := make(chan struct{})
 
+	ctx, cancel := context.WithCancel(context.Background())
+	if ovsClient == nil {
+		var err error
+		ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
+		if err != nil {
+			panic("failed to initialize libovsdb vswitchd client")
```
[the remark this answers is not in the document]
oshoval: i can change the function to return err, it doesnt do it atm
and then we can call cancel (with panic we didn't need of course)
on the good path, we do call cancel on `Stop` method

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: from what i read we need it to get the new data, but i will make sure
the current places on code that use it, do call it as well fwiw

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -639,3 +639,7 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 		)
 	})
 })
+
+func getExternalIdsCmd() string {
```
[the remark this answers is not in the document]
oshoval: oops, we dont need this func at all anymore, since it belong to the fexec mock which is dropped now

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: no need imho, since we are already in a small helper

oshoval:
Addressed comments
beside https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000#discussion_r1939156053 that i will check now

oshoval:
fixed linter (call cancel even that it is stale in case of this error)

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: interesting, with quick and dirty code, seems you are right

oshoval:
something is weird, i think this PR uncovers some bug maybe

lets say the mapping is missing, i delete the VM gracefully, add the mapping
it cause the ovnkube-nodes to crash, and to be created again, causing few seconds downtime
for example if i try to create new VM during this downtime it will fail because kubevirt webhook has timeout (seems the cluster is busy)

the biggest problem here is the fatal that cause the pod to crash

```
I0203 14:12:34.416534 1605933 event.go:377] Event(v1.ObjectReference{Kind:"Pod", Namespace:"ovn-kubernetes", Name:"ovnkube-node-b6qww", UID:"dfaed835-3763-48a1-bad6-402af41a6d8c", APIVersion:"v1", ResourceVersion:"3209554", FieldPath:""}): type: 'Warning' reason: 'ErrorReconcilingPod' failed to find bridge mapping for network: "localnet-network", node "ovn-control-plane", ovn-bridge-mappings: foo:bar,physnet:breth0
I0203 14:12:34.416481 1605933 obj_retry.go:620] Update event received for *factory.egressIPPod ovn-kubernetes/ovnkube-node-b6qww
I0203 14:12:34.631613 1605933 ovs.go:159] Exec(189): /usr/bin/ovs-vsctl --timeout=15 --if-exists get Interface patch-breth0_ovn-worker2-to-br-int ofport
I0203 14:12:34.641814 1605933 ovs.go:162] Exec(189): stdout: ""
I0203 14:12:34.641878 1605933 ovs.go:163] Exec(189): stderr: ""
E0203 14:12:34.641902 1605933 openflow_manager.go:265] Fatal error: patch port patch-breth0_ovn-worker2-to-br-int ofport changed from 15 to 
=============== pid 1605933 terminated ========== 

```

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -105,6 +109,11 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) AddResource(obj interfa
 			return fmt.Errorf("could not cast %T object to Node", obj)
 		}
 		return h.oc.addUpdateNodeEvent(node)
+	case factory.PodType:
+		if err := h.oc.checkBridgeMapping(obj); err != nil {
+			return fmt.Errorf("failed AddResource %w", err)
+		}
+		return h.oc.AddSecondaryNetworkResourceCommon(h.objType, obj)
```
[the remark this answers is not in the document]
maiqueb: I am OK with what you had though. Whatever you prefer. 

I think it might be more readable, but, there's the obvious tradeoff of if something else is added in the wrong place, the code will start misbehaving.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -186,21 +201,48 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) IsObjectInTerminalState
...
+	if ovsClient == nil {
+		var err error
+		ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
+		if err != nil {
+			cancel()
+			return nil, err
+		}
+	}
+
+	openvSwitch, err := ovsops.GetOpenvSwitch(ovsClient)
```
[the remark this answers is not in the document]
oshoval: interesting, but do we even support non IC atm ?
there are lots of known bugs there, so not sure we need to dive there ?

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +398,40 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	ovnBridgeMappings := oc.openvSwitch.ExternalIDs["ovn-bridge-mappings"]
+
+	networkName := oc.getNetworkName()
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		networkBridgeAssociation := strings.Split(bridgeMapping, ":")
+		if len(networkBridgeAssociation) == 2 && networkBridgeAssociation[0] == networkName {
+			return nil
+		}
+	}
```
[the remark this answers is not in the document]
oshoval: i prefer not on this PR please
if needed a follow-up can do it focused

go-controller/pkg/ovn/secondary_localnet_network_controller.go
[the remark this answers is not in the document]
qinqon: Well, the thing is that maybe we are doing all this at the wrong controller even if it works for IC, proper controller is the node controller I think.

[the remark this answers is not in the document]
maiqueb: Localnet is for sure a feature that should work in both IC and non-IC deployments.

And I think it's even weirder - if the network does not have subnets defined, I think this runs on the ovnkube-controllers. That's OK.

But if it has subnets, something we document, and is supported (only) for pods, it'll run as @qinqon 's saying: on the ovnkube-control-plane, which runs in a random node. 

@oshoval could you run a quick experiment and confirm this suspicion ? 

@qinqon good catch !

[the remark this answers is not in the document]
oshoval: i think we indeed saw this once, with other feature that depends on subnets was either on the ovnkube-control-plane or on the nodes

if `go-controller/hybrid-overlay/pkg/controller/ovn_node_linux.go` is the right place which always node specific
and would solve both IC, non IC by a chance, lets just move there because it is better anyhow ?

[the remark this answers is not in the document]
maiqueb: And for sure `go-controller/hybrid-overlay/pkg/controller/ovn_node_linux.go` is **not** the proper place ! That's for hybrid overlay (windows nodes IIUC). 

You're after something similar, but around the CNI package. I think it would have been https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/cni/helper_linux.go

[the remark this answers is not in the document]
oshoval: maybe we just need to read the pod, get the localnet network from there (bit more complex than current)
read the ovs db and assert

first question i guess is to find the right place(s), and then to select the one that is nicest
is the file above right even? (just if you know by heart please)

oshoval:
rebased only

maiqueb:
I'm OK if this PR focuses exclusively in IC. 

Seems there's no good place (other than CNI ...) to actually run non-IC for localnet in all nodes.

@jcaamano thoughts ? Is there a code path for non-IC in localnet that is executed in **all** nodes ? (other than CNI :) )

jcaamano:
> I'm OK if this PR focuses exclusively in IC.
> 
> Seems there's no good place (other than CNI ...) to actually run non-IC for localnet in all nodes.
> 
> @jcaamano thoughts ? Is there a code path for non-IC in localnet that is executed in **all** nodes ? (other than CNI :) )

Not sure I understand but you have a `SecondaryNodeNetworkController`. If you want one specific for localnet you can just add it like we have for the OVN controllers.

maiqueb:
> > I'm OK if this PR focuses exclusively in IC.
> > Seems there's no good place (other than CNI ...) to actually run non-IC for localnet in all nodes.
> > @jcaamano thoughts ? Is there a code path for non-IC in localnet that is executed in **all** nodes ? (other than CNI :) )
> 
> Not sure I understand but you have a `SecondaryNodeNetworkController`. If you want one specific for localnet you can just add it like we have for the OVN controllers.

Tks @jcaamano, your reply is what we're looking for.

oshoval:
After research and our discussion, created https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5055
Our suggestion is that this PR will focus on IC only, and if desired we can later extend it to non IC (as it require some deeper changes comparing to what this feature gives).

Will fix some bugs found on this PR, so PR is not ready yet
Thanks

go-controller/pkg/ovn/secondary_localnet_network_controller.go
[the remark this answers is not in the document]
oshoval: ref https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000#issuecomment-2653402719

oshoval:
Addressed Quique's cosmetic comments 
Added validation that the pod runs on the specific node where the controller run (to avoid cross checking other node's pods)
Added validation that the Nad is relevant to the pod, else it will cross check the mapping for each pod
Refactor some functions as helpers (as Miguel asked once on one of them)

oshoval:
fixed unit tests, and added a negative test
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/8c6d4b639ff5a6eddf5adcf4dd43ddb926018df8..e6c42b9a889d7941b9be3801b0fb1f32a29dc7c0

go-controller/pkg/ovn/address_set/fake_address_set.go
```diff
@@ -248,6 +249,13 @@ func (f *FakeAddressSetFactory) EventuallyExpectAddressSetWithAddresses(dbIDsOrN
 	}).Should(gomega.Succeed())
 }
 
+func (f *FakeAddressSetFactory) EventuallyExpectAddressSetWithAddressesFailure(dbIDsOrNsName any, addresses []string) {
```
maiqueb, line 252: should we maybe use ginkgo helper here ?

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -540,9 +541,10 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 		})
 
 		ginkgo.DescribeTable("correctly adds and deletes pod IPs from secondary network namespace address set",
-			func(topology string, remote bool) {
+			func(topology string, remote bool, expectMissingMappingFailure bool) {
```
maiqueb, line 544: oh boy ... could we maybe turn this into a builder ?... If you do, instead of adding a boolean, pass directly the expected failure or something like that.

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -540,9 +541,10 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
...
 				app.Action = func(ctx *cli.Context) error {
 					var err error
+					os.Setenv("K8S_NODE", "node1")
```
maiqueb, line 547: shouldn't we clear this also on defer cleanup ?...

I worry this'll be a side effect that might impact other tests running in parallel.

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -616,8 +624,13 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 						gomega.Expect(err).NotTo(gomega.HaveOccurred())
 					}
 
-					ocInfo.asf.EventuallyExpectAddressSetWithAddresses(namespaceName1, []string{"10.1.1.1"})
-
+					if !expectMissingMappingFailure {
+						ocInfo.asf.EventuallyExpectAddressSetWithAddresses(namespaceName1, []string{"10.1.1.1"})
+					} else {
+						ocInfo.asf.EventuallyExpectAddressSetWithAddressesFailure(namespaceName1, []string{"10.1.1.1"})
+						gomega.Expect(fakeOvn.fakeRecorder.Events).To(gomega.HaveLen(1))
```
maiqueb, line 631: this one is redundant right ? We can have only the one below.

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -616,8 +624,13 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
...
+						gomega.Expect(<-fakeOvn.fakeRecorder.Events).To(gomega.Equal(`Warning ErrorReconcilingPod failed to find bridge mapping for network: "network1", node "node1", ovn-bridge-mappings: ""`))
```
maiqueb, line 632: OK, let's discuss the error msg here ...

```suggestion
						gomega.Expect(<-fakeOvn.fakeRecorder.Events).To(gomega.Equal(`Warning ErrorReconcilingPod failed to find bridge mapping for network: "network1" on node "node1"; Current ovn-bridge-mappings: ""`))
```

It's pretty much the same, but I think it reads better.

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -630,12 +643,13 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
...
-			ginkgo.Entry("on remote zone for layer2 topology", ovntypes.Layer2Topology, true),
-			ginkgo.Entry("on local zone for localnet topology", ovntypes.LocalnetTopology, false),
-			ginkgo.Entry("on remote zone for localnet topology", ovntypes.LocalnetTopology, true),
+			ginkgo.Entry("on local zone for layer3 topology", ovntypes.Layer3Topology, false, false),
+			ginkgo.Entry("on remote zone for layer3 topology", ovntypes.Layer3Topology, true, false),
+			ginkgo.Entry("on local zone for layer2 topology", ovntypes.Layer2Topology, false, false),
+			ginkgo.Entry("on remote zone for layer2 topology", ovntypes.Layer2Topology, true, false),
+			ginkgo.Entry("on local zone for localnet topology", ovntypes.LocalnetTopology, false, false),
+			ginkgo.Entry("on remote zone for localnet topology", ovntypes.LocalnetTopology, true, false),
+			ginkgo.Entry("on remote zone for localnet topology, skip bridge mapping, expect failure", ovntypes.LocalnetTopology, true, true),
```
maiqueb, line 652: Thanks for adding this one ! Nice.

go-controller/pkg/ovn/ovn_test.go
```diff
@@ -564,7 +566,19 @@ func (o *FakeOVN) NewSecondaryNetworkController(netattachdef *nettypes.NetworkAt
...
+			testDB := []libovsdbtest.TestData{
+				&vswitchd.OpenvSwitch{
+					ExternalIDs: o.externalIDS,
+				},
+			}
+			dbSetup := libovsdbtest.TestSetup{
+				OVSData: testDB,
+			}
+			ovsClient, _, err := libovsdbtest.NewOVSTestHarness(dbSetup)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
```
maiqueb, line 578: maybe move this into an helper to hide away some of this complexity ? 

Something like `setupOVSWithExternalIDs` or similar.

nit: you don't actually need the `testDB` variable, do you ? You can just inline it below, right ?

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +402,77 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+
+	if !util.PodScheduled(pod) {
+		return nil
+	}
+
+	if oc.nodeName != pod.Spec.NodeName {
+		return nil
+	}
+
+	knownNad, err := oc.podAttachedToKnownNAD(pod)
```
maiqueb, line 428:
```suggestion
	isNADKnown, err := oc.podAttachedToKnownNAD(pod)
```

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +402,77 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+		return err
+	}
+	if !knownNad {
+		return nil
+	}
+
+	return oc.networkMappedToBridge(pod)
+}
+
+func (oc *SecondaryLocalnetNetworkController) podAttachedToKnownNAD(pod *corev1.Pod) (bool, error) {
```
maiqueb, line 439:
```suggestion
func (oc *SecondaryLocalnetNetworkController) isPodAttachedToKnownNAD(pod *corev1.Pod) (bool, error) {
```

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -186,21 +203,49 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) IsObjectInTerminalState
...
+	cnci *CommonNetworkControllerInfo,
+	netInfo util.NetInfo,
+	networkManager networkmanager.Interface,
+	ovsClient client.Client,
+) (*SecondaryLocalnetNetworkController, error) {
 
 	stopChan := make(chan struct{})
 
+	ctx, cancel := context.WithCancel(context.Background())
+	if ovsClient == nil {
```
qinqon, line 230: There is something weird here if `ovsClient != nil` we are going to build a cancel context not related to ovsClient, I understand that this is for unit test and all, but feels like cancel context from ovsclient has to be passed with it or build if nil.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +402,77 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+}
+
+func (oc *SecondaryLocalnetNetworkController) checkBridgeMapping(obj interface{}) error {
+	if !config.OVNKubernetesFeature.EnableInterconnect {
+		return nil
+	}
+
+	pod, isPod := obj.(*corev1.Pod)
+	if !isPod {
+		return fmt.Errorf("could not cast %T object to *knet.Pod", obj)
```
qinqon, line 417: ovn-k error are most of the time prefixed with "failed"
```suggestion
		return fmt.Errorf("failed casting %T object to *knet.Pod", obj)
```

maiqueb (changes requested):
Pretty much just nitpicking on your unit tests ... 

Everything else looks good to me. Most of my comments are opinionated.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -353,11 +402,77 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+		return nil
+	}
+
+	knownNad, err := oc.podAttachedToKnownNAD(pod)
+	if err != nil {
+		return err
+	}
+	if !knownNad {
+		return nil
+	}
```
qinqon, line 434: Do we really need this ? if we are at this controller it means the pod has already a localnet NAD assigned.

qinqon (changes requested):

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: we dont cancel it for UT (UT cancel it via a chan, shared between the fake server and client)
it is because otherwise the linter fails for context that wasnt cancel.
We do it only if the current function have failures, otherwise we set it in the struct, and Stop function will call it.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -222,6 +267,9 @@ func NewSecondaryLocalnetNetworkController(
 				},
 			},
 		},
+		openvSwitch:     openvSwitch,
+		ovsClientCancel: cancel,
+		nodeName:        os.Getenv("K8S_NODE"),
```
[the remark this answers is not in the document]
oshoval: but it is a syscall that cost a little,
and on IC, we just skip currently the logic, so found it better to do it one timer
it also available both on OpenShift and K8s
unless you insist of course

go-controller/pkg/ovn/multipolicy_test.go
oshoval: it checks the Event was fired, the previous check that there was an error
they are not the same (i.e if we remove the recordError, it can still fail the flow because the mapping wasnt found and the function returns error)
behind the scene, it also helps us, because if the test runs too fast, the event wont manage to be recorded,
so the above line also check Consistently for 2 seconds that there is a failure, which adds the required delay
(reminds the sleep we have between the nad creation and the pod creation in the tests)

go-controller/pkg/ovn/secondary_localnet_network_controller.go
qinqon: Ahh right, all good.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: we initialize it once, instead calling it everytime,
there is quite lot of logic there
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/dd2a70c928d362a8d471352e7ec6e07164064e9f/go-controller/pkg/libovsdb/libovsdb.go#L231
i think it is nicer as now (Miguel suggested that)

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: btw if we can improve the controller to get events on specific pods only (i.e if they have some specific label or such), it can improve stuff in general, but this isnt the scope of this PR

go-controller/pkg/ovn/secondary_localnet_network_controller.go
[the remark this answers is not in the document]
qinqon: then rename it to something specific like controllerPodNodeName, so is clear what node are we talking about.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
qinqon: Ahh wait a minute... looks like we don't use the client anymore after initializing it, we just retrieve the OpenSwitch stuff.

This means that we can be not detecting the problem since we are not doing GetOpenvSwitch at the check function.

So even if they create the ovn-mapping after controller starts it's not going to be detected.

qinqon (changes requested):
We should have the ovsClient as the member of the controller not the OpenvSwitch struct to read the state everytime we check.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: i made sure, we do get updates in real time with the current PR
the current logic is good enough in this aspect in production code

go-controller/pkg/ovn/address_set/fake_address_set.go
oshoval: wdym please ?
but imo if it is nit, i would prefer to keep it simple, unless there is something neat and not complex

go-controller/pkg/ovn/secondary_localnet_network_controller.go
qinqon: That's weird, like OpenvSwitch object get updated behind the scenes ? if this is the case then I am fine, although if feels a little dangerous.

oshoval: yep it does,
i tried to run the current solution, let it fail once, add the mapping 
recreate the pod, and it now found the ovs mapping

go-controller/pkg/ovn/address_set/fake_address_set.go
maiqueb: since you're using an expectation from within an helper, this will mess up the stack trace of the failure.

You either use `GinkgoHelper` *or* set an offset on the caller - https://pkg.go.dev/github.com/onsi/gomega#ConsistentlyWithOffset

go-controller/pkg/ovn/secondary_localnet_network_controller.go
oshoval: about ovsClient versus openvSwitch
GetOpenvSwitch does a list, which returns a pointer to the cache, it can answer your question about how it updates,
Miguel suggested and i agree that we can just skip that
please decide, my vote is keep as is, we don't need to be over afraid imho

go-controller/pkg/ovn/multipolicy_test.go
maiqueb: I think I didn't manage to express myself properly, I apologize for that. 

I mean to say `gomega.Expect(fakeOvn.fakeRecorder.Events).To(gomega.HaveLen(1))` can be removed, and you can leave only `gomega.Expect(<-fakeOvn.fakeRecorder.Events).To(gomega.Equal(`Warning ErrorReconcilingPod failed to find bridge mapping for network: "network1", node "node1", ovn-bridge-mappings: ""`))` . Right ? 

Like ... both of these check the event was fired - one checks the event queue has a single event, the other that we receive the event we're after. I think we can leave only the former.

oshoval: well if there arent event it will block (unless we use select) as far as i know, moreover we want that there will be only one event, else the other might be noise, and also their order can cause noise

go-controller/pkg/ovn/secondary_localnet_network_controller.go
qinqon: I understand how it updates it but that means we depend on the implementation, like for example they can change the implementation in the future and return new pointers instead of exposing internal cache.

maiqueb: let's do whatever it is other people are doing. That seems to be what @qinqon 's suggesting, i.e. calling `GetOpenvSwitch` .

oshoval:
Thanks for the review
addressed comments (1st push, 2nd is rebase) beside
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000#discussion_r1957956090
will address it soon

qinqon (approved):
LGTM

oshoval:
Thanks
something is weird
```
2025-02-17T16:44:56.8706555Z   [38;5;9m[FAIL][0m [0mOVN master EgressIP Operations cluster default network [38;5;243mOn node UPDATE [0m[secondary host network] should perform proper OVN transactions when namespace and pod is created after node egress label switch [38;5;9m[1m[It] interconnect enabled; node1 in global and node2 in remote zones[0m
2025-02-17T16:44:56.8708242Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:2133[0m
```
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13374607714/job/37350900984?pr=5000
fails lately always it seems, will check if it pass on clean main,
locally it passed as far as i tried so far

oshoval:
did the refactor you asked hopefully @maiqueb 
(just `go-controller/pkg/ovn/multipolicy_test.go` was changed, rest by mistake are rebase)

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -630,12 +644,52 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
...
+	b.remote = true
+	return b
+}
+
+func (b *TestCaseBuilder) ExpectFailure(msg string) *TestCaseBuilder {
+	b.expectedError = msg
+	return b
+}
+
+func (b *TestCaseBuilder) Build() ginkgo.TableEntry {
```
maiqueb, line 681:
```suggestion
func (b *TestCaseBuilder) Succeed() ginkgo.TableEntry {
```

Or `Success`. Something to indicate it "worked as expected".

go-controller/pkg/ovn/multipolicy_test.go
maiqueb: OK, I give up.

go-controller/pkg/ovn/multipolicy_test.go
oshoval: this func "builds" the entry
```
return ginkgo.Entry(
		fmt.Sprintf("on %s zone for %s topology", zone, b.topology), b.topology, b.remote, b.mapping, b.expectedError,
	)
```
one of the lines should fail, so Succeed might be weird ? (also since it just render, it makes build to sounds weird imho)
anyhow whatever you prefer

one option is to create `Failed` that equal this func, and then we can rename it to Succeed
and use `Failed` for the one expects failure

maiqueb: So, imho, you should also use `build` to build the entry that expects failure, right ?

oshoval: i do aint it ? just on sep line
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/files#diff-6866db9fa9a4045b3db6ed1e842dc2a1c90a1565581a2e8db5f0c36adfa6e943R655

maiqueb: :facepalm: 

Right, sorry. 

I had missed the `Build` on the separate line, my bad. You can disregard this conversation.

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -630,12 +644,52 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
...
 })
+
+type TestCaseBuilder struct {
+	topology      string
+	remote        bool
+	expectedError string
+	mapping       string
+}
+
+func NewTestCase(topology string) *TestCaseBuilder {
```
maiqueb, line 667: How about we try this one instead ? https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c

maiqueb: this is opinionated, you can choose to ignore.

oshoval: will consider for next times if its ok
thanks

maiqueb: All good.

maiqueb (approved):
Thank you !

maiqueb:
@oshoval could you check the failing e2e tests are flakes ? 

@ricky-rav could you review this PR ?... 

Thanks in advance !

oshoval:
Thanks
all seems to be flakes / errors that are not related to this PR
rebasing for rerun

go-controller/pkg/ovn/multipolicy_test.go
ricky-rav: Uhm ok, so we are making sure there's only one event and then that it's the event we're expecting. All good.

ricky-rav (approved):
/LGTM

Nothing to add, the existing conversation already covered the main points to discuss. 
Thanks, Or!

oshoval:
Thanks, rebased

oshoval:
the failure is a known flake
https://github.com/ovn-kubernetes/ovn-kubernetes/issues?q=is%3Aissue%20state%3Aopen%20disabling%20egress%20nodes%20impeding%20GRCP%20health%20check%20

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -186,21 +203,43 @@ func (h *secondaryLocalnetNetworkControllerEventHandler) IsObjectInTerminalState
...
+	networkManager networkmanager.Interface,
+	ovsClient client.Client,
+) (*SecondaryLocalnetNetworkController, error) {
 
 	stopChan := make(chan struct{})
 
+	ctx, cancel := context.WithCancel(context.Background())
+	if ovsClient == nil {
+		var err error
+		ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
```
jcaamano, line 232: We should be using the OVS client that we already initialize in `ovnkube.go`. It brings a cache with it and we don't want dup caches.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -360,11 +403,82 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	if oc.controllerNodeName != pod.Spec.NodeName {
+		return nil
+	}
+
+	isNADKnown, err := oc.isPodAttachedToKnownNAD(pod)
+	if err != nil {
+		return err
+	}
+	if !isNADKnown {
+		return nil
```
jcaamano, line 434: What happens if the NAD becomes known and our business logic should apply? We don't reconcile this. I guess not a major issue if we are just reporting an event...until we do something more and then it becomes an issue.
I guess I am just interested if this was considered for now.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -360,11 +403,82 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		networkBridgeAssociation := strings.Split(bridgeMapping, ":")
+		if len(networkBridgeAssociation) == 2 && networkBridgeAssociation[0] == networkName {
+			return nil
+		}
+	}
+
+	err = fmt.Errorf("failed to find bridge mapping for network: %q on node %q; Current ovn-bridge-mappings: %q", networkName, oc.controllerNodeName, ovnBridgeMappings)
+	oc.recordPodErrorEvent(pod, err)
```
jcaamano, line 475: So if I get this right, the only thing we pretend to do with all this is report an event? We are not blocking or affecting the status of the pod otherwise? If the bridge mapping is not there as reported, wouldn't the CNI side of things fail anyway? Kinda interest on what the global scenario looks like.

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -360,11 +403,82 @@ func (oc *SecondaryLocalnetNetworkController) newRetryFramework(
...
+	return err
+}
+
+func (oc *SecondaryLocalnetNetworkController) findNetworkName() string {
```
jcaamano, line 479: nit: `gtPhysicalNetworkName`

jcaamano (changes requested):

go-controller/pkg/ovn/secondary_localnet_network_controller.go
maiqueb: Right @jcaamano .

It could make sense to flat out crashloop the pod - but we plan in the future to impact scheduling. Meaning, the workload would only be scheduled on a node featuring the required mapping.

This would be a first step, to show the users why their pod's network is broken.

jcaamano: Interested in `If the bridge mapping is not there as reported, wouldn't the CNI side of things fail anyway?` also

maiqueb: no, it doesn't fail. The localnet port will be created, but the patch port will not, which will black-hole inter-node traffic, while working for workloads scheduled in the same node.

jcaamano: What prevents us to do exactly the same check on the CNI side which would be more straight forward because everything its synchronous there? What are we missing?

maiqueb: Do we even have access to OVS from the CNI side of things ?

maiqueb: @oshoval would you remind me if we actually evaluated doing this from the CNI side ? 

I don't have anything recorded on my notes, and I'm afraid to mix this up w/ the ipv6 story.

oshoval: dont fully remember, if i will recall i will update

but on the CNI, unless the following info is missing:
We will need to find the network that is used by the nads pointed by the NetworkSelectionElements, and to check if this one is a local net

Note, i think i did we dont just report an event but also fails the creation, but will need to double check it if desired

maiqueb: IIUC you can use the `networkManager` [here](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/cni/cni.go#L118) to get the NetInfo, from which you can access the topology.

maiqueb: > Note about https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/files#r2000973063, i think i did we dont just report an event but also fails the creation, but will need to double check it if desired

Please check ! This one's important.

maiqueb: Regarding the architecture of this suggestion, doing it in the CNI will pretty much involve adding code common to all the type of networks, which will surely add a performance penalty (... for all types of networks). I think we are using listers, hence, it shouldn't be much, but still ... it's something. At the very least - IMHO - we need to make sure we're not impacting the default cluster network.

Furthermore, the current proposal, which only touches localnet specific code. IMHO, this suggestion is cleaner, and potentially performs better. (I am aware that it's hard to argue for / against without figures ...)

Not exactly sure implementing it in the CNI is more straightforward, but I can see a tradeoff. 
@jcaamano let us know if you really want us to explore this path further.

oshoval: > This implementation here is racey so not functionally correct. If we are not aware of the NAD yet, we might not report the event when we should have. If we are not aware of the bridge mapping yet, we might report the event when we shouldn't have. Once we report, we can't unreport.

If there is no NAD, the pod wont get scheduled because other things and there is retry isnt it ? (i think i saw a retry, we can even try to create the mapping after the event and see if the pod self heal, maybe tried already)
If there is no bridge mapping yet, it is marginal that can also happen on CNI, and there is retry

qinqon: I remember we found that doing it at CNI was quite involved since we have to add listeners to the CNI server, although it's true that it will be clear to the user (no ovn mapping, no pod).

Since this PR is quite old maybe now things has improve and we need less plubmbing and CNI server ?

oshoval: > Please check ! This one's important.

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/files#diff-87fd62ad4d80ef1e2da6f02a7bdc5e4910bd30b1565e1042feba949dfdd23772R117
from what i remember and as we see here 
error + event was the requirements (at least back then of course)

jcaamano: > If there is no NAD, the pod wont get scheduled

But the important thing here is the order in which different components become aware of things, not he order in which they actually happen. So the pod can be scheduled, but ourselves may not yet be aware of the existence of the NAD.

> we have to add listeners to the CNI server

Can you elaborate? I don't see the need, perhaps I am missing something.

> https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/files#diff-87fd62ad4d80ef1e2da6f02a7bdc5e4910bd30b1565e1042feba949dfdd23772R117

So the only effect that has is recording the event up to 15 times or so which are our maximum retries. This is not failing anything else since we are not doing anything else. Meaning, not failing anything from the user perspective.

maiqueb: sure ! right. I'm on a roll today :sweat_smile: 

To sum up: we get access to the topology attribute for free, from the configuration. There's no performance impacts of doing this in the CNI side. As a bonus, this means that we don't have to reconcile NADs at all.

@oshoval I think we should at least try to do this on the CNI side of things.

oshoval:
Creating https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145 instead (via CNI), WIP
closing this one, thanks

PR #5001, opened 2025-01-23
Accept no active network during pod deletion in the IP allocator

kyrtapz:
There is a chance that the pod deletion will come after the primary network is already removed so we cannot rely on `getActiveNetworkForPod` returning no error during pod deletion.

### DIFF
--- a/go-controller/pkg/clustermanager/pod/allocator.go
+++ b/go-controller/pkg/clustermanager/pod/allocator.go
@@ -204,9 +204,33 @@ func (a *PodAllocator) reconcile(old, new *corev1.Pod, releaseFromAllocator bool
 		return nil
 	}
 
-	activeNetwork, err := a.getActiveNetworkForPod(pod)
-	if err != nil {
-		return fmt.Errorf("failed looking for an active network: %w", err)
+	var activeNetwork util.NetInfo
+	var err error
+
+	if new != nil {
+		activeNetwork, err = a.getActiveNetworkForPod(pod)
+		if err != nil {
+			return fmt.Errorf("failed looking for an active network: %w", err)
+		}
+	} else if a.netInfo.IsPrimaryNetwork() {
+		// During pod deletion, the UDN might already be removed. To handle this, check if the activeNetwork
+		// is managed by the current allocator using pod annotations. If not, exit as there is nothing to do.
+		// There is no need to handle non-primary networks as they are already present in the network-selection
+		// annotations handled in GetPodNADToNetworkMappingWithActiveNetwork.
+		podNetworks, err := util.UnmarshalPodAnnotationAllNetworks(pod.Annotations)
+		if err != nil {
+			return err
+		}
+		for nadName := range podNetworks {
+			if a.netInfo.HasNAD(nadName) {
+				activeNetwork = a.netInfo
+				break
+			}
+		}
+		if activeNetwork == nil {
+			klog.V(5).Infof("Skipping pod cleanup for %s/%s on network %s", pod.Namespace, pod.Name, a.netInfo.GetNetworkName())
+			return nil
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


go-controller/pkg/clustermanager/pod/allocator.go
```diff
@@ -204,9 +204,33 @@ func (a *PodAllocator) reconcile(old, new *corev1.Pod, releaseFromAllocator bool
...
+				break
+			}
+		}
+		if activeNetwork == nil {
+			klog.V(5).Infof("Skipping pod cleanup for %s/%s on network %s", pod.Namespace, pod.Name, a.netInfo.GetNetworkName())
+			return nil
+		}
 	}
 
 	onNetwork, networkMap, err := util.GetPodNADToNetworkMappingWithActiveNetwork(pod, a.netInfo, activeNetwork)
```
trozet, line 236: @maiqueb @qinqon wouldn't we also want to release IPs for pods on secondary networks as well? This only gets active primary network and releases it?

trozet: or maybe it does include all secondary networks as well? It's really not clear to me in the function:

```
func GetPodNADToNetworkMappingWithActiveNetwork(pod *kapi.Pod, nInfo NetInfo, activeNetwork NetInfo) (bool, map[string]*nettypes.NetworkSelectionElement, error) {
	on, networkSelections, err := GetPodNADToNetworkMapping(pod, nInfo)
	if err != nil {
		return false, nil, err
	}

	if activeNetwork == nil {
		return on, networkSelections, nil
	}

	if activeNetwork.IsDefault() ||
		activeNetwork.GetNetworkName() != nInfo.GetNetworkName() ||
		nInfo.TopologyType() == types.LocalnetTopology {
		return on, networkSelections, nil
	}

	// Add the active network to the NSE map if it is configured
	activeNetworkNADs := activeNetwork.GetNADs()
	if len(activeNetworkNADs) < 1 {
		return false, nil, fmt.Errorf("missing NADs at active network %q for namespace %q", activeNetwork.GetNetworkName(), pod.Namespace)
	}
	activeNetworkNADKey := strings.Split(activeNetworkNADs[0], "/")
	if len(networkSelections) == 0 {
		networkSelections = map[string]*nettypes.NetworkSelectionElement{}
	}
	networkSelections[activeNetworkNADs[0]] = &nettypes.NetworkSelectionElement{
		Namespace: activeNetworkNADKey[0],
	}

###TROZET: How is this safe to overwrite the first index?^

	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) {
		if wasPersistentIPRequested {
			networkSelections[activeNetworkNADs[0]].IPAMClaimReference = ipamClaimName
		}
	}

	return true, networkSelections, nil
}
```

trozet: oh i guess if we get thsi far in the function, then this must be a primary UDN netinfo, and there cant be an attachment to a primary UDN  as well as another attachment to the same network via a secondary NAD?

trozet: nvm i'm an idiot...tgif :)

go-controller/pkg/clustermanager/pod/allocator.go
```diff
@@ -204,9 +204,33 @@ func (a *PodAllocator) reconcile(old, new *corev1.Pod, releaseFromAllocator bool
...
+		if err != nil {
+			return err
+		}
+		for nadName := range podNetworks {
+			if a.netInfo.HasNAD(nadName) {
+				activeNetwork = a.netInfo
+				break
+			}
+		}
+		if activeNetwork == nil {
```
trozet, line 230: @maiqueb @jcaamano do we support a network being primary and secondary for the pod?

trozet (approved):

PR #5004, opened 2025-01-27
Handles unspecified protocol in network policy port

trozet:
Fixes a null pointer exception when network policy port has no protocol. If the protocol is missing in the network policy port definition, it should be assumed to be TCP.

Hit the crash downstream in:
https://issues.redhat.com/browse/OCPBUGS-49393

### DIFF
--- a/go-controller/pkg/ovn/gress_policy.go
+++ b/go-controller/pkg/ovn/gress_policy.go
@@ -14,6 +14,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 
+	v1 "k8s.io/api/core/v1"
 	knet "k8s.io/api/networking/v1"
 	utilnet "k8s.io/utils/net"
 )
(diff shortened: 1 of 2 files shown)


go-controller/pkg/ovn/gress_policy.go
```diff
@@ -102,12 +103,16 @@ func (gp *gressPolicy) addPeerAddressSets(asHashNameV4, asHashNameV6 string) {
 // If the port is not specified, it implies all ports for that protocol
 func (gp *gressPolicy) addPortPolicy(portJSON *knet.NetworkPolicyPort) {
 	var pp *libovsdbutil.NetworkPolicyPort
+	protocol := v1.ProtocolTCP
```
jluhrsen, line 106: so it's always gotta be TCP?

trozet: yeah kube defaults it to TCP for network policy for us in the API, but multinetwork policy does not

trozet:
> Thank you.
> 
> @trozet are you planning on addressing the scenario of the empty pod selector as well ?

I wasn't planning to @maiqueb Let me go back and revisit it. I guess it can be a different PR.

maiqueb:
> > Thank you.
> > @trozet are you planning on addressing the scenario of the empty pod selector as well ?
> 
> I wasn't planning to @maiqueb Let me go back and revisit it. I guess it can be a different PR.

Surely. It can take place in a follow-up.

PR #5007, opened 2025-01-28
Reconcile pending pods when a NAD is added to an existing network

kyrtapz:
Previously, if a new NAD was added to an existing network after a pod referencing it, the pod would never start. This is fixed by reconciling pending pods when the secondary network controller reconciles a new NAD.

The e2e test is failing without the controller change, the pod never becomes ready.

cc: @trozet @jcaamano

### DIFF
--- a/go-controller/pkg/clustermanager/network_cluster_controller.go
+++ b/go-controller/pkg/clustermanager/network_cluster_controller.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/klog/v2"
 
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/id"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/ip/subnet"
 	annotationalloc "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/pod"
@@ -438,10 +439,16 @@ func (ncc *networkClusterController) Cleanup() error {
 }
 
 func (ncc *networkClusterController) Reconcile(netInfo util.NetInfo) error {
-	// update network informaiton, point of no return
+	reconcilePendingPods := !ncc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+	// update network information, point of no return
 	err := util.ReconcileNetInfo(ncc.ReconcilableNetInfo, netInfo)
 	if err != nil {
 		klog.Errorf("Failed to reconcile network %s: %v", ncc.GetNetworkName(), err)
+	}
+	if reconcilePendingPods {
+		if err := objretry.RequeuePendingPods(ncc.kube, ncc.GetNetInfo(), ncc.retryPods); err != nil {
+			klog.Errorf("Failed to requeue pending pods for network %s: %v", ncc.GetNetworkName(), err)
+		}
 	}
 	return nil
 }
--- a/go-controller/pkg/ovn/base_network_controller.go
+++ b/go-controller/pkg/ovn/base_network_controller.go
@@ -12,6 +12,18 @@ import (
 	"github.com/ovn-org/libovsdb/ovsdb"
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 
+	kapi "k8s.io/api/core/v1"
+	knet "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/sets"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	ref "k8s.io/client-go/tools/reference"
+	"k8s.io/klog/v2"
+	utilnet "k8s.io/utils/net"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/allocator/pod"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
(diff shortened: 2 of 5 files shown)


go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -197,7 +198,11 @@ type BaseSecondaryNetworkController struct {
 }
 
 func (oc *BaseSecondaryNetworkController) Reconcile(netInfo util.NetInfo) error {
-	return util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	err := util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	if err != nil {
+		return err
+	}
+	return oc.addAllPendingPods()
```
jcaamano, line 205: Just note that this Reconcile might be called when something else on the NetInfo changed other than the NADs. So you might be calling `addAllPendingPods` when there is no need.
If you are fine with that, ok.

Other option is to compare the NADs in the new NetInfo against the old NetInfo before `ReconcileNetInfo, and just call `addAllPendingPods` if those specifically changed. That makes `ReconcileNetInfo` a point of no return, since further comparison would see no change: if `addAllPendingPods` fails, then the only thing you could do is log.

There is no mandate, but the second approach is the one we have followed elsewhere.

kyrtapz: Done.
Updated the PR to handle the cluster-manager ctrl too.

go-controller/pkg/retry/obj_retry.go
```diff
@@ -755,3 +761,41 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
...
+		if err != nil {
+			return nil, err
+		}
+		allPods = append(allPods, pods...)
+	}
+	return allPods, nil
+}
+
+// RequeuePendingPodsForNetwork enqueues all Pending pods from the namespaces associated with the given netInfo into the retryPods.
+func RequeuePendingPodsForNetwork(kubeClient kube.InterfaceOVN, netInfo util.NetInfo, retryPods *RetryFramework) error {
```
maiqueb, line 782: how will this work for pods located on other namespaces, but attempting to attach themselves to a network defined on a shared namespace ? For instance, the `default` namespace ?

go-controller/pkg/retry/obj_retry.go
```diff
@@ -755,3 +761,41 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
 
 	return handler, nil
 }
+
+// getPendingPodsForNetwork returns pods from netInfo namespaces that are in the Pending state
+func getPendingPodsForNetwork(kubeClient kube.InterfaceOVN, netInfo util.NetInfo) ([]*kapi.Pod, error) {
+	var allPods []*kapi.Pod
+
+	for _, namespace := range netInfo.GetNamespaces() {
```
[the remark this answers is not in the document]
jcaamano: Yea so probably we should rename it `GetNADNamespaces`. In order to know what pods are served by a network, this is only useful fro primary networks.

go-controller/pkg/retry/obj_retry.go
kyrtapz: I wasn't aware that this was possible. Modified the code to handle all pending pods.

go-controller/pkg/retry/obj_retry.go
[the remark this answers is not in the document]
kyrtapz: Renamed `GetNamespaces` to `GetNADNamespaces` and modified the code to handle all pending pods.

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -438,11 +439,15 @@ func (ncc *networkClusterController) Cleanup() error {
...
 func (ncc *networkClusterController) Reconcile(netInfo util.NetInfo) error {
-	// update network informaiton, point of no return
+	reconcilePendingPods := !ncc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+	// update network information, point of no return
 	err := util.ReconcileNetInfo(ncc.ReconcilableNetInfo, netInfo)
 	if err != nil {
 		klog.Errorf("Failed to reconcile network %s: %v", ncc.GetNetworkName(), err)
 	}
+	if reconcilePendingPods {
+		return objretry.RequeuePendingPodsForNetwork(ncc.kube, ncc.GetNetInfo(), ncc.retryPods)
```
jcaamano, line 449: Would just log this error.

Returning this error would cause a noop retry on which `reconcilePendingPods` would no longer be true and nothing would be done.

The real issue lies on the retry framework, re-queue should just not fail.

jcaamano: Same for other instances.

go-controller/pkg/retry/obj_retry.go
```diff
@@ -755,3 +761,40 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
 
 	return handler, nil
 }
+
+// getPendingPodsForNetwork returns all pods that are in the Pending state
+func getPendingPodsForNetwork(kubeClient kube.InterfaceOVN) ([]*kapi.Pod, error) {
```
jcaamano, line 766: `getPendingPodsForNetwork` mighty be a tad misleading. You didn't like just `getPendingPods`?

go-controller/pkg/retry/obj_retry.go
```diff
@@ -755,3 +761,40 @@ func (r *RetryFramework) WatchResourceFiltered(namespaceForFilteredHandler strin
...
+
+	// NOTE: A pod may reference a NAD from a different namespace, so check all pending pods.
+	allPods, err := getPendingPodsForNetwork(kubeClient)
+	if err != nil {
+		return err
+	}
+
+	for _, pod := range allPods {
+		pod := *pod
+		klog.Infof("Adding pending pod %s/%s to retryPods for network %s", pod.Namespace, pod.Name, netInfo.GetNetworkName())
```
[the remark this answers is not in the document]
jcaamano: probably move this log to the network controller itself

jcaamano (approved):

trozet (approved):

go-controller/pkg/retry/obj_retry.go
trozet: @maiqueb we don't support multus namespace isolation being false. If we did that it breaks the entire paradigm of UDN. A pod in another NS could connect to a S-UDN created by a tenant in their NS.

maiqueb: OK, I guess I understand the reasoning ... but won't that break existing users ?... 

I can straight away tell that in OpenShift virt plenty of customers use a single shared NAD (in the default namespace) and attach to that from multiple namespaces. 

This happens for secondary UDNs only FWIW.

PR #5013, opened 2025-01-29
Fix getting active network secondary udn

trozet:
Fixes issues where we are calling getActiveNetworkForNamespace on secondary UDN controllers. There is no point in doing this and can cause errors for resources like pods if we are handling pods slowly and the namespace has already been deleted. Remember secondary UDN/NAD can use other namespaces, so NAD/network controller may still be functional trying to add pods for another namespace and that namespace may no longer exist.

Fixes issues where we are starting services and multinetworkpolicy controllers when we shouldn't be.

There were duplicate functions for GetNetworkRole, with ambiguous logic that needed a note to caution developers against when to use it. Consolidated the function into a single util, and made the logic more explicit. Now if a GetNetworkRole is called on a pod that the controller is not serving, it will return "none", indicating to the caller that this  pod is not served by this controller and can be ignored.

### DIFF
--- a/go-controller/pkg/clustermanager/pod/allocator.go
+++ b/go-controller/pkg/clustermanager/pod/allocator.go
@@ -109,53 +109,17 @@ func (a *PodAllocator) getActiveNetworkForPod(pod *corev1.Pod) (util.NetInfo, er
 
 }
 
-// GetNetworkRole returns the role of this controller's
-// network for the given pod
-// Expected values are:
-// (1) "primary" if this network is the primary network of the pod.
-//
-//	The "default" network is the primary network of any pod usually
-//	unless user-defined-network-segmentation feature has been activated.
-//	If network segmentation feature is enabled then any user defined
-//	network can be the primary network of the pod.
-//
-// (2) "secondary" if this network is the secondary network of the pod.
-//
-//	Only user defined networks can be secondary networks for a pod.
-//
-// (3) "infrastructure-locked" is applicable only to "default" network if
-//
-//	a user defined network is the "primary" network for this pod. This
-//	signifies the "default" network is only used for probing and
-//	is otherwise locked for all intents and purposes.
-//
-// NOTE: Like in other places, expectation is this function is always called
-// from controller's that have some relation to the given pod, unrelated
-// networks are treated as secondary networks so caller has to be careful
+// GetNetworkRole returns the role of this controller's network for the given pod
 func (a *PodAllocator) GetNetworkRole(pod *corev1.Pod) (string, error) {
-	if !util.IsNetworkSegmentationSupportEnabled() {
-		// if user defined network segmentation is not enabled
-		// then we know pod's primary network is "default" and
-		// pod's secondary network is NOT its primary network
-		if a.netInfo.IsDefault() {
-			return types.NetworkRolePrimary, nil
-		}
-		return types.NetworkRoleSecondary, nil
-	}
-	activeNetwork, err := a.getActiveNetworkForPod(pod)
+	role, err := util.GetNetworkRole(a.netInfo, a.networkManager.GetActiveNetworkForNamespace, pod)
 	if err != nil {
+		if util.IsUnprocessedActiveNetworkError(err) {
+			a.recordPodErrorEvent(pod, err)
+		}
 		return "", err
 	}
-	if activeNetwork.GetNetworkName() == a.netInfo.GetNetworkName() {
-		return types.NetworkRolePrimary, nil
-	}
-	if a.netInfo.IsDefault() {
-		// if default network was not the primary network,
-		// then when UDN is turned on, default network is the
-		// infrastructure-locked network forthis pod
-		return types.NetworkRoleInfrastructure, nil
-	}
-	return types.NetworkRoleSecondary, nil
+
+	return role, nil
 }
 
 // Reconcile allocates or releases IPs for pods updating the pod annotation
@@ -208,9 +172,11 @@ func (a *PodAllocator) reconcile(old, new *corev1.Pod, releaseFromAllocator bool
 	var err error
 
 	if new != nil {
-		activeNetwork, err = a.getActiveNetworkForPod(pod)
-		if err != nil {
-			return fmt.Errorf("failed looking for an active network: %w", err)
+		if a.netInfo.IsPrimaryNetwork() {
+			activeNetwork, err = a.getActiveNetworkForPod(pod)
+			if err != nil {
+				return fmt.Errorf("failed looking for an active network: %w", err)
+			}
 		}
 	} else if a.netInfo.IsPrimaryNetwork() {
 		// During pod deletion, the UDN might already be removed. To handle this, check if the activeNetwork
@@ -363,12 +329,18 @@ func (a *PodAllocator) allocatePodOnNAD(pod *corev1.Pod, nad string, network *ne
 		idAllocator = a.idAllocator.ForName(name)
 	}
 
-	// don't reallocate to new IPs if currently annotated IPs fail to alloccate
+	// don't reallocate to new IPs if currently annotated IPs fail to allocate
 	reallocate := false
 	networkRole, err := a.GetNetworkRole(pod)
 	if err != nil {
 		return err
 	}
+
+	if networkRole == types.NetworkRoleNone {
+		// pod not on this controller, nothing to do
+		return nil
+	}
+
 	updatedPod, podAnnotation, err := a.podAnnotationAllocator.AllocatePodAnnotationWithTunnelID(
 		ipAllocator,
 		idAllocator,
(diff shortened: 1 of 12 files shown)


trozet:
@maiqueb FYI

kyrtapz (commented):
Overall looks good to me, I don't have any blocking comments but this looks concerning:
```
 Summarizing 2 Failures:
  [FAIL] Network Segmentation a user defined primary network created using NetworkAttachmentDefinitions is isolated from the default network [It] with L2 primary UDN
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:427
  [FAIL] e2e egress IP validation Network Segmentation: IPv6 L2 role primary [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts [It] disabling egress nodes impeding GRCP health check
```
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13042343967/job/36392676780?pr=5013

I will re-run to see if it is just a flake.

go-controller/pkg/clustermanager/pod/allocator.go
```diff
@@ -363,12 +329,18 @@ func (a *PodAllocator) allocatePodOnNAD(pod *corev1.Pod, nad string, network *ne
...
 	reallocate := false
 	networkRole, err := a.GetNetworkRole(pod)
 	if err != nil {
 		return err
 	}
+
+	if networkRole == types.NetworkRoleNone {
+		// pod not on this controller, nothing to do
+		return nil
+	}
```
jcaamano, line 342: Could this happen earlier? Like the very first thing to check when handling the event?

go-controller/pkg/clustermanager/pod/allocator.go
```diff
@@ -208,9 +172,11 @@ func (a *PodAllocator) reconcile(old, new *corev1.Pod, releaseFromAllocator bool
 	var err error
 
 	if new != nil {
-		activeNetwork, err = a.getActiveNetworkForPod(pod)
-		if err != nil {
-			return fmt.Errorf("failed looking for an active network: %w", err)
+		if a.netInfo.IsPrimaryNetwork() {
+			activeNetwork, err = a.getActiveNetworkForPod(pod)
```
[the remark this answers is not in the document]
jcaamano: I honestly can't understand what's the deal with `activeNetwork` here and why we care about it. Can't comment much on it.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -257,9 +257,12 @@ func (bsnc *BaseSecondaryNetworkController) ensurePodForSecondaryNetwork(pod *co
 		return err
 	}
 
-	activeNetwork, err := bsnc.networkManager.GetActiveNetworkForNamespace(pod.Namespace)
-	if err != nil {
-		return fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
+	var activeNetwork util.NetInfo
+	if bsnc.IsPrimaryNetwork() {
+		activeNetwork, err = bsnc.networkManager.GetActiveNetworkForNamespace(pod.Namespace)
```
jcaamano, line 262: Nothing against the change but couldn't grasp what problems we have when calling `GetActiveNetworkForNamespace` on secondary networks (other than the performance price on a miss, which would be unrelated to the fact that NADs can live on a different namespace).

go-controller/pkg/clustermanager/pod/allocator.go
jcaamano: Noting that there seems to be some overlap with what is done in `GetPodNADToNetworkMappingWithActiveNetwork` which also checks if the pod is on the network. Can be improved in the future.

go-controller/pkg/ovn/base_network_controller_secondary.go
jcaamano: Okay I wasn't updated on the fact that we are now getting the namespace in `GetActiveNetworkForNamespace`.

jcaamano: I will invite for drinks the day we don't start the networks on initial sync.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -257,9 +257,12 @@ func (bsnc *BaseSecondaryNetworkController) ensurePodForSecondaryNetwork(pod *co
 		return err
 	}
 
-	activeNetwork, err := bsnc.networkManager.GetActiveNetworkForNamespace(pod.Namespace)
-	if err != nil {
-		return fmt.Errorf("failed looking for the active network at namespace '%s': %w", pod.Namespace, err)
+	var activeNetwork util.NetInfo
+	if bsnc.IsPrimaryNetwork() {
```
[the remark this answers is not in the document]
trozet: > Just a note: With this change pods will have secondary networks wired, even if they are created in namespaces labeled for primary UDNs without any CRs defined. This shouldn't be an issue, as the default network checks for the udn and the pod is not going to start anyway.

They will be wired from an OVN standpoint, but CNI will never add the secondary interfaces to the pod, because it will fail on the primary network CNI ADD first, if the primary network was not able to be provisioned.

> On the other hand, doesn't the fact that the default network controller calls GetActiveNetworkForNamespace imply that it might fail to reconcile existing namespaces if any were deleted before being processed?
This is less likely to happen since default will reconcile all namespaces/pods only on startup.

Yes this is true, but if the namespace is being deleted, then the pods are also being deleted for primary network. The namespace in KAPI cannot be removed until all the pods that belong to it are removed. So we are really talking about a scenario where we listed pods, and took so long to process them that a namespace and all pods in a namespace were terminated. These pods should not be wired. So the DNC should fail to start, retry to start again and list all pods and hopefully by that time our cache is consistent with the kube state.

As we have found with downstream slow CI platforms this backing up of processing UDNs is more likely to exhibit the scenario, but with these changes we limit the blast radius to only primary networks, which are tied to specific namespaces.

go-controller/pkg/util/multi_network.go
```diff
@@ -1413,3 +1413,70 @@ func CanServeNamespace(network NetInfo, namespace string) bool {
...
+// (1) "primary" if this network is the primary network of the pod.
+//
+//	The "default" network is the primary network of any pod usually
+//	unless user-defined-network-segmentation feature has been activated.
+//	If network segmentation feature is enabled then any user defined
+//	network can be the primary network of the pod.
+//
+// (2) "secondary" if this network is the secondary network of the pod.
+//
+//	Only user defined networks can be secondary networks for a pod.
```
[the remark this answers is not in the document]
trozet: right this is not completely true it should say NAD as well

go-controller/pkg/clustermanager/pod/allocator.go
trozet: yes i didnt look to see if it could be moved up earlier in the functions as an optimization. if there are other changes needed to this PR i will take a look at that possibility.

go-controller/pkg/ovn/base_network_controller_secondary.go
trozet: LOL. Sorry I didn't write out the full explanation (lack of time). I think you get the point now. Right now we also are slow in bringing up nertwork controllers on slow platforms. We see it can take up to a minute. I think the performance fixes we recently merged for filtering out objects will help, and potentially these fixes will also have an impact.

I think the next step is to change network manger to be multithreaded, so it can handle starting multiple controllers at once. With @dceara changes for multiplexing internal informers there should be room for multiple network controllers to start simultaneously and not lock up internal informers in the process. From my estimation  the control plane side with kube can scale well there. On the flip side, with OVN, most controllers will be doing initial adds and potentially fighting for libovsdb txn locks. There may be diminishing returns there.  However, we have had the same contention in the past with multiple controllers fighting for libovsdb lock to program their stuff in parallel (like services controller vs pod vs network policy, etc), and we saw gains by increasing threadiness there up to 15 (past 15 showed no improvement from my original testing). So in my opinion it would stand to reason that increasing from 1 to at least some value in single digits or low double digits would help our performance with UDN.

However, I'm a bit worried about enabling threadiness for network manager and causing unforeseen deadlock. I think we probably need a bunch of perf/scale testing on it and downstream CI mileage to confirm the benefits and rule out deadlock in that code.

jcaamano: Mutithreading network manager is reasonable, I am just not sure it should be our next step.

I would rather try to look into why we need to start all controllers on initial sync (which wont be multithreaded anyway, or is that what you mean?). It might be that other controllers that depend on that could be enhanced to no longer require it.

Another thought is that I wouldn't like to start overloading network manager with more functionality. For example, I am not sure if it is the best place for this namespace check. What we want to do is to avoid attaching pods to an unsuitable network. The only thing we can't reconcile for the attachment is whatever the CNI does. To stop the CNI, we need to stop the pod annotation. So we would just need to do the namespace check whatever controller is handling the event that annotates. Perhaps `GetNetworkRole` is a better place for this check.

And maybe we should return a specific error that lets us ignore the outcome in different ways depending on the circumstances. Otherwise we are turning a purely informational method `GetActiveNetworkForNamespace` into something that can only be called on certain contexts to prevent it failing on a generic way.

jcaamano (approved):

PR #5015, opened 2025-01-30
ovspinning: Enable the feature at runtime

zeeke:
## 📑 Description
A configuration manager like the `machine-config-operator` is often used to enable the OVS CPU pinning feature. During reconciliation loop, the configurator might remove and recreate a file for a very short time. When it comes to the `/etc/openvswitch/enable_dynamic_cpu_affinity` file, it might disable the CPU pinning permanently and an ovnkube-controller restart is needed.

Make the OVS CPU pinning be enabled without restarting `ovnkube-controller` pod. Watch the parent folder instead of the enabler file because if the file gets deleted, the file watch will not emit any further fsevent.

Fixes #

## ✅ Checks

- [x] My code requires tests

### DIFF
--- a/go-controller/pkg/node/ovspinning/ovspinning_linux_test.go
+++ b/go-controller/pkg/node/ovspinning/ovspinning_linux_test.go
@@ -68,7 +68,7 @@ func TestAlignCPUAffinity(t *testing.T) {
 	}
 
 	// Disable the feature by making the enabler file empty
-	os.WriteFile(featureEnablerFile, []byte(""), 0)
+	err = os.WriteFile(featureEnablerFile, []byte(""), 0)
 	assert.NoError(t, err)
 
 	var tmpCPUset unix.CPUSet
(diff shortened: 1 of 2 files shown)


go-controller/pkg/node/ovspinning/ovspinning_linux_test.go
```diff
@@ -191,7 +218,7 @@ func mockFeatureEnableFile(t *testing.T, data string) func() {
 	previousValue := featureEnablerFile
 	featureEnablerFile = f.Name()
 
-	os.WriteFile(featureEnablerFile, []byte(data), 0)
+	os.WriteFile(featureEnablerFile, []byte(data), 0777)
```
martinkennelly, line 221: Why 777?

zeeke: No specific reason. That flag is used only when the file doesn't exist and it is created by the WriteFile. Reverting to 0

PR #5016, opened 2025-01-30
crd, cudn: Ensure matching topology and topology-configuration

ormergi:
## 📑 Description

Add missing API validations to ensure the topology and the topology configuration are matching.
Similar to the validation rules on UserDefiendNetwork CRD [[1]](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/crd/userdefinednetwork/v1/udn.go#L18-L19).

For example: the following spec is invalid and should be blocked by the cluster API:
```
apiVersion: k8s.ovn.org/v1
...
spec
  network:
    topology: Layer2
    layer3:
    ...
```

[1] https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/crd/userdefinednetwork/v1/udn.go#L18-L19

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
+++ b/dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
@@ -329,6 +329,14 @@ spec:
                 - topology
                 type: object
                 x-kubernetes-validations:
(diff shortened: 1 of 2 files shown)


trozet (approved):
i see UDN has the same checks, lgtm

PR #5017, opened 2025-01-30
Don't retry initial sync for secondary networks

jcaamano:
When secondary networks fail to start, network manager will retry. There is no point in retrying on the factory with synchronous timeouts blocking everything else.

Change also to be more lenient on network controller initial network sync failure given that there are no synchronous retries.  So network controller initial sync is now best effort and on failures it may impact other controllers that depend on it. However the best path would be for these other controller to actually not depend on it.


go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -240,19 +247,44 @@ func (c *networkController) syncAll() error {
...
+					}
+				},
+				func() error { return c.syncNetwork(networkName) },
+			)
+			if errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
+				klog.V(5).Infof("Ignoring network %q since %q does not manage it", networkName, c.name)
+				return nil
+			}
+			return err
+		})
```
ricky-rav, line 281: Neat! I like your solution. This also addresses this bug of mine: https://issues.redhat.com/browse/OCPBUGS-45239

ricky-rav (approved):
/LGTM

go-controller/pkg/networkmanager/network_controller.go
jcaamano: @ricky-rav can you explain how this addresses your issue? This will still crash-loop if the attempts to start a network fail. This is just moving the retry logic from the factory to here.

PR #1072, opened 2020-02-18
fix IPv6 subnet dynamic addressing

danwinship:
`other-config:subnet` for IPv4 addressing takes a CIDR string (eg, `"10.128.0.0/23"`) but `other-config:ipv6_prefix` for IPv6 addressing takes just the IP address part (`"fd01:0:0:1::"`) and assumes `/64`. (This got broken when we were cleaning up the original hacky IPv6-only patches to make them support both IPv4 and IPv6 to submit upstream.)

Also, since IPv6 only supports /64 hostsubnets, update the config docs/validation accordingly.

@dcbw @russellb @girishmg @trozet

PR #5018, opened 2025-01-30
OCPBUGS-49621: Fix UDN and CUDN subnet validation

ricky-rav:
Fix UDN and CUDN subnet validation

In the previous validation rule, IPv6 subnets would mistakenly invalidate the CR.

I couldn't find any tests for this in the PR that introduced the changes (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4933), I'm not sure if that was intended.

### DIFF
--- a/dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
+++ b/dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
@@ -291,7 +291,7 @@ spec:
                               self.hostSubnet > cidr(self.cidr).prefixLength()'
                           - message: HostSubnet must < 32 for ipv4 CIDR
                             rule: '!has(self.hostSubnet) || !isCIDR(self.cidr) ||
-                              (cidr(self.cidr).ip().family() == 4 && self.hostSubnet
+                              (cidr(self.cidr).ip().family() != 4 || self.hostSubnet
                               < 32)'
                         maxItems: 2
                         minItems: 1
--- a/dist/templates/k8s.ovn.org_userdefinednetworks.yaml.j2
+++ b/dist/templates/k8s.ovn.org_userdefinednetworks.yaml.j2
@@ -238,7 +238,7 @@ spec:
                           > cidr(self.cidr).prefixLength()'
                       - message: HostSubnet must < 32 for ipv4 CIDR
                         rule: '!has(self.hostSubnet) || !isCIDR(self.cidr) || (cidr(self.cidr).ip().family()
-                          == 4 && self.hostSubnet < 32)'
+                          != 4 || self.hostSubnet < 32)'
                     maxItems: 2
                     minItems: 1
                     type: array
--- a/go-controller/pkg/crd/userdefinednetwork/v1/shared.go
+++ b/go-controller/pkg/crd/userdefinednetwork/v1/shared.go
@@ -69,7 +69,7 @@ type Layer3Config struct {
 }
 
 // +kubebuilder:validation:XValidation:rule="!has(self.hostSubnet) || !isCIDR(self.cidr) || self.hostSubnet > cidr(self.cidr).prefixLength()", message="HostSubnet must be smaller than CIDR subnet"
-// +kubebuilder:validation:XValidation:rule="!has(self.hostSubnet) || !isCIDR(self.cidr) || (cidr(self.cidr).ip().family() == 4 && self.hostSubnet < 32)", message="HostSubnet must < 32 for ipv4 CIDR"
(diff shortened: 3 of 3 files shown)


dceara:
> Do we support an ipv6 host prefix length other than /64? Looking into the history this was an OVN limitation:
> 

More of an IPv6 EUI-64 limitation IMO but ok.

> #1072
> 
> @dceara I assume this is no longer the case?

@trozet, it's still the case:
https://github.com/ovn-org/ovn/blob/d95d43943dc1b7c3a80f984ff378559f3606c3be/northd/ipam.c#L220-L227

ricky-rav:
I did some testing with the last changes I've just pushed (without the second rule on IPv6, which was redundant):

```
## IPv4 and IPv6 are correct
$ kubectl apply -f - <<EOF
apiVersion: k8s.ovn.org/v1
metadata:
  name: l3-udn
  namespace: cudn-l3-ns
spec:
  namespaceSelector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - cudn-l3-ns
  network:
    topology: Layer3
    layer3:
      role: Primary
      subnets:
      - cidr: 10.10.0.0/16
        hostSubnet: 31
      - cidr: 2010:100:200::0/48
        hostSubnet: 127
EOF
clusteruserdefinednetwork.k8s.ovn.org/l3-udn created


$ kubectl delete clusteruserdefinednetwork l3-udn
clusteruserdefinednetwork.k8s.ovn.org "l3-udn" deleted


## IPv4 is not correct
$ kubectl apply -f - <<EOF
apiVersion: k8s.ovn.org/v1
metadata:
  name: l3-udn
  namespace: cudn-l3-ns
spec:
  namespaceSelector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - cudn-l3-ns
  network:
    topology: Layer3
    layer3:
      role: Primary
      subnets:
      - cidr: 10.10.0.0/16
        hostSubnet: 32
      - cidr: 2010:100:200::0/48
        hostSubnet: 127
EOF
The ClusterUserDefinedNetwork "l3-udn" is invalid: spec.network.layer3.subnets[0]: Invalid value: "object": HostSubnet must < 32 for ipv4 CIDR

## IPv6 is not correct
$ kubectl apply -f - <<EOF
apiVersion: k8s.ovn.org/v1
metadata:
  name: l3-udn
  namespace: cudn-l3-ns
spec:
  namespaceSelector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - cudn-l3-ns
  network:
    topology: Layer3
    layer3:
      role: Primary
      subnets:
      - cidr: 10.10.0.0/16
        hostSubnet: 31
      - cidr: 2010:100:200::0/48
        hostSubnet: 128
EOF
The ClusterUserDefinedNetwork "l3-udn" is invalid: spec.network.layer3.subnets[1].hostSubnet: Invalid value: 128: spec.network.layer3.subnets[1].hostSubnet in body should be less than or equal to 127


## both IPv4 and IPv6 are not correct
$ kubectl apply -f - <<EOF
apiVersion: k8s.ovn.org/v1
metadata:
  name: l3-udn
  namespace: cudn-l3-ns
spec:
  namespaceSelector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values:
      - cudn-l3-ns
  network:
    topology: Layer3
    layer3:
      role: Primary
      subnets:
      - cidr: 10.10.0.0/16
        hostSubnet: 32
      - cidr: 2010:100:200::0/48
        hostSubnet: 128
EOF
The ClusterUserDefinedNetwork "l3-udn" is invalid: 
* spec.network.layer3.subnets[1].hostSubnet: Invalid value: 128: spec.network.layer3.subnets[1].hostSubnet in body should be less than or equal to 127
* spec.network.layer3.subnets[0]: Invalid value: "object": HostSubnet must < 32 for ipv4 CIDR

```

trozet (approved):

trozet:
note we should have an e2e test that tries to createa  dualstack UDN so we would have caught this. We can add it in a follow up @ricky-rav

tssurya:
we have a dualstack lane right? I would except us to already have coverage for this....too bad if that's not the case

ricky-rav:
> note we should have an e2e test that tries to createa dualstack UDN so we would have caught this. We can add it in a follow up @ricky-rav

I'm on it!
I've created an upstream issue for that: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5023

ricky-rav:
> > note we should have an e2e test that tries to createa dualstack UDN so we would have caught this. We can add it in a follow up @ricky-rav
> 
> I'm on it! I've created an upstream issue for that: #5023

Here it is: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5030

PR #5019, opened 2025-01-30
Ensure controllers stop gracefully

trozet:
With secondary network controllers, if they start and fail, they return error, but their resources are not cleaned up. For example, they might leave handlers attached to the retry factory after they return out. Change this behavior to stop the controller and cleanup after a failure.

Additionally, stop methods throughout our code were not safe to execute more than one time. They try to close a channel which will panic if it is already closed. Encapsulate all stop methods with sync.Once so that even if Stop is called multiple times, it will only ever shutdown a single time for each controller.


go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -342,6 +342,10 @@ func (c *networkController) ensureNetwork(network util.MutableNetInfo) error {
 
 	err = nc.Start(context.Background())
 	if err != nil {
+		nc.Stop()
+		if err := nc.Cleanup(); err != nil {
+			klog.Errorf("Failed to cleanup network %s: %v", networkName, err)
+		}
```
jcaamano, line 348: This is what I would do:

- I would not call `Stop` here. It might work nicely now but semantically an implementation might need to tell apart if its being told to stop after successfully starting vs after unsuccessfully starting. The implementation is already aware about the latter case so `Stop` can be reserved just for the former case. `Cleanup` is fine since there is no semantic contract that something needs to be started in order to be cleaned up.
- I would add network state with `stoppedAndDeleting=true` if cleanup failed to retry cleanup and block it from starting until that has happened.

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -455,26 +455,28 @@ func (oc *SecondaryLayer3NetworkController) Start(ctx context.Context) error {
...
-		oc.watchFactory.RemovePodHandler(oc.podHandler)
-	}
-	if oc.nodeHandler != nil {
-		oc.watchFactory.RemoveNodeHandler(oc.nodeHandler)
-	}
-	if oc.namespaceHandler != nil {
-		oc.watchFactory.RemoveNamespaceHandler(oc.namespaceHandler)
-	}
+	oc.stopOnce.Do(func() {
+		klog.Infof("Stop secondary %s network controller of network %s", oc.TopologyType(), oc.GetNetworkName())
```
jcaamano, line 459: `Stopping layer 3 network controller for secondary network %s`?

go-controller/pkg/ovn/secondary_localnet_network_controller.go
```diff
@@ -308,7 +308,7 @@ func (oc *SecondaryLocalnetNetworkController) Init() error {
 }
 
 func (oc *SecondaryLocalnetNetworkController) Stop() {
-	klog.Infof("Stoping controller for secondary network %s", oc.GetNetworkName())
+	klog.Infof("Stopping localnet controller for secondary network %s", oc.GetNetworkName())
```
jcaamano, line 311: `Stopping localnet network controller for secondary network %s`?

PR #5020, opened 2025-01-31
Change dynamic_neigh_routers to false for the Interconnect topology.

abhat:
In OVN Interconnect topology having dynamic_neigh_routers option on the gateway router will cause mac bindings to expire toward the join switch port. This change disables that option for IC.

### DIFF
--- a/go-controller/pkg/ovn/egressip_test.go
+++ b/go-controller/pkg/ovn/egressip_test.go
@@ -277,6 +277,15 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 				UUID: node1.Name + "-UUID",
 				Name: node1.Name,
 			}
+			dynamicNeighRouters := "true"
+			if config.OVNKubernetesFeature.EnableInterconnect {
+				dynamicNeighRouters = "false"
+			}
+
+			logicalRouterOptions := map[string]string{
+				"dynamic_neigh_routers": dynamicNeighRouters,
+			}
+
 			fakeOvn.startWithDBSetup(
 				libovsdbtest.TestSetup{
 					NBData: []libovsdbtest.TestData{
@@ -290,9 +299,10 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 							UUID: types.OVNClusterRouter + "-UUID",
 						},
 						&nbdb.LogicalRouter{
-							Name:  types.GWRouterPrefix + node1.Name,
-							UUID:  types.GWRouterPrefix + node1.Name + "-UUID",
-							Ports: []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node1.Name + "-UUID"},
+							Name:    types.GWRouterPrefix + node1.Name,
+							UUID:    types.GWRouterPrefix + node1.Name + "-UUID",
+							Ports:   []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node1.Name + "-UUID"},
+							Options: logicalRouterOptions,
 						},
 						&nbdb.LogicalSwitchPort{
 							UUID: types.EXTSwitchToGWRouterPrefix + types.GWRouterPrefix + node1Name + "-UUID",
@@ -392,10 +402,11 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 					},
 				},
 				&nbdb.LogicalRouter{
-					Name:  types.GWRouterPrefix + node1.Name,
-					UUID:  types.GWRouterPrefix + node1.Name + "-UUID",
-					Ports: []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node1.Name + "-UUID"},
-					Nat:   []string{"egressip-nat-UUID"},
+					Name:    types.GWRouterPrefix + node1.Name,
+					UUID:    types.GWRouterPrefix + node1.Name + "-UUID",
+					Ports:   []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node1.Name + "-UUID"},
+					Nat:     []string{"egressip-nat-UUID"},
+					Options: logicalRouterOptions,
 				},
 				&nbdb.LogicalRouter{
 					Name:     types.OVNClusterRouter,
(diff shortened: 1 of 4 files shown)


go-controller/pkg/ovn/egressip_test.go
```diff
@@ -1956,14 +1966,16 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+							Options: logicalRouterOptions,
 						},
 						&nbdb.LogicalRouter{
-							Name:  types.GWRouterPrefix + node2.Name,
-							UUID:  types.GWRouterPrefix + node2.Name + "-UUID",
-							Ports: []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node2.Name + "-UUID"},
+							Name:    types.GWRouterPrefix + node2.Name,
+							UUID:    types.GWRouterPrefix + node2.Name + "-UUID",
+							Ports:   []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node2.Name + "-UUID"},
+							Options: logicalRouterOptions,
```
ricky-rav, line 1978: Do we expect `dynamic_neigh_routers` to have an impact on egress ip? Is that why you added the option here in egressip_test.go and not in the other unit tests we have?

ricky-rav:
> 2. https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13116656239/job/36656733564?pr=5020 which is [e2e (control-plane, HA, shared, ipv4, snatGW, 1br, ic-disabled, 1,2,3)](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13116656239/job/36656733564?pr=5020#logs) and its failing because of:
> 
> ```
> 2025-02-04T15:42:25.6292666Z �[38;5;9m�[1mSummarizing 1 Failure:�[0m
> 2025-02-04T15:42:25.6293800Z   �[38;5;9m[FAIL]�[0m �[0mServices �[38;5;9m�[1m[It] Allow connection to an external IP using a source port that is equal to a node port�[0m
> 2025-02-04T15:42:25.6295526Z   �[38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/service.go:93�[0m
> ```
> 
> I have not seen this flake before @ricky-rav can you please look into this second flake?

Sure! I've created an upstream issue for that and added my analysis: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5034
Basically the test that fails is using the same port as the previous test and ovnk hasn't released the port yet. 

Let's continue the discussion there.

go-controller/pkg/ovn/egressip_test.go
abhat: Not particularly. I think there are other egress IP tests that need to be patched to use the Options field too. But a grep across the repo didn't show me other files affected. I would like some feedback though on if hybrid overlay should also have this change?

go-controller/pkg/ovn/gateway.go
```diff
@@ -276,9 +276,14 @@ func (gw *GatewayManager) GatewayInit(
 		physicalIPs[i] = ip.IP.String()
 	}
 
+	dynamicNeighRouters := "true"
+	if config.OVNKubernetesFeature.EnableInterconnect {
+		dynamicNeighRouters = "false"
```
tssurya, line 281: what are the implications of setting this to false.. I couldn't find the history commit where we set this to true.. 
```
       options : dynamic_neigh_routers: optional string, either true or false
              If  set  to  true, the router will resolve neighbor routers’ MAC
              addresses only by dynamic ARP/ND, instead of prepopulating  sta‐
              tic  mappings  for all neighbor routers in the ARP/ND Resolution
              stage. This reduces number of flows, but  requires  ARP/ND  mes‐
              sages to resolve the IP-MAC bindings when needed. It is false by
              default. It is recommended to set to true when a large number of
              logical  routers  are  connected  to the same logical switch but
              most of them never need to send traffic between each  other.  By
              default,  ovn-northd  does  not  create mappings to NAT and load
              balancer addresess. However, for NAT and load balancer addresses
              that have the add_route option  added,  ovn-northd  will  create
              logical flows that map NAT and load balancer IP addresses to the
              appropriate  MAC  address. Setting dynamic_neigh_routers to true
              will prevent the automatic creation of these logical flows.
```
so according to this true helps with scale aspects of number of flows?
functionally speaking is there any difference between false and true for OVNK I wonder - i.e internal OVNK topology neighbors we anyway configure them statically but for customer envs when nodes are connected to other routers does this need any specific config change for the customer to discover the neighbor routers compared to how it happens today?

@dceara @almusil could you please explain what OVN does exactly with regards to:
```
 instead of prepopulating  static  mappings  for all neighbor routers in the ARP/ND Resolution stage. 
```
when we set this option to false? when does the ARP/ND stage happen for this

tssurya (commented):
@abhat :+1: 
`dynamic_neigh_routers option on the gateway router will cause mac bindings to expire toward the join switch port.`
how often is the expiry happening is it using `mac_binding_age_threshold` 5mins timer?

abhat:
> @abhat `dynamic_neigh_routers option on the gateway router will cause mac bindings to expire toward the join switch port.` how often is the expiry happening is it using `mac_binding_age_threshold` 5mins timer?

Yes

go-controller/pkg/ovn/egressip_test.go
abhat: Yes, or for winc containers use-cases.

go-controller/pkg/ovn/gateway.go
igsilya: When set to `false`, or not set at all (default), OVN will create logical flows for each specific router IP with the known MAC address of the router port.  It may be a lot of flows if there are many routers connected to each other, see https://mail.openvswitch.org/pipermail/ovs-discuss/2020-May/049995.html .
But in OVN-IC case with 1 node per AZ, this results with just 2 logical flows.

If set to `true`, OVN will ARP for those IPs instead and create dynamic MAC bindings from ARP replies.

go-controller/pkg/ovn/egressip_test.go
ricky-rav: Uhm. I wonder if this option in unit tests makes any difference :)  If it does, then all gateway routers in all unit tests in the repo should be updated, right? Either that or I'm missing a few pieces here. :)

trozet: the other unit tests probably rely on generateGatewayInitExpectedNB which is in gateway_test.go. egress IP must be a one off.

trozet: Not sure I understand the context around hybrid overlay. We don't create a router for that.

trozet (approved):

Issue #5024, opened 2025-02-03
Kubevirt VM Live Migration link is broken


dror1212:
### What would you like to see added/updated/deleted in our docs?

The Kubevirt VM Live Migration link in https://ovn-kubernetes.io/ is broken (https://ovn-kubernetes.io/design/live-migration.md) since the path it tries to find in the [mkdocs](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/mkdocs.yml) file does not exist
```
    - Kubevirt VM Live Migration: design/live-migration.md
```

So I guess this section can be removed entirely or create an empty file like [design/gateway-modes.md](https://github.com/ovn-kubernetes/ovn-kubernetes/tree/master/docs/design/gateway-modes.md).

### Why is this needed?

Currently this path in the docs is broken.

PR #5025, opened 2025-02-03
docs: add missing live-migration.md file

dror1212:
Add a missing file design/live-migration.md so the docs is not broken

Fixes: #5024


dror1212:
@pperiyasamy

PR #5026, opened 2025-02-03
stop adding events to NAD if the network type is not ovn-k

SchSeba:
## 📑 Description

If NADs like bridge,macvlan or others exist we should not record an error event for it

Also in case the NAD is not ovn-k for example multus we support chain plugins.

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/config/cni.go
+++ b/go-controller/pkg/config/cni.go
@@ -120,10 +120,6 @@ func parseNetConfSingle(bytes []byte) (*ovncnitypes.NetConf, error) {
 }
 
 func parseNetConfList(confList *libcni.NetworkConfigList) (*ovncnitypes.NetConf, error) {
-	if len(confList.Plugins) > 1 {
-		return nil, ErrorChainingNotSupported
-	}
-
 	netconf := &ovncnitypes.NetConf{MTU: Default.MTU}
 	if err := json.Unmarshal(confList.Plugins[0].Bytes, netconf); err != nil {
 		return nil, err
@@ -132,6 +128,10 @@ func parseNetConfList(confList *libcni.NetworkConfigList) (*ovncnitypes.NetConf,
 	// skip non-OVN NAD
 	if netconf.Type != "ovn-k8s-cni-overlay" {
 		return nil, ErrorAttachDefNotOvnManaged
+	}
+
+	if len(confList.Plugins) > 1 {
+		return nil, ErrorChainingNotSupported
 	}
 
 	netconf.Name = confList.Name
--- a/go-controller/pkg/networkmanager/nad_controller.go
+++ b/go-controller/pkg/networkmanager/nad_controller.go
@@ -274,6 +274,11 @@ func (c *nadController) syncNAD(key string, nad *nettypes.NetworkAttachmentDefin
 	if nad != nil {
 		nadNetwork, err = util.ParseNADInfo(nad)
 		if err != nil {
+			// in case the type for the NAD is not ovn-k we should not record the error event
+			if err.Error() == util.ErrorAttachDefNotOvnManaged.Error() {
+				return nil
+			}
+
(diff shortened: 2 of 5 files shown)


go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -1,6 +1,7 @@
 package networkmanager
 
 import (
+	errors2 "errors"
```
maiqueb, line 4: maybe `coreerrors` or something else.

This is a nit you can ignore.

maiqueb (changes requested):
Could we have a unit test added to cover these changes ?

SchSeba:
Hi @maiqueb,
thanks for the comment when you have time please take a look

maiqueb:
@SchSeba your PR doesn't seem to be passing the unit tests. 

Is it related to your PR ?

SchSeba:
Hi @maiqueb @trozet @tssurya can you please take a look on this PR?

Thanks!

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -273,6 +273,11 @@ func (c *nadController) syncNAD(key string, nad *nettypes.NetworkAttachmentDefin
 	if nad != nil {
 		nadNetwork, err = util.ParseNADInfo(nad)
 		if err != nil {
+			// in case the type for the NAD is not ovn-k we should not record the error event
+			if err.Error() == util.ErrorAttachDefNotOvnManaged.Error() {
```
tssurya, line 277: I knew this looked too familiar somewhere..

@martinkennelly ptal https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4846/files#diff-0b74e004432877a3ca7b23367059bb077ffa13c01d26f6e3a0f37d09acc67403R164
I thought we closed this PR because you saw the new syncNAD changes happen and didn't see this happening anymore?

@SchSeba : https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4846#issuecomment-2551576818
so is my understanding correct that error event here is causing noise for you? @martinkennelly maybe we should have merged your PR haha :)

if we still need this can we do `if err != nil && !errors.Is(err, config.ErrorAttachDefNotOvnManaged) {` ? its more cleaner...

tssurya (changes requested):

go-controller/pkg/config/cni.go
```diff
@@ -134,6 +130,10 @@ func parseNetConfList(confList *libcni.NetworkConfigList) (*ovncnitypes.NetConf,
 		return nil, ErrorAttachDefNotOvnManaged
 	}
 
+	if len(confList.Plugins) > 1 {
```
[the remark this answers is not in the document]
oshoval: we just added this condition / error `ErrorChainingNotSupported` to declare we don't support 
plugin chaining
all seems fine for me in this aspect, the error is used if `if len(confList.Plugins) > 1` condition is met

go-controller/pkg/networkmanager/nad_controller.go
SchSeba: I think it will not be the same here @tssurya.

if I do `if err != nil && !errors.Is(err, config.ErrorAttachDefNotOvnManaged)` and the error is `config.ErrorAttachDefNotOvnManaged` it will add it to the list `nadNetworkName = nadNetwork.GetNetworkName()`

tssurya: /assign @martinkennelly

martinkennelly (approved):

tssurya:
@SchSeba : can you please rebase on latest master and I can get this in

SchSeba:
Hi @tssurya done sorry I miss your message about the rebase

PR #5027, opened 2025-02-03
factory/handler: avoid deadlock on shutdown.

npinaeva:
Calling unlock in defer will only be executed before return, thus holding all informer locks at the same time.

e.g. if you try to execute Shutdown from 2 threads in parallel, and informers are not iterated in the same order, you get a deadlock.

```
thread1

for inf := [inf1, inf2]
  inf.Lock() // locks inf1
```

```
thread2

for inf := [inf2, inf1]
  inf.Lock() //locks inf2
```


npinaeva:
cc @dceara

npinaeva:
hit https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4092

dceara (approved):
Oops, looks good to me, thanks for fixing it @npinaeva!

trozet (approved):

PR #5028, opened 2025-02-03
change slack url to point to the cloud-native workspace

girishmg:
Currently all our documentation is pointing to an archived slack workspace and channel. We need to move it to the new slack_url: https://cloud-native.slack.com/archives/C08452HR8V6


flavio-fernandes (approved):
/lgtm

PR #5029, opened 2025-02-04
observability-lib/model: reduce dependencies by duplicating constants

jotak:
## 📑 Description

Reducing dependencies for projects pulling observability-lib/model (e.g. netobserv)
Currently there's a lot of dependencies just because of using constants (strings) from `libovsdbops` and `nbdb`.

This change duplicates those constants, and protects from potential future renames with a unit test.

## ✅ Checks

- [x] My code requires tests


go-controller/observability-lib/model/network_event.go
```diff
@@ -2,9 +2,30 @@ package model
...
 	"fmt"
+)
+
+const (
+	// Constants are duplicated to minimize dependencies
+	// When adding constants here, add them in network_event_test.go too
 
-	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
+	// libovsdb constants: cf github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops
```
npinaeva, line 11: what does "cf" mean?

npinaeva (commented):
Makes sense to me, can you tell how much dependencies you managed to save like this and if you expect more changes to be needed in the future?

go-controller/observability-lib/model/network_event.go
jotak: huh
I've been using it very frequently, probably borrowed from French, and you're the first one asking ... and it seems like I should stop using it lol
I use it in the meaning of "See also" ... which, according to Wiktionary, is [non-standard & proscribed](https://en.wiktionary.org/wiki/cf.#English)

jotak:
> Makes sense to me, can you tell how much dependencies you managed to save like this

It's all the dependencies that you can see here: https://github.com/netobserv/network-observability-cli/pull/135/files
Which is more than 600 files.


> and if you expect more changes to be needed in the future?

If you're asking about changes like that, I don't see anything needed now. But I suspect there might be at some point new OVN features that may come with new observability items to plug here.

go-controller/observability-lib/model/network_event.go
```diff
@@ -23,39 +44,41 @@ type ACLEvent struct {
...
+	case netpolNamespaceOwnerType:
 		msg = fmt.Sprintf("network policies isolation in namespace %s, direction %s", e.Namespace, e.Direction)
-	case libovsdbops.EgressFirewallOwnerType:
+	case egressFirewallOwnerType:
 		msg = fmt.Sprintf("egress firewall in namespace %s", e.Namespace)
-	case libovsdbops.UDNIsolationOwnerType:
+	case udnIsolationOwnerType:
 		msg = fmt.Sprintf("UDN isolation of type %s", e.Name)
+	default:
+		msg = e.Actor
```
npinaeva, line 81: now thinking about this again: do you think it makes sense to deplucate actor into the message vs just leaving msg empty? I think empty message gives a more clear signal that we don't know how to format messages from this actor and actor will be in the event struct anyway?

npinaeva (commented):
I wasn't sure if just this change will be enough to get rid of all dependencies, but if that works, then great! One mote nit: can you squash the commits into one?

go-controller/observability-lib/model/network_event.go
jotak: done .. and squashed commits

npinaeva (approved):
perfect, thank you!

npinaeva:
All test failures are known and were updated with the link to the failing job

trozet (approved):

PR #3322, opened 2022-12-14
Move from math/rand to crypto/rand package for random numbers generation

Pardhakeswar:
Signed-off-by: Pardhakeswar Pacha <ppacha@nvidia.com>

**- What this PR does and why is it needed**

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

PR #5031, opened 2025-02-04
Improve multiplexed internal informer performance

trozet:
Using cryptorand ends up making a syscall to udev/random and takes too much CPU since this is called on every enqueue event. We don't need cryptographic random numbers as we are not generating tokens or certificates.

In the future, we should look at implementing a heap and storing queues by size on the heap. That would make it faster than the current method, but for now this is the least risky option.

Reverts PR:  https://github.com/ovn-kubernetes/ovn-kubernetes/pull/3322

@girishmg @Pardhakeswar fyi

image

Additionally includes a performance fix to limit how many events we need to process on the 201 internal informers we create.

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -190,6 +190,10 @@ const (
 	defaultHandlerPriority int = 0
 	// lowest priority among various handlers (See GetHandlerPriority for more information)
 	minHandlerPriority int = 4
+
+	// used to determine if an internal informer has handlers attached to it or not
+	hasNoHandler uint32 = 0
+	hasHandler   uint32 = 1
 )
 
 var (
@@ -1295,6 +1299,12 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 	intInf.Lock()
 	defer intInf.Unlock()
 
+	// we are going to add a handler, we need to update the atomic signal that handlers exist now
+	// so that we do not miss events after we list current items.
+	// We need to do this after we get internal informer lock, to preserve that we can be the only one updating
+	// the atomic and preserve known state of the atomic while the handler is going to be added in the future
+	hadZeroHandlers := atomic.CompareAndSwapUint32(&intInf.hasHandlers, hasNoHandler, hasHandler)
+
 	items := make([]interface{}, 0)
 	for _, obj := range inf.inf.GetStore().List() {
 		if filterFunc(obj) {
@@ -1313,6 +1323,10 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 			return true, nil
 		})
 		if err != nil {
+			// handler is not going to be added, restore previous value if needed
+			if hadZeroHandlers {
+				atomic.StoreUint32(&intInf.hasHandlers, hasNoHandler)
+			}
 			return nil, err
 		}
 	}
--- a/go-controller/pkg/factory/handler.go
+++ b/go-controller/pkg/factory/handler.go
@@ -107,20 +107,22 @@ type queueMapEntry struct {
 
 type internalInformer struct {
 	sync.RWMutex
-	oType    reflect.Type
+	oType reflect.Type
+	// keyed by priority - used to track the handler's priority of being invoked.
+	// example: a handler with priority 0 will process the received event first
+	// before a handler with priority 1, 0 being the highest priority.
+	// NOTE: we can have multiple handlers with the same priority hence the value
+	// is a map of handlers keyed by its unique id.
 	handlers map[int]map[uint64]*Handler
 	// queueMap handles distributing events across a queued handler's queues
 	queueMap *queueMap
+	// hasHandlers is an atomic used to determine if this internal informer actually has handlers attached to it or not
+	hasHandlers uint32
 }
 
 type informer struct {
-	oType reflect.Type
-	inf   cache.SharedIndexInformer
-	// keyed by priority - used to track the handler's priority of being invoked.
-	// example: a handler with priority 0 will process the received event first
-	// before a handler with priority 1, 0 being the higest priority.
-	// NOTE: we can have multiple handlers with the same priority hence the value
-	// is a map of handlers keyed by its unique id.
+	oType  reflect.Type
+	inf    cache.SharedIndexInformer
 	lister listerInterface
 	// initialAddFunc will be called to deliver the initial list of objects
 	// when a handler is added
@@ -133,7 +135,7 @@ type informer struct {
 func (inf *internalInformer) forEachQueuedHandler(f func(h *Handler)) {
 	inf.RLock()
 	defer inf.RUnlock()
-	for priority := 0; priority <= minHandlerPriority; priority++ { // loop over priority higest to lowest
+	for priority := 0; priority <= minHandlerPriority; priority++ { // loop over priority highest to lowest
 		for _, handler := range inf.handlers[priority] {
 			f(handler)
 		}
@@ -192,7 +194,9 @@ func (i *informer) removeHandler(handler *Handler) {
 
 		intInf.Lock()
 		defer intInf.Unlock()
-		removed := 0
+		removed := false
+		// track overall how many handlers this internal informer has
+		numHandlers := 0
 		for priority := range intInf.handlers { // loop over priority
 			if _, ok := intInf.handlers[priority]; !ok {
 				continue // protection against nil map as value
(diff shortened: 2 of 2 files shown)


trozet:
@dceara fyi

trozet:
egress ip flake, rerunning

trozet:
rebased to pick up egress ip flake fix

go-controller/pkg/factory/factory.go
```diff
@@ -1313,6 +1323,10 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 			return true, nil
 		})
 		if err != nil {
+			// handler is not going to be added, restore previous value if needed
+			if hadZeroHandlers {
+				atomic.StoreUint32(&intInf.hasHandlers, hasNoHandler)
```
trozet, line 1328: i just realized on handler removal i need to set it back to 0 if there are no handlers left on the internal informer. I'll do a follow up after I see how CI goes here

trozet:
tests look good upstream and downstream: https://github.com/openshift/ovn-kubernetes/pull/2442

going to push another revision to handle removal of secondary controllers/handlers

go-controller/pkg/factory/factory.go
trozet: done

go-controller/pkg/factory/factory.go
```diff
@@ -1295,6 +1299,12 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 	intInf.Lock()
 	defer intInf.Unlock()
 
+	// we are going to add a handler, we need to update the atomic signal that handlers exist now
+	// so that we do not miss events after we list current items.
```
dceara, line 1303: Nit: maybe change this to "we need to update the atomic signal that handlers exist now, before adding the actual handler, to ensure that we do not miss events while listing current items".

I'm not sure whether that sounds better or not, I'll leave it up to you.

go-controller/pkg/factory/factory.go
```diff
@@ -1295,6 +1299,12 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
...
+	// We need to do this after we get internal informer lock, to preserve that we can be the only one updating
+	// the atomic and preserve known state of the atomic while the handler is going to be added in the future
```
dceara, line 1305: I'm not sure what you meant with "while the handler is going to be added in the future".  Did you mean that you want to store the old value so you can rollback in case of error?

dceara (approved):
I nit-picked a bit but the change looks correct to me.  There should be no window in which we can lose incoming events as far as I can tell.

go-controller/pkg/factory/factory.go
trozet: Yeah your text is clearer. I can fix the comment later in another pr

go-controller/pkg/factory/handler.go
```diff
@@ -192,19 +194,28 @@ func (i *informer) removeHandler(handler *Handler) {
 
 		intInf.Lock()
 		defer intInf.Unlock()
-		removed := 0
+		removed := false
+		// track overall how many handlers this internal informer has
+		numHandlers := 0
```
tssurya, line 199: unrelated but what's the purpose of 
```

type WatchFactory struct {
	handlerCounter *handlerCounter
```
handler counter in the wf which we seem to also be using to count the number of handlers? I guess that is total across all informers?

tssurya (approved):
code looks correct to me

The only doubt I have is the following:
Let's take ANP as an example which is level driven
So I do the following for handler creation:

1. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/pkg/ovn/controller/admin_network_policy/admin_network_policy_controller.go#L141
2. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L581
3. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L624

so that's where we do the list and add from what I can tell...

so we are never going to use our custom:

`func (wf *WatchFactory) addHandler(` here and hence `intInf.hasHandlers` will always be zero right?

but given we create the ANP informers from watch factory using https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/pkg/factory/factory.go#L416 which is also creating the initialAddFunc + https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/pkg/factory/handler.go#L549 where we will simply keep returning since hasHandlers will always be zero right?

functionally I guess we are not using those handlers so we are good, but if there is a way to identify 0 v/s not used totally that would be nice unless I am missing something.

go-controller/pkg/factory/handler.go
```diff
@@ -260,7 +260,7 @@ func (qm *queueMap) getNewQueueNum() uint32 {
 	if numEventQueues == 1 {
 		return 0
 	}
-	startIdx = uint32(cryptorand.Intn(int64(numEventQueues - 1)))
+	startIdx = uint32(rand.Intn(int(numEventQueues - 1)))
```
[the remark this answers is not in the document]
trozet: yeah sorry the picture is cut off, io.Read will end up calling into unix fd

go-controller/pkg/factory/factory.go
```diff
@@ -1295,6 +1299,12 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 	intInf.Lock()
 	defer intInf.Unlock()
 
+	// we are going to add a handler, we need to update the atomic signal that handlers exist now
```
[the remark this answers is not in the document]
trozet: after processExisting (our sync function) it will then do the initial add of the listed items. If we didnt pre-emptively set the signal that we should receive events before we list...this could happen:

1. baseline scenario, no handlers on internal informer, no events are being sent to the queues
2. event A happens, ignored
3. we start to add a new handler, list existing items -> A
4. handler starts being added, doing Sync and processing initial add from 3 (just A)
5. while 4 is happening, event B happens. It is ignored by the kapi informer from being sent to the queue of the internal informer because hasHandlers is 0
6. handler add completes, and now new event C occurs and is sent to the internal informer

we missed event B, so we have to pre-emptively say, hey we know we are going to add a handler so we should start receiving events now to not miss any

go-controller/pkg/factory/handler.go
trozet: i guess it could probably removed, it doesnt look like its being used. It makes more sense to have the counter on the internal informer, since that is the thing we need to lock to protect write access

trozet:
> code looks correct to me
> 
> The only doubt I have is the following: Let's take ANP as an example which is level driven So I do the following for handler creation:
> 
> 1. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/pkg/ovn/controller/admin_network_policy/admin_network_policy_controller.go#L141
> 2. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L581
> 3. https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/vendor/k8s.io/client-go/tools/cache/shared_informer.go#L624
> 
> so that's where we do the list and add from what I can tell...
> 
> so we are never going to use our custom:
> 
> `func (wf *WatchFactory) addHandler(` here and hence `intInf.hasHandlers` will always be zero right?
> 
> but given we create the ANP informers from watch factory using
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/pkg/factory/factory.go#L416
> 
> which is also creating the initialAddFunc +
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/bd5fff3a438d598419f0bd04890a336b12c93a08/go-controller/pkg/factory/handler.go#L549
> 
> where we will simply keep returning since hasHandlers will always be zero right?
> functionally I guess we are not using those handlers so we are good, but if there is a way to identify 0 v/s not used totally that would be nice unless I am missing something.

I guess the question is why does ANP create a queued informer on the ovnk custom handler code when it is level driven and adds its own event handlers directly :) ?

tssurya:
lol firstly what are you doing ONLINE? :P

> I guess the question is why does ANP create a queued informer on the ovnk custom handler code when it is level driven and adds its own event handlers directly :) ?

exactly something unrelated to this PR deff but need to also see this for other level driven controllers on how its done. but that's a FUP.

go-controller/pkg/factory/factory.go
tssurya: yeah +1 to updating this

PR #5032, opened 2025-02-05
Ensure UDN peer namespace address sets added on the gress policy ACL

pperiyasamy:
The network policy controller previously added address sets for both default and UDN networks though target namespace is managed by primary UDN. With [PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4861) as an optimization, Address Set for default network is no longer available for UDN networked peer namespace, so needed to add a check while creating and deleting ACL for gress policy to skip empty hash when trying to query for AS for such peer namespace on default network (or it is being avoided with network check before updating ACL). Anyway this PR has checks for both.

The network policy controller used to populate corresponding peer UDN namespace address sets in its ACL which were always there for existing namespace because secondary network controllers created those address sets for the network beforehand though it can be associated with a different network later on. But this is no longer the case for UDN with [PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4861), it has an address set created only for the first namespace associated with the network, for the remaining namespaces belonging to the same network address set creation happens only at the time the first pod is created on it. Hence this PR hooks up peer namespace into gress ACL while setting up a network for a pod.

## Summary by CodeRabbit

* **New Features**
  * Expanded network segmentation policy tests to include additional namespaces and more comprehensive connectivity scenarios.

* **Bug Fixes**
  * Improved validation in network policy address set management to prevent operations with incomplete or invalid namespace information.

* **Refactor**
  * Enhanced error handling and modularized reconciliation logic for network controllers, improving reliability and maintainability.
  * Refactored peer namespace handling in network policies to use retry frameworks for more robust reconciliation.
  * Updated namespace handling to support immediate requeueing for network policy processing in primary networks.

* **Tests**
  * Added new test cases covering more namespaces and network policy behaviors to ensure correct isolation and traffic allowance.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller.go
+++ b/go-controller/pkg/ovn/base_network_controller.go
@@ -260,6 +260,10 @@ func (oc *BaseNetworkController) doReconcile(reconcileRoutes, reconcilePendingPo
 		}
 	}
 
+	// reconciles namespaces that were added to the network, this will trigger namespace add event and
+	// network controller creates the address set for the namespace.
+	// To update gress policy ACLs with peer namespace address set, invoke requeuePeerNamespace method after
+	// address set is created for the namespace.
 	namespaceAdded := false
 	for _, ns := range reconcileNamespaces {
 		namespace, err := oc.watchFactory.GetNamespace(ns)
@@ -276,11 +280,6 @@ func (oc *BaseNetworkController) doReconcile(reconcileRoutes, reconcilePendingPo
 	}
 	if namespaceAdded {
 		oc.retryNamespaces.RequestRetryObjs()
-	}
-
-	err := oc.requeuePeerNamespaces(reconcileNamespaces)
-	if err != nil {
-		klog.Infof("Failed to retry network policy peer namespaces for network %s: %v", oc.GetNetworkName(), err)
 	}
 }
 
--- a/go-controller/pkg/ovn/base_network_controller.go
+++ b/go-controller/pkg/ovn/base_network_controller.go
@@ -191,8 +191,7 @@ type BaseNetworkController struct {
 
 func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed func(string)) error {
 	// gather some information first
-	var err error
-	var retryNodes []*corev1.Node
+	var reconcileNodes []string
 	oc.localZoneNodes.Range(func(key, _ any) bool {
 		nodeName := key.(string)
 		wasAdvertised := util.IsPodNetworkAdvertisedAtNode(oc, nodeName)
@@ -201,41 +200,57 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
 			// noop
 			return true
 		}
-		var node *corev1.Node
-		node, err = oc.watchFactory.GetNode(nodeName)
-		if err != nil {
-			return false
-		}
-		retryNodes = append(retryNodes, node)
+		reconcileNodes = append(reconcileNodes, nodeName)
 		return true
 	})
-	if err != nil {
-		return fmt.Errorf("failed to reconcile network %s: %w", oc.GetNetworkName(), err)
-	}
 	reconcileRoutes := oc.routeImportManager != nil && oc.routeImportManager.NeedsReconciliation(netInfo)
 	reconcilePendingPods := !oc.IsDefault() && !oc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+	reconcileNamespaces := sets.NewString()
+	if oc.IsPrimaryNetwork() {
+		// since CanServeNamespace filters out namespace events for namespaces unknown
+		// to be served by this primary network, we need to reconcile namespaces once
+		// the network is reconfigured to serve a namespace.
+		reconcileNamespaces = sets.NewString(netInfo.GetNADNamespaces()...).Difference(
+			sets.NewString(oc.GetNADNamespaces()...))
+	}
 
 	// set the new NetInfo, point of no return
-	err = util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
+	err := util.ReconcileNetInfo(oc.ReconcilableNetInfo, netInfo)
 	if err != nil {
 		return fmt.Errorf("failed to reconcile network information for network %s: %v", oc.GetNetworkName(), err)
 	}
 
+	oc.doReconcile(reconcileRoutes, reconcilePendingPods, reconcileNodes, setNodeFailed, reconcileNamespaces.List())
+
+	return nil
+}
+
+// doReconcile performs the reconciliation after the controller NetInfo has already being
+// updated with the changes. What needs to be reconciled should already be known and
+// provided on the arguments of the method. This method returns no error and logs them
+// instead since once the controller NetInfo has been updated there is no point in retrying.
+func (oc *BaseNetworkController) doReconcile(reconcileRoutes, reconcilePendingPods bool,
+	reconcileNodes []string, setNodeFailed func(string), reconcileNamespaces []string) {
 	if reconcileRoutes {
-		err = oc.routeImportManager.ReconcileNetwork(oc.GetNetworkName())
+		err := oc.routeImportManager.ReconcileNetwork(oc.GetNetworkName())
 		if err != nil {
 			klog.Errorf("Failed to reconcile network %s on route import controller: %v", oc.GetNetworkName(), err)
 		}
 	}
 
-	for _, node := range retryNodes {
-		setNodeFailed(node.Name)
+	for _, nodeName := range reconcileNodes {
+		setNodeFailed(nodeName)
+		node, err := oc.watchFactory.GetNode(nodeName)
+		if err != nil {
+			klog.Infof("Failed to get node %s for reconciling network %s: %v", nodeName, oc.GetNetworkName(), err)
+			continue
+		}
 		err = oc.retryNodes.AddRetryObjWithAddNoBackoff(node)
 		if err != nil {
-			klog.Errorf("Failed to retry node %s for network %s: %v", node.Name, oc.GetNetworkName(), err)
+			klog.Errorf("Failed to retry node %s for network %s: %v", nodeName, oc.GetNetworkName(), err)
 		}
 	}
-	if len(retryNodes) > 0 {
+	if len(reconcileNodes) > 0 {
 		oc.retryNodes.RequestRetryObjs()
 	}
 
@@ -245,7 +260,28 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
 		}
 	}
 
-	return nil
+	namespaceAdded := false
+	for _, ns := range reconcileNamespaces {
+		namespace, err := oc.watchFactory.GetNamespace(ns)
+		if err != nil {
+			klog.Infof("Failed to get namespace %s for reconciling network %s: %v", ns, oc.GetNetworkName(), err)
+			continue
+		}
+		err = oc.retryNamespaces.AddRetryObjWithAddNoBackoff(namespace)
+		if err != nil {
+			klog.Infof("Failed to retry namespace %s for network %s: %v", ns, oc.GetNetworkName(), err)
+			continue
+		}
+		namespaceAdded = true
+	}
+	if namespaceAdded {
+		oc.retryNamespaces.RequestRetryObjs()
+	}
+
+	err := oc.requeuePeerNamespaces(reconcileNamespaces)
+	if err != nil {
+		klog.Infof("Failed to retry network policy peer namespaces for network %s: %v", oc.GetNetworkName(), err)
+	}
 }
 
 // BaseSecondaryNetworkController structure holds per-network fields and network specific
--- a/go-controller/pkg/ovn/base_network_controller_policy.go
+++ b/go-controller/pkg/ovn/base_network_controller_policy.go
@@ -1496,20 +1496,10 @@ func (bnc *BaseNetworkController) peerNamespaceUpdate(np *networkPolicy, gp *gre
 	return err
 }
 
-// requeuePeerNamespaces enqueues the namespace into network policy peer namespace
+// requeuePeerNamespace enqueues the namespace into network policy peer namespace
 // retry framework object(s) which need to be retried immediately with add event.
-func (bnc *BaseNetworkController) requeuePeerNamespaces(namespaces []string) error {
+func (bnc *BaseNetworkController) requeuePeerNamespace(namespace *corev1.Namespace) error {
 	var errors []error
-	var peerNamespaces []*corev1.Namespace
-	for _, ns := range namespaces {
-		namespace, err := bnc.watchFactory.GetNamespace(ns)
-		if err != nil {
-			errors = append(errors, fmt.Errorf("failed to retrieve namespace %s for reconciling network %s: %w",
-				ns, bnc.GetNetworkName(), err))
-			continue
-		}
-		peerNamespaces = append(peerNamespaces, namespace)
-	}
 	npKeys := bnc.networkPolicies.GetKeys()
 	for _, npKey := range npKeys {
 		err := bnc.networkPolicies.DoWithLock(npKey, func(npKey string) error {
@@ -1519,26 +1509,23 @@ func (bnc *BaseNetworkController) requeuePeerNamespaces(namespaces []string) err
 			}
 			np.RLock()
 			defer np.RUnlock()
+			if np.deleted {
+				return nil
+			}
 			var errors []error
 			for _, reconcilePeerNamespace := range np.reconcilePeerNamespaces {
-				namespaceAdded := false
-				for _, namespace := range peerNamespaces {
-					// Filter out namespace when it's labels not matching with network policy peer namespace
-					// selector.
-					if !reconcilePeerNamespace.handler.FilterFunc(namespace) {
-						continue
-					}
-					err := reconcilePeerNamespace.retryFramework.AddRetryObjWithAddNoBackoff(namespace)
-					if err != nil {
-						errors = append(errors, fmt.Errorf("failed to retry peer namespace %s for network policy %s on network %s: %w",
-							namespace.Name, npKey, bnc.GetNetworkName(), err))
-						continue
-					}
-					namespaceAdded = true
+				// Filter out namespace when it's labels not matching with network policy peer namespace
+				// selector.
+				if !reconcilePeerNamespace.handler.FilterFunc(namespace) {
+					continue
 				}
-				if namespaceAdded {
-					reconcilePeerNamespace.retryFramework.RequestRetryObjs()
+				err := reconcilePeerNamespace.retryFramework.AddRetryObjWithAddNoBackoff(namespace)
+				if err != nil {
+					errors = append(errors, fmt.Errorf("failed to retry peer namespace %s for network policy %s on network %s: %w",
+						namespace.Name, npKey, bnc.GetNetworkName(), err))
+					continue
 				}
+				reconcilePeerNamespace.retryFramework.RequestRetryObjs()
 			}
 			return utilerrors.Join(errors...)
 		})
@@ -1587,11 +1574,13 @@ func (bnc *BaseNetworkController) addPeerNamespaceHandler(
 	// a new peer namespace is newly created later under UDN network, it gets reconciled and
 	// address set is created for the namespace. so we must reconcile it for network policy
 	// as well to update gress policy ACL with matching peer namespace address set.
-	np.Lock()
-	np.reconcilePeerNamespaces = append(np.reconcilePeerNamespaces,
-		&peerNamespacesRetry{retryFramework: retryPeerNamespaces,
-			handler: namespaceHandler})
-	np.Unlock()
+	if bnc.IsPrimaryNetwork() {
+		np.Lock()
+		np.reconcilePeerNamespaces = append(np.reconcilePeerNamespaces,
+			&peerNamespacesRetry{retryFramework: retryPeerNamespaces,
+				handler: namespaceHandler})
+		np.Unlock()
+	}
 
 	return nil
 }
--- a/go-controller/pkg/ovn/base_network_controller_policy.go
+++ b/go-controller/pkg/ovn/base_network_controller_policy.go
@@ -10,7 +10,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	knet "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
@@ -162,10 +161,8 @@ type networkPolicy struct {
 
 	// network policy owns only 1 local pod handler
 	localPodHandler *factory.Handler
-	// peer namespace handlers
-	nsHandlerList []*factory.Handler
 	// peer namespace reconcilers
-	reconcilePeerNamespaces []*reconcilePeerNamespaces
+	reconcilePeerNamespaces []*peerNamespacesRetry
 	// peerAddressSets stores PodSelectorAddressSet keys for peers that this network policy was successfully added to.
 	// Required for cleanup.
 	peerAddressSets []string
@@ -190,9 +187,9 @@ type networkPolicy struct {
 	cancelableContext *util.CancelableContext
 }
 
-type reconcilePeerNamespaces struct {
-	retryNamespaces   *retry.RetryFramework
-	namespaceSelector *metav1.LabelSelector
+type peerNamespacesRetry struct {
+	retryFramework *retry.RetryFramework
+	handler        *factory.Handler
 }
 
 func NewNetworkPolicy(policy *knet.NetworkPolicy) *networkPolicy {
@@ -204,8 +201,7 @@ func NewNetworkPolicy(policy *knet.NetworkPolicy) *networkPolicy {
 		egressPolicies:          make([]*gressPolicy, 0),
 		isIngress:               policyTypeIngress,
 		isEgress:                policyTypeEgress,
-		nsHandlerList:           make([]*factory.Handler, 0),
-		reconcilePeerNamespaces: make([]*reconcilePeerNamespaces, 0),
+		reconcilePeerNamespaces: make([]*peerNamespacesRetry, 0),
 		localPods:               sync.Map{},
 	}
 	return np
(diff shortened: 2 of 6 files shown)


tssurya:
```
2025-02-05T17:55:48.3621888Z [38;5;9m[1mSummarizing 3 Failures:[0m
2025-02-05T17:55:48.3625212Z   [38;5;9m[FAIL][0m [0mOVN Kubevirt Operations [38;5;243mduring execution [0mreconcile migratable vm pods [38;5;9m[1m[It] for failing live migration[0m
2025-02-05T17:55:48.3626955Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/kubevirt_test.go:741[0m
2025-02-05T17:55:48.3630027Z   [38;5;9m[FAIL][0m [0mOVN master EgressIP Operations cluster default network [38;5;243mOn node UPDATE [0m[secondary host network] should perform proper OVN transactions when namespace and pod is created after node egress label switch [38;5;9m[1m[It] interconnect enabled; node1 in global and node2 in remote zones[0m
2025-02-05T17:55:48.3632701Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:2060[0m
2025-02-05T17:55:48.3635249Z   [38;5;9m[FAIL][0m [0mOVN master EgressIP Operations cluster default network [38;5;243mWatchEgressNodes running with WatchEgressIP [38;5;9m[1m[It] should re-balance EgressIPs when their node is removed[0m
2025-02-05T17:55:48.3637345Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/egressip_test.go:12240[0m
```
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4721#issuecomment-2637711488 nice

jcaamano:
I don't think this is how it should be done, at least not as a first approach.

It looks like we should remove `PeerNamespaceSelectorType` handling from `networkControllerPolicyEventHandler` and instead just rely on the existing `AddressSetNamespaceAndPodSelectorType`/`AddressSetPodSelectorType` combo: we listen to namespace events for selected namespaces, and when a namespace appears, we listen to any pod event on that namespace (having just a namespace selector is equivalent o having a namespace selector and a `select everything/default` pod selector).

Then when handling the pod event itself, we need to tell apart if it matched a specific pod selector or an `everything/default` pod selector: for the former we keep the current handling of adding the pod IPs to the peer address set; for the latter we update the ACLs to match on the namespace address set itself.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1393,6 +1408,14 @@ func (bnc *BaseNetworkController) handlePeerNamespaceSelectorAdd(np *networkPoli
...
 	for _, obj := range objs {
 		namespace := obj.(*kapi.Namespace)
+		netinfo, err := bnc.networkManager.GetActiveNetworkForNamespace(namespace.Name)
+		if err != nil {
+			klog.Warningf("Could not get active network for namespace %s: %v", namespace.Name, err)
+			continue
+		}
+		if bnc.GetNetworkName() != netinfo.GetNetworkName() {
+			continue
+		}
```
pperiyasamy, line 1418: this check must be removed as it's breaking mutlinet policies on secondary networks.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1541,6 +1618,11 @@ func (bnc *BaseNetworkController) shutdownHandlers(np *networkPolicy) {
 		bnc.watchFactory.RemoveNamespaceHandler(handler)
 	}
 	np.nsHandlerList = make([]*factory.Handler, 0)
+
+	for _, handler := range np.nsPodHandlers {
+		bnc.watchFactory.RemoveNamespaceHandler(handler)
```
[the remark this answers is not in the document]
pperiyasamy: good catch, it's `RemovePodHandler`, done.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1410,11 +1418,83 @@ func (bnc *BaseNetworkController) handlePeerNamespaceSelectorAdd(np *networkPoli
 			errors = append(errors, err)
 		}
 	}
+	// check if UDN label is present on netpol namespace.
```
jcaamano: So if I get this correctly, the call above in line 1405, which we are doing regardless of this being the default network or not, might fail because the address set does not exist? Which would mean there will be retries to handle this. Does it make sense to return error if we expect that the address set might not yet exist and we are setting up a separate handler to take care if it when it does?

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1410,11 +1418,83 @@ func (bnc *BaseNetworkController) handlePeerNamespaceSelectorAdd(np *networkPoli
...
+	if np.deleted {
+		np.RUnlock()
+		return nil
+	}
+	updated := false
+	var errors []error
+	for _, obj := range objs {
+		pod := obj.(*corev1.Pod)
+		// addNamespaceAddressSet is safe for concurrent use, doesn't require additional synchronization
+		nsUpdated, err := gp.addNamespaceAddressSet(pod.Namespace, bnc.addressSetFactory)
```
jcaamano: The namespace address set might not exist yet. Is there any state on the pod that we can access (cheaply) that we can use if the address set might exist already? To avoid spurious errors and retries.

go-controller/pkg/ovn/network_controller_policy_event_handler.go
```diff
@@ -127,16 +130,22 @@ func (h *networkControllerPolicyEventHandler) GetResourceFromInformerCache(key s
...
+		return h.bnc.handlePeerNamespacePodAddUpdate(extraParameters.np, extraParameters.gp, obj)
 
 	case factory.AddressSetNamespaceAndPodSelectorType:
-		peerAS := h.extraParameters.(*PodSelectorAddrSetHandlerInfo)
-		return h.bnc.handleNamespaceAddUpdate(peerAS, obj)
+		if peerAS, ok := h.extraParameters.(*PodSelectorAddrSetHandlerInfo); ok {
+			return h.bnc.handleNamespaceAddUpdate(peerAS, obj)
+		}
+		extraParameters := h.extraParameters.(*NetworkPolicyExtraParameters)
+		return h.bnc.handlePeerNamespaceAdd(extraParameters.np, extraParameters.gp, obj)
```
jcaamano: I am not sure this is exactly what I suggested. 

I suggested one handler for namespaces and one handler for pods. But this approach has 2 of each. So in total 4 retry frameworks/handlers instead of 2. These times the number of gress policies in a policy, times the number of policies in the system, might not be something easily ignored. As a reference, before we had 3: two namespace handlers and a pod handler. We had a chance to go to 2, but we went to 4 instead.

The way it was chosen to be implemented is also confusing. If we just had the 2 of handlers I suggested, it does make sense to have only two *Types (AddressSetNamespaceAndPodSelectorType and AddressSetPodSelectorType). Having 4, then it would make sense to have 4 *Types. Instead we are reusing the 2 former *Types, but using extraParameters type to tell the two handlers apart. *Types is what makes it easy to understand what is being handled (among other things), however the extraParameters does not, as they are general purpose reused in different handlers. With this I want to say that the code turns out to be more confusing as well. Comments did not help me much deciphering this either.

@pperiyasamy 
Do you have any comments on the above? Perhaps it is not possible to do what I am suggesting?
Given that we are increasing the number of handlers, do you think it would be interesting to to do a scale run of this? Maybe it was already done?
Can you add pertinent comments, both in the newNetpolRetryFramework block, as well as here, explaining what each of the two subtypes of handlers, based on extra parameters, are doing? Just so we can navigate this code better.

Thanks!

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: Yes, The line 1405 doesn't fail, but still address set doesn't exist  for the UDN namespace, `as.GetASHashNames()` returning empty string yet. so retries on the namespace handler won't help.
This happens when UDN has more than one namespace and no pods available on those namespaces yet. Hence we needed pod handler to get namespace address set and update the ACL accordingly.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: sure, using `bnc.getNamespaceLocked(pod.Namespace, true)` method now, it worked on my local testing. hope this is fine.

go-controller/pkg/ovn/network_controller_policy_event_handler.go
pperiyasamy: yes @jcaamano your observation is correct, it's now having 4 handler for a UDN network policy. Tried to fix the issue with 2 handlers with https://github.com/ovn-kubernetes/ovn-kubernetes/commit/a216b9a97d3d817499993c4a61a8389acd6b6da5, but got into deadlock problem on np lock (https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/base_network_controller_policy.go#L970 and https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/base_network_controller_policy.go#L1463) while syncing existing namespace. I think this is reason why we are introduced a separate peer namespace handler already. This new pod handler is spin up only when needed and also being done only for NP belonging to a UDN namespace, so may have impact on UDN deployments. WDYT ?
Have added relevant comments in newNetpolRetryFramework block and in each of the two subtypes of handlers. PTAL.

go-controller/pkg/ovn/base_network_controller_policy.go
jcaamano: > so retries on the namespace handler won't help

why are we retrying then?

go-controller/pkg/ovn/base_network_controller_policy.go
jcaamano: That's not exactly what I suggested and incurs in locking and might actually be more expensive. If there is no `state on the pod that we can access (cheaply) `, then perhaps it is just better to avoid retrying if `addNamespaceAddressSet` returns a libovsdb not found error, as I suggested o the previous comment as well?

jcaamano: In any case, any of this would only work if we are guaranteed to get a pod event when the address set gets created, which might not be true. What guarantees this?

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -245,6 +253,27 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
...
+			for _, retryNamespace := range retryNamespaces {
+				err = retryNamespace.AddRetryObjWithAddNoBackoff(nsObj)
+				if err != nil {
+					klog.Infof("Failed to retry namespace %s for network policy %s on network %s: %v",
+						ns, npKey, oc.GetNetworkName(), err)
+				}
+			}
+		}
+	}
+
```
jcaamano: I would rather have the list of `retryNamespaces` to be part of each `networkPolicy` for which we already have a list in the base network controller.

I would also probably extract this to a separate `requeueNamespaces` method for cleanness. In that method you would just iterate the `networkPolicy`s and call a method upon them that would internally requeue.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -214,6 +220,8 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
 	}
 	reconcileRoutes := oc.routeImportManager != nil && oc.routeImportManager.NeedsReconciliation(netInfo)
 	reconcilePendingPods := !oc.IsDefault() && !oc.ReconcilableNetInfo.EqualNADs(netInfo.GetNADs()...)
+	reconcileNamespaces := sets.NewString(netInfo.GetNADNamespaces()...).Difference(
+		sets.NewString(oc.GetNADNamespaces()...))
```
jcaamano: This is a good place to check if this is a primary network and do nothing otherwise. Along with a comment explaining why we need to do it: 
since `CanServeNamespace` filters out namespace events for namespaces unknown to be served by this primary network, we need to reconcile namespaces once the network is reconfigured to serve a namespace.

test/e2e/network_segmentation_policy.go
```diff
@@ -271,6 +278,35 @@ var _ = ginkgo.Describe("Network Segmentation: Network Policies", func() {
...
+
+				ginkgo.By("asserting the *orange client* pod can not contact the allow server pod exposed endpoint")
+				gomega.Eventually(func() error {
+					return reachServerPodFromClient(cs, allowServerPodConfig, clientPodConfig, allowServerPodIP, port)
+				}, 1*time.Minute, 6*time.Second).ShouldNot(gomega.Succeed())
+
+				ginkgo.By("asserting the *orange client* pod can not contact deny server pod exposed endpoint")
+				gomega.Eventually(func() error {
+					return reachServerPodFromClient(cs, denyServerPodConfig, clientPodConfig, denyServerPodIP, port)
+				}, 1*time.Minute, 6*time.Second).ShouldNot(gomega.Succeed())
```
jcaamano, line 349: I wonder if this test is actually reproducing the original issue and exercising the new code. 
So without the changes, this test would not pass?
Wouldn't the test actually be required to create the NADs at different times?

pperiyasamy: yes, The issue is reproduced after having `allowTrafficToPodFromNamespacePolicy` network policy in the test which is having only peer namespace selectors. 
The NADs are created after namespaces so new network controller changes would reconcile those namespaces.
now gressPolicy#addNamespaceAddressSet throws error when empty hash is found, so netpol ACL would get updated either because of pod is created before reattempts expires (or) network controller would have reconciled it.

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: sure, moved `retryNamespaces` to be part of each `networkPolicy` and having a new `requeuePeerNamespace` method to reconcile peer namespace for network policies.

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: ok, added this comment.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -245,6 +253,23 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
 		}
 	}
 
+	for _, ns := range reconcileNamespaces.List() {
```
jcaamano: This method is tarting to get too big. One neat idea would be to have a separate method
```
// doReconcile performs the reconciliation after the controller NetInfo has already being
// updated with the changes. What needs to be reconciled should already be known and 
// provided on the arguments of the method. This method returns no error and logs them
// instead since once the controller NetInfo has been updated there is no point in retrying.
func (oc *BaseNetworkController) doReconcile(reconcileRoutes, reconcilePendingPods bool, reconcileNodes, reconcileNamespaces []string) {
    ....
}
```
including everything under line 232

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1490,6 +1494,39 @@ func (bnc *BaseNetworkController) peerNamespaceUpdate(np *networkPolicy, gp *gre
...
+// requeuePeerNamespace enqueues the namespace into network policy peer namespace
+// retry framework object(s) which need to be retried immediately with add event.
+func (bnc *BaseNetworkController) requeuePeerNamespace(namespace *corev1.Namespace) error {
+	if namespace == nil {
+		return nil
+	}
+	npKeys := bnc.networkPolicies.GetKeys()
+	var errors []error
+	for _, npKey := range npKeys {
+		err := bnc.networkPolicies.DoWithLock(npKey, func(npKey string) error {
```
jcaamano, line 1505: Just an observation here. I wonder why we have both `DoWithLock` and `np.Mutex`. Seems like we don't need both of them in general.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1522,7 +1559,14 @@ func (bnc *BaseNetworkController) addPeerNamespaceHandler(
...
 	}
-
+	// Add peer namespace retry framework object into np.retryPeerNamespaces list so that
+	// when a new peer namespace is newly created later under UDN network, it gets reconciled
+	// and network policy ACL is updated with namespace address set.
+	np.Lock()
+	if util.IsNetworkSegmentationSupportEnabled() && bnc.IsPrimaryNetwork() {
+		np.retryPeerNamespaces = append(np.retryPeerNamespaces, retryPeerNamespaces)
+	}
+	np.Unlock()
```
jcaamano: lock inside the if block

jcaamano: explain in the comment why you are doing this only for primary networks

test/e2e/network_segmentation_policy.go
jcaamano: That's what I meant. 
How do you control in the test if it is either because of pod is created before reattempts expires (or) network controller would have reconciled it? My guess is you want to test the reconciliation code path, not the retry code path.
I could ask it another way: if you remove the code changes but keep this test changes, does the test case fail? How?

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: AFAIU `DoWithLock` is needed to synchronize accessing map entry whereas `np.Mutex` is used to synchrnonize accessing network policy object itself. this is the pattern used across np code, so not changing that in this PR.

test/e2e/network_segmentation_policy.go
pperiyasamy: This NAD is being created for every namespace  (totally 4 namespaces) in this test, so it is going to reconcile 3 remaining namespaces. so reconcile code changes will get exercised from the test.

without code changes the new test changes would fail. that is because ngress ACL entry would have wrong address set match `"$"` for peer address set.

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: ok, done.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: ok, added few more comments on this.

go-controller/pkg/ovn/base_network_controller_policy.go
jcaamano: I don't think that's it. `GetKeys` and `Load` already constitute a synchronized access to the map. `DoWithLock` should cover up for `np.Mutex` if used broadly and correctly, and additionally prevent for a key being removed from the map while being used. So I guess I am missing the point of `np.Mutex`.
However, I didn't ask to do anything, hence just the observation. Agree to continue the current usage pattern for now.

test/e2e/network_segmentation_policy.go
jcaamano: > without code changes the new test changes would fail. that is because ngress ACL entry would have wrong address set match `"$"` for peer address set.

wouldn't it retry and eventually succeed?

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -206,36 +206,55 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
...
+
+	return nil
+}
+
+// doReconcile performs the reconciliation after the controller NetInfo has already being
+// updated with the changes. What needs to be reconciled should already be known and
+// provided on the arguments of the method. This method returns no error and logs them
+// instead since once the controller NetInfo has been updated there is no point in retrying.
+func (oc *BaseNetworkController) doReconcile(reconcileRoutes, reconcilePendingPods bool,
+	reconcileNodes []*corev1.Node, setNodeFailed func(string), reconcileNamespaces []string) {
```
jcaamano: nit: suggested here `[]string` for nodes as well so that we move `oc.watchFactory.GetNode(nodeName)` also here and simplify the other method more

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -245,7 +260,28 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
...
 
-	return nil
+	namespaceAdded := false
+	for _, ns := range reconcileNamespaces {
+		namespace, err := oc.watchFactory.GetNamespace(ns)
+		if err != nil {
+			klog.Infof("Failed to get namespace %s for reconciling network %s: %v", ns, oc.GetNetworkName(), err)
+			continue
+		}
+		err = oc.retryNamespaces.AddRetryObjWithAddNoBackoff(namespace)
```
pperiyasamy, line 274: will have to reconcile only namespaces which match on network policy peer namespace selector, otherwise it adds other UDN namespaces as well into gress policy ACL, will fix it tomorrow.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -186,17 +190,23 @@ type networkPolicy struct {
 	cancelableContext *util.CancelableContext
 }
 
+type reconcilePeerNamespaces struct {
+	retryNamespaces   *retry.RetryFramework
+	namespaceSelector *metav1.LabelSelector
+}
```
jcaamano: The `factory.Handler` already has a `filterFunc` that you can use. I think it might be more beneficial to correlate the `RetryFramework` and `Handler` together and export a method on the `Handler` to get that `filterFunc` so that you can use it.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -245,7 +260,28 @@ func (oc *BaseNetworkController) reconcile(netInfo util.NetInfo, setNodeFailed f
...
+		if err != nil {
+			klog.Infof("Failed to retry namespace %s for network %s: %v", ns, oc.GetNetworkName(), err)
+			continue
+		}
+		namespaceAdded = true
+	}
+	if namespaceAdded {
```
jcaamano, line 281: nit: I would just use `len(reconcileNamespaces) > 0` as we do with the nodes

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1490,6 +1496,56 @@ func (bnc *BaseNetworkController) peerNamespaceUpdate(np *networkPolicy, gp *gre
...
+			var errors []error
+			for _, reconcilePeerNamespace := range np.reconcilePeerNamespaces {
+				namespaceAdded := false
+				for _, ns := range namespaces {
+					namespace, err := bnc.watchFactory.GetNamespace(ns)
+					if err != nil {
+						errors = append(errors, fmt.Errorf("failed to retrieve peer namespace %s for network policy %s on network %s: %w",
+							ns, npKey, bnc.GetNetworkName(), err))
+						continue
+					}
```
jcaamano: Would it make sense to gather namespace objects out of the top level loop? Otherwise you are doing it many times for the same namespace while holding locks.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1490,6 +1496,56 @@ func (bnc *BaseNetworkController) peerNamespaceUpdate(np *networkPolicy, gp *gre
 	return err
 }
 
+// requeuePeerNamespaces enqueues the namespace into network policy peer namespace
+// retry framework object(s) which need to be retried immediately with add event.
+func (bnc *BaseNetworkController) requeuePeerNamespaces(namespaces []string) error {
+	npKeys := bnc.networkPolicies.GetKeys()
+	var errors []error
+	for _, npKey := range npKeys {
+		err := bnc.networkPolicies.DoWithLock(npKey, func(npKey string) error {
```
jcaamano, line 1505: One of the concerns I have is the potential contention that all the locking causes to the network manager reconciliation loop.

Thinking of alternatives...would it make more sense to call this from the namespace reconciliation loop, as we reconcile each namespace?

go-controller/pkg/ovn/base_network_controller.go
pperiyasamy: yes, but in a worst case, there is a chance that none of the namespaces (see the above error handling block) being added to retry framework , hence using `namespaceAdded` flag here.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: I think namespace objs are retrieved from informer cache here, so thought it is not read heavy, but if you still insist, then move it to a outer loop without np lock.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: even if we call from namespace reconciliation loop, we need to iterate over `networkPolicies` and must use locking order `namespace Lock -> oc.networkPolicies key Lock -> networkPolicy.Lock` to access `networkPolicy`, isn't it ?

jcaamano: Right. But you don't block the network manager reconciliation loop in that case. So you block one aspect of the affected network, instead of other/new unrelated networks.

go-controller/pkg/ovn/base_network_controller_policy.go
jcaamano: Might not be heavy, but it's really straight forward to make it more efficient just in case.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -680,6 +680,14 @@ func (bsnc *BaseSecondaryNetworkController) AddNamespaceForSecondaryNetwork(ns *
...
 	}
 	defer nsUnlock()
+	// Enqueue the UDN namespace into network policy controller if it needs to be
+	// processed by network policy peer namespace handlers.
+	if bsnc.IsPrimaryNetwork() {
+		err = bsnc.requeuePeerNamespace(ns)
+		if err != nil {
+			return fmt.Errorf("failed to requeue peer namespace %s: %v", ns.Name, err)
+		}
+	}
```
jcaamano, line 690: Is there any specific reason we are holding the nsInfo lock while calling this?

test/e2e/network_segmentation_policy.go
```diff
@@ -271,6 +278,72 @@ var _ = ginkgo.Describe("Network Segmentation: Network Policies", func() {
...
+				gomega.Consistently(func() v1.PodPhase {
+					updatedPod, err := cs.CoreV1().Pods(clientPodConfig.namespace).Get(context.Background(),
+						clientPodConfig.name, metav1.GetOptions{})
+					if err != nil {
+						return v1.PodFailed
+					}
+					return updatedPod.Status.Phase
+				}, 1*time.Minute, 6*time.Second).Should(gomega.Equal(v1.PodPending))
+
+				ginkgo.By("creating NAD for red and orange namespaces and check pod moves into running state")
```
jcaamano, line 303: Can you add some comment here explaining what you are testing. Say something like "The pod won't run and the namespace address set won't be created until the NAD for the network is added to the namespace and we test here that once that happens the policy is reconciled to account for it "

jcaamano (approved):

PR #5036, opened 2025-02-05
[WIP] Make DBUS controller more efficient

trozet:
The previous code was iterating a for loop every 50ms to try to detect if there was a state change to an internal cache provided by coreos dbus package. This consume too much CPU. Instead, only leverage the coreos package to handle connection setup, then directly use dbus to receive systemd unit events. This drastically reduces CPU and still detects when Kubelet is restarted.


trozet:
@npinaeva

go-controller/pkg/node/udn_isolation.go
```diff
@@ -293,52 +295,77 @@ func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
...
-		return fmt.Errorf("failed to subscribe to systemd events: %w", err)
+		return err
 	}
-	// interval is important here as we need to catch the restart state, before it is running again
-	events, errChan := conn.SubscribeUnitsCustom(50*time.Millisecond, 0, func(u1, u2 *dbus.UnitStatus) bool { return *u1 != *u2 },
-		func(s string) bool {
-			return s != "kubelet.service"
-		})
+
+	signalChan := make(chan *dbus.Signal, 10)
```
npinaeva, line 322: every event that happens when the channel is full will spin up a goroutine, so considering I have got 40 events per second on a regular kind cluster, it may make sense to increase this number to 50

go-controller/pkg/node/udn_isolation.go
```diff
@@ -293,52 +295,77 @@ func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
...
-						waitingForActive = true
-					} else if waitingForActive {
-						klog.Infof("Kubelet was restarted, re-applying UDN host isolation")
-						err = m.updateKubeletCgroup()
-						if err != nil {
-							klog.Errorf("Failed to re-apply UDN host isolation: %v", err)
-						} else {
-							waitingForActive = false
+			case signal := <-signalChan:
+				klog.V(5).Infof("D-Bus event received: %#v", signal)
```
npinaeva, line 335: I think this may be too much logging?

go-controller/pkg/node/udn_isolation.go
```diff
@@ -293,52 +295,77 @@ func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
...
+				// Extract unit name from path
+				unitPath := signal.Path
+				parts := strings.Split(string(unitPath), "/")
+				if len(parts) < 6 || parts[4] != "unit" {
```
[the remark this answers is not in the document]
npinaeva: is it the same on kind vs OCP?

go-controller/pkg/node/udn_isolation.go
```diff
@@ -293,52 +295,77 @@ func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
...
+					continue
+				}
+				escapedUnit := parts[5]
+				unitName := strings.ReplaceAll(escapedUnit, "_2e", ".")
+
+				if unitName == "kubelet.service" {
+					changes := signal.Body[1].(map[string]dbus.Variant)
+					if state, exists := changes["ActiveState"]; exists {
+						newState := state.Value().(string)
+						if newState == "active" {
```
npinaeva, line 349: on my kind cluster, this causes 4 cgroup updates for every kubelet restart. That is why I had this `waitingForActive` logic.

go-controller/pkg/node/udn_isolation.go
```diff
@@ -293,52 +295,77 @@ func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
...
 		}
 	}()
+
+	conn.BusObject().Call("org.freedesktop.DBus.AddMatch", 0,
+		"type='signal',interface='org.freedesktop.DBus.Properties',member='PropertiesChanged'")
+
+	sigObj := conn.Object("org.freedesktop.systemd1", dbus.ObjectPath("/org/freedesktop/systemd1"))
+	if err := sigObj.Call("org.freedesktop.systemd1.Manager.Subscribe", 0).Store(); err != nil {
+		return fmt.Errorf("failed to subscribe to systemd: %w", err)
+	}
```
npinaeva, line 367: why is this part set up after the signal handler is added?

go-controller/pkg/node/udn_isolation.go
```diff
@@ -293,52 +295,77 @@ func (m *UDNHostIsolationManager) updateKubeletCgroup() error {
 // If a new cgroup is created, you load the filtering policy for the new cgroup and then add
 // processes to that cgroup. You only have to follow the right sequence to avoid problems.
 func (m *UDNHostIsolationManager) runKubeletRestartTracker(ctx context.Context) (err error) {
-	conn, err := dbus.NewSystemdConnectionContext(ctx)
+
+	conn, err := dbus.Dial("unix:path=/run/systemd/private", dbus.WithContext(ctx))
 	if err != nil {
-		return fmt.Errorf("failed to connect to systemd: %w", err)
+		return err
```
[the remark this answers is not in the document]
trozet: what do you suggest?

go-controller/pkg/node/udn_isolation.go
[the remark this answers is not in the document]
trozet: I think it is the same. I have a downstream PR https://github.com/openshift/ovn-kubernetes/pull/2440

I will launch in cluster bot and test.

go-controller/pkg/node/udn_isolation.go
trozet: its V5, you still think I should remove it?

go-controller/pkg/node/udn_isolation.go
trozet: I'll bump it to 100, which looks like waht coreos was using anyway

go-controller/pkg/node/udn_isolation.go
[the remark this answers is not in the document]
npinaeva: leave/update extra message wrap here and for every system error saying which operation returned error

go-controller/pkg/node/udn_isolation.go
trozet: I see the same thing in my setup, but the cost of executing 4 times on a kubelet restart seems minimal when i pprof'ed.

go-controller/pkg/node/udn_isolation.go
trozet: so this is the same way it happens in the coreos code. There they call "dispatch()" function which sets up the signal, and then they Subscribe after. My understanding is you want to be able to receive and handle events before you start subscribing, so that you dont miss any events.

go-controller/pkg/node/udn_isolation.go
npinaeva: I wasn't sure what you can debug with this. Unexpected path that is not parsed correctly?

go-controller/pkg/node/udn_isolation.go
npinaeva: i mean why make 4 system calls when you can do 1? in case you miss deactivating events? is it likely to happen? or some other reason?

go-controller/pkg/node/udn_isolation.go
npinaeva: can't say I understand hoe that helps not missing events and why it is important on startup, but also not a big deal :)

go-controller/pkg/node/udn_isolation.go
trozet: right, i thought in the future if its not working, its probably going to be us parsing events incorrectly, so this would help us identify if that was the bug

go-controller/pkg/node/udn_isolation.go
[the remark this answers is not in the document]
trozet: oh sorry i was confused and looking at the wrong code!

go-controller/pkg/node/udn_isolation.go
trozet: exactly...i'm worried about missing events. I dont want to internally track state in ovnk. I think it could be error prone. Kubelet doesn't restart often so when it happens its not a big hit to just reprogram the nftables rules 4x. We can explore maybe examining the properties event in more detail to see if there is something else we could filter further on. I'll try to take a look at that, but I dont think this is a huge issue.

trozet:
downstream looks good: https://github.com/openshift/ovn-kubernetes/pull/2442

npinaeva (approved):

PR #5039, opened 2025-02-07
Enrich CNI request pod UID error with the mismatched UIDs

martinkennelly:
I am trying to debug an issue where this error occurs and I dont easily know what pod UIDs that were compared.

Downstream issue; https://issues.redhat.com/browse/OCPBUGS-48760

### DIFF
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -101,7 +101,8 @@ func (pr *PodRequest) checkOrUpdatePodUID(pod *kapi.Pod) error {
 		pr.PodUID = string(pod.UID)
 	} else if string(pod.UID) != pr.PodUID {
 		// Exit early if the pod was deleted and recreated already
-		return fmt.Errorf("pod deleted before sandbox %v operation began", pr.Command)
+		return fmt.Errorf("pod deleted before sandbox %v operation began. Request Pod UID %s is different from "+
+			"the Pod UID (%s) retrieved from the informer/API", pr.Command, pr.PodUID, pod.UID)
 	}
 	return nil
 }


go-controller/pkg/cni/cni.go
```diff
@@ -101,7 +101,8 @@ func (pr *PodRequest) checkOrUpdatePodUID(pod *kapi.Pod) error {
 		pr.PodUID = string(pod.UID)
 	} else if string(pod.UID) != pr.PodUID {
 		// Exit early if the pod was deleted and recreated already
-		return fmt.Errorf("pod deleted before sandbox %v operation began", pr.Command)
+		return fmt.Errorf("pod deleted before sandbox %v operation began. Request Pod UID %s is different from the"+
+			" informer/API servers (%s)", pr.Command, pr.PodUID, pod.UID)
```
tssurya, line 105:
```suggestion
			" what informer/API holds (%s)", pr.Command, pr.PodUID, pod.UID)
```
nit ^
I think its a tiny bit confusing when you say different from the informer/API servers... feels like their version of the pod UID bit is missing

tssurya (approved):

jluhrsen:
@martinkennelly , if we don't **need** this PR to actually be in the shipped code-base you can just file an openshift/ovnk PR and we can run the jobs on that PR to get our data. just a thought, since it's not the fastest way to get something approved u/s then in to a d/s merge PR and then approved and merged.

jluhrsen:
> @martinkennelly , if we don't **need** this PR to actually be in the shipped code-base you can just file an openshift/ovnk PR and we can run the jobs on that PR to get our data. just a thought, since it's not the fastest way to get something approved u/s then in to a d/s merge PR and then approved and merged.

couldn't resist and made [this](https://github.com/openshift/ovn-kubernetes/pull/2451) on 4.18

tssurya:
I don't see any harm in changing a single line log to add two extra params to it.. so merging this

PR #5040, opened 2025-02-07
[UDN: host isolation] don't return error for pods with no IPs assigned.

npinaeva:
## 📑 Description

Avoid spamming errors for pods that just don't have IPs assigned yet.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/udn_isolation.go
+++ b/go-controller/pkg/node/udn_isolation.go
@@ -491,6 +491,10 @@ func (m *UDNHostIsolationManager) getPodInfo(podKey string, pod *v1.Pod) (*podIn
 	// only add pods with primary UDN
 	primaryUDN, err := m.isPodPrimaryUDN(pod)
 	if err != nil {
+		if util.IsAnnotationNotSetError(err) {
+			// pod IPs were not assigned yet, expecting an update event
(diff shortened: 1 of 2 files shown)


npinaeva:
cc @martinkennelly

go-controller/pkg/node/udn_isolation.go
```diff
@@ -491,6 +491,10 @@ func (m *UDNHostIsolationManager) getPodInfo(podKey string, pod *v1.Pod) (*podIn
 	// only add pods with primary UDN
 	primaryUDN, err := m.isPodPrimaryUDN(pod)
 	if err != nil {
+		if util.IsAnnotationNotSetError(err) {
```
jcaamano, line 494: Just a heads that (unfortunately) this won't be a `IsAnnotationNotSetError` if the annotation is set but just not for a specific network.

jcaamano (approved):

go-controller/pkg/node/udn_isolation.go
npinaeva: yeah that part I expected, but I always ask for "default" network, so this shouldn't be a problem

PR #5041, opened 2025-02-07
Allow default network pods to reach localnet on the same node

ricky-rav:
Add flows at prio=102 to allow traffic from the OVN network to the host network to be handled normally by the bridge instead of forwarding it directly to the NIC (which would be done by the flow at prio=100). This allows pods in the default network to reach localnet pods on the same node.

Add e2e tests to check connections from default network to localnet on different nodes and on the same node.

Testing downstream here: https://github.com/openshift/ovn-kubernetes/pull/2467

Follow-up PR for localnet->host: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5093 

Fixes #OCPBUGS-43004

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -837,7 +837,7 @@ func (npw *nodePortWatcher) AddService(service *corev1.Service) error {
 		// No endpoint object exists yet so default to false
 		hasLocalHostNetworkEp = false
 	} else {
-		nodeIPs := npw.nodeIPManager.ListAddresses()
+		nodeIPs, _ := npw.nodeIPManager.ListAddresses()
 		localEndpoints = npw.GetLocalEligibleEndpointAddresses(epSlices, service)
 		hasLocalHostNetworkEp = util.HasLocalHostNetworkEndpoints(localEndpoints, nodeIPs)
 	}
@@ -931,7 +931,7 @@ func (npw *nodePortWatcher) deleteConntrackForService(service *corev1.Service) e
 	}
 	if util.ServiceTypeHasNodePort(service) {
 		// remove conntrack entries for NodePorts
-		nodeIPs := npw.nodeIPManager.ListAddresses()
+		nodeIPs, _ := npw.nodeIPManager.ListAddresses()
 		for _, nodeIP := range nodeIPs {
 			for _, svcPort := range service.Spec.Ports {
 				if err := util.DeleteConntrackServicePort(nodeIP.String(), svcPort.NodePort, svcPort.Protocol,
@@ -1014,7 +1014,7 @@ func (npw *nodePortWatcher) SyncServices(services []interface{}) error {
 			klog.V(5).Infof("No endpointslice found for service %s in namespace %s during sync", service.Name, service.Namespace)
 			continue
 		}
-		nodeIPs := npw.nodeIPManager.ListAddresses()
+		nodeIPs, _ := npw.nodeIPManager.ListAddresses()
 		localEndpoints := npw.GetLocalEligibleEndpointAddresses(epSlices, service)
 		hasLocalHostNetworkEp := util.HasLocalHostNetworkEndpoints(localEndpoints, nodeIPs)
 		npw.getAndSetServiceInfo(name, service, hasLocalHostNetworkEp, localEndpoints)
@@ -1113,7 +1113,7 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) e
 	}
 
 	klog.V(5).Infof("Adding endpointslice %s in namespace %s", epSlice.Name, epSlice.Namespace)
-	nodeIPs := npw.nodeIPManager.ListAddresses()
+	nodeIPs, _ := npw.nodeIPManager.ListAddresses()
 	epSlices, err := npw.watchFactory.GetServiceEndpointSlices(svc.Namespace, svc.Name, netInfo.GetNetworkName())
 	if err != nil {
 		// No need to continue adding the new endpoint slice, if we can't retrieve all slices for this service
@@ -1267,7 +1267,7 @@ func (npw *nodePortWatcher) UpdateEndpointSlice(oldEpSlice, newEpSlice *discover
 	}
 
 	// Update rules and service cache if hasHostNetworkEndpoints status changed or localEndpoints changed
-	nodeIPs := npw.nodeIPManager.ListAddresses()
+	nodeIPs, _ := npw.nodeIPManager.ListAddresses()
 	epSlices, err := npw.watchFactory.GetServiceEndpointSlices(newEpSlice.Namespace, namespacedName.Name, netInfo.GetNetworkName())
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
@@ -1766,7 +1766,7 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 	return dftFlows, nil
 }
 
-func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
+func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, error) {
 	// CAUTION: when adding new flows where the in_port is ofPortPatch and the out_port is ofPortPhys, ensure
 	// that dl_src is included in match criteria!
 	ofPortPhys := bridge.ofPortPhys
@@ -1849,6 +1849,11 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
 							"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:%s",
 							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
 							netConfig.masqCTMark, ofPortPhys))
+
+					// Allow OVN->Host traffic on the same node
+					if config.Gateway.Mode == config.GatewayModeShared || config.Gateway.Mode == config.GatewayModeLocal {
+						dftFlows = append(dftFlows, ovnToHostNetworkNormalActionFlows(netConfig, bridgeMacAddress, hostSubnets, false)...)
+					}
 				} else {
 					//  for UDN we additionally SNAT the packet from masquerade IP -> node IP
 					dftFlows = append(dftFlows,
@@ -1940,6 +1945,11 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
 						fmt.Sprintf("cookie=%s, priority=100, in_port=%s, dl_src=%s, ipv6, "+
 							"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:%s",
 							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone, netConfig.masqCTMark, ofPortPhys))
+
+					// Allow OVN->Host traffic on the same node
+					if config.Gateway.Mode == config.GatewayModeShared || config.Gateway.Mode == config.GatewayModeLocal {
+						dftFlows = append(dftFlows, ovnToHostNetworkNormalActionFlows(netConfig, bridgeMacAddress, hostSubnets, true)...)
+					}
 				} else {
 					//  for UDN we additionally SNAT the packet from masquerade IP -> node IP
 					dftFlows = append(dftFlows,
@@ -2116,6 +2126,67 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
 	}
 
 	return dftFlows, nil
+}
+
+// ovnToHostNetworkNormalActionFlows returns the flows that allow IP{v4,v6} traffic from the OVN network to the host network
+// when the destination is on the same node as the sender. This is necessary for pods in the default network to reach
+// localnet pods on the same node, when the localnet is mapped to breth0. The expected srcMAC is the MAC address of breth0
+// and the expected hostSubnets is the host subnets found on the node primary interface.
+func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC string, hostSubnets []*net.IPNet, isV6 bool) []string {
+	var inPort, ctMark, ipFamily, ipFamilyDest string
+	var flows []string
+
+	if config.Gateway.Mode == config.GatewayModeShared {
+		inPort = netConfig.ofPortPatch
+		ctMark = netConfig.masqCTMark
+	} else if config.Gateway.Mode == config.GatewayModeLocal {
+		inPort = "LOCAL"
+		ctMark = ctMarkHost
+	} else {
+		return nil
+	}
+
+	if isV6 {
+		ipFamily = "ipv6"
+		ipFamilyDest = "ipv6_dst"
+	} else {
+		ipFamily = "ip"
+		ipFamilyDest = "nw_dst"
+	}
+
+	for _, hostSubnet := range hostSubnets {
+		if (hostSubnet.IP.To4() == nil) != isV6 {
+			continue
+		}
+		// IP traffic from the OVN network to the host network should be handled normally by the bridge instead of
+		// being output directly to the NIC by the existing flow at prio=100.
+		flows = append(flows,
+			fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, %s, %s=%s, "+
+				"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+				defaultOpenFlowCookie,
+				inPort,
+				srcMAC,
+				ipFamily,
+				ipFamilyDest,
+				hostSubnet.String(),
+				config.Default.ConntrackZone,
+				ctMark))
+	}
+
+	if isV6 {
+		// Neighbor discovery in IPv6 happens through ICMPv6 messages to a special destination (ff02::1:ff00:0/104),
+		// which has nothing to do with the host subnets we're matching against in the flow above at prio=102.
+		// Let's allow neighbor discovery by matching against icmp type and in_port.
+		for _, icmpType := range []int{types.NeighborSolicitationICMPType, types.NeighborAdvertisementICMPType} {
+			flows = append(flows,
+				fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, icmp6, icmpv6_type=%d, "+
+					"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+					defaultOpenFlowCookie, inPort, srcMAC, icmpType,
+					config.Default.ConntrackZone, ctMark))
+		}
+	}
+
+	return flows
 }
 
 func setBridgeOfPorts(bridge *bridgeConfiguration) error {
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -124,20 +124,22 @@ func (c *addressManager) delAddr(ipnet net.IPNet, linkIndex int) bool {
 }
 
 // ListAddresses returns all the addresses we know about
-func (c *addressManager) ListAddresses() []net.IP {
+func (c *addressManager) ListAddresses() ([]net.IP, []*net.IPNet) {
 	c.Lock()
 	defer c.Unlock()
 	addrs := sets.List(c.cidrs)
-	out := make([]net.IP, 0, len(addrs))
+	addresses := make([]net.IP, 0, len(addrs))
+	networkAddresses := make([]*net.IPNet, 0, len(addrs))
 	for _, addr := range addrs {
-		ip, _, err := net.ParseCIDR(addr)
+		ip, networkAddress, err := net.ParseCIDR(addr)
 		if err != nil {
 			klog.Errorf("Failed to parse %s: %v", addr, err)
 			continue
 		}
-		out = append(out, ip)
+		addresses = append(addresses, ip)
+		networkAddresses = append(networkAddresses, networkAddress)
 	}
-	return out
+	return addresses, networkAddresses
 }
 
 type subscribeFn func() (bool, chan netlink.AddrUpdate, error)
--- a/go-controller/pkg/node/openflow_manager.go
+++ b/go-controller/pkg/node/openflow_manager.go
@@ -212,7 +212,7 @@ func (c *openflowManager) Run(stopChan <-chan struct{}, doneWg *sync.WaitGroup) 
 
 // updateBridgeFlowCache generates the "static" per-bridge flows
 // note: this is shared between shared and local gateway modes
-func (c *openflowManager) updateBridgeFlowCache(extraIPs []net.IP) error {
+func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []*net.IPNet) error {
 	// protect defaultBridge config from being updated by gw.nodeIPManager
 	c.defaultBridge.Lock()
 	defer c.defaultBridge.Unlock()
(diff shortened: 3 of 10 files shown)


ricky-rav:
@maiqueb here I tried to reproduce the scenario described in https://issues.redhat.com/browse/OCPBUGS-43004. Basically I took one of the e2e localnet tests you wrote, removed the attachment from the existing client and added a second client.

I launch a server pod in the localnet secondary network and two client pods in the default cluster network: one on the same node as the server pod, one on a different node.

Neither client pod can reach the server, lol. When dumping traffic on the client node, I see that the client gets a RST from the server IP:
```
17:38:36.244591 c1d27cde83a304d P   IP 10.244.1.17.3131 > 60.128.0.2.9000: Flags [S], seq 2467024935, win 65280, options [mss 1360,sackOK,TS val 3698885092 ecr 0,nop,wscale 7], length 0
17:38:36.247574 eth0  Out IP 172.18.0.3.3131 > 60.128.0.2.9000: Flags [S], seq 2467024935, win 65280, options [mss 1360,sackOK,TS val 3698885092 ecr 0,nop,wscale 7], length 0
17:38:36.595514 eth0  In  IP 60.128.0.2.9000 > 172.18.0.3.3131: Flags [R.], seq 0, ack 2467024936, win 0, length 0
17:38:36.596407 c1d27cde83a304d Out IP 60.128.0.2.9000 > 10.244.1.17.3131: Flags [R.], seq 0, ack 2467024936, win 0, length 0
```
The same happens when client and server are on the same node.

Could you please take a brief look?  Did I miss something?

Thanks!

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1936,6 +1958,45 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
...
+							config.Default.ConntrackZone, netConfig.masqCTMark)
+
+						if config.Gateway.Mode == config.GatewayModeLocal {
+							newFlow = fmt.Sprintf("cookie=%s, priority=102, in_port=LOCAL, dl_src=%s, ipv6, ipv6_dst=%s, "+
+								"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+								defaultOpenFlowCookie, bridgeMacAddress, hostSubnet,
+								config.Default.ConntrackZone, ctMarkHost)
+						}
+
+						dftFlows = append(dftFlows, newFlow)
```
qinqon, line 1981: ditto for ipv6

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1936,6 +1958,45 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
...
+
+						if config.Gateway.Mode == config.GatewayModeLocal {
+							newFlow = fmt.Sprintf("cookie=%s, priority=102, in_port=LOCAL, dl_src=%s, icmp6, icmpv6_type=%d, "+
+								"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+								defaultOpenFlowCookie, bridgeMacAddress, icmpType,
+								config.Default.ConntrackZone, ctMarkHost)
+						}
+
+						dftFlows = append(dftFlows, newFlow)
+					}
```
qinqon, line 1998: Don't we need similar stuff for ipv4 ICMP ? like we ping from default pod network to localnet same node ? 

Also add a comment here similar to the ones befefore and the shared + gateway if statements.

go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -139,6 +139,22 @@ func (c *addressManager) ListAddresses() []net.IP {
...
+	out := make([]*net.IPNet, 0, len(addrs))
+	for _, addr := range addrs {
+		_, cidr, err := net.ParseCIDR(addr)
+		if err != nil {
+			klog.Errorf("Failed to parse %s: %v", addr, err)
+			continue
+		}
+		out = append(out, cidr)
+	}
+	return out
```
qinqon, line 155: I understand that this is whow it's done at `ListAddresses` but this should be done when c.cidr is set and store at the addresssManager as `set.Set<*net.IPNet>` or `[]*net.IPNet`, it do not make sense to do this list to set + parse at every retrievel.

Also we should use here a read/write lock instead of a normal lock.

Just ranting, up to you for a commit refactoring this.

test/e2e/multihoming.go
```diff
@@ -1812,3 +1930,75 @@ func createMultiNetworkPolicy(mnpClient mnpclient.K8sCniCncfIoV1beta1Interface,
 	)
 	return err
 }
+
+func computeIPWithOffset(baseAddr string, increment int) (string, error) {
```
qinqon, line 1934: We don't need this function since there is a `generator/ip/ip_generator.go` package for this

test/e2e/multihoming.go
```diff
@@ -254,12 +260,123 @@ var _ = Describe("Multi Homing", func() {
...
+				serverIPs, err := podIPsForAttachment(cs, f.Namespace.Name, serverPod.GetName(), netConfig.name)
+				Expect(err).NotTo(HaveOccurred())
+				for _, serverIP := range serverIPs {
+					By(fmt.Sprintf("asserting the *client* can contact the server pod exposed endpoint: %q", serverIP))
+					Eventually(func() error {
+						return reachServerPodFromClient(cs, serverPodConfig, clientPodConfig, serverIP, port)
+					}, 2*time.Minute, 6*time.Second).Should(Succeed())
+				}
+			},
+			ginkgo.Entry(
```
qinqon, line 332: I don't see this testing ipv4,ipv6 ICMP (ping and RS/RA)

qinqon (changes requested):

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1845,6 +1845,28 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
...
+							"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, hostSubnet,
+							config.Default.ConntrackZone, netConfig.masqCTMark)
+
+						if config.Gateway.Mode == config.GatewayModeLocal {
+							newFlow = fmt.Sprintf("cookie=%s, priority=102, in_port=LOCAL, dl_src=%s, ip, nw_dst=%s, "+
+								"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+								defaultOpenFlowCookie, bridgeMacAddress, hostSubnet,
+								config.Default.ConntrackZone, ctMarkHost)
+						}
```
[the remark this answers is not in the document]
ricky-rav: True, it would also make the differences between the two gw modes easier to spot.

test/e2e/multihoming.go
ricky-rav: ops I had no idea

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: This needs more explanation, I forgot to add enough comments in the code, sorry.
So for ICMP (v4) we don't need any special flows: it's on top of IPv4, which we allow explicitly with the `ip` filter in the flow. 

Same goes for ICMPv6, as long as the destination involved is within the host subnet, of course.
However neighbor discovery in IPv6 happens through ICMPv6 messages to a special destination (`ff02::1:ff00:0/104`). The flows I'm adding in this PR match against the host subnet, so all neighbor discovery messages in V6 didn't match against the new flows and were getting redirected to the host NIC, thanks to the existing flow at prio=100.  I discussed this with @trozet and we decided it'd be a better solution to just match against the icmp type and the in-port here.

go-controller/pkg/node/node_ip_handler_linux.go
ricky-rav: aah I see what you mean. I admit it's not the most elegant solution, but I could only find uglier ways :| 
Since there are customers waiting for this fix, maybe we can postpone this to my followup PR (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5093) ?

test/e2e/localnet-underlay.go
```diff
@@ -13,41 +13,56 @@ import (
...
+		defaultNetworkBridgeMapping(),
+		bridgeMapping(networkName, bridgeName),
+	)
+}
+
+func setupUnderlayOnSecondaryBridge(ovsPods []v1.Pod, portName string, nadConfig networkAttachmentConfig) error {
 	for _, ovsPod := range ovsPods {
-		if err := addOVSBridge(ovsPod.Name, bridgeName); err != nil {
+
+		if err := addOVSBridge(ovsPod.Name, secondaryBridge); err != nil {
```
maiqueb, line 33: IMHO, you should receive the name of the bridge as the func's input parameter. Please name it `bridgeName`.

test/e2e/localnet-underlay.go
```diff
@@ -13,41 +13,56 @@ import (
...
-	add        = "add-br"
-	del        = "del-br"
+	defaultOvsBridge = "breth0"
+	secondaryBridge  = "ovsbr1"
+	add              = "add-br"
+	del              = "del-br"
 )
 
-func setupUnderlay(ovsPods []v1.Pod, portName string, nadConfig networkAttachmentConfig) error {
+func configureBridgeMappingsForNode(ovsPod v1.Pod, networkName, bridgeName string) error {
```
maiqueb, line 22: I think it weird that a function named `configureBridgeMappingsForNode` does not receive the node name as a parameter. 

And I also find it weird it simply wraps another function call, omitting the fact it configures the cluster default network mapping, *plus* the mapping you want to have configured.

test/e2e/multihoming.go
```diff
@@ -254,12 +261,129 @@ var _ = Describe("Multi Homing", func() {
...
+				}
+				if serverPodConfig.namespace == "" {
+					serverPodConfig.namespace = f.Namespace.Name
+				}
+
+				By("setting up the localnet underlay")
+				pods := ovsPods(cs)
+				Expect(pods).NotTo(BeEmpty())
+
+				Expect(setupUnderlayOnDefaultBridge(pods, netConfig)).To(Succeed())
```
maiqueb, line 294: where do you tear this mapping down ? IIUC, it is being left configured.

test/e2e/multihoming.go
```diff
@@ -1812,3 +1937,72 @@ func createMultiNetworkPolicy(mnpClient mnpclient.K8sCniCncfIoV1beta1Interface,
...
+			return "", fmt.Errorf("overflow: IP address exceeds bounds")
+		}
+	}
+
+	return netip.PrefixFrom(ip, addr.Bits()).String(), nil
+}
+
+// Given a node name and an offset, generateIPsFromNodePrimaryIfAddr returns an IPv4 and an IPv6 address
+// at the provided offset from the primary interface addresses found on the node.
+func generateIPsFromNodePrimaryIfAddr2(cs clientset.Interface, nodeName string, offset int) ([]string, error) {
```
maiqueb, line 1961: why is this function name suffixed with a 2 ? Am I missing something ?

test/e2e/multihoming_utils.go
```diff
@@ -130,14 +130,15 @@ func generateNetAttachDef(namespace, nadName, nadSpec string) *nadapi.NetworkAtt
...
-	requiresExtraNamespace bool
+	attachments                  []nadapi.NetworkSelectionElement
+	containerCmd                 []string
+	name                         string
+	namespace                    string
+	nodeSelector                 map[string]string
+	isPrivileged                 bool
+	labels                       map[string]string
+	requiresExtraNamespace       bool
+	needsIPRequestFromHostSubnet bool
```
maiqueb, line 141: isn't there a better way to model this without adding yet another boolean ? 

Can't you define here a static ip which you'd pass to the factory func ?

maiqueb (changes requested):
Some questions.

test/e2e/multihoming.go
ricky-rav: Sure! Let me fix that also for the case of secondary bridge case... I've just noticed that removing the bridge is not enough, we should explicitly set the desired bridge mapping.

test/e2e/multihoming.go
ricky-rav: My bad.

test/e2e/multihoming_utils.go
ricky-rav: Ah I went for a boolean because the logic to pick an IP within the host subnets, which I implemented in `addIPRequestToPodConfig`, can return an error. If I call that when I create the ginkgo.Entry, I'll get a `No Assertions in Container Nodes`.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1845,6 +1845,30 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
 							"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:%s",
 							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
 							netConfig.masqCTMark, ofPortPhys))
+
+					// IP traffic from the OVN network to the host network should be handled normally by the bridge instead of
+					// being forwarded directly to the NIC (prio=100 flow above). This allows pods in the default network to reach
+					// localnet pods on the same node.
+					if config.Gateway.Mode == config.GatewayModeShared || config.Gateway.Mode == config.GatewayModeLocal {
```
qinqon, line 1852: Let's just check if gateway is not disabled here with `config.Gateway.Mode != config.GatewayModeDisabled`
Then we do specific things inside.

But maybe we are planning on having a more gateway modes (BGP?)

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1936,6 +1960,52 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
...
+							}
+							inPort := ""
+							ctMark := ""
+							if config.Gateway.Mode == config.GatewayModeShared {
+								inPort = netConfig.ofPortPatch
+								ctMark = netConfig.masqCTMark
+							} else if config.Gateway.Mode == config.GatewayModeLocal {
+								inPort = "LOCAL"
+								ctMark = ctMarkHost
+							}
```
qinqon, line 1980: We are using this at ipv4 too, let's encpasulate on function or create a map to define inport and ctmark per gateway mode (also I think this kind of mapping is useful at other part of the code here).

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1936,6 +1960,52 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
...
+						for _, icmpType := range []int{types.NeighborSolicitationICMPType, types.NeighborAdvertisementICMPType} {
+							inPort := ""
+							ctMark := ""
+							if config.Gateway.Mode == config.GatewayModeShared {
+								inPort = netConfig.ofPortPatch
+								ctMark = netConfig.masqCTMark
+							} else if config.Gateway.Mode == config.GatewayModeLocal {
+								inPort = "LOCAL"
+								ctMark = ctMarkHost
+							}
```
qinqon, line 2000: ditto, rule of three applies.

go-controller/pkg/node/node_ip_handler_linux.go
qinqon: I am ok with follow up, thanks!

qinqon (changes requested):

test/e2e/multihoming.go
ricky-rav: In the end I imported the package you suggested, but in general we should avoid as much as possible to import code from ovn-k, since that's what we're testing here :)

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: Not sure, but I prefer to be explicit on what's allowed :)

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: yes, I captured that in a separate function in the end, as suggested by Miguel

ricky-rav:
I've addressed comments, but I broke something... I'll go back to it this afternoon.

go-controller/pkg/node/gateway_shared_intf.go
qinqon: fair enough

qinqon (approved):
LGTM

go-controller/pkg/node/node_ip_handler_linux.go
maiqueb: I'm unsure here, but I'll defer to a maintainer.

test/e2e/multihoming_utils.go
maiqueb: yeah, for that to work, you'd need to pass the function as argument, then execute it in the table's body.

I guess this can be refactored later on ... But the more and more booleans we add here, the harder it'll be. 

Another question: why don't you just hard code an IP ? i.e. just choose something from the range.

test/e2e/multihoming.go
maiqueb: We're now tearing down the underlay using a `defer`ed call.

test/e2e/localnet-underlay.go
maiqueb: tks, I'm resolving this, since it's now IMHO simpler.

maiqueb (approved):
There's a broken network segmentation lane: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13861148464/job/38791349932?pr=5041

Just asking you to ensure this isn't related to the PR. 

Other than that, LGTM.

ricky-rav:
> There's a broken network segmentation lane: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13861148464/job/38791349932?pr=5041
> 
> Just asking you to ensure this isn't related to the PR.

yeah, it's a known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4936

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2114,6 +2134,62 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
...
+			dstIP,
+			config.Default.ConntrackZone,
+			ctMark))
+	if isIPv6 {
+		// Neighbor discovery in IPv6 happens through ICMPv6 messages to a special destination (ff02::1:ff00:0/104),
+		// which has nothing to do with the host subnets we're matching against in the flow above at prio=102.
+		// Let's allow neighbor discovery by matching against icmp type and in_port.
+		for _, icmpType := range []int{types.NeighborSolicitationICMPType, types.NeighborAdvertisementICMPType} {
+			flows = append(flows,
+				fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, icmp6, icmpv6_type=%d, "+
```
trozet, line 2183: so this function is going to be called on potentially multiple IP addresses in a for loop, but this particular flow is agnostic to the IP, so the same flow will be appended multiple times. Maybe we should move this out of this function and put it into the commonFlows itself under the ipv6 part.

go-controller/pkg/node/node_ip_handler_linux.go
trozet: I don't think we need a RWMutex here yet. However, I think maybe an easy optimization is that ListNetworkAddresses seems to be called everywhere ListAddresses is called. In ListAddresses we do the same parsing, why not just make ListAddresses return ([]net.IP, []*net.IPNet)? That will make this PR not incur any perf regression.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1762,7 +1762,7 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 	return dftFlows, nil
 }
 
-func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
+func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]string, error) {
```
trozet, line 1765: hostSubnets seems like a misnomer here, this is actually hostCIDRs? A real IP/prefix, not a subnet like 10.244.0.0/24?

go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -139,6 +139,22 @@ func (c *addressManager) ListAddresses() []net.IP {
 	return out
 }
 
+func (c *addressManager) ListNetworkAddresses() []*net.IPNet {
```
trozet, line 142: does this include the IP of a localnet pod? we store those in addressManager as host IPs??

trozet (changes requested):

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: uhm no no, here we're really looking for the `network address /  prefix`. For instance in a dual stack cluster, I get: `172.18.0.0/16 fc00:f853:ccd:e793::/64`. Is `subnet` correct here?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2114,6 +2134,62 @@ func commonFlows(bridge *bridgeConfiguration) ([]string, error) {
 	return dftFlows, nil
 }
 
+// ovnToHostNetworkNormalActionFlows returns the flows that allow IP{v4,v6} traffic from the OVN network to the host network
+// when the destination is on the same node as the sender. This is necessary for pods in the default network to reach
+// localnet pods on the same node.
+func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC, dstIP string) []string {
```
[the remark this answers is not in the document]
ricky-rav: Sure! It's actually the network address (e.g.  172.18.0.0/16) of the IPs configured on the node primary interface

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: Ops. Good catch. I've reworked the function a little bit to pass the host cidrs and do everything inside the function instead of splitting it across different spots in the source code.

go-controller/pkg/node/node_ip_handler_linux.go
ricky-rav: uhm no, it's orthogonal to what we do for localnet.

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: Let's decide what terms to use here for `hostSubnets` above, because I'm not consistent across the code. The options I see are:
- `nodeSubnet`: I have a feeling this is used for the pod subnet
- `nodeCIDR`: not sure
- `hostSubnet`
- `hostCIDR` 

WDYT?

trozet: if you are getting 172.18.0.0/16, then the term hostSubnet is fine. But across your commit here I think the consistency breaks. Let me walk through it again now that I know this is a subnet.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -217,19 +217,21 @@ func (c *openflowManager) Run(stopChan <-chan struct{}, doneWg *sync.WaitGroup)
 
 // updateBridgeFlowCache generates the "static" per-bridge flows
 // note: this is shared between shared and local gateway modes
-func (c *openflowManager) updateBridgeFlowCache(extraIPs []net.IP) error {
+func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []*net.IPNet) error {
+	klog.Infof("RICCARDO Host IPs: %v, Host Networks: %v", hostIPs, hostSubnets)
```
trozet, line 221: remove

go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -123,18 +123,36 @@ func (c *addressManager) delAddr(ipnet net.IPNet, linkIndex int) bool {
...
 			continue
 		}
-		out = append(out, ip)
+		addresses = append(addresses, ip)
+		networkAddresses = append(networkAddresses, networkAddress)
+	}
+	return addresses, networkAddresses
+}
+
+func (c *addressManager) ListNetworkAddresses() []*net.IPNet {
```
trozet, line 144: we dont need this function anymore right?

trozet (changes requested):

go-controller/pkg/node/gateway_shared_intf.go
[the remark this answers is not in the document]
trozet: I see you changed this to hostSubnets, so that is good now.

go-controller/pkg/node/gateway_shared_intf.go
trozet: looks like your variable names are consistent now. I think it looks good.

trozet (approved):

PR #5042, opened 2025-02-07
temp

tssurya:
can't repro locally, let's try here


tssurya:
passed: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13206257919/job/36869981559?pr=5042

PR #5046, opened 2025-02-10
Perf improvement: egress ip/qos reroute rules

trozet:
These things were constantly being executed on every node update. Bring them inline with what we do for other node syncs, keep a cache of what needs to be sync'ed. The impact of this was especially bad for UDN.

image

### DIFF
--- a/go-controller/pkg/ovn/default_network_controller.go
+++ b/go-controller/pkg/ovn/default_network_controller.go
@@ -131,6 +131,7 @@ type DefaultNetworkController struct {
 	hybridOverlayFailed         sync.Map
 	syncZoneICFailed            sync.Map
 	syncHostNetAddrSetFailed    sync.Map
+	syncEIPNodeRerouteFailed    sync.Map
 
 	// variable to determine if all pods present on the node during startup have been processed
 	// updated atomically
@@ -843,14 +844,20 @@ func (h *defaultNetworkControllerEventHandler) AddResource(obj interface{}, from
 				_, hoSync := h.oc.hybridOverlayFailed.Load(node.Name)
 				_, zoneICSync := h.oc.syncZoneICFailed.Load(node.Name)
 				nodeParams = &nodeSyncs{
-					nodeSync,
-					clusterRtrSync,
-					mgmtSync,
-					gwSync,
-					hoSync,
-					zoneICSync}
+					syncNode:              nodeSync,
+					syncClusterRouterPort: clusterRtrSync,
+					syncMgmtPort:          mgmtSync,
+					syncGw:                gwSync,
+					syncHo:                hoSync,
+					syncZoneIC:            zoneICSync}
 			} else {
-				nodeParams = &nodeSyncs{true, true, true, true, config.HybridOverlay.Enabled, config.OVNKubernetesFeature.EnableInterconnect}
+				nodeParams = &nodeSyncs{
+					syncNode:              true,
+					syncClusterRouterPort: true,
+					syncMgmtPort:          true,
+					syncGw:                true,
+					syncHo:                config.HybridOverlay.Enabled,
+					syncZoneIC:            config.OVNKubernetesFeature.EnableInterconnect}
 			}
 
 			if err = h.oc.addUpdateLocalNodeEvent(node, nodeParams); err != nil {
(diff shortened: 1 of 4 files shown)


go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -591,12 +610,20 @@ func (oc *SecondaryLayer2NetworkController) addUpdateLocalNodeEvent(node *corev1
 			}
 		}
 
-		if config.OVNKubernetesFeature.EnableEgressIP {
+		if config.OVNKubernetesFeature.EnableEgressIP && nSyncs.syncReroute {
```
tssurya, line 613: not entirely related to this change:
I find it confusing as to how come for default net controller we don't use this syncReroute bool flag but we seem to use them in this fashion for 2ndary controllers... but even more so from the `NodeType` handler? 
I can see answer to this is because we seem to be plumbing EIP Node related things from Node handler itself? -> but given the event driven model here.. generally speaking in DNC we prioritize the nodeHandler to execute the nodeAdd successfully before EgressNodeHandler which has handler priority 2 IIRC so for example in this case if a node add failed above or say management port config setup failed for the node I don't see how we should add it for egressIP as a good candidate...
i guess I am probably overthinking this given this UDN code is anyways doing very less work ?

Is UDN EIP using DNC's EgressNodeType `addEgressNode` to do the garp on GR? so half from nodetype and half from DNC's nodetype?

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -182,12 +185,15 @@ func (h *secondaryLayer3NetworkControllerEventHandler) UpdateResource(oldObj, ne
 					nodeSubnetChanged ||
 					hostCIDRsChanged(oldNode, newNode) ||
 					nodeGatewayMTUSupportChanged(oldNode, newNode)
+				_, failed = h.oc.syncRerouteFailed.Load(newNode.Name)
+				syncReroute := failed || util.NodeHostCIDRsAnnotationChanged(oldNode, newNode)
```
tssurya, line 189: this remote->local zone change logic you had for DNC is not needed here?

tssurya (commented):
a few questions otherwise pretty easy must have change 
++ nice improvement

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -1066,24 +1090,24 @@ func (h *defaultNetworkControllerEventHandler) UpdateResource(oldObj, newObj int
...
-		if util.NodeHostCIDRsAnnotationChanged(oldNode, newNode) {
+		// update the nodeIP in the default-reRoute (102 priority) destination address-set
+		if failed || util.NodeHostCIDRsAnnotationChanged(oldNode, newNode) {
 			klog.Infof("Egress IP detected IP address change for node %s. Updating no re-route policies", newNode.Name)
 			err := h.oc.eIPC.ensureDefaultNoRerouteNodePolicies()
 			if err != nil {
 				return err
 			}
 		}
+		h.oc.syncRerouteFailed.Delete(newNode.Name)
```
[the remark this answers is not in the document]
trozet: hmm that's a good point. It's not specific to this PR, but yeah if the node is deleted the cache goes stale cause we dont clean it up. If a new node is added later, we dont use fromRetry, and i guess however far it makes it it will clean up the cache and delete those entries. If it fails it will store them to true anyway. So it seems like it will overwrite the cache correctly.

I guess the real downside here would be if nodes would come and go often, then we would have stale data in our cache and it would become a memory problem, but since that doesn't really happen with nodes we haven't noticed it.

To be technically correct we probably should delete the entries out of the cache when a node gets deleted. I can post another commit to do that.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
trozet: this is probably more of a question for @martinkennelly who implemented EgressIP with UDN. The UDN controllers dont seem to trigger on EgressNode types and handle the reroute config inside of Node type. Either way the same config is done on DNC and UDN ovn instances.

go-controller/pkg/ovn/secondary_layer3_network_controller.go
trozet: yeah that was intentional. I didn't think we support (or want to) migrating from non-IC to IC with UDNs enabled.

go-controller/pkg/ovn/default_network_controller.go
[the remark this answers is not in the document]
tssurya: @trozet ack new commit makes it easier for review deff, ping me back after you repush and I can approve this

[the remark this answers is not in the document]
trozet: so I realized the sync maps are deleted on node delete. They are just not obvious because it happens inside of the delete functions themselves. Updating my initial commit to do the same thing for the new reroute map.

tssurya (approved):

tssurya:
known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13379228903/job/37365696474?pr=5046 there is already an issue for that.

PR #5048, opened 2025-02-11
Use CNCF hosted DCO check

jcaamano:
The probot DCO app is unmaintained and does not support merge queue.


jcaamano:
https://github.com/cncf/sandbox/issues/302#issuecomment-2650918282

PR #5049, opened 2025-02-11
Make retest command work with multiple workflows

jcaamano:
Hopefully fixes the problems that we were having with it, or otherwise provides more debuggability.

### DIFF
--- a/.github/actions/retest-action/entrypoint.sh
+++ b/.github/actions/retest-action/entrypoint.sh
@@ -6,12 +6,12 @@ set -ex
 # Prerequisites check
 ##############################
 
-if ! jq -e '.issue.pull_request' ${GITHUB_EVENT_PATH}; then
+if ! jq -e '.issue.pull_request' "${GITHUB_EVENT_PATH}"; then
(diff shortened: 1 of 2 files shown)


.github/actions/retest-action/entrypoint.sh
```diff
@@ -52,58 +54,81 @@ send_comment() {
...
+    RESPONSE_MESSAGE=$(jq -r '.message' .action-response.json)
+    send_comment "Oops, something went wrong when triggering workflow run\n${url}\n~~~\n${RESPONSE_MESSAGE}\n~~~\n"
+    break
+  fi
+  touch triggered.data
+  echo "$url" | sed -e 's|/api.github.com/repos/|/github.com/|' -e 's|/[^/]*$||' >> triggered.data
+  rm .action-response.json
+done
+
+cat triggered.data
```
npinaeva, line 123: I think when `url.data` is empty, `triggered.data` file doesn't exist, and since we have `set -ex` in the beginning of the script, it just exists right after
`cat: can't open 'triggered.data': No such file or directory` and no message is sent?

PR #5051, opened 2025-02-11
Adds logging for informer waitForCacheSync

trozet:
The logging will help us determine how long it took to process initial items from the informer into the informer cache.

Note, qos was duplicated as syncing twice in the factory, removed the duplicate.

## Summary by CodeRabbit

* **Refactor**
  * Unified startup synchronization flow with consolidated error propagation and timing for overall startup.
  * Added per-component success logging to standardize observability of cache synchronization.

* **Bug Fixes**
  * Improved detection and reporting of cache-sync failures during startup to reduce silent failures.
  * Enhanced, descriptive error messages and startup timing information to speed troubleshooting when background components fail to initialize.

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -527,153 +527,113 @@ func NewOVNKubeControllerWatchFactory(ovnClientset *util.OVNKubeControllerClient
 
 // Start starts the factory and begins processing events
 func (wf *WatchFactory) Start() error {
+	start := time.Now()
 	klog.Info("Starting watch factory")
 	wf.iFactory.Start(wf.stopChan)
-	for oType, synced := range waitForCacheSyncWithTimeout(wf.iFactory, wf.stopChan) {
-		if !synced {
-			return fmt.Errorf("error in syncing cache for %v informer", oType)
-		}
+	if err := waitForCacheSyncWithTimeout(wf.iFactory, wf.stopChan); err != nil {
+		return err
 	}
 	if config.OVNKubernetesFeature.EnableAdminNetworkPolicy && wf.anpFactory != nil {
 		wf.anpFactory.Start(wf.stopChan)
-		for oType, synced := range waitForCacheSyncWithTimeout(wf.anpFactory, wf.stopChan) {
-			if !synced {
-				return fmt.Errorf("error in syncing cache for %v informer", oType)
-			}
+		if err := waitForCacheSyncWithTimeout(wf.anpFactory, wf.stopChan); err != nil {
+			return err
 		}
 	}
 	if config.OVNKubernetesFeature.EnableEgressIP && wf.eipFactory != nil {
 		wf.eipFactory.Start(wf.stopChan)
-		for oType, synced := range waitForCacheSyncWithTimeout(wf.eipFactory, wf.stopChan) {
-			if !synced {
-				return fmt.Errorf("error in syncing cache for %v informer", oType)
-			}
+		if err := waitForCacheSyncWithTimeout(wf.eipFactory, wf.stopChan); err != nil {
+			return err
 		}
 	}
 	if config.OVNKubernetesFeature.EnableEgressFirewall && wf.efFactory != nil {
 		wf.efFactory.Start(wf.stopChan)
-		for oType, synced := range waitForCacheSyncWithTimeout(wf.efFactory, wf.stopChan) {
-			if !synced {
-				return fmt.Errorf("error in syncing cache for %v informer", oType)
-			}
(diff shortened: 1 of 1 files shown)


go-controller/pkg/factory/factory.go
```diff
@@ -527,153 +527,113 @@ func NewOVNKubeControllerWatchFactory(ovnClientset *util.OVNKubeControllerClient
...
-		for oType, synced := range waitForCacheSyncWithTimeout(wf.networkQoSFactory, wf.stopChan) {
-			if !synced {
-				return fmt.Errorf("error in syncing cache for %v informer", oType)
-			}
+		if err := waitForCacheSyncWithTimeout(wf.frrFactory, wf.stopChan); err != nil {
+			return err
 		}
 	}
-
+	klog.Infof("Watch Factory start up complete, took: %s", time.Since(start))
```
npinaeva, line 636: not sure how you want to use it, but it could be also useful as a metric

go-controller/pkg/factory/factory.go
```diff
@@ -1919,10 +1879,18 @@ type waitForCacheSyncer interface {
...
+func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) error {
+	// Give some small amount of time for sync. It helps significantly reduce unit tests time.
 	time.Sleep(5 * time.Millisecond)
-	return factory.WaitForCacheSync(util.GetChildStopChanWithTimeout(stopCh, types.InformerSyncTimeout))
+
+	for oType, synced := range factory.WaitForCacheSync(util.GetChildStopChanWithTimeout(stopCh, types.InformerSyncTimeout)) {
+		if !synced {
+			return fmt.Errorf("error in syncing cache for %v informer", oType)
+		} else {
+			klog.Infof("%s informer cache synced successfully", oType)
```
tssurya, line 1890: I assume this is intentionally to also know when the sync was successful for each resource?
This might spam a bit during startup but should be a one time thing

go-controller/pkg/factory/factory.go
```diff
@@ -1919,10 +1879,18 @@ type waitForCacheSyncer interface {
...
-	// Give some small time for sync. It helps significantly reduce unit tests time
+func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) error {
+	// Give some small amount of time for sync. It helps significantly reduce unit tests time.
 	time.Sleep(5 * time.Millisecond)
-	return factory.WaitForCacheSync(util.GetChildStopChanWithTimeout(stopCh, types.InformerSyncTimeout))
+
+	for oType, synced := range factory.WaitForCacheSync(util.GetChildStopChanWithTimeout(stopCh, types.InformerSyncTimeout)) {
+		if !synced {
+			return fmt.Errorf("error in syncing cache for %v informer", oType)
+		} else {
```
tssurya, line 1889: nit: I don't think you need else here since you are returning in L1888 so for all other cases you want to log - will make it look cleaner

go-controller/pkg/factory/factory.go
```diff
@@ -1919,10 +1879,18 @@ type waitForCacheSyncer interface {
 	WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool
 }
 
-func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) map[reflect.Type]bool {
-	// Give some small time for sync. It helps significantly reduce unit tests time
+func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struct{}) error {
+	// Give some small amount of time for sync. It helps significantly reduce unit tests time.
```
tssurya, line 1883: oh cause other wise unit tests keep retrying for the object?

tssurya (approved):
one small nit if you want to fix it

go-controller/pkg/factory/factory.go
trozet: right, iirc we had a bug where the informer cache syncs were taking a long time and we couldn't really tell which one it was. This will tell us how long each one is taking so we can root cause it easier.

go-controller/pkg/factory/factory.go
trozet: yeah, maybe in the future if we need it

go-controller/pkg/factory/factory.go
trozet: i'm not sure, i didn't add that bit.

go-controller/pkg/factory/factory.go
trozet: I agree with you, but i think it's OK and not worth pushing another iteration for.

PR #5057, opened 2025-02-12
Fix CM-EIP controller UT flake

### LINKED ISSUE #4968: OVN cluster-manager EgressIP Operations WatchEgressNodes running with WatchEgressIP [It] should result in error and event if specified egress IP is a cluster node IP
### Which jobs are flaking?

OVN cluster-manager EgressIP Operations WatchEgressNodes running with WatchEgressIP [It] should result in error and event if specified egress IP is a cluster node IP

### Which tests are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/12805667661/job/35702528425?pr=4933

16-Jan-2025

```
[FAILED] Timed out after 1.001s.
  Expected
      <chan string | len:4, cap:10>: 0xc0006a8fc0
  to have length 3
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/clustermanager/egressip_controller_test.go:1259 @ 01/16/25 09:21:05.385
```

_No response_

tssurya:
During UDN we added syncEgressIPMarkAllocator
which is called when WatchEIP is started as a global syncFunc
that causes an extra patch operation that
causes an extra event. So update the test to
reflect that.

Fixes #4968 

@martinkennelly PTAL

### DIFF
--- a/go-controller/pkg/clustermanager/egressip_controller_test.go
+++ b/go-controller/pkg/clustermanager/egressip_controller_test.go
@@ -1256,7 +1256,39 @@ var _ = ginkgo.Describe("OVN cluster-manager EgressIP Operations", func() {
 				gomega.Expect(fakeClusterManagerOVN.eIPC.nodeAllocator.cache).To(gomega.HaveKey(node2.Name))
 
 				gomega.Eventually(getEgressIPStatusLen(egressIPName)).Should(gomega.Equal(0))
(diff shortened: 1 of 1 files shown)


martinkennelly (approved):
makes sense - i didnt consider the sync func event :) thanks!

tssurya:
this PR only changes UTs.. the E2E flakes should have no impact here.

PR #5060, opened 2025-02-13
Fix updateNetworkStatus

trozet:
When a node would no longer have an error, the code would search for the next node that has an error and report that. However, it would not update the error message and report the error message for the old node. It was also not detecting the next error node properly, by not checking if its error message was empty or not.

### DIFF
--- a/go-controller/pkg/clustermanager/network_cluster_controller.go
+++ b/go-controller/pkg/clustermanager/network_cluster_controller.go
@@ -278,17 +278,21 @@ func (ncc *networkClusterController) updateNetworkStatus(nodeName string, handle
 
 	ncc.nodeErrorsLock.Lock()
 	defer ncc.nodeErrorsLock.Unlock()
+
 	if ncc.nodeErrors[nodeName] == errorMsg {
 		// error message didn't change for that node, no need to update
 		return nil
 	}
 
+	// identify current error node or set the currently reported error node as this node
 	reportedErrorNode := ncc.reportedErrorNode
 	if ncc.reportedErrorNode == "" && errorMsg != "" {
 		reportedErrorNode = nodeName
 	}
+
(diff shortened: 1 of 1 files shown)


trozet:
@npinaeva @jcaamano PTAL

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -317,7 +324,12 @@ func (ncc *networkClusterController) updateNetworkStatus(nodeName string, handle
 	if err := ncc.statusReporter(netName, "NetworkClusterController", condition, events...); err != nil {
 		return fmt.Errorf("failed to report network status: %w", err)
 	}
-	ncc.nodeErrors[nodeName] = errorMsg
+
+	if nodeHasError {
```
npinaeva, line 328: I think this part is all you need to fix, but just make it
```
if errorMsg == "" {
    delete(ncc.nodeErrors, nodeName)
} else {
    ncc.nodeErrors[nodeName] = errorMsg
}
```
and then everything else should work? It was not expected that `nodeErrors` map has empty errors.

trozet: i dont think so because lets say we process node1, it no longer has an error, so now we search for the next error node and we find one. We don't use its error message on line 315-319, we use the original error message that was for node1

npinaeva: that is I think expected (maybe I should improve the function comment to make it more obvious)
so the idea is that UDN condition doesn't change until we pick a new `reportedErrorNode`, but we still report the event with the original error message for the original node (if error is not empty)

trozet: i see i'll fix it and add a comment

npinaeva (approved):
thank you for fixing this and for extra comments!

PR #5061, opened 2025-02-13
Make network controllers have infinite attempts

trozet:
Currently network controllers give up at 15 tries on an object. Since zone/node controllers wait on cluster manager to populate node information, and cluster manager may lag behind other controllers, it can result in some UDNs never getting rendered if a controller gives up. This leads to a state where the UDN never came up, and pods are infinitely retrying and failing to start.


trozet:
@jcaamano @tssurya PTAL

jcaamano (approved):

PR #5062, opened 2025-02-13
Add better logging for cluster manager network controller

trozet:
So we can actually see what happens when cluster manager starts a new network controller


trozet:
@jcaamano @tssurya PTAL

jcaamano (approved):
Too much logging for my taste. I would move most of these to `V(5)`, otherwise it might be difficult for inexperienced eyes to pick up on otherwise obvious errors. I have not better argument so I'll approve.

PR #5068, opened 2025-02-17
Removes checking for mac address changes, minor fixes to nodeSubnetChanged

trozet:
We were seeing this error in downstream a lot:

2025-02-14T00:49:29.505243189Z E0214 00:49:29.505221 2349982 ovn.go:423] Failed to unmarshal old node ip-10-0-50-90.us-west-2.compute.internal annotation: unexpected end of JSON input

This was coming from the macAddressChanged function. I realized there is no point in checking if the mac address changed, as we do not store it anymore on the node annotation. The only use for it is for reading it on the default network controller if it existed for legacy reasons.

Also, some minor fixes to nodeSubnetChanged.


jcaamano:
LGTM

jcaamano (approved):

PR #5070, opened 2025-02-17
Limit cluster manager node allocator updates/logs

trozet:
On every node update we were syncing the node in cluster manager. While there were checks in place to limit updating the node annotation, there were not checks in place to limit the other functionality (like marking subnets allocated). This code would execute everytime, which would spam the logs with messages like:

2025-02-14T01:25:53.598240753Z I0214 01:25:53.598230       1 node_allocator.go:510] Allowed existing subnets [10.132.5.0/24] on node ip-10-0-58-12.us-west-2.compute.internal
2025-02-14T01:25:53.598305025Z I0214 01:25:53.598279       1 node_allocator.go:510] Allowed existing subnets [10.132.8.0/24] on node ip-10-0-114-225.us-west-2.compute.internal
2025-02-14T01:25:53.598305025Z I0214 01:25:53.594125       1 node_allocator.go:488] Valid subnet 10.132.21.0/24 allocated on node ip-10-0-58-12.us-west-2.compute.internal
2025-02-14T01:25:53.598305025Z I0214 01:25:53.594137       1 node_allocator.go:488] Valid subnet 10.132.28.0/24 allocated on node ip-10-0-58-12.us-west-2.compute.internal
2025-02-14T01:25:53.598305025Z I0214 01:25:53.594143       1 node_allocator.go:488] Valid subnet 10.132.4.0/24 allocated on node ip-10-0-58-12.us-west-2.compute.internal
2025-02-14T01:25:53.598305025Z I0214 01:25:53.594148       1 node_allocator.go:488] Valid subnet 10.132.6.0/24 allocated on node ip-10-0-58-12.us-west-2.compute.internal

This floods the log. The "Valid subnet" just happens when the subnet is marked as allocated. It doesn't mean anything new was allocated. Removed this log. Allowed existing subnets message jsut means the existing subnets on the node were already allocated. These log messages also dont reference network name, so they are pretty useless. Logs remain which indicate if new subnets were allocated and for what network.

Additionally we dont need to run the update logic if the node was already sync'ed on node add. Once the node is allocated, nothing changes on the node that would force us to need to allocate again (other than a node going from hybrid overlay -> ovn). Added a sync map to track if a node needs to be updated again.

Did the same thing for ZCC (zone cluster controller):

trozet@fedora:~/go/src/github.com/ovn-org/ovn-kubernetes/go-controller$ go test -mod=vendor -v ./pkg/clustermanager -ginkgo.v -ginkgo.focus=".*Node subnet allocations.*Linux nodes$" | grep -i "setting annotations"
I0218 11:18:14.168191  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:3] on node node1
I0218 11:18:14.168187  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:2] on node node2
I0218 11:18:14.168200  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:4] on node node3
I0218 11:18:14.168964  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:3] on node node1
I0218 11:18:14.169120  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:4] on node node3
I0218 11:18:14.169152  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:2] on node node2
I0218 11:18:14.169395  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:3] on node node1
I0218 11:18:14.169430  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:2] on node node2
I0218 11:18:14.169492  310203 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:4] on node node3

After patch:
trozet@fedora:~/go/src/github.com/ovn-org/ovn-kubernetes/go-controller$ go test -mod=vendor -v ./pkg/clustermanager -ginkgo.v -ginkgo.focus=".*Node subnet allocations.*Linux nodes$" | grep -i "setting annotations"
I0218 13:03:29.871106  653449 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:2] on node node3
I0218 13:03:29.871132  653449 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:3] on node node1
I0218 13:03:29.871146  653449 kube.go:130] Setting annotations map[k8s.ovn.org/node-id:4] on node node2

Finally, simplified some of the logic in the sync node network annotations. No need to annotate the network id on the node unless it already existed and is somehow incorrect. Also only release the tunnel ID if it was allocated and failed to be annotated.


go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -557,19 +562,29 @@ func (h *networkClusterControllerEventHandler) UpdateResource(oldObj, newObj int
...
+		_, nodeFailed := h.nodeSyncFailed.Load(newNode.GetName())
+		if !nodeFailed && util.NoHostSubnet(oldNode) != util.NoHostSubnet(newNode) {
+			// no other node updates would require us to reconcile again
+			return nil
+		}
+		err = h.ncc.nodeAllocator.HandleAddUpdateNodeEvent(newNode)
 		if err == nil {
-			h.clearInitialNodeNetworkUnavailableCondition(node)
+			h.clearInitialNodeNetworkUnavailableCondition(newNode)
+			h.nodeSyncFailed.Delete(newNode.GetName())
```
npinaeva, line 585: don't you need to add the nodeto `nodeSyncFailed` if `err != nil`?

go-controller/pkg/clustermanager/zone_cluster_controller.go
```diff
@@ -131,6 +131,23 @@ func (zcc *zoneClusterController) Stop() {
...
+	}
+
+	if _, ok := node.Annotations[util.OvnNodeID]; !ok {
+		return true
+	}
+	if config.OVNKubernetesFeature.EnableInterconnect {
+		if _, ok := node.Annotations[util.OvnTransitSwitchPortAddr]; !ok {
+			return true
+		}
+	}
```
npinaeva, line 147: doesn't this part also mean that the node will be in `nodeSyncFailed` map?

go-controller/pkg/clustermanager/network_cluster_controller.go
```diff
@@ -571,7 +571,11 @@ func (h *networkClusterControllerEventHandler) UpdateResource(oldObj, newObj int
 			return fmt.Errorf("could not cast %T object to *corev1.Node", newObj)
 		}
 		_, nodeFailed := h.nodeSyncFailed.Load(newNode.GetName())
-		if !nodeFailed && util.NoHostSubnet(oldNode) != util.NoHostSubnet(newNode) {
+		// Note: (trozet) It might be pedantic to check if the NeedsNodeAllocation. This assumes one of the following:
+		// 1. we missed an add event (bug in kapi informer code)
+		// 2. a user removed the annotation on the node
```
npinaeva, line 576: current checks will only work if the annotation was removed, but not if it is incorrect (can happen on both missed kapi event after a restart and user changing annotation instead of deleting it).
not sure how much of a problem it solves for the price

test/e2e/network_segmentation.go
```diff
@@ -1057,20 +1057,21 @@ spec:
...
 				}
 				Expect(found).To(BeTrue(), "NetworkAllocationSucceeded condition should be False when cluster has > 2 nodes")
+				By("checking that a NetworkAllocationFailed error event was also reported")
 				events, err := cs.CoreV1().Events(f.Namespace.Name).List(context.Background(), metav1.ListOptions{})
 				Expect(err).NotTo(HaveOccurred())
 				found = false
 				for _, event := range events.Items {
 					if event.Reason == "NetworkAllocationFailed" && event.LastTimestamp.After(time.Now().Add(-30*time.Second)) &&
-						strings.Contains(event.Message, "error allocating network") {
+						strings.Contains(event.Message, "error allocating subnet") {
```
npinaeva, line 1074: this may need a change in origin? :)

npinaeva (commented):
I guess we need to decide whether we want to handle unexpected changes by user (admin?) for our annotations, which is a side-effect of this pr

trozet:
> I guess we need to decide whether we want to handle unexpected changes by user (admin?) for our annotations, which is a side-effect of this pr

I don't think that is a consistent behavior that we follow now for other annotations. So I don't think we need to start doing it now. I'm open to hearing other opinions though if someone thinks differently.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -406,10 +416,12 @@ func (na *NodeAllocator) updateNodeNetworkAnnotationsWithRetry(nodeName string,
 			return fmt.Errorf("failed to update node %q annotation LRPAddrAnnotation %s: %w",
 				node.Name, util.JoinIPNets(joinAddr, ","), err)
 		}
-		cnode.Annotations, err = util.UpdateNetworkIDAnnotation(cnode.Annotations, networkName, networkId)
-		if err != nil {
-			return fmt.Errorf("failed to update node %q network id annotation %d for network %s: %w",
-				node.Name, networkId, networkName, err)
+		if networkId != util.NoID {
```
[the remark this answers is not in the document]
trozet: I think it was just not used for anything other than tunnel ID until now, but it is as generic as the InvalidID constant, which is also used for networkID:

```
	// InvalidID signifies its an invalid network id or invalid tunnel id
	InvalidID = -1

	// NoID signifies its an empty tunnel id (its reserved as un-usable when the allocator is created)
	NoID = 0
```

So I think it's OK to use it for networkID, but as you mentioned in your other comment using a boolean is just as good (if not clearer). So will change it to that.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -301,14 +303,22 @@ func (na *NodeAllocator) syncNodeNetworkAnnotations(node *corev1.Node) error {
 		}
 	}
 
+	// only update node annotation with ID if it had it before and does not match
+	// NoID means it will not update
+	// InvalidID means node did not have the annotation
+	annoUpdateID := util.NoID
+	if networkID != util.InvalidID && networkID != na.networkID {
+		annoUpdateID = na.networkID
+	}
```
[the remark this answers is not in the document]
trozet: networkID was previously annotated on the node and then moved to be annotated on the NAD. I think the transition was done in stages to only use the one on the NAD. So to just preserve backwards compatibility we handle updating the node version. I guess I can just add a comment explaining this with a TODO in the future to remove updating it on the node altogether.

go-controller/pkg/clustermanager/node/node_allocator.go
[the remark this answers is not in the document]
trozet: Actually I think I should keep the ID usage, and that I just missed passing it to updateNodeNetworkAnnotationsWithRetry, which will then conditionalize whether or not it should annotate the value.

go-controller/pkg/clustermanager/zone_cluster_controller.go
trozet: that's checked where this function is called:
```
		if !nodeFailed && !needsZoneAllocation(node) {
			// node ID and transit switch IP are assigned by us and cannot change
			return nil
		}
```

or you mean there is no point in checking node.Annotations[util.OvnTransitSwitchPortAddr] because it would have failed and been on nodeSyncFailed? I guess either way it doesn't hurt to have the check here.

go-controller/pkg/clustermanager/network_cluster_controller.go
trozet: yeah I was a bit torn on this one, so I tried to compromise with a somewhat fast check and only solving if the annotations were removed. We don't support anything if a user changes the ovnk annotations. I guess we should also not support if they are removed, but I'm worried something may happen where we are essentially "stuck" and the node has no annotation, so if you are OK with it I'd prefer to have this code at least for now

test/e2e/network_segmentation.go
trozet: and thats enough for me to revert the message back tot he old one :)

trozet:
looks like there are more places in the code relying on the network id on the node :(

I'll have to fix these too...

go-controller/pkg/clustermanager/node/node_allocator.go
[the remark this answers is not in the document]
npinaeva: alright, new idea: what if instead of `annoUpdateID` (this name somehow kills my brain) you do this
```
        // Only update networkID annotation if the node had this annotation before and it doesn't match
	// Use NoID to indicate that the annotation update should be skipped
        // maybe some more comments here
	if networkID == util.InvalidID || networkID == na.networkID {
		networkID = util.NoID
	}
```
then replace `annoUpdateID` with `networkID`?

go-controller/pkg/clustermanager/zone_cluster_controller.go
npinaeva: yep, the latter
All conditions of this function are already covered by other checks (including HO check in the very beginning of `handleAddUpdateNodeEvent`), and I find it confusing that it is only called on update and not add events. 
It can be also hard to maintain, e.g. if we add some annotation in the future, I am not sure if we will remember to update this function as well (and there is no need as long as we keep `nodeSyncFailed`)

go-controller/pkg/util/multi_network.go
```diff
@@ -1111,18 +1112,35 @@ func GetAnnotatedNetworkName(netattachdef *nettypes.NetworkAttachmentDefinition)
...
+		}
+	}
+
+	n, err := newNetInfo(netconf)
+	if err != nil {
+		return nil, err
+	}
+	if id > NoID {
+		n.SetNetworkID(id)
+	}
```
npinaeva, line 1141: Do you think we should follow exactly the same algorithm as https://github.com/ovn-kubernetes/ovn-kubernetes/blob/ab2bbed40acccf88e56cf9bfa8ff1452acfa72fc/go-controller/pkg/networkmanager/nad_controller.go#L556-L574? or if you decided to make changes, why?

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -450,14 +450,6 @@ func (oc *SecondaryLayer3NetworkController) newRetryFramework(
 // Start starts the secondary layer3 controller, handles all events and creates all needed logical entities
 func (oc *SecondaryLayer3NetworkController) Start(_ context.Context) error {
 	klog.Infof("Start secondary %s network controller of network %s", oc.TopologyType(), oc.GetNetworkName())
-	_, err := oc.getNetworkID()
-	if err != nil {
-		return fmt.Errorf("unable to set networkID on secondary L3 controller for network %s, err: %w", oc.GetNetworkName(), err)
-	}
-	if err = oc.init(); err != nil {
```
trozet, line 457: oops I shouldn't have removed this

go-controller/pkg/util/multi_network.go
trozet: hmm good point. I didn't think about it that way. I was simply thinking "well if i find it, then i'll set it", but I need to consider what happens when I don't set it. In that case the struct gets default int value of 0 (NoID). Maybe that is OK, but it is inconsistent with the nad_controller code which represents an unset/missing value as -1 (InvalidID). To make it consistent I'll follow the same pattern you pasted.

trozet: Looks like making the change to default to invalidID breaks this unit test:

2025-03-05T22:02:02.4546773Z === RUN   TestNADController/NAD_on_default_network_is_tracked_with_default_controller
2025-03-05T22:02:02.4553734Z I0305 22:02:02.455183   38736 network_controller.go:246] [ network controller]: syncing all networks
2025-03-05T22:02:02.4555716Z I0305 22:02:02.455250   38736 network_controller.go:257] [ network controller]: finished syncing all networks. Time taken: 68.679µs
2025-03-05T22:02:02.4556992Z I0305 22:02:02.455299   38736 controller.go:156] Starting controller [ network controller] with 1 workers
2025-03-05T22:02:02.4563229Z I0305 22:02:02.456156   38736 nad_controller.go:363] : will wait for cluster manager to allocate an ID before ensuring network default
2025-03-05T22:02:03.4571914Z     nad_controller_test.go:596: 
2025-03-05T22:02:03.4572654Z         Timed out after 1.001s.
2025-03-05T22:02:03.4575017Z         The function passed to Eventually failed at /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/networkmanager/nad_controller_test.go:525 with:
2025-03-05T22:02:03.4576882Z         Expected
2025-03-05T22:02:03.4577718Z             <[]util.NetInfo | len:0, cap:0>: []
2025-03-05T22:02:03.4578437Z         to have length 1
2025-03-05T22:02:03.4578967Z === RUN   TestNADController/NAD_added


https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/nad_controller.go#L363

looks like this check needs to also consider if its default network as it states in the comments

trozet: and according to other code, the default network should use 0:

		// Reserve the ID of the default network
		err := c.networkIDAllocator.ReserveID(types.DefaultNetworkName, DefaultNetworkID)

DefaultNetworkID = 0

trozet: fixed it to handle default network differently, all other networks return InvalidID if not found in annotation

go-controller/pkg/clustermanager/node/node_allocator.go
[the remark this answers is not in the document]
npinaeva: now that I know `DefaultNetworkID == NoID`, I don't think we should use `NoID` for networks at all, it just makes it confusing (not to mention the fact, that DefaultNetworkID is also passed to this function, so defaultNetworkID will be treated as don't update)
I think you could reuse `InvalidID` here like so
```
        // Only update networkID annotation if the node had this annotation before (networkID == util.InvalidID) and it doesn't match
	// Use InvalidID to indicate that the annotation update should be skipped
        // maybe some more comments here
	if  networkID == na.networkID {
		networkID = util.InvalidID
	}
```
then update `updateNodeNetworkAnnotationsWithRetry` function to check for InvalidID on networkID instead of noID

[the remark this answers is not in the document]
trozet: earlier in the function: ParseNetworkIDAnnotation will return invalidID if the annotation wasn't found. If I just do logic you suggested we miss out on knowing if networkID was annotated on the node or not. So I think checking InvalidID still makes sense. Like:

```
	// only update node annotation with ID if it had it before (networkID at this point should not be InvalidID)
	// and does not match.
	// InvalidID means do not update the annotation
	if networkID != types.InvalidID && networkID != na.networkID {
		networkID = na.networkID
	}
```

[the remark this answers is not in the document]
trozet: unfortunately "then update updateNodeNetworkAnnotationsWithRetry function to check for InvalidID on networkID instead of noID" doesn't work, because InvalidID is passed for the Cleanup function, and it expects there that this value will signal to "remove the networkID from the annotation. Where as here we are trying to use the value to indicate "do not update this annotation". I think there is no getting around using an additional value here, but I will make it not share the same value as DefaultNetworkID.

go-controller/pkg/util/node_annotations.go
```diff
@@ -1330,12 +1356,15 @@ func parseNetworkMapAnnotation(nodeAnnotations map[string]string, annotationName
...
+		if IsAnnotationNotSetError(err) {
+			return types.NoNetworkID, err
+		}
+		return types.InvalidID, err
 	}
 
 	networkID, ok := networkIDsMap[netName]
 	if !ok {
-		return InvalidID, newAnnotationNotSetError("node %q has no %q annotation for network %s", node.Name, ovnNetworkIDs, netName)
+		return types.NoNetworkID, newAnnotationNotSetError("node %q has no %q annotation for network %s", node.Name, ovnNetworkIDs, netName)
```
npinaeva, line 1367: I find this function returning different errors confusing, not sure why anyone would want to distinguish empty annotation that will return InvalidID vs unset annotation that will return NoNetworkID. It also complicates error handling on the caller side (e.g. currently in `syncNodeNetworkAnnotations` you only check for NoNetworkID, but I think InvalidID should have the same effect ).

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -301,14 +340,21 @@ func (na *NodeAllocator) syncNodeNetworkAnnotations(node *corev1.Node) error {
 		}
 	}
 
+	// only update node annotation with ID if it had it before (networkID is not NoNetworkID)
+	// and does not match.
+	// NoNetworkID means do not update the annotation
+	if networkID != types.NoNetworkID && networkID != na.networkID {
+		networkID = na.networkID
+	}
```
npinaeva, line 348: I think you may need an
```
else {
    networkID = types.NoNetworkID
}
```
to skip update when `networkID == na.NetworkID`?

npinaeva (commented):
nice ID refactoring, makes more sense now!

npinaeva (approved):
what a nice cleanup!

Issue #5074, opened 2025-02-18
unit test deadlock: handles a OVN node is switched to a HO node


trozet:
### Which jobs are flaking?

unit test

### Which tests are flaking?

handles a OVN node is switched to a HO node 

### Since when has it been flaking?

not sure

### Reason for failure (if possible)

UT will timeout: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13375667465/job/37364903921?pr=5068

This is due to a dead lock where after the test, handlers are attempted to be removed, but we cannot get internal informer lock:

```
2025-02-17T22:43:08.4894516Z goroutine 3458743 [sync.RWMutex.Lock, 15 minutes]:
2025-02-17T22:43:08.4894684Z sync.runtime_SemacquireRWMutex(0xc07e79a594?, 0x1?, 0x44cec9?)
2025-02-17T22:43:08.4894874Z    /opt/hostedtoolcache/go/1.22.0/x64/src/runtime/sema.go:87 +0x25
2025-02-17T22:43:08.4894971Z sync.(*RWMutex).Lock(0xc07e79a580)
2025-02-17T22:43:08.4895161Z    /opt/hostedtoolcache/go/1.22.0/x64/src/sync/rwmutex.go:151 +0x8b
2025-02-17T22:43:08.4895544Z github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory.(*informer).removeAllHandlers(0xc084e42d80)
2025-02-17T22:43:08.4895855Z    /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/factory/handler.go:455 +0xc5
2025-02-17T22:43:08.4896131Z github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory.(*informer).shutdown(0xc084e42d80)
2025-02-17T22:43:08.4896434Z    /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/factory/handler.go:466 +0x27
2025-02-17T22:43:08.4896956Z github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory.(*WatchFactory).Shutdown(0xc083faa000)
2025-02-17T22:43:08.4897274Z    /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/factory/factory.go:1077 +0x105
[...]

PR #5075, opened 2025-02-18
Fixes CreateOrReplaceLogicalRouterStaticRouteWithPredicate timeout

trozet:
The function was not using a context with timeout. This would cause the event handler to wait forever in unit tests when the test was shutting down. Factory would shut down trying to remove all handlers, trying to get internal informer lock, but it would already be locked while an event was being processed waiting for libovsdb cache to become consistent.

Fixes: #5074

### DIFF
--- a/go-controller/pkg/libovsdb/ops/router.go
+++ b/go-controller/pkg/libovsdb/ops/router.go
@@ -614,16 +614,29 @@ func GetRouterLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Cli
 	lrsrs := []*nbdb.LogicalRouterStaticRoute{}
 	for _, uuid := range r.StaticRoutes {
 		lrsr := &nbdb.LogicalRouterStaticRoute{UUID: uuid}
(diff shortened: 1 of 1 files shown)


go-controller/pkg/libovsdb/ops/router.go
```diff
@@ -614,18 +614,31 @@ func GetRouterLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Cli
 	lrsrs := []*nbdb.LogicalRouterStaticRoute{}
 	for _, uuid := range r.StaticRoutes {
 		lrsr := &nbdb.LogicalRouterStaticRoute{UUID: uuid}
-		err := nbClient.Get(context.Background(), lrsr)
+		validRoute, err := routeExistsWithPredicate(nbClient, lrsr, p)
```
npinaeva, line 617: i would just call it routeExists instead of validRoute, but that is a nit

npinaeva (approved):
this shutdown deadlock was so annoying!

PR #5080, opened 2025-02-19
Serialize egress IP handling

trozet:
Today we have an event driven strategy for egress IP where any of these events can force egress IP reconcilation:
1. egressIP object change
2. namespace change
3. pod change

All 3 of these are called from different threads. These threads all block each other by using a single global mutex called podAssignmentMutex. The problem here is that pod and namespace handlers find what egressIP affects their objects, and then waits for the mutex, to then program egressIP state. They essentially can operate on old data, which causes configuration issues between the three threads. Take this example:

1. Egress IP exists, with status node 1
2. Simultaneously pod is x is created in y namespace, selected by egress IP
3. Pod and namespace handlers fire, get current Egress IP on node1, wait for podAssignmentMutex to unlock to configure
4. Egress IP is updated with status node 2
5. Egress IP thread fires on the update to node 2, it gets podAssignmentMutex Lock before pod and namespace handlers

The following is the result:
1. Egress IP gets added for node 2 by egress IP thread
2. Pod/namespace handlers eventually fire with old egress IP status of node1 and configures egress IP for node 1 as well for the pod

To fix this, adds a lock per egressIP to the controller. Now each of these threads will have to lock the egressIP object itself, and then once they have they lock, they get the latest egressIP object from teh informer cache to ensure they are operating on current data.

Also, a second commit removes a global podAssignmentMutex and instead switches to a per key lockable syncmap. This should allow parallel access to the map and speed up things a bit.

Fixes: #4092

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -169,6 +169,8 @@ type EgressIPController struct {
 	watchFactory *factory.WatchFactory
 	// event recorder used to post events to k8s
 	recorder record.EventRecorder
+	// used as a locking mechanism to serialize egress IP processing on a per egress IP basis
+	egressIPCache *syncmap.SyncMap[bool]
 	// podAssignmentMutex is used to ensure safe access to podAssignment.
 	// Currently WatchEgressIP, WatchEgressNamespace and WatchEgressPod could
 	// all access that map simultaneously, hence why this guard is needed.
@@ -219,6 +221,7 @@ func NewEIPController(
 		kube:               kube,
 		watchFactory:       watchFactory,
 		recorder:           recorder,
+		egressIPCache:      syncmap.NewSyncMap[bool](),
 		podAssignmentMutex: &sync.Mutex{},
 		nodeUpdateMutex:    &sync.Mutex{},
 		podAssignment:      map[string]*podAssignmentState{},
@@ -254,6 +257,15 @@ func NewEIPController(
 //
 //	We only care about `Spec.NamespaceSelector`, `Spec.PodSelector` and `Status` field
 func (e *EgressIPController) reconcileEgressIP(old, new *egressipv1.EgressIP) (err error) {
+	var egressIPName string
+	if old != nil {
+		egressIPName = old.Name
+	} else {
+		egressIPName = new.Name
+	}
+	e.egressIPCache.LockKey(egressIPName)
+	defer e.egressIPCache.UnlockKey(egressIPName)
+
 	// CASE 1: EIP object deletion, we need to teardown database configuration for all the statuses
 	if old != nil && new == nil {
 		removeStatus := old.Status.Items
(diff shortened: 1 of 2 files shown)


trozet:
@martinkennelly PTAL

go-controller/pkg/ovn/egressip.go
```diff
@@ -593,71 +618,90 @@ func (e *EgressIPController) reconcileEgressIPPod(old, new *corev1.Pod) (err err
...
-				oldMatches := old != nil && podSelector.Matches(oldPodLabels)
-				// If the podSelector doesn't match the pod, then continue
-				// because this EgressIP intends to match other pods in that
-				// namespace and not this one. Other EgressIP objects might
-				// match the pod though so we need to check that.
-				if !newMatches && !oldMatches {
-					continue
+				return err
+			}
+			if namespaceSelector.Matches(namespaceLabels) {
```
npinaeva, line 635: so here you use namespaceLabels, that you've got before locking the eIP, so current ns labels could be already different.
it can go like 
```
pod event handler:
  get ns labels = {label1: a}
ns event handler:
  newLabels = {label1: b}
  lock eip1
    do update for label1=b
  unlock
pod event handler:
  lock eip1
    do update for label1=a
  unlock
```
I think you need to re-get related objects after the eip lock?

npinaeva: this could also cause temporary problems for ns handler: while same-ns events are serialized, you could get
```
ns event handler
  newLabels = {label1: a}
// here label1=b ns update comes, but it waits 
// but here can come pod update
pod hander
  ns labels = {label1: b}
  eip lock
    do update for ns label1=b
  eip unlock
ns event handler
  eip lock
    do update for ns label1=a
  eip unlock
// here is the time when "incorrect" config is applied, so pod could be added for eip on pod update, then removed on the ns label1=a update, and then added back on ns label1=b update
here ns event handler will do label1=b update
```
since it is a temporary misconfig, maybe not a big problem?

go-controller/pkg/ovn/egressip.go
```diff
@@ -2182,6 +2241,7 @@ func (pas *podAssignmentState) Clone() *podAssignmentState {
 // the standby egressIP. This must always be called with a lock on podAssignmentState mutex
 // This is special case function called only from deleteEgressIPAssignments, don't use this for normal setup
 // Any failure from here will not be retried, its a corner case undefined behaviour
+// must be called with podAssignment lock
```
npinaeva, line 2244: with podAssignment lock on podKey?

npinaeva (commented):
I was mostly looking at the locking part, hoping everything else is handled correctly ;)

go-controller/pkg/ovn/egressip.go
trozet: so I was intentionally not trying to get the latest objects for the type of the object handler itself. So like egressIP handler should rely on the retry framework egress IP object it was given. I felt like retrieving the latest was kind of bypassing our retry framework. That being said, I think your first example is a valid issue. The second one I'm OK with as long as we reach eventual consistency.

So to solve the first example, I think we need to retrieve the non-handler type object after egress IP lock. In this case the pod event handler should retrieve ns labels after egress IP handler lock, and vice versa, namespace event handler should only get pods after egress ip handler lock.

That way eventually the world will be consistent and we refrain from overriding main-type objects provided by retry framework. wdyt?

npinaeva: yep, I think this should work

trozet: done

go-controller/pkg/ovn/egressip.go
```diff
@@ -514,28 +522,45 @@ func (e *EgressIPController) reconcileEgressIPNamespace(old, new *corev1.Namespa
...
 	}
 	for _, egressIP := range egressIPs {
-		namespaceSelector, err := metav1.LabelSelectorAsSelector(&egressIP.Spec.NamespaceSelector)
-		if err != nil {
-			return err
-		}
-		if namespaceSelector.Matches(oldLabels) && !namespaceSelector.Matches(newLabels) {
-			ni, err := e.networkManager.GetActiveNetworkForNamespace(namespaceName)
+		if err := e.egressIPCache.DoWithLock(egressIP.Name, func(key string) error {
+			eIP, err := e.watchFactory.GetEgressIP(key)
```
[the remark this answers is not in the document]
trozet: done

npinaeva (approved):
there are a couple more comments that need to updated from "podAssignment lock" to "podAssignment lock on podKey",e.g.
```
addPodEgressIPAssignments
deletePodEgressIPAssignments
getNetworkFromPodAssignment
```
 but otherwise this should work :)

trozet:
@npinaeva looks like i have a conflict now, so will fix the other references as well

npinaeva (approved):
`addPodEgressIPAssignments` still says "requires holding the podAssignmentMutex lock", do with that information what you will :D

PR #5086, opened 2025-02-26
Lint improvements for unit tests

jcaamano:
Enable linting for unit tests, add test specific linters.

I suggest the reviewer to do a leap of faith and not do thorough review of the changes, 99% percent of them are suggested or automatically applied by the linters.

Some highlights:

- Numerous fixes of incorrect use of `Eventually` (passing a value instead of a function).
- Numerous fixes of ignored errors, in some cases tests passing for the incorrect reasons.
- Numerous fixes to assertions and presentation of failures
- Numerous fixes on usage of test helpers (`GinkgoRecover`, `t.Helper`, ...) 

I would suggest normal attention to these changes that some level of refactoring involved:
`go-controller/pkg/node/gateway_localnet_linux_test.go`
`go-controller/pkg/node/egress_service_test.go`

### DIFF
--- a/go-controller/pkg/node/gateway_localnet_linux_test.go
+++ b/go-controller/pkg/node/gateway_localnet_linux_test.go
@@ -2000,7 +2000,7 @@ var _ = Describe("Node Operations", func() {
 					f4 := iptV4.(*util.FakeIPTables)
 					err = f4.MatchState(expectedTables, nil)
 					g.Expect(err).NotTo(HaveOccurred())
-				})
+				}).Should(Succeed())
 
 				// TODO Make delete operation fail, check retry entry, run a successful delete
 				return nil
--- a/go-controller/pkg/node/gateway_localnet_linux_test.go
+++ b/go-controller/pkg/node/gateway_localnet_linux_test.go
@@ -9,8 +9,7 @@ import (
 	"github.com/urfave/cli/v2"
 	"github.com/vishvananda/netlink"
 
-	kapi "k8s.io/api/core/v1"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8stypes "k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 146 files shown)


go-controller/pkg/node/gateway_localnet_linux_test.go
```diff
@@ -285,10 +283,8 @@ var _ = Describe("Node Operations", func() {
 		app.Name = "test"
 		app.Flags = config.Flags
 		fExec = ovntest.NewFakeExec()
-		fakeOvnNode = NewFakeOVNNode(fExec)
-		fakeOvnNode.fakeExec.AddFakeCmd(&ovntest.ExpectedCmd{
-			Cmd: "ovs-vsctl --timeout=15 --no-heading --data=bare --format=csv --columns name list interface",
-		})
+		err := util.SetExec(fExec)
```
pperiyasamy, line 286: seem like we can remove `FakeOVNNode` related functions in `ovn_test.go` as it is not used anymore by tests.

go-controller/pkg/node/gateway_localnet_linux_test.go
```diff
@@ -285,10 +283,8 @@ var _ = Describe("Node Operations", func() {
 		app.Name = "test"
 		app.Flags = config.Flags
 		fExec = ovntest.NewFakeExec()
-		fakeOvnNode = NewFakeOVNNode(fExec)
-		fakeOvnNode.fakeExec.AddFakeCmd(&ovntest.ExpectedCmd{
```
pperiyasamy, line 289: why this command is removed now ? is this specific to fake node ?
Also i would like understand why we no longer depend on using watch factory exists with fakeOvnNode.

go-controller/pkg/node/gateway_localnet_linux_test.go
```diff
@@ -394,18 +389,20 @@ var _ = Describe("Node Operations", func() {
...
-				fNPW.watchFactory = fakeOvnNode.watcher
-				Expect(startNodePortWatcher(fNPW, fakeOvnNode.fakeClient, &fakeMgmtPortConfig)).To(Succeed())
+				stopChan := make(chan struct{})
+				fakeClient := util.GetOVNClientset(&service).GetNodeClientset()
+				wf, err := factory.NewNodeWatchFactory(fakeClient, "node")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(wf.Start()).To(Succeed())
+				defer func() {
+					close(stopChan)
+					wf.Shutdown()
```
pperiyasamy, line 399: this is kind of boilerplate code in every test, but seems like we can't avoid as each test dealing with different service and endpointslice objs. hope that is ok.

pperiyasamy (commented):
what about Lint improvements for E2E tests ?

jcaamano:
> what about Lint improvements for E2E tests ?

Different PR of course. I will eventually get to it if no one else does it.

go-controller/pkg/node/gateway_localnet_linux_test.go
jcaamano: True, it is unused. But it is an exported utility so in terms of linter that's ok. I decided to let it be as the utility is similar to the one we have on ovn package and I think it might be useful in the future..

go-controller/pkg/node/gateway_localnet_linux_test.go
jcaamano: Yes, it is specific to fake node.I don't know what you mean with `using watch factory exists with fakeOvnNode`?

go-controller/pkg/node/gateway_localnet_linux_test.go
jcaamano: It certainly could be better. But I have to contain the changes to reasonable fixes to the lint issues, otherwise the work would be unbearable, and the review too!

go-controller/pkg/node/gateway_localnet_linux_test.go
jcaamano: Within `fakeOVN.Start`, the default network controller `PreStart` was failing. With lint we are force to check for this error. Rather than ignoring the errors, it's clear we don't need the default network controller for anything as the test still passes. So instead of using fakeOVN and everything that comes with, use just what we need. We are using the factory, but it is setup in the test rather than using fakeOVN.

pperiyasamy: ok @jcaamano , that explains. thanks.

pperiyasamy (approved):
LGTM

PR #5093, opened 2025-03-04
SDN-5345: allow localnet -> host network on the same node

ricky-rav:
Building on top of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5041 to allow localnet -> host network on the same node.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1846,25 +1846,31 @@ func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]strin
 							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, config.Default.ConntrackZone,
 							netConfig.masqCTMark, ofPortPhys))
 					// IP traffic from the OVN network to the host network should be handled normally by the bridge instead of
-					// being forwarded directly to the NIC (prio=100 flow above). This allows pods in the default network to reach
-					// localnet pods on the same node.
+					// being forwarded directly to the NIC (prio=100 flow above). This allows:
+					// (a) default network pods -> localnet pod (same node)
+					// (b) localnet pod -> host (same node)
 					for _, hostSubnet := range hostSubnets {
 						if hostSubnet.IP.To4() == nil {
 							continue
 						}
-						newFlow := fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, ip, nw_dst=%s, "+
-							"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
-							defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, hostSubnet,
-							config.Default.ConntrackZone, netConfig.masqCTMark)
-
-						if config.Gateway.Mode == config.GatewayModeLocal {
-							newFlow = fmt.Sprintf("cookie=%s, priority=102, in_port=LOCAL, dl_src=%s, ip, nw_dst=%s, "+
+						newFlows := []string{}
+						// scenario (a) for shared gw mode
+						if config.Gateway.Mode == config.GatewayModeShared {
+							newFlows = append(newFlows,
+								fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, ip, nw_dst=%s, "+
+									"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+									defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, hostSubnet,
+									config.Default.ConntrackZone, netConfig.masqCTMark))
+						}
+						// scenario (a) for local gw mode
+						// scenario (b) for both gw modes
+						newFlows = append(newFlows,
+							fmt.Sprintf("cookie=%s, priority=102, in_port=LOCAL, dl_src=%s, ip, nw_dst=%s, "+
 								"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
 								defaultOpenFlowCookie, bridgeMacAddress, hostSubnet,
-								config.Default.ConntrackZone, ctMarkHost)
-						}
+								config.Default.ConntrackZone, ctMarkHost))
 
-						dftFlows = append(dftFlows, newFlow)
+						dftFlows = append(dftFlows, newFlows...)
 					}
 
 				} else {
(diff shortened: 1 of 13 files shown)


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1846,25 +1846,31 @@ func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]strin
...
+						// scenario (a) for shared gw mode
+						if config.Gateway.Mode == config.GatewayModeShared {
+							newFlows = append(newFlows,
+								fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, ip, nw_dst=%s, "+
+									"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+									defaultOpenFlowCookie, netConfig.ofPortPatch, bridgeMacAddress, hostSubnet,
+									config.Default.ConntrackZone, netConfig.masqCTMark))
+						}
+						// scenario (a) for local gw mode
+						// scenario (b) for both gw modes
```
maiqueb, line 1866: maybe it's just me and my ignorance is getting in the way, but it's not immediately clear why for both GW modes we **need** this second flow.

ricky-rav: For localnet<->host, regardless of the gateway mode,  we need the same flow that allows OVN->localnet, with `in_port=LOCAL` and the host subnet as dest ip in the filter. I can try to rephrase it better in the comments.

maiqueb: please do; thanks for explaining.

ricky-rav:
@qinqon @maiqueb PTAL
Thanks!

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2154,38 +2151,67 @@ func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC
...
+			ipFamilyDest,
+			destIP,
+			config.Default.ConntrackZone,
+			ctMark)
+	}
+
+	// Traffic path (a): OVN->host for shared gw mode
+	if config.Gateway.Mode == config.GatewayModeShared {
+		for _, hostSubnet := range hostSubnets {
+			if (hostSubnet.IP.To4() == nil) != isV6 {
```
qinqon, line 2171: Is it worth it to use `utilnet.IsIPv6(ip)` to simplify this a little ?

qinqon (approved):
LGTM

Added a nit not a must

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: sure!

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2128,23 +2130,18 @@ func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]strin
...
 
-// ovnToHostNetworkNormalActionFlows returns the flows that allow IP{v4,v6} traffic from the OVN network to the host network
-// when the destination is on the same node as the sender. This is necessary for pods in the default network to reach
-// localnet pods on the same node, when the localnet is mapped to breth0. The expected srcMAC is the MAC address of breth0
-// and the expected hostSubnets is the host subnets found on the node primary interface.
-func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC string, hostSubnets []*net.IPNet, isV6 bool) []string {
-	var inPort, ctMark, ipFamily, ipFamilyDest string
+// hostNetworkNormalActionFlows returns the flows that allow IP{v4,v6} traffic:
+// a. from the OVN network to the host network
+// b. from the host to itself (through breth0)
```
trozet, line 2135: i don't know what this case is. If a host is trying to talk to itself, it would never go out an interface. Can you help me understand case b?

The original PR 5041 was for allowing a pod on a default network to reach a localnet pod. Which I think is what you mean by case a? If so, then I would expect case b to be allowing localnet to reach host. Is that right? If so, that isn't host to host, that is localnet pod to host.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2154,38 +2151,67 @@ func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC
...
+	}
+
+	// Traffic path (a): OVN->host for local gw mode, equivalent to host->host
+	// Traffic path (b): Host->host for both gw modes
 	for _, hostSubnet := range hostSubnets {
-		if (hostSubnet.IP.To4() == nil) != isV6 {
+		if utilnet.IsIPv6(hostSubnet.IP) != isV6 {
 			continue
 		}
-		// IP traffic from the OVN network to the host network should be handled normally by the bridge instead of
```
trozet, line 2161: i think you should retain this comment where it belongs in the new patch

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2154,38 +2151,67 @@ func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC
...
-		// being output directly to the NIC by the existing flow at prio=100.
-		flows = append(flows,
-			fmt.Sprintf("cookie=%s, priority=102, in_port=%s, dl_src=%s, %s, %s=%s, "+
-				"actions=ct(commit, zone=%d, exec(set_field:%s->ct_mark)), output:NORMAL",
+		flows = append(flows, formatFlow("LOCAL", hostSubnet.String(), ctMarkHost))
```
trozet, line 2184: so you are saying traffic from LOCAL port should be forwarded normally?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2154,38 +2151,67 @@ func ovnToHostNetworkNormalActionFlows(netConfig *bridgeUDNConfiguration, srcMAC
 		ipFamilyDest = "nw_dst"
 	}
 
+	formatFlow := func(inport, destIP, ctMark string) string {
```
trozet, line 2154: you should add a generic comment about what this flow does, and how its priority plays with other flows. Like your previous comment of:

"instead of being output directly to the NIC by the existing flow at prio=100."

trozet (changes requested):

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: Ah I stretched the terminology a bit too much. This is traffic flow is for localnet to host, but it applies to everything that comes from `inPort=LOCAL`: OVN->localnet when on local gw mode, localnet->host on both gw modes.

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: yes!

trozet (approved):

PR #5010, opened 2025-01-28
crds,cudn: Support localnet topology

ormergi:
## 📑 Description

This PR implements localnet support on top of the ClusterUserDefinedNetwork CRD, following https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5087
Tracking issue https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5085

With this PR changes the CDUN CRD will feature option for creating localnet topology: 
```
apiVersion: k8s.ovn.org/v1
kind: ClusterUserDefinedNetwork
metadata:
  name: test-net
spec:
  namespaceSelector:
    matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: In
      values: ["red", "blue"]
  network:
    topology: Localnet
    localnet:
      role: Secondary
      physicalNetworkName: tenantblue
      subnets: ["192.168.0.0/16", "2001:dbb::/64"]
      excludeSubnets: ["192.168.50.0/24", "2001:dbb::0/128"]
      vlan: 
        mode: Access
        access: {id: 200}
      ipam:
        lifecycle: Persistent
```
The CUDN CRD controller will create the following NAD should be create in the selected namespaces "red":
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  name: test-net
  namespace: red
  labels:
    k8s.ovn.org/user-defined-network: ""
  finalizers:
  - k8s.ovn.org/user-defined-network-protection
  ownerReferences:
  - apiVersion: k8s.ovn.org/v1
    blockOwnerDeletion: true
    controller: true
    kind: ClusterUserDefinedNetwork
    name: test-net
    uid: c27b7c6d-440e-4b31-b7b6-30a09aa2f378
spec:
  config: '{
    "cniVersion":"1.0.0",
    "type":"ovn-k8s-cni-overlay",
    "netAttachDefName":"red/test-net",
    "name":"cluster_udn_test-net",
    "role":"secondary",
    "topology":"localnet",
    "physicalNetworkName":"tenantblue",
    "subnets":"192.168.0.0/16,2001:dbb::/64",
    "excludeSubnets":"192.168.50.0/24,2001:dbb::0/128",
    "allowPersistentIPs":true,
    "mtu":1500,
    "vlanID":200
  }'
```

Fixes #

## Additional Information for reviewers

- Since localnet topology is not part of the SDN infran the MTU is not required to be adjusted to compensate for the SDN overhead, and use the default MTU - 1500.
[...]

PR #5098, opened 2025-03-07
crds, cudn: Enable spec mutation for localnet topology configurations

ormergi:
## 📑 Description

**Rebased on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5010, only the last commit is relevant**

This PR resolves second implemntion phase of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5087
Enable localnet topology configuration changes on day2 for CUDN CRs.

Fixes #

Layer2 and Layer3 typologies remain immutable.

## ✅ Checks

- [ ] My code requires tests


go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
```diff
@@ -100,6 +102,7 @@ type LocalnetConfig struct {
 	// The network will be assigned to pods who has the `k8s.v1.cni.cncf.io/networks` annotation in place pointing
 	// to subject.
 	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="layer2 topology config is immutable"
```
npinaeva, line 105: I guess these messages need to be fixed to say that actual immutable field name instead of "layer2 topology"

go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
```diff
@@ -158,6 +164,7 @@ type LocalnetConfig struct {
 	//    When set with 'Persistent', the assigned IP addresses will be persisted in `ipamclaims.k8s.cni.cncf.io` object.
 	// 	  Useful for VMs, IP address will be persistent after restarts and migrations. Supported when `ipam.mode` is `Enabled`.
 	//
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="layer2 topology config is immutable"
```
npinaeva, line 167: for all optional, but immutable fields we need an extra validation on the `LocalnetConfig`, like so
`// +kubebuilder:validation:XValidation:rule="has(oldSelf.ipam) == has(self.ipam)", message="ipam is immutable"`
otherwise you can remove ipam completely, at least if it looks like
```
      ipam: 
        mode: Enabled
```
and it will be allowed. Or even better you can update
```
localnet:
  ipam:
    mode: Disabled
```
to
```
localnet:
  subnets: ["10.10.0.0/24"]
```
because ipam and subnets are optional fields :)

ormergi:
Closing this PR for now, because we like to get users feedback on the API before allowing mutability, which has potential to introduce major complexity to the project.

PR #5102, opened 2025-03-11
Bump to k8s v1.32.3

jcaamano:
**For go-controller:**

go get k8s.io/api@v0.32.3
go get k8s.io/client-go@v0.32.3
go get k8s.io/component-helpers@v0.32.3
go get k8s.io/kubernetes@v1.32.3
go get k8s.io/apiextensions-apiserver@v1.32.3
go get k8s.io/apiextensions-apiserver@v0.32.3
go get sigs.k8s.io/controller-runtime@v0.20.3
go mod tidy
go mod vendor

GO version was bumped to v1.23, aligned linter version. Had to move
away from giantswarm since it was no longer providing useful v1.23
images and it's unknown to me why we would not be using a more official
image anyway.

Had to bump metallb to v0.14.9 as it would no longer build otherwise.

kind was updated to v0.27 and kindest/node to v1.32.3 as well. kindest/node
v1.32+ is only compatible with kind v0.27+ but metallb relies on their
own incompatible build of kind so had to patch it to use our kind
installation.

Fixed lint issues.

Updated contrib/kind.sh, CI e2e manifests and codegen.

**For test/e2e:**

go get k8s.io/kubernetes@v1.32.2
go get k8s.io/api@v0.32.2
go get k8s.io/client-go@v0.32.2
go get k8s.io/pod-security-admission@v0.32.2
go get k8s.io/kubectl@v0.32.2
go get sigs.k8s.io/controller-runtime@v0.20.3
go get k8s.io/mount-utils@v0.32.2
go get k8s.io/cri-client@v0.32.2
go get k8s.io/dynamic-resource-allocation@v0.32.2
go get k8s.io/kube-scheduler@v0.32.2
go get k8s.io/csi-translation-lib@v0.32.2
go get k8s.io/cloud-provider@v0.32.2
go get k8s.io/controller-manager@v0.32.2
go get sigs.k8s.io/apiserver-network-proxy/konnectivity-client@v0.32.0
go mod tidy

Updated kind, kubectl and k8s test binaries.

controller-manager no longer registers the kubeconfig flag so we can go
back to using the test framework utility methods for it rather than
working around it.

Added skips for new Alpha/Beta features introduced in v1.32

Fixed a kubevirt test issue which could not activate DHCP on a UDN veth
because the device was unmanaged after upgrade to kindest/node v1.32+

Worked around two metallb issues
https://github.com/metallb/metallb/issues/2723
https://github.com/metallb/metallb/issues/2724

While fixing the metallb issues, I noticed that relevant tests cases would occasionally fail because upon ovnkube-node restart, sometimes a crashloop would be triggered. Looking into it, it was due to a failure condition in UDN gateway DeleteNetwork related to flushing the nft mark chain of the UDN, which fails when it doesn't exist. This caused DeleteNetwork to be called again, attempting to close a clsoed channel and hence a panic. I fixed the panic. However this deserves a second look because we might be in an infinite loop attempting to delete the network. Porbably deleting a chain does not require to flush it before hand. cc @kyrtapz


jcaamano:
rebased conflict :/

tssurya:
FYI I am mid reviews... i know you hit conflict again but hold :)

tssurya (commented):
Don't have time for the whole thing yet. Has to wait till next week.. - ofc someone else can review it..
But I notice the anp conformance folder hasn't been updated which is usually done during bumps

contrib/kind-common
```diff
@@ -168,13 +169,15 @@ EOF
   # Override GOBIN until https://github.com/metallb/metallb/issues/2218 is fixed.
   GOBIN="" inv dev-env -n ovn -b frr -p bgp -i "${ip_family}"
 
+  docker network rm clientnet || true
```
qinqon, line 172: `docker network rm -f` instead of the `|| true`

```
Options:
  -f, --force   Do not error if the network does not exist
```

contrib/kind-common
```diff
@@ -168,13 +169,15 @@ EOF
...
   docker network create --subnet="${METALLB_CLIENT_NET_SUBNET_IPV4}" ${ipv6_network} --driver bridge clientnet
   docker network connect clientnet frr
   if  [ "$KIND_IPV6_SUPPORT" == true ]; then
     # Enable IPv6 forwarding in FRR
     docker exec frr sysctl -w net.ipv6.conf.all.forwarding=1
   fi
   # Note: this image let's us use it also for creating load balancer backends that can send big packets
+  docker rm -f lbclient || true
```
qinqon, line 180: since we have `-f` we don't need `|| true`

test/e2e/go.mod
```diff
@@ -1,27 +1,31 @@
 module github.com/ovn-org/ovn-kubernetes/test/e2e
 
-go 1.22.0
+go 1.23.0
+
+toolchain go1.23.6
```
qinqon, line 5: Can't we just use go 1.23.6 so we don't have the toolchain thing ?

test/e2e/kubevirt/net.go
```diff
@@ -15,7 +15,12 @@ import (
 )
 
 func GenerateAddressDiscoveryConfigurationCommand(iface string) string {
+	// since kindest/node v1.32+, which was upgraded with containerd v2, /sys is
+	// mounted into kind pods as read/write which activates udev in them
+	// including a rule that sets as unmanaged any veth device with a name not
+	// starting with 'eth*'. To workaround, we force the device as managed here.
 	return fmt.Sprintf(`
+nmcli d set %[1]s managed true
```
qinqon, line 23: Thanks so much!!!

test/scripts/upgrade-ovn.sh
```diff
@@ -6,6 +6,13 @@ set -ex
 export KUBECONFIG=${KUBECONFIG:-${HOME}/ovn.conf}
 export OVN_IMAGE=${OVN_IMAGE:-ovn-daemonset-fedora:pr}
 
+set -ex
+ARCH=""
+case $(uname -m) in
+    x86_64)  ARCH="amd64" ;;
+    aarch64) ARCH="arm64"   ;;
+esac
+
```
qinqon, line 15: I don't know if it's worth it but we can use bash hash tables here.
```bash
declare -A archs=( ["x86_64"]="amd64" ["aarch64"]="arm64")
ARCH="${archs[$(uname -m)]}"
```

qinqon (changes requested):

test/e2e/go.mod
jcaamano: This change is automated. Best not fight the automation. What is the problem with the toolchain thing?

go-controller/Makefile
```diff
@@ -11,8 +11,8 @@ PKGS ?=
 GOPATH ?= $(shell go env GOPATH)
 TEST_REPORT_DIR?=$(CURDIR)/_artifacts
 export TEST_REPORT_DIR
-GO_VERSION ?= 1.22.0
-GO_DOCKER_IMG = quay.io/giantswarm/golang:${GO_VERSION}
+GO_VERSION ?= 1.23
```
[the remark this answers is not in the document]
jcaamano: will defer to next chance

contrib/kind-common
qinqon: Just a super nit, good for follow ups, 
LGTM

contrib/kind-common
qinqon: Just a super nit, good for follow ups, 
LGTM

test/e2e/go.mod
qinqon: Fair enough

test/scripts/upgrade-ovn.sh
qinqon: Not a must you can skip this

contrib/kind-common
jcaamano: will follow up

contrib/kind-common
jcaamano: will follow up

test/scripts/upgrade-ovn.sh
jcaamano: just a different way to do the same thing? Or maybe I am missing a benefit?

go-controller/Makefile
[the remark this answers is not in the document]
qinqon: Is ok for follow up

tssurya:
I haven't checked if the comments were addressed: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5102#pullrequestreview-2698085307 but hoping they were since I didn't see replies

jcaamano:
> I haven't checked if the comments were addressed: [#5102 (review)](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5102#pullrequestreview-2698085307) but hoping they were since I didn't see replies

:worried: I checked that some did not apply any longer, not sure I checked all of them.

PR #5103, opened 2025-03-11
 OVN IC mode should handle user provided encapIPs

crnithya:
## 📑 Description
This PR adds support for handling user provided encap IPs in remote nodes during OVN IC mode operation

Fixes #

## ✅ Checks

- [x] My code requires tests

### DIFF
--- a/go-controller/pkg/libovsdb/libovsdb.go
+++ b/go-controller/pkg/libovsdb/libovsdb.go
@@ -140,10 +140,6 @@ func NewSBClientWithConfig(cfg config.OvnAuthConfig, promRegistry prometheus.Reg
 	enableMetricsOption := client.WithMetricsRegistryNamespaceSubsystem(promRegistry,
 		"ovnkube", "master_libovsdb")
 
-	dbModel.SetIndexes(map[string][]model.ClientIndex{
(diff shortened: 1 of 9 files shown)


crnithya:
@cathy-zhou PTAL

crnithya:
Summary
At present in IC mode, the remote node's encapip is assumed to be the remote node's primary IP and the local node's SBDB is programmed as such, which is not correct. It is also assumed that a node will only have a single encapip, though multiple encapIps are allowed by OVN. The DBmodel is indexed to only allow update of existing encapIP for a chassis, ensuring that only a single entry exists. 
To fix these, a new node annotation "k8s.ovn.org/node-encap-ips" is introduced to capture the encapips set on the node. The encapIps are obtained from "EffectiveEncapIP" field, which contains the list of encapips provided by the user or the node's primary address. The remote node's chassis record is created/updated with the entire list of encapIps at that time. The ovsdb operations model used is such that the list of encapips for the chassis can only be updated as a whole, though the individual encapips can be added later if the chassis already exists. This will take care of any changes to remote node's encapip list. Since there is reference checking and automatic garbage collection for ovsdb operations, the unreferenced encapips will be automatically deleted.

cathy-zhou:
lgtm

PR #5106, opened 2025-03-12
Multinetpol: update to the latest API version that adds endPort

npinaeva:
## 📑 Description

[multinetpol] Update to use the latest 1.0.1 release.
In 1.0.1 `endPort` support was added.

## ✅ Checks

- [X] My code requires tests
- [X] if so, I have added and/or updated the tests as required

## How to verify it

Follow e2e test example for e2e. Or apply
```
kind: Namespace
apiVersion: v1
metadata:
  name: test
---
metadata:
  namespace: test
  name: tenantblue
spec:
  config: |2
    {
            "cniVersion": "1.0.0",
            "name": "tenant-blue",
            "topology":"layer3",
            "subnets": "10.200.0.0/16/24",
            "mtu": 1300,
            "netAttachDefName": "test/tenantblue",
            "role": "secondary"
    }
---
apiVersion: k8s.cni.cncf.io/v1beta1
kind: MultiNetworkPolicy
metadata:
  namespace: test
  name: end-port
  annotations:
    k8s.v1.cni.cncf.io/policy-for: test/tenantblue
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  ingress: 
    - from:
        - ipBlock: 
            cidr: 10.244.0.0/24
      ports:
        - protocol: TCP
          port: 32000
          endPort: 32100   
```
and check that ACL is created like this
```
_uuid               : 27e22799-e440-40bb-b51e-08869ce147b4
action              : allow-related
direction           : to-lport
external_ids        : {direction=Ingress, gress-index="0", ip-block-index="0", "k8s.ovn.org/id"="tenant-blue-network-controller:NetworkPolicy:test:end-port:Ingress:0:tcp:0", "k8s.ovn.org/name"="test:end-port", "k8s.ovn.org/owner-controller"=tenant-blue-network-controller, "k8s.ovn.org/owner-type"=NetworkPolicy, port-policy-protocol=tcp}
label               : 0
log                 : false
match               : "ip4.src == 10.244.0.0/24 && tcp && 32000<=tcp.dst<=32100 && outport == @a15274770323027050471"
meter               : acl-logging
name                : "NP:test:end-port:Ingress:0"
options             : {}
priority            : 1001
sample_est          : []
sample_new          : []
severity            : []
tier                : 2
```


test/e2e/multihoming.go
```diff
@@ -1165,7 +1165,45 @@ var _ = Describe("Multi Homing", func() {
...
+					multiNetIngressLimitingPolicy(
+						secondaryNetworkName,
+						metav1.LabelSelector{
+							MatchLabels: map[string]string{"app": "stuff-doer"},
+						},
+						metav1.LabelSelector{
+							MatchLabels: map[string]string{"role": "trusted"},
+						},
+						// build a random range around the port we are actually trying to allow without explicitly setting it
+						multiNetPolicyPortRange(port-3, port+5),
```
pperiyasamy, line 1206: nit. can you also add a test to cover multinetpolicy with port range which is outside `port` value 9000 ?

npinaeva: sure, why not :)
hopefully I understood this correctly, ptal

pperiyasamy: thanks @npinaeva ! that's what i wanted :)

pperiyasamy (approved):

test/e2e/multihoming.go
```diff
@@ -1169,7 +1169,7 @@ var _ = Describe("Multi Homing", func() {
 					),
 				),
 				ginkgo.Entry(
-					"for a routed topology when the multi-net policy describes the allow-list using pod selectors",
+					"using pod selectors for a routed topology",
```
tssurya, line 1210: for my own understanding: what's a routed topology in this context?

go-controller/pkg/ovn/base_network_controller_multipolicy.go
```diff
@@ -77,6 +77,7 @@ func convertMultiNetPolicyToNetPolicy(mpolicy *mnpapi.MultiNetworkPolicy, allowP
 			ingress.Ports[j] = knet.NetworkPolicyPort{
 				Protocol: mport.Protocol,
 				Port:     mport.Port,
+				EndPort:  mport.EndPort,
```
tssurya, line 80: yayy I wish all features were as simple as this one line hehe

go-controller/pkg/ovn/base_network_controller_multipolicy.go
tssurya, line 1: self-note-for-review: I don't see a need for unit tests for this change given the netpol side has many and the underlying logic is the same + e2e tests have been added to cover port range with MNPs.

test/e2e/multihoming.go
```diff
@@ -1670,6 +1708,39 @@ var _ = Describe("Multi Homing", func() {
...
+					multiNetIngressLimitingPolicy(
+						secondaryNetworkName,
+						metav1.LabelSelector{
+							MatchLabels: map[string]string{"app": "stuff-doer"},
+						},
+						metav1.LabelSelector{
+							MatchLabels: map[string]string{"role": "trusted"},
+						},
+						// build a port range that doesn't include server port
+						multiNetPolicyPortRange(port-10, port-1),
```
tssurya, line 1741: so in this test you are creating a MNP with an allow rule that matches the range of ports between 8990 and 8999 IIUC which looks good.
but the test itself seems to check

```
By("asserting the *client* pod can't contact the server pod exposed endpoint when using ingress deny-all")
					Eventually(func() error {
						return reachToServerPodFromClient(cs, serverPodConfig, clientPodConfig, serverIP, port)
					}, 2*time.Minute, 6*time.Second).Should(Not(Succeed()))
```
should not connect? Is there a test for should connect to the range we allow ? maybe I missed that part.
but OK I think this test is checking the ACL we added is still not allowing 9000 which is case gotcha

test/e2e/multihoming.go
```diff
@@ -1165,7 +1165,45 @@ var _ = Describe("Multi Homing", func() {
...
+						containerCmd: httpServerContainerCmd(port),
+						labels:       map[string]string{"app": "stuff-doer"},
+					},
+					multiNetIngressLimitingPolicy(
+						secondaryNetworkName,
+						metav1.LabelSelector{
+							MatchLabels: map[string]string{"app": "stuff-doer"},
+						},
+						metav1.LabelSelector{
+							MatchLabels: map[string]string{"role": "trusted"},
```
tssurya, line 1203: so following up on this test from my previous question the policy created here will hit:
```
By("asserting the *allowed-client* pod can contact the server pod exposed endpoint")
					Eventually(func() error {
						return reachToServerPodFromClient(cs, serverPodConfig, allowedClientPodConfig, serverIP, port)
					}, 2*time.Minute, 6*time.Second).Should(Succeed())
```
this right, making sure all the label match semantics we have is doing what I think it is. this role:trusted would match the allowed client so now the port range allow ACL will be hit correct?

tssurya (commented):
overall LGTM
thank you Nadia! I left some questions to ensure I understand those pieces correctly, once you confirm them I can merge this

self-note: given everything around https://ovn-kubernetes.io/features/multiple-networks/multi-network-policies/ is the same no need for new docs, change is very straight forward here.

test/e2e/multihoming.go
npinaeva: I think that means Layer3, like that one with the router :D

test/e2e/multihoming.go
npinaeva: exactly! I have added 2 tests: one positive with the port range that includes server port and another one is negative that doesn't include server port. I added it to the existing tests as you noted, one of them asserts allowed connection and another denied. `role:trusted` is a label that the allowed/client pod has in both cases

test/e2e/multihoming.go
tssurya: ah gotcha thanks

Issue #5108, opened 2025-03-12
mkdocs deploy is broken


tssurya:
### Which jobs are flaking?

github pages

### Which tests are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13819043099/job/38659596577

### Since when has it been flaking?

since today

### Reason for failure (if possible)

awesome-pages is not found.. need to point the workflow to requirements file perhaps

PR #5110, opened 2025-03-12
awesome-pages became awesome-nav

tssurya:
MKDocs deploy is broken, we need a CI lane to pre-catch things like this but for now this fix should do it.
I'll try to add a CI lane for deploy test for website but as of now changes to docs is not rendering on website so let's fix that

Fixes: #5108

Screenshot from 2025-03-13 09-23-31

### DIFF
--- a/MEETINGS.md
+++ b/MEETINGS.md
@@ -12,11 +12,11 @@ The meetings last up to 1 hour.
 
 ## Meeting location
 
-Platform: CNCF Zoom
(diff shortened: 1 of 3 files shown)


tssurya:
ahhhh @kyrtapz found the RCA: https://github.com/mkdocs/catalog/commit/a932ced9e6c1302be031090614088bfbee615df2

tada very nice :P awesome-pages is NOW awesome-nav

kyrtapz (approved):
I think that as a followup I think we should pin mkdocs and all of the plugins to a specific version to avoid these kind of issues.

tssurya:
unrelated flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5112

PR #5111, opened 2025-03-13
Remove default route if UDN network is advertised on its own VRF

arghosh93:
Add or remove default route from a vrf device based on the network is advertised on its own network or default network.

Default route from a VRF corresponding to a UDN does not get removed if the network is advertised on default VRF.

default via 172.18.0.1 dev breth0 mtu 1400

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -359,13 +359,13 @@ func setupUDPAggregationUplink(ifname string) error {
 func gatewayInitInternal(nodeName, gwIntf, egressGatewayIntf string, gwNextHops []net.IP, nodeSubnets, gwIPs []*net.IPNet,
 	advertised bool, nodeAnnotator kube.Annotator) (
 	*bridgeConfiguration, *bridgeConfiguration, error) {
-	gatewayBridge, err := bridgeForInterface(gwIntf, nodeName, types.PhysicalNetworkName, nodeSubnets, gwIPs, advertised)
+	gatewayBridge, err := bridgeForInterface(gwIntf, nodeName, types.PhysicalNetworkName, nodeSubnets, gwIPs, gwNextHops, advertised)
 	if err != nil {
 		return nil, nil, fmt.Errorf("bridge for interface failed for %s: %w", gwIntf, err)
 	}
 	var egressGWBridge *bridgeConfiguration
 	if egressGatewayIntf != "" {
-		egressGWBridge, err = bridgeForInterface(egressGatewayIntf, nodeName, types.PhysicalNetworkExGwName, nodeSubnets, nil, false)
+		egressGWBridge, err = bridgeForInterface(egressGatewayIntf, nodeName, types.PhysicalNetworkExGwName, nodeSubnets, nil, nil, false)
 		if err != nil {
 			return nil, nil, fmt.Errorf("bridge for interface failed for %s: %w", egressGatewayIntf, err)
 		}
@@ -550,6 +550,7 @@ type bridgeConfiguration struct {
 	ofPortHost  string
 	netConfig   map[string]*bridgeUDNConfiguration
 	eipMarkIPs  *markIPsCache
+	nextHops    []net.IP
 }
 
 func (b *bridgeConfiguration) getGatewayIface() string {
@@ -590,7 +591,10 @@ func (b *bridgeConfiguration) updateInterfaceIPAddresses(node *corev1.Node) ([]*
 	return ifAddrs, nil
 }
 
-func bridgeForInterface(intfName, nodeName, physicalNetworkName string, nodeSubnets, gwIPs []*net.IPNet,
+func bridgeForInterface(intfName, nodeName,
+	physicalNetworkName string,
+	nodeSubnets, gwIPs []*net.IPNet,
+	gwNextHops []net.IP,
 	advertised bool) (*bridgeConfiguration, error) {
 	var intfRep string
 	var err error
@@ -608,6 +612,9 @@ func bridgeForInterface(intfName, nodeName, physicalNetworkName string, nodeSubn
 			types.DefaultNetworkName: defaultNetConfig,
 		},
 		eipMarkIPs: newMarkIPsCache(),
+	}
+	if len(gwNextHops) > 0 {
+		res.nextHops = gwNextHops
 	}
 	res.netConfig[types.DefaultNetworkName].advertised.Store(advertised)
 
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -670,7 +671,31 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link) 
 			}
 		}
 	}
-
+	// Add unreachable route to enure that kernel always finds a match to the VRF table rather than
+	// referring to default VRF table and send traffic via unwanted interfaces and to unwanted gateway.
+	// non 0 link index for an unreachable or blackhole IPv4 route returns 'invalid argument'
+	hasV4Subnet, hasV6Subnet := udng.IPMode()
+	if hasV4Subnet {
+		_, v4AnyCIDR, _ := net.ParseCIDR("0.0.0.0/0")
+		retVal = append(retVal, netlink.Route{
+			Dst:      v4AnyCIDR,
+			Table:    udng.vrfTableId,
+			Priority: 4278198272,
+			Type:     unix.RTN_UNREACHABLE,
+		})
+	}
+	// link index for an unreachable IPv6 route always get set to 1. Link index 1 refers to default loopback
+	// device at all time. Reference: https://docs.kernel.org/networking/vrf.html#using-iproute2-for-vrfs
+	if hasV6Subnet {
+		_, v6AnyCIDR, _ := net.ParseCIDR("::/0")
+		retVal = append(retVal, netlink.Route{
+			LinkIndex: types.LoopbackInterfaceIndex,
+			Dst:       v6AnyCIDR,
+			Table:     udng.vrfTableId,
+			Priority:  4278198272,
+			Type:      unix.RTN_UNREACHABLE,
+		})
+	}
 	return retVal, nil
 }
 
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"slices"
 	"sync/atomic"
 	"time"
 
@@ -81,6 +82,12 @@ type UserDefinedNetworkGateway struct {
 	// reconcile channel to signal reconciliation of the gateway on network
 	// configuration changes
 	reconcile chan struct{}
+
+	// vrfTableId holds the route table ID corresponding to management port interface of the network
+	vrfTableId int
+
+	// gwInterfaceIndex holds the link index of gateway interface
+	gwInterfaceIndex int
 }
 
 // UTILS Needed for UDN (also leveraged for default netInfo) in bridgeConfiguration
@@ -250,19 +257,29 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
 		return nil, fmt.Errorf("unable to dereference default node network controller gateway object")
 	}
 
+	if gw.openflowManager == nil {
+		return nil, fmt.Errorf("openflow manager has not been provided for network: %s", netInfo.GetNetworkName())
+	}
+	intfName := gw.openflowManager.defaultBridge.getGatewayIface()
+	link, err := util.GetNetLinkOps().LinkByName(intfName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get link for %s, error: %v", intfName, err)
+	}
+
 	return &UserDefinedNetworkGateway{
-		NetInfo:       netInfo,
-		node:          node,
-		nodeLister:    nodeLister,
-		kubeInterface: kubeInterface,
-		vrfManager:    vrfManager,
-		masqCTMark:    masqCTMark,
-		pktMark:       pktMark,
-		v4MasqIPs:     v4MasqIPs,
-		v6MasqIPs:     v6MasqIPs,
-		gateway:       gw,
-		ruleManager:   ruleManager,
-		reconcile:     make(chan struct{}, 1),
+		NetInfo:          netInfo,
+		node:             node,
+		nodeLister:       nodeLister,
+		kubeInterface:    kubeInterface,
+		vrfManager:       vrfManager,
+		masqCTMark:       masqCTMark,
+		pktMark:          pktMark,
+		v4MasqIPs:        v4MasqIPs,
+		v6MasqIPs:        v6MasqIPs,
+		gateway:          gw,
+		ruleManager:      ruleManager,
+		reconcile:        make(chan struct{}, 1),
+		gwInterfaceIndex: link.Attrs().Index,
 	}, nil
 }
 
@@ -317,21 +334,23 @@ func (udng *UserDefinedNetworkGateway) addMarkChain() error {
 // AddNetwork will be responsible to create all plumbings
 // required by this UDN on the gateway side
 func (udng *UserDefinedNetworkGateway) AddNetwork() error {
+	if udng.openflowManager == nil {
+		return fmt.Errorf("openflow manager has not been provided for network: %s", udng.NetInfo.GetNetworkName())
+	}
 	// port is created first and its MAC address configured. The IP(s) on that link are added after enslaving to a VRF device (addUDNManagementPortIPs)
 	// because IPv6 addresses are removed by the kernel (if not link local) when enslaved to a VRF device.
-	// Add the routes after setting the IP(s) to ensure that the default subnet route towards the mgmt network exists.
+	// Add the routes(AddVRFRoutes) after setting the IP(s) to ensure that the default subnet route towards the mgmt network exists.
 	mplink, err := udng.addUDNManagementPort()
 	if err != nil {
 		return fmt.Errorf("could not create management port netdevice for network %s: %w", udng.GetNetworkName(), err)
 	}
 	vrfDeviceName := util.GetNetworkVRFName(udng.NetInfo)
-	vrfTableId := util.CalculateRouteTableID(mplink.Attrs().Index)
-	routes, err := udng.computeRoutesForUDN(vrfTableId, mplink)
+	routes, err := udng.computeRoutesForUDN(mplink)
 	if err != nil {
 		return fmt.Errorf("failed to compute routes for network %s, err: %v", udng.GetNetworkName(), err)
 	}
-	if err = udng.vrfManager.AddVRF(vrfDeviceName, mplink.Attrs().Name, uint32(vrfTableId), nil); err != nil {
-		return fmt.Errorf("could not add VRF %d for network %s, err: %v", vrfTableId, udng.GetNetworkName(), err)
+	if err = udng.vrfManager.AddVRF(vrfDeviceName, mplink.Attrs().Name, uint32(udng.vrfTableId), nil); err != nil {
+		return fmt.Errorf("could not add VRF %d for network %s, err: %v", udng.vrfTableId, udng.GetNetworkName(), err)
 	}
 	if err = udng.addUDNManagementPortIPs(mplink); err != nil {
 		return fmt.Errorf("unable to add management port IP(s) for link %s, for network %s: %w", mplink.Attrs().Name, udng.GetNetworkName(), err)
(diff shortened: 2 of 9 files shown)


go-controller/pkg/node/gateway_udn.go
```diff
@@ -877,3 +933,25 @@ func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRule() error {
...
+	routes, err := udng.vrfManager.GetVRFRoutes(vrfDeviceName)
+	if err != nil {
+		return fmt.Errorf("unable to get all routes for network %s, err: %v", udng.GetNetworkName(), err)
+	}
+	var routeToBeDeleted []netlink.Route
+	for _, route := range routes {
+		if route.Dst.IP == nil && route.Dst.Mask == nil ||
+			route.Dst.IP.Equal(net.IPv4zero) ||
+			route.Dst.IP.Equal(net.IPv6zero) {
+			routeToBeDeleted = append(routeToBeDeleted, route)
```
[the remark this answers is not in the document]
arghosh93: yes, makes sense. Thanks!

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -297,6 +297,53 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
...
+
+	return vrfm.sync(vrf)
+}
+
+// GetVRFRoutes retrives all routes for a VRF
+func (vrfm *Controller) GetVRFRoutes(name string) ([]netlink.Route, error) {
+	vrfm.mu.Lock()
+	defer vrfm.mu.Unlock()
+
+	vrfLink, err := util.GetNetLinkOps().LinkByName(name)
```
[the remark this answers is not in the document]
arghosh93: We are getting cached routes only. The link index is required to fetch cached routes from the map.

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -297,6 +297,53 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
...
+				vrf.routes = append(vrf.routes[:i], vrf.routes[i+1:]...)
+			}
+		}
+		if err := vrfm.routeManager.Del(route); err != nil {
+			return fmt.Errorf("failed to delete route for VRF device %s, err: %w", route, err)
+		}
+
+	}
+
+	return vrfm.sync(vrf)
```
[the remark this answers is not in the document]
arghosh93: yes, I agree. I dont need to sync again as I am removing the route explicitly and also removing from vrf.routes. Thanks!

go-controller/pkg/node/gateway_udn.go
```diff
@@ -666,6 +657,53 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(vrfTableId int, mpLin
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get link for %s, error: %v", intfName, err)
+	}
+
+	mgmtPortName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(udng.GetNetworkID()))
+	mplink, err := util.LinkByName(mgmtPortName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get link for %s, error: %v", mgmtPortName, err)
+	}
+	vrfTableId := util.CalculateRouteTableID(mplink.Attrs().Index)
```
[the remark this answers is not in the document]
arghosh93: Thats a good suggestion but in that case we may have to create management port while initializing the UserDefinedNetworkGateway struct instead of creating it during AddNetwork(). would that be okay?
or, is it fine to update the struct just after creating management port ?

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -297,6 +297,53 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
...
+
+	vrf, ok := vrfm.vrfs[vrfLink.Attrs().Index]
+	if !ok {
+		return fmt.Errorf("failed to find VRF %s", name)
+	}
+	for _, route := range routes {
+		for i, cachedRoute := range vrf.routes {
+			if route.Dst.String() == cachedRoute.Dst.String() && route.Table == cachedRoute.Table {
+				vrf.routes = append(vrf.routes[:i], vrf.routes[i+1:]...)
+			}
```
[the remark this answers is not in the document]
jcaamano: What about using `slices.Delete` or `slices.DeleteFunc`?

go-controller/pkg/node/gateway_udn.go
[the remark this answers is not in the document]
jcaamano: Set vrfTableId after the management port is created. If something tries to use it before that, detect it and error out.

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -207,7 +210,8 @@ func (c *Controller) applyRoute(link netlink.Link, gwIP net.IP, subnet *net.IPNe
 	return nil
 }
 
-func (c *Controller) netlinkAddRoute(link netlink.Link, gwIP net.IP, subnet *net.IPNet, mtu int, srcIP net.IP, table int) error {
+func (c *Controller) netlinkAddRoute(link netlink.Link, gwIP net.IP, subnet *net.IPNet,
```
martinkennelly, line 213: do we copy the unreachable route to the EIP multi NIC table? Its linked to mgnt port so i wont think we do but id have to check.

martinkennelly: i see no changes to the unit tests for multi nic controller so i guess it changes nothing ?

arghosh93: I am not creating unreachable route to default VRF. So, its not getting copied also.
I am creating unreachable route for a VRF specific to any UDN.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -877,3 +925,37 @@ func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRule() error {
...
+	vrfDeviceName := util.GetNetworkVRFName(udng.NetInfo)
+	defaultRoute, err := udng.getDefaultRoute(false)
+	if err != nil {
+		return fmt.Errorf("unable to get default route for network %s, err: %v", udng.GetNetworkName(), err)
+	}
+	if err = udng.vrfManager.DeleteVRFRoutes(vrfDeviceName, defaultRoute); err != nil {
+		return fmt.Errorf("unable to delete routes for network %s, err: %v", udng.GetNetworkName(), err)
+	}
+	return nil
+}
```
jcaamano, line 961: I don't think we need a separate `removeDefaultRouteFromVRF` method if you don't also have a `removeDefaultRouteToVRF`.

What bothers me a bit though is the `getDefaultRoute` call which incurs in a netlink call that might be unnecessary if no route needs to be added/removed. Perhaps we could cache the MP link index as well to avoid that. That shouldn't change from under our feet, should it?

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -297,6 +298,35 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
...
+	}
+	for _, route := range routes {
+		for i, cachedRoute := range vrf.routes {
+			if route.Dst.String() == cachedRoute.Dst.String() && route.Table == cachedRoute.Table {
+				vrf.routes = slices.Delete(vrf.routes, i, i+1)
+			}
+		}
+		if err := vrfm.routeManager.Del(route); err != nil {
+			return fmt.Errorf("failed to delete route for VRF device %s, err: %w", route, err)
+		}
```
jcaamano, line 323: Perhaps we could avod this call and the subsequent netlink calls if the route was not in the cache.

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -223,18 +227,31 @@ func (c *Controller) netlinkAddRoute(link netlink.Link, gwIP net.IP, subnet *net
...
 	}
+	if priority != 0 {
+		newNlRoute.Priority = priority
+	}
+	if rtype != 0 {
+		newNlRoute.Type = rtype
+	}
+	if scope != netlink.Scope(0) {
+		newNlRoute.Scope = scope
+	}
```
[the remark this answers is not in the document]
arghosh93: yes yes, missed it!

go-controller/pkg/node/vrfmanager/vrf_manager.go
arghosh93: Hmm thats true, can we deal with this in a separate PR?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -656,13 +651,68 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(vrfTableId int, mpLin
...
+	if hasV6Subnet {
+		_, v6AnyCIDR, _ := net.ParseCIDR("::/0")
+		retVal = append(retVal, netlink.Route{
+			LinkIndex: mpLink.Attrs().Index,
+			Dst:       v6AnyCIDR,
+			Table:     udng.vrfTableId,
+			Priority:  4278198272,
+			Type:      unix.RTN_UNREACHABLE,
+		})
+	}
```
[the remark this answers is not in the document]
arghosh93: Link index we are passing with these routes will be overridden before actually creating the route using `NetLinkOps.RouteAdd() ` but it is required to be present as routeManager returns error from various functions if link index is not valid.
Instead of overriding link index at routeManager we could set correct link index here but for IPv4 unreachable or blackhole route we need to set link index 0 but getting link for link index 0 returns an error.

go-controller/pkg/node/gateway.go
```diff
@@ -359,13 +359,13 @@ func setupUDPAggregationUplink(ifname string) error {
...
 	*bridgeConfiguration, *bridgeConfiguration, error) {
-	gatewayBridge, err := bridgeForInterface(gwIntf, nodeName, types.PhysicalNetworkName, nodeSubnets, gwIPs, advertised)
+	gatewayBridge, err := bridgeForInterface(gwIntf, nodeName, types.PhysicalNetworkName, nodeSubnets, gwIPs, advertised, gwNextHops)
 	if err != nil {
 		return nil, nil, fmt.Errorf("bridge for interface failed for %s: %w", gwIntf, err)
 	}
 	var egressGWBridge *bridgeConfiguration
 	if egressGatewayIntf != "" {
-		egressGWBridge, err = bridgeForInterface(egressGatewayIntf, nodeName, types.PhysicalNetworkExGwName, nodeSubnets, nil, false)
+		egressGWBridge, err = bridgeForInterface(egressGatewayIntf, nodeName, types.PhysicalNetworkExGwName, nodeSubnets, nil, false, gwNextHops)
```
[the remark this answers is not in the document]
arghosh93: yaa, we should set gwNextHops nil in case egressGatewayIntf is there. Next hop for the other interface may not be same.

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -297,6 +298,35 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
...
+		return fmt.Errorf("failed to retrieve VRF device %s, err: %v", name, err)
+	}
+
+	vrf, ok := vrfm.vrfs[vrfLink.Attrs().Index]
+	if !ok {
+		return fmt.Errorf("failed to find VRF %s", name)
+	}
+	for _, route := range routes {
+		for i, cachedRoute := range vrf.routes {
+			if route.Dst.String() == cachedRoute.Dst.String() && route.Table == cachedRoute.Table {
```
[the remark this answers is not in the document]
arghosh93: yes, `routemanager.RoutePartiallyEqual()` seems more appropriate to compare routes before deleting.
yes, I agree that `netlink.Route`  should not be used to store cached data. We probably need a separate PR to change this to a vanilla struct and store.
I am inclined to use `slices.Delete() ` for now  and later when we start caching routes as struct literal instead of `netlink.Route` , we may also start converting to be deleted route into a struct before actual deletion.

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -355,3 +376,15 @@ func RoutePartiallyEqual(r, x netlink.Route) bool {
 		r.Flags == x.Flags &&
 		r.MTU == x.MTU
 }
+
+// update link index for unreachable or blackhole route based on IP mode. non 0 link index for an
+// unreachable or blackhole IPv4 route returns 'invalid argument'. Whereas link index for an unreachable
+// or blackhole IPv6 route always get set to 1.
+func updateLinkIndexBasedOnRouteTypes(rtype int, subnet *net.IPNet, r *netlink.Route) *netlink.Route {
```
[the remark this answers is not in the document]
arghosh93: ip-route documentation does not have any mention of this fact. But this seems logical to me as blackhole or unreachable route effectively discards traffic and doesn't require an actual outgoing interface.
I will update the comment mentioning that link index 1 refers to default loopback device at all time.

go-controller/pkg/node/gateway_udn.go
arghosh93: you mean gateway interface ID , right?
sounds good!

go-controller/pkg/node/vrfmanager/vrf_manager.go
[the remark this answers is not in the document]
jcaamano: In the meantime you can probably define an anonymous struct in this method, should be straight forward:
https://go.dev/talks/2012/10things.slide

go-controller/pkg/node/routemanager/route_manager.go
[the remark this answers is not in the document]
arghosh93: Ack

go-controller/pkg/node/gateway_udn.go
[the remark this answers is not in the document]
jcaamano: routeManager shouldn't require a link and it should be relatively easy to remove that requirement from it and certainly what we should be doing. If you want to defer, please add a comment here explaining.

go-controller/pkg/node/gateway_udn.go
jcaamano: no, management port index

go-controller/pkg/node/vrfmanager/vrf_manager.go
jcaamano: Why not now? You just need to move this call to the block above.

go-controller/pkg/node/gateway_udn.go
arghosh93: `getDefaultRoute()` tries to get `netlink.Link` for gateway interface and add `LinkIndex`  at default gateway `netlink.Route`.  We pass management port link from `AddNetwork() ` otherwise and get management port index using mpLink.Attrs().Index.

jcaamano: ah, sorry, right, the gateway interface index.

go-controller/pkg/node/vrfmanager/vrf_manager.go
arghosh93: We will have to remove the route from the cache and also call `routeManager.Del(route)` to delete a route. vrf manager sync function cant remove a route and if we update sync function to delete route then it would definitely result into more netlink calls.
 --> You just need to move this call to the block above.
Are you just asking to make it like below?
```
for _, route := range routes {
		for i, cachedRoute := range vrf.routes {
			if routemanager.RoutePartiallyEqual(route, cachedRoute) {
				vrf.routes = slices.Delete(vrf.routes, i, i+1)
                                if err := vrfm.routeManager.Del(route); err != nil {
			           return fmt.Errorf("failed to delete route for VRF device %s, err: %w", route, err)
		                }
			}
		}
	}
```

go-controller/pkg/node/vrfmanager/vrf_manager.go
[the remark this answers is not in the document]
arghosh93: Jaime, are you strongly suggesting to use `slices.DeleteFunc()`? I did not quietly understand your last comment.

go-controller/pkg/node/gateway_udn.go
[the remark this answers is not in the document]
arghosh93: We do need link index, right? to add route for management interface or default gateway interface.
Are you suggesting that we should not return error in case link index is not present in a `netlink.Route` ?

go-controller/pkg/node/vrfmanager/vrf_manager.go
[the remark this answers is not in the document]
jcaamano: Yes,  something like
```
func (vrfm *Controller) DeleteVRFRoutes(name string, routes []netlink.Route) error {
	vrfm.mu.Lock()
	defer vrfm.mu.Unlock()
        type route struct {
             // specific fields
        }
        deleteRoutesRequested := sets.Set[route].New()
        for i, r := range routes {
            deleteRoutesRequested.Insert(route{
               field1: r.field1
               // ...
            })
        }
        var toDelete []netlink.Route
        var err error
        slices.DeleteFunc(vrf.routes, func(r) bool {
            if err != nil {
               return false
            }
            delete :=  deleteRoutesRequested.Has(
               route{
                  field1: r.field1
                  // ...
             })
             if !delete {
                return false
             }
             err = vrfm.routeManager.Del(route)
             return err != nil
        })
      return err
}
```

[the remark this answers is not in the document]
arghosh93: Jaime, I would like to keep it simple. Anyway we will have to do two things:
- Delete from VRF manager route cache
- Delete the route using routemanager. Which also removes the route from routemanager cache. 
Use of a struct and `slices.Deetefunc()` seems complicated to me and i am not feeling confident about it.

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -286,16 +284,8 @@ func (c *Controller) sync() {
...
-				if err != nil {
-					if util.GetNetLinkOps().IsLinkNotFoundError(err) {
-						deletedLinkIndexes = append(deletedLinkIndexes, linkIndex)
-					} else {
-						klog.Errorf("Route Manager: failed to apply route (%s) because unable to retrieve associated link: %v", managedRoute.String(), err)
-					}
-					continue
-				}
-				if err := c.applyRoute(link, managedRoute.Gw, managedRoute.Dst, managedRoute.MTU, managedRoute.Src, managedRoute.Table); err != nil {
+				if err := c.applyRoute(managedRoute.LinkIndex, managedRoute.Gw, managedRoute.Dst, managedRoute.MTU, managedRoute.Src, managedRoute.Table,
```
jcaamano, line 287: `deletedLinkIndexes` is unused now? 
I think the safest thing to do is to keep all link index checks in place except when `linkIndex==0`.
Down the road (meaning different PR) I don't think route_manager should be concerned about link indexes.

go-controller/pkg/node/vrfmanager/vrf_manager.go
```diff
@@ -297,6 +298,34 @@ func (vrfm *Controller) AddVRFRoutes(name string, routes []netlink.Route) error
...
+	}
+
+	vrf, ok := vrfm.vrfs[vrfLink.Attrs().Index]
+	if !ok {
+		return fmt.Errorf("failed to find VRF %s", name)
+	}
+	for _, route := range routes {
+		for i, cachedRoute := range vrf.routes {
+			if routemanager.RoutePartiallyEqual(route, cachedRoute) {
+				vrf.routes = slices.Delete(vrf.routes, i, i+1)
```
jcaamano, line 318: Are you sure it is safe to remove an item from a slice while you are iterating it?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -250,6 +258,17 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
...
+	var link netlink.Link
+	if gw.openflowManager != nil {
+		intfName := gw.openflowManager.defaultBridge.getGatewayIface()
+		link, err = util.GetNetLinkOps().LinkByName(intfName)
+		if err != nil {
+			return nil, fmt.Errorf("unable to get link for %s, error: %v", intfName, err)
+		}
+	} else {
+		return nil, fmt.Errorf("openflow manager has not been invoked for network: %s", netInfo.GetNetworkName())
+	}
```
jcaamano, line 270: nit: avoid unneccessary indentation, just do 
```
if gw.openflowManager == nil {
	return nil, fmt.Errorf("openflow manager has not been invoked for network: %s", netInfo.GetNetworkName())
}
intfName := gw.openflowManager.defaultBridge.getGatewayIface()
link, err = util.GetNetLinkOps().LinkByName(intfName)
if err != nil {
    return nil, fmt.Errorf("unable to get link for %s, error: %v", intfName, err)
}
```

go-controller/pkg/node/gateway_udn.go
```diff
@@ -81,6 +83,12 @@ type UserDefinedNetworkGateway struct {
 	// reconcile channel to signal reconciliation of the gateway on network
 	// configuration changes
 	reconcile chan struct{}
+
+	// vrfTableId holds the link index of management port interface for this network
+	vrfTableId int
+
+	// gwInterfaceId holds the link index of gateway interface
+	gwInterfaceId int
```
jcaamano, line 91: nit: `gwInterfaceIndex`

Future work: this could be cached in default network bridgeConfiguration, since it is the same for all

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -353,5 +343,21 @@ func RoutePartiallyEqual(r, x netlink.Route) bool {
 		r.Gw.Equal(x.Gw) &&
 		r.Table == x.Table &&
 		r.Flags == x.Flags &&
-		r.MTU == x.MTU
+		r.MTU == x.MTU &&
+		r.Type == x.Type &&
+		r.Priority == x.Priority &&
+		r.Scope == x.Scope
+}
```
jcaamano, line 350: Actually, I am a bit unsure why are we looking at all these fields. `netlinkDelRoute` only looks at `LinkIndex`, `Dst` and `Table` and we shnould probably stick to that here to make our lives way easier.
Additionally, we should use and filter by `Protocol` static to avoid handling routes added by other protocols (i.e. BGP)
Let's bring up this discussion with @martinkennelly

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -223,21 +211,31 @@ func (c *Controller) netlinkAddRoute(link netlink.Link, gwIP net.IP, subnet *net
 	if mtu != 0 {
 		newNlRoute.MTU = mtu
 	}
+	if priority != 0 {
+		newNlRoute.Priority = priority
+	}
+	if rtype != 0 {
+		newNlRoute.Type = rtype
+	}
+	if scope != netlink.Scope(0) {
```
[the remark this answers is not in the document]
arghosh93: While creating resources or an egress IP , We copy routes from default table and add those into VRF table specific to egress IP. Ref: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/controllers/egressip/egressip.go#L642-L666
while copying routes from default table, Route.Type is set to unix.RTN_LOCAL. when route Type is unix.RTN_LOCAL, we also need to set Route.Scope to netlink.SCOPE_HOST. Otherwise we would get invalid argument error.

go-controller/pkg/node/vrfmanager/vrf_manager.go
arghosh93: i cant think of any but my knowledge is limited. If we remove from the slice and then a sync happens before the route actually gets removed from the node, then also cache is not gonna get updated with the route to be deleted. 
If you think otherwise then I will try to formulate the delete function in the way you suggested earlier.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -81,6 +82,12 @@ type UserDefinedNetworkGateway struct {
 	// reconcile channel to signal reconciliation of the gateway on network
 	// configuration changes
 	reconcile chan struct{}
+
+	// vrfTableId holds the link index of management port interface for this network
+	vrfTableId int
```
martinkennelly, line 88: go doc test is confusing and wrong

go-controller/pkg/node/routemanager/route_manager.go
martinkennelly: because those fields are used by consumers of route manager and it was a lazy way check.

go-controller/pkg/node/routemanager/route_manager.go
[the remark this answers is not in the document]
jcaamano: I don't see anywhere that `unix.RTN_LOCAL` is being set for Type. I also don't see anywhere that `netlink.SCOPE_HOST` is being set for Scope. I can't really think of a reason why that would have anything to do with this PR. So I can't really follow your reasoning.

go-controller/pkg/node/routemanager/route_manager.go
jcaamano: That doesn't really address the concern I have here. Let me explain better.

Look at `delRoute` method:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/27033e321c7e881a2a5a356b062ceee796b7d50d/go-controller/pkg/node/routemanager/route_manager.go#L117-L150

It fist calls `netlinkDelRoute` which will delete routes based on `LinkIndex`, `Dst` and `Table` only. It then calls `RoutePartiallyEqual` to remove from cache which considers all these fields.

So it might happen that we delete routes with netlink that we then don't delete from the cache. What sense does that make?

jcaamano: So the question is: should we alight these two to the minimum common denominator of fields?

go-controller/pkg/node/routemanager/route_manager.go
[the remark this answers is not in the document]
arghosh93: yes, we dont create any route with either Type set to `unix.RTN_LOCAL` or scope set to `netlink.SCOPE_HOST`. However these fields get set after creation of some routes as per default behavior. Now while creating egress IP we copy routes from one table to another and during that Type(I had to include this for unreachable routes) also comes with it.
Now there is a requirement that if you are setting Type `unix.RTN_LOCAL` then you cant have scope undefined and must set it to `netlink.SCOPE_HOST `. 
Thats the reason, I am passing Scope along with Type to avoid `invalid argument` error.

go-controller/pkg/node/gateway_udn.go
arghosh93: Will rectify

go-controller/pkg/node/routemanager/route_manager.go
arghosh93: The only downside I can think of this change is that interface would be wrongly set at the route if we do a mistake while preparing netlink.Route.
I can keep the LinkIndex check only at `addRoute()` and `delRoute()` function just to make sure we are passing a valid LinkIndex.
However I am inclined to pass LinkIndex instead of netlink.Link from  `addRoute()` and `delRoute()`.

go-controller/pkg/node/routemanager/route_manager.go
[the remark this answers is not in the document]
arghosh93: FYI, https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/controllers/egressip/egressip.go#L642-L666

go-controller/pkg/node/routemanager/route_manager.go
arghosh93: Ack, done

arghosh93:
single stack IPv6 interconnect cluster build is failing on master branch as well.

jcaamano (approved):

Issue #5082, opened 2025-02-24
OVN nftables migration broke Submariner


aswinsuryan:
### What would you like to be added?

Submariner[1] connects two clusters and ensures the pod and services connectivity between the clusters.

When submarine traffic enters the clusters with OVN-Kubernetes it gets SNATed at ovn-k8s-mp0 and since OVN moved to nftables there is no way to override this behaviour. Earlier with iptables, we avoided this SNAT with a higher priority rule in the POSTROUTING chain. With nftables , even if we add a higher priority rule, all the rules will be executed before taking a final decision, so there is no way to skip the SNAT.

If OVN-Kubernetes could expose a generic mark value that Submariner could tag the traffic with and avoid an SNAT, it would help us solve the issue.

[1]https://submariner.io/

### Why is this needed?

Submariner wants to preserve the source IP when the traffic reaches the destination pod as some use cases rely on the source IP.

yboaron:
This issue is still relevant, and https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5113 is intended to resolve it.

nyechiel:
@yboaron with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5113 now merged, can this issue be closed now?

PR #5113, opened 2025-03-13
Enable SNAT bypass in mgmtport-snat chain for specified subnets

yboaron:
## 📑 Description

This PR enables adding subnets to the don't SNAT list via node annotations, allowing applications to designate traffic that should bypass SNAT.

Fixes # https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5082

Currently traffic gets SNATed at ovn-k8s-mp0 within the mgmtport-snat chain. Since OVNK has transitioned to nftables, this behavior can no longer be overridden.

Previously, with iptables, SNAT could be avoided by adding a higher-priority rule in the POSTROUTING chain. However, with nftables, all rules are evaluated before making a final decision, making it impossible to skip SNAT.

Some applications, like Submariner, need to preserve the source IP when traffic reaches the destination pod, as certain use cases depend on it.

This PR allows users to add subnets to don't SNAT list through node annotations.    

-->

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/obj_retry_node.go
+++ b/go-controller/pkg/node/obj_retry_node.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/managementport"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/retry"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
@@ -121,7 +122,7 @@ func (h *nodeEventHandler) AreResourcesEqual(obj1, obj2 interface{}) (bool, erro
 		if !ok {
 			return false, fmt.Errorf("could not cast obj2 of type %T to *kapi.Node", obj2)
 		}
-		return reflect.DeepEqual(node1.Status.Addresses, node2.Status.Addresses), nil
+		return reflect.DeepEqual(node1.Status.Addresses, node2.Status.Addresses) && reflect.DeepEqual(node1.Annotations, node2.Annotations), nil
 
 	default:
 		return false, fmt.Errorf("no object comparison for type %s", h.objType)
@@ -175,6 +176,13 @@ func (h *nodeEventHandler) AddResource(obj interface{}, _ bool) error {
 		node := obj.(*corev1.Node)
 		// if it's our node that is changing, then nothing to do as we dont add our own IP to the nftables rules
 		if node.Name == h.nc.name {
+			if util.NodeDontSNATSubnetAnnotationExist(node) {
+				err := managementport.UpdateNoSNATSubnetsSets(node, util.ParseNodeDontSNATSubnetsList)
+				if err != nil {
+					return fmt.Errorf("error updating no snat subnets sets: %w", err)
+				}
+			}
+
 			return nil
 		}
 		return h.nc.addOrUpdateNode(node)
@@ -218,6 +226,15 @@ func (h *nodeEventHandler) UpdateResource(oldObj, newObj interface{}, _ bool) er
 
 		// if it's our node that is changing, then nothing to do as we dont add our own IP to the nftables rules
 		if newNode.Name == h.nc.name {
+
+			// if node's dont SNAT subnet annotation changed sync nftables
+			if !reflect.DeepEqual(oldNode.Annotations, newNode.Annotations) &&
+				util.NodeDontSNATSubnetAnnotationChanged(oldNode, newNode) {
+				err := managementport.UpdateNoSNATSubnetsSets(newNode, util.ParseNodeDontSNATSubnetsList)
+				if err != nil {
+					return fmt.Errorf("error updating no snat subnets sets: %w", err)
+				}
+			}
 			return nil
 		}
 
@@ -273,6 +290,10 @@ func (h *nodeEventHandler) DeleteResource(obj, _ interface{}) error {
 
 	case factory.NodeType:
 		h.nc.deleteNode(obj.(*corev1.Node))
+		_ = managementport.UpdateNoSNATSubnetsSets(obj.(*corev1.Node), func(_ *corev1.Node) ([]string, error) {
+			return []string{}, nil
+		})
+
 		return nil
 
 	default:
--- a/go-controller/pkg/util/node_annotations.go
+++ b/go-controller/pkg/util/node_annotations.go
@@ -154,6 +154,9 @@ const (
 
 	// ovnNodeEncapIPs is used to indicate encap IPs set on the node
 	OVNNodeEncapIPs = "k8s.ovn.org/node-encap-ips"
+
+	// OvnNodeDontSNATSubnets is a user assigned source subnets that should avoid SNAT at ovn-k8s-mp0 interface
+	OvnNodeDontSNATSubnets = "k8s.ovn.org/node-ingress-snat-exclude-subnets"
 )
 
 type L3GatewayConfig struct {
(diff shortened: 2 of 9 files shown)


trozet (commented):
So can you explain what kind of traffic from submariner needs to access the pod network without being SNAT'ed? Is the destination traffic always pods local to the node where the traffic arrives? If not, how does routing not occur asymmetrically for pod replies?

I had a PR in progress to skip SNAT for any packet destined to a pod local to the node. Wondering if that would solve this problem.

Let's assume that it doesn't and you really need to be able to specify some traffic that should never be SNAT'ed. Then I think the right approach would be to make a new config value here:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L288

maybe called IngressSubnetsWithoutNAT []*net.IPNet or something. Then a user can declare a slice of subnets that when that traffic ingresses we skip SNAT on the mgmt port.

yboaron:
Thanks for your feedback @trozet 

> So can you explain what kind of traffic from submariner needs to access the pod network without being SNAT'ed?

**short background:**

Submariner connects networks across clusters within a Clusterset, enabling seamless IP communication between Pods and Services across these clusters.

In each cluster, a designated node marked as submariner-gw, submariner-gw establishes tunnels to gateway nodes in other clusters. The submariner-routeagent sets up the necessary infrastructure to route cross-cluster traffic from all nodes to the gateway node.

Let's consider two clusters with the following configurations:

Cluster1: podCIDR - 10.1.0.0/16, serviceCIDR - 172.16.0.0/16
Cluster2: podCIDR - 10.2.0.0/16, serviceCIDR - 172.17.0.0/16
In both clusters, node1 acts as the gateway node.

Suppose a user workload pod1 runs on node2 in cluster1, and it needs to communicate with pod2 on node2 in cluster2. The data path flow will be:

Egress from cluster1: pod1 → node2@cluster1 → node1@cluster1 (gateway) → Submariner tunnel to cluster2
Ingress cluster2: Submariner tunnel → node1@cluster2 (gateway) → node2@cluster2 → pod2

After tunnel decapsulation on the receiving gateway, Submariner relies on the cluster's CNI to route traffic to the target pod. Since some inter-cluster applications running on top of submariner are sensitive to SNAT, Submariner needs to prevent SNAT for cross-cluster traffic.

For the scenario above:

In cluster2, rules must prevent SNAT for traffic originating from cluster1's subnet.
Similarly, in cluster1, rules must prevent SNAT for traffic originating from cluster2's subnet.

Additionally, since clusters can be added or removed from the Clusterset dynamically, SNAT prevention rules should be updated each time a cluster joins or leaves the clusterset. SNAT rules (subnets) cannot be pre-configured.

> Is the destination traffic always pods local to the node where the traffic arrives?

Nope, 

>  If not, how does routing not occur asymmetrically for pod replies?

For egress, Submariner configures routes in OVN to k8s-mp0 interface on the GW node, to reach the host network stack on the GW node, then Submariner tunnel sends traffic to remote cluster.

For ingress, when packets are received at the GW node, Submariner tunnel decapsulate it and lets OVNK route it to the destination pod, currently Submariner only adds iptables rules to prevent SNAT (append accept rule to nat/postrouting before OVNK SNAT rule applied, but after OVNK switch to nftables , we can't use this approach anymore).


>  Then I think the right approach would be to make a new config value here:

> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L288

> maybe called IngressSubnetsWithoutNAT []*net.IPNet or something. Then a user can > declare a slice of subnets that when that traffic ingresses we skip SNAT on the mgmt port.


 Is this field is predefined ? , would it be possible to dynamically add or remove subnets from this slice? as mentioned, if we choose the don't-SNAT subnet list approach, OVNK would need to expose an API that allows Submariner to update the list.

Is there any OVNK CRD we can get this field into ?

go-controller/pkg/node/obj_retry_node.go
```diff
@@ -216,18 +218,29 @@ func (h *nodeEventHandler) UpdateResource(oldObj, newObj interface{}, _ bool) er
...
 		// if it's our node that is changing, then resync routes/nftables for all nodes
 		if newNode.Name == h.nc.name {
-			nodes, err := h.nc.watchFactory.GetNodes()
-			if err != nil {
-				return fmt.Errorf("error retrieving nodes: %w", err)
-			}
-			for _, node := range nodes {
-				if node.Name == h.nc.name {
-					continue
+			if !reflect.DeepEqual(oldNode.Status.Addresses, newNode.Status.Addresses) {
```
trozet, line 221: also need to add handling logic to AddResource and DeleteResoruce

go-controller/pkg/util/node_annotations.go
```diff
@@ -151,6 +151,9 @@ const (
 	//		"l2-network-b":"10"}
 	// }",
 	ovnUDNLayer2NodeGRLRPTunnelIDs = "k8s.ovn.org/udn-layer2-node-gateway-router-lrp-tunnel-ids"
+
+	// OvnNodeDontSNATSubnets is a user assigned subnets that should avoid SNAT at ovn-k8s-mp0 interface
+	OvnNodeDontSNATSubnets = "k8s.ovn.org/node-dontsnat-subnets"
```
trozet, line 156: how about "k8s.ovn.org/node-ingress-snat-exclude-subnets" as a better name?

trozet (changes requested):
I think you are on the right track. Need to add unit tests, and an e2e test. Also, look at pkg/ovnwebhook/nodeadmission.go. We have checks in there to ensure if ovnk-node is allowed to update certain annotations. In your case you want to prevent ovnk from ever changing this annotation. This is a security feature in case a node was compromised and a user tried to change its node's annotations to modify networking configuration.

trozet:
btw I'm in the middle of rebasing my PR, should be posted shortly

trozet (commented):
looks good, just need to fix lint problems and add testing

go-controller/pkg/util/node_annotations.go
```diff
@@ -154,6 +154,9 @@ const (
 
 	// ovnNodeEncapIPs is used to indicate encap IPs set on the node
 	OVNNodeEncapIPs = "k8s.ovn.org/node-encap-ips"
+
+	// OvnNodeDontSNATSubnets is a user assigned subnets that should avoid SNAT at ovn-k8s-mp0 interface
```
trozet: maybe specify here if this is subnet is a destination or a source

trozet (commented):
code lgtm, but it is suspicious that so many ci lanes failed to install kind. Going to rerun and deploy myself

yboaron:
> code lgtm, but it is suspicious that so many ci lanes failed to install kind. Going to rerun and deploy myself

Oh, I think I found the root cause [1]. For nftables rules pointing to sets with IPv6 subnets,  should use 'ip6' instead of 'ip'. will fix it.

[1] 
 https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5113/commits/1bc14911ca8ae771c154c3a5591e33ee110dae7f#diff-e81aac60334a7aaa1911c9e86f197350a235e5ca8a2242b5787f8343c1c59fe8R467

shettyg:
@trozet 
> I had a PR in progress to skip SNAT for any packet destined to a pod local to the node. Wondering if that would solve this problem.
>btw I'm in the middle of rebasing my PR, should be posted shortly

Did you post this PR by any chance?

yboaron:
> @trozet
> 
> > I had a PR in progress to skip SNAT for any packet destined to a pod local to the node. Wondering if that would solve this problem.
> > btw I'm in the middle of rebasing my PR, should be posted shortly
> 
> Did you post this PR by any chance?

@shettyg , 

BTW, for use cases like Submariner, packet arrives at the gateway node (which is one of the cluster nodes designated as the gateway) and then needs to be routed to dest pod that could be running on any node in the cluster, not just the local (gateway) node. so skipping SNAT for packets destined to pod local to the node wouldn't address our use case.

trozet:
> @trozet
> 
> > I had a PR in progress to skip SNAT for any packet destined to a pod local to the node. Wondering if that would solve this problem.
> > btw I'm in the middle of rebasing my PR, should be posted shortly
> 
> Did you post this PR by any chance?

Not yet. Thanks for reminding me I need to post this soon.

trozet (approved):

PR #5117, opened 2025-03-14
Update website copy of the meeting times

tssurya:
Some of the governance docs which should stay in root folder as per policies if they need to appear in the website also have a copy under docs folder.. so not updating this makes the website go stale :/

### DIFF
--- a/docs/governance/MEETINGS.md
+++ b/docs/governance/MEETINGS.md
@@ -6,15 +6,17 @@ All are welcome to join our meetings! If you want to discuss something with the 
 
 ## Meeting time
 
-We meet alternate Monday's at 10:00 AM US Pacific Time.
+We meet alternate Monday's at 6:00 PM CET/CEST.
 In order to figure out when our next meeting is, please check our agenda for previous meeting history.
 The meetings last up to 1 hour.
 
 ## Meeting location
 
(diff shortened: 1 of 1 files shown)


tssurya:
@jluhrsen when you add CI can you also consider this aspect? we should be checking if updates done to root files are changed in website paths OR let's come up with a smarter way to do symlinks perhaps?

jluhrsen:
> @jluhrsen when you add CI can you also consider this aspect? we should be checking if updates done to root files are changed in website paths OR let's come up with a smarter way to do symlinks perhaps?

@tssurya , I can add some checks around this, BUT why not just symlink these files so you don't have to update both? they appear to have identical contents.

jluhrsen:
> > @jluhrsen when you add CI can you also consider this aspect? we should be checking if updates done to root files are changed in website paths OR let's come up with a smarter way to do symlinks perhaps?
> 
> @tssurya , I can add some checks around this, BUT why not just symlink these files so you don't have to update both? they appear to have identical contents.

@tssurya , [here is a PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5118) that symlinks all the ./doc/governance/* files

flavio-fernandes (approved):
/lgtm

PR #5128, opened 2025-03-18
E2Es for net seg: remove KinD specific Node selector

martinkennelly:
Blocking downstream enablement of this test.


flavio-fernandes (approved):
/lgtm

trozet (approved):

PR #5129, opened 2025-03-18
E2Es: reduce the amount of env vars

martinkennelly:
Using env vars is troublesome to config the e2es.
We should be inspecting the cluster under test and trying to figure out what config/features are available.

This is by no means a full conversion but enough to unblock me downstream.


martinkennelly:
closed in favour of #5225 cc @jluhrsen

PR #5130, opened 2025-03-18
SGW Support for BGP support for L2 UDN

kyrtapz:
Add BGP support for Layer2 primary UDN in shared gateway mode.
Local gateway mode support will be added separately.

### DIFF
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -19,7 +19,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	meta "k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -319,6 +319,8 @@ type selectedNetworks struct {
 	hostNetworkSubnets map[string][]string
 	// prefixLength is a map of selected network to their prefix length
 	prefixLength map[string]uint32
+	// networkType is a map of selected network to their topology
+	networkTopology map[string]string
 }
 
 // generateFRRConfigurations generates FRRConfigurations for the route
@@ -346,9 +348,10 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements) 
 	// validate and gather information about the networks
 	networkSet := sets.New[string]()
 	selectedNetworks := &selectedNetworks{
-		networkVRFs:    map[string]string{},
-		networkSubnets: map[string][]string{},
-		prefixLength:   map[string]uint32{},
+		networkVRFs:     map[string]string{},
+		networkSubnets:  map[string][]string{},
+		prefixLength:    map[string]uint32{},
+		networkTopology: map[string]string{},
 	}
 	for _, nad := range nads {
 		networkName := util.GetAnnotatedNetworkName(nad)
@@ -363,10 +366,14 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements) 
 		if !network.IsDefault() && !network.IsPrimaryNetwork() {
 			return nil, nil, fmt.Errorf("%w: selected network %q is not the default nor a primary network", errConfig, networkName)
 		}
-		if network.TopologyType() != types.Layer3Topology {
-			// TODO don't really know what to do with layer 2 topologies yet
+		if network.TopologyType() != types.Layer3Topology && network.TopologyType() != types.Layer2Topology {
 			return nil, nil, fmt.Errorf("%w: selected network %q has unsupported topology %q", errConfig, networkName, network.TopologyType())
 		}
+
+		if config.Gateway.Mode == config.GatewayModeLocal && network.TopologyType() == types.Layer2Topology {
+			return nil, nil, fmt.Errorf("%w: BGP is currenty not supported for Layer2 networks in local gateway mode, network: %s", errConfig, network.GetNetworkName())
+		}
+
 		vrf := util.GetNetworkVRFName(network)
 		if vfrNet, hasVFR := selectedNetworks.networkVRFs[vrf]; hasVFR && vfrNet != networkName {
 			return nil, nil, fmt.Errorf("%w: vrf %q found to be mapped to multiple networks %v", errConfig, vrf, []string{vfrNet, networkName})
@@ -374,6 +381,7 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements) 
 		networkSet.Insert(networkName)
 		selectedNetworks.vrfs = append(selectedNetworks.vrfs, vrf)
 		selectedNetworks.networkVRFs[vrf] = networkName
+		selectedNetworks.networkTopology[networkName] = network.TopologyType()
 		// TODO check overlaps?
 		for _, cidr := range network.Subnets() {
 			subnet := cidr.CIDR.String()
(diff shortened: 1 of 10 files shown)


contrib/kind-common
```diff
@@ -739,6 +739,8 @@ deploy_bgp_external_server() {
     ip_family="ipv4"
     ipv6_network=""
   fi
+  docker rm -f bgpserver || true
```
tssurya, line 742: hmmm I added this into the delete code for the cluster...

test/e2e/route_advertisements.go
```diff
@@ -222,3 +227,265 @@ var _ = ginkgo.Describe("BGP: Pod to external server when default podNetwork is
...
+					gomega.Eventually(func() bool {
+						routes, err := runCommand(cmd...)
+						framework.ExpectNoError(err, "failed to get BGP routes from node")
+						framework.Logf("Routes in node %s", routes)
+						return strings.Contains(routes, nodeIPv6LLA)
+					}, 30*time.Second).Should(gomega.BeTrue())
+				}
+			}
+
+			ginkgo.By("routes to the CUDN network are advertised to external frr router")
```
tssurya, line 381: oops

test/e2e/route_advertisements.go
```diff
@@ -222,3 +227,265 @@ var _ = ginkgo.Describe("BGP: Pod to external server when default podNetwork is
...
+			ginkgo.By("ensure route advertisement matching CUDN was created successfully")
+			gomega.Eventually(func() string {
+				reason, err := e2ekubectl.RunKubectl("", "get", "ra", ra.Name, "-o", "jsonpath={.status.conditions[?(@.type=='Accepted')].reason}")
+				if err != nil {
+					return ""
+				}
+				return reason
+			}, 30*time.Second, time.Second).Should(gomega.Equal("Accepted"))
+
+			gomega.Expect(len(serverContainerIPs)).To(gomega.BeNumerically(">", 0))
```
tssurya, line 337: repeat

test/e2e/route_advertisements.go
```diff
@@ -430,42 +412,81 @@ var _ = ginkgo.Describe("BGP: Pod to external server when CUDN Layer3 Network is
...
-  advertisements:
-    - "PodNetwork"`
+func generateL2Subnets(v4, v6 string) udnv1.DualStackCIDRs {
+	var subnets udnv1.DualStackCIDRs
+	if isIPv4Supported() {
+		subnets = append(subnets, udnv1.CIDR(v4))
+	} else if isIPv6Supported() {
+		subnets = append(subnets, udnv1.CIDR(v6))
+	}
+	return subnets
```
qinqon, line 491: This will always return a single stack ipv4 or ipv6 it should be something like 
```golang
var subnets udnv1.DualStackCIDRs
	if isIPv4Supported() {
		subnets = append(subnets, udnv1.CIDR(v4))
	} 

       if isIPv6Supported() {
		subnets = append(subnets, udnv1.CIDR(v6))
	}

	return subnets
```

test/e2e/route_advertisements.go
```diff
@@ -430,42 +412,81 @@ var _ = ginkgo.Describe("BGP: Pod to external server when CUDN Layer3 Network is
...
-    matchExpressions:
-    - key: kubernetes.io/metadata.name
-      operator: In
-      values: [ ` + targetNs + ` ]
-  network:
-    topology: Layer3
-    layer3:
-      role: Primary
-      subnets: ` + generateCIDRforClusterUDN("103.103.0.0/16", "2014:100:200::0/60")
+func generateL3Subnets(v4, v6 udnv1.Layer3Subnet) []udnv1.Layer3Subnet {
```
qinqon, line 475: ditto

qinqon (changes requested):
e2e dual stack is broken

test/e2e/route_advertisements.go
kyrtapz: Thanks, fixed!

kyrtapz:
image
:man_dancing:

qinqon (approved):
LGTM

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -482,22 +488,28 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
...
 			}
 		}
 		return eipsByNodesByNetworks[nodeName], nil
 	}
 
-	// helper to gather host subnets and egress ips as prefixes
-	getPrefixes := func(nodeName string, network string) ([]string, error) {
-		// gather host subnets
+	// helper to gather subnets and egress ips as prefixes
+	getPrefixes := func(nodeName string, network util.NetInfo) ([]string, error) {
```
jcaamano, line 500: Why not use `selectedNetworks.networkSubnets` instead of NetInfo?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -398,7 +404,7 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
 	// ordered
 	slices.Sort(selectedNetworks.vrfs)
 	slices.Sort(selectedNetworks.subnets)
-	selectedNetworks.networks = sets.List(networkSet)
+	selectedNetworks.networks = networks
```
jcaamano, line 407: This was previously an ordered list. With this change this will not be iterated in order. Have you crosschecked this is not an issue? The idea of having ordered things is to produce consistent FRRConfigurations and avoid unnecessary updates just because we only change the order in which we list things.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: I need the topology, the alternative would be to add yet another map for that to selectedNetworks, is that something you prefer to using netInfo?

kyrtapz: Or are you saying that the logic of determining which subnet to use should happen before calling getPrefixes?

jcaamano: Right. I think that is what makes sense, `selectedNetworks` is meant to contain all the pre-processed information in a way that its easy to consume. I would change it to keep the same strategy going on.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -490,14 +498,24 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
...
+			if networkTopology == types.Layer2Topology {
+				subnets = networkSubnets
+				if len(subnets) == 0 {
+					return nil, fmt.Errorf("%w: no layer2 subnets found", errConfig)
+				}
+			} else {
+				subnets, err = getHostSubnets(nodeName, network)
+				if err != nil || len(subnets) == 0 {
+					return nil, fmt.Errorf("%w: will wait for subnet annotation to be set for node %q and network %q: %w", errConfig, nodeName, network, err)
+				}
```
jcaamano, line 518: I guess I was expecting something else and I probably didn't express myself well enough.
`selectedNetworks.hostNetworkSubnets` can already be prepopulated for L2 networks so we wouldn't really need to call `getPrefixes`.

jcaamano: Scratch that. I guess this is true but would require some level of shifting to also know the node names. Had I known this I probably would not have asked for any change from your original proposal. Sorry. What do you want to do?

kyrtapz: I would just keep the current approach if that is ok with you.

kyrtapz:
This is weird: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14404688972/job/40400404436?pr=5130
L3 tests failed because CUDN status was not updated in time.
Looking at the ovn-controlplane logs we gather in CI they start after the CUDN was created, probably suggesting a crash?
Looking into it.

jcaamano (approved):

maiqueb:
@kyrtapz do we already have a PR for this PR's LGW equivalent ?

kyrtapz:
> @kyrtapz do we already have a PR for this PR's LGW equivalent ?

There is a PR https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5160.
It requires a discussion around the approach though.

Issue #4142, opened 2024-02-08
Flake: ANP: `AdminNetworkPolicyIngressSCTP` 


martinkennelly:
Test: 
```
2024-02-07T13:29:53.3214245Z         --- FAIL: TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressSCTP/Should_support_an_'deny-ingress'_policy_for_SCTP_protocol;_ensure_rule_ordering_is_respected (0.19s)
```

https://github.com/ovn-org/ovn-kubernetes/actions/runs/7814875289/job/21318044260?pr=4061

```
=== RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressSCTP/Should_support_an_'allow-ingress'_policy_for_SCTP_protocol;_ensure_rule_ordering_is_respected
2024-02-07T13:26:29.9926003Z === RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressSCTP/Should_support_an_'allow-ingress'_policy_for_SCTP_protocol_at_the_specified_port
2024-02-07T13:26:33.1522114Z === RUN   TestNetworkPolicyV2Conformance/AdminNetworkPolicyIngressSCTP/Should_support_an_'deny-ingress'_policy_for_SCTP_protocol;_ensure_rule_ordering_is_respected
2024-02-07T13:26:33.2574606Z     admin-network-policy-core-ingress-sctp-rules.go:118: FAILED Command was [/agnhost connect --timeout=3s --protocol=sctp [fd00:10:244:3::193]:9003]
2024-02-07T13:26:33.2578431Z     admin-network-policy-core-ingress-sctp-rules.go:118: Expected connection to fail from network-policy-conformance-gryffindor/harry-potter-0 to fd00:10:244:3::193, but instead it successfully connected.
2024-02-07T13:26:33.2581502Z     admin-network-policy-core-ingress-sctp-rules.go:120: 
2024-02-07T13:26:33.2583135Z         	Error Trace:	/home/runner/go/pkg/mod/sigs.k8s.io/network-policy-api@v0.1.2/conformance/tests/admin-network-policy-core-ingress-sctp-rules.go:120
2024-02-07T13:26:33.2584530Z         	Error:      	Not equal: 
2024-02-07T13:26:33.2584993Z         	            	expected: true
2024-02-07T13:26:33.2585449Z         	            	actual  : false
[...]

PR #5132, opened 2025-03-19
dist/images: fix cleanup command for northd

TOMOFUMI-KONDO:
ovs-appctl is used to exit northd when SIGTERM is notified, but northd can be stopped by ov"n"-appctl not ov"s"-appctl.

With the existing code, northd is not stopped gracefully when SIGTERM is notified, and may be killed after graceful period.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/dist/images/ovnkube.sh
+++ b/dist/images/ovnkube.sh
@@ -1012,7 +1012,7 @@ local-sb-ovsdb() {
 
 # v1.0.0 - Runs northd on master. Does not run nb_ovsdb, and sb_ovsdb
 run-ovn-northd() {
-  trap 'ovs-appctl -t ovn-northd exit >/dev/null 2>&1; exit 0' TERM
+  trap 'ovn-appctl -t ovn-northd exit >/dev/null 2>&1; exit 0' TERM
   check_ovn_daemonset_version "1.0.0"
   rm -f ${OVN_RUNDIR}/ovn-northd.pid
   rm -f ${OVN_RUNDIR}/ovn-northd.*.ctl


dist/images/ovnkube.sh
```diff
@@ -1012,7 +1012,7 @@ local-sb-ovsdb() {
 
 # v1.0.0 - Runs northd on master. Does not run nb_ovsdb, and sb_ovsdb
 run-ovn-northd() {
-  trap 'ovs-appctl -t ovn-northd exit >/dev/null 2>&1; exit 0' TERM
+  trap 'ovn-appctl -t ovn-northd exit >/dev/null 2>&1; exit 0' TERM
```
martinkennelly, line 1015: not an expert here regarding shutting down northd. I see [1] ovs-appctl used in the man page.
How do you know northd isnt gracefully stopped?

[1] https://man7.org/linux/man-pages/man8/ovn-northd.8.html

martinkennelly: cc @dceara  WDYT?

TOMOFUMI-KONDO: I see.
ovn-appctl and ovs-appctl do the same thing when the target is the same. I didn't know that.

I was using the "ovnkube.sh" script in a different environment than ovn-kubernetes, and didn't mount /var/run/ovn and /var/run/openvswitch to the same directory, so ovs-appctl couldn't find the target of "ovn-northd" in /var/run/openvswitch because it's in the only /var/run/ovn directory.

I tried graceful shutdown in the environment created by ovn-kubernetes helm chart, it worked correctly. So there was no problem. I was wrong.

Thanks for your work for clarity.

npinaeva:
known flakes:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4142

PR #5095, opened 2025-03-06
Add L3 UDN support to BGP routeimport

jcaamano:
When adding BGP support for UDNs, we forgot about a couple of details related to routeimport:
* Registering the network with the route import controller
* The fact that the VRF name for CUDNs is the CUND name itself and does not follow the same template as for UDNs.

PR #5133, opened 2025-03-19
 Add L3 UDN support to BGP routeimport V2

jcaamano:
When adding BGP support for UDNs, we forgot about a couple of details related to routeimport:

- Registering the network with the route import controller
- The fact that the VRF name for CUDNs is the CUND name itself and does not follow the same template as for UDNs.

This implementation is functionally equivalent to that of the original attempt https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5095 for those that followed that one. However that approach had a problem when dealing with potentially unknown CUDN VRFs at the time of handling the table->network cache:  we could be caching information about VRFs we might not care about at all, and also we would not able to base this cache on network IDs.

Instead, syncNetwork always goes to netlink as the source of truth to learn the table of a network since the VRF of a network is pre-defined. The table->network cache is then only maintained for known networks and only used as a hint for syncRouteUpdate to reconcile the associated network if known.


go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -343,7 +323,27 @@ func (c *controller) syncNetwork(network string) error {
 	if info == nil {
 		return nil
 	}
-	router := info.GetNetworkScopedGWRouterName(c.node)
+
+	// get the table from the network VRF. Note we go to netlink for this as
```
kyrtapz, line 327: why not try `c.tables` before getting it from netlink?

go-controller/pkg/ovn/routeimport/route_import.go
```diff
@@ -343,7 +323,27 @@ func (c *controller) syncNetwork(network string) error {
...
+	// mantain c.tables
+	table, err := c.getRoutingTableForNetwork(network)
+	if err != nil {
+		return fmt.Errorf("failed to get VRF table from network: %w", err)
+	}
+	if table == noTable {
+		return nil
+	}
+
+	// sneakily set the hint for syncRouteUpdate. Handles this sequence of events:
```
kyrtapz, line 339: I don't like that we have to do it but I guess this is still better than figuring out the network based on the tableID from a route update..

jcaamano: I think this is a good compromise

go-controller/pkg/ovn/routeimport/route_import.go
jcaamano: Because it is a cache after all, and the information might be old, so not as trustable. This became more of a concern when we can handle VRFs with any name, and not just UDN specifically named VRFs that nobody is supposed to tinker with. Most likely using `c.tables` is fine, but why not take the safe approach and have less concerns unless we know we can't?

kyrtapz: I am not against using it, it is just an extra netlink call.
I don't think that the table is just a hint though, in most cases it determines whether we call syncNetwork at all so we already rely on it heavily.

jcaamano: Discussed offline but clarifying. 
Yeah, with hint meaning it just used to hint what network to reconcile, which can be spuriously wrong but of no consequence because syncNetwork does not rely on that cache.
Staleness of a table<->network association is temporary until we receive a link update or network update.

I didn't want to give the impression of taking the decision to add an extra netlink call lightly. The more I think about it, the more I would use netlink transactions if they existed. Even inclined to add an extra netlink call just before we push the route ops but that's maybe taking it too far.

kyrtapz:
I think we should remove this skip with this pr:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/route_advertisements.go#L283-L288

kyrtapz (dismissed):

jcaamano:
The GRPC health check flakes are bothering me to no end so attempting to fix in the last commit.

kyrtapz (approved):

kyrtapz:
Thanks for fixing the EIP issue, Hope that is the end of that saga :crossed_fingers:

PR #5081, opened 2025-02-24
kubevirt, e2e: Add localnet ipamless live migration

qinqon:
## 📑 Description
The current localnet kubevirt live migration tests are implemented using persistent IPs, this change add similar version of those tests but for ipamless localnet where VMI and pod IPs are statically assigned.

PR #5138, opened 2025-03-20
Explicitly probe for frr-k8s webhook readiness

jcaamano:


kyrtapz (approved):
The fact that the webhook doesn't check it's own readiness seems like a bug to me, the endpoint will be added to the service even though it might be non-operational.
Shall we raise the issue in frr-k8s and link ithere?

An alternative to your solution would be to patch the upstream yaml with a readiness probe to the webhook port but I don't mind this approach either.

jcaamano:
oh no conflicts

kyrtapz (approved):

jcaamano:
kv-live-migration failures are known issue
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5081

PR #5142, opened 2025-03-24
E2Es: Test when CDN "PodNetwork" is advertised with EIP

martinkennelly:


martinkennelly:
Rebased. No changes except I dropped a commit from @tssurya  that unblocked testing but is merged into main.

martinkennelly:
Rebased

Issue #4228, opened 2024-03-15
UT flake: [Fail] EgressIP selectors [It] configures multiple IPv6 EIPs on different links, multiple namespaces and multiple pods


tssurya:
[[Fail] EgressIP selectors [It] configures multiple IPv6 EIPs on different links, multiple namespaces and multiple

npinaeva:
hasn't been updated for a while, closing

PR #5143, opened 2025-03-24
e2e, kubevirt: Add a case for layer2 + BGP

qinqon:
## 📑 Description

Depends-On: 
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5130

Fixes https://issues.redhat.com/browse/SDN-5767

## ✅ Checks

- [ ] My code requires tests


qinqon:
Still, one of the iperf3 clients get broken, looks like it closes after execCommand

qinqon:
Localnet test is broken, possible related to the new fedora image that already contains iperf3

test/e2e/kubevirt.go
```diff
@@ -1649,6 +1659,41 @@ runcmd:
 				Eventually(clusterUserDefinedNetworkReadyFunc(fr.DynamicClient, cudn.Name), 5*time.Second, time.Second).Should(gomega.Succeed())
 				udnName = cudn.Name
 				networkName = util.GenerateCUDNNetworkName(cudn.Name)
+				if td.ingress == "routed" {
+					ginkgo.By("create router advertisement")
+					raClient, err := raclientset.NewForConfig(fr.ClientConfig())
+					Expect(err).NotTo(HaveOccurred())
+
+					raApplyCfg := applycfgrav1.RouteAdvertisements(udnName).
```
kyrtapz, line 1695: Super NIT: with the latest API  you can just create the object, no need for apply configuration. It doesn't hurt though.

test/e2e/kubevirt.go
```diff
@@ -1095,15 +1110,33 @@ passwd:
...
-dnf install -y psmisc procps
+#!/bin/bash -xe
 iface=$(ifconfig  |grep flags |grep -v "eth0\|lo" | sed "s/: .*//")
+iface=${iface:-eth0}
+
 ipv4=$(ifconfig $iface | grep "inet "|awk '{print $2}'| sed "s#/.*##")
-ipv6=$(ifconfig $iface | grep inet6 |grep -v fe80 |awk '{print $2}'| sed "s#/.*##")
 if [ "$ipv4" != "" ]; then
 	iperf3 -s -D --bind $ipv4 --logfile /tmp/test_${ipv4}_iperf3.log
+	sleep 1
```
kyrtapz, line 1120: :sweat_smile: why do wee need all the sleeps?

test/e2e/kubevirt.go
```diff
@@ -1698,8 +1836,10 @@ runcmd:
 						Should(Equal([]string{expectedIPv6GatewayPath}), "should filter remote ipv6 gateway nexthop")
 				}
 				step = by(vmi.Name, fmt.Sprintf("Check north/south traffic before %s %s", td.resource.description, td.test.description))
-				startNorthSouthIngressIperfTraffic(externalContainerName, nodeIPs, svc.Spec.Ports[0].NodePort, step)
-				checkNorthSouthIngressIperfTraffic(externalContainerName, nodeIPs, svc.Spec.Ports[0].NodePort, step)
+				output, err := kubevirt.RunCommand(vmi, "/tmp/iperf-server.sh &", time.Minute)
```
kyrtapz, line 1839: Why is the script ran in background? isn't the iperf3 in it already supposed to be ran in background?

kyrtapz (commented):
Overall looks good.
I haven't reviewed the iperf changes in the last commit in depth.

.github/workflows/test.yml
```diff
@@ -453,7 +453,7 @@ jobs:
...
           - {"target": "external-gateway",  "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv6",      "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv4",      "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-disabled", "num-workers": "3", "network-segmentation": "enable-network-segmentation"}
-          - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "num-workers": "3", "network-segmentation": "enable-network-segmentation"}
+          - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "num-workers": "3", "network-segmentation": "enable-network-segmentation", "routeadvertisements": "true"}
```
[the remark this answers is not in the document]
qinqon: I think the problem is that the job names do not take into account the this bar 

```
```

Is this in purpose or should I fix it ?

test/e2e/kubevirt.go
```diff
@@ -1605,9 +1616,47 @@ runcmd:
 				Expect(setupUnderlay(nodes, secondaryBridge, secondaryInterfaceName, netConfig)).To(Succeed())
 			}
 
-			By("Creating NetworkAttachmentDefinition")
-			nad = generateNAD(netConfig)
-			Expect(crClient.Create(context.Background(), nad)).To(Succeed())
+			if td.topology == "layer2" {
```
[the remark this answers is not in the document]
qinqon: I would love to move all this to CUDN, but Localnet is not there yet, and for RAs we have to use CUDN.

test/e2e/kubevirt.go
```diff
@@ -1605,9 +1616,47 @@ runcmd:
...
+						NamespaceSelector: metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{{
+							Key:      "kubernetes.io/metadata.name",
+							Operator: metav1.LabelSelectorOpIn,
+							Values:   []string{fr.Namespace.Name},
+						}}},
+					},
+				}
+				Expect(crClient.Create(context.Background(), cudn)).To(Succeed())
+				DeferCleanup(func() {
+					if e2eframework.TestContext.DeleteNamespace && (e2eframework.TestContext.DeleteNamespaceOnFailure || !CurrentSpecReport().Failed()) {
```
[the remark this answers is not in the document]
qinqon: The pattern we have follow at kubevirt tests is that if something is wrong with the test you run it with `--delete-namespace=false` or `--delete-namespace-on-failure=false` but sometimes you need stuff outside of the namespace to have the test running (CUDN, RAs, underlay bridges, etc...).

test/e2e/kubevirt.go
qinqon: To detect stuff like binding errors, iperf3 -D do not fail if it cannot bind the socket.

test/e2e/kubevirt.go
qinqon: Good catch!, I think it was a blocking script at the beginning and I forgot to remove this

.github/workflows/test.yml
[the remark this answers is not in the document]
kyrtapz: sooo after looking into it a bit more github just shows all values from the matrix.
The problem is they are truncated and I am not sure if there is any way around it :facepalm: 
an example of a job that advertises default is called:
[e2e (shard-conformance, HA, shared, ipv4, noSnatGW, 1br, ic-single-node-zones, advertise-default)](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14734860761/job/41398596691?pr=5143#logs)

so if github UI wouldn't be so dumb yours would probably end with `,"true")`
I don't mind keeping it as is for now, just found it interesting.

test/e2e/kubevirt/udn.go
```diff
@@ -0,0 +1,68 @@
...
+			Subnets:             subnets,
+			IPAM:                ipam,
+			PhysicalNetworkName: networkName,
+		}
+	}
+
+	return cudn, networkName
+}
+
+func capitalizeFirstLetter(s string) string {
```
kyrtapz, line 61: is this used anywhere?

test/e2e/kubevirt.go
```diff
@@ -1775,70 +1872,81 @@ runcmd:
...
+				role:     udnv1.NetworkRolePrimary,
 			}),
 			Entry(nil, testData{
 				resource: virtualMachineWithUDN,
 				test:     liveMigrate,
-				topology: "layer2",
-				role:     "primary",
+				topology: udnv1.NetworkTopologyLayer2,
+				role:     udnv1.NetworkRolePrimary,
+				ingress:  "routed",
```
kyrtapz, line 1922: What happens in the lane where BGP is disabled?

kyrtapz (dismissed):
minor followups, other than that LGTM

kyrtapz:
@qinqon could you add a check to the N/S test for routed networks that confirms the source IP is the PodIP?

test/e2e/kubevirt.go
qinqon: It get skip for non IC
```
	if td.role == udnv1.NetworkRolePrimary && !isInterconnectEnabled() {
				const upstreamIssue = "https://github.com/ovn-org/ovn-kubernetes/issues/4528"
				e2eskipper.Skipf(
					"The egress check of tests are known to fail on non-IC deployments. Upstream issue: %s", upstreamIssue,
				)
			}

```

qinqon:
[rebased](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/6d6a0ba84c6eca3f40c887dbfb3c99e729a75b22..2a9735a42100647957bc68ed51b0d9bbe9cb5044)

.github/workflows/test.yml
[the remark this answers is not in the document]
qinqon: The bgp non virt jobs do just test BGP, but this is not the chase here, we start a BGP cluster and run all the virt tests there (that happend to have a routed ingress test there).

qinqon:
/retest-failed

Unspected flake ?

kyrtapz:
> @kyrtapz thinking about it, I think is not needed, the fact that we are connecting from external container to that IP is enough, if somehow the response packages change the IP the tcp connection is broken and we detect it.

I meant that it would good to have a check on the server side that would confirm that we are seeing the pod IP instead of the node IP which is the case in non-bgp scenarios. I think it is a valuable addition, 
If you think that is difficult to do I am fine with deferring/not doing it.

qinqon:
> > @kyrtapz thinking about it, I think is not needed, the fact that we are connecting from external container to that IP is enough, if somehow the response packages change the IP the tcp connection is broken and we detect it.
> 
> I meant that it would good to have a check on the server side that would confirm that we are seeing the pod IP instead of the node IP which is the case in non-bgp scenarios. I think it is a valuable addition, If you think that is difficult to do I am fine with deferring/not doing it.

In the case of of kubevirt test the VM is the server, that's why we don't need it, the connection IP is the one we expect.

qinqon:
@kyrtapz comments addressed, can you take another look ?

kyrtapz (approved):
thanks for adding the egress test!
LGTM

qinqon:
/retest-failed

Not related

```
 025-05-13T08:31:23.6464795Z [38;5;9m[1mSummarizing 1 Failure:[0m
2025-05-13T08:31:23.6467069Z   [38;5;9m[FAIL][0m [0mEgressIP selectors [38;5;9m[1m[It] configures one IPv6 EIP and one Pod[0m
2025-05-13T08:31:23.6467982Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:607[0m
2025-05-13T08:31:23.6468559Z 
2025-05-13T08:31:23.6469154Z [38;5;9m[1mRan 18 of 18 Specs in 6.832 seconds[0m
```

test/e2e/kubevirt.go
```diff
@@ -85,6 +90,14 @@ func newControllerRuntimeClient() (crclient.Client, error) {
 	if err != nil {
 		return nil, err
 	}
+	err = udnv1.AddToScheme(scheme)
+	if err != nil {
+		return nil, err
+	}
+	err = rav1.AddToScheme(scheme)
+	if err != nil {
```
maiqueb, line 98: ditto / nit: please inline

test/e2e/kubevirt.go
```diff
@@ -353,24 +366,48 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
 			}
 		}
 
-		startNorthSouthIngressIperfTraffic = func(containerName string, addresses []string, port int32, stage string) error {
+		startNorthSouthIperfTraffic = func(execFn func(cmd string) (string, error), addresses []string, port int32, logPrefix, stage string) error {
```
maiqueb, line 369: nit: consider adding a type for the execFn attribute - `type execFn func(cmd string) (string, error)` 

not a must, but I think it pays off.

test/e2e/kubevirt.go
```diff
@@ -353,24 +366,48 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
...
+				By(fmt.Sprintf("start from %s: %s", address, stage))
+				output, err = execFn(fmt.Sprintf("nohup iperf3 -t 0 -c %[1]s -p %[2]d --logfile %[3]s &", address, port, iperfLogFile))
+				if err != nil {
+					return fmt.Errorf("failed at starting iperf3 in background %s: %w", output, err)
 				}
 			}
 			return nil
 		}
 
+		startNorthSouthIngressIperfTraffic = func(containerName string, addresses []string, port int32, stage string) error {
```
maiqueb, line 395: nit: maybe `startNorthSouthIperfIngressTraffic` ?

test/e2e/kubevirt.go
```diff
@@ -353,24 +366,48 @@ var _ = Describe("Kubevirt Virtual Machines", func() {
...
+			}
+			return startNorthSouthIperfTraffic(execFn, addresses, port, "ingress", stage)
+		}
+
+		startNorthSouthEgressIperfTraffic = func(vmi *kubevirtv1.VirtualMachineInstance, addresses []string, stage string) error {
+			GinkgoHelper()
+			execFn := func(cmd string) (string, error) {
+				return kubevirt.RunCommand(vmi, cmd, 5*time.Second)
+			}
+			return startNorthSouthIperfTraffic(execFn, addresses, 5201, "egress", stage)
```
maiqueb, line 408: maybe we should *not* hardcode the port number ? i.e. follow-up executions of failures will fail if this hangs ... 

And we run tests that fail twice (i.e. the second execution would surely fail if this had hanged.

test/e2e/kubevirt.go
```diff
@@ -1461,7 +1548,18 @@ fi
...
+			// - In Fedora 39, the "may-fail" option is configured by cloud-init in
+			//   NetworkManager. This causes the entire interface to remain inactive
+			//   if no IPv6 address is assigned.
+			networkDataIPv4 = `version: 2
+ethernets:
+  eth0:
+    dhcp4: true
+`
+
+			networkDataDualSatck = `version: 2
```
maiqueb, line 1562:
```suggestion
			networkDataDualStack = `version: 2
```

test/e2e/kubevirt.go
```diff
@@ -1474,24 +1572,21 @@ chpasswd: { expire: False }
 `
 
 			userDataWithIperfServer = userData + fmt.Sprintf(`
-packages:
-  - iperf3
```
maiqueb, line 1478: I assume this is now present in the new test img you're pulling ?

test/e2e/kubevirt.go
```diff
@@ -1602,27 +1695,61 @@ runcmd:
...
+			iperfServerTestPods, err = createIperfServerPods(selectedNodes, cudn.Name, td.role, []string{})
 			Expect(err).NotTo(HaveOccurred())
 
+			routerContainerName := "frr"
 			externalContainerName := namespace + "-iperf"
-			externalContainerIPV4Address, externalContainerIPV6Address := createIperfExternalContainer(externalContainerName)
+			containerNetwork := "kind"
+			if td.ingress == "routed" {
+				containerNetwork = "bgpnet"
+			}
```
maiqueb, line 1734: nit: maybe encapsulate these details in a `td` method ?

test/e2e/kubevirt.go
```diff
@@ -1602,27 +1695,61 @@ runcmd:
...
 			workerNodeList, err := fr.ClientSet.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{LabelSelector: labels.FormatLabels(map[string]string{"node-role.kubernetes.io/worker": ""})})
 			Expect(err).NotTo(HaveOccurred())
 			selectedNodes = workerNodeList.Items
 			Expect(selectedNodes).NotTo(BeEmpty())
 
-			iperfServerTestPods, err = createIperfServerPods(selectedNodes, netConfig, []string{})
+			iperfServerTestPods, err = createIperfServerPods(selectedNodes, cudn.Name, td.role, []string{})
 			Expect(err).NotTo(HaveOccurred())
 
+			routerContainerName := "frr"
```
maiqueb, line 1729: can't this be a const ?

test/e2e/kubevirt.go
```diff
@@ -1558,11 +1653,15 @@ runcmd:
 			description string
 			resource    resourceCommand
 			test        testCommand
-			topology    string
-			role        string
+			topology    udnv1.NetworkTopology
+			role        udnv1.NetworkRole
+			ingress     string
```
maiqueb, line 1658: maybe add a new enum type for the ingress types ? Or just `routed` since I assume you have `routed` and empty ingress.

test/e2e/kubevirt.go
```diff
@@ -1643,24 +1770,29 @@ runcmd:
...
 			expectedAddreses := virtualMachineAddressesFromStatus(vmi, expectedNumberOfAddresses)
 			expectedAddresesAtGuest := expectedAddreses
-			testPodsIPs := podsMultusNetworkIPs(iperfServerTestPods, podNetworkStatusByNetConfigPredicate(netConfig))
-
-			step = by(vmi.Name, "Expose VM iperf server as a service")
-			svc, err := fr.ClientSet.CoreV1().Services(namespace).Create(context.TODO(), composeService("iperf3-vm-server", vmi.Name, 5201), metav1.CreateOptions{})
-			Expect(svc.Spec.Ports[0].NodePort).NotTo(Equal(0), step)
+			testPodsIPs := podsMultusNetworkIPs(iperfServerTestPods, podNetworkStatusByNetConfigPredicate(namespace, cudn.Name, strings.ToLower(string(td.role))))
+
+			serverPort := int32(5201)
```
maiqueb, line 1780: I *really* dislike this hard-coded port number. 

At the very least, let's make the issue more prominent by specifying it as a const and re-using it.

But I'm hoping to see a random port number.

test/e2e/kubevirt.go
```diff
@@ -1643,24 +1770,29 @@ runcmd:
...
+			if td.ingress != "routed" {
+				step = by(vmi.Name, "Expose VM iperf server as a service")
+				svc, err := fr.ClientSet.CoreV1().Services(namespace).Create(context.TODO(), composeService("iperf3-vm-server", vmi.Name, 5201), metav1.CreateOptions{})
+				Expect(err).ToNot(HaveOccurred())
+				Expect(svc.Spec.Ports[0].NodePort).NotTo(Equal(0), step)
+				serverPort = svc.Spec.Ports[0].NodePort
+				nodes, err := e2enode.GetBoundedReadySchedulableNodes(context.TODO(), fr.ClientSet, 1)
+				Expect(err).NotTo(HaveOccurred())
+				serverIPs = e2enode.CollectAddresses(nodes, v1.NodeInternalIP)
+			}
```
maiqueb, line 1791: maybe encapsulate this into an helper ?

qinqon:
flake -> https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4228
/retest

test/e2e/multihoming_utils.go
```diff
@@ -226,12 +226,12 @@ func podNetworkStatus(pod *v1.Pod, predicates ...func(nadapi.NetworkStatus) bool
 	return netStatusMeetingPredicates, nil
 }
 
-func podNetworkStatusByNetConfigPredicate(netConfig networkAttachmentConfig) func(nadapi.NetworkStatus) bool {
+func podNetworkStatusByNetConfigPredicate(namespace, name, role string) func(nadapi.NetworkStatus) bool {
```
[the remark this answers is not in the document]
qinqon: same thing, when using with CUDN we don't want to depend on `networkAttachmentConfig` CUDN has a proper golang struct.

test/e2e/kubevirt.go
qinqon: I don't think is it worth it, there is just one "routed" test

test/e2e/kubevirt/udn.go
```diff
@@ -0,0 +1,57 @@
...
+	udnv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+)
+
+func GenerateCUDN(namespace, name string, topology udnv1.NetworkTopology, role udnv1.NetworkRole, subnets udnv1.DualStackCIDRs) (*udnv1.ClusterUserDefinedNetwork, string) {
+	cudn := &udnv1.ClusterUserDefinedNetwork{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace + "-" + name,
+			Labels: map[string]string{
+				"name": namespace + "-" + name,
```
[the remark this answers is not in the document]
qinqon: This allow the RouteAdvertisement CRD to match the CUDN label.

test/e2e/kubevirt.go
qinqon: Yep, now that we have a proper fedora 39 test tooling image (we full IPv6 support) we can use it, so we don't have to install iperf3 ourselves.

test/e2e/kubevirt.go
qinqon: I am going to just remove it and inline it at function call, we are just using it at one place.

test/e2e/kubevirt.go
qinqon: I can use a random number

tssurya (approved):
has two +1's did not do an indepth review e2e only changes looks good to me.

maiqueb (approved):
Tks.

PR #5145, opened 2025-03-24
localnet: Make sure bridge mapping is valid

oshoval:
For localnet, the ovs `external_ids:ovn-bridge-mappings` should have entry for the network 
(either name, or physicalNetworkName if used). 
Validate it.

## 📑 Description

Fixes #

## ✅ Checks

- [x] My code requires tests

### DIFF
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -521,6 +521,12 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 		}()
 	}
 
+	ovsClient, err := libovsdb.NewOVSClient(ctx.Done())
+	if err != nil {
+		cancel()
+		return fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
+	}
+
 	if runMode.node {
 		wg.Add(1)
 		go func() {
@@ -541,7 +547,8 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 				runMode.identity,
 				wg,
 				eventRecorder,
-				routemanager.NewController())
+				routemanager.NewController(),
+				ovsClient)
 			if err != nil {
 				nodeErr = fmt.Errorf("failed to create node network controller: %w", err)
 				return
@@ -565,12 +572,6 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 	// Note: for ovnkube node mode dpu-host no metrics is required as ovs/ovn is not running on the node.
 	if config.OvnKubeNode.Mode != types.NodeModeDPUHost && config.Metrics.OVNMetricsBindAddress != "" {
 		metricsScrapeInterval := 30
-		defer cancel()
-
-		ovsClient, err := libovsdb.NewOVSClient(ctx.Done())
-		if err != nil {
-			return fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
-		}
 		if config.Metrics.ExportOVSMetrics {
 			metrics.RegisterOvsMetricsWithOvnMetrics(ovsClient, metricsScrapeInterval, ctx.Done())
 		}
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -22,6 +22,8 @@ import (
 	"k8s.io/klog/v2"
 	kexec "k8s.io/utils/exec"
 
+	"github.com/ovn-org/libovsdb/client"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/clustermanager"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controllermanager"
@@ -521,12 +523,7 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 		}()
 	}
 
-	ovsClient, err := libovsdb.NewOVSClient(ctx.Done())
-	if err != nil {
-		cancel()
-		return fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
-	}
-
+	var ovsClient client.Client
 	if runMode.node {
 		wg.Add(1)
 		go func() {
@@ -541,6 +538,11 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 			// register ovnkube node specific prometheus metrics exported by the node
 			metrics.RegisterNodeMetrics(ctx.Done())
 
+			ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
+			if err != nil {
+				nodeErr = fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
+				return
+			}
 			nodeControllerManager, err := controllermanager.NewNodeControllerManager(
 				ovnClientset,
 				watchFactory,
@@ -548,7 +550,8 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 				wg,
 				eventRecorder,
 				routemanager.NewController(),
-				ovsClient)
+				ovsClient,
+			)
 			if err != nil {
 				nodeErr = fmt.Errorf("failed to create node network controller: %w", err)
 				return
@@ -572,6 +575,14 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 	// Note: for ovnkube node mode dpu-host no metrics is required as ovs/ovn is not running on the node.
 	if config.OvnKubeNode.Mode != types.NodeModeDPUHost && config.Metrics.OVNMetricsBindAddress != "" {
 		metricsScrapeInterval := 30
+		defer cancel()
+
+		if ovsClient == nil {
+			ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
+			if err != nil {
+				return fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
+			}
+		}
 		if config.Metrics.ExportOVSMetrics {
 			metrics.RegisterOvsMetricsWithOvnMetrics(ovsClient, metricsScrapeInterval, ctx.Done())
 		}
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	current "github.com/containernetworking/cni/pkg/types/100"
@@ -14,11 +15,14 @@ import (
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 
+	"github.com/ovn-org/libovsdb/client"
+
 	ovncnitypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/udn"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/kube"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/kubevirt"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops/ovs"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
@@ -107,8 +111,8 @@ func (pr *PodRequest) checkOrUpdatePodUID(pod *corev1.Pod) error {
 	return nil
 }
 
-func (pr *PodRequest) cmdAdd(kubeAuth *KubeAPIAuth, clientset *ClientSet, networkManager networkmanager.Interface) (*Response, error) {
-	return pr.cmdAddWithGetCNIResultFunc(kubeAuth, clientset, getCNIResult, networkManager)
+func (pr *PodRequest) cmdAdd(kubeAuth *KubeAPIAuth, clientset *ClientSet, networkManager networkmanager.Interface, ovsClient client.Client) (*Response, error) {
+	return pr.cmdAddWithGetCNIResultFunc(kubeAuth, clientset, getCNIResult, networkManager, ovsClient)
 }
 
 func (pr *PodRequest) cmdAddWithGetCNIResultFunc(
@@ -116,6 +120,7 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(
 	clientset *ClientSet,
 	getCNIResultFn getCNIResultFunc,
 	networkManager networkmanager.Interface,
+	ovsClient client.Client,
 ) (*Response, error) {
 	namespace := pr.PodNamespace
 	podName := pr.PodName
@@ -172,6 +177,15 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(
 	if !config.UnprivilegedMode {
 		//TODO: There is nothing technical to run this at unprivileged mode but
 		//      we will tackle that later on.
+
+		netName := pr.netName
+		if pr.CNIConf.PhysicalNetworkName != "" {
+			netName = pr.CNIConf.PhysicalNetworkName
+		}
+		if err := checkBridgeMapping(ovsClient, pr.CNIConf.Topology, netName); err != nil {
+			return nil, fmt.Errorf("failed bridge mapping validation: %w", err)
+		}
+
 		response.Result, err = getCNIResultFn(pr, clientset, podInterfaceInfo)
 		if err != nil {
 			return nil, err
@@ -304,6 +318,7 @@ func HandlePodRequest(
 	clientset *ClientSet,
 	kubeAuth *KubeAPIAuth,
 	networkManager networkmanager.Interface,
+	ovsClient client.Client,
 ) ([]byte, error) {
 	var result, resultForLogging []byte
 	var response *Response
@@ -312,7 +327,7 @@ func HandlePodRequest(
 	klog.Infof("%s %s starting CNI request %+v", request, request.Command, request)
 	switch request.Command {
 	case CNIAdd:
-		response, err = request.cmdAdd(kubeAuth, clientset, networkManager)
+		response, err = request.cmdAdd(kubeAuth, clientset, networkManager, ovsClient)
 	case CNIDel:
 		response, err = request.cmdDel(clientset)
 	case CNICheck:
@@ -419,4 +434,27 @@ func (pr *PodRequest) buildPodInterfaceInfo(annotations map[string]string, podAn
 		pr.netName,
 		pr.CNIConf.MTU,
 	)
+}
+
+func checkBridgeMapping(ovsClient client.Client, topology string, networkName string) error {
+	if topology != types.LocalnetTopology || networkName == types.DefaultNetworkName {
+		return nil
+	}
+
+	openvSwitch, err := ovs.GetOpenvSwitch(ovsClient)
+	if err != nil {
+		return fmt.Errorf("failed getting openvswitch: %w", err)
+	}
+
+	ovnBridgeMappings := openvSwitch.ExternalIDs["ovn-bridge-mappings"]
+
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		networkBridgeAssociation := strings.Split(bridgeMapping, ":")
+		if len(networkBridgeAssociation) == 2 && networkBridgeAssociation[0] == networkName {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("failed to find bridge mapping for network: %q; Current ovn-bridge-mappings: %q", networkName, ovnBridgeMappings)
 }
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -455,6 +455,6 @@ func checkBridgeMapping(ovsClient client.Client, topology string, networkName st
 			return nil
 		}
 	}
-
-	return fmt.Errorf("failed to find bridge mapping for network: %q; Current ovn-bridge-mappings: %q", networkName, ovnBridgeMappings)
+	klog.V(5).Infof("Failed to find bridge mapping for network: %q, current OVN bridge-mappings: (%s)", networkName, ovnBridgeMappings)
+	return fmt.Errorf("failed to find OVN bridge-mapping for network: %q", networkName)
 }
--- a/go-controller/pkg/cni/types.go
+++ b/go-controller/pkg/cni/types.go
@@ -14,6 +14,8 @@ import (
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 
+	"github.com/ovn-org/libovsdb/client"
+
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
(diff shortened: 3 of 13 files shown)


go-controller/pkg/cni/types.go
```diff
@@ -166,6 +166,9 @@ type PodRequest struct {
 
 	// the DeviceInfo struct
 	deviceInfo nadapi.DeviceInfo
+
+	// the topology type
+	toplogy string
```
maiqueb, line 171: drop this, it is not needed.

go-controller/pkg/cni/udn/primary_network.go
```diff
@@ -25,6 +25,10 @@ func NewPrimaryNetwork(networkManager networkmanager.Interface) *UserDefinedPrim
 	}
 }
 
+func (p *UserDefinedPrimaryNetwork) TopologyType() string {
+	return p.activeNetwork.TopologyType()
+}
+
```
maiqueb, line 31: drop this, it is not needed.

go-controller/pkg/cni/cni.go
```diff
@@ -345,6 +349,10 @@ func HandlePodRequest(
 // instance of the pod in the apiserver, see checkCancelSandbox for more info.
 // If kube api is not available from the CNI, pass nil to skip this check.
 func getCNIResult(pr *PodRequest, getter PodInfoGetter, podInterfaceInfo *PodInterfaceInfo) (*current.Result, error) {
+	if err := checkBridgeMapping(pr.toplogy, pr.netName); err != nil {
```
maiqueb, line 352: You already have the topology defined in the CNI conf. Did you try using it and it didn't work ? 

```suggestion
	if err := checkBridgeMapping(pr.CNIConf.Topology, pr.netName); err != nil {
```

maiqueb (changes requested):
Please check this version out. 

I think it should just work.

You're returning an error if the mapping is not found, which I think will crash-loop the pod, as we discussed with @jcaamano previously. 

Please provide some data on what happens when you schedule a pod and hte mapping is not available - do we see a proper error message when we describe the pod ?

maiqueb:
IMHO you'll also need an e2e test to assert this behavior - at the very least, add a new localnet describe block to https://github.com/ovn-kubernetes/ovn-kubernetes/blob/a3570ad9188f21b41b5ed2968d546a3161eaffb4/go-controller/pkg/cni/cni_test.go#L1 with some unit tests that exercise this - i.e. expect this error when the mapping is not found; expect nil when the mapping is found.

maiqueb:
> Please check this version out.
> 
> I think it should just work.
> 
> You're returning an error if the mapping is not found, which I think will crash-loop the pod, as we discussed with @jcaamano previously.
> 
> Please provide some data on what happens when you schedule a pod and hte mapping is not available - do we see a proper error message when we describe the pod ?

Just deployed the changes I proposed on top of yours, and the pod gets stuck in `ContainerCreating` and the following error is shown on a pod describe:
```
  Normal   AddedInterface          10s  multus   Add eth0 [10.244.0.3/24] from ovn-kubernetes
  Warning  FailedCreatePodSandBox  9s   kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d": plugin type="multus" name="multus-cni-network" failed (add): [default/tinypod/575876eb-a636-4c88-846e-fe45d2954fb7:tenantblack]: error adding container to network "tenantblack": CNI request failed with status 400: '[default/tinypod a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d network tenantblack NAD default/localnet-network] [default/tinypod a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d network tenantblack NAD default/localnet-network] failed bridge mapping validation: failed to find bridge mapping for network: "tenantblack"; Current ovn-bridge-mappings: "physnet:breth0"
```

oshoval:
> > Please check this version out.
> > I think it should just work.
> > You're returning an error if the mapping is not found, which I think will crash-loop the pod, as we discussed with @jcaamano previously.
> > Please provide some data on what happens when you schedule a pod and hte mapping is not available - do we see a proper error message when we describe the pod ?
> 
> Just deployed the changes I proposed on top of yours, and the pod gets stuck in `ContainerCreating` and the following error is shown on a pod describe:
> 
> ```
>   Normal   AddedInterface          10s  multus   Add eth0 [10.244.0.3/24] from ovn-kubernetes
>   Warning  FailedCreatePodSandBox  9s   kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d": plugin type="multus" name="multus-cni-network" failed (add): [default/tinypod/575876eb-a636-4c88-846e-fe45d2954fb7:tenantblack]: error adding container to network "tenantblack": CNI request failed with status 400: '[default/tinypod a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d network tenantblack NAD default/localnet-network] [default/tinypod a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d network tenantblack NAD default/localnet-network] failed bridge mapping validation: failed to find bridge mapping for network: "tenantblack"; Current ovn-bridge-mappings: "physnet:breth0"
> ```

good enough then right ?
thanks

maiqueb:
> > > Please check this version out.
> > > I think it should just work.
> > > You're returning an error if the mapping is not found, which I think will crash-loop the pod, as we discussed with @jcaamano previously.
> > > Please provide some data on what happens when you schedule a pod and hte mapping is not available - do we see a proper error message when we describe the pod ?
> > 
> > 
> > Just deployed the changes I proposed on top of yours, and the pod gets stuck in `ContainerCreating` and the following error is shown on a pod describe:
> > ```
> >   Normal   AddedInterface          10s  multus   Add eth0 [10.244.0.3/24] from ovn-kubernetes
> >   Warning  FailedCreatePodSandBox  9s   kubelet  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d": plugin type="multus" name="multus-cni-network" failed (add): [default/tinypod/575876eb-a636-4c88-846e-fe45d2954fb7:tenantblack]: error adding container to network "tenantblack": CNI request failed with status 400: '[default/tinypod a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d network tenantblack NAD default/localnet-network] [default/tinypod a482587c210a6a6145534d4e5c45be4758accda356c7eb58c9f776d1e437941d network tenantblack NAD default/localnet-network] failed bridge mapping validation: failed to find bridge mapping for network: "tenantblack"; Current ovn-bridge-mappings: "physnet:breth0"
> > ```
> 
> good enough then right ? thanks

And the pod went into `Running` once I've manually added that mapping to the node where the pod was scheduled on.

IMHO yes.

Please focus on the tests.

oshoval:
Thanks, addressed comments / added UT
they are enough imho
testing directly `checkBridgeMapping` cases, and not the caller of that
because the use is simple (it just calls it) and even the current UT use stub for getCNIResult (`getCNIResultStub`)
so it isn't straight forward to test it, and also not that important

maiqueb:
> Thanks, addressed comments / added UT they are enough imho testing directly `checkBridgeMapping` cases, and not the caller of that because the use is simple (it just calls it) and even the current UT use stub for getCNIResult (`getCNIResultStub`) so it isn't straight forward to test it, and also not that important

But why do you need to invoke `checkBridgeMapping` from `getCNIResult` ? Can't you invoke that from the caller, which would enable you to test the integration of this function in the CNI flow ? 

What you'll run into trouble with then if you need to somehow mock the OVS DB - since you have that available in the unit tests.

oshoval:
The caller is `CmdAdd`, and since we can't change its prototype, it doesn't make sense to move it there
(also imho too "high" level there), moreover it will need mock indeed, and it is better to use [harness](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145/files#diff-d38f452089bd09509003f301b9158ea207fb6865229d13e22d769357dff391bfR389) as done now (and on other tests),
the current call is very simple, so on those cases, it doesnt make sense to over engineer
we just create the ovs client and call the function, returning err in case of an error

oshoval:
fixed gci lint problems

oshoval:
should we address https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/#discussion_r2000962389 in this PR as well ?
for that will need somehow to export the ovsClient, so few files can use it

oshoval:
WIP - need to fix the e2e tests as done on previous PR
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14155936643/job/39655738416?pr=5145

oshoval:
rebased 
fixed e2e tests according previous PR - lets see it pass

oshoval:
need to see this one is not real problem
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14156782786/job/39657620377?pr=5145
(wasnt on the previous PR)

oshoval:
```
Summarizing 1 Failure:
  [FAIL] Multi Homing multiple pods connected to the same OVN-K secondary network localnet OVN-K secondary network with a trunked configuration [BeforeEach] the same bridge mapping can be shared by a separate VLAN by using the physical network name attribute
  /root/go/src/github.com/ovn-kubernetes/test/e2e/multihoming.go:1261
```
seems it fails locally, even on clean main branch

trying to check why it has the bridge already created even that i run one test only,
wondering if we will see it on CI on other PRs as well
```
  STEP: setting up the localnet underlay with a trunked configuration @ 03/31/25 13:17:37.679
  [FAILED] in [BeforeEach] - /root/go/src/github.com/ovn-kubernetes/test/e2e/multihoming.go:1267 @ 03/31/25 13:17:37.832
  [PANICKED] in [AfterEach] - /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.linux-amd64/src/runtime/panic.go:261 @ 03/31/25 13:17:40.042

...

  [BeforeEach] /root/go/src/github.com/ovn-kubernetes/test/e2e/multihoming.go:1251
  [It] /root/go/src/github.com/ovn-kubernetes/test/e2e/multihoming.go:1295

  [FAILED] configuring the OVS bridge
      <*errors.errorString | 0xc0018149b0>: 
      failed to ADD OVS bridge ovsbr1: failed to run "kubectl -n ovn-kubernetes exec ovs-node-9s45z -- ovs-vsctl add-br ovsbr1": exit status 1 (ovs-vsctl: cannot create a bridge named ovsbr1 because a bridge named ovsbr1 already exists
      command terminated with exit code 1
      )
      {
          s: "failed to ADD OVS bridge ovsbr1: failed to run \"kubectl -n ovn-kubernetes exec ovs-node-9s45z -- ovs-vsctl add-br ovsbr1\": exit status 1 (ovs-vsctl: cannot create a bridge named ovsbr1 because a bridge named ovsbr1 already exists\ncommand terminated with exit code 1\n)",
      }
```

oshoval:
if using `export FLAKE_ATTEMPTS=0`
we can see this hidden problem
```
  [FAILED] in [BeforeEach] - /root/go/src/github.com/ovn-kubernetes/test/e2e/multihoming.go:1261 @ 03/31/25 14:46:48.677
  I0331 14:46:48.729550 3783263 util.go:1047] Unexpected error: 
      <*errors.errorString | 0xc000d4e070>: 
      failed to run "docker exec ovn-control-plane mkdir -p /var/log/e2e-dbs/the_same_bridge_mapping_can_be_shared_by_a_separate_VLAN_by_using_the_physical_network_name_attribute-multi-homing-6439": exit status 1 (Error response from daemon: client version 1.44 is too new. Maximum supported API version is 1.41
      )
      {
          s: "failed to run \"docker exec ovn-control-plane mkdir -p /var/log/e2e-dbs/the_same_bridge_mapping_can_be_shared_by_a_separate_VLAN_by_using_the_physical_network_name_attribute-multi-homing-6439\": exit status 1 (Error response from daemon: client version 1.44 is too new. Maximum supported API version is 1.41\n)",
      }
  [FAILED] in [JustAfterEach] - /root/go/src/github.com/ovn-kubernetes/test/e2e/util.go:1047 @ 03/31/25 14:46:48.729
  [PANICKED] in [AfterEach] - /root/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.0.linux-amd64/src/runtime/panic.go:261 @ 03/31/25 14:46:48.729
  STEP: dump namespace information after failure @ 03/31/25 14:46:48.73
```

so using the following, before cluster deployment, fixes the problem, maybe the github CI bootstrap was updated behind the scene?
```
export DOCKER_API_VERSION=1.43
sudo systemctl restart docker
```

qinqon:
> interesting point when doing it from CNI (that didnt happen on the previous PR design) is that the prefix is important i.e
> 
> ```
> ... failed to find bridge mapping for network: "q2cl9_different-network"; Current ovn-bridge-mappings: "physnet:breth0,different-network:ovsbr1"
> ```
> 
> there is a random internal prefix that is added somewhere in the flow, in this example `q2cl9_`
> 
> There are 2 options:
> 
> 1. in the failing test i can either read the NAD and add according the actual network name create the mapping
> 2. to remove the prefix in CNI if exists (i.e remove all before `_`, including it)
> 
> wdyt?

First you need to understand what this prefix is.

oshoval:
@qinqon can you please review ?
need to see all test pass, but meanwhile if you want we can start the review
thanks

go-controller/pkg/cni/cni.go
```diff
@@ -345,6 +350,23 @@ func HandlePodRequest(
...
+	}
+
+	netName := pr.netName
+	if pr.CNIConf.PhysicalNetworkName != "" {
+		netName = pr.CNIConf.PhysicalNetworkName
+	}
+	if err := checkBridgeMapping(ovsClient, pr.CNIConf.Topology, netName); err != nil {
+		return nil, fmt.Errorf("failed bridge mapping validation: %w", err)
+	}
+
```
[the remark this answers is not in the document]
oshoval: we already discussed with Miguel and saw the `NewOVSClient` is the preferred way, as it has cache
there was a PR that adapted from the old method to the new method as far as i remember
also please note that in the previous review, Jaime was not against it, at most asked to reuse the ovsClient that was already created on other place 

but it means we will need to create a singleton fetcher for that, if they are in the same container, do we want this ?
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000/#discussion_r2000962389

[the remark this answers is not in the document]
qinqon: at least can we move those ovs operations to ovs.go ?

go-controller/pkg/cni/cni.go
```diff
@@ -420,3 +442,26 @@ func (pr *PodRequest) buildPodInterfaceInfo(annotations map[string]string, podAn
...
+	ovnBridgeMappings := openvSwitch.ExternalIDs["ovn-bridge-mappings"]
+
+	bridgeMappings := strings.Split(ovnBridgeMappings, ",")
+	for _, bridgeMapping := range bridgeMappings {
+		networkBridgeAssociation := strings.Split(bridgeMapping, ":")
+		if len(networkBridgeAssociation) == 2 && networkBridgeAssociation[0] == networkName {
+			return nil
+		}
+	}
```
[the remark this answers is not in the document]
oshoval: we dont need to use ovs.go please see https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145#discussion_r2022704248

unless you insist, i dont think we need intermediate conversion

go-controller/pkg/cni/cni.go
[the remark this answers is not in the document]
oshoval: i think it will mix the methods which isnt good
it can be confusing, even now we dont have them mixed where `NewOVSClient` is used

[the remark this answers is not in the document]
qinqon: To be honest I would just make this PR simple to fix the issue at hand and do like the other ovs operations, and if you want after merging it refactor all the ovs operations with ovs client, but whatever maintainers prefer.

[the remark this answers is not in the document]
oshoval: we already had agreement that this is the preferred method, i dont think we need to go back please

go-controller/pkg/cni/cni.go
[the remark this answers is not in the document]
qinqon: ack

go-controller/pkg/cni/cni.go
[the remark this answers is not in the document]
oshoval: > UPDATE: at privileged mode we use the CNI server that is running at ovnkube-node controllers

does it mean we should strive to create singleton then ?
what about unprivileged mode, if one mode works and the other isnt, it means that we either forget about it, or simply the singleton will run in two places, each one of them will init it in the worse case, which is fine

qinqon (dismissed):
LGTM

We have alreay review 99% of this.

go-controller/pkg/cni/cni.go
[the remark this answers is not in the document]
qinqon: It means that we may need to detect the mode and re-use the controller cache if we don't do that we end with a pair of ovs caches at privileged mode.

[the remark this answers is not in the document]
oshoval: what about having `once.Do(func() {`
and the rest is automatic, i.e:
if they both run from the same container, it will simply reuse the already created instance on follow-up calls,
otherwise each first call will init it, on different containers, without no need to detect the mode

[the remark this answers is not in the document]
oshoval: np thanks, once Jaime will review we can see if still needed or not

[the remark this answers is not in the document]
oshoval: @jcaamano @qinqon @maiqueb
wrt Jaime's comment
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5000#discussion_r2000962389

Do we still want to do it on this PR ?
considering what Quique said, that there is privileged and unprivileged mode,
we can create a struct with `once.Do`  (sync pkg), that initialize the client on the first run, and return it on subsequent ones.
It can be agnostic to the mode iiuc, because if it runs on two containers, each one will simply init the client for its own, otherwise the 2nd call onwards will just use the singleton

Thanks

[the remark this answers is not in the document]
maiqueb: I think we should do it now @oshoval

oshoval:
all tests passed, ptal (just rebased and added the 2nd commit to address [this](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145#discussion_r2031320764))

maiqueb (approved):
thank you. 

@jcaamano / @ricky-rav are you OK with these changes ?

oshoval:
On unpriviledge mode, there can be a subtle bug, that if the first time failed to init the client, we will reach a limbo state, where the pod is alive, but it always fail to get the ovsClient, because we use sync once.
Fixed it to allow reinit attempts in case of a failure

oshoval:
@maiqueb ptal due to the latest changes

ricky-rav (dismissed):
/LGTM
Thanks, Or!

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -561,16 +562,17 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 		}()
 	}
 
+	ovsClient, err := ovs.GetOVSClient(ctx)
```
jcaamano, line 565: Prefer if you pass this just like we do with everything else.

NewNodeControllerManager -> NewDefaultNodeNetworkController -> NewCNIServer. And form here you just pass ti around same as the clientset.

I don't have anything against singletons, but I would prefer not start popping them without some kind of central registry design or something. Otherwise its easy to lose track of them.

I think passing it around is the best approach for now.

jcaamano: Also I don't think you need to change anything relative to `cancel()`. Just leave as it was. If anything, you can remove the call there was online 568 which should be completely unnecessary.

jcaamano (changes requested):

jcaamano:
Also. Are the IC only support references in the commit & PR message accurate? I don't see why this wouldn't work in non-IC.

oshoval:
> Also. Are the IC only support references in the commit & PR message accurate? I don't see why this wouldn't work in non-IC.

interesting indeed, since now it is on CNI level, CNI is per node anyhow which is the thing we need

oshoval:
> Also. Are the IC only support references in the commit & PR message accurate? I don't see why this wouldn't work in non-IC.

Interesting, indeed, checking on non IC works, since it is from the CNI now
is it correct that on CNI it always runs on the specific node ? (i guess because it should enter its network ns, but best to ask if it is the case on all the sub configurations)
will update code / commits accordingly

oshoval:
addressed comments, lets see tests are passing
non IC is supported now, nice catch
thanks

oshoval:
all passed beside known flakes
changes as Jaime asked - support non IC (just removed the condition) and pass the client via functions instead of singleton

oshoval:
@jcaamano @maiqueb @qinqon can you please take a look ?
Thanks

test/e2e/multihoming.go
maiqueb, line 1: shouldn't we add a test entry that triggers the failure ? I'm somehow missing that.

oshoval: imo we don't need to test both bad and good path
we test the good path e2e, and the unit test tests the bad path

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -521,6 +521,12 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 		}()
 	}
 
+	ovsClient, err := libovsdb.NewOVSClient(ctx.Done())
+	if err != nil {
+		cancel()
+		return fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
+	}
+
```
jcaamano, line 529: This should only be created when `runMode.node`

The usage we had below for metrics seems to only be IC compatible.

I would either return an error if metrics have been requested on non-node configuration or alternatively not start the metric server or not register the metrics in that case.

But it seems a somewhat incorrect metrics design. @martinkennelly is there anything I am missing here?

test/e2e/multihoming.go
maiqueb: I think it would be a good test to add, but, at least it's covered in unit tests. 

Up to the maintainer to say if this is really needed.

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/27c795586501edd4c8627e0e9cb8ec721fa2343b..2fe493dd1b64bb547fc71cf824f40631095bad67):
- Init. OVS client when only when necessary, resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145#pullrequestreview-2787344038 
- In the CNI bridge-mapping check, remove existing node bridge-mapping from the error message in order to avoid exposing this info to non-admin users, as the error is propagated to pod event from kubelet any user can have access to it (using kubectl describe pod..)

@jcaamano @martinkennelly PTAL

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/27c795586501edd4c8627e0e9cb8ec721fa2343b..6878d61eec42ccbeb0785149ff0fe6e35af70e89):
- Initialize OVS client when needed, resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145#pullrequestreview-2787344038
- Avoid exposing the existing nodes OVN bridge-mapping in CNI error message, as the error eventually propogated as pod event any user can access (using kubectl describe pod ..)

go-controller/pkg/cni/cni.go
```diff
@@ -420,3 +435,26 @@ func (pr *PodRequest) buildPodInterfaceInfo(annotations map[string]string, podAn
...
+	klog.Errorf("Failed to find bridge mapping for network: %q, current OVN bridge-mappings: (%s)", networkName, ovnBridgeMappings)
```
jcaamano, line 458: nit: logging should happen top level. Difficult for me to tell if this is not being logged twice.

ormergi: Unfortunately the existing bridge mapping info exist in this context, this is why I added the log message here.
I checked how logs will look in case of an error, ovnkube-contoller logs will have the following messages:
```
E0424 15:49:43.805811  117892 cni.go:458] Failed to find bridge mapping for network: "mylocalnet", current OVN bridge-mappings: (physnet:breth0)
I0424 15:49:43.805823  117892 cni.go:348] [default/test 9acfbd9f1e5a4e5c6ada7e352cb43fc16ec03c8c3bbcf9c46894bfffddcdd074 network localnet NAD default/localnet] ADD finished CNI request [default/test 9acfbd9f1e5a4e5c6ada7e352cb43fc16ec03c8c3bbcf9c46894bfffddcdd074 network localnet NAD default/localnet], result "", err failed bridge mapping validation: failed to find OVN bridge-mapping for network: "mylocalnet"
```
First one the log message in this scope.
The second message comes from the CNI top-level log message, which doesnt include the node info.

jcaamano: I suggest to include the existing bridge mappings on the error or if that is too verbose then keep the log but change it to v(5)?

go-controller/cmd/ovnkube/ovnkube.go
martinkennelly: whats metrics are you referring to that are ic only?

go-controller/pkg/cni/cni.go
ormergi: I am trying to prevent having the nodes bridge-mapping in the error message, because the emitted pod event message derives from that error including the node info, which is accesiable for non-admin users.

I changed the log level to v(5) as you suggest.

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/6878d61eec42ccbeb0785149ff0fe6e35af70e89..5ac9a14bec3be41edb8b7308302f581c7193e7bc): resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5145#discussion_r2058714603
@jcaamano PTAL :pray:

jcaamano (approved):

PR #5146, opened 2025-03-24
Convert to DaemonSet and fix update strategy of ovnkube-identity pods

crnithya:
Ovnkube-identity pods should run on all control-plane nodes and require that a newly created pod is fully functional before the old one can be terminated and hence require additional pods during upgrades

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/dist/templates/ovnkube-identity.yaml.j2
+++ b/dist/templates/ovnkube-identity.yaml.j2
@@ -1,7 +1,7 @@
 # ovnkube-identity
 # starts ovnkube-identity
 # it is run on the master(s).
-kind: Deployment
+kind: DaemonSet
 apiVersion: apps/v1
 metadata:
   name: ovnkube-identity
@@ -9,18 +9,16 @@ metadata:
   namespace: ovn-kubernetes
   annotations:
     kubernetes.io/description: |
-      This Deployment launches the ovnkube-identity networking component.
+      This DaemonSet launches the ovnkube-identity networking component on control-plane nodes.
 spec:
-  progressDeadlineSeconds: 600
-  replicas: {{ ovn_master_count | default(1|int) }}
   revisionHistoryLimit: 10
   selector:
     matchLabels:
       name: ovnkube-identity
-  strategy:
+  updateStrategy:
     rollingUpdate:
-      maxSurge: 0
-      maxUnavailable: 1
+      maxSurge: 100%
+      maxUnavailable: 0
     type: RollingUpdate
   template:
     metadata:
@@ -35,20 +33,9 @@ spec:
       serviceAccountName: ovnkube-identity
       hostNetwork: true
       dnsPolicy: Default
-
-      # required to be scheduled on a linux node with node-role.kubernetes.io/control-plane label and
-      # only one instance of ovnkube-control-plane pod per node
-      affinity:
-        nodeAffinity:
-          requiredDuringSchedulingIgnoredDuringExecution:
-            nodeSelectorTerms:
-              - matchExpressions:
-                  - key: node-role.kubernetes.io/control-plane
-                    operator: Exists
-                  - key: kubernetes.io/os
-                    operator: In
-                    values:
-                      - "linux"
+      nodeSelector:
+        node-role.kubernetes.io/control-plane: ""
+        kubernetes.io/os: "linux"
       containers:
       - name: ovnkube-identity
         image: "{{ ovn_image | default('docker.io/ovnkube/ovn-daemonset:latest') }}"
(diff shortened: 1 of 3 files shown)


crnithya:
@kyrtapz, @flavio-fernandes PTAL

flavio-fernandes (approved):
/lgtm
TY @crnithya !!!

cathy-zhou:
lgtm

crnithya:
Thank you @cathy-zhou and @flavio-fernandes

kyrtapz (approved):
Thanks! this should help with upgrades upstream.

crnithya:
Thank you @kyrtapz

flavio-fernandes:
Flake:  https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4936
```
  [FAIL] e2e egress IP validation Network Segmentation: IPv4 L2 role primary [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts [It] disabling egress nodes impeding GRCP health check
```

dist/templates/ovnkube-identity.yaml.j2
```diff
@@ -19,8 +19,8 @@ spec:
       name: ovnkube-identity
   strategy:
     rollingUpdate:
-      maxSurge: 0
-      maxUnavailable: 1
+      maxSurge: 100%
```
npinaeva, line 22: can you please explain a bit more about the upgrade scenario for 3 control-plane node case? afaiu ovnkube-identity is running on all control-plane nodes with its own leaderelection, so do we want to have 6 copies of the pod and then start deleting the old ones?

dist/templates/ovnkube-identity.yaml.j2
```diff
@@ -49,6 +49,17 @@ spec:
                     operator: In
                     values:
                       - "linux"
+        podAntiAffinity:
```
npinaeva, line 52: was that part always supposed to be here, but we never noticed because we only use single-master deployments?

flavio-fernandes: Yes, that is very plausible. When this is deployed via CNO, we have this section when replicas is greater than 1, but it does not hurt to always have it:
https://github.com/openshift/cluster-network-operator/blob/1daca87a5c6ebd5e6d7b5bcc4374119b59061af6/bindata/network/node-identity/managed/node-identity.yaml#L66-L81

crnithya:
@kyrtapz, I do have follow up question. Like @npinaeva mentioned, shouldn't the kube-id pod be running on all control-plane nodes? Being a deployment, we can set kube-id replica count to be equal to control-plane nodes and together with podanitiaffinity we can achieve this, but why isn't this a daemonset?
The new update strategy also has issues if the replica count is less than control-plane nodes, then during upgrades, the excess pod from maxsurge could land on the node which doesn't have kube-id pod running currently and another one previously running could be stuck in "terminating" state because it requires a working kube-id pod on that node.

kyrtapz:
> @kyrtapz, I do have follow up question. Like @npinaeva mentioned, shouldn't the kube-id pod be running on all control-plane nodes? Being a deployment, we can set kube-id replica count to be equal to control-plane nodes and together with podanitiaffinity we can achieve this, but why isn't this a daemonset? The new update strategy also has issues if the replica count is less than control-plane nodes, then during upgrades, the excess pod from maxsurge could land on the node which doesn't have kube-id pod running currently and another one previously running could be stuck in "terminating" state because it requires a working kube-id pod on that node.

The webhook has to run on every control-plane node so it makes sense to make it a daemonset, the existing deployment has the replica count set to the  control-plane node count so it is ok for setups where the count of those doesn't change but I agree we can make it a daemonset to handle cases where new ctrl nodes are added day-2.

kyrtapz:
> @kyrtapz I see we actually use Daemonset for self-hosted setup here https://github.com/openshift/cluster-network-operator/blob/1daca87a5c6ebd5e6d7b5bcc4374119b59061af6/bindata/network/node-identity/self-hosted/node-identity.yaml#L1, but deployment for managed, and I can't figure out why we use deployment upstream when it is closer to the self-hosted setup. Please help

Managed is a special case where the webhook is a cluster-networked pod running on a different cluster so it doesn't have the same requirement of being colocated with the kapi pod as standalone deployments do.

crnithya:
@kyrtapz, Can I go ahead and change this to daemonset then?

crnithya:
@kyrtapz @npinaeva @flavio-fernandes @cathy-zhou PTAL, I reworked this to convert to daemonset per our discussion. Thank you

kyrtapz:
> @kyrtapz @npinaeva @flavio-fernandes @cathy-zhou PTAL, I reworked this to convert to daemonset per our discussion. Thank you

Thank you @crnithya! LGTM

crnithya:
Thank you @kyrtapz and @flavio-fernandes

npinaeva:
that is very clean fix, thanks all!

PR #5150, opened 2025-03-26
kind: Remove passt

qinqon:
## 📑 Description
We are in the process of removing passt from ipam controller we should remove it from kind.sh


maiqueb (approved):
Thank you for unblocking the lanes.

jcaamano (approved):

PR #4069, opened 2024-01-03
Make pod to host access route via overlay

trozet:
We have had a myriad of issues where pods trying to reach kubernetes hosts run into problems, due to this traffic being treated as north/south egress. Some of these problems include trying to avoid egress features like egress services, or egress IP when trying to talk to another kubernetes nodes. For all intents and purposes, an ovn networked pod trying to communicate with a host network pod or process on another node can be treated as internal traffic to the cluster.

With this patch, logical route policies are added to ovn_cluster_router that reference and address set for all known host ips. Now if any pod is trying to reach a known IP on a kubernetes node, it will be routed via geneve to the ovn-k8s-mp0 port on that node. This is true for OVN IC or traditional deployments.

PR #5153, opened 2025-03-27
Guard against PMTUD route caching for Kubernetes Node IPs

trozet:
There are several traffic paths that can trigger a k8s node to get an icmp needs frag/packet too big message and install a lower MTU cached route in the kernel. This is occurs due to paths where the geneve tunnel is going to be traversed (100 bytes lower than underlay). Once this cached route is installed, it can have detrimental effects to the entire cluster.

An attempt to solve this by forcing all node destined traffic to go via geneve was done here:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4069

However that PR is quite complex, and there are issues around upgrade. Not to mention potential regressions with changing traffic patterns with OVN-Kubernetes.

~~This PR offers a more conservative solution. For k8s nodes that are all on the same network, we can simply install /32 or /128 routes to each node, with a special attribute "mtu lock" on the route. This ensures that:~~

~~"If the modifier lock is used then no path MTU discovery will be performed and all the packets will be sent without the DF bit set for the IPv4 case or fragmented to the MTU for the IPv6 case."~~

~~Additionally, needs frag packets received from other nodes will not trigger a cached route to be installed.~~

Turns out that with IPv6 MTU lock does not prevent packet too big messages from installing cached routes. Therefore using mtu lock was abandoned, and a pure nftables approach now being used.

For k8s nodes that are not on the same network (separated by another router), we cannot assume that there wont be an MTU drop in between the k8s nodes. For example:

node A -------1500 MTU-----router ----------1400 MTU ------router ------1500 MTU ----node B

In this case a PMTUD message from one of the routers is legitimate, and we should lower the MTU in that case. However, it is still illegitimate for one of the k8s nodes to send a PMTUD message in this scenario. Therefore to avoid the k8s node from sending PMTUD messages in the separate subnet case, we install nftables rules that drop those packets on egress to known node IPs. 

E2Es are added that cover 3 of the paths that trigger a lower MTU cached route. ~~netlink library is updated to support MTU lock. Additionally, some logic is added to routeManager to label routes OVNK installs with a unique protocol, so that we can effectively remove stale routes that exist after OVNK was restarting.~~

Update: will move the routeManager label/cleanup to another PR @martinkennelly 

Downstream bugs:
https://issues.redhat.com/browse/OCPBUGS-44285
https://issues.redhat.com/browse/OCPBUGS-52503
https://issues.redhat.com/browse/OCPBUGS-37733


go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1440,6 +1448,116 @@ func (nc *DefaultNodeNetworkController) WatchNamespaces() error {
...
+				return err
+			}
+		}
+		// TODO(trozet): else case...create nftables rules to block egress of ICMP needs frag
+
+	}
+	return nil
+}
+
+func (nc *DefaultNodeNetworkController) deleteNode(node *corev1.Node) {
```
martinkennelly, line 1562: wheres the sync func upon restart? how you handling missed Node delete events?

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -652,4 +662,365 @@ var _ = Describe("Node", func() {
...
+
+		Context("with a cluster in IPv4 mode", func() {
+			const (
+				ethName     string = "eth0"
+				nodeIP      string = "169.254.254.60"
+				ethCIDR     string = nodeIP + "/24"
+				otherNodeIP string = "169.254.254.61"
+				fullMask           = 32
+			)
+			// var eth0MAC string
```
martinkennelly, line 719: clean

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -652,4 +662,365 @@ var _ = Describe("Node", func() {
...
+			var link netlink.Link
+
+			BeforeEach(func() {
+				config.IPv4Mode = true
+				config.IPv6Mode = false
+
+				// Note we must do this in default netNS because
+				// nc.WatchNodes() will spawn goroutines which we cannot lock to the testNS
```
martinkennelly, line 727: hmm.. youre creating a link on the test network namespace

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -652,4 +662,365 @@ var _ = Describe("Node", func() {
...
+				ovntest.AddLink(ethName)
```
martinkennelly, line 728: cleanup?

martinkennelly: isnt the link name really common and you could hit a name clash with existing link

go-controller/pkg/node/routemanager/route_manager.go
```diff
@@ -202,6 +238,7 @@ func (c *Controller) applyRoute(link netlink.Link, gwIP net.IP, subnet *net.IPNe
 		netlinkRoute.Src = src
 		netlinkRoute.Gw = gwIP
 		netlinkRoute.MTULock = mtuLock
+		netlinkRoute.Protocol = OVNKRouteProtocol
```
martinkennelly, line 241: NICEE!! I have been wondering how to track these routes and never thought of this.

I looked at netlink code and i think this maps to the following:
```
/* rtm_protocol */

#define RTPROT_UNSPEC        0
#define RTPROT_REDIRECT        1        /* Route installed by ICMP redirects;
                                   not used by current IPv4 */
#define RTPROT_KERNEL        2        /* Route installed by kernel                */
#define RTPROT_BOOT        3        /* Route installed during boot                */
#define RTPROT_STATIC        4        /* Route installed by administrator        */

/* Values of protocol >= RTPROT_STATIC are not interpreted by kernel;
   they are just passed from user and back as is.
   It will be used by hypothetical multiple routing daemons.
   Note that protocol values should be standardized in order to
   avoid conflicts.
 */
```

martinkennelly: Therefore value 99 (OVNKRouteProtocol) is ignored by kernel.

go-controller/pkg/node/default_node_network_controller.go
martinkennelly: ah its in a later commit with route manager.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1304,6 +1304,10 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
 		ovspinning.Run(nc.stopChan)
 	}()
 
+	if err := nc.routeManager.CleanupStaleRoutes(); err != nil {
```
martinkennelly, line 1319: as long as route manager is correctly programmed we wont have disruption right? but how do you know it is correctly programmed here by all route manager consumers? e.g. EIP controller for multi NIC.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1440,6 +1464,156 @@ func (nc *DefaultNodeNetworkController) WatchNamespaces() error {
...
+				return fmt.Errorf("unable to list host routes: %w", err)
+			}
+			linkIndex := -1
+			existingRoute := netlink.Route{}
+			for _, route := range routes {
+				if existingHostSubnet.Contains(route.Dst.IP) {
+					linkIndex = route.LinkIndex
+					existingRoute = route
+					break
+				}
```
martinkennelly, line 1515: can you stop programming the route if an existing route exists that matches what we want?

go-controller/go.mod
```diff
@@ -148,4 +148,5 @@ replace (
 	github.com/coreos/go-iptables => github.com/trozet/go-iptables v0.0.0-20240328221912-077e672b3808
 	github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
 	github.com/j-keck/arping => github.com/JacobTanenbaum/arping v0.0.0-20240209152419-3987db83bd51
+	github.com/vishvananda/netlink => github.com/trozet/netlink v0.0.0-20250326135057-2f783e771ab6
```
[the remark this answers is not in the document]
trozet: wow that was fast, will do!

go-controller/pkg/node/default_node_network_controller_test.go
trozet: yeah oops, i should make it some dummy loopback interface...will change it

go-controller/pkg/node/routemanager/route_manager.go
trozet: yeah I thought of it cause i saw that FRR/Quagga will program routes with protocol as "bgp" or "quagga" sometimes. Seems like this is a safe method for us to use, as long as no one else uses 99.

go-controller/pkg/node/default_node_network_controller.go
trozet: My assumption was by the time we get here all the watchers have started up, so kapi objects have been processed, and now it is safe to run the cleanup. Is that not true?

go-controller/pkg/node/default_node_network_controller.go
trozet: hmm i guess i could use routemanager equal method to see if they are the same and then skip processing it if so. That compares mtu lock as well. I'll look into it.

trozet:
> For the scenario where we have a network inbetween the k8 nodes with a different MTU and one of the routes in that network sends an ICMP, the route on the node should be updated right but how is that done when the mtu is locked? missing that.

For nodes in a different subnet we dont add an "mtu lock" route. We add the nftables rules instead. So imagine our node is 172.18.0.2/24, and there are 2 other nodes 172.18.0.3/24 and 172.19.0.3/24. It would look like this:

ip route add 172.18.0.3/32 dev eth0 mtu lock 1500
nft add element inet ovn-kubernetes no-pmtud-remote-node-ips-v4 { 172.19.0.3 }

martinkennelly:
> > For the scenario where we have a network inbetween the k8 nodes with a different MTU and one of the routes in that network sends an ICMP, the route on the node should be updated right but how is that done when the mtu is locked? missing that.
> 
> For nodes in a different subnet we dont add an "mtu lock" route. We add the nftables rules instead. So imagine our node is 172.18.0.2/24, and there are 2 other nodes 172.18.0.3/24 and 172.19.0.3/24. It would look like this:
> 
> ip route add 172.18.0.3/32 dev eth0 mtu lock 1500
> nft add element inet ovn-kubernetes no-pmtud-remote-node-ips-v4 { 172.19.0.3 }

I meant a router in the infra providers cluster networks sends an ICMP back. Thats legitimate right and we should update the route MTU but I am wonder how does that happen when the route has a MTU lock attribute specified.

trozet:
So upon further testing I found that mtu lock doesn't always work. For example with IPv6, even with mtu lock, the kernel will install a cached route and lower MTU when receiving packet too big ICMPv6 message:

[root@ovn-worker2 ~]# ip -6 route show cache
fc00:f853:ccd:e793::4 dev breth0 metric 1024 expires 592sec mtu lock 1400 pref medium
[root@ovn-worker2 ~]# ip -6 route show
fc00:f853:ccd:e793::3 dev breth0 proto 201 metric 1024 mtu lock 1500 pref medium
fc00:f853:ccd:e793::4 dev breth0 proto 201 metric 1024 mtu lock 1500 pref medium

According to some documentation online, this is intentional because IPv6 doesn't support fragmentation, so it has to respect the PMTUD message.

Therefore I think we need to abandon relying on MTU lock, and use the nftables method that is implemented for remote nodes for local nodes as well. Will push a follow up.

go-controller/pkg/node/node_nftables.go
```diff
@@ -0,0 +1,100 @@
...
+// Relies on the sets from setupPMTUDNFTSets.
+func setupPMTUDNFTChain() error {
+	counterIfDebug := ""
+	if config.Logging.Level > 4 {
+		counterIfDebug = "counter"
+	}
+
+	nft, err := nodenft.GetNFTablesHelper()
+	if err != nil {
+		return err
```
martinkennelly, line 52: failed to get ..

go-controller/pkg/node/obj_retry_node.go
```diff
@@ -147,13 +163,20 @@ func (h *nodeEventHandler) GetResourceFromInformerCache(key string) (interface{}
...
+func (h *nodeEventHandler) AddResource(obj interface{}, _ bool) error {
 	switch h.objType {
 	case factory.NamespaceExGwType,
 		factory.EndpointSliceForStaleConntrackRemovalType:
 		// no action needed upon add event
 		return nil
 
+	case factory.NodeType:
+		node := obj.(*corev1.Node)
+		if node.Name == h.nc.name {
```
martinkennelly, line 175: nice ! add reasoning why we are skipping for future readers?

go-controller/pkg/node/obj_retry_node.go
```diff
@@ -187,6 +210,31 @@ func (h *nodeEventHandler) UpdateResource(oldObj, newObj interface{}, _ bool) er
...
+				}
+				if err := h.nc.addOrUpdateNode(node); err != nil {
+					return fmt.Errorf("error updating node %s: %w", node.Name, err)
+				}
+			}
+			return nil
+		}
+
+		// remote node that is changing
+		h.nc.deleteNode(oldNode)
```
martinkennelly, line 235: why delete first? couldnt that cause us to leak out messages?

go-controller/pkg/ovn/master_test.go
```diff
@@ -335,6 +335,9 @@ func addNodeLogicalFlowsHelper(testData []libovsdbtest.TestData, expectedOVNClus
 		MAC:            node.NodeLRPMAC,
 		Networks:       []string{node.NodeGWIP},
 		GatewayChassis: []string{chassisName + "-UUID"},
+		Options: map[string]string{
+			"gateway_mtu": "1400",
```
martinkennelly, line 339: why 1400? Can it change during the life time of the cluster?

test/e2e/pod.go
```diff
@@ -357,3 +364,445 @@ var _ = ginkgo.Describe("Pod to pod TCP with low MTU", func() {
...
+		echoServicePortMax        = 31299
+	)
+
+	var ipCmd = []string{"ip"}
+	var cs clientset.Interface
+	var echoMtuRegex = regexp.MustCompile(`expires.*mtu.*1400`)
+	f := wrappedTestFramework("icmp-needs-frag")
+	cleanupFn := func() {
+		ovnKubeNodePods, err := f.ClientSet.CoreV1().Pods(ovnNamespace).List(context.TODO(), metav1.ListOptions{
+			LabelSelector: "name=ovnkube-node",
```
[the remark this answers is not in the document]
trozet: i'll fix it now

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1440,6 +1460,95 @@ func (nc *DefaultNodeNetworkController) WatchNamespaces() error {
...
+				Set: types.NFTNoPMTUDRemoteNodeIPsv6,
+				Key: []string{nodeIP.String()},
+			})
+		}
+	}
+
+	gw := nc.Gateway.(*gateway)
+	gw.openflowManager.updateBridgePMTUDFlowCache(getPMTUDKey(node.Name), addrs)
+
+	if len(nftElems) > 0 {
```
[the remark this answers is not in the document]
trozet: hmm so UpdateNFTElements only will add or update a given element, it wont remove nft config. But you made me think...what happens if the nodeIP changes? I think I cover that in the node update logic where we delete then re-add:
```
		// remote node that is changing
		h.nc.deleteNode(oldNode)
		return h.nc.addOrUpdateNode(newNode)
```

go-controller/pkg/node/default_node_network_controller.go
trozet: actually now that i pulled out route manager stuff, i realize i need a sync for nftables with pmtud...i'll add it. Thanks for catching it

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1440,6 +1460,95 @@ func (nc *DefaultNodeNetworkController) WatchNamespaces() error {
...
+		}
+		nodeIP := net.ParseIP(address.Address)
+		if nodeIP == nil {
+			continue
+		}
+		mask := 32
+		if utilnet.IsIPv6(nodeIP) {
+			mask = 128
+		}
+		_, nodeCIDR, err := net.ParseCIDR(fmt.Sprintf("%s/%d", nodeIP, mask))
```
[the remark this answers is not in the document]
trozet: I see there is already a util function: GetIPNetFullMaskFromIP, but looking back at the code... I don't think I even need this anymore now that I removed the route manager stuff. Will fix.

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -652,4 +678,566 @@ var _ = Describe("Node", func() {
...
+				err = netlink.LinkSetUp(link)
+				Expect(err).NotTo(HaveOccurred())
+
+				// Add an IP address
+				addr, err := netlink.ParseAddr(ethCIDR)
+				Expect(err).NotTo(HaveOccurred())
+				addr.Scope = int(netlink.SCOPE_UNIVERSE)
+				err = netlink.AddrAdd(link, addr)
+				Expect(err).NotTo(HaveOccurred())
+				fmt.Print(netlink.RouteList(link, netlink.FAMILY_V6))
```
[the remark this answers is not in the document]
trozet: leftover i think from the route manager stuff. will remove

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -652,4 +678,566 @@ var _ = Describe("Node", func() {
...
+					err = nc.WatchNodes()
+					Expect(err).NotTo(HaveOccurred())
+					nftRules := v4PMTUDNFTRules + `
+add element inet ovn-kubernetes no-pmtud-remote-node-ips-v4 { 169.254.253.61 }
+`
+					err = nodenft.MatchNFTRules(nftRules, nft.Dump())
+					Expect(err).NotTo(HaveOccurred())
+					gw := nc.Gateway.(*gateway)
+					By("start up should add openflow rules for remote node")
+					flows := gw.openflowManager.getFlowsByKey(getPMTUDKey(remoteNodeName))
```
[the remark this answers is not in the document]
trozet: No I don't think it is. nc.WatchNodes() wont return until all initial adds are done, and here i'm checking that the flows are in the cache, not rendered into OVS. So the add should have at least added everything into the cache by now.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2128,6 +2143,21 @@ func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]strin
...
+func pmtudDropFlows(bridge *bridgeConfiguration, ipAddrs []string) []string {
+	var flows []string
+	if config.Gateway.Mode != config.GatewayModeShared {
+		return nil
+	}
+	for _, addr := range ipAddrs {
+		for _, netConfig := range bridge.patchedNetConfigs() {
+			flows = append(flows,
+				generateICMPFragmentationFlow(addr, outputPortDrop, netConfig.ofPortPatch, pmtudOpenFlowCookie, 700))
+		}
```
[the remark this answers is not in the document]
trozet: I wanted it to be higher than 500, which is the highest we use right now for host<->OVN traffic. I don't think it should hurt for this to have higher priority, PMTUD to known k8s node addresses should always be dropped, so I made it highest priority.

go-controller/pkg/node/node_nftables.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/knftables"
+
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	nodenft "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/nftables"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+)
+
+const nftPMTUDChain = "no-pmtud"
+
+// setupPMTUDNFTSets sets up the NFT sets that contain remote node IPs not in the local node's subnet
```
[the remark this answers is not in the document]
trozet: sorry, this was leftover from when i was trying to do the 2 pronged approach with mtu lock on routes, will fix it

go-controller/pkg/node/obj_retry_node.go
trozet: no real reason from a functional perspective. I guess from a code perspective delete does not throw error, so we can at least delete and clean up if addOrUpdate errors out.

go-controller/pkg/ovn/master_test.go
martinkennelly: LOL

go-controller/pkg/node/node_nftables.go
```diff
@@ -0,0 +1,100 @@
...
+// Relies on the sets from setupPMTUDNFTSets.
+func setupPMTUDNFTChain() error {
+	counterIfDebug := ""
+	if config.Logging.Level > 4 {
+		counterIfDebug = "counter"
+	}
+
+	nft, err := nodenft.GetNFTablesHelper()
+	if err != nil {
+		return fmt.Errorf("failed to get nftables helper")
```
martinkennelly, line 52: include the error message

test/e2e/pod.go
```diff
@@ -204,7 +211,7 @@ var _ = ginkgo.Describe("Pod to external server PMTUD", func() {
 						// bug: https://issues.redhat.com/browse/OCPBUGS-7609.
 						// TODO: Revisit this once https://bugzilla.redhat.com/show_bug.cgi?id=2169839 is fixed.
 						ovnKubeNodePods, err := f.ClientSet.CoreV1().Pods(ovnNamespace).List(context.TODO(), metav1.ListOptions{
-							LabelSelector: "name=ovnkube-node",
+							LabelSelector: "app=ovnkube-node",
```
martinkennelly, line 214: thanks - helps downstream :)

test/e2e/service.go
```diff
@@ -395,7 +404,26 @@ var _ = ginkgo.Describe("Services", func() {
 				//                0.0.0.0/0                172.18.0.1 dst-ip rtoe-GR_ovn-worker
 				// This time, the packet will leave the rtoj port and it will be fragmented.
 				ginkgo.It("queries to the nodePort service shall work for UDP", func() {
-					for _, size := range []string{"small", "large"} {
+					packetSizes := []string{"small", "large"}
+					// If gateway mode is shared, and endpoint is OVN networked, host networked originated packets
+					// exceeding pod MTU will not be delivered. This is because ICMP needs frag will be sent back to the original
+					// OCP node by OVN (even if DF bit is not set) and the node will refuse to install an MTU cache route.
```
martinkennelly, line 410:
```suggestion
					//  Kubernetes Node by OVN (even if DF bit is not set) and the node will refuse to install an MTU cache route.
```

go-controller/pkg/node/obj_retry_node.go
trozet: done

martinkennelly:
/lgtm

but conditional on ci - its bricked it seems

trozet:
going to try to rebase it to see if that fixes CI

trozet:
rebased, there was a trivial conflict I had to fix. nothing tangible changed from previous lgtm

PR #5159, opened 2025-04-01
Copy OVS RPMs to be installed when building from source

kyrtapz:
Use the OVS RPMs built from source.
cc @dceara


dceara (approved):
Looks good to me, thanks!

@qinqon WDYT?

qinqon:
@kyrtapz did you tested ? we don't have CI for it.

kyrtapz:
> @kyrtapz did you tested ? we don't have CI for it.

Yes, I've tested it locally. 
Without this change there is no point of building the OVS RPMs if we are not going to use them.

jcaamano (approved):

qinqon (approved):
LGTM

trozet (approved):

PR #5160, opened 2025-04-01
Add allow-stateless ACLs for advertised L2 UDNs

kyrtapz:
Based on: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5130

With BGP layer2 networks we advertise the whole subnet from every node, this naturally causes asymmetric routing which breaks in local gateway mode due to OVN dropping packets with an invalid conntrack state. 

Example flow:
```mermaid
flowchart TD    
    External ~~~  ovn-worker
    External ~~~ ovn-worker2
    subgraph ovn-worker2
        C1[eth0] --> C2[mpX- DROP]
    end
    subgraph ovn-worker
        A1[pod] --> A2[mpX]
        A2 --> A3[eth0]
    end
    A3 -- SYN --> External
    External -- SYN/ACK --> C1
```

OVN drops the traffic due to allow-related ACLs that cause the assymetric reply to be commited to conntrack and then dropped since the state is invalid:
```
recirc_id(0x1f50),in_port(9),ct_state(-new-est-rpl+inv+trk),ct_mark(0/0x1),eth(),eth_type(0x0800),ipv4(frag=no), packets:2, bytes:148, used:0.712s, flags:S., actions:drop
```

To work around the issue this PR is adding two `allow-stateless` ACLs that avoid committing to conntrack for the traffic going to and from pod subnet  through the management port.

**The remaining concern is that this change will affect network policy enforcement, I've picked the priority that is below our default deny to avoid that but I would like some extra eyes on this.** 

kudos to @dceara for all the help with this!

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller.go
+++ b/go-controller/pkg/ovn/base_network_controller.go
@@ -746,6 +746,14 @@ func (bnc *BaseNetworkController) syncNodeManagementPort(node *corev1.Node, swit
 	var v4Subnet *net.IPNet
 	addresses := macAddress.String()
 	mgmtPortIPs := []net.IP{}
+
+	if config.Gateway.Mode == config.GatewayModeLocal &&
+		bnc.TopologyType() == types.Layer2Topology &&
+		util.IsPodNetworkAdvertisedAtNode(bnc.GetNetInfo(), node.Name) {
+		if err := bnc.addNetworkSubnetAllowACL(node.Name, switchName, hostSubnets); err != nil {
+			return nil, err
+		}
+	}
 	for _, hostSubnet := range hostSubnets {
 		mgmtIfAddr := util.GetNodeManagementIfAddr(hostSubnet)
 		addresses += " " + mgmtIfAddr.IP.String()
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -373,10 +373,6 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements) 
 			return nil, nil, fmt.Errorf("%w: selected network %q has unsupported topology %q", errConfig, networkName, network.TopologyType())
 		}
 
-		if config.Gateway.Mode == config.GatewayModeLocal && network.TopologyType() == types.Layer2Topology {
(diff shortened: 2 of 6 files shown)


kyrtapz:
frr failed to start in the sgw bgp [job](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14198816035/job/39827826470?pr=5160)
```
2025-04-02T08:00:11.4736206Z Error response from daemon: No such container: frr
2025-04-02T08:00:11.4748848Z + docker run -d --privileged --network kind --rm --ulimit core=-1 --name frr --volume /tmp/frr-GKQuyYgfhB:/etc/frr quay.io/frrouting/frr:9.1.2
2025-04-02T08:00:11.4852843Z Unable to find image 'quay.io/frrouting/frr:9.1.2' locally
2025-04-02T08:00:12.6656341Z docker: Error response from daemon: received unexpected HTTP status: 502 Bad Gateway.
2025-04-02T08:00:12.6656826Z See 'docker run --help'.
```

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -746,6 +746,14 @@ func (bnc *BaseNetworkController) syncNodeManagementPort(node *corev1.Node, swit
 	var v4Subnet *net.IPNet
 	addresses := macAddress.String()
 	mgmtPortIPs := []net.IP{}
+
+	if config.Gateway.Mode == config.GatewayModeLocal &&
+		bnc.TopologyType() == types.Layer2Topology &&
+		util.IsPodNetworkAdvertisedAtNode(bnc.GetNetInfo(), node.Name) {
+		if err := bnc.addNetworkSubnetAllowACL(node.Name, switchName, hostSubnets); err != nil {
```
kyrtapz, line 753: if we go with this approach this has to be cleaned up if the network is moved to not advertised.

maiqueb:
Does this fix have an associated issue we can track ? @kyrtapz

kyrtapz:
> Does this fix have an associated issue we can track ? @kyrtapz

This PR is one of the ways we can enable L2 advertised networks in LGW. We are likely not going to go with it due to the fact that it affects ETP=Local services and we might initially do go with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5351 instaed. 
There are no upstream specific issue tracking this work.

kyrtapz:
Closed in favor of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5351.

PR #5164, opened 2025-04-03
node, metrics: Replace depracated option with equivalent

almusil:
The option "ovn-openflow-probe-interval" is ignored from ovn24.03. Replace it with "ovn-bridge-remote-probe-interval" which serves similar purpose with the difference that from 24.09 it affects all connections to br-int, not only the main thread.

At the same time change the default to 0 which is disabled, the inactivity probe shouldn't be needed for connections over unix socket.


almusil:
cc @dceara

dceara (approved):
Looks good to me, thanks!

go-controller/pkg/config/config.go
```diff
@@ -67,7 +67,7 @@ var (
 		EncapIP:                      "",
 		EncapPort:                    DefaultEncapPort,
 		InactivityProbe:              100000, // in Milliseconds
-		OpenFlowProbe:                180,    // in Seconds
+		OpenFlowProbe:                0,      // in Milliseconds
```
flavio-fernandes, line 70: NIT: should we rename this config to `BridgeRemoteProbe` ?

flavio-fernandes (commented):
/lgtm @almusil !
While changing the value from 180 to 0 you mention that it does not matter for unix socket, right? Is that the case because you are assuming that we are using ovn-ic?

almusil:
> /lgtm @almusil ! While changing the value from 180 to 0 you mention that it does not matter for unix socket, right? Is that the case because you are assuming that we are using ovn-ic?

Hi @flavio-fernandes,

ovn-k is using other type of connection to br-int on non-ic deployments? This affects only br-int connection from ovn-controller.

trozet (approved):

PR #5165, opened 2025-04-04
Fix EIP "noreroute" LRPs when network contains multiple subnets of the same IP fa…

martinkennelly:
…mily

Previous to this change, the code made an incorrect assumption that there would be one subnet per IP family.

KinD only supports 1 IP per family therefore no E2Es update.

I am not happy with this PR and its clear the LRP DB IDs should be refactored but I am time constrained.

/hold until QE and downstream test.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -3121,7 +3121,7 @@ func createDefaultNoRerouteServicePolicies(nbClient libovsdbclient.Client, netwo
 		}
 		dbIDs := getEgressIPLRPNoReRoutePodToJoinDbIDs(IPFamilyValueV4, network, controller)
 		match := fmt.Sprintf("ip4.src == %s && ip4.dst == %s", v4Subnet.String(), v4JoinSubnet.String())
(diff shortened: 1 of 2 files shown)


martinkennelly:
Thanks to jaime for the reorientation with this PR. Made it simpler. I am writing a unit test now.

martinkennelly:
KinD doesnt allow multi cluster subnets of the same IP type therefore we must rely on manual testing + QE.
I will document the tests i am now currently performing manually with a downstream image with this PR.

Issue starts in 4.18 when the constraint was introduced to only allow one cluster subnet per IP family. Issue seen from upgrade from 4.17 to 4.18.
I will:
1. launch openshift 4.17 with multiple cluster subnets of the same IP and upgrade to 4.18 (without fix) and witness error
2. launch openshift 4.17 with multiple cluster subnets of the same IP and upgrade to 4.18 (with fix) and witness no error
3. launch openshift 4.18 cluster with fix and witness no error

QE will perform regression EIP testing on 4.18 with this fix and also retest step 2
QE will upgrade from 4.18.2 (without fix) -> 4.18 (with fix) and perform EIP regression testing.
QE will also launch a fresh cluster in 4.18 with this fix.

jcaamano (approved):

PR #5166, opened 2025-04-07
Universal network selector API

jcaamano:
Adds a universal network selector API to used across all out other APIs. 
Updates RouteAdvertisements API to use this universal network selector API.
Switch other label selectors in the RouteAdvertisements API to the standard convention.


go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -994,6 +982,43 @@ func (c *Controller) updateRAStatus(ra *ratypes.RouteAdvertisements, hadUpdates
...
+		case apitypes.ClusterUserDefinedNetworks:
+			nadSelector, err := metav1.LabelSelectorAsSelector(&networkSelector.ClusterUserDefinedNetworkSelector.NetworkSelector)
+			if err != nil {
+				return nil, err
+			}
+			nads, err := c.nadLister.List(nadSelector)
+			if err != nil {
+				return nil, err
+			}
+			for _, nad := range nads {
```
trozet, line 1006: I think to be completely sure the nad matches the cudn you should cross-reference the ownerReferences->uid and ensure it matches the uid of the CUDN:


```
trozet@fedora:~/go/src/github.com/ovn-org/ovn-kubernetes/contrib$ oc get network-attachment-definition -A -o yaml
apiVersion: v1
items:
- apiVersion: k8s.cni.cncf.io/v1
  metadata:
    annotations:
      k8s.ovn.org/network-id: "1"
      k8s.ovn.org/network-name: cluster_udn_l3-primary
    creationTimestamp: "2025-04-07T21:45:53Z"
    finalizers:
    - k8s.ovn.org/user-defined-network-protection
    generation: 1
    labels:
      k8s.ovn.org/user-defined-network: ""
    name: l3-primary
    namespace: udn-test
    ownerReferences:
    - apiVersion: k8s.ovn.org/v1
      blockOwnerDeletion: true
      controller: true
      name: l3-primary
      uid: d20fb3f6-b019-49a5-8fb7-7bdfd7868dda <----------should match CUDN
    resourceVersion: "15419"
    uid: de9db26b-2e0e-4bf1-8232-554dd497ff94
  spec:
    config: '{"cniVersion":"1.0.0","joinSubnets":"100.65.0.0/16,fd99::/64","name":"cluster_udn_l3-primary","netAttachDefName":"udn-test/l3-primary","role":"primary","subnets":"10.20.0.0/16","topology":"layer3","type":"ovn-k8s-cni-overlay"}'
kind: List


apiVersion: v1
items:
- apiVersion: k8s.ovn.org/v1
  metadata:
    creationTimestamp: "2025-04-07T21:45:28Z"
    finalizers:
    - k8s.ovn.org/user-defined-network-protection
    generation: 1
    name: l3-primary
    resourceVersion: "17959"
    uid: d20fb3f6-b019-49a5-8fb7-7bdfd7868dda

````

**Review by trozet — COMMENTED** [review]
do you plan on adding another commit, or a follow up PR to copy the name of the object as a label? similar to what k8s namespaces do

**Comment by jcaamano** [comment]
> do you plan on adding another commit, or a follow up PR to copy the name of the object as a label? similar to what k8s namespaces do

I had no plans. I guess that can come in a follow up PR.

File: go-controller/pkg/clustermanager/routeadvertisements/controller.go (hunk 1 above)
Comment (jcaamano) [reply go-controller/pkg/clustermanager/routeadvertisements/controller.go:1006]: I will check that the owner reference is there in case anyone has decided to follow the same naming pattern of the CUDN (which likely is problematic elsewhere anyway). I don't think we need to match the UID itself. If we do, probably selecting NADs is not the best approach after all. But I don't see why we would need to do that.

File: go-controller/pkg/crd/types/networkselector.go [hunk 2]
```diff
@@ -0,0 +1,88 @@
...
+// +kubebuilder:validation:XValidation:rule="!has(self.primaryUserDefinedNetworkSelector) || has(self.networkSelectionType) && self.networkSelectionType == 'PrimaryUserDefinedNetworks'",message="Can only specify primaryUserDefinedNetworkSelector when networkSelectionType is PrimaryUserDefinedNetworks"
+// +kubebuilder:validation:XValidation:rule="!has(self.secondaryUserDefinedNetworkSelector) || has(self.networkSelectionType) && self.networkSelectionType == 'SecondaryUserDefinedNetworks'",message="Can only specify secondaryUserDefinedNetworkSelector when networkSelectionType is SecondaryUserDefinedNetworks"
+// +kubebuilder:validation:XValidation:rule="!has(self.networkAttachmentDefinitionSelector) || has(self.networkSelectionType) && self.networkSelectionType == 'NetworkAttachmentDefinitions'",message="Can only specify networkAttachmentDefinitionSelector when networkSelectionType is NetworkAttachmentDefinitions"
+type NetworkSelector struct {
+	// networkSelectionType determines the type of networks selected.
+	// +kubebuilder:validation:Required
+	NetworkSelectionType NetworkSelectionType `json:"networkSelectionType"`
```
kyrtapz, line 24: Have you considered using a `unionDiscriminator`? I think this could help remove the CEL validation above.

go-controller/pkg/crd/types/networkselector.go
```diff
@@ -0,0 +1,88 @@
...
+// +kubebuilder:validation:MinItems=1
+// +kubebuilder:validation:MaxItems=5
+// +patchMergeKey=networkSelectionType
+// +patchStrategy=merge
+// +listType=map
+// +listMapKey=networkSelectionType
+type NetworkSelectors []NetworkSelector
+
+// NetworkSelector selects a set of networks.
+// +kubebuilder:validation:XValidation:rule="!has(self.clusterUserDefinedNetworkSelector) || has(self.networkSelectionType) && self.networkSelectionType == 'ClusterUserDefinedNetworks'",message="Can only specify clusterUserDefinedNetworkSelector when networkSelectionType is ClusterUserDefinedNetworks"
```
kyrtapz, line 17: I think that without using an applyconfiguration it would be difficult to create the NetworkSelector because the Selectors will always be set to an empty value, `omitempty` doesn't work on structs.

go-controller/pkg/crd/types/networkselector.go
jcaamano: Can you give me a reference?

go-controller/pkg/crd/types/networkselector.go
jcaamano: Can you expand. I don't really follow?

go-controller/pkg/crd/types/networkselector.go
kyrtapz: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/96b8798640db4314dd915716b9f8546dd88ebec2/go-controller/pkg/crd/userdefinednetwork/v1/cudn.go#L40-L61

go-controller/pkg/crd/types/networkselector.go
kyrtapz: When using a client-go that tries to set a `NetworkSelector` through the go structs it will always send all selectors because an empty label selector is not actually omitted, it results in `selector: {}` json representation.

go-controller/pkg/crd/types/networkselector.go
jcaamano: Thanks. So I did think somethign like this might be available but I could find doc in  I can't find doc
https://book.kubebuilder.io/introduction?search=unionDiscriminator
Is this openshift specific? Is there doc?

kyrtapz: I don't think it is a kubebuilder feature, here is a KEP I found but it doesn't exactly match what we used for UDN
https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/1027-api-unions/README.md

Seems we used the OCP verison:
https://github.com/openshift/enhancements/blob/master/dev-guide/api-conventions.md#discriminated-unions

kyrtapz: here is an upstream example that seems the same to what we are using: https://github.com/kubernetes/kubernetes/blob/ab3e83f73424a18f298a0050440af92d2d7c4720/pkg/apis/core/types.go#L4020-L4031

jcaamano: I can throw it in.
But precisely what validation do you think it replaces? Can we be 100% sure of what it is doing to remove that validation?

go-controller/pkg/crd/types/networkselector.go
jcaamano: So you are saying that if we try to create something like this in go it will fail
```
ratypes.RouteAdvertisements{
			Name: tra.Name,
		},
		Spec: ratypes.RouteAdvertisementsSpec{
			TargetVRF:      tra.TargetVRF,
			Advertisements: []ratypes.AdvertisementType{ratypes.PodNetwork},
			NetworkSelectors: apitypes.NetworkSelectors{apitypes.NetworkSelector{NetworkSelectionType: apitypes.DefaultNetwork}}
		},
	}
```
because the other NetworkSelector fields are not really empty. I will try it out. 

There is an `omitzero` option, wonder if that would work. I don't see any other alternatives, do you?

kyrtapz: Exactly, something very similar already happens for RAs: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5130#discussion_r2007116178

The only way out of this is to use pointers, this seems to be a requirement for discriminated unions so two birds with one stone?

go-controller/pkg/crd/types/networkselector.go
jcaamano: No changes at all in the schema with or without it. I will just throw it in to be aligned with the UDN spec.

go-controller/pkg/crd/types/networkselector.go
jcaamano: > Exactly, something very similar already happens for RAs: [#5130 (comment)](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5130#discussion_r2007116178)

I guess I am fixing this as well.

jcaamano:
My intention all along was to follow standard conventions with the label selectors.
But I guess I was wrong on which those where:
* no label selects nothing
* `{}` selects all
Fixed in latest commit.

jcaamano:
> My intention all along was to follow standard conventions with the label selectors. But I guess I was wrong on which those where:
> 
> * no label selects nothing
> * `{}` selects all
>   Fixed in last commit.

Actually not done yet. Want to make the labels required and not pointers. Still same semantics as detailed above.

The thing is that while I think that for `nodeSelector` might make sense to have a default that selects all nodes, I don't think the same thing makes sense for `frrConfigurationSelector` because most of the time this is not what you want to do. Since I don't like to have different selector semantics across the same API, I think it is best to stick with making them all required.

jcaamano:
> > My intention all along was to follow standard conventions with the label selectors. But I guess I was wrong on which those where:
> > 
> > * no label selects nothing
> > * `{}` selects all
> >   Fixed in last commit.
> 
> Actually not done yet. Want to make the labels required and not pointers. Still same semantics as detailed above.
> 
> The thing is that while I think that for `nodeSelector` might make sense to have a default that selects all nodes, I don't think the same thing makes sense for `frrConfigurationSelector` because most of the time this is not what you want to do. Since I don't like to have different selector semantics across the same API, I think it is best to stick with making them all required.

There, done.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
trozet: We are matching on a CUDN here, so imo we need to be strict on correlating NADs to CUDN. If someone created a NAD that matched the cudn naming scheme, we should not select that.

As far as selecting CUDN rather than NAD, at the end of the day I think you are going to need to find the NADs. If you only selected CUDN, then you need to watch namespaces and reconcile which namespaces map to the CUDN depending on the controller logic. This work is already handled by the NAD controller. Everything results in a NAD, and controllers basically operate on a NAD level. I think it makes sense to look for NADs that are created and derived from a CUDN, since the existence of those NADs tell us the validation has already been done and the NAD/namespace setup is proper.

jcaamano:
```
If someone created a NAD that matched the cudn naming scheme, we should not select that.
```
If I check owner ref `apiVersion` and `kind` and no longer match on the cudn naming scheme, this is no longer an issue. What issue do you want to prevent by matching specifically the UID?

kyrtapz:
Forgot to remove this CEL?
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/06c26bc40d3793ff4c4c01d9a6fc61dd444eb0cf/go-controller/pkg/crd/routeadvertisements/v1/types.go#L26

jcaamano:
> Forgot to remove this CEL?
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/06c26bc40d3793ff4c4c01d9a6fc61dd444eb0cf/go-controller/pkg/crd/routeadvertisements/v1/types.go#L26
> 
> With the new approach I guess we need a CEL that verifies that the labelselector is empty.

We check this internally, unsure about a new CEL.

jcaamano:
> Forgot to remove this CEL?
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/06c26bc40d3793ff4c4c01d9a6fc61dd444eb0cf/go-controller/pkg/crd/routeadvertisements/v1/types.go#L26
> 
> With the new approach I guess we need a CEL that verifies that the labelselector is empty.

Just removed the CEL. I am not sure how to make a future-proof CEL for a label selector. This is anyway checked internally an error status set on the RA.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -994,6 +984,44 @@ func (c *Controller) updateRAStatus(ra *ratypes.RouteAdvertisements, hadUpdates
...
+				// check this NAD is controlled by a CUDN
+				controller := metav1.GetControllerOfNoCopy(nad)
+				isCUDN := controller != nil && controller.Kind == cudnController.Kind && controller.APIVersion == cudnController.GroupVersion().String()
+				if !isCUDN {
+					continue
+				}
+				selected = append(selected, nad)
```
kyrtapz, line 1015: Shouldn't we also exclude any NADs that do not represent a primary network?

kyrtapz:
> nodeSelector

We could just check that the length of matchLabels and matchExpressions both 0.

jcaamano:
> > > Forgot to remove this CEL?
> > > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/06c26bc40d3793ff4c4c01d9a6fc61dd444eb0cf/go-controller/pkg/crd/routeadvertisements/v1/types.go#L26
> > > 
> > > With the new approach I guess we need a CEL that verifies that the labelselector is empty.
> > 
> > 
> > Just removed the CEL. I am not sure how to make a future-proof CEL for a label selector. This is anyway checked internally and error status set on the RA.
> 
> We could just check that the length of matchLabels and matchExpressions are both 0. Having a CEL the user would get a more direct feedback but I am ok to keep it as is if you prefer.

That won't work if they add a field to LabelSelector in the future additional to `matchLabels` and `matchExpressions`. That should be opaque to us. That is what I meant as feature-prrof, there is no LabelSelector API usable for us at this level.

jcaamano:
> > > > Forgot to remove this CEL?
> > > > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/06c26bc40d3793ff4c4c01d9a6fc61dd444eb0cf/go-controller/pkg/crd/routeadvertisements/v1/types.go#L26
> > > > 
> > > > With the new approach I guess we need a CEL that verifies that the labelselector is empty.
> > > 
> > > 
> > > Just removed the CEL. I am not sure how to make a future-proof CEL for a label selector. This is anyway checked internally and error status set on the RA.
> > 
> > 
> > We could just check that the length of matchLabels and matchExpressions are both 0. Having a CEL the user would get a more direct feedback but I am ok to keep it as is if you prefer.
> 
> That won't work if they add a field to LabelSelector in the future additional to `matchLabels` and `matchExpressions`. That should be opaque to us. That is what I meant as future-proof, there is no LabelSelector API usable for us at this level.

I guess for that specific case it would probably not be a problem if we catch it later. What I am concerned about in general is the lack of an API that generally tells me when the selector restricts the selected space or not, like we have in go bindings. I am alone in this concern then we can add it.

jcaamano:
> > > > > Forgot to remove this CEL?
> > > > > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/06c26bc40d3793ff4c4c01d9a6fc61dd444eb0cf/go-controller/pkg/crd/routeadvertisements/v1/types.go#L26
> > > > > 
> > > > > With the new approach I guess we need a CEL that verifies that the labelselector is empty.
> > > > 
> > > > 
> > > > Just removed the CEL. I am not sure how to make a future-proof CEL for a label selector. This is anyway checked internally and error status set on the RA.
> > > 
> > > 
> > > We could just check that the length of matchLabels and matchExpressions are both 0. Having a CEL the user would get a more direct feedback but I am ok to keep it as is if you prefer.
> > 
> > 
> > That won't work if they add a field to LabelSelector in the future additional to `matchLabels` and `matchExpressions`. That should be opaque to us. That is what I meant as future-proof, there is no LabelSelector API usable for us at this level.
> 
> I guess for that specific case it would probably not be a problem if we catch it later. What I am concerned about in general is the lack of an API that generally tells me when the selector restricts the selected space or not, like we have in go bindings. If am alone in this concern then we can add it.

Added by non-popular demand

go-controller/pkg/clustermanager/routeadvertisements/controller.go
trozet: sure I guess that would also suffice

go-controller/pkg/clustermanager/routeadvertisements/controller.go
trozet: good point, as well as localnet types?

jcaamano: That filtering is already done in `generateFRRConfigurations`. Filtering applied in this method only corresponds to the what the selector itself is telling us. This potentially can be common code in the future for any APIs using the selector.

jcaamano: Aside from that, if they select something that we don't support, we return error `generateFRRConfigurations`, we don't ignore it.

flavio-fernandes (commented):
/LGTM

flavio-fernandes:
Perma-failing on unrelated issue:  https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5170

```
Summarizing 2 Failures:
  [FAIL] [sig-network] Services [It] should respect internalTrafficPolicy=Local Pod and Node, to Pod (hostNetwork: true) [sig-network]
  k8s.io/kubernetes/test/e2e/network/service.go:2696
  [FAIL] [sig-network] Services [It] should respect internalTrafficPolicy=Local Pod (hostNetwork: true) to Pod [sig-network]
  k8s.io/kubernetes/test/e2e/network/service.go:2594
```

kyrtapz (approved):

jcaamano:
Well had to revert because it interleaved with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5167

flavio-fernandes:
Note: Work being continued in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5166

PR #5168, opened 2025-04-08
kind: check minimum version available

jcaamano:


contrib/kind.sh
```diff
@@ -468,6 +468,13 @@ check_dependencies() {
     exit 1
   fi
 
+  kindMin="0.27.0"
+  kindCur=$(kind version | grep -Po '\d+\.\d+.\d+' | head -1)
```
ormergi, line 472: nit: We can use -q flag to get clean kind version:
```
kind version -q
0.27.0
```

ormergi (approved):
Thanks for the PR!
It will surly save troubleshooting when kind version is old.

ormergi (approved):

kyrtapz (dismissed):
Super NIT: I don't think we are using camelCase for variables in bash? I don't mind to keep it as is

LGTM

PR #5169, opened 2025-04-09
Import BGP routes to UDN when EgressIP is advertised

pliurh:
BGP routes will be imported to the UDN VRF and GR, when EgressIP for the UDN is advertised to the default VRF. This ensures that the egress traffic can find the right egress route.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## How to verify it

1. Create a CUDN

2. Create a namespace that uses the CUDN as the primary network.

3. Create an EIP that selects the pods in the namespace

4. Create an RA that selects the CUDN and only advertises the EIP to the default VRF

5. Create a FRRConfiguration that can connect to an external BGP router. The BGP router is different from the default route of the node.

6. Create a test pod in the namespace. And try to curl an external host that is only reachable via the BGP router.

### DIFF
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -714,11 +714,6 @@ func (c *Controller) generateFRRConfiguration(
 			continue
 		}
 
-		// we won't do imports if the pod network is not advertised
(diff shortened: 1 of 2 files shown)


tssurya:
Please don't use RH JIRA's upstream - it hurts the community aspect.

Issue #5183, opened 2025-04-15
Upgrade OVN tests are failing while waiting for deployments.apps ovnkube-identity


flavio-fernandes:
### What happened?

When kube-id changed from deploy to ds, we missed updating a script that still expects it to be a deployments.

Introduced here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5146

Ref failure: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14479210371/job/40613851848?pr=5182

```
+ echo 'info: waiting for '\''kubectl get deployments.apps ovnkube-identity -n ovn-kubernetes'\'' to succeed...'
+ sleep 1
+ true
+ kubectl get deployments.apps ovnkube-identity -n ovn-kubernetes
Error from server (NotFound): deployments.apps "ovnkube-identity" not found
+ (( retries += 1 ))
+ [[ 11 -gt 10 ]]
+ echo 'error: '\''kubectl get deployments.apps ovnkube-identity -n ovn-kubernetes'\'' did not succeed, failing'
+ exit 1
make: *** [Makefile:18: upgrade-ovn] Error 1
error: 'kubectl get deployments.apps ovnkube-identity -n ovn-kubernetes' did not succeed, failing
make: Leaving directory '/home/runner/work/ovn-kubernetes/ovn-kubernetes/test'
Error: Process completed with exit code 2.
``` 

### What did you expect to happen?

ci to pass

### How can we reproduce it (as minimally and precisely as possible)?

Run make a pr

### Anything else we need to know?

ez fix!

### OVN-Kubernetes version

latest master, after https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5146

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### OVN version

<details>

```console
$ oc rsh -n ovn-kubernetes ovnkube-node-xxxxx (pick any ovnkube-node pod on your cluster)
$ rpm -q ovn
# paste output here
```

</details>

### OVS version

<details>

```console
$ oc rsh -n ovn-kubernetes ovs-node-xxxxx (pick any ovs pod on your cluster)
$ rpm -q openvswitch
# paste output here
```

</details>

### Platform

<details>

Is it baremetal? GCP? AWS? Azure?
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
[...]

Issue #5170, opened 2025-04-09
`should respect internalTrafficPolicy=Local Pod`


tssurya:
### Which jobs are flaking?

shard-conformance, v6

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14355614780/job/40246109108?pr=4878

### Which tests are flaking?

```
Summarizing 2 Failures:
  [FAIL] [sig-network] Services [It] should respect internalTrafficPolicy=Local Pod (hostNetwork: true) to Pod [sig-network]
  k8s.io/kubernetes/test/e2e/network/service.go:2594
  [FAIL] [sig-network] Services [It] should respect internalTrafficPolicy=Local Pod and Node, to Pod (hostNetwork: true) [sig-network]
  k8s.io/kubernetes/test/e2e/network/service.go:2696
```

### Since when has it been flaking?

Since rebase maybe? we need to check if these are new tests.

flavio-fernandes:
Also seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14361895500/job/40267312151?pr=4508

flavio-fernandes:
And another one:  https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14361895500/job/40273662258?pr=4508

flavio-fernandes:
And also here:  https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14365091868/job/40277645470?pr=5166

jcaamano:
Runner image update happened
From
```
  Image: ubuntu-22.04
  Version: 20250323.1.0
  Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20250323.1/images/ubuntu/Ubuntu2204-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20250323.1
```
https://github.com/actions/runner-images/blob/ubuntu22/20250323.1/images/ubuntu/Ubuntu2204-Readme.md
to
```
  Image: ubuntu-22.04
  Version: 20250406.1.0
  Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20250406.1/images/ubuntu/Ubuntu2204-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20250406.1
```
[...]

Issue #4806, opened 2024-10-25
[TIMEDOUT] [sig-network] Services [It] should not be able to connect to terminating and unready endpoints if PublishNotReadyAddresses is false [sig-network]   k8s.io/kubernetes/test/e2e/network/service.go:1997


tssurya:
### Which jobs are flaking?

Shard-Conformance

### Which tests are flaking?

```
Summarizing 1 Failure:
  [TIMEDOUT] [sig-network] Services [It] should not be able to connect to terminating and unready endpoints if PublishNotReadyAddresses is false [sig-network]
  k8s.io/kubernetes/test/e2e/network/service.go:1997
```

### Since when has it been flaking?

hard to tell, sorry

### Anything else we need to know?

https://github.com/ovn-org/ovn-kubernetes/actions/runs/11517587527/job/32063721731?pr=4777

flavio-fernandes:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13399840958/job/37429305872?pr=4508

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13817315281/job/38655973334?pr=5106

flavio-fernandes:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14496319721/job/40667151360?pr=5184

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22764254284/job/66028956542?pr=6073

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22853911545/job/66290197780?pr=6086
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6086#issuecomment-4024454042

pperiyasamy:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/23756744815/job/69222234036?pr=6034

PR #5184, opened 2025-04-15
upgrade-ovn.sh: ovnkube-identity is now a DaemonSet

flavio-fernandes:
Starting from:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5146 ovnkube-identity pods are now created via a DaemonSet.

The test in upgrade-ovn.sh needed to be updated accordingly.

Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5183

### DIFF
--- a/contrib/kind-helm.sh
+++ b/contrib/kind-helm.sh
@@ -403,7 +403,6 @@ helm install ovn-kubernetes . -f "${value_file}" \
           --set k8sAPIServer=${API_URL} \
           --set podNetwork="${NET_CIDR_IPV4}/24" \
           --set serviceNetwork=${SVC_CIDR_IPV4} \
-          --set ovnkube-identity.replicas=${MASTER_REPLICAS} \
           --set ovnkube-master.replicas=${MASTER_REPLICAS} \
           --set global.image.repository=$(get_image) \
           --set global.image.tag=$(get_tag) \
--- a/docs/features/infrastructure-security-controls/node-identity.md
+++ b/docs/features/infrastructure-security-controls/node-identity.md
@@ -65,7 +65,7 @@ Some of the allowed annotations have additional checks; for instance, the IP add
 must match the node's [k8s.ovn.org/node-subnets](https://github.com/ovn-org/ovn-kubernetes/blob/5d56a53df520a085e629cdc71be092afed9c3f0f/go-controller/pkg/util/subnet_annotations.go#L15-L39) networks.
 
 
-## Deployment
+## DaemonSet
 
 In Kind,
 the feature is enabled by default and can be disabled with `--disable-ovnkube-identity` when creating the cluster.\
--- a/docs/installation/launching-ovn-kubernetes-with-helm.md
+++ b/docs/installation/launching-ovn-kubernetes-with-helm.md
@@ -71,10 +71,10 @@ networking:
 
 This chart does not use a `values.yaml` by default. You must specify a values file during installation.
 
-- Run `helm install` with propery `k8sAPIServer`, `ovnkube-identity.replicas`, image repo and tag
+- Run `helm install` with propery `k8sAPIServer` image repo and tag
 ```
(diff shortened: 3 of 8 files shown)


crnithya:
@flavio-fernandes sorry about missing these while making the switch to daemonset. 
Can you also remove the kube-id replicas under

- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-no-ic.yaml#L170
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-single-node-zone.yaml#L158
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-multi-node-zone.yaml#L157
I really appreciate you fixing these.

tssurya:
So why didn't the upgrade lane catch it in @crnithya 's PR ?
I thought we build and upgrade to the new version ?
/cc @npinaeva

npinaeva:
Strange unit test flake, filed https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5185

tssurya:
> upgrade job checks master -> PR upgrade path. At the original PR time that means deployment -> daemonset upgrade, which makes `get deployment` command work because deployment from the master branch still exists. After the PR was merged, it switched to daemonset -> daemonset upgrade and `get deployment` started to fail thanks @flavio-fernandes for the fix!

so is this a backwards incompatible change? Can someone actually upgrade from OVNK release 1.0.0 to the new release I will cut for 1.0.1 with this new PR in it?

tssurya:
> > upgrade job checks master -> PR upgrade path. At the original PR time that means deployment -> daemonset upgrade, which makes `get deployment` command work because deployment from the master branch still exists. After the PR was merged, it switched to daemonset -> daemonset upgrade and `get deployment` started to fail thanks @flavio-fernandes for the fix!
> 
> so is this a backwards incompatible change? Can someone actually upgrade from OVNK release 1.0.0 to the new release I will cut for 1.1.0 with nithya's new PR in it? Maybe based on what you said since 1.0.0 has the deployment still it will work just fine?
How is our support for upgrades upstream? we need to discuss this in upstream meeting next time cc @girishmg

flavio-fernandes:
> @flavio-fernandes sorry about missing these while making the switch to daemonset. Can you also remove the kube-id replicas under
> 
> * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-no-ic.yaml#L170
> * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-single-node-zone.yaml#L158
> * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-multi-node-zone.yaml#L157
>   I really appreciate you fixing these.

fised.

npinaeva:
@flavio-fernandes I think we need to fix this part of the script too https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/scripts/upgrade-ovn.sh#L48
iiuc it should say `run_kubectl get daemonsets.apps $1 -n ovn-kubernetes` instead of using `ovnkube-node` as a name

flavio-fernandes:
> @flavio-fernandes I think we need to fix this part of the script too https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/scripts/upgrade-ovn.sh#L48 iiuc it should say `run_kubectl get daemonsets.apps $1 -n ovn-kubernetes` instead of using `ovnkube-node` as a name

🤦 

fixed.

flavio-fernandes:
- Known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5170
- Known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4806

PR #5190, opened 2025-04-21
docs: Use official logo

flavio-fernandes:
The upstream documentation should use the official logo for the project, which is published at:

https://github.com/cncf/artwork/tree/main/projects/ovn-kubernetes


tssurya:
ah thanks for adding this I actually had a PR for it here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5083

PR #5192, opened 2025-04-22
Isolate default network services from UDN pods

tssurya:
## 📑 Description

Currently the behaviour of an advertised UDN is that
UDN pods are able to talk any service in the default
network in local gateway mode.

Expected behaviour is that advertised UDN pods should
only be able to reach kapi and dns services in the
default network and rest of them should not be
reachable.

Solution adopted is to change all the relevant flows
that were using masIP subnet for services isolation
on br-ex to use the podsubnets.

Caveat is: if user advertises overlapping podSubnets
then things will fall apart and assumption is users
can't do that.

Flow details (post fix I don't want to call out the
current behaviour to avoid confusions -> current behaviour
was using the default network flows for UDN pods which was
wrong):

Onward packet

1) cookie=0xdeff105, duration=1472.742s, table=0, n_packets=9, n_bytes=666,
   priority=550,ip,in_port=LOCAL,nw_src=103.103.0.0/16,nw_dst=10.96.0.0/16 actions=ct(commit,table=2,zone=64001)
this flow is same as that for UDNs with the masSubnet as the src

2) cookie=0xdeff105, duration=1472.742s, table=2, n_packets=9, n_bytes=666,
   priority=200,ip,nw_src=103.103.0.0/16 actions=mod_dl_dst:02:42:ac:12:00:03,output:3
this flow is also same flow as that for UDNs with the masIP as the src

So in this above way ^ any traffic from any advertised UDNs towards clusterIP
range will always be taken into the patch port of that respective UDN
and this guarantees isolation towards services belonging in other UDNs
since that given UDN won't have any LBs for those services

Return packet

3) There is no need for returning any packets correctly unless its for
kapi and/or dns

for UDNs, I see this flow:
cookie=0xdeff105, duration=684.087s, table=0, n_packets=0, n_bytes=0,
idle_age=684, priority=500,ip,in_port=2,nw_src=10.96.0.0/16,nw_dst=169.254.0.0/17 actions=ct(table=3,zone=64001,nat)
 cookie=0xdeff105, duration=684.050s, table=0, n_packets=0, n_bytes=0,
idle_age=684, priority=500,ip,in_port=4,nw_src=10.96.0.0/16,nw_dst=169.254.0.0/17 actions=ct(table=3,zone=64001,nat)

but for BGP I change these two flows:
cookie=0xdeff105, duration=5216.234s, table=2, n_packets=1067, n_bytes=120337,
priority=300,ip,nw_src=169.254.0.0/17,nw_dst=10.96.0.1 actions=mod_dl_dst:96:fd:12:44:a1:83,output:"patch-breth0_ov"
cookie=0xdeff105, duration=5216.234s, table=2, n_packets=0, n_bytes=0,
priority=300,ip,nw_src=169.254.0.0/17,nw_dst=10.96.0.10 actions=mod_dl_dst:96:fd:12:44:a1:83,output:"patch-breth0_ov"

into:

 cookie=0xdeff105, duration=2319.685s, table=2, n_packets=496, n_bytes=67111,
priority=300,ip,nw_dst=10.96.0.1 actions=mod_dl_dst:02:42:ac:12:00:03,output:"patch-breth0_ov"
 cookie=0xdeff105, duration=2319.685s, table=2, n_packets=0, n_bytes=0,
priority=300,ip,nw_dst=10.96.0.10 actions=mod_dl_dst:02:42:ac:12:00:03,output:"patch-breth0_ov"

that is generic enough for both UDN and BGP packets to match on.

There is scope for scale improvement by combining the onward packet flows
at tables 0 and 2 into one entity, but maybe that could also be done for UDNs?

There is now isolation established between UDN pods and default network
because except kapi and dns all other traffic matching clusterIP is sent
to the respective UDN patch ports where its black-holed (in l3 it goes
back and forth from breth0 to GR and back where its dropped using 105 flow).

However in L2 we do see the expected bad behaviour where packet is looping from
management port -> breth0 -> GR -> management port -> breth0 and so on
which is a never ending loop
If we decide to perhaps ban the circular looping and optimize the flow better
using a proper drop rule:

inport=LOCAL, srcIP=UDNSubnet, dstIP=serviceSubnet -> 1 per advertised UDN “DROP”
that can be considered as a future improvement here

dump:
```
tcpdump: listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
20:13:18.881036 33dfd1008b804_3 P   ifindex 15 0a:58:67:67:00:05 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 64, id 25988, offset 0, flags [DF], proto TCP (6), length
60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x1614 (incorrect -> 0x5b5d), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale
 7], length 0
20:13:18.881626 ovn-k8s-mp1 In  ifindex 8 0a:58:64:41:00:03 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 63, id 25988, offset 0, flags [DF], proto TCP (6), length 60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x5b5d (correct), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale 7], length
0
20:13:18.881651 breth0 Out ifindex 4 02:42:ac:12:00:02 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 62, id 25988, offset 0, flags [DF], proto TCP (6), length 60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x5b5d (correct), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale 7], length
0
20:13:18.882283 ovn-k8s-mp1 In  ifindex 8 0a:58:64:41:00:03 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 61, id 25988, offset 0, flags [DF], proto TCP (6), length 60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x5b5d (correct), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale 7], length
0
20:13:18.882298 breth0 Out ifindex 4 02:42:ac:12:00:02 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 60, id 25988, offset 0, flags [DF], proto TCP (6), length 60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x5b5d (correct), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale 7], length
0
20:13:18.882402 ovn-k8s-mp1 In  ifindex 8 0a:58:64:41:00:03 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 59, id 25988, offset 0, flags [DF], proto TCP (6), length 60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x5b5d (correct), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale 7], length
0
20:13:18.882414 breth0 Out ifindex 4 02:42:ac:12:00:02 ethertype IPv4 (0x0800), length 80: (tos 0x0, ttl 58, id 25988, offset 0, flags [DF], proto TCP (6), length 60)
    103.103.0.5.36363 > 10.96.164.25.80: Flags [S], cksum 0x5b5d (correct), seq 2541324161, win 65280, options [mss 1360,sackOK,TS val 984084514 ecr 0,nop,wscale 7], length
```

$ oc rsh -n blue blue3
/ # curl --local-port 36363 10.96.164.25:80
curl: (7) Failed to connect to 10.96.164.25 port 80 after 3 ms: Host is unreachable

Signed-off-by: Surya Seetharaman <suryaseetharaman.9@gmail.com>

## ✅ Checks

- [x] My code requires tests

## How to verify it
E2E and unit tests were done

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -411,18 +411,33 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 			}
 
 			ipPrefix := "ip"
-			masqueradeSubnet := config.Gateway.V4MasqueradeSubnet
 			if !utilnet.IsIPv4String(service.Spec.ClusterIP) {
 				ipPrefix = "ipv6"
-				masqueradeSubnet = config.Gateway.V6MasqueradeSubnet
 			}
 			// table 2, user-defined network host -> OVN towards default cluster network services
 			defaultNetConfig := npw.ofm.defaultBridge.getActiveNetworkBridgeConfig(types.DefaultNetworkName)
-
-			npw.ofm.updateFlowCacheEntry(key, []string{fmt.Sprintf("cookie=%s, priority=300, table=2, %s, %s_src=%s, %s_dst=%s, "+
+			// sample flow: cookie=0xdeff105, duration=2319.685s, table=2, n_packets=496, n_bytes=67111, priority=300,
+			//              ip,nw_dst=10.96.0.1 actions=mod_dl_dst:02:42:ac:12:00:03,output:"patch-breth0_ov"
+			// This flow is used for UDNs and advertised UDNs to be able to reach kapi and dns services alone on default network
+			flows := []string{fmt.Sprintf("cookie=%s, priority=300, table=2, %s, %s_dst=%s, "+
 				"actions=set_field:%s->eth_dst,output:%s",
-				defaultOpenFlowCookie, ipPrefix, ipPrefix, masqueradeSubnet, ipPrefix, service.Spec.ClusterIP,
-				npw.ofm.getDefaultBridgeMAC().String(), defaultNetConfig.ofPortPatch)})
+				defaultOpenFlowCookie, ipPrefix, ipPrefix, service.Spec.ClusterIP,
+				npw.ofm.getDefaultBridgeMAC().String(), defaultNetConfig.ofPortPatch)}
+			if util.IsRouteAdvertisementsEnabled() {
+				// if the network is advertised, then for the reply from kapi and dns services to go back
+				// into the UDN's VRF we need flows that statically send this to the local port
+				// sample flow: cookie=0xdeff105, duration=264.196s, table=0, n_packets=0, n_bytes=0, priority=490,ip,
+				//              in_port="patch-breth0_ov",nw_src=10.96.0.10,actions=ct(table=3,zone=64001,nat)
+				// this flow is meant to match all advertised UDNs and then the ip rules on the host will take
+				// this packet into the corresponding UDNs
+				// NOTE: We chose priority 490 to differentiate this flow from the flow at priority 500 added for the
+				// non-advertised UDNs reponse for debugging purposes:
+				// sample flow for non-advertised UDNs: cookie=0xdeff105, duration=684.087s, table=0, n_packets=0, n_bytes=0,
+				//				idle_age=684, priority=500,ip,in_port=2,nw_src=10.96.0.0/16,nw_dst=169.254.0.0/17 actions=ct(table=3,zone=64001,nat)
+				flows = append(flows, fmt.Sprintf("cookie=%s, priority=490, in_port=%s, ip, ip_src=%s,actions=ct(zone=%d,nat,table=3)",
+					defaultOpenFlowCookie, defaultNetConfig.ofPortPatch, service.Spec.ClusterIP, config.Default.HostMasqConntrackZone))
+			}
+			npw.ofm.updateFlowCacheEntry(key, flows)
 		}
 	}
 	return utilerrors.Join(errors...)
@@ -1593,6 +1608,37 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 					"actions=ct(commit,zone=%d,table=2)",
 					defaultOpenFlowCookie, ofPortHost, protoPrefix, protoPrefix,
 					masqSubnet, protoPrefix, svcCIDR, config.Default.HostMasqConntrackZone))
+			if util.IsRouteAdvertisementsEnabled() {
+				// If the UDN is advertised then instead of matching on the masqSubnet
+				// we match on the UDNPodSubnet itself and we also don't SNAT to 169.254.0.2
+				// sample flow: cookie=0xdeff105, duration=1472.742s, table=0, n_packets=9, n_bytes=666, priority=550
+				//              ip,in_port=LOCAL,nw_src=103.103.0.0/16,nw_dst=10.96.0.0/16 actions=ct(commit,table=2,zone=64001)
+				for _, netConfig := range bridge.patchedNetConfigs() {
+					if netConfig.isDefaultNetwork() {
+						continue
+					}
+					if netConfig.advertised.Load() {
+						var udnAdvertisedSubnets []*net.IPNet
+						for _, clusterEntry := range netConfig.subnets {
+							udnAdvertisedSubnets = append(udnAdvertisedSubnets, clusterEntry.CIDR)
+						}
+						// Filter subnets based on the clusterIP service family
+						// NOTE: We don't support more than 1 subnet CIDR of same family type; we only pick the first one
+						matchingIPFamilySubnet, err := util.MatchFirstIPNetFamily(utilnet.IsIPv6CIDR(svcCIDR), udnAdvertisedSubnets)
+						if err != nil {
+							klog.Infof("Unable to determine UDN subnet for the provided family isIPV6: %t, %v", utilnet.IsIPv6CIDR(svcCIDR), err)
+							continue
+						}
+
+						// Use the filtered subnet for the flow compute instead of the masqueradeIP
+						dftFlows = append(dftFlows,
+							fmt.Sprintf("cookie=%s, priority=550, in_port=%s, %s, %s_src=%s, %s_dst=%s, "+
+								"actions=ct(commit,zone=%d,table=2)",
+								defaultOpenFlowCookie, ofPortHost, protoPrefix, protoPrefix,
+								matchingIPFamilySubnet.String(), protoPrefix, svcCIDR, config.Default.HostMasqConntrackZone))
+					}
+				}
+			}
 		}
 
 		masqDst := masqIP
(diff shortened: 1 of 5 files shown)


test/e2e/route_advertisements.go
```diff
@@ -837,9 +837,11 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
 				func(ipFamilyIndex int) (clientName string, clientNamespace string, dst string, expectedOutput string, expectErr bool) {
```
tssurya, line 837: I think I can easily add a kapi server test to ensure that part works

tssurya:
@kyrtapz this is ready for your eagle eyes

kyrtapz:
@tssurya the service isolation tests are failing in lgw: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14603856307/job/40969980347?pr=5192

test/e2e/route_advertisements.go
kyrtapz: Good idea! lets do that :)

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -411,18 +411,29 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
...
 			// table 2, user-defined network host -> OVN towards default cluster network services
 			defaultNetConfig := npw.ofm.defaultBridge.getActiveNetworkBridgeConfig(types.DefaultNetworkName)
-
-			npw.ofm.updateFlowCacheEntry(key, []string{fmt.Sprintf("cookie=%s, priority=300, table=2, %s, %s_src=%s, %s_dst=%s, "+
+			// sample flow: cookie=0xdeff105, duration=2319.685s, table=2, n_packets=496, n_bytes=67111, priority=300,
+			//              ip,nw_dst=10.96.0.1 actions=mod_dl_dst:02:42:ac:12:00:03,output:"patch-breth0_ov"
+			// This flow is used for UDNs and advertised UDNs to be able to reach kapi and dns services alone on default network
+			flows := []string{fmt.Sprintf("cookie=%s, priority=300, table=2, %s, %s_dst=%s, "+
 				"actions=set_field:%s->eth_dst,output:%s",
-				defaultOpenFlowCookie, ipPrefix, ipPrefix, masqueradeSubnet, ipPrefix, service.Spec.ClusterIP,
```
kyrtapz, line 424: Can we still use the old flow when route advertisement is not enabled?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1593,6 +1604,36 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
+					if netConfig.isDefaultNetwork() {
+						continue
+					}
+					if netConfig.advertised.Load() {
+						var udnAdvertisedSubnets []*net.IPNet
+						for _, clusterEntry := range netConfig.subnets {
+							udnAdvertisedSubnets = append(udnAdvertisedSubnets, clusterEntry.CIDR)
+						}
+						// Filter subnets based on the clusterIP service family
+						// NOTE: We don't support more than 1 subnet CIDR of same family type; we only pick the first one
```
kyrtapz, line 1622: I wasn't aware, are we really not supporting it?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1593,6 +1604,36 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
+						matchingIPFamilySubnet, err := util.MatchFirstIPNetFamily(utilnet.IsIPv6CIDR(svcCIDR), udnAdvertisedSubnets)
```
kyrtapz, line 1623: What about single stack UDNs on dualstack clusters?

go-controller/pkg/node/gateway_shared_intf.go
tssurya: its more like we don't have tests for it... - at least for primary UDNs I don't see a single test that has more than 1 cidr

go-controller/pkg/node/gateway_shared_intf.go
tssurya: good point. I guess I can just ignore the error then and continue with a warning?

go-controller/pkg/node/gateway_shared_intf.go
tssurya: yeah that is what I intended... hmm what did I do here.. why is this flow changing?

tssurya: ah no because we decided we don't have a use case for matching on srcIP right?

go-controller/pkg/node/gateway_shared_intf.go
tssurya: ah found it:
```
	// Dual-stack clusters may set 2 subnets (one for each IP family), otherwise only 1 subnet is allowed.
	// Given subnet is split into smaller subnets for every node.
	//
	// +kubebuilder:validation:MinItems=1
	// +kubebuilder:validation:MaxItems=2
	// +required
	// +kubebuilder:validation:XValidation:rule="size(self) != 2 || !isCIDR(self[0].cidr) || !isCIDR(self[1].cidr) || cidr(self[0].cidr).ip().family() != cidr(self[1].cidr).ip().family()", message="When 2 CIDRs are set, they must be from different IP families"
	Subnets []Layer3Subnet `json:"subnets,omitempty"`
```
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/ad9b16886f60f8d1cb743a05c566437835af68e5/go-controller/pkg/crd/userdefinednetwork/v1/shared.go#L51

tssurya: we have max length set to 2 items and that 1 subnet must be from each diff family in validation. I remember as part of primary UDN design we don't allow more than 1 from same family discussion that happened

go-controller/pkg/node/gateway_shared_intf.go
tssurya: log and continue

tssurya: Info and continue

test/e2e/route_advertisements.go
tssurya: done

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1706,10 +1748,26 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
+			if util.IsRouteAdvertisementsEnabled() && netConfig.advertised.Load() {
+				var udnAdvertisedSubnets []*net.IPNet
+				for _, clusterEntry := range netConfig.subnets {
+					udnAdvertisedSubnets = append(udnAdvertisedSubnets, clusterEntry.CIDR)
+				}
+				// Filter subnets based on the clusterIP service family
+				// NOTE: We don't support more than 1 subnet CIDR of same family type; we only pick the first one
+				matchingIPFamilySubnet, err := util.MatchFirstIPNetFamily(false, udnAdvertisedSubnets)
+				if err != nil {
+					return nil, fmt.Errorf("unable to determine IPv4 UDN subnet: %v", err)
```
kyrtapz, line 1761: Why do you return an error here?

tssurya: as discussed, changing this and also the one below to not return error, but continue
reason: supporting singlestack UDNs on dualstack

PR #5194, opened 2025-04-22
Add Nadia Pinaeva to maintainer's list

tssurya:
Thanks @dcbw for all your contributions! Whenever you choose to come back the process to get you back to maintainer should be fast as per our governance.

Welcome @npinaeva to maintainer role. It's more than just PR merges LOL, Happy to have you share that burden :)

cc @trozet @girishmg @jcaamano

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -2,11 +2,17 @@ The current Maintainers Group for the ovn-kubernetes Project consists of:
 
 | Name | Employer | Responsibilities |
 | ---- | -------- | ---------------- |
-| [Dan Williams](https://github.com/dcbw) | Red Hat  | All things ovnkube |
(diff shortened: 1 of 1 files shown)


npinaeva:
thank you for following-up on this @tssurya!
I am wondering if need to update https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/OWNERS too or whether we use it at all?

tssurya:
> thank you for following-up on this @tssurya! I am wondering if need to update https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/OWNERS too or whether we use it at all?

AFAIK it's useless, not used. @trozet do you use it for random review allocation? 
All internal review/approver things are controlled using github team settings.

Regardless if it is needed to be synced I can

trozet (approved):

PR #5200, opened 2025-04-24
Configure isolation and routes during network addition

kyrtapz:
This change fixes a case where we add a network that is already advertised. We should configure the missing bits on startup too, not only during reconciliation.

I was wondering about extracting and reusing the code already present in `doReconcile` but it is going to require some refactor around `udng.gateway.Reconcile` that also calls `openflowManager.updateBridgeFlowCache` so I want to defer that for the future. 

cc: @tssurya @arghosh93 @jcaamano

### DIFF
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -363,17 +363,16 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	isNetworkAdvertised := util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)
 
 	// create the iprules for this network
-	err = udng.updateUDNVRFIPRule()
-	if err != nil {
+	if err = udng.updateUDNVRFIPRules(isNetworkAdvertised); err != nil {
 		return fmt.Errorf("failed to update IP rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
 	if err = udng.updateAdvertisedUDNIsolationRules(isNetworkAdvertised); err != nil {
-		return fmt.Errorf("failed to configure isolation rules for network %s: %w", udng.GetNetworkName(), err)
+		return fmt.Errorf("failed to update isolation rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
 	if err := udng.updateUDNVRFIPRoute(isNetworkAdvertised); err != nil {
-		return fmt.Errorf("failed to configure ip routes for network %s: %w", udng.GetNetworkName(), err)
+		return fmt.Errorf("failed to update ip routes for network %s: %w", udng.GetNetworkName(), err)
 	}
 
 	// add loose mode for rp filter on management port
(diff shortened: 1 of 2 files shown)


go-controller/pkg/node/gateway_udn.go
```diff
@@ -359,11 +359,23 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	if err = udng.vrfManager.AddVRFRoutes(vrfDeviceName, routes); err != nil {
 		return fmt.Errorf("could not add VRF %s routes for network %s, err: %v", vrfDeviceName, udng.GetNetworkName(), err)
 	}
+
+	isNetworkAdvertised := util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)
```
tssurya, line 363: so now I am getting a bit confused, are we going to introduce the notion of BGP into `AddNetwork` itself?
to answer myself - yes because if the RA is created before the CUDN @kyrtapz want's that info of RA being present being done during CUDN creation I guess?

I guess this is because `doReconcile` is not getting called when a normal CUDN network get's added freshly since there is no previous state.

its a bit sad that now we have two different event ordered paths for advertised udns add -> this function AddNetwork and non-advertised udn adds followed by update to make them advertised -> doReconcile

Maybe AddorUpdateNetwork was the better design choice but that is not related to this specific PR. I guess we need to make gateway code also all level driven at some point.

tssurya: I assume no problem is there on delete ordering ?

- RA is deleted which calls doReconcile which puts flows back to UDN-y and then CUDN is deleted so delNetwork is called
- CUDN is deleted alltogether so delNetwork is called .. hm in this case you correctly already remove the subnets i guess

tssurya (commented):
I see missing unit test coverage @arghosh93 as part of original PR in gateway side:

do we have coverage for : RA created and then CUDN is created @arghosh93 ? can't find any.. I won't block this PR for that because that UT was supposed to be added in the original PR  (I have opened a downstream tracker for this)

tssurya (approved):

go-controller/pkg/node/gateway_udn.go
```diff
@@ -359,11 +359,22 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
...
-	if err != nil {
+	if err = udng.updateUDNVRFIPRules(isNetworkAdvertised); err != nil {
 		return fmt.Errorf("failed to update IP rules for network %s: %w", udng.GetNetworkName(), err)
 	}
+
+	if err = udng.updateAdvertisedUDNIsolationRules(isNetworkAdvertised); err != nil {
+		return fmt.Errorf("failed to update isolation rules for network %s: %w", udng.GetNetworkName(), err)
+	}
+
+	if err := udng.updateUDNVRFIPRoute(isNetworkAdvertised); err != nil {
```
arghosh93, line 374: It is a repetitive task in my opinion. Here you are basically removing the default route based on whether the network is advertised or not. But we are not adding default route in first place while the network is advertised.
We are calling computeRoutesForUDN() to compute routes for a network and there we call getDefaultRoute() to get default route. 
getDefaultRoute() returns nil if network is advertised but not on default vrf. Otherwise it returns expected default route.
```
func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool) ([]netlink.Route, error) {
	vrfs := udng.GetPodNetworkAdvertisedOnNodeVRFs(udng.node.Name)
	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
	if isNetworkAdvertised && !slices.Contains(vrfs, types.DefaultNetworkName) {
		return nil, nil
	}
```

kyrtapz: This is basically a noop in most cases but is required to cover a corner case where a network stops or starts being advertised when ovn-k is out. in such case we need to make sure the state is correct on startup to avoid any staleness.

go-controller/pkg/node/gateway_udn.go
kyrtapz: Delete is covered already yeah.
We do need to handle advertisement config on Add as well to cover the case you mentioned. this was already done for IP rules but was less explicit(I refactored that in the second commit).

go-controller/pkg/node/gateway_udn.go
arghosh93: ohh okay, understood

kyrtapz:
> I see missing unit test coverage @arghosh93 as part of original PR in gateway side:
> 
> do we have coverage for : RA created and then CUDN is created @arghosh93 ? can't find any.. I won't block this PR for that because that UT was supposed to be added in the original PR (I have opened a downstream tracker for this)

I don't think we have any, I can take the task of adding the UT for this case since.

PR #5211, opened 2025-04-28
Run all control plane tests for bgp lane

jcaamano:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


jcaamano:
Why does it always happen to me :sob:

test/scripts/e2e-cp.sh
```diff
@@ -187,10 +187,10 @@ FOCUS=$(echo ${@:1} | sed 's/ /\\s/g')
 pushd e2e
 
 go mod download
-go test -test.timeout 180m -v . \
+go test -test.timeout 185m -v . \
```
pperiyasamy: nit: may be it's good add a comment on why test.timeout is higher than ginkgo.timeout value ?

contrib/kind-common
```diff
@@ -835,7 +842,7 @@ EOF
         # Add IPv4 route
         if [ -n "$ipv4_subnet" ] && [ -n "$node_ipv4" ]; then
           echo "Adding IPv4 route for $node ($node_ipv4): $ipv4_subnet"
-          sudo ip route add $ipv4_subnet via $node_ipv4
+          sudo ip route replace $ipv4_subnet via $node_ipv4
```
pperiyasamy, line 845: any specific reason for using ip route replace command now ?

test/scripts/e2e-cp.sh
```diff
@@ -33,147 +33,152 @@ queries to the hostNetworked server pod on another node shall work for UDP|\
 ipv4 pod"
 
 SKIPPED_TESTS=""
+skip() {
```
pperiyasamy, line 36: nice cleanup!

contrib/kind-common
jcaamano: so that it doesn't fail we you re-run this locally

test/scripts/e2e-cp.sh
```diff
@@ -184,13 +184,18 @@ export NUM_NODES=2
 
 FOCUS=$(echo ${@:1} | sed 's/ /\\s/g')
 
+# Ginkgo test timeout needs to be lower than both github's timeout and go test
+# timeout to be able to get proper Ginkgo output when it happens.
+TEST_TIMEOUT=${TEST_TIMEOUT:-180}
```
tssurya, line 194: so this is the timeout for a single test suite? so individual test suite max will take 180?

test/scripts/e2e-cp.sh
```diff
@@ -184,13 +184,18 @@ export NUM_NODES=2
...
+GO_TEST_TIMEOUT=$((TEST_TIMEOUT + 5))
```
tssurya, line 195: and this is the timeout for the entire test process? so TOTAL_TEST_TIMEOUT?

.github/workflows/test.yml
```diff
@@ -616,7 +616,9 @@ jobs:
     - name: Run Tests
       # e2e tests take ~60 minutes normally, 120 should be more than enough
       # set 3 hours for control-plane tests as these might take a while
-      timeout-minutes: ${{ matrix.target == 'control-plane' && 180 || matrix.target == 'control-plane-helm' && 180 || matrix.target == 'external-gateway' && 180 || 120 }}
+      # give 10m extra to give ginkgo chance to timeout before github so that we
```
tssurya, line 619: ok so self:notes
CP tests:
each test suite timeout is 180
total test suite timeout is 185
github actions timeout for the CI lane is 190 huh..

shard tests:
each test suite timeout is 120
github timeout is 130

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1676,7 +1676,7 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
 			// at the GR load balancer or switch load balancer. It means the correct port wasn't provided.
 			// nodeCIDR->serviceCIDR traffic flow is internal and it shouldn't be carried to outside the cluster
 			dftFlows = append(dftFlows,
-				fmt.Sprintf("cookie=%s, priority=105, in_port=%s, %s, %s_dst=%s,"+
+				fmt.Sprintf("cookie=%s, priority=115, in_port=%s, %s, %s_dst=%s,"+
```
tssurya, line 1679: original bump to 115 makes sense to me.
I know its not related to your change here
but while at it.
I think we should also bump
```
 cookie=0xdeff105, duration=52513.775s, table=0, n_packets=0, n_bytes=0, priority=105,pkt_mark=0x3f0,ip,in_port="patch-breth0_ov",dl_src=fa:4e:7b:13:f6:9c actions=ct(commit,
zone=64000,nat(src=172.18.0.2),exec(load:0x1->NXM_NX_CT_MARK[])),output:eth0                                                                                                 
 cookie=0xdeff105, duration=19801.234s, table=0, n_packets=0, n_bytes=0, priority=105,pkt_mark=0x3f0,ip,in_port=100,dl_src=fa:4e:7b:13:f6:9c actions=ct(commit,zone=64000,nat
(src=172.18.0.2),exec(load:0x65->NXM_NX_CT_MARK[])),output:eth0                                                                                                              
 cookie=0xdeff105, duration=19779.945s, table=0, n_packets=0, n_bytes=0, priority=105,pkt_mark=0x3f0,ip,in_port=101,dl_src=fa:4e:7b:13:f6:9c actions=ct(commit,zone=64000,nat
(src=172.18.0.2),exec(load:0x66->NXM_NX_CT_MARK[])),output:eth0
```
to 115? maybe it can be seperate fix but just noticed that the 109 flows for ICNI/BGP are going to send the packets out eth0 but in case of EIP pods that are trying to talk to other nodes in the cluster we need to SNAT to nodeIP if the mark matches so I supposed these can also be at 115..

looks like there are some docs here: https://ovn-kubernetes.io/features/cluster-egress-controls/egress-ip/#pod-to-node-ip-traffic which is nice

as a stretch goal if you can add a section for clusterIP or BGP flows in devel docs that would be nice - but improving docs is more of a wider community effort for documenting flows I suppose.

.github/workflows/test.yml
```diff
@@ -464,15 +464,15 @@ jobs:
           - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv6", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "bgp", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "snatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation"}
-          - {"target": "bgp", "ha": "noHA", "gateway-mode": "shared",  "ipfamily": "ipv4", "disable-snat-multiple-gws": "snatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation"}
```
tssurya, line 467:
```
Make both dualstack, not much value testing IPv4 single stack
```
are you sure? sometimes singlestack v6 is forgotten to be handled by folks but perhaps singlestackv4 in itself doesn't add though it is the most used setup today from what I know.

.github/workflows/test.yml
```diff
@@ -464,15 +464,15 @@ jobs:
...
+          - {"target": "bgp", "ha": "noHA", "gateway-mode": "shared",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation"}
```
tssurya, line 467:
```
Make one of the lanes noSnatGW to get signal from that as well
```
its cause we use in downstream and for the sake of BGP here - this is moot correct? today if default net  is advertised using bgp the per-pod-snat or pod-subnet snat is always OFF right? so just like ICNI disable-snat is what we use but switched off for the whole network?

for udns we only add the snat to whole subnet today not per-pod snat so even there for bgp i assume its just turned off?

.github/workflows/test.yml
```diff
@@ -464,15 +464,15 @@ jobs:
...
     needs: [ build-pr ]
     env:
       JOB_NAME: "${{ matrix.target }}-${{ matrix.ha }}-${{ matrix.gateway-mode }}-${{ matrix.ipfamily }}-${{ matrix.disable-snat-multiple-gws }}-${{ matrix.second-bridge }}-${{ matrix.ic }}"
       OVN_HYBRID_OVERLAY_ENABLE: ${{ (matrix.target == 'control-plane' || matrix.target == 'control-plane-helm') && (matrix.ipfamily == 'ipv4' || matrix.ipfamily == 'dualstack' ) }}
-      OVN_MULTICAST_ENABLE:  "${{ matrix.target == 'control-plane' || matrix.target == 'control-plane-helm' || matrix.target == 'network-segmentation' }}"
-      OVN_EMPTY_LB_EVENTS: "${{ matrix.target == 'control-plane' || matrix.target == 'control-plane-helm' }}"
+      OVN_MULTICAST_ENABLE:  "${{ matrix.target == 'control-plane' || matrix.target == 'control-plane-helm' || matrix.target == 'network-segmentation' || matrix.target == 'bgp' }}"
+      OVN_EMPTY_LB_EVENTS: "${{ matrix.target == 'control-plane' || matrix.target == 'control-plane-helm' || matrix.target == 'bgp' }}"
```
tssurya, line 475: so this is mainly for CP regression tests but only for default network i guess so we don't need net-seg enabled here.

contrib/kind-common
```diff
@@ -751,6 +751,13 @@ deploy_bgp_external_server() {
     echo "FRR kind network IPv6: ${bgp_network_frr_v6}"
     $OCI_BIN exec bgpserver ip -6 route replace default via "$bgp_network_frr_v6"
   fi
+  # disable the default route to make sure the container only routes accross
+  # directly connected or learnt networks (doing this at the very end since
+  # docker changes the routing table when a new network is connected)
+  docker exec frr ip route delete default
```
tssurya, line 757: don't we want to gate this with if  [ "$PLATFORM_IPV6_SUPPORT" == true ] ; then and if  [ "$PLATFORM_IPV4_SUPPORT" == true ] ; then? based on kind.sh -i6 -n4 flags?

test/scripts/e2e-cp.sh
tssurya: yeaa looks more clean now!

test/scripts/e2e-cp.sh
```diff
@@ -33,147 +33,152 @@ queries to the hostNetworked server pod on another node shall work for UDP|\
...
+    # TODO: perhaps the secondary network attached pods should not be attached to default network
+    skip "Multi Homing A single pod with an OVN-K secondary network attached to a localnet network mapped to breth0 can be reached by a client pod in the default network on the same node"
+    skip "Multi Homing A single pod with an OVN-K secondary network attached to a localnet network mapped to breth0 can be reached by a client pod in the default network on a different node"
+  
+    # these tests require metallb but the configuration we do for it is not compatible with the configuration we do to advertise the default network
+    # TODO: consolidate configuration
+    skip "Load Balancer Service Tests with MetalLB"
+    skip "EgressService"
+
+    # tests that specifically expect the node SNAT to happen
```
tssurya, line 156: oh for EIP.. the ones that expect the node SNAT to happen is cause of the flow in br-ex being lower prio i assume?

tssurya:
thanks for splitting into commits, it really helped, some inline questions but I am mostly OK on the core changes.

test/scripts/e2e-cp.sh
jcaamano: No. Ginkgo runs inside a `go test` process that has its own timeout. Both are total timeouts. But Ginkgo is more like an inner timeout, and go's is more like an outer timeout.

.github/workflows/test.yml
jcaamano: But dualstack is like IPv4 single stack with IPv6 stack added. So in terms of integration it adds no value. It is useful to test for IPv4 specific bits in our code but there are not much and are already covered nicely in unit tests. So I think adding IPv6 coverage to both gateway modes adds more value. 

We don't support IPv6 single stack right now so I didn't consider that. Although it would probably work in kind since docker does not really offer a real single stack IPv6 network but again, not much more value than dualstack if the network ends up being primarily an IPv4 network.

.github/workflows/test.yml
jcaamano: Right so the kind of SNATs that need to be handled (deleted or added) depending on this config are different. In this case I found the coverage interesting since we are integrating with other parts of the code that are not BGP specific.

.github/workflows/test.yml
jcaamano: hmm no control-plane lane enables net-seg? But yeah, I didn't need that for any test running on the lane

contrib/kind-common
jcaamano: No need. Both stacks are enabled anyway and the command doesn't fail if the route doesn't exist.

test/scripts/e2e-cp.sh
jcaamano: No, they just expect the SNAT because that is what happens if the network is not advertised. The tests are not prepared for an advertised network.

go-controller/pkg/node/gateway_shared_intf.go
jcaamano: Do you think something is broken? The 109 flow matches on pod subnet, but the EIP 105 flows should be for traffic already SNATed to EIP. Regardless. I don't think this is a "while at it" thing. The flow I changed is because something was broken for which I had coverage. With these flows, what is broken? Is there coverage? Do we need to add new tests? If I were just to move them then I would have the feeling of doing so blindly.

But If you think I should just bump them (not to 115 because it would defeat the purpose, 113 or something) without any other concern, let me know, will do that.

Same goes for documentation. In this case we are just missing overall BGP documentation upstream, I will add a story for it since I do think we need it

test/scripts/e2e-cp.sh
tssurya: ack

.github/workflows/test.yml
tssurya:
```
 It is useful to test for IPv4 specific bits in our code but there are not much and are already covered nicely in unit tests. So I think adding IPv6 coverage to both gateway modes adds more value.
```
ack agree I would rather have a new lane added is what i meant than removing v4 but its fine. I don't have a strong opinion

test/scripts/e2e-cp.sh
tssurya: but pod to other node should always be SNATed to nodeIP right for EIP? isn't that the flow 105 in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5211#discussion_r2137041123

go-controller/pkg/node/gateway_shared_intf.go
tssurya: I didn't mean a while we are at it thing, I assumed the reason why https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5211#discussion_r2137064167 pod to other node for EIP had to be skipped is because of this - but if you are sure that's not related I am fine not changing it. (I thought things had to be skipped because of this flow but perhaps not.)

cause IIRC we mark the pod to other node in the cluster traffic for EIPs and then match on mark and SNAT to nodeIP. doesn't matter if BGP advertised or not right? EIP matched pod to other nodes in cluster is considered internal to cluster logic so it should be SNATed to nodeIP (I could be mis understanding the e2e that got skipped so feel free to correct me)

jcaamano: Yeah, it doesn't matter if it is advertised with BGP or not.

And the truth is we didn't realize. Otherwise we would have probably attempted to avoid this SNAT for BGP advertised networks. But right now is more aligned to our current intention of SNATing all internal traffic anyways, so all good.

test/scripts/e2e-cp.sh
jcaamano: I am not sure if we are having a two sided discussion for the same topic, in that case we should continue only one of them.

Note that ~~all~~ most EIPs tests toggle EIPs on and off, and when they are off they check that the traffic is SNATed to node IP. So not all the tests I am skipping here are just intra-cluster EIP tests, they are probably not skipped for that reason, it's just a generic thing that test might at some point, for whatever reason, expect an SNAT that doesn't happen if the network is advertised.

go-controller/pkg/node/gateway_shared_intf.go
tssurya: ack sgtm.

test/scripts/e2e-cp.sh
tssurya: yeah both threads were converged. I trust you know why you skipped them and that fault is in the tests and not in code fully

tssurya (approved):

tssurya:
~@jcaamano : I have merged this
But I still think this needs a FUP issue tracker to investigate the skipping of:~
```
    skip "e2e egress IP validation Cluster Default Network \[OVN network\] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts"
```
~so maybe some of those tests are not ready for advertised networks but the piece I was concerned with which is the conflict between:~
```
cookie=0xdeff105, duration=52513.775s, table=0, n_packets=0, n_bytes=0, priority=105,pkt_mark=0x3f0,ip,in_port="patch-breth0_ov",dl_src=fa:4e:7b:13:f6:9c actions=ct(commit,
zone=64000,nat(src=172.18.0.2),exec(load:0x1->NXM_NX_CT_MARK[])),output:eth0 
```
and
```
cookie=0xdeff105, duration=2308.615s, table=0, n_packets=4, n_bytes=376, priority=109,ipv6,in_port="patch-breth0_ov",dl_src=96:b0:34:18:12:7c,ipv6_src=fd00:10:244:1::/64 actions=ct(commit,zone=64000,exec(load:0x1->NXM_NX_CT_MARK[])),output:eth0
```
~still needs to be sorted out
because the test that checks the SNAT is happening to nodeIP for EIP is what is also skipped. So while I understand we are not there yet and maybe this falls under the pod to other host (but also for EIP) SNAT category its good to have it tracked somewhere on investigating the priority of the 105 flow being 113 or something else~

```
The 109 flow matches on pod subnet, but the EIP 105 flows should be for traffic already SNATed to EIP.
```
~while this is also true, I expect also the corner case since the marking is done on LRP level that pkt might not be snated but yes its not that predominant. - but its a lower hanger and can be tracked later.~

EDIT: Given LRP/SNAT is same transact we should be good this scenario won't happen.

Issue #5212, opened 2025-04-28
Make OVNObservability feature configurable


npinaeva:
### What would you like to be added?

OVNObservability is a feature that provides visibility for ovn-kubernetes-managed network by using sampling mechanism.
That means, that network packets generate samples with user-friendly description to give more information about what and why is happening in the network.
Current functionality only allows enabling sampling for all supported features at 100% sampling rate at the same time, the goal is provide an API for more fine-grained configuration.

### Why is this needed?

More granularity allows cluster admin to only observe the features they care about and control the amount of extra load generated by the sampling.

PR #5213, opened 2025-04-28
[OKEP] Add an OVNObservability API enhancement

npinaeva:
EP for https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5212

taken over from https://github.com/openshift/enhancements/pull/1693

Moved some comments from the previous EP ^
cc @dave-tucker

### DIFF
--- a/docs/okeps/okep-5212-ovnobserv-api.md
+++ b/docs/okeps/okep-5212-ovnobserv-api.md
@@ -1,0 +1,241 @@
+# OKEP-5212: OVNObservability API
+
+* Issue: [#5212](https://github.com/ovn-org/ovn-kubernetes/issues/5212)
+
+## Problem Statement
+
+Current functionality only allows enabling sampling for all supported features at 100% sampling rate at the same time, 
+the goal is to provide an API for more fine-grained configuration.
+
+## Goals
+
+- Design a CRD to configure OVN Observability for given user stories
+
+## Non-Goals
+
+- Discuss OVN Observability feature in general
+
+## Introduction
+
+OVN Observability is a feature that provides visibility for ovn-kubernetes-managed network by using sampling mechanism.
+That means, that network packets generate samples with user-friendly description to give more information about what 
+and why is happening in the network.
+This feature is designed to be used with the [netobserv-operator](https://github.com/netobserv/network-observability-operator)
+
+## User-Stories/Use-Cases
+
+- Observability: As a cluster admin, I want to enable OVN Observability integration with netobserv so that I can use network policy correlation feature.
+- Debuggability: As a cluster admin, I want to debug a networking issue using OVN Observability feature.
+
+## Proposed Solution
+
+There is a number of k8s features that can be used to configure OVN Observability right now, and some more that we expect
+to be added in the future.
+Currently supported features are:
+- NetworkPolicy
+- (Baseline)AdminNetworkPolicy
+- EgressFirewall
+- Primary UDN network isolation from the default network
+- Multicast ACLs
+
+
+In the future more observability features, like egressIP and services, but also networking-specific features
+like logical port ingress/egress will be added.
+
+Every feature may set a sampling probability, that defines how often a packet will be sampled, in percent. 100% means every packet
+will be sampled, 0% means no packets will be sampled.
+
+Currently only on/off mode is supported, where on mode turns all available features with 100% probability.
+
+
+### Observability use case
+
+For observability use case, a cluster admin can set a sampling probability for each feature across the whole cluster.
+As sampling configuration may affect cluster performance, not every admin may want all features enabled with 100% probability.
+We expect this type of configuration to change rarely.
+
+This configuration should be handled by ovn-kubernetes to configure sampling, and by netobserv ebpf agent to "subscribe" to
+the samples enabled by this configuration.
+
+### Debuggability use case
+
+Debuggability is expected to be turned on during debugging sessions, and turned off after the session is finished.
+It will most likely have 100% probability for required features, and also may enable more networking-related features, like
+logical port ingress/egress for cluster admins. Debugging configuration is expected to have more granular filters to
+debug specific problems with minimal overhead. For example, per-node or per-namespace filtering may be required.
+
+Debuggability sampling may be used with or without Network Observability Operator being installed:
+- without netobserv, samples may be printed or forwarded to a file by ovnkube-observ binary, which is a part of ovnkube-node pod on every node.
+- with neobserv, samples may be displayed in a similar to observability use case way, but may need another interface to ensure
+  cluster users won't be distracted by samples generated for debugging.
+
+This configuration should be handled by ovn-kubernetes to configure sampling, and by netobserv ebpf agent OR ovnkube-observ script
+to "subscribe" to the samples enabled by this configuration.
+
+### Workflow Description
+
+1. Cluster admin wants to enable OVN Observability for admin-managed features.
+2. Cluster admin created a CR with 100% probabilities for (Baseline)AdminNetworkPolicy and EgressFirewall.
+3. Messages like "Allowed by admin network policy test, direction Egress" appear on netobserv UI.
+4. Cluster admin sees CPU usage by ovnkube pods grows up 10%, and reduces the sampling probability to 50%.
+5. Still most of the flows have correct enrichment with lower overhead.
+6. Admin is happy. Until...
+7. A user from project "bigbank" reports that they can't access the internet from pod A.
+8. Cluster admin enables debugging for project "bigbank" with 100% probability for NetworkPolicies.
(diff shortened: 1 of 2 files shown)


docs/okeps/okep-5212-ovnobserv-api.md
```diff
@@ -0,0 +1,241 @@
...
+	// +optional
+	Status ObservabilityStatus `json:"status,omitempty"`
+}
+
+type ObservabilitySpec struct {
+	// CollectorID is the ID of the collector that should be unique across the cluster, and will be used to receive configured samples.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=1
+	// +required
+	CollectorID int32 `json:"collectorID"`
```
npinaeva, line 131: > So when this CR is created, couldn't OVNK assign an ID, rather than the user having to provide one?

we could do that, but it will require a centralized (aka cluster-manager) watcher to assign IDs. Then we will need to report that ID back via status, so that it can be used with debug script.
I am not sure if this is worth an extra centralized component to automate this part, wdyt?

docs/okeps/okep-5212-ovnobserv-api.md
```diff
@@ -0,0 +1,241 @@
...
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// Filter allows to apply ObservabilityConfig in a granular manner.
+// Currently, it supports node and namespace based filtering.
+// If both node and namespace filters are specifies, they are logically ANDed.
+// +kubebuilder:validation:MinProperties=1
+type Filter struct {
```
npinaeva, line 167: > hmm but the sampling configuration is per feature. So I guess with this design if I want to sample a specific feature on 1 node, and another feature on another node, I would make 2 CRs?

yep, the problem here is that I don't know what kind of filtering will actually be useful, and I would like to avoid over-complicating the API if possible.
I could move namespace-based filter under FeatureConfig, but I think nodeSelector still makes more sense on the config-level (we could add node-based filtering to per-feature filter later if needed). Do you think it will be good enough for the first version?

docs/okeps/okep-5212-ovnobserv-api.md
dave-tucker: Could you not make `ObservabilityStatus` something like this?

```
// ObservabilityStatus contains the observed status of the ObservabilityConfig.
type ObservabilityStatus struct {
        // A collection of CollectorIDs, indexed by nodeName
        CollectorIDs map[string]int32 
	// +listType=map
	// +listMapKey=type
	Conditions []metav1.Condition `json:"conditions,omitempty"`
}
```

Then it's easy enough for users to look up the ID by node name.

dave-tucker, line 131:
```suggestion
    // +kubebuilder:validation:Maximum=4294967295
	CollectorID int64 `json:"collectorID"`
```
Because int32 caps out at 2,147,483,647.

From man ovn-nb(5)
```
set_id: integer, in range 1 to 4,294,967,295
```

dave-tucker: I've thought about this some more having tried to implement this and I'm now -1 on my own suggestion 😆 
Logic as it stands in the OKEP is fine. Given there are over 4M unique IDs I don't think this will be a problem... we store less in the `Status` and it's easier for users of the CRD to work with.

docs/okeps/okep-5212-ovnobserv-api.md
dave-tucker: > Per-namespace filtering can only be applied to the namespaced Observability Features, like NetworkPolicy and EgressFirewall.

I took that to mean that we should validate and reject a CRD like this:
```
apiVersion: observability.ovn.org/v1alpha1
kind: ObservabilityConfig
metadata:
  name: invalid-namespace-filter
spec:
  collectorID: 1
  features:
    - feature: NetworkPolicy
      probability: 100
    - feature: AdminNetworkPolicy
      probability: 100
  filter:
    namespaces:
      - foo
```
because ANP is not namespace-scoped. would be good to clarify the behaviour in the OKEP

npinaeva: yeah good point, I can't tell at this point how that was supposed to work :D 
Also I am not even sure if namespace-based filtering is useful, as said before this is a bit of guessing, so the per-feature sampling rate settings are definitely useful, and for extra filter maybe we should only add node-based for now?

docs/okeps/okep-5212-ovnobserv-api.md
npinaeva: I am not really opposed to changing this, but we could also decide that for this feature 32 bits is enough considering there will probably be only 1 or 2 collectors?

docs/okeps/okep-5212-ovnobserv-api.md
npinaeva: oh past me actually did clarify that:
```
- Per-namespace filtering can only be applied to the namespaced Observability Features, like NetworkPolicy and EgressFirewall.
  Cluster-scoped features, like (Baseline)AdminNetworkPolicy, can only be additionally filtered based on name. Label selectors
  can't be used, as observability module doesn't watch any objects.
```

docs/okeps/okep-5212-ovnobserv-api.md
```diff
@@ -0,0 +1,241 @@
...
+	UDNIsolation       ObservabilityFeature = "UDNIsolation"
+	MulticastIsolation ObservabilityFeature = "MulticastIsolation"
+)
+```
+
+For now, we only have use cases for 2 different `ObservabilityConfig`s in a cluster, and ovn-kubernetes will likely have a
+limit for the number of CRs that will be handled.
+
+For the integration with netobserv, we could use a label on `ObservabilityConfig` to signal that configured samples should be
+reflected via netobserv. For example, `netobserv.io/ovn-observability: "true"`.
```
jotak, line 207: I don't remember why we said that, what would be the purpose for netobserv to watch this configuration ? Is it for getting the CollectorID ? Or there was more than that ?

The alternative being that the admin manually sets the same CollectorID on both sides (it's currently hardcoded in netobserv, but would be something to add to the FlowCollector API)

docs/okeps/okep-5212-ovnobserv-api.md
jotak, line 131: quick question about ids and multiple collectors, @npinaeva @dave-tucker , as I'm taking over this work, we're only talking here about the collector id but is there any change to bring the the groupID (which are currently hard-coded, 123 in ovnk-observ and 10 in netobserv) ? Are they meant to stay hardcoded?

npinaeva: groupID is only important for the samples receiver to know on which psample group to listen.
So it's the samples listener saying "send the samples of collector X to groupID Y". Here it is important to avoid overlapping group IDs (e.g. ovnkube-observ and netobserv should use different group IDs), but otherwise it's a local decision and can even be different on every node. So I think it can stay hardcoded

jotak:
LGTM
@npinaeva @dave-tucker do you see anything in the current ongoing discussions that need a change?

docs/okeps/okep-5212-ovnobserv-api.md
jotak: would there be a reason for users to have an id above the int32 range? E.g. while interfacing with _something_ that already works with ovn-nb and that generates int64 ids ? Could that be a potential use case?

docs/okeps/okep-5212-ovnobserv-api.md
jotak: maybe we can mention that a warning will be surfaced (via the Status/Conditions?) in case of inconsistent config

docs/okeps/okep-5212-ovnobserv-api.md
jotak: (anyway, that's a discussion for netobserv, not important for the implementation in ovn-k)

docs/okeps/okep-5212-ovnobserv-api.md
npinaeva: not that I am aware of.. this ID is only used in ovn-k and then in the local node collectors, so I don't see any external dependencies

docs/okeps/okep-5212-ovnobserv-api.md
npinaeva: yeah it is to remove that hardcoded value (10 I guess in netobserv)
instead netobserv would create this CR with any ID (even 10 :)) and then use it internally

dave-tucker:
LGTM also 😎

dave-tucker (dismissed):

npinaeva:
ok re-pushed after more than a year of waiting :D only a rebase

npinaeva:
ok rabbit has got much more strict since the time this PR was created :D
see the last diff to check some changes

jotak (approved):

PR #5216, opened 2025-04-28
[SDN-3551] LB eps not updated correctly while target port changes

cathy-zhou:
When service endpoint target port updated to a different value, at the time during rolling update, old target port and new target port are both used but associated with different set of target IPs. Current ovn-k8s does not support it. It randomly chose one port to associted with all the current target IPs.

This commit fixes that.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Consolidated load‑balancer targets so multiple rules aggregate backends; made endpoint detection port‑aware to prevent overwritten or missing backends and ensure correct per-port return traffic.

* **New Features**
  * Per‑port endpoint aggregation and per‑port LB handling for NodePort/Local flows, including per‑port masquerade and template value aggregation.
  * Consistent port-key generation for service ports.

* **Tests**
  * Added unit and e2e tests for multi‑port backends and rolling updates that change target ports.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -340,11 +340,10 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 					if len(nodeportFlows) > 0 {
 						npw.ofm.updateFlowCacheEntry(key, nodeportFlows)
 						npw.ofm.updateGroupCacheEntry(key, nodeportGroups)
-						continue
+					} else {
+						npw.ofm.updateGroupCacheEntry(key, nil)
 					}
-					npw.ofm.updateGroupCacheEntry(key, nil)
-				}
-				if config.Gateway.Mode == config.GatewayModeShared {
+				} else if config.Gateway.Mode == config.GatewayModeShared {
 					// case2 (see function description for details)
 					npw.ofm.updateFlowCacheEntry(key, []string{
 						// table=0, matches on service traffic towards nodePort and sends it to OVN pipeline
@@ -524,7 +523,6 @@ func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service,
 		// And then ensure that return traffic is UnDNATed correctly back
 		// to the ingress / external IP
 		isServiceTypeETPLocal := util.ServiceExternalTrafficPolicyLocal(service)
-		hostNetworkFlowsAdded := false
 		if isServiceTypeETPLocal && hasLocalHostNetworkEp {
 			// Get the target port (this is important for named ports).
 			svcPortKey := util.GetServicePortKey(svcPort.Protocol, svcPort.Name)
@@ -558,12 +556,10 @@ func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service,
 			externalIPFlows = append(externalIPFlows, hostNetworkFlows...)
 			if len(hostNetworkFlows) > 0 {
 				npw.ofm.updateGroupCacheEntry(key, hostNetworkGroups)
-				hostNetworkFlowsAdded = true
 			} else {
 				npw.ofm.updateGroupCacheEntry(key, nil)
 			}
-		}
-		if !hostNetworkFlowsAdded && config.Gateway.Mode == config.GatewayModeShared {
+		} else if config.Gateway.Mode == config.GatewayModeShared {
 			// add the ICMP Fragmentation flow for shared gateway mode.
 			icmpFlow := nodeutil.GenerateICMPFragmentationFlow(externalIPOrLBIngressIP, netConfig.OfPortPatch, npw.ofportPhys, cookie, 110)
 			externalIPFlows = append(externalIPFlows, icmpFlow)
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1396,6 +1396,7 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) e
 	}
 
 	if out.hasLocalHostNetworkEp != hasLocalHostNetworkEp ||
+		(hasLocalHostNetworkEp && !reflect.DeepEqual(out.localEndpoints, localEndpoints)) ||
 		(!util.LoadBalancerServiceHasNodePortAllocation(svc) && !reflect.DeepEqual(out.localEndpoints, localEndpoints)) {
 		klog.V(5).Infof("Endpointslice %s ADD event in namespace %s is updating rules", epSlice.Name, epSlice.Namespace)
 		if err = delServiceRules(svc, out.localEndpoints, npw); err != nil {
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -316,24 +316,6 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 						errors = append(errors, err)
 						continue
 					}
-					if len(lbe) > 1 {
-						klog.Warningf("Multiple target ports for service %s/%s port %s, using first",
-							service.Namespace, service.Name, svcPortKey)
-					}
-					// The return traffic, matches on the flowProtocol + targetPort. That's different from cookie which
-					// uses svcPort.NodePort.
-					targetPort := lbe[0].Port
-					targetPortCookie, err := svcToCookie(service.Namespace, service.Name, flowProtocol, targetPort)
-					if err != nil {
-						klog.Warningf("Unable to generate target port cookie for svc: %s, %s, %d, error: %v",
-							service.Namespace, service.Name, targetPort, err)
-						targetPortCookie = "0"
-					}
-
-					// case1 (see function description for details)
-					var nodeportFlows []string
-					klog.V(5).Infof("Adding flows on breth0 for Nodeport Service %s with targetPort %d in "+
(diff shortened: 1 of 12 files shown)


test/e2e/service.go
```diff
@@ -1613,6 +1615,320 @@ spec:
...
+
+		ginkgo.By(fmt.Sprintf("Triggering rolling update: changing target port from %d to %d", oldTargetPort, newTargetPort))
+		deployment, err = cs.AppsV1().Deployments(namespace).Get(ctx, deploymentName, metav1.GetOptions{})
+		framework.ExpectNoError(err)
+		deployment.Spec.Template.Spec.Containers[0].Command = []string{
+			"/agnhost", "serve-hostname", fmt.Sprintf("--port=%d", newTargetPort),
+		}
+		deployment.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort = newTargetPort
+		_, err = cs.AppsV1().Deployments(namespace).Update(ctx, deployment, metav1.UpdateOptions{})
+		framework.ExpectNoError(err)
```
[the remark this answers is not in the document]
cathy-zhou: fixed. please check

test/e2e/service.go
```diff
@@ -1613,6 +1615,320 @@ spec:
...
+				}
+			}
+
+			if portNumbers.Has(oldTargetPort) && portNumbers.Has(newTargetPort) {
+				framework.Logf("Detected both target ports in endpoint slices: %v", portNumbers.UnsortedList())
+				return true, nil
+			}
+			framework.Logf("Current target ports in endpoint slices: %v (waiting for both %d and %d)",
+				portNumbers.UnsortedList(), oldTargetPort, newTargetPort)
+			return false, nil
```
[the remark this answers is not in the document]
cathy-zhou: fixed. please check

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -338,25 +324,40 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
...
+								etpSvcOpenFlowCookie),
+							// table 0, Matches on return traffic, i.e traffic coming from the host networked pod's port, and unDNATs
+							// Use targetPortCookie, as the flow will be the same if the target port + protocol are the same.
+							fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%d, actions=ct(zone=%d nat,table=7)",
+								targetPortCookie, flowProtocol, targetPort, config.Default.HostNodePortConntrackZone),
+							// table 7, Sends the packet back out eth0 to the external client. Note that the constant etp svc
+							// cookie is used since this would be same for all such services.
+							fmt.Sprintf("cookie=%s, priority=110, table=7, "+
+								"actions=output:%s", etpSvcOpenFlowCookie, npw.ofportPhys))
+					}
```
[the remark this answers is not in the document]
cathy-zhou: can you check the current changes

ricky-rav:
@cathy-zhou FYI your e2e tests are passing on a cluster built out of the master branch

ricky-rav:
@cathy-zhou let's try running your e2e tests on our CI against master: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6257/commits

npinaeva:
@cathy-zhou so if the tests are passing on master that should mean that there is either no bug or the test doesn't really catch the problem?

cathy-zhou:
> @cathy-zhou so if the tests are passing on master that should mean that there is either no bug or the test doesn't really catch the problem?

it was incorrect e2e test. I changed the e2e test yesterday and verified it does not pass in master head

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -312,9 +312,13 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 						errors = append(errors, err)
 						continue
 					}
+					if len(lbe) > 1 {
+						klog.Warningf("Multiple target ports for service %s/%s port %s, using first",
+							service.Namespace, service.Name, svcPortKey)
+					}
```
trozet: So it looks like you are leaving this as is because you would have to handle load balancing to multiple ports if there were 2 host networked pods.

Also, this code needs changing because you could have a scenario where you have 2 pods on this node, one is ovn-networked, and one is host networked. In that case if you just pick the first port, you could pick the ovn networked pod accidentally right? So I think you need to check the endpoint is actually host networked here and pick the first one of those.

go-controller/pkg/util/util.go
```diff
@@ -691,38 +691,56 @@ type IPPort struct {
 	Port int32
 }
 
-// LBEndpoints contains load balancer endpoint information with IPv4 and IPv6 addresses.
+// LBEndpointEntry contains load balancer endpoint information for a single target port
```
trozet, line 694: I was very lost at first by the commit message, but I think I understand it now. IIUC this is specific to named ports, because endpoints can vary their numerical port based on the named port, which otherwise is not possible if target port was numerical. Also, the premise that there is a rolling update seems like only one possible scenario. I think in steady state 2 endpoints could have the same named port and point to different ports (without any rolling update) right?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -345,25 +327,50 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
...
+						targetPortCookie, err := svcToCookie(service.Namespace, service.Name, flowProtocol, targetPort)
+						if err != nil {
+							klog.Warningf("Unable to generate target port cookie for svc: %s, %s, %d, error: %v",
+								service.Namespace, service.Name, targetPort, err)
+							targetPortCookie = "0"
+						}
+
+						klog.V(5).Infof("Adding flows on breth0 for Nodeport Service %s with targetPort %d in "+
+							"Namespace: %s since ExternalTrafficPolicy=local", service.Name, targetPort, service.Namespace)
+						if i == 0 {
```
trozet: still doesn't solve multiple target ports for new incoming connections. I think we need to use a openflow group here and load balance between multiple host network endpoints

test/e2e/service.go
```diff
@@ -1619,8 +1621,372 @@ spec:
...
+				framework.Logf("Detected both target ports in endpoint slices: %v", ports.UnsortedList())
+				return true, nil
+			}
+			framework.Logf("Current target ports in endpoint slices: %v (waiting for both %d and %d)",
+				ports.UnsortedList(), oldTargetPort, newTargetPort)
+			return false, nil
+		})
+		framework.ExpectNoError(err, "timed out waiting for both target ports to appear in endpoint slices")
+
+		ginkgo.By("Pausing the rolling update to freeze the multi-port state")
```
trozet, line 1788: I think the test would be better to not do a rolling update and just add a second pod endpoint with a different named port (or maybe a few) and then validate traffic on the original connection is still working correctly.

trozet (changes requested):
I would like to fix the entire bug in this PR so I asked AI to write the openflow group part that would get rid of having to only select the first target port for ingress traffic. It looks OK to me, but needs testing:

https://github.com/ovn-kubernetes/ovn-kubernetes/commit/25ce18f45de42b43a61e72821909328b4e7c1654

Also, need to make sure it will worth with hardware offload, so probably need to run it on the upstream rig.

You can also modify the e2e test you added to ingress multiple new connections and make sure you hit both backends.

go-controller/pkg/node/gateway_shared_intf.go
cathy-zhou: yes, thanks for catching this.

cathy-zhou:
> Also, need to make sure it will worth with hardware offload, so probably need to run it on the upstream rig.

but the flow test group is only used when there are multiple target ports, right, so even it does not work with hardware offload, it is not really an regression.

I will cherry pick your commit into my PR. Thanks!

test/e2e/service.go
cathy-zhou: the initial intend of this PR is for a downstream jira that "OVN LB endpoints not updating properly when the service endpoints change their target ports during rolling update", that is why I added this test case. I kept this test case and added two more test cases for services with multiple endpoints using different numerical values for the same named port.

go-controller/pkg/node/gateway_shared_intf.go
cathy-zhou: cherry-picked your commit into my PR.

cathy-zhou: @trozet please take a look. 3 new test cases are added and they did help to find several issues. I've add fix as new commits after the cherry-picking commits so it is easier for you to review the changes. Once the review is completed, I can squash them into the old commits before merging into upstream.

go-controller/pkg/util/util.go
cathy-zhou: the fix started from a downstream jira that needs the load balancer to be updated correctly with multiple target port when an service is upraded.

test/e2e/service.go
```diff
@@ -1615,8 +1617,883 @@ spec:
...
+			}
+			if !ports.Has(newTargetPort) {
+				return fmt.Errorf("new target port %d not found in endpoint slices: %v", newTargetPort, ports.UnsortedList())
+			}
+			return nil
+		}, 2*time.Minute, framework.Poll).Should(gomega.Succeed(),
+			"after rolling update completes, all endpoint slices should use the new target port")
+	})
+
+	// This test verifies that a ClusterIP service with multiple endpoints
```
trozet, line 1946: thanks for the additional tests

trozet (approved):

PR #5219, opened 2025-04-30
Sync existing nodes at node allocator initialization

pliurh:
## 📑 Description
Initialized the node allocator by syncing the existing nodes. This will mark the existing subnets as allocated in the allocator. It prevents node allocation from allocating a subnet owned by an existing node, if the node is added while the cluster manager is restarting.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## How to verify it
Add a new node to the cluster while restarting the ovnkube-control-plane pod.


go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -411,10 +428,28 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
 					klog.Errorf("Failed to parse hybrid overlay for node %s: %v", node.Name, err)
 				} else if hostSubnet != nil {
 					klog.V(5).Infof("Node %s contains subnets: %v", node.Name, hostSubnet)
-					if err := na.hybridOverlaySubnetAllocator.ReleaseNetworks(node.Name, hostSubnet); err != nil {
+					if err := na.hybridOverlaySubnetAllocator.MarkAllocatedNetworks(node.Name, hostSubnet); err != nil {
 						klog.Errorf("Failed to mark the subnet %v as allocated in the hybrid subnet allocator for node %s: %v", hostSubnet, node.Name, err)
 					}
 				}
+			} else if config.HybridOverlay.Enabled && config.HybridOverlay.ClusterSubnets == nil {
```
kyrtapz, line 435: looking at the implementation of `hasHybridOverlayAllocation()` have you missed the secondary network check here?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -411,10 +428,28 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
...
+				// this is a hybrid overlay node but not managed by the hybrid overlay subnet allocator
+				// so we need to mark the host subnet as allocated in the cluster subnet allocator if it overlaps with the cluste subnets.
+				hostSubnet, err := houtil.ParseHybridOverlayHostSubnet(node)
+				if err != nil {
+					klog.Errorf("Failed to parse hybrid overlay for node %s: %v", node.Name, err)
+				} else if hostSubnet != nil {
+					for _, subnet := range config.Default.ClusterSubnets {
+						maskLength, _ := hostSubnet.Mask.Size()
+						if subnet.CIDR.Contains(hostSubnet.IP) && subnet.HostSubnetLength == maskLength {
```
kyrtapz, line 444: Should we care about a case where the host subnet mask length doesn't match but there is an overlap?

go-controller/pkg/clustermanager/node/node_allocator_test.go
```diff
@@ -59,6 +73,58 @@ func TestController_allocateNodeSubnets(t *testing.T) {
 			wantStr:       []string{"172.16.0.0/24"},
 			allocated:     1,
 			wantErr:       false,
+			existingNodes: []*corev1.Node{
+				testNode,
+			},
+		},
+		{
+			name:          "new node, IPv4 only cluster, the test node is added before node allocator starts",
```
kyrtapz, line 81: Thanks for the UTs!

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -411,10 +428,28 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
 					klog.Errorf("Failed to parse hybrid overlay for node %s: %v", node.Name, err)
 				} else if hostSubnet != nil {
 					klog.V(5).Infof("Node %s contains subnets: %v", node.Name, hostSubnet)
-					if err := na.hybridOverlaySubnetAllocator.ReleaseNetworks(node.Name, hostSubnet); err != nil {
+					if err := na.hybridOverlaySubnetAllocator.MarkAllocatedNetworks(node.Name, hostSubnet); err != nil {
```
kyrtapz, line 431: uh oh, this is an unrelated bug fix right?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -112,6 +112,23 @@ func (na *NodeAllocator) Init() error {
 		}
 	}
 
+	// initialized the subnet allocator by syncing the existing nodes
+	// this will mark the existing subnets as allocated in the allocator
+	existingNodes, err := na.nodeLister.List(labels.Everything())
```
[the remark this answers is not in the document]
pliurh: I just want to ensure that `na.Sync` is invoked before `HandleAddUpdateNodeEvent` is invoked, but I found that it may not be necessary with the code you linked.

go-controller/pkg/clustermanager/node/node_allocator.go
pliurh: right.

go-controller/pkg/clustermanager/node/node_allocator.go
pliurh: In SDN migration scenario, the mask length shall always be the same. But you're right, we shall also consider the other scenarios.

go-controller/pkg/clustermanager/node/subnet_allocator.go
```diff
@@ -456,3 +474,46 @@ func (snr *subnetAllocatorRange) releaseAllNetworks(owner string) {
 		}
 	}
 }
+
+// listAllNetworks returns all networks in the range
+func (snr *subnetAllocatorRange) listAllNetworks() []*net.IPNet {
```
kyrtapz, line 479: this function reuses logic from https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/clustermanager/node/subnet_allocator.go#L368-L428 so I think we should be good here.

kyrtapz (approved):
LGTM

go-controller/pkg/clustermanager/node/subnet_allocator.go
jcaamano: I guess we could extract this to a common function like `range(func(*net.IPNet) bool)`. We might even want to export it.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -230,6 +230,11 @@ func (na *NodeAllocator) HandleAddUpdateNodeEvent(node *corev1.Node) error {
 				}
 				return fmt.Errorf("failed to set hybrid overlay annotations for node %s: %v", node.Name, err)
 			}
+		} else if config.HybridOverlay.Enabled && !na.netInfo.IsSecondary() && config.HybridOverlay.ClusterSubnets == nil {
```
jcaamano, line 233: This could be extracted to an analogous `hasHybridOverlay`

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -634,3 +645,21 @@ func (na *NodeAllocator) hasJoinSubnetAllocation() bool {
 	// we allocate join subnets for L3/L2 primary user defined networks or default network
 	return na.netInfo.IsDefault() || (util.IsNetworkSegmentationSupportEnabled() && na.netInfo.IsPrimaryNetwork())
 }
+
+func (na *NodeAllocator) markAllocatedNetworksForUnmanagedHONode(node *corev1.Node, nodeSubnets []*net.IPNet) {
+	hostSubnet, err := houtil.ParseHybridOverlayHostSubnet(node)
+	if err != nil {
```
jcaamano, line 651: nit 
```
if err != nil {
    klog....
    return
}
if hostSubnet == nil {
   return
}
...
```

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -634,3 +645,21 @@ func (na *NodeAllocator) hasJoinSubnetAllocation() bool {
...
+		klog.Errorf("Failed to parse hybrid overlay for node %s: %v", node.Name, err)
+	} else if hostSubnet != nil {
+		for _, nodeSubnet := range nodeSubnets {
+			// check if the host subnet overlaps with any node subnet that is managed by the cluster subnet allocator
+			// if it does, mark it as allocated in the cluster subnet allocator
+			if utilnet.IsIPv4CIDR(nodeSubnet) == utilnet.IsIPv4CIDR(hostSubnet) && (nodeSubnet.Contains(hostSubnet.IP) || hostSubnet.Contains(nodeSubnet.IP)) {
```
jcaamano, line 657: If would be very nice to have an util function that checks if two sets of networks overlap. It could most likely be based on the already available `IsContainedInAnyCIDR`

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -397,6 +402,8 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
 
 	networkName := na.netInfo.GetNetworkName()
 
+	nodeSubnets := na.clusterSubnetAllocator.ListAllIPv4Networks()
```
jcaamano, line 405: Call this closer to where it is being used

jcaamano: Also before iterating all the allocatable networks, it might be worth it to first check against `netInfo.Subnets()`

jcaamano (changes requested):
What happens if an HO node is removed from the cluster? Shouldn't we set the subnet back as available?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -634,3 +645,21 @@ func (na *NodeAllocator) hasJoinSubnetAllocation() bool {
 	// we allocate join subnets for L3/L2 primary user defined networks or default network
 	return na.netInfo.IsDefault() || (util.IsNetworkSegmentationSupportEnabled() && na.netInfo.IsPrimaryNetwork())
 }
+
+func (na *NodeAllocator) markAllocatedNetworksForUnmanagedHONode(node *corev1.Node, nodeSubnets []*net.IPNet) {
```
jcaamano, line 649: Return error from here. Handle logging in the caller.

go-controller/pkg/clustermanager/node/node_allocator.go
pliurh: We only need to call this function once. It returns all possible node subnets. Therefore, I don't think we can move it in the iteration below.

go-controller/pkg/clustermanager/node/subnet_allocator.go
pliurh: I extract the common logic in function `listAllNetworks` and `allocateNetwork` to a new function. Not sure if that is what you meant.

go-controller/pkg/clustermanager/node/subnet_allocator.go
```diff
@@ -456,3 +456,58 @@ func (snr *subnetAllocatorRange) releaseAllNetworks(owner string) {
...
+	}
+
+	return &net.IPNet{
+		IP:   genIP,
+		Mask: net.CIDRMask(int(snr.subnetBits)+netMaskSize, addrLen),
+	}
+}
+
+// listAllNetworks returns all networks in the range
+func (snr *subnetAllocatorRange) listAllNetworks() []*net.IPNet {
```
jcaamano, line 491: I was thinking on somehting more along this line

```
func (snr *subnetAllocatorRange) foreach(next uint32, do func(*net.IPNet) bool) {
	netMaskSize, addrLen := snr.network.Mask.Size()
	numSubnets := uint32(1) << snr.subnetBits
	if snr.subnetBits > 24 {
		// We need to make sure that the uint32 math below won't overflow. If
		// snr.subnetBits > 32 then numSubnets has already overflowed, but also if
		// numSubnets is between 1<<24 and 1<<32 then "base << (snr.hostBits % 8)"
		// below could overflow if snr.hostBits%8 is non-0. So we cap numSubnets
		// at 1<<24. "16M subnets ought to be enough for anybody."
		numSubnets = 1 << 24
	}

	var i uint32
	for i = 0; i < numSubnets; i++ {
		n := (i + next) % numSubnets
		base := n
		if snr.leftShift != 0 {
			base = ((base << snr.leftShift) & snr.leftMask) | ((base >> snr.rightShift) & snr.rightMask)
		} else if addrLen == 128 && snr.subnetBits >= 16 {
			// Skip the 0 subnet (and other subnets with all 0s in the low word)
			// since the extra 0 word will get compressed out and make the address
			// look different from addresses on other subnets.
			if (base & 0xFFFF) == 0 {
				continue
			}
		}

		genIP := append([]byte{}, []byte(snr.network.IP)...)
		subnetBits := base << (snr.hostBits % 8)
		b := (uint32(addrLen) - snr.hostBits - 1) / 8
		for subnetBits != 0 {
			genIP[b] |= byte(subnetBits)
			subnetBits >>= 8
			b--
		}

		genSubnet := &net.IPNet{IP: genIP, Mask: net.CIDRMask(int(snr.subnetBits)+netMaskSize, addrLen)}
                if !do(genSubnet) {
                   return
                }
       }
}

// allocateNetwork returns a new subnet, or nil if the range is full
func (snr *subnetAllocatorRange) allocateNetwork(owner string) *net.IPNet {
	// Return an already allocated subnet instead of creating a new one if a
	// combination of subnet and node name already exists in the cache
	for nodeSubnet, nodeName := range snr.allocMap {
		if nodeName == owner {
			_, subnet, err := net.ParseCIDR(nodeSubnet)
			if err != nil {
				klog.Errorf("Failed to parse subnet %s for node %s: %v", nodeSubnet, nodeName, err)
				continue
			}
			return subnet
		}
	}

        var subnet *net.IPNet
        foreach(snr.next, func(ipnet *IPNet {
		if _, ok := snr.allocMap[genSubnet.String()]; !ok {
			snr.allocMap[genSubnet.String()] = owner
			snr.next = n + 1
			snr.used++
                        subnet = ipnet
			return false
		}
                return true
        })

        if subnet == nil {
            snr.next = 0
        }

        return subnet
}

// listAllNetworks returns all networks in the range
func (snr *subnetAllocatorRange) listAllNetworks() []*net.IPNet {
	var subnets []*net.IPNet
        foreach(0, func(ipnet *IPNet {
		subnets = append(subnets, genSubnet)
                return true
        })
        return subnets
}
```

That `foreach` can be generically useful to iterate networks in the range from any given point.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -230,6 +236,13 @@ func (na *NodeAllocator) HandleAddUpdateNodeEvent(node *corev1.Node) error {
 				}
 				return fmt.Errorf("failed to set hybrid overlay annotations for node %s: %v", node.Name, err)
 			}
+		} else if na.hasHybridOverlayAllocationUnmanaged() {
+			// this is a hybrid overlay node but not managed by the hybrid overlay subnet allocator
+			// Hybrid overlay is only available for IPv4 for now, so we only need to check for IPv4 subnets
+			nodeSubnets := na.clusterSubnetAllocator.ListAllIPv4Networks()
+			if err := na.markAllocatedNetworksForUnmanagedHONode(node); err != nil {
+				klog.Errorf("Failed to mark the subnet %v as allocated in the cluster subnet allocator for node %s: %v", nodeSubnets, node.Name, err)
```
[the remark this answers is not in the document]
pliurh: No. Will return an error here.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -634,3 +654,39 @@ func (na *NodeAllocator) hasJoinSubnetAllocation() bool {
...
+	for _, clusterSubnet := range na.netInfo.Subnets() {
+		if util.CIDRsOverlap(hostSubnet, clusterSubnet.CIDR) {
+			overlap = true
+			break
+		}
+	}
+	if !overlap {
+		// if the host subnet does not overlap with any cluster subnet, return
+		return nil
+	}
```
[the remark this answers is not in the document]
pliurh: The hybrid overlay node subnet (hostSubnet) may get a different mask length from the nodeSubnets. So one may overlap with multiple nodeSubnets. `CIDRsOverlap` cannot just return the first network, it has to return a list. Then we will still need a `for` here.

go-controller/pkg/clustermanager/node/subnet_allocator.go
pliurh: I combine your code with mine.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -49,6 +49,8 @@ type NodeAllocator struct {
 	netInfo util.NetInfo
 }
 
+var nodeSubnets []*net.IPNet
```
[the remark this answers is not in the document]
jcaamano: I guess it doesn't need to be passed as a parameter, just be a field within?

go-controller/pkg/util/net.go
```diff
@@ -285,6 +285,17 @@ func ContainsCIDR(ipnet1, ipnet2 *net.IPNet) bool {
 	return mask1 <= mask2 && ipnet1.Contains(ipnet2.IP)
 }
 
+// CIDRsOverlap returns true if cidr1 and cidr2 overlap
```
jcaamano, line 288: nit: fix comment

go-controller/hack/lint.sh
```diff
@@ -7,7 +7,7 @@ if [ "$#" -ne 1 ]; then
 fi
 
 $1 run --security-opt label=disable --rm \
-  -v  ${HOME}/.cache/golangci-lint:/cache -e GOLANGCI_LINT_CACHE=/cache \
+  -v  ${HOME}/.cache:/cache -e GOLANGCI_LINT_CACHE=/cache \
```
jcaamano, line 10: where does this change come from?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -230,6 +237,13 @@ func (na *NodeAllocator) HandleAddUpdateNodeEvent(node *corev1.Node) error {
 				}
 				return fmt.Errorf("failed to set hybrid overlay annotations for node %s: %v", node.Name, err)
 			}
+		} else if na.hasHybridOverlayAllocationUnmanaged() {
+			// this is a hybrid overlay node but not managed by the hybrid overlay subnet allocator
+			// Hybrid overlay is only available for IPv4 for now, so we only need to check for IPv4 subnets
+			nodeSubnets := na.clusterSubnetAllocator.ListAllIPv4Networks()
```
jcaamano, line 243: you are already initializing nodeSubnets in `markAllocatedNetworksForUnmanagedHONode` so it wouldn't be needed here

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -380,7 +394,7 @@ func (na *NodeAllocator) HandleDeleteNode(node *corev1.Node) error {
 		return nil
 	}
 
-	if na.hasNodeSubnetAllocation() {
+	if na.hasNodeSubnetAllocation() || na.hasHybridOverlayAllocationUnmanaged() {
 		na.clusterSubnetAllocator.ReleaseAllNetworks(node.Name)
```
jcaamano, line 398: does this mechanism of releasing subnets work for subnets that may have different sizes than the cluster subnets?

pliurh: Yes, `ReleaseAllNetworks` can release all overlapped nodeSubnets occupied by one hybrid overlay node.

go-controller/hack/lint.sh
pliurh: removed

go-controller/pkg/clustermanager/node/subnet_allocator.go
```diff
@@ -456,3 +446,64 @@ func (snr *subnetAllocatorRange) releaseAllNetworks(owner string) {
...
+		genIP[b] |= byte(subnetBits)
+		subnetBits >>= 8
+		b--
+	}
+
+	return &net.IPNet{
+		IP:   genIP,
+		Mask: net.CIDRMask(int(snr.subnetBits)+netMaskSize, addrLen),
+	}
+}
```
jcaamano, line 478: nit: extra space

jcaamano:
registered flake
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5272

PR #5221, opened 2025-04-30
docs, okep: routed ingress support for vms without IPAM

### LINKED ISSUE #5217: Support of routed ingress for L2 UDN without dynamic IP allocation

Allow importing existing VMs from another platform into Kubernetes, connected to an overlay network. This includes:

- Specifying the IP address of the VM for OVNK to use (but not assign) via persistent IP or some other mechanism
- Detect/expose the MAC for the VM to be configured in OVNK for the VM (or disable port security)
- When creating the UDN, be able to specify the gateway IP address, so the migrated VM can keep the same default route.
- When creating the UDN, allow excludeSubnets to be used with L2 or L3 UDNs to ensure OVNK does not use an IP address from the range that VMs have already been assigned outside of the cluster.
- When creating the UDN, we may need a new IPAM mode to indicate to OVNK that it should reserve IPs for internal use (gateway IP potentially, and mp0 IP), but it should not auto assign IPs to pods. Note, we may not need this, but listing it just in case.

### Why is this needed?

Usually VMs running in virtualization platforms are using static IP allocation. This allocation is done by the guest itself, i.e. there is no centralized IPAM.

We are seeing increased interest in SDN for VMs in Kubernetes, and while primary UDN provides lot of the needed functionality, for our typical VMs it may provide too much. The fact that primary UDN handles IP address assignment prevents users from preserving their statically allocated IPs they used on their previous virtualization platform.

maiqueb:
## 📑 Description

This adds a new enhancement describing the feature design to have direct routed ingress for virtlualization workloads without IP allocation from OVN-Kubernetes.

Fixes #5217

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
+++ b/docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
@@ -1,0 +1,116 @@
+# OKEP-5217: Routed ingress support for L2 UDN without dynamic IP allocation
+
+* Issue: [#5217](https://github.com/ovn-org/ovn-kubernetes/issues/5217)
+
+## Problem Statement
+
+There is an increased interest for VMs in Kubernetes, which is generating the
+desire to import VMs from another virtualization platforms.
+
+VMs running in virtualization platforms are usually using static IP allocation,
+where IPs are assigned directly in the guest.
+
+One of the features most sought after by virtualization users is routed ingress
+for L2 UDN, something which - right now - OVN-Kubernetes allows using the BGP
+feature.
+
+The current gap in OVN-Kubernetes is to provide routed ingress to VMs whose IPs
+were assigned statically in the guest. This enhancement plans to bridge that
+gap.
+
+## Goals
+
+- Detect/expose the MAC for the VM to be configured in OVNK for the VM (or disable port security)
+- When creating the UDN, be able to specify the gateway IP address, so the migrated VM can keep the same default route.
+- When creating the UDN, allow excludeSubnets to be used with L2 or L3 UDNs to ensure OVNK does not use an IP address from the range that VMs have already been assigned outside of the cluster.
+- When creating the UDN, we may need a new IPAM mode to indicate to OVNK that it should reserve IPs for internal use (gateway IP potentially, and mp0 IP), but it should not auto assign IPs to pods. Note, we may not need this, but listing it just in case.
+
+## Non-Goals
+
+- `NetworkPolicies`, `MultiNetworkPolicies`, or `Services` to work, given
+OVN-Kubernetes has no awareness of the assigned IP addresses. A way to
+introspect / discover the IP address on the guest may be required to implement
+any of those features.
+
+## Follow-up goals
+
+- Specifying the IP address of the VM for OVNK to use (but not assign) via persistent IP or some other mechanism
+
+## Limitations
+
+This feature is scoped for **virtualization** workloads. Hence, the only type
+of overlay being targeted is `layer2`.
+
+## Introduction
+
+TODO
+
+## User-Stories/Use-Cases
+
+- As a traditional **VM owner** I want to import a VM - whose IPs were
+statically configured - from an existing virtualization platform into
+Kubernetes, attaching it to an overlay network. I want to ingress/egress using
+the same IP address (the one assgined to the VM).
+
+- As a VM owner, I want my VM's interfaces MAC addresses to stay the same when
+the VM is migrated, restarted, or stopped then started, so that the established
+TCP connections do not break.
+
+- As a VM owner, I want to express the intent for my overlay UDN to have a
+subnet defined, but not to feature IP address allocation, so that I can
+set whatever IP addresses I want within the guest.
+
+## Proposed Solution
+
+What is the proposed solution to solve the problem statement?
+
+### API Details
+
+(... details, can point to PR PoC with changes but this section has to be
+explained in depth including details about each API field and validation
+details)
+
+* add details if ovnkube API is changing
+
+### Implementation Details
+
+(... details on what changes will be made to ovnkube to achieve the
+proposal; go as deep as possible; use diagrams wherever it makes sense)
+
+* add details for differences between default mode and interconnect mode if any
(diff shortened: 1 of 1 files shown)


docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
```diff
@@ -0,0 +1,116 @@
...
+for L2 UDN, something which - right now - OVN-Kubernetes allows using the BGP
+feature.
+
+The current gap in OVN-Kubernetes is to provide routed ingress to VMs whose IPs
+were assigned statically in the guest. This enhancement plans to bridge that
+gap.
+
+## Goals
+
+- Detect/expose the MAC for the VM to be configured in OVNK for the VM (or disable port security)
```
qinqon, line 23: We should generalize this, "detect", "expose" or "port security" are implementation details, so we can decide later on the enhancement the alternatives

maybe something like 
```
Allow UDN traffic for virtujal machine guest internal mac addresses.
```

docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
```diff
@@ -0,0 +1,116 @@
...
+- When creating the UDN, be able to specify the gateway IP address, so the migrated VM can keep the same default route.
+- When creating the UDN, allow excludeSubnets to be used with L2 or L3 UDNs to ensure OVNK does not use an IP address from the range that VMs have already been assigned outside of the cluster.
```
qinqon, line 25: Isn't this enhancement just for layer2 ? I was not expected layer3 to be mention here, in fact this is specified at the limitations.

Also if we are going to discover the VM's internal IP then we don't need the excludeSubnets at UDN, ovn-k should just reserve it after reading for whatever mechanism we decide here, or are we talking about other IPs ?

docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
```diff
@@ -0,0 +1,116 @@
...
+## Non-Goals
+
+- `NetworkPolicies`, `MultiNetworkPolicies`, or `Services` to work, given
+OVN-Kubernetes has no awareness of the assigned IP addresses. A way to
+introspect / discover the IP address on the guest may be required to implement
+any of those features.
+
+## Follow-up goals
+
+- Specifying the IP address of the VM for OVNK to use (but not assign) via persistent IP or some other mechanism
```
qinqon, line 37: I would add this as part of the enhancement instead of follow up, it may simplify the mechanism, and make sense to tackle together with mac address.

docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
```diff
@@ -0,0 +1,116 @@
...
+- When creating the UDN, be able to specify the gateway IP address, so the migrated VM can keep the same default route.
+- When creating the UDN, allow excludeSubnets to be used with L2 or L3 UDNs to ensure OVNK does not use an IP address from the range that VMs have already been assigned outside of the cluster.
+- When creating the UDN, we may need a new IPAM mode to indicate to OVNK that it should reserve IPs for internal use (gateway IP potentially, and mp0 IP), but it should not auto assign IPs to pods. Note, we may not need this, but listing it just in case.
+
+## Non-Goals
+
+- `NetworkPolicies`, `MultiNetworkPolicies`, or `Services` to work, given
+OVN-Kubernetes has no awareness of the assigned IP addresses. A way to
+introspect / discover the IP address on the guest may be required to implement
+any of those features.
```
qinqon, line 33: If this is really the case, I don't see the benefict of layer2 vs localnet.

docs/okeps/okep-5217-routed-ingress-l2-udn-without-ipam.md
```diff
@@ -0,0 +1,116 @@
...
+
+- As a VM owner, I want my VM's interfaces MAC addresses to stay the same when
+the VM is migrated, restarted, or stopped then started, so that the established
+TCP connections do not break.
+
+- As a VM owner, I want to express the intent for my overlay UDN to have a
+subnet defined, but not to feature IP address allocation, so that I can
+set whatever IP addresses I want within the guest.
+
+## Proposed Solution
```
qinqon, line 63: The cheapest solution would be to add mac address to IPAMClaim next to the discovered address, so users create IPAMClaim for VMs with static ip/mac and VM follow it.

Another alternative I have see proposed is using ` v1.multus-cni.io/default-network:`, but feels better to have a proper IPAMClaim CRD instead of the NSE annotation struct.

Maybe we can mix the two and ipam extensions renders IPAMClaim into NSE and annotate the virt-launcher, similar to what ovn-k is doing, rendering CUDN/UDN into NAD.

qinqon (changes requested):
First pass, please refasctor it first with what was decided already related to having the IP address from the beginning.

trozet:
does  #5238 supersede this?

maiqueb:
> does #5238 supersede this?

No - this is about having an L2 overlay with a subnet defined, but without OVN-K allocating IP addresses for the workloads. i.e. egress is configured for that subnet, but OVN-K does not handle IP allocation.

trozet:
> > does #5238 supersede this?
> 
> No - this is about having an L2 overlay with a subnet defined, but without OVN-K allocating IP addresses for the workloads. i.e. egress is configured for that subnet, but OVN-K does not handle IP allocation.

Isn't that the same as 5238 with just a giant excludeSubnets and always specifying the static IP for the Pod?

maiqueb:
> > > does #5238 supersede this?
> > 
> > 
> > No - this is about having an L2 overlay with a subnet defined, but without OVN-K allocating IP addresses for the workloads. i.e. egress is configured for that subnet, but OVN-K does not handle IP allocation.
> 
> Isn't that the same as 5238 with just a giant excludeSubnets and always specifying the static IP for the Pod?

I actually have seen this is a new IPAM mode, which we discussed in a previous [openshift enhancement](https://github.com/openshift/enhancements/pull/1725). I guess that could be an option though. A very creative option.

PR #5227, opened 2025-05-05
gateway_mtu should be set on router ports only if node supports it

crnithya:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


crnithya:
@trozet @girishmg PTAL

trozet (approved):

crnithya:
Thank you @trozet and @girishmg

PR #5232, opened 2025-05-07
Fix ovnkube-node startup when UDN runs out of host subnet

pperiyasamy:
When L3 Primary UDN host subnet is exhausted, host subnet may not be allocated for the node(s) by cluster manager, then node is not annotated with subnet on the k8s.ovn.org/node-subnets. When ovnkube-node is restarted, then it goes into CLBO because subnet is not found for the network on the annotation, but it okay UDN to fail on the node, but ovnkube-node should come up and continue to work for other networks. So this PR relaxes subnet annotation check for the UDN network on network controller's syncAll method and makes ovnkube-controller to come up without an error.

## Summary by CodeRabbit

- New Features
  - Event-driven reconciliation for User Defined Networks (UDN), improving responsiveness to node changes.
  - Cluster-aware CIDR selection with support for limited host-subnet configurations.

- Bug Fixes
  - More reliable UDN startup and reconciliation; falls back to self-recovery instead of failing.
  - Reduced unnecessary reconciliations for non-applicable topologies.
  - De-duplication and safe merging of VRF routes to avoid conflicts.
  - Consistent host-subnet manifest formatting.

- Tests
  - Expanded end-to-end coverage for UDN (limited host-subnet, EndpointSlice mirroring, preconfigured L2).
  - Manifest generation updated to honor supported network configurations.

### DIFF
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"slices"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/vishvananda/netlink"
@@ -83,6 +84,8 @@ type UserDefinedNetworkGateway struct {
 	// reconcile channel to signal reconciliation of the gateway on network
 	// configuration changes
 	reconcile chan struct{}
+	// ensure run() is only started once per instance
+	runOnce sync.Once
 
 	// vrfTableId holds the route table ID corresponding to management port interface of the network
 	vrfTableId int
@@ -201,9 +204,16 @@ func (udng *UserDefinedNetworkGateway) addMarkChain() error {
 	return nft.Run(context.TODO(), tx)
 }
 
-// AddNetwork will be responsible to create all plumbings
+// SyncNetwork will be responsible to create all plumbings
 // required by this UDN on the gateway side
 func (udng *UserDefinedNetworkGateway) AddNetwork() error {
+	err := udng.syncNetwork()
+	// run gateway reconciliation loop on network configuration changes
+	udng.run()
+	return err
+}
+
+func (udng *UserDefinedNetworkGateway) syncNetwork() error {
 	if udng.openflowManager == nil {
 		return fmt.Errorf("openflow manager has not been provided for network: %s", udng.NetInfo.GetNetworkName())
 	}
@@ -290,9 +300,6 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 	if err := udng.addMarkChain(); err != nil {
 		return fmt.Errorf("failed to add the service masquerade chain: %w", err)
 	}
-
-	// run gateway reconciliation loop on network configuration changes
-	udng.run()
 
 	return nil
 }
@@ -760,29 +767,31 @@ func addRPFilterLooseModeForManagementPort(mgmtPortName string) error {
 }
 
 func (udng *UserDefinedNetworkGateway) run() {
-	go func() {
-		for range udng.reconcile {
-			err := retry.OnError(
-				wait.Backoff{
-					Duration: 10 * time.Millisecond,
-					Steps:    4,
-					Factor:   5.0,
-				},
-				func(error) bool {
-					select {
-					case _, open := <-udng.reconcile:
-						return open
-					default:
-						return true
-					}
-				},
-				udng.doReconcile,
-			)
-			if err != nil {
-				klog.Errorf("Failed to reconcile gateway for network %s: %v", udng.GetNetworkName(), err)
+	udng.runOnce.Do(func() {
+		go func() {
+			for range udng.reconcile {
+				err := retry.OnError(
+					wait.Backoff{
+						Duration: 10 * time.Millisecond,
+						Steps:    4,
+						Factor:   5.0,
+					},
+					func(error) bool {
+						select {
+						case _, open := <-udng.reconcile:
+							return open
+						default:
+							return true
+						}
+					},
+					udng.syncNetwork,
+				)
+				if err != nil {
+					klog.Errorf("Failed to reconcile gateway for network %s: %v", udng.GetNetworkName(), err)
+				}
 			}
-		}
-	}()
+		}()
+	})
 }
 
 func (udng *UserDefinedNetworkGateway) Reconcile() {
@@ -790,47 +799,6 @@ func (udng *UserDefinedNetworkGateway) Reconcile() {
 	case udng.reconcile <- struct{}{}:
 	default:
 	}
-}
-
-func (udng *UserDefinedNetworkGateway) doReconcile() error {
-	klog.Infof("Reconciling gateway with updates for UDN %s", udng.GetNetworkName())
-
-	// shouldn't happen
-	if udng.openflowManager == nil || udng.openflowManager.defaultBridge == nil {
-		return fmt.Errorf("openflow manager with default bridge configuration has not been provided for network %s", udng.GetNetworkName())
-	}
-
-	udng.updateAdvertisementStatus()
-
-	// update bridge configuration
-	netConfig := udng.openflowManager.defaultBridge.GetNetworkConfig(udng.GetNetworkName())
-	if netConfig == nil {
-		return fmt.Errorf("missing bridge configuration for network %s", udng.GetNetworkName())
-	}
-	netConfig.Advertised.Store(udng.isNetworkAdvertised)
-
-	if err := udng.updateUDNVRFIPRules(); err != nil {
-		return fmt.Errorf("error while updating ip rule for UDN %s: %s", udng.GetNetworkName(), err)
-	}
-
-	if err := udng.updateUDNVRFIPRoute(); err != nil {
-		return fmt.Errorf("error while updating ip route for UDN %s: %s", udng.GetNetworkName(), err)
-	}
-
-	// add below OpenFlows based on the gateway mode and whether the network is advertised or not:
-	// table=1, n_packets=0, n_bytes=0, priority=16,ip,nw_dst=128.192.0.2 actions=LOCAL (Both gateway modes)
-	// table=1, n_packets=0, n_bytes=0, priority=15,ip,nw_dst=128.192.0.0/14 actions=output:3 (shared gateway mode)
-	// necessary service isolation flows based on whether network is advertised or not
-	if err := udng.openflowManager.updateBridgeFlowCache(udng.nodeIPManager.ListAddresses()); err != nil {
-		return fmt.Errorf("error while updating logical flow for UDN %s: %s", udng.GetNetworkName(), err)
-	}
-	// let's sync these flows immediately
-	udng.openflowManager.requestFlowSync()
-
-	if err := udng.updateAdvertisedUDNIsolationRules(); err != nil {
-		return fmt.Errorf("error while updating advertised UDN isolation rules for network %s: %w", udng.GetNetworkName(), err)
-	}
-	return nil
 }
 
 // updateUDNVRFIPRules updates IP rules for a network depending on whether the
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"reflect"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -22,11 +24,13 @@ import (
 	"sigs.k8s.io/knftables"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/generator/udn"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/kube"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/iprulemanager"
 	nodenft "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/nftables"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/vrfmanager"
+	ovnretry "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/retry"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
@@ -96,6 +100,9 @@ type UserDefinedNetworkGateway struct {
 	// save BGP state at the start of reconciliation loop run to handle it consistently throughout the run
 	isNetworkAdvertisedToDefaultVRF bool
 	isNetworkAdvertised             bool
+
+	// retry framework for nodes
+	retryNodes *ovnretry.RetryFramework
 }
 
 func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeLister listers.NodeLister,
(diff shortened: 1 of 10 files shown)


pperiyasamy:
This fix causes cluster unreachable at times when a UDN network (which doesn't have enough subnet to allocate for a node) is created/deleted multiple times. Need to investigate why that happens.

pperiyasamy:
fix is working fine, will add E2E to cover exhausted UDN subnet scenario.

pperiyasamy:
some net-seg tests are flaking, from the initial look those are not related to PR change. running failed jobs again.

pperiyasamy:
> some net-seg tests are flaking, from the initial look those are not related to PR change. running failed jobs again.

The following test is failing consistently.

```
2025-07-31T07:09:16.7229585Z   I0731 07:09:16.722420 62653 network_segmentation_services.go:442] Unexpected error: Failed to verify that Client udn-services-868/udn-client should connect to NodePort service udn-services-868/udn-service on fc00:f853:ccd:e793::4:31264 (node ovn-worker, other node): 
2025-07-31T07:09:16.7231047Z       <wait.errInterrupted>: 
2025-07-31T07:09:16.7231485Z       timed out waiting for the condition
2025-07-31T07:09:16.7231832Z       {
2025-07-31T07:09:16.7232416Z           cause: <*errors.errorString | 0xc0004843b0>{
2025-07-31T07:09:16.7233081Z               s: "timed out waiting for the condition",
2025-07-31T07:09:16.7233417Z           },
2025-07-31T07:09:16.7233587Z       }
2025-07-31T07:09:16.7234666Z   [38;5;9m[FAILED][0m in [It] - /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_services.go:442 [38;5;243m@ 07/31/25 07:09:16.722[0m
2025-07-31T07:09:17.5210856Z   [1mSTEP:[0m dump namespace information after failure [38;5;243m@ 07/31/25 07:09:17.52[0m
2025-07-31T07:09:17.5212006Z   [1mSTEP:[0m Destroying namespace "udn-services-868" for this suite. [38;5;243m@ 07/31/25 07:09:17.52[0m
2025-07-31T07:09:17.5260135Z [38;5;9m• [FAILED] [54.813 seconds][0m
2025-07-31T07:09:17.5262293Z [0mNetwork Segmentation: services [38;5;243mon a user defined primary network [0mshould be reachable through their cluster IP, node port and load balancer [38;5;9m[1m[It] L3 primary UDN, cluster-networked pods, NodePort service[0m [38;5;204m[Feature:NetworkSegmentation][0m
```

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16618453571/job/47091502173?pr=5232

Able to repro the issue at local kind non-ic cluster. This failure happens when this test runs after ovnkube-node pod restart caused new E2E test https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5232/commits/8b34fe1c7b197da1b86b468f089ffb3ffeed702b. Still trying to figure out the root cause.

cc @ricky-rav @tssurya PTAL.

ricky-rav:
@pperiyasamy so are you saying that without the new e2e test you're adding, the existing e2e test on UDN services works?  And that it doesn't work especially because you restart  all ovnkube-node pods?

I'm asking because we already restart ovnkube-node, but only on the node where the client is running:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6082160282cf263efd162c0d04ca81ae3c9ecda7/test/e2e/network_segmentation_services.go#L262

pperiyasamy:
> so are you saying that without the new e2e test you're adding, the existing e2e test on UDN services works? And that it doesn't work especially because you restart all ovnkube-node pods?

@ricky-rav yes, this specific service test doesn't work only for this PR and after an ovnkube-node pod restart happened from the new E2E test.

pperiyasamy:
@ricky-rav the CI is green now, PTAL for code review.

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,13 +245,32 @@ func (c *networkController) syncAll() error {
 	start := time.Now()
 	klog.Infof("%s: syncing all networks", c.name)
 	for _, network := range validNetworks {
-		err := c.syncNetwork(network.GetNetworkName())
+		// Ensure node is allocated with host subnet for the network, When subnet is not allocated
```
ricky-rav:
```suggestion
		// Ensure node is allocated with host subnet for the network. When subnet is not allocated
```

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,13 +245,32 @@ func (c *networkController) syncAll() error {
...
+		// for the node then skip syncing the network and proceed with syncing remaining network.
+		// Return the error at the end when k8s.ovn.org/node-subnets annotation is not set for a
+		// network and let the caller to decide on what to do with it.
```
ricky-rav:
```suggestion
		// network and let the caller decide what to do with it.
```

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,13 +245,32 @@ func (c *networkController) syncAll() error {
...
+				return fmt.Errorf("error retrieving node %s object while syncing network %s: %w",
+					c.node, networkName, err)
+			}
+			_, err = util.ParseNodeHostSubnetAnnotation(node, networkName)
+			if err != nil && util.IsAnnotationNotSetError(err) {
+				klog.Errorf("Ignoring syncing network %s because k8s.ovn.org/node-subnets annotation is not set on node %s",
+					networkName, c.node)
+				continue
+			}
+		}
```
ricky-rav: Do you think we can move this block to a separate function, like `IsNetworkReadyForSync` or a similar name?

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,13 +245,32 @@ func (c *networkController) syncAll() error {
...
+		networkName := network.GetNetworkName()
+		if c.node != "" && util.IsNetworkSegmentationSupportEnabled() && network.IsPrimaryNetwork() &&
+			network.TopologyType() == types.Layer3Topology {
+			node, err := c.nodeLister.Get(c.node)
+			if err != nil {
+				return fmt.Errorf("error retrieving node %s object while syncing network %s: %w",
+					c.node, networkName, err)
+			}
+			_, err = util.ParseNodeHostSubnetAnnotation(node, networkName)
+			if err != nil && util.IsAnnotationNotSetError(err) {
```
ricky-rav: what happens if `err != nil && !util.IsAnnotationNotSetError(err)`? Should we still sync the network as in your changes or should we skip it?

test/e2e/network_segmentation.go
```diff
@@ -1808,6 +1809,47 @@ spec:
 			),
 		)
 	})
+
+	Context("UDN with limited host subnet", func() {
+		It("check ovnkube-node pod restart when node is not assigned with UDN host subnet", func() {
```
ricky-rav:
```suggestion
		It("check ovnkube-node pod restart when node is not assigned a UDN host subnet", func() {
```

test/e2e/network_segmentation.go
```diff
@@ -1808,6 +1809,47 @@ spec:
...
+			framework.ExpectNoError(err)
+			if len(nodes.Items) < 3 {
+				ginkgo.Skip("requires at least 3 Nodes")
+			}
+
+			By("creating the network with limited host subnet configuration")
+			netConfig := networkAttachmentConfigParams{
+				name:     nadName,
+				topology: "layer3",
+				cidr:     correctCIDRFamily(cs, "192.168.100.0/25/26", "2014:100:200::/63/64"),
```
ricky-rav, line 1929: Can you add a brief comment with the reasoning behind your e2e test? Here I understand that you're assigning a pod subnet of /25 to the cluster and of /26 to each node. So there will be only two /26 subnets available and the third node will be left without a subnet. The same goes for IPv6. Is this correct?

test/e2e/network_segmentation.go
```diff
@@ -1808,6 +1809,47 @@ spec:
...
+			for i, node := range nodes.Items {
+				_, err = util.ParseNodeHostSubnetAnnotation(&nodes.Items[i], fmt.Sprintf("%s_%s", f.Namespace.Name, nadName))
+				if err != nil && util.IsAnnotationNotSetError(err) {
+					Expect(restartOVNKubeNodePod(cs, deploymentconfig.Get().OVNKubernetesNamespace(), node.Name)).ShouldNot(HaveOccurred(), "restart of OVNKube node pod must succeed")
+					restartpod = true
+					continue
+				}
+				Expect(err).ShouldNot(HaveOccurred(), "should not be other than host subnet annotation not set error")
+			}
+			Expect(restartpod).To(BeTrue())
```
ricky-rav: Here we're expecting to restart only 1 pod, right? So maybe `restartpod` should be an int that we expect to equal 1.

ricky-rav (changes requested):
Overall looks good, I've left a few comments to improve clarity. Thanks, @pperiyasamy !

ricky-rav:
> > so are you saying that without the new e2e test you're adding, the existing e2e test on UDN services works? And that it doesn't work especially because you restart all ovnkube-node pods?
> 
> @ricky-rav yes, this specific service test doesn't work only for this PR and after an ovnkube-node pod restart happened from the new E2E test.

BTW how did you fix this?

go-controller/pkg/networkmanager/network_controller.go
pperiyasamy: sure, moved this new block into a `isNetworkReadyForSync`  method.

test/e2e/network_segmentation.go
pperiyasamy: yes @ricky-rav , you are right, added comment to reflect this.

test/e2e/network_segmentation.go
pperiyasamy: there can be more than 3 nodes, so adding check now with `>= 1`.

go-controller/pkg/networkmanager/network_controller.go
pperiyasamy: this PR is only handling pod crash only upon missing host subnet annotation, so it fixes a d/s issue https://issues.redhat.com/browse/OCPBUGS-55105. should we cover more cases ?

pperiyasamy:
> > > so are you saying that without the new e2e test you're adding, the existing e2e test on UDN services works? And that it doesn't work especially because you restart all ovnkube-node pods?
> > 
> > 
> > @ricky-rav yes, this specific service test doesn't work only for this PR and after an ovnkube-node pod restart happened from the new E2E test.
> 
> BTW how did you fix this?

Previously `syncAll() `throws such errors back to the caller and it was suppressed in nad controller's Start [method](https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/nad_controller.go#L158).
But this doesn't work and causing E2E test failure, so doing it with `syncAll()` method itself now.

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,19 +245,48 @@ func (c *networkController) syncAll() error {
...
 		if err != nil {
-			return fmt.Errorf("failed to sync network %s: %w", network.GetNetworkName(), err)
+			return fmt.Errorf("failed to sync network %s: %w", networkName, err)
 		}
 	}
 	klog.Infof("%s: finished syncing all networks. Time taken: %s", c.name, time.Since(start))
 	return nil
 }
 
+func (c *networkController) isNetworkReadyForSync(network util.NetInfo) (bool, error) {
```
ricky-rav: I would add a comment to describe the purpose of this function. I think you had a comment in the previous version, I would have included that here.

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,19 +245,48 @@ func (c *networkController) syncAll() error {
...
+		node, err := c.nodeLister.Get(c.node)
+		if err != nil {
+			return false, fmt.Errorf("error retrieving node %s object while syncing network %s: %w",
+				c.node, networkName, err)
+		}
+		_, err = util.ParseNodeHostSubnetAnnotation(node, networkName)
+		if err != nil && util.IsAnnotationNotSetError(err) {
+			klog.Errorf("Network %s is not ready for sync because k8s.ovn.org/node-subnets annotation is not set on node %s",
+				networkName, c.node)
+			return false, nil
```
ricky-rav: Here we shouldn't hide the error from `util.ParseNodeHostSubnetAnnotation`. When `util.IsAnnotationNotSetError` returns false, we should propagate it to the caller.

go-controller/pkg/networkmanager/network_controller.go
ricky-rav: Let's propagate the error at least :)

test/e2e/network_segmentation.go
ricky-rav: All right, so in that case I would be even more precise and keep track of:
- nodes for which `_, err = util.ParseNodeHostSubnetAnnotation` didn't return an error; its number must equal 2
- nodes for which _, err = util.ParseNodeHostSubnetAnnotation returned an error of type annotation not set, just like you're doing now.

ricky-rav (changes requested):
A couple more things to polish, otherwise looks good.

go-controller/pkg/networkmanager/network_controller.go
pperiyasamy: yes @ricky-rav , right. missed it, now propagating other errors.

go-controller/pkg/networkmanager/network_controller.go
pperiyasamy: done.

test/e2e/network_segmentation.go
pperiyasamy: sure, good catch, updated it.

ricky-rav (commented):
Looks good now, thanks Peri!
I will approve it once CI is green :)

test/e2e/network_segmentation.go
```diff
@@ -1908,6 +1909,53 @@ spec:
...
+					restartPodCount++
+					continue
+				}
+				Expect(err).ShouldNot(HaveOccurred(), "should not be other than host subnet annotation not set error")
+				hostsubnetAnnotationSetNodeCount++
+			}
+			Expect(restartPodCount).Should(BeNumerically(">=", 1))
+			Expect(hostsubnetAnnotationSetNodeCount).To(Equal(2))
+		})
+	})
```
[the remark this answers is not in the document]
pperiyasamy: yes, right, done.

pperiyasamy:
@ricky-rav the CI is looking good now, the E2E needed an update for latest master (that is when CIDR contains subnet for both cluster as well as node, then filtering on supported network config won't work, so it's okay skip the filtering).

test/e2e/network_segmentation.go
```diff
@@ -2503,3 +2558,25 @@ func unmarshalPodAnnotationAllNetworks(annotations map[string]string) (map[strin
...
+
+// takes ipv4 and ipv6 cidrs and returns the correct type for the cluster under test
+// this function is useful when cidr has both cluster and host subnets included.
+// example: "192.168.100.0/25/26", "2014:100:200::/63/64"
+func correctCIDRFamily(cs clientset.Interface, ipv4CIDR, ipv6CIDR string) string {
+	return strings.Join(selectCIDRs(cs, ipv4CIDR, ipv6CIDR), ",")
+}
+
+// takes ipv4 and ipv6 cidrs and returns the correct type for the cluster under test
+func selectCIDRs(cs clientset.Interface, ipv4CIDR, ipv6CIDR string) []string {
```
ricky-rav, line 2576: here you probably want to use the existing `filterCIDRs`

ricky-rav (dismissed):
/LGTM
Thanks, @pperiyasamy

test/e2e/network_segmentation.go
pperiyasamy: @ricky-rav right, but `filterCIDRs` don't work when CIDR having two two slashes (/), so using these new functions to support it.

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,19 +245,53 @@ func (c *networkController) syncAll() error {
...
 		}
 	}
 	klog.Infof("%s: finished syncing all networks. Time taken: %s", c.name, time.Since(start))
 	return nil
 }
 
+// isNetworkReadyForSync checks if the network is ready for sync. when a node is not allocated
+// with the host subnet for the given network, then the network can not be ready for sync, so
+// return false.
+func (c *networkController) isNetworkReadyForSync(network util.NetInfo) (bool, error) {
```
npinaeva: One thing that bothers me is that this is a very specific filtering, so if any other error happens it will still crashloop, and the better way is to send failed-on-initial-sync objects to the normal retry loop. But the controller mentions
```
 Certain downstream controllers that handle configuration for multiple networks depend on being
	// aware of all the existing networks on initialization. To achieve that, we need to start existing
// networks synchronously. Otherwise, these controllers might incorrectly assess valid configuration
```
so potentially other controller may cleanup something they shouldn't have because of it...
@jcaamano any opinion on this?

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -245,19 +245,53 @@ func (c *networkController) syncAll() error {
...
-		err := c.syncNetwork(network.GetNetworkName())
+		networkName := network.GetNetworkName()
+		ready, err := c.isNetworkReadyForSync(network)
+		if err != nil {
+			return err
+		}
+		// When network is not ready, then skip syncing the network and proceed with syncing
+		// remaining networks.
+		if !ready {
+			continue
```
npinaeva: I am also wondering how you will trigger network creation when the node gets the right label. It looks like `syncNetwork` will only be called next time on NAD update and nothing watches for the node events?

go-controller/pkg/networkmanager/network_controller.go
jcaamano: I don't think this is good for two reasons.

First, there is an assumption that `syncAll` should sync all networks or return an error (except for networks this process does not manage at all). This change does not comply with the assumption, I guess without addressing the reason the assumption is there in the first place.

Second, making this controller aware of what other controllers are doing (parsing subnet annotations?) takes it down the route of non-maintainability. This controller has a clear responsibility and we should not add more on top without proper abstractions. I would say this needs a different design at the very least.

npinaeva: agree!
@pperiyasamy another approach could be to find the code that returns an error when annotation is not set and replace it with warning and `return nil` (or return `AnnotationNotSetError`). There is no point in returning an error (and retrying) when annotations are unset, because only an update event will "fix" the problem, and retry can't help. So this logic would help reduce the number of retries and "fix" the initialSync, because it won't return an error

go-controller/pkg/networkmanager/network_controller.go
pperiyasamy: I think so, it looks like `syncNetwork` is invoked only for NAD event as seen [here](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/nad_controller.go#L115), so should we handle node events too for label update ?

go-controller/pkg/networkmanager/network_controller.go
pperiyasamy: @npinaeva this is the source (https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_udn.go#L409) of the error, but if we ignore returning error from here, should we handle networks for node events ?

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -50,11 +50,29 @@ func newNetworkController(name, zone, node string, cm ControllerManager, wf watc
...
+		Lister:         nc.nodeLister.List,
+		Reconcile:      func(string) error { return nodeReconciler },
+		ObjNeedsUpdate: nodeNeedsUpdate,
+		Threadiness:    1,
+	}
+	nc.nodeController = controller.NewController(
+		nc.name,
+		nodeConfig,
+	)
+
```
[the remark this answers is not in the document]
pperiyasamy: right, fixed it.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -207,6 +207,15 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
...
 	}
+	_, err := util.ParseNodeHostSubnetAnnotation(udng.node, udng.GetNetworkName())
+	if err != nil && util.IsAnnotationNotSetError(err) {
+		klog.Warningf("Network %s is not ready for sync because k8s.ovn.org/node-subnets annotation is not set on node",
+			udng.GetNetworkName())
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("failed to parse k8s.ovn.org/node-subnets annotation for the network %s: %w",
+			udng.GetNetworkName(), err)
+	}
```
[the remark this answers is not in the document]
pperiyasamy: yes, done.

pperiyasamy:
@npinaeva @jcaamano updated the PR with commit [d9edfac](https://github.com/ovn-kubernetes/ovn-kubernetes/commit/d9edfac8b426ae4abd42f3991dd59f4adb7bd859), this may address your concerns. can you check ?
Meanwhile i'm investigating E2E failures for the new change.

go-controller/pkg/networkmanager/network_controller.go
npinaeva: yeah it is a bit of an unfortunate design, so to successfully create nodeUDNController, all labels have to be already present on the node and the only retry mechanism is the one in `network_controller`. Which also means you can't return no error because it will remove it from the infinite retry loop.
My best idea for now with the least amount of change is to add an option to `nadController` to return an error vs simply log and continue on `syncAll` and afaiu this option should be fine for `NodeControllerManager`.
I don't think it builds any cache on startup and seems fairly decoupled from the other controllers, but also let's ask @jcaamano if that makes sense, as he is more familiar with that code

jcaamano: I talked offline with @pperiyasamy. It seems that the error comes from `UserDefinedNetworkGateway.AddNetwork`. We could do two things:

1. `UserDefinedNetworkGateway` already has a simple reconciliation loop, so we could leverage it for retries as long as whatever it is we retry is idempotent; and avoid returning an error.
2. If we want to take it further than just retries, then we could replace the simple channel based reconciliation loop of `UserDefinedNetworkGateway` for a level driven controller monitoring node updates. This is what it should have been done since the beginning I guess.

In terms of this `syncAll` in network controller, it should have never existed :stuck_out_tongue: and fixing the above would be part of the effort to get rid of it. I guess an additional option here is to just log errors and continue, instead of crash-looping. But I have the feeling that is kinda silencing the problem rather than fixing it, particularly when we have a better alternative available.

pperiyasamy:
The latest PR update has UT failures, will revisit it later.

go-controller/pkg/networkmanager/network_controller.go
npinaeva: hmm I am not sure why syncAll should have never existed? Every controller should have a syncAll to cleanup stale resources on restart?

jcaamano: Sorry, wanted to say calling `syncNetwork` from `syncAll`, not just `syncAll`.

Issue #5231, opened 2025-05-07
node-ip-mac-migration test cases perma-failing on IC


jcaamano:
### What is this issue about?

These test cases
```
  [FAIL] Node IP and MAC address migration when the node IPv4 address is updated when no EgressIPs are configured [It] makes sure that the cluster is still operational (update kubelet first, the IP address later)
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/node_ip_mac_migration.go:289
  [FAIL] Node IP and MAC address migration when the node IPv4 address is updated when no EgressIPs are configured [It] makes sure that the cluster is still operational (update the IP address first, kubelet later)
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/node_ip_mac_migration.go:289
  [FAIL] Node IP and MAC address migration when the node IPv4 address is updated when EgressIPs are configured [It] makes sure that the EgressIP is still operational (update kubelet first, the IP address later)
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/node_ip_mac_migration.go:393
  [FAIL] Node IP and MAC address migration when the node IPv4 address is updated when EgressIPs are configured [It] makes sure that the EgressIP is still operational (update the IP address first, kubelet later)
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/node_ip_mac_migration.go:393
```
seem to be perma-failing in IC configuration
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/14805582718/job/41573968507#step:15:2831

It looks like it started happening around this PR being merged 
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5103

PR #5236, opened 2025-05-07
Update node-encap-ips annotation when primary ip changes

crnithya:
## 📑 Description
This PR fixes the issue of node-encap-ips not getting updated when the node's primary IP changes. This breaks geneve traffic in IC as the nodes rely on this annotation to update their records of remote nodes. 

Fixes # 

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -260,7 +260,7 @@ func (c *addressManager) handleNodePrimaryAddrChange() {
 	}
 	if nodePrimaryAddrChanged && config.Default.EncapIP == "" {
 		klog.Infof("Node primary address changed to %v. Updating OVN encap IP.", c.nodePrimaryAddr)
-		updateOVNEncapIPAndReconnect(c.nodePrimaryAddr)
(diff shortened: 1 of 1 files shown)


crnithya:
Fixes https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5231

crnithya:
@girishmg @cathy-zhou PTAL

cathy-zhou:
the logic is correct, but I am wondering if you need to add error handling.

jcaamano:
> the logic is correct, but I am wondering if you need to add error handling.

error handling in address manager could be improved in general but I think the added code is inline to what is being currently done in there. Did you have a different approach in mind?

cathy-zhou:
> error handling in address manager could be improved in general but I think the added code is inline to what is being currently done in there. Did you have a different approach in mind?

I meant it needs to retry in case of error instead of logging an error and carrying on. But I agree it is general issue in address manager.

crnithya:
I agree, @cathy-zhou, lets open a new issue to properly address this error handling.

jcaamano (approved):

crnithya:
Thank you @cathy-zhou and @jcaamano

Issue #4348, opened 2024-05-10
CI flake: Admin net pol `TestNetworkPolicyV2Conformance` test attempting to use port already in-use


martinkennelly:
```
2024-05-09T17:45:20.1404158Z === RUN   TestNetworkPolicyV2Conformance
2024-05-09T17:45:20.1407058Z     network_policy_v2_test.go:31: Configuring environment for network policy V2 API conformance tests
2024-05-09T17:45:20.1421212Z     network_policy_v2_test.go:53: Starting the network policy V2 conformance test suite
2024-05-09T17:45:20.1422439Z     suite.go:125: Test Setup: Applying base manifests
2024-05-09T17:45:20.2616014Z     apply.go:124: Creating network-policy-conformance-gryffindor Namespace
2024-05-09T17:45:20.2686256Z     apply.go:124: Creating network-policy-conformance-slytherin Namespace
2024-05-09T17:45:20.2850881Z     apply.go:124: Creating network-policy-conformance-hufflepuff Namespace
2024-05-09T17:45:20.2958762Z     apply.go:124: Creating network-policy-conformance-ravenclaw Namespace
2024-05-09T17:45:20.3033351Z     apply.go:124: Creating network-policy-conformance-forbidden-forrest Namespace
2024-05-09T17:45:20.3185458Z     apply.go:124: Creating harry-potter StatefulSet
2024-05-09T17:45:20.3249033Z     apply.go:124: Creating draco-malfoy StatefulSet
2024-05-09T17:45:20.3312093Z     apply.go:124: Creating cedric-diggory StatefulSet
2024-05-09T17:45:20.3584068Z     apply.go:124: Creating luna-lovegood StatefulSet
2024-05-09T17:45:20.4050777Z     apply.go:124: Creating centaur StatefulSet
2024-05-09T17:45:20.4482467Z     suite.go:128: Test Setup: Ensuring Namespaces and Pods from base manifests are ready
2024-05-09T17:45:20.4551943Z     helper.go:123: StatefulSet replicas in namespace network-policy-conformance-gryffindor not rolled out yet. 0/2 replicas are available.
2024-05-09T17:45:21.4627860Z     helper.go:123: StatefulSet replicas in namespace network-policy-conformance-gryffindor not rolled out yet. 0/2 replicas are available.
....
....
[...]

PR #5241, opened 2025-05-08
ovnkube.sh: get_node_zone: default to node name as zone 

flavio-fernandes:
If the `k8s.ovn.org/zone-name` label is not set on the node, the fallback logic now uses the node name as the zone when `ovn_enable_interconnect` is true. Otherwise, the zone defaults to "global" as before.

Also updated the empty string check to use `-z`, which is more idiomatic in Bash.

### DIFF
--- a/dist/images/ovnkube.sh
+++ b/dist/images/ovnkube.sh
@@ -803,8 +803,12 @@ function memory_trim_on_compaction_supported {
 function get_node_zone() {
   zone=$(kubectl --subresource=status --server=${K8S_APISERVER} --token=${k8s_token} --certificate-authority=${K8S_CACERT} \
      get node ${K8S_NODE} -o=jsonpath={'.metadata.labels.k8s\.ovn\.org/zone-name'})
-  if [ "$zone" == "" ]; then
-    zone="global"
+  if [ -z "$zone" ]; then
+    if [[ ${ovn_enable_interconnect} == "true" ]]; then
+      zone="${K8S_NODE}"
+    else
+      zone="global"
+    fi
   fi
   echo "$zone"
 }


dist/images/ovnkube.sh
```diff
@@ -803,8 +803,12 @@ function memory_trim_on_compaction_supported {
 function get_node_zone() {
   zone=$(kubectl --subresource=status --server=${K8S_APISERVER} --token=${k8s_token} --certificate-authority=${K8S_CACERT} \
      get node ${K8S_NODE} -o=jsonpath={'.metadata.labels.k8s\.ovn\.org/zone-name'})
-  if [ "$zone" == "" ]; then
-    zone="global"
+  if [ -z "$zone" ]; then
+    if [[ ${ovn_enable_interconnect} == "true" ]]; then
```
flavio-fernandes, line 807: I know this may be controversial because it would disrupt the upgrade detection, right? One solution for that would be to also consider the number of nodes per zone -- if such info was made available to `ovnkube.sh`. That way we could safely determine that the label is not present and we still want to go ahead with ovn-ic.

npinaeva:
Summary from my side: Currently when nodes don't have any labels in ic mode we default to global zone, which is the closest option to pre-IC world. Since helm unfortunately can't label nodes, helm install has to work without zone labels, so currently there is no way to do helm deploy with single node zones (which is what we actually use in our clusters).
So the question is whether it is reasonable to default to single-node-per-zone and node single-zone config. I don't think it will break any upgrades (considering ovnkube.sh is only used for local/CI deployments)

girishmg (approved):
LGTM.

tssurya:
@npinaeva / @girishmg : Gentle reminder: This is part of the 1.1 release that I want to cut tomorrow, please try to get this merged today. If not I need to push release again to next week.
As maintainer I think both of you should be able to rebase this on master?

npinaeva:
known flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4348

trozet (approved):

PR #5247, opened 2025-05-14
Allow ICMP and ICMPv6 regardless of network policy

trozet:
Adds a global config knob "--allow-icmp-netpol" which adds permissive ACL rules for icmp || icmp6 when network policy default deny rules are generated.

### DIFF
--- a/contrib/kind-helm.sh
+++ b/contrib/kind-helm.sh
@@ -83,6 +83,7 @@ set_default_params() {
   export KIND_IPV4_SUPPORT=true
 
   export OVN_ENABLE_DNSNAMERESOLVER=${OVN_ENABLE_DNSNAMERESOLVER:-false}
+  export OVN_ENABLE_ICMP_NETPOL=${OVN_ENABLE_ICMP_NETPOL:-false}
 }
 
 usage() {
@@ -126,6 +127,7 @@ usage() {
     echo "-wk  | --num-workers                 Number of worker nodes. DEFAULT: 2 workers"
     echo "-cn  | --cluster-name                Configure the kind cluster's name"
     echo "-dns | --enable-dnsnameresolver      Enable DNSNameResolver for resolving the DNS names used in the DNS rules of EgressFirewall."
+    echo "--allow-icmp-netpol                   Allows ICMP and ICMPv6 traffic globally, regardless of network policy rules"
     echo "-ic  | --enable-interconnect         Enable interconnect with each node as a zone (only valid if OVN_HA is false)"
     echo "-npz | --nodes-per-zone              Specify number of nodes per zone (Default 0, which means global zone; >0 means interconnect zone, where 1 for single-node zone, >1 for multi-node zone). If this value > 1, then (total k8s nodes (workers + 1) / num of nodes per zone) should be zero."
     echo ""
@@ -188,6 +190,8 @@ parse_args() {
                                                   ;;
             -dns | --enable-dnsnameresolver )     OVN_ENABLE_DNSNAMERESOLVER=true
                                                   ;;
+            --allow-icmp-netpol )                 OVN_ENABLE_ICMP_NETPOL=true
+                                                  ;;
             -ic | --enable-interconnect )         OVN_ENABLE_INTERCONNECT=true
                                                   ;;
             -npz | --nodes-per-zone )             shift
@@ -228,6 +232,7 @@ print_params() {
      echo "KIND_NUM_MASTER = $KIND_NUM_MASTER"
      echo "KIND_NUM_WORKER = $KIND_NUM_WORKER"
      echo "OVN_ENABLE_DNSNAMERESOLVER= $OVN_ENABLE_DNSNAMERESOLVER"
+     echo "OVN_ENABLE_ICMP_NETPOL= $OVN_ENABLE_ICMP_NETPOL"
      echo "OVN_ENABLE_INTERCONNECT = $OVN_ENABLE_INTERCONNECT"
      if [[ $OVN_ENABLE_INTERCONNECT == true ]]; then
        echo "KIND_NUM_NODES_PER_ZONE = $KIND_NUM_NODES_PER_ZONE"
@@ -421,6 +426,7 @@ helm install ovn-kubernetes . -f "${value_file}" \
           --set global.emptyLbEvents=$(if [ "${OVN_EMPTY_LB_EVENTS}" == "true" ]; then echo "true"; else echo "false"; fi) \
           --set global.enableDNSNameResolver=$(if [ "${OVN_ENABLE_DNSNAMERESOLVER}" == "true" ]; then echo "true"; else echo "false"; fi) \
           --set global.enableNetworkQos=$(if [ "${OVN_NETWORK_QOS_ENABLE}" == "true" ]; then echo "true"; else echo "false"; fi) \
+          --set global.allowICMPNetpol=$(if [ "${OVN_ENABLE_ICMP_NETPOL}" == "true" ]; then echo "true"; else echo "false"; fi) \
           ${ovnkube_db_options}
 EOF
        )
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -154,6 +154,7 @@ usage() {
     echo "-obs | --observability                Enable OVN Observability feature."
     echo "-rae | --enable-route-advertisements  Enable route advertisements"
     echo "-adv | --advertise-default-network    Applies a RouteAdvertisements configuration to advertise the default network on all nodes"
+    echo "--allow-icmp-netpol                   Allows ICMP and ICMPv6 traffic globally, regardless of network policy rules"
     echo ""
 }
 
@@ -360,6 +361,8 @@ parse_args() {
                                                 ;;
             -dns | --enable-dnsnameresolver )   OVN_ENABLE_DNSNAMERESOLVER=true
                                                 ;;
+            --allow-icmp-netpol )               OVN_ENABLE_ICMP_NETPOL=true
+                                                ;;
             -h | --help )                       usage
                                                 exit
                                                 ;;
@@ -447,6 +450,7 @@ print_params() {
      echo "KIND_NUM_WORKER = $KIND_NUM_WORKER"
      echo "OVN_MTU= $OVN_MTU"
      echo "OVN_ENABLE_DNSNAMERESOLVER= $OVN_ENABLE_DNSNAMERESOLVER"
+     echo "OVN_ENABLE_ICMP_NETPOL= $OVN_ENABLE_ICMP_NETPOL"
      echo ""
 }
 
@@ -609,6 +613,7 @@ set_default_params() {
   OVN_ENABLE_INTERCONNECT=${OVN_ENABLE_INTERCONNECT:-false}
   OVN_ENABLE_OVNKUBE_IDENTITY=${OVN_ENABLE_OVNKUBE_IDENTITY:-true}
   OVN_NETWORK_QOS_ENABLE=${OVN_NETWORK_QOS_ENABLE:-false}
+  OVN_ENABLE_ICMP_NETPOL=${OVN_ENABLE_ICMP_NETPOL:-false}
 
 
   if [ "$OVN_COMPACT_MODE" == true ] && [ "$OVN_ENABLE_INTERCONNECT" != false ]; then
@@ -911,7 +916,8 @@ create_ovn_kube_manifests() {
     --mtu="${OVN_MTU}" \
     --enable-dnsnameresolver="${OVN_ENABLE_DNSNAMERESOLVER}" \
     --mtu="${OVN_MTU}" \
-    --enable-observ="${OVN_OBSERV_ENABLE}"
+    --enable-observ="${OVN_OBSERV_ENABLE}" \
+    --allow-icmp-netpol="${OVN_ENABLE_ICMP_NETPOL}"
   popd
 }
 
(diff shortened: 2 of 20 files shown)


mrobson:
hope you're doing well @trozet - miss your expertise!

What are the changes of this still merging in via natural process? Otherwise I will pivot :).

Thanks!

trozet:
sorry @mrobson I missed your message. @tssurya informed me this is needed. Will revive it now

trozet:
Will need to rebase once the CI fix goes in, and https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5880 goes in to fix kind configuration. Will address rabbit feedback as well

tssurya:
shall we add some E2Es as well?

go-controller/pkg/config/config.go
```diff
@@ -475,6 +475,7 @@ type OVNKubernetesFeatureConfig struct {
 	EnableServiceTemplateSupport    bool `gcfg:"enable-svc-template-support"`
 	EnableObservability             bool `gcfg:"enable-observability"`
 	EnableNetworkQoS                bool `gcfg:"enable-network-qos"`
+	AllowICMPNetworkPolicy          bool `gcfg:"allow-icmp-network-policy"`
```
arkadeepsen, line 510: Most of the existing config knobs starts with `enable`. Should this one be renamed to `EnableICMPNetworkPolicy` (`enable-icmp-network-policy`)?

go-controller/pkg/config/config.go
```diff
@@ -1194,6 +1195,12 @@ var OVNK8sFeatureFlags = []cli.Flag{
 		Destination: &cliConfig.OVNKubernetesFeature.EnableStatelessNetPol,
 		Value:       OVNKubernetesFeature.EnableStatelessNetPol,
 	},
+	&cli.BoolFlag{
+		Name:        "allow-icmp-network-policy",
+		Usage:       "Configure to use enable ICMP and ICMPv6 traffic to not be denied by network policy.",
+		Destination: &cliConfig.OVNKubernetesFeature.AllowICMPNetworkPolicy,
+		Value:       OVNKubernetesFeature.AllowICMPNetworkPolicy,
+	},
```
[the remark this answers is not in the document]
arkadeepsen: I also think the description can be improved.

contrib/kind-helm.sh
```diff
@@ -190,6 +191,8 @@ parse_args() {
                                                   OVN_ENABLE_INTERCONNECT=false
                                                   CENTRAL_ARG_PROVIDED=true
                                                   ;;
+            --allow-icmp-netpol )                 OVN_ENABLE_ICMP_NETPOL=true
```
arkadeepsen:
```suggestion
            --enable-icmp-netpol )                 OVN_ENABLE_ICMP_NETPOL=true
```
nit: Same suggestion here as well.

dist/images/daemonset.sh
```diff
@@ -615,6 +620,9 @@ echo "ovn_network_qos_enable: ${ovn_network_qos_enable}"
 ovn_enable_dnsnameresolver=${OVN_ENABLE_DNSNAMERESOLVER}
 echo "ovn_enable_dnsnameresolver: ${ovn_enable_dnsnameresolver}"
 
+ovn_allow_icmp_netpol=${OVN_ENABLE_ICMP_NETPOL}
+echo "ovn_allow_icmp_netpol: ${ovn_allow_icmp_netpol}"
```
arkadeepsen:
```suggestion
ovn_enable_icmp_netpol=${OVN_ENABLE_ICMP_NETPOL}
echo "ovn_enable_icmp_netpol: ${ovn_enable_icmp_netpol}"
```
nit

dist/images/ovnkube.sh
```diff
@@ -326,6 +327,8 @@ ovn_enable_svc_template_support=${OVN_ENABLE_SVC_TEMPLATE_SUPPORT:-true}
 ovn_network_qos_enable=${OVN_NETWORK_QOS_ENABLE:-false}
 # OVN_ENABLE_DNSNAMERESOLVER - enable dns name resolver support
 ovn_enable_dnsnameresolver=${OVN_ENABLE_DNSNAMERESOLVER:-false}
+# OVN_ENABLE_ICMP_NETPOL - allow ICMP/ICMPv6 with network policy
+ovn_allow_icmp_netpol=${OVN_ENABLE_ICMP_NETPOL:-false}
```
arkadeepsen:
```suggestion
ovn_enable_icmp_netpol=${OVN_ENABLE_ICMP_NETPOL:-false}
```
nit

dist/images/ovnkube.sh
```diff
@@ -1426,6 +1429,12 @@ ovn-master() {
   fi
   echo "ovn_enable_dnsnameresolver_flag=${ovn_enable_dnsnameresolver_flag}"
 
+  ovn_allow_icmp_netpol_flag=
+  if [[ ${ovn_allow_icmp_netpol} == "true" ]]; then
+	  ovn_allow_icmp_netpol_flag="--allow-icmp-network-policy"
+  fi
+  echo "ovn_allow_icmp_netpol_flag=${ovn_allow_icmp_netpol_flag}"
```
arkadeepsen, line 1511:
```suggestion
  ovn_enable_icmp_netpol_flag=
  if [[ ${ovn_enable_icmp_netpol} == "true" ]]; then
	  ovn_enable_icmp_netpol_flag="--enable-icmp-network-policy"
  fi
  echo "ovn_enable_icmp_netpol_flag=${ovn_enable_icmp_netpol_flag}"
```
nit

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -383,16 +386,22 @@ func (bnc *BaseNetworkController) defaultDenyPortGroupName(namespace string, acl
 }
 
 func (bnc *BaseNetworkController) buildDenyACLs(namespace, pgName string, aclLogging *libovsdbutil.ACLLoggingLevels,
-	aclDir libovsdbutil.ACLDirection) (denyACL, allowACL *nbdb.ACL) {
+	aclDir libovsdbutil.ACLDirection) []*nbdb.ACL {
 	denyMatch := libovsdbutil.GetACLMatch(pgName, "", aclDir)
-	allowMatch := libovsdbutil.GetACLMatch(pgName, arpAllowPolicyMatch, aclDir)
+	allowARPMatch := libovsdbutil.GetACLMatch(pgName, arpAllowPolicyMatch, aclDir)
+	allowICMPMatch := libovsdbutil.GetACLMatch(pgName, icmpAllowPolicyMatch, aclDir)
```
arkadeepsen: The `allowICMPMatch` variable can be moved inside the if condition which check whether the enable ICMP flag is set to true or not.

go-controller/pkg/ovn/multipolicy_test.go
```diff
@@ -540,6 +540,93 @@ var _ = ginkgo.Describe("OVN MultiNetworkPolicy Operations", func() {
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 		})
 
+		ginkgo.It("correctly creates and deletes network policy and multi network policy with the same policy - allow ICMP", func() {
```
arkadeepsen: DescribeTable would be better here since this test is almost exactly the same with the previous one apart from the `config.OVNKubernetesFeature.AllowICMPNetworkPolicy = true` line.

go-controller/pkg/ovn/policy_stale_test.go
```diff
@@ -277,6 +291,34 @@ var _ = ginkgo.Describe("OVN Stale NetworkPolicy Operations", func() {
 			gomega.Eventually(fakeOvn.nbClient).Should(libovsdbtest.HaveData(expectedData...))
 		})
 
+		ginkgo.It("reconciles an existing networkPolicy updating stale ACLs - Allow ICMP", func() {
```
arkadeepsen: DescribeTable can be used here as well.

go-controller/pkg/ovn/policy_test.go
```diff
@@ -804,6 +851,35 @@ var _ = ginkgo.Describe("OVN NetworkPolicy Operations", func() {
 			gomega.Expect(app.Run([]string{app.Name})).To(gomega.Succeed())
 		})
 
+		ginkgo.It("reconciles an existing networkPolicy with empty db - Allow ICMP", func() {
```
arkadeepsen: DescribeTable can be used here as well.

tssurya:
@trozet this is important for us timeline wise, will you have time to update the PR?

go-controller/pkg/config/config.go
trozet: I'm not really a fan of enable here, because it doesn't really distinguish between the action of allowing or denying. Enable is kind of ambiguous. What does it mean for icmp in network policy? is it denied? is it allowed?

I'm all for another name that makes more sense.

arkadeepsen: Makes sense. I agree with the thought process. We can go with "allow".

In the scripts some of the variables are using "enable" and some are using "allow" for this feature. I think we can use "allow" everywhere consistently.

trozet:
next push will be rebase

contrib/kind.sh
```diff
@@ -125,7 +125,11 @@ echo "-dug | --dynamic-udn-removal-grace-period <seconds>     Configure the grac
 echo "-adv | --advertise-default-network            Applies a RouteAdvertisements configuration to advertise the default network on all nodes"
 echo "-rud | --routed-udn-isolation-disable         Disable isolation across BGP-advertised UDNs (sets advertised-udn-isolation-mode=loose). DEFAULT: strict."
 echo "-mps | --multi-pod-subnet                     Use multiple subnets for the default cluster network"
+<<<<<<< HEAD
 echo "-noe | --no-overlay-enable                    Enable no overlay"
+=======
+echo "--allow-icmp-netpol                           Allows ICMP and ICMPv6 traffic globally, regardless of network policy rules"
+>>>>>>> dc29a3087f (Adds kind.sh and helm support for --allow-icmp-network-policy)
```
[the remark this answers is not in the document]
arkadeepsen: The rebase wasn't clean here.

dist/images/ovnkube.sh
```diff
@@ -2432,7 +2455,7 @@ ovnkube-controller-with-node() {
     ${network_qos_enabled_flag} \
     ${ovn_enable_dnsnameresolver_flag} \
     ${ovn_disable_requestedchassis_flag} \
-    ${cluster_access_opts} \
```
arkadeepsen, line 2435: Not sure if this should be removed here.

test/e2e/network_policy_icmp.go
```diff
@@ -0,0 +1,58 @@
...
+		ginkgo.By("creating server and client pods")
+		serverPodName := "icmp-server"
+		clientPodName := "icmp-client"
+		serverCmd := []string{"/bin/bash", "-c", "/agnhost netexec --http-port 8000"}
+		clientCmd := []string{"/agnhost", "pause"}
+
+		serverPod, err := createGenericPod(f, serverPodName, "", namespace, serverCmd)
+		framework.ExpectNoError(err, "failed to create server pod")
+		_, err = createGenericPod(f, clientPodName, "", namespace, clientCmd)
+		framework.ExpectNoError(err, "failed to create client pod")
```
arkadeepsen: I think the node selectors should also be added in this e2e test as well so that the pods are scheduled on different nodes.

test/e2e/network_policy_icmp.go
```diff
@@ -0,0 +1,58 @@
...
+
+		clientConfig := podConfiguration{name: clientPodName, namespace: namespace}
+		serverConfig := podConfiguration{name: serverPodName, namespace: namespace}
+
+		ginkgo.By("verifying TCP is denied by the default deny policy")
+		gomega.Eventually(func() error {
+			return pokePod(f, clientPodName, serverPod.Status.PodIP)
+		}, 1*time.Minute, 6*time.Second).ShouldNot(gomega.Succeed())
```
arkadeepsen, line 56: Shouldn't this check pass consistently rather than eventually?

test/e2e/network_segmentation_policy.go
```diff
@@ -207,6 +207,120 @@ var _ = ginkgo.Describe("Network Segmentation: Network Policies", feature.Networ
...
+					gomega.Eventually(func() error {
+						return pingServerPodFromClient(cs, serverPodConfig, clientPodConfig, serverIP)
+					}, 1*time.Minute, 6*time.Second).Should(gomega.Succeed())
+				}
+
+				ginkgo.By("asserting the *client* pod can not reach the server pod HTTP endpoint due to default deny policy")
+				for _, serverIP := range serverIPs {
+					gomega.Eventually(func() error {
+						return reachServerPodFromClient(cs, serverPodConfig, clientPodConfig, serverIP, port)
+					}, 1*time.Minute, 6*time.Second).ShouldNot(gomega.Succeed())
```
arkadeepsen, line 277: Here also shouldn't we use Consistenly instead of Eventually?

arkadeepsen (commented):
> If the config knob is toggled on and off, will the stale ICMP allow rules be deleted?

Should there be an unit test where the initial DB starts with stale ICMP allow ACLs and then when the controller starts with allow-icmp-netpol set to false, these ACLs get removed? This will verify the above scenario of the config knob being toggled on and off works correctly.

test/e2e/network_policy_icmp.go
trozet: we need an eventually first, cause its not guaranteed that network policy is in place as soon as the pod starts (although it is usually true). But I can add a consistently after to make sure it continues to be blocked.

arkadeepsen:
> > If the config knob is toggled on and off, will the stale ICMP allow rules be deleted?
> 
> Should there be an unit test where the initial DB starts with stale ICMP allow ACLs and then when the controller starts with allow-icmp-netpol set to false, these ACLs get removed? This will verify the above scenario of the config knob being toggled on and off works correctly.

I am mostly good with the PR except for the cleaning up of stale ICMP ACLs. Do you think we should we should have a test for this?

PR #5254, opened 2025-05-15
contrib: Enable working with local-registry when using podman

ormergi:
## 📑 Description

When podman is installed, it impossible to work with a local registry because the automation fail to push images to the local registry.

On podman push, it defaults to secure connection.
In our case the local registry uses an insecure connection result in podman push failures making it impossible to work with the local registry when podman is installed.

Set podman to skip secure connection check when pushing OVN-K images to the local registry.

The automation inspect the built ovnkube-image digest (SHA) 
and pass it to the daemonset manifest, in order to ensure the latest built image is deployed.

Some container runtime (e.g. podman) may not retain the same digest, result in having
one image digest in the local runtime image and different one on the local registry.
Which makes OVN-K to not become ready.

In order to overcome this, get the actual image digest that exist in the local registry.

Fixes #

This change introduce new dependency for the project - [skopeo](https://github.com/containers/skopeo)

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -822,6 +822,11 @@ set_ovn_image() {
 }
 
 build_ovn_image() {
+  local push_args
+  if [ "$OCI_BIN" == "podman" ]; then
+    push_args="--tls-verify=false"
+  fi
+
   if [ "$OVN_IMAGE" == local ]; then
     set_ovn_image
 
@@ -834,14 +839,14 @@ build_ovn_image() {
     # store in local registry
     if [ "$KIND_LOCAL_REGISTRY" == true ];then
       echo "Pushing built image to local $OCI_BIN registry"
-      $OCI_BIN push "${OVN_IMAGE}"
+      $OCI_BIN push "${push_args}" "${OVN_IMAGE}"
     fi
   # We should push to local registry if image is not remote
   elif [ "${OVN_IMAGE}" != "" -a "${KIND_LOCAL_REGISTRY}" == true ] && (echo "$OVN_IMAGE" | grep / -vq); then
     local local_registry_ovn_image="localhost:5000/${OVN_IMAGE}"
     $OCI_BIN tag "$OVN_IMAGE" $local_registry_ovn_image
     OVN_IMAGE=$local_registry_ovn_image
-    $OCI_BIN push $OVN_IMAGE
+    $OCI_BIN push "${push_args}" "$OVN_IMAGE"
   fi
 }
 
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -504,6 +504,11 @@ check_dependencies() {
   	  echo "Dependency not met: Neither docker nor podman found"
   	  exit 1
   fi
+
+  if ! command_exists skopeo; then
+      echo "Dependency not met: skopeo not installed. Run the following command to install it: 'sudo dnf install skopeo'"
+      exit 1
+  fi
 }
 
 OPENSSL=""
@@ -853,8 +858,9 @@ build_ovn_image() {
 create_ovn_kube_manifests() {
     local ovnkube_image=${OVN_IMAGE}
     if [ "$KIND_LOCAL_REGISTRY" == true ];then
-      # When updating with local registry we have to reference the sha
-      ovnkube_image=$($OCI_BIN inspect --format='{{index .RepoDigests 0}}' $OVN_IMAGE)
+      # When updating with local registry we have to reference the image digest (SHA)
+      # Check the image digest in the local registry because it might be different then the digest in the local container runtime
+      ovnkube_image=$(skopeo inspect --format "{{.Name}}@{{.Digest}}" --tls-verify=false  "docker://$OVN_IMAGE")
     fi
     pushd ${DIR}/../dist/images
     if [ "$OVN_ENABLE_INTERCONNECT" == true ]; then


ormergi:
@maiqueb @jcaamano could you please have a look?

ormergi:
@tssurya @jcaamano could you please have a look?

maiqueb:
@tssurya / @jcaamano the podman users workflow is broken if we use the local registry.

Could we get this merged ? I confirm this PR makes it work.

contrib/kind.sh
```diff
@@ -504,6 +504,11 @@ check_dependencies() {
   	  echo "Dependency not met: Neither docker nor podman found"
   	  exit 1
   fi
+
+  if ! command_exists skopeo; then
```
tssurya: is this something that only podman needs?

contrib/kind.sh
```diff
@@ -822,6 +827,11 @@ set_ovn_image() {
 }
 
 build_ovn_image() {
+  local push_args=""
+  if [ "$OCI_BIN" == "podman" ]; then
+    push_args="--tls-verify=false"
```
tssurya, line 833: maybe we should add a comment to "TODO: investinate if the localregistry must be using a secure connection instead?" Does this mean docker is ok with insecure conn as the default?

tssurya (commented):
im not a podman expert, but should we change one of the CI lanes to use podman instead of docker to ensure kind with podman continues to work with future changes and we don't break something else?

maiqueb:
> im not a podman expert, but should we change one of the CI lanes to use podman instead of docker to ensure kind with podman continues to work with future changes and we don't break something else?

If we're doing that, let's please do it in a follow-up. 
Having to ensure all branches I use have this PR in is getting tiring ...

**EDIT:** and FWIW, I wouldn't bother. Let podman users complain, and provide fixes. 
... but surely in an ideal world automation would let us know when something's broken.

tssurya:
> > im not a podman expert, but should we change one of the CI lanes to use podman instead of docker to ensure kind with podman continues to work with future changes and we don't break something else?
> 
> If we're doing that, let's please do it in a follow-up. Having to ensure all branches I use have this PR in is getting tiring ...
> 
> **EDIT:** and FWIW, I wouldn't bother. Let podman users complain, and provide fixes. ... but surely in an ideal world automation would let us know when something's broken.

As a maintainer I have to ask for tests/CI always on all PRs :) just doing my bit 
I'll wait for @ormergi's thoughts/opinions as well

contrib/kind.sh
ormergi: Yes, its required for podman needs, when working with local registry.
Due to the difference how Docker and Podman persist images.

contrib/kind.sh
ormergi: It just that Docker doesnt perform TLS check by default, but podman does.
The local registry endpoint is insecure (IMO its fine for development environment), thus when podman is being used we need to specify the above flag.
WDYT?

ormergi:
> im not a podman expert, but should we change one of the CI lanes to use podman instead of docker to ensure kind with podman continues to work with future changes and we don't break something else?

Switching to podman in Github Action is not straight-forward and require some additional changes, trial and error, etc.. 
I posted this PR to unblock devs who use podman in their development environment, hopefully changing transitioning podman on CI can be done later on :pray:

maiqueb:
> > im not a podman expert, but should we change one of the CI lanes to use podman instead of docker to ensure kind with podman continues to work with future changes and we don't break something else?
> 
> Switching to podman in Github Action is not straight-forward and require some additional changes, trial and error, etc.. I posted this PR to unblock devs who use podman in their development environment, hopefully changing transitioning podman on CI can be done later on 🙏

For what it's worth, I'm another upstream user whose workflow is severely impaired by this.

contrib/kind.sh
tssurya: what I meant was should this check be done only if its on podman @ormergi ? example I use docker.. so i don't think my script should stop if i don't have skopeo right?
right now it looks like we have converged to using skipeo for both modes?

contrib/kind.sh
tssurya: ack, whatever you said here in the comment: "docker doesn't perform tls check by default only podman does" -> hence we need to disable it for podman -> if you can add that as a comment on the script it would be good info.

contrib/kind.sh
```diff
@@ -834,22 +844,23 @@ build_ovn_image() {
...
 }
 
 create_ovn_kube_manifests() {
     local ovnkube_image=${OVN_IMAGE}
     if [ "$KIND_LOCAL_REGISTRY" == true ];then
-      # When updating with local registry we have to reference the sha
-      ovnkube_image=$($OCI_BIN inspect --format='{{index .RepoDigests 0}}' $OVN_IMAGE)
+      # When updating with local registry we have to reference the image digest (SHA)
+      # Check the image digest in the local registry because it might be different then the digest in the local container runtime
+      ovnkube_image=$(skopeo inspect --format "{{.Name}}@{{.Digest}}" --tls-verify=false  "docker://$OVN_IMAGE")
```
tssurya: so instead of docker inspect we will need to do skopeo inspect? can we change this only for podman?

contrib/kind.sh
ormergi: Done

contrib/kind.sh
ormergi: Right, skopeo is necessary when using podman only.
I removed this check and changed it to use skopeo only when podman is used.

contrib/kind.sh
ormergi: Done

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/f14644f57063c01b142849d7716bbf9d3211464e..ba0ee5002e24f3e1a0646961b144da1f774f6659): resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5254#pullrequestreview-2958685053 https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5254#pullrequestreview-2945776920
- skopeo check is removed
- skopeo is used only when podman is used 
@tssurya PTAL

tssurya:
> [Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/f14644f57063c01b142849d7716bbf9d3211464e..ba0ee5002e24f3e1a0646961b144da1f774f6659): resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5254#pullrequestreview-2958685053 https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5254#pullrequestreview-2945776920
> - skopeo check is removed
> - skopeo is used only when podman is used 
> @tssurya PTAL

Thanks Or! But why did you remove the check for whether skopeo is installed OR not ? Shouldn't we put that back in check dependency and if its Podman then ensure skopeo exits ? 
My suggestion was to make the check conditional for podman but looks like now it's gone

ormergi:
> Thanks Or! But why did you remove the check for whether skopeo is installed OR not ? Shouldn't we put that back in check dependency and if its Podman then ensure skopeo exits ? My suggestion was to make the check conditional for podman but looks like now it's gone

I got it all wrong, sorry for this back and forward.
I reverted the check, this time its conditional for podman.
@tssurya PTAL

contrib/kind.sh
```diff
@@ -504,6 +504,11 @@ check_dependencies() {
   	  echo "Dependency not met: Neither docker nor podman found"
   	  exit 1
   fi
+
+  if command_exists podman && ! command_exists skopeo; then
```
tssurya, line 508: awesome looks good now

contrib/kind.sh
tssurya: thank you

contrib/kind.sh
```diff
@@ -834,22 +845,28 @@ build_ovn_image() {
...
-      $OCI_BIN push "${OVN_IMAGE}"
+      $OCI_BIN push "${push_args}" "${OVN_IMAGE}"
     fi
   # We should push to local registry if image is not remote
   elif [ "${OVN_IMAGE}" != "" -a "${KIND_LOCAL_REGISTRY}" == true ] && (echo "$OVN_IMAGE" | grep / -vq); then
     local local_registry_ovn_image="localhost:5000/${OVN_IMAGE}"
     $OCI_BIN tag "$OVN_IMAGE" $local_registry_ovn_image
     OVN_IMAGE=$local_registry_ovn_image
-    $OCI_BIN push $OVN_IMAGE
+    $OCI_BIN push "${push_args}" "$OVN_IMAGE"
```
tssurya: @ormergi one last ask I am not bash expert so feel free to not do this shouldn't it be "${OVN_IMAGE}" or $OVN_IMAGE here? the braces don't matter?

contrib/kind.sh
```diff
@@ -834,22 +845,28 @@ build_ovn_image() {
...
 }
 
 create_ovn_kube_manifests() {
     local ovnkube_image=${OVN_IMAGE}
     if [ "$KIND_LOCAL_REGISTRY" == true ];then
-      # When updating with local registry we have to reference the sha
-      ovnkube_image=$($OCI_BIN inspect --format='{{index .RepoDigests 0}}' $OVN_IMAGE)
+      # When updating with local registry we have to reference the image digest (SHA)
+      # Check the image digest in the local registry because it might be different then the digest in the local container runtime
+      if [ "$OCI_BIN" == "podman" ]; then
```
tssurya, line 864: thanks!

tssurya (approved):
@ormergi changes look correct to me now and lgtm. thanks for sticking with me
nit: consider squashing the commits since the push-args added in first commit get used actually only in 2nd commit properly

one last question here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5254/files#r2176605886

contrib/kind.sh
ormergi: Braces usually doesnt matter unless some string manipulation or slicing is necessary.
The convention in the file is in consistent and there is no bash linter to enforce it, I am not sure whats better :sweat_smile:

ormergi: I removed the braces in the changes of this PR, keep this PR changes consistent as least.

tssurya: yeah this file is not having bash linter :( but yea I don't think in the end like you said it matters with string manipulation for bash for braces and no braces are valid looks like
@coderabbitai what do you think?

tssurya (approved):

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/15992999092/job/45112247004?pr=5254
unrelated flake that needs investigation: 
```
Summarizing 1 Failure:
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_localnet.go:87

Ran 93 of 504 Specs in 4289.944 seconds
FAIL! -- 92 Passed | 1 Failed | 2 Flaked | 2 Pending | 409 Skipped
--- FAIL: TestE2E (4290.03s)
```

ormergi:
@tssurya thanks for merging this, Kubevirt org project would benefit from it as well.

PR #5258, opened 2025-05-19
Adds support for selecting PUDN with BGP RA

trozet:


trozet:
going to hold on this until we have a more concrete use case from users

PR #5207, opened 2025-04-25
Don't addup failedAttempts in pod-add-retry when it's namespace is already terminating

tssurya:
Prior to this change, we were:

```
if entry.failedAttempts >= MaxFailedAttempts {
			klog.Warningf("Dropping retry entry for %s %s: exceeded number of failed attempts",
				r.ResourceHandler.ObjType, objKey)
			r.DeleteRetryObj(key)
			metrics.MetricResourceRetryFailuresCount.Inc()
			if entry.newObj != nil {
				r.ResourceHandler.RecordErrorEvent(entry.newObj, "RetryFailed",
					fmt.Errorf("failed to reconcile and retried %d times for object: %v", MaxFailedAttempts, entry.newObj))
			}
			return
		}
```
incrementing the resourceRetry metric whenever a pod add was done
on a terminating namespace.

```
This leads to:
W0425 05:11:51.799928    3490 obj_retry.go:265] Dropping retry entry for *v1.Pod e2e-test-resolve-local-names-jnkkz/resolve-t42vd: exceeded number of failed attempts
W0425 05:11:51.799923    3490 obj_retry.go:265] Dropping retry entry for *v1.Pod e2e-test-resolve-local-names-jnkkz/resolve-template-xgc8c: exceeded number of failed attempts
W0425 05:11:51.799936    3490 obj_retry.go:265] Dropping retry entry for *v1.Pod e2e-test-resolve-local-names-jnkkz/resolve-template-nx8rc: exceeded number of failed attempts
```

and alerts since the pod will never get scheduled anyways

```
2025-04-25T05:02:45.312400267Z E0425 05:02:45.311904       1 framework.go:1316] "Plugin Failed" err="pods \"resolve-template-nx8rc\" is forbidden: unable to create new content in namespace e2e-test-resolve-local-names-jnkkz because it is being terminated" plugin="DefaultBinder" pod="e2e-test-resolve-local-names-jnkkz/resolve-template-nx8rc" node="ip-10-0-5-38.us-west-1.compute.internal"
2025-04-25T05:02:45.312400267Z I0425 05:02:45.311931       1 schedule_one.go:1005] "Failed to bind pod" pod="e2e-test-resolve-local-names-jnkkz/resolve-template-nx8rc"
```

following the crumb of one such pod event:
```
I0425 05:09:51.535888    3898 obj_retry.go:302] Retry object setup: *v1.Pod e2e-test-resolve-local-names-jnkkz/resolve-t42vd
[...]

PR #5260, opened 2025-05-20
Do not retry pods that are not scheduled

trozet:
In our pod handlers we have code that checks if a pod is scheduled. If the pod is not scheduled, then we do not add the pod to our retry framework. However, there are times where we automatically enqueue all pods on a node or across the cluster. This can be from when a new node comes up (pre-IC) and need to add all pods on that node. Another use case is when a new UDN/NAD is created and the controller spins up...we add all pods then.

We shouldn't be queuing pods to the retry framework that are not scheduled. It's a waste of operations. However, even if we do enqueue them, we definitely should not be treating a non-scheduled resource as an error and retrying it again later.

This commit changes the retry framework to detect the above case, and log an error. It does not trigger retrying of the resource, which may perpetually fail and then then cause OVNKubernetesResourceRetryFailure.

### DIFF
--- a/go-controller/pkg/retry/obj_retry.go
+++ b/go-controller/pkg/retry/obj_retry.go
@@ -320,7 +320,11 @@ func (r *RetryFramework) resourceRetry(objKey string, now time.Time) {
 		}
 		if r.ResourceHandler.NeedsUpdateDuringRetry && entry.config != nil && entry.newObj != nil {
 			klog.Infof("%v retry: updating object %s", r.ResourceHandler.ObjType, objKey)
-			if err := r.ResourceHandler.UpdateResource(entry.config, entry.newObj, true); err != nil {
+			if !r.ResourceHandler.IsResourceScheduled(entry.newObj) {
+				// unscheduled resources (pods) will be retried again later we do not track these as failures, and should not retry.
+				// we should avoid queuing objects to the retry handler that are not scheduled. Thus treat this as an error.
+				klog.Errorf("%v retry: cannot update object that is not scheduled: %s", r.ResourceHandler.ObjType, objKey)
+			} else if err := r.ResourceHandler.UpdateResource(entry.config, entry.newObj, true); err != nil {
 				entry.timeStamp = time.Now()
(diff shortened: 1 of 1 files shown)


go-controller/pkg/retry/obj_retry.go
```diff
@@ -320,7 +320,11 @@ func (r *RetryFramework) resourceRetry(objKey string, now time.Time) {
 		}
 		if r.ResourceHandler.NeedsUpdateDuringRetry && entry.config != nil && entry.newObj != nil {
 			klog.Infof("%v retry: updating object %s", r.ResourceHandler.ObjType, objKey)
-			if err := r.ResourceHandler.UpdateResource(entry.config, entry.newObj, true); err != nil {
+			if !r.ResourceHandler.IsResourceScheduled(entry.newObj) {
+				// unscheduled resources (pods) will be retried again later we do not track these as failures, and should not retry.
+				// we should avoid queuing objects to the retry handler that are not scheduled. Thus treat this as an error.
```
ricky-rav, line 325: So here and in the cases below we're relying on the API server to generate an update event when the pod is finally scheduled, right?

ricky-rav (approved):
/LGMT

go-controller/pkg/retry/obj_retry.go
```diff
@@ -320,7 +320,11 @@ func (r *RetryFramework) resourceRetry(objKey string, now time.Time) {
...
+				klog.Errorf("%v retry: cannot update object that is not scheduled: %s", r.ResourceHandler.ObjType, objKey)
```
martinkennelly, line 326: its not an error ? clog our logs.

trozet: it's an error to us because we should not have queued an object to the retry framework that is not scheduled. Someone forgot to check in their code before they added it to the retry queue.

tssurya:
I had a different angle fix here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5207 already for this issue which was tracked.
Looks like that just became moot.

PR #5261, opened 2025-05-20
Retry all pods smarter

trozet:
Related to investigating the root cause for: #5260. This commit removes adding pods that are not scheduled to the retry framework. When the pod is scheduled the controller will receive an event.

Additionally these functions that add pods were using the kubeclient instead of informer cache. That means everytime a UDN was added we would issue kubeclient command to get all pods, which is really bad for performance.

### DIFF
--- a/go-controller/hybrid-overlay/pkg/controller/ho_node_windows.go
+++ b/go-controller/hybrid-overlay/pkg/controller/ho_node_windows.go
@@ -57,7 +57,7 @@ func newNodeController(kube kube.Interface,
 			"UDP port. Please make sure you install all the KB updates on your system.")
 	}
 
-	node, err := kube.GetNode(nodeName)
+	node, err := kube.GetNodeForWindows(nodeName)
 	if err != nil {
 		return nil, err
 	}
@@ -345,7 +345,7 @@ func (n *NodeController) initSelf(node *corev1.Node, nodeSubnet *net.IPNet) erro
 	}
 
 	// Add existing nodes
-	nodes, err := n.kube.GetNodes()
+	nodes, err := n.kube.GetNodesForWindows()
 	if err != nil {
 		return fmt.Errorf("error in initializing/fetching nodes: %v", err)
 	}
@@ -370,7 +370,7 @@ func (n *NodeController) uninitSelf(node *corev1.Node) error {
 		networkName, n.networkID, node.Name)
 
 	// Remove existing nodes
-	nodes, err := n.kube.GetNodes()
+	nodes, err := n.kube.GetNodesForWindows()
 	if err != nil {
 		return fmt.Errorf("failed to get nodes: %v", err)
 	}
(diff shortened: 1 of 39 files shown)


trozet:
@kyrtapz PTAL

kyrtapz:
> Additionally these functions that add pods were using the kubeclient instead of informer cache. That means everytime a UDN was added we would issue kubeclient command to get all pods, which is really bad for performance.

oof, although I think https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5261/files#diff-9f0f2e18136afe48ce19d510c3b76a8b074775d3d6d7f4718239ab17ab7a702dR478 will only happen in a rare scenario where a NAD is added after the network was created the `addAllPodsOnNode` is not that lucky. 
I also think we should just get rid of https://github.com/ovn-kubernetes/ovn-kubernetes/blob/75fe04ca6b6df3dc5fcce8c5e81b125c3692ccb9/go-controller/pkg/kube/kube.go#L351-L362. 
Even with your PR it is still being used here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/gateway.go#L164
If there is any reason to keep it lets please, please rename it to something more scary to avoid falling in this trap again in the future. Maybe something like `GetPodsUncached`? 
Going even further I think this should also apply to all of the other getters we define on the kube interface.

trozet:
> > > Additionally these functions that add pods were using the kubeclient instead of informer cache. That means everytime a UDN was added we would issue kubeclient command to get all pods, which is really bad for performance.
> > 
> > 
> > oof! 😨 I think https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5261/files#diff-9f0f2e18136afe48ce19d510c3b76a8b074775d3d6d7f4718239ab17ab7a702dR478 will only happen in a rare scenario where a NAD is added after the network was created and the API will only return pending pods(still an unnecessary call). `addAllPodsOnNode` is not that lucky...
> > I also think we should just get rid of
> > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/75fe04ca6b6df3dc5fcce8c5e81b125c3692ccb9/go-controller/pkg/kube/kube.go#L351-L362
> > 
> > .
> > Even with your PR it is still being used here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/gateway.go#L164
> > If there is any reason to keep it lets please, please rename it to something more scary to avoid falling in this trap again in the future. Maybe something like `GetPodsUncached`?
> > Going even further I think this should also apply to all of the other getters we define on the kube interface.
> 
> good catch. I will remove it form here as well and remove the function

Actually the function is still used by ovn db manager. I'll leave it for now and we can remove it once we drop support for the legacy arch.

go-controller/pkg/ovn/controller/egressservice/.egressservice_zone_node.go.swp
kyrtapz, line 1: oops

kyrtapz (approved):
lgtm

PR #5140, opened 2025-03-21
UDN Isolation with BGP: Remove support for receiving advertised routes from remote nodes

tssurya:
CREDIT: To @trozet for suggesting the idea around SNATing all node related traffic to nodeIP for BGP and @jcaamano for coming up with design: https://issues.redhat.com/browse/OCPBUGS-56506?focusedId=27304963&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-27304963

## 📑 Description

Today when default network or UDN networks are
advertised using RAs the nodes also learn the
routes of other nodes' pod subnets in the cluster.

Example snippet of exposing a UDN network on
default vrf (non-vrflite usecase):

```
root@ovn-worker2:/# ip r show table 1048
default via 172.18.0.1 dev breth0 mtu 1400
10.96.0.0/16 via 169.254.0.4 dev breth0 mtu 1400
10.244.0.0/24 nhid 39 via 172.18.0.4 dev breth0 proto bgp metric 20
10.244.2.0/24 nhid 40 via 172.18.0.3 dev breth0 proto bgp metric 20
103.103.0.0/24 nhid 39 via 172.18.0.4 dev breth0 proto bgp metric 20
103.103.1.0/24 nhid 40 via 172.18.0.3 dev breth0 proto bgp metric 20
169.254.0.3 via 203.203.1.1 dev ovn-k8s-mp12
169.254.0.34 dev ovn-k8s-mp12 mtu 1400
172.26.0.0/16 nhid 41 via 172.18.0.5 dev breth0 proto bgp metric 20
203.203.0.0/24 nhid 39 via 172.18.0.4 dev breth0 proto bgp metric 20
203.203.0.0/16 via 203.203.1.1 dev ovn-k8s-mp12
203.203.1.0/24 dev ovn-k8s-mp12 proto kernel scope link src 203.203.1.2
local 203.203.1.2 dev ovn-k8s-mp12 proto kernel scope host src 203.203.1.2
broadcast 203.203.1.255 dev ovn-k8s-mp12 proto kernel scope link src 203.203.1.2
203.203.2.0/24 nhid 40 via 172.18.0.3 dev breth0 proto bgp metric 20

root@ovn-worker2:/# ip r show table 1046
default via 172.18.0.1 dev breth0 mtu 1400
10.96.0.0/16 via 169.254.0.4 dev breth0 mtu 1400
[...]

PR #5268, opened 2025-05-26
[DNM] Avoid importing BGP routes for UDNs

pperiyasamy:
This PR would skip importing BGP routes into the node for RA belonging to a UDN network, so it gets uniform route handling for both local and remote UDN network host subnet routes. The route handling for default pod network RAs is not changed and it would still import routes to keep traffic working for the default network.

### DIFF
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -684,21 +684,29 @@ func (c *Controller) generateFRRConfiguration(
 					Prefixes: advertisePrefixes,
 				},
 			}
-			neighbor.ToReceive = frrtypes.Receive{
-				Allowed: frrtypes.AllowedInPrefixes{
-					Mode: frrtypes.AllowRestricted,
-				},
-			}
-			if advertisements.Has(ratypes.PodNetwork) {
-				for _, prefix := range receivePrefixes {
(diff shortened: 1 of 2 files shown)


tssurya:
we'll continue with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140

pperiyasamy:
> Why is it required to keep importing the routes for the default network?

@trozet we needed to import CDN host subnet routes, otherwise CDN pod -> host's reply traffic takes asymmetric path via `ovn-k8s-mp0` and dropped, Thank you @tssurya for helping to troubleshoot. 
So we either need to keep importing CDN host subnet route or use SNAT (to node ip) for this traffic to work. 
This solution is still broken for pod to host traffic in a UDN network. @tssurya would address it with SNAT (to node ip) to handle this traffic for UDN network on top of this change. Does that make sense ?

> 
> Also on a different topic. I can see a future with no overlay mode, and no EVPN, where we will want to import these routes for native routing to work without any overlay. Just fyi.

OK, I understand. then let RA controller import the routes when UDN network's transport mode is set to `none`.

tssurya:
Yea I'll close this PR in favor of mine.. I had only reopened this to show everything passes in CI based on slack discussion. But we know the necessary pod to host tests don't exist for advertised UDNs which is why everything is passing here.

Issue #5239, opened 2025-05-08
ovn-kubernetes in host mode advanced uplink detection


aserdean:
### What would you like to be added?

in the case we have a host with two links that can serve as gateway, instead of choosing the default gateway interface by default, we want to prioritize the interface which is part of the DPU and can also serve as a gateway

### Why is this needed?

hardware enablement in order to ensure the acceleration of the high speed traffic for the N/S communication.

aserdean:
Adding more information about the issue.

Assume we have a worker running ovn-k in host mode, `br-dpu` is my default default GW but I want ovnkube to pick `ens5f0np0`  the PF of the card:
```
depuser@worker1:~$ lspci -vvv | grep -i nox
89:00.0 Ethernet controller: Mellanox Technologies MT43244 BlueField-3 integrated ConnectX-7 network controller (rev 01)
	Subsystem: Mellanox Technologies MT43244 BlueField-3 integrated ConnectX-7 network controller
```

I hopefully attached all the information needed, if not please feel free to ask for additional output

```
user@worker1:~$ ip -4 a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
6: ens5f0np0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc mq state UP group default qlen 1000
    altname enp137s0f0np0
    inet 10.0.120.2/29 metric 100 brd 10.0.120.7 scope global dynamic ens5f0np0
       valid_lft 2660sec preferred_lft 2660sec
    inet 169.254.0.2/17 brd 169.254.127.255 scope global ens5f0np0
       valid_lft forever preferred_lft forever
8: br-dpu: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9000 qdisc noqueue state UP group default qlen 1000
    inet 10.0.110.21/24 metric 100 brd 10.0.110.255 scope global dynamic br-dpu
       valid_lft 560sec preferred_lft 560sec
11: ovn-k8s-mp0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8940 qdisc mq state UP group default qlen 1000
    altname enp137s0f0v1
    inet 10.233.67.2/24 brd 10.233.67.255 scope global ovn-k8s-mp0
[...]

PR #5269, opened 2025-05-27
ovn-ic enablement on the DPU

aserdean:
This PR is a draft to discuss about ovn-ic feature on the DPU


go-controller/pkg/config/config.go
```diff
@@ -462,6 +462,8 @@ type GatewayConfig struct {
 	EgressGWInterface string `gcfg:"egw-interface"`
 	// NextHop is the gateway IP address of Interface; will be autodetected if not given
 	NextHop string `gcfg:"next-hop"`
+	// UplinkPort is the port used as the uplink on the gateway interface bridge.
```
trozet: this needs more context. For example we already have:
```
	// Interface is the network interface to use for the gateway in "shared" mode
	Interface string `gcfg:"interface"`
```

and then 

```
	// GatewayAcceleratedInterface is the optional network interface to use for gateway traffic acceleration.
	// This is typically a VF or SF device. When specified it would be used as the in_port for Openflow rules
	// on the external bridge. The Host IP would be on this device.
	// Should be used mutually exclusive to the `--gateway-interface` flag.
	GatewayAcceleratedInterface string `gcfg:"gateway-accelerated-interface"`
```
^ is this one DPU mode only?

now you are adding another one that at first glance appears to be the same thing as Interface.  Is this the outcome of https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5239 ?

If so why not just pass Interface?

go-controller/pkg/util/node_annotations.go
```diff
@@ -524,6 +529,36 @@ func SetNodePrimaryIfAddrs(nodeAnnotator kube.Annotator, ifAddrs []*net.IPNet) (
...
+		primaryIfAddrAnnotation.IPv4 = nodeIPNetv4.String()
+	}
+	if nodeIPNetv6 != nil {
+		primaryIfAddrAnnotation.IPv6 = nodeIPNetv6.String()
+	}
+	return nodeAnnotator.Set("k8s.ovn.org/node-primary-ifaddr-dpu", primaryIfAddrAnnotation)
+}
+
+func GetDpuNodeIfAddrAnnotation(node *corev1.Node) (*primaryIfAddrAnnotation, error) {
+	nodeIfAddrAnnotation, ok := node.Annotations["k8s.ovn.org/node-primary-ifaddr-dpu"]
```
trozet, line 551: make this annotation a constant

go-controller/pkg/node/openflow_manager.go
```diff
@@ -239,7 +239,30 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+	}
+	encapIP = strings.TrimSuffix(encapIP, "\n")
+
+	if config.OvnKubeNode.Mode == types.NodeModeDPU {
+		c.updateFlowCacheEntry("NORMAL", []string{
+			fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction),
+			fmt.Sprintf("table=0,priority=999,udp,tp_dst=67,actions=%s\n", util.NormalAction),
+			fmt.Sprintf("table=0,priority=999,udp,nw_dst=%s,tp_dst=6081,actions=LOCAL\n", encapIP),
+			fmt.Sprintf("table=0,priority=998,ip,nw_dst=%s,actions=%s\n", encapIP, util.NormalAction),
+		})
```
trozet: why?

go-controller/pkg/node/gateway_init.go
```diff
@@ -488,29 +502,43 @@ func interfaceForEXGW(intfName string) string {
...
 	if err != nil {
 		return err
 	}
 
-	ifAddrs, err := getNetworkInterfaceIPAddresses(gatewayIntf)
-	if err != nil {
-		return err
+	if err := util.SetDpuNodePrimaryIfAddrs(nodeAnnotator, ifAddrs); err != nil {
+		klog.Errorf("Unable to set primary IP net label on node, err: %v", err)
+	}
```
trozet: i'm not sure why this is being done to add a new annotation and not use the kube node IP of host cluster. I'm guessing you are trying to use a different interface than the kapi interface. We have some limited support already for doing that by using GatewayConfig->EgressGWInterface. I think we should get a better understand of the things happening in this commit in the DPU meeting.

go-controller/pkg/config/config.go
aserdean: That's a fair point, will check the code behind it maybe this part is no longer needed (the initial intention for the accelerated interface was to be also allocate the SF)

In my case I want the actual uplink to be an OVS patch port since the actual uplink will be behind a different bridge.

go-controller/pkg/node/openflow_manager.go
aserdean: The 67 port is to allow dhcp traffic to be passed to the PF (physical function).

6081 is to send the geneve traffic to the bridge (local port).

The last one which matches the encap ip is to be able to ping the local port from the PF.

If it is helpful I can add the bridge configuration and the IP management.

go-controller/pkg/node/gateway_init.go
aserdean: @trozet : you are on point here, I am trying to make OVNK aware about two networks the k8s API network (slow) and the data network, the one which we are trying to accelerate.

Both of them have external network and can egress traffic however only one can be accelerated.

go-controller/pkg/config/config.go
aserdean:
```
if config.Gateway.GatewayAcceleratedInterface != "" {
		// Try to get representor for the specified gateway device.
		// If function succeeds, then it is either a valid switchdev VF or SF, and we can use this accelerated device
		// for node IP, Host Ofport for Openflow etc.
		// If failed - error for improper configuration option
		intfRep, err = getRepresentor(config.Gateway.GatewayAcceleratedInterface)
		if err != nil {
			return nil, fmt.Errorf("gateway accelerated interface %s is not valid: %w", config.Gateway.GatewayAcceleratedInterface, err)
		}
		gwIntf = config.Gateway.GatewayAcceleratedInterface
		isGWAcclInterface = true
		klog.Infof("For gateway accelerated interface %s representor: %s", config.Gateway.GatewayAcceleratedInterface, intfRep)
	} else {
		intfRep, err = getRepresentor(gwIntf)
		if err == nil {
			isGWAcclInterface = true
		}
	}
```

unfortunately there are a lot of checks in order to figure out if the interface is backed by a representor and I cannot use a simple patchport.
lets discuss this further during the DPU sync.

go-controller/pkg/node/gateway_init.go
trozet: can you please create a diagram to show the host/dpu the vf rep/PF ports and how the OVS bridges are connected? I think an enhancement first would be really helpful before we jump into the code, but if you don't want to do that then we really need a visual and good description in the PR of what you are trying to achieve.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,29 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+		}
+		encapIP, _, err := util.RunOVSVsctl(checkCmd...)
+		if err != nil {
+			return err
+		}
+		encapIP = strings.TrimSuffix(encapIP, "\n")
+		c.updateFlowCacheEntry("NORMAL", []string{
+			fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction),
+			fmt.Sprintf("table=0,priority=999,udp,tp_dst=67,actions=%s\n", util.NormalAction),
+			fmt.Sprintf("table=0,priority=999,udp,nw_dst=%s,tp_dst=6081,actions=LOCAL\n", encapIP),
```
trozet, line 263: we already set geneve flows here:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_shared_intf.go#L1481C10-L1482C11

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,29 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+			"external_ids:ovn-encap-ip",
+		}
+		encapIP, _, err := util.RunOVSVsctl(checkCmd...)
+		if err != nil {
+			return err
+		}
+		encapIP = strings.TrimSuffix(encapIP, "\n")
+		c.updateFlowCacheEntry("NORMAL", []string{
+			fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction),
+			fmt.Sprintf("table=0,priority=999,udp,tp_dst=67,actions=%s\n", util.NormalAction),
```
trozet, line 262: can you elaborate on why you need to match dhcp here? why would it need to be normal action, and why would you only install flows for 67 and not 68?

npinaeva:
Do you think this PR should also include a docs update to make it easier to review?

go-controller/pkg/config/config.go
aserdean: dropped in favour of using the OVS metadata, example:

```
ovs-vsctl br-set-external-id br-ovn bridge-id br-ovn
ovs-vsctl br-set-external-id br-ovn bridge-uplink puplinkbrovn
```

if someone wants to pursue adding a direct argument from ovnkube he could use the code above

go-controller/pkg/node/openflow_manager.go
aserdean: Will remove it, I think I originally added it to use the IP instead of the MAC. 
I opened this specific part under the following PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5301

go-controller/pkg/node/openflow_manager.go
aserdean: Sure its a local DHCP server that is serving on the ovnk managed bridge (br-ex i.e.) ip address for the PF_rep/PF.

The reason for the priority is to make sure that the DHCP server receives the packet first.

We did not have issues on sending the request (68) but just receiving it (67).

aserdean:
> Do you think this PR should also include a docs update to make it easier to review?

At the moment I am trying to split PR into more digestible pieces with proper commit description, hopefully it will suffice.

So far I have:
1. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5298
2. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5301

The plan is to have docs after the changes go in due to time constraints.

PR #5274, opened 2025-05-29
remote node deletion failure due to libovsdb integrity violation error

cathy-zhou:
ovnkube-controller is trying to delete logical static route from the router it does not belong, which ends with the error:

"referential integrity violation: cannot delete
Logical_Router_Static_Route row ... because of 1 remaining references"

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
+++ b/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
@@ -515,7 +515,9 @@ func (zic *ZoneInterconnectHandler) cleanupNode(nodeName string) error {
 		return err
 	}
 
-	// Delete any static routes in the cluster router for this node
+	// Delete any static routes in the cluster router for this node.
+	// skip types.NetworkExternalID check in the predicate function as this static route may be deleted
+	// before types.NetworkExternalID external-ids is set correctly during upgrade.
 	p := func(lrsr *nbdb.LogicalRouterStaticRoute) bool {
 		return lrsr.ExternalIDs["ic-node"] == nodeName
 	}
@@ -573,11 +575,15 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
 	addRoute := func(prefix, nexthop string) error {
 		logicalRouterStaticRoute := nbdb.LogicalRouterStaticRoute{
 			ExternalIDs: map[string]string{
-				"ic-node": node.Name,
+				"ic-node":               node.Name,
+				types.NetworkExternalID: zic.GetNetworkName(),
 			},
 			Nexthop:  nexthop,
 			IPPrefix: prefix,
 		}
+		// Note that because logical router static routes were originally created without types.NetworkExternalID
+		// external-ids, skip types.NetworkExternalID check in the predicate function to replace existing static route
+		// with correct external-ids on an upgrade scenario.
 		p := func(lrsr *nbdb.LogicalRouterStaticRoute) bool {
 			return lrsr.IPPrefix == prefix &&
 				lrsr.Nexthop == nexthop &&
@@ -613,6 +619,8 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
 
 // deleteLocalNodeStaticRoutes deletes the static routes added by the function addRemoteNodeStaticRoutes
 func (zic *ZoneInterconnectHandler) deleteLocalNodeStaticRoutes(node *corev1.Node, nodeTransitSwitchPortIPs []*net.IPNet) error {
+	// skip types.NetworkExternalID check in the predicate function as this static route may be deleted
+	// before types.NetworkExternalID external-ids is set correctly during upgrade.
 	deleteRoute := func(prefix, nexthop string) error {
 		p := func(lrsr *nbdb.LogicalRouterStaticRoute) bool {
 			return lrsr.IPPrefix == prefix &&


cathy-zhou:
@npinaeva @jcaamano Please take a look

jcaamano:
shouldn't we have some unit test coverage here?

cathy-zhou:
> shouldn't we have some unit test coverage here?

@jcaamano done. please take a look.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -538,15 +538,17 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
 	addRoute := func(prefix, nexthop string) error {
 		logicalRouterStaticRoute := nbdb.LogicalRouterStaticRoute{
 			ExternalIDs: map[string]string{
-				"ic-node": node.Name,
+				"ic-node":               node.Name,
+				types.NetworkExternalID: zic.GetNetworkName(),
```
jcaamano, line 579: So if this external ID wasn't already being set, what happens on upgrade to this change?  Wouldn't this create a new LRSR and the old one be left over completely unmanaged?

cathy-zhou: fixed, please take a look.

jcaamano: This is a tricky fix. 

What ensures that `CreateOrReplaceLogicalRouterStaticRouteWithPredicate` below only finds 1 route instead of many. If it finds many, it will fail.

For example, can you add a test where you preload the data, with both LRSR for red and blue already present, as if ovnk were just being restarted?

cathy-zhou: @jcaamano the implementation of CreateOrReplaceLogicalRouterStaticRouteWithPredicate() is to find the logical static route **from the specified logical router only** with the predicate, so your concern won't be a problem.

jcaamano: Actually, you are going to keep overwriting the same route with a different network name. So you will still have the problem as we will continue to have a single LRSR for all networks:
* for some networks no LRSR will be deleted
* for one actual network you will get the same problem as before

So there are two options:
* Assume that the LRSR with no ID is for the default network, and create new LRSR for other networks. You can arrange the predicate to have this effect.
* Go into `DeleteLogicalRouterStaticRoutesWithPredicate` and remove the part of the op that actually deletes the LRSR. Removing the reference form the router should be enough, as garbage collection on the server side will remove the LRSR when it is not referenced.

I prefer the second option, as its is more efficient to have less repeated data on the server, particularly on an non-indexed table. However, you might have to clean up some tests that don't expect this to happen. Also would add a comment here if the route is expected to be shared by multiple networks.

jcaamano: I missed the fact that `CreateOrReplaceLogicalRouterStaticRouteWithPredicate` operates in the context of a single router.

We agreed offline that although the current solution works, it would add more value to also have `DeleteLogicalRouterStaticRoutesWithPredicate` operate only in the context of the router passed-

cathy-zhou: @jcaamano fixed based on our discussion. please take a look

go-controller/pkg/libovsdb/ops/router.go
```diff
@@ -775,32 +775,21 @@ func DeleteLogicalRouterStaticRoutesWithPredicate(nbClient libovsdbclient.Client
...
 
 // DeleteLogicalRouterStaticRoutesWithPredicateOps looks up logical router static
-// routes from the cache based on a given predicate, and returns the ops to delete
-// them and remove them from the provided logical router
+// routes from the logical router of the specified name based on a given predicate,
+// and returns the ops to delete them and remove them from the provided logical router
 func DeleteLogicalRouterStaticRoutesWithPredicateOps(nbClient libovsdbclient.Client, ops []ovsdb.Operation, routerName string, p logicalRouterStaticRoutePredicate) ([]ovsdb.Operation, error) {
-	router := &nbdb.LogicalRouter{
-		Name: routerName,
+	deleted, err := GetRouterLogicalRouterStaticRoutesWithPredicate(nbClient, &nbdb.LogicalRouter{Name: routerName}, p)
```
jcaamano: nit: I think it is better to find a different name that `deleted` here. It gives the impression the routes were already deleted and makes you look twice at the method that is being called here. I would just call it `routes` or `lrsrs`.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -538,11 +538,15 @@ func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node,
...
 			},
 			Nexthop:  nexthop,
 			IPPrefix: prefix,
 		}
+		// Note that because logical router static route was created without types.NetworkExternalID external-ids,
```
jcaamano: nit: 
```
// Note that because logical router static routes were originally created without types.NetworkExternalID external-ids,
// skip types.NetworkExternalID check in the predicate function to replace existing static route with correct
// external-ids on an upgrade scenario.
```

go-controller/pkg/libovsdb/ops/router.go
cathy-zhou: fixed

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
cathy-zhou: fixed

jcaamano (dismissed):

jcaamano:
> There is a small caveat. If there is an upgrade and just following it, a cleanup for a stale node, then the route wont have the external ID.

> I think that operating in the context should handle the original problem well enough so I would probably avoid matching against the external ID for some time on the delete operations and add a comment reasoning why.

Sorry this thought cross my mind now and not earlier.

cathy-zhou:
> I think that operating in the context should handle the original problem well enough so I would probably avoid matching against the external ID for some time on the delete operations and add a comment reasoning why.

fixed. please take a look. but I am wondering if it is better to update all existing lrsrs without network external-ids to add that exteranal-ids (upgrade logic) in syncNodes().

jcaamano:
> > I think that operating in the context should handle the original problem well enough so I would probably avoid matching against the external ID for some time on the delete operations and add a comment reasoning why.
> 
> fixed. please take a look. but I am wondering if it is better to update all existing lrsrs without network external-ids to add that exteranal-ids (upgrade logic) in syncNodes().

Not sure, syncNodes only handles stale nodes so I don0't its worth the complication.

However, wouldn't we have the same problem on the other delete operation as well? Don't we need to remove the external ID from that predicate too?

cathy-zhou:
> However, wouldn't we have the same problem on the other delete operation as well? Don't we need to remove the external ID from that predicate too?

done. thanks for pointing this out.

jcaamano:
This looks like a flake that might have been introduced recently, maybe in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5113
```
  [FAILED] in [It] - /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:499 @ 06/11/25 09:27:28.705
• [FAILED] [1.303 seconds]
EgressIP selectors [It] configures multiple IPv6 EIPs on different links, multiple namespaces and multiple pods
/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:998

  [FAILED] Timed out after 1.000s.
      <*errors.errorString | 0xc0757d0ee0>: 
      expected and found IPTable rule(s) count do not match: expected 4 IPtable rule(s) but got 2:
      expected IPtable rule(s) '[{0xc000550900 [{67 0 0 0xc00053a540 [] [] [] 0 1024 0 1067 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {67 0 0 0xc00053a570 [] [] [] 0 256 0 1067 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {67 0 0 0xc00053bbc0 [] [] [] 0 256 0 1067 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0}] 0xc00053cfc0 dummy1 [{testpod1 {[-s fd46::1/128 -o dummy1 -j SNAT --to-source 2001::1:0:8a2e:370:2]} 0xc0002fd680} {testpod2 {[-s fd46::2/128 -o dummy1 -j SNAT --to-source 2001::1:0:8a2e:370:2]} 0xc0002fd740}]} {0xc000550a80 [{48 0 0 0xc00053a540 [] [] [] 0 1024 0 1048 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {48 0 0 0xc00053a570 [] [] [] 0 256 0 1048 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {48 0 0 0xc00053bce0 [] [] [] 0 256 0 1048 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0}] 0xc00053d020 dummy2 [{testpod3 {[-s fd46::3/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]} 0xc0002fd800} {testpod4 {[-s fd46::4/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]} 0xc0002fd8c0}]}]'
      found IPTable rule(s) '[{[-s fd46::4/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]} {[-s fd46::3/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]}]'
      {
          s: "expected and found IPTable rule(s) count do not match: expected 4 IPtable rule(s) but got 2:\nexpected IPtable rule(s) '[{0xc000550900 [{67 0 0 0xc00053a540 [] [] [] 0 1024 0 1067 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {67 0 0 0xc00053a570 [] [] [] 0 256 0 1067 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {67 0 0 0xc00053bbc0 [] [] [] 0 256 0 1067 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0}] 0xc00053cfc0 dummy1 [{testpod1 {[-s fd46::1/128 -o dummy1 -j SNAT --to-source 2001::1:0:8a2e:370:2]} 0xc0002fd680} {testpod2 {[-s fd46::2/128 -o dummy1 -j SNAT --to-source 2001::1:0:8a2e:370:2]} 0xc0002fd740}]} {0xc000550a80 [{48 0 0 0xc00053a540 [] [] [] 0 1024 0 1048 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {48 0 0 0xc00053a570 [] [] [] 0 256 0 1048 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0} {48 0 0 0xc00053bce0 [] [] [] 0 256 0 1048 1 0 0 <nil> <nil> <nil> <nil> 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0  0}] 0xc00053d020 dummy2 [{testpod3 {[-s fd46::3/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]} 0xc0002fd800} {testpod4 {[-s fd46::4/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]} 0xc0002fd8c0}]}]'\nfound IPTable rule(s) '[{[-s fd46::4/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]} {[-s fd46::3/128 -o dummy2 -j SNAT --to-source 2001::2:0:8a2e:370:2]}]'",
      }
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:499 @ 06/11/25 09:27:28.705
```
cc @yboaron @trozet

jcaamano:
Now a different flake, maybe related to https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5286 ? I don't know what's going on @cathy-zhou but I had to rebase because I saw related changes on HEAD. I think we need to sort this out (don't think it's related to your changes but we can't merge in this state).

```
EgressIP selectors [It] configures one uncompressed IPv6 EIP and one Pod
/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:788

  [FAILED] Timed out after 1.001s.
      <*errors.errorString | 0xc05017c950>: 
      failed to find rule ip rule 6000: from fd46::1/128 to all table 1067 
      {
          s: "failed to find rule ip rule 6000: from fd46::1/128 to all table 1067 ",
      }
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:607 @ 06/11/25 10:11:47.545
```

PR #5275, opened 2025-05-29
Fix node update check for network cluster controller

trozet:
Introduced in 836ec3680aa33300d631b3e7f3bd3069bbfed2b9

This would just cause node updates to fire HandleAddUpdateNodeEvent everytime as the code prior to the aforementioned commit would have.

### DIFF
--- a/go-controller/pkg/clustermanager/node/node_allocator.go
+++ b/go-controller/pkg/clustermanager/node/node_allocator.go
@@ -195,27 +195,24 @@ func (na *NodeAllocator) NeedsNodeAllocation(node *corev1.Node) bool {
 	}
 
 	// ovn node check
-	// allocation is all or nothing, so if one field was allocated from:
-	// nodeSubnets, joinSubnet, layer 2 tunnel id, then all of them were
 	if na.hasNodeSubnetAllocation() {
-		if util.HasNodeHostSubnetAnnotation(node, na.netInfo.GetNetworkName()) {
-			return false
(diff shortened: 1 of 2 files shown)


trozet:
@kyrtapz PTAL

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -195,27 +195,24 @@ func (na *NodeAllocator) NeedsNodeAllocation(node *corev1.Node) bool {
 	}
 
 	// ovn node check
-	// allocation is all or nothing, so if one field was allocated from:
-	// nodeSubnets, joinSubnet, layer 2 tunnel id, then all of them were
 	if na.hasNodeSubnetAllocation() {
-		if util.HasNodeHostSubnetAnnotation(node, na.netInfo.GetNetworkName()) {
-			return false
+		if !util.HasNodeHostSubnetAnnotation(node, na.netInfo.GetNetworkName()) {
```
kyrtapz, line 199: just a note, although I agree with your reasoning for changing this logic it might have adverse impact on scale since we are now going to parse all annotations.

kyrtapz:
I haven't seen this flake before:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/15336692727/job/43156284284?pr=5275
might be related.

go-controller/pkg/clustermanager/node/node_allocator.go
trozet: ack. For now we will just play it safe to avoid regression. In the future we can mitigate our json unmarshal issues with a different CRD/solution.

PR #4533, opened 2024-07-16
Direct ingress for default network in SGW mode with route advertisements

jcaamano:
Network manager will, upon ensuring a network, fetch relevant
information from RouteAdvertisements (such as VRFs being advertised on
and selected nodes to advertise EIPs) applying to the network and set
it on the corresponding NetInfo used to create the network controller,
or triggering a reconciliation if it was already running.

This relies on annotations set by cluster manager on NADs pointing to
the route advertising configuration that applies to the network which
will come in a future PR/commit.

This will also be done for the default network. This PR assumes
cluster manager will create a dummy NAD for the default network in ovn-k
namespace were the annotaiton will be set. The NAD controller will
process the NAD and ensure the network on the network controller, just
like with any other network. The network controller gains access to the
default network controller, however, it treats that controller
differently: it's not started or stopped, obviously, just reconciled.

Add the ability for network controllers to reconcile some network
information changes. Currently reconciles:
- for zone network controllers to configure or not the pod IP to node IP
  SNAT on the GR for a nodes local to the zone
- for node network controllers to configure or not br-ex flows to
  redirect pod IP ingress traffic to the OVN network (except ingress to
  the management ip address). Only done for the default network, UDNs
  will be handled on a later commit.

Support for reconciling NAD changes is out of scope.

This should be enough to provide direct ingress capabilities for the
default network in SGW mode.

Note that non-primary network controllers don't reconcile anything as
route advertising is not supported on them. Also cluster manager network
controllers don't reconcile much as they don't have the need.

Currently based on top of https://github.com/ovn-org/ovn-kubernetes/pull/4472

PR #5276, opened 2025-05-30
Advertised UDN pod networks isolation configurable

pperiyasamy:
This PR provides a global routed udn isolation config knob (`routed-udn-isolation`) that use can configure to disable isolation between pod networks across BGP advertised networks. 
This enables flexibility for the provider networks to connect two different UDN networks with required routing configuration. 
By default `routed-udn-isolation` is `Enabled` and UDN networks are isolated each other.

## Summary by CodeRabbit

* **New Features**
  * Configurable Advertised UDN Isolation Mode ("strict" default, "loose" option) exposed via CLI, env, Helm and kind; runtime gating now controls advertised-network isolation vs connectivity.

* **Bug Fixes**
  * Fixed config tag for EphemeralPortRange.

* **Documentation**
  * Updated EgressIP docs to reflect new flow priorities and advertised UDN behavior.

* **Tests**
  * e2e: added bgp-loose-isolation matrix entry, ADVERTISED_UDN_ISOLATION_MODE wiring, consolidated FRR route helpers, and updated test names/expectations.

* **Chores**
  * CI/e2e: job/env wiring and timeouts extended; kind tooling and manifests updated to pass the new option; runtime cleanup behavior adjusted for loose mode.

### DIFF
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -767,10 +767,10 @@ deploy_bgp_external_server() {
     echo "FRR bgp network IPv6: ${bgp_network_frr_v6}"
     $OCI_BIN exec bgpserver ip -6 route replace default via "$bgp_network_frr_v6"
   fi
-  if [ "$ROUTED_UDN_ISOLATION" == "Disabled" ]; then
+  if [ "$ADVERTISED_UDN_ISOLATION_MODE" == "loose" ]; then
     kind_network_frr_v4=$($OCI_BIN inspect -f '{{.NetworkSettings.Networks.kind.IPAddress}}' frr)
     echo "FRR kind network IPv4: ${kind_network_frr_v4}"
-    # If routed UDN isolation is disabled, we need to set the default gateway for the nodes in the cluster
+    # If UDN isolation is in loose disabled, we need to set the default gateway for the nodes in the cluster
     # to the FRR router so that cross-UDN traffic can be routed back to the pods in the cluster in the loose mode.
     echo "Setting default gateway for nodes in the cluster to FRR router IPv4: ${kind_network_frr_v4}"
     set_nodes_default_gw "$kind_network_frr_v4"
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -53,6 +53,7 @@ usage() {
     echo "                 [-ic | --enable-interconnect]"
     echo "                 [-uae | --preconfigured-udn-addresses-enable]"
     echo "                 [-rae | --enable-route-advertisements]"
+    echo "                 [-rud | --routed-udn-isolation-disable]"
     echo "                 [-adv | --advertise-default-network]"
     echo "                 [-nqe | --network-qos-enable]"
     echo "                 [--isolated]"
@@ -127,6 +128,7 @@ echo "-obs | --observability                        Enable OVN Observability fea
 echo "-uae | --preconfigured-udn-addresses-enable   Enable connecting workloads with preconfigured network to user-defined networks"
 echo "-rae | --enable-route-advertisements          Enable route advertisements"
 echo "-adv | --advertise-default-network            Applies a RouteAdvertisements configuration to advertise the default network on all nodes"
+echo "-rud | --routed-udn-isolation-disable         Disable isolation across BGP-advertised UDNs (sets advertised-udn-isolation-mode=loose). DEFAULT: strict."
 echo ""
 }
 
@@ -316,7 +318,7 @@ parse_args() {
                                                   ;;
             -adv | --advertise-default-network) ADVERTISE_DEFAULT_NETWORK=true
                                                   ;;
-            -rnd | --routed-udn-isolation-disable)  ROUTED_UDN_ISOLATION=Disabled
+            -rud | --routed-udn-isolation-disable) ADVERTISED_UDN_ISOLATION_MODE=loose
                                                   ;;
             -ce | --enable-central )              OVN_ENABLE_INTERCONNECT=false
                                                   CENTRAL_ARG_PROVIDED=true
@@ -419,7 +421,7 @@ print_params() {
      echo "ENABLE_MULTI_NET = $ENABLE_MULTI_NET"
      echo "ENABLE_NETWORK_SEGMENTATION= $ENABLE_NETWORK_SEGMENTATION"
      echo "ENABLE_ROUTE_ADVERTISEMENTS= $ENABLE_ROUTE_ADVERTISEMENTS"
-     echo "ROUTED_UDN_ISOLATION= $ROUTED_UDN_ISOLATION"
+     echo "ADVERTISED_UDN_ISOLATION_MODE= $ADVERTISED_UDN_ISOLATION_MODE"
      echo "ADVERTISE_DEFAULT_NETWORK = $ADVERTISE_DEFAULT_NETWORK"
      echo "ENABLE_PRE_CONF_UDN_ADDR = $ENABLE_PRE_CONF_UDN_ADDR"
      echo "OVN_ENABLE_INTERCONNECT = $OVN_ENABLE_INTERCONNECT"
@@ -666,7 +668,7 @@ set_default_params() {
     echo "Preconfigured UDN addresses requires interconnect to be enabled (-ic)"
     exit 1
   fi
-  ROUTED_UDN_ISOLATION=${ROUTED_UDN_ISOLATION:-Enabled}
+  ADVERTISED_UDN_ISOLATION_MODE=${ADVERTISED_UDN_ISOLATION_MODE:-strict}
   ADVERTISE_DEFAULT_NETWORK=${ADVERTISE_DEFAULT_NETWORK:-false}
   OVN_COMPACT_MODE=${OVN_COMPACT_MODE:-false}
   if [ "$OVN_COMPACT_MODE" == true ]; then
@@ -920,7 +922,7 @@ create_ovn_kube_manifests() {
     --preconfigured-udn-addresses-enable="${ENABLE_PRE_CONF_UDN_ADDR}" \
     --route-advertisements-enable="${ENABLE_ROUTE_ADVERTISEMENTS}" \
     --advertise-default-network="${ADVERTISE_DEFAULT_NETWORK}" \
-    --routed-udn-isolation="${ROUTED_UDN_ISOLATION}" \
+    --advertised-udn-isolation-mode="${ADVERTISED_UDN_ISOLATION_MODE}" \
     --ovnkube-metrics-scale-enable="${OVN_METRICS_SCALE_ENABLE}" \
     --compact-mode="${OVN_COMPACT_MODE}" \
     --enable-interconnect="${OVN_ENABLE_INTERCONNECT}" \
--- a/go-controller/pkg/config/config.go
+++ b/go-controller/pkg/config/config.go
@@ -134,6 +134,7 @@ var (
 	// OVNKubernetesFeatureConfig holds OVN-Kubernetes feature enhancement config file parameters and command-line overrides
 	OVNKubernetesFeature = OVNKubernetesFeatureConfig{
 		EgressIPReachabiltyTotalTimeout: 1,
+		AdvertisedUDNIsolationMode:      AdvertisedUDNIsolationModeStrict,
 	}
 
 	// OvnNorth holds northbound OVN database client and server authentication and location details
@@ -416,27 +417,27 @@ type OVNKubernetesFeatureConfig struct {
 	// EgressIP feature is enabled
 	EnableEgressIP bool `gcfg:"enable-egress-ip"`
 	// EgressIP node reachability total timeout in seconds
-	EgressIPReachabiltyTotalTimeout int    `gcfg:"egressip-reachability-total-timeout"`
-	EnableEgressFirewall            bool   `gcfg:"enable-egress-firewall"`
-	EnableEgressQoS                 bool   `gcfg:"enable-egress-qos"`
-	EnableEgressService             bool   `gcfg:"enable-egress-service"`
-	EgressIPNodeHealthCheckPort     int    `gcfg:"egressip-node-healthcheck-port"`
-	EnableMultiNetwork              bool   `gcfg:"enable-multi-network"`
-	EnableNetworkSegmentation       bool   `gcfg:"enable-network-segmentation"`
-	EnablePreconfiguredUDNAddresses bool   `gcfg:"enable-preconfigured-udn-addresses"`
-	EnableRouteAdvertisements       bool   `gcfg:"enable-route-advertisements"`
-	EnableMultiNetworkPolicy        bool   `gcfg:"enable-multi-networkpolicy"`
-	EnableStatelessNetPol           bool   `gcfg:"enable-stateless-netpol"`
-	EnableInterconnect              bool   `gcfg:"enable-interconnect"`
-	EnableMultiExternalGateway      bool   `gcfg:"enable-multi-external-gateway"`
-	EnablePersistentIPs             bool   `gcfg:"enable-persistent-ips"`
-	EnableDNSNameResolver           bool   `gcfg:"enable-dns-name-resolver"`
-	EnableServiceTemplateSupport    bool   `gcfg:"enable-svc-template-support"`
-	EnableObservability             bool   `gcfg:"enable-observability"`
-	EnableNetworkQoS                bool   `gcfg:"enable-network-qos"`
-	RoutedUDNIsolation              string `gcfg:"routed-udn-isolation"`
+	EgressIPReachabiltyTotalTimeout int  `gcfg:"egressip-reachability-total-timeout"`
+	EnableEgressFirewall            bool `gcfg:"enable-egress-firewall"`
+	EnableEgressQoS                 bool `gcfg:"enable-egress-qos"`
+	EnableEgressService             bool `gcfg:"enable-egress-service"`
+	EgressIPNodeHealthCheckPort     int  `gcfg:"egressip-node-healthcheck-port"`
+	EnableMultiNetwork              bool `gcfg:"enable-multi-network"`
+	EnableNetworkSegmentation       bool `gcfg:"enable-network-segmentation"`
+	EnablePreconfiguredUDNAddresses bool `gcfg:"enable-preconfigured-udn-addresses"`
+	EnableRouteAdvertisements       bool `gcfg:"enable-route-advertisements"`
+	EnableMultiNetworkPolicy        bool `gcfg:"enable-multi-networkpolicy"`
+	EnableStatelessNetPol           bool `gcfg:"enable-stateless-netpol"`
+	EnableInterconnect              bool `gcfg:"enable-interconnect"`
+	EnableMultiExternalGateway      bool `gcfg:"enable-multi-external-gateway"`
+	EnablePersistentIPs             bool `gcfg:"enable-persistent-ips"`
+	EnableDNSNameResolver           bool `gcfg:"enable-dns-name-resolver"`
+	EnableServiceTemplateSupport    bool `gcfg:"enable-svc-template-support"`
+	EnableObservability             bool `gcfg:"enable-observability"`
+	EnableNetworkQoS                bool `gcfg:"enable-network-qos"`
 	// This feature requires a kernel fix https://github.com/torvalds/linux/commit/7f3287db654395f9c5ddd246325ff7889f550286
 	// to work on a kind cluster. Flag allows to disable it for current CI, will be turned on when github runners have this fix.
+	AdvertisedUDNIsolationMode string `gcfg:"advertised-udn-isolation-mode"`
 }
 
 // GatewayMode holds the node gateway mode
@@ -452,10 +453,10 @@ const (
 )
 
 const (
-	// RoutedUDNIsolationEnabled pod isolation across advertised UDN networks is enabled.
-	RoutedUDNIsolationEnabled = "Enabled"
-	// RoutedUDNIsolationDisabled pod isolation across advertised UDN networks is disabled.
-	RoutedUDNIsolationDisabled = "Disabled"
+	// AdvertisedUDNIsolationModeStrict pod isolation across advertised UDN networks is enabled.
+	AdvertisedUDNIsolationModeStrict = "strict"
+	// AdvertisedUDNIsolationModeLoose pod isolation across advertised UDN networks is disabled.
+	AdvertisedUDNIsolationModeLoose = "loose"
 )
 
 // GatewayConfig holds node gateway-related parsed config file parameters and command-line overrides
@@ -508,7 +509,7 @@ type GatewayConfig struct {
 	// EphemeralPortRange is the range of ports used by egress SNAT operations in OVN. Specifically for NAT where
 	// the source IP of the NAT will be a shared Node IP address. If unset, the value will be determined by sysctl lookup
 	// for the kernel's ephemeral range: net.ipv4.ip_local_port_range. Format is "<min port>-<max port>".
-	EphemeralPortRange string `gfcg:"ephemeral-port-range"`
+	EphemeralPortRange string `gcfg:"ephemeral-port-range"`
 }
 
 // OvnAuthConfig holds client authentication and location details for
@@ -1113,10 +1114,10 @@ var OVNK8sFeatureFlags = []cli.Flag{
 		Value:       OVNKubernetesFeature.EnableRouteAdvertisements,
 	},
 	&cli.StringFlag{
-		Name:        "routed-udn-isolation",
-		Usage:       "Configure to use pod isolation for BGP advertised UDN networks",
-		Destination: &cliConfig.OVNKubernetesFeature.RoutedUDNIsolation,
-		Value:       OVNKubernetesFeature.RoutedUDNIsolation,
+		Name:        "advertised-udn-isolation-mode",
+		Usage:       "Configure to use pod isolation for BGP advertised UDN networks. Valid values are 'strict' or 'loose'.",
+		Destination: &cliConfig.OVNKubernetesFeature.AdvertisedUDNIsolationMode,
+		Value:       OVNKubernetesFeature.AdvertisedUDNIsolationMode,
 	},
 	&cli.BoolFlag{
 		Name:        "enable-stateless-netpol",
@@ -2029,6 +2030,10 @@ func buildOVNKubernetesFeatureConfig(cli, file *config) error {
 	// And CLI overrides over config file and default values
 	if err := overrideFields(&OVNKubernetesFeature, &cli.OVNKubernetesFeature, &savedOVNKubernetesFeature); err != nil {
 		return err
+	}
+	if OVNKubernetesFeature.AdvertisedUDNIsolationMode != AdvertisedUDNIsolationModeStrict && OVNKubernetesFeature.AdvertisedUDNIsolationMode != AdvertisedUDNIsolationModeLoose {
+		return fmt.Errorf("invalid advertised-udn-isolation-mode %q: expect one of %s or %s",
+			OVNKubernetesFeature.AdvertisedUDNIsolationMode, AdvertisedUDNIsolationModeStrict, AdvertisedUDNIsolationModeLoose)
 	}
 	return nil
 }
--- a/go-controller/pkg/ovn/secondary_layer2_network_controller.go
+++ b/go-controller/pkg/ovn/secondary_layer2_network_controller.go
@@ -590,7 +590,7 @@ func (oc *SecondaryLayer2NetworkController) addUpdateLocalNodeEvent(node *corev1
 				if err != nil {
 					return err
 				}
-				shouldIsolate := isUDNAdvertised && config.OVNKubernetesFeature.RoutedUDNIsolation == config.RoutedUDNIsolationEnabled
+				shouldIsolate := isUDNAdvertised && config.OVNKubernetesFeature.AdvertisedUDNIsolationMode == config.AdvertisedUDNIsolationModeStrict
 				if shouldIsolate {
 					if err = oc.addAdvertisedNetworkIsolation(node.Name); err != nil {
 						return err
--- a/go-controller/pkg/ovn/secondary_layer3_network_controller.go
+++ b/go-controller/pkg/ovn/secondary_layer3_network_controller.go
@@ -908,7 +908,7 @@ func (oc *SecondaryLayer3NetworkController) addNode(node *corev1.Node) ([]*net.I
 		if err := oc.addOrUpdateUDNNodeSubnetEgressSNAT(hostSubnets, node, isUDNAdvertised); err != nil {
 			return nil, err
 		}
-		shouldIsolate := isUDNAdvertised && config.OVNKubernetesFeature.RoutedUDNIsolation == config.RoutedUDNIsolationEnabled
+		shouldIsolate := isUDNAdvertised && config.OVNKubernetesFeature.AdvertisedUDNIsolationMode == config.AdvertisedUDNIsolationModeStrict
 		if shouldIsolate {
 			if err = oc.addAdvertisedNetworkIsolation(node.Name); err != nil {
 				return nil, err
--- a/test/e2e/route_advertisements.go
+++ b/test/e2e/route_advertisements.go
@@ -20,10 +20,10 @@ import (
 	apitypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/types"
 	udnv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1"
 	udnclientset "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/clientset/versioned"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/deploymentconfig"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/feature"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/images"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider"
 	infraapi "github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider/api"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/label"
@@ -854,7 +854,10 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
 						curlOutput string
 						curlErr    bool
 					)
-					if os.Getenv("ROUTED_UDN_ISOLATION") == "Disabled" {
+					// Test behavior depends on the ADVERTISED_UDN_ISOLATION_MODE environment variable:
+					// - "loose": Pod connectivity is allowed, test expects success
+					// - anything else (including unset): Treated as "strict", pod connectivity is blocked
+					if os.Getenv("ADVERTISED_UDN_ISOLATION_MODE") == "loose" {
 						clientPodStatus, err := getPodAnnotationForAttachment(clientPod, namespacedName(clientPod.Namespace, cudnATemplate.Name))
 						framework.ExpectNoError(err)
 
@@ -881,7 +884,7 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
 						curlOutput string
 						curlErr    bool
 					)
-					if os.Getenv("ROUTED_UDN_ISOLATION") == "Disabled" {
+					if os.Getenv("ADVERTISED_UDN_ISOLATION_MODE") == "loose" {
 						clientPodStatus, err := getPodAnnotationForAttachment(clientPod, namespacedName(clientPod.Namespace, cudnATemplate.Name))
 						framework.ExpectNoError(err)
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -483,7 +483,7 @@ jobs:
           - {"target": "network-segmentation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv6", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "bgp", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "snatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation", "dns-name-resolver": "enable-dns-name-resolver"}
           - {"target": "bgp", "ha": "noHA", "gateway-mode": "shared",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation", "dns-name-resolver": "enable-dns-name-resolver"}
-          - {"target": "bgp-loose-isolation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "snatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation", "routed-udn-isolation": "Disabled"}
+          - {"target": "bgp-loose-isolation", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "snatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "network-segmentation": "enable-network-segmentation", "advertised-udn-isolation-mode": "loose"}
           - {"target": "traffic-flow-test-only","ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "traffic-flow-tests": "1-24", "network-segmentation": "enable-network-segmentation"}
           - {"target": "tools", "ha": "noHA", "gateway-mode": "local", "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "network-segmentation": "enable-network-segmentation"}
     needs: [ build-pr ]
@@ -515,7 +515,7 @@ jobs:
       ENABLE_ROUTE_ADVERTISEMENTS: "${{ matrix.routeadvertisements != '' }}"
       ADVERTISE_DEFAULT_NETWORK:  "${{ matrix.routeadvertisements == 'advertise-default' }}"
       ENABLE_PRE_CONF_UDN_ADDR: "${{ matrix.ic == 'ic-single-node-zones' && (matrix.target == 'network-segmentation' || matrix.network-segmentation == 'enable-network-segmentation') }}"
-      ROUTED_UDN_ISOLATION: "${{ matrix.routed-udn-isolation }}"
+      ADVERTISED_UDN_ISOLATION_MODE: "${{ matrix.advertised-udn-isolation-mode }}"
     steps:
 
     - name: Install VRF kernel module
--- a/dist/images/daemonset.sh
+++ b/dist/images/daemonset.sh
@@ -74,7 +74,7 @@ OVN_NETWORK_SEGMENTATION_ENABLE=
 OVN_PRE_CONF_UDN_ADDR_ENABLE=
 OVN_ROUTE_ADVERTISEMENTS_ENABLE=
 OVN_ADVERTISE_DEFAULT_NETWORK=
-OVN_ROUTED_UDN_ISOLATION=
+OVN_ADVERTISED_UDN_ISOLATION_MODE=
 OVN_V4_JOIN_SUBNET=""
 OVN_V6_JOIN_SUBNET=""
 OVN_V4_MASQUERADE_SUBNET=""
@@ -284,8 +284,8 @@ while [ "$1" != "" ]; do
   --advertise-default-network)
     OVN_ADVERTISE_DEFAULT_NETWORK=$VALUE
     ;;
-    --routed-udn-isolation)
-    OVN_ROUTED_UDN_ISOLATION=$VALUE
+    --advertised-udn-isolation-mode)
+    OVN_ADVERTISED_UDN_ISOLATION_MODE=$VALUE
     ;;
   --egress-service-enable)
     OVN_EGRESSSERVICE_ENABLE=$VALUE
@@ -482,8 +482,8 @@ ovn_route_advertisements_enable=${OVN_ROUTE_ADVERTISEMENTS_ENABLE}
 echo "ovn_route_advertisements_enable: ${ovn_route_advertisements_enable}"
 ovn_advertise_default_network=${OVN_ADVERTISE_DEFAULT_NETWORK}
 echo "ovn_advertise_default_network: ${ovn_advertise_default_network}"
-ovn_routed_udn_isolation=${OVN_ROUTED_UDN_ISOLATION}
-echo "ovn_routed_udn_isolation: ${ovn_routed_udn_isolation}"
+ovn_advertised_udn_isolation_mode=${OVN_ADVERTISED_UDN_ISOLATION_MODE}
+echo "ovn_advertised_udn_isolation_mode: ${ovn_advertised_udn_isolation_mode}"
 ovn_hybrid_overlay_net_cidr=${OVN_HYBRID_OVERLAY_NET_CIDR}
 echo "ovn_hybrid_overlay_net_cidr: ${ovn_hybrid_overlay_net_cidr}"
 ovn_disable_snat_multiple_gws=${OVN_DISABLE_SNAT_MULTIPLE_GWS}
@@ -626,7 +626,7 @@ ovn_image=${ovnkube_image} \
   ovn_network_segmentation_enable=${ovn_network_segmentation_enable} \
   ovn_pre_conf_udn_addr_enable=${ovn_pre_conf_udn_addr_enable} \
   ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
-  ovn_routed_udn_isolation=${ovn_routed_udn_isolation} \
+  ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
   ovn_egress_service_enable=${ovn_egress_service_enable} \
   ovn_ssl_en=${ovn_ssl_en} \
   ovn_remote_probe_interval=${ovn_remote_probe_interval} \
@@ -681,7 +681,7 @@ ovn_image=${ovnkube_image} \
   ovn_multi_network_enable=${ovn_multi_network_enable} \
   ovn_network_segmentation_enable=${ovn_network_segmentation_enable} \
   ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
-  ovn_routed_udn_isolation=${ovn_routed_udn_isolation} \
+  ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
   ovn_egress_service_enable=${ovn_egress_service_enable} \
   ovn_ssl_en=${ovn_ssl_en} \
   ovn_remote_probe_interval=${ovn_remote_probe_interval} \
@@ -781,7 +781,7 @@ ovn_image=${ovnkube_image} \
   ovn_multi_network_enable=${ovn_multi_network_enable} \
   ovn_network_segmentation_enable=${ovn_network_segmentation_enable} \
   ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
-  ovn_routed_udn_isolation=${ovn_routed_udn_isolation} \
+  ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
   ovn_egress_service_enable=${ovn_egress_service_enable} \
   ovn_ssl_en=${ovn_ssl_en} \
   ovn_master_count=${ovn_master_count} \
@@ -832,7 +832,7 @@ ovn_image=${ovnkube_image} \
   ovn_network_segmentation_enable=${ovn_network_segmentation_enable} \
   ovn_pre_conf_udn_addr_enable=${ovn_pre_conf_udn_addr_enable} \
   ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
-  ovn_routed_udn_isolation=${ovn_routed_udn_isolation} \
+  ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
   ovn_egress_service_enable=${ovn_egress_service_enable} \
   ovn_ssl_en=${ovn_ssl_en} \
   ovn_master_count=${ovn_master_count} \
@@ -914,7 +914,7 @@ ovn_image=${ovnkube_image} \
   ovn_network_segmentation_enable=${ovn_network_segmentation_enable} \
   ovn_pre_conf_udn_addr_enable=${ovn_pre_conf_udn_addr_enable} \
   ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
-  ovn_routed_udn_isolation=${ovn_routed_udn_isolation} \
+  ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
   ovn_egress_service_enable=${ovn_egress_service_enable} \
   ovn_ssl_en=${ovn_ssl_en} \
   ovn_remote_probe_interval=${ovn_remote_probe_interval} \
@@ -983,7 +983,7 @@ ovn_image=${ovnkube_image} \
   ovn_network_segmentation_enable=${ovn_network_segmentation_enable} \
   ovn_pre_conf_udn_addr_enable=${ovn_pre_conf_udn_addr_enable} \
   ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
-  ovn_routed_udn_isolation=${ovn_routed_udn_isolation} \
+  ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
   ovn_ssl_en=${ovn_ssl_en} \
   ovn_remote_probe_interval=${ovn_remote_probe_interval} \
   ovn_monitor_all=${ovn_monitor_all} \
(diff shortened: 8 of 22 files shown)


tssurya:
@pperiyasamy can you now please rebase on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140 and see how CI flares here tomorrow for loose mode?

tssurya:
> @pperiyasamy can you now please rebase on #5140 and see how CI flares here tomorrow for loose mode?

also since CI is down you might have to manually test this :D I'll fix UTs tomorrow but functional testing would be good to see

pperiyasamy:
> also since CI is down you might have to manually test this :D I'll fix UTs tomorrow but functional testing would be good to see

@tssurya Tested https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140 PR locally with loose isolation mode, functionality seems to be intact with this mode.  The E2Es from the commit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5276/commits/d43d70a3d7d9865baa3d8dead79b5e293f54de7d are passing. great work !

The below E2Es are failing for same node case.

```
[ReportAfterSuite] OVN-Kubernetes e2e JUnit report
/mnt/ovn-kubernetes/test/e2e/testcontext.go:103
[ReportAfterSuite] PASSED [0.010 seconds]
[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report
autogenerated by Ginkgo
[ReportAfterSuite] PASSED [0.053 seconds]

Summarizing 3 Failures:
  [FAIL] BGP: isolation between advertised networks Layer3 connectivity between networks [It] UDN pod to the same node nodeport service in different UDN should work
  /mnt/ovn-kubernetes/test/e2e/route_advertisements.go:820
  [FAIL] BGP: isolation between advertised networks Layer2 connectivity between networks [It] UDN pod to the same node nodeport service in default network should work
  /mnt/ovn-kubernetes/test/e2e/route_advertisements.go:820
  [FAIL] BGP: isolation between advertised networks Layer2 connectivity between networks [It] UDN pod to the same node nodeport service in different UDN should work
  /mnt/ovn-kubernetes/test/e2e/route_advertisements.go:820

Ran 42 of 516 Specs in 1467.199 seconds
FAIL! -- 39 Passed | 3 Failed | 2 Pending | 472 Skipped

```

go-controller/pkg/config/config.go
```diff
@@ -1111,6 +1119,12 @@ var OVNK8sFeatureFlags = []cli.Flag{
 		Destination: &cliConfig.OVNKubernetesFeature.EnableRouteAdvertisements,
 		Value:       OVNKubernetesFeature.EnableRouteAdvertisements,
 	},
+	&cli.StringFlag{
+		Name:        "routed-udn-isolation",
+		Usage:       "Configure to use pod isolation for BGP advertised UDN networks",
+		Destination: &cliConfig.OVNKubernetesFeature.RoutedUDNIsolation,
+		Value:       OVNKubernetesFeature.RoutedUDNIsolation,
+	},
```
jcaamano: This doesn't seem to be aligned to what we are doing on CNO side: https://github.com/openshift/cluster-network-operator/pull/2714#discussion_r2213382893

tssurya:
@pperiyasamy are coderabbit comments addressed here? and PR needs rebase also FYI

test/e2e/route_advertisements.go
```diff
@@ -343,8 +331,179 @@ var _ = ginkgo.Describe("BGP: Pod to external server when CUDN network is advert
...
+					// Delete IPv6 route if supported
+					if isIPv6Supported(f.ClientSet) && externalServerV6CIDR != "" && frrContainerIPv6 != "" {
+						_, err := e2ekubectl.RunKubectl(ovnKubernetesNamespace, "exec", dbPod, "-c", dbContainerName, "--",
+							"ovn-nbctl", "--no-leader-only", "lr-route-del", "GR_"+grName, externalServerV6CIDR)
+						if err != nil {
+							framework.Logf("Warning: Failed to delete IPv6 BGP route from GR_%s: %v", grName, err)
+						}
+					}
+				}
+			}
```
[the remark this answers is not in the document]
pperiyasamy: this suggestion is not valid for changes in the PR.

test/e2e/route_advertisements.go
```diff
@@ -858,26 +1012,56 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+						// With the above underlay routing configuration client pod can reach server pod.
+						curlOutput = clientPodStatus.IPs[ipFamilyIndex].IP.String()
+						curlErr = false
+					} else {
+						curlOutput = curlConnectionTimeoutCode
+						curlErr = true
+					}
+					return clientPod.Name, clientPod.Namespace, net.JoinHostPort(srvPodStatus.IPs[ipFamilyIndex].IP.String(), "8080") + "/clientip",
+						curlOutput, curlErr
 				}),
```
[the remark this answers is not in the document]
pperiyasamy: the current change is ok, when `isolationMode` is set with "loose" mode, then curl command should pass, all other cases are considered as "strict" mode and it is handled appropriately, so this is also a bit reliable and readable.

contrib/kind.sh
```diff
@@ -341,6 +341,8 @@ parse_args() {
                                                   ;;
             -adv | --advertise-default-network) ADVERTISE_DEFAULT_NETWORK=true
                                                   ;;
+            -rui | --routed-udn-isolation)      ROUTED_UDN_ISOLATION=Disabled
+                                                  ;;
```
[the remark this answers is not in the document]
pperiyasamy: let's not make it more complex, because `ROUTED_UDN_ISOLATION` is enabled by default, when user wants to disable it, then it can be done with `-rui` option.

go-controller/pkg/config/config.go
pperiyasamy: @pliurh aligned it with commit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5276/commits/db63749eff05b498c77668467041448bc7708ff2 as per latest CNO PR https://github.com/openshift/cluster-network-operator/pull/2752.

test/e2e/service.go
```diff
@@ -1791,9 +1791,9 @@ metadata:
 		bgpRouteCommand := strings.Split(fmt.Sprintf("ip%s route show %s", ipVer, svcLoadBalancerIP), " ")
 		cmd = append(cmd, bgpRouteCommand...)
 
-		backendNodeIP, err := getNodeIP(f.ClientSet, backendNodeName)
+		backendNodeIP, err := GetNodeIP(f.ClientSet, backendNodeName)
 		framework.ExpectNoError(err, fmt.Sprintf("failed to get node's %s node ip address", backendNodeName))
-		nonBackendNodeIP, err := getNodeIP(f.ClientSet, backendNodeName)
+		nonBackendNodeIP, err := GetNodeIP(f.ClientSet, backendNodeName)
 		framework.ExpectNoError(err, fmt.Sprintf("failed to get node's %s node ip address", backendNodeName))
```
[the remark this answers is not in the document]
pperiyasamy: not a valid comment anymore.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -684,21 +684,28 @@ func (c *Controller) generateFRRConfiguration(
...
+								LE:     selectedNetworks.prefixLength[prefix],
+								GE:     selectedNetworks.prefixLength[prefix],
+							},
+						)
+					}
+				}
+			} else {
+				klog.Infof("Skip receiving prefixes from neighbor %s for user-defined pod network RA %s",
+					neighbor.Address, ra.Name)
 			}
```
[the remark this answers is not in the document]
pperiyasamy: not a valid comment anymore.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -831,12 +831,38 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+func getClusterNodesDestinationBasedSNATMatch(ipv4ClusterNodeIPAS, ipv6ClusterNodeIPAS string) string {
+	if config.IPv4Mode && config.IPv6Mode {
+		return fmt.Sprintf("ip4.dst == $%s || ip6.dst == $%s", ipv4ClusterNodeIPAS, ipv6ClusterNodeIPAS)
+	} else if config.IPv4Mode {
+		return fmt.Sprintf("ip4.dst == $%s", ipv4ClusterNodeIPAS)
+	} else if config.IPv6Mode {
+		return fmt.Sprintf("ip6.dst == $%s", ipv6ClusterNodeIPAS)
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
pperiyasamy: not a valid comment anymore.

contrib/kind.sh
```diff
@@ -347,6 +347,8 @@ parse_args() {
                                                   ;;
             -adv | --advertise-default-network) ADVERTISE_DEFAULT_NETWORK=true
                                                   ;;
+            -rui | --routed-udn-isolation)      ROUTED_UDN_ISOLATION=Disabled
```
jcaamano: it this not aligned with the commit message which says `-rnd` or `--routed-udn-isolation-disable`?

go-controller/pkg/controllermanager/controller_manager.go
```diff
@@ -532,6 +532,24 @@ func (cm *ControllerManager) Reconcile(_ string, _, _ util.NetInfo) error {
 
 func (cm *ControllerManager) configureAdvertisedNetworkIsolation() error {
 	addressSetFactory := addressset.NewOvnAddressSetFactory(cm.nbClient, config.IPv4Mode, config.IPv6Mode)
-	_, err := addressSetFactory.EnsureAddressSet(ovn.GetAdvertisedNetworkSubnetsAddressSetDBIDs())
-	return err
+	if config.OVNKubernetesFeature.UDNIsolationMode == config.UDNIsolationModeStrict {
+		_, err := addressSetFactory.EnsureAddressSet(ovn.GetAdvertisedNetworkSubnetsAddressSetDBIDs())
+		return err
+	}
+	klog.Infof("Ensure global advertised networks addressset and tier-0 drop ACLs are removed in loose UDN isolation mode")
```
jcaamano: is this log really necessary at this level?

go-controller/pkg/ovn/udn_isolation.go
```diff
@@ -307,6 +308,23 @@ func BuildAdvertisedNetworkSubnetsDropACL(advertisedNetworkSubnetsAddressSet add
 // pass   "(ip[4|6].src == <UDN_SUBNET> && ip[4|6].dst == <UDN_SUBNET>)"                1100
 // drop   "(ip[4|6].src == $<ALL_ADV_SUBNETS> && ip[4|6].dst == $<ALL_ADV_SUBNETS>)"    1050
 func (bnc *BaseNetworkController) addAdvertisedNetworkIsolation(nodeName string) error {
+	if config.OVNKubernetesFeature.UDNIsolationMode == config.UDNIsolationModeLoose {
+		klog.Infof("The network %s is configured with loose isolation mode, so delete tier-0 pass ACL rule if it exists",
```
jcaamano: is this log really necessary at this level?

go-controller/pkg/ovn/udn_isolation.go
```diff
@@ -309,8 +309,20 @@ func BuildAdvertisedNetworkSubnetsDropACL(advertisedNetworkSubnetsAddressSet add
...
+		// method.
+		passACLs, _ := bnc.getPassACLsForAdvertisedNetwork()
+		if len(passACLs) == 0 {
+			return nil
+		}
+		p := func(sw *nbdb.LogicalSwitch) bool { return sw.Name == bnc.GetNetworkScopedSwitchName(nodeName) }
+		err := libovsdbops.RemoveACLsFromLogicalSwitchesWithPredicate(bnc.nbClient, p, passACLs...)
+		if err != nil {
+			return fmt.Errorf("failed to delete tier-0 pass ACLs for network %s: %w", bnc.GetNetworkName(), err)
+		}
```
jcaamano: I am confused: why are we just not calling `addAdvertisedNetworkIsolation` or `deleteAdvertisedNetworkIsolation` depending on the value `config.OVNKubernetesFeature.RoutedUDNIsolation` instead of adding removal code in this `addAdvertisedNetworkIsolation`  which makes less sense.

go-controller/pkg/controllermanager/controller_manager.go
```diff
@@ -531,11 +531,25 @@ func (cm *ControllerManager) Reconcile(_ string, _, _ util.NetInfo) error {
...
+	dropACLIDs := ovn.GetAdvertisedNetworkSubnetsDropACLdbIDs()
+	dropACLPredicate := libovsdbops.GetPredicate[*nbdb.ACL](dropACLIDs, nil)
+	dropACLs, _ := libovsdbops.FindACLsWithPredicate(cm.nbClient, dropACLPredicate)
+	if len(dropACLs) > 0 {
+		p := func(_ *nbdb.LogicalSwitch) bool { return true }
+		err := libovsdbops.RemoveACLsFromLogicalSwitchesWithPredicate(cm.nbClient, p, dropACLs...)
+		if err != nil {
+			return err
+		}
+	}
```
jcaamano: why are we just not calling addAdvertisedNetworkIsolation or deleteAdvertisedNetworkIsolation depending on the value config.OVNKubernetesFeature.RoutedUDNIsolation which is what originally handled the drop ACLs instead of removing drop ACLs here which originally it is just a method handling the address set?

About the address set itself: previously we were not removing it if RA was disabled. I guess we should do that as well?

go-controller/pkg/config/config.go
```diff
@@ -1102,6 +1113,12 @@ var OVNK8sFeatureFlags = []cli.Flag{
 		Destination: &cliConfig.OVNKubernetesFeature.EnableRouteAdvertisements,
 		Value:       OVNKubernetesFeature.EnableRouteAdvertisements,
 	},
+	&cli.StringFlag{
+		Name:        "udn-isolation-mode",
```
jcaamano: it strikes me now that this could probably be clearer as `advertised-udn-isolation-mode`. Thoughts?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -786,21 +772,35 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
-					ipv := getIPv(mgmtIP)
+				mgmtIP := util.GetNodeManagementIfAddr(cidr)
+				if mgmtIP == nil {
+					return nil, fmt.Errorf("unable to determine management IP for subnet %s.", cidr)
+				}
+				dftFlows = append(dftFlows,
+					fmt.Sprintf("cookie=%s, priority=301, table=0, in_port=%s, %s, %s_dst=%s, "+
+						"actions=output:%s",
+						nodetypes.DefaultOpenFlowCookie, ofPortPhys, ipv, ipv, mgmtIP.IP, nodetypes.OvsLocalPort),
+				)
```
jcaamano: Previously we were not adding this flow in SGW mode, as the 300 flow was sufficient. Any specific reason you decided to add it now regardless of the mode? Can you add the explanation to the commit message?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -786,21 +772,35 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+				// skip adding egress flow if the network is advertised in LGW mode
+				// and the network is not the default network
+				if isAdvertisedLGW && !netConfig.IsDefaultNetwork() {
+					continue
 				}
+				// ICNIv2 and advertised UDN networks requires that local pod traffic can leave the node without SNAT.
+				// We match on the pod subnets and forward the traffic to the physical interface
+				dftFlows = append(dftFlows, fmt.Sprintf("cookie=%s, priority=104, in_port=%s, dl_src=%s, %s, %s_src=%s, "+
+					"actions=output:%s",
+					nodetypes.DefaultOpenFlowCookie, netConfig.OfPortPatch, bridgeMacAddress, ipv, ipv, cidr, ofPortPhys))
```
jcaamano: Previously this flow was not added if the 103 flow was not added.

Previously this flow was added on ~~SGW~~ LGW mode as well. 

Can you explain and add to the commit message your reasoning for making these changes?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Previously, we were also adding this flow in SGW mode, but with the output LOCAL. I didn't change the behavior.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: The (priority=103, previous priority=109) flow is only added when EgressIP is enabled. But this flow (priority=104, previous priority=109) is for ICNI or advertised UDN regardless of whether EgressIP is enabled. So these 2 flows shall be added independently. 
And both flows are also added on LGW mode in the new code.

go-controller/pkg/ovn/udn_isolation.go
pliurh: @pperiyasamy can you chime in?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
jcaamano: Sorry. I meant to say LGW mode.

The flow was guarded by a condition
```
if output == netConfig.OfPortPatch {
```
And that itself depended on 
```
			if isNetworkAdvertised && config.Gateway.Mode == config.GatewayModeLocal {
				// except if advertised through BGP, go to kernel
				// TODO: MEG enabled pods should still go through the patch port
				// but holding this until
				// https://issues.redhat.com/browse/FDP-646 is fixed, for now we
				// are assuming MEG & BGP are not used together
				output = nodetypes.OvsLocalPort
			}
```
so it wouldn't be added in LGW mode, just SGW mode. Now it is being added in both modes.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
jcaamano: > The (priority=103, previous priority=109) flow is only added when EgressIP is enabled. 

I think you meant to say (priority=103, previous priority=***104***)?

> But this flow (priority=104, previous priority=109) is for ICNI or advertised UDN regardless of whether EgressIP is enabled.

Why? 
Before it was just to avoid the previous 104 flow when EIP was enabled.
I guess it is because we no longer need to depend on CT mark to forward the reply traffic.
But there is the alternative of adding it only if EIP is enabled or if the network is advertised instead of always.

> And both flows are also added on LGW mode in the new code.

Just on the code above, you have
```
				// skip adding egress flow if the network is advertised in LGW mode
				// and the network is not the default network
				if isAdvertisedLGW && !netConfig.IsDefaultNetwork() {
					continue
				}
```
So it is not being added for advertised CUDN in LGW mode, right? 
But the previous priority=109 was, right?
Why the change?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Well, you're right, this flow is not required in LGW mode, it has the same output port as the 300 flow.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: > I think you meant to say (priority=103, previous priority=104)?

Yes.

>I guess it is because we no longer need to depend on CT mark to forward the reply traffic.

Yes.

> But there is the alternative of adding it only if EIP is enabled or if the network is advertised instead of always.

This is the current behavior. We don't add the 104 flow if `!disableSNATMultipleGWs && !isNetworkAdvertised`. But the 103 flow shall be added regardless. I need to move that code.

>But the previous priority=109 was, right?
Why the change?

Yes. ~~But this flow won't be hit in LGW mode, as the `in_port` doesn't match. So I removed it from LGM mode.~~ The traffic path of EgressIP for UDN is the same in LGW and SGW modes. I'll change it back.

go-controller/pkg/controllermanager/controller_manager.go
pperiyasamy: ok, this was done in the assumption that both ACLs and AS must be removed when RoutedUDNIsolation is disabled. as you said this address set can be just there and only ACLs must be removed. so think we can just invoke `addAdvertisedNetworkIsolation` or `deleteAdvertisedNetworkIsolation` methods accordingly for node events, i think that also should take care cleaning up things for existing advertised UDN networks as well. WDYT ?

go-controller/pkg/ovn/udn_isolation.go
pperiyasamy: same comment as before.

go-controller/pkg/ovn/udn_isolation.go
pperiyasamy: yes, moved to log level 5.

go-controller/pkg/controllermanager/controller_manager.go
pperiyasamy: moved to log level 5.

contrib/kind.sh
pperiyasamy: right, updated it.

go-controller/pkg/config/config.go
pliurh: Yeah, it's more accurate.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: fixed.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: fixed.

go-controller/pkg/controllermanager/controller_manager.go
jcaamano: Sorry, I didn't explain myself very well.

The AS must be removed when RoutedUDNIsolation is disabled. In a similar way, I also think the AS  must be removed when RouteAdvertisemets is disabled, shouldn't it?

On top of that, I don't think how we handle the AS has anything to do with my other part of the comment: we can just call  addAdvertisedNetworkIsolation or deleteAdvertisedNetworkIsolation as previously implemented depending on the value config.OVNKubernetesFeature.RoutedUDNIsolation and regardless of how we handle the AS, can't we?

jcaamano: Ok so after discussing this offline, the best thing to do is to leave the AS in place. An empty AS is not going to break the bank.

So recap, no changes needed for `addAdvertisedNetworkIsolation`  or `deleteAdvertisedNetworkIsolation` which should just be called according to isolation mode.

I think this kind of pattern is wrong:
```
					if util.IsRouteAdvertisementsEnabled() {
						if err = oc.deleteAdvertisedNetworkIsolation(node.Name); err != nil {
							return err
						}
					}
```
shouldn't we clean up when RouteAdvertisements is disabled, just as we should when isolation mode changes to strict?

I we need to optimize the delete path, we could consider leaving the ACLs in place as well. The delete logic itself could be optimized further: instead of running the predicate over the full ACL table, we could only run it by the ACLs assciated to he network switch.

pperiyasamy: ok @jcaamano, updated the PR for not have any changes to `addAdvertisedNetworkIsolation` or `deleteAdvertisedNetworkIsolation` and calling them accordingly depending on the value `config.OVNKubernetesFeature.RoutedUDNIsolation`. and leaving AS lying there without any subnet IPs in it in case of loose isolation mode. The changes seem to work with my kind cluster for both old and new advertised UDNs.
regarding delete code path optimization and refactoring invocation logic, will work with @kyrtapz on this.

pperiyasamy:
@pliurh The UT changes [here](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5276/commits/468b539957006950299ea30b59d310c0ae164556#diff-31c03edf99cf8745fccaa45cb21c0a10a04a9c470d5a5b150bb79a8006dcb5fa) is failing now, I'm not sure if missed something during my rebase, PTAL.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,96 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+				}
+				// MEG and advertised UDN networks requires that local pod traffic can leave the node without SNAT.
+				// We match on the pod subnets and forward the traffic to the physical interface.
+				// Select priority 104 for the senario when both EgressIP and advertised UDN are active:
+				// 1. Override egressIP drop flows (priority 103)
+				// 2. Still allow egressIP flows at priority 105
+				dftFlows = append(dftFlows,
+					fmt.Sprintf("cookie=%s, priority=104, in_port=%s, dl_src=%s, %s, %s_src=%s, "+
+						"actions=output:%s",
+						nodetypes.DefaultOpenFlowCookie, netConfig.OfPortPatch, bridgeMacAddress, ipv, ipv, subnet, ofPortPhys))
```
jcaamano: Bear with me, I still don't understand why this is being done. I understand what you are doing, which is what the comment explains, but not why. I am not sure if I am expressing correctly it is not the `what` which I don't understand, its the `why`

~~Let's imagine EIP is disabled, so neither 104 or 103 flows will be there. What egress flow are we going to hit for an advertised CUDN?~~

~~Now let's imagine EIP is enabled and 103 flow is there.
For an advertised CUDN in SGW mode 104 flow will be there and will be hit.~~
For an advertised CUDN on LGW mode 104 flow won't be ~~there~~ hit, what will we hit for egress? 

Is it because 104 matches on patch port that you are considering we don't need it in LGW mode as egress traffic will come through LOCAL port?

But we decided we don't want to depend on conntrack to do plain ingress/egress for advertise networks. So if we don't have something like 104 flow for LGW mode, won't we hit flows on later stages that depend on conntrack?

jcaamano: ~~There is a bit I missunderstood, `(isAdvertisedLGW || !disableSNATMultipleGWs)` can only be true for the default network, I will edit my comment above with this consideration.~~

jcaamano: Another scenario: 
default network advertised in SGW mode with MEG disabled: what egress flow do we hit?

pliurh: > For an advertised CUDN on LGW mode 104 flow won't be there hit, what will we hit for egress?

It will hit `priority=100,ip,in_port=LOCAL actions=ct(commit,zone=64000,exec(load:0x2->NXM_NX_CT_MARK[])),output:eth0`
And the reply traffic will hit `priority=100,ct_state=+rel+trk,ct_mark=0x2,ip actions=LOCAL`

> Is it because 104 matches on patch port that you are considering we don't need it in LGW mode as egress traffic will come through LOCAL port?

Yes.

> won't we hit flows on later stages that depend on conntrack?

Yes, we will hit the 100 flows. However, it seems that the 100 flows are not designed for our case, although it works. I think we shall add 104 flow for LGW mode as well.

> default network advertised in SGW mode with MEG disabled: what egress flow do we hit?

The 104 flow.

jcaamano: > Yes, we will hit the 100 flows. However, it seems that the 100 flows are not designed for our case, although it works. I think we shall add 104 flow for LGW mode as well.

That would make more sense yeah, we decided to not use conntrack but I guess that is a decision that applies both to SGW and LGW modes unless there is a reason not to.

> > default network advertised in SGW mode with MEG disabled: what egress flow do we hit?
> 
> The 104 flow.

and EIP disabled I meant.

go-controller/pkg/ovn/secondary_layer2_network_controller.go
```diff
@@ -591,16 +591,17 @@ func (oc *SecondaryLayer2NetworkController) addUpdateLocalNodeEvent(node *corev1
 				if err != nil {
 					return err
 				}
-				if !isUDNAdvertised {
-					if util.IsRouteAdvertisementsEnabled() {
+				shouldIsolate := isUDNAdvertised && config.OVNKubernetesFeature.UDNIsolationMode == config.UDNIsolationModeStrict
+				if util.IsRouteAdvertisementsEnabled() {
```
jcaamano: nit: remove this check (just as it is not being done for the default network)

go-controller/pkg/ovn/secondary_layer3_network_controller.go
```diff
@@ -909,15 +909,16 @@ func (oc *SecondaryLayer3NetworkController) addNode(node *corev1.Node) ([]*net.I
 		if err := oc.addOrUpdateUDNNodeSubnetEgressSNAT(hostSubnets, node, isUDNAdvertised); err != nil {
 			return nil, err
 		}
-		if !isUDNAdvertised {
-			if util.IsRouteAdvertisementsEnabled() {
-				if err := oc.deleteAdvertisedNetworkIsolation(node.Name); err != nil {
+		shouldIsolate := isUDNAdvertised && config.OVNKubernetesFeature.UDNIsolationMode == config.UDNIsolationModeStrict
+		if util.IsRouteAdvertisementsEnabled() {
```
jcaamano: nit: remove this check (just as it is not being done for the default network)

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: > and EIP disabled I meant.

Still the 104 flow. I'll change the code, the 104 flow will only be skipped if `!disableSNATMultipleGWs && !config.OVNKubernetesFeature.EnableEgressIP`.

go-controller/pkg/config/config.go
pliurh: fixed

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+				if (disableSNATMultipleGWs && config.OVNKubernetesFeature.EnableEgressIP) || isNetworkAdvertised {
+					// MEG and advertised UDN networks requires that local pod traffic can leave the node without SNAT.
+					// We match on the pod subnets and forward the traffic to the physical interface.
+					// Select priority 104 for the senario when both EgressIP and advertised UDN are active:
+					// 1. Override egressIP drop flows (priority 103)
+					// 2. Still allow egressIP flows at priority 105
+					dftFlows = append(dftFlows,
+						fmt.Sprintf("cookie=%s, priority=104, in_port=%s, dl_src=%s, %s, %s_src=%s, "+
+							"actions=output:%s",
+							nodetypes.DefaultOpenFlowCookie, input, bridgeMacAddress, ipv, ipv, subnet, ofPortPhys))
```
jcaamano, line 826: Worth noting that we are short-circuiting the 102 flows which allow the default network to inter-connect with a localnet network afaik (we already did with MEG & SGW mode, now also in LGW mode) but I don't see that as a working configuration for an advertised network so it is fine by me.

pliurh: Added to the comment.

jcaamano: I didn't really ask to add the comment sorry, I was just doing a highlight.
This is not accurate as far as the reason I had in mind.
```
// But it is not a problem since localnet networks cannot be advertised.
```
So I thought that the advertised default network would have to be on the same subnet as the nodes/localnet for that to work and I don't think we allow that kind of configuration.

However I just now realized that we are SNATing E/W even for an advertised network so in that case this 104 flow wouldn't even be matched and is not short-circuiting anything.

Sorry for all this confusion.

jcaamano:
These coderabbitai comments might be worth addressing:


https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5276/files#r2265875816

```
820-823: Typo in comment (“senario” → “scenario”)
```
```
232-232: Consider adding validation test for invalid mode values
```
```
857-869: Consider documenting the environment variable behavior
```

pliurh:
@trozet PTAL

contrib/kind-common
```diff
@@ -758,22 +758,52 @@ deploy_bgp_external_server() {
...
+  local route_cmd="route replace default via"
+
+  # Check if $gw is IPv6 (contains ':')
+  if [[ "$gw" == *:* ]]; then
+    ip_cmd="ip -6"
+  fi
+
+  KIND_NODES=$(kind_get_nodes)
+  for n in $KIND_NODES; do
+    $OCI_BIN exec "$node" $ip_cmd $route_cmd "$gw"
```
jcaamano: s/n/node

jcaamano:
LGTM
Let's see if either @trozet or @tssurya can throw another pair of eyes to those pesky bridge flows.

docs/features/cluster-egress-controls/egress-ip.md
tssurya, line 1: the commit message tells what the commit is doing, but doesn't mention what problem the commit is trying to solve - its hard for me to isolated-ly review this commit if I don't know why these flow changes are being made.
@pliurh what are the problems this commit is trying to solve?

tssurya, line 1: the commit message tells what the commit is doing - ie the changes proposed, but doesn't mention what problem the commit is trying to solve - its hard for me to isolated-ly review this commit if I don't know why these flow changes are being made.
@pliurh what are the problems this commit is trying to solve?

IIUC priority 15 and 16 flows are changing to priority 300 and 301.
Why did we have to change/touch flows are priority 109 and 104 as part of this commit?

test/e2e/external_gateways.go
```diff
@@ -857,7 +857,7 @@ var _ = ginkgo.Describe("External Gateway", feature.ExternalGateway, func() {
 				podConnEntriesWithMACLabelsSet := pokeConntrackEntries(nodeName, addresses.srcPodIP, protocol, macAddressGW)
 				gomega.Expect(podConnEntriesWithMACLabelsSet).To(gomega.Equal(2))
 				totalPodConnEntries := pokeConntrackEntries(nodeName, addresses.srcPodIP, protocol, nil)
-				gomega.Expect(totalPodConnEntries).To(gomega.Equal(6)) // total conntrack entries for this pod/protocol
+				gomega.Expect(totalPodConnEntries).To(gomega.Equal(4)) // total conntrack entries for this pod/protocol
```
tssurya, line 860: I don't follow why the number of conntrack entries are reducing here? its not mentioned in the commit message

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
-	}
-
 	if ofPortPhys != "" {
+		defaultNetConfig := b.netConfig[types.DefaultNetworkName]
+		// table 0, Ingress/Egress flows for MEG enabled pods and advertised UDNs
+		// priority 300: Ingress traffic to MEG pods and advertised UDNs
+		// priority 301: Ingress traffic to node management traffic
+		// priority 104: Egress traffic from advertised UDNs or MEG enabled pods
+		// priority 103: For egressIP, drop packets coming from pods from other nodes headed externally that were not SNATed.
+		// example flows in SGW mode EIP enabled:
```
tssurya, line 747: its weird when we put the flows wrt to whether EgressIPs are enabled or disabled. aren't the 300 and 301 flows mainly for MEG and BGP? so 300 is always going to be there - either its towards patch port for SGW OR towards LOCAL for LGW. - is there a dependency on whether EIP is enabled or not?

one thing i don't follow - why do we need 301 in SGW mode? which traffic matches on magementPortIP in SGW to send to local? I know that priority 15 was sgw and priority 16 was lgw before surprised to see priority 16 aka 301 in sgw?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+		//   table=0, n_packets=0, n_bytes=0, priority=104,ip,in_port=4,dl_src=02:42:ac:12:00:03,nw_src=<nodeSubnet> actions=output:eth0
+		// example flows in LGW mode EIP disabled:
+		//   table=0, n_packets=0, n_bytes=0, priority=300,ip,in_port=eth0,nw_dst=<nodeSubnet> actions=output:LOCAL
+		//   table=0, n_packets=0, n_bytes=0, priority=104,ip,in_port=LOCAL,dl_src=02:42:ac:12:00:03,nw_src=<nodeSubnet> actions=output:eth0
 		for _, netConfig := range b.patchedNetConfigs() {
 			isNetworkAdvertised := netConfig.Advertised.Load()
 			// disableSNATMultipleGWs only applies to default network
 			disableSNATMultipleGWs := netConfig.IsDefaultNetwork() && config.Gateway.DisableSNATMultipleGWs
+
+			if config.OVNKubernetesFeature.EnableEgressIP {
```
tssurya, line 768: the main change here is that this flow only used to be done for default network and now we are doing it also for UDN networks right? not just for default networks?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+				// Due to the fact that ovn-controllers on different nodes apply the changes independently,
+				// there is a chance that the pod traffic will reach the egress node before it configures the SNAT flows.
+				// Drop pod traffic that is not SNATed
+				for _, clusterEntry := range netConfig.Subnets {
+					cidr := clusterEntry.CIDR
+					ipv := getIPv(cidr)
+					// table 0, drop packets coming from pods headed externally that were not SNATed.
+					dftFlows = append(dftFlows,
+						fmt.Sprintf("cookie=%s, priority=103, in_port=%s, %s, %s_src=%s, actions=drop",
```
tssurya, line 777: why is the priority 104/103 instead of 109/104? - we leave some room on purpose between flows in case in future we need that extra priority gap usually.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+				input = nodetypes.OvsLocalPort
 			}
-			for _, clusterEntry := range netConfig.Subnets {
-				cidr := clusterEntry.CIDR
-				ipv := getIPv(cidr)
+			for _, subnet := range netConfig.NodeSubnets {
+				ipv := getIPv(subnet)
 				dftFlows = append(dftFlows,
-					fmt.Sprintf("cookie=%s, priority=15, table=1, %s, %s_dst=%s, "+
+					fmt.Sprintf("cookie=%s, priority=300, table=0, in_port=%s, %s, %s_dst=%s, "+
```
tssurya, line 801: so this is the main change here right? we were routing packets between UDNs and with this change, we ensure we only match on packets coming from outside. -> at least mentioning the problem we are trying to solve in commit message around this is worth it IMHO - sorry if you have mentioned it somewhere and I missed it.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
 				// but holding this until
 				// https://issues.redhat.com/browse/FDP-646 is fixed, for now we
 				// are assuming MEG & BGP are not used together
 				output = nodetypes.OvsLocalPort
+				input = nodetypes.OvsLocalPort
 			}
-			for _, clusterEntry := range netConfig.Subnets {
-				cidr := clusterEntry.CIDR
-				ipv := getIPv(cidr)
+			for _, subnet := range netConfig.NodeSubnets {
```
tssurya, line 798: for l2 the node subnets is the subnets right?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+						nodetypes.DefaultOpenFlowCookie, ofPortPhys, ipv, ipv, subnet, output))
 				// except node management traffic
-				for _, subnet := range netConfig.NodeSubnets {
-					mgmtIP := util.GetNodeManagementIfAddr(subnet)
-					ipv := getIPv(mgmtIP)
+				mgmtIP := util.GetNodeManagementIfAddr(subnet)
+				if mgmtIP == nil {
+					return nil, fmt.Errorf("unable to determine management IP for subnet %s", subnet.String())
+				}
+				if config.Gateway.Mode != config.GatewayModeLocal {
```
tssurya, line 809: so it looks like instead of adding just priority=16 flows in lgw, we have added the full nodesubnet match and we are reusing the MEG flows for BGP - did I get that right? that way we don't need that specific mgmtIP flows..

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+							nodetypes.DefaultOpenFlowCookie, ofPortPhys, ipv, ipv, mgmtIP.IP, nodetypes.OvsLocalPort),
 					)
 				}
+
+				if (disableSNATMultipleGWs && config.OVNKubernetesFeature.EnableEgressIP) || isNetworkAdvertised {
+					// MEG and advertised UDN networks requires that local pod traffic can leave the node without SNAT.
+					// We match on the pod subnets and forward the traffic to the physical interface.
+					// Select priority 104 for the scenario when both EgressIP and advertised UDN are active:
+					// 1. Override egressIP drop flows (priority 103)
+					// 2. Still allow egressIP flows at priority 105
```
tssurya, line 822: the 105 is the egressIP flows to SNAT to nodeIP if the destination is nodeIPs right?

go-controller/pkg/node/gateway_udn_test.go
```diff
@@ -1143,7 +1143,7 @@ var _ = Describe("UserDefinedNetworkGateway", func() {
 
 			Expect(udnGateway.AddNetwork()).To(Succeed())
 			flowMap = udnGateway.gateway.openflowManager.flowCache
-			Expect(flowMap["DEFAULT"]).To(HaveLen(71))                                      // 18 UDN Flows, 5 advertisedUDN flows, and 2 packet mark flows (IPv4+IPv6) are added by default
+			Expect(flowMap["DEFAULT"]).To(HaveLen(73))                                      // 18 UDN Flows, 5 advertisedUDN flows, and 2 packet mark flows (IPv4+IPv6) are added by default
```
tssurya, line 1146: what are the two extra flows you are adding here that weren't already present before? is it the two flows that are moved from table=1 to table=0? is this unit test only looking at table-0 flows?

docs/features/cluster-egress-controls/egress-ip.md
tssurya, line 1: changes need update to either BGP enhancement or to the BGP user/devel docs for flows.
with this change, the OKEP BGP effectively becomes stale in the flows section: https://ovn-kubernetes.io/okeps/okep-5296-bgp/#ingress-via-ovs-bridge

pliurh: This commit aims to unify the traffic path for MEG-enabled pods and advertised UDN pods, as in both cases, SNAT will be skipped for pod outbound traffic. Since the egress traffic won't be SNATed, we don't need to set the CT mark, so we can handle both the ingress/egress in table 0.

pliurh: > IIUC priority 15 and 16 flows are changing to priority 300 and 301.

Yes, they are also moved from table 1 to table 0. 

> Why did we have to change/touch flows are priority 109 and 104 as part of this commit?

These egress flows used to be created for MEG only. However, I found we also need these flows to allow cross-UDN traffic in the loose mode. And we don't need to set the CT mark for such non-SNATed traffic.

test/e2e/external_gateways.go
pliurh: We no longer set/match CT mark for egress/ingress MEG traffic in the new flows.

docs/features/cluster-egress-controls/egress-ip.md
tssurya: @pliurh ah thanks for explaining, which CT Mark are you talking about - is it the 0x1/0x2 ct mark on the way out? so we moved the flows from table 15 and 16 to table-0? for BGP advertised networks?

so before we used to do:
```
 cookie=0xdeff105, duration=2244.222s, table=0, n_packets=34761, n_bytes=339531916, idle_age=0, priority=50,ip actions=ct(table=1,zone=64000,nat)
and then
 cookie=0xdeff105, duration=2244.222s, table=1, n_packets=0, n_bytes=0, idle_age=2244, priority=100,ct_state=+rel+trk,ct_mark=0x1,ip actions=output:2
or
 cookie=0xdeff105, duration=2244.222s, table=1, n_packets=1410, n_bytes=1905156, idle_age=1, priority=100,ct_state=+est+trk,ct_mark=0x1,ip actions=output:2
```
for outbound for MEG and BGP?

and now we do:

```
		//   table=0, n_packets=0, n_bytes=0, priority=104,ip,in_port=4,dl_src=02:42:ac:12:00:03,nw_src=<nodeSubnet> actions=output:eth0
```
in table 0 and bypass the mark and send bits?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Correct. And no CT mark anymore.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Yes.

test/e2e/external_gateways.go
tssurya: so we don't have an entry in the 64000 zone for MEG correct?

docs/features/cluster-egress-controls/egress-ip.md
tssurya: `However, I found we also need these flows to allow cross-UDN traffic in the loose mode.`

ah this is the bit I can't remember.. can you expand on this more.. in the past we had routing between UDNs happening and we stopped that using the inport match, then after that we had EIP rules dropping the packets is it?
and so you had to bring up the rules for BGP to 300 priority that is higher priority than the drop which adds specific matches for the traffic outbound ?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
tssurya: > I didn't really ask to add the comment sorry, I was just doing a highlight. This is not accurate as far as the reason I had in mind.
> 
> ```
> // But it is not a problem since localnet networks cannot be advertised.
> ```
> 
> So I thought that the advertised default network would have to be on the same subnet as the nodes/localnet for that to work and I don't think we allow that kind of configuration.
> 
> However I just now realized that we are SNATing E/W even for an advertised network so in that case this 104 flow wouldn't even be matched and is not short-circuiting anything.
> 

SNATing E/W for advertised network hmm wydm? pod2pod is not being SNATed right?

docs/features/cluster-egress-controls/egress-ip.md
tssurya: note that this is not just a BGP OKEP problem - but in general a OKEP will go stale problem.
so if we have devel docs for flows we should be fine.

go-controller/pkg/node/gateway_udn_test.go
pliurh: Yes.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: The 105 flows are for EIPs in UDNs.

docs/features/cluster-egress-controls/egress-ip.md
tssurya: @pliurh confirmed offline that my understanding is correct.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+			input := netConfig.OfPortPatch
+			isAdvertisedLGW := isNetworkAdvertised && config.Gateway.Mode == config.GatewayModeLocal
+			if isAdvertisedLGW {
 				// except if advertised through BGP, go to kernel
 				// TODO: MEG enabled pods should still go through the patch port
 				// but holding this until
 				// https://issues.redhat.com/browse/FDP-646 is fixed, for now we
 				// are assuming MEG & BGP are not used together
 				output = nodetypes.OvsLocalPort
+				input = nodetypes.OvsLocalPort
```
tssurya, line 796: what if default network is advertised.. and its lgw? then do you change the input port to LOCAL here also for the MEG flow? because in that case the inport should still be patchport since MEG in LGW also behaves like SGW

go-controller/pkg/node/bridgeconfig/bridgeflows.go
tssurya: so before only MEG was using 109 flows for outbound:
```
 cookie=0xdeff105, duration=3469.562s, table=0, n_packets=0, n_bytes=0, idle_age=3469, priority=109,ip,in_port=2,dl_src=de:92:30:6b:e5:bb,nw_src=10.244.0.0/24 actions=ct(commit,zone=64000,exec(load:0x1->NXM_NX_CT_MARK[])),output:1
```
now BGP also uses the same flow. we stop conntrack mark and stop sending it to commit into 64000 because its not needed since we never SNAT this traffic. Moreever this traffic works the same way whether its for lgw or sgw and we always want MEG traffic to back to patch port so we don't really need a clear distinction between 0x1 and 0x2.
so we changed this to also be for BGP.

But we have 105 EIP flow to SNAT to nodeIP if dstIP=nodeIP and now we won't hit that if we keep the flow at 109 for EIP+BGP case. so we lowered flow from 109 to 104. and in turn the 104 became 103.

I assume direct podIP ingress w/o MEG/BGP is not supported so no need to check support for that here I guess?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
tssurya: so before this change the MEG flow for default network was:
```
 cookie=0xdeff105, duration=43.928s, table=0, n_packets=0, n_bytes=0, idle_age=43, priority=109,ip,in_port=2,dl_src=32:2d:b7:8b:e5:34,nw_src=10.244.2.0/24 actions=ct(commit,zone=64000,exec(load:0x1->NXM_NX_CT_MARK[])),output:1
```
and then we had BGP flows for advertised networks.

after this change, it became:
```
 cookie=0xdeff105, duration=257.815s, table=0, n_packets=0, n_bytes=0, priority=104,ip,in_port=LOCAL,dl_src=36:69:56:ae:51:25,nw_src=10.244.0.0/24 actions=output:eth0
```
so that means MEG and BGP together being used is not something we support right? 
the patchport is always br-int-default-network patch port for MEG.. and it becomes LOCAL now for BGP use case.

so is my understanding correct that on advertised default network MEG is not supported?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
tssurya: its not specific to UDNs, but yes we want it for both UDNs and default.

docs/features/cluster-egress-controls/egress-ip.md
pliurh: Yes. You're right.

test/e2e/external_gateways.go
pliurh: Correct.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: We don't support MEG with the advertised default network.

docs/features/cluster-egress-controls/egress-ip.md
pliurh: We shall add this to the BGP developer doc.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Yes.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Yeah, you're right.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: > is there a dependency on whether EIP is enabled or not?

The 103 flow is to drop packets from pods in other nodes. This can only happen when EIP is enabled.

> why do we need 301 in SGW mode?

The original 16 flow was added by https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4533. This PR just keeps the original logic. However, this flow might be a mistake. It may break the EIP healthcheck. We shall address it in a different PR.

docs/features/cluster-egress-controls/egress-ip.md
pliurh: As we don't need set/match CT mark anymore. We set 300 priority flow to ensure the ingress traffic can be matched as early as possible in table 0 for ingress. The 109 and 104 flows used to be only for MEG egress. However, in the new design, we want this flow for advertised networks. And we want EIP can work with advertised networks. So we need to change the priority from 109 to 104 which is less than the 105 flows for EIP. Therefore, the 109 and 104 flows are changed to 104 and 103 respectively.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
tssurya: @jcaamano has opened a FUP issue for us to track this. resolving.

pliurh:
Tested LGW with RA=true and MEG=true manually, traffic works for non-MEG and MEG pods.

tssurya:
> Tested LGW with RA=true and MEG=true manually, traffic works for non-MEG and MEG pods.

thank you @pliurh !

tssurya:
LGTM for the flows commit - i.e 5th commit.
deferring to @jcaamano for approval since he has reviewed all commits.

go-controller/pkg/ovn/udn_isolation.go
```diff
@@ -371,38 +371,45 @@ func (bnc *BaseNetworkController) addAdvertisedNetworkIsolation(nodeName string)
...
-	passACLPredicate := libovsdbops.GetPredicate[*nbdb.ACL](passACLIDs, nil)
-	passACLs, err := libovsdbops.FindACLsWithPredicate(bnc.nbClient, passACLPredicate)
-	if err != nil {
-		return fmt.Errorf("unable to find the pass ACL for advertised network %s: %w", bnc.GetNetworkName(), err)
+	dropACLIDs := GetAdvertisedNetworkSubnetsDropACLdbIDs()
+	// Create a combined predicate to find both ACLs in a single lookup
+	combinedACLPredicate := func(acl *nbdb.ACL) bool {
+		// Check if ACL matches either pass or drop ACL IDs
+		passMatch := libovsdbops.GetPredicate[*nbdb.ACL](passACLIDs, nil)(acl)
+		dropMatch := libovsdbops.GetPredicate[*nbdb.ACL](dropACLIDs, nil)(acl)
```
jcaamano: Can we take these two lines out of the predicate?

go-controller/pkg/ovn/udn_isolation.go
```diff
@@ -371,38 +371,45 @@ func (bnc *BaseNetworkController) addAdvertisedNetworkIsolation(nodeName string)
...
+		return passMatch || dropMatch
 	}
 
-	dropACLIDs := GetAdvertisedNetworkSubnetsDropACLdbIDs()
-	dropACLPredicate := libovsdbops.GetPredicate[*nbdb.ACL](dropACLIDs, nil)
-	dropACLs, err := libovsdbops.FindACLsWithPredicate(bnc.nbClient, dropACLPredicate)
+	// Find both ACLs in a single lookup
+	allACLsToRemove, err := libovsdbops.FindACLsWithPredicate(bnc.nbClient, combinedACLPredicate)
```
jcaamano, line 401: @pperiyasamy would you work in a FUP PR to constraint the predicate search to ACLs on the network's switch instead of all ACLs?

jcaamano (commented):
One comment after final review.

@pliurh please review @coderabbitai's "Duplicated comments" to check if there is anyone worth addressing.

jcaamano (approved):

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -737,70 +737,94 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
...
+					// MEG and advertised UDN networks requires that local pod traffic can leave the node without SNAT.
+					// We match on the pod subnets and forward the traffic to the physical interface.
+					// Select priority 104 for the scenario when both EgressIP and advertised UDN are active:
+					// 1. Override egressIP drop flows (priority 103)
+					// 2. Still allow egressIP flows at priority 105
+					dftFlows = append(dftFlows,
+						fmt.Sprintf("cookie=%s, priority=104, in_port=%s, dl_src=%s, %s, %s_src=%s, "+
+							"actions=output:%s",
+							nodetypes.DefaultOpenFlowCookie, input, bridgeMacAddress, ipv, ipv, subnet, ofPortPhys))
+				}
```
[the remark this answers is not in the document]
tssurya: @coderabbitai: what is the earlier consensus you are referring to?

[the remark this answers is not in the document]
tssurya: @coderabbitai: point me to the comment or discussion that made you consider this aspect that 104 flow gating is too restrictive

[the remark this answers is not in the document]
tssurya: no you are right to question that match condition. MEG and EgressIPs are two different features, there is no need to gate a MEG flow based on whether EgressIP is enabled OR not.
@coderabbitai: can you please open a followup issue to address this?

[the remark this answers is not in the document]
tssurya: @coderabbitai: please assign that followup issue to @pliurh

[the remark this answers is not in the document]
tssurya: @coderabbitai: please link the issue to this thread for history once you have opened it.

PR #5278, opened 2025-06-02
Some quality of life improvements for layer 3 controllers node handling

trozet:
During update node events, local and remote addOrUpdate functions are called. There are a series of sync checks used to know what to configure. However, in some cases log messages were being printed no matter what, and hybrid overlay was being processed on every node event.

This cleans things up so that hybrid overlay is only sync'ed when necessary, and logs are only printed when work is being done to add the local or remote node.

Also adds a couple commits to optimize zone ic handling.

### DIFF
--- a/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
+++ b/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
@@ -13,6 +13,7 @@ import (
 	utilnet "k8s.io/utils/net"
 
 	libovsdbclient "github.com/ovn-org/libovsdb/client"
+	"github.com/ovn-org/libovsdb/ovsdb"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
 	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
@@ -568,6 +569,7 @@ func (zic *ZoneInterconnectHandler) cleanupNodeTransitSwitchPort(nodeName string
 // ip4.dst == 10.244.0.0/24 , nexthop = 100.88.0.2
 // ip4.dst == 100.64.0.2/16 , nexthop = 100.88.0.2  (only for default primary network)
 func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node, nodeTransitSwitchPortIPs, nodeSubnets, nodeGRPIPs []*net.IPNet) error {
+	ops := make([]ovsdb.Operation, 0, 2)
 	addRoute := func(prefix, nexthop string) error {
 		logicalRouterStaticRoute := nbdb.LogicalRouterStaticRoute{
 			ExternalIDs: map[string]string{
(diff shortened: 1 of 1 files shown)


trozet:
@kyrtapz fyi

trozet:
last commit is not right, i'll fix it in the morning

go-controller/pkg/ovn/hybrid.go
```diff
@@ -137,12 +137,18 @@ func (oc *DefaultNetworkController) handleHybridOverlayPort(node *corev1.Node, a
 }
 
 func (oc *DefaultNetworkController) deleteHybridOverlayPort(node *corev1.Node) error {
-	klog.Infof("Removing node %s hybrid overlay port", node.Name)
 	portName := util.GetHybridOverlayPortName(node.Name)
 	lsp := nbdb.LogicalSwitchPort{Name: portName}
-	sw := nbdb.LogicalSwitch{Name: oc.GetNetworkScopedSwitchName(node.Name)}
-	if err := libovsdbops.DeleteLogicalSwitchPorts(oc.nbClient, &sw, &lsp); err != nil {
-		return err
+	if _, err := libovsdbops.GetLogicalSwitchPort(oc.nbClient, &lsp); err != nil {
```
kyrtapz, line 142: Is it really faster when doing a get+delete compared to a delete when the element doesn't exist?
I am guessing it will be a little slower if the element does exist.

go-controller/pkg/ovn/master.go
```diff
@@ -509,12 +523,10 @@ func (oc *DefaultNetworkController) addUpdateLocalNodeEvent(node *corev1.Node, n
 	}
 
 	// since the nodeSync objects are created knowing if hybridOverlay is enabled this should work
-	if nSyncs.syncHo {
+	if nSyncs.syncHo && config.HybridOverlay.Enabled {
```
kyrtapz, line 526: Is it possible for `syncHo` to be `true` if `config.HybridOverlay.Enabled` is not?
Also, why is it necessary to have two if blocks for `nSyncs.syncHo`?

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -218,7 +218,7 @@ func (zic *ZoneInterconnectHandler) AddRemoteZoneNode(node *corev1.Node) error {
 	if err := zic.createRemoteZoneNodeResources(node, nodeID); err != nil {
 		return fmt.Errorf("creating interconnect resources for remote zone node %s for the network %s failed : err - %w", node.Name, zic.GetNetworkName(), err)
 	}
-	klog.Infof("Creating Interconnect resources for node %v took: %s", node.Name, time.Since(start))
+	klog.Infof("Creating Interconnect resources for node %q on network %q took: %s", node.Name, zic.GetNetworkName(), time.Since(start))
```
kyrtapz, line 264: This wasn't changed in this PR but do you think we need to print how much time it took to create IC resources for a remote node at normal level? 
Considering we already have `Creating interconnect resources for remote zone node %s for the network %s` I wonder if we should move this one to v5.

go-controller/pkg/ovn/master.go
```diff
@@ -470,6 +470,16 @@ type nodeSyncs struct {
 	syncReroute           bool
 }
 
+func nodeNeedsSync(syncs *nodeSyncs) bool {
```
kyrtapz, line 473: NIT: maybe it would look better as a receiver function on `nodeSyncs`? something like `nSyncs.NeedsSync()`

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -207,15 +207,57 @@ func (zic *ZoneInterconnectHandler) AddLocalZoneNode(node *corev1.Node) error {
...
-	klog.Infof("Creating interconnect resources for remote zone node %s for the network %s", node.Name, zic.GetNetworkName())
 
 	nodeID := util.GetNodeID(node)
 	if nodeID == -1 {
 		// Don't consider this node as cluster-manager has not allocated node id yet.
 		return fmt.Errorf("failed to get node id for node - %s", node.Name)
 	}
 
-	if err := zic.createRemoteZoneNodeResources(node, nodeID); err != nil {
+	nodeSubnets, err := util.ParseNodeHostSubnetAnnotation(node, zic.GetNetworkName())
```
kyrtapz, line 218: Would it be possible to check OVN to see if this is already configured to avoid parsing of the annotations?
The risk is DB/cache access time nad annotations changing of course(not sure if we support that).

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -568,6 +569,7 @@ func (zic *ZoneInterconnectHandler) cleanupNodeTransitSwitchPort(nodeName string
 // ip4.dst == 10.244.0.0/24 , nexthop = 100.88.0.2
 // ip4.dst == 100.64.0.2/16 , nexthop = 100.88.0.2  (only for default primary network)
 func (zic *ZoneInterconnectHandler) addRemoteNodeStaticRoutes(node *corev1.Node, nodeTransitSwitchPortIPs, nodeSubnets, nodeGRPIPs []*net.IPNet) error {
+	ops := make([]ovsdb.Operation, 0, 2)
```
kyrtapz, line 572: NIT: the capacity here won't be correct in dual stack, it doesn't matter much though.

go-controller/pkg/ovn/hybrid.go
trozet: i think you are right, but the key part is where is the work being done? I think the txn lock is precious for nbdb for write, so read is much cheaper on our cache (multiple simultaneous access, no interaction with nbdb). We want to avoid sending a txn to nbdb if we dont have to.

go-controller/pkg/ovn/master.go
trozet: yeah this was tricky. If you look at the resource handler:
```
				hoNeedsCleanup := false
				if !config.HybridOverlay.Enabled {
					// check if the node has the stale annotations on it to signal that we need to clean up
					if _, exists := newNode.Annotations[hotypes.HybridOverlayDRIP]; exists {
						hoNeedsCleanup = true
					}
					if _, exist := newNode.Annotations[hotypes.HybridOverlayDRMAC]; exist {
						hoNeedsCleanup = true
					}
				}
...
syncHo: hoSync || hoNeedsCleanup,
```

if the node was hybrid overlay but then it restarts and its disabled, we need to clean up. This particular block fo code allocates the DRIP, and should only be executed if hybrid overlay is enabled. The second block should execute no matter what and the logic there determines if it should add or clean up stuff.

trozet: note this was the same order the code was in before, perhaps it could be optimized...dont really want to touch it in this PR

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
trozet: probably a good idea, or eliminate the other log

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
trozet: but the remote node wont exist until we create it with this path right?

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
trozet: ok if i push another revision i'll bump it to 4

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
kyrtapz: I suggested looking the entry in the DB to check if it is already configured.

go-controller/pkg/ovn/master.go
kyrtapz: I am ok to keep it as is for now.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
trozet: we do that i the resource handler today to only trigger these functions based on certain annotation changes, so by the time we get here we should know something changed and we need to parse. Albeit some of these operations are lumped together and we could make it more granular to eliminate additional work that might not be required. I'll look at that as a follow up.

PR #5284, opened 2025-06-06
Add the IP rule for a UDN only when it is advertised to the default VRF

pliurh:
## 📑 Description
When an UDN is advertised to a non default VRF, we shall not add the ip rule to the default VRF. Otherwise if another UDN is advertised to the default VRF with the same subnet. The ingress traffic intended for the second UDN cannot be correctly routed to its respective VRF.

## ✅ Checks

- [ ] My code requires tests

## How to verify it

1. Create UDN-A. Configure VRF-lite for a UDN-A
2. Create UDN-B with the same subnet as UDN-A. Make sure the node-subets from both UDNs are the same for a node.
3. Advertise UDN-B to the default VRF.
4. Create pod-a and pod-b connect to UDN-A and UDN-B respectively. Make sure pod-a and pod-b are scheduled to the same node.
5. Try to access an external host from pod-b in UDN-B.

### DIFF
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -89,6 +89,10 @@ type UserDefinedNetworkGateway struct {
 
 	// gwInterfaceIndex holds the link index of gateway interface
 	gwInterfaceIndex int
+
+	// save BGP state at the start of reconciliation loop run to handle it consistently throughout the run
+	isNetworkAdvertisedToDefaultVRF bool
+	isNetworkAdvertised             bool
 }
 
 // UTILS Needed for UDN (also leveraged for default netInfo) in bridgeConfiguration
@@ -366,18 +370,18 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 		return fmt.Errorf("could not add VRF %s routes for network %s, err: %v", vrfDeviceName, udng.GetNetworkName(), err)
 	}
 
-	isNetworkAdvertised := util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)
+	udng.updateAdvertisementStatus()
 
 	// create the iprules for this network
-	if err = udng.updateUDNVRFIPRules(isNetworkAdvertised); err != nil {
+	if err = udng.updateUDNVRFIPRules(); err != nil {
 		return fmt.Errorf("failed to update IP rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
-	if err = udng.updateAdvertisedUDNIsolationRules(isNetworkAdvertised); err != nil {
+	if err = udng.updateAdvertisedUDNIsolationRules(); err != nil {
 		return fmt.Errorf("failed to update isolation rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
-	if err := udng.updateUDNVRFIPRoute(isNetworkAdvertised); err != nil {
+	if err := udng.updateUDNVRFIPRoute(); err != nil {
 		return fmt.Errorf("failed to update ip routes for network %s: %w", udng.GetNetworkName(), err)
 	}
 
@@ -455,18 +459,16 @@ func (udng *UserDefinedNetworkGateway) DelNetwork() error {
 		}
 	}
 
-	if util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name) {
-		err := udng.updateAdvertisedUDNIsolationRules(false)
-		if err != nil {
-			return fmt.Errorf("failed to remove advertised UDN isolation rules for network %s: %w", udng.GetNetworkName(), err)
-		}
+	err := udng.deleteAdvertisedUDNIsolationRules()
+	if err != nil {
+		return fmt.Errorf("failed to remove advertised UDN isolation rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
 	if err := udng.delMarkChain(); err != nil {
 		return err
 	}
 	// delete the management port interface for this network
-	err := udng.deleteUDNManagementPort()
+	err = udng.deleteUDNManagementPort()
 	if err != nil {
 		return err
 	}
@@ -622,8 +624,7 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link) 
 
 	// Route2: Add default route: default via 172.18.0.1 dev breth0 mtu 1400
 	// necessary for UDN CNI and host-networked pods default traffic to go to node's gatewayIP
-	isNetworkAdvertised := util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)
-	defaultRoute, err := udng.getDefaultRoute(isNetworkAdvertised)
+	defaultRoute, err := udng.getDefaultRouteWithAdvertisedCheck()
 	if err != nil {
 		return nil, fmt.Errorf("unable to add default route for network %s, err: %v", udng.GetNetworkName(), err)
 	}
@@ -724,15 +725,7 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link) 
 	return retVal, nil
 }
 
-func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool) ([]netlink.Route, error) {
-	vrfs := udng.GetPodNetworkAdvertisedOnNodeVRFs(udng.node.Name)
-	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
-	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
-	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
-	if isNetworkAdvertised && !slices.Contains(vrfs, types.DefaultNetworkName) {
-		return nil, nil
-	}
-
+func (udng *UserDefinedNetworkGateway) getDefaultRoute() ([]netlink.Route, error) {
 	networkMTU := udng.NetInfo.MTU()
 	if networkMTU == 0 {
 		networkMTU = config.Default.MTU
@@ -755,6 +748,16 @@ func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool)
 		})
 	}
 	return retVal, nil
+}
+
+func (udng *UserDefinedNetworkGateway) getDefaultRouteWithAdvertisedCheck() ([]netlink.Route, error) {
+	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
+	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
+	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
+	if udng.isNetworkAdvertised && !udng.isNetworkAdvertisedToDefaultVRF {
+		return nil, nil
+	}
+	return udng.getDefaultRoute()
 }
 
 // getV4MasqueradeIP returns the V4 management port masqueradeIP for this network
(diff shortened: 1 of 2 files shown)


go-controller/pkg/node/gateway_udn.go
```diff
@@ -984,21 +983,19 @@ func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRules(isNetworkAdvertised b
...
-				util.GetNetworkVRFName(udng.NetInfo), udng.GetNetworkName(), err)
-		}
+	}
+	defaultRoute, err := udng.getDefaultRoute(isNetworkAdvertised)
+	if err != nil {
+		return fmt.Errorf("unable to get default route for network %s, err: %v", udng.GetNetworkName(), err)
+	}
+	if err = udng.vrfManager.AddVRFRoutes(util.GetNetworkVRFName(udng.NetInfo), defaultRoute); err != nil {
+		return fmt.Errorf("error while adding default route to VRF %s corresponding to network %s, err: %v",
+			util.GetNetworkVRFName(udng.NetInfo), udng.GetNetworkName(), err)
```
jcaamano: Before we had an `if/else` bit not now. If `removeDefaultRouteFromVRF` doesn't fail, wouldn't you be adding the default route back?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -827,7 +826,7 @@ func (udng *UserDefinedNetworkGateway) constructUDNVRFIPRules(isNetworkAdvertise
 		}
 	}
 	switch {
-	case !isNetworkAdvertised:
+	case !isNetworkAdvertisedToDefaultVRF:
```
jcaamano: Making sure we are on the same page. You decided to add the masquerade ip rule, but that wouldn't be strictly needed. This could be a tri-state:
```
case isNetworkAdvertised && !isNetworkAdvertisedToDefaultVRF:
   // delete masq and subnet ip rule
case isNetworkAdvertised:
  // delete masq ip rule, add subnet ip rule
default:
  // delete subnet ip rule, add masq ip rule
```
Also note my shift to remove negations

jcaamano (commented):
A bit surprised to see this does not affect any existing unit test.

go-controller/pkg/node/gateway_udn.go
pliurh: You're right.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -725,11 +725,10 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link)
 }
 
 func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool) ([]netlink.Route, error) {
-	vrfs := udng.GetPodNetworkAdvertisedOnNodeVRFs(udng.node.Name)
 	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
 	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
 	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
-	if isNetworkAdvertised && !slices.Contains(vrfs, types.DefaultNetworkName) {
+	if isNetworkAdvertised && !udng.isNetworkAdvertisedToDefaultVRF() {
```
jcaamano: There is, throughout the methods of this file, a mix of:
* passing only `isNetworkAdvertised`
* passing both `isNetworkAdvertised` and `isNetworkAdvertisedToDefaultVRF`
* passing nothing and directly querying NetInfo in a utility method

There's two factors: 
* The dynamic information of NetInfo is live, it can change at any time so two consecutive calls might yield different results. Eventually everything will be reconciled to a consistent state since the reconciliation loop is triggered every time the information changes. However, it would still be nice to remain consistent within a given loop run, if at all possible and where reasonable.
* There's also the general consistency of the code in this file

So due to the previous considerations, I think I would try to follow a different pattern. One idea is to get at the top level of the reconciliation loop (doReconcile/AddNetwork) the list of VRFs we are advertising on (GetPodNetworkAdvertisedOnNodeVRFs) and pass that around, instead of the inconsistently passing none, one or two boolean.

If we were to not care and take the opposite approach, we would just query NetInfo directly when needed instead of passing any booleans around since the information is always readily available on methods of UserDefinedNetworkGateway.

Current state of affairs just seem to be inconsistent at least in the code level.

I don't think this is an issue you introduced but gets exacerbated by the introduction of a new boolean.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -827,12 +829,18 @@ func (udng *UserDefinedNetworkGateway) constructUDNVRFIPRules(isNetworkAdvertise
...
-	default:
-		addIPRules = append(addIPRules, subnetIPRules...)
+	case isNetworkAdvertisedToDefaultVRF:
+		// the network is advertised to the default VRF
 		delIPRules = append(delIPRules, masqIPRules...)
+		addIPRules = append(addIPRules, subnetIPRules...)
+	default:
+		// the network is not advertised
+		delIPRules = append(delIPRules, subnetIPRules...)
+		addIPRules = append(addIPRules, masqIPRules...)
```
jcaamano: nit: so this would be maybe more straight forward
```
case isNetworkAdvertisedToDefaultVRF:
  // network is advertised on the default VRF
  // add subnet ip rule, delete masq ip rule
case isNetworkAdvertised:
  // network is advertised but not on the default VRF
  // delete subnet ip rule, delete masq ip rule
default:
  // network is not advertised
  // delete subnet ip rule, add masq ip rule
```
sorry if I misguided you before

go-controller/pkg/node/gateway_udn.go
```diff
@@ -984,13 +992,12 @@ func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRules(isNetworkAdvertised b
 // Add or remove default route from a vrf device based on the network is
 // advertised on its own network or default network
 func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRoute(isNetworkAdvertised bool) error {
-	vrfs := udng.GetPodNetworkAdvertisedOnNodeVRFs(udng.node.Name)
-	if isNetworkAdvertised && !slices.Contains(vrfs, types.DefaultNetworkName) {
+	if isNetworkAdvertised && !udng.isNetworkAdvertisedToDefaultVRF() {
```
jcaamano: nit: if you agree, switch/case patterns read better for me. Otherwise ignore.

go-controller/pkg/node/gateway_udn.go
pliurh: fixed.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -89,6 +89,9 @@ type UserDefinedNetworkGateway struct {
 
 	// gwInterfaceIndex holds the link index of gateway interface
 	gwInterfaceIndex int
+
+	isNetworkAdvertisedToDefaultVRF bool
```
jcaamano, line 94: Can you add a comment here? Something like
```
save BGP state at the start of reconciliation loop run to handle it consistently throughout the run
```

go-controller/pkg/node/gateway_udn.go
```diff
@@ -725,11 +727,10 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link)
 }
 
 func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool) ([]netlink.Route, error) {
-	vrfs := udng.GetPodNetworkAdvertisedOnNodeVRFs(udng.node.Name)
 	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
 	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
 	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
-	if isNetworkAdvertised && !slices.Contains(vrfs, types.DefaultNetworkName) {
+	if isNetworkAdvertised && !udng.isNetworkAdvertisedToDefaultVRF {
```
jcaamano: use `udng.isNetworkAdvertised` and don't pass it as parameter?

pliurh: [Here](https://github.com/ovn-org/ovn-kubernetes/blob/6a01b3a9fd0d3c05314dbc2b754eea68d6c4764a/go-controller/pkg/node/gateway_udn.go#L1008), `getDefaultRoute` is invoked with a constant value. So we cannot use `udng.isNetworkAdvertised` in this function.

jcaamano: I would just have a `getDeefaultRoute()` method that checks no booleans, and then a `getDefaultRouteIfNotAdvertisedOnDefaultVRF()` that cheks the booleans and calls the former to be used in the single place we actually need it.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -757,6 +752,16 @@ func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool)
 	return retVal, nil
 }
 
+func (udng *UserDefinedNetworkGateway) getDefaultRouteWithAdvertisedCheck() ([]netlink.Route, error) {
```
jcaamano, line 753: lol 
I liked my method name suggestion better because it was more clear on what it does and it is an un-exported method anyway.
I guess you didn't. 
I will learn to accept this is subjective so let's go with your option.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -456,7 +460,7 @@ func (udng *UserDefinedNetworkGateway) DelNetwork() error {
 	}
 
 	if util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name) {
-		err := udng.updateAdvertisedUDNIsolationRules(false)
+		err := udng.deleteAdvertisedUDNIsolationRules()
```
jcaamano: OK, this opens a new can of worms :P Bear with me.

I think this delete call should be unconditional and ensure there's no network leftover.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -1034,37 +1059,15 @@ func (udng *UserDefinedNetworkGateway) removeDefaultRouteFromVRF() error {
...
-				tx.Delete(elem)
-			}
-		}
-
-		if tx.NumOperations() == 0 {
-			return nil
-		}
-		return nft.Run(context.TODO(), tx)
+	if !udng.isNetworkAdvertised {
+		return udng.doDeleteAdvertisedUDNIsolationRules(nft, tx)
```
jcaamano: This setup works but I don't think it adds value and retracts from code clarity. Why not have separate add/delete methods each with its own NFT transaction and update method would just  
```
switch {
case udng.isNetworkAdvertised:
    addAdvertisedUDNIsolationNFTRules
default:
    deleteAdvertisedUDNIsolationNFTRules
}
```

jcaamano (commented):
I am getting into subjective territory. Feel free to reply you don't agree if that's the case. It's just that some stuff looks strange to my eyes.

go-controller/pkg/node/gateway_udn.go
pliurh: I think we should keep the condition check, but replace `util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)` with `udng.isNetworkAdvertised`. Since the isolation rules are added only when the network is advertised.

jcaamano: I will leave it up to you, but `deleteAdvertisedUDNIsolationRules` is (and should be) idempotent so unconditional call is definitely the safest approach if we are actually deleting the network.

jcaamano (approved):

trozet:
@tssurya @jcaamano seems like this PR should have never existed in the first place as 2 UDNs with overlapping subnets should not be allowed to be advertised to the same VRF.

jcaamano:
> @tssurya @jcaamano seems like this PR should have never existed in the first place as 2 UDNs with overlapping subnets should not be allowed to be advertised to the same VRF.

This PR addresses overlapping subnets **not** advertised on the same VRF.

PR #5287, opened 2025-06-08
workflow: Add fix missing and apt update before trying to install VRF…

aserdean:
try to fix CI


.github/workflows/test.yml
```diff
@@ -200,6 +200,7 @@ jobs:
       if: steps.is_pr_image_build_needed.outputs.PR_IMAGE_RESTORED != 'true' && success()
       run: |
         set -x
+        sudo apt update
         sudo apt-get install linux-modules-extra-$(uname -r) -y
```
tssurya, line 204: thanks @aserdean !

PR #4286, opened 2024-04-17
Refactor egress IP healthcheck

jcaamano:
This PR refactors the egress IP health check with the following objectives:

- have the same runtime health check for egress ip and egress service
- remove the hardcoded 5 second health check period and 2 minimum
  attempts that was making the quickest possible reaction time to range
  from 7s to 12s without providing finer control
- health check each node in separate goroutines, preventing timeouts
  on a node's health check to delay other node's health check
- TENTATIVE: have different unavailability tiers, so that an ovnk rollout
   does not cause unnecessary service disruption

Adds a new centralized controller in charge of health checks. Listens to
node events to cancel health check of deleted nodes or to become aware of
node configuration changes that impact the health check function.

Provides an API for parties interested in node health information,
like egress IP or egress service controllers. These parties will use this
API to express and upkeep the interest on the health information of
specific nodes, which will serve as an additional input to the health
check controller reconciliation loop. The API itself is level driven:
interested parties will get notified of health state changes but not of
the details, interested parties are expected to query the controller for
those details on their reconciliation loops.

The health check itself is done on a separate goroutine for each node.
The health check period is hardcoded to 1 second instead of the previous
5 seconds. The health check probe timeout is no longer governed by the
timeout configuration parameter and is also hardcoded to 1 second which
was the previous default. There is no hardcoded number of minimum
consecutive failed attempts to declare a node unreachable: the timeout
configuration parameter now measures the amount of time that needs to
pass since the last time a node was observed available before declaring
a node unreachable situating the range where this happens at (value, value+1).

Changes the default timeout configuration parameter value to 10s. With the
new implementation, an unreachable node is actually declared unreachable
after 10s-11s of being last observed available, which is similar to the previous
behavior where that range had minimum and default of 7s-12s. Of course, now
users have the option of changing this configuration down to 1, making this
range 1s-2s.

This is up to commit `Change EgressIPReachabiltyTotalTimeout to 10s`

After that, a compromise exercise is done to attempt to avoid disruption on ovnk
rollouts. A new UNAVAILABLE (vs UNREACHABLE) health state is introduced that
expresses a situation where the node is alive with an apparently functioning OVN dataplane
(given that the health check itself is done through the management port) even though
ovn-k is not running (similar interpretation we had with the legacy health check), where
other controllers might not want to unschedule their services to avoid unnecessary
disruptions. As mentioned, this is a compromise were we try to improve expected
scenarios potentially worsening the effect of other unexpected scenarios.
[...]

PR #5292, opened 2025-06-09
Gate ovn-controller start until ovnkube controller syncs following a reboot

martinkennelly:
For IC mode, after a Node reboot, SB DB may contain stale control plane data because ovnkube-controller has not config'd OVN and changes propagated to SB DB. The issue we see downstream is that when an egress Node which hosts an EgressIP IP, and when it reboots, ovn-controller will connect to SB DB which will contain stale SNAT to support EgressIP. ovn-controller will GARP for the EgressIP IP even though the EgressIP IP may have already been assigned to another Node. 

The issue may also be present for a simple pod restart if the EIP moves while ovnkube-controller is restarting. This PR wont fix that scenario.

For restart scenario, gate ovn-controller until ovnkube controller has syncd and the changes have reached SB DB for IC mode only. Its up to an external entity to clear the ovn-remote value before ovnkube controller starts for the reboot scenario because the ovn-remote is not cleared when ovnkube exists because we dont want to disrupt traffic for simple ovnkube pod restart case.

Other options were considered including wiping the OVN DBs on startup before ovnkube-node pod starts but we may wipe out dynamic info like learned MACs. Adding non-reboot persistant storage was also considered but was overly complex.

Also, we do not want to gate starting ovn-controller for a simple ovnkube pod restart because it would affect the data plane whereas post reboot, there will not be any active workloads.

Downstream bug: https://issues.redhat.com/browse/OCPBUGS-42303

### DIFF
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -515,6 +515,15 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 				return
 			}
 
+			// SB DB maybe stale if IC is enabled and ovn-controller will only be allowed to connect here.
+			if config.OVNKubernetesFeature.EnableInterconnect && config.OvnKubeNode.Mode != types.NodeModeDPUHost {
+				for _, auth := range []config.OvnAuthConfig{config.OvnNorth, config.OvnSouth} {
+					if err := auth.SetDBAuth(); err != nil {
+						controllerErr = fmt.Errorf("unable to set the authentication towards OVN local dbs")
+					}
+				}
+			}
+
 			// record delay until ready
 			metrics.MetricOVNKubeControllerReadyDuration.Set(time.Since(startTime).Seconds())
 
--- a/go-controller/cmd/ovnkube-identity/ovnkubeidentity.go
+++ b/go-controller/cmd/ovnkube-identity/ovnkubeidentity.go
@@ -329,7 +329,7 @@ func runWebhook(ctx context.Context, restCfg *rest.Config) error {
 	nodeWebhook := admission.WithCustomValidator(
 		scheme.Scheme,
 		&corev1.Node{},
-		ovnwebhook.NewNodeAdmissionWebhook(cliCfg.enableInterconnect, cliCfg.enableHybridOverlay, cliCfg.extraAllowedUsers.Value()...),
+		ovnwebhook.NewNodeAdmissionWebhook(cliCfg.enableHybridOverlay, cliCfg.extraAllowedUsers.Value()...),
 	).WithRecoverPanic(true)
 
 	nodeHandler, err := admission.StandaloneWebhook(
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -869,15 +869,6 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
 			return fmt.Errorf("timed out waiting for the node zone %s to match the OVN Southbound db zone, err: %v, err1: %v", config.Default.Zone, err, err1)
 		}
 
-		// if its nonIC OR IC=true and if its phase1 OR if its IC to IC upgrades
-		if !config.OVNKubernetesFeature.EnableInterconnect || sbZone == types.OvnDefaultZone || util.HasNodeMigratedZone(node) { // if its nonIC or if its phase1
-			for _, auth := range []config.OvnAuthConfig{config.OvnNorth, config.OvnSouth} {
-				if err := auth.SetDBAuth(); err != nil {
-					return err
-				}
-			}
-		}
-
 		err = setupOVNNode(node)
 		if err != nil {
 			return err
@@ -953,8 +944,9 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
 		return fmt.Errorf("failed to set node %s annotations: %w", nc.name, err)
 	}
 
-	// Connect ovn-controller to SBDB
-	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
+	// If IC disabled, connect ovn-controller to SB/NB DB. If IC, OVNKube-controller will start it post sync to ensure
+	// ovn-controller doesn't consume stale SB DB data.
+	if !config.OVNKubernetesFeature.EnableInterconnect && config.OvnKubeNode.Mode != types.NodeModeDPUHost {
 		for _, auth := range []config.OvnAuthConfig{config.OvnNorth, config.OvnSouth} {
 			if err := auth.SetDBAuth(); err != nil {
 				return fmt.Errorf("unable to set the authentication towards OVN local dbs")
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -745,60 +745,6 @@ func getOVNSBZone() (string, error) {
 	return dbZone, nil
 }
 
-/** HACK BEGIN **/
-// TODO(tssurya): Remove this HACK a few months from now.
-// checkOVNSBNodeLRSR returns true if the logical router static route for the
-// the given nodeSubnet is present in the SBDB
-func checkOVNSBNodeLRSR(nodeSubnet *net.IPNet) bool {
-	var matchv4, matchv6 string
-	v6 := true
-	v4 := true
-	if config.IPv6Mode && utilnet.IsIPv6CIDR(nodeSubnet) {
-		matchv6 = fmt.Sprintf("match=\"reg7 == 0 && ip6.dst == %s\"", nodeSubnet)
-		stdout, stderr, err := util.RunOVNSbctl("--bare", "--columns", "_uuid", "find", "logical_flow", matchv6)
-		klog.Infof("Upgrade Hack: checkOVNSBNodeLRSR for node - %s : match %s : stdout - %s : stderr - %s : err %v",
-			nodeSubnet, matchv6, stdout, stderr, err)
-		v6 = (err == nil && stderr == "" && stdout != "")
-	}
-	if config.IPv4Mode && !utilnet.IsIPv6CIDR(nodeSubnet) {
-		matchv4 = fmt.Sprintf("match=\"reg7 == 0 && ip4.dst == %s\"", nodeSubnet)
-		stdout, stderr, err := util.RunOVNSbctl("--bare", "--columns", "_uuid", "find", "logical_flow", matchv4)
-		klog.Infof("Upgrade Hack: checkOVNSBNodeLRSR for node - %s : match %s : stdout - %s : stderr - %s : err %v",
-			nodeSubnet, matchv4, stdout, stderr, err)
-		v4 = (err == nil && stderr == "" && stdout != "")
-	}
-	return v6 && v4
-}
-
-func fetchLBNames() string {
-	stdout, stderr, err := util.RunOVNSbctl("--bare", "--columns", "name", "find", "Load_Balancer")
-	if err != nil || stderr != "" {
-		klog.Errorf("Upgrade hack: fetchLBNames could not fetch services %v/%v", err, stderr)
-		return stdout // will be empty and we will retry
-	}
-	klog.Infof("Upgrade Hack: fetchLBNames: stdout - %s : stderr - %s : err %v", stdout, stderr, err)
-	return stdout
-}
-
-// lbExists returns true if the OVN load balancer for the corresponding namespace/name
-// was created
-func lbExists(lbNames, namespace, name string) bool {
-	stitchedServiceName := "Service_" + namespace + "/" + name
-	match := strings.Contains(lbNames, stitchedServiceName)
-	klog.Infof("Upgrade Hack: lbExists for service - %s/%s/%s : match - %v",
-		namespace, name, stitchedServiceName, match)
-	return match
-}
-
-func portExists(namespace, name string) bool {
-	lspName := fmt.Sprintf("logical_port=%s", util.GetLogicalPortName(namespace, name))
-	stdout, stderr, err := util.RunOVNSbctl("--bare", "--columns", "_uuid", "find", "Port_Binding", lspName)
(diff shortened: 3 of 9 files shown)


pperiyasamy:
>Other options were considered including wiping the OVN DBs on startup before ovnkube-node pod starts but we may wipe out dynamic info like learned MACs.

@martinkennelly Does SNAT entry in sbdb cause GARP requests ? If that is the case, did you consider wiping out only EIP SNATs in NBDB at the time of ovnkube-controller shutdown ? Let them get setup after startup. or does it have a pod traffic hit ?

martinkennelly:
> > Other options were considered including wiping the OVN DBs on startup before ovnkube-node pod starts but we may wipe out dynamic info like learned MACs.
> 
> @martinkennelly Does SNAT entry in sbdb cause GARP requests ? If that is the case, did you consider wiping out only EIP SNATs in NBDB at the time of ovnkube-controller shutdown ? Let them get setup after startup. or does it have a pod traffic hit ?

yea, there still could be traffic flowing through the data plate and i wanted to prevent further disruption.
I also did not have any good ideas to detect shut down and pod restart only cases. I dont want to disrupt traffic for simple pod restart case.

martinkennelly:
/hold

.. until verified downstream with reboot and test cases. I will update the test definition later here.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -149,6 +156,73 @@ func delPidfile(pidfile string) {
...
+
+// doesOVNKubeControllerSyncPostBootFileExist checks if a sentinel file exists. File will not exist post Node restart. OVNKube controller creates the file post sync and after a reboot.
+func doesOVNKubeControllerSyncPostBootFileExist() (bool, error) {
+	exists := false
+	_, err := os.Stat(ovnkubeControllerSyncPostBootFilePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return exists, fmt.Errorf("failed to stat file at path %q: %w", ovnkubeControllerSyncPostBootFilePath, err)
+		}
+		exists = true
```
pperiyasamy: this is bit confusing, exists flag is set to true when file doesn't exist, but the method name suggests otherwise.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -149,6 +156,73 @@ func delPidfile(pidfile string) {
...
+	nbGlobal, err := libovsdbops.GetNBGlobal(nbClient, nbGlobal)
+	if err != nil {
+		return fmt.Errorf("failed to find OVN Northbound NB_Global table"+
+			" entry: %v", err)
+	}
+	// increment nb_cfg value by 1. When northd consumes updates from NB DB, it will copy this value to SB DBs SB_Global table nb_cfg field.
+	ops, err := nbClient.Where(nbGlobal).Mutate(nbGlobal, model.Mutation{
+		Field:   &nbGlobal.NbCfg,
+		Mutator: ovsdb.MutateOperationAdd,
+		Value:   1,
```
pperiyasamy: nit; may add comment that no other updates happening for `nbGlobal.NbCfg` field at this time.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -149,6 +156,73 @@ func delPidfile(pidfile string) {
...
+// for northd to copy it to SB DB. Poll SB DB until context is cancelled.
+func waitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Client) error {
+	// 1. Get value of nb_cfg
+	// 2. Increment value of nb_cfg
+	// 3. Wait until value appears in SB DB after northd copies it.
+	nbGlobal := &nbdb.NBGlobal{}
+	nbGlobal, err := libovsdbops.GetNBGlobal(nbClient, nbGlobal)
+	if err != nil {
+		return fmt.Errorf("failed to find OVN Northbound NB_Global table"+
+			" entry: %v", err)
```
pperiyasamy: why not using %w for err ?

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -149,6 +156,73 @@ func delPidfile(pidfile string) {
...
+			// northd hasn't added an entry yet
+			if errors.Is(err, client.ErrNotFound) {
+				return false, nil
+			}
+			return false, fmt.Errorf("failed to get sb_global table entry from SB DB: %w", err)
+		}
+		return sbGlobal.NbCfg >= expectedNbCfgValue, nil // we only need to ensure it is greater than or equal to the expected value
+	})
+	if err != nil {
+		return fmt.Errorf("failed while waiting for nb_cfg value greater than or rqual %d in sb db sb_global table: %w", expectedNbCfgValue, err)
```
pperiyasamy: typo: equal

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -149,6 +156,73 @@ func delPidfile(pidfile string) {
 	}
 }
 
+// ovnkubeControllerSyncPostBootFilePath is located in storage that does not persist following a reboot. ovn-controller start
+// is predicated on the existence of this file.
+const ovnkubeControllerSyncPostBootFilePath = "/tmp/ovnkube-controller-sync-post-boot"
+
+func ensureOVNKubeControllerSyncPostBootFile() error {
+	if err := os.WriteFile(ovnkubeControllerSyncPostBootFilePath, []byte(time.Now().String()), 0o644); err != nil {
```
pperiyasamy: what if root user accidentally deletes the file ?

go-controller/cmd/ovnkube/ovnkube.go
martinkennelly: ah my bad. I renamed the method and didnt change the bool.

go-controller/cmd/ovnkube/ovnkube.go
martinkennelly: its actually fine if theres updates at the same time. May I ask why it wouldnt be fine?

go-controller/cmd/ovnkube/ovnkube.go
martinkennelly: For upstream, if the user deletes the file, and if the ovnkube node pod restarts, things are fine. if only the ovn-controller restarts then it wont start until ovnkube-controller restarts. The scenario can be handled with a process watching the file... but i think its overkill but up for opinions.

Downstream, I plan to put a timeout on the start of ovn-controller.

pperiyasamy: ok, I like the approach of adding timeout for ovn-controller start, why don't we need that in upstream as well. ?

go-controller/cmd/ovnkube/ovnkube.go
pperiyasamy: that may be because check on `sbGlobal.NbCfg >= expectedNbCfgValue` caused by other update.

martinkennelly: nb_cfg is only ever increment by any ovnk code.

We dont care above others incrementing it because when we get the nb_cfg value initally, we know the NB DB is HOT with all the data we want to copy to SB DB, so if we see that value + 1 or greater in SB DB, then we know whatever was in the NB DB for the initial nb_cfg value is in SB DB. Make sense?

go-controller/cmd/ovnkube/ovnkube.go
martinkennelly: i wanted to highlight any func error directly upstream so no timeout. is ok?

martinkennelly:
@pperiyasamy thank you!!

martinkennelly:
/hold


I need to make mount a non-persistent storage from the host.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -149,6 +156,77 @@ func delPidfile(pidfile string) {
 	}
 }
 
+// ovnkubeControllerSyncPostBootFilePath is located in storage that does not persist following a reboot. ovn-controller start
+// is predicated on the existence of this file.
+const ovnkubeControllerSyncPostBootFilePath = "/tmp/ovnkube-controller-sync-post-boot"
```
martinkennelly, line 161: replace with host mounted storage that wont persist during reboot.

martinkennelly:
I went with the solution of gating ovn-controller only on restart to minimise perceived disruption.
I went with the following approach over using "tmp" storage that will be wiped on the host following a reboot because its complicated updating all the manifests for both containers and its also gated on IC. Also, Id have to make it a config flag because maybe we dont want to mount all of tmp into storage so we would have to precreate a directory there that would be mounted. We cannot use the pod storage because I didnt want to gate ovn-controller for simple restart after consulting @trozet  and OVN folks because they had concerns over disruption for new flows.

martinkennelly:
As @jcaamano pointed out, my current solution relies on something downstream currently that clears ovn-remote before ovnkube is started on boot so ovn-controller wont connect to stale SB DB. Upstream theres no concept of rebooting a Node.

pperiyasamy:
> As @jcaamano pointed out, my current solution relies on something downstream currently that clears ovn-remote before ovnkube is started on boot so ovn-controller wont connect to stale SB DB. Upstream theres no concept of rebooting a Node.

+1 for using ovs-configuration systemd service which clears up ovn-remote column from ovsdb. it's a neat solution.

jcaamano:
> As @jcaamano pointed out, my current solution relies on something downstream currently that clears ovn-remote before ovnkube is started on boot so ovn-controller wont connect to stale SB DB. Upstream theres no concept of rebooting a Node.

Yeah, using ovs-configuration means we are not providing a solution upstream. Also not a fan of solving one repo's problems in a different repo. We are actively trying to rely less and less on ovs-configuration and adding things to it does not help. 

I would try to solve this in the pod's manifest. For example, we could write a timestamp before starting ovnkube-controller and another when ovnkube-controller has synchronized with SBDB. This timestamps could just be stat on a file. We delay ovn-controller from starting until either of these two happens:
* the latter timestamp is newer that the former
* a maximum acceptable time has passed

The problem with the manifest solution is that there is no good way to tell apart a node restart vs a pod restart, not aware of any downward API for this. There seems to be interest in handling these two situations differently but I am unaware why. I understand there is some level of disruption that might be perceived differently with a node restart than with a pod restart, however we can gate that with the "maximum acceptable time" mentioned above. Depending on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4286 to handle the pod restart differently is pulling in a lot of things when we might not be ready for them.

Another option is to write a script that cleans the `"nat-addresses": "router"` option from the switch ports. This script can be invoked before ovn-controller is started and should prevent stale GARPs to be sent until OVNK updates the DB.

martinkennelly:
> > As @jcaamano pointed out, my current solution relies on something downstream currently that clears ovn-remote before ovnkube is started on boot so ovn-controller wont connect to stale SB DB. Upstream theres no concept of rebooting a Node.
> 
> Yeah, using ovs-configuration means we are not providing a solution upstream. Also not a fan of solving one repo's problems in a different repo. We are actively trying to rely less and less on ovs-configuration and adding things to it does not help.
> 
> I would try to solve this in the pod's manifest. For example, we could write a timestamp before starting ovnkube-controller and another when ovnkube-controller has synchronized with SBDB. This timestamps could just be stat on a file. We delay ovn-controller from starting until either of these two happens:
> 
> * the latter timestamp is newer that the former
> * a maximum acceptable time has passed
> 

For some CUs, its never acceptable to pass the timeout threshold and GARP.

> The problem with the manifest solution is that there is no good way to tell apart a node restart vs a pod restart, not aware of any downward API for this. There seems to be interest in handling these two situations differently but I am unaware why. I understand there is some level of disruption that might be perceived differently with a node restart than with a pod restart, however we can gate that with the "maximum acceptable time" mentioned above. Depending on #4286 to handle the pod restart differently is pulling in a lot of things when we might not be ready for them.
> 
> Another option is to write a script that cleans the `"nat-addresses": "router"` option from the switch ports. This script can be invoked before ovn-controller is started and should prevent stale GARPs to be sent until OVNK updates the DB.


detecting whether its a pod restart or a node restart is what i wanted to avoid with my current approach in this PR and also with your #4286  healthcheck refactor solves nearly all the garp issues we have.. its still possible to occur when ovnkube-controller is down and the EIP moves during a pod restart even with #4286 but thats future work. #4286 makes it unlikely but something we should test at scale.

martinkennelly:
If we always gate ovn-controller and thats it - it solves all the EIP issues.

Also, by gate, i mean start ovn-controller and dont allow it to download SB DB because our current implementation of ovnkube-controlle requires config of the Node and config ovs - that will fail without ovn-controller running. It maybe refactored but I am not sure.

jcaamano:
> For some CUs, its never acceptable to pass the timeout threshold and GARP.

we can make it configurable.

The option to unset nat_addresses might not be straight forward as that is NBDB. We would need to operate with the equivalent of it in SBDB which might not be a direct mapping. But maybe we can run this idea by the OVN team and it lights another bulb.

jcaamano:
> Also, by gate, i mean start ovn-controller and dont allow it to download SB DB because our current implementation of ovnkube-controlle requires config of the Node and config ovs - that will fail without ovn-controller running. It maybe refactored but I am not sure.

Ah right. So from the manifests, we would just need to gate ovn-controller until SBDB starts and we had had the chance to clean the remote. So on the script that starts ovn-controller, we loop until we can connect to SBDB and clean up the remote. Only after we have done that, we start ovn-controller. That doesn't sound too complicated or at least, not as complicated as requiring an involvem,ent of ovs-configure. 

BTW, are you able to clean up the remote from DB without the DB actually running. If that's the case you could do that with an init container as well.

What I am more concerned with long term. is that we are assuming ovn-controller is able to configure whatever we need it to configure without downloading the SBDB. This can change any time and then what?

It seems like we need an RFE for OVN to handle this better in any case.

jcaamano:
If we are decided to treat pod restart differently than node restart, the we can use an `emptyDir`volume mount with `medium` set to `Memory`.

martinkennelly:
> If we are decided to treat pod restart differently than node restart, the we can use an `emptyDir`volume mount with `medium` set to `Memory`.

I tried that originally with mount defined as the following and attach to containers but memory didnt persist pod delete and recreate. I guess it works for container restart but havent tested.
```yaml
      - name: mem-storage
        emptyDir:
          medium: Memory
```

martinkennelly:
closed in favor of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5373

PR #5295, opened 2025-06-11
EgressIP: fix startup syncer to parse IP instead of CIDR

martinkennelly:
Initial implementations erroneously assumed a CIDR.

This was not caught by unit tests because the unit test also contained the assumption of only logical IP with no mask.

It was not caught by upstream CI because we have no reboot tests.

TBD; figure out how to test it downstream / upstream.


martinkennelly:
I am trying to think how to improve the testing for this. First of all I should have returned an error if logical IP wasnt as expected. That would have caught this issue sooner.  Second, I have QE writing a test downstream to contineously arping during a Node reboot / upgrade secenario and ensure the expected MACs.

martinkennelly:
I made this error because I saw we added the mask in the code here:
```go
func (e *EgressIPController) buildSNATFromEgressIPStatus(ni util.NetInfo, podIP net.IP, status egressipv1.EgressIPStatusItem, egressIPName, podNamespace, podName string) (*nbdb.NAT, error) {
	logicalIP := &net.IPNet{
		IP:   podIP,
		Mask: util.GetIPFullMask(podIP),
	}
	ipFamily := IPFamilyValueV4
	if utilnet.IsIPv6CIDR(logicalIP) {
		ipFamily = IPFamilyValueV6
	}
	externalIP := net.ParseIP(status.EgressIP)
	logicalPort := ni.GetNetworkScopedK8sMgmtIntfName(status.Node)
	externalIds := getEgressIPNATDbIDs(egressIPName, podNamespace, podName, ipFamily, e.controllerName).GetExternalIDs()
	nat := libovsdbops.BuildSNAT(&externalIP, logicalIP, logicalPort, externalIds)
	return nat, nil
}
```
LogicalIP accepts CIDR or IP but for EIP its only ever going to be an IP not a CIDR.

martinkennelly:
CI looks good :)

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
```diff
@@ -101,7 +101,8 @@ func (syncer *LRPSyncer) syncEgressIPReRoutes() error {
 			podInfo, err := cache.getPod(podIP)
 			if err != nil {
 				klog.Infof("Failed to find Logical Switch Port cache entry for pod IP %s: %v", podIP.String(), err)
-				continue
+				// pod not found, add dummy metadata that will be cleaned up by EIP controller sync.
+				podInfo = podNetInfo{namespace: "UNKNOWN", name: "UNKNOWN"}
```
trozet, line 105: so you set these values as UNKNOWN, then that generates the corresponding external ids, then you update the policy with this? I'm not following how this triggers the syncer to remove it.

go-controller/pkg/ovn/external_ids_syncer/nat/nat_sync.go
```diff
@@ -86,10 +87,10 @@ func (n *NATSyncer) syncEgressIPNATs() error {
...
 			continue
 		}
-		podIP, _, err := net.ParseCIDR(nat.LogicalIP)
-		if err != nil {
-			klog.Errorf("Failed to process logical IP %q of NAT %s", nat.LogicalIP, nat.UUID)
-			continue
+		// for egress IP, the logicalIP does not contain a mask.
+		podIP := net.ParseIP(nat.LogicalIP)
+		if podIP == nil {
+			return fmt.Errorf("failed to process logical IP %q of NAT %s", nat.LogicalIP, nat.UUID)
```
trozet, line 93: why return here?

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
martinkennelly: when eip syncer starts then it wont find a namespace + pod of that name (its also not possible)
I didnt want to do removal within this func or pkg and let it up to the eip controller.

go-controller/pkg/ovn/external_ids_syncer/nat/nat_sync.go
martinkennelly: id rather block than ignore the error. it would be a programming error (Or user messing with DB) and i want it highlighted.

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
martinkennelly: eip controller will only react to ovn constructs it owns - thats why i added this dummy external IDs with the owner being the eip controller.

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
```diff
@@ -101,7 +101,8 @@ func (syncer *LRPSyncer) syncEgressIPReRoutes() error {
 			podInfo, err := cache.getPod(podIP)
 			if err != nil {
 				klog.Infof("Failed to find Logical Switch Port cache entry for pod IP %s: %v", podIP.String(), err)
-				continue
+				// pod not found, add dummy metadata that will be cleaned up by EIP controller sync.
```
pperiyasamy, line 104: for my clarification: 

1. Looks like old LRPs containing external ID with `name` key as EIP entry, is that valid assumption ?
2. pod cache is built from nbdb's LSP, is it guaranteed cache upto date ? 
3.  For the dual stack case, does both v4 and v6 cache have entry for same pod ?

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
pperiyasamy, line 105: it's going to have same external ID for all stale LRPs, is that fine for EIP controller to sync?

go-controller/pkg/ovn/external_ids_syncer/nat/nat_sync_test.go
pperiyasamy, line 1: nit. you may add unit test for stale snat.

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
martinkennelly: 1. yes
2. yes, nb db is syncd before we run this code and we build the cache before this code is exec
3. we have only a single entry in the cache per pod for both ip families

go-controller/pkg/ovn/external_ids_syncer/nat/nat_sync_test.go
martinkennelly: will do - ill add a mix of objs with metadata and objs without.

go-controller/pkg/ovn/external_ids_syncer/logical_router_policy/logical_router_policy_sync.go
martinkennelly: ill add a unit test to the sync code in the ovn eip controller - thanks

martinkennelly: i only manually tested it. i should have written a ut. apologies.

trozet (approved):

Issue #5296, opened 2025-06-11
OKEP: Upstream BGP enhancement


trozet:
### What would you like to be added?

Migrating the OpenShift enhancement to upstream

### Why is this needed?

So that future modifications to the enhancement can be done upstream and not be Red Hat/OpenShift specific.

PR #5289, opened 2025-06-09
[WIP] No-overlay mode enhancement

pliurh:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

PR #5246, opened 2025-05-13
[WIP] OKEP for Connecting UDNs

tssurya:
Closes #5224

PR #5173, opened 2025-04-10
Universal network selector API V2

jcaamano:
Adds a universal network selector API to used across all out other APIs.
Updates RouteAdvertisements API to use this universal network selector API.
Switch other label selectors in the RouteAdvertisements API to the standard convention.
Adapt e2e tests to new API

PR #5297, opened 2025-06-11
OKEP 5296: Upstreams BGP enhancement

trozet:
Migrating from openshift/enhancements

Closes: #5296

## Summary by CodeRabbit

* **Documentation**
  * Added a detailed enhancement proposal outlining BGP routing integration into OVN-Kubernetes, enabling dynamic route advertisement and import with provider networks.
  * Included use cases such as improved throughput, egress IP mobility, and route management automation.
  * Updated documentation navigation to include the new BGP enhancement proposal for easier access.


docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+   and unnecessary for networks where only their respective per-node pod subnet is being advertised to the BGP fabric.
+2. If running shared gateway mode, the reply packet would be routed via br-ex with the default gateway configured in
+   ovn_cluster_router.
+
+For local gateway mode, changes will need to be made to skip the SNAT as it is not necessary and provides pods with the
+true source IP of the sender.
+
+
+Additionally, modifications will need to be made in the kernel routing table to
+leak routes to the pod subnets from each user-defined VRF into the default VRF routing table.
```
tssurya, line 621: I would love to see a "section on isolation here" - all the things we designed on the fly - and future improvements we plan: https://issues.redhat.com/browse/CORENET-6044 the knob is just temporary and we want to do proper API I assume one day

hopefully I didn't miss this when I ctrl-f searched for it .. if so apologies

tssurya:
Also we need a section for LGW L2 BGP design - is that already added ?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+traffic, as well as third party operators to handle external network integration into the cluster. The purpose of this
+enhancement is to introduce BGP as a supported routing protocol with OVN-Kubernetes. The extent of this support will
+allow OVN-Kubernetes to integrate into different BGP user environments, enabling it to dynamically expose cluster scoped
+network entities into a provider’s network, as well as program BGP learned routes from the provider’s network into OVN.
+
+## Goals
+
+* To provide a user facing API to allow configuration of iBGP or eBGP peers, along typical BGP configuration including
+  communities, route filtering, etc.
+* Support for advertising Egress IP addresses.
```
dave-tucker, line 28: As /32? A summary? Or is it not important to differentiate?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+* To enable BFD to BGP peers.
+* Leveraging BGP to allow for no overlay encapsulation with east/west traffic.
```
dave-tucker:
```suggestion
To allow east/west traffic without encapsulation.
```

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+* ECMP routing support within OVN for BGP learned routes.
```
dave-tucker, line 31: Not directly related, but it might be helpful ti add a glossary to this OKEP.
We've already got BGP, BGPd, BFD, iBGP, eBGP within the first few paragraphs 😓

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+* Support for advertising user-defined networks via BGP as long as there is no subnet overlap over the default VRF.
+* Allowing for VRF-Lite type of VPN where the user maps interfaces on the host to user-defined VRFs/networks and
+  advertises VPN routes via BGP sessions over said VRFs.
+
+## Non-Goals
+
+* Support of any other routing protocol.
```
dave-tucker: Do you want to rule this out forever? I know other protocols don't have L2VPN/L3VPN support, but route import/export with OSPF seems like something that might be useful for some use-cases (i.e local gateway route management in a DC that's not running BGP).

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+any other VRF.
+
+The frrConfigurationSelector is used in order to determine which FRRConfiguration CR to use for building the OVN-Kubernetes
+driven FRRConfiguration. OVN-Kubernetes needs to leverage a pre-existing FRRConfiguration to be able to find required
+pieces of configuration like BGP peering, etc. If more than one FRRConfiguration is found matching the selector, then
+an error will be propagated to the RouteAdvertisements CR and no configuration shall be done.
+
+#### No Tunnel/Overlay Mode
+
+Changes will be made to the network CRD(s) in order to specify the method of transport for east/west traffic. A new
```
dave-tucker: `network CRD(s)` = `udn/cudn CRDs`?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+    # frrConfigurationSelector:
+    targetVRF: default
+    advertisements:
+      podNetwork: true
+      egressIP: true
+```      
+
+In the above example, an optional networkSelector may also be optionally supplied which will match namespaces/networks.
+When omitted, RouteAdvertisements will be applied to the default cluster network. If a selector is used that selects the
+default cluster network, it will be enabled for any namespace using the cluster default network. Additionally,
```
dave-tucker, line 189: I assume the use-case for that is if you aren't scheduling workloads on the worker nodes, but wanted to confirm that was the case.

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+advertisements may be limited to specific nodes via the nodeSelector.
+
+A networkSelector may select more than one network, including user-defined networks. In such a case the network subnets
+will be checked by OVN-Kubernetes to determine if there is any overlap of the IP subnets. If so, an error status will be
+reported to the CRD and no BGP configuration will be done by OVN-Kubernetes.
```
dave-tucker: This is a little "odd". Mainly because k8s doesn't have networks 😄.
Let's say I label a `cudn` with `some-other-label=foo` and a namespace with `some-other-label=foo`

```yaml
networkSelector:
  matchLabels: 
      some-other-label: foo
```

Is that going to include my namespace(s)+ cudn(s)? Is that by design?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+The "targetVRF" key is used to determine which VRF the routes should be advertised in. The default value is "default", in
+which case routes will be leaked into the default VRF. One use case for this would be when a user wants to define a
+network with a specific IP addressing scheme, and then wants to advertise the pod IPs into the provider BGP network without
+VPN. Note that by using route leaking with a user-defined network, the network is no longer fully
+isolated, as now any other networks also leaked or attached to that VRF may reach this user-defined network. If a user
+attempts to leak routes into a targetVRF for a user-defined network whose IP subnet would collide with another,
+OVN-Kubernetes will report an error to the RouteAdvertisement status. Alternatively, a user may specify the value "auto",
+which in which case OVN-Kubernetes will advertise routes in the VRF that corresponds to the selected network. Any other
+values other than "auto" or "default" will result in an error. There is no support at this time for routing leaking into
+any other VRF.
```
dave-tucker, line 213: The `auto` magic worries me a bit. How do you determine `the VRF that corresponds to the selected network`?
From the examples it looks like you assume that `k8s.ovn.org/metadata.name` is a valid VRF.
What if I'm not using `k8s.ovn.org/metadata.name` - i.e. because I'm including some UDNs and some namespaces in the same VRF?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+
+#### Route Advertisements
+
+When OVN-Kubernetes detects that a FRR-K8S CR has been created for BGP peering, OVN-Kubernetes will by default
+advertise the pod subnet for each node, via creating an additive, per node FRR-K8S CR that is managed by OVN-Kubernetes.
+OVN-Kubernetes CRD:
+
+```yaml
+apiVersion: k8s.ovn.org/v1
+kind: RouteAdvertisements
```
dave-tucker, line 169:
```suggestion
kind: VirtualRoutingForwarding
```
Maybe we could introduce VRF as a top-level concept?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+metadata:
+  name: default
+spec:
+    # networkSelector:
+    # nodeSelector:
+    # frrConfigurationSelector:
+    targetVRF: default
```
dave-tucker: If we do ☝️ then you can get rid of `targetVRF` since the name of this object is the name of the VRF.
You'd replace that with:
```yaml
import:
  - red
```
Where `red` is the name of the VRF whose routes you want to import.
That makes import/export behviour more explicit + easier to validate.

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+    podNetwork: true
+  networkSelector:
+    matchLabels: 
+      k8s.ovn.org/metadata.name: blue
+  nodeSelector:
+    matchLabels:
+      kubernetes.io/hostname: ovn-worker
+  frrConfigurationSelector:
+    matchLabels:
+      routeAdvertisements: default
```
dave-tucker, line 289: could we not match on `app.kubernetes.io/name: metallb-ovn-worker`?
I was wondering what this special `routeAdvertisements` label was for.

docs/okeps/okep-5296-bgp.md
dave-tucker: For example, what if you wanted something like `route-maps` to filter which routes get imported?
If we had a top-level VRF object, we'd have something to associate route-maps to.

It would be really cool to then be able to map IP addresses for each node via configMap or somesuch. Then ovn-k8s could handle the vrf creation too.

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+  bgp:
+    routers:
+    - asn: 64512
+      neighbors:
+      - address: 172.18.0.5
+        asn: 64512
+        disableMP: false
+        holdTime: 1m30s
+        keepaliveTime: 30s
+        passwordSecret: {}
```
dave-tucker, line 250: I wonder if we should take a position on BGP Transport Security 😉

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+  to use a pure routing implementation to handle advertising egress IP movement across nodes.
+
+## Proposed Solution
+
+OVN-Kubernetes will leverage other projects that already exist to enable BGP in Linux. FRR will be used as the BGP
+speaker, as well as provide support for other protocols like BFD. For FRR configuration, the
+MetalLB project has already started an API to be able to configure FRR:
+[https://github.com/metallb/frr-k8s](https://github.com/metallb/frr-k8s). While some of the configuration support for
+FRR may be directly exposed by FRR-K8S API, it may also be the case that some intermediary CRD provided by
+OVN-Kubernetes is required to integrate OVN-Kubernetes networking concepts into FRR.
```
dave-tucker, line 124: > For FRR configuration, the
MetalLB project has already started an API to be able to configure FRR:
[https://github.com/metallb/frr-k8s](https://github.com/metallb/frr-k8s).

Playing devils advocate here.
What do we gain vs. deploying FRR in a DaemonSet and configure it directly from OVN-K8s?
The potential downside here is that managing a dependency on another operator isn't ideal for a number of reasons. In your examples, the FRR instance is in the `metallb-system` namepsace? What happens if a user decides to uninstall MetalLB (and FRR is cleaned up too). Upgrading MetalLB could cause an outage for Pod-to-Pod networking if it changes the FRR daemonset + you're relying on BGP for Pod-to-pod without encapsulation.

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,601 @@
...
+  networkSelector:
+    matchExpressions:
+      - { key: k8s.ovn.org/metadata.name, operator: In, values: [blue,red] } 
+  nodeSelector:
+    matchLabels:
+      kubernetes.io/hostname: ovn-worker
+  frrConfigurationSelector:
+    matchLabels:
+      routeAdvertisements: vpn-blue-red
+```
```
dave-tucker, line 410: This feels like it should be 2 CRDs.
`RouteAdvertisements/red` and `RouteAdvertisements/blue`.
The "magic" that maps to a VRF based on label feels like a bit of a footgun to me.

docs/okeps/okep-5296-bgp.md
trozet: I didn't add anything other than what was in the openshift enhancement, but you are right this is a good opportunity to do so. I'll add that section.

docs/okeps/okep-5296-bgp.md
trozet: /32 since Egress IPs can be on different nodes but each node/IP combo can only handle its egress IP. I don't think it is important to differentiate at that is point in the doc, but I'll update another section to call out detail.

docs/okeps/okep-5296-bgp.md
trozet: yeah, we already have some users asking about OSPF support. I will move this into Future Goals.

docs/okeps/okep-5296-bgp.md
trozet: I think 2 different things are being conflated here:
1. Where FRR runs.
2. Using FRR-K8S as a mechanism to configure FRR.

It's confusing because FRR-K8S does not rely on metallb at all. The value that FRR-K8S brings to OVNK is that we can abstract the manual config a user would have to accomplish the same configuration in FRR-K8S. Depending on that project is a little bit of a risk, but I think we have had success with it so far. FRR-K8S gives us a more reliable way to configure FRR. It also takes a union of all the CRDs and idempotently ensures that configuration is accurate in FRR. So if there are multiple writers to FRR, it can aggregate all of it.

How FRR is ran is outside the scope of OVNK. We changed MetalLB so that you can run MetalLB without it managing an instance of FRR. In fact with Openshift now, CNO will deploy FRR as a daemonset. When MetalLB is started, it will simply use the FRR already deployed by CNO. Running FRR as a daemonset by CNO is also not ideal. We really should be running FRR as a systemd service on start up so that we can handle day 0 BGP bootstrapped environments.

docs/okeps/okep-5296-bgp.md
trozet: So we already implemented this CRD RouteAdvertisements:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/crd/routeadvertisements/v1/types.go

This is just upstreaming the enhancement that we already have in openshift:
https://github.com/openshift/enhancements/blob/master/enhancements/network/bgp-ovn-kubernetes.md

To your point about route maps, FRR-K8S has some support for filtering already:
https://github.com/metallb/frr-k8s?tab=readme-ov-file#receiving-prefixes-from-a-given-neighbor

In our RA crd we dont expose that directly, but a user could create an FRR-K8S crd to accomplish it. If we added support for routemaps in RA, it would probably be something more abstract that then OVNK will dynamically figure out what route maps to crate.

Today whenever a UDN is created, a VRF is automatically created by ovnk in OVN and in the Linux host. The RA simply gives a way for users to select the UDN and then either advertise that VRF via leaking it into the default VRF, or advertising the VRF on its own VPN (either via VRF-Lite or in the future, EVPN).

docs/okeps/okep-5296-bgp.md
trozet: yeah I guess the targetVRF is really today where we want to "export". There are some plans in the future to enable routeLeaking and importing from other VRFs:

https://docs.google.com/drawings/d/1psQNwTVDFDfvwWaob1qS6lgJ5-4gq9_3l6uazaDz2LU/edit?usp=sharing

UDNConnect is something @tssurya is working on:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5246

routeLeaking in the RA context would mean exporting routes from different UDNs that are connected into the targetVRF. Using "auto" here comes to our advantage because we can select multiple networks and then advertise all of them out of each automatically determined VRF/VPN. Maybe that is a clunky way of doing it and there are better options. This was just the first idea that popped in my head...at 4:30am one night....

Either way we can't change the CRD implementation we have now, but if you see room for improvement we can build on it and make additions.

docs/okeps/okep-5296-bgp.md
trozet: The use case here is if someone is not using a UDN, and just wants to use BGP with the cluster default network (CDN). Then the CDN subnet will be advertised and pods on the default network in any namespace can be addressed directly. It also opens the door for using no overlay mode so that we can give folks the option to not use Geneve, even without UDN.

docs/okeps/okep-5296-bgp.md
trozet: maybe I should rephrase it to just say:

A networkSelector may select more than one user-defined network. ?

We debated in the community for a while on how networkSelectors should be written. The outcome was here:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5173

Depending on the network type, there are different kinds of selectors. Today I think with RA we only allow you to advertise the default cluster network or CUDNs. We are going to add support for primary UDNs later. If you added a selector for CUDN, then we would go find CUDNs that match that label selector, then check that all of those networks do not have IP overlap if they are being advertised into the same VRF. Once we add support for primary UDN (PUDN), then if you had selectors that included PUDN (really a namespaceSelector) and CUDN, then we would find all of those CUDN and UDN crds and ensure they do not have subnet overlap if advertised into the same VRF.

docs/okeps/okep-5296-bgp.md
trozet: For each CUDN/UDN we create the VRF internally. Albeit it is pretty bad naming scheme like mpx-1000-vrf or something like that. That way when someone puts auto and selects networks, we can go determine which vrf it is since we know internally which one it maps to. It isn't ideal that a user has no control over the VRF name. A lot of network admins want to use the same VRF name for good reason across all their PE devices and network. I think @jcaamano introduced a way for a user to influence the VRF name when creating a UDN. @jcaamano do you remember those PRs and our plans for introducing a new field in the CRD?

I remember there are limitations to the number of characters in a VRF device in Linux, which was one of the constraints.

docs/okeps/okep-5296-bgp.md
trozet: well this is an admin CRD at least. The passwordSecret enables MD5 protection on the bgp session. We can also combine it with IPSec, but I guess that is outside the scope of this enhancement.

docs/okeps/okep-5296-bgp.md
trozet: maybe this is just a terrible example. The purpose here is that you might have multiple frr-k8s configurations for the same node. Imagine you have 2 frr-k8s CRs, each one peers to a different BGP neighbor. You only wan to select one of them to advertise your UDN on. This gives you that flexibilty.

docs/okeps/okep-5296-bgp.md
trozet: I think we did this because:
1. A user cannot define/know the VRF name.
2. If you want to select a bunch of UDNs, put them eventually into EVPN and only select a few nodes to do it on, this allows you to do all of that in one place.

Nice side effect with my other comment about enabling route leaking (exporting) of multiple UDNs across each others VPN.

@jcaamano has a better memory than me, maybe he can also give his opinion of why we did it this way.

booxter:
The latest OVN releases (25.03+ I believe?) introduced native BGP support for learning and advertisement of routes (`Advertised_Route`, `Learned_Route` objects). OVN maintains a VRF where routes are updated as needed. Then a BGP daemon translates VRF configuration into BGP messages. In this scenario, the interface between the daemon and OVN is a standard Linux kernel feature (VRF).

In contract, this proposal couples the feature with a particular implementation interface (FRR CRD).

Should this proposal consider leveraging the native OVN support as an alternative path?

trozet:
> BGP support for learning and advertisement of routes (`Advertised_Route`, `Learned_Route` objects). OVN maintains a VRF where routes are updated as needed. Then a BGP daemon translates VRF configuration into BGP messages. In this scenario, the interface between the daemon and OVN is a standard Linux kernel feature (VRF).
> 
> In contract, this proposal couples the feature with a particular implementation interface (FRR CRD).
> 
> Should this proposal consider leveraging the native OVN support as an alternative path?

When it was originally written the BGP support in OVN didn't exist yet, but we can put that in as a future alternative of switching to it. From my conversations in the past with @dceara I thought the OVN implementation was very similar to what we are doing...watching netlink for route updates from FRR and then configuring it into the NBDB. I don't see enough value proposition right now to see us switching to it.

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+| **BFD** (Bidirectional Forwarding Detection) | A network protocol that quickly detects faults in the path between two forwarding devices. It enables sub-second failure detection, often used with BGP or OSPF for faster convergence.                                         |
+| **iBGP** (Internal BGP)                      | A BGP session between routers within the same autonomous system (AS). It is used to distribute external routes internally, usually requiring a full mesh or route reflectors.                                                   |
+| **eBGP** (External BGP)                      | A BGP session between routers in different autonomous systems. It is the primary method for routing between networks on the Internet.                                                                                           |
+| **OSPF** (Open Shortest Path First)          | A link-state interior gateway protocol (IGP) used to distribute IP routing information within a single autonomous system. It calculates the shortest path to each destination using Dijkstra's algorithm and converges quickly. |
```
tssurya, line 14: nice glossary!

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+## Future Goals
+
+* Support EVPN configuration and integration with a user’s DC fabric, along with MAC-VRFs and IP-VRFs.
+* Support iBGP with route reflectors.
+* Potentially advertising other IP addresses, including the Kubernetes API VIP across the BGP fabric.
+* Optimize Layer 2 network routing.
+* Allow selecting only a subset of nodes to advertise BGP.
+* Support other routing protocols like OSPF.
+
```
tssurya, line 53: Till this section its straight from downstream except the support for additional protocols like OSPF which got added to future goals. LGTM

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+* As a user I want to be able to leverage my existing BGP network to dynamically learn routes to pods in my Kubernetes
+  cluster.
+* As a user, rather than having to maintain routes with NM State manually in each Kubernetes node, as well as being
+  constrained to using local gateway mode for respecting user defined routes; I want to use BGP so that I can dynamically
+  advertise egress routes for the Kubernetes pod traffic in either gateway mode.
+* As a user where maximum throughput is a priority, I want to reduce packet overhead by not having to encapsulate
+  traffic with Geneve.
+* As an egress IP user, I do not want to have to restrict my nodes to the same layer 2 segment and prefer
+  to use a pure routing implementation to handle advertising egress IP movement across nodes.
```
tssurya, line 115: Again here as well straight from downstream enhancement LGTM

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+## Proposed Solution
+
+OVN-Kubernetes will leverage other projects that already exist to enable BGP in Linux. FRR will be used as the BGP
```
tssurya:
```suggestion
OVN-Kubernetes will leverage other CNCF projects that already exist to enable BGP in Linux. FRR will be used as the BGP
```
nit

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+FRR may be directly exposed by FRR-K8S API, it may also be the case that some intermediary CRD provided by
+OVN-Kubernetes is required to integrate OVN-Kubernetes networking concepts into FRR.
+
+Functionally, FRR will handle advertising and importing routes and configuring those inside a Linux VRF. OVN-Kubernetes
+will be responsible for listening on netlink and configuring logical routers in OVN with routes learned
+by FRR. OVN-Kubernetes will manage FRR configuration through the FRR-K8S API in order to advertise routes outside of the
+node onto the provider network.
+
+There should be no changes required in FRR. FRR-K8S may need extended APIs to cover the OVN-Kubernetes use cases
+proposed in this enhancement. OVN will require no changes.
```
tssurya: do we need to reconsider the future where OVN has support for learning BGP routers from the host? so whatever  copying is done by OVN-Kubernetes today _can be done by OVN_?
but OK I think we can keep the enhancement as close as possible to the original to be honest and not deviate much so let this be..

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+3. Create the OVN-Kubernetes RouteAdvertisements CR to configure what routes and where to advertise them.
+4. Verify peering and that routes have been propagated correctly.
+5. If desired, change the transport type from geneve on the network to none in order to no longer use an overlay.
+
+For detailed examples, see the [BGP Configuration](#bgp-configuration) section.
+
+### API Extensions
+FRR-K8S API will be used in order to create BGP Peering and configure other BGP related configuration. A
+RouteAdvertisements CRD will be introduced in order to determine which routes should be advertised for specific networks.
+Additionally, the network CRD will be modified in order to expose a new transport field to determine if encapsulation
```
tssurya, line 156: should we link to the no-overlay mode enhancement PR here? or remove it from this OKEP since we didn't implement it?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+should be used for east/west traffic.
```
tssurya, line 157: I guess since we omitted the CNO line here, its good to add that the lifecycle management of FRR-K8s will not be handled by us? totally end user responsibility?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+kind: RouteAdvertisements
+metadata:
+  name: default
+spec:
+    # networkSelector:
+    # nodeSelector:
+    # frrConfigurationSelector:
+    targetVRF: default
+    advertisements:
+      podNetwork: true
```
tssurya: do we want to keep this formatting correct to the real deal or maybe its OK since we are adding user docs as well for BGP via https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5423

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+      egressIP: true
+```      
+
+In the above example, an optional networkSelector may also be optionally supplied which will match namespaces/networks.
```
tssurya: too many optional's? and I think this is not true currently right? with the new selector changes we did we require to submit the defaultnetwork field ?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+When omitted, RouteAdvertisements will be applied to the default cluster network. If a selector is used that selects the
+default cluster network, it will be enabled for any namespace using the cluster default network. Additionally,
+advertisements may be limited to specific nodes via the nodeSelector.
+
+A networkSelector may select more than just the cluster default network, including user-defined networks. In such a case
```
tssurya: should we remove UDNs from scope? or we keep things as is in the OKEP and its OK if some things are not implemented yet I gues?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+the network subnets will be checked by OVN-Kubernetes to determine if there is any overlap of the IP subnets. If so, an
+error status will be reported to the CRD and no BGP configuration will be done by OVN-Kubernetes.
+
+Multiple CRs may not select the same network. If this happens OVN-Kubernetes will report an error to the
+RouteAdvertisements CR and will refuse to apply the config.
+
+The CRD will support enabling advertisements for pod subnet and egress IPs. Note, MetalLB still handles advertising
+LoadBalancer IP so there is no conflict of responsibilities here. When the pod network is set to be advertised, there is
```
tssurya, line 200: not sure I follow this bit with MetalLB why is that bit called our with regards to responsibilities?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+no longer a need to SNAT pod IPs for this network to the node IP. Therefore, when pod network advertisements are enabled,
+the traffic from these pods will no longer be SNAT'ed on egress.
+
+The "targetVRF" key is used to determine which VRF the routes should be advertised in. The default value is "default", in
```
tssurya, line 204: IIUC this field is only really relevant for VRFLite or is that not true?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+any other VRF.
+
+The frrConfigurationSelector is used in order to determine which FRRConfiguration CR to use for building the OVN-Kubernetes
+driven FRRConfiguration. OVN-Kubernetes needs to leverage a pre-existing FRRConfiguration to be able to find required
+pieces of configuration like BGP peering, etc. If more than one FRRConfiguration is found matching the selector, then
+an error will be propagated to the RouteAdvertisements CR and no configuration shall be done.
+
+#### No Tunnel/Overlay Mode
+
+Changes will be made to the CUDN and UDN CRDs in order to specify the method of transport for east/west traffic. A new
```
tssurya: do we keep this and link to no-overlay enhancement PR for more details? https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5289

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+            mode: filtered
+        toReceive:
+          allowed:
+            mode: filtered
+  nodeSelector:
+    matchLabels:
+       kubernetes.io/hostname: ovn-worker
+```
+
+OVNKube-Controller will check that if this FRRConfiguration applies to its node, ovn-worker. For this example, a user-defined
```
tssurya: nit for CNCF correctness here and every to use either `OVN-Kubernetes` or `ovn-kubernetes` for project name and 
`ovnkube-controller`? or `OVN-Kubernetes Controller`

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+  nodeSelector:
+     matchLabels:
+        kubernetes.io/hostname: ovn-worker
+```
+
+OVN-Kubernetes uses the configuration already in place in FRR and the desired RouteAdvertisements to generate the above
+FRRConfiguration. For filtering or choosing what routes to receive, a user should do that in the FRRConfiguration they
+declare for peering.
+
+Note, VRF-Lite is only available when using Local Gateway Mode (routingViaHost: true) in OVN-Kubernetes.
```
tssurya: routingViaHost:true is an openshift thing, let's remove it here

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+#### Preserving UDN Isolation
+
+By definition, UDNs are expected to provide network isolation. With the ability to advertise UDNs within the same BGP
+VRF, it opens the door for users on one UDN to try to access another UDN via external routing. This may be undesirable
+for users who are simply exposing their UDN onto a VRF to allow direct ingress to those pods from outside the cluster,
+and not necessarily wanting to allow UDNs to communicate. Therefore by default, access between externally routed UDNs will
+be blocked by ACLs internal to OVN. This will protect users from accidentally exposing their UDNs to other tenants.
```
tssurya, line 464: thank you for adding this section!

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+In the future, an option will be added to allow a cluster admin to specify that UDNs exposed via BGP may be externally
+routed between each other. Note, this security limitation only applies to pod addresses. If a group of pods are exposed
```
tssurya: you mean isolation only applies to pod network and for services its open?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+#### BGP Route Advertisements
+
+##### Layer 3 Topology
+
+With a Layer 3 network type, each BGP node should advertise the node subnet assigned to it. Since the cluster supernet
+is divided into per node subnets, it makes sense to advertise those from each node. This will result in direct routing
+to the node that hosts destination pod. In the future when nodeSelector is supported, all BGP speakers will also need
+to advertise the cluster supernet. This will allow ECMP routes via any BGP enabled node for reachability to nodes not
+participating in BGP.
```
tssurya, line 483: should we add a note on "don't learn routes of remote node subnets" in this case of in the isolation section above?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+##### Layer 2 Topology
+
+Layer 2 networks are more complicated than Layer 3 due to the nature of a single subnet spanning multiple nodes. In other
+words the cluster supernet is not subnetted out on a per-node basis. Therefore, every node must advertise the cluster supernet.
+This results in asynchronous routing, and in practical terms, packets having to endure an extra hop before making it to their
```
tssurya, line 489: do we need to mention the OVN switch off invalid conntrack option we did in LGW here in design? for asynchronous traffic to work?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+
+A user can also configure RouteAdvertisements for namespaces affected by MEG. Since MEG directly uses pod IPs
+(in disable-snat-multiple-gws mode), the external gateway needs to know where to route pod IPs for ingress and egress
+reply traffic. Traditionally this is done by the gateway having its own operator to detect pod subnets via kubernetes
+API. With BGP, this is no longer necessary. A user can simply configure RouteAdvertisements, and the pod subnet routes
+will be dynamically learned by an external gateway capable of BGP peering.
+
+##### Egress IP
+
+EgressIP feature that is dynamically moved between nodes. By enabling the feature in RouteAdvertisements, OVN-Kubernetes
```
tssurya, line 538: hmm there is a section on EgressIP just above this one.. too close?

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+ 2(patch-breth0_ov): addr:ca:2c:82:e9:06:42
+     config:     0
+     state:      0
+     speed: 0 Mbps now, 0 Mbps max
+
+
+[root@ovn-worker ~]# ovs-ofctl dump-flows breth0 |grep 10.244
+ cookie=0xdeff105, duration=437.393s, table=0, n_packets=4, n_bytes=392, idle_age=228, priority=109,ip,in_port=2,dl_src=02:42:ac:12:00:03,nw_src=10.244.1.0/24 actions=ct(commit,zone=64000,exec(load:0x1->NXM_NX_CT_MARK[])),output:1
+ cookie=0xdeff105, duration=437.393s, table=0, n_packets=0, n_bytes=0, idle_age=437, priority=104,ip,in_port=2,nw_src=10.244.0.0/16 actions=drop
+ cookie=0xdeff105, duration=437.393s, table=1, n_packets=4, n_bytes=392, idle_age=228, priority=15,ip,nw_dst=10.244.0.0/16 actions=output:2
```
tssurya, line 596: I think the devel docs should update exactly later on the flow details of how we changed it specially now in Peng's PR

docs/okeps/okep-5296-bgp.md
```diff
@@ -0,0 +1,663 @@
...
+* Testing using transport none for some networks, and Geneve for non-BGP enabled networks.
+* Testing to cover BGP functionality including MEG, Egress IP, Egress QoS, etc.
+* Scale testing to determine impact of FRR-K8S footprint on large scale deployments.
+
+### Documentation Details
+
+* ovn-kubernetes.io will be updated with a BGP user guide
+* Additional dev guides will be added to the repo to show how the internal design of BGP is implemented.
+
+* New proposed additions to ovn-kubernetes.io for end users
```
tssurya: nit: remove the template sections here..

tssurya (changes requested):
I focussed on newer sections like isolation/layer2/layer3 etc and for what was agreed upon in downstream enhancement I did not poke any holes there because its all already implemented.

Left some comments/suggestions but overall LGTM.

tssurya:
@trozet Reminder: I want to cut release tomorrow, can you please rebase and push review comments?

docs/okeps/okep-5296-bgp.md
trozet: I think OVN has this capability now, but yeah the way we implemented it did not require that. It can still change in the future.

docs/okeps/okep-5296-bgp.md
trozet: yeah I will remove this sentence, and remove other references to the no overlay in the implementation. I'll list no overlay as a future goal.

docs/okeps/okep-5296-bgp.md
trozet: I'll add a comment about this.

docs/okeps/okep-5296-bgp.md
trozet: yeah I will fix

docs/okeps/okep-5296-bgp.md
trozet: will fix

docs/okeps/okep-5296-bgp.md
trozet: because MetalLB provisions and creates the FRR config to advertise the loadbalancer IP. OVNK will not advertise any loadbalancer IPs.

docs/okeps/okep-5296-bgp.md
trozet: it will also be relevant for EVPN

docs/okeps/okep-5296-bgp.md
trozet: no i dont want to reference a PR inside of an enhancement. I feel like it would be better to reference the document. So when no overlay eventually merges, we can update this doc to point to it.

docs/okeps/okep-5296-bgp.md
trozet: pod addresses and cluster IP addresses, but not loadbalancer, nodeport, external IP. I'll re-word this.

docs/okeps/okep-5296-bgp.md
trozet: ive added a sentence for that. However with no overlay we will have to enable route importing between nodes. I don't want to go too far into the weeds on that and the no overlay PR can address those changes.

docs/okeps/okep-5296-bgp.md
trozet: this is a sub-section of feature compat. I get that it is awkward, but let's not change it now.

docs/okeps/okep-5296-bgp.md
tssurya: makes sense

docs/okeps/okep-5296-bgp.md
tssurya: ah I see gotcha!

docs/okeps/okep-5296-bgp.md
tssurya: makes sense!

tssurya (approved):
thanks @trozet !

needs a rebase on master to fix conflict, but once that's done I'll merge this.

PR #2782, opened 2022-01-31
Fixes finding default gateway for configured GW interface

trozet:
The code for determining the default gateway was not taking into
consideration if the gateway interfaces was provided via config and
would return any default gateway found on the host. This is an issue
when a system has more than 1 default gateway, as well as if a user
provided an interface other than the real default gateway interface.

If the gateway interface is provided via config, the code now scopes
default routes found to only that interface.

Signed-off-by: Tim Rozet <trozet@redhat.com>

PR #5298, opened 2025-06-12
ovnkube.sh: Add new overwriting options for the gateway options and kubernetes node name

aserdean:
This commit adds:
  a) options to change ovn_gateway_opts and ovn_gateway_router_subnet by a container inside the same POD.
     the idea is that a init container can do an IP allocation write the output to a file and we will
     consume those values from the file.
  b) in case of ovnkube in DPU mode, we are running ovnkube on behalf of a different host, however the
     way we identify that is using the DPU hostname. to bypass the latter we will use the OVS metadata
     external_ids:host-k8s-nodename. This is already used by the ovn-node (OVN central where we have a single
     global zone).
  c) extend stateless network policies for ovnkube running in different mode types: ovn-master, ovnkube-controller
     and ovnkube-controller-with-node. this is useful for offloading RDMA traffic.

Signed-off-by: Alin Gabriel Serdean <aserdean@ovn.org>

### DIFF
--- a/dist/images/ovnkube.sh
+++ b/dist/images/ovnkube.sh
@@ -324,15 +324,14 @@ ovn_nohostsubnet_label=${OVN_NOHOSTSUBNET_LABEL:-""}
 # should be set to true when dpu nodes are in the cluster
 ovn_disable_requestedchassis=${OVN_DISABLE_REQUESTEDCHASSIS:-false}
 
-# external_ids:host-k8s-nodename is set on an Open_vSwitch enabled system if the ovnkube pod
-# should function on behalf of a different host than external_ids:host
+# external_ids:host-k8s-nodename is set on an Open_vSwitch enabled system if the ovnkube stack
+# should function on behalf of a different host than external_ids:hostname. This includes
+# all the components that belond in an ovnkube stack (i.e. NB DB, SB DB, ovnkube etc)
 # overwrite the K8S_NODE env var with the one found within the OVS metadata in this case
-if [[ ${ovnkube_node_mode} == "dpu" ]]; then
-  K8S_NODE=$(ovs-vsctl --if-exists get Open_vSwitch . external_ids:host-k8s-nodename | tr -d '\"')
-  if [[ ${K8S_NODE} == "" ]]; then
-    echo "Trying to run in DPU mode and couldn't get the required Host K8s Nodename. Exiting..."
-    exit 1
-  fi
+ovn_k8s_node=$(ovs-vsctl --if-exists get Open_vSwitch . external_ids:host-k8s-nodename | tr -d '\"')
+if [[ ! -z $ovn_k8s_node ]]; then
+  echo "host-k8s-nodename is set, overriding K8S_NODE with $ovn_k8s_node"
+  K8S_NODE=$ovn_k8s_node
 fi
 
 # Determine the ovn rundir.
@@ -2423,10 +2422,12 @@ ovn-node() {
     wait_for_event ovs_ready
   fi
 
-  if [[ ${ovnkube_node_mode} != "dpu-host" ]] && [[ ${ovn_enable_interconnect} != "true" ]]; then
+  if [[ ${ovnkube_node_mode} == "dpu-host" ]] && [[ ${ovn_enable_interconnect} == "true" ]]; then
     # ready_to_start_node checks for the NB/SB readiness state.
     # This is not available on the DPU host when interconnect is enabled,
     # because the DBs will run locally on the DPU
+    echo "skipping ready_to_start_node on DPU Host and when interconnect is true"
+  else
     echo "=============== ovn-node - (wait for ready_to_start_node)"
     wait_for_event ready_to_start_node
   fi
--- a/docs/features/hardware-offload/dpu-support.md
+++ b/docs/features/hardware-offload/dpu-support.md
@@ -17,3 +17,39 @@ on the embedded CPU.
 Any vendor that manufactures a DPU which supports the above model should work with current design.
 
 Design document can be found [here](https://docs.google.com/document/d/11IoMKiohK7hIyIE36FJmwJv46DEBx52a4fqvrpCBBcg/edit?usp=sharing).
+
+## OVN Kubernetes in a DPU-Accelerated Environment
+
+The **ovn-kubernetes** deployment will have two parts one on the host and another on the DPU side.
+
+
+These aforementioned parts are expected to be deployed also on two different Kubernetes clusters, one for the host and another for the DPUs.
+
+
+### Host Cluster
+---
+
+#### OVN Kubernetes control plane related component
+- ovn-cluster-manager
+
+#### OVN Kubernetes components on a Standard Host (Non-DPU)
+- local-nb-ovsdb
+- local-sb-ovsdb
+- run-ovn-northd
+- ovnkube-controller-with-node
+- ovn-controller
+- ovs-metrics
+
+#### OVN Kubernetes component on a DPU-Enabled Host
+- ovn-node
+
+### DPU Cluster
+---
+
+#### OVN Kubernetes components
+- local-nb-ovsdb 
+- local-sb-ovsdb
+- run-ovn-northd
+- ovnkube-controller-with-node
+- ovn-controller
+- ovs-metrics


aserdean:
not sure why DCO is failing since the commit is signed.

dist/images/ovnkube.sh
```diff
@@ -324,6 +336,14 @@ ovn_nohostsubnet_label=${OVN_NOHOSTSUBNET_LABEL:-""}
 # should be set to true when dpu nodes are in the cluster
 ovn_disable_requestedchassis=${OVN_DISABLE_REQUESTEDCHASSIS:-false}
 
+# external_ids:host-k8s-nodename is set on an Open_vSwitch enabled system if the ovnkube pod
+# should function on behalf of a diiferent host than external_ids:host
```
trozet: nit: different

dist/images/ovnkube.sh
```diff
@@ -324,6 +336,14 @@ ovn_nohostsubnet_label=${OVN_NOHOSTSUBNET_LABEL:-""}
...
+# overwrite the K8S_NODE env var with the one found within the OVS metadata in this case
+node_name=$(ovs-vsctl --if-exists get Open_vSwitch . external_ids:host-k8s-nodename | tr -d '\"')
+if [[ ! -z $node_name ]]; then
+  K8S_NODE=$node_name
```
trozet: so you are detecting it here, but in other cases (like if dpu mode) you are exiting if you cant find it. It's a little awkward because you are setting it here as a global variable, and then in the functions -- they are not using a local variable.

Why don't we remove the other checks, and just do it one time here to set K8S_NODE, then add a check to see if ovnkube_node_mode == dpu and this node_name is empty, exit 1?

dist/images/ovnkube.sh
```diff
@@ -2355,7 +2404,10 @@ ovn-controller() {
   wait_for_event ovs_ready
 
   echo "=============== ovn-controller - (wait for ready_to_start_node)"
-  wait_for_event ready_to_start_node
+  if [[ ${ovnkube_node_mode} != "dpu-host" ]] && [[ ${ovn_enable_interconnect} != "true" ]]; then
+    echo "=============== ovn-node - (wait for ready_to_start_node)"
+    wait_for_event ready_to_start_node
+  fi
```
trozet: how come this is needed? ovn-controller wouldn't run on the host right? Also doesn't the ready_to_start_node function just skip if it is not IC from within the function?

dist/images/ovnkube.sh
```diff
@@ -2400,7 +2452,10 @@ ovn-node() {
   fi
 
   echo "=============== ovn-node - (wait for ready_to_start_node)"
-  wait_for_event ready_to_start_node
+  if [[ ${ovnkube_node_mode} != "dpu-host" ]] && [[ ${ovn_enable_interconnect} != "true" ]]; then
+    echo "=============== ovn-node - (wait for ready_to_start_node)"
+    wait_for_event ready_to_start_node
+  fi
```
trozet: we dont use ovnkube-node with IC right? I think we run it as part of ovnkube-controller combined now:
ovnkube-controller-with-node

trozet (changes requested):

dist/images/ovnkube.sh
aserdean: fair, will remove this

dist/images/ovnkube.sh
aserdean: In my case I am using it in on the host side in conjunction with ovn-ic on the DPU side.

If I comment this part out as is the dpu-host side will fail to become ready:
```
kubectl get pods -A | grep dpu-host
ovn-kubernetes            ovn-kubernetes-node-dpu-host-9xjxw                            0/1     Running             0                5m8s
ovn-kubernetes            ovn-kubernetes-node-dpu-host-nw562                            0/1     Running             0                5m16s
```

dist/images/ovnkube.sh
aserdean: > Why don't we remove the other checks, and just do it one time here to set K8S_NODE, then add a check to see if ovnkube_node_mode == dpu and this node_name is empty, exit 1?

makes, sense, originally I added it as a global var in the case I missed any of the modes.
but this makes sense to remove a bunch of the duplicate code.

aserdean: follow-up question should we do this sorta thing for other settings i.e. stateless net policy?

dist/images/ovnkube.sh
```diff
@@ -324,6 +336,17 @@ ovn_nohostsubnet_label=${OVN_NOHOSTSUBNET_LABEL:-""}
 # should be set to true when dpu nodes are in the cluster
 ovn_disable_requestedchassis=${OVN_DISABLE_REQUESTEDCHASSIS:-false}
 
+# external_ids:host-k8s-nodename is set on an Open_vSwitch enabled system if the ovnkube pod
+# should function on behalf of a different host than external_ids:host
+# overwrite the K8S_NODE env var with the one found within the OVS metadata in this case
+if [[ ${ovnkube_node_mode} == "dpu" ]]; then
+  K8S_NODE=$(ovs-vsctl --if-exists get Open_vSwitch . external_ids:host-k8s-nodename | tr -d '\"')
+  if [[ ${K8S_NODE} == "" ]]; then
+    echo "Trying to run in DPU mode and ouldn't get the required Host K8s Nodename. Exiting..."
```
npinaeva: nit: couldn't

dist/images/ovnkube.sh
npinaeva: please add a smalls section here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/features/hardware-offload/dpu-support.md to show which parts of ovn-k run in which containers/pods on the node and DPU

aserdean: is it okay to show them in the case OVN IC only?

npinaeva: yeah if that is the only one you are actually testing, makes sense. Just make it clear that this is the only supported/tested setup

aserdean: oki

aserdean: added

dist/images/ovnkube.sh
trozet: I don't see why not. As long as it applies to all modes lets just make it common.

trozet:
@npinaeva will wait for your review/merge

dist/images/ovnkube.sh
```diff
@@ -2404,10 +2391,7 @@ ovn-controller() {
   wait_for_event ovs_ready
 
   echo "=============== ovn-controller - (wait for ready_to_start_node)"
-  if [[ ${ovnkube_node_mode} != "dpu-host" ]] && [[ ${ovn_enable_interconnect} != "true" ]]; then
-    echo "=============== ovn-node - (wait for ready_to_start_node)"
-    wait_for_event ready_to_start_node
-  fi
```
npinaeva: doesn't this totally revert the change from the first commit? Maybe it makes sense to squash the first 2 commits?

aserdean: indeed, will squash them

dist/images/ovnkube.sh
```diff
@@ -2400,7 +2436,10 @@ ovn-node() {
   fi
 
   echo "=============== ovn-node - (wait for ready_to_start_node)"
-  wait_for_event ready_to_start_node
+  if [[ ${ovnkube_node_mode} != "dpu-host" ]] && [[ ${ovn_enable_interconnect} != "true" ]]; then
```
npinaeva: since it took me some time to figure this out, let's add a comment here saying that `ready_to_start_node` checks the nb/sb db readinness, but in dpu ovn-ic mode dpu-host can't reach local dpu db, hence the check should be skipped

npinaeva (commented):
Almost there!
This is much easier to understand now, thank you for the comments and docs!

romfreiman:
> Almost there! This is much easier to understand now, thank you for the comments and docs!

@aserdean I told you just to use copliot  and ask it to add docs :)

trozet:
@npinaeva I got clarification from @aserdean on why this part is needed:

```
a) options to change ovn_gateway_opts and ovn_gateway_router_subnet by a container inside the same POD.
the idea is that a init container can do an IP allocation write the output to a file and we will
consume those values from the file.
```
For the unique use case with DPU, the requirements/details are the following:

1. This change is specific to ovnkube running down in the dpu.
2. When ovnkube on dpu is started, it specifies gateway-interface as br-ovn, an OVS bridge that is created by a script before ovnkube is launched.
3. br-ovn may not yet have an IP address, when ovnkube-dpu starts. This is OK as ovnkube-dpu would just fail to start and retry until the IP is there.
4. once br-ovn has an IP address, ovnkube would start and then it would try to find the appropriate route to use for the nexthop from the linux table on the dpu.
5. However, rather than using the default route on the dpu, @aserdean wants to be able to specify potentially a different IP that has a route but it is not a default route on the dpu.

Thinking about this problem more generically, I remember we had users who had multiple default gateway routes via different interfaces on their host before. I fixed this here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/2782

That means when we lookup the default route we would only scope our lookup to br-ovn interface. Therefore one potential alternative to the behavior in this PR would be to ensure that br-ovn interface on the dpu has a default gateway route with a higher metric to the next hop. Then our auto-detection would work. @aserdean any thoughts there?

aserdean:
> That means when we lookup the default route we would only scope our lookup to br-ovn interface. Therefore one potential alternative to the behavior in this PR would be to ensure that br-ovn interface on the dpu has a default gateway route with a higher metric to the next hop. Then our auto-detection would work. @aserdean any thoughts there?

`br-ovn` does not have a default route and I would avoid adding a default route just to enable detection:
```
root@worker1-0000-89-00:/home/ubuntu# ip r
default via 10.0.110.254 dev br-comm-ch proto dhcp src 10.0.110.66 metric 100 
10.0.110.0/24 dev br-comm-ch proto kernel scope link src 10.0.110.66 metric 100 
10.0.110.0/24 via 10.0.120.3 dev br-ovn metric 10000 
10.0.110.252 dev br-comm-ch proto dhcp scope link src 10.0.110.66 metric 100 
10.0.110.254 dev br-comm-ch proto dhcp scope link src 10.0.110.66 metric 100 
10.0.120.0/29 dev br-ovn proto kernel scope link src 10.0.120.1 
10.0.120.0/22 via 10.0.120.3 dev br-ovn 
10.244.0.0/24 dev cni0 proto kernel scope link src 10.244.0.1 
10.244.1.0/24 via 10.244.1.0 dev flannel.1 onlink 
192.168.100.0/30 dev tmfifo_net0 proto kernel scope link src 192.168.100.2 
```

aserdean:
> > That means when we lookup the default route we would only scope our lookup to br-ovn interface. Therefore one potential alternative to the behavior in this PR would be to ensure that br-ovn interface on the dpu has a default gateway route with a higher metric to the next hop. Then our auto-detection would work. @aserdean any thoughts there?
> 
> `br-ovn` does not have a default route and I would avoid adding a default route just to enable detection:

@trozet 

Maybe a follow up question would be why are we trying to enforce the autodetection?

I understand that it is a nice UX for users to not specify fields but in the case the user knows exactly what he wants there why should we not allow him to do that?

If the filepath is not a good approach, alternatively  we could read that data from the OVS metadata.

trozet:
> > > That means when we lookup the default route we would only scope our lookup to br-ovn interface. Therefore one potential alternative to the behavior in this PR would be to ensure that br-ovn interface on the dpu has a default gateway route with a higher metric to the next hop. Then our auto-detection would work. @aserdean any thoughts there?
> > 
> > 
> > `br-ovn` does not have a default route and I would avoid adding a default route just to enable detection:
> 
> @trozet
> 
> Maybe a follow up question would be why are we trying to enforce the autodetection?
> 
> I understand that it is a nice UX for users to not specify fields but in the case the user knows exactly what he wants there why should we not allow him to do that?
> 
> If the filepath is not a good approach, alternatively we could read that data from the OVS metadata.

You are right we shouldn't force auto detection and allow the user to provide the value. We do that already with providing the argument for --gateway-nexthop. In OVNK you can also put the values in a config file:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L2255C3-L2255C4

So why not just make the init container write the value in the config file we already support today?

aserdean:
> So why not just make the init container write the value in the config file we already support today?



> > > > That means when we lookup the default route we would only scope our lookup to br-ovn interface. Therefore one potential alternative to the behavior in this PR would be to ensure that br-ovn interface on the dpu has a default gateway route with a higher metric to the next hop. Then our auto-detection would work. @aserdean any thoughts there?
> > > 
> > > 
> > > `br-ovn` does not have a default route and I would avoid adding a default route just to enable detection:
> > 
> > 
> > @trozet
> > Maybe a follow up question would be why are we trying to enforce the autodetection?
> > I understand that it is a nice UX for users to not specify fields but in the case the user knows exactly what he wants there why should we not allow him to do that?
> > If the filepath is not a good approach, alternatively we could read that data from the OVS metadata.
> 
> You are right we shouldn't force auto detection and allow the user to provide the value. We do that already with providing the argument for --gateway-nexthop. In OVNK you can also put the values in a config file:
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L2255C3-L2255C4
> 
> So why not just make the init container write the value in the config file we already support today?

That is a very good point, I will adapt the init container to set the gw opts and see if we have any gaps.

But beside that is there anything else that should be changed?

aserdean:
> > So why not just make the init container write the value in the config file we already support today?
> 
> > > > > That means when we lookup the default route we would only scope our lookup to br-ovn interface. Therefore one potential alternative to the behavior in this PR would be to ensure that br-ovn interface on the dpu has a default gateway route with a higher metric to the next hop. Then our auto-detection would work. @aserdean any thoughts there?
> > > > 
> > > > 
> > > > `br-ovn` does not have a default route and I would avoid adding a default route just to enable detection:
> > > 
> > > 
> > > @trozet
> > > Maybe a follow up question would be why are we trying to enforce the autodetection?
> > > I understand that it is a nice UX for users to not specify fields but in the case the user knows exactly what he wants there why should we not allow him to do that?
> > > If the filepath is not a good approach, alternatively we could read that data from the OVS metadata.
> > 
> > 
> > You are right we shouldn't force auto detection and allow the user to provide the value. We do that already with providing the argument for --gateway-nexthop. In OVNK you can also put the values in a config file:
> > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/config/config.go#L2255C3-L2255C4
> > So why not just make the init container write the value in the config file we already support today?
> 
> That is a very good point, I will adapt the init container to set the gw opts and see if we have any gaps.
> 
> But beside that is there anything else that should be changed?

@trozet just managed to test this and it worked like a charm, thank you for pointing out.

I guess the only downside is that we do not see the values on the ovnkube process arguments, so debugging info will come only from the log. 

please let me know if there any other concerns I need to address.

npinaeva (dismissed):
great job, thank you!

dist/images/ovnkube.sh
```diff
@@ -2399,8 +2422,13 @@ ovn-node() {
     wait_for_event ovs_ready
   fi
 
-  echo "=============== ovn-node - (wait for ready_to_start_node)"
-  wait_for_event ready_to_start_node
+  if [[ ${ovnkube_node_mode} != "dpu-host" ]] && [[ ${ovn_enable_interconnect} != "true" ]]; then
```
npinaeva: ok. too fast, I think this may be the reason for test failure, should this condition use `||` instead of `&&`?

aserdean: fair point, will test on my cluster and lets see if the CI passes.

PR #3040, opened 2022-06-16
EGW: Delete Stale Conntrack Entries

tssurya:
OVNK parts work, ARP-ing works, conntrack deletion works -> tested on KIND and user lab
However in order for this PR to work completely we will need [an upstream PR to netlink library for label matches.](https://github.com/vishvananda/netlink/pull/784) to merge first and then I can remove my 2nd commit.

Logic:

1. annotate the namespace each time an exgw is added/deleted with list of ips
2. add new informer for namespace on node side checking only if gw ip annotation changed OR routing-externalgws annotation changed
3. ovnkube node on namespace change, iterates through all the ips and initiates an arp request via ovnk and collects the MACs
4. once all the responses come back, we have all the known macs
5. we search for ct entries for any pod ip belonging to the namespace, if ct_label is loaded with a mac not in our list we flush it
6. we run the above in a goroutine as well set which will run every 5mins looping through all relevant namespaces.

Brainstormed-By: @trozet , @dcbw 

NOTE for reviewers: Added/Updated unit tests, and two new e2e test's in `external_gateways.go` file for the namespace annotation case and external gateway annotation case.

PR #5300, opened 2025-06-12
External gateway: Remove routes for external gateway pods in terminating or not ready state

arkadeepsen:
## 📑 Description

This PR fixes the issue of traffic still being routed to external gateway pods even when the pods are in terminating or not ready state. Whenever the pods go into terminating or not ready state the routes to the pods are removed so that no traffic is routed to them.

Fixes #

## ✅ Checks

- [x] My code requires tests

## How to verify it

For verification both unit and e2e tests are added.


martinkennelly:
are we copying the semantics of service endpoints? Ricky is awesome for knowing the  semantics there.

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
```diff
@@ -28,6 +30,13 @@ func (m *externalPolicyManager) syncPod(pod *corev1.Pod, routeQueue workqueue.Ty
 }
 
 func getExGwPodIPs(gatewayPod *corev1.Pod, networkName string) (sets.Set[string], error) {
+	// If an external gateway pod is in terminating or not ready state then don't return the
+	// IPs for the external gateway pod
+	if util.PodTerminating(gatewayPod) || util.PodNotReady(gatewayPod) {
```
pperiyasamy: isn't pod not running in both cases ? if that is the case, can't just use `!util.PodRunning(gatewayPod)` ?

go-controller/pkg/ovn/controller/apbroute/external_controller_pod_test.go
```diff
@@ -448,6 +448,141 @@ var _ = Describe("OVN External Gateway pod", func() {
...
+				[]*namespaceWithPods{namespaceGWWith1Pod}, false)
+
+			expectedPolicy2, expectedRefs2 = expectedPolicyStateAndRefs(
+				[]*namespaceWithPods{namespaceTargetWithPod},
+				nil,
+				[]*namespaceWithPods{namespaceGWWith1Pod}, false)
+
+			eventuallyExpectNumberOfPolicies(2)
+			eventuallyExpectConfig(dynamicPolicy.Name, expectedPolicy1, expectedRefs1)
+			eventuallyExpectConfig(dynamicPolicyDiffTargetNS.Name, expectedPolicy2, expectedRefs2)
```
pperiyasamy, line 580: it may be good to add another test step here to get `pod1` back into running state and validate dynamic GW is updated with both pods.

pperiyasamy:
will review E2E a bit later.

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
arkadeepsen: In both the cases, pod can remain in running state thus using `!util.PodRunning(gatewayPod)` may not have the desired result.

When the pod is marked for deletion with a finalizer, the pod will continue to remain in Running state up until the finalizer is removed from the pod. The deletion timestamp will be added to the pod in a such case.

When the readiness probe fails for some reason, the pod will be marked as not ready. However, the pod can remain in Running state even after being marked as not ready.

arkadeepsen:
> are we copying the semantics of service endpoints? Ricky is awesome for knowing the semantics there.

In this PR, the additional check of whether the ex gw pod is in terminating or not ready state is added. The service endpoints semantics is not used in the PR.

go-controller/pkg/ovn/controller/apbroute/external_controller_pod_test.go
arkadeepsen: Added the test step.

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
pperiyasamy: ok, so trying to understand, can't it still serve as a gateway though it's running during that period ?

test/e2e/external_gateways.go
```diff
@@ -874,10 +884,55 @@ var _ = ginkgo.Describe("External Gateway", func() {
...
 				if addresses.srcPodIP == "" || addresses.nodeIP == "" {
 					skipper.Skipf("Skipping as pod ip / node ip are not set pod ip %s node ip %s", addresses.srcPodIP, addresses.nodeIP)
 				}
+
+				if removalType == GatewayNotReady {
+					ginkgo.By(fmt.Sprintf("Delete second external gateway pod %s from ns %s", gatewayPodName2, servingNamespace))
+					err = f.ClientSet.CoreV1().Pods(servingNamespace).Delete(context.TODO(), gatewayPodName2, metav1.DeleteOptions{})
+					framework.ExpectNoError(err, "Delete the gateway pod failed: %v", err)
+					// give some time to handle pod delete event
+					time.Sleep(5 * time.Second)
```
pperiyasamy: why there is a hard 5s sleep here ? can't it covered under below Eventually block ?

test/e2e/external_gateways.go
```diff
@@ -874,10 +884,55 @@ var _ = ginkgo.Describe("External Gateway", func() {
...
+					gomega.Eventually(func() bool {
+						_, err := f.ClientSet.CoreV1().Pods(servingNamespace).Get(context.Background(), gatewayPodName2, metav1.GetOptions{})
```
pperiyasamy: i think `err` must not be a new variable here because you are trying to log this at 904.

test/e2e/external_gateways.go
```diff
@@ -874,10 +884,55 @@ var _ = ginkgo.Describe("External Gateway", func() {
...
+								},
+							},
+							InitialDelaySeconds: 5,
+							PeriodSeconds:       5,
+							FailureThreshold:    1,
+						}
+					})
+					framework.ExpectNoError(err, "Create and annotate the external gw pod to manage the src app pod namespace, failed: %v", err)
+					// give some time to handle pod create event
+					time.Sleep(5 * time.Second)
```
pperiyasamy: same comments here.

test/e2e/external_gateways.go
```diff
@@ -924,15 +979,54 @@ var _ = ginkgo.Describe("External Gateway", func() {
...
+				case GatewayUpdate:
 					ginkgo.By("Remove second external gateway pod's routing-namespace annotation")
 					annotatePodForGateway(gatewayPodName2, servingNamespace, "", addresses.gatewayIPs[1], false)
+				case GatewayDeletionTimestamp:
+					ginkgo.By("Setting finalizer then deleting external gateway pod with grace period to set deletion timestamp")
+					p := getGatewayPod(f, servingNamespace, gatewayPodName2)
+					p.Finalizers = append(p.Finalizers, "k8s.ovn.org/external-gw-pod-finalizer")
+					updatePod(f, p)
+					// give some time for the pod update event
+					time.Sleep(5 * time.Second)
```
pperiyasamy: you are not actually validating if the pod is updated with the finalizer, 5s sleep will not guarantee that.

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
arkadeepsen: Even if the gateway pod is in running state while it's terminating or in not ready state, it may not be able to serve traffic. In terminating state, it might be gracefully shutting down the processes running inside the pod and during that time it may not serve the traffic. In the case of gateway not being ready, there are chances that even if the pod is still in running state it may not serve the traffic as the readiness probe has failed. Thus, rather than waiting for the pod to completely stop running, it's being preemptively removed from the list of active gateways to avoid the scenarios where a gateway is marked as active however it does not serve the traffic that is being forwarded to it.

test/e2e/external_gateways.go
arkadeepsen: Removed the sleep.

test/e2e/external_gateways.go
arkadeepsen: Removed the sleep.

test/e2e/external_gateways.go
arkadeepsen: Added an eventually block to validate it.

test/e2e/external_gateways.go
```diff
@@ -2027,8 +2172,58 @@ var _ = ginkgo.Describe("External Gateway", func() {
...
+						p = getGatewayPod(f, servingNamespace, gatewayPodName2)
+						p.Finalizers = []string{}
+						updatePod(f, p)
+					}()
+				case GatewayNotReady:
+					ginkgo.By("Remove /tmp/ready in external gateway pod so that readiness probe fails")
+					_, err = e2ekubectl.RunKubectl(servingNamespace, "exec", gatewayPodName2, "--", "rm", "/tmp/ready")
+					framework.ExpectNoError(err, fmt.Sprintf("unable to remove /tmp/ready in pod: %s, err: %v", gatewayPodName2, err))
+					// give some time to handle pod event
+					time.Sleep(5 * time.Second)
```
pperiyasamy: Nit: remove this hard sleep and move it to eventually block.

arkadeepsen: Changed hard sleep to eventually block.

arkadeepsen:
> If my above understanding is correct ^ why can't user use the feature with `bfdEnabled` which means if the serving pods are not healthy, then the heartbeats would fail right? So the ECMP routes are going to not be added automatically and cache wouldn't contain those entries right?

This was also asked by Tim in the downstream bug [here](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27194118&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27194118).
Currently, the ECMP conntrack entries are not flushed as mentioned by [Lorenzo](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27194835&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27194835) and [Dumitru](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27197763&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27197763).
Tim suggested [2 step approach](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27203864&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27203864) for fixing the bug.

tssurya:
> > If my above understanding is correct ^ why can't user use the feature with `bfdEnabled` which means if the serving pods are not healthy, then the heartbeats would fail right? So the ECMP routes are going to not be added automatically and cache wouldn't contain those entries right?
> 
> This was also asked by Tim in the downstream bug [here](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27194118&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27194118). Currently, the ECMP conntrack entries are not flushed as mentioned by [Lorenzo](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27194835&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27194835) and [Dumitru](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27197763&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27197763). Tim suggested [2 step approach](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27203864&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27203864) for fixing the bug.

haha that still doesn't answer my question, so I know OVN's history around conntrack flushing. OVN doesn't do it today yes which is why we have https://github.com/ovn-kubernetes/ovn-kubernetes/pull/3040 and did the whole annotation for that to happen.. so its not about the conntrack flushing piece which I know I had opened https://bugzilla.redhat.com/show_bug.cgi?id=2096233 which then got reverted due to perf issues and now we have https://issues.redhat.com/browse/FDP-56 which we need to enable but for that we need https://issues.redhat.com/browse/OCPBUGS-49889 - so finally it is OVN-K doing the flush I am fine with that..

So I am not sure I understand why user can't use bfd so we would know the heart beats fail and ECMP routes would be removed right? This fix is not fixing conntrack right? It's trying to fix removal of routes + IPs from annotation? and removing that will auto-fix conntrack flush which will happen

tssurya:
> > > If my above understanding is correct ^ why can't user use the feature with `bfdEnabled` which means if the serving pods are not healthy, then the heartbeats would fail right? So the ECMP routes are going to not be added automatically and cache wouldn't contain those entries right?
> > 
> > 
> > This was also asked by Tim in the downstream bug [here](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27194118&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27194118). Currently, the ECMP conntrack entries are not flushed as mentioned by [Lorenzo](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27194835&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27194835) and [Dumitru](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27197763&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27197763). Tim suggested [2 step approach](https://issues.redhat.com/browse/OCPBUGS-56202?focusedId=27203864&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27203864) for fixing the bug.
> 
> haha that still doesn't answer my question, so I know OVN's history around conntrack flushing. OVN doesn't do it today yes which is why we have #3040 and did the whole annotation for that to happen.. so its not about the conntrack flushing piece which I know I had opened https://bugzilla.redhat.com/show_bug.cgi?id=2096233 which then got reverted due to perf issues and now we have https://issues.redhat.com/browse/FDP-56 which we need to enable but for that we need https://issues.redhat.com/browse/OCPBUGS-49889 - so finally it is OVN-K doing the flush I am fine with that..
> 
> So I am not sure I understand why user can't use bfd so we would know the heart beats fail and ECMP routes would be removed right? This fix is not fixing conntrack right? It's trying to fix removal of routes + IPs from annotation? and removing that will auto-fix conntrack flush which will happen

oh sorry nvm I missed the point:
```
Also, power outage on a node should be a rare event. Seems like the 3rd option (using BFD) would help here. [Dumitru Ceara](https://issues.redhat.com/secure/ViewProfile.jspa?name=dceara%40redhat.com) [lorenzo bianconi](https://issues.redhat.com/secure/ViewProfile.jspa?name=lorenzobianconi) do you recall if when BFD goes down and the ecmp-symmetric route is flushed, do we also flush stored conntrack entries tracking the MAC?
```

so yes exactly this is my question as well.. so when BFDEnabled is set via OVN-K is the problem here that OVN-K itself doesn't know something failed and so we rely on OVN to do flushing and yes that bit I understand..

So IIUC -> even if BGP detects hop is down, and OVN deletes ECMP, flush doesn't happen via OVN and OVNK also doesnt remove the nexthop from its own annotation flush logic... ?

go-controller/pkg/ovn/controller/apbroute/external_controller.go
```diff
@@ -565,10 +565,13 @@ func (m *externalPolicyManager) onPodUpdate(oldObj, newObj interface{}) {
...
 		return
 	}
-	// if labels AND assigned Pod IPs AND the multus network status annotations are the same, skip processing changes to the pod.
+	// if labels AND assigned Pod IPs AND the multus network status annotations AND
+	// pod status conditions AND deletion timestamp are the same, skip processing changes to the pod.
 	if reflect.DeepEqual(o.Labels, n.Labels) &&
 		reflect.DeepEqual(o.Status.PodIPs, n.Status.PodIPs) &&
-		reflect.DeepEqual(o.Annotations[nettypes.NetworkStatusAnnot], n.Annotations[nettypes.NetworkStatusAnnot]) {
+		reflect.DeepEqual(o.Annotations[nettypes.NetworkStatusAnnot], n.Annotations[nettypes.NetworkStatusAnnot]) &&
+		reflect.DeepEqual(o.Status.Conditions, n.Status.Conditions) &&
```
tssurya: pod's conditions could fluctuate right? is this what is used to tell if the pod went unhealthy?
why did we add deletion timestamp here ? can you add a comment behind why you added the new match conditions?

go-controller/pkg/util/kube.go
```diff
@@ -738,6 +738,16 @@ func PodTerminating(pod *corev1.Pod) bool {
 	return pod.DeletionTimestamp != nil
 }
 
+// PodNotReady checks if the pod is not ready
+func PodNotReady(pod *corev1.Pod) bool {
```
tssurya: nit: like above utils go with `PodReady` and then do `!PodReady()` for better readability because otherwise we'd do `!PodNotReady` double negation which can be avoided 

and I'd call it `IsPodReady` ?

```
import "k8s.io/apimachinery/pkg/api/meta"

func IsPodReady(pod *corev1.Pod) bool {
    condition := meta.FindStatusCondition(pod.Status.Conditions, string(corev1.PodReady))
    return condition != nil && condition.Status == metav1.ConditionTrue
}
```
something like that?

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
```diff
@@ -28,6 +30,13 @@ func (m *externalPolicyManager) syncPod(pod *corev1.Pod, routeQueue workqueue.Ty
...
+		klog.Warningf("External gateway pod cannot serve traffic; it's in terminating or not ready state: %s", gatewayPod.Name)
```
tssurya: add the namespace name as well to this warning log - there would be pods with same names in the cluster in different namespaces

go-controller/pkg/ovn/egressgw.go
```diff
@@ -49,6 +49,13 @@ func (oc *DefaultNetworkController) addPodExternalGW(pod *corev1.Pod) error {
 
 	klog.Infof("External gateway pod: %s, detected for namespace(s) %s", pod.Name, podRoutingNamespaceAnno)
 
+	// If an external gateway pod is in terminating or not ready state then don't add the
+	// routes for the external gateway pod
+	if util.PodTerminating(pod) || util.PodNotReady(pod) {
+		klog.Warningf("External gateway pod cannot serve traffic; it's in terminating or not ready state: %s", pod.Name)
```
tssurya: same here add namespace name

go-controller/pkg/ovn/ovn.go
```diff
@@ -131,6 +135,14 @@ func (oc *DefaultNetworkController) ensurePod(oldPod, pod *corev1.Pod, addPort b
 		return oc.ensureRemotePodIP(oldPod, pod, addPort)
 	}
 
+	// If an external gateway pod is in terminating or not ready state then remove the
+	// routes for the external gateway pod
+	if util.PodTerminating(pod) || util.PodNotReady(pod) {
```
tssurya: and if it came back into ready state I guess it would hit line L185 and go onto to add things?

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
tssurya: i think this check should happen before getIPs is called? i.e similar to how its done inside addPodExternalGW ?
we shouldn't even call getIPs if we know we don't want to call it

go-controller/pkg/ovn/controller/apbroute/external_controller_pod_test.go
tssurya: yup was just about to ask this and saw peri already did :) nice!

go-controller/pkg/ovn/controller/apbroute/external_controller_pod_test.go
```diff
@@ -448,6 +448,163 @@ var _ = Describe("OVN External Gateway pod", func() {
...
+
+			if terminating {
+				By("Removing deletion timestamp for the ex gw pod")
+				setPodDeletionTimestamp(pod1, nil, fakeClient)
+			} else {
+				By("Updating the ex gw pod status to mark it as ready")
```
tssurya, line 586: ah this is where you bring the pod back to ready? i.e if terminating flag is false then we mark not ready and then ready here..?

go-controller/pkg/ovn/egressgw_test.go
```diff
@@ -1817,6 +1818,591 @@ var _ = ginkgo.Describe("OVN Egress Gateway Operations", func() {
 				},
 			),
 		)
+		ginkgo.DescribeTable("reconciles a host networked pod in terminating or not ready state acting as a exgw for another namespace for existing pod",
```
tssurya, line 1822: loved the tests :) well done!

go-controller/pkg/ovn/egressgw_test.go
```diff
@@ -1817,6 +1818,591 @@ var _ = ginkgo.Describe("OVN Egress Gateway Operations", func() {
...
+							"pod":       "true",
+							"namespace": namespaceName,
+						},
+						Name: "namespace1_myPod",
+						Options: map[string]string{
+							"iface-id-ver":      "myPod",
+							"requested-chassis": "node1",
+						},
+						PortSecurity: []string{"0a:58:0a:80:01:03 10.128.1.3"},
+					},
```
tssurya, line 1985: ack in the afterupdate DB we can see LRSR is gone

tssurya:
@coderabbitai review

tssurya (changes requested):

tssurya:
overall logic looks good to me I need to review the E2E's as well but @arkadeepsen amazing job with all the tests.. its so nice to see a PR which is test rich :)

go-controller/pkg/util/kube.go
```diff
@@ -738,6 +738,16 @@ func PodTerminating(pod *corev1.Pod) bool {
...
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type == corev1.PodReady {
+			return condition.Status == corev1.ConditionFalse
+		}
+	}
+	return false
```
tssurya: also if no ready condition is found.. then we are treating it as "it is ready?" is that true to assume it?

go-controller/pkg/util/kube.go
tssurya: I got this from kubernetes utils design:
```
"k8s.io/kubectl/pkg/util/podutils" which gives 

// IsPodReady returns true if a pod is ready; false otherwise.
func IsPodReady(pod *corev1.Pod) bool {
	return isPodReadyConditionTrue(pod.Status)
}
// IsPodReadyConditionTrue returns true if a pod is ready; false otherwise.
func isPodReadyConditionTrue(status corev1.PodStatus) bool {
	condition := getPodReadyCondition(status)
	return condition != nil && condition.Status == corev1.ConditionTrue
}

// GetPodReadyCondition extracts the pod ready condition from the given status and returns that.
// Returns nil if the condition is not present.
func getPodReadyCondition(status corev1.PodStatus) *corev1.PodCondition {
	_, condition := getPodCondition(&status, corev1.PodReady)
	return condition
}
```

go-controller/pkg/ovn/controller/apbroute/external_controller.go
arkadeepsen: I have added the reason for the new matches. PLMK if I should add more details.

go-controller/pkg/util/kube.go
arkadeepsen: I have removed the function altogether. Using `IsPodReadyConditionTrue()` from `k8s.io/kubernetes/pkg/api/v1/pod`.

go-controller/pkg/ovn/ovn.go
arkadeepsen: Yes.

go-controller/pkg/util/kube.go
arkadeepsen: Removed the function altogether.

go-controller/pkg/ovn/controller/apbroute/external_controller_pod.go
arkadeepsen: The function `getExGwPodIPs` is being used in multiple places. That's why added the check inside the function to reduce the duplication of code.

arkadeepsen:
> overall logic looks good to me I need to review the E2E's as well but @arkadeepsen amazing job with all the tests.. its so nice to see a PR which is test rich :)

@tssurya I have incorporated the suggestions by coderabbit for the e2e tests. Got a green signal in the CI as well. PTAL.

go-controller/pkg/ovn/controller/apbroute/external_controller.go
tssurya: I have one last question here.. shouldn't we be explicitly then checking if the podready status changed? checking the whole of status conditions means we could get a lot more noise than we actually care about? so i'd rather you check on pod ready v/s not ready

tssurya: so it should be podBecameReady || podBecameNotReady

go-controller/vendor/k8s.io/kubernetes/pkg/api/v1/pod/util.go
```diff
@@ -0,0 +1,418 @@
...
+	return false
+}
+
+// IsPodReady returns true if a pod is ready; false otherwise.
+func IsPodReady(pod *v1.Pod) bool {
+	return IsPodReadyConditionTrue(pod.Status)
+}
+
+// IsPodTerminal returns true if a pod is terminal, all containers are stopped and cannot ever regress.
+func IsPodTerminal(pod *v1.Pod) bool {
```
tssurya, line 323: I am OK to get this vendored code in.. because the file looks small..
but in the past as you can see we have preferred to copy over the utils we want.. - to reduce image size
hope you ran go mod tidy etc?
also if we bring this file in this file i think then we can stop having copies of these utils in our own kube utils file

arkadeepsen: > - to reduce image size
> hope you ran go mod tidy etc?

Yes. Ran both `go mod tidy` and `go mod vendor`.



> also if we bring this file in this file i think then we can stop having copies of these utils in our own kube utils file

I checked the file and in this PR we don't need any other utility function. The `IsPodTerminal()` checks if the pod is already in a terminal state, whereas we need to check whether the pod is currently in a terminating state or not using the `DeletionTimestamp` field, which we're getting from the `PodTerminating()` function.

go-controller/pkg/ovn/controller/apbroute/external_controller.go
arkadeepsen: Ack. Will update the code to only reconcile the pod when the ready status changes.

arkadeepsen: Updated the code to check the changes in PodReady status condition.

test/e2e/external_gateways.go
```diff
@@ -3595,3 +3624,133 @@ func resetGatewayAnnotations(f *framework.Framework) {
...
+		gomega.Eventually(func() bool {
+			p, err = f.ClientSet.CoreV1().Pods(servingNamespace).Get(context.Background(), gatewayPodName, metav1.GetOptions{})
+			if err != nil {
+				return false
+			}
+			return strings.Contains(strings.Join(p.GetFinalizers(), ","), "k8s.ovn.org/external-gw-pod-finalizer")
+		}).Should(gomega.Equal(true), fmt.Sprintf("Update second external gateway pod %s from ns %s with finalizer, failed: %v", gatewayPodName, servingNamespace, err))
+
+		p = getGatewayPod(f, servingNamespace, gatewayPodName)
+		err = e2epod.DeletePodWithGracePeriod(context.Background(), f.ClientSet, p, 1000)
```
tssurya, line 3715: why 1000? - just to trigger a deletion timestamp being set and ensuring we react to it?

tssurya (approved):

test/e2e/external_gateways.go
arkadeepsen: Yes. Kept a large value so that the e2e test can work correctly.

PR #5301, opened 2025-06-12
openflow_manager: Allow critical host traffic in DPU mode

aserdean:
In DPU mode, add specific OpenFlow rules to allow traffic to the host stack.

This includes adding high-priority rules for DHCP (UDP/67) and for IP traffic destined to the node's , ensuring the physical function remains accessible.

Trying an ASCII representation of the bridge layout

```
                                                     PF
        Host                                          |
         ------------------------------------------------
        DPU                                           |
                                               PF_repr_plugged_in_br_ovn
                                                      |
                                DHCP server LOCAL---br-ovn 
                                       

```

## Summary by CodeRabbit

* **Documentation**
  * Added detailed ASCII diagrams to illustrate DPU support configurations and DHCP serving scenarios in the hardware offload documentation.

* **Bug Fixes**
  * Improved handling of network traffic flows for DPU mode, ensuring correct prioritization for DHCP server and local tunnel endpoint traffic.


aserdean:
image
ASCII art did not cut it added an extra image

aserdean:
The extra actions are due to the DHCP server, while not impacting the table negatively it implies something "custom"

Maybe this could be a future feature to allow users to add **custom flow actions** on the managed ovnkube bridge via Kubernetes API.


CC: @girishmg @trozet

npinaeva:
> This could be a future feature to allow users to add **custom flow actions** on the managed ovnkube bridge via Kubernetes API.
> 
> CC: @girishmg @trozet

I think something along those lines could be useful as a comment in the code too

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+			return err
+		}
+		encapIP = strings.TrimSuffix(encapIP, "\n")
+
+		// The extra actions, compared to the full mode, are due to a DHCP server on the bridge.
+		// While not impacting the other flows negatively it implies something "custom"
+		// this could be a future feature to allow users to add custom flow actions on the managed ovnkube bridge via Kubernetes API.
+		c.updateFlowCacheEntry("NORMAL", []string{
+			// Add normal action on table 0 (act as a switch)
+			fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction),
```
trozet, line 267: you dont need \n on the end of the flows

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+			"get",
+			"Open_vSwitch",
+			".",
+			"external_ids:ovn-encap-ip",
+		}
+		encapIP, _, err := util.RunOVSVsctl(checkCmd...)
+		if err != nil {
+			return err
+		}
+		encapIP = strings.TrimSuffix(encapIP, "\n")
```
trozet: nit: TrimSpace

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+
+		// The extra actions, compared to the full mode, are due to a DHCP server on the bridge.
```
trozet, line 262: so in DPU mode are you connecting a DHCP server to the ovs bridge to serve DHCP to the host? and that is going via br-ovn linux interface?  what IP address does br-ovn have?

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+		// this could be a future feature to allow users to add custom flow actions on the managed ovnkube bridge via Kubernetes API.
+		c.updateFlowCacheEntry("NORMAL", []string{
+			// Add normal action on table 0 (act as a switch)
+			fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction),
+			// Add OpenFlow action with high priority that allows a normal action for DHCP servers
+			// that are serving on the ovnkube managed bridge.
+			fmt.Sprintf("table=0,priority=999,udp,tp_dst=67,actions=%s\n", util.NormalAction),
+			// Add OpenFlow action with high priority that allows the PF (physical function) to the
+			// local_ip.
+			fmt.Sprintf("table=0,priority=998,ip,nw_dst=%s,actions=%s\n", encapIP, util.NormalAction),
```
trozet, line 273: it's not clear to me who's IP the encap IP is. Is it the host PF? or the br-ovn DPU? The first flow says if it is a packet sent to a dhcp server then normal action. The second flow says if its a packet to the encap IP normal action. It's not making sense to me what the purpose of the 2nd flow is.

Maybe putting some IPs in your diagram and showing which one is which, and the analogy to br-ex in a regular deployment will help me understand.

trozet: Note we could also update the same DPU markdown file with your diagram and explaining the DHCP server, what it is used for and where the IP addresses live for encap, what flows are added and why they are needed. Similar to:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/design/service-traffic-policy.md#local-gateway-mode

aserdean: `it's not clear to me who's IP the encap IP is. Is it the host PF? or the br-ovn DPU? `
br-ovn
`The second flow says if its a packet to the encap IP normal action. It's not making sense to me what the purpose of the 2nd flow is.`
The issue is when I send traffic from the PF to the VTEP w/o this action it will go out the uplink instead of going to br-ovn

will add a picture to try to make it clearer.

go-controller/pkg/node/openflow_manager.go
aserdean: `so in DPU mode are you connecting a DHCP server to the ovs bridge to serve DHCP to the host?`
To serve the PF on the host
`and that is going via br-ovn linux interface?`
Yes
`what IP address does br-ovn have?`
An IP on the same subnet as the PF will, but hopefully the diagram will make things clearer.

go-controller/pkg/node/openflow_manager.go
aserdean: why do we originally set it?
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/4de8fd5f958956874bce147e01f11bd0049abcd9/go-controller/pkg/node/openflow_manager.go#L248

go-controller/pkg/node/openflow_manager.go
aserdean: image
ping from the host to `10.0.120.1` will not work unless I have the second OpenFlow flow

aserdean: added diagrams on docs via ASCII art.

my mermaid skills are not good enough :)

go-controller/pkg/node/openflow_manager.go
trozet: I think it is a bug

docs/features/hardware-offload/dpu-support.md
```diff
@@ -14,6 +14,118 @@ Similarly to Smart-NICs, a DPU follows the kernel switchdev model.
...
+│                                                                                      │                    │
+└──────────────────────────────────────────────────────────────────────────────────────┼────────────────────┘
+                                                                                       │                     
+┌──────────────────────────────────────────────────────────────────────────────────────┼────────────────────┐
+│DPU                                                                                   │                    │
+│                                                                                      ▼                    │
+│     ┌────────────────────┐           ┌──────────────────────────────────────────┬──────────┬────────┐     │
+│     │                    │           │                                          │pf0hpf_rep│        │     │
+│     │                    │           │ ┌─────────────────────────────────────┐  └──────────┘        │     │
+│     │                    │           │ │    DHCP Server Serving on br-ex     │                      │     │
```
trozet: This DHCP server is not really in or on br-ex. It really exists as a host dnsmasq process that interacts vai the br-ex LOCAL port. Can we change the picture to depict that?

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+
+		// The extra actions, compared to the full mode, are due to a DHCP server on the bridge.
+		// While not impacting the other flows negatively it implies something "custom"
+		// this could be a future feature to allow users to add custom flow actions on the managed ovnkube bridge via Kubernetes API.
+		c.updateFlowCacheEntry("NORMAL", []string{
+			// Add normal action on table 0 (act as a switch)
+			fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction),
+			// Add OpenFlow action with high priority that allows a normal action for DHCP servers
+			// that are serving on the ovnkube managed bridge.
+			fmt.Sprintf("table=0,priority=999,udp,tp_dst=67,actions=%s\n", util.NormalAction),
```
trozet, line 270: So I think the reason you need this flow is because in br-ex when we see IP traffic, we send to table 1 for IP handling. In that table we do not handle broadcast traffic. My understanding of why you are adding the flow is 1 or more clients on the br-ex bridge will send DHCP discover/request to 255.255.255.255:67. I think it would make more sense to add a generic flow that matches on 255.255.255.255 and does a normal action. This should be enabled in table 1, and should not depend on DPU.

I could see a case where a localnet port VM tries to DHCP Request, and the DHCP server is running on the host. This would also solve that case I think. wdyt @ricky-rav ?

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+			// Add OpenFlow action with high priority that allows the PF (physical function) to the
+			// local_ip.
+			fmt.Sprintf("table=0,priority=998,ip,nw_dst=%s,actions=%s\n", encapIP, util.NormalAction),
```
trozet, line 273: So with this flow you are trying to ensure that packets from the PF destined to br-ovn IP can be sent. I think you should set action to be LOCAL.

Just to make sure I'm on the same page here. The node IP for the host k8s node is the PF IP right?

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
 	}
 	dftFlows = append(dftFlows, dftCommonFlows...)
 
-	c.updateFlowCacheEntry("NORMAL", []string{fmt.Sprintf("table=0,priority=0,actions=%s\n", util.NormalAction)})
+	if config.OvnKubeNode.Mode == types.NodeModeDPU {
+		// Retrieve the local_ip for the tunnel endpoint which will be used to create OpenFlow actions
```
trozet, line 249: I think you should not use the encap IP here. This implies that the encap IP is the same as the br-ovn IP. Which may not always be the case. I think you should just get the IP of the br-ovn interface.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+		}
+		encapIP, _, err := util.RunOVSVsctl(checkCmd...)
+		if err != nil {
+			return err
+		}
+		encapIP = strings.TrimSpace(encapIP)
+
+		// The extra actions, compared to the full mode, are due to a DHCP server on the bridge.
+		// While not impacting the other flows negatively it implies something "custom"
+		// this could be a future feature to allow users to add custom flow actions on the managed ovnkube bridge via Kubernetes API.
```
trozet, line 264: I would remove these comments. I think the behavior of allowing broadcast to 255.255.255.255 is generic and legit in any scenario and is not "custom".

For the second flow, I think that one is DPU specific where your gateway uplink interface is not the same as LOCAL. I can see that being common across all DPUs, so we can consider that not "custom", and just part of the DPU design.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -245,7 +245,38 @@ func (c *openflowManager) updateBridgeFlowCache(hostIPs []net.IP, hostSubnets []
...
+		c.updateFlowCacheEntry("NORMAL", []string{
```
trozet, line 265: I would maybe use they key DPU here and it is only going to have one flow for now to allow IP traffic destined to LOCAL to go straight there from any interface.

trozet (changes requested):

go-controller/pkg/node/openflow_manager.go
aserdean: fair enough

go-controller/pkg/node/openflow_manager.go
aserdean: for this particular case I agree, but as discussed offline this is something that we might need in the future

docs/features/hardware-offload/dpu-support.md
aserdean: sure will do so soon

go-controller/pkg/node/openflow_manager.go
aserdean: I am not sure I follow, I think we still the original normal action: 
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5301/files#diff-997669eacbce30bde27feb1ebccea9ca61a85d01c3d4647c1ab1ab405cbfbddcL248 right?

Should that be common and the additional needed flow(s) should be under `c.updateFlowCacheEntry("DPU", []string{` ?

go-controller/pkg/node/openflow_manager.go
aserdean: Ack, should I remove it?

go-controller/pkg/node/openflow_manager.go
aserdean: Just to be sure we are on the same page.

With the current proposed flows, we validated both usecases (local DHCP server and external DHCP server) and also that all traffic patterns are hardware offloaded.

Now that being said, if we change the proposed flows this will require significant effort to retest and revalidate the results.

Is there a document which explains the pipeline of the ovnk managed bridge with different features?

go-controller/pkg/node/openflow_manager.go
aserdean: There is some method behind the NORMAL madness, if enabled, it will help with the MAC learning (FDB entries) on the bridge.

There are good parts about that, it can aid us i.e. for the ARP responder needed for the VTEP intf but also it has its pitfalls i.e.: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5318

Setting to LOCAL, could work to enable the traffic pattern described above, but I don't know the full impact this will entail because we will use those learned FDB entries.

Just as a reminder in my case I have **two** node IPs, lets name the first one for ease `eth0` and the second the PF `pf0hpf`.

`eth0` - will have the K8s node IP and is our default gw
`pf0hpf` - will be represent our highspeed network

girishmg:
Long term we should move to the two OVS bridge model that we were discussing downstream, @aserdean. Something like this below:

This should fix the various issues we are seeing with Shared gateway mode when the Host and OVN pipeline are stepping on each other toes.

```
                                   +----+         +----+ +----+             
                                   | PF |         |    | |    |             
                                   +--+-+         |Pod1| |Pod2|             
                                      |           |    | |    |             
                                      |           +--+-+ +-+--+             
                                      |              |     |        Host    
               -----------------------+--------------+-----+-------------   
                                      |              |     |        DPU     
                                      |              |     |                
                                      |            +-+--+ ++---+            
                                      |           ++Rep +-+Rep +--+         
                                      |           |+----+ +----+  |         
                                      |           |    br-int     |         
                                      |           +----+----------+         
                                      |              patch-port             
                                      |                |                    
                                    +-+--+             |                    
                                  +-+Rep +-------------+----+               
                                  | +----+  br-ovn          |               
+-----------+                     | (OVN-K8s manages stuff) |               
|DHCP-Server|                     |                         |               
+-----------+                     +------+------------------+               
   DPU-IP    VTEP-IP                     |                                  
       ^         ^                  patch-port                              
       |         |                       |                                  
     +-+---------+-----------------------+----+                             
     |            br-dpu (NORMAL)             |                             
     +---------------------------+------------+                             
                                 |                                          
                              uplink                                        
                                 |                                          
                                 v                                          

```

aserdean:
> Long term we should move to the two OVS bridge model that we were discussing downstream, @aserdean. Something like this below:
> 
> This should fix the various issues we are seeing with Shared gateway mode when the Host and OVN pipeline are stepping on each other toes.
> 
> ```
>                                    +----+         +----+ +----+             
>                                    | PF |         |    | |    |             
>                                    +--+-+         |Pod1| |Pod2|             
>                                       |           |    | |    |             
>                                       |           +--+-+ +-+--+             
>                                       |              |     |        Host    
>                -----------------------+--------------+-----+-------------   
>                                       |              |     |        DPU     
>                                       |              |     |                
>                                       |            +-+--+ ++---+            
>                                       |           ++Rep +-+Rep +--+         
>                                       |           |+----+ +----+  |         
>                                       |           |    br-int     |         
>                                       |           +----+----------+         
>                                       |              patch-port             
>                                       |                |                    
>                                     +-+--+             |                    
>                                   +-+Rep +-------------+----+               
>                                   | +----+  br-ovn          |               
> +-----------+                     | (OVN-K8s manages stuff) |               
> |DHCP-Server|                     |                         |               
> +-----------+                     +------+------------------+               
>    DPU-IP    VTEP-IP                     |                                  
>        ^         ^                  patch-port                              
>        |         |                       |                                  
>      +-+---------+-----------------------+----+                             
>      |            br-dpu (NORMAL)             |                             
>      +---------------------------+------------+                             
>                                  |                                          
>                               uplink                                        
>                                  |                                          
>                                  v                                          
> ```

@girishmg I agree, this is something that we are looking for in phase 2 where we will have two ovnks on the DPU

docs/features/hardware-offload/dpu-support.md
aserdean: done

go-controller/pkg/node/openflow_manager.go
trozet: We can add e2e tests to ensure your use case doesn't break in the future, and then you dont need to test anything manually.

1. For external DHCP traffic, when IP packets come into the node we always send to table 1 (as long as they dont match k8s service):

```
cookie=0xdeff105, duration=139.971s, table=0, n_packets=33504, n_bytes=145629806, idle_age=0, priority=50,ip,in_port=1 actions=ct(table=1,zone=64000,nat)
```

here in table 1 we process IP traffic. We first check if this is reply traffic, if so send it to the corresponding port. Otherwise if it is pod address traffic, send it to OVN. Geneve traffic, send it local. Destination mac== shared mac, send it local. Otherwise, send it action NORMAL.

This why your external DHCP server just works without any flow changes. It hits this last normal flow.

We can test this by creating a localnet VM, crafting a DHCP offer externally and sending it, and making sure it arrives at the localnet VM. There may even already be a test for this.

2. For internal server, in table 1 we can add a flow targeting 255.255.255.255, action FLOOD. I was thinking FLOOD will be better than NORMAL, because NORMAL incurs some CPU overhead in OVS.

Then we can create a test where we send a dhcp discover message from a localnet VM and ensure it gets sent to the LOCAL port. Or vice versa, we can send a dhcp discover from the LOCAL port and ensure the localnet VM receives it.

There is only one document that I know that explains the flows for services:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/design/service-traffic-policy.md

I'll try to leverage AI to see if it can write a generic br-ex openflow pipeline doc for me :)

aserdean: > We can add e2e tests to ensure your use case doesn't break in the future, and then you dont need to test anything manually.

I don't think we are aligned.
What I am trying to emphasize is that with the current proposed OpenFlow changes **we know** traffic is properly offloaded, if we change the pipeline or diverge from these changes, **we would have to revalidate** that the same traffic patterns are still properly offloaded.

Adding the E2E would be great to ensure the traffic pattern for DHCP will work, but in order to validate the offloading bits it requires a CI which we currently do not have.

go-controller/pkg/node/openflow_manager.go
trozet: I posted a fix for 5318. Thanks for bringing it to my attention. IMO LOCAL should be used for the IP flow. There should be other flows that handle ARP/L2. But regardless, I would think https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5313 would fix this part. I need to look at it tmrw when I'm fresh.

go-controller/pkg/node/openflow_manager.go
trozet: oh yeah sorry, i commented on the wrong line. The additional DPU flow should be like you mentioned.

go-controller/pkg/node/openflow_manager.go
trozet: sure

go-controller/pkg/node/openflow_manager.go
trozet: I get it and feel your pain, but that is a bigger problem than just this PR. We change flows in the project and we don't know when we break offload due to lack of coverage. It's also hard for me to review for something I cant setup myself and inspect flows/traffic path.

I try to consider offload when flow changes happen and predict whether or not it will break offload. 

Looking at your flow again it really isn't obvious to me why it is needed. I think I was wrong about table 1 here. The packet is coming from LOCAL, but in your case LOCAL is not the ofPortHost.. the pf0hpf_rep is. That would mean the packet would hit:

```
cookie=0x0, duration=3337.722s, table=0, n_packets=18463, n_bytes=49618677, priority=0 actions=NORMAL
```
which is essentially the same as your flow.

Maybe we can do a debugging session or something and we can trace the packet to see what is going on.

aserdean: > I get it and feel your pain, but that is a bigger problem than just this PR. We change flows in the project and we don't know when we break offload due to lack of coverage. It's also hard for me to review for something I cant setup myself and inspect flows/traffic path.

Since that CI will not be available immediately, I am reweighing things and I think @girishmg is right, maybe we should bite the bullet and go with a 3 bridge layout since that would simplify the anomalies that can occur.

Will try to see what that entails on our side and test it out.

> That would mean the packet would hit:

Was anything merged in the last two weeks that might fix that traffic pattern?

Either way, will rebase and use ofproto/trace with a raw packet to see what flows are hit for completeness.

trozet: yes I think there are a few things that will fix your traffic pattern:
1. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5313
2. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5334
3. One more PR to match broadcast traffic and flood it.
4. The earlier PRs may not take into account that in DPU mode, ofportHost should be used and not hardcode LOCAL. Might need to fix this.

I'll be on PTO next week, but when I get back I can probably just deploy kind and add a dummy interface to ovs as my "PF" and then run ovnk in DPU mode. Then I can see the flows and I will know what we need to fix. If you could also send me a ovs-ofctl dump-flows br-ovn on your side, or give me tmate access to a cluster when I get back that would also be helpful.

aserdean:
> > Long term we should move to the two OVS bridge model that we were discussing downstream, @aserdean. Something like this below:
> > This should fix the various issues we are seeing with Shared gateway mode when the Host and OVN pipeline are stepping on each other toes.
> > ```
> >                                    +----+         +----+ +----+             
> >                                    | PF |         |    | |    |             
> >                                    +--+-+         |Pod1| |Pod2|             
> >                                       |           |    | |    |             
> >                                       |           +--+-+ +-+--+             
> >                                       |              |     |        Host    
> >                -----------------------+--------------+-----+-------------   
> >                                       |              |     |        DPU     
> >                                       |              |     |                
> >                                       |            +-+--+ ++---+            
> >                                       |           ++Rep +-+Rep +--+         
> >                                       |           |+----+ +----+  |         
> >                                       |           |    br-int     |         
> >                                       |           +----+----------+         
> >                                       |              patch-port             
> >                                       |                |                    
> >                                     +-+--+             |                    
> >                                   +-+Rep +-------------+----+               
> >                                   | +----+  br-ovn          |               
> > +-----------+                     | (OVN-K8s manages stuff) |               
> > |DHCP-Server|                     |                         |               
> > +-----------+                     +------+------------------+               
> >    DPU-IP    VTEP-IP                     |                                  
> >        ^         ^                  patch-port                              
> >        |         |                       |                                  
> >      +-+---------+-----------------------+----+                             
> >      |            br-dpu (NORMAL)             |                             
> >      +---------------------------+------------+                             
> >                                  |                                          
> >                               uplink                                        
> >                                  |                                          
> >                                  v                                          
> > ```
> 
> @girishmg I agree, this is something that we are looking for in phase 2 where we will have two ovnks on the DPU

@girishmg @trozet @npinaeva : I am abandoning this PR because I managed to successfully implement the design above in our product DPF and tested it successfully.

⚠️ There is a caveat with this solution: br-ovn in the diagram must have an IP and unfortunately it can't be a link local address. 

I might need to disable this check : 
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/release-1.0/go-controller/pkg/node/gateway_init.go#L349-L352
since w/o an IP the following will be triggered:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/release-1.0/go-controller/pkg/node/gateway_init.go#L108-L112 

when ovnk is running in DPU mode.

PR #5302, opened 2025-06-12
fix udn e2e test for multiple cidr ranges

flavio-fernandes:
The network segmentation test now iterates through each configured CIDR and ensures at least one pod IP of the same address family falls within it. This validates that pod IPs match the configured ranges correctly.


flavio-fernandes:
abort: this is not the bots you are looking for. :)

PR #5307, opened 2025-06-15
gateway: Refactor gateway initialization and DPU host handling

aserdean:
- Improve DPU host gateway configuration with better IP handling
- Fix node interface address handling for DPU mode
- Clean up gateway interface selection logic for DPU and DPU-HOST
- Update node annotations for DPU host mode
- Remove redundant code and improve error handling
- Fix typos and improve code readability

This change simplifies the gateway initialization flow and improves the handling of DPU (Data Processing Unit) host configurations.

The main idea is that the DPU in host mode is the one that dictates how the network looks and the DPU side just follows.

### DIFF
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -1028,7 +1028,7 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
 
 	// Complete gateway initialization
 	if config.OvnKubeNode.Mode == types.NodeModeDPUHost {
-		err = nc.initGatewayDPUHost(nc.nodeAddress)
+		err = nc.initGatewayDPUHost(nc.nodeAddress, nodeAnnotator)
 		if err != nil {
 			return err
 		}
--- a/go-controller/pkg/node/gateway_init.go
+++ b/go-controller/pkg/node/gateway_init.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/vishvananda/netlink"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 
@@ -324,7 +326,7 @@ func (nc *DefaultNodeNetworkController) initGatewayPreStart(
 	subnets []*net.IPNet,
 	nodeAnnotator kube.Annotator,
 	mgmtPort managementport.Interface,
-	kubeNodeIP net.IP,
+	_ net.IP,
 ) (*gateway, error) {
 
 	klog.Info("Initializing Gateway Functionality for Gateway PreStart")
@@ -351,7 +353,19 @@ func (nc *DefaultNodeNetworkController) initGatewayPreStart(
 	// For DPU need to use the host IP addr which currently is assumed to be K8s Node cluster
 	// internal IP address.
 	if config.OvnKubeNode.Mode == types.NodeModeDPU {
-		ifAddrs, err = getDPUHostPrimaryIPAddresses(kubeNodeIP, ifAddrs)
+		var node *corev1.Node
+		if node, err = nc.watchFactory.GetNode(nc.name); err != nil {
+			return nil, fmt.Errorf("error retrieving node %s: %v", nc.name, err)
+		}
+		nodeAddrStr, err := util.GetNodeIfAddrAnnotation(node)
+		if err != nil {
+			return nil, err
+		}
+		nodeAddr, _, err := net.ParseCIDR(nodeAddrStr.IPv4)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse node IP address. %v", nodeAddrStr)
+		}
+		ifAddrs, err = getDPUHostPrimaryIPAddresses(nodeAddr, ifAddrs)
 		if err != nil {
 			return nil, err
 		}
@@ -488,22 +502,22 @@ func interfaceForEXGW(intfName string) string {
 	return intfName
 }
 
-func (nc *DefaultNodeNetworkController) initGatewayDPUHost(kubeNodeIP net.IP) error {
+func (nc *DefaultNodeNetworkController) initGatewayDPUHost(kubeNodeIP net.IP, nodeAnnotator kube.Annotator) error {
 	// A DPU host gateway is complementary to the shared gateway running
 	// on the DPU embedded CPU. it performs some initializations and
 	// watch on services for iptable rule updates and run a loadBalancerHealth checker
 	// Note: all K8s Node related annotations are handled from DPU.
 	klog.Info("Initializing Shared Gateway Functionality on DPU host")
 	var err error
+	var gwIntf, gatewayIntf string
+	gatewayIntfTemp := config.Gateway.Interface
 
-	// Force gateway interface to be the interface associated with kubeNodeIP
-	gwIntf, err := getInterfaceByIP(kubeNodeIP)
+	gwIntf, err = getInterfaceByIP(kubeNodeIP)
 	if err != nil {
 		return err
 	}
 	config.Gateway.Interface = gwIntf
-
-	_, gatewayIntf, err := getGatewayNextHops()
+	_, gatewayIntf, err = getGatewayNextHops()
 	if err != nil {
 		return err
 	}
@@ -511,6 +525,37 @@ func (nc *DefaultNodeNetworkController) initGatewayDPUHost(kubeNodeIP net.IP) er
 	ifAddrs, err := getNetworkInterfaceIPAddresses(gatewayIntf)
 	if err != nil {
 		return err
+	}
+
+	nodeIPNetv4, _ := util.MatchFirstIPNetFamily(false, ifAddrs)
+	nodeAddrSet := sets.New[string](nodeIPNetv4.String())
+
+	if gatewayIntfTemp != "" && gatewayIntfTemp != gatewayIntf {
+		config.Gateway.Interface = gatewayIntfTemp
+		gatewayIntf = config.Gateway.Interface
+		gwIntf = gatewayIntf
+
+		ifAddrs, err = getNetworkInterfaceIPAddresses(gatewayIntf)
+		if err != nil {
+			return err
+		}
+
+		nodeIPNetv4, _ = util.MatchFirstIPNetFamily(false, ifAddrs)
+		nodeAddrSet.Insert(nodeIPNetv4.String())
+	}
+
+	if err := util.SetNodePrimaryIfAddrs(nodeAnnotator, ifAddrs); err != nil {
+		klog.Errorf("Unable to set primary IP net label on node, err: %v", err)
+		return err
+	}
+
+	if err := util.SetNodeHostCIDRs(nodeAnnotator, nodeAddrSet); err != nil {
+		klog.Errorf("Unable to set host-cidrs on node, err: %v", err)
+		return err
+	}
+
+	if err := nodeAnnotator.Run(); err != nil {
+		return fmt.Errorf("failed to set node %s annotations: %w", nc.name, err)
 	}
 
 	// Delete stale masquerade resources if there are any. This is to make sure that there
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -2445,6 +2445,9 @@ func newGateway(
 			gwBridge.eipMarkIPs = gw.bridgeEIPAddrManager.GetCache()
 		}
 		gw.nodeIPManager = newAddressManager(nodeName, kube, mgmtPort, watchFactory, gwBridge)
+		if gw.nodeIPManager == nil {
+			return fmt.Errorf("failed to allocate newAddressManager")
+		}
 
 		if config.OvnKubeNode.Mode == types.NodeModeFull {
 			// Delete stale masquerade resources if there are any. This is to make sure that there
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -317,12 +317,8 @@ func (c *addressManager) updateNodeAddressAnnotations() error {
 
 func (c *addressManager) updateHostCIDRs(ifAddrs []*net.IPNet) error {
 	if config.OvnKubeNode.Mode == types.NodeModeDPU {
-		// For DPU mode, here we need to use the DPU host's IP address which is the tenant cluster's
-		// host internal IP address instead.
-		// Currently we are only intentionally supporting IPv4 for DPU here.
-		nodeIPNetv4, _ := util.MatchFirstIPNetFamily(false, ifAddrs)
-		nodeAddrSet := sets.New[string](nodeIPNetv4.String())
-		return util.SetNodeHostCIDRs(c.nodeAnnotator, nodeAddrSet)
+		// The host CIDR should be updated only by the full and host mode
+		return nil
 	}
 
 	return util.SetNodeHostCIDRs(c.nodeAnnotator, c.cidrs)
--- a/go-controller/pkg/ovn/controller/services/lb_config.go
+++ b/go-controller/pkg/ovn/controller/services/lb_config.go
@@ -77,6 +77,7 @@ func makeNodeSwitchTargetIPs(node string, c *lbConfig) (targetIPsV4, targetIPsV6
 func makeNodeRouterTargetIPs(node *nodeInfo, c *lbConfig, hostMasqueradeIPV4, hostMasqueradeIPV6 string) (targetIPsV4, targetIPsV6 []string, v4Changed, v6Changed bool) {
 	targetIPsV4 = c.clusterEndpoints.V4IPs
 	targetIPsV6 = c.clusterEndpoints.V6IPs
+	var v4Updated, v6Updated bool
 
 	if c.externalTrafficLocal {
 		// For ExternalTrafficPolicy=Local, remove non-local endpoints from the router/switch targets
@@ -92,9 +93,15 @@ func makeNodeRouterTargetIPs(node *nodeInfo, c *lbConfig, hostMasqueradeIPV4, ho
 	}
 
 	// Any targets local to the node need to have a special
-	// harpin IP added, but only for the router LB
-	targetIPsV4, v4Updated := util.UpdateIPsSlice(targetIPsV4, node.l3gatewayAddressesStr(), []string{hostMasqueradeIPV4})
-	targetIPsV6, v6Updated := util.UpdateIPsSlice(targetIPsV6, node.l3gatewayAddressesStr(), []string{hostMasqueradeIPV6})
+	// hairpin IP added, but only for the router LB
+	if config.OvnKubeNode.Mode == types.NodeModeDPU {
+		// hairpin traffic of the hostaddress in case of the DPU
+		targetIPsV4, v4Updated = util.UpdateIPsSlice(targetIPsV4, node.hostAddressesStr(), []string{hostMasqueradeIPV4})
+		targetIPsV6, v6Updated = util.UpdateIPsSlice(targetIPsV6, node.hostAddressesStr(), []string{hostMasqueradeIPV6})
+	} else {
+		targetIPsV4, v4Updated = util.UpdateIPsSlice(targetIPsV4, node.l3gatewayAddressesStr(), []string{hostMasqueradeIPV4})
+		targetIPsV6, v6Updated = util.UpdateIPsSlice(targetIPsV6, node.l3gatewayAddressesStr(), []string{hostMasqueradeIPV6})
+	}
 
 	// Local endpoints are a subset of cluster endpoints, so it is enough to compare their length
 	v4Changed = len(targetIPsV4) != len(c.clusterEndpoints.V4IPs) || v4Updated
--- a/go-controller/pkg/ovn/default_network_controller.go
+++ b/go-controller/pkg/ovn/default_network_controller.go
@@ -1006,10 +1006,16 @@ func (h *defaultNetworkControllerEventHandler) UpdateResource(oldObj, newObj int
 			// Also check if node subnet changed, so static routes are properly set
 			// Also check if the node is used to be a hybrid overlay node
 			syncZoneIC = syncZoneIC || h.oc.isLocalZoneNode(oldNode) || nodeSubnetChange || zoneClusterChanged ||
-				switchToOvnNode || nodeEncapIPsChanged
+				switchToOvnNode || nodeEncapIPsChanged || primaryAddrChanged(oldNode, newNode)
 			if syncZoneIC {
 				klog.Infof("Node %q in remote zone %q, network %q, needs interconnect zone sync up. Zone cluster changed: %v",
 					newNode.Name, util.GetNodeZone(newNode), h.oc.GetNetworkName(), zoneClusterChanged)
+			}
+			if nodeChassisChanged(oldNode, newNode) {
+				if err := h.oc.zoneChassisHandler.DeleteRemoteZoneNode(oldNode); err != nil {
+					aggregatedErrors = append(aggregatedErrors, err)
+				}
+				syncZoneIC = true
 			}
 			if err := h.oc.addUpdateRemoteNodeEvent(newNode, syncZoneIC); err != nil {
 				aggregatedErrors = append(aggregatedErrors, err)
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -563,26 +563,29 @@ func (b *bridgeConfiguration) getGatewayIface() string {
 func (b *bridgeConfiguration) updateInterfaceIPAddresses(node *corev1.Node) ([]*net.IPNet, error) {
 	b.Lock()
 	defer b.Unlock()
-	ifAddrs, err := getNetworkInterfaceIPAddresses(b.getGatewayIface())
-	if err != nil {
-		return nil, err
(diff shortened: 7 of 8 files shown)


aserdean:
cc: @trozet @npinaeva

go-controller/pkg/node/gateway_init.go
```diff
@@ -324,7 +326,7 @@ func (nc *DefaultNodeNetworkController) initGatewayPreStart(
 	subnets []*net.IPNet,
 	nodeAnnotator kube.Annotator,
 	mgmtPort managementport.Interface,
-	kubeNodeIP net.IP,
+	_ net.IP,
```
npinaeva: don't you want to just remove this parameter?

go-controller/pkg/node/gateway_init.go
```diff
@@ -513,6 +527,37 @@ func (nc *DefaultNodeNetworkController) initGatewayDPUHost(kubeNodeIP net.IP) er
 		return err
 	}
 
+	nodeIPNetv4, _ := util.MatchFirstIPNetFamily(false, ifAddrs)
+	nodeAddrSet := sets.New[string](nodeIPNetv4.String())
+
+	if gatewayIntfTemp != "" && gatewayIntfTemp != gatewayIntf {
+		config.Gateway.Interface = gatewayIntfTemp
```
npinaeva: didn't you just assign
`gatewayIntfTemp := config.Gateway.Interface`? and now the other way around?

go-controller/pkg/node/gateway_init.go
```diff
@@ -513,6 +527,37 @@ func (nc *DefaultNodeNetworkController) initGatewayDPUHost(kubeNodeIP net.IP) er
 		return err
 	}
 
+	nodeIPNetv4, _ := util.MatchFirstIPNetFamily(false, ifAddrs)
+	nodeAddrSet := sets.New[string](nodeIPNetv4.String())
+
+	if gatewayIntfTemp != "" && gatewayIntfTemp != gatewayIntf {
```
npinaeva: could you please explain what is this logic doing and in which cases this condition can be false? Maybe an example of how these params/interfaces/addresses look like on a real cluster could be useful, otherwise it is a bit hard to follow

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -1006,11 +1006,17 @@ func (h *defaultNetworkControllerEventHandler) UpdateResource(oldObj, newObj int
 			// Also check if node subnet changed, so static routes are properly set
 			// Also check if the node is used to be a hybrid overlay node
 			syncZoneIC = syncZoneIC || h.oc.isLocalZoneNode(oldNode) || nodeSubnetChange || zoneClusterChanged ||
-				switchToOvnNode || nodeEncapIPsChanged
+				switchToOvnNode || nodeEncapIPsChanged || primaryAddrChanged(oldNode, newNode)
```
npinaeva: `primaryAddrChanged` checks `node.Status` to find primary IP, while you seem to be using `node-primary-ifaddr` annotation in the PR, please help me understand how you use both

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2445,6 +2445,9 @@ func newGateway(
 			gwBridge.eipMarkIPs = gw.bridgeEIPAddrManager.GetCache()
 		}
 		gw.nodeIPManager = newAddressManager(nodeName, kube, mgmtPort, watchFactory, gwBridge)
+		if gw.nodeIPManager == nil {
```
npinaeva: do you think it makes sense to refactor `newAddressManager to return an error? shouldn't be too difficult

go-controller/pkg/node/gateway_init.go
```diff
@@ -488,22 +502,22 @@ func interfaceForEXGW(intfName string) string {
...
-	// Force gateway interface to be the interface associated with kubeNodeIP
-	gwIntf, err := getInterfaceByIP(kubeNodeIP)
+	gwIntf, err = getInterfaceByIP(kubeNodeIP)
 	if err != nil {
 		return err
 	}
 	config.Gateway.Interface = gwIntf
-
-	_, gatewayIntf, err := getGatewayNextHops()
+	_, gatewayIntf, err = getGatewayNextHops()
```
npinaeva: I am thinking if this is the best way to get interface name... especially considering that we need to reset config variable for that
do you execute this part of the function https://github.com/ovn-kubernetes/ovn-kubernetes/blob/4de8fd5f958956874bce147e01f11bd0049abcd9/go-controller/pkg/node/gateway_init.go#L161 or something else?

aserdean: this interface is not necessarily under an OVS bridge.

will check to see what else is available in the codebase

go-controller/pkg/node/gateway_init.go
aserdean: as discussed offline we overwrite it here:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5307/files#diff-e9790f3508dd0c36e1fd2634870b3066acebe3d2d370d7c9e276ddb8e2278b21R519

and we are dancing around it because of the usage of: `getGatewayNextHops`

go-controller/pkg/node/gateway_init.go
aserdean: sure will add a couple of comments  what is trying to be achieved.

go-controller/pkg/node/gateway_shared_intf.go
aserdean: sure, should that be part of this PR though?

I originally added this part because I made a test panic and I added this check, however now it initializes itself properly and can safely remove the check.

What do you prefer?

go-controller/pkg/node/gateway_init.go
```diff
@@ -351,7 +353,19 @@ func (nc *DefaultNodeNetworkController) initGatewayPreStart(
...
+		}
+		nodeAddrStr, err := util.GetNodeIfAddrAnnotation(node)
+		if err != nil {
+			return nil, err
+		}
+		nodeAddr, _, err := net.ParseCIDR(nodeAddrStr.IPv4)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse node IP address. %v", nodeAddrStr)
+		}
+		ifAddrs, err = getDPUHostPrimaryIPAddresses(nodeAddr, ifAddrs)
```
trozet: So I think your PR description needs a lot more detail than it does now. Maybe I'm wrong, but my interpretation of these changes are as follows:
1. You are no longer using the kubeNodeIP. This is the kubernetes node IP of the host. I'm guessing this is because you are bringing this inline with the other PR where you want to tell the host to use an interface other than the one use for kapi access.
2. You are relying now on k8s.ovn.org/node-primary-ifaddr to provide you with the ip address of that interface on the host. You are adding code to SetNodePrimaryIfAddrs in the host path now.

I think there are several issues here:

1. A few lines down from this (line 360) we SetNodePrimaryIfAddrs with the value from ifAddrs, err = getNetworkInterfaceIPAddresses(gatewayIntf). Won't this override the value that you just got?
2. The ifAddrs were also passed to getDPUHostPrimaryIPAddresses, but were essentially ignored if config.Gateway.RouterSubnet was configured. Are you setting config.Gateway.RouterSubnet for your use case? I'm not sure why this config.Gateway.RouterSubnet is even needed, if the model is the host tells the dpu what to use...then why not just use the value specified by the host? Why try to validate it is in a certain subnet?

It seems like this PR is changing behavior, and we need to have a good description of what behavior you are trying to change and why in the PR.

go-controller/pkg/node/gateway_shared_intf.go
trozet: DPU mode is the only mode that allows returning nil for newAddressManager. I'm not sure if that is intentional and we do not need AddressManager for DPU mode or just a bug in the code. I see it klogs an error so I'm leaning towards it is needed and is a bug. However looking in the addressManager->sync() is skipped if in dpu mode. The current DPU path seems to just updateHostCIDRs once, I'm not sure why we need to fake creating a AddressManager for that. I would expect the host side to handle updating the HostCIDRs if the only ip addresses contained there are ones from the host and not the dpu.

@aserdean wdyt?

aserdean: Sounds reasonable to me but I don't want to handle that as part of this PR and indeed the hostcidrs should be updated only by the host as the name implies.

go-controller/pkg/node/gateway_init.go
aserdean: > 1. You are no longer using the kubeNodeIP. This is the kubernetes node IP of the host. I'm guessing this is because you are bringing this inline with the other PR where you want to tell the host to use an interface other than the one use for kapi access.

Not necessarily. The behavior I am looking for is the following.

On the host side:
- If no GW interface is specified, or if it is the one backing the kubernetesnodeIP, on ovnk DPU host mode  use the old behavior kubeNodeIP, set and consume the annotations as needed without affecting the functionality.


If a GW interface is specified and it is different than the one from kubeNodeIP:
- get cirds for kubenodeIP, get cidrs for the gw intf specified via gw options and store both to: `k8s.ovn.org/host-cidrs`
- the gw intf cidr gets stored into: k8s.ovn.org/node-primary-ifaddr

On the DPU side:
- always use `k8s.ovn.org/node-primary-ifaddr` to figure out what to do

In my mind this translates into the following:
- `k8s.ovn.org/host-cidrs`: which interfaces on the host side we want ovn-k to be aware of
- `k8s.ovn.org/node-primary-ifaddr`: this is my default interface of the host

I am setting `config.Gateway.RouterSubnet` blindly in my case and I did not fully try to understand what was the reasoning behind that code originally. We can raise it during the community meeting to see if it can be removed

go-controller/pkg/ovn/default_network_controller.go
aserdean: thanks for the pointer, confusing naming :)

will add an equivalent for the `node-primary-ifaddr`

aserdean:
@trozet re:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5307#discussion_r2155369360
since for some reason I can't reply to it...

I think that and the RouterSubnet needs an issue that needs to be revisited.


Again, I am not trying to complicate the PR more that it is. I will drop that additional check for the time being.


In general on the DPU side we seem to have dead code that should be removed.
please let me know if the above is reasonable?

go-controller/pkg/node/gateway_shared_intf.go
npinaeva: I have tried to move some things around, but with the current code state it is quite difficult. I will follow up with some refactoring, and we can leave it as is for now

trozet:
@coderabbit-ai re-review

go-controller/pkg/node/gateway_init.go
trozet: still need to address this

go-controller/pkg/node/gateway_init.go
trozet: It's still kind of silly that we are getting the address from the annotation, then writing the annotation again a few lines down. Especially if the model is supposed to be that the host is indicating what the primary-addr is. I don't see why the DPU side would ever rewrite that annotation.  @aserdean are you planning on fixing this now?

@adrianchiris I think you originally wrote a lot of this code. Maybe you can tell us the purpose of the RouterSubnet and why the validation of the IP was necessary in GetDPUHostPrimaryIPAddresses?

go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -317,12 +317,8 @@ func (c *addressManager) updateNodeAddressAnnotations() error {
 
 func (c *addressManager) updateHostCIDRs(ifAddrs []*net.IPNet) error {
 	if config.OvnKubeNode.Mode == types.NodeModeDPU {
-		// For DPU mode, here we need to use the DPU host's IP address which is the tenant cluster's
-		// host internal IP address instead.
-		// Currently we are only intentionally supporting IPv4 for DPU here.
-		nodeIPNetv4, _ := util.MatchFirstIPNetFamily(false, ifAddrs)
-		nodeAddrSet := sets.New[string](nodeIPNetv4.String())
-		return util.SetNodeHostCIDRs(c.nodeAnnotator, nodeAddrSet)
+		// The host CIDR should be updated only by the full and host mode
```
trozet: look at the lint error above, need to fix it.

go-controller/pkg/ovn/default_network_controller.go
trozet: still need to address this one

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -1006,11 +1006,17 @@ func (h *defaultNetworkControllerEventHandler) UpdateResource(oldObj, newObj int
...
 			if syncZoneIC {
 				klog.Infof("Node %q in remote zone %q, network %q, needs interconnect zone sync up. Zone cluster changed: %v",
 					newNode.Name, util.GetNodeZone(newNode), h.oc.GetNetworkName(), zoneClusterChanged)
 			}
+			if nodeChassisChanged(oldNode, newNode) {
```
trozet: why this change?

go-controller/pkg/node/gateway_init.go
```diff
@@ -488,22 +502,22 @@ func interfaceForEXGW(intfName string) string {
...
 	// watch on services for iptable rule updates and run a loadBalancerHealth checker
 	// Note: all K8s Node related annotations are handled from DPU.
 	klog.Info("Initializing Shared Gateway Functionality on DPU host")
 	var err error
+	var gwIntf, gatewayIntf string
+	gatewayIntfTemp := config.Gateway.Interface
 
-	// Force gateway interface to be the interface associated with kubeNodeIP
-	gwIntf, err := getInterfaceByIP(kubeNodeIP)
+	gwIntf, err = getInterfaceByIP(kubeNodeIP)
```
trozet: dont call this gwIntf, call it like kubeIntf or something

go-controller/pkg/node/gateway_init.go
trozet: @npinaeva I think the part of another comment that explains it was:

```
On the host side:

If no GW interface is specified, or if it is the one backing the kubernetesnodeIP, on ovnk DPU host mode use the old behavior kubeNodeIP, set and consume the annotations as needed without affecting the functionality.

If a GW interface is specified and it is different than the one from kubeNodeIP:

get cirds for kubenodeIP, get cidrs for the gw intf specified via gw options and store both to: k8s.ovn.org/host-cidrs
the gw intf cidr gets stored into: k8s.ovn.org/node-primary-ifaddr
```

I think we need to change this a bit. If your goal is to do this logic and you are having issues because by this point in time the interface has already been potentially moved to an OVS bridge, then why not just execute this logic before that happens? 

In gateway_init.go -> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_init.go#L355 (same place you made earlier changes), why not just add an if dpuhostmode block, get the kubeNodeIP intf, run your logic there, then write the annotations?

trozet (changes requested):

go-controller/pkg/node/gateway_init.go
adrianchiris: >@adrianchiris I think you originally wrote a lot of this code. Maybe you can tell us the purpose of the RouterSubnet and why the validation of the IP was necessary in GetDPUHostPrimaryIPAddresses?


Here is what i remember:

we need to configure the OVN GR for that node/dpu (by setting the gateway config annot).
for that we need to have some subnet and an IP within this subnet to perform NAT when egressing to underlay.

at the time the k8s node IP (of the host) was used and RouterSubnet was the means to get the subnet where this IP resided. 

if RouterSubnet was not specified then we try to derive the subnet from the gateway interface netdev

go-controller/pkg/node/gateway_init.go
aserdean: ack, will move it one level up so it will be clear what we try to do in host mode and in dpu mode

go-controller/pkg/ovn/default_network_controller.go
aserdean: right, will fold the changes in the next incremental

go-controller/pkg/node/gateway_init.go
aserdean: >@aserdean are you planning on fixing this now?

Will fix it in another MR.

We might want to use it on the DPU in the case we do not have an IP address on the OVNK managed bridge.

Will create an issue for that effort that and reference the comment there.

aserdean:
thank you for the review, will incorporate the comments and fold them in a new revision

go-controller/pkg/ovn/default_network_controller.go
aserdean: since we have a dpu, which consists the ovs side and this is pinned to a host, if the dpu is reprovisioned the system-id will change but the node will not. this will handle that use-case

go-controller/pkg/node/gateway_init.go
aserdean: >If your goal is to do this logic and you are having issues because by this point in time the interface has already been potentially moved to an OVS bridge,

FWIW I do not have issues regarding ovs. Again on the DPU host there is no OVS installed only on the dpu, but I guess the code would look better as you suggested.

go-controller/pkg/node/gateway_init.go
```diff
@@ -350,13 +351,38 @@ func (nc *DefaultNodeNetworkController) initGatewayPreStart(
...
+		}
+
+		// Set the parsed IP as the network address
+		nodeAddrs.IP = nodeIP
+
+		// Create a new slice and replace ifAddrs with the DPU host address
+		// This overrides the gateway interface addresses for DPU mode
+		var gwIps []*net.IPNet
+		ifAddrs = append(gwIps, nodeAddrs)
 	}
```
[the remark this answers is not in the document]
aserdean: we only set IPv4 inside the codebase, no need to complicate this logic at the moment.

go-controller/pkg/node/gateway_init.go
```diff
@@ -359,43 +385,79 @@ func interfaceForEXGW(intfName string) string {
...
+	ifAddrs, err := nodeutil.GetNetworkInterfaceIPAddresses(kubeIntf)
 	if err != nil {
 		return err
 	}
 
-	ifAddrs, err := nodeutil.GetNetworkInterfaceIPAddresses(gatewayIntf)
-	if err != nil {
+	// Extract the IPv4 address from the interface addresses for node annotation
+	nodeIPNet, _ := util.MatchFirstIPNetFamily(false, ifAddrs)
+	nodeAddrSet := sets.New[string](nodeIPNet.String())
```
npinaeva, line 412: that means we don't support ipv6-only clusters?

npinaeva: also looks like we only ever add ipv4 IPs to `nodeAddrSet`?

npinaeva: So is there ever a valid config where `OVNNodePrimaryDPUAddr` will have an ipv6-only address?

go-controller/pkg/ovn/default_network_controller.go
npinaeva: Can you add a comment explaining this? Also, do we only need this in DPUMode?

aserdean: >Can you add a comment explaining this?

Sure
>Also, do we only need this in DPUMode?

Correct we need this just for dpumode, since its the only combo in which the uuid is not tied to the worker hostname.

go-controller/pkg/node/gateway_init.go
aserdean: >that means we don't support ipv6-only clusters?

correct, the current DPU mode codebase side is IPv4 opinionated. we could tackle that in the future, please let me know if I should open an issue in order to track that.

go-controller/pkg/ovn/default_network_controller.go
npinaeva: Should we then add the DPU mode check here?

go-controller/pkg/node/gateway_init.go
npinaeva: I would prefer being open about that in the code too, so if we know that `GetNodePrimaryDPUAddrAnnotation` will never have ipv6 address, there is no need to have all the `if`s and doing stuff like https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5307/files#diff-e9790f3508dd0c36e1fd2634870b3066acebe3d2d370d7c9e276ddb8e2278b21R237-R240
we could just say it is always ipv4

aserdean: oki, I can make some comments to make that clear on the helpers as well

go-controller/pkg/ovn/default_network_controller.go
aserdean: yes, already added as part of the next incremental and rebase.

go-controller/pkg/node/gateway_init.go
```diff
@@ -219,13 +220,38 @@ func (nc *DefaultNodeNetworkController) initGatewayPreStart(
...
+		nodeIfAddr, err := util.GetNodePrimaryDPUAddrAnnotation(node)
 		if err != nil {
 			return nil, err
 		}
+		nodeAddrStr := nodeIfAddr.IPv4
+		if nodeAddrStr == "" {
+			nodeAddrStr = nodeIfAddr.IPv6
+		}
+
+		// Parse the IPv4 address string into IP and network components
```
trozet, line 243: this says ipv4 but the above logic says ipv6

go-controller/pkg/node/gateway_init.go
trozet: should we also check if dpu mode && !config.IPv4Mode { return err} ?

trozet (changes requested):
minor comments about ip family handling, other than that lgtm

go-controller/pkg/util/node_annotations.go
```diff
@@ -97,6 +97,9 @@ const (
 	// OVNNodeHostCIDRs is used to track the different host IP addresses and subnet masks on the node
 	OVNNodeHostCIDRs = "k8s.ovn.org/host-cidrs"
 
+	// OVNNodePrimaryDPUAddr is used to track the primary DPU address on the node
+	OVNNodePrimaryDPUAddr = "k8s.ovn.org/primary-dpu-addr"
```
trozet: would this actually make more sense as primary-dpu-host-addr?

go-controller/pkg/ovn/controller/services/lb_config.go
```diff
@@ -92,9 +93,18 @@ func makeNodeRouterTargetIPs(node *nodeInfo, c *lbConfig, hostMasqueradeIPV4, ho
...
+	hostMasqueradeIPsV6 := node.l3gatewayAddressesStr()
+
+	if config.OvnKubeNode.Mode == types.NodeModeDPU {
+		// hairpin traffic of the hostaddress in case of the DPU
+		hostMasqueradeIPsV4 = node.hostAddressesStr()
+		hostMasqueradeIPsV6 = node.hostAddressesStr()
+	}
+
+	targetIPsV4, v4Updated = util.UpdateIPsSlice(targetIPsV4, hostMasqueradeIPsV4, []string{hostMasqueradeIPV4})
+	targetIPsV6, v6Updated = util.UpdateIPsSlice(targetIPsV6, hostMasqueradeIPsV6, []string{hostMasqueradeIPV6})
```
trozet: so really what you are doing here is trying to ignore using the l3 gateway address and include other host ips to find and replace addresses with masquerade IPs for the GR loadbalancer VIP. In shared gateway mode we assume that the gwIntf ==kube api IP and that assumption is why you need to change this code. I think your change is valid even in a *non* DPU case. I would:

1. remove this function: node.l3gatewayAddressesStr()
2. always use node.hostAddressesStr() for UpdateIPSlice

Then any IP used on the host will be substituted for the masquerade IP, which what needs to happen to make host->service work for any interface that is used as KAPI.

go-controller/pkg/node/gateway_init.go
aserdean: its both ipv6 and 4 https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5307/files, but will drop the v6 part for now

go-controller/pkg/ovn/controller/services/lb_config.go
aserdean: as discussed offline will make the change, lets hope the CI will be happy :)

go-controller/pkg/util/node_annotations.go
aserdean: updated + helpers

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -965,8 +965,12 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
 
 	// First part of gateway initialization. It will be completed by (nc *DefaultNodeNetworkController) Start()
 	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
+		// IPv6 is not supported in DPU enabled nodes, error out if ovnkube is not set in IPv4 mode
+		if config.IPv6Mode {
+			return fmt.Errorf("IPv6 mode is not supported on a DPU enabled node")
```
trozet, line 970: this is failing for even full mode clusters:

E0722 13:17:19.422863    2839 node_controller_manager.go:168] Stopping node network controller manager, err=failed to init default node network controller: IPv6 mode is not supported on a DPU enabled node

the check above is != DPUHost, need to check also if DPUMode

trozet (approved):

PR #5310, opened 2025-06-16
EIP OVN controller: stop pod config flap

martinkennelly:
func was refactored erroneously when network
comparisson was refactored. The if comparisson went from:

!cachedNetwork.Equals(ni)

to:

util.AreNetworksCompatible(cachedNetwork, ni)

Disruption can be seen for brief periods of time.

cc @jcaamano 

Introduced in PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4533

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of egress IP assignment cleanup and SNAT rule application to address edge cases in network compatibility and topology.

* **Tests**
  * Added a test to ensure pod and namespace label updates do not unintentionally alter egress IP configurations.
  * Introduced a comprehensive test verifying correct OVN configuration cleanup when pods associated with EgressIP are deleted across multiple networks and nodes.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -2451,11 +2451,18 @@ func (e *EgressIPController) deletePodEgressIPAssignment(ni util.NetInfo, egress
 		return err
 	}
 	var ops []ovsdb.Operation
-	if !loadedPodNode || isLocalZonePod { // node is deleted (we can't determine zone so we always try and nuke OR pod is local to zone)
+	// For CDN only, add SNATs to support external GW feature
+	if ni.IsDefault() && (!loadedPodNode || isLocalZonePod) {
 		ops, err = e.addExternalGWPodSNATOps(ni, nil, pod.Namespace, pod.Name, status)
 		if err != nil {
 			return err
 		}
+	}
+	// Following cases will ensure removal of a pod LRP
+	// Case 1 - node where pod is hosted is not known
+	// Case 2 - pod is within the local zone
+	// case 3 - a local zone node is egress node and pod is attached to layer 2. For layer2, there is always an LRP attached to the egress Node GW router
+	if !loadedPodNode || isLocalZonePod || (isLocalZoneEgressNode && ni.IsSecondary() && ni.TopologyType() == types.Layer2Topology) {
 		ops, err = e.deleteReroutePolicyOps(ni, ops, status, egressIPName, nextHopIP, routerName, pod.Namespace, pod.Name)
 		if errors.Is(err, libovsdbclient.ErrNotFound) {
 			// if the gateway router join IP setup is already gone, then don't count it as error.
--- a/go-controller/pkg/ovn/egressip_udn_l2_test.go
+++ b/go-controller/pkg/ovn/egressip_udn_l2_test.go
@@ -2558,4 +2558,490 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 		})
 	})
+
+	ginkgo.Context("Pod delete", func() {
+		ginkgo.It("should delete UDN and CDN config", func() {
+			// create a single EIP IP selecting multiple pods both local and remote.
+			// Delete pods and ensure OVN DB is as expected
+			app.Action = func(ctx *cli.Context) error {
+				// Node 1 is local, Node 2 is remote
+				egressIP := "192.168.126.101"
+				node1IPv4 := "192.168.126.202"
+				node1IPv4CIDR := node1IPv4 + "/24"
+				node2IPv4 := "192.168.126.51"
+				node2IPv4CIDR := node2IPv4 + "/24"
+				_, node1CDNSubnet, _ := net.ParseCIDR(v4Node1Subnet)
+				_, node1UDNSubnet, _ := net.ParseCIDR(v4Node1Net1)
+				nadName := util.GetNADName(eipNamespace2, nadName1)
+				egressCDNNamespace := newNamespaceWithLabels(eipNamespace, egressPodLabel)
+				egressUDNNamespace := newUDNNamespaceWithLabels(eipNamespace2, egressPodLabel)
+				egressPodCDNLocal := *newPodWithLabels(eipNamespace, podName, node1Name, podV4IP, egressPodLabel)
+				egressPodUDNLocal := *newPodWithLabels(eipNamespace2, podName2, node1Name, v4Pod1IPNode1Net1, egressPodLabel)
+				egressPodCDNRemote := *newPodWithLabels(eipNamespace, podName3, node2Name, podV4IP2, egressPodLabel)
+				setPrimaryNetworkAnnot(&egressPodCDNRemote, ovntypes.DefaultNetworkName, fmt.Sprintf("%s%s", podV4IP2, util.GetIPFullMaskString(podV4IP2)))
+				egressPodUDNRemote := *newPodWithLabels(eipNamespace2, podName4, node2Name, v4Pod2IPNode2Net1, egressPodLabel)
+				setPrimaryNetworkAnnot(&egressPodUDNRemote, nadName, fmt.Sprintf("%s%s", v4Pod2IPNode2Net1, util.GetIPFullMaskString(v4Pod2IPNode2Net1)))
+				netconf := ovncnitypes.NetConf{
+					NetConf: cnitypes.NetConf{
+						Name: networkName1,
+						Type: "ovn-k8s-cni-overlay",
+					},
+					Role:     ovntypes.NetworkRolePrimary,
+					Topology: ovntypes.Layer2Topology,
+					NADName:  nadName,
+					Subnets:  v4Net1,
+				}
+				nad, err := newNetworkAttachmentDefinition(
+					eipNamespace2,
+					nadName1,
+					netconf,
+				)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				nad.Annotations = map[string]string{ovntypes.OvnNetworkIDAnnotation: secondaryNetworkID}
+				netInfo, err := util.NewNetInfo(&netconf)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+				node1Annotations := map[string]string{
+					"k8s.ovn.org/node-primary-ifaddr":             fmt.Sprintf("{\"ipv4\": \"%s\", \"ipv6\": \"%s\"}", node1IPv4CIDR, ""),
+					"k8s.ovn.org/node-subnets":                    fmt.Sprintf("{\"default\":\"%s\",\"%s\":\"%s\"}", v4Node1Subnet, networkName1, v4Node1Net1),
+					"k8s.ovn.org/node-transit-switch-port-ifaddr": fmt.Sprintf("{\"ipv4\":\"%s/16\"}", v4Node1Tsp),
+					"k8s.ovn.org/zone-name":                       node1Name,
+					"k8s.ovn.org/node-chassis-id":                 "473ca66d-d800-472f-b289-1ab81ae7f21c",
+					"k8s.ovn.org/remote-zone-migrated":            node1Name,
+					"k8s.ovn.org/node-gateway-router-lrp-ifaddrs": fmt.Sprintf("{\"default\":{\"ipv4\":\"%s\"},\"%s\":{\"ipv4\":\"%s\"}}", node1DefaultRtoJIPCIDR, networkName1, node1Network1RtoSIPCIDR),
+					util.OVNNodeHostCIDRs:                         fmt.Sprintf("[\"%s\"]", node1IPv4CIDR),
+					util.OvnNodeL3GatewayConfig: fmt.Sprintf(`{"%s":{"mode":"local","mac-address":"7e:57:f8:f0:3c:49", "ip-address":"%s", "next-hop":"%s", "next-hops": ["%s"]},
+"default":{"mode":"local","mac-address":"7e:57:f8:f0:3c:49", "ip-address":"192.168.126.12/24", "next-hop": "192.168.126.1", "next-hops": ["192.168.126.1"]}}`, networkName1, v4Net1, gwIP, gwIP),
+				}
+				labels := map[string]string{
+					"k8s.ovn.org/egress-assignable": "",
+				}
+				node1 := getNodeObj(node1Name, node1Annotations, labels)
+				node2Annotations := map[string]string{
+					"k8s.ovn.org/node-primary-ifaddr":             fmt.Sprintf("{\"ipv4\": \"%s\", \"ipv6\": \"%s\"}", node2IPv4CIDR, ""),
+					"k8s.ovn.org/node-subnets":                    fmt.Sprintf("{\"default\":\"%s\",\"%s\":\"%s\"}", v4Node2Subnet, networkName1, v4Node2Net1),
+					"k8s.ovn.org/node-transit-switch-port-ifaddr": fmt.Sprintf("{\"ipv4\":\"%s/16\"}", v4Node2Tsp),
+					"k8s.ovn.org/zone-name":                       node2Name,
+					"k8s.ovn.org/node-chassis-id":                 "473ca66d-d800-472f-b289-1ab81ae7f21c",
+					"k8s.ovn.org/remote-zone-migrated":            node2Name,
+					"k8s.ovn.org/node-gateway-router-lrp-ifaddrs": fmt.Sprintf("{\"default\":{\"ipv4\":\"%s\"},\"%s\":{\"ipv4\":\"%s\"}}", node2DefaultRtoJIPCIDR, networkName1, node2Network1RtoSIPCIDR),
+					util.OVNNodeHostCIDRs:                         fmt.Sprintf("[\"%s\"]", node2IPv4CIDR),
+					util.OvnNodeL3GatewayConfig: fmt.Sprintf(`{"%s":{"mode":"local","mac-address":"7e:57:f8:f0:3c:49", "ip-address":"%s", "next-hop":"%s", "next-hops": ["%s"]},
+"default":{"mode":"local","mac-address":"7e:57:f8:f0:3c:49", "ip-address":"192.168.126.12/24", "next-hop": "192.168.126.1", "next-hops": ["192.168.126.1"]}}`, networkName1, v4Net1, gwIP, gwIP),
+				}
+				node2 := getNodeObj(node2Name, node2Annotations, nil)
+				oneNodeStatus := []egressipv1.EgressIPStatusItem{
+					{
+						Node:     node1Name,
+						EgressIP: egressIP,
+					},
+				}
+				eIP := egressipv1.EgressIP{
+					ObjectMeta: newEgressIPMetaWithMark(egressIPName, eIP1Mark),
+					Spec: egressipv1.EgressIPSpec{
+						EgressIPs: []string{egressIP},
+						PodSelector: metav1.LabelSelector{
+							MatchLabels: egressPodLabel,
+						},
+						NamespaceSelector: metav1.LabelSelector{
+							MatchLabels: egressPodLabel,
+						},
+					},
+					Status: egressipv1.EgressIPStatus{
+						Items: oneNodeStatus,
+					},
+				}
+
+				initialDB := []libovsdbtest.TestData{
+					//CDN start
+					&nbdb.LogicalRouterPort{
+						UUID:     ovntypes.GWRouterToJoinSwitchPrefix + ovntypes.GWRouterPrefix + node1.Name + "-UUID",
+						Name:     ovntypes.GWRouterToJoinSwitchPrefix + ovntypes.GWRouterPrefix + node1.Name,
+						Networks: []string{node1DefaultRtoJIPCIDR},
+					},
+					&nbdb.LogicalRouter{
+						Name: ovntypes.OVNClusterRouter,
+						UUID: ovntypes.OVNClusterRouter + "-UUID",
+					},
+					&nbdb.LogicalRouter{
+						Name:  ovntypes.GWRouterPrefix + node1.Name,
+						UUID:  ovntypes.GWRouterPrefix + node1.Name + "-UUID",
+						Ports: []string{ovntypes.GWRouterToJoinSwitchPrefix + ovntypes.GWRouterPrefix + node1.Name + "-UUID"},
+					},
+					&nbdb.LogicalSwitchPort{
+						UUID:      "k8s-" + node1Name + "-UUID",
+						Name:      "k8s-" + node1Name,
+						Addresses: []string{"fe:1a:b2:3f:0e:fb " + util.GetNodeManagementIfAddr(node1CDNSubnet).IP.String()},
+					},
+					&nbdb.LogicalSwitch{
+						UUID:  node1Name + "-UUID",
+						Name:  node1Name,
+						Ports: []string{"k8s-" + node1Name + "-UUID"},
+					},
+					// UDN start
+					&nbdb.LogicalRouterPort{
+						UUID:     ovntypes.RouterToSwitchPrefix + networkName1_ + layer2SwitchName + "-UUID",
+						Name:     ovntypes.RouterToSwitchPrefix + networkName1_ + layer2SwitchName,
+						Networks: []string{node1Network1RtoSIPCIDR},
+					},
+					&nbdb.LogicalRouter{
+						UUID:        netInfo.GetNetworkScopedGWRouterName(node1.Name) + "-UUID",
+						Name:        netInfo.GetNetworkScopedGWRouterName(node1.Name),
+						Ports:       []string{ovntypes.RouterToSwitchPrefix + networkName1_ + layer2SwitchName + "-UUID"},
+						ExternalIDs: map[string]string{ovntypes.NetworkExternalID: networkName1, ovntypes.TopologyExternalID: ovntypes.Layer2Topology},
+					},
+					&nbdb.LogicalSwitchPort{
+						UUID:      "k8s-" + networkName1_ + node1Name + "-UUID",
+						Name:      "k8s-" + networkName1_ + node1Name,
+						Addresses: []string{"fe:1a:b2:3f:0e:fb " + util.GetNodeManagementIfAddr(node1UDNSubnet).IP.String()},
+					},
+					&nbdb.LogicalSwitch{
+						UUID:        netInfo.GetNetworkScopedSwitchName(node1.Name) + "-UUID",
+						Name:        netInfo.GetNetworkScopedSwitchName(node1.Name),
+						Ports:       []string{"k8s-" + networkName1_ + node1Name + "-UUID"},
+						ExternalIDs: map[string]string{ovntypes.NetworkExternalID: networkName1, ovntypes.TopologyExternalID: ovntypes.Layer2Topology},
+					},
+				}
+				fakeOvn.startWithDBSetup(
+					libovsdbtest.TestSetup{
+						NBData: initialDB,
+					},
+					&corev1.NodeList{
+						Items: []corev1.Node{node1, node2},
+					},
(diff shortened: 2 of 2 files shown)


martinkennelly:
introduced in commit 2f860b91340a6777b3bf632adf98cec3c19ec8c2

martinkennelly:
working on how to test this scenario.

martinkennelly:
Adding UT

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+								"name": egressNamespace.Name,
+							},
+						},
+						PodSelector: metav1.LabelSelector{
+							MatchLabels: egressPodLabel,
+						},
+					},
+				}
+				i, n, _ := net.ParseCIDR(podV4IP + "/23")
+				n.IP = i
```
maiqueb, line 3491: imho, these could be better named. Or even better wrapped up in an helper func somewhere. 

Actually, I think there's already an helper out there for this.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+								Name: "k8s-" + node.Name,
+								Addresses: []string{"fe:1a:b2:3f:0e:fb " + util.GetNodeManagementIfAddr(nodeSubnetV4).IP.String(),
+									"fe:1a:b2:3f:0e:fb " + util.GetNodeManagementIfAddr(nodeSubnetV6).IP.String()},
+							},
+							&nbdb.LogicalSwitch{
+								UUID:  node.Name + "-UUID",
+								Name:  node.Name,
+								Ports: []string{"k8s-" + node.Name + "-UUID"},
+							},
+						},
```
maiqueb, line 3461: can we maybe move this to a parametrized helper ? 

I find it hard to review like this. Not a must, but I think it'll help w/ the test readability.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+						UUID: "k8s-" + node.Name + "-UUID",
+						Name: "k8s-" + node.Name,
+						Addresses: []string{"fe:1a:b2:3f:0e:fb " + util.GetNodeManagementIfAddr(nodeSubnetV4).IP.String(),
+							"fe:1a:b2:3f:0e:fb " + util.GetNodeManagementIfAddr(nodeSubnetV6).IP.String()},
+					},
+					&nbdb.LogicalSwitch{
+						UUID:  node.Name + "-UUID",
+						Name:  node.Name,
+						Ports: []string{"k8s-" + node.Name + "-UUID"},
+					},
```
maiqueb, line 3537: I think these should be wrapped up in functions, so it becomes easier to compare against the initial state, which you've defined above.

maiqueb (dismissed):
I think the unit test's readability can be improved, but the fix looks legit - good catch.

martinkennelly:
@maiqueb i dont have time to refactor the whole suite tbh to make it consistent.

martinkennelly:
rebased only @tssurya

martinkennelly:
@tssurya  waiting on your review - thanks.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1083,7 +1083,7 @@ func (e *EgressIPController) deletePodEgressIPAssignments(ni util.NetInfo, name
 func (e *EgressIPController) deletePreviousNetworkPodEgressIPAssignments(ni util.NetInfo, name string, statusesToRemove []egressipv1.EgressIPStatusItem, pod *corev1.Pod) {
```
tssurya, line 1083: sorry this is not at all related to your PR
how can a network change for a pod? :o do we mean advertised v/s non-advertised?
or is it the network was deleted and recreated or something some corner case?

tssurya: ah I see this is the logic that checks if old and new are "same network" or "different networks"?
`canReconcile` is checking if its the same network that got an update? i hope this can happen only for NADs right? not for UDN CRs since that's immutable?
can you provide an example scenario of when this can happen and we hit the "not compatible" condition?

go-controller/pkg/ovn/egressip.go
```diff
@@ -1083,7 +1083,7 @@ func (e *EgressIPController) deletePodEgressIPAssignments(ni util.NetInfo, name
...
 	cachedNetwork := e.getNetworkFromPodAssignment(getPodKey(pod))
 	if cachedNetwork != nil {
-		if util.AreNetworksCompatible(cachedNetwork, ni) {
+		if !util.AreNetworksCompatible(cachedNetwork, ni) {
```
tssurya, line 1086: makes sense , if not compatible then remove the config.. cause we can't reconcile things?

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+						Items: []corev1.Namespace{*egressNamespace},
+					},
+					&corev1.PodList{
+						Items: []corev1.Pod{egressPod},
+					},
+					&corev1.NodeList{
+						Items: []corev1.Node{node},
+					},
+				)
+				// matches only on namespace
```
tssurya: hmm seems to match on namespace and pod?

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+						Name:  node.Name,
+						Ports: []string{"k8s-" + node.Name + "-UUID"},
+					},
+				}
+				gomega.Eventually(fakeOvn.nbClient).Should(libovsdbtest.HaveData(expectedDatabaseState))
+				// create no-op updates that trigger reconcile for the selected pod async but update should continue to select the pod and not alter pod config
+				go func() {
+					ns, err := fakeOvn.fakeClient.KubeClient.CoreV1().Namespaces().Get(context.TODO(), egressNamespace.Name, metav1.GetOptions{})
+					if err != nil {
+						panic(err.Error())
```
tssurya: hmm panic sounds dangerous..why panic?

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+					}
+					pod = pod.DeepCopy()
+					pod.Labels["newlabel"] = "noop"
+					_, err = fakeOvn.fakeClient.KubeClient.CoreV1().Pods(egressNamespace.Name).Update(context.TODO(), pod, metav1.UpdateOptions{})
+					if err != nil {
+						panic(err.Error())
+					}
+				}()
+				// ensure the DBs are unaltered
```
tssurya, line 3573: before this change `AreNetworksCompatible` would return true? and we were removing pod config and it would flap?
like i don't understand how the pod update event is triggering a "changed network" event which is what we really want right?

go-controller/pkg/ovn/egressip.go
martinkennelly: I dont know. I didnt think we could get into the scenario where the net is imcompat. I can spend some time understanding this if you want.

go-controller/pkg/ovn/egressip_test.go
martinkennelly: ah ack. oops - will correct.

go-controller/pkg/ovn/egressip_test.go
martinkennelly: because i didnt have a simpler approach to signally a go thread failed. I dont think I can do Expect within a go routine.

go-controller/pkg/ovn/egressip_test.go
martinkennelly: > before this change AreNetworksCompatible would return true? and we were removing pod config and it would flap?

its not - i just want to make sure that during the reconciliation loop, theres no reconfig. I didnt want to write a unit test as specific as what you stated.

tssurya: aaah I see..I'd actually rather have a test that actually tests the modularity of what is expected if the if condition matches - not the reverse.. though I understand pod config not being deleted is a way for checking this to ensure no flaps happen - I don't see why we need a new unit test for this then? cause I am sure at least one of the existing tests must have a "pod update" path - there are existing tests with label selectors that can be reused to also test this bit right?

go-controller/pkg/ovn/egressip.go
tssurya: nah I think faster approach is asking @jcaamano maybe you know what scenario you had in mind when you wrote these bits for reconcile..- when can we get into a situation of "incompatible network update" for a pod ?

tssurya:
holly molly CI on net-seg is all red for ~no reason?~ EIP tests failing
```
  [FAIL] e2e egress IP validation Network Segmentation: IPv4 L2 role primary [It] Should validate the egress IP SNAT functionality for stateful-sets
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/egressip.go:1364
  [FAIL] e2e egress IP validation Network Segmentation: IPv6 L2 role primary [It] Should validate the egress IP SNAT functionality for stateful-sets
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/egressip.go:1364
```
and
```
Summarizing 1 Failure:
  [FAIL] e2e egress IP validation Network Segmentation: IPv4 L2 role primary [It] Should validate the egress IP SNAT functionality for stateful-sets
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/egressip.go:1364
```
that looks suspicious..
on all lanes... i wonder what changed suddenly?

martinkennelly:
Looking now - sorry for delay.

martinkennelly:
CI looks good and no flaking after fix for the stateful set test for EIP feature.

martinkennelly:
@tssurya looking good

martinkennelly:
cc @tssurya

tssurya:
Yup on it

go-controller/pkg/ovn/egressip.go
```diff
@@ -2451,7 +2451,8 @@ func (e *EgressIPController) deletePodEgressIPAssignment(ni util.NetInfo, egress
 		return err
 	}
 	var ops []ovsdb.Operation
-	if !loadedPodNode || isLocalZonePod { // node is deleted (we can't determine zone so we always try and nuke OR pod is local to zone)
+
+	if !loadedPodNode || isLocalZonePod || (isLocalZoneEgressNode && ni.IsSecondary() && ni.TopologyType() == types.Layer2Topology) { // node is deleted (we can't determine zone so we always try and nuke OR pod is local to zone)
```
tssurya: can you please update commit message to describe what was the stale scenario, why CI was failing etc?

(isLocalZoneEgressNode && ni.IsSecondary() && ni.TopologyType() == types.Layer2Topology)

this bit.. not sure why it needs a specific match here? the 1st two matches are about "not able to determine pod zone" let's try to delete OR "pod is local to this zone" so we must delete the policy...
what's this 3rd case?
I think we should have code comments here for CASE1, CASE2, CASE3 its getting complicated to read

go-controller/pkg/ovn/egressip_test.go
martinkennelly: > aaah I see..I'd actually rather have a test that actually tests the modularity of what is expected if the if condition matches - not the reverse.. though I understand pod config not being deleted is a way for checking this to ensure no flaps happen - I don't see why we need a new unit test for this then? cause I am sure at least one of the existing tests must have a "pod update" path - there are existing tests with label selectors that can be reused to also test this bit right?

my bug is regarding pod config flapping so i tesed to ensure that doesn't occur.
I couldnt think of scenario of network changing anyway. I didnt add that code. I meant to ask jaime but he was on pto.

go-controller/pkg/ovn/egressip.go
martinkennelly: sure.

ya, its a mess. 

I was hoping i wasnt the person asked to do it :) Ill do it now.

go-controller/pkg/ovn/egressip.go
```diff
@@ -2451,11 +2451,18 @@ func (e *EgressIPController) deletePodEgressIPAssignment(ni util.NetInfo, egress
...
 		ops, err = e.addExternalGWPodSNATOps(ni, nil, pod.Namespace, pod.Name, status)
 		if err != nil {
 			return err
 		}
+	}
+	// Following cases will ensure removal of a pod LRP
+	// Case 1 - node where pod is hosted is not known
+	// Case 2 - pod is within the local zone
+	// case 3 - a local zone node is egress node and pod is attached to layer 2. For layer2, there is always an LRP attached to the egress Node GW router
+	if !loadedPodNode || isLocalZonePod || (isLocalZoneEgressNode && ni.IsSecondary() && ni.TopologyType() == types.Layer2Topology) {
```
tssurya, line 2465: self-note: isLocalZoneEgressNode here is true if the EgressNode is local to this zone's controller.

tssurya: self-note-2: ni.IsSecondary()  is redundant as we don't support layer2 for CDNs, but its good to have this check for visibility. changing this to IsPrimaryNetwork is also accurate as secondary networks don't have egress/egressip support today.

tssurya: @martinkennelly to confirm if the remove LRP is removed for L3 in addition to L2? maybe L2488 has a ifelse for L3 delete already?

go-controller/pkg/ovn/egressip_udn_l2_test.go
```diff
@@ -2558,4 +2558,490 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
...
+					},
+					getNoReRouteReplyTrafficPolicyForController(netInfo.GetNetworkName(), DefaultNetworkControllerName),
+					getDefaultQoSRule(false, netInfo.GetNetworkName(), DefaultNetworkControllerName),
+					egressIPServedPodsASUDNv4,
+					udnEnabledSvcV4,
+				}
+				ginkgo.By("ensure expected equals actual")
+				gomega.Eventually(fakeOvn.nbClient).Should(libovsdbtest.HaveData(expectedDatabaseState))
+				ginkgo.By("deleting all EgressIP seelected pods")
+				deletePod(egressPodCDNLocal.Namespace, egressPodCDNLocal.Name, fakeOvn.fakeClient.KubeClient)
```
tssurya, line 2915: why do we have CDN pods in L2 tests?
answer: because we want to make sure we don't break CDNs when changing UDN stuff (came from previous UDN review comments from Tim)

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 		)
 	})
 
+	ginkgo.Context("IPv4 on pod UPDATE", func() {
```
martinkennelly, line 3413: regarding the concern that this is already completed in other tests raised in a call with surya, theres an IPv6 pod update context section but none for IPv4. The other contexts are primary focused on everything but pod update and I want a specific test for my bug instead of adding more complexity to existing unit tests. I think this is structured ok after reading the entire test file to see if its appropriate.  I wanted to create a no-op events that triggered the issue.

tssurya: ack fair enough. thanks for looking through the tests to ensure there isn't another dupe

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+					if err != nil {
+						panic(err.Error())
+					}
+					pod = pod.DeepCopy()
+					pod.Labels["newlabel"] = "noop"
+					_, err = fakeOvn.fakeClient.KubeClient.CoreV1().Pods(egressNamespace.Name).Update(context.TODO(), pod, metav1.UpdateOptions{})
+					if err != nil {
+						panic(err.Error())
+					}
+				}()
```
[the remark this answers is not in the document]
tssurya: @coderabbitai can you provide link to the go best practices for this? 
@martinkennelly should we consider using channels? :) panics do seem a bit bold :D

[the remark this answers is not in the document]
tssurya: @coderabbitai can you provide link to the go best practices for this?

go-controller/pkg/ovn/egressip_test.go
tssurya: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5310/commits/fbbd0660413b1417df40273acb06d90b4163db40#r2176752616 part of my comment was resolved here

go-controller/pkg/ovn/egressip_test.go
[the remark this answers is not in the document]
tssurya: @coderabbitai why should i use chanels versus panics?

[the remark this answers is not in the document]
tssurya: @martinkennelly let's go with channels instead of panic

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -3410,6 +3410,167 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+					},
+					&nbdb.LogicalSwitch{
+						UUID:  node.Name + "-UUID",
+						Name:  node.Name,
+						Ports: []string{"k8s-" + node.Name + "-UUID"},
+					},
+				}
+				gomega.Eventually(fakeOvn.nbClient).Should(libovsdbtest.HaveData(expectedDatabaseState))
+				// create no-op updates that trigger reconcile for the selected pod async but update should continue to select the pod and not alter pod config
+				go func() {
```
tssurya, line 3544: @martinkennelly also add a comment for why you had to do the update inside another goroutine and the race you are trying to hit here

go-controller/pkg/ovn/egressip_udn_l2_test.go
```diff
@@ -2558,4 +2558,490 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
...
+				node1IPv4CIDR := node1IPv4 + "/24"
+				node2IPv4 := "192.168.126.51"
+				node2IPv4CIDR := node2IPv4 + "/24"
+				_, node1CDNSubnet, _ := net.ParseCIDR(v4Node1Subnet)
+				_, node1UDNSubnet, _ := net.ParseCIDR(v4Node1Net1)
+				nadName := util.GetNADName(eipNamespace2, nadName1)
+				egressCDNNamespace := newNamespaceWithLabels(eipNamespace, egressPodLabel)
+				egressUDNNamespace := newUDNNamespaceWithLabels(eipNamespace2, egressPodLabel)
+				egressPodCDNLocal := *newPodWithLabels(eipNamespace, podName, node1Name, podV4IP, egressPodLabel)
+				egressPodUDNLocal := *newPodWithLabels(eipNamespace2, podName2, node1Name, v4Pod1IPNode1Net1, egressPodLabel)
```
tssurya, line 2579: so pod1 and pod3 are CDN pods and pod2 is UDN L2 pod.

go-controller/pkg/ovn/egressip_test.go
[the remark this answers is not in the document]
martinkennelly: done - thanks @coderabbitai

go-controller/pkg/ovn/egressip_udn_l2_test.go
```diff
@@ -2558,4 +2558,490 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
...
+						},
+					},
+					getNoReRouteReplyTrafficPolicy(ovntypes.DefaultNetworkName, DefaultNetworkControllerName),
+					getDefaultQoSRule(false, ovntypes.DefaultNetworkName, DefaultNetworkControllerName),
+					egressSVCServedPodsASv4,
+					egressIPServedPodsASCDNv4,
+					egressNodeIPsASv4,
+
+					// UDN
+					getReRoutePolicyForController(egressIPName, eipNamespace2, podName2, v4Pod1IPNode1Net1, eIP1Mark, IPFamilyValueV4, []string{gwIP}, netInfo.GetNetworkName(), DefaultNetworkControllerName),
```
tssurya, line 2852: so this is the UDN L2 pod LRP here

go-controller/pkg/ovn/egressip_udn_l2_test.go
```diff
@@ -2558,4 +2558,490 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
...
+					},
+					&nbdb.LogicalRouter{
+						UUID:        netInfo.GetNetworkScopedGWRouterName(node1.Name) + "-UUID",
+						Name:        netInfo.GetNetworkScopedGWRouterName(node1.Name),
+						Ports:       []string{ovntypes.RouterToSwitchPrefix + networkName1_ + layer2SwitchName + "-UUID"},
+						ExternalIDs: map[string]string{ovntypes.NetworkExternalID: netInfo.GetNetworkName(), ovntypes.TopologyExternalID: ovntypes.Layer2Topology},
+						Policies: []string{"udn-default-no-reroute-node-UUID", "udn-default-no-reroute-UUID",
+							"udn-no-reroute-service-UUID", "udn-enabled-svc-no-reroute-UUID",
+							fmt.Sprintf("%s-no-reroute-reply-traffic", netInfo.GetNetworkName()),
+							getReRoutePolicyUUID(eipNamespace2, podName2, IPFamilyValueV4, netInfo.GetNetworkName()),
```
tssurya, line 2891: and this is the UDN LRP for L2 UDN pod2

go-controller/pkg/ovn/egressip_udn_l2_test.go
```diff
@@ -2558,4 +2558,490 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
...
+						Networks: []string{node1Network1RtoSIPCIDR},
+					},
+					&nbdb.LogicalRouter{
+						UUID:        netInfo.GetNetworkScopedGWRouterName(node1.Name) + "-UUID",
+						Name:        netInfo.GetNetworkScopedGWRouterName(node1.Name),
+						Ports:       []string{ovntypes.RouterToSwitchPrefix + networkName1_ + layer2SwitchName + "-UUID"},
+						ExternalIDs: map[string]string{ovntypes.NetworkExternalID: netInfo.GetNetworkName(), ovntypes.TopologyExternalID: ovntypes.Layer2Topology},
+						Policies: []string{"udn-default-no-reroute-node-UUID", "udn-default-no-reroute-UUID",
+							"udn-no-reroute-service-UUID", "udn-enabled-svc-no-reroute-UUID",
+							fmt.Sprintf("%s-no-reroute-reply-traffic", netInfo.GetNetworkName()),
```
tssurya, line 3020: and LRP gone.. yayyyy

tssurya (commented):
commit2 is LGTM. nice bug fix and nice test.
comments are just walking through the test workflow nothing to react on for the author

go-controller/pkg/ovn/egressip.go
martinkennelly: For L3 UDN, and if we are only talking about the gateway LRPs, we add LRP to pkt mark the pod and `loadedEgressNode` && `isLocalZoneEgressNode`:
```
func addPodEgressIPAssignment() {
....
ops, err = e.createGWMarkPolicyOps(ni, ops, podIPs, status, mark, pod.Namespace, pod.Name, egressIPName)
				if err != nil {
					return fmt.Errorf("unable to create GW router LRP ops to packet mark pod %s/%s: %v", pod.Namespace, pod.Name, err)
				}
```

On the delete path we just nuke entries if `loadedEgressNode` && `isLocalZoneEgressNode`:
```
func deletePodEgressIPAssignment()
if ni.IsSecondary() && ni.TopologyType() == types.Layer3Topology {
  ops, err = e.deleteGWMarkPolicyOps(ni, ops, status, pod.Namespace, pod.Name, egressIPName)
  if err != nil {
           return fmt.Errorf("unable to create GW router packet mark LRPs delete ops for pod %s/%s: %v", pod.Namespace, pod.Name, err)
  }
}
```

So we are good for that.

martinkennelly:
@trozet Tim, you and I discussed the possibility of stale eip config here if we failed to get an active net: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4530#discussion_r1842983546

I added this commit on top in response: 35d3732ec2ab9818041dd0f67f445e42c093ca3e

Here, we save the network the pod was config'd on originally to the eip cache and when we try to setup a new pod, we look for an entry in that eip cache and if one is present, and its network is different, we try to tear down the eip config.

Lets chat when youre back regarding if this is needed.

go-controller/pkg/ovn/egressip.go
tssurya: `hmm on L3 UDNs marking is done both on ovn-cluster-router and GR?`
yes this is correct for my future self:

1. remote EIP pods don't have LRPs where marking can be done - we only do LRSR's so we do a LRP on GR for marking in that case
2. local EIP pods have LRPs in CR but I Think here we do 1008 marking ? so maybe that's why we also have GR marking in this case.
anyhow that explains why marking is done on GR for L3 UDNs.

tssurya:
> @trozet Tim, you and I discussed the possibility of stale eip config here if we failed to get an active net: [#4530 (comment)](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4530#discussion_r1842983546)
> 
> I added this commit on top in response: [35d3732](https://github.com/ovn-kubernetes/ovn-kubernetes/commit/35d3732ec2ab9818041dd0f67f445e42c093ca3e)
> 
> Here, we save the network the pod was config'd on originally to the eip cache and when we try to setup a new pod, we look for an entry in that eip cache and if one is present, and its network is different, we try to tear down the eip config.
> 
> Lets chat when youre back regarding if this is needed.

thanks @martinkennelly ! @trozet yea our doubt is can the network actually be different? if so when?

go-controller/pkg/ovn/egressip.go
tssurya: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5310#issuecomment-3023236903 it seems the code came from here @trozet can help chime in when is back.

tssurya:
```
Summarizing 1 Failure:
  [FAIL] e2e br-int flow monitoring export validation Should validate flow data of br-int is sent to an external gateway [It] with sflow
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:1988

Ran 125 of 504 Specs in 3101.048 seconds
FAIL! -- 124 Passed | 1 Failed | 2 Pending | 377 Skipped
``` 
unrelated flake retesting after run succeeds

go-controller/pkg/ovn/egressip.go
```diff
@@ -2451,11 +2451,18 @@ func (e *EgressIPController) deletePodEgressIPAssignment(ni util.NetInfo, egress
 		return err
 	}
 	var ops []ovsdb.Operation
-	if !loadedPodNode || isLocalZonePod { // node is deleted (we can't determine zone so we always try and nuke OR pod is local to zone)
+	// For CDN only, add SNATs to support external GW feature
+	if ni.IsDefault() && (!loadedPodNode || isLocalZonePod) {
```
tssurya, line 2455: this is the per pod SNATs huh? ok makes sense since we don't have it for UDNs..but its a bit weird small catches like this are slipping into this PR :D looks like we have some gaps in testing that can come later. I consider this fix outside this PR's scope but OK to get this in logic wise

tssurya (approved):

Issue #5283, opened 2025-06-05
vlan interface is not interpreted properly


zlangi:
### What happened?

I have been trying to deploy the ovn-k8s via helm chart to a single node to test the ovs-doca functionalities on a ConnectX7. In our deployment we are using vlans even our management network is a vlan where the k8s nodes are talking to each other.
currently I have link redundancy set up on this connectx7 so it's represented as bond0 interface where I have the management ip address of 10.25.41.24 tagged by the vlan of 2726.
I deployed a new k8s deployment via kubeadm to the single node then I deployed the cni the following way: helm install ovn-kubernetes . -f values-no-ic.yaml --set k8sAPIServer="https://10.25.41.24:6443/" --set ovnkube-identity.replicas=$(kubectl get node -l [node-role.kubernetes.io/control-plane](http://node-role.kubernetes.io/control-plane) --no-headers | wc -l) --set global.image.repository=[ghcr.io/ovn-kubernetes/ovn-kubernetes/ovn-kube-ubuntu](http://ghcr.io/ovn-kubernetes/ovn-kubernetes/ovn-kube-ubuntu) --set global.image.tag=master
all components started except the ovnkube-node component went into a crash loop with the following:
`I0520 10:53:24.349566   15411 gateway_init.go:330] Initializing Gateway Functionality for Gateway PreStart
I0520 10:53:24.350208   15411 helper_linux.go:118] Found default gateway interface brbond0.2726 10.25.41.1
I0520 10:53:24.350607   15411 gateway_init.go:367] Preparing Gateway
I0520 10:53:24.350624   15411 gateway_shared_intf.go:2372] Creating new gateway
I0520 10:53:24.350669   15411 ovs.go:159] Exec(14): /usr/bin/ovs-vsctl --timeout=15 port-to-br brbond0.2726
I0520 10:53:24.355703   15411 ovs.go:162] Exec(14): stdout: ""
I0520 10:53:24.355720   15411 ovs.go:163] Exec(14): stderr: "ovs-vsctl: no port named brbond0.2726\n"
I0520 10:53:24.355726   15411 ovs.go:165] Exec(14): err: exit status 1
I0520 10:53:24.355736   15411 ovs.go:159] Exec(15): /usr/bin/ovs-vsctl --timeout=15 br-exists brbond0.2726
I0520 10:53:24.359466   15411 ovs.go:162] Exec(15): stdout: ""
[...]

PR #5312, opened 2025-06-17
Use forward slash as path separator for some sysctl commands

hedgieinsocks:
## 📑 Description
Convert `.` path separators to `/` when enabling forwarding for bridgeName, interfaceName and mgmtPortName to avoid errors when those names contain `.` characters, e.g. `foo.200`

Fixes: #5283

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/gateway_init.go
+++ b/go-controller/pkg/node/gateway_init.go
@@ -27,8 +27,10 @@ import (
 func bridgedGatewayNodeSetup(nodeName, bridgeName, physicalNetworkName string) (string, error) {
 	// IPv6 forwarding is enabled globally
 	if config.IPv4Mode {
-		stdout, stderr, err := util.RunSysctl("-w", fmt.Sprintf("net.ipv4.conf.%s.forwarding=1", bridgeName))
-		if err != nil || stdout != fmt.Sprintf("net.ipv4.conf.%s.forwarding = 1", bridgeName) {
+		// we use forward slash as path separator to allow dotted bridgeName e.g. foo.200
+		stdout, stderr, err := util.RunSysctl("-w", fmt.Sprintf("net/ipv4/conf/%s/forwarding=1", bridgeName))
+		// systctl output enforces dot as path separator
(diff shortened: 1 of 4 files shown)


trozet (changes requested):
thanks @hedgieinsocks for working on this. There is an issue with how sysctl returns output, leading to deploys failing:

F0618 20:08:08.892258    1968 ovnkube.go:138] failed to run ovnkube: [failed to start network controller: failed to start default network controller: error running OVN Kubernetes Services controller: handler {0x1ea65c0 0x1ea62a0 0x1ea6240} was not added to shared informer because it has stopped already, failed to start node network controller: failed to init default node network controller: bridge for interface failed for breth0: failed to set up shared interface gateway: could not set the correct forwarding value for interface breth0: stdout: net.ipv4.conf.breth0.forwarding = 1, stderr: , err: <nil>]
info: Waiting for process_ready ovnkube-controller-with-node to come up, waiting 5s ...

Example:

trozet@fedora:~/go/src/github.com/ovn-org/ovn-kubernetes/contrib$ sudo sysctl -w  net/ipv4/conf/eno1.100/forwarding=1
net.ipv4.conf.eno1/100.forwarding = 1

In this case even though I use slashes, sysctl actually prints periods and subsitutes my subinterface period with a slash. Perhaps its better to leave the other uses of . alone and only focus on the places where an interface is used. We could use os.Read/Write to directly read and write the values to those files, rather than relying on sysctl. Might make it easier than having to handle this whacky behavior from sysctl for vlan subinterfaces.

hedgieinsocks:
Hm, weird, either I musunderstood man  https://man7.org/linux/man-pages/man5/sysctl.d.5.html#top_of_page or missed some detail. It couldn't be some misinformation.

hedgieinsocks:
Oh, I guess I understood this whacky situation. The sysctl output seems to always use dots for path separators no matter the input. But the command itself succeeds.

So we have only 3 (I think) places where we use arbitrary names and where it can fail. So I will try to "normalize" only them. 

As for rewriting RunSysctl and read/write directly, I'm afraid it would be more than I'm ready to take.

trozet (approved):

PR #5315, opened 2025-06-18
[DNM] Gate ovn controller always for IC until ovnk syncs - option 2

martinkennelly:
For IC, allow ovn-controller to connect to OVN DB only when ovnkube controller is up and has syncd and changes propagated to SB DB and ready to be consumed by ovn-controller.


martinkennelly:
closed in favor of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5373

PR #5320, opened 2025-06-19
udn, pre assigned port net ids: provision the default net NAD CR

maiqueb:
## 📑 Description

The "pre assigned port net ids" feature requires a NAD for the `default` network to be provisioned. This PR provisions that NAD whenever the cluster manager starts.

Fixes #

Depends-on: 
- #5332 

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Added automatic provisioning of a default network attachment when specific features are enabled.
  * Introduced configuration checks to determine if preconfigured user-defined network addresses are supported.

* **Bug Fixes**
  * Ensured that the default network attachment is not created when related features are disabled.

* **Tests**
  * Improved test coverage for scenarios involving automatic provisioning of network attachments based on feature flags.

### DIFF
--- a/go-controller/pkg/clustermanager/clustermanager_test.go
+++ b/go-controller/pkg/clustermanager/clustermanager_test.go
@@ -34,10 +34,9 @@ const (
 
 var _ = ginkgo.Describe("Cluster Manager", func() {
 	var (
-		app      *cli.App
-		f        *factory.WatchFactory
-		stopChan chan struct{}
-		wg       *sync.WaitGroup
+		app *cli.App
+		f   *factory.WatchFactory
+		wg  *sync.WaitGroup
 	)
 
 	const (
@@ -54,12 +53,10 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
 		app = cli.NewApp()
 		app.Name = "test"
 		app.Flags = config.Flags
-		stopChan = make(chan struct{})
 		wg = &sync.WaitGroup{}
 	})
 
 	ginkgo.AfterEach(func() {
-		close(stopChan)
 		if f != nil {
 			f.Shutdown()
 		}
@@ -1436,4 +1433,102 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
 		})
 	})
 
+	ginkgo.Context("starting the cluster manager", func() {
+		const networkName = "default"
+
+		var fakeClient *util.OVNClusterManagerClientset
+
+		ginkgo.BeforeEach(func() {
+			fakeClient = util.GetOVNClientset().GetClusterManagerClientset()
+		})
+
+		ginkgo.When("the required features are not enabled", func() {
+			ginkgo.It("does *not* automatically provision a NAD for the default network", func() {
+				app.Action = func(ctx *cli.Context) error {
+					_, err := config.InitConfig(ctx, nil, nil)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					f, err = factory.NewClusterManagerWatchFactory(fakeClient)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					clusterMngr, err := clusterManager(fakeClient, f)
+					gomega.Expect(clusterMngr).NotTo(gomega.BeNil())
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+					gomega.Expect(clusterMngr.Start(ctx.Context)).To(gomega.Succeed())
+
+					_, err = fakeClient.NetworkAttchDefClient.
+						K8sCniCncfIoV1().
+						NetworkAttachmentDefinitions(config.Kubernetes.OVNConfigNamespace).
+						Get(
+							context.Background(),
+							networkName,
+							metav1.GetOptions{},
+						)
+					gomega.Expect(err).To(
+						gomega.MatchError("network-attachment-definitions.k8s.cni.cncf.io \"default\" not found"),
+					)
+
+					return nil
+				}
+				gomega.Expect(app.Run([]string{app.Name})).To(gomega.Succeed())
+			})
+		})
+
+		ginkgo.When("the multi-network, network-segmentation, and preconfigured-udn-addresses features are enabled", func() {
+			ginkgo.BeforeEach(func() {
+				config.OVNKubernetesFeature.EnableMultiNetwork = true
+				config.OVNKubernetesFeature.EnableNetworkSegmentation = true
+				config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses = true
+			})
+
+			ginkgo.It("automatically provisions a NAD for the default network", func() {
+				app.Action = func(ctx *cli.Context) error {
+					_, err := config.InitConfig(ctx, nil, nil)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					f, err = factory.NewClusterManagerWatchFactory(fakeClient)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					clusterMngr, err := clusterManager(fakeClient, f)
+					gomega.Expect(clusterMngr).NotTo(gomega.BeNil())
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					c, cancel := context.WithCancel(ctx.Context)
+					defer cancel()
+					gomega.Expect(clusterMngr.Start(c)).To(gomega.Succeed())
+					defer clusterMngr.Stop()
+
+					nad, err := fakeClient.NetworkAttchDefClient.
+						K8sCniCncfIoV1().
+						NetworkAttachmentDefinitions(config.Kubernetes.OVNConfigNamespace).
+						Get(
+							context.Background(),
+							networkName,
+							metav1.GetOptions{},
+						)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					const expectedNADContents = `{"cniVersion": "0.4.0", "name": "ovn-kubernetes", "type": "ovn-k8s-cni-overlay"}`
+					gomega.Expect(nad.Spec.Config).To(gomega.Equal(expectedNADContents))
+
+					return nil
+				}
+				gomega.Expect(app.Run([]string{app.Name})).To(gomega.Succeed())
+			})
+		})
+	})
+
 })
+
+func clusterManager(client *util.OVNClusterManagerClientset, f *factory.WatchFactory) (*ClusterManager, error) {
+	if err := f.Start(); err != nil {
+		return nil, fmt.Errorf("failed to start the CM watch factory: %w", err)
+	}
+
+	clusterMngr, err := NewClusterManager(client, f, "identity", nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to start the CM watch factory: %w", err)
+	}
+
+	return clusterMngr, nil
+}
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -1016,7 +1016,7 @@ func (c *Controller) getSelectedNADs(networkSelectors apitypes.NetworkSelectors)
 		case apitypes.DefaultNetwork:
 			// if we are selecting the default networkdefault network label,
 			// make sure a NAD exists for it
-			nad, err := c.getOrCreateDefaultNetworkNAD()
+			nad, err := util.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient)
 			if err != nil {
 				return nil, fmt.Errorf("failed to get/create default network NAD: %w", err)
 			}
@@ -1045,34 +1045,6 @@ func (c *Controller) getSelectedNADs(networkSelectors apitypes.NetworkSelectors)
 	}
 
 	return selected, nil
-}
-
-// getOrCreateDefaultNetworkNAD ensure that a well-known NAD exists for the
-// default network in ovn-k namespace.
-func (c *Controller) getOrCreateDefaultNetworkNAD() (*nadtypes.NetworkAttachmentDefinition, error) {
-	nad, err := c.nadLister.NetworkAttachmentDefinitions(config.Kubernetes.OVNConfigNamespace).Get(types.DefaultNetworkName)
-	if err != nil && !apierrors.IsNotFound(err) {
-		return nil, err
-	}
-	if nad != nil {
-		return nad, nil
-	}
-	return c.nadClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(config.Kubernetes.OVNConfigNamespace).Create(
-		context.Background(),
-		&nadtypes.NetworkAttachmentDefinition{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      types.DefaultNetworkName,
-				Namespace: config.Kubernetes.OVNConfigNamespace,
-			},
-			Spec: nadtypes.NetworkAttachmentDefinitionSpec{
-				Config: fmt.Sprintf("{\"cniVersion\": \"0.4.0\", \"name\": \"ovn-kubernetes\", \"type\": \"%s\"}", config.CNI.Plugin),
-			},
-		},
-		// note we don't set ourselves as field manager for this create as we
-		// want to process the resulting event that would otherwise be filtered
-		// out in nadNeedsUpdate
-		metav1.CreateOptions{},
-	)
 }
 
 // getEgressIPsByNodesByNetworks iterates all existing egress IPs that apply to
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -150,6 +150,12 @@ func (c *Controller) Run() error {
 		return fmt.Errorf("unable to start user-defined network controller: %v", err)
 	}
 
+	if util.IsPreconfiguredUDNAddressesEnabled() {
+		if _, err := util.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient); err != nil {
+			return fmt.Errorf("failed to ensure default network nad exists: %w", err)
+		}
+	}
+
 	return nil
 }
 
--- a/go-controller/pkg/util/multi_network.go
+++ b/go-controller/pkg/util/multi_network.go
@@ -1378,6 +1378,12 @@ func IsRouteAdvertisementsEnabled() bool {
 	return config.OVNKubernetesFeature.EnableMultiNetwork && config.OVNKubernetesFeature.EnableRouteAdvertisements
 }
 
+// IsPreconfiguredUDNAddressesEnabled indicates if user defined IPs / MAC
+// addresses can be set in primary UDNs
+func IsPreconfiguredUDNAddressesEnabled() bool {
+	return IsNetworkSegmentationSupportEnabled() && config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses
+}
+
 func DoesNetworkRequireIPAM(netInfo NetInfo) bool {
 	return !((netInfo.TopologyType() == types.Layer2Topology || netInfo.TopologyType() == types.LocalnetTopology) && len(netInfo.Subnets()) == 0)
 }
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller_test.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller_test.go
@@ -1051,7 +1051,7 @@ func TestController_reconcile(t *testing.T) {
 			g.Expect(err).ToNot(gomega.HaveOccurred())
 			// prime the default network NAD
 			if defaultNAD == nil {
-				defaultNAD, err = c.getOrCreateDefaultNetworkNAD()
+				defaultNAD, err = util.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient)
 				g.Expect(err).ToNot(gomega.HaveOccurred())
 				// update it with the annotation that network manager would set
 				defaultNAD.Annotations = map[string]string{types.OvnNetworkNameAnnotation: types.DefaultNetworkName}
--- a/go-controller/pkg/util/nad.go
+++ b/go-controller/pkg/util/nad.go
@@ -1,0 +1,46 @@
+package util
+
+import (
+	"context"
+	"fmt"
+
+	nadtypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
+	nadclientset "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned"
+	nadlisters "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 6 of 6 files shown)


maiqueb:
@qinqon / @kyrtapz FYI. 

I plan on rebasing this on top of the PR that adds the OVN-K feature configuration flag, thus ensuring the default network NAD is only provisioned if this feature is enabled.

RamLavi:
Is `pre assigned port net ids` feature a formal name? If you mean the routed ingress, can you stick to that name or perhaps link it on the PR desc? It threw me off on a journey to find this new feature.
Also, `port net ids` you mean IPs and MACs right?

go-controller/pkg/util/multi_network.go
```diff
@@ -1557,3 +1557,9 @@ func ParseNetworkName(networkName string) (udnNamespace, udnName string) {
 	}
 	return "", ""
 }
+
+// IsPrimaryNetworkCustomizationEnabled indicates if user defined IPs / MAC
+// addresses can be set in primary UDNs
+func IsPrimaryNetworkCustomizationEnabled() bool {
+	return IsNetworkSegmentationSupportEnabled() && config.OVNKubernetesFeature.EnableCustomNetworkConfig
+}
```
qinqon: Is it worth it to move this to the feature gate PR ?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -674,6 +675,13 @@ func (c *Controller) updateClusterUDNStatus(cudn *userdefinednetworkv1.ClusterUs
 	return nil
 }
 
+func (c *Controller) EnsureDefaultNetworkNAD() error {
+	if _, err := defaultnetworkcustomization.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient); err != nil {
+		return fmt.Errorf("unable to ensure default network nad exists: %w", err)
+	}
```
qinqon: Let's use "failed" prefix like 90% of errors do that and is easier to grep
```
failed ensuring...
``

File: go-controller/pkg/clustermanager/routeadvertisements/controller.go [hunk 3]
```diff
@@ -1012,7 +1013,7 @@ func (c *Controller) getSelectedNADs(networkSelectors apitypes.NetworkSelectors)
 		case apitypes.DefaultNetwork:
 			// if we are selecting the default networkdefault network label,
 			// make sure a NAD exists for it
-			nad, err := c.getOrCreateDefaultNetworkNAD()
+			nad, err := defaultnetworkcustomization.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient)
 			if err != nil {
 				return nil, fmt.Errorf("failed to get/create default network NAD: %w", err)
```
qinqon, line 1021: Change it to `failed to ensure...`

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return nadClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(config.Kubernetes.OVNConfigNamespace).Create(
+		context.Background(),
+		&nadtypes.NetworkAttachmentDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      types.DefaultNetworkName,
+				Namespace: config.Kubernetes.OVNConfigNamespace,
+			},
+			Spec: nadtypes.NetworkAttachmentDefinitionSpec{
+				Config: fmt.Sprintf("{\"cniVersion\": \"0.4.0\", \"name\": \"ovn-kubernetes\", \"type\": \"%s\"}", config.CNI.Plugin),
```
qinqon: I know we are just moving code around, but now that we are a it It may be better to use the NetConf struct and marshal it, I think there are helpers around for that.

qinqon (changes requested):
very nitty

go-controller/pkg/util/multi_network.go
maiqueb: I don't care. Could go either way - but adding code that is not used feels weird.

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
maiqueb: I'm literally moving code around.

This can happen in a follow-up if you're up to it.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
maiqueb: this code was already there.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
maiqueb: Done.

qinqon (dismissed):
LGTM

go-controller/pkg/clustermanager/clustermanager.go
```diff
@@ -216,6 +216,12 @@ func (cm *ClusterManager) Start(ctx context.Context) error {
 		if err := cm.userDefinedNetworkController.Run(); err != nil {
 			return err
 		}
+
+		if util.IsPrimaryNetworkCustomizationEnabled() {
+			if err := cm.userDefinedNetworkController.EnsureDefaultNetworkNAD(); err != nil {
```
ormergi: Note that the cluster-manager factory (`NewClusterManager`)  already has clients, and it pass the relevent ones to each controller.
It seems `userDefinedNetworkController` have this method because the NAD client is not accessible in this context.
Please consider avoiding this "round trip" by adding the NAD client to `ClusterManager{}`, this way it can call the func at `ensurenad.go` directly.

go-controller/pkg/util/multi_network.go
```diff
@@ -1557,3 +1557,9 @@ func ParseNetworkName(networkName string) (udnNamespace, udnName string) {
 	}
 	return "", ""
 }
+
+// IsPrimaryNetworkCustomizationEnabled indicates if user defined IPs / MAC
+// addresses can be set in primary UDNs
+func IsPrimaryNetworkCustomizationEnabled() bool {
```
ormergi: nits: 
Do we need to update the name following the FG name change?
Maybe `IsPreconfiguredUDNAddressEnabled`?
I would move it closer to `IsPrimaryNetworkCustomizationEnabled`

go-controller/pkg/clustermanager/clustermanager_test.go
```diff
@@ -1436,4 +1433,102 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
...
+						gomega.MatchError("network-attachment-definitions.k8s.cni.cncf.io \"default\" not found"),
+					)
+
+					return nil
+				}
+				gomega.Expect(app.Run([]string{app.Name})).To(gomega.Succeed())
+			})
+		})
+
+		ginkgo.When("the multi-network, network-segmentation, are customized-network features are enabled", func() {
```
ormergi: nit:
```suggestion
ginkgo.When("the multi-network, network-segmentation, are preconfigurfed-udn-addresses features are enabled", func() {
```

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -674,6 +675,13 @@ func (c *Controller) updateClusterUDNStatus(cudn *userdefinednetworkv1.ClusterUs
 	return nil
 }
 
+func (c *Controller) EnsureDefaultNetworkNAD() error {
+	if _, err := defaultnetworkcustomization.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient); err != nil {
+		return fmt.Errorf("failed to ensure default network nad exists: %w", err)
+	}
+	return nil
+}
+
```
ormergi: This file is already bloated, please consider moving this method to new file.
Or removing altogether please see https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5320#discussion_r2197035039

go-controller/pkg/clustermanager/clustermanager.go
```diff
@@ -216,6 +216,12 @@ func (cm *ClusterManager) Start(ctx context.Context) error {
 		if err := cm.userDefinedNetworkController.Run(); err != nil {
 			return err
 		}
+
+		if util.IsPrimaryNetworkCustomizationEnabled() {
```
ormergi: What about PersistnetIPs FG, isnt it required for this feature to work?

go-controller/pkg/clustermanager/clustermanager.go
maiqueb: ~~if you prefer I can directly invoke `defaultnetworkcustomization.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient)` and not introduce this helper method.~~

Changing factory functions because of this seems overkill. IMHO.

I'll at least wait for a maintainer's opinion on this; it can be done.

go-controller/pkg/clustermanager/clustermanager.go
maiqueb: No, this feature is about allowing a "user" to specify MACs and/or IPs for primary UDN attachments. Allowing VM's interface IPs to persist is a different feature, which we (OpenShift Virt) also want to use.

go-controller/pkg/util/multi_network.go
maiqueb: > Do we need to update the name following the FG name change?
Maybe IsPreconfiguredUDNAddressEnabled?

I can do that. 

> I would move it closer to `IsPrimaryNetworkCustomizationEnabled`

Would you elaborate why ? I fail to see the advantage. Still, I assume this is a nit.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
maiqueb: I think it's good as it is. Feels very strange to have a NAD client and lister in cluster manager just because of this.

I can see creating a required NAD for customizing the primary UDN as responsibility of the UDN controller. 

IMHO, it's up for a maintainer to decide. It can go either way.

go-controller/pkg/clustermanager/clustermanager_test.go
maiqueb: Done. (sans typo)

go-controller/pkg/util/multi_network.go
maiqueb: Did both.

go-controller/pkg/clustermanager/clustermanager_test.go
```diff
@@ -54,12 +53,10 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
 		app = cli.NewApp()
 		app.Name = "test"
 		app.Flags = config.Flags
-		stopChan = make(chan struct{})
```
qinqon, line 57: This may be right, but do not look related to the change.

go-controller/pkg/clustermanager/clustermanager_test.go
```diff
@@ -1436,4 +1433,102 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
...
+
+					f, err = factory.NewClusterManagerWatchFactory(fakeClient)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					clusterMngr, err := clusterManager(fakeClient, f)
+					gomega.Expect(clusterMngr).NotTo(gomega.BeNil())
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					c, cancel := context.WithCancel(ctx.Context)
+					defer cancel()
```
qinqon, line 1497: Do we really need this ?

go-controller/pkg/clustermanager/clustermanager_test.go
maiqueb: trying to be good boyscout here. if you want, I can put this useless channel back into the file. I won't bother pushing a dedicated PR to address that.

go-controller/pkg/clustermanager/clustermanager_test.go
maiqueb: I'm just copying the pattern I've seen in the other tests in this file, since I assume that gives me higher changes of this test being accepted.

go-controller/pkg/clustermanager/clustermanager_test.go
qinqon: Nah all good.

go-controller/pkg/clustermanager/clustermanager_test.go
qinqon: fair enough

qinqon (approved):

go-controller/pkg/clustermanager/clustermanager.go
```diff
@@ -216,6 +216,12 @@ func (cm *ClusterManager) Start(ctx context.Context) error {
 		if err := cm.userDefinedNetworkController.Run(); err != nil {
 			return err
 		}
+
+		if util.IsPreconfiguredUDNAddressesEnabled() {
+			if err := cm.userDefinedNetworkController.EnsureDefaultNetworkNAD(); err != nil {
```
kyrtapz: NIT: any particular reason for not calling this from the inside of `userDefinedNetworkController.Run()`?

kyrtapz (dismissed):
LGTM

go-controller/pkg/clustermanager/clustermanager.go
maiqueb: I'm executing from the place where I saw other people checking if the feature's requirements are being called. 

I can move it there, but I don't see the advantage.

ormergi: If you like to keep UDN CRDs controller to mange the default NAD instead of cluster-manager controller, having it in Run() looks like a better fit to me because it needs to be done once when the starts, Run() does that.

maiqueb:
Rebased latest main in this [forced push](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/3a3a45c8799c15d7edd59303a23190f02fbac99f..1e8f2875663da813b796e0987c43453d9fbb155f).

go-controller/pkg/clustermanager/clustermanager.go
tssurya: I'd prefer to move it inside the `userDefinedNetworkController.Run()`call. (similar to how it was being done within RA not in CM)
Why? because we seem to be leveraging that controller after it has started to then call EnsureDefaultNAD.

So rule of thumb for maintainence is usually anything on the CM level that we are "starting or running" can be done in CM.. any functionality that is actually done within the specific controller that has already started up can be moved within.

go-controller/pkg/clustermanager/clustermanager_test.go
```diff
@@ -1436,4 +1433,102 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
...
+
+					const expectedNADContents = `{"cniVersion": "0.4.0", "name": "ovn-kubernetes", "type": "ovn-k8s-cni-overlay"}`
+					gomega.Expect(nad.Spec.Config).To(gomega.Equal(expectedNADContents))
+
+					return nil
+				}
+				gomega.Expect(app.Run([]string{app.Name})).To(gomega.Succeed())
+			})
+		})
+	})
```
tssurya, line 1519: nice tests! thank you

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
```diff
@@ -0,0 +1,44 @@
+package defaultnetworkcustomization
```
tssurya: ~do we really want a new package just for this util? `defaultnetworkcustomization` is going to be confusing given its not a specific feature...
given we are leveraging cm.UserDefinedNetworkController.Run() does it make more sense to just put this 1 util into pkg/clustermanager/userdefinednetwork/nad.go ? I understand the reason you have it here is because its part of "Default Network" but the multus meaning of this is "whatever is the primary network for the pod - whether its PDUN and CDN" and in that sense given we are planning to use this feature with L2 UDNs for me its better to scope it within userdefinednetwork package itself unless there are more utils planned to be added to this file in future PRs.~

so I assume the reason we put this in a new file is because its used from both RA feature and from UDNPreconfigured feature? - so we are not exactly always creating it.. its still bounded by those two features I guess and that is also why placing this independently under defaultnetworkcustomization is also confusing

Are we planning on adding more things into "ensurenad.go?" the file name for me is very very narrow scoped..

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
```diff
@@ -0,0 +1,44 @@
...
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+)
+
+// EnsureDefaultNetworkNAD ensure that a well-known NAD exists for the
+// default network in ovn-k namespace.
```
tssurya: and this will be used for the "active primary network in a namespace right?" not just the default network? - can we call out the scope change from just being for only _default- CDN_ to being for whatever is the default network of the pod in the function comments?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -1016,7 +1017,7 @@ func (c *Controller) getSelectedNADs(networkSelectors apitypes.NetworkSelectors)
 		case apitypes.DefaultNetwork:
 			// if we are selecting the default networkdefault network label,
 			// make sure a NAD exists for it
-			nad, err := c.getOrCreateDefaultNetworkNAD()
+			nad, err := defaultnetworkcustomization.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient)
```
tssurya: so I guess handles the case where let's say UDN feature is not enabled but BGP is used with default network which requires this NAD.. so this would create it?

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
tssurya: @coderabbitai: do you have some suggestions for a good name for the file or a better place to put this function?

tssurya: maybe `util.go` is not a bad place.. I see other features having similar functionality related functions there: 

1. pkg/util/egressfirewall.go
2. pkg/util/dns.go
3. pkg/util/arp.go

maiqueb: > so I assume the reason we put this in a new file is because its used from both RA feature and from UDNPreconfigured feature? - so we are not exactly always creating it.. its still bounded by those two features I guess and that is also why placing this independently under defaultnetworkcustomization is also confusing

Yes, I've moved this to a common place to allow me to reuse that code.

> Are we planning on adding more things into "ensurenad.go?" the file name for me is very very narrow scoped..

No, I don't plan on adding anything else here. I just don't see what's the issue with having a narrow scoped file name. It's not like we pay more for each file where the code is written. IMHO, that's just good organization. 

> maybe util.go is not a bad place.. I see other features having similar functionality related functions there:

    pkg/util/egressfirewall.go
    pkg/util/dns.go
    pkg/util/arp.go

`util` packages, IMHO, are junkyards waiting to be. It's literally the place where everyone throws things they don't know what to name, or where unrelated code ends up being. Most common issue I see with it is low cohesion. 

High cohesion means that a module's responsibilities are closely related and focused, which promotes maintainability, reusability, and code readability. Utils packages violate this principle by grouping unrelated functions together simply because they don't fit elsewhere. 

Plus ... it's a sisifean task to try to refactor it afterwards, because of the dependencies. I usually avoid putting anything in utils.

@tssurya this is my personal opinion - I'm just a code contributor here, and I will not maintain this code. Let me know where you want this to be, and I'll move it there.

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
maiqueb: > and this will be used for the "active primary network in a namespace right?" not just the default network? - can we call out the scope change from just being for only default- CDN to being for whatever is the default network of the pod in the function comments?

Unfortunately, I don't follow your argument. 

This is for customizing the default cluster network attachment (a multus feature). We bend the rules a lil' bit by making this custom data apply to the active network for a namespace, but the point being is that "a well-known NAD" - i.e. `default` - "exists for the default network in ovn-k namespace".

go-controller/pkg/clustermanager/routeadvertisements/controller.go
maiqueb: yes

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
tssurya: let's put it in `pkg/util/nad.go`

seems like here is another file with 1 util: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/util/udn/udn.go and another file here for pods.go: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/util/pod.go with 1 util function.. so that seems to be pattern right now.

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
tssurya: > > and this will be used for the "active primary network in a namespace right?" not just the default network? - can we call out the scope change from just being for only default- CDN to being for whatever is the default network of the pod in the function comments?
> 
> Unfortunately, I don't follow your argument.
> 
> This is for customizing the default cluster network attachment (a multus feature). We bend the rules a lil' bit by making this custom data apply to the active network for a namespace, 

exactly what you said here ^ to reflect that as an additional sentence in the comment on what the default NAD is going to be used for in ovnk

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
maiqueb: I still don't think this is good in the long run, since IMHO it will negatively impact the overall maintainability of the code, but, complying w/ the maintainers request. 

Done.

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
maiqueb: Done.

go-controller/pkg/clustermanager/clustermanager.go
maiqueb: Had forgotten about this one. Moved it into the UDN controller.

Done.

tssurya:
```
Summarizing 1 Failure:
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_localnet.go:94
```
is this a new flake?
hmm https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16595804839/job/46944471798?pr=5320

maiqueb:
> ```
> Summarizing 1 Failure:
>   [FAIL] Network Segmentation: Localnet [It] using ClusterUserDefinedNetwork CR, pods in different namespaces, should communicate over localnet topology
>   /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation_localnet.go:94
> ```
> 
> is this a new flake? hmm https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16595804839/job/46944471798?pr=5320

I guess to - I don't see how my changes can impact localnet code. 

I'll dig deeper if this fails again.

Issue #4221, opened 2024-03-13
Flake: Should validate flow data of br-int is sent to an external gateway [It] with sflow


kyrtapz:
Seen here: https://github.com/ovn-org/ovn-kubernetes/actions/runs/8249902255/job/22564616111

npinaeva:
hit here https://github.com/ovn-org/ovn-kubernetes/actions/runs/8539445598/job/23395440649

qinqon:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/13264316559/job/37029808158?pr=4899

tssurya:
seen here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16673702405/job/47197338115?pr=5321 on BGP lanes frequently.

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16807306450/job/47637490580?pr=5385

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20023857810/job/57417512549?pr=5786

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20370980823/job/58537934602?pr=5796

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20935529810/job/60157871267?pr=5840

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21069078660/job/60594765824?pr=5541

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21069078660/job/60594765833?pr=5541

tssurya:
@pliurh if https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5848 is a fix for this flake, then its good practice to link that PR to the flake and assign yourself to this issue

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21076102893/job/60619027511?pr=5852

pliurh:
Various reasons caused the flakes. #5848 can fix the racing condition for `waitForRollout()`, but it does not resolve other issues.

kyrtapz:
[...]

PR #5321, opened 2025-06-19
Allow configuring static IP and MAC with primary UDNs.

qinqon:
## 📑 Description
Allow a pod's primary Layer 2 UDN network to select static IP and MAC addresses using the Multus default-network annotation.

Fixes https://issues.redhat.com/browse/CORENET-6131

Depends-on:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5320

Notes:
- Collision detection is not included in this PR and will be addressed in a subsequent update.
  - https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5411 
- No additional checks will be added to the CNI, as CNI code respects OVN pod annotations and disregards IP requests.
- Negative test cases for pods will be incorporated by the QE team in a follow-up PR.
- NAD client was bumped but we needed to ping Pin containernetworking/cni v1.2.3 at go.mod looks like future versions will break ovnk

## ✅ Checks

- [x] My code requires tests

## How to verify it
This include unit test and e2e tests

## Summary by CodeRabbit

* **New Features**
  * Added support for static IP and MAC address assignment on user-defined network interfaces for pods and KubeVirt VMs.
  * Enhanced network status reporting to include MTU information.
  * Introduced new end-to-end tests covering static IP/MAC assignment and validation for both pods and VMs.
  * Improved feature flag and environment variable handling for enabling preconfigured UDN addresses.
  * Extended CNI plugin support with new commands (GC, STATUS) and improved error handling.
  * Added network namespace support for multiple OS platforms with platform-specific implementations.
  * Refined network selection logic to incorporate default network annotations when feature enabled.
  * Added utility functions for filtering IPs by cluster IP family support and generating address annotations.

* **Bug Fixes**
  * Improved validation and error reporting for static IP requests in certain network scenarios.

* **Dependency Updates**
  * Upgraded CNI and network-attachment-definition-client dependencies to latest versions.

* **General Improvements**
  * Enhanced test coverage for custom network configurations and static address scenarios.
  * Modernized file I/O and version handling in CNI libraries.
  * Improved utility functions for IP family detection and annotation generation.
  * Refined CNI plugin initialization and configuration parsing with support for multiple CNI versions and flags.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -467,7 +467,7 @@ jobs:
           - {"target": "node-ip-mac-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv6",      "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-disabled"}
           - {"target": "node-ip-mac-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4",      "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "compact-mode",      "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv4",      "disable-snat-multiple-gws": "snatGW",   "second-bridge": "1br", "ic": "ic-disabled"}
-          - {"target": "multi-homing",      "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones"}
+          - {"target": "multi-homing",      "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "network-segmentation": "enable-network-segmentation"}
           - {"target": "multi-node-zones",  "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv4",      "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-multi-node-zones", "num-workers": "3", "num-nodes-per-zone": "2"}
           - {"target": "external-gateway",  "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4",      "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "2br", "ic": "ic-single-node-zones"}
           - {"target": "external-gateway",  "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv4",      "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
@@ -513,7 +513,7 @@ jobs:
       TRAFFIC_FLOW_TESTS: "${{ matrix.traffic-flow-tests }}"
       ENABLE_ROUTE_ADVERTISEMENTS: "${{ matrix.routeadvertisements != '' }}"
       ADVERTISE_DEFAULT_NETWORK:  "${{ matrix.routeadvertisements == 'advertise-default' }}"
-      ENABLE_PRE_CONF_UDN_ADDR: "${{ ( matrix.target == 'network-segmentation' || matrix.target == 'kv-live-migration' ) && matrix.ic == 'ic-single-node-zones' }}"
+      ENABLE_PRE_CONF_UDN_ADDR: "${{ ( ( matrix.target == 'multi-homing' && matrix.network-segmentation == 'enable-network-segmentation' ) || matrix.target == 'kv-live-migration' ) && matrix.ic == 'ic-single-node-zones' }}"
     steps:
 
     - name: Install VRF kernel module


contrib/kind-common
```diff
@@ -434,6 +434,16 @@ install_multus() {
...
+
+    run_kubectl apply -f - <<EOF
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  name: default
+  namespace: ovn-kubernetes
+spec:
+  config: '{"cniVersion": "0.4.0", "name": "ovn-kubernetes", "type": "ovn-k8s-cni-overlay"}'
+EOF
```
maiqueb: maybe you could rebase this on top of #5320 and avoid having to use this ?

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1049,6 +1077,7 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
 			)
 
 			if err != nil {
+				g.Expect(test.expectedError).ToNot(gomega.BeNil(), "unexpected error: %v", err)
```
maiqueb: I am puzzled here ... so if there's an error, we expect the error to not be nil ?... 

that's strange. Is this just for debug or something ?

go-controller/pkg/util/multi_network.go
```diff
@@ -1344,18 +1344,31 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
 		Name:      activeNetworkNADKey[1],
 	}
+	defaultNSE, err := GetK8sPodDefaultNetworkSelection(pod)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed getting default-network annotation for pod %q: %w", pod.Namespace+"/"+pod.Name, err)
+	}
+
+	// If there are static IPs and MACs at the default NSE, override the active NSE with them
+	// TODO: Discover default network "name" and "namespace"
+	if defaultNSE != nil && defaultNSE.Namespace == "ovn-kubernetes" && defaultNSE.Name == "default" {
```
maiqueb: I think you can hardcode "default" NAD metadata name, and can rely on `config.Kubernetes.OVNConfigNamespace` to get the namespace.

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1011,6 +1011,34 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
...
+			},
+			inputPrimaryUDNConfig: &ovncnitypes.NetConf{
+				NetConf:  cnitypes.NetConf{Name: networkName},
+				Topology: ovntypes.Layer2Topology,
+				NADName:  GetNADName(namespaceName, attachmentName),
+				Role:     ovntypes.NetworkRolePrimary,
+			},
+			inputPodAnnotations: map[string]string{
+				nadv1.NetworkAttachmentAnnot: GetNADName(namespaceName, "another-network"),
+				DefNetworkAnnotation:         `[{"namespace": "ovn-kubernetes", "name": "default", "ips": ["192.168.0.3"], "mac": "aa:bb:cc:dd:ee:ff"}]`,
```
maiqueb: could we maybe turn this test into a dual-stack test ? so we test the most complex alternative out there.

contrib/kind-common
qinqon: Ahh didn't know it was there already, let's add the dep thanks!

maiqueb: Tks; this way, you help me test my stuff :stuck_out_tongue_closed_eyes:

go-controller/pkg/util/multi_network.go
qinqon: You have being a living LLM to ask for :-)

maiqueb: lol - luckily, the PR you're now based on top of made use of this ... It was super fresh in my mind.

go-controller/pkg/util/multi_network_test.go
qinqon: This means that if we have an error but the test it was not expected to it, we should fail if we don't do that
the `gomega.MatchError(test.expectedError)` panics since `test.expectedError` is nil.

contrib/kind-common
qinqon: quid pro quo

go-controller/pkg/util/multi_network_test.go
maiqueb: OK got it. 

Wouldn't it be more readable if you did the opposite ? meaning something like:
```golang
if test.expectedError != nil {
    g.Expect(err).To(gomega.MatchError(test.expectedError))
} else {
    g.Expect(err).NotTo(gomega.HaveOccurred())
}
```

This is just an opinion - I now understand what you're going for.

qinqon: Ahh that's way better, I will do it.

qinqon:
Looks like bumping net-attach-def client breaks networkmanage unit tests

qinqon:
when an existing NAD is passed to the default network then it replace pod request nad name 

```
ovnk ovn-worker2 logs ovnkube-node ovnkube-controller |grep DELETE
I0701 11:17:54.410482   61361 cniserver.go:131] DELETEME, cr.Config: %s{"cniVersion":"1.0.0","name":"default-network-annotation-7332_secondaryl2network","netAttachDefName":"default-network-annotation-7332/secondaryl2network","role":"secondary","subnets":"103.0.0.0/16,2014:100:200::0/60","topology":"layer2","type":"ovn-k8s-cni-overlay"}
```

go-controller/go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
 	github.com/bhendo/go-powershell v0.0.0-20190719160123-219e7fb4e41e
 	github.com/cenkalti/backoff/v4 v4.3.0
-	github.com/containernetworking/cni v1.1.2
+	github.com/containernetworking/cni v1.2.0-rc1
 	github.com/containernetworking/plugins v1.2.0
```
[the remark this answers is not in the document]
maiqueb: @qinqon can you act on this ? It does make sense to me.

go-controller/go.mod
```diff
@@ -146,6 +145,7 @@ require (
 
 replace (
 	github.com/Microsoft/hcsshim => github.com/Microsoft/hcsshim v0.8.20
+	github.com/containernetworking/cni => github.com/containernetworking/cni v1.2.3
```
maiqueb: yeah, as AI said, I think we should require this version directly.

go-controller/pkg/clustermanager/clustermanager_test.go
```diff
@@ -1476,7 +1476,7 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
 			ginkgo.BeforeEach(func() {
 				config.OVNKubernetesFeature.EnableMultiNetwork = true
 				config.OVNKubernetesFeature.EnableNetworkSegmentation = true
-				config.OVNKubernetesFeature.EnableCustomNetworkConfig = true
+				config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses = true
```
maiqueb: you need to rebase the latest iteration of the dependent PRs. This is accounted for in them.

go-controller/pkg/util/multi_network.go
```diff
@@ -1344,18 +1344,44 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
+			if defaultNSE != nil {
+				if defaultNSE.Namespace != config.Kubernetes.OVNConfigNamespace {
+					return false, nil, fmt.Errorf("unexpected default NSE namespace %q, expected %q", defaultNSE.Namespace, config.Kubernetes.OVNConfigNamespace)
+				}
+				if defaultNSE.Name != types.DefaultNetworkName {
+					return false, nil, fmt.Errorf("unexpected default NSE name %q, expected %q", defaultNSE.Name, types.DefaultNetworkName)
+				}
+				activeNSE.IPRequest = defaultNSE.IPRequest
+				activeNSE.MacRequest = defaultNSE.MacRequest
+			}
```
maiqueb: this does make sense to me, but could we maybe abstract it away in an helper ? 

It would make it easier to unit test, and allow the casual reader to not have to focus on the details.

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -575,6 +580,130 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+			},
+			args: args{
+				network: &nadapi.NetworkSelectionElement{
+					IPRequest: []string{"192.168.0.101/24"},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			// on secondary networks with IPAM and layer2 topology, expect error when EnablePreconfiguredUDNAddresses is enabled
```
maiqueb: can we re-phrase this comment to highlight what makes it fail ? i.e. something like:
"when EnablePreconfiguredUDNAddresses is enable, on layer2 networks with IPAM, fail for secondary networks".

Not a must. And I'm not sure this comment reads that well.

test/e2e/kubevirt.go
```diff
@@ -1982,13 +2019,31 @@ ip route add %[3]s via %[4]s
...
 				topology: udnv1.NetworkTopologyLayer2,
 				role:     udnv1.NetworkRolePrimary,
 				ingress:  "routed",
 			}),
+			Entry(nil, testData{
+				resource: virtualMachineWithUDN,
+				test:     liveMigrate,
+				topology: udnv1.NetworkTopologyLayer2,
+				role:     udnv1.NetworkRolePrimary,
+			}),
```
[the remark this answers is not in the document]
maiqueb: oh, interesting. Is this true @qinqon ?

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -330,13 +331,28 @@ func allocatePodAnnotationWithRollback(
...
+			return
+		}
+		if netInfo.TopologyType() != types.Layer2Topology {
+			// Static IP requests with IPAM are only supported on layer2 topology networks.
+			// On other topologies, we cannot distinguish between already allocated IPs and
+			// IPs excluded from allocation, making it impossible to safely honor static IP
+			// requests when IPAM is enabled.
+			err = fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s: layer2 topology is required, but network has topology %q", podDesc, netInfo.TopologyType())
+			return
+		}
```
maiqueb: nit: maybe we should encapsulate this in an helper that receives the config and the `NetInfo` and makes a decision ? 

It would IMHO lead to simpler code.

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -785,33 +915,41 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+					tt.netInfo, err = util.NewNetInfo(&ovncnitypes.NetConf{
+						Topology: types.Layer2Topology,
+						NetConf: cnitypes.NetConf{
+							Name: network.Name,
+						},
+						NADName:            tt.nadName,
+						Subnets:            subnets,
+						AllowPersistentIPs: tt.persistentIPAllocation,
+						Role:               tt.role,
+					})
```
maiqueb, line 932: maybe if we encapsulate this into some helpers this code becomes more readable ? 

Like ... IMHO these unit tests have gotten quite complex over the years .

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -871,6 +1009,15 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 				tt.role,
 			)
 
+			if tt.wantErr {
+				// check the expected error after we have checked above that the
+				// rollback has behaved as expected
```
maiqueb: wait a minute ... at the very least this comment needs to be updated - you're no longer asserting rollbacks are performed on errors. 

... and that's a bit concerning ... Why are you moving this here ?... Don't we still have errors that we expect but IPs that need to be returned to the pool ? 

Could it be we need to split the errors into two types (errors that produce rollbacks vs errors that don't ?)

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
```diff
@@ -4,13 +4,13 @@ import (
...
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
 	nadtypes "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	nadclientset "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned"
 	nadlisters "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
maiqueb: yeah, I think this is addressed on the latest iteration of my PR that provisions the default network attachment NAD whenever the feature is enabled.

test/e2e/kubevirt.go
```diff
@@ -1811,6 +1845,12 @@ ip route add %[3]s via %[4]s
 			step = by(vmi.Name, "Wait for addresses at the virtual machine")
 			expectedNumberOfAddresses := len(cidrs)
 			expectedAddreses := virtualMachineAddressesFromStatus(vmi, expectedNumberOfAddresses)
+			if _, ok := vmi.Annotations[kubevirtAddressesAnnotation]; ok {
```
maiqueb: nit: please change from OK to something with semantic meaning, like `hasIPRequests` - or similar.

go-controller/pkg/allocator/pod/pod_annotation_test.go
qinqon: Moving it back where it was.

go-controller/pkg/allocator/pod/pod_annotation_test.go
qinqon: It's quite declarative, adding indirections to tests complicate following them.

go-controller/pkg/clustermanager/defaultnetworkcustomization/ensurenad.go
qinqon: will check

test/e2e/kubevirt.go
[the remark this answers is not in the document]
qinqon: @maiqueb yep and fixed it

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: I think the current state is already quite complex and hard to follow. 

Adding these 3 nested levels of if conditions will hardly make this easier to follow. 

Still, only a recommendation ... I'm not a maintainer.

qinqon:
[push](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5321):
- Support the default-network NSE ipamclaimref

kyrtapz:
@qinqon you are yet another victim of the recent E2Es refactoring :D 

```
Error: ./kubevirt.go:1636:15: undefined: selectCIDRs
Error: ./kubevirt.go:1855:43: undefined: selectCIDRs
Error: ./kubevirt.go:2232:4: declared and not used: cidr
Error: ./kubevirt.go:2232:27: undefined: selectCIDRs
Error: ./kubevirt.go:2271:138: undefined: cidrs
Error: ./network_segmentation_default_network_annotation.go:57:15: undefined: generateL2Subnets
Error: ./network_segmentation_default_network_annotation.go:102:9: undefined: selectCIDRs
```

go-controller/pkg/util/multi_network.go
```diff
@@ -1349,18 +1364,39 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
-	networkSelections[activeNetworkNADs[0]] = &nettypes.NetworkSelectionElement{
+
+	activeNSE := &nettypes.NetworkSelectionElement{
 		Namespace: activeNetworkNADKey[0],
 		Name:      activeNetworkNADKey[1],
 	}
 
-	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) {
+	// Feature gate integration: EnablePreconfiguredUDNAddresses controls default network IP/MAC transfer to active network
+	if config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses {
```
kyrtapz: Lets use `IsPreconfiguredUDNAddressesEnabled` instead.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -330,13 +353,11 @@ func allocatePodAnnotationWithRollback(
...
 	}
+
 	if hasIPAM && hasStaticIPRequest {
-		// for now we can't tell apart already allocated IPs from IPs excluded
-		// from allocation so we can't really honor static IP requests when
-		// there is IPAM as we don't really know if the requested IP should not
-		// be allocated or was already allocated by the same pod
-		err = fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s", podDesc)
-		return
+		if err = validateStaticIPRequest(netInfo, podDesc); err != nil {
```
kyrtapz: Just to make sure we are all on the same page, this is all we need to ensure that the requested IP ends up on the IPAMClaim too right? 
Is the scenario when someone requests a different IP than the one that's set already on the existing IPAMClaim covered as well?

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -207,6 +208,28 @@ func allocatePodAnnotationWithTunnelID(
 	return pod, podAnnotation, nil
 }
 
+// validateStaticIPRequest checks if a static IP request can be honored when IPAM is enabled for the given network.
+func validateStaticIPRequest(netInfo util.NetInfo, podDesc string) error {
+	// Allow static IPs with IPAM only for primary networks with layer2 topology when EnablePreconfiguredUDNAddresses is enabled
+	// Feature gate integration: EnablePreconfiguredUDNAddresses controls static IP allocation with IPAM
+	if !config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses {
```
kyrtapz: Lets use `IsPreconfiguredUDNAddressesEnabled` instead.

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: > Just to make sure we are all on the same page, this is all we need to ensure that the requested IP ends up on the IPAMClaim too right?

Yes, with this IPAMClaims status store the IP from IPRequest

> Is the scenario when someone requests a different IP than the one that's set already on the existing IPAMClaim covered as well?

No, that's not covered, since we don't support ipamclaims for pods and new VMs cannot reference a pre existing IPAMClaim I think we don't need to cover it.

Still do you think we need to add it ? I can added if you see is needed

maiqueb: This is an interesting train of thought. Tks @kyrtapz for making me think this through.

Looking at the code below, I see the following code:
```golang
	     if hasIPRequest {
			tentative.IPs, err = util.ParseIPNets(network.IPRequest)
			if err != nil {
				klog.Warningf("Failed parsing IPRequest %+v for pod %s: %v", network.IPRequest, podDesc, err)
				return
			}
		} else if hasIPAMClaim {
			tentative.IPs, err = util.ParseIPNets(ipamClaim.Status.IPs)
			if err != nil {
				return
			}
		}
```

Which means - IIUC - that:
- the IP request will always take precedence to the IPAMClaim IP. 
- the IP requests cannot change on an existing pod - there's (or there will be) the Kubernetes VAP to prevent it
- thus, the only way we can find ourselves in the situation described - IPRequests != IPAMClaim.Status.IPs - is if the IPAMClaim IPs change. 

How can that happen ? 
I'm trusting a mix of RBAC - the component that *owns* the status is OVN-Kubernetes - and proper test coverage to ensure that doesn't happen. 

Maybe we could add some CEL validation (please bear with me, maybe this suggestion is not possible) to the IPAMClaim status to prevent the status IPs from changing ? It would need to accept the *first* status IPs definition, but from then, keep in immutable. How could we express that ?

qinqon: > Maybe we could add some CEL validation (please bear with me, maybe this suggestion is not possible) to the IPAMClaim status to prevent the status IPs from changing ? It would need to accept the _first_ status IPs definition, but from then, keep in immutable. How could we express that ?

@ormergi is the CEL expert, do you know if we can enforce a value being changed from empty just once ?

qinqon: An idea from LLMs:

```golang
// If oldSelf.myAttribute was empty/unset, allow setting to any value (including empty initially)
// OR if oldSelf.myAttribute was already set, then self.myAttribute must be immutable (equal to oldSelf.myAttribute)
((!has(oldSelf.myAttribute) || oldSelf.myAttribute == '') && (has(self.myAttribute))) || (has(oldSelf.myAttribute) && oldSelf.myAttribute != '' && self.myAttribute == oldSelf.myAttribute)
```

ormergi: The following rule should  enforce imutability: 
```
	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="Status is immutable"
```

ormergi: Making the status immutable sounds odd and some what out of the convention where it should refect the current state, semantically if its immutable it will never reflect a current state :sweat_smile: 

I need to go over this PR to suggest something meaningful

EDIT: from top of my head, in a scenario the annotation ip-request mismatch the IP in the referenced IPAMClaim, I would call that undefined state where something went wrong in the process by one of the components that interacts with ovn-k, thus I think it should return with an error.

maiqueb: will this allow the controller to set the state initially ? i.e. define the IPs once ? 

Keep in mind we first (ipam-ext) creates the IPAMClaim, then a separate component - OVN-K - mutates this ipam claim, filling out its status attribute.

ormergi: It should prevent status for changing once the CR is created, so I dont think this will go..

maiqueb: Right, that's my point.

1. we don't want the status to immutable - we need to update conditions on the status
2. we want to allow the controller to define the value of the status.IPs attribute *after* the IPAMClaim has been created - *exactly* one time.

those are my reqs ... 

I'd say this is still "just" hardening ( ensuring OVN-K is resilient).

ormergi: I think the following rule should work as long the IPAMClaim CR status.IPs is empty once created and remain empty until OVN-K updates it:
```
// +kubebuilder:validation:XValidation:rule="self == '''' ||  self == oldSelf", message="status.IPs cannot be changed once set"
```

Hardening the status up to it can be updated one time only feel wrong to me, by the convention status should reflect the current state, having it immutable (eventually) it cannot reflect a current state - semantically.

The way I see it, when IP-request mismatch the referenced IPAMClaim spec, its an undefined state caused by some component that did wrong. OVN-K should blow up in that case and return an error. 
Do you think this can work?

ormergi:
@qinqon #5332 is merged could you please rebase and drop the first commit?

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1011,10 +1012,127 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
 				},
 			},
 		},
+		{
```
ormergi, line 1015: Please consider adding test for when default NSE is absent, and if possible when fetching the default NSE fails.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -207,6 +207,28 @@ func allocatePodAnnotationWithTunnelID(
...
+		// Static IP requests with IPAM are only supported on primary networks
+		return fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s: only supported on primary networks", podDesc)
+	}
+	if netInfo.TopologyType() != types.Layer2Topology {
+		// Static IP requests with IPAM are only supported on layer2 topology networks.
+		// On other topologies, we cannot distinguish between already allocated IPs and
+		// IPs excluded from allocation, making it impossible to safely honor static IP
+		// requests when IPAM is enabled.
+		return fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s: layer2 topology is required, but network has topology %q", podDesc, netInfo.TopologyType())
+	}
```
ormergi, line 228: nit: please consider combining the condition to one with an error saying only primary L2 is supported.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -207,6 +207,28 @@ func allocatePodAnnotationWithTunnelID(
 	return pod, podAnnotation, nil
 }
 
+// validateStaticIPRequest checks if a static IP request can be honored when IPAM is enabled for the given network.
+func validateStaticIPRequest(netInfo util.NetInfo, podDesc string) error {
```
ormergi: Please consider validating the default NSE IP-request match the referenced IPAMClaim CR IP request.

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -575,6 +580,130 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+				Role: types.NetworkRolePrimary,
+			},
+			wantReleasedIPsOnRollback: ovntest.MustParseIPNets("192.168.0.101/24"),
+		},
+		{
+			// on networks with IPAM and layer2 topology, expect error when EnablePreconfiguredUDNAddresses is false
+			name: "expect error, static IP with IPAM on layer2 when EnablePreconfiguredUDNAddresses is false",
+			ipam: true,
+			// enablePreconfiguredUDNAddresses defaults to false
+			netConf: &ovncnitypes.NetConf{
```
ormergi: Shouldn't the netConf have role primary?

test/e2e/kubevirt.go
```diff
@@ -1619,6 +1625,32 @@ write_files:
 				},
 			}
 
+			virtualMachineWithUDNAndStaticIPsAndMAC = resourceCommand{
+				description: "VirtualMachine with interface binding for UDN and statics IPs and MAC",
+				cmd: func() string {
+					GinkgoHelper()
+					if !isPreConfiguredUdnAddressesEnabled() {
+						Skip("ENABLE_PRE_CONF_UDN_ADDR not configured")
```
ormergi, line 1630: Please consider replacing the skip with actual assertion, let the tests runner skip the tests according to env.

ormergi (commented):
Very nice!
This is first round, I need to go over e2e again, please see my comments

go-controller/pkg/util/multi_network.go
```diff
@@ -1321,6 +1321,21 @@ func GetPodNADToNetworkMapping(pod *corev1.Pod, nInfo NetInfo) (bool, map[string
 	return true, networkSelections, nil
 }
 
+// overrideActiveNSEWithDefaultNSE overrides the provided active NetworkSelectionElement with the IP and MAC requests from
+// the default NetworkSelectionElement after validating its namespace and name.
+func overrideActiveNSEWithDefaultNSE(defaultNSE, activeNSE *nettypes.NetworkSelectionElement) error {
```
maiqueb, line 1314: nit: wouldn't it be preferable to make this function return an NSE over mutating the contents of the active NSE ? 

just an opinion - you can ignore it.

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1011,10 +1012,127 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
...
 			g := gomega.NewWithT(t)
+
+			originalEnablePreconfiguredUDNAddresses := config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses
+			originalEnableMultiNetwork := config.OVNKubernetesFeature.EnableMultiNetwork
+			originalEnableNetworkSegmentation := config.OVNKubernetesFeature.EnableNetworkSegmentation
+			t.Cleanup(func() {
+				config.OVNKubernetesFeature.EnablePreconfiguredUDNAddresses = originalEnablePreconfiguredUDNAddresses
+				config.OVNKubernetesFeature.EnableMultiNetwork = originalEnableMultiNetwork
+				config.OVNKubernetesFeature.EnableNetworkSegmentation = originalEnableNetworkSegmentation
+			})
```
maiqueb: I think you can just choose to run the function `config.PrepareTestConfig()` : https://github.com/ovn-kubernetes/ovn-kubernetes/blob/f09eca1cfaa23a691141b626fea8f1df498769a6/go-controller/pkg/config/config.go#L652

At least it's what I see in the other tests. 

... yes, if one test doesn't call it, it risks failing ... 

just an opinion.

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1048,11 +1166,14 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
 				primaryUDNNetInfo,
 			)
 
-			if err != nil {
+			if test.expectedError != nil {
+				g.Expect(err).To(gomega.HaveOccurred(), "unexpected success operation, epecting error")
 				g.Expect(err).To(gomega.MatchError(test.expectedError))
```
maiqueb, line 1187: aren't these redundant ? i.e. I'd just leave the expectation with the `MatchError` matcher.

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1048,11 +1166,14 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
...
+			} else {
+				g.Expect(err).ToNot(gomega.HaveOccurred())
+				g.Expect(isAttachmentRequested).To(gomega.Equal(test.expectedIsAttachmentRequested))
+				g.Expect(networkSelectionElements).To(gomega.Equal(test.expectedNetworkSelectionElements))
```
maiqueb, line 1191: quick question: the order here is kept since the NSEs are a slice, right ? 

initially I was wondering if using a `ConsistOf` matcher wouldn't be safer, but I think it doesn't matter.

go-controller/pkg/allocator/pod/pod_annotation.go
maiqueb: > convention status should reflect the current state, having it immutable (eventually) it cannot reflect a current state - semantically.

Point being is we do not reconfigure the attachment under any circumstance, and we do not allow the cluster default attachment customization to change over the life time of a pod. 

> The way I see it, when IP-request mismatch the referenced IPAMClaim spec, its an undefined state caused by some component that did wrong. OVN-K should blow up in that case and return an error.

How would we recover from that error ? 

**EDIT:** disclaimer - I am in no way saying this is required. I just thought it would make it more resilient.
For sure this should be carefully thought over.

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: good point.

go-controller/pkg/allocator/pod/pod_annotation.go
maiqueb: @ormergi you lost me at "IPAMClaim CR IP request". AFAIK there isn't an IP request attribute in IPAMClaims.

Do you mean in the status ? If so, it would OK - we'd just have to *also* allow it if `status.ips` are empty. Right ?

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -575,6 +580,130 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+			enablePreconfiguredUDNAddresses: true,
+			role:                            types.NetworkRolePrimary, // has to be primary network for default routes to be set
+			netConf: &ovncnitypes.NetConf{
+				Topology: types.Layer2Topology,
+				NetConf: cnitypes.NetConf{
+					Name: "network",
+				},
+				NADName: util.GetNADName("namespace", "network"),
+				Subnets: "192.168.0.0/24,2001:db8::/64",
+				Role:    types.NetworkRolePrimary,
```
maiqueb: this is fishy. We're passing the role attribute directly, but that is part of the `NetConf` struct. 

At the very least - IMHO - we can only pass this attribute via one way.

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -575,6 +580,130 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
+			args: args{
+				network: &nadapi.NetworkSelectionElement{
+					IPRequest: []string{"192.168.0.101/24"},
+				},
+			},
+			role:    types.NetworkRoleSecondary,
+			wantErr: true,
+		},
+		{
+			// on primary networks with IPAM and layer2 topology, expect success when EnablePreconfiguredUDNAddresses is enabled
```
maiqueb, line 614: I am sooooo confused. This has too many parameters, making it hard to follow. Or I'm daft. 

Isn't this the same test you have added on line 585 ?

test/e2e/util.go
```diff
@@ -1169,6 +1169,11 @@ func isLocalGWModeEnabled() bool {
 	return present && val == "local"
 }
 
+func isPreConfiguredUdnAddressesEnabled() bool {
+	val, present := os.LookupEnv("ENABLE_PRE_CONF_UDN_ADDR")
+	return present && val == "true"
```
maiqueb: shouldn't we actually introspect the cluster to figure this out ? 

It would prepare us for the future, but, I'd say it is not a must.

test/e2e/kubevirt.go
```diff
@@ -1619,6 +1625,32 @@ write_files:
...
+					}
+					staticIPs, err := json.Marshal(map[string][]string{
+						"net1": filterCIDRs(fr.ClientSet, staticIPv4, staticIPv6),
+					})
+					Expect(err).ToNot(HaveOccurred())
+					annotations := map[string]string{
+						kubevirtAddressesAnnotation: string(staticIPs),
+					}
```
maiqueb: maybe turn this into an helper func with a proper name ?

test/e2e/kubevirt.go
```diff
@@ -1619,6 +1625,32 @@ write_files:
...
+					vm = fedoraWithTestToolingVM(nil /*labels*/, annotations, nil, /*nodeSelector*/
+						kubevirtv1.NetworkSource{
+							Pod: &kubevirtv1.PodNetwork{},
+						}, userDataWithIperfServer, networkDataDualStack)
+					vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].Bridge = nil
+					vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].Binding = &kubevirtv1.PluginBinding{Name: "l2bridge"}
+					vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress = staticMAC
```
maiqueb, line 1642: nit: maybe just setting this once using an `Interface` struct ? you'd then have a well named helper return that `Interface`, something like: `vm.Spec.Template.Spec.Domain.Devices.Interfaces[0] = primaryUDNIface(withMACAddr(staticMAC))`

you can ignore this comment.

test/e2e/network_segmentation_default_network_annotation.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		Expect(netStatus[0].IPs).To(ConsistOf(exposedIPs), "Should have the IPs specified in the default network annotation")
+		Expect(strings.ToLower(netStatus[0].Mac)).To(Equal(strings.ToLower(tc.mac)), "Should have the MAC specified in the default network annotation")
+
+	},
+
+		Entry("should create the pod with the specified static IP and MAC address", testCase{
+			ips: []string{"103.0.0.3/16", "2014:100:200::3/60"},
+			mac: "02:A1:B2:C3:D4:E5",
+		}),
```
maiqueb, line 104: I mean ... if we're only adding a single test ... the table is the definition of overkill isn't it ? 

IMHO it reads a lot worse. 

Are there expected entries to be added in the future ? If so, then OK - everything's fine.

qinqon:
> @qinqon #5332 is merged could you please rebase and drop the first commit?

@ormergi the first commit is a different dependency(https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5320), I rebased long time ago with your changes.

go-controller/pkg/util/multi_network_test.go
qinqon: > Please consider adding test for when default NSE is absent

Expect the new test the other ones are already testing without default NSE

> and if possible when fetching the default NSE fails

You mean for example testing with a wrong json at the default-network ? I will
do that

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: I will add a validation to check that IPRequest is the same as IPAMClaim.Status.IPs ]

ormergi: Sorry for confusing, yes I meant the status.IPs

test/e2e/kubevirt.go
qinqon: > Please consider replacing the skip with actual assertion, let the tests runner skip the tests according to env.

@ormergi what do you mean ?

go-controller/pkg/util/multi_network_test.go
qinqon: Ginkgo segfaults if err is nil

go-controller/pkg/util/multi_network_test.go
qinqon: This is comparing maps not slices

maiqueb: I thought `Equal` wouldn't work on maps, but I see it in gomega's unit tests so ... It must. 
My bad.

go-controller/pkg/allocator/pod/pod_annotation_test.go
qinqon: @maiqueb I am going to remove the "netConf" option totally from the tests, I was just missing the allowPersistentIPs flag on the test
to make it work

go-controller/pkg/allocator/pod/pod_annotation_test.go
qinqon: One test check ipRequest + mac the other just IPRequest

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: that works for me - we either pass netconf w/ everything in it, *or* we pass the individual knobs and compose a netconf in the test. 

Doing both makes this even harder to follow.

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: could we kind of tinker with the test names or comments to reflect that ? 

honestly, I get headaches from looking at these test entries.

**EDIT:** I think moving the tests together would be helpful, and either aligning the comment (or removing it altogether) would make it easier to understand.

test/e2e/kubevirt.go
qinqon: I am not going to change that at this PR

test/e2e/network_segmentation_default_network_annotation.go
qinqon: The idea is to add the negative test at the same table in the future, so they
have a placeholder kind of.

go-controller/pkg/util/multi_network.go
qinqon: I don't thing it's worth it, it means cloning active and returning it.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -1016,7 +1017,7 @@ func (c *Controller) getSelectedNADs(networkSelectors apitypes.NetworkSelectors)
 		case apitypes.DefaultNetwork:
 			// if we are selecting the default networkdefault network label,
 			// make sure a NAD exists for it
-			nad, err := c.getOrCreateDefaultNetworkNAD()
+			nad, err := defaultnetworkcustomization.EnsureDefaultNetworkNAD(c.nadLister, c.nadClient)
```
RamLavi: `getOrCreateDefaultNetworkNAD()` is also used on pkg/clustermanager/routeadvertisements/controller_test.go:1054.

qinqon: This is not from my PR.

maiqueb: @RamLavi please check the file you've linked. It's changed there. 

Are you sure something's wrong ?

kyrtapz (approved):
LGTM

maiqueb (dismissed):
tks.

go-controller/go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/asaskevich/govalidator v0.0.0-20210307081110-f21760c49a8d
 	github.com/bhendo/go-powershell v0.0.0-20190719160123-219e7fb4e41e
 	github.com/cenkalti/backoff/v4 v4.3.0
-	github.com/containernetworking/cni v1.1.2
+	github.com/containernetworking/cni v1.2.0-rc1
```
tssurya: hmm is this an unreleased versionof containernetworking/cni? :o can we bump it to a released version instead? and is this dependency caused by bumping network-attachment-definition-client to 1.7.7?

go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
tssurya, line 1: self-note: OK so this bump is to get https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/commit/372cca29d6a433d58b85e8ffcff3dccb879f9603

go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
```diff
@@ -20,10 +20,12 @@ func main() {
 
 	p := cni.NewCNIPlugin("")
 	c.Action = func(_ *cli.Context) error {
-		skel.PluginMain(
-			p.CmdAdd,
-			p.CmdCheck,
-			p.CmdDel,
+		skel.PluginMainFuncs(
```
tssurya, line 23: this is because we also bumped CNI version looks like?

go-controller/go.mod
```diff
@@ -73,6 +73,7 @@ require (
 )
 
 require (
+	github.com/Masterminds/semver/v3 v3.2.1 // indirect
```
tssurya: why is semver dependency coming is it because of cni version or network-attachement-definition bump?

go-controller/go.mod
tssurya: ah commit2 is bumping it to 1.2.3

go-controller/go.mod
tssurya: ah commit2 is removing the semver dependency altogether

go-controller/go.mod
tssurya, line 1: on the commit message you mention:
```
breaks ovn-k nad
parssing, looks like there is a new MarshalJSON fuction at CNI that make
it ignore the kustom ovn-k fields at nad.spec.config
```
can you point to this function in CNI repo? is this a problem only on the rc-1 release and not on the 2.3 release right? its really weird go mod tidy picked a rc1 release w/o picking the proper one

go-controller/pkg/util/multi_network.go
```diff
@@ -1349,18 +1364,39 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
 		Namespace: activeNetworkNADKey[0],
 		Name:      activeNetworkNADKey[1],
 	}
 
-	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) {
+	// Feature gate integration: EnablePreconfiguredUDNAddresses controls default network IP/MAC transfer to active network
+	if IsPreconfiguredUDNAddressesEnabled() {
+		// Limit the static ip and mac requests to the layer2 primary UDN when EnablePreconfiguredUDNAddresses is enabled, we
+		// don't need to explicitly check this is primary UDN since
+		// the "active network" concept is exactly that.
```
tssurya, line 1365: ack so we don't only limit this to CUDN, we are doing it both for L2 P-UDN and P-CUDN? hope enhancement agreed on the same thing? (IIRC only the BGP bits were scoped to CUDNs rest are for both UDN and CUDN @kyrtapz ?)

go-controller/pkg/util/multi_network.go
```diff
@@ -1349,18 +1364,39 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
+
+	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) && activeNSE.IPAMClaimReference == "" {
 		ipamClaimName, wasPersistentIPRequested := pod.Annotations[OvnUDNIPAMClaimName]
 		if wasPersistentIPRequested {
-			networkSelections[activeNetworkNADs[0]].IPAMClaimReference = ipamClaimName
+			activeNSE.IPAMClaimReference = ipamClaimName
 		}
 	}
 
+	networkSelections[activeNetworkNADs[0]] = activeNSE
```
tssurya, line 1387: so for secondary network Pods I see behaviour stays the same, we seem good.

go-controller/pkg/util/multi_network.go
tssurya, line 1365: when we say we don't need to explicitly check if its primary UDN since active network concept does that,, we mean:
`activeNetwork.TopologyType() == types.Layer2Topology` -> its not like default network is Layer2 and secondary networks active network will be default network or the primary one anyways ?

go-controller/pkg/util/multi_network.go
```diff
@@ -1349,18 +1364,39 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
+	if IsPreconfiguredUDNAddressesEnabled() {
+		// Limit the static ip and mac requests to the layer2 primary UDN when EnablePreconfiguredUDNAddresses is enabled, we
+		// don't need to explicitly check this is primary UDN since
+		// the "active network" concept is exactly that.
+		if activeNetwork.TopologyType() == types.Layer2Topology {
+			defaultNSE, err := GetK8sPodDefaultNetworkSelection(pod)
+			if err != nil {
+				return false, nil, fmt.Errorf("failed getting default-network annotation for pod %q: %w", pod.Namespace+"/"+pod.Name, err)
+			}
+			// If there are static IPs and MACs at the default NSE, override the active NSE with them
```
tssurya, line 1371: today we don't support active NSE even having a way to specific IPrequests or MAC etc right? meaning we never added support for NSE on primary UDN pods unlike for secondary UDN pods.. so here the goal is that usually at this stage these fields in the activeNSE must be empty correct?

activeNSE := &nettypes.NetworkSelectionElement{
		Namespace: activeNetworkNADKey[0],
	}
so usually we only set these two values here I guess ^

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1011,10 +1012,143 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
...
+				Topology: ovntypes.Layer2Topology,
+				NADName:  GetNADName(namespaceName, attachmentName),
+				Role:     ovntypes.NetworkRolePrimary,
+			},
+			inputPodAnnotations: map[string]string{
+				nadv1.NetworkAttachmentAnnot: GetNADName(namespaceName, "another-network"),
+				DefNetworkAnnotation:         `[{"foo}`,
+			},
+			enablePreconfiguredUDNAddresses: true,
+			expectedError:                   fmt.Errorf(`failed getting default-network annotation for pod "/test-pod": %w`, fmt.Errorf(`GetK8sPodDefaultNetwork: failed to parse CRD object: parsePodNetworkAnnotation: failed to parse pod Network Attachment Selection Annotation JSON format: unexpected end of JSON input`)),
```
tssurya, line 1134: do we need a false negative test if the network role is not primary?

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -207,6 +207,28 @@ func allocatePodAnnotationWithTunnelID(
...
+	// Allow static IPs with IPAM only for primary networks with layer2 topology when EnablePreconfiguredUDNAddresses is enabled
+	// Feature gate integration: EnablePreconfiguredUDNAddresses controls static IP allocation with IPAM
+	if !util.IsPreconfiguredUDNAddressesEnabled() {
+		// Feature is disabled, reject static IPs with IPAM
+		return fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s (custom network configuration disabled)", podDesc)
```
tssurya, line 216: nit: in using the exact flag name in the message to people know what to do to get this
but otherwise looks good to me in terms of backwards compatibility

go-controller/pkg/allocator/pod/pod_annotation.go
tssurya, line 1:
```
```
at the top of this file I see this annotation present ^ should we update its description to say active networks for the pod not just the CDN?

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -330,13 +352,11 @@ func allocatePodAnnotationWithRollback(
 		}
 		hasIPAMClaim = ipamClaim != nil && len(ipamClaim.Status.IPs) > 0
 	}
+
 	if hasIPAM && hasStaticIPRequest {
```
tssurya, line 363: wow so we used to support `IPRequest` via the multus NSE - but I assume this code exists and was only used for secondary network pods? at this spot?
so up until this PR `hasStaticIPRequest` was only applicable to secondary network pod rights? and that too only when `hasIPAM`  is false meaning ipamlessMode for l2/localnet with 0 subnets I guess

tssurya (commented):
@qinqon great job! I don't have any major concerns, just some questions for my own understanding.

Also did you add validation for this: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5321/commits/f42a91e9c5b2f44c86c55be7556c0b0f9da0b4dd#r2224681260 if so where?

OK this PR was simple enough, it does the 2 main things:

1. allows IP Request and IPAMClaim to co-exist for L2 Primary UDNs
2. allows the re-use of `v1.multus-cni.io/default-network` as the active NSE for primary UDN L2 pods - if we were to add the annotation on the primary network like we do for secondary - its going to delay pod setup which is also why for P-UDNs we decided to plumb in 2 interfaces together at the same time with 1 OVNK IPAM annotation - so we stick with the same thing but re-use the default network annotation+default NAD instead of re-doing `k8s.v1.cni.cncf.io/networks` annotation - as agreed on in OKEP design for this feature. 

Missing bits:
3. Doesn't do any validation on conflicts with the IPRequest field - blindly uses it @RamLavi 's PR will bring the validation in
4. Doesn't do any validation on conflicts with the MAC field - blindly uses it - needs a FUP.
5. E2E's which @asood-rh should add in FUP

go-controller/pkg/allocator/pod/pod_annotation.go
tssurya: I don't see this fixup.. lmk if I missed it somewhere as I had the same question

tssurya:
@qinqon if you get some time - ensuring when the feature is enabled and the pod network is a secondary network nothing changes for secondary pods would be good to confirm - if there are tests already for this then good.

go-controller/go.mod
qinqon: > ah commit2 is bumping it to 1.2.3

Yes, will close resolve this thread

go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
qinqon: Right, I will resolve this

go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
qinqon: Yep, is this ok ? or you see some issue with it ?

go-controller/go.mod
qinqon: Yep, resolving this

go-controller/go.mod
qinqon: If I remember correctly using cni 1.3.0 (latest stable) was breaking ovn-k compilation, but I have try it again and it's compiling allright
maybe something changed at ovn-kubernetes and not it's fine ? 

Long Story Short I am not able to reproduce the compilation error, I will pin "cni" to the latest stable wich is "1.3.0"

go-controller/pkg/util/multi_network.go
qinqon: > ack so we don't only limit this to CUDN, we are doing it both for L2 P-UDN and P-CUDN? hope enhancement agreed on the same thing? (IIRC only the BGP bits were scoped to CUDNs rest are for both UDN and CUDN @kyrtapz ?)

Yes, I think is better to limit the feature surface, if needed we can open up a follow up PR with proper testing.

qinqon: > when we say we don't need to explicitly check if its primary UDN since active network concept does that,, we mean:
`activeNetwork.TopologyType() == types.Layer2Topology` -> its not like default network is Layer2 and secondary networks active network will be default network or the primary one anyways ?

The activeNetwork is the primary network is not related to secondary networks at all, like a secondary network's primary network can be whatever users wants, for example a pod on layer3 
primary network that happends to also have a secondary layer2 network.

go-controller/pkg/util/multi_network.go
qinqon: Right, what we did here is to extend the logic a little and pass NSE to active network so is no longer always empty.

go-controller/pkg/util/multi_network_test.go
qinqon: Do you mean adding default-network annotation pointing to a secondary network ?

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: What people should do depends on lot on the product for example, openshift action is different from upstream
kind action, I think keeping it generic is good enough.

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: Make sense updating...

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: That's about right, resolving the thread

qinqon:
> @qinqon if you get some time - ensuring when the feature is enabled and the pod network is a secondary network nothing changes for secondary pods would be good to confirm - if there are tests already for this then good.

We are activating the feature at virt lanes and we are  testing VMs with just secondary network too everything looks good. 
@tssurya is it worth it to activate the feature at the multihoming tests too ? to check we don't break them ?

qinqon:
Addresses review from @tssurya except possible new unit test for false negative
[force-push](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/e1f2300f0a512934e7946479491f0ae57da570c6..d8f25425462ba6adeb5acb516fb0d56c67831136)

qinqon:
[rebased](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/d8f25425462ba6adeb5acb516fb0d56c67831136..8e7acc31990d3707a8c3fb60ad84e8391e5dfdca)

go-controller/pkg/util/multi_network_test.go
tssurya: yes! - i kinda meant a secondary network pod having a NSE on its secondary network and then also having the defaultnetwork on its primary - like both annotations together

go-controller/pkg/util/multi_network.go
tssurya: makes sense! thanks

go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
tssurya: no all good, resolving it

go-controller/go.mod
tssurya: ack that makes sense ! thanks

tssurya:
> > @qinqon if you get some time - ensuring when the feature is enabled and the pod network is a secondary network nothing changes for secondary pods would be good to confirm - if there are tests already for this then good.
> 
> We are activating the feature at virt lanes and we are testing VMs with just secondary network too everything looks good. @tssurya is it worth it to activate the feature at the multihoming tests too ? to check we don't break them ?

yea I think so! at least one of the multihoming lanes can have this feature activated yea - perhaps there we don't active UDN either so might need that as well even if we don't create UDNs

go-controller/go.mod
qinqon: Well my memory is pretty bad looks like the issue get reproduced on unit test not compile time, after bumping I get 
```
goroutine 244487 [running]:
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config.parseNetConfList(0xc000ed2120)
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/config/cni.go:124 +0x37d
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config.ParseNetConf({0xc000065810, 0x50, 0x50})
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/config/cni.go:80 +0x67
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util.ParseNetConf(0xc00652d560)
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util/multi_network.go:1145 +0x6d
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util.ParseNADInfo(0xc00652d560)
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util/multi_network.go:1112 +0x3c
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager.(*nadController).syncNAD(0xc00719edd0, {0xc003c72060, 0x16}, 0xc00652d560)
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/networkmanager/nad_controller.go:275 +0xa5
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager.(*nadController).sync(0xc00719edd0, {0xc003c72060, 0x16})
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/networkmanager/nad_controller.go:260 +0x63e
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controller.(*controller[...]).processNextQueueItem(0x0)
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controller/controller.go:253 +0x1b5
github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controller.(*controller[...]).startWorkers.func1()
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controller/controller.go:163 +0xd8
created by github.com/ovn-org/ovn-kubernetes/go-controller/pkg/controller.(*controller[...]).startWorkers in goroutine 237769
	/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controller/controller.go:160 +0x2f0
```

I will go back to 1.2.3

qinqon:
[Going back to cni 1.2.3](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/8e7acc31990d3707a8c3fb60ad84e8391e5dfdca..a3de8680a5a8f599a34abd710938720993c2774a)

go-controller/pkg/util/multi_network_test.go
qinqon: The unit test before this one is already doing that I think 
```golang
{
			desc: "the network configuration for a primary layer2 UDN receive pod requesting IP, MAC and IPAMClaimRef on default network annotation for it",
			inputNetConf: &ovncnitypes.NetConf{
				NetConf:  cnitypes.NetConf{Name: networkName},
				Topology: ovntypes.Layer2Topology,
				NADName:  GetNADName(namespaceName, attachmentName),
				Role:     ovntypes.NetworkRolePrimary,
			},
			inputPrimaryUDNConfig: &ovncnitypes.NetConf{
				NetConf:  cnitypes.NetConf{Name: networkName},
				Topology: ovntypes.Layer2Topology,
				NADName:  GetNADName(namespaceName, attachmentName),
				Role:     ovntypes.NetworkRolePrimary,
			},
			inputPodAnnotations: map[string]string{
				nadv1.NetworkAttachmentAnnot: GetNADName(namespaceName, "another-network"),
				DefNetworkAnnotation:         `[{"namespace": "ovn-kubernetes", "name": "default", "ips": ["192.168.0.3/24", "fda6::3/48"], "mac": "aa:bb:cc:dd:ee:ff", "ipam-claim-reference": "my-ipam-claim"}]`,
			},
			expectedIsAttachmentRequested: true,
			expectedNetworkSelectionElements: map[string]*nadv1.NetworkSelectionElement{
				"ns1/attachment1": {
					Name:               "attachment1",
					Namespace:          "ns1",
					IPRequest:          []string{"192.168.0.3/24", "fda6::3/48"},
					MacRequest:         "aa:bb:cc:dd:ee:ff",
					IPAMClaimReference: "my-ipam-claim",
				},
			},
			enablePreconfiguredUDNAddresses: true,
		},

```

qinqon:
> > > @qinqon if you get some time - ensuring when the feature is enabled and the pod network is a secondary network nothing changes for secondary pods would be good to confirm - if there are tests already for this then good.
> > 
> > 
> > We are activating the feature at virt lanes and we are testing VMs with just secondary network too everything looks good. @tssurya is it worth it to activate the feature at the multihoming tests too ? to check we don't break them ?
> 
> yea I think so! at least one of the multihoming lanes can have this feature activated yea - perhaps there we don't active UDN either so might need that as well even if we don't create UDNs

Ok make sense will do at a new commit.

qinqon:
/retest

It's not related

```golang
  Unexpected error:
      <*errors.errorString | 0xc0007a3260>: 
      failed to run [killall iperf3] at VMI worker-fgzp2, error: killall iperf3 failed
      {
          s: "failed to run [killall iperf3] at VMI worker-fgzp2, error: killall iperf3 failed",
      }
  occurred
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:2320 @ 08/01/25 12:42:40.725
```

tssurya:
BGP CI failure looks like known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16673702405/job/47197338115?pr=5321 - https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4221

go-controller/go.mod
tssurya: ack I see thanks for the snippet.
I think this info should be in commit message so others can see why we pinned it.
from commit message it looks like even compile would fail - but perhaps we don't use unmarshall in code but only in unit test huh?

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -226,6 +226,13 @@ func validateStaticIPRequest(netInfo util.NetInfo, podDesc string) error {
 		// requests when IPAM is enabled.
 		return fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s: layer2 topology is required, but network has topology %q", podDesc, netInfo.TopologyType())
 	}
+	if ipamClaim != nil && len(ipamClaim.Status.IPs) > 0 {
+		for _, ipRequest := range network.IPRequest {
+			if !util.IsItemInSlice(ipamClaim.Status.IPs, ipRequest) {
+				return fmt.Errorf("cannot allocate a static IP request with IPAM for pod %q: the pod references an ipam claim with IPs not containing the requested IP %q", podDesc, ipRequest)
```
tssurya, line 232: it would be nice to get a pod event out of this for better user experience but that's a bit out of scope here and as discussed with @kyrtapz we can have another card to track that work.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -226,6 +226,13 @@ func validateStaticIPRequest(netInfo util.NetInfo, podDesc string) error {
 		// requests when IPAM is enabled.
 		return fmt.Errorf("cannot allocate a static IP request with IPAM for pod %s: layer2 topology is required, but network has topology %q", podDesc, netInfo.TopologyType())
 	}
+	if ipamClaim != nil && len(ipamClaim.Status.IPs) > 0 {
+		for _, ipRequest := range network.IPRequest {
```
tssurya, line 230: do we need to check network is not nil? it seems like we are passing a pointer.

.github/workflows/test.yml
```diff
@@ -513,6 +513,7 @@ jobs:
       TRAFFIC_FLOW_TESTS: "${{ matrix.traffic-flow-tests }}"
       ENABLE_ROUTE_ADVERTISEMENTS: "${{ matrix.routeadvertisements != '' }}"
       ADVERTISE_DEFAULT_NETWORK:  "${{ matrix.routeadvertisements == 'advertise-default' }}"
+      ENABLE_PRE_CONF_UDN_ADDR: "${{ ( matrix.target == 'network-segmentation' || matrix.target == 'kv-live-migration' ) && matrix.ic == 'ic-single-node-zones' }}"
```
tssurya: don't we need the multi-homing lane here?

go-controller/pkg/allocator/pod/pod_annotation.go
kyrtapz: I don't think this will ever be nil and we are not checking for it in `allocatePodAnnotationWithRollback` either.

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: Let's add a card and create a follow up PR then, so we can continue with this feature.

kyrtapz (approved):
LGTM

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: If we are hear we have ipRequest as stated by "validateIPRequest" so network cannot be nil.

.github/workflows/test.yml
qinqon: Done at different commit 
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5321/commits/d49c46c030a6dd54d879e2da9143815fc290c80a

qinqon:
@coderabbitai open an issue to track the error at "kv-live-migration" job, stating the "killall iperf3" error

qinqon:
@coderabbitai also add the label "kind/ci-flake" to the issue you just open to track the kv-live-migration job failure

go-controller/pkg/allocator/pod/pod_annotation.go
tssurya: ack sounds good, thanks for checking this.

tssurya:
The virt CI issue is not related to this PR: `failed to run [killall iperf3] at VMI worker-fgzp2, error: killall iperf3 failed` -> that was also seen on another PR and we have opened a new flake for it.

PR #5327, opened 2025-06-24
ovnkube in dpu host mode: advanced gateway detection

aserdean:
This patch introduces a new way to detect the gateway interface in the case ovnkube is running in DPU-HOST mode.

Introduce the argument `derive-from-mgmt-port`, if this is specified as the gateway interface, we will identify the physical function of the device used as an management port accelerated interface.

Signed-off-by: Alin Gabriel Serdean <aserdean@nvidia.com>
Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5239

### DIFF
--- a/docs/features/hardware-offload/dpu-gateway-interface.md
+++ b/docs/features/hardware-offload/dpu-gateway-interface.md
@@ -1,0 +1,208 @@
+# DPU Gateway Interface Configuration
+
+## Overview
+
+In DPU (Data Processing Unit) host mode deployments, OVN-Kubernetes supports automatic gateway interface resolution from PCI address. This feature is particularly useful when the management port is a Virtual Function (VF) and you want to automatically select the corresponding Physical Function (PF) interface as the gateway.
+
+## Background
+
+In DPU deployments, the host typically has access to Virtual Functions (VFs) for management purposes, while the Physical Functions (PFs) are used for external connectivity. The "derive-from-mgmt-port" feature allows OVN-Kubernetes to automatically discover and configure the appropriate PF interface as the gateway interface based on the VF used for the management port.
+
+## How It Works
+
+When configured with `--gateway-interface=derive-from-mgmt-port`, OVN-Kubernetes performs the following steps:
+
+1. **Management Port Resolution**: Gets the management port network device name (specified by `--ovnkube-node-mgmt-port-netdev`)
+2. **VF PCI Address Retrieval**: Retrieves the PCI address of the management port device (VF)
+3. **PF PCI Address Resolution**: Gets the Physical Function (PF) PCI address from the Virtual Function (VF) PCI address
+4. **Network Device Discovery**: Retrieves all network devices associated with the PF PCI address
+5. **Interface Selection**: Selects the first available network device as the gateway interface
+
+## Configuration
+
+### Command Line Options
+
+```bash
+--ovnkube-node-mode=dpu-host
+--ovnkube-node-mgmt-port-netdev=pf0vf0
+--gateway-interface=derive-from-mgmt-port
+```
+
+### Configuration File
+
+```ini
+[OvnKubeNode]
+mode=dpu-host
+mgmt-port-netdev=pf0vf0
+
+[Gateway]
+interface=derive-from-mgmt-port
+```
+
+### Helm Configuration
+
+```yaml
+ovnkube-node:
+  mode: dpu-host
+  mgmtPortNetdev: pf0vf0
+  
+gateway:
+  interface: derive-from-mgmt-port
+```
+
+## Example Scenario
+
+Consider a DPU setup with the following configuration:
+
+- **Management port device**: `pf0vf0` (Virtual Function)
+- **VF PCI address**: `0000:01:02.3`
+- **PF PCI address**: `0000:01:00.0`
+- **Available PF interfaces**: `eth0`, `eth1`
+
+With `--gateway-interface=derive-from-mgmt-port`, OVN-Kubernetes will:
+
+1. Start with the management port device `pf0vf0`
+2. Get its PCI address `0000:01:02.3`
+3. Resolve the PF PCI address to `0000:01:00.0`
+4. Find all network devices associated with PF `0000:01:00.0`: `eth0`, `eth1`
+5. Select `eth0` (first device) as the gateway interface
+
+## Requirements
+
+### Hardware Requirements
+
+- SR-IOV capable network interface card
+- Virtual Function (VF) and Physical Function (PF) setup
+- Management port configured as a VF
+
+### Software Requirements
+
+- SR-IOV utilities available on the system
+- OVN-Kubernetes running in DPU host mode
+- Proper VF/PF driver support
+
+### Configuration Requirements
+
+- Must be used in DPU host mode (`--ovnkube-node-mode=dpu-host`)
+- Management port netdev must be specified (`--ovnkube-node-mgmt-port-netdev`)
+- Gateway interface must be set to `derive-from-mgmt-port`
+
+## Error Handling
+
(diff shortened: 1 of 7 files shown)


romfreiman:
@coderabbitai review

aserdean:
> I'm joining the party a bit late, sorry in advance if anything I'm saying has already been discussed! So according to @trozet 's comment ([#5239 (comment)](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5239#issuecomment-2985277801)), the requirements were:
> 
> > * ovnkube-host cannot know the name of the PF when it first starts
> > * Sometime later, the DPU sets up the PF and presents it into the host
> > * ovnkube-host needs to know which interface this is and set it as the gateway interface
> 
> Here we're initializing the gateway interface for ovnk DPU Host without waiting for the PF to be presented by ovnk DPU. Does it mean ovnk DPU host will fail at this point until the device is ready?

o/

correct, if and only if `config.Gateway.Interface` is set to `from-pci-address`, and that is totally acceptable.

to be on the same page ovnk in dpu host mode will also fail if `config.OvnKubeNode.MgmtPortNetdev` is not properly set
which is a check prior to the newly introduced one.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1028,6 +1028,29 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
 
 	// Complete gateway initialization
 	if config.OvnKubeNode.Mode == types.NodeModeDPUHost {
+		if config.Gateway.Interface == "from-pci-address" {
```
wizhaoredhat: As discussed on the call with Alin and Igal. 
- We need "from-pci-address" as a const.
- Since ""from-pci-address" is the API to enable North-South acceleration of flows through the gateway, it is ok for the API to look like this for now. With Alin we discussed consolidating all the DPU based knobs/parameters into a single parameter; however this refactor would be huge.

For now, let's document this "from-pci-address" API change. Igal will need to add this API into Cluster Network Operator.

Please also make "from-pci-address" a const variable as well.

wizhaoredhat:
@trozet  / @npinaeva  This PR looks good to me, but needs some minor tweeks like const variable and documentation.

aserdean:
thank you for the reviews will rebase and address comments after I finish a round of testing

docs/features/hardware-offload/dpu-gateway-interface.md
```diff
@@ -0,0 +1,195 @@
...
+# Get PF PCI address (if available)
+cat /sys/class/net/pf0vf0/device/physfn/address
+```
+
+## Integration with Existing Features
+
+### Gateway Accelerated Interface
+
+The "from-pci-address" feature can be used in conjunction with gateway accelerated interface configuration. When both are configured, the automatically resolved gateway interface will be used for hardware acceleration.
+
```
trozet: how is it automatically resolved? can you go into more detail here on how they interact together?

docs/features/hardware-offload/dpu-gateway-interface.md
```diff
@@ -0,0 +1,195 @@
+# DPU Gateway Interface Configuration
+
+## Overview
```
trozet, line 3: great docs! thanks!

docs/features/hardware-offload/dpu-gateway-interface.md
```diff
@@ -0,0 +1,195 @@
...
+5. **Interface Selection**: Selects the first available network device as the gateway interface
+
+## Configuration
+
+### Command Line Options
+
+```bash
+--ovnkube-node-mode=dpu-host
+--ovnkube-node-mgmt-port-netdev=pf0vf0
+--gateway-interface=from-pci-address
```
trozet: so you are using the mgmt port VF to derive the PF used for the gateway interface. I'm OK with that, but lets make the value more specific. So instead of "from-pci-address" can we make it "from-mgmt-port-pci-address" ?

ricky-rav: or `derive-from-mgmt-port`, because the input given by the user is just the device name for the management port (`pf0vf0` from above)

docs/features/hardware-offload/dpu-gateway-interface.md
```diff
@@ -0,0 +1,195 @@
...
+4. **Network Device Discovery**: Retrieves all network devices associated with the PF PCI address
+5. **Interface Selection**: Selects the first available network device as the gateway interface
+
+## Configuration
+
+### Command Line Options
+
+```bash
+--ovnkube-node-mode=dpu-host
+--ovnkube-node-mgmt-port-netdev=pf0vf0
```
ricky-rav, line 27: Should we care to mention that `ovnkube-node-mgmt-port-dp-resource-name` can also be specified and the corresponding device name will be derived from the device plugin?

aserdean: I can add a mention to it, but that information should be part of the mgmt-port docs which we do not have AFAIK.

docs/features/hardware-offload/dpu-gateway-interface.md
aserdean: I like `derive-from-mgmt-port` more since that is what we actually do

docs/features/hardware-offload/dpu-gateway-interface.md
aserdean: sure, will add it to the doc

aserdean:
> A couple of comments inline. Other than that, with using the mgmt netdev port to derive the PF...we don't expect the interface to have been created when dpu-host initially comes up, so how do we already know the mgmt port will be "pf0vf0"? Is that just a consistent naming convention?

So the management port can be specified in two ways:

--ovnkube-node-mgmt-port-netdev - a name of an interface 
--ovnkube-node-mgmt-port-dp-resource-name - SRIOV Device Plugin Pool Name (higher prio)

On a non homogenous system we can't ensure naming without udev rules and it is expected for users to use the SRIOV DP resource name. I don't think this is documented but will try to add a note about that in the new docs.

>Also, do we want dpu-host to wait for a certain time before it exits? If the pod crashes a bunch of times kubelet may put it into exponential backoff and take longer for the entire system to converge.

It is a valid concern, but I think that the infrastructure layer should be the one ensuring interfaces are there before the node joins and the pod is started.
For now, although the pod crashes, this serves our use-case well. If this will become an issue we can add more logic around it.

docs/features/hardware-offload/dpu-gateway-interface.md
aserdean: added a more detailed description here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5327/files#diff-595f8346396e001af4f37fe7c4b2d6082bdb8af661658f99671f775f6b2f4d53R182-R186

go-controller/pkg/types/const.go
```diff
@@ -187,6 +187,9 @@ const (
 	NodeModeDPU     = "dpu"
 	NodeModeDPUHost = "dpu-host"
 
+	// Gateway interface configuration
+	FromPCIAddress = "derive-from-mgmt-port"
```
ricky-rav: So this constant can also be renamed `DeriveFromMgmtPort`

ricky-rav (dismissed):
/LGTM
Thanks, @aserdean !

aserdean:
Based on some offline discussions, added: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5388, to document `ovnkube-node-mgmt-port-dp-resource-name`

This will be addressed in another PR.

trozet (approved):

PR #5333, opened 2025-06-26
update documentation on deploying OVN K8s with KIND

### LINKED ISSUE #5322: Issues found while deploying OVN-Kubernetes with KIND on Ubuntu 24.04
### What would you like to see added/updated/deleted in our docs?

There are several issues I noticed when using the guide to deploy with KIND.

**Wrong github repo link in docs**

`git clone github.com/ovn-org/ovn-kubernetes; `
should actually be
`git clone git@github.com:ovn-kubernetes/ovn-kubernetes.git;  `
or
`git clone https://github.com/ovn-kubernetes/ovn-kubernetes.git;`

**Jinjanate install on ubuntu 24.04 fails**
Image

There should be a note for ubuntu users, the fix is:
```
sudo apt-get install pipx
pipx install jinjanator[yaml] 
pipx ensurepath
```

**GO related docs**
Nitpick, but go is not mentioned in the prerequisites. I believe it would be a best practice to at least mention a minimum version. I went with 1.24.4 installed via official guide.

For the podman kind run, if installing go via [official go guide](https://go.dev/doc/install), you need to tell sudo to preserve the path so it can have access to the go binary, since /etc/profile is loaded only on a login shell:

`sudo --preserve-env=PATH ./kind.sh -ep podman`

Also, I found that I need to recursively change ownership on the .kube folder for it to run correctly:

`sudo chown $(id -u):$(id -g) -R ~/.kube`

### Why is this needed?

It's important to have good documentation and it lowerd the barrier of entry for new potential contributors.

GeorgianaTurcsanyi:
## 📑 Description

This PR updates the documentation for deploying OVN-Kubernetes via KIND to fix some problems I found while reproducing the documentation.

Fixes #5322 

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/docs/installation/launching-ovn-kubernetes-on-kind.md
+++ b/docs/installation/launching-ovn-kubernetes-on-kind.md
@@ -18,15 +18,16 @@ KIND (Kubernetes in Docker) deployment of OVN kubernetes is a fast and easy mean
 - jq
 - openssl
 - openvswitch
+- Go 1.23.0 or above
 
-**NOTE :**  In certain operating systems such as CentOS 8.x, pip2 and pip3 binaries are installed instead of pip. In such situations create a softlink for "pip" that points to "pip2".
+**NOTE :** In certain operating systems such as CentOS 8.x, pip2 and pip3 binaries are installed instead of pip. In such situations create a softlink for "pip" that points to "pip2".
 
 For OVN kubernetes KIND deployment, use the `kind.sh` script.
 
 First Download and build the OVN-Kubernetes repo: 
 
 ```
-git clone github.com/ovn-org/ovn-kubernetes; 
+git clone https://github.com/ovn-kubernetes/ovn-kubernetes.git; 
 cd ovn-kubernetes
 ```
 The `kind.sh` script builds OVN-Kubernetes into a container image. To verify
@@ -51,6 +52,13 @@ $ pushd contrib
 $ export KUBECONFIG=${HOME}/ovn.conf
 $ ./kind.sh
 $ popd
+```
+
+**NOTE:** If you run into issues with installing jinjanate on Ubuntu due to [PEP-0668](https://peps.python.org/pep-0668/) you can work around via:
+```
+sudo apt-get install pipx
+pipx install jinjanator[yaml] 
+pipx ensurepath
 ```
 
 ### Run the KIND deployment with podman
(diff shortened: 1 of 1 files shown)


docs/installation/launching-ovn-kubernetes-on-kind.md
```diff
@@ -53,6 +54,13 @@ $ ./kind.sh
 $ popd
 ```
 
+**NOTE:** If you run into issues with installing jinjanate on Ubuntu due to [PEP-0668](https://peps.python.org/pep-0668/) you can work around via:
```
GeorgianaTurcsanyi: Considering this is a PEP issue, I am wondering if the correct way to fix this is to require pipx as prerequisite and fix kind.sh to use pipx instead.

docs/installation/launching-ovn-kubernetes-on-kind.md
```diff
@@ -18,15 +18,16 @@ KIND (Kubernetes in Docker) deployment of OVN kubernetes is a fast and easy mean
 - jq
 - openssl
 - openvswitch
+- Go 1.23.0 or above
```
GeorgianaTurcsanyi, line 21: I took the version from the go mods here, if you think this should be different please let me know.

docs/installation/launching-ovn-kubernetes-on-kind.md
trozet: I think it makes sense to convert it to use pipx

GeorgianaTurcsanyi: > I think it makes sense to convert it to use pipx


Here we go:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5333/files#diff-74a1a8552feeb3203caa0b2f9cc5c9c0eeca76451e37f1a61865c598b3247444R501-R505

docs/installation/launching-ovn-kubernetes-on-kind.md
```diff
@@ -14,19 +14,21 @@ KIND (Kubernetes in Docker) deployment of OVN kubernetes is a fast and easy mean
       sudo firewall-cmd --permanent --add-port=11337/tcp; sudo firewall-cmd --reload
       ```
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
-- Python and pip
+- Python and pipx
 - jq
 - openssl
 - openvswitch
+- Go 1.23.0 or above
+- For podman users: skopeo
```
GeorgianaTurcsanyi, line 22: Added via https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5254, thought it is nice to mention it since the kind deployment fails for podman users if it is not installed.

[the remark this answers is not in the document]
GeorgianaTurcsanyi: Inaccurate statement, resolving

contrib/kind.sh
```diff
@@ -459,8 +459,7 @@ print_params() {
 
 install_jinjanator_renderer() {
   # ensure jinjanator renderer installed
-  pip install wheel --user
-  pip freeze | grep jinjanator || pip install jinjanator[yaml] --user
+  pipx install jinjanator[yaml]
   export PATH=~/.local/bin:$PATH
 }
```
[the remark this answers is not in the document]
trozet: @GeorgianaTurcsanyi should we include this suggested change?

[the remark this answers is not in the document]
GeorgianaTurcsanyi: > @GeorgianaTurcsanyi should we include this suggested change?

I'll add it, with the caveat that for podman users, since they run this script with sudo, what this will do is add jinjanate to the bashrc of the root user, which I don't know how useful it is. But it is indeed cleaner to have the binary in your PATH if you install it anyway.

trozet (approved):
thanks @GeorgianaTurcsanyi !

Issue #5318, opened 2025-06-18
Overlay pod to Host Network Pod/NodePort in a different node causes flooding


hareeshpc:
### What happened?

I am running iperf tests for different kubernetes flows. 

For flows which involve
a) overlay network pod to a host network pod (Different node)
b) overlay network pod to a NodePort (Different node) backed by an overlay network pod

notice that we are hitting this openflow rule[0] on the external bridge managed by ovn-kubernetes
```
cookie=0xdeff105, duration=210055.800s, table=0, n_packets=8412875379, n_bytes=9022421316566, priority=102,ip,in_port="patch-brp0_c-14",dl_src=86:87:74:ba:cf:99,nw_dst=20.0.0.0/24 actions=ct(commit,zone=64000,exec(load:0x1->NXM_NX_CT_MARK[])),NORMAL
```
20.0.0.0/24 is my underlay network connected to external bridge brp0

My underlay node IP (20.0.0.2/24) is a VF(pf0vf1) connected to the external bridge
```
                                                    
       │pf0vf1                                      
       │20.0.0.2/24                                 
   ┌───┼──────────────┐         ┌────────────────┐  
   │                  │         │                │  
   │      brp0        ┼──patch──┤    br-int      │  
   │                  │         │                │  
   │                  │         │                │  
   └────────┬─────────┘         └────────────────┘  
            uplink(p0)                              
            │                                       
         ───┴────────────                           
            L2 underlay                             
            20.0.0.0/24                             
 ```
My flow's Dest IP address is 20.0.0.3 

I see the following datapath flow
```
[...]

PR #5334, opened 2025-06-26
Fixes FDB learning and usage of NORMAL action

trozet:
Commit f978967 caused a regression in performance. As the below issue
describes, the egress traffic from OVN will now use NORMAL action, which
will cause an FDB lookup and then FLOOD if not found. This always ends
up being the case because the reply ARP packet from the physical port is
flooded to the patch port and the LOCAL port. This causes an increase in
CPU and unnecessarily flooding packets.

We need layer 2 packets destined to the shared gateway mac to go to both
the host and OVN. This is so both can receive ARP replies, etc. However,
we also need the FDB entry in OVS to get updated, for our new
functionality with using the NORMAL action.

To fix this, add a static FDB entry for LOCAL, then modify the layer 2
flooding flow actions from "output:patch,LOCAL" to
"output:patch,NORMAL". Since the FDB entry is bound in the table to
LOCAL, it is effectively forwarding the packets the same as before, but
with the added bonus of FDB learning on ingress.

Fixes: #5318

## Summary by CodeRabbit

* **New Features**
  * Improved gateway initialization by setting a static MAC address entry for the LOCAL port on the gateway bridge.

* **Bug Fixes**
  * Adjusted packet forwarding behavior for shared gateway bridges to enhance compatibility and reliability.

* **Tests**
  * Enhanced test coverage to simulate and verify the addition of static forwarding database (FDB) entries during gateway setup.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1899,8 +1899,8 @@ func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]strin
 		}
 		actions += strip_vlan + "NORMAL"
 		dftFlows = append(dftFlows,
-			fmt.Sprintf("cookie=%s, priority=10, table=0, in_port=%s, %s dl_dst=%s, actions=%s",
-				defaultOpenFlowCookie, ofPortPhys, match_vlan, bridgeMacAddress, actions))
+			fmt.Sprintf("cookie=%s, priority=10, table=0, %s dl_dst=%s, actions=%s",
+				defaultOpenFlowCookie, match_vlan, bridgeMacAddress, actions))
 	}
 
 	// table 0, check packets coming from OVN have the correct mac address. Low priority flows that are a catch all


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1897,7 +1897,7 @@ func commonFlows(hostSubnets []*net.IPNet, bridge *bridgeConfiguration) ([]strin
 		for _, netConfig := range bridge.patchedNetConfigs() {
 			actions += "output:" + netConfig.ofPortPatch + ","
 		}
-		actions += strip_vlan + "output:" + ofPortHost
+		actions += strip_vlan + "NORMAL"
```
ricky-rav, line 1900: I see, so ARP replies were hitting the `output:patch,LOCAL` rule, which doesn't populate the FDB, but only outputs to OVN and the host. So also all subsequent outgoing packets destined to the provider network would flood all ports with the `prio=102` rule I introduced. The `output:patch,LOCAL` rule now makes sure that ARP replies will first populate the FDB and then the `prio=102` rule will know to which port it should forward traffic.

Is that correct?

ricky-rav:
Now OVN->localnet is broken https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/15913896902/job/44888804199?pr=5334 :(

go-controller/pkg/node/gateway_shared_intf.go
trozet: right, except the updated rule is now output:patch,NORMAL

trozet:
> Now OVN->localnet is broken https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/15913896902/job/44888804199?pr=5334 :(

Thanks @ricky-rav. I figured out why this is. The GR in  OVN is never getting the ARP reply because we now populate the FDB with a static entry locked to the LOCAL port. That means when the GR arps for the localnet VM IP, the localnet VM ARP replies, but it gets sent to the host via the normal action. We basically need the same behavior as I added in my 2nd commit (send to patch and NORMAL), but for the localnet VM as well. So adding a flow manually:

```
ovs-ofctl add-flow breth0 table=0,priority=5,dl_dst=4e:9b:8c:55:2a:d3,actions=output:"patch-breth0_ov",NORMAL
```

makes the test pass. The flow we have in place today matches on the physical port:
```
cookie=0xdeff105, duration=63410.430s, table=0, n_packets=3674, n_bytes=154564, priority=10,in_port=eth0,dl_dst=4e:9b:8c:55:2a:d3 actions=output:"patch-breth0_ov",NORMAL
```

we can potentially drop the match on physical port to fix this problem.

trozet:
looks like a flake:
```
2025-06-27T17:06:52.8647351Z   [38;5;9m[FAILED][0m in [It] - /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/gateway_localnet_linux_test.go:1477 [38;5;243m@ 06/27/25 17:06:52.864[0m
2025-06-27T17:06:52.8648662Z [38;5;9m• [FAILED] [2.200 seconds][0m
2025-06-27T17:06:52.8649786Z [0mNode Operations [38;5;243mon delete [38;5;9m[1m[It] deletes iptables rules with ExternalIP[0m
2025-06-27T17:06:52.8651165Z [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/gateway_localnet_linux_test.go:1437[0m
2025-06-27T17:06:52.8652173Z 
2025-06-27T17:06:52.8652505Z   [38;5;9m[FAILED] Timed out after 2.000s.
2025-06-27T17:06:52.8653074Z   Executed commands do not match expected commands!
2025-06-27T17:06:52.8653721Z   Executed unexpected command /fake-bin/ovs-ofctl show 
2025-06-27T17:06:52.8654089Z 
2025-06-27T17:06:52.8654582Z   [00]   /fake-bin/ovs-ofctl show 
2025-06-27T17:06:52.8655054Z   [01]   /fake-bin/ovs-ofctl show 
2025-06-27T17:06:52.8655491Z   [02] + /fake-bin/ovs-ofctl show 
2025-06-27T17:06:52.8655736Z 
2025-06-27T17:06:52.8655865Z   Expected
2025-06-27T17:06:52.8656178Z       <bool>: false
2025-06-27T17:06:52.8656748Z   to be true[0m
```

ricky-rav (approved):
/LGTM
Thanks, Tim!

PR #5339, opened 2025-06-30
scale test fixes: memory leak + changed ofport

npinaeva:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of patch port changes for non-default networks, providing clearer warning messages and expecting bridge configuration updates instead of exiting.
  * Enhanced shutdown process for service controllers by explicitly removing event handlers and cleaning up resources, reducing the risk of resource leaks.

* **Improvements**
  * Updated network reconciliation logic to prevent updates when network IDs differ, ensuring more reliable network management.

### DIFF
--- a/go-controller/pkg/node/openflow_manager.go
+++ b/go-controller/pkg/node/openflow_manager.go
@@ -275,12 +275,12 @@ func checkPorts(netConfigs []*bridgeUDNConfiguration, physIntf, ofPortPhys strin
 
 		}
 		if netConfig.ofPortPatch != curOfportPatch {
-			if netConfig.isDefaultNetwork() || curOfportPatch != "" {
+			if netConfig.isDefaultNetwork() {
 				klog.Errorf("Fatal error: patch port %s ofport changed from %s to %s",
 					netConfig.patchPort, netConfig.ofPortPatch, curOfportPatch)
 				os.Exit(1)
 			} else {
-				klog.Warningf("Patch port %s removed for existing network", netConfig.patchPort)
+				klog.Warningf("UDN patch port %s changed for existing network from %v to %v. Expecting bridge config update.", netConfig.patchPort, netConfig.ofPortPatch, curOfportPatch)
 			}
 		}
 	}
--- a/go-controller/pkg/ovn/controller/services/services_controller.go
+++ b/go-controller/pkg/ovn/controller/services/services_controller.go
@@ -168,6 +168,11 @@ type Controller struct {
 	useTemplates bool
 
 	netInfo util.NetInfo
+
+	// handlers stored for shutdown
+	nodeHandler     cache.ResourceEventHandlerRegistration
+	svcHandler      cache.ResourceEventHandlerRegistration
+	endpointHandler cache.ResourceEventHandlerRegistration
 }
 
 // Run will not return until stopCh is closed. workers determines how many
@@ -180,15 +185,15 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}, wg *sync.WaitGroup
 		// wait until we're told to stop
 		<-stopCh
 
-		klog.Infof("Shutting down controller %s for network=%s", controllerName, c.netInfo.GetNetworkName())
-		c.queue.ShutDown()
+		c.Cleanup()
 	}()
 
 	c.useLBGroups = useLBGroups
 	c.useTemplates = useTemplates
 	klog.Infof("Starting controller %s for network=%s", controllerName, c.netInfo.GetNetworkName())
 
-	nodeHandler, err := c.nodeTracker.Start(c.nodeInformer)
+	var err error
+	c.nodeHandler, err = c.nodeTracker.Start(c.nodeInformer)
 	if err != nil {
 		return err
 	}
@@ -197,12 +202,12 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}, wg *sync.WaitGroup
 	c.startupDoneLock.Lock()
 	c.startupDone = false
 	c.startupDoneLock.Unlock()
-	if !util.WaitForHandlerSyncWithTimeout(nodeControllerName, stopCh, types.HandlerSyncTimeout, nodeHandler.HasSynced) {
+	if !util.WaitForHandlerSyncWithTimeout(nodeControllerName, stopCh, types.HandlerSyncTimeout, c.nodeHandler.HasSynced) {
 		return fmt.Errorf("error syncing node tracker handler")
 	}
 
 	klog.Infof("Setting up event handlers for services for network=%s", c.netInfo.GetNetworkName())
-	svcHandler, err := c.serviceInformer.Informer().AddEventHandler(factory.WithUpdateHandlingForObjReplace(cache.ResourceEventHandlerFuncs{
+	c.svcHandler, err = c.serviceInformer.Informer().AddEventHandler(factory.WithUpdateHandlingForObjReplace(cache.ResourceEventHandlerFuncs{
 		AddFunc:    c.onServiceAdd,
 		UpdateFunc: c.onServiceUpdate,
 		DeleteFunc: c.onServiceDelete,
@@ -212,7 +217,7 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}, wg *sync.WaitGroup
 	}
 
 	klog.Infof("Setting up event handlers for endpoint slices for network=%s", c.netInfo.GetNetworkName())
-	endpointHandler, err := c.endpointSliceInformer.Informer().AddEventHandler(factory.WithUpdateHandlingForObjReplace(
+	c.endpointHandler, err = c.endpointSliceInformer.Informer().AddEventHandler(factory.WithUpdateHandlingForObjReplace(
 		// Filter out endpointslices that don't belong to this network (i.e. keep only kube-generated endpointslices if
 		// on default network, keep only mirrored endpointslices for this network if on UDN)
 		util.GetEndpointSlicesEventHandlerForNetwork(
@@ -227,7 +232,7 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}, wg *sync.WaitGroup
 	}
 
 	klog.Infof("Waiting for service and endpoint handlers to sync for network=%s", c.netInfo.GetNetworkName())
-	if !util.WaitForHandlerSyncWithTimeout(controllerName, stopCh, types.HandlerSyncTimeout, svcHandler.HasSynced, endpointHandler.HasSynced) {
+	if !util.WaitForHandlerSyncWithTimeout(controllerName, stopCh, types.HandlerSyncTimeout, c.svcHandler.HasSynced, c.endpointHandler.HasSynced) {
 		return fmt.Errorf("error syncing service and endpoint handlers")
 	}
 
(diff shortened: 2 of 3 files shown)


npinaeva:
cc @dceara to comment on the second commit, as you are the last person who touched that openflow manager part here https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4786

go-controller/pkg/util/multi_network.go
```diff
@@ -822,6 +822,11 @@ func (nInfo *secondaryNetInfo) canReconcile(other NetInfo) bool {
 	if nInfo == nil && other == nil {
 		return true
 	}
+	// if network ID has changed, it means the network was re-created, and all controllers
+	// should execute delete+create instead of update
+	if nInfo.id != types.InvalidID && other.GetNetworkID() != types.InvalidID && nInfo.id != other.GetNetworkID() {
```
npinaeva: maybe that is an overkill and I need to do `if nInfo.GetNetworkID() != other.GetNetworkID()`, idk

dceara: Hi @npinaeva!  Will this cause `openflowManager.delNetwork(nInfo)` to be called eventually?  I'm not sure I see any way the `bridgeConfiguration.ofPortPatch` can change without a `delNetwork()` call.

npinaeva: yep! the networkID check from that commit is what makes sure network will be re-created by all controllers (mentioned in the description node-nad-controller is the one that calls openflow_manager eventually)

dceara: I see, then the change looks good to me, thanks!

go-controller/pkg/node/openflow_manager.go
```diff
@@ -275,12 +275,12 @@ func checkPorts(netConfigs []*bridgeUDNConfiguration, physIntf, ofPortPhys strin
...
 		}
 		if netConfig.ofPortPatch != curOfportPatch {
-			if netConfig.isDefaultNetwork() || curOfportPatch != "" {
+			if netConfig.isDefaultNetwork() {
 				klog.Errorf("Fatal error: patch port %s ofport changed from %s to %s",
 					netConfig.patchPort, netConfig.ofPortPatch, curOfportPatch)
 				os.Exit(1)
 			} else {
-				klog.Warningf("Patch port %s removed for existing network", netConfig.patchPort)
+				klog.Warningf("UDN patch port %s changed for existing network from %v to %v. Expecting bridge config update.", netConfig.patchPort, netConfig.ofPortPatch, curOfportPatch)
```
kyrtapz, line 283: Isn't this going to cause unnecessary logs on network startup? I think we should still check if the `curOfportPatch` is not empty.

Additionally, would it be possible to emit a warning event in this case? While I completely agree we should not exit on udn patch port missmatch I think an event would be a god way to highlight an important issue.

go-controller/pkg/util/multi_network.go
```diff
@@ -822,6 +822,11 @@ func (nInfo *secondaryNetInfo) canReconcile(other NetInfo) bool {
 	if nInfo == nil && other == nil {
 		return true
 	}
+	// if network ID has changed, it means the network was re-created, and all controllers
+	// should execute delete+create instead of update
+	if nInfo.GetNetworkID() != types.InvalidID && other.GetNetworkID() != types.InvalidID && nInfo.GetNetworkID() != other.GetNetworkID() {
```
kyrtapz, line 827: I don't think we should ignore the case of going from a valid id to invalid one.
```
if nInfo.GetNetworkID() != types.InvalidID && nInfo.GetNetworkID() != other.GetNetworkID() {
```
should be enough imo.

go-controller/pkg/ovn/controller/services/services_controller.go
```diff
@@ -255,6 +260,27 @@ func (c *Controller) Run(workers int, stopCh <-chan struct{}, wg *sync.WaitGroup
 	return nil
 }
 
+func (c *Controller) Cleanup() {
+	klog.Infof("Shutting down controller %s for network=%s", controllerName, c.netInfo.GetNetworkName())
+	c.queue.ShutDown()
+
+	if c.nodeHandler != nil {
```
kyrtapz, line 267: Were you able to determine if other UDN controllers suffer from this?
I think https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/ovn/controller/udnenabledsvc/udn_enabled_svc.go#L84 might be the same?

kyrtapz (commented):
Overall the changes look good to me. 

The solution in the second commit is very similar to what we were trying here
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5337

While it does not include https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5337/files#diff-0b74e004432877a3ca7b23367059bb077ffa13c01d26f6e3a0f37d09acc67403R435-R465
I think it is fine as the NAD controller should be really fast to reconcile any changes (compared to the network ctrl). We can always add it as a followup.

go-controller/pkg/node/openflow_manager.go
npinaeva: A couple lines above there is a check 
```
if netConfig.ofPortPatch == "" {
			continue
		}
```
so it won't do anything until netconfig is updated with the patch port, the check with the empty port was only allowing port mismatch when the actual port is removed but the netConfig is not. I want to allow an actual mismatch when ofPort is newly configured, but the netConfig hasn't been updated yet.
I think an event I can try

go-controller/pkg/ovn/controller/services/services_controller.go
npinaeva: good point, I was lazy :)

go-controller/pkg/util/multi_network.go
npinaeva: That makes sense, but technically `canReconcile` doesn't say which one is old and which is new. Also if we consider this quick re-create event, a newly created network can have invalidID (until it is assigned a new one) and the not-yet-deleted network will have the old ID. So going to invalid ID could also be a new network event?

go-controller/pkg/node/openflow_manager.go
kyrtapz: > A couple lines above there is a check

Hmm, this is not the same as `curOfportPatch` that is retrieved from OVS?
edit: are you saying that it's not possible to have the netconfig one set and the one in ovs unset?

go-controller/pkg/util/multi_network.go
kyrtapz: > That makes sense, but technically canReconcile doesn't say which one is old and which is new.

Although we use it this way that is a good point, `AreNetworksCompatible` doesn't differentiate new and current technically.
 
IIUC it is hard to get a `InvalidID` with UDN because we are now setting the ID as an annotation on the NAD when it is being created. 
But in general if a network switches IDs I would expect for it to be recreated, even if the it switches to/from is set to `InvalidID`.

go-controller/pkg/node/openflow_manager.go
npinaeva: yeah you have to dig though to figure this out :D
so the netconfig ofport is set by gateway initFunc, and openflow_manager check_ports is started to gateway.Start, which happens after

go-controller/pkg/ovn/controller/services/services_controller.go
npinaeva: ok, I checked that all controllers that use `AddEventHandler` are only stopped with the whole binary, e.g. this one is only created once with the controller_managaer here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/256b38cc4bd347b5f8ce652c07195f868df614ce/go-controller/pkg/controllermanager/controller_manager.go#L468-L476
even though a lot of controllers using this ^ directly could be refactored, I am not doing it in this pr :)

go-controller/pkg/node/openflow_manager.go
npinaeva: I have also tried adding an event, but that requires passing a lot of info, like event recorder and at least nodename (if we record event on the node or UDN if we record an event for UDN).
I am working on refactoring this part right now (first part here https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5345), so I will try to add an event recorder later when it is easier if that is ok

go-controller/pkg/util/multi_network.go
npinaeva: @kyrtapz I think removing this check breaks normal NAD use case (not the UDN-created NAD, but manually created NAD)
thinking about the network re-create use case, even if new networkID is invalid, it will get a new ID eventually, and we will re-create the network later?

kyrtapz: I guess that's ok to keep the check as is, it is not really likely the network will switch to an invalid ID anyway.
How does it break a normal NAD use case though? I am curious.

kyrtapz:
I haven't seen this failure before:
```
```
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16141339356/job/45551429357?pr=5339

While I doubt it is relate I think it would be best to confirm.

kyrtapz (approved):

trozet:
> Overall the changes look good to me.
> 
> The solution in the second commit is very similar to what we were trying here #5337
> 
> While it does not include https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5337/files#diff-0b74e004432877a3ca7b23367059bb077ffa13c01d26f6e3a0f37d09acc67403R435-R465 I think it is fine as the NAD controller should be really fast to reconcile any changes (compared to the network ctrl). We can always add it as a followup.

@kyrtapz I think we still need the other part. The nad controller could be backed up with a stale network on it, then a pod comes into be wired and we attach to the wrong network. Right? I don't think it hurts to add it.

go-controller/pkg/node/openflow_manager.go
```diff
@@ -275,12 +275,12 @@ func checkPorts(netConfigs []*bridgeUDNConfiguration, physIntf, ofPortPhys strin
 
 		}
 		if netConfig.ofPortPatch != curOfportPatch {
-			if netConfig.isDefaultNetwork() || curOfportPatch != "" {
+			if netConfig.isDefaultNetwork() {
```
trozet, line 278: Just to highlight this for the future...the check ports is here in case ovn-controller has a bug and recreates the port:

https://github.com/openshift/ovn-kubernetes/commit/0b04aa4f4f7310f402cb65f3fe62317a5fe33ce5

"it so happens that ovn-controller, when it hits a snag, it can tear
down this patch cable under the foot of ovnkube-node and add it back.
this changes the `ofport` number and all that openflow rules are now
invalid"

This commit was from a long time ago, perhaps this is no longer a concern. I haven't seen it happen in years as a real failure. I think it is safe to ignore this for  now with UDNs, if we ever see a real bug in ovn from this then we can refer back to this comment and see if we want to add the functionality back.

kyrtapz: Thanks for the context!
I do think that even if there is a bug in ovn-controller in the future we should not crash ovn-k because some  UDN failed, we could certainly try to be more vocal through events as I suggested above but that can be done "in the future".

PR #5346, opened 2025-07-03
VF gateway trigger errors message in updateServiceFlowCache

cathy-zhou:
When gateway accelerated interface is used, we noticed the error messages 'gateway_shared_intf.go:392] Unable to get port list from bridge. ... failed to get list of ports on bridge "enp1s0f0v0":, stderr: "ovs-ofctl: enp1s0f0v0 is not a bridge or a socket\n" ...'.

Also, bridgeConfiguration.getGatewayIface() function is confusing, as b.gwIface is always non-empty, so one can just directly use b.gwIface.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -460,7 +460,7 @@ func (g *gateway) GetGatewayBridgeIface() string {
 }
 
 func (g *gateway) GetGatewayIface() string {
-	return g.openflowManager.defaultBridge.getGatewayIface()
+	return g.openflowManager.defaultBridge.gwIface
 }
 
 // getMaxFrameLength returns the maximum frame size (ignoring VLAN header) that a gateway can handle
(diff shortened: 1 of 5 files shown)


cathy-zhou:
@winsopc @wizhaoredhat @hareeshpc can you please take a look?

wizhaoredhat:
LGTM

trozet (changes requested):
@cathy-zhou I think there is one place you missed in:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/default_node_network_controller.go#L1189

this should be changed to use bridge name

winsopc:
> @cathy-zhou I think there is one place you missed in:
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/default_node_network_controller.go#L1189
> 
> this should be changed to use bridge name

hi @trozet ,  I worked on downstream part of the gateway support.
this part should be gateway interface which has the IP address.  if we use bridge name, netlink will return error
eg.

` failed to add route (gw: 169.254.169.4, subnet 10.223.0.0/16, mtu 1500, src IP 169.254.169.2): network is unreachable`

winsopc:
LGTM

trozet:
@winsopc OK but then I think we should still change the code from bridge->interface for clarity. Like this function:

```
func configureSvcRouteViaBridge(routeManager *routemanager.Controller, bridge string) error {
	return configureSvcRouteViaInterface(routeManager, bridge, DummyNextHopIPs())
}
```

Let's just eliminate the above function (only used in 1 place) and call configureSvcRouteViaInterface directly:
```
	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
		if config.OvnKubeNode.Mode == types.NodeModeFull {
			// Configure route for svc towards shared gateway interface
			if err := configureSvcRouteViaInterface(nc.routeManager, nc.Gateway.GetGatewayIface(), DummyNextHopIPs()); err != nil {
				return err
			}
		}
	}
```

PR #5314, opened 2025-06-18
Revert "Add the IP rule for a UDN only when it is advertised to the default VRF"

jcaamano:
This reverts commit deff5e64ccc6069437bf7abf08f61522f73501a1.

Breaks traffic flows to KAPI, DNS on VRF-Lite scenarios. Requires and SNAT that is being worked on [1].

1. https://issues.redhat.com/browse/OCPBUGS-56506?focusedId=27440592&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-27440592

## Summary by CodeRabbit

* **Refactor**
  * Network advertisement status is now computed dynamically on demand instead of using cached values.
  * Simplified logic for managing network rules and routes based on advertisement status passed as a parameter.
  * Removed redundant methods and streamlined control flow related to network advertisement handling.

* **Tests**
  * Updated tests to pass advertisement status as a parameter rather than setting internal flags.
  * Consolidated and removed redundant test cases for improved clarity and maintenance.

PR #5348, opened 2025-07-03
Vrf lite e2e

jcaamano:
Adds L3 VRF-Lite test cases covering:

- positive connectivity tests from/towards external server on CUDN VRF
- positive connectivity tests towards other pod/cluster IP service on the CUDN, same/different node
- negative connectivity tests from/towards external server on different VRF
- negative connectivity tests towards other pod/cluster IP service on other default/L3 CUDN/L3 CUDN advertised/L3 CUDN advertised VRF Lite/ L2 UDN; same/different node

Includes a fix on RA controller not updating the RA status when an external circumstance corrects its status without further updates done by the controller itself.

Includes https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5314 temporarily for tests to pass.


test/e2e/testdata/routeadvertisements/frr-k8s/frrconf.yaml.tmpl
```diff
@@ -0,0 +1,44 @@
...
+{{- end }}
+spec:
+  bgp:
+    routers:
+{{- range $v := .Routers }}
+    - asn: 64512
+{{- if .VRF }}
+      vrf: {{ .VRF }}
+{{- end }}
+      neighbors:
```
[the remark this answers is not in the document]
jcaamano: This is valid yaml and can be parsed.

test/e2e/egressip.go
```diff
@@ -753,13 +753,13 @@ var _ = ginkgo.DescribeTableSubtree("e2e egress IP validation", feature.EgressIP
...
+			Network: primaryProviderNetwork, CmdArgs: getAgnHostHTTPPortBindCMDArgs(primaryTargetExternalContainerPort), ExtPort: primaryTargetExternalContainerPort}
 		primaryTargetExternalContainer, err = providerCtx.CreateExternalContainer(primaryTargetExternalContainerSpec)
 		framework.ExpectNoError(err, "failed to create external target container on primary network", primaryTargetExternalContainerSpec.String())
 
 		primaryDeniedExternalContainerPort := infraprovider.Get().GetExternalContainerPort()
 		primaryDeniedExternalContainerSpec := infraapi.ExternalContainer{Name: deniedTargetNodeName, Image: images.AgnHost(),
-			Network: primaryProviderNetwork, Args: getAgnHostHTTPPortBindCMDArgs(primaryDeniedExternalContainerPort), ExtPort: primaryDeniedExternalContainerPort}
+			Network: primaryProviderNetwork, CmdArgs: getAgnHostHTTPPortBindCMDArgs(primaryDeniedExternalContainerPort), ExtPort: primaryDeniedExternalContainerPort}
 		primaryDeniedExternalContainer, err = providerCtx.CreateExternalContainer(primaryDeniedExternalContainerSpec)
 		framework.ExpectNoError(err, "failed to create external denied container on primary network", primaryDeniedExternalContainer.String())
```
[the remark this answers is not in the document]
jcaamano: This PR only refactored the field name from `Args` to `CmdArgs`. Any new uses of this field introduced in this PR already account for the image entry point.

jcaamano:
> **Actionable comments posted: 3**
> 
> 🔭 Outside diff range comments (1)
> > go-controller/pkg/node/gateway_udn.go (1)> `209-210`: **Fix incorrect atomic store in shallowCopy method.**
> > > The advertised value is being stored back to the source instead of the copy.
> > > ```diff
> > > -	netConfig.advertised.Store(netConfig.advertised.Load())
> > > +	copy.advertised.Store(netConfig.advertised.Load())
> > > ```

Fixing this.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -955,10 +955,18 @@ func (c *Controller) updateRAStatus(ra *ratypes.RouteAdvertisements, hadUpdates
...
 	condition := meta.FindStatusCondition(ra.Status.Conditions, "Accepted")
-	updateStatus := hadUpdates || condition == nil || condition.ObservedGeneration != ra.Generation
-	updateStatus = updateStatus || err != nil
-
+	switch {
+	case condition == nil:
+		fallthrough
+	case condition.ObservedGeneration != ra.Generation:
+		fallthrough
+	case (err == nil) != (condition.Status == metav1.ConditionTrue):
```
kyrtapz, line 961: does this basically handle a case where the status was changed by something else? 
I guess I don't understand what do you mean by:
```
The status of a RouteAdvertisements might change due to external
conditions to the controller.
```

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -119,7 +119,18 @@ func (k *kind) GetK8HostPort() uint16 {
 
 func (k *kind) NewTestContext() api.Context {
 	ck := &contextKind{Mutex: sync.Mutex{}}
-	ginkgo.DeferCleanup(ck.CleanUp)
+	if !framework.TestContext.DeleteNamespace {
```
kyrtapz: Remove? :)

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -119,7 +119,18 @@ func (k *kind) GetK8HostPort() uint16 {
...
+
+	}
+	ginkgo.DeferCleanup(func() error {
+		if !framework.TestContext.DeleteNamespace {
```
kyrtapz: Can you provide a little more context around why you want to avoid infra cleanup in these case?
I guess it is for debugging right? While I like that addition I don't think it is related to the framework ns deletion.

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -427,13 +427,13 @@ func (c *contextKind) cleanUp() error {
...
 	nameFormat                     = "{{.Name}}"
 	inspectNetworkIPAMJSON         = "{{json .IPAM.Config }}"
-	inspectNetworkIPv4GWKeyStr     = "{{ .NetworkSettings.Networks.%s.Gateway }}"
-	inspectNetworkIPv4AddrKeyStr   = "{{ .NetworkSettings.Networks.%s.IPAddress }}"
-	inspectNetworkIPv4PrefixKeyStr = "{{ .NetworkSettings.Networks.%s.IPPrefixLen }}"
-	inspectNetworkIPv6GWKeyStr     = "{{ .NetworkSettings.Networks.%s.IPv6Gateway }}"
-	inspectNetworkIPv6AddrKeyStr   = "{{ .NetworkSettings.Networks.%s.GlobalIPv6Address }}"
-	inspectNetworkIPv6PrefixKeyStr = "{{ .NetworkSettings.Networks.%s.GlobalIPv6PrefixLen }}"
-	inspectNetworkMACKeyStr        = "{{ .NetworkSettings.Networks.%s.MacAddress }}"
+	inspectNetworkIPv4GWKeyStr     = "{{ with index .NetworkSettings.Networks \"%s\" }}{{ .Gateway }}{{ end }}"
```
kyrtapz: Why not `%q`? 
Can you also share what `with index` changes here?

test/e2e/network_segmentation.go
```diff
@@ -1881,31 +1881,17 @@ func generateLayer3Subnets(cidrs string) []string {
...
 
 // userDefinedNetworkReadyFunc returns a function that checks for the NetworkCreated condition in the provided cluster udn
 func clusterUserDefinedNetworkReadyFunc(client dynamic.Interface, name string) func() error {
+	return networkReadyFunc(client.Resource(clusterUDNGVR), name)
+}
+
+func networkReadyFunc(client dynamic.ResourceInterface, name string) func() error {
 	return func() error {
-		cUDN, err := client.Resource(clusterUDNGVR).Get(context.Background(), name, metav1.GetOptions{}, "status")
+		cUDN, err := client.Get(context.Background(), name, metav1.GetOptions{}, "status")
```
kyrtapz, line 1900: Nice! I wasn't aware we could go this generic.

test/e2e/testdata/routeadvertisements/frr/daemons.tmpl
```diff
@@ -0,0 +1,82 @@
...
+fabricd=no
+vrrpd=no
+
+#
+# If this option is set the /etc/init.d/frr script automatically loads
+# the config via "vtysh -b" when the servers are started.
+# Check /etc/pam.d/frr if you intend to use "vtysh"!
+#
+vtysh_enable=yes
+zebra_options="  -A 127.0.0.1 -s 90000000"
```
kyrtapz, line 41: Considering the amount of ipv4 addresses here do you think this will work well in a purely ipv6 environment? Maybe this should be templated based on the IP family of the cluster?

test/e2e/util.go
```diff
@@ -1342,9 +1362,62 @@ func matchIPv6StringFamily(ipStrings []string) (string, error) {
...
+		case utilnet.IsIPv4String(ip):
+			ipv4 = append(ipv4, ip)
+		case utilnet.IsIPv6String(ip):
+			ipv6 = append(ipv6, ip)
+		}
+	}
+	return
+}
+
+func getFirstCIDROfFamily(family utilnet.IPFamily, ipnets []*net.IPNet) *net.IPNet {
```
kyrtapz, line 1404: NIT: We already have https://github.com/ovn-kubernetes/ovn-kubernetes/blob/192f575d351acc9bc049db213bf65f25a315e2c6/go-controller/pkg/util/net.go#L212

test/e2e/util.go
```diff
@@ -1424,3 +1497,38 @@ func getAgnHostHTTPPortBindFullCMD(port uint16) []string {
...
+	}
+	defer f.Close()
+	err = templates.ExecuteTemplate(f, name, data)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// podIPsForUserDefinedPrimaryNetwork returns the v4 or v6 IPs for a pod on the UDN
```
kyrtapz: What about https://github.com/ovn-kubernetes/ovn-kubernetes/blob/58a10f88e1f4520aaf809dcb51a7515ad364e2a4/test/e2e/network_segmentation.go#L2251? :) 
I would like for us to avoid adding yet another function doing something really similar.

test/e2e/route_advertisements.go
```diff
@@ -1053,6 +1058,596 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+		// FIXME: each test brings its own topology up, and sometimes zebra on
+		// external FRR container fails to start on the first attempt for
+		// unknown reasons delaying the overall availability, so we need to use
+		// long timeouts
+		timeout    = 240 * time.Second
+		timeoutNOK = 10 * time.Second
+	)
+	testPodToHostnameAndExpect := func(src *corev1.Pod, dstIP, expect string) {
+		ginkgo.GinkgoHelper()
+		hostname, err := e2epodoutput.RunHostCmdWithRetries(
```
kyrtapz, line 1089: Have you considered using https://github.com/ovn-kubernetes/ovn-kubernetes/blob/58a10f88e1f4520aaf809dcb51a7515ad364e2a4/test/e2e/route_advertisements.go#L757-L786?
It seems that we are repeating the same logic in each `test...` function.

test/e2e/route_advertisements.go
```diff
@@ -1053,6 +1058,596 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+			Topology: udnv1.NetworkTopologyLayer3,
+			Layer3: &udnv1.Layer3Config{
+				Role: "Primary",
+				Subnets: generateL3Subnets(
+					udnv1.Layer3Subnet{CIDR: cudnCIDRv4, HostSubnet: 24},
+					udnv1.Layer3Subnet{CIDR: cudnCIDRv6, HostSubnet: 64}),
+			},
+		}
+	)
+	networksToTest := []ginkgo.TableEntry{
```
kyrtapz, line 1234: What's the point of defining this here instead of adding the entry directly?
So Layer 2 doesn't work even in shared GW now?

test/e2e/route_advertisements.go
```diff
@@ -1053,6 +1058,596 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+				nodeList, err := f.ClientSet.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				network, err := infraprovider.Get().GetNetwork(testNetworkName)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				for _, node := range nodeList.Items {
+					iface, err := infraprovider.Get().GetK8NodeNetworkInterface(node.Name, network)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+					_, err = infraprovider.Get().ExecK8NodeCommand(node.Name, []string{"ip", "link", "set", "dev", iface.InfName, "master", testNetworkName})
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+					// quirk: need to reset IPv6 address
```
kyrtapz, line 1286: Just for my own knowledge, why is that?

test/e2e/route_advertisements.go
```diff
@@ -1053,6 +1058,596 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+							},
+						}
+					)
+
+					otherNetworksToTest := []ginkgo.TableEntry{
+						ginkgo.Entry("Default", defaultNetwork, nil),
+						ginkgo.Entry("Layer 3 UDN non advertised", udn, otherLayer3NetworkSpec),
+						ginkgo.Entry("Layer 3 CUDN advertised", cudnAdvertised, otherLayer3NetworkSpec),
+						ginkgo.Entry("Layer 3 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer3NetworkSpec),
+						ginkgo.Entry("Layer 2 UDN non advertised", udn, otherLayer2NetworkSpec),
```
kyrtapz: Why there is no variant for advertised L2 network?

test/e2e/route_advertisements.go
```diff
@@ -1053,6 +1058,596 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+
+										ginkgo.DescribeTable("The pod on the tested network cannot reach the service on the other network",
+											func(family utilnet.IPFamily) {
+												if !isIPFamilySupported(family) {
+													e2eskipper.Skipf("IP family %v not supported", family)
+												}
+												ginkgo.By("Ensuring a request from the tested network pod cannot reach the other network pod")
+												clusterIP := getFirstIPStringOfFamily(family, service.Spec.ClusterIPs)
+												gomega.Expect(clusterIP).ToNot(gomega.BeEmpty())
+												gomega.Consistently(func(g gomega.Gomega) {
```
kyrtapz: Most of the connectivity tests have their own wrapper but a few don't, was there any specific reason?
Maybe you could use `testContainerToClientIPNOK` here?

test/e2e/route_advertisements.go
```diff
@@ -1053,6 +1058,596 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+		// long timeouts
+		timeout    = 240 * time.Second
+		timeoutNOK = 10 * time.Second
+	)
+	testPodToHostnameAndExpect := func(src *corev1.Pod, dstIP, expect string) {
+		ginkgo.GinkgoHelper()
+		hostname, err := e2epodoutput.RunHostCmdWithRetries(
+			src.Namespace,
+			src.Name,
+			fmt.Sprintf("curl --max-time 2 -g -q -s http://%s/hostname", net.JoinHostPort(dstIP, "8080")),
```
kyrtapz: Can you define `8080` as a variable and maybe randomize it?

test/e2e/route_advertisements.go
```diff
@@ -1074,3 +1669,519 @@ func generateL2Subnets(v4, v6 string) udnv1.DualStackCIDRs {
...
+			{
+				NeighborsIPv4: neighborsIPv4,
+				NetworksIPv4:  networksIPv4,
+				NeighborsIPv6: neighborsIPv6,
+				NetworksIPv6:  networksIPv6,
+			},
+		},
+	}
+
+	templates.DefinedTemplates()
```
kyrtapz: What's this for?

test/e2e/route_advertisements.go
```diff
@@ -1074,3 +1669,519 @@ func generateL2Subnets(v4, v6 string) udnv1.DualStackCIDRs {
...
+	name string,
+	cudnType bool,
+	networkSpec *udnv1.NetworkSpec,
+	networkLabels map[string]string,
+) error {
+	var gvr schema.GroupVersionResource
+	var gvk schema.GroupVersionKind
+	var obj runtime.Object
+	var client dynamic.ResourceInterface
+	switch cudnType {
```
kyrtapz: why not use an `if cudnType`? :sweat_smile:

kyrtapz:
> Fixing this

@coderabbitai good bot!

kyrtapz (commented):
Overall coverage and logic looks good to me but I have some concerns about the utils added that are already present elsewhere or could be repurposed. 
Love the commit split!

go-controller/pkg/clustermanager/routeadvertisements/controller.go
jcaamano: Let's say for example that you have an Accepted RouteAdvertisements. And that then you delete the FRRConfiguration that was being selected. The RouteAdvertisements would change to Not Accepted complaining about it, but otherwise it keeps whatever configuration that was applied in place. Let's say that you add that FRRConfiguration back. There is no different config to apply but the status would not be set back to Accepted without this fix.

test/e2e/infraprovider/providers/kind/kind.go
jcaamano: It is because these booleans are tied to the `--delete-namespace-on-failure` or the `--delete-namespace` arguments to `go test` which is how you control if you want to leave things be after the test case runs to be able to troubleshoot. 

Don't understand what you mean when you say that framework naemspace deletion is not related to debugging, it is, isn't it?

test/e2e/infraprovider/providers/kind/kind.go
jcaamano: > Why not `%q`?

No particular reason, I guess I could try that.

> Can you also share what `with index` changes here?

It's equivalent but if the network name has a hyphen, the previous template doesn't work.

test/e2e/testdata/routeadvertisements/frr/daemons.tmpl
jcaamano: I am not sure I understand.

test/e2e/util.go
jcaamano: It's not the same signature. Also I don't think we should promote the use of non testing ovnk code.

test/e2e/util.go
jcaamano: I understand. I strongly didn't like the existing one. I can see where porting the old uses to the new one takes me if you are up for that.

test/e2e/route_advertisements.go
jcaamano: I did consider it. I even considered refactoring the existing tests so it would only be a matter of adding mine as new table entries. But it was a lot of work.

I am sure neither this function nor the one I added constitutes the only code testing connectivity. 

When I see the function defined within a test, I understand it has not been coded with the mindset of it being reused in other tests. I particularly don't like how `clientNamespace` is used as a magic value to determine what you are testing against.

So that sets me up for a refactor changing all its current uses. For a function that really doesn't do much I am not sure it is worth it.

test/e2e/route_advertisements.go
jcaamano: > What's the point of defining this here instead of adding the entry directly?

Readibility I guess? It is also among the suggested patterns in ginkgo docs: 
https://onsi.github.io/ginkgo/#table-specs-patterns

> So Layer 2 doesn't work even in shared GW now?

VRF-Lite, thus also these tests, are for LGW mode only.

test/e2e/route_advertisements.go
jcaamano: Attaching the interface to the VRF removes its IPv6 address so I set it back. As to why that happens, I don't know :/

test/e2e/route_advertisements.go
jcaamano: These test are for LGW mode and we don't support that advertised topology.

test/e2e/route_advertisements.go
jcaamano: Yeah, I can define a variable for it. But why randomize it? The server is listening in 8080, you want to randomize that too? What purpose?

test/e2e/route_advertisements.go
jcaamano: That slipped, will remove.

test/e2e/route_advertisements.go
jcaamano: I guess I am a fan(atic) of switch statement and their readability vs if/else but I really meant to do 
```
switch {
case cudnType:
default:
}
```

test/e2e/route_advertisements.go
jcaamano: In general If I don't think it is likely a test will be reused I don't define a wrapper for it. But for this one you are highlighting there is already a wrapper so I should use it.

test/e2e/route_advertisements.go
```diff
@@ -1074,3 +1664,518 @@ func generateL2Subnets(v4, v6 string) udnv1.DualStackCIDRs {
...
+	Routers []templateInputRouter
+}
+
+// for routeadvertisements test cases we generate configuration from templates embed in the program
+//
+//go:embed testdata/routeadvertisements
+var ratestdata embed.FS
+var tmplDir = filepath.Join("testdata", "routeadvertisements")
+
+const frrImage = "quay.io/frrouting/frr:9.1.3"
```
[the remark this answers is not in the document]
jcaamano: We are not ready to update to frr 10 yet.

test/e2e/util.go
jcaamano: Refactored the function for more core re-usability on its own commit.

test/e2e/infraprovider/providers/kind/kind.go
jcaamano: To describe what I would find as a problem about this for what is supposed to be an agnostic platform implementation, is the dependency towards `k8s.io/kubernetes/test/e2e/framework"` which was already there.
I just checked and is just for logs so it should be trivial to remove. So decided to remove this to not make it any more harder.

jcaamano:
```
[e2e (bgp, noHA, local, dualstack, snatGW, 1br, ic-single-node-zones, advertise-default, enable-ne...](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16414579590/job/46379411894?pr=5348#logs)
succeeded 6 minutes ago in 3h 31m 1s
...
Run Tests                                                                  3 23m 3s
```

My bump to 200m is probably too tight. I am going to bump it 10m more.

Otherwise, for now we should consider trimming the testing matrix:
```
					otherNetworksToTest := []ginkgo.TableEntry{
						ginkgo.Entry("Default", defaultNetwork, nil),
						ginkgo.Entry("Layer 3 UDN non advertised", udn, otherLayer3NetworkSpec),
						ginkgo.Entry("Layer 3 CUDN advertised", cudnAdvertised, otherLayer3NetworkSpec),
						ginkgo.Entry("Layer 3 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer3NetworkSpec),
						ginkgo.Entry("Layer 2 UDN non advertised", udn, otherLayer2NetworkSpec),
						ginkgo.Entry("Layer 2 CUDN advertised", cudnAdvertised, otherLayer2NetworkSpec),
						ginkgo.Entry("Layer 2 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer2NetworkSpec),
					}
```

We could potentially go for
```
					otherNetworksToTest := []ginkgo.TableEntry{
						ginkgo.Entry("Default", defaultNetwork, nil),
						ginkgo.Entry("Layer 3 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer3NetworkSpec),
						ginkgo.Entry("Layer 2 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer2NetworkSpec),
					}
```

I could tag the other combinations as `EXTENDED` so that we can still call them up if needed.

test/e2e/testdata/routeadvertisements/frr/daemons.tmpl
kyrtapz: will this config work for ipv6 single-stack clusters?

test/e2e/route_advertisements.go
kyrtapz: If you think this improves readability I am fine to keep it, I don't think this is something we are doing anywhere else though so it is an exception.

test/e2e/route_advertisements.go
kyrtapz: I don't like hardcoded/common ports in tests because it is harder to look through logs when something fails and it there is always a higher chance of conflicts with something that exists.
I don't mind keeping it as is here.

kyrtapz:
> ```
> [e2e (bgp, noHA, local, dualstack, snatGW, 1br, ic-single-node-zones, advertise-default, enable-ne...](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16414579590/job/46379411894?pr=5348#logs)
> succeeded 6 minutes ago in 3h 31m 1s
> ...
> Run Tests                                                                  3 23m 3s
> ```
> 
> My bump to 200m is probably too tight. I am going to bump it 10m more.
> 
> Otherwise, for now we should consider trimming the testing matrix:
> 
> ```
> 					otherNetworksToTest := []ginkgo.TableEntry{
> 						ginkgo.Entry("Default", defaultNetwork, nil),
> 						ginkgo.Entry("Layer 3 UDN non advertised", udn, otherLayer3NetworkSpec),
> 						ginkgo.Entry("Layer 3 CUDN advertised", cudnAdvertised, otherLayer3NetworkSpec),
> 						ginkgo.Entry("Layer 3 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer3NetworkSpec),
> 						ginkgo.Entry("Layer 2 UDN non advertised", udn, otherLayer2NetworkSpec),
> 						ginkgo.Entry("Layer 2 CUDN advertised", cudnAdvertised, otherLayer2NetworkSpec),
> 						ginkgo.Entry("Layer 2 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer2NetworkSpec),
> 					}
> ```
> 
> We could potentially go for
> 
> ```
> 					otherNetworksToTest := []ginkgo.TableEntry{
> 						ginkgo.Entry("Default", defaultNetwork, nil),
> 						ginkgo.Entry("Layer 3 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer3NetworkSpec),
> 						ginkgo.Entry("Layer 2 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer2NetworkSpec),
> 					}
> ```
> 
> I could tag the other combinations as `EXTENDED` so that we can still call them up if needed.

+1 to limiting `otherNetworksToTest`. The ones proposed are the most complex and we cannot realistically test everything.
There is a high chance we are going to add more tests to the BGP lanes too.

.github/workflows/test.yml
```diff
@@ -422,8 +422,8 @@ jobs:
   e2e:
     name: e2e
     runs-on: ubuntu-24.04
-    # 30 mins for kind, 180 mins for control-plane tests, 10 minutes for all other steps
-    timeout-minutes: 220
+    # 30 mins for kind, 220 mins for control-plane tests, 10 minutes for all other steps
```
kyrtapz: NIT: 30+220+10 != 270

kyrtapz (approved):
LGTM

test/e2e/route_advertisements.go
jcaamano: There is always a first time for everything ;) 
Just a different way of writing code in any case, I am not aware of us having style rules around this.

jcaamano:
> +1 to limiting `otherNetworksToTest`. The ones proposed are the most complex and we cannot realistically test everything.
> There is a high chance we are going to add more tests to the BGP lanes too.

Doing this then. Will adjust timeouts after checking the run time.

jcaamano:
> > +1 to limiting `otherNetworksToTest`. The ones proposed are the most complex and we cannot realistically test everything.
> > There is a high chance we are going to add more tests to the BGP lanes too.
> 
> Doing this then. Will adjust timeouts after checking the run time.

96 tests less executed, ~1h runtime less

test/e2e/testdata/routeadvertisements/frr/daemons.tmpl
jcaamano: I wouldn't be concerned about this for the time being. It is what frr-k8s has hardcoded everywhere and it seems to work when deployed on single stack ipv6 or at least we haven't been hitting any issue related to this that I am aware.. Also we are not deploying single stack IPv6 docker networks and last I checked not long ago this wasn't possible, might have changed.

test/scripts/e2e-cp.sh
```diff
@@ -138,6 +146,11 @@ if [ "$ENABLE_ROUTE_ADVERTISEMENTS" != true ]; then
   skip $BGP_TESTS
 else
   if [ "$ADVERTISE_DEFAULT_NETWORK" = true ]; then
+    # Filter out extended RouteAdvertisements tests to keep job run time down
+    if [ "$ENABLE_NETWORK_SEGMENTATION" = true ]; then
+      skip_label "Feature:RouteAdvertisements && EXTENDED"
```
kyrtapz, line 151: Why are EXTENDED skipped only for BGP? Shouldn't this be universal?
I know that this is only used for BGP e2es right now but I don't think this is specific to the feature in general.
We can address it as a followup.

test/e2e/route_advertisements.go
```diff
@@ -1545,12 +1547,14 @@ var _ = ginkgo.Describe("BGP: For a VRF-Lite configured network", func() {
...
 						ginkgo.Entry("Default", defaultNetwork, nil),
-						ginkgo.Entry("Layer 3 UDN non advertised", udn, otherLayer3NetworkSpec),
-						ginkgo.Entry("Layer 3 CUDN advertised", cudnAdvertised, otherLayer3NetworkSpec),
 						ginkgo.Entry("Layer 3 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer3NetworkSpec),
-						ginkgo.Entry("Layer 2 UDN non advertised", udn, otherLayer2NetworkSpec),
-						ginkgo.Entry("Layer 2 CUDN advertised", cudnAdvertised, otherLayer2NetworkSpec),
 						ginkgo.Entry("Layer 2 CUDN advertised VRF-Lite", cudnAdvertisedVRFLite, otherLayer2NetworkSpec),
+						// The following testcases are labeled as extended,
+						// might not be run on all jobs
+						ginkgo.Entry("Layer 3 UDN non advertised", udn, otherLayer3NetworkSpec, label.Extended()),
```
trozet, line 1554: It seems like this test doesn't make any sense, as I don't see how a non-advertised layer 3 network would  test anything for VRF-Lite.

trozet (approved):
the last commit doesn't make a lot of sense to me. A better commit message would have helped I think. The PR overall has a lot of good stuff in (more tests are great!) so will merge it. I think we should probably follow up with this extended thing and figure out what the purpose of it is and why we are skipping tests in CI.

tssurya:
@trozet got to this before me :) all good
thanks @jcaamano for adding these tests

test/e2e/route_advertisements.go
jcaamano: These are isolation tests between a VRF-Lite network and a Layer 3 UDN non advertised network.

jcaamano:
> the last commit doesn't make a lot of sense to me. A better commit message would have helped I think. The PR overall has a lot of good stuff in (more tests are great!) so will merge it. I think we should probably follow up with this extended thing and figure out what the purpose of it is and why we are skipping tests in CI.

Sorry for that. I had a commit message but on the way I rebased and edited the commit and forgot to add it back. The reason is testing all combinations in the matrix increases the job runtime >3h. Patryk and I thought that testing all variations was overkill and decided to trim down the tests but still have them somewhat available. We used a label for this.

PR #5351, opened 2025-07-04
Enable Layer2 route advertisements on LGW

kyrtapz:
Configure use_ct_inv_match=false on nbdb when route advertisements are enabled in LGW to support assymetric traffic on Layer2 advertised networks.


dist/images/ovnkube.sh
```diff
@@ -987,6 +987,11 @@ local-nb-ovsdb() {
   wait_for_event attempts=3 process_ready ovnnb_db
   echo "=============== nb-ovsdb (unix sockets only) ========== RUNNING"
 
+ [[ "local" == "${OVN_GATEWAY_MODE}" &&  "true" == "${OVN_ROUTE_ADVERTISEMENTS_ENABLE}" ]] && {
+    ovn-nbctl set NB_Global . options:use_ct_inv_match=false
+    echo "=============== nb-ovsdb ========== reconfigured for route advertisements"
+  }
```
maiqueb, line 995: what will this look like in OpenShift ? I assume CNO will have to set something similar on the OVS nodes of the cluster. 

I do wonder if it makes sense to expose this global OVS table option via nmstate,  and have it configurable via NNCP (nmstate policy).

kyrtapz: There will be a CNO change yeah, this is not really something that matters for upstream though.
Here is the pr: https://github.com/openshift/cluster-network-operator/pull/2735

> I do wonder if it makes sense to expose this global OVS table option via nmstate, and have it configurable via NNCP (nmstate policy).

This is an OVN level setting which is applied in the container running nbdb, not on the host.

kyrtapz:
The `UDN pod to a different node should work` test fails https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16200112782/job/45738981781?pr=5351 because it expects both ipv6 and ipv4 connectivity to fail but v6 works.
That is because with my PR we allow for assymetric traffic and the reply comes back through the mpX interface as there is no rp filtering set for ipv6.

test/e2e/route_advertisements.go
```diff
@@ -521,6 +516,10 @@ var _ = ginkgo.Describe("BGP: Pod to external server when CUDN network is advert
 var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks",
 	func(cudnATemplate, cudnBTemplate *udnv1.ClusterUserDefinedNetwork) {
 		const curlConnectionTimeoutCode = "28"
+		const (
+			ipFamilyV4 = iota
+			ipFamilyV6 = 1
```
trozet: you dont need to put ipFamilyV6 =1, just put ipFamilyV6 so it automatically enumerates one higher than iota

test/e2e/route_advertisements.go
```diff
@@ -963,14 +965,20 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
-						// rp_filter is set to 1 in breth0 and there is an iprule that sends the packet to mpX interface so kernel sees the packet
-						// having return path different from the incoming interface.
-						// The SNAT to nodeIP should fix this.
-						// this causes curl timeout with code 28
-						errBool = true
-						out = curlConnectionTimeoutCode
+						// FIXME: this should be removed once we add the SNAT for pod->node traffic
+						// We now permit asymmetric traffic on LGW. This prevents the issue from occurring with IPv6.
+						// However, for IPv4 LGW rp_filter is still blocking the replies.
+						// The situation is different on SGW as we don't allow asymmetric traffic at all, which is why IPv6 traffic fails there too.
```
trozet, line 953: why do we allow loose mode for ipv6 on rp_filter but not ipv4?

dist/images/ovnkube.sh
```diff
@@ -989,6 +989,11 @@ local-nb-ovsdb() {
   wait_for_event attempts=3 process_ready ovnnb_db
   echo "=============== nb-ovsdb (unix sockets only) ========== RUNNING"
 
+ [[ "local" == "${OVN_GATEWAY_MODE}" &&  "true" == "${OVN_ROUTE_ADVERTISEMENTS_ENABLE}" ]] && {
+    ovn-nbctl set NB_Global . options:use_ct_inv_match=false
```
trozet, line 993: note: this is not hw offloadable, but lucky for us this is local gw mode only change so all good

test/e2e/route_advertisements.go
```diff
@@ -963,14 +965,20 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+						if ipFamilyIndex == ipFamilyV4 || !isLocalGWModeEnabled() {
+							// FIXME: fix assymmetry in L2 UDNs
```
tssurya, line 955: I think we are going to have conflicts for sure hehe.. I hope LGW with L2 doesn't give me any surprises for https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140/commits/9e2923971a6ab970d31d88aba1dfa057cc377092#diff-00535718a80e537658f346b0cbae8ced845dcc0cccc9cf79bd9fe610d09a6497R966

it all gets fixed here ^.. so between these two PRs I think we need to decide which PR should land first.

dist/images/ovnkube.sh
jcaamano: I think while talking with @kyrtapz last week we realized we might need this in SGW mode as well, but we missed it because it was masqueraded with the rp_filter issue. I will let him confirm though.

test/e2e/route_advertisements.go
kyrtapz: Lol, of course. My IDE was trying to be helpful, fixed.

test/e2e/route_advertisements.go
kyrtapz: We don't allow it, we do nothing. For IPv4 the default value is "strict mode" and that's why it doesn't work there.
This is all going to change with the SNATing that @tssurya is working on so I don't think we should be changing it here.

dist/images/ovnkube.sh
kyrtapz: No, we don't want to do it on SGW since @tssurya  is addressing this issue by using SNAT to node IP for these specific traffic flows.

test/e2e/route_advertisements.go
```diff
@@ -906,7 +899,16 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
-					return podsNetA[0].Name, podsNetA[0].Namespace, net.JoinHostPort(svcNetB.Spec.ClusterIPs[ipFamilyIndex], "8080") + "/clientip", curlConnectionTimeoutCode, true
+					err := true
+					out := curlConnectionTimeoutCode
+					if cudnATemplate.Spec.Network.Topology == udnv1.NetworkTopologyLayer2 && isLocalGWModeEnabled() {
+						// FIXME: prevent looping of traffic in L2 UDNs
+						// bad behaviour: packet is looping from management port -> breth0 -> GR -> management port -> breth0 and so on
+						// which is a never ending loop
+						// this causes curl timeout with code 7 host unreachable instead of code 28
+						out = ""
+					}
```
trozet: this seems like something a user could exploit to attack the dataplane. What is the plan to address this?

trozet (commented):
@kyrtapz I'd like to see a more detailed explanation in the PR description. IIRC the invalid CT state dropping packets is specifically for the mpx LSP CT zone. Is that correct? That's what we are targeting here right? established packets going to another node and entering ovn via an mpx that has no CT state.

Invalid CT state packets are not dropped in routers.

test/e2e/route_advertisements.go
kyrtapz: This issue is limited to L2 advertised pods in LGW.
I think one potential solution would be to add ACLs that do not allow traffic to leave through mpX if the DST is a service IP(with the exception of kapi/dns).
I can create an issue for it if that works for you.

kyrtapz: @jcaamano pointed out offline that we might also consider modifying the flows we originally added here https://github.com/ovn-kubernetes/ovn-kubernetes/commit/d63887ed167da260d3f26c71ec06e520d89a4b0f:
```
table=0,priority=550,ip,in_port=LOCAL,nw_src=<UDN_SUBNET>,nw_dst=<SVC_SUBNET>,actions=ct(commit,table=2,zone=64001)
table=2,priority=200,ip,nw_src=<UDN_SUBNET> actions=...,output:"<UDN_PATCH_PORT>"
```
so that instead of sending the traffic to the patch port we just drop it.
We will need to ensure it doesn't break anything else.

trozet: can you provide a better description of the traffic flow here? A packet destined to a cluster IP service on the same UDN should be DNAT'ed on pod egress to the switch. If it's a different-UDN service, it would get rerouted out mpx ->breth0 -> other UDN GR -> DNAT to pod IP... what am I missing?

kyrtapz: It's like this:
udn-mpX->breth0->udn-GR(see the flows above)->udn-mpX(caused by this route: `10.10.0.0/16 10.10.0.2 src-ip`->....

So the traffic is looping in the same UDN.
Because of the flows mentioned above this will never get forwarded to other networks.

kyrtapz: Trying the approach of dropping directly here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5386/commits/a8dd3c6ebb1cbcddecd1d19e0524e4836e56a046

trozet: I would prefer to see that fix included in this PR before this is merged.

kyrtapz: Done.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -406,9 +406,9 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 
 	// Add flows for default network services that are accessible from UDN networks
 	if util.IsNetworkSegmentationSupportEnabled() {
-		// The flow added below has a higher priority than the per UDN service flow:
-		//   priority=200, table=2, ip, ip_src=169.254.0.<UDN>, actions=set_field:<bridge-mac>->eth_dst,output:<UDN-patch-port>
-		// This ordering ensures that traffic to UDN allowed default services goes to the the default patch port.
+		// The flow added below has a higher priority than the per UDN service isolation flow:
+		//   priority=200, table=2, ip, ip_src=169.254.0.<UDN>, actions=drop
+		// This ordering ensures that traffic to UDN allowed default services goes to the default patch port.
```
trozet, line 411: note to self: this change is OK because:
1. the original behavior implemented by @dceara  was to match on masquerade IP from the host and then dispatch the packet to the right patch port
2. This behavior was modified by @kyrtapz to set pkt marks on service packets and then use another flow to dispatch them here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4792

So now these masq flows should not be used by anything and we should drop the traffic, rather than letting it hit the next rule which would forward to CDN patch port.

trozet (approved):

go-controller/pkg/node/gateway_shared_intf.go
kyrtapz: Agreed. 

I think the original change had some potential to be used for host-networked UDN pod -> UDN service but as I as said in the commit message this is not there yet and we can reuse the marking approach.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1788,13 +1788,17 @@ func flowsForDefaultBridge(bridge *bridgeConfiguration, extraIPs []net.IP) ([]st
...
 				fmt.Sprintf("cookie=%s, priority=200, table=2, ip, ip_src=%s, "+
-					"actions=set_field:%s->eth_dst,output:%s",
-					defaultOpenFlowCookie, srcIPOrSubnet,
-					bridgeMacAddress, netConfig.ofPortPatch))
+					"actions=drop",
+					defaultOpenFlowCookie, srcIPOrSubnet))
+
 			dftFlows = append(dftFlows,
-				fmt.Sprintf("cookie=%s, priority=200, table=2, ip, pkt_mark=%s, "+
+				fmt.Sprintf("cookie=%s, priority=250, table=2, ip, pkt_mark=%s, "+
```
tssurya, line 1801: this change is conflicting with my changes here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140/commits/8cd1414725b62ee5bc859452eb0efb9bccd374ed
why are we fixing looping issue in this PR and not as a FUP... I actually neded to lower the priority of this rule and here its been increased?

PR #5356, opened 2025-07-07
persistentIPs: Deprecate primary UDN IPAMClaim annotation - 'k8s.ovn.org/primary-udn-ipamclaim'

ormergi:
## 📑 Description

This PR implements the part of deprecating the UDN IPAMClaim annotation in favor of the default NSE extension for IPAMClaim https://github.com/kyrtapz/ovn-kubernetes/blob/master/docs/okeps/okep-5233-preconfigured-udn-addresses.md#pod-network-identity

The NetworkSelectionElement (NSE) API has been extended and features IPAM claim
reference attribute enabling associating pod with IPAMClaim CR.

Deprecate the UDN IPAMClaim annotation - `k8s.ovn.org/primary-udn-ipamclaim`.

The defaullt NSE `v1.multus-cni.io/default-network` should be used on
order to associate pod with IPAM claim CR.

When default NSE doesn't specify IPAMClaim reference, falling back to
UDN IPAMClaim annotation

In next release the UDN IPAMClaim annotation can be removed altogether.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Errors resolving a pod's default-network selection are now propagated earlier; persistent pod IP handling is more reliable.

* **Refactor**
  * Primary-network persistent-IP selection now prefers the default-network annotation (ipam-claim-reference) and falls back to the legacy annotation for compatibility; default-network static IP/MAC values can override active network requests.

* **Tests**
  * Expanded coverage for annotation precedence, default-network sourcing, primary-network behaviors, and legacy-annotation interactions.

* **Chores**
  * Legacy primary-IP annotation marked deprecated; guidance moved to using the standard default-network annotation.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_user_defined.go
+++ b/go-controller/pkg/ovn/base_network_controller_user_defined.go
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
 	var ipamClaimName string
 	var wasPersistentIPRequested bool
 	if bsnc.IsPrimaryNetwork() {
-		// primary network ipam reference claim is on the annotation
-		ipamClaimName, wasPersistentIPRequested = pod.Annotations[util.OvnUDNIPAMClaimName]
+		// 'k8s.ovn.org/primary-udn-ipamclaim' annotation has been deprecated. Maintain backward compatibility by
+		// using it as a fallback; when defaultNSE.IPAMClaimReference is set, it takes precedence.
+		if desiredClaimName, isIPAMClaimRequested := pod.Annotations[util.DeprecatedOvnUDNIPAMClaimName]; isIPAMClaimRequested && desiredClaimName != "" {
+			wasPersistentIPRequested = true
+			ipamClaimName = desiredClaimName
+		}
+		defaultNSE, err := util.GetK8sPodDefaultNetworkSelection(pod)
+		if err != nil {
+			return false, err
+		}
+		if defaultNSE != nil && defaultNSE.IPAMClaimReference != "" {
+			wasPersistentIPRequested = true
+			ipamClaimName = defaultNSE.IPAMClaimReference
+		}
 	} else {
 		// secondary network the IPAM claim reference is on the network selection element
 		nadKeys := strings.Split(nadNamespacedName, "/")
--- a/go-controller/pkg/ovn/multihoming_test.go
+++ b/go-controller/pkg/ovn/multihoming_test.go
@@ -471,7 +471,8 @@ func newMultiHomedPod(testPod testPod, multiHomingConfigs ...userDefinedNetInfo)
 	for _, multiHomingConf := range multiHomingConfigs {
 		if multiHomingConf.isPrimary {
 			if multiHomingConf.ipamClaimReference != "" {
-				pod.Annotations[util.OvnUDNIPAMClaimName] = multiHomingConf.ipamClaimReference
+				pod.Annotations[util.DeprecatedOvnUDNIPAMClaimName] = multiHomingConf.ipamClaimReference
+				pod.Annotations[util.DefNetworkAnnotation] = generateDefaultNSEAnnotation(multiHomingConf)
 			}
 			continue // these will be automatically plugged in
 		}
@@ -498,6 +499,22 @@ func newMultiHomedPod(testPod testPod, multiHomingConfigs ...userDefinedNetInfo)
 		}
 	}
 	return pod
+}
+
+func generateDefaultNSEAnnotation(multiHomingConf userDefinedNetInfo) string {
+	nadNamePair := strings.Split(multiHomingConf.nadName, "/")
+	if len(nadNamePair) != 2 {
+		panic("failed to generate default NSE: invalid NAD name. Expected NAD name format: '<namespace>/<name>'")
+	}
+	bytes, err := json.Marshal([]nadapi.NetworkSelectionElement{{
+		Namespace:          nadNamePair[0],
+		Name:               nadNamePair[1],
+		IPAMClaimReference: multiHomingConf.ipamClaimReference,
+	}})
+	if err != nil {
+		panic(fmt.Errorf("failed to generate default NSE: %v", err))
+	}
+	return string(bytes)
 }
 
 func dummyOVNPodNetworkAnnotations(secondaryPodInfos map[string]*udnPodInfo, multiHomingConfigs []userDefinedNetInfo) string {
--- a/go-controller/pkg/util/multi_network.go
+++ b/go-controller/pkg/util/multi_network.go
@@ -1543,7 +1543,6 @@ func overrideActiveNSEWithDefaultNSE(defaultNSE, activeNSE *nettypes.NetworkSele
 	}
 	activeNSE.IPRequest = defaultNSE.IPRequest
 	activeNSE.MacRequest = defaultNSE.MacRequest
-	activeNSE.IPAMClaimReference = defaultNSE.IPAMClaimReference
 	return nil
 }
 
@@ -1586,29 +1585,38 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo, 
 		Name:      activeNADKey.Name,
 	}
 
+	isPersistentIPsPrimaryNetwork := nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo)
+	var defaultNSE *nettypes.NetworkSelectionElement
+	if isPersistentIPsPrimaryNetwork || IsPreconfiguredUDNAddressesEnabled() {
+		defaultNSE, err = GetK8sPodDefaultNetworkSelection(pod)
+		if err != nil {
+			return false, nil, fmt.Errorf("failed getting default-network annotation for pod %q: %w", pod.Namespace+"/"+pod.Name, err)
+		}
+	}
+
+	if isPersistentIPsPrimaryNetwork {
+		// 'k8s.ovn.org/primary-udn-ipamclaim' annotation has been deprecated. Maintain backward compatibility by
+		// using it as a fallback; when defaultNSE.IPAMClaimReference is set, it takes precedence.
+		if ipamClaimName, wasPersistentIPRequested := pod.Annotations[DeprecatedOvnUDNIPAMClaimName]; wasPersistentIPRequested {
+			activeNSE.IPAMClaimReference = ipamClaimName
+		}
+		if defaultNSE != nil && defaultNSE.IPAMClaimReference != "" {
+			activeNSE.IPAMClaimReference = defaultNSE.IPAMClaimReference
+		}
+	}
+
 	// Feature gate integration: EnablePreconfiguredUDNAddresses controls default network IP/MAC transfer to active network
 	if IsPreconfiguredUDNAddressesEnabled() {
 		// Limit the static ip and mac requests to the layer2 primary UDN when EnablePreconfiguredUDNAddresses is enabled, we
 		// don't need to explicitly check this is primary UDN since
 		// the "active network" concept is exactly that.
 		if activeNetwork.TopologyType() == types.Layer2Topology {
-			defaultNSE, err := GetK8sPodDefaultNetworkSelection(pod)
-			if err != nil {
-				return false, nil, fmt.Errorf("failed getting default-network annotation for pod %q: %w", pod.Namespace+"/"+pod.Name, err)
-			}
 			// If there are static IPs and MACs at the default NSE, override the active NSE with them
 			if defaultNSE != nil {
 				if err := overrideActiveNSEWithDefaultNSE(defaultNSE, activeNSE); err != nil {
 					return false, nil, err
 				}
 			}
-		}
-	}
-
-	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) && activeNSE.IPAMClaimReference == "" {
-		ipamClaimName, wasPersistentIPRequested := pod.Annotations[OvnUDNIPAMClaimName]
-		if wasPersistentIPRequested {
-			activeNSE.IPAMClaimReference = ipamClaimName
 		}
 	}
 
--- a/go-controller/pkg/util/multi_network_test.go
+++ b/go-controller/pkg/util/multi_network_test.go
@@ -1107,7 +1107,103 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
 			},
 			inputPodAnnotations: map[string]string{
 				nadv1.NetworkAttachmentAnnot: GetNADName(namespaceName, "another-network"),
-				OvnUDNIPAMClaimName:          "the-one-to-the-left-of-the-pony",
+				DefNetworkAnnotation:         `[{"ipam-claim-reference":"the-one-to-the-left-of-the-pony","namespace":"ns1","name":"attachment1"}]`,
+			},
+			expectedIsAttachmentRequested: true,
+			expectedNetworkSelectionElements: map[string]*nadv1.NetworkSelectionElement{
+				"ns1/attachment1": {
+					Name:               "attachment1",
+					Namespace:          "ns1",
+					IPAMClaimReference: "the-one-to-the-left-of-the-pony",
+				},
+			},
+		},
+		{
+			desc: "the network configuration for a primary layer2 UDN features allow persistent IPs, and the pod requests it." +
+				"Using deprecated UDN IPAMClaim annotation",
+			// verify backward compatibility for deprecated annotation 'k8s.ovn.org/primary-udn-ipamclaim'
+			inputNetConf: &ovncnitypes.NetConf{
+				NetConf:            cnitypes.NetConf{Name: networkName},
+				Topology:           ovntypes.Layer2Topology,
+				NADName:            GetNADName(namespaceName, attachmentName),
+				Role:               ovntypes.NetworkRolePrimary,
+				AllowPersistentIPs: true,
+			},
+			inputPrimaryUDNConfig: &ovncnitypes.NetConf{
+				NetConf:            cnitypes.NetConf{Name: networkName},
+				Topology:           ovntypes.Layer2Topology,
+				NADName:            GetNADName(namespaceName, attachmentName),
+				Role:               ovntypes.NetworkRolePrimary,
+				AllowPersistentIPs: true,
+			},
+			inputPodAnnotations: map[string]string{
+				nadv1.NetworkAttachmentAnnot:  GetNADName(namespaceName, "another-network"),
+				DeprecatedOvnUDNIPAMClaimName: "the-one-to-the-left-of-the-pony",
+			},
+			expectedIsAttachmentRequested: true,
+			expectedNetworkSelectionElements: map[string]*nadv1.NetworkSelectionElement{
+				"ns1/attachment1": {
+					Name:               "attachment1",
+					Namespace:          "ns1",
+					IPAMClaimReference: "the-one-to-the-left-of-the-pony",
+				},
+			},
+		},
+		{
+			desc: "the network configuration for a primary layer2 UDN features allow persistent IPs, and the pod requests it." +
+				"Pod has both defaultNSE with ipam-claim reference and UDNIPAMClaim annotations, specifying the same IPAMClaim CR",
+			// verify backward compatibility for deprecated annotation 'k8s.ovn.org/primary-udn-ipamclaim'
+			inputNetConf: &ovncnitypes.NetConf{
+				NetConf:            cnitypes.NetConf{Name: networkName},
+				Topology:           ovntypes.Layer2Topology,
+				NADName:            GetNADName(namespaceName, attachmentName),
+				Role:               ovntypes.NetworkRolePrimary,
+				AllowPersistentIPs: true,
+			},
+			inputPrimaryUDNConfig: &ovncnitypes.NetConf{
+				NetConf:            cnitypes.NetConf{Name: networkName},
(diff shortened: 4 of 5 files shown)


go-controller/pkg/ovn/multihoming_test.go
```diff
@@ -473,7 +473,8 @@ func newMultiHomedPod(testPod testPod, multiHomingConfigs ...userDefinedNetInfo)
 	for _, multiHomingConf := range multiHomingConfigs {
 		if multiHomingConf.isPrimary {
 			if multiHomingConf.ipamClaimReference != "" {
-				pod.Annotations[util.OvnUDNIPAMClaimName] = multiHomingConf.ipamClaimReference
+				pod.Annotations[util.DeprecatedOvnUDNIPAMClaimName] = multiHomingConf.ipamClaimReference
+				pod.Annotations[util.DefNetworkAnnotation] = generateDefaultNSEAnnotation(multiHomingConf)
 			}
```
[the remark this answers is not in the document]
ormergi: Done

go-controller/pkg/util/multi_network.go
```diff
@@ -1511,16 +1510,20 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
 	}
 
+	defaultNSE, err := GetK8sPodDefaultNetworkSelection(pod)
+	if err != nil {
+		return false, nil, fmt.Errorf("failed getting default-network annotation for pod %q: %w", pod.Namespace+"/"+pod.Name, err)
+	}
+	if defaultNSE != nil {
+		activeNSE.IPAMClaimReference = defaultNSE.IPAMClaimReference
+	}
+
```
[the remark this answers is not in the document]
ormergi: We cant do that because default NSE can point to some other udn that serves as the default network

ormergi:
@qinqon / @maiqueb PTAL

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
...
+		if err != nil {
+			return false, err
+		}
+		if defaultNSE != nil && defaultNSE.IPAMClaimReference != "" {
+			wasPersistentIPRequested = true
+			ipamClaimName = defaultNSE.IPAMClaimReference
+		}
+		// 'k8s.ovn.org/primary-udn-ipamclaim' annotation has been deprecated. Maintain backward compatibility by
+		// respecting 'k8s.ovn.org/primary-udn-ipamclaim' value if present; it overrides the defaultNSE value.
+		if desiredClaimName, ipamClaimRequested := pod.Annotations[util.DeprecatedOvnUDNIPAMClaimName]; ipamClaimRequested && desiredClaimName != "" {
```
kyrtapz: I don't think the order is correct. I think that the new way of requesting the IPAMClaims should take precedence and if it is not set then we can read the deprecated annotation.

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
...
+			wasPersistentIPRequested = true
+			ipamClaimName = desiredClaimName
```
maiqueb: I **wonder** if we want to overwrite the values here ...

Maybe the most sensible approach is to fail if these are different ?... 

While saying that, I understand that for this to be backwards compatible, it needs to consume whatever's in the "old" annotation. 

Still, let's make the experiment of thinking this through: 
- let's say we have a running pod w/ the default net annotation + a primary UDN ipam claim requested. 
- that would "stop" working if start failing if the annotations have different values

I guess it all boils down to if having a launcher pod w/ the default net annotation + a primary UDN ipam claim requested is possibly. 

I'm inclined to say your current proposal is preferable (I'm just thinking out loud, trying to see some alternatives / options).

go-controller/pkg/ovn/multihoming_test.go
```diff
@@ -502,6 +503,22 @@ func newMultiHomedPod(testPod testPod, multiHomingConfigs ...userDefinedNetInfo)
 	return pod
 }
 
+func generateDefaultNSEAnnotation(multiHomingConf userDefinedNetInfo) string {
+	nadNamePair := strings.Split(multiHomingConf.nadName, "/")
+	if len(nadNamePair) != 2 {
```
maiqueb, line 506: we can't have the format of omitting the namespace, right ? Since the NAD must be in the ovn-k ns, right ?

go-controller/pkg/ovn/base_network_controller_user_defined.go
qinqon: Are we going to find this situation ? I mean the kubevirt ipam controller is going to re-generate the annotations after upgrade so it will set the proper one instead of the ipam claim ref specic.

maiqueb: it will, won't it ? 

I don't think we ever wrote code in ipam-ext to deprecate the primary udn annotation. 
I might be wrong though.

go-controller/pkg/ovn/multihoming_test.go
ormergi: Correct
Setting no namespace would result on pod error not finding the NAD, and we dont support spesifying a "local" NAD for the default network.

go-controller/pkg/ovn/base_network_controller_user_defined.go
ormergi: ipam-ext still support the annotation although we have plans to deprecate it in ipam-ext as well.

I think that in order to retain existing behavior we should respect the ipam claim req over default NSE even when values are different.

go-controller/pkg/ovn/multihoming_test.go
maiqueb: OK, let's move on then. Tks.

go-controller/pkg/ovn/base_network_controller_user_defined.go
maiqueb: I agree with @ormergi

qinqon: fair enough

ormergi:
> /lgtm

no prow in this repo :x

ormergi:
@kyrtapz could you please have a look? :pray:

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
 	var ipamClaimName string
 	var wasPersistentIPRequested bool
 	if bsnc.IsPrimaryNetwork() {
-		// primary network ipam reference claim is on the annotation
-		ipamClaimName, wasPersistentIPRequested = pod.Annotations[util.OvnUDNIPAMClaimName]
```
kyrtapz, line 538: Hmm, isn't this something we missed previously? I am surprised this function only read this annotation. Looks like a bug you just fixed!

go-controller/pkg/util/multi_network.go
```diff
@@ -1511,16 +1510,20 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
 		Name:      activeNADKey.Name,
 	}
 
+	defaultNSE, err := GetK8sPodDefaultNetworkSelection(pod)
```
kyrtapz: Why was this moved outside of the `IsPreconfiguredUDNAddressesEnabled()` block? We don't want to respect the default nse if the feature is disabled or it is a Layer3 network.

go-controller/pkg/util/multi_network.go
```diff
@@ -1530,9 +1533,10 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
 		}
 	}
 
-	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) && activeNSE.IPAMClaimReference == "" {
-		ipamClaimName, wasPersistentIPRequested := pod.Annotations[OvnUDNIPAMClaimName]
-		if wasPersistentIPRequested {
+	if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) {
```
kyrtapz: Same as https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5356/files#r2489483609

ormergi: Done

go-controller/pkg/util/multi_network.go
ormergi: My bad, we dont really need this.
Fixed.

go-controller/pkg/ovn/base_network_controller_user_defined.go
ormergi: Done

ormergi:
[Rebase](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/5b9ce6386a6359b9084771c3004e9271cbb99420..fb3e2b2a889e2494da8851251aeb26267e0b9ac1)

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
...
+			ipamClaimName = desiredClaimName
+		}
+		defaultNSE, err := util.GetK8sPodDefaultNetworkSelection(pod)
+		if err != nil {
+			return false, err
+		}
+		if defaultNSE != nil && defaultNSE.IPAMClaimReference != "" {
+			wasPersistentIPRequested = true
+			ipamClaimName = defaultNSE.IPAMClaimReference
+		}
```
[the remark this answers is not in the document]
ormergi: Following discussion offline, we concluded default NSE should take precedence when set over the deprecated UDN IPAMClaim annotation

ormergi:
Regarding https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5356#discussion_r2518158897
> Why was this moved outside of the `IsPreconfiguredUDNAddressesEnabled()` block? We don't want to respect the default nse if the feature is disabled or it is a Layer3 network.

Looks like we need just the fetching part outside the FG scope, so the fetched default NSE can be reused later on.
The part that assign the IPAMClaim ref in the default NSE should not relay on the PreconfiguredUDNAddress FG in order to replace the deprecated annotation (which realy on the persistentIPs FG)
@kyrtapz @maiqueb WDYT?

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
 	var ipamClaimName string
 	var wasPersistentIPRequested bool
 	if bsnc.IsPrimaryNetwork() {
-		// primary network ipam reference claim is on the annotation
-		ipamClaimName, wasPersistentIPRequested = pod.Annotations[util.OvnUDNIPAMClaimName]
+		// 'k8s.ovn.org/primary-udn-ipamclaim' annotation has been deprecated. Maintain backward compatibility by
+		// using it as a fallback; when defaultNSE.IPAMClaimReference is set, it takes precedence.
+		if desiredClaimName, ipamClaimRequested := pod.Annotations[util.DeprecatedOvnUDNIPAMClaimName]; ipamClaimRequested && desiredClaimName != "" {
```
maiqueb: nit: add some prefix to `ipamClaimRequested` to hint it as a bool ?

go-controller/pkg/util/multi_network_test.go
```diff
@@ -1107,7 +1107,102 @@ func TestGetPodNADToNetworkMappingWithActiveNetwork(t *testing.T) {
...
+				NetConf:            cnitypes.NetConf{Name: networkName},
+				Topology:           ovntypes.Layer2Topology,
+				NADName:            GetNADName(namespaceName, attachmentName),
+				Role:               ovntypes.NetworkRolePrimary,
+				AllowPersistentIPs: true,
+			},
+			inputPodAnnotations: map[string]string{
+				nadv1.NetworkAttachmentAnnot:  GetNADName(namespaceName, "another-network"),
+				DefNetworkAnnotation:          `[{"ipam-claim-reference":"the-one-to-the-left-of-the-pony","namespace":"ns1","name":"attachment1"}]`,
+				DeprecatedOvnUDNIPAMClaimName: "the-one-to-the-left-of-the-pony",
```
maiqueb, line 1173: wait, shouldn't these have different values ? so we test that we're picking up the *right* annotation value ? 

I now read the test's name. I miss what is the purpose of this test.

maiqueb (approved):
tks

go-controller/pkg/util/multi_network_test.go
ormergi: This one is about a scenerio where both default NSE and UND IPAMClaim annotations exist with referencing the same value.

I fixed the test description.

go-controller/pkg/ovn/base_network_controller_user_defined.go
ormergi: Done

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/a0e19bbd2c4f7ac854861e564ad56903a9b336ef..2d853b8228db1f2dc26b09933f350f8be9dda87e): resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5356#pullrequestreview-3528963027

go-controller/pkg/util/multi_network_test.go
maiqueb: OK, I guess we're good then. Tks. 
Resolving this comment.

go-controller/pkg/util/multi_network.go
kyrtapz: This is still always ran?

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -534,8 +534,20 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
...
 	if bsnc.IsPrimaryNetwork() {
-		// primary network ipam reference claim is on the annotation
-		ipamClaimName, wasPersistentIPRequested = pod.Annotations[util.OvnUDNIPAMClaimName]
+		// 'k8s.ovn.org/primary-udn-ipamclaim' annotation has been deprecated. Maintain backward compatibility by
+		// using it as a fallback; when defaultNSE.IPAMClaimReference is set, it takes precedence.
+		if desiredClaimName, isIPAMClaimRequested := pod.Annotations[util.DeprecatedOvnUDNIPAMClaimName]; isIPAMClaimRequested && desiredClaimName != "" {
+			wasPersistentIPRequested = true
+			ipamClaimName = desiredClaimName
+		}
+		defaultNSE, err := util.GetK8sPodDefaultNetworkSelection(pod)
```
kyrtapz, line 543: We should only check it for L2 networks and when `IsPreconfiguredUDNAddressesEnabled` is true

kyrtapz (commented):
just two remaining bits and we should be good to merge here.

go-controller/pkg/ovn/base_network_controller_user_defined.go
ormergi: I think it diverge from the current behavior with the UDN IPAMClaim annotation, which is available regardless of the IsPreconfiguredUDNAddressesEnabled FG.
Default NSE should supersede the UDN IPAMClaim annotation, so it should be available for usage as before, right? (regardless of IsPreconfiguredUDNAddressesEnabled FG)

@maiqueb @kyrtapz WDYT?

go-controller/pkg/util/multi_network.go
ormergi: Yes it will always run.

Did you mean this should run at two occurrences, primary+persistentIPs (1593) and when IsPreconfiguredUDNAddressesEnabled (line 1605)?

ormergi:
@kyrtapz please see https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5356/files#r2596166029

go-controller/pkg/ovn/base_network_controller_user_defined.go
kyrtapz: Not quite - `IsPreconfiguredUDNAddressesEnabled` gates this feature differently than the legacy annotation.

I want to keep this simple: either the default network annotation supports both IPAMClaim and static IPs, or it supports neither. Supporting IPAMClaim but not static IPs creates yet another edge case that can break.

maiqueb: Wait a sec ... 

We're deprecating this annotation. Meaning, we need to preserve its functionality. Meaning, the users need to be able to use this feature (preserve IP addresses) without specifying addresses (as they do today), otherwise, we'll be breaking an existing feature.

I think being backwards compatible beats simplicity.

kyrtapz: I am not arguing that. What I am saying is that I don't want for us to parse the default nse when the feature is not enabled.

maiqueb: ah. Then I agree. Let's not. 

Sorry for having misunderstood your point.

ormergi: > I am not arguing that. What I am saying is that I don't want for us to parse the default nse when the feature is not enabled.

I pushed new changes to address that

ormergi:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/2d853b8228db1f2dc26b09933f350f8be9dda87e..c93a0e8d3588045509450dd5e6fb00ba9823ec1a): resolve https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5356#pullrequestreview-3543775161
@kyrtapz PTAL :pray:

go-controller/pkg/ovn/base_network_controller_user_defined.go
maiqueb: @kyrtapz o/

I just thought about this some more yesterday ...

I don't think we can tie parsing the default nse to the `PreconfiguredUDNAddresses` feature gate. 

Let's assume that we (in a couple releases) actually remove the "old" annotation. Meaning, the "new" annotation is the only way to instruct OVN-K it needs to persist IP addresses for given pods.

If we take the path you're suggesting, how will the user be able to configure persistent IP addresses for a primary L2 UDN with subnets ? Meaning, the user wants to have both persistent IPs on their VMs, but also wants to have OVN-K allocating the IPs for their pods. AFAIU, that would no longer be possible, thus breaking an existing feature.

ormergi:
I mistakenly pushed new changes and rebased at once, please see the relevant diff in the following link (it take few moments for Github to scroll down to the relevant hunk) https://github.com/ovn-kubernetes/ovn-kubernetes/compare/c93a0e8d3588045509450dd5e6fb00ba9823ec1a..fc147e2b60073bf141b0f14f33c841a02ce01c38#diff-204cead8f2f1e61a6569211a8c762e70d43a88b06137e1a3cb2eae7ae0a57786L1588-L1613

ormergi:
@kyrtapz I resolved last comment [diff](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5356#issuecomment-3665930133), PTAL

go-controller/pkg/ovn/base_network_controller_user_defined.go
kyrtapz: We've agreed offline to always read the default nse to cover this scenario.

go-controller/pkg/util/multi_network.go
```diff
@@ -1586,16 +1585,32 @@ func GetPodNADToNetworkMappingWithActiveNetwork(pod *corev1.Pod, nInfo NetInfo,
...
+
+	if isPersistentIPsPrimaryNetwork {
+		// 'k8s.ovn.org/primary-udn-ipamclaim' annotation has been deprecated. Maintain backward compatibility by
+		// using it as a fallback; when defaultNSE.IPAMClaimReference is set, it takes precedence.
+		if ipamClaimName, wasPersistentIPRequested := pod.Annotations[DeprecatedOvnUDNIPAMClaimName]; wasPersistentIPRequested {
+			activeNSE.IPAMClaimReference = ipamClaimName
+		}
+		if defaultNSE != nil && defaultNSE.IPAMClaimReference != "" {
+			activeNSE.IPAMClaimReference = defaultNSE.IPAMClaimReference
+		}
```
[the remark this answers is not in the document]
kyrtapz: While it seems we were doing that check in the past:
```
if nInfo.IsPrimaryNetwork() && AllowsPersistentIPs(nInfo) && activeNSE.IPAMClaimReference == "" {
		if wasPersistentIPRequested {
			activeNSE.IPAMClaimReference = ipamClaimName
		}
	}
```
I don't think this is a possible scenario because we are creating activeNSE just above.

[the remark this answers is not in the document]
kyrtapz: This could be simplified to:
```
activeNSE.IPAMClaimReference = pod.Annotations[DeprecatedOvnUDNIPAMClaimName]
```
but it's a nit.

kyrtapz (approved):
@qinqon ptal as it seems you have a `requested changes` on this pr.

kyrtapz:
Thanks @ormergi !

PR #5368, opened 2025-07-14
virt: Delete LSP at external process killing VM

qinqon:
## 📑 Description

When something like OOM killer kill the running virt-launcher pod code enter on [1] this prevent ovnk from deleting the LSP so when VM is restarted the traffic is blocked.

This change just return nil virtual machine status, since there is no live migration going on.

[1] https://github.com/openshift/ovn-kubernetes/blob/release-4.18/go-controller/pkg/kubevirt/pod.go#L475

Fixes #

## ✅ Checks

- [x] My code requires tests

## How to verify it
It includes unit and e2e tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for live migration status, preventing unnecessary errors when no active migration pods are found.

* **Tests**
  * Enhanced readiness checks with flexible condition handling for virtual machines.
  * Updated test VM creation and recovery procedures to improve stability.
  * Added post-migration network traffic verification and recovery validation.
  * Adopted a new pointer helper for consistent field assignments in tests.
  * Added test coverage for live migration status with all pods completed.

* **Refactor**
  * Scoped test variables locally to improve test code clarity and isolation.

* **New Features**
  * Added capability to forcibly terminate virt-launcher processes on nodes to support test scenarios.

### DIFF
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -470,11 +470,15 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod) 
 
 	targetPod := vmPods[len(vmPods)-1]
 	livingPods := filterNotComplete(vmPods)
+
+	// If there is no living pod we should state no live migration status
+	if len(livingPods) == 0 {
+		return nil, nil
+	}
+
+	// There is a living pod but is not the target one so the migration
+	// has failed.
 	if util.PodCompleted(targetPod) {
-		// if target pod failed, then there should be only one living source pod.
-		if len(livingPods) != 1 {
-			return nil, fmt.Errorf("unexpected live migration state: should have a single living pod")
-		}
 		return &LiveMigrationStatus{
 			SourcePod: livingPods[0],
 			TargetPod: targetPod,
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -50,7 +50,7 @@ import (
 	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	testutils "k8s.io/kubernetes/test/utils"
 	utilnet "k8s.io/utils/net"
-	"k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	butaneconfig "github.com/coreos/butane/config"
@@ -794,9 +794,9 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 			}).WithPolling(time.Second).WithTimeout(time.Minute).Should(Succeed())
 		}
 
-		waitVirtualMachineInstanceReadiness = func(vmi *kubevirtv1.VirtualMachineInstance) {
+		waitVirtualMachineInstanceReadinessWith = func(vmi *kubevirtv1.VirtualMachineInstance, conditionStatus corev1.ConditionStatus) {
 			GinkgoHelper()
-			By(fmt.Sprintf("Waiting for readiness at virtual machine %s", vmi.Name))
+			By(fmt.Sprintf("Waiting for readiness=%q at virtual machine %s", conditionStatus, vmi.Name))
 			Eventually(func() []kubevirtv1.VirtualMachineInstanceCondition {
 				err := crClient.Get(context.Background(), crclient.ObjectKeyFromObject(vmi), vmi)
 				Expect(err).To(SatisfyAny(
@@ -807,8 +807,18 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 			}).WithPolling(time.Second).WithTimeout(5 * time.Minute).Should(
 				ContainElement(SatisfyAll(
 					HaveField("Type", kubevirtv1.VirtualMachineInstanceReady),
-					HaveField("Status", corev1.ConditionTrue),
+					HaveField("Status", conditionStatus),
 				)))
+		}
+
+		waitVirtualMachineInstanceReadiness = func(vmi *kubevirtv1.VirtualMachineInstance) {
+			GinkgoHelper()
+			waitVirtualMachineInstanceReadinessWith(vmi, corev1.ConditionTrue)
+		}
+
+		waitVirtualMachineInstanceFailed = func(vmi *kubevirtv1.VirtualMachineInstance) {
+			GinkgoHelper()
+			waitVirtualMachineInstanceReadinessWith(vmi, corev1.ConditionFalse)
 		}
 
 		waitVirtualMachineAddresses = func(vmi *kubevirtv1.VirtualMachineInstance) []kubevirt.Address {
@@ -903,7 +913,7 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 							NetworkSource: networkSource,
 						},
 					},
-					TerminationGracePeriodSeconds: pointer.Int64(5),
+					TerminationGracePeriodSeconds: ptr.To(int64(5)),
 					Volumes: []kubevirtv1.Volume{
 						{
 							Name: "containerdisk",
@@ -929,7 +939,7 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 					GenerateName: vmi.GenerateName,
 				},
 				Spec: kubevirtv1.VirtualMachineSpec{
-					Running: pointer.Bool(true),
+					RunStrategy: ptr.To(kubevirtv1.RunStrategyAlways),
 					Template: &kubevirtv1.VirtualMachineInstanceTemplateSpec{
 						ObjectMeta: metav1.ObjectMeta{
 							Annotations: vmi.Annotations,
@@ -1414,8 +1424,8 @@ fi
 					Name: "force-post-copy",
 				},
 				Spec: kvmigrationsv1alpha1.MigrationPolicySpec{
-					AllowPostCopy:           pointer.Bool(true),
-					CompletionTimeoutPerGiB: pointer.Int64(1),
+					AllowPostCopy:           ptr.To(true),
+					CompletionTimeoutPerGiB: ptr.To(int64(1)),
 					BandwidthPerMigration:   &bandwidthPerMigration,
 					Selectors: &kvmigrationsv1alpha1.Selectors{
 						VirtualMachineInstanceSelector: kvmigrationsv1alpha1.LabelSelector{
@@ -2219,15 +2229,20 @@ chpasswd: { expire: False }
 			networkData, err := staticIPsNetworkData(selectCIDRs(vmiIPv4, vmiIPv6))
 			Expect(err).NotTo(HaveOccurred())
 
-			vmi := fedoraWithTestToolingVMI(nil /*labels*/, nil /*annotations*/, nil /*nodeSelector*/, kubevirtv1.NetworkSource{
+			vm := fedoraWithTestToolingVM(nil /*labels*/, nil /*annotations*/, nil /*nodeSelector*/, kubevirtv1.NetworkSource{
 				Multus: &kubevirtv1.MultusNetwork{
 					NetworkName: cudn.Name,
 				},
 			}, userData, networkData)
 			// Harcode mac address so it's the same after live migration
-			vmi.Spec.Domain.Devices.Interfaces[0].MacAddress = vmiMAC
-			createVirtualMachineInstance(vmi)
-
+			vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].MacAddress = vmiMAC
+			createVirtualMachine(vm)
+			vmi := &kubevirtv1.VirtualMachineInstance{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: namespace,
+					Name:      vm.Name,
+				},
+			}
 			waitVirtualMachineInstanceReadiness(vmi)
 			Expect(crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)).To(Succeed())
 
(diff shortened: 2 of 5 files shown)


go-controller/pkg/kubevirt/pod.go
```diff
@@ -470,11 +470,14 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod)
 
 	targetPod := vmPods[len(vmPods)-1]
 	livingPods := filterNotComplete(vmPods)
+
+	// if there is no living pods, it can be that something has killed
+	// the vm like OOM killer so there is no live migration going on
+	if len(livingPods) == 0 {
+		return nil, nil
+	}
+
```
ormergi: Can we add UT to cover scenario where no pod is running as soon migration is completed?

test/e2e/kubevirt.go
```diff
@@ -929,7 +930,7 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 					GenerateName: vmi.GenerateName,
 				},
 				Spec: kubevirtv1.VirtualMachineSpec{
-					Running: pointer.Bool(true),
+					RunStrategy: ptr.To(kubevirtv1.RunStrategyRerunOnFailure),
```
maiqueb: what about a strategy of `always` ? wouldn't it behave the same way ? 
(I think so.)

go-controller/pkg/kubevirt/pod.go
```diff
@@ -470,11 +470,14 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod)
 
 	targetPod := vmPods[len(vmPods)-1]
 	livingPods := filterNotComplete(vmPods)
+
+	// if there is no living pods, it can be that something has killed
+	// the vm like OOM killer so there is no live migration going on
+	if len(livingPods) == 0 {
+		return nil, nil
```
maiqueb, line 476: Please bear with me for a sec here ...

In https://github.com/ovn-kubernetes/ovn-kubernetes/blob/0b513c6319e3258309ffbba0e67ce7a5d427b17e/go-controller/pkg/ovn/base_network_controller_secondary.go#L933, we're ensuring that in the scenario of a migration failure, the src pod LSP is "enabled".

We're OK because we're not in a migration right ? i.e. there's no src pod LSP to enable back, right ?

test/e2e/kubevirt.go
```diff
@@ -2253,11 +2285,34 @@ chpasswd: { expire: False }
...
 			step = by(vmi.Name, "Check east/west traffic after virtual machine instance live migration")
 			checkEastWestIperfTraffic(vmi, testPodsIPs, step)
+
+			By("Stop iperf3 traffic before force killing vm, so iperf3 server do not get stuck")
+			output, err = kubevirt.RunCommand(vmi, "killall iperf3", 5*time.Second)
+			Expect(err).ToNot(HaveOccurred(), output)
+
+			step = by(vmi.Name, fmt.Sprintf("Force kill qemu at node %q where VM is running on", vmi.Status.NodeName))
+			output, err = infraprovider.Get().ExecK8NodeCommand(vmi.Status.NodeName, []string{"bash", "-c", `kill -9 $(pidof qemu-kvm)`})
+			Expect(err).ToNot(HaveOccurred(), output)
```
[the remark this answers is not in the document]
qinqon: This is a controller environment where there is just one kubevirt VM running and the kubevirt qemu binary name is always the same, @coderabbitai mark this as resolved

go-controller/pkg/kubevirt/pod.go
qinqon: That's it, we are covering the case where none of the virt-launcher pods are running.

test/e2e/kubevirt.go
qinqon: yep, I will use Always, my trick is to force failure with `kill -9`

go-controller/pkg/kubevirt/pod.go
qinqon: Added

test/e2e/kubevirt.go
```diff
@@ -2253,11 +2268,40 @@ chpasswd: { expire: False }
...
+			output, err = kubevirt.RunCommand(vmi, "killall iperf3", 5*time.Second)
+			Expect(err).ToNot(HaveOccurred(), output)
+
+			step = by(vmi.Name, fmt.Sprintf("Force kill qemu at node %q where VM is running on", vmi.Status.NodeName))
+			output, err = infraprovider.Get().ExecK8NodeCommand(vmi.Status.NodeName, []string{"bash", "-c", `kill -9 $(pidof virt-launcher)`})
+			Expect(err).ToNot(HaveOccurred(), output)
+
+			step = by(vmi.Name, "Waiting for failed restarted VMI to reach ready state")
+			waitVirtualMachineInstanceFailed(vmi)
+			waitVirtualMachineInstanceReadiness(vmi)
```
ormergi, line 2289: Waiting for intermediate state such as this one (wait for fail and then for readiness) has potential to introduce flakes, please consider changing that test in a way to wont wait for the state flip.
Does waiting for VM readiness would work?
Would wait for the process to be gone instead? (if doesnt occur immediately)

test/e2e/kubevirt.go
```diff
@@ -2253,11 +2268,40 @@ chpasswd: { expire: False }
...
 
 			step = by(vmi.Name, "Check east/west traffic after virtual machine instance live migration")
 			checkEastWestIperfTraffic(vmi, testPodsIPs, step)
+
+			By("Stop iperf3 traffic before force killing vm, so iperf3 server do not get stuck")
+			output, err = kubevirt.RunCommand(vmi, "killall iperf3", 5*time.Second)
+			Expect(err).ToNot(HaveOccurred(), output)
+
+			step = by(vmi.Name, fmt.Sprintf("Force kill qemu at node %q where VM is running on", vmi.Status.NodeName))
+			output, err = infraprovider.Get().ExecK8NodeCommand(vmi.Status.NodeName, []string{"bash", "-c", `kill -9 $(pidof virt-launcher)`})
```
ormergi: Please consider killing the tested virt-launcher to prevent unexpected deletions or side effects

test/e2e/kubevirt.go
```diff
@@ -2253,11 +2268,40 @@ chpasswd: { expire: False }
...
+
+			step = by(vmi.Name, "Restart iperf traffic after forcing a vm failure")
+			Expect(startEastWestIperfTraffic(vmi, testPodsIPs, step)).To(Succeed(), step)
+			checkEastWestIperfTraffic(vmi, testPodsIPs, step)
+
+			by(vmi.Name, "Running live migration after forcing vm failure")
+			td(vmi)
+
+			step = by(vmi.Name, "Check east/west traffic for failed virtual machine after live migration")
+			checkEastWestIperfTraffic(vmi, testPodsIPs, step)
```
ormergi, line 2303: Why do we need to check migration again?
Isnt connectivity check is good enough indication that the network plumbing is in good shape?

test/e2e/kubevirt.go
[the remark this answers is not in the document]
maiqueb: wait a minute, I'm not so sure. 

Don't we run tests in parallel ? If so, it is possible for another test with VMs to run at the same time, and, for a VM to be scheduled in the same node isn't it ? IIUC, this would kill *all* QEMU processes running on the node, thus impacting other tests.
@qinqon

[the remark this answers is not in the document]
qinqon: > wait a minute, I'm not so sure.
> 
> Don't we run tests in parallel ? If so, it is possible for another test with VMs to run at the same time, and, for a VM to be scheduled in the same node isn't it ? IIUC, this would kill _all_ QEMU processes running on the node, thus impacting other tests. @qinqon

We don't run test in paralle, if we enter that territory we have a lot of stuff going on not just here, like the ovn-k e2e tests are not prepared for parallel stuff 

UPDATE: I can improve this by checking exactly what virt-launcher to kill

test/e2e/kubevirt.go
qinqon: Waiting just for readiness do not work because it takes some time to transition to Failed, so we either add an sleep or check for failed so we know then it has to go to ready.

test/e2e/kubevirt.go
qinqon: nope, the traffic problem occurs when the dangling LSP is at the same node as the proper one, so we need to live migrate so we make sure we check that.

maiqueb: > Why do we need to check migration again?

Good question :) 

> Isnt connectivity check is good enough indication that the network plumbing is in good shape?

I would have been even lazier, and just test that the VM starts up again, and goes to ready state. 
But I agree that checking traffic is better (but might be redundant, since we test vm startup elsewhere).

qinqon: > > Why do we need to check migration again?
> 
> Good question :)
> 

As said without another migration we do not reproduce the issue where the VM is moved to the node where the danling LSP is.

> > Isnt connectivity check is good enough indication that the network plumbing is in good shape?
> 
> I would have been even lazier, and just test that the VM starts up again, and goes to ready state. But I agree that checking traffic is better (but might be redundant, since we test vm startup elsewhere).

Checking VM start will prove nothing, we either check danling LSPs direclty with ovn-nbctl (that's not right from integration tests) or we test what we know is failing wich is the traffic.

maiqueb: I thought the pods failed to start .

test/e2e/kubevirt.go
```diff
@@ -2253,11 +2268,47 @@ chpasswd: { expire: False }
...
 
 			step = by(vmi.Name, "Check east/west traffic after virtual machine instance live migration")
 			checkEastWestIperfTraffic(vmi, testPodsIPs, step)
+
+			By("Stop iperf3 traffic before force killing vm, so iperf3 server do not get stuck")
+			output, err = kubevirt.RunCommand(vmi, "killall iperf3", 5*time.Second)
+			Expect(err).ToNot(HaveOccurred(), output)
+
+			step = by(vmi.Name, fmt.Sprintf("Force kill qemu at node %q where VM is running on", vmi.Status.NodeName))
+			// /usr/bin/virt-launcher --qemu-timeout 312s --name worker-dcf9j --uid bcf975f4-7bdd-4264-948b-b6080320e38a --namespace kv-live-migration-2575 --kubevirt-share-dir /var/run/kubevirt --ephemeral-disk-dir /var/run/kubevirt-ephemeral-disks --container-disk-dir /var/run/kubevirt/container-disks --grace-period-seconds 20 --hook-sidecars 0 --ovmf-path /usr/share/OVMF --run-as-nonroot
```
ormergi: nit: is it a leftover?

ormergi (approved):

test/e2e/kubevirt.go
qinqon: Done

test/e2e/kubevirt.go
qinqon: Is an example of how the process looks like so it's easier to understand the pgrep command.

test/e2e/kubevirt.go
qinqon: > I thought the pods failed to start .

Problem is that traffic fails since the same logical switch has two ports with the same mac and one lsp up=true and the other up=false

test/e2e/kubevirt.go
```diff
@@ -2253,11 +2268,47 @@ chpasswd: { expire: False }
...
+
+			killScript := fmt.Sprintf(`
+pid=$(pgrep -f 'virt-launcher .*--name %s.*--namespace %s'|grep -v $$)
+ps aux |grep virt-launcher
+kill -9 $pid
+`, vmi.Name, vmi.Namespace)
+			output, err = infraprovider.Get().ExecK8NodeCommand(vmi.Status.NodeName, []string{"bash", "-xe", "-c", killScript})
+			Expect(err).ToNot(HaveOccurred(), output)
```
maiqueb: huge thanks. 

Could we maybe encapsulate this in an helper, thus simplifying the test code body ?

test/e2e/kubevirt.go
maiqueb: OK, got it.

maiqueb (approved):
tks. just left an opinionated nit you can choose to ignore.

test/e2e/kubevirt.go
qinqon: > huge thanks.
> 
> Could we maybe encapsulate this in an helper, thus simplifying the test code body ?

sure, not a biggy

qinqon:
[rebased](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/bd318efddaba0f7698f80e86405e4b5a6e7a721a..8fa0ee8e90b36232801c3689c693f4098f78fe49)

maiqueb:
I see some failures loading the docker img in CI (all the failed lanes have failed for that). 

Can we somehow re-run them ?

trozet:
> I see some failures loading the docker img in CI (all the failed lanes have failed for that).
> 
> Can we somehow re-run them ?

yes you can use /retest-failed or I will restart them once CI completes

ormergi:
The job `ovn-ci / e2e (network-segmentation, noHA, shared, ipv6, noSnatGW, 1br, ic-single-node-zones` failed, looking into the logs it seems CUDN CR readiness timeout the tests setup phase.

Looking into the test artifacts, we see  various informers getting restarted by the time the test wait for the CUDN CR readiness, including the CUDN informer: 
`unable to decode an event from the watch stream: http2: client connection lost`

Although its odd, it doesnt seem related to the PR changes.
/retest-failed

---
More details:
job: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16319367889/job/46158427300?pr=5368
artifacts: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16319367889/artifacts/3553023422
In the artifacts tar, open file `ovn-control-plane/ovn-kubenetes/ovnkube-cluster-manager.log`, look around the test failure time `2025-07-17T09:19:47.2650184Z`, various informer getting restarted due to the mentioned error.

trozet:
flake?
Localnet [It] using ClusterUserDefinedNetwork CR, pods in different namespaces, should communicate over localnet topology

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16319367889/job/46158427300?pr=5368

going to re-run

qinqon:
> flake? Localnet [It] using ClusterUserDefinedNetwork CR, pods in different namespaces, should communicate over localnet topology
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16319367889/job/46158427300?pr=5368
> 
> going to re-run

@trozet the change only affect at pods annotated with kubevirt virtual machine, that's not the case at the CI error, so it's not related.

PR #5369, opened 2025-07-14
[OVN build] Fetch ovs package from the ovn submodule.

npinaeva:
Thus, we make sure to use the compatible ovs version by default, otherwise OVN may not be built correctly.
When I tried to build ovn using current version, I have got a failure https://productionresultssa8.blob.core.windows.net/actions-results/f1d79533-2387-49f1-a0e0-f3d41be2feba/workflow-job-run-ea8873ba-a1db-5e2f-87e3-50ec1407f82f/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-07-14T07%3A55%3A10Z&sig=L14t06WvfeqZabFlC1NrWKqM1ha2LU7d81QAS2f%2B7c8%3D&ske=2025-07-14T19%3A23%3A50Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-07-14T07%3A23%3A50Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-07-14T07%3A45%3A05Z&sv=2025-05-05

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/dist/images/Dockerfile.fedora
+++ b/dist/images/Dockerfile.fedora
@@ -24,14 +24,28 @@ ENV PYTHONDONTWRITEBYTECODE yes
 RUN INSTALL_PKGS="git rpm-build dnf-plugins-core" && \
     dnf install --best --refresh -y --setopt=tsflags=nodocs $INSTALL_PKGS
 
+# Clone OVN Source Code.
+ARG OVN_REPO=https://github.com/ovn-org/ovn.git
+ARG OVN_GITREF=main
+WORKDIR /root
+RUN mkdir ovn && pushd ovn && \
+    git init && \
+    git remote add origin $OVN_REPO && \
+    git fetch origin ${OVN_GITREF} --depth 1 && \
+    git reset --hard FETCH_HEAD && \
+    popd
+
 # Clone OVS Source Code.
 ARG OVS_REPO=https://github.com/openvswitch/ovs.git
-ARG OVS_GITREF=branch-*
+# OVS_GITREF can be set to a specific commit or branch, otherwise the version pinned by OVN will be used.
+ARG OVS_GITREF=""
 WORKDIR /root
-RUN mkdir ovs && pushd ovs && \
+RUN OVS_OVN_GITREF=$(cd ovn && git submodule status ovs|cut -c 2-|cut -d ' ' -f 1) && \
+    mkdir ovs && pushd ovs && \
     git init && \
     git remote add origin $OVS_REPO && \
-    git fetch $OVS_REPO $(git ls-remote origin "${OVS_GITREF}" | sort -V -k2 | tail -1 | awk '{print $1}') --depth 1 && \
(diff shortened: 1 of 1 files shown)


dist/images/Dockerfile.fedora
```diff
@@ -24,21 +24,21 @@ ENV PYTHONDONTWRITEBYTECODE yes
...
+RUN mkdir ovn && pushd ovn && \
     git init && \
-    git remote add origin $OVS_REPO && \
-    git fetch $OVS_REPO $(git ls-remote origin "${OVS_GITREF}" | sort -V -k2 | tail -1 | awk '{print $1}') --depth 1 && \
+    git remote add origin $OVN_REPO && \
+    git fetch origin ${OVN_GITREF} --depth 1 && \
     git reset --hard FETCH_HEAD && \
-    echo "1" && \
-    find rhel && \
+    git submodule update --init --recursive && \
```
almusil: nit: I would add the `--depth 1` here too.

almusil (commented):
From my point of view the change is good, however it removes the possibility of mixing different ovs and ovn version. @qinqon you did the original changes, was there any intention behind that?

dist/images/Dockerfile.fedora
```diff
@@ -24,21 +24,21 @@ ENV PYTHONDONTWRITEBYTECODE yes
 RUN INSTALL_PKGS="git rpm-build dnf-plugins-core" && \
     dnf install --best --refresh -y --setopt=tsflags=nodocs $INSTALL_PKGS
 
-# Clone OVS Source Code.
-ARG OVS_REPO=https://github.com/openvswitch/ovs.git
-ARG OVS_GITREF=branch-*
```
qinqon: Can we still keeping the old behaviour if users want to specify OVS ? so we can do something like 
```
ARG OVS_REPO=submodule
```
And later on we check `if [ "OVS_REPO" == "submodule ]"` the code you are doing if not the old way.

npinaeva: I can do that, but the question is whether someone will ever use it? maybe you have an example of when you needed a newer ovs version than the one that OVN uses?

qinqon: > I can do that, but the question is whether someone will ever use it? maybe you have an example of when you needed a newer ovs version than the one that OVN uses?

When you need a fix at OVS but not a OVN (way less frequent though), so to test that the fix should be backport to the stable branch where OVN is taking OVS from and then we will pass OVS_REPO and OVS_GITREF to that and OVS from submodule will be replaced by it.

npinaeva: alright, I just changes how the default OVS version is fetched, but you can still set another one.
e.g. if you do `OVS_GITREF="branch-3.5"` the build will fail

qinqon: Good enough

PR #5373, opened 2025-07-15
ovn-controller: block ARP replies and router advertisements during startup

martinkennelly:
Prevent ovn-controller from sending stale ARP replies and router
advertisements by adding drop flows on external bridge patch ports
until ovnkube-controller synchronizes the southbound database - henceforth
known as "drop flows".

This addresses race conditions where ovn-controller processes outdated
SB DB state before ovnkube-controller updates it, particularly affecting
EIP SNAT configurations attached to logical router ports.
Fixes: https://issues.redhat.com/browse/FDP-1537

ovnkube-controller controls the lifecycle of the drop flows.
ovs / ovn-controller running is required to configure external bridge.
Downstream, the external bridge maybe precreated and ovn-controller
will use this.

This fix considers three primary scenarios: node, container and pod restart.

On Node restart means the ovs flows installed priotior to reboot on the node are
cleared but the external bridge exists. Add the flows before ovnkube controller
with node starts. The reason to add it here is that our gateway code depends
on ovn-controller started and running...
There is now a race here between ovn-controller starting
(and garping) before we set this flow but I think the risk is low however
it needs serious testing. The reason I did not naturally at the drop
flows before ovn-controller started is because I have no way to detect
if its a node reboot or pod reboot and i dont want to inject drop flows
for simple ovn-controller container restart which could disrupt traffic.
ovnkube-controller starts, we create a new gateway and apply flows the same
flows in-order to ensure we always drop ARP reply / RA when ovnkube controller
hasn't sync.

Remove the flows when ovnkube-controller has syncd. There is also a race here
between ovnkube-controller removing the flows and ovn-controller GARPing with
stale SB DB info. There is no easy way to detect what SB DB data ovn-controller
has consumed.

On Pod restart, we add the drop flows before exit. ovnkube-controller-with-node
will also add it before it starts the go code.

We will revert all this when https://issues.redhat.com/browse/FDP-1537 lands and we refactor eip controller 

TBD; e2e


martinkennelly:
cc @pliurh @crnithya PTAL

martinkennelly:
Rebased only

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -498,6 +509,15 @@ func generateIPFragmentReassemblyFlow(ofPortPhys string) []string {
 	return flows
 }
 
+func generateGratuitousARPDropFlow(inPort string, priority int) string {
+	// set to op code 1 - see rfc5227 particularly section:
+	// Why Are ARP Announcements Performed Using ARP Request Packets and Not ARP Reply Packets?
+	// ovn follows this practise of using op code 1
+	// TODO: match on dst mac ?
+	return fmt.Sprintf("cookie=%s,table=0,priority=%d,in_port=%s,arp,arp_op=1,actions=drop",
```
pliurh: This blocks all ARP requests. A GARP has the sender and target IP addresses set to the same value. Can we use `arp_spa=arp_tpa` to filter out GARP only?

go-controller/pkg/controllermanager/node_controller_manager.go
```diff
@@ -224,15 +225,45 @@ func (ncm *NodeControllerManager) Start(ctx context.Context) (err error) {
 			return fmt.Errorf("failed to own priority %d for IP rules: %v", node.UDNMasqueradeIPRulePriority, err)
 		}
 	}
+
+	// start workaround and remove when ovn has native support for silencing GARPs for LRPs
+	// https://issues.redhat.com/browse/FDP-1537
+addDropFlowsLoop:
```
pliurh: The name `addDropFlowsLoop` is kind of misleading. Aren't we trying to remove the drop flow here?

pliurh (commented):
This patch is for the cluster default network. Can such issue also happened in EIP with UDN?

go-controller/pkg/node/bridgeconfig/bridgeflows.go
martinkennelly: True - Let me ask someone from the OVN team to double check it.

go-controller/pkg/controllermanager/node_controller_manager.go
martinkennelly: 100% - dumb naming - let me change!

go-controller/pkg/node/bridgeconfig/bridgeflows.go
martinkennelly: We'd have to match on the IP address for support.
It gets overly complex if I attempt to do that considering we are also adding drops in shell files.
I am keeping it simple until we have ovn native support for silencing garps for EIP IPs on startup.
WDYT?

martinkennelly:
first push was addressing @pliurh  comment regarding badly named var
second push was rebase on main

martinkennelly:
Downstream verification underway.
I manually tested it and it looked good.

martinkennelly:
> This patch is for the cluster default network. Can such issue also happened in EIP with UDN?

No, because ovn-controller doesnt generate GARPs for pods selected by UDNs because we dont add a SNAT for UDN Pods. We add the EIP to the bridge and GARP is generated there but this bug is only for ovn-controller.

go-controller/pkg/node/bridgeconfig/bridgeflows.go
pliurh: Yeah, it makes sense.

pliurh:
LGTM

martinkennelly:
/assign @jcaamano 

^ you were selected in team meeting - lmk if you dont have time.

martinkennelly:
Downstream verification is good for pod and node reboot.

martinkennelly:
Rebased only

martinkennelly:
Downstream, we need to clear `ovn-remote` so ovn-controller doesnt connect before ovnkube-controller is able to add the drop flows. The work around will be removed when ovn has native support. OVNKube controller then adds `ovn-remote` on startup. It would be nice if we had a deployment model upstream like we do downstream so i could include said changes here too.

martinkennelly:
TD; add docs.

dist/images/ovnkube.sh
```diff
@@ -1717,7 +1730,10 @@ ovnkube-controller() {
 }
 
 ovnkube-controller-with-node() {
-  trap 'kill $(jobs -p) ; rm -f /etc/cni/net.d/10-ovn-kubernetes.conf ; exit 0' TERM
+  # send sig term to background job (ovnkube-node process), remove CNI conf and resume background job until it ends.
+  # currently we only send ovnkube node process to background, therefore when using fg we are resuming this process and
```
trozet: i dont see you using fg to bring to foreground?

trozet (commented):
"The reason I did not naturally at the drop
flows before ovn-controller started is because I have no way to detect
if its a node reboot or pod reboot and i dont want to inject drop flows
for simple ovn-controller container restart which could disrupt traffic."

So from what I gather from this we do not want to inject drop flows if ovn-controller restarts, but if ovnkube-controller restarts we do? Why is one OK and the other isn't?

I think if you manually add the flow during ovn-controller start up, within a few seconds ovnkube-controller will overwrite any flow that was manually added anyway. We are only blocking arp request/GARP flows from OVN...GARPs are not that often, and it shouldn't block ingress ARP requests and OVN replies. It could block ARP requests for next hop egress gateways.

It would be nice to narrow the flow further to only GARPs, but not sure it is possible.

trozet:
Perhaps to make it safer, you could add another, higher priority flow that ALLOWS arp op_code for the node IP as arp_spa? Then at least we are only dropping GARP for non-Node IP.

martinkennelly:
Will do.

martinkennelly:
> "The reason I did not naturally at the drop
> flows before ovn-controller started is because I have no way to detect
> if its a node reboot or pod reboot and i dont want to inject drop flows
> for simple ovn-controller container restart which could disrupt traffic."
> 
> So from what I gather from this we do not want to inject drop flows if ovn-controller restarts, but if ovnkube-controller restarts we do? Why is one OK and the other isn't?

Because sb db is hot and we dont have to be concerned about GARPs

> I think if you manually add the flow during ovn-controller start up, within a few seconds ovnkube-controller will overwrite any flow that was manually added anyway. We are only blocking arp request/GARP flows from OVN...GARPs are not that often, and it shouldn't block ingress ARP requests and OVN replies. It could block ARP requests for next hop egress gateways.
>
> It would be nice to narrow the flow further to only GARPs, but not sure it is possible.

Its not possible for what I want to do and I asked OVN team. I'll paste the info shortly to this issue and we can discuss that response from them. The drop follow should only allow request arps with a specific Mac (ff:ff..)

Within the flow generation for ovnkube, I'll add a higher prio to allow node IP. Must check if possible/clean.

trozet:
@martinkennelly can you expand a little more on "Because sb db is hot and we dont have to be concerned about GARPs"?

I think you are saying that we dont need to worry cause when ovn-controller comes back up, it will have the latest info and program immediately the latest state (hot SBDB). But what about the case where, ovn-controller goes down and is stuck for a while, OVNK moves the egress IP somewhere else...are we not worried about this scenario because GARP is handled via ovn-controller as packet out? so if ovn-controller is down, no GARPs will be sent anyway?

martinkennelly:
> @martinkennelly can you expand a little more on "Because sb db is hot and we dont have to be concerned about GARPs"?
> 
> I think you are saying that we dont need to worry cause when ovn-controller comes back up, it will have the latest info and program immediately the latest state (hot SBDB). But what about the case where, ovn-controller goes down and is stuck for a while, OVNK moves the egress IP somewhere else...are we not worried about this scenario because GARP is handled via ovn-controller as packet out? so if ovn-controller is down, no GARPs will be sent anyway?


If ovnkube controller is running, then I think for now,  that sb db is configured correctly and ok for ovn controller to reconnect to sb db.

martinkennelly:
Its still possible we garp incorrectly in that scenario you described if ovnkube controller didnt react fast enough to an EIP Node reallocation from that node to another node but its not something I want to focus on due to the complexity to cover that edge case when theres a native ovn fix coming.

martinkennelly:
I'll try to ensure a higher prio flow allows node IPs.

martinkennelly:
Added higher prio flows to allow GARPs when src IPv4 is  Node IP(s). Thanks @trozet

martinkennelly:
Feeling a bit more comfortable now with this change Tim.. I hope this work around can be reverted in the next month or two.

martinkennelly:
@trozet PTAL - thanks

martinkennelly:
Ill update docs on following PR..

dist/images/ovnkube.sh
martinkennelly: Ill update this

martinkennelly:
Removed the reference to fg in commit message and the ovnkube.sh script.

martinkennelly:
@trozet Ready for re-review - thanks

go-controller/pkg/node/bridgeconfig/bridgeflows.go
```diff
@@ -882,6 +915,55 @@ func (b *BridgeConfiguration) PMTUDDropFlows(ipAddrs []string) []string {
...
+// Remove when https://issues.redhat.com/browse/FDP-1537 is available
+func (b *BridgeConfiguration) allowNodeIPGARPFlows(nodeIPs []net.IP) []string {
+	if config.Gateway.Mode != config.GatewayModeShared || !config.IPv4Mode {
+		return nil
+	}
+	const priority = 499
+	var flows []string
+
+	defaultNetInfo := util.DefaultNetInfo{}
+	defaultNetPatchPortName := defaultNetInfo.GetNetworkScopedPatchPortName(b.bridgeName, b.nodeName)
```
trozet, line 950: with egress IP and UDN, how do GARPs work there? Is the NAT done with br-ex flows? or is it done in the GR?

martinkennelly: For UDNs, We add the EIP IP and  also flows to snat to ext bridge. Garp handled by that bridge when we add the IP.

martinkennelly: For UDNs, NATs done on breth0.

martinkennelly: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5557

Issue #5375, opened 2025-07-15
Support trust zone assignment to Nodes in IC mode


booxter:
### What would you like to be added?

Support a new annotation to `Node` that would allow to assign particular trust zones to Chassis.

### Why is this needed?

By default, OVN controller builds tunnels between all chassis, even if these
tunnels are, by policy, not to be used. These tunnels are unnecessary in some
scenarios (Edge or when nodes are dedicated to run workloads for an isolated
tenant) and may potentially expose a node with a rogue workload to other nodes.

## Goals

Allow cluster admins to dedicate `Node`s to particular trust zones,
making sure that no traffic will be tunneled between disjoint zones.

Make sure that a rogue workload cannot force other nodes to establish
unintended tunnels to the rogue node.

PR #5376, opened 2025-07-15
okep: Trust Zones for IC mode (#5375)

booxter:
## 📑 Description

OKEP for Transport Zones for IC mode feature (#5375).

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Documentation**
  * Added a new enhancement proposal, "Trust Zones for Interconnect," describing a design to restrict tunnel creation between nodes in OVN-Kubernetes IC mode.
  * Introduced a new node annotation for assigning nodes/chassis to one or more trust zones (API and design documented).
  * Updated site navigation to include the new proposal under "Enhancement Proposals."

### DIFF
--- a/docs/okeps/okep-5375-trust-zones.md
+++ b/docs/okeps/okep-5375-trust-zones.md
@@ -18,8 +18,8 @@ considerations are a secondary benefit.
 
 ## Goals
 
-Allow cluster admins to dedicate chassis to particular trust zones, making sure
-that no traffic will be tunneled between disjoint zones.
+Allow cluster admins to dedicate chassis to particular trust zones, only building
+Geneve tunnels between directly connected zones.
 
 Make sure that a rogue chassis cannot force other chassis to create unintended
 tunnels to the rogue chassis.
@@ -35,60 +35,180 @@ rationale below in the `Risks, Known Limitations and Mitigations` section.)
 ## Introduction
 
 Inter-chassis tunnels carry some cost, in terms of security, performance, and
-stability. Even when a tunnel is not used, chassis attempt to create
-corresponding Interfaces.
+scale. Even when a tunnel is not used, chassis attempt to create corresponding
+Interfaces.
 
 This is unnecessary and may even expose victim workloads to an attacker if they
-manage to hijack a chassis, since the attacker then has direct tunnel to all
-other chassis.
+manage to hijack a chassis, since the attacker then has direct tunnels to all
+chassis.
 
-This OKEP proposes to expose trust zone functionality through a `Node`
-annotation. When set, the annotation will control which chassis should have
-tunnels between them, and which should not.
+This OKEP exposes trust zone functionality through a new `TrustZone` CRD. When
+created, a `TrustZone` CR will affect which chassis have tunnels between them,
+and which don't.
 
-Only cluster admin should be able to set this annotation.
+Only cluster admin should be able to create `TrustZone` CRs.
 
 ## User-Stories/Use-Cases
 
-### Story 1: Deny tunnel traffic between some chassis
+### Story 1: Isolate a set of chassis in a trust zone
 
-As a cluster admin, I want to apply an annotation to some chassis that will
-distinguish them into a separate trust zone, so that only chassis that belong
-to the same zone will create tunnel endpoints between them.
+As a cluster admin, I want to assign a set of chassis to a trust zone, so that
+only a chassis that belongs to the trust zone creates tunnel endpoints to other
+chassis from the zone.
 
-### Story 2: Group chassis in multiple trust zones
+### Story 2: Group a set of chassis in multiple trust zones
 
 As a cluster admin, I want to apply unique trust zones to disjoint subsets of
-chassis, so that each subset can host workloads that can communicate inside the
-same subset but cannot communicate between disjoint subsets. (It's up to me as
-an admin to make sure trust zone assignment reflects respective node selectors
-used.)
-
-### Story 3: Set trust zones for DPU chassis
-
-As a cluster admin, I want to set trust zones for on-DPU OVN chassis that may
-be different from trust zones that the on-Host OVN chassis is part of. By doing
-so, I am able to isolate traffic handled by Host `ovnkube-controller` from
-traffic handled by DPU `ovnkube-controller` on VTEP level.
-
-### Story 4: Dedicate chassis to gateway appliances (pods)
-
-As a cluster admin, I want to isolate most chassis into disjoint trust zones
-while also dedicating a subset of shared "gateway" chassis that would be
-present in multiple (maybe all) trust zones. These dedicated chassis will run
-gateway appliances (pods) to handle external as well as internal traffic for
-disjoint trust zones.
-
-To make it work, some routing mechanism between isolated trust zones and their
-respective gateway appliance pods is needed. (One option may be
-`AdminPolicyBasedExternalRoutes`. There may be alternatives using BGP and other
-routing protocols.)
-
-**Defining such redirection mechanism is out of scope for this OKEP.**
+chassis, so that each subset can tunnel traffic internally but not between
+zones.
 
 ## Proposed Solution
 
-### Core OVN
+### At a Glance
+
+OVN-Kubernetes will maintain the list of remote `Chassis` in SBDB so that only
+chassis that share a trust zone with the local chassis will have a
+corresponding `Chassis` record in SBDB. Since there won't be `Chassis` records
+for chassis from disjoint trust zones, no tunnels will be created.
+
+To guarantee that a rogue chassis won't be able to change trust zone membership
+for itself or other chassis, changing `TrustZone` CRs will not be denied using
+RBAC unless the user is a cluster admin.
+
+### API Details
+
+#### TrustZone CRD
+
+A new `TrustZone` CRD will be created in `k8s.ovn.org/v1alpha1` API group as
+follows:
+
+```
+apiVersion: k8s.ovn.org/v1alpha1
+kind: TrustZone
+metadata:
+  name: trust-zone-a
+spec:
+  nodeSelector:
+    matchLabels:
+      tenant: tenant-a
+status:
+  members:
+    - ovnkube-pod-1
+    - ovnkube-pod-2
+    - ovnkube-pod-3
+  conditions:
+    - type: Ready
+      status: "True"
+      reason: AllTunnelsEstablished
+      lastTransitionTime: "2025-08-13T15:04:05Z"
+```
+
+In the above example, the `TrustZone` named `trust-zone-a` selects nodes that
+are assigned to a tenant using a `tenant` label. The `status.members` field
+lists the names of `ovnkube-node` pods that belong to the trust zone. The
+`status.conditions` field indicates whether all tunnels between members of the
+trust zone have been successfully established.
+
+```
+apiVersion: k8s.ovn.org/v1alpha1
+kind: TrustZone
+metadata:
+  name: trust-zone-a
+spec:
+  nodeSelector:
+    matchExpressions:
+      - key: k8s.ovn.org/zone-name
+        operator: In
+        values:
+        - tenant-a-node1
+        - tenant-a-node2
+        - tenant-a-node3
+status:
+  members:
+    - ovnkube-pod-1
+    - ovnkube-pod-2
+    - ovnkube-pod-3
+  conditions:
+    - type: Ready
+      status: "True"
+      reason: AllTunnelsEstablished
+      lastTransitionTime: "2025-08-13T15:04:05Z"
+```
+
+In the above example, the `TrustZone` named `trust-zone-a` selects nodes that
+match one of `zone-name` labels. The `status.members` field lists the names of
+`ovnkube-node` pods that belong to the trust zone. The `status.conditions`
+field indicates whether all tunnels between members of the trust zone have been
+successfully established.
+
+In the future, the `TrustZone` CRD may be extended to support additional
+selector types, e.g. to match against `ovnkube-node` pod labels. (This may be
+useful when running multiple `ovnkube-node` pods that belong to the same node
+but different trust zones.)
+
+### Implementation Details
+
+A new controller will be added to `ovnkube-node` to watch for `TrustZone` CRs.
+All `ovnkube-node` pods will watch for all `TrustZone` CRs in the cluster.
+
+#### DB configuration
+
+On `TrustZone` CR created, `ovnkube-node` will create `Chassis` records
+for remote chassis that have at least one common trust zone set, and remove
+`Chassis` records for remote chassis that do not have any common trust zones
+set.
+
+On `TrustZone` CR deleted, `ovnkube-node` will create `Chassis` records for
+remote chassis that also have no trust zones set, and remove `Chassis` records
+for remote chassis that have any trust zones set.
+
+In addition to the `Chassis` records, `ovnkube-node` will also insert or
+delete any other records that pertain to remote chassis, namely:
+
+* Logical Switch Port in the transit switch.
+* Logical Router Port in the distributed cluster router.
+* Static Route in the distributed cluster router.
+
+This will be achieved by extending `UpdateResource` handler for `NodeType` to
+call the `deleteNodeEvent` code path as needed.
+
+#### Admission enforcement
+
+Only cluster admin should be able to create or delete `TrustZone` CRs. This
+will be enforced using RBAC.
+
+### Testing Details
+
+Unit tests should be updated to confirm that `UpdateResource` handler for
+`NodeType` inserts and deletes remote `Chassis` and related records as per the
+rules described above.
+
+E2E test scenario should confirm inter-zone connectivity guarantees are
+enforced. Pods should successfully communicate if they run on different chassis
+that are not part of any trust zones, or if they are part of the same trust
+zones. Pods should fail to communicate if they run on chassis that belong to
+different trust zones.
+
+### Documentation Details
+
+A new page in `api-reference` guide should be added to describe the new
+`TrustZone` CRD.
+
+A new page in `design` guide should be added to briefly explain the intent and
+design of the trust zone feature from the perspective of an operator.
+
+## Risks, Known Limitations and Mitigations
+
+The feature is not implemented for multi-node IC or centralized modes that use
+shared SBDB.
+
+## OVN Kubernetes Version Skew
+
+Proposed for 1.2.
+
+## Alternatives
+
+### OVN Transport Zones
 
 OVN [has](https://mail.openvswitch.org/pipermail/ovs-dev/2019-April/358314.html)
 support for Transport Zones. Transport zones are configured in local OVSDB by
@@ -109,220 +229,23 @@ chassis. But since peer chassis with `transport_zones` set will not set up
 tunnels to these implicitly marked chassis, no tunnels between marked and
 unmarked chassis will be created either.
 
-### OVN-Kubernetes
+Trust zones could have been implemented using OVN Transport Zones too. In
+general, it's a viable approach, but it has some drawbacks:
 
-Instead of using OVN Transport Zones directly, OVN-Kubernetes will maintain the
-list of remote `Chassis` in SBDB so that only chassis that share a trust zone
-with the local chassis will have a corresponding `Chassis` record in SBDB.
-Since there won't be `Chassis` records for chassis from disjoint trust zones,
-no tunnels will be created.
-
-To guarantee that a rogue chassis won't be able to change its `Node` annotation
-to expose the chassis to illegal trust zones, an admission hook should be
-installed to deny such updates. (Note: this is already the case.)
-
-### API Details
-
-#### DPUs
-
-The proposed solution should work in the presence of a DPU chassis, with and
-without a Host chassis. DPU chassis don't have a separate `Node` object, so
-they may have to use its Host `Node` object to communicate about trust zones.
-
-There may be more than one DPU chassis per Host. The proposed solution
-should support specifying trust zones for DPU chassis that are different from
-other chassis that belong to the same `Node` (whether on-DPU or Host).
-
-(In a similar vein, the proposed API format should be flexible enough for
-multiple virtual OVN chassis running on the same Host  or per-VTEP trust
-zones.)
-
-#### Node annotation
-
-To serve all possible permutations of chassis (DPU vs Host; numbers of each
-type; multiple VTEPs), a new `Node` object annotation of map type will be
-added.
-
-Each key in the map will represent a particular endpoint entity. In the scope
-of this OKEP, "endpoint entity" refers to an OVN chassis, but its definition
-may later be expanded to e.g. support per VTEP trust zones. For OVN chassis,
-`Chassis` `system-id` will be used as the key.
-
-Each value of the map will be a list of trust zones that the endpoint entity
-(chassis) belongs to.
-
-A default key `__default` will be used to represent the default trust zones for
-endpoints that do not have a specific key set.
-
----
-
-For example, if all OVN chassis on a `Node` are part of the same trust zones,
-one can set the annotation like this:
-
-```yaml
-apiVersion: v1
-kind: Node
-metadata:
-  annotations:
-    k8s.ovn.org/trust-zones: '{"__default": ["tz1","tz2","tz3"]}'
-```
-
-This setting will serve scenarios with a single OVN chassis on the `Node`
-(either on-DPU or on-Host); or with multiple OVN chassis that all belong to the
-same trust zones.
-
----
-
-If multiple OVN chassis are present on a `Node`, and they belong to different
-trust zones, the following annotation can be used:
-
-```yaml
-apiVersion: v1
-kind: Node
-metadata:
-  annotations:
-    k8s.ovn.org/trust-zones: '{"<system-id1>": ["tz1"], "<system-id2>": ["tz2"]}'
-```
-
-or:
-
-```yaml
-apiVersion: v1
-kind: Node
-metadata:
-  annotations:
-    k8s.ovn.org/trust-zones: '{"__default": ["tz1"], "<system-id2>": ["tz2"]}'
-```
-
-(in the latter case, all chassis but `<system-id2>` will be part of `tz1`).
-
----
-
-In the future, if per-VTEP trust zones are needed, the annotation can be
-extended to support multiple VTEPs per chassis, e.g.:
-
-```yaml
-apiVersion: v1
-kind: Node
-metadata:
-  annotations:
-    k8s.ovn.org/trust-zones: '{"<system-id1>": {"vtep1": ["tz1"], "vtep2": ["tz2"]}}'
-```
-
-or:
-
-```yaml
-apiVersion: v1
-kind: Node
-metadata:
-  annotations:
-    k8s.ovn.org/trust-zones: '{"<system-id1:vtep1>": ["tz1"], "<system-id1:vtep2>": ["tz2"]}'
-```
-
-(Defining the exact key format for this scenario is beyond the scope of this
-OKEP.)
-
----
-
-Note: the `trust-zones` annotation will be ignored for Centralized mode.
-
-### Implementation Details
-
-#### DB configuration
-
-On `Node` `trust-zones` annotation update, `ovnkube-controller` will:
-
-1. If the annotation is not set, create `Chassis` records for remote chassis
-   that also have no trust zones set, and remove `Chassis` records for remote
-   chassis that have any trust zones set.
-
-2. If the annotation is set, create `Chassis` records for remote chassis that
-   have at least one common trust zone set, and remove `Chassis` records for
-   remote chassis that do not have any common trust zones set.
-
-Since we are not going to use OVN Transport Zones feature directly,
-`ovnkube-controller` `chassis_handler.go` will **not** need access to local
-OVSDB client to update trust zone assignment for the local chassis.
-
-Attention should be given to transition of chassis between remote and local
-states.
-
-In addition to the `Chassis` records, `ovnkube-controller` will also insert or
-delete any other records that pertain to remote chassis, namely:
-
-* Logical Switch Port in the transit switch.
-* Logical Router Port in the distributed cluster router.
-* Static Route in the distributed cluster router.
-
-This will be achieved by extending `UpdateResource` handler for `NodeType` to
-call the `deleteNodeEvent` code path as needed.
-
-#### Admission enforcement
-
-The `nodeadmission.go` hook already enforces that a `ovnkube-controller` can
-only update a particular explicit list of annotations, so no changes are needed
-to guard against a rogue chassis updating its own trust zones.
-
-```console
-[root@ovn-worker ~]# kubectl annotate nodes ovn-worker k8s.ovn.org/trust-zones='{"__default": ["tz1"]}'
-Error from server (Forbidden): nodes "ovn-worker" is forbidden: User "system:serviceaccount:ovn-kubernetes:ovn-node" cannot patch resource "nodes" in API group "" at the cluster scope
-```
-
-Note: since the rogue chassis, by definition, has access to its own SBDB in
-single-node IC mode, it can always affect its list of `Chassis` records. But
-it's not enough to mislead other intact chassis that use separate databases,
-which is required to create tunnels between rogue and intact chassis.
-
-### Testing Details
-
-Unit tests should be updated to confirm that `UpdateResource` handler for
-`NodeType` inserts and deletes remote `Chassis` and related records as per the
-rules described above.
-
-E2E test scenario should confirm inter-zone connectivity guarantees are
-enforced. Pods should successfully communicate if they run on different chassis
-that are not part of any trust zones, or if they are part of the same trust
-zones. Pods should fail to communicate if they run on chassis that belong to
-different trust zones.
-
-### Documentation Details
-
-A new page in `api-reference` guide should be added to list supported `Node`
-annotations. (Only the newly proposed annotation has to be described there.)
-
-A new page in `design` guide should be added to briefly explain the intent and
-design of the trust zone feature from the perspective of an operator.
-
-## Risks, Known Limitations and Mitigations
-
-The feature is not implemented for multi-node IC or centralized modes that use
-shared SBDB.
-
-## OVN Kubernetes Version Skew
-
-Proposed for 1.2.
-
-## Alternatives
-
-### OVN Transport Zones
-
-As discussed above, trust zones could have been implemented using OVN Transport
-Zones too. In general, it's a viable approach, but it has some drawbacks:
-
-* `ovnkube-controller` would need to access local OVSDB to update
+* `ovnkube-node` would need to access local OVSDB to update
   `external-ids:ovn-transport-zones` setting, which is not currently the
   case. This would require additional code to pass a local OVSDB client into
   zone IC handler. Only interacting with SBDB simplifies the implementation a
   bit.
+
 * The proposed approach allows to avoid creating remote `Chassis` and its
   supporting resources (LSP, LRP, Static Routes) for chassis that are disjoint
   from the local chassis. This should help with scalability, as the number of
   redundant records in SBDB is reduced.
 
-Note: future developments like per-VTEP trust zones may require using OVN
-Transport Zones, but this OKEP does not cover such use cases. If so, both the
-approach proposed in this OKEP and OVN Transport Zones may be combined for
-multi-layered isolation and control plane scaling benefits.
+Utilizing OVN Transport Zones directly may be revisited in the future if these
+allow for even more granular segmentation than what we can achieve natively in
+OVN-K.
 
 ### Using hypervisor firewall for isolation
 
@@ -335,14 +258,6 @@ results in the large number of firewall rules.
 
 The firewall approach is not mutually exclusive with this OKEP and may be
 applied on top of it for security in depth, if needed.
-
-### Deriving trust zones from logical topology
-
-Trust zone assignment could probably have been determined through complex
-calculation of logical topology adjacency, e.g. based on UDN attachments. Even
-if technically possible, it would complicate the logic and may not have covered
-all possible administrative policy needs that cluster admins may want to impose
-on inter-chassis communication.
 
 ## References
 


docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,188 @@
...
+nodes are dedicated to run workloads that belong to a specific isolated
+tenant). Not just redundant, they may potentially expose a rogue node to other
+victim nodes.
+
+## Goals
+
+Allow cluster admins to dedicate `Node`s to particular OVN transport zones,
+making sure that no traffic will be tunneled between disjoint zones.
+
+(For IC) Make sure that a rogue workload cannot force other nodes to establish
```
girishmg: NIT: perhaps you meant, s/rogue workload/rogue node/

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Story 1: Deny tunnel traffic between some nodes
+
+As a cluster admin, I want to apply an annotation to some `Node`s that will
+distinguish them into a separate transport zone, so that only chassis that
+belong to the same zone will create tunnel endpoints between them.
+
+### Story 2: Group nodes in multiple transport zones
+
+As a cluster admin, I want to apply different transport zones to some subsets
```
girishmg: s/transport zones/transport zone/

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,188 @@
...
+manage to hijack a node since the attacker then has direct tunnel to all other
+nodes.
+
+This OKEP proposes to expose transport zone functionality through a `Node`
+annotation. When set, the annotation will control which `Node`s should have
+tunnels between them, and which should not.
+
+Only cluster admin should be able to set this annotation.
+
+## User-Stories/Use-Cases
```
girishmg: We need to also support an use-case where a Node is part of two or more transport zones.

Say, a node is running Egress Gateway service pod. All the pods from different transport zones might be routed to this egress gateway service pod through overlay and from this pod the traffic is SNATed and sent out to the Internet.

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,188 @@
...
+always affect its own view of zone membership of all chassis. But it's not
+enough to mislead other intact `Node`s, which is required to establish tunnels
+between rogue and intact nodes.
+
+Note: the discussion above only applies to Interconnect mode where SBDBs are
+disjoint. In Centralized mode, a rogue chassis can update its own transport
+zones to e.g. add it to all other known transport zones (which can be extracted
+from the shared `Chassis` table), which will force all other chassis to
+establish tunnels to the rogue chassis. **I am not aware of a solution to this
+attack scenario for Centralized mode.**
```
girishmg: I think we should say that this feature will work `securely` only for the single-node IC zone, where-in every K8s node is its own zone.

If the deployers end up running IC for bunch of nodes, where-in, there is going to be central OVN NB/SB for bunch of these nodes, then this will not work.

docs/okeps/okep-5375-transport-zones.md
booxter: I think this is an expansion / clarification to Story 3. I will reword and expand the existing story to make the mechanism of "dedication of a subset of nodes as routing points" clearer. Let me know if this is sufficient, or you would like me to separate the story into two separate stories - one for communication between segregated nodes, and one for node-to-egress.

docs/okeps/okep-5375-transport-zones.md
booxter: This is a great point. I was assuming a single-node IC zone when I wrote it, but indeed, OVN-K allows to centralize multiple nodes in a single IC zone, in which case the whole IC zone shares the destiny.

I will expand this text; I will also move it down into Risks and Limitations section where it probably belongs.

docs/okeps/okep-5375-transport-zones.md
booxter: I meant multiple zones, as in: each subset gets its own zone, so plural seems fitting, though wording is probably not great. I reworded it a bit to make it clearer. Let me know if it's any better.

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,195 @@
...
+By default, OVN controller builds tunnels between all chassis, even if these
+tunnels are, by administrative policy or configuration, not to be used. These
+tunnels are unnecessary in some scenarios (Edge computing, or when particular
+nodes are dedicated to run workloads that belong to a specific isolated
+tenant). Not just redundant, they may potentially expose a rogue node to other
+victim nodes.
+
+## Goals
+
+Allow cluster admins to dedicate `Node`s to particular OVN transport zones,
```
trozet: This renders weird: `Node`s

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,195 @@
...
+attributes (e.g. UDN attachments).
+
+Resolving the rogue node problem for Centralized mode. (See rationale below.)
+
+## Introduction
+
+Inter-chassis tunnel establishment carries some cost, in terms of security,
+performance, and stability. Even when a tunnel is not used, chassis attempt to
+establish them. If DC network policy forbids them, controllers will still
+attempt to do so, probably timing out while doing so.
```
trozet: what do you mean by establish tunnels here? The tunnel is UDP/Geneve and has no connection establishment AFAIK.

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,195 @@
...
+distinguish them into a separate transport zone, so that only chassis that
+belong to the same zone will create tunnel endpoints between them.
+
+### Story 2: Group nodes in multiple transport zones
+
+As a cluster admin, I want to apply unique transport zones to disjoint subsets
+of `Node`s, so that each subset can host workloads that can communicate inside
+the same subset but cannot communicate between disjoint subsets.
+
+### Story 3: Enforce centralized N-S tunneling through dedicated nodes
```
trozet: It's not really N-S tunneling. I think you are describing dedicated gateway nodes. This use case would require adding routes to nodes so that they can get to the gateway nodes. Is that going to be covered as part of this okep?

There are plans already to make this same idea possible with BGP. When a nodeSelector in an RA selects only a few nodes (BGP gateways), then all other nodes will be configured with routes to ECMP to those gateways. We should align the efforts.

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,195 @@
...
+points for any traffic between segregated nodes, or between the nodes and the
+Internet. These routing nodes may run gateways or any other additional network
+services needed to enforce administrative policies, e.g. DPI.
+
+## Proposed Solution
+
+### Core OVN
+
+OVN [has](https://mail.openvswitch.org/pipermail/ovs-dev/2019-April/358314.html)
+support for Transport Zones. Transport zones are configured in local OVSDB by setting:
```
trozet: with IC, why do you need to use OVN for this? Couldn't you just identify in OVNK which nodes are part of same zone, then only add remote chassis for those nodes?

Also, isn't this very similar to multi-node-zone which we are trying to remove from the repo cause we said there is no use case :)

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,195 @@
...
+
+(For IC) Make sure that a rogue node cannot force other nodes to establish
+unintended tunnels to the rogue node.
+
+## Non-Goals
+
+Automatically determining zone assignment for `Node`s based on workload
+attributes (e.g. UDN attachments).
+
+Resolving the rogue node problem for Centralized mode. (See rationale below.)
```
trozet: why implement this at all for centralized mode? Is this a downstream only feature that you are trying to upstream to gain parity with upstream?

docs/okeps/okep-5375-transport-zones.md
```diff
@@ -0,0 +1,195 @@
...
+
+As a cluster admin, I want to apply an annotation to some `Node`s that will
+distinguish them into a separate transport zone, so that only chassis that
+belong to the same zone will create tunnel endpoints between them.
+
+### Story 2: Group nodes in multiple transport zones
+
+As a cluster admin, I want to apply unique transport zones to disjoint subsets
+of `Node`s, so that each subset can host workloads that can communicate inside
+the same subset but cannot communicate between disjoint subsets.
```
trozet: how is this visible or controlled to a user? If they launch pods that end up on two different transport zones, they cant communicate east/west?

docs/okeps/okep-5375-transport-zones.md
booxter: In the scenario I was thinking about, funneling of traffic to dedicated gateway nodes would be forced by an admin routing policy. I imagine the mechanism to inject routes to upstream gateway nodes could be different too.

Now that I think about it, what I'm describing here is already possible with OVN-K, and it's only the VTEP isolation stories (1 and 2) that make sense in the context of the OKEP. I will drop this "dedicated gateway nodes" story (3) completely.

---

Regardless, I'm curious to learn more about your BGP plans. Are these written down somewhere?

docs/okeps/okep-5375-transport-zones.md
booxter: Sorry, wrong word.

This was also written under assumption that OVN-K may rely on BFD / HA Chassis, but apparently it's not the case, so I'm also removing the notion of time outs.

docs/okeps/okep-5375-transport-zones.md
booxter: Yes, directly controlling the list of remote chassis created may be an option. It may even simplify the code since we would not need to touch local ovsdb (and hence we would not need to pass its client through call stack and handlers...)

One potential concern I have is whether any other part of the code may currently expect presence of remote chassis records in the database. I'm very new to the code base, so perhaps you can help me to estimate it.

---

I'll need some help to understand the "multi-node-zone" comment. Do you mean IC zones with multiple nodes that share north services (dbs+northd)? If so, they seem different because they allow Nodes to update SBDB that is then read by other Nodes. (Please confirm this is what you refer to. I'm very new here and may easily misread.)

docs/okeps/okep-5375-transport-zones.md
booxter: Sorry, very new to the project and not sure what API parity standards are. :) I assumed that I would have to handle both deployment scenarios in some form, even if doesn't bring much security benefit. Code wise, handling the annotation (by setting TZs in local ovsdb) vs. explicitly ignoring it are very similar in terms of complexity (both are simple once you have IC handled).

I've also heard that Central mode is going away, but I know little about exact timelines and the mode status in terms of expected parity. Is it captured somewhere I can double-check?

---

Not sure how exactly to address the upstream parity question, but: this is a net-new code, still to be written and used in any context (see WIP at #5378).

docs/okeps/okep-5375-transport-zones.md
booxter: If pods land into different TZs, they - by design - should not communicate with each other.

In the scenario I'm considering, **end users don't have direct access to Pods API** so they cannot control where their workload lands. Instead, a higher level service takes orders from end users and sets Node selectors so that the workloads land in the right spot in the cluster. It's up to the higher level service to make sure workloads that should communicate have at least one shared TZ for their respective Nodes.

What is the definition of the "right" spot may depend on a particular use case. For example, an end user (or their collective) may receive dedicated cluster nodes from their admin to run their workloads. The admin would ensure that these nodes belong to the same TZs to enable cross-node communication. The admin would then also ensure that end user workloads land on one of these nodes, to guarantee communication between legal peers and denying it otherwise.

---

I'll mention that it's up to cluster admin to guarantee correct placement and zone assignment.

docs/okeps/okep-5375-transport-zones.md
girishmg: Supporting `ovn-central` mode is not a downstream requirement.

We already don't support UDNs for `ovn-central` case, so we have precedence where we are not adding new features to `ovn-central` mode.

Let's keep this feature only for ovn-k8s Interconnect mode.

docs/okeps/okep-5375-transport-zones.md
girishmg: @trozet within the context of single-node IC we are trying to implement who can form VTEP tunnels with whom. We don't want to use 'multi-node-zone' because RAFT OVN DBs will be back in the picture, and we don't want to go there.

docs/okeps/okep-5375-transport-zones.md
girishmg: Sorry, I don't see Story 3 anymore. Where are you capturing the fact that a node can be in multiple OVN-transport zones.

docs/okeps/okep-5375-transport-zones.md
girishmg: @booxter we still need to capture a use case where a single node could be part of two transport-zones. 

An example of that would be CoreDNS Pods where-in Pods from different transport-zones will need to communicate over the Overlay with CoreDNS Pods. So, Control Plane nodes in K8s cluster will have a leg in all the transport-zones.

booxter: Indeed, I lost this part when ripping the story off. Restoring.

docs/okeps/okep-5375-transport-zones.md
booxter: I will rename the OKEP to explicitly mention it's for IC only. I will also update the spec document to say that the feature is not supported for Centralized, and why.

I'm also stating there that the annotation will be *ignored* in Centralized mode. But please let me know if it should be explicitly denied through an admission hook.

docs/okeps/okep-5375-transport-zones.md
trozet: OK that's good -- no multi-node-zone or central support. So for IC I think it would make sense to just not add remote chassis, remote ports, pod/eip routes for nodes that are not in the local node's zone. It seems natural and will help with scale.

We will need to think about how egress features like egress IP and egress Service will work. There may be others.

I guess also it would be good to know the primary motivation for introducing this new feature. Is it to help with scale? Is it to dedicate only certain nodes as gateways?

docs/okeps/okep-5375-transport-zones.md
trozet: I don't think the BGP plans are documented upstream yet. I'm working on getting the BGP enhancement merged upstream: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5297

I was going to post another enhancement after that migration to talk about how dedicated nodes would work. But in essence with IC its like this... lets say you have 3 gateway nodes (gateway1..3) and we are talking about a layer 3 type of topology. On every node we typically have an OVN switch, ovn_cluster_router and gateway router (GR). In the scenario where every node is a gateway node, we route egress traffic up to the GR and it goes out the physical NIC.

However, in this scenario we wont have a GR non-gateway nodes. So we need to add ecmp routes to all non-gateway nodes in the ovn_cluster_router for egress to one of the gateway nodes:

default via gateway1 ecmp
default via gateway2 ecmp
default via gateway3 ecmp

With BGP, we dont SNAT packets to node IP on egress either which further complicates things because the reply packet or any other ingress packet might go to any one of the 3 gateways. Therefore we have to be careful about conntrack.

Anyway the point is if we are breaking up nodes into groups, we need to also cover how the underlying networking changes. I can help with that once I understand the use cases a bit more.

girishmg: > With BGP, we dont SNAT packets to node IP on egress either which further complicates things because the reply packet or any other ingress packet might go to any one of the 3 gateways. Therefore we have to be careful about conntrack.

We will not be able to use this downstream because we have a strict requirement that no` overlay packet` can be seen on the underlay without an encapsulation.

docs/okeps/okep-5375-transport-zones.md
girishmg: The primary motivation is for Security.

If someone escapes out of the Pod is on the host, then they shouldn't not be able to send VTEP packets out to Nodes outside of their subset of nodes.

docs/okeps/okep-5375-transport-zones.md
booxter: To close the loop here, the latest revision has this section now:

```
### Story 3: Dedicate gateway nodes to communicate with isolated transport zones

As a cluster admin, I want to isolate most nodes into disjoint transport zones
while also dedicating a subset of shared "gateway" nodes that would be present
in multiple (maybe all) transport zones. These dedicated gateway nodes would
then handle external and internal communications across disjoint transport
zones.

To make it work, some routing mechanism between isolated transport zones and
their respective gateway nodes is needed. (One option may be
`AdminPolicyBasedExternalRoutes`). Defining such mechanism is out of scope for
this OKEP.
```

docs/okeps/okep-5375-transport-zones.md
trozet: Egress packets don't have encapsulation. With BGP you can still use Geneve overlay for east/west traffic.

docs/okeps/okep-5375-transport-zones.md
trozet: If someone hacks a node (and has kapi access) then nothing stops them from being able to add the remote tunnel for another node and send packets via geneve. Also, besides geneve the worker node would be able to send any traffic it wants to the other node to attack it. Maybe a better approach to totally isolate nodes is to use a firewall on the nodes to block traffic from any nodes it should not communicate with? We still do "transport zone" groupings and then apply ingress firewalls to block traffic from worker nodes that should not be able to talk to this node.

booxter: The assumption here is that we can forbid unsafe zone membership changes via kapi from the node itself. And that the existing annotation admission hook for Nodes already does it, implicitly (only explicitly listed annotations are allowed to be changed). Apart from the new Node-level annotation, what other changes can affect tunnel interfaces on other nodes?

What a node can affect is its own SBDB Chassis table, but because nodes are decoupled from each other, it's not enough for other nodes to accept tunneled traffic from the offensive node.

girishmg: > If someone hacks a node (and has kapi access) then nothing stops them from being able to add the remote tunnel for another node and send packets via geneve. 

but the remote node's SBDB doesn't consider the local node as being within its transport zone, so it will drop the packet.

> Also, besides geneve the worker node would be able to send any traffic it wants to the other node to attack it.

Yes, they can formulate the traffic and send it, but again they remote node will drop all these packets since the local node is not part of the zone. This is all done at the hardware level.

I think applying ACLs through OVS OpenFlow rules (we will need a lot of OpenFlow rules -- say 1000s of nodes and 100s of transport zones) at the node level is still a possibility, however achieving the same through transport zones is much simpler.

docs/okeps/okep-5375-transport-zones.md
girishmg: Let's not bring BGP to OKEP, because it is not related to the transport-zone OKEP. Besides, @booxter removed that 3rd usecase for gateways.

booxter: In the scenario we are looking at, we are going to route between regular and gateway pods in overlay: tenant routing policy redirects external traffic through a gateway pod's (V)IP. The gateway pod can then make a routing / inspection decision, then - if admitted - forward it externally. I believe only at this point frames will be decapsulated and forwarded in underlay.

With BGP fabric, it is probably different since pod IPs are routed by the fabric itself.

---

I wonder if "dedicated gateway node" is too vague a term to use for both deployments...

trozet: so are we removing the use case of allowing dedicated gateway nodes to route between transport zones?

docs/okeps/okep-5375-transport-zones.md
trozet: sorry maybe I'm not being clear. The point I'm trying to make is:
1. The main intention of this enhancement is to provide security from hijacked nodes talking to other nodes.
2. The solution proposed for this is to use OVN transport zones to drop geneve traffic from isolated nodes.
3. IMO a better solution is to not scope to geneve and instead block *all* traffic from an isolated node.

docs/okeps/okep-5375-transport-zones.md
booxter: Story 3 is still in the OKEP, so I'll try to make it more explicit as to what it's trying to achieve.

---

I'm going to disambiguate the terms a bit in the new version of the OKEP, namely by suggesting that these dedicated nodes with run "gateway appliance pods" to handle external traffic from isolated TZs. This should make it clear that there's another bump in the wire (a gateway pod) in the considered design.

I hope this will make it clearer that we are not talking about mere Logical Router structure.

---

I'm also going to mention BGP as one direction to implement routing between isolated TZs and gateway appliances located in other "gateway TZs". But I'm also stating that defining the exact mechanism for redirection is out of scope.

docs/okeps/okep-5375-transport-zones.md
booxter: I'm going to mention the firewall in Alternatives section and explain how it relates to the OKEP, namely, that it may be a viable approach but has its performance implications, and that it is not in conflict with the OKEP and could be potentially applied in lieu or on top of TZs, if needed for security in depth or other considerations.

---

As to the attack vector concerns, I think the OKEP already explains the assumption well in `Admission enforcement` and a bit more in `Risks, Known Limitations and Mitigations`, so I won't expand it unless we have some unaddressed concerns still. (Please speak up.)

booxter:
I've pushed a new revision. Here are the major changes:

- Adopted the alternative design proposed by @trozet to directly manage the list of remote `Chassis` records so that nodes from disjoint TZs don't have Chassis, LRP, LSP, Static_Route records.
- Renamed the feature into "Trust Zones" to reflect the fact we are not using native OVN Transport Zones here anymore. (I hope to avoid confusion that OVN-K has with Interconnect...)
- Mentioned the idea to use firewall rules to block intra-node traffic in Alternatives section.
- Tried to disambiguate Story 3 by stating that external and cross-TZ traffic would be handled by dedicated nodes running **gateway appliance pods**.
- Listed 1.2 as the target version, as per zoom discussion we had yesterday.

Please take another look.

(Note: I put the switch from OVN Transport Zones to OVN-K Chassis record management into a separate patch on top of the original one, in case we'd like to revert it as a result of the discussion. I will squash them once we are in clear agreement as to the direction.)

girishmg:
My two cents on this proposal:

OVN-native transport zones influences how the VTEP tunnels are setup between the chassises.

Very recently we added support for multiple VTEPs per OVN chassis. Say, we update OVN-Native support for multiple-VTEPs, then I should be able to do something like this on the same OVN chassis (micro-segmentation).

```
+---------------------------+                              +---------------------------+
|     Host1 (chassis-1)     |                              |     Host2 (chassis-2)     |
|                           |                              |                           |
|                        +--VTEP1---+    +-----+     +--VTEP1---+                      |
|                        |   CX2    |----+     +---->|   CX2    |                      |
|                        +--+-------+    |     |     +-----+----+                      |
|                           |            | TZ1 |           |                           |
| +-------------------+  +--VTEP2---+    |     |     +---VTEP2--+ +------------------+ |
| | ovnk8s-controller |  |   CX3    |----+     +----->   CX3    | |ovnk8s-controller | |
| +-------------------+  +--+-------+    |     |     +-----+----+ +------------------+ |
|                           |            +-----+           |                           |
+---------+---VTEP0--+------+                              |          +---VTEP0--+     |
          |          |                                     +----------+          +-----+
          |         ++------------------------------------------------++         |      
          |   NIC2  |                 trused-zones=TZ0                 |  NIC2   |      
          |         ++------------------------------------------------++         |      
          |          |                                                |          |      
          +----------+                                                +----------+      
```



or like this:

```
+---------------------------+                              +---------------------------+
|     Host1 (chassis-1)     |                              |     Host2 (chassis-2)     |
|                           |                              |                           |
|                        +--VTEP1---+    +-----+     +--VTEP1---+                      |
|                        |   CX2    |----+ TZ1 +---->|   CX2    |                      |
|                        +--+-------+    +-----+     +-----+----+                      |
|                           |                              |                           |
| +-------------------+  +--VTEP2---+    +-----+     +---VTEP2--+ +------------------+ |
| | ovnk8s-controller |  |   CX3    |----+ TZ2 +----->   CX3    | |ovnk8s-controller | |
| +-------------------+  +--+-------+    +-----+     +-----+----+ +------------------+ |
|                           |                              |                           |
+---------+---VTEP0--+------+                              |          +---VTEP0--+     |
          |          |                                     +----------+          +-----+
          |         ++------------------------------------------------++         |      
          |   NIC2  |                 trused-zones=TZ0                 |  NIC2   |      
          |         ++------------------------------------------------++         |      
          |          |                                                |          |      
          +----------+                                                +----------+      
```



I am NOT opposed to the new proposal of implementing trusted-zones in OVN-K8s IC. However,
IMHO we will also need OVN native TZ if and when we add support for trusted-zone per-VTEP.
So, we will need both:
1. ovn-k8s IC TZ to leverage not creating logical resources for remote nodes not part of the same TZ
2. ovn TZ to leverage per-VTEP tunnel creation

I was talking with @booxter about considering Host with DPU in the mix as well. As you can see below the ovnkube-controller running on the DPU provides primary interface networking and the ovnkube-controller running on the Host is doing the secondary networks or east-west networking via VTEP1.

Ideally, we want all the DPU chassis to be in one TZ and all the host chassis in on another TZ.



```
+-----------------------------------------------------------------------+    
|                          +---------------------+ +-------------------+|    
|                          |ovnk8s-node full mode| |  ovn-controller   ||    
|                          |    (enables E/W)    | | (chassis - Host1) ||    
|                          +---------------------+ +-------------------+|    
|                                     +-----------+                     |    
|  +---------------+                  |         +-+-+    +------+       |    
|  |    Kubelet    |                  |         |VF3+---->br-int|       |    
|  +---------------+                  |         +-+-+    +----+-+       |    
|                                     |    Pod    |           |         |    
|                       Management    |           |   +-VTEP1-+----+ +--+---+
|      HostIP              Port       |           |   |   breth1   | | eth1 |
|   +--------+                +---+   |   +---+   |   +------------+ +--+---+
|   |   PF   |                |VF0|   +---+VF2+---+                     |    
|   +--------+                +-+-+       +-+-+                         |    
|       |                       |           |             Host          |    
+-------+-----------------------+-----------+---------------------------+    
|       |                       |           |             DPU           |    
|   +---+----+                +-v-----------v-----+                     |    
|   |br-host |                |  ovnk8s-node-dpu  |                     |    
|   +---+----+                |   (enables N/S)   |                     |    
|       |           VTEP0     +-------------------+                     |    
|   +---+----+    +--------+  +-------------------+                     |    
|   |br-phys +----+ br-dpu |  |  ovn-controller   |                     |    
|   +--+-----+    +--------+  | (chassis - DPU1)  |                     |    
|      |                      +-------------------+                     |    
|     +--+                                                              |    
+-----+P0+--------------------------------------------------------------+    
      +--+                                                                   

```

This OKEP minimally should tackle the Primary Interface for the Pod for both Hosts with and without DPU.

booxter:
Thanks @girishmg for the insights. The current proposal doesn't satisfactorily handle the scenario when both DPU and Host run ovnkube-controller, each belonging to its own TZs. Let me expand it. See below my current thinking on how the interface should look like.

---

**On native OVN Transport Zone support.**

My understanding is that we may (will?) need to manage `Chassis:transport_zones` when we implement per-VTEP TZs, but also that it's out of scope for this OKEP. I think it's fine to address this scenario in a separate OKEP. (Please chime in if you believe it must be addressed in the scope of this OKEP though.)

---

**On annotation interface.**

We discussed this OKEP with @girishmg today a bit, and one thing we agreed on was that either way we'll need to handle unique TZs set for DPU(s) vs Host, and DPU chassis don't (always) have corresponding `Node`s, so their TZs will need to be communicated through the same Host `Node` object (extracted via `K8S_NODE` variable) - meaning DPU chassis will need a separate vessel to pass TZ membership info for DPU(s) vs Host.

We were considering several options, e.g. introducing `dpu-trust-zones` in contrast to `trust-zones` (perhaps with a fallback to the latter if the former is not defined), or - in reverse - introduce `host-trust-zones` to override TZs for the Host.

One other complication worth considering upfront is a scenario with multiple DPUs on the same Host (or even multiple virtual OVN chassis on the same Host?) I'm not saying necessarily handling these scenarios in code (some scenarios may be too weird or useless to be used in the real world), but just consider them when defining the TZ annotation interface.

One solution could be **expanding the proposed annotation format from a simple list into a map**, reusing the same annotation for more elaborate scenarios:

- (1) **simplest case** - e.g. just 1 DPU; or just Host; or DPUs+Host sharing same TZs (same TZs for all chassis):

```
trusted-zones={
  default: [tz0,tz1,tz2]
}
``` 

- (2) **more elaborate** - separate TZs per Host/DPU (TZs per chassis):


```
trusted-zones={
  default: [tz0],
  dpu1-chassis-name: [tz1],
  dpu2-chassis-name: [tz2]
}
```

or:

```
trusted-zones={
   host-chassis-name: [tz0],
   dpu1-chassis-name: [tz1],
   dpu2-chassis-name: [tz2]
}
```

- (3) **most elaborate** - per-VTEP TZs (TZs per chassis-vtep pair):

```
trusted-zones={
  host-chassis-name:vtep0: [tz0],
  host-chassis-name:vtep1: [tz1],
  host-chassis-name:vtep2: [tz2],
  dpu1-chassis-name:vtep2: [tz2],
  dpu1-chassis-name:vtep3: [tz3],
  ...
}
```

(not sure if chassis:vtep syntax is best we can do...)

We'll start with trusted-zones (1) and (2). (2) is needed to conform with this request: "all the DPU chassis to be in one TZ and all the host chassis in on another TZ."

---

I am going to send an updated proposal tomorrow that should adopt the new annotation format described above. If you have any initial comments before then, feel free to share.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,350 @@
...
+for other chassis that also belong to at least one of the listed zones.
+
+When `transport_zones` are not set, tunnel Interfaces are created for all other
+chassis. But since peer chassis with `transport_zones` set will not set up
+tunnels to these implicitly marked chassis, no tunnels between marked and
+unmarked chassis will be created either.
+
+### OVN-Kubernetes
+
+Instead of using OVN Transport Zones directly, OVN-Kubernetes will maintain the
```
booxter: FYI we can also do both if we anticipate further developments for per-VTEP TZs. I have both implemented in the latest push at https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5378 for your reference.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,350 @@
...
+    k8s.ovn.org/trust-zones: '{"__default": ["tz1","tz2","tz3"]}'
+```
+
+This setting will serve scenarios with a single OVN chassis on the `Node`
+(either on-DPU or on-Host); or with multiple OVN chassis that all belong to the
+same trust zones.
+
+---
+
+If multiple OVN chassis are present on a `Node`, and they belong to different
```
booxter: Reading through the code and trying to implement it (and talking to Nithya R), I now realize that OVN-K is perhaps not ready for multi-chassis Nodes, yet, so the initial implementation may assume node-chassis-id is the target key for now. A separate initiative may be needed to support multiple chassis per single Node. But the annotation syntax will be ready to support it when ready.

booxter: Adjusted the OKEP to clarify multi-chassis is a future scenario; added details about how chassis id is extracted (from `node-chassis-id` annotation.)

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+As a cluster admin, I want to isolate most chassis into disjoint trust zones
+while also dedicating a subset of shared "gateway" chassis that would be
+present in multiple (maybe all) trust zones. These dedicated chassis will run
+gateway appliances (pods) to handle external as well as internal traffic for
+disjoint trust zones.
+
+To make it work, some routing mechanism between isolated trust zones and their
+respective gateway appliance pods is needed. (One option may be
+`AdminPolicyBasedExternalRoutes`. There may be alternatives using BGP and other
+routing protocols.)
```
trozet: I think this story just adds confusion and is the same as 2. Adding multiple trust zones to the gateway appliance zone will add remote chassis (and thus geneve tunnels ) for multiple groups to the gateway zone.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+same subset but cannot communicate between disjoint subsets. (It's up to me as
+an admin to make sure trust zone assignment reflects respective node selectors
+used.)
+
+### Story 3: Set trust zones for DPU chassis
+
+As a cluster admin, I want to set trust zones for on-DPU OVN chassis that may
+be different from trust zones that the on-Host OVN chassis is part of. By doing
+so, I am able to isolate traffic handled by Host `ovnkube-controller` from
+traffic handled by DPU `ovnkube-controller` on VTEP level.
```
trozet: This story lacks context. In OVNK upstream we had agreed the only way we would deploy DPU from this point forward was with the 2 cluster approach. Which means with 2 cluster, this problem is solved with just trust zones used in two different clusters. However, this is implying single cluster for DPU along with multiple east/west NICs on the host. We need to come up with (at least) documentation on how this type of deployment is done and how the OVNK components interact. The dpu-host ovnk process and the ovnkube-node full process will both need to run on the host. DPU and host will need to use different SRIOV pools, and multus will have to chain CNI call them. It's also not clear if there will be any issues with the other annotations we put on the node and how the 2 ovnks will collide trying to use them.

I think without solving this first, we cannot possibly know how to solve your story for this new feature. I would defer trying to tackle this story now, and just design the API around being flexible enough to extend it in the future.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+Not just redundant, they may potentially expose a rogue chassis to other victim
+chassis that perhaps never meant to communicate with it.
+
+Security is the primary driver for this feature, while scalability
+considerations are a secondary benefit.
+
+## Goals
+
+Allow cluster admins to dedicate chassis to particular trust zones, making sure
+that no traffic will be tunneled between disjoint zones.
```
trozet: This goal seems ambiguous to me. From what I understand this goal should be about dedicating nodes to a particular trust zone, and only building Geneve tunnels between directly connected zones. I would keep it consolidated to that kind of a goal. Nothing about this implementation stops traffic from being sent. I can take over a node and craft a geneve packet towards a node IP in another trust zone and it will receive it.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+
+If multiple OVN chassis are present on a `Node`, and they belong to different
+trust zones, the following annotation can be then used:
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  annotations:
+    k8s.ovn.org/trust-zones: '{"<system-id1>": ["tz1"], "<system-id2>": ["tz2"]}'
```
trozet: how is an admin going to know and track the system-ids of every host/dpu in the cluster? That doesn't seem practical.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+
+In the future, if per-VTEP trust zones are needed, the annotation can be
+extended to support multiple VTEPs per chassis, e.g.:
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  annotations:
+    k8s.ovn.org/trust-zones: '{"<system-id1>": {"vtep1": ["tz1"], "vtep2": ["tz2"]}}'
```
trozet: this wont work because now you have changed the underlying data structure type

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+metadata:
+  annotations:
+    k8s.ovn.org/trust-zones: '{"__default": ["tz1"], "<system-id2>": ["tz2"]}'
+```
+
+(in the latter case, all chassis but `<system-id2>` will be part of `tz1`).
+
+---
+
+In the future, if per-VTEP trust zones are needed, the annotation can be
```
trozet: how can we even do per vtep trust zones? The multi-vtep implementation creates a mesh of all encap ips between 2 chassis I think. How can we segment it further?

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,365 @@
...
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  annotations:
+    k8s.ovn.org/trust-zones: '{"<system-id1:vtep1>": ["tz1"], "<system-id1:vtep2>": ["tz2"]}'
+```
+
+(Defining the exact key format for this scenario is beyond the scope of this
+OKEP.)
```
trozet: I think we do need to figure out the API now and make sure it is extensible in the future. I'm thinking it is going to be better to put this into a CRD.  Maybe something like:

```
apiVersion: networking.ovn.org/v1alpha1
kind: TrustZone
metadata:
  name: zone-a
spec:
  scope: host                # host | dpu | both
  nodeSelector:
    matchLabels:
      zone-group: group-a
  vtepSelector:              # future
    encapIPs:
      - 192.0.2.10
      - 192.0.2.11
    pfNames:
      - ens2f0
      - ens3f1
  description: "east/west trust zone for tenant A hosts"
status:
  members:
    - node1
    - node2
    - node3
  tunnelsEstablished:
    - "node1 - node2"
    - "node1 - node3"
    - "node2 - node3"
  conditions:
    - type: Ready
      status: "True"
      reason: AllTunnelsEstablished
      lastTransitionTime: "2025-08-13T15:04:05Z"
```

In the future if we also tie a UDN to a VTEP, maybe we also throw a networkSelector into the mix somewhere.

trozet: another thought -- instead of scope: host/dpu we could provide another selector "ovnkubeSelector", which ovnkubes will use to determine if this trustzone applies to their instance. OVNK pods would be labeled, and then using Downward API we can pass the config down to the OVNK instance. That would be more flexible than categorizing based on hardcoded values. I really like this idea.

booxter: I like the CRD idea. Indeed, this API went through a set of iterative extensions that brought us from a simple list annotation to a franken-blob of json map that is painful to read and update. At this point CRD may indeed be a worthy upgrade. I will adopt it in the next revision.

---

On `ovnkubeSelector`: agreed we may need additional selectors to distinguish between several `ovnkube-node` pods that "belong" to the same `Node`. I am not sure it has to be a `ovnkube` specific match (perhaps a standard `podSelector` would do?) That said...

---

As long as we have means to extend the API to new scenarios (CRD), I think it's not in scope for this OKEP to address all of them in detail and we can largely leave them out.

Namely, for the cluster-less DPU scenario I alluded to in the OKEP before, the exact `TrustZone` API would probably align with a particular way the scenario would be implemented in general, **regardless of Trust Zones**. (I haven't initially realized that this scenario is not supported yet and incorrectly assumed it.)

Same goes for segmentation on sub-chassis level (multi-VTEPs?) and any other future segmentation work.

So: I will leave discussion of these scenarios out of the next iteration of the OKEP, only mentioning they exist and alluding to the means of extension (selectors) without prescribing particular API forms yet.

docs/okeps/okep-5375-trust-zones.md
booxter: I will leave this out, it's a distraction to the OKEP.

docs/okeps/okep-5375-trust-zones.md
booxter: Well, it COULD work since JSON per se doesn't force a particular type form on us. But yeah it would look ugly and require conditional unmarshalling.

I'm adopting your CRD suggestion, so it won't be an issue.

docs/okeps/okep-5375-trust-zones.md
booxter: I incorrectly assumed `system-id`s are `hostname`s, but they are not. So it should really have said `hostname1` etc. instead. These admins know, even when `system-id=random` is used.

---

That said, we are switching to CRD and `nodeSelectors`. For multi-chassis `Node`s, we can then add new selectors to match particular chassis / `ovnkube-node`s (e.g. `podSelector`).

docs/okeps/okep-5375-trust-zones.md
booxter: Changed the wording to positive statement about effective connectivity:

```
Allow cluster admins to dedicate chassis to particular trust zones, only building
Geneve tunnels between directly connected zones.
```

instead of suggesting traffic per se can be denied.

docs/okeps/okep-5375-trust-zones.md
booxter: I'm removing the notion of DPUs from the OKEP because the cluster-less DPU scenario is not supported by OVN-K at the moment and will need a separate OKEP to address it in general. Only then we can consider extending `TrustZone` API to segment out such DPU chassis.

docs/okeps/okep-5375-trust-zones.md
booxter: Removed.

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,265 @@
...
+
+### At a Glance
+
+OVN-Kubernetes will maintain the list of remote `Chassis` in SBDB so that only
+chassis that share a trust zone with the local chassis will have a
+corresponding `Chassis` record in SBDB. Since there won't be `Chassis` records
+for chassis from disjoint trust zones, no tunnels will be created.
+
+To guarantee that a rogue chassis won't be able to change trust zone membership
+for itself or other chassis, changing `TrustZone` CRs will not be denied using
```
girishmg, line 75: s/will not/will/g?

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,265 @@
...
+    - ovnkube-pod-3
+  conditions:
+    - type: Ready
+      status: "True"
+      reason: AllTunnelsEstablished
+      lastTransitionTime: "2025-08-13T15:04:05Z"
+```
+
+In the above example, the `TrustZone` named `trust-zone-a` selects nodes that
+are assigned to a tenant using a `tenant` label. The `status.members` field
```
girishmg, line 107: what if the rogue user labels node-X to be part of `tenant-a`?

an attacker escapes out of the VM/Pod and is on the node and uses the `kubelet` kube configuration to label the node to be part of the different trust zone

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,265 @@
...
+  members:
+    - ovnkube-pod-1
+    - ovnkube-pod-2
+    - ovnkube-pod-3
+  conditions:
+    - type: Ready
+      status: "True"
+      reason: AllTunnelsEstablished
+      lastTransitionTime: "2025-08-13T15:04:05Z"
+```
```
girishmg, line 136: Say, there is one node (node-1) that is running CoreDNS Pod and all the other Chassis (1000 of them) will need to tunnel with this node because it is running a shared service. How do I do that with the above CR?

Looks like I will have to create 1000 Trust zone objects, right?

name: tz-a + CoreDNS node
spec.nodeSelector: (match all nodes in tz-a + CoreDNS node)

name: tz-b + CoreDNS node
spec.nodeSelector: (match all nodes in tz-a + CoreDNS node)

docs/okeps/okep-5375-trust-zones.md
```diff
@@ -0,0 +1,265 @@
...
+In the above example, the `TrustZone` named `trust-zone-a` selects nodes that
+match one of `zone-name` labels. The `status.members` field lists the names of
+`ovnkube-node` pods that belong to the trust zone. The `status.conditions`
+field indicates whether all tunnels between members of the trust zone have been
+successfully established.
+
+In the future, the `TrustZone` CRD may be extended to support additional
+selector types, e.g. to match against `ovnkube-node` pod labels. (This may be
+useful when running multiple `ovnkube-node` pods that belong to the same node
+but different trust zones.)
```
girishmg, line 147: I think this will be needed in the first iteration itself.

girishmg:
@booxter @trozet we also need to handle the scenario below

```
            +--------------------------------+              +------------------------------------+      
            |                 Node1 (kubelet)|              |Node2 (kubelet)                     |      
            |                                |              |                                    |      
+-----------+--------------------------------+--------------+------------------------------------+-----+
|                                           Host K8s Cluster                                           |
+-----------+--------------------------------+--------------+------------------------------------+-----+
            |                                |              |                                    |      
            | +-----------+                  |              |                      +-----------+ |      
            | |          ++---+ ovnkube-node-|              |ovnkube-node-      +--+-+         | |      
            | |    Pod   |net0|--------------+--East/West---+------------------>|net0|  Pod    | |      
            | |          ++---+ full  IC mode|  (overlay)   |full IC mode       +--+-+         | |      
            | | +----+    |                  |              |                      |+----+     | |      
            | +-+eth0+----+                  |              |                      ++eth0+-----+ |      
            |   +--+-+                       |              |                       +--+-+       |      
            +------+-------------------------+              +--------------------------+---------+      
            |      |                     DPU1|              |DPU2                      |         |      
            |      |                         |              |                          |         |      
            |      |         ovnkube-node-dpu|              |                          |         |      
            |      |                  IC mode|              |ovnkube-node-dpu          |         |      
            |      |                         |              |IC mode                   |         |      
+-----------+------+-------------------------+--------------+--------------------------+---------+-----+
|                                           DPU K8s Cluster                                            |
+-----------+------+-------------------------+--------------+--------------------------+---------+-----+
            |      |                         |              |                          |         |      
            +------+-------------------------+              +--------------------------+---------+      
                   |                                                                   |                
                   |                                                                   |                
                   |                  North/South (overlay)                            |                
            -------+-------------------------------------------------------------------+-----           
            
```

where-in we need to make sure that we have a way to isolate the N/S and E/W tunnels as wells, if required. There are two instances of onvkube-node - one on the DPU and another on the Host. Note that the DPU is not part of the Host K8s cluster and is in it's own cluster.

docs/okeps/okep-5375-trust-zones.md
booxter: I will double check but I assume `./go-controller/pkg/ovnwebhook/nodeadmission.go` should make sure that non-whitelisted annotations won't be allowed for modification by a rogue ovnkube-node.

docs/okeps/okep-5375-trust-zones.md
booxter: You will create as many TrustZone CRs as there are unique trust zones (tenants) in the cluster. For example, if you have 10 tenants, then you will create 10 TrustZone CRs, each of the form:

`name: tenant-X`
`spec.nodeAnnotationSelector: (match all nodes of tenant-X + CoreDNS node)`

(If there's only one tenant, then you don't need any TrustZones since there's nothing to isolate, but it's a degenerate case.)

---

To compare with the previous annotations design, there you will always have to maintain N separate annotations, where N is the number of nodes in the cluster, regardless of the number of trust zones.

docs/okeps/okep-5375-trust-zones.md
booxter: @girishmg Is the contention here that, in contrast to **annotations**, we don't have a webhook that would disallow modifying node labels (`./go-controller/pkg/ovnwebhook/nodeadmission.go`)?

We could probably build another webhook that would not admit label modifications. But perhaps a more straightforward path would be to switch to **`nodeAnnotationSelector`** (I already mix the two below where I match against `k8s.ovn.org/zone-name` in `matchExpressions` that - at least traditionally - is used for labels.) Then rely on the existing annotation admission webhook to enforce its immutability for `ovnkube-node`. (Same for the existing pod annotations admission hook.)

docs/okeps/okep-5375-trust-zones.md
booxter: I am happy to include it, but I think @trozet [suggested](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5376#discussion_r2274608409) that OVN-K (upstream) only supports separate clusters for DPUs, in which case, DPUs could consult `TrustZone` CRs in their own cluster. (Which doesn't seem the best idea to me - using two sources of truth and two endpoints to orchestrate - but I may miss context.)

@trozet have I understood your suggestion correctly? Do you think that maybe pulling zone info from two sources may be too complicated and we should instead capture pod annotations selector in the OKEP?

booxter:
> we also need to handle the scenario below
> 
> [...snip...]
> 
> where-in we need to make sure that we have a way to isolate the N/S and E/W tunnels as wells, if required. There are two instances of onvkube-node - one on the DPU and another on the Host. Note that the DPU is not part of the Host K8s cluster and is in it's own cluster.

@trozet suggested [here](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5376#discussion_r2274608409) that in this scenario, there may be two separate sets of `TrustZone` CRs - one for Host cluster and one for DPU cluster. I am not sure I agree this is a good approach (it would mean that on-DPU `ovnkube-node` would have to read from both APIs - and that the admin would have to adjust two sources of `TrustZone`s). But I think the main point Tim was trying to convey was that the solution for 2-cluster and cluster-less DPU scenarios might be different and would require separate prototyping / documentation.

I think overall, expanding the CRD for `pod(Annotation)Selector` could be a better option than splitting the knowledge about trust zones between the two `etcd`s. It could then look like:

```
apiVersion: k8s.ovn.org/v1alpha1
kind: TrustZone
metadata:
  name: trust-zone-a-ew
spec:
  nodeAnnotationSelector:
    matchExpressions:
      - key: k8s.ovn.org/zone-name
        operator: In
        values:
        - tenant-a-node1
        - tenant-a-node2
        - tenant-a-node3
  podAnnotationSelector:
    matchAnnotations:
      k8s.ovn.org/role: host
---
apiVersion: k8s.ovn.org/v1alpha1
kind: TrustZone
metadata:
  name: trust-zone-a-ns
spec:
  nodeAnnotationSelector:
    matchExpressions:
      - key: k8s.ovn.org/zone-name
        operator: In
        values:
        - tenant-a-node1
        - tenant-a-node2
        - tenant-a-node3
        - shared-node
  podAnnotationSelector:
    matchAnnotations:
      k8s.ovn.org/role: dpu
```

The former would be used by on-Host `ovnkube-node`; the latter by the DPU.

docs/okeps/okep-5375-trust-zones.md
girishmg: >  there you will always have to maintain N separate annotations, where N is the number of nodes in the cluster, regardless of the number of trust zones.

there will be one annotation on each node, but that annotation will have  the trust-zone name(s).

Red-tenant nodes:
```
node1`annotation: k8s.ovn.org/trust-zone: red-tenant,default
node2`annotation: k8s.ovn.org/trust-zone: red-tenant,default
```

blue-tenant nodes:
```
nodeA`annotation: k8s.ovn.org/trust-zone: blue-tenant,default
nodeB`annotation: k8s.ovn.org/trust-zone: blue-tenant,default
```

default nodes (running CoreDNS)
```
nodeX`annotation: k8s.ovn.org/trust-zone: red-tenant,blue-tenant,default

or simply

nodeX`annotation: k8s.ovn.org/trust-zone: *
```

girishmg: also, the annotation name is fixed and values can be centrally populated.

i understand the benefits of a CR and how it provides `status` and such, but don't completely agree that a CRD solution for everything.

girishmg:
These are the scenarios:

1. For K8s cluster with DPUs, the ovnkube-node Pod (full mode) will watch TrustZone CRs and does the right thing. 

2. For all practical purposes, the ovnkube-node-dpu Pod/Process running on the DPU cluster is a K8s client of the Host Cluster. So, this Pod has Kubeconfig file of the host cluster and it will be listening on the TrustZone CRs from the host cluster.

3. the cluster-less DPU is same case as (1) above. It is a K8s client of the Host cluster. Just that it is running as a static pod.

4. two instances of ovn-k - one instance {dpu-host mode and dpu mode combination} and 2nd instance {full mode on host}. This instance, 

5. For isolating nodes in DPU cluster in their own trust zone,  the ovnkube-node daemonset Pod will be responsible for it. This pod will be watching for TrustZone CRs in the DPU Cluster.

I think there is a possibility to unify the solution for (1), (2), (3), and (4) above.  Let me think and get back.

booxter:
One drawback of selector design is that reconciliation logic becomes more complex: on `TrustZone` CR update, each `ovnkube-node` will have to:

2.1. If the current `ovnkube-node` is a member of the selected group, create/update chassis and other related resources for each other `ovnkube-node` that also matches the selector.
2.2. If the current `ovnkube-node` is not a member of the selected group, then for each other `ovnkube-node` that matches the selected group, determine if it is not covered by any other existing `TrustZone` that the current `ovnkube-node` also matches for. If not, remove the peer chassis and related resources.

---

What's different with annotations is that it's significantly easier to determine whether remaining trust zones for two `ovnkube-node` instances still overlap (in which case no removal should happen) - in case of lists (or map of lists) it's a matter of finding a common subset for two pre-populated lists. In case of selector-based design, each node membership in any particular trust zone is never explicitly stored, so it has to be calculated from selector rules.

---

There were several reasons of going with CRDs:

1. Extensibility for future segmentation flavors (per-VTEP / per-network was suggested before) becomes easier with a CRD. A list/map of list annotation may require a change in the form of the json blob (e.g. adding another level in the map to sub-segment per VTEP).

2. From user perspective, it may be easier to deal with CRs instead of marshalled json blobs.

3. @trozet also suggested a CRD may be less computationally intensive than a JSON annotation. I am not sure this is the case due to the complexity of determining whether two nodes have overlapping trust zones (as described above).

I think if we address (1) meaningfully with annotations, CRDs will become unnecessary.

---

For example, (I'm leaving string escapes and other fluff out for easier reading)

All `ovnkube-node`s belong to `tz1` and `tz-shared`:

```
kind: Node
metadata:
  annotations:
    ovn.k8s.org/trust-zones: "{__default: [tz1,tz-shared]}"
```

---

In the following example, `ovnkube-node` pod annotated as `ovn.k8s.org/node-identity=ovnkube-node-ew` will belong to `tz1`, while another `ovnkube-node` pod annotated as `ovn.k8s.org/node-identity=ovnkube-node-ns` (running on DPU) will belong to `tz1` and `tz-common`.

```
kind: Node
metadata:
  annotations:
    ovn.k8s.org/trust-zones: "{ovnkube-node-ew: [tz1], ovnkube-node-ns: [tz1,tz-shared]}"
```

This scenario implies that a deployment tool for `ovnkube-node`s annotates different pods with unique "identities" that will then allow to distinguish between them. E.g.,

```
kind: Pod
metadata:
   name: ovnkube-node-host-full
   annotations:
     ovn.k8s.org/node-identity: ovnkube-node-ew
---
kind: Pod
metadata:
   name: ovnkube-node-dpu
   annotations:
     ovn.k8s.org/node-identity: ovnkube-node-ns
```

---

Then in the future, if we need to support a sub-node segmentation scheme, then the `trust-zones` annotation interface remains the same, but the "identity" mechanism may be extended to specify unique identities for each VTEP / network. For example,

```
kind: Node
metadata:
  annotations:
    ovn.k8s.org/trust-zones: "{ovnkube-net1: [tz1], ovnkube-net2: [tz2], ...}"
---
kind: Pod
metadata:
   name: ovnkube-node-host-full
   annotations:
     ovn.k8s.org/network-identities: "{network1: ovnkube-net1, network2: ovnkube-net2, ...}"
```

---

**On two-cluster scenarios.** Regardless of whether we go with CRDs or annotations, I think it's important to keep all information relevant to zone membership in a single place (meaning, the Host cluster API) and not use the DPU management cluster Nodes for this. It simplifies implementation (`ovnkube-node`s on DPUs keep interacting with Host cluster only, using their kubeconfig) as well as use (admin manages zones using a single control plane).

---

docs/okeps/okep-5375-trust-zones.md
trozet: selecting on annotation is not a common kubernetes practice. I think it would make sense to just add label checks in the nodeadmission webhook we already use today.

docs/okeps/okep-5375-trust-zones.md
trozet: imo that does not mean an annotation is better than a CRD...it means that perhaps there is room to improve on the API of the CRD so that we can allow for asymmetrical relationships...

So we can change the API to be something like:

```
kind: TrustZone
metadata:
  name: coredns-access
spec:
  nodeSelector:     # all workers
    matchLabels:
      role: worker
  peerSelector:     # CoreDNS node(s)
    matchLabels:
      runs-coredns: "true"
```

Now we have an API that selects nodes to be affected by this CRD, and then lists peers they can each peer with. To create a symmetrical relationship, just use the same selector for both nodeSelector and peerSelector.

In my mind  the reasons for the CRD:
1. It is unclear if we want to segment trust zones further in the future to be per VTEP, or per OVNK instance. These would require data structure changes in the annotation, or deprecating the annotation and adding another annotation. This is less flexible than extending the API of a CRD.
2. It looked like that if the annotation was going to be extended in the future, we would have nested maps in the annotation. This is something we shot ourselves in the foot with UDN, the cost of unmarshaling nested json maps and comparing is too expensive.
3. It looked like using selectors was going to be easier than managing raw data of hostnames.
4. The user API experience is better and less error prone with a CRD.

docs/okeps/okep-5375-trust-zones.md
trozet: So we in the upstream context of DPU, we had agreed on a 2 cluster model. I didn't want to confuse other folks upstream by bringing a single cluster model back into the mix, but as @girishmg alluded to, I'm not sure it is avoidable at this point for this feature.

I'll try to clarify the 2 cluster model. In this case we have a host cluster and a DPU cluster. In the DPU there are two OVNK stacks (one for host and one for DPU). In this scenario, traffic always goes out the DPU. The OVNK on the DPU for the  host cluster listens for k8s events only from the host cluster, while the DPU cluster OVNK listens for events on the DPU cluster. Pods can be launched on the host, but they can also be launched in the DPU itself. In this case the clusters are totally separate. You can read more about this here:
https://docs.google.com/document/d/1MDZlu4oHL3RCWndbSgC-IGLgs1QfnB1l47nPqtM5iNo/edit?tab=t.0#heading=h.g8u53k9ds9s5

Anyway in that case, there are 2 clusters so you dont need to worry about TrustZones selecting different OVNK instances.

Now, when we talk about a single cluster with DPU, and having secondary interfaces on the host, we now have OVNK on the DPU that is wiring pods towards DPU and OVNK on the host wiring up secondary interfaces. That means we have VTEPs leaving the host and VTEPs leaving the DPU, depending on which UDN we are referring to. Since we have 2 OVNKs serving different UDNs, we need to segment TrustZones based on which OVNK we need to apply the TrustZone to. In that sense it makes sense to add another selector to the CRD, becoming something like:

```
kind: TrustZone
metadata:
  name: coredns-access
spec:
  nodeSelector:     # all workers
    matchLabels:
      role: worker
  peerSelector:     # CoreDNS node(s)
    matchLabels:
      runs-coredns: "true"
  ovnkubeSelector:
     matchLabels:
       ovnk-type: host
```

then applying the right labels to the ovnk pods. The ovnk pods then use downward API to know what their label is and can ignore TrustZones for other OVNKs.

trozet:
"also suggested a CRD may be less computationally intensive than a JSON annotation. I am not sure this is the case due to the complexity of determining whether two nodes have overlapping trust zones (as described above)."

@booxter I think what you described above is not quite how it works. There are nuances to everything and we can walkthrough it in detail. Essentially though, when there is a CRD change it does not mean you walk through and evaluate each condition. You only need to check if what you care about changed. So for example if TrustZone changes its metadata name, I don't ever evaluate does this affect me other than did the selectors change. If a node's label changes then the nodeWatcher handles checking to see if that actually changed, and if it did, then it will re-evaluate.

With evaluation or re-valuation the label selector built ins are very fast. They are idiomatic to Kubernetes. We can explore the code further, but I think it is O(N) where N is the number of requirements in your selector. I think trying to unmarshal and do string parsing is going to be much slower.

I just asked chatgpt and it says:

Label selector match (labels.Selector.Matches()) with 2 requirements → ~50–100 ns.

JSON unmarshal of a small map ({"zone":"a","role":"worker"}) → ~3–10 µs (microseconds

For a fun exercise we can benchmark it ourselves with whatever our proposed CRD and annotations are.

trozet:
@booxter I wrote and tested out some benchmarking:
https://gist.github.com/trozet/da60dee8e170583b92e1ecbd28bdbe7d

It consists of 3 tests:
1. Unmarshaling a trustZone map annotation for a remote node and checking to see if it applies to our node (remote node event)
2. Using a labelSelector to identify if a change in the selector now applies to a peer node (TrustZone CRD selector change)
3. Using a labelSelector for a remote node and checking to see if it applies ot our node (remote node event)

Results:
/home/trozet/.cache/JetBrains/GoLand2024.2/tmp/GoLand/___gobench_benchmark.test -test.v -test.paniconexit0 -test.bench . -test.run ^$
goos: linux
goarch: amd64
pkg: benchmark
cpu: Intel(R) Core(TM) Ultra 9 285K
BenchmarkJSONUnmarshal
BenchmarkJSONUnmarshal-24                              	 2370673	       503.5 ns/op
BenchmarkLabelSelectorTrustZoneCRDSelectorChange
BenchmarkLabelSelectorTrustZoneCRDSelectorChange-24    	 1706650	       682.7 ns/op
BenchmarkLabelSelectorNodeChange
BenchmarkLabelSelectorNodeChange-24                    	131465083	         8.974 ns/op
PASS

What we can take from this is that:
1. The cost of rebuilding the selector and evaluating is roughly the same as unmarshaling and comparing. However, this is only done on a CRD selector change, which is not nearly as common of an event as things being selected changing.
2. The cost of using a label selector to detect a node change is 2 orders of magnitude faster than using an annotation.

That being said, I think there are a lot of places in our code where we do not cache selectors appropriately and we re-compile them everytime. I'm going to look into that as a potential perf improvement. @tssurya @kyrtapz @jcaamano FYI

booxter:
@trozet thanks for grounding the discussion into benchmarks. I will have to take a closer look and play with the code, but see below.

---

I think the benchmark doesn't cover the whole process that we'll have to go through on TZ update. This is because a mere label match against a single TZ doesn't give an answer as to whether a particular Node shares **any** trust zones with another Node. (It only gives an answer as to whether a Node shares **this particular** trust zone with another Node. There may be **more** Trust Zone CRs that affect the final result.)

With CRD, each node will need to maintain selector matches for all TZs for all Nodes, because the final answer to the question "Does the local node have to maintain a chassis record for the remote node?" is calculated from `n` answers to the question "Does the local node match TZn selectors *and* Does the remote node match TZn selectors?" where `n` is the number of TZs (tenants) in the cluster. **O(n^2)?**

**A lot of these intermediate answers can be cached**. Maintaining this cache is not free though (in terms of LOCs and process state). It's NxM matrix (N: number of TZs; M: number of Nodes). Is it what you are referring to when you said "_there are a lot of places in our code where we do not cache selectors appropriately and we re-compile them everytime_"?

(I think the `peerSelector` suggestion doesn't change the fact that label matching will have to be executed for each Node.)

---

In contrast, with annotation list, the cost of an answer to the question: "Does the local node have to maintain a chassis record for the remote node?" is: 2 JSON unmarshalling ops + 1 intersection op for two zone lists (for remote node update) or N JSON unmarshalling + N intersections (for local node update). **O(n). No caching needed.** (Though we could shave `1` / `N -1` unmarshallings if we cache results per watched Node.)

---

I will model the above assumptions in a benchmark on Monday. But please let me know if I'm missing something obvious in my reasoning.

[UPD](https://github.com/booxter/bench-selectors-vs-annotations-trust-zones/blob/master/bench_k8s_object_test.go)

```
./run.sh
goos: darwin
goarch: arm64
cpu: Apple M4 Pro
BenchmarkAnnotationAnySharedTrustZone-14    	   2479	   469781 ns/op
BenchmarkSelectorHostnameOneOf-14           	    435	  2720073 ns/op
PASS
ok  	command-line-arguments	3.790s
```

---

Regardless, I think there are good reasons to consider CRDs beyond computation complexity. I'd rather focus on which interface makes more sense and can evolve.

From user perspective, it's not clear whether "marking nodes that belong to particular trust zones" or "declaring trust zones that comprise of particular nodes" is easier to reason about. If we are talking only about 1:1 mapping between Nodes and trust zones, then arguably "marking" is easier. But "declaring zones" may be easier to deal with if a Node belongs to more than one zone. (The burden of intersection calculation is then offloaded from the API user to OVN-K.)

PR #5395, opened 2025-07-17
[WIP] primaryUDN, layer2: MAC and IP conflict detection

RamLavi:
## 📑 Description
This PR is introducing a MAC and IP detection mechanism to primaryUDN L2 networks.
It is built on top of #5321, only 4 last commits are relevant:
*  util: Add NetworkPoolManager for IP/MAC conflict detection
*  allocator: Init NetworkPoolManager with reserved IPs
*  allocator: release pool resources on pod deletion
* allocator: Add IP/MAC conflict detection on pod creation

Fixes #CORENET-6160

## ✅ Checks

- [ ] My code requires tests


maiqueb:
I think we should reach out to the epic owner / maintainer before to understand if we want to use a cached version (your pool based approach) versus a fully stateless approach, in which we'll read from the informers per request. 

The simplicity of the latter approach sounds appealing to me ... While I'm fully aware it comes with a performance cost.

RamLavi:
[Update] the IP conflict will be addressed in a different PR.

PR #5385, opened 2025-07-16
[WIP] debug e2e

npinaeva:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

PR #5396, opened 2025-07-17
Remove OVNNodeGRLRPAddrs annotation

npinaeva:
## 📑 Description

Stop using OVNNodeGRLRPAddrs annotaion.
This annotation was always evaluated based on node-id, and
not actually allocated.
That means every node can evaluate it locally

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Deterministic gateway router IPs derived from node IDs for default and user-defined networks.
  * Automatic cleanup of deprecated node annotations at startup.
  * KubeVirt live-migration now derives gateway IP/MAC directly for more reliable reconcilation.

* **Refactor**
  * Centralized gateway IP/route utilities; controllers and networking logic updated to use them.
  * Simplified node-change and reroute detection by removing join‑CIDR-based triggers.

* **Tests**
  * Test data and helpers switched to node ID–based annotations and new gateway IP utilities.


npinaeva:
@coderabbitai review

npinaeva:
@coderabbitai full review

test/e2e/kubevirt/net.go
```diff
@@ -65,37 +67,54 @@ func RetrieveIPv6Gateways(vmi *v1.VirtualMachineInstance) ([]string, error) {
...
-		lrpJoinIPString = lrpJoinAddress.IPv6
+	if len(lrpJoinAddress) == 0 {
+		return "", fmt.Errorf("missing lrp join ip at node %q", node.Name)
 	}
 
-	if lrpJoinIPString == "" {
-		return "", fmt.Errorf("missing lrp join ip at node %q with network %q", node.Name)
-	}
+	return util.IPAddrToHWAddr(*lrpJoinAddress[0]).String(), nil
+}
```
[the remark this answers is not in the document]
npinaeva: It follow existing  pattern from `GenerateGatewayMAC`

npinaeva:
it begs for https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5385 to be merged

go-controller/pkg/generator/udn/join_ips.go
```diff
@@ -0,0 +1,90 @@
...
+	// we allocate join subnets for L3/L2 primary user defined networks or default network
+	if !(netInfo.IsDefault() || (util.IsNetworkSegmentationSupportEnabled() && netInfo.IsPrimaryNetwork())) {
+		return gwRouterAddrs, nil
+	}
+	// Allocate the IP address(es) for the node Gateway router port connecting
+	// to the Join switch
+	nodeID, err := util.GetNodeIDWithError(node)
+	if err != nil {
+		// Don't consider this node as cluster-manager has not allocated node id yet.
+		return nil, fmt.Errorf("failed to generate gateway router port addressed for node %s: %w", node.Name, err)
```
trozet: nit: s/addressed/address

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -1078,8 +1078,7 @@ func (h *defaultNetworkControllerEventHandler) UpdateResource(oldObj, newObj int
 		}
 
 		_, syncEIPNodeFailed := h.oc.syncEIPNodeFailed.Load(newNode.Name)
-		// update only if the GR join IP changed for default network
-		if syncEIPNodeFailed || joinCIDRChanged(oldNode, newNode, h.oc.GetNetworkName()) {
+		if syncEIPNodeFailed {
```
trozet, line 1081: well now the IP is tied to the nodeID, so dont you need to check NodeIDAnnotationChanged ? Probably need to check all other places where you stopped checking of LRP IPs changed.

go-controller/pkg/util/node_annotations.go
```diff
@@ -1319,6 +1122,23 @@ func GetNodeID(node *corev1.Node) int {
 	return id
 }
 
+// GetNodeIDWithError returns the node ID from the OvnNodeID annotation.
+// Unlike GetNodeID, this function returns specific errors:
```
trozet: GetNodeID looks like it's only used in a handful of places, and almost all of them return an error. Why not just make GetNodeID now return an error instead of making a new function.

go-controller/pkg/generator/udn/join_ips.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	if gwRouterIPv6 == nil {
+		return nil, fmt.Errorf("failed to find an IPv6 address for gateway router in node: %s, net: %s",
+			node.Name, netInfo.GetNetworkName())
+	}
+	return gwRouterIPv6, nil
+}
+
+// TODO this can be moved to netInfo to avoid instantiating the same IPGenerator
```
trozet, line 53: is that part of this PR or a follow up?

go-controller/pkg/podannotation/external_gw_conntrack.go
```diff
@@ -1,7 +1,7 @@
 //go:build linux
 // +build linux
 
-package util
+package podannotation
```
trozet: isn't it weird to put this external gw conntrack code under podannotation? I don't see what it has to do with it.

go-controller/pkg/podannotation/pod_annotation.go
```diff
@@ -1,10 +1,11 @@
-package util
+package podannotation
```
trozet: all our annotations like node_annotations, namespace_annotations are in util. What is the circular dependency that made you do this second commit?

go-controller/pkg/util/pod_annotation.go
```diff
@@ -595,7 +596,8 @@ func AddRoutesGatewayIP(
 			// Until https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4876 is fixed, it is limited to IC only
 			if config.OVNKubernetesFeature.EnableInterconnect {
 				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
-					joinAddrs, err := ParseNodeGatewayRouterJoinAddrs(node, netinfo.GetNetworkName())
+					// TODO this doesn't work
+					joinAddrs, err := udn.GetGWRouterIPs(node, netinfo.GetNetInfo())
```
trozet: is this the cyclical dependency?

test/e2e/egressip.go
```diff
@@ -661,7 +662,7 @@ var _ = ginkgo.DescribeTableSubtree("e2e egress IP validation", feature.EgressIP
 			if err != nil {
 				return false, err
 			}
-			ips, err := util.DefaultNetworkPodIPs(pod)
+			ips, err := podannotation.DefaultNetworkPodIPs(pod)
```
trozet: i think in E2E we are trying to avoid using OVNK functional code

trozet (changes requested):

go-controller/pkg/generator/udn/join_ips.go
npinaeva: Not implemented yet, and I am not sure how big of a performance problem this is.
We do the same thing for masq IPs, so if we ever change it, makes sense to do it for both.

go-controller/pkg/ovn/default_network_controller.go
npinaeva: I don't think this check was ever needed, since both nodeID and joinIPs never change (once they have been set). And if they weren't set, AddNode will fail and the node will be in the `syncEIPNodeFailed` map.

go-controller/pkg/podannotation/external_gw_conntrack.go
npinaeva: yeah so the problem is that it is the only file that was using podannotation within util package, so I just moved it together. It probably should go to its own package, but even better would be to get rid of the old egressgw implementation and then cleanup.
I guess just moving it to its own package is the best I can do for now

go-controller/pkg/podannotation/pod_annotation.go
npinaeva: the difference is that podannotation stores the maximum amount of info that we allocate in tricky ways. And we have this https://github.com/ovn-kubernetes/ovn-kubernetes/blob/a54f16862ed0cb5238896f6c42af1dd5dcac58d3/go-controller/pkg/ovn/base_network_controller_pods.go#L560-L563 which is 2 copies of almost the same code but that is hard to actually unify. Then there are function calls from different packages tot he podannotation package.
Anyway, it is messy and I am not sure yet what is a reasonable way to work it around

test/e2e/egressip.go
npinaeva: then i am lucky that it is not a part of this PR anymore :P

go-controller/pkg/podannotation/pod_annotation.go
npinaeva: ok I found a less disruptive way to fix this, ptal

go-controller/pkg/ovn/default_network_controller.go
trozet: Agree, but I think a long time ago we used to allocate GR IPs, and there were bugs where sometimes they would change. I hope this check was there due to that and not some other reason we are missing. Let's hope that is why it was there :) You could look up the history to find out for sure if we are scared of breaking something...

npinaeva: you tell me how it was intended https://github.com/ovn-kubernetes/ovn-kubernetes/commit/188309de2876c686fb9c5e19e8efd01af3226873 :)

npinaeva:
@trozet there were some network problems, so all tests were randomly failing, should be ready for another review now

go-controller/pkg/ovn/default_network_controller.go
trozet: LOL ok I guess it was recently from me heh... I guess I was only looking at the annotation consumed and then adding checks around it and not if it was practical or not. That makes me feel better about removing the check now!

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -456,3 +459,187 @@ func allocatePodAnnotationWithRollback(
...
+			IP:   ip,
+			Mask: util.GetIPFullMask(ip),
+		},
+		NextHop: gatewayIP,
+	}
+}
+
+// addRoutesGatewayIP updates the provided pod annotation for the provided pod
+// with the gateways derived from the allocated IPs
+func AddRoutesGatewayIP(
```
trozet, line 503: nice moving it to allocator/pod...I think this makes a lot of sense

go-controller/pkg/util/node_annotations.go
```diff
@@ -507,7 +507,7 @@ func UpdateUDNLayer2NodeGRLRPTunnelIDs(annotations map[string]string, netName st
 	return annotations, nil
 }
 
-type primaryIfAddrAnnotation struct {
+type PrimaryIfAddrAnnotation struct {
 	IPv4 string `json:"ipv4,omitempty"`
 	IPv6 string `json:"ipv6,omitempty"`
 }
```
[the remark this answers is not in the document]
npinaeva: here is an example where this is used for JoinNetwork IPs https://github.com/ovn-kubernetes/ovn-kubernetes/blob/5155c91f862ad81585d7af7d5b36ae8a7c85dd72/go-controller/pkg/util/node_annotations.go#L831-L854

test/e2e/kubevirt/net.go
```diff
@@ -65,37 +67,54 @@ func RetrieveIPv6Gateways(cli *Client, vmi *v1.VirtualMachineInstance) ([]string
...
+	nodeID, err := util.GetNodeID(node)
 	if err != nil {
-		return "", err
+		// Don't consider this node as cluster-manager has not allocated node id yet.
+		return nil, err
 	}
-	if len(joinAddresses) == 0 {
-		return "", fmt.Errorf("missing join addresses at node %q for network %q", node.Name, networkName)
+	udnJoinNetv4 := "100.65.0.0/16"
+	udnJoinNetv6 := "fd99::/64"
```
trozet: Even though this function is called GetDefaultUDN...it is consumed implicitly by the test. Therefore someone who writes another kubevirt test has no idea that under the covers default values are being used, and they might add a test where they specify the UDN join subnet to the CR. I think we should find the proper network join subnet here and not assume default.

go-controller/pkg/util/node_annotations.go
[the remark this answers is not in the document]
npinaeva: Updated

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -75,27 +68,10 @@ func NewNodeAllocator(networkID int, netInfo util.NetInfo, nodeLister listers.No
...
-			na.nodeGWRouterLRPIPv6Generator = nodeGWRouterLRPIPv6Generator
-		}
-	}
-
 	if !na.hasNodeSubnetAllocation() {
 		return nil
 	}
+	na.CleanupStaleAnnotation()
 
 	clusterSubnets := na.netInfo.Subnets()
```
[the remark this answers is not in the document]
npinaeva: I wasn't trying to remove OVNNodeGRLRPAddrs, just found one unused function

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -456,3 +459,187 @@ func allocatePodAnnotationWithRollback(
...
+				if _, isIPv6Mode := netinfo.IPMode(); isIPv6Mode {
+					joinAddrs, err := udn.GetGWRouterIPs(node, netinfo.GetNetInfo())
+					if err != nil {
+						if util.IsAnnotationNotSetError(err) {
+							return types.NewSuppressedError(err)
+						}
+						return fmt.Errorf("failed parsing node gateway router join addresses, network %q, %w", netinfo.GetNetworkName(), err)
+					}
+					podAnnotation.GatewayIPv6LLA = util.HWAddrToIPv6LLA(util.IPAddrToHWAddr(joinAddrs[0].IP))
+				}
```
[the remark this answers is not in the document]
npinaeva: there is a return for non-primary networks a couple lines above

go-controller/pkg/clustermanager/node/node_allocator.go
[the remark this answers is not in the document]
npinaeva: it is only used for cleanup as expected

trozet (approved):

PR #4634, opened 2024-08-16
Fix several issues for supporting DPUs

crnithya:
#### Which issue(s) this PR fixes

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

PR #5400, opened 2025-07-18
Add support for --disable-requestedchassis flag in ovnkube controller

aserdean:
This commit adds conditional logic to pass the --disable-requestedchassis flag to the ovnkube controller when the ovn_disable_requestedchassis environment variable is set to "true".

The flag is added to the ovnkube-controller-with-node function in dist/images/ovnkube.sh, following the same pattern as other similar configuration flags like --enable-stateless-netpol.

This flag is extremely useful when ovnkube is running in DPU mode since its chassis name will differ from the hostname of the DPU host.


ricky-rav:
I see that this flag was introduced last year with commit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4634/commits/2ba317f91722bd218add70e56823ab24ff0f733f#diff-6e6b5cc7e9fa3e942ef0cffb333f8cfb1495ed1fcabafb3a9cbdcfeeae2980f7, where it was also added to a bunch of scripts for the centralized ovnk architecture (non-IC): 
- dist/templates/ovnkube-master.yaml.j2
- helm/ovn-kubernetes/charts/ovnkube-master/templates/deployment-ovnkube-master.yaml
- helm/ovn-kubernetes/values-no-ic.yaml

Do we need to make similar changes for the corresponding files for ovnk IC?

aserdean:
> I see that this flag was introduced last year with commit [2ba317f#diff-6e6b5cc7e9fa3e942ef0cffb333f8cfb1495ed1fcabafb3a9cbdcfeeae2980f7](https://github.com/ovn-kubernetes/ovn-kubernetes/commit/2ba317f91722bd218add70e56823ab24ff0f733f#diff-6e6b5cc7e9fa3e942ef0cffb333f8cfb1495ed1fcabafb3a9cbdcfeeae2980f7), where it was also added to a bunch of scripts for the centralized ovnk architecture (non-IC):
> 
> * dist/templates/ovnkube-master.yaml.j2

Isn't that deprecated and we are focusing only on the helm parts?

> * helm/ovn-kubernetes/charts/ovnkube-master/templates/deployment-ovnkube-master.yaml
> * helm/ovn-kubernetes/values-no-ic.yaml
> 
> Do we need to make similar changes for the corresponding files for ovnk IC?

The plan was to update the docs and the helm part with the OVN IC support on DPU after this PR and #5307 is merged. Please let me know if I should create an issue to track that effort and tag this conversation there.

girishmg (approved):

Issue #5358, opened 2025-07-08
`configures one IPv6 EIP and multiple namespaces and multiple pods`  failed to find rule ip rule


tssurya:
### Which jobs are flaking?

Unit test

### Which tests are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16145763765/job/45564066713?pr=5140

### Since when has it been flaking?

hit it for 2 weeks now 

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140#issuecomment-2996125495

### Reason for failure (if possible)

```
2025-07-08T14:29:06.8515481Z [38;5;9m• [FAILED] [1.273 seconds][0m
2025-07-08T14:29:06.8517385Z [0mEgressIP selectors [38;5;9m[1m[It] configures one IPv6 EIP and multiple namespaces and multiple pods[0m
2025-07-08T14:29:06.8519001Z [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:911[0m
2025-07-08T14:29:06.8519882Z 
2025-07-08T14:29:06.8520201Z   [38;5;9m[FAILED] Timed out after 1.001s.
2025-07-08T14:29:06.8520717Z   Expected success, but got an error:
2025-07-08T14:29:06.8521411Z       <*errors.errorString | 0xc054158760>: 
2025-07-08T14:29:06.8522228Z       failed to find rule ip rule 6000: from fd46::1/128 to all table 1067 
2025-07-08T14:29:06.8522795Z       {
2025-07-08T14:29:06.8523621Z           s: "failed to find rule ip rule 6000: from fd46::1/128 to all table 1067 ",
2025-07-08T14:29:06.8524274Z       }[0m
2025-07-08T14:29:06.8525932Z   [38;5;9mIn [1m[It][0m[38;5;9m at: [1m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/controllers/egressip/egressip_test.go:607[0m [38;5;243m@ 07/08/25 14:29:06.85[0m
2025-07-08T14:29:06.8527818Z [38;5;243m------------------------------[0m
```

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5274#issuecomment-2962103019

trozet:
[...]

PR #5402, opened 2025-07-19
Retry selected netlink ops because netlink notified us that config changed during operation and results unreliable

martinkennelly:
Netlink results can be incomplete or inconsistent when the dump operation is interrupted or raced with concurrent modifications. This often occurs when the data being retrieved via Netlink is changing while the dump is in progress. Libraries like `netlinksafe` handle this by returning EINTR (interrupted) and requiring a retry. 

A minor version change  to the netlink package (https://github.com/vishvananda/netlink/pull/925  v1.2.1) exposes `NLM_F_DUMP_INTR` in some netlink responses as an `EINTR` error return, with no data.

This has been causing failures in CI testing with message `results may be incomplete or inconsistent`.
No issues seen in producition or test environments but I think for a very busy node this could be a problem and increase cpu load.

Solution for now is to retry the requests on `EINTR` before returning the error. This was added to container networking plugins pkg.

Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5358

### DIFF
--- a/go-controller/pkg/config/cni.go
+++ b/go-controller/pkg/config/cni.go
@@ -74,21 +74,13 @@ func WriteCNIConfig() error {
 // ParseNetConf parses config in NAD spec
 func ParseNetConf(bytes []byte) (*ovncnitypes.NetConf, error) {
 	var netconf *ovncnitypes.NetConf
-
 	confList, err := libcni.ConfListFromBytes(bytes)
-	if err == nil {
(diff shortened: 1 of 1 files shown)


martinkennelly:
@trozet this is an annoying log spam that looks like a bug down stream and CI issues.

 I need to rebase. What you think of it and are you interested in this approach ?

PR #5405, opened 2025-07-21
Bump fedora from 41 -> 42

trozet:
Brings in a new OVS that includes fix:
https://github.com/openvswitch/ovs/commit/a119828ea608c38611f6ee60e55a7376ca471d6f


aserdean:
lgtm

Issue #5389, opened 2025-07-16
[gateway] gateway IP address check can be skipped  when ovnkube-node is running in DPU mode


aserdean:
### What is this issue about?

As per the discussion: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5301#issuecomment-3035597742

Opening an issue to track the disabling of the check on the gateway address used by ovnk in dpu mode.

This check should be disabled in case we have a bridge layout that will include an additional steering bridge.

The same could apply when ovnkube is running in full mode.

aserdean:
cc: @girishmg @trozet @npinaeva

PR #5408, opened 2025-07-22
refactor(node): improve gateway initialization readability

aserdean:
Extract interface address logic into dedicated helper function to improve code readability and maintainability in gateway initialization.

- Add getInterfaceAddressesForNodeMode() helper function
- Replace conditional logic with switch statement for better extensibility
- Simplify initGatewayPreStart() by removing duplicate error handling
- Improve code organization and reduce cognitive complexity

fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5389


go-controller/pkg/node/gateway_init.go
```diff
@@ -188,6 +187,39 @@ func configureSvcRouteViaInterface(routeManager *routemanager.Controller, iface
 	return nil
 }
 
+// getInterfaceAddressesForNodeMode returns the appropriate interface addresses based on the node mode
+func getInterfaceAddressesForNodeMode(nc *DefaultNodeNetworkController, gatewayIntf string) ([]*net.IPNet, error) {
```
npinaeva: should this be a method of `DefaultNodeNetworkController` ?

npinaeva: also trying to find a better name (considering it doesn't just gets an interface address, `GetNetworkInterfaceIPAddresses` does that)
maybe `getNodePrimaryIfAddrs`?

aserdean: > should this be a method of DefaultNodeNetworkController ?
not sure.

I think for the time its better for it to be a helper, but if we would like to use it somewhere else, it would make sense to move it to DefaultNodeNetworkController

> also trying to find a better name (considering it doesn't just gets an interface address, GetNetworkInterfaceIPAddresses does that)
maybe getNodePrimaryIfAddrs?

`getNodePrimaryIfAddrs` sounds better indeed

npinaeva: > I think for the time its better for it to be a helper, but if we would like to use it somewhere else, it would make sense to move it to DefaultNodeNetworkController

I would suggest then passing `watchFactory` and `nodeName` as args, instead of passing the whole `DefaultNodeNetworkController` to the helper

PR #5413, opened 2025-07-25
Cluster manager shouldn't error out on localnets with no subnet

ricky-rav:
Localnets without a subnet specified are perfectly valid, but cluster manager will keep erroring out because of this check in `isTopologyManaged`:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/clustermanager/secondary_network_cluster_manager.go#L85-L88
This produces a red herring in the logs, where we retry to create a secondaryNetworkClusterManager 15 times until we give up:

`I0724 18:43:21.377878      16 controller.go:257] Controller [clustermanager-nad-controller network controller]: error found while processing gh25b_tenant-blue: [clustermanager-nad-controller network controller]: failed to ensure network gh25b_tenant-blue: failed to create network gh25b_tenant-blue: no cluster network controller to manage topology`

Let's not consider this an error when ensuring a network.

## Summary by CodeRabbit

* **Bug Fixes**
  * Creation now skips networks whose topology the controller does not manage and logs the skip for better observability.
  * Synchronization no longer silently ignores "topology not managed" conditions; such sync-time issues are now surfaced for investigation.
  * Keeps creation-time skips distinct from sync-time error handling to avoid masking sync failures.

### DIFF
--- a/go-controller/pkg/networkmanager/network_controller.go
+++ b/go-controller/pkg/networkmanager/network_controller.go
@@ -373,10 +373,6 @@ func (c *networkController) syncAll() error {
 	klog.Infof("%s: syncing all networks", c.name)
 	for _, network := range validNetworks {
 		err := c.syncNetwork(network.GetNetworkName())
-		if errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
-			klog.V(5).Infof("Ignoring network %q since %q does not manage it", network.GetNetworkName(), c.name)
-			continue
-		}
 		if err != nil {
 			return fmt.Errorf("failed to sync network %s: %w", network.GetNetworkName(), err)
 		}
@@ -465,6 +461,11 @@ func (c *networkController) ensureNetwork(network util.MutableNetInfo) error {
 	// otherwise setup & start the new network controller
 	nc, err := c.cm.NewNetworkController(network)
 	if err != nil {
+		if errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
+			klog.V(5).Infof("Network %q of topology %s is not managed by %s, skipping it",
+				networkName, network.TopologyType(), c.name)
+			return nil
+		}
 		return fmt.Errorf("failed to create network %s: %w", networkName, err)
 	}
 


ricky-rav:
@qinqon PTAL

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -339,6 +339,10 @@ func (c *networkController) ensureNetwork(network util.MutableNetInfo) error {
 	// otherwise setup & start the new network controller
 	nc, err := c.cm.NewNetworkController(network)
 	if err != nil {
+		if errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
+			klog.Infof("Network %s of topology %s is not managed by %s, skipping it", networkName, network.TopologyType(), c.name)
```
pperiyasamy: can this be v5 log level ?

go-controller/pkg/networkmanager/network_controller.go
```diff
@@ -339,6 +339,10 @@ func (c *networkController) ensureNetwork(network util.MutableNetInfo) error {
 	// otherwise setup & start the new network controller
 	nc, err := c.cm.NewNetworkController(network)
 	if err != nil {
+		if errors.Is(err, ErrNetworkControllerTopologyNotManaged) {
```
pperiyasamy, line 464: The `syncAll` method is still handling `ErrNetworkControllerTopologyNotManaged` error, i think that is not needed anymore.

pperiyasamy, line 464: Nit: `err != nil && errors.Is(err, ErrNetworkControllerTopologyNotManaged`

ricky-rav: True

ricky-rav: uhm no, because I need to return the error if it's not of type "topology not managed"

go-controller/pkg/networkmanager/network_controller.go
ricky-rav: yup

ricky-rav:
Ops, I forgot to separate the new changes from the rebase. Hopefully with only a few lines modified, it's not too annoying. Sorry, @pperiyasamy ! :)

ricky-rav:
@trozet PTAL
Thanks!

ricky-rav:
@kyrtapz when you have time, could you take a look? Thanks!

PR #5416, opened 2025-07-25
Move masquarade SNAT to the nftables

npinaeva:
## 📑 Description

Based on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5140

Move SNAT for management port from ovn routers to the nftables.
We are on a mission to eventually move all SNATs to nftables, this is a first step.
This change is required for the future layer2 topology change to avoid disruption.
Upgrade logic ensures that we only keep the conditional snat if the network
already existed and pods are already wired. During an upgrade, the node
will be drained and pods will be moved. When OVNK comes up at that point
we no longer have pods on the node, and now we can remove this SNAT in
OVN, and use the one in the host nftables.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Improved tracking of all node IPs (IPv4 and IPv6) in nftables sets for enhanced network rule management.
  * Introduced explicit masquerade nftables chains and packet marks for user-defined networks.
  * Added node pod presence checks and conditional egress SNAT management in secondary network controllers.
  * Added helper functions to detect pods on secondary networks and manage egress SNAT rules per node.

* **Bug Fixes**
  * Strengthened equality checks for network rules by including mark and family fields to ensure precise rule reconciliation.

* **Tests**
  * Enhanced test coverage for nftables sets, masquerade IP rules, and stale SNAT entry cleanup in Layer2 and Layer3 secondary network controllers.
  * Updated test setups to reflect new masquerade subnet configurations and network ID annotations.
  * Removed obsolete NAT entries from test databases to align with updated logic.
  * Added new tests for cleanup of stale egress SNAT entries in secondary network controllers.

* **Chores**
  * Removed special-case IPv6 test logic for Layer2 networks in route advertisement tests.
  * Refined comments and test expectations for clarity in masquerade and SNAT-related tests.

### DIFF
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -37,6 +37,10 @@ const (
 	// pktMarkBase is the base value for packet mark assigned to user defined networks
 	// Each network has a packet mark equal to base + network-id
 	pktMarkBase = 4096
+	// masqCTMarkBase is the base value for conntrack mark assigned to user defined networks
+	// Each network has a masquerade conntrack mark equal to base + network-id
+	// It limits the amount of supported UDNs to 100000 and is used by nftables rules
+	masqMarkBase = pktMarkBase + 100000
 	// waitForPatchPortTimeout is the maximum time we wait for a UDN's patch
 	// port to be created by OVN.
 	waitForPatchPortTimeout = 30 * time.Second
@@ -65,6 +69,9 @@ type UserDefinedNetworkGateway struct {
 	// pktMark hold the packets mark value for this network
 	// which is used for directing traffic towards the UDN
 	pktMark uint
+	// masqPktMark holds the masquerade packets mark value for this network
+	// which used by nftables rules to mark packets that are going to be masqueraded
+	masqPktMark uint
 	// v4MasqIPs holds the IPv4 masquerade IPs for this network
 	v4MasqIPs *udn.MasqueradeIPs
 	// v6MasqIPs holds the IPv6 masquerade IPs for this network
@@ -103,6 +110,7 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
 	networkID := netInfo.GetNetworkID()
 	masqCTMark := ctMarkUDNBase + uint(networkID)
 	pktMark := pktMarkBase + uint(networkID)
+	masqPktMark := masqMarkBase + uint(networkID)
 	if config.IPv4Mode {
 		v4MasqIPs, err = udn.AllocateV4MasqueradeIPs(networkID)
 		if err != nil {
@@ -138,6 +146,7 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
 		vrfManager:       vrfManager,
 		masqCTMark:       masqCTMark,
 		pktMark:          pktMark,
+		masqPktMark:      masqPktMark,
 		v4MasqIPs:        v4MasqIPs,
 		v6MasqIPs:        v6MasqIPs,
 		gateway:          gw,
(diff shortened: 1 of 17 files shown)


go-controller/pkg/node/gateway_udn.go
```diff
@@ -147,6 +147,164 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
...
+		counterIfDebug = "counter"
+	}
+	mpxInterfaceName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(udng.GetNetworkID()))
+
+	nft, err := nodenft.GetNFTablesHelper()
+	if err != nil {
+		return err
+	}
+	tx := nft.NewTransaction()
+	// TODO: check priority here and make sure there wont be collisions with egress service/ip.
```
npinaeva: I think it should be fine, @trozet did you manage to think of any conflicts?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -147,6 +147,164 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
...
+
+	podSubnets, err := udng.getLocalSubnets()
+	if err != nil {
+		return fmt.Errorf("failed to get pod subnets for network %s: %w", udng.GetNetworkName(), err)
+	}
+
+	if len(podSubnets) == 0 {
+		return fmt.Errorf("cannot determine pod subnets while configuring masquerade nftables chain for network: %s", udng.GetNetworkName())
+	}
+	// TODO agree on the right offset = the number of supported networks
```
npinaeva: do we go with 100K as an offset and as a network limit?

go-controller/pkg/node/secondary_node_network_controller_test.go
```diff
@@ -459,7 +459,9 @@ var _ = Describe("SecondaryNodeNetworkController: UserDefinedPrimaryNetwork Gate
 					udnRules = append(udnRules, rule)
 				}
 			}
-			Expect(udnRules).To(HaveLen(3))
+
+			// 3 rules per ip family: see constructUDNVRFIPRules for more details
+			Expect(udnRules).To(HaveLen(6))
```
npinaeva: For anyone thinking: we only added 2 new rules, why is it +3 in the test?
It is because we never actually were adding ipv6 pktMark rule before (due to `areNetlinkRulesEqual` bug that is fixed now) and expected rule number shouldn've been 4 before this change (2 for ipv4 and 2 for ipv6 makes sense right?)

trozet:
@npinaeva can you rebase now that the other PR merged? I'll review tonight

npinaeva:
> @npinaeva can you rebase now that the other PR merged? I'll review tonight

The other PR hasn't merged yet, but the failures were for another reason.
I have added the fix with the last commit, but now it seems that having SNAT on the node side is uglier than on the OVN side (we need to add new nft sets with node IPs and then mark packets here and there), which brought me to the original need for moving SNATs to the host (because this PR was based on assumption that we agreed to move SNATs to the host anyway, but now I am not sure if that is the case).

The only reason I managed to understand by now is that we **think** conditional SNAT breaks HW offload, but I don't think this is confirmed (there is a flag `options:ct-commit-all="true"` that could be helpful, talking to Han now to figure out if that is the case). Are there any other reasons to move SNAT?

trozet:
Oh looks like this ct-commit-all is a new feature in ovn:
https://github.com/ovn-org/ovn/blob/ab19375413cfc6429172d3101f125a8426a20b25/NEWS#L128

That would be great if it fixes the offload problem. We still need to figure out the migration of snat for layer 2 upgrade.

npinaeva:
> Oh looks like this ct-commit-all is a new feature in ovn: https://github.com/ovn-org/ovn/blob/ab19375413cfc6429172d3101f125a8426a20b25/NEWS#L128
> 
> That would be great if it fixes the offload problem. We still need to figure out the migration of snat for layer 2 upgrade.

Right, but I think migration of snat  for layer 2 upgrade is not a good enough reason for SNAT move, unless there is something else I am missing, wdyt? I mean I could just migrate it to the new transit router.

trozet:
> > Oh looks like this ct-commit-all is a new feature in ovn: https://github.com/ovn-org/ovn/blob/ab19375413cfc6429172d3101f125a8426a20b25/NEWS#L128
> > That would be great if it fixes the offload problem. We still need to figure out the migration of snat for layer 2 upgrade.
> 
> Right, but I think migration of snat for layer 2 upgrade is not a good enough reason for SNAT move, unless there is something else I am missing, wdyt? I mean I could just migrate it to the new transit router.

Agree. We just need to figure out our disruption story there.

tssurya:
> > Oh looks like this ct-commit-all is a new feature in ovn: https://github.com/ovn-org/ovn/blob/ab19375413cfc6429172d3101f125a8426a20b25/NEWS#L128
> > That would be great if it fixes the offload problem. We still need to figure out the migration of snat for layer 2 upgrade.
> 
> Right, but I think migration of snat for layer 2 upgrade is not a good enough reason for SNAT move, unless there is something else I am missing, wdyt? I mean I could just migrate it to the new transit router.

++++++ agree

PR #5441, opened 2025-08-04
e2e, kv: Print vm run command error result

qinqon:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


test/e2e/kubevirt/console.go
```diff
@@ -94,12 +94,12 @@ func runCommand(virtctlPath string, vmi *v1.VirtualMachineInstance, command stri
 			},
 		}},
 	}, timeout)
-	if err != nil {
```
kyrtapz, line 97: is `results` always not nil and has at least one element even if err is not nil? you are not checking for it with the new change.

kyrtapz (changes requested):

test/e2e/kubevirt/console.go
qinqon: Yes, there is always something if you have run a command, it has multiple just for the case where you run more than one command, but since we are running at least one always there is always one there.

Issue #5435, opened 2025-07-30
flaky: [It] using ClusterUserDefinedNetwork CR, pods in different namespaces, should communicate over localnet topology


dave-tucker:
### Which jobs are flaking?

Seen here first:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5320#issuecomment-3133223858

Then here:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16619770678/job/47023257809?pr=5432

### Which tests are flaking?

`Network Segmentation: Localnet [It] using ClusterUserDefinedNetwork CR, pods in different namespaces, should communicate over localnet topology`

### Since when has it been flaking?

29 Jul 2025

ormergi:
In both recorded incidents the mentioned test failed early, on waiting for the CUDN CR readiness, we see the status is not updated yet and the test time out after 5s.

Looking into both incidents, ovnkube-cluster-manager get into unhealthy state.
There is a blip in the log around the time the test wait for CR readiness, followed by the following error and  informers crash and restart:
```
W0730 12:15:58.734336      20 reflector.go:492] pkg/crd/egressqos/v1/apis/informers/externalversions/factory.go:140: watch of *v1.EgressQoS ended with: an error on the server ("unable to decode an event from the watch stream: http2: client connection lost") has prevented the request from succeeding
W0730 12:15:58.734336      20 reflector.go:492] pkg/crd/egressqos/v1/apis/informers/externalversions/factory.go:140: watch of *v1.EgressQoS ended with: an error on the server ("unable to decode an event from the watch stream: http2: client connection lost") has prevented the request from succeeding
```
[job](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16619770678/job/47023257809?pr=5432) 
Result in outage of ovn-control-plan around the time test waits on the CR readiness.

In both incident some part of a test form egress IP suite ran although marked as skipped.

--- 
More details here:
On failing job https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16619770678/job/47023257809?pr=5432
First time the test failed TS 12:15:28
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16619770678/job/47023257809?pr=5432#step:15:13203
On t-5s we see the CR creation https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16619770678/job/47023257809?pr=5432#step:15:13202
[...]

PR #5447, opened 2025-08-05
ovnkube: add error logging on exit

kyrtapz:
Add error logging on exit to avoid a scenario where ovnkube exits with an error without providing any info in the logs.


booxter (approved):

RamLavi (approved):
/lgtm

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -110,7 +110,11 @@ func main() {
 	ctx := context.Background()
 	ctx, cancel := context.WithCancel(ctx)
 	c.Action = func(ctx *cli.Context) error {
```
tssurya: my understanding is `c.Action` get's triggered via 
```
	if err := c.RunContext(ctx, os.Args); err != nil {
		klog.Exit(err)
	}
```
where we print the error and exit.
@kyrtapz will look into why we don't print the error there and the weirdness its causing.
but this fix is not necessarily wrong - just want to understand why the existing klog.Exit doesn't work and why we need another klog inside Action func.

tssurya (changes requested):
had a question inline

go-controller/cmd/ovnkube/ovnkube.go
kyrtapz: Great spot @tssurya. The issue was a bit more nuanced and I updated the PR description.
Basically if the metrics server ovs client failed it exited early without waiting for the other go routines/returning other errors.
This should be fixed with the current changes in this pr.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -579,21 +579,23 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
 	if config.OvnKubeNode.Mode != types.NodeModeDPUHost && config.Metrics.OVNMetricsBindAddress != "" {
 		metricsScrapeInterval := 30
-		defer cancel()
 
 		if ovsClient == nil {
 			ovsClient, err = libovsdb.NewOVSClient(ctx.Done())
 			if err != nil {
-				return fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
+				ovsCLIErr = fmt.Errorf("failed to initialize libovsdb vswitchd client: %w", err)
+				cancel()
```
tssurya, line 587: oh we were returning here from entire `runOvnKube`... w/o even handling the errors from other goroutines!

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -579,21 +579,23 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
 			}
 		}
-		if config.Metrics.ExportOVSMetrics {
-			metrics.RegisterOvsMetricsWithOvnMetrics(ovsClient, metricsScrapeInterval, ctx.Done())
+		if ovsClient != nil {
```
tssurya, line 590: do we need some check for if ovsClient is nil then we also have a ovsCLI failure propagated saying client was nil? or is that an overkill?

tssurya (approved):
one question inline otherwise lgtm

go-controller/cmd/ovnkube/ovnkube.go
kyrtapz: There is no way for ovsClient to  be nil if ovsCLIErr is not so I think we are good.

go-controller/cmd/ovnkube/ovnkube.go
kyrtapz: yep :(

go-controller/cmd/ovnkube/ovnkube.go
tssurya: awesome I thought the same but I asked cause you were specially checking if client  != nil so wasn't aware when and if that could even happen?

kyrtapz: Because I removed the early exit there is a chance that ovsClient will be nil later, but not without `ovsCLIErr` being set so we should be good.

tssurya:
the two CI failures are known flakes

1. https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16774079096/job/47498031767?pr=5447 is https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5456
2. https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16774079096/job/47498031773?pr=5447 is https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5435

go-controller/cmd/ovnkube/ovnkube.go
tssurya: oh yesss makes sense... I forgot we'd be continuing now that we don't return. Resolving.

PR #5450, opened 2025-08-05
kind.sh: Allow to override duration to sleep at the end

booxter:
For example, to avoid sleeping, set KIND_SETTLE_DURATION=0.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


contrib/kind-common
```diff
@@ -35,6 +35,7 @@ if_error_exit() {
 set_common_default_params() {
   KIND_IMAGE=${KIND_IMAGE:-kindest/node}
   K8S_VERSION=${K8S_VERSION:-v1.32.3}
+  KIND_SETTLE_DURATION=${KIND_SETTLE_DURATION:-30}
 }
```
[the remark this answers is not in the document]
booxter: This script is for devs only; not much harm if they see a failure on invalid variable set.

contrib/kind-common
```diff
@@ -35,6 +35,7 @@ if_error_exit() {
 set_common_default_params() {
   KIND_IMAGE=${KIND_IMAGE:-kindest/node}
   K8S_VERSION=${K8S_VERSION:-v1.32.3}
+  KIND_SETTLE_DURATION=${KIND_SETTLE_DURATION:-30}
```
trozet, line 38: so I added this sleep a long time ago, before IC when it would take a little while for things to converge with raft and OVNK was slower. I think we were seeing issues in CI where tests would start executing and the system would still be churning. I think it might be a good exercise to play with lowering the default value in the future and seeing if it affects CI results at all (but our that would mean our CI needs to be stable).

trozet (dismissed):

trozet:
@booxter can you rebase please?

contrib/kind-common
booxter: Right. For now I'm just trying to speed up feedback loop in my development where I use the kind.sh script  to respin images. Separately, we can consider lowering defaults up to zero here in CI.

trozet (approved):

Issue #5436, opened 2025-07-31
Load Balancer Service Tests with MetalLB [It] Should ensure load balancer service works with pmtud


npinaeva:
### Which jobs are flaking?

control-plane
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16631051829/job/47099375650?pr=5396

### Which tests are flaking?

Load Balancer Service Tests with MetalLB [It] Should ensure load balancer service works with pmtud

### Since when has it been flaking?

idk

### Reason for failure (if possible)

``` 
 STEP: all 3 nodeIP routes are advertised correctly by metalb BGP routes @ 07/31/25 09:24:41.789
  I0731 09:24:41.850354 75508 service.go:1802] Routes in FRR 192.168.10.0 nhid 57 proto bgp metric 20 
  	nexthop via 172.18.0.3 dev eth0 weight 1 
  	nexthop via 172.18.0.4 dev eth0 weight 1 

  I0731 09:24:43.901955 75508 service.go:1802] Routes in FRR 192.168.10.0 nhid 57 proto bgp metric 20 
  	nexthop via 172.18.0.3 dev eth0 weight 1 
  	nexthop via 172.18.0.4 dev eth0 weight 1 

  I0731 09:24:45.956951 75508 service.go:1802] Routes in FRR 192.168.10.0 nhid 57 proto bgp metric 20 
  	nexthop via 172.18.0.3 dev eth0 weight 1 
  	nexthop via 172.18.0.4 dev eth0 weight 1 

  I0731 09:24:48.011748 75508 service.go:1802] Routes in FRR 192.168.10.0 nhid 57 proto bgp metric 20 
  	nexthop via 172.18.0.3 dev eth0 weight 1 
  	nexthop via 172.18.0.4 dev eth0 weight 1 

  I0731 09:24:50.065337 75508 service.go:1802] Routes in FRR 192.168.10.0 nhid 57 proto bgp metric 20 
  	nexthop via 172.18.0.3 dev eth0 weight 1 
  	nexthop via 172.18.0.4 dev eth0 weight 1 

  I0731 09:24:52.118940 75508 service.go:1802] Routes in FRR 192.168.10.0 nhid 57 proto bgp metric 20 
  	nexthop via 172.18.0.3 dev eth0 weight 1 
  	nexthop via 172.18.0.4 dev eth0 weight 1 

[...]

PR #5452, opened 2025-08-06
Prevent UDN deletion race condition by checking controller state

kyrtapz:
Adds `GetActiveNetwork` method to verify network controllers are stopped before allowing UDN/CUDN deletion, preventing race conditions in a scenario where a network is recreated before the cluster-manager controller stopped running.
Replaces custom condition update logic with standard k8s `meta.SetStatusCondition` to avoid updating condtions even if they haven't changed. Use rate limiting instead of fixed interval re-queuing. 

## Summary by CodeRabbit

* **New Features**
  * Prevent deletion of user-defined network resources while their network controllers are active.
  * Added ability to query whether a network controller is active.
  * Use rate-limited retries during reconciliation and modernized status condition handling.

* **Bug Fixes**
  * More reliable removal of network resources by detecting active controllers.
  * Consistent, sorted pod lists in error messages.

* **Tests**
  * Tests updated to use the network-manager interface for controller scenarios.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -38,6 +38,7 @@ import (
 	userdefinednetworkinformer "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/informers/externalversions/userdefinednetwork/v1"
 	userdefinednetworklister "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/listers/userdefinednetwork/v1"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/metrics"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
 
@@ -72,6 +73,8 @@ type Controller struct {
 	// trying to create an object with the same name.
 	createNetworkLock sync.Mutex
 
+	networkManager networkmanager.Interface
+
 	udnClient         userdefinednetworkclientset.Interface
 	udnLister         userdefinednetworklister.UserDefinedNetworkLister
 	cudnLister        userdefinednetworklister.ClusterUserDefinedNetworkLister
@@ -93,6 +96,7 @@ func New(
 	udnInformer userdefinednetworkinformer.UserDefinedNetworkInformer,
 	cudnInformer userdefinednetworkinformer.ClusterUserDefinedNetworkInformer,
 	renderNadFn RenderNetAttachDefManifest,
+	networkManager networkmanager.Interface,
 	podInformer corev1informer.PodInformer,
 	namespaceInformer corev1informer.NamespaceInformer,
 	eventRecorder record.EventRecorder,
@@ -108,6 +112,7 @@ func New(
 		renderNadFn:                 renderNadFn,
 		podInformer:                 podInformer,
 		namespaceInformer:           namespaceInformer,
+		networkManager:              networkManager,
 		networkInUseRequeueInterval: defaultNetworkInUseCheckInterval,
 		namespaceTracker:            map[string]sets.Set[string]{},
 		eventRecorder:               eventRecorder,
@@ -410,6 +415,12 @@ func (c *Controller) syncUserDefinedNetwork(udn *userdefinednetworkv1.UserDefine
 				return nil, fmt.Errorf("failed to delete NetworkAttachmentDefinition [%s/%s]: %w", udn.Namespace, udn.Name, err)
 			}
 
+			// Ensure that the network controller is stopped(GetActiveNetwork returns nil) before allowing the UDN
+			// to be removed.
+			if c.networkManager.GetActiveNetwork(util.GenerateUDNNetworkName(udn.Namespace, udn.Name)) != nil {
+				return nil, &networkInUseError{err: fmt.Errorf("cannot remove UDN, controller for network %s is still running", util.GenerateUDNNetworkName(udn.Namespace, udn.Name))}
+			}
+
 			controllerutil.RemoveFinalizer(udn, template.FinalizerUserDefinedNetwork)
 			udn, err := c.udnClient.K8sV1().UserDefinedNetworks(udn.Namespace).Update(context.Background(), udn, metav1.UpdateOptions{})
 			if err != nil {
@@ -580,6 +591,12 @@ func (c *Controller) syncClusterUDN(cudn *userdefinednetworkv1.ClusterUserDefine
 
 			if len(errs) > 0 {
 				return nil, errors.Join(errs...)
+			}
+
+			// Ensure that the network controller is stopped(GetActiveNetwork returns nil) before allowing the cUDN
+			// to be removed.
+			if c.networkManager.GetActiveNetwork(util.GenerateCUDNNetworkName(cudn.Name)) != nil {
+				return nil, &networkInUseError{err: fmt.Errorf("cannot remove cluster UDN, controller for network %s is still running", util.GenerateCUDNNetworkName(cudn.Name))}
 			}
 
 			var err error
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -15,6 +15,7 @@ import (
 	netv1lister "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -87,8 +88,6 @@ type Controller struct {
 	eventRecorder               record.EventRecorder
 }
 
-const defaultNetworkInUseCheckInterval = 1 * time.Minute
-
 func New(
 	nadClient netv1clientset.Interface,
 	nadInfomer netv1infomer.NetworkAttachmentDefinitionInformer,
@@ -104,18 +103,17 @@ func New(
 	udnLister := udnInformer.Lister()
 	cudnLister := cudnInformer.Lister()
 	c := &Controller{
-		nadClient:                   nadClient,
-		nadLister:                   nadInfomer.Lister(),
-		udnClient:                   udnClient,
-		udnLister:                   udnLister,
-		cudnLister:                  cudnLister,
-		renderNadFn:                 renderNadFn,
-		podInformer:                 podInformer,
-		namespaceInformer:           namespaceInformer,
-		networkManager:              networkManager,
-		networkInUseRequeueInterval: defaultNetworkInUseCheckInterval,
-		namespaceTracker:            map[string]sets.Set[string]{},
-		eventRecorder:               eventRecorder,
+		nadClient:         nadClient,
+		nadLister:         nadInfomer.Lister(),
+		udnClient:         udnClient,
+		udnLister:         udnLister,
+		cudnLister:        cudnLister,
+		renderNadFn:       renderNadFn,
+		podInformer:       podInformer,
+		namespaceInformer: namespaceInformer,
+		networkManager:    networkManager,
+		namespaceTracker:  map[string]sets.Set[string]{},
+		eventRecorder:     eventRecorder,
 	}
 	udnCfg := &controller.ControllerConfig[userdefinednetworkv1.UserDefinedNetwork]{
 		RateLimiter:    workqueue.DefaultTypedControllerRateLimiter[string](),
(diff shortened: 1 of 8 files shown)


go-controller/pkg/networkmanager/api.go
```diff
@@ -42,6 +42,11 @@ type Interface interface {
 	// DoWithLock takes care of locking and unlocking while iterating over all role primary user defined networks.
 	DoWithLock(f func(network util.NetInfo) error) error
 	GetActiveNetworkNamespaces(networkName string) ([]string, error)
+
+	// GetControllerNetwork returns the netinfo currently held by the controller for the given network.
+	// This may differ from the netinfo returned by GetNetwork which reflects the API state.
+	// Returns nil if there is no running controller for the provided network.
+	GetControllerNetwork(network string) util.NetInfo
```
jcaamano: `GetControllerNetwork` doesn't sound right to me. Why not `GetNetworkState` and place it just next to `GetNetwork`?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -410,6 +415,10 @@ func (c *Controller) syncUserDefinedNetwork(udn *userdefinednetworkv1.UserDefine
 				return nil, fmt.Errorf("failed to delete NetworkAttachmentDefinition [%s/%s]: %w", udn.Namespace, udn.Name, err)
 			}
 
+			if c.networkManager.GetControllerNetwork(util.GenerateUDNNetworkName(udn.Namespace, udn.Name)) != nil {
```
jcaamano: Do you think it would do good to have comments here. This can be a hotspot where people might look at it an ask "why are we waiting?"

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -389,7 +387,7 @@ func (c *Controller) reconcileUDN(key string) error {
 
 	var networkInUse *networkInUseError
 	if errors.As(syncErr, &networkInUse) {
-		c.udnController.ReconcileAfter(key, c.networkInUseRequeueInterval)
+		c.udnController.ReconcileRateLimited(key)
```
jcaamano, line 391: While the change seems reasonable, I am missing if there was an actual need to make the change or not. If not, did you check for all other uses of `ReconcileAfter` if they could reasonable be changed as well? I wouldn't like for `ReconcileAfter` to remain there if it is most likely an improper use most of the time.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -450,40 +448,40 @@ func (c *Controller) updateUserDefinedNetworkStatus(udn *userdefinednetworkv1.Us
...
+	udn, err = c.udnClient.K8sV1().UserDefinedNetworks(udn.Namespace).ApplyStatus(context.Background(), udnApplyConf, opts)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
 		}
-		klog.Infof("Updated status UserDefinedNetwork [%s/%s]", udn.Namespace, udn.Name)
+		return fmt.Errorf("failed to update UserDefinedNetwork status: %w", err)
 	}
-
+	klog.Infof("Updated status UserDefinedNetwork [%s/%s]", udn.Namespace, udn.Name)
```
jcaamano, line 481: noop comment: this triggers my "excessive logging" thought

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -450,40 +448,40 @@ func (c *Controller) updateUserDefinedNetworkStatus(udn *userdefinednetworkv1.Us
...
 	return nil
 }
 
-func newNetworkCreatedCondition(nad *netv1.NetworkAttachmentDefinition, syncError error) *metav1.Condition {
+func newNetworkCreatedCondition(nad *netv1.NetworkAttachmentDefinition, syncError error) metav1.Condition {
```
jcaamano: nit: seems unnecessary change, wouldn't we benefit from no copying with no drawbacks?

go-controller/pkg/networkmanager/api.go
kyrtapz: Can you elaborate what is wrong with the name? Personally I find it less generic/vague than `GetNetworkState`.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
kyrtapz: Great idea, will do.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
kyrtapz: > While the change seems reasonable, I am missing if there was an actual need to make the change or not.

Without this change it can be the case that in normal operation the controller is still running when the UDN delete is processed which would cause a huge delay of 1 minute.

> If not, did you check for all other uses of ReconcileAfter if they could reasonable be changed as well? I wouldn't like for ReconcileAfter to remain there if it is most likely an improper use most of the time.

I don't like `ReconcileAfter` in principle so I think it is a good idea to get rid of it. Actually I am not a fan of not doing rate limited enqueuing by default but that is a major change I don't want to get into in this pr. 
The only other component using `ReconcileAfter` is the route_import controller and I would like to hear from you whether it is desired there.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
kyrtapz: It was there before and I just moved it. I don't entirely agree it is a noop as there were instances where these types of logs were detrimental in RCAing races with API updates. I do agree it shares the annoying trend in our repo of logging that we have updated something without stating what was the change.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
kyrtapz: I don't think this has any meaningful impact but I don't mind changing it back. I think  I modified it because `meta.SetStatusCondition` takes a value.

go-controller/pkg/networkmanager/api.go
jcaamano: I can't outright understand what it tries to say. What is `Controller` referring to? We are not getting any controller.

kyrtapz: We are getting the network used by the controller but I understand that it can be unclear. I don't see how `GetNetworkState` is better though, maybe `GetNetworkControllerState`?
Would something like `GetControllerNetInfo` be a easier to understand?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: > I am not a fan of not doing rate limited enqueuing by default

+1

> The only other component using `ReconcileAfter` is the route_import controller and I would like to hear from you whether it is desired there.

For of its uses it is reasonable: when we receive a route update we will potentially receive more route updates immediately after that one so I wait a bit to reconcile.
For all its other uses, it falls in the category of your other comment that reconciles should be rate limited by default.

I guess no action on this front then for now.

> Without this change it can be the case that in normal operation the controller is still running when the UDN delete is processed which would cause a huge delay of 1 minute.

Not sure I fully understand.
`defaultNetworkInUseCheckInterval` is 1 minute so you are saying that is too much of a delay?
If you use the queue rate limiter the requeue can be immediate with no delay depending on what the controller might be doing so I guess that's the drawback?
Have you considered reducing `defaultNetworkInUseCheckInterval`?

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: with noop I mean I don't expect you to do anything, just a thought of mine.

However in regards to your comment, I think we should just make proper use of levels and be able to reproduce the issues. Logging on every update is too much logging in my view. If we did that everywhere it would be bad.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: up to you, just a nit.

go-controller/pkg/networkmanager/api.go
jcaamano: > network used by the controller

hmmm what controller? I am just missing to what controller you a referring to. Do you mean the one in `network_controller.go` in the package? I see that as an implementation detail of the package, the caller is just calling network manager, and it has no idea there is a controller in there or what can that possibly mean towards the functionality of the method. It is only meaningful if you are aware how the package is implemented and I had to think hard :p

`GetNetworkState` does look good to me, because if there is state it kinda means the network is active, and if not, it means the network is not active. But I can give you other ideas in order of preference: `GetActiveNetwork`, `GetActiveNetworkState`, `GetRunningNetworkState`, `GetRunningNetwork` ....

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: Also realized, if you are going to be calling `ReconcileRateLimited`, isn't just that the same as just returning the error and not calling anything at all there?

go-controller/pkg/networkmanager/api.go
kyrtapz: I think that `GetControllerNetwork` is less abstract but more clear to ovn-k developers. Going with `GetNetworkState`(a method of an interface called `Interface`) makes the code more abstract but less clear and can be kinda understood depending on who looks at it. If the abstraction is the main focus of this interface I can change it to `GetNetworkState` as it is vague enough to be interpreted as anything - I am not sure what a network state really is.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
kyrtapz: > Not sure I fully understand.
defaultNetworkInUseCheckInterval is 1 minute so you are saying that is too much of a delay?

It is yes. 

> If you use the queue rate limiter the requeue can be immediate with no delay depending on what the controller might be doing so I guess that's the drawback?

Only initially, if the reconciliation fails then the delay increases. Not sure what is the drawback you have in mind. 

> Have you considered reducing defaultNetworkInUseCheckInterval?

I don't see the benefit of using constant intervals, with rate limiting we retry with increasing intervals which is fast initially and then slower over time where the retry being successful decreases. 

> Also realized, if you are going to be calling ReconcileRateLimited, isn't just that the same as just returning the error and not calling anything at all there?

My understanding was that our plain reconciliation isn't rate limited?

go-controller/pkg/networkmanager/api.go
kyrtapz: Considering the already existing `GetActiveNetworkNamespaces` I would be inclined to go with `GetActiveNetwork` instead of `GetNetworkState`

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: > Only initially, if the reconciliation fails then the delay increases. Not sure what is the drawback you have in mind.

The drawback of retying before it can be reasonably expected for whatever cause to be to holding off the delete of the network to have cleared up. From the initial rate limit ~10ms to 1min there is a big gap.

> My understanding was that our plain reconciliation isn't rate limited?

A retry for an error is rate limited:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/68bd339b032cdd2881cf5fa3951c8dc2b04494f7/go-controller/pkg/controller/controller.go#L258

So just returning the error would have the same effect.

kyrtapz: > A retry for an error is rate limited:

I've missed that initially, sorry. I don't think the behavior is the same in this case since this controller(as most of our controllers) stops retrying after 15 tries. For this specific error we want to keep retrying with the hopes that:
- The controller stopped(very likely)
- The pods were eventually removed(less likely and can take a long time).

kyrtapz: Considering that `networkInUse` is the only condition we want to retry forever i would like to keep the current apporach, wdyt?

kyrtapz:
[Changes](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/aea1e27b7406fb7a362f02f1d87abf6c9675a364..2f1987c623d208e0b403b8e6aa77a8055eb38bbb):
- Added some comments
- Renamed GetControllerNetwork to GetActiveNetwork and moved it right under GetNetwork
- Reverted back to returning a pointer in newNetworkCreatedCondition.

go-controller/pkg/networkmanager/api.go
kyrtapz: Done.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
jcaamano: If it were me I would just set the controller config max attempts to `InfiniteAttempts`.

But I hear you and this now becomes a nit, so I am ok with your decision and hope that if we ever do it this doesn't go missed to be cleaned up, even if harmless.

kyrtapz:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16901696473/job/47884107325?pr=5452 failed on a known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5436

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -410,6 +414,12 @@ func (c *Controller) syncUserDefinedNetwork(udn *userdefinednetworkv1.UserDefine
 				return nil, fmt.Errorf("failed to delete NetworkAttachmentDefinition [%s/%s]: %w", udn.Namespace, udn.Name, err)
 			}
 
+			// Ensure that the network controller is stopped(GetActiveNetwork returns nil) before allowing the UDN
+			// to be removed.
```
jcaamano, line 418: Can you enrich this comment a bit. I will make a proposal.
```
This allows for annotations of the old network to be removed before a potential new network with the same name is added preventing the new network to use the old annotations on the node side
```

PR #5463, opened 2025-08-08
SNAT traffic from advertised UDNs towards UDN enabled default services

jcaamano:
Just as we currently do with traffic towards nodes.

Specifically this allows for networks advertised with a VRF-Lite configuration with a subnet overlap to reach these services. Otherwise the return path could hit an ip rule corresponding to a different advertised network forwarding it to an inappropriate destination.

This undoes the revert of from https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5284. This PR was removing the ip rule for a VRF-Lite configured subnet, precisely to allow overlaps as well. However doing that without the SNAT implemented in this PR broke VRF-Lite configured network to kapi/dns traffic. For the original commit,  some minor conflicts are resolved here related to the fact that networks advertised on the default VRF only had the network subnet ip rule at the time, but since then that changed to having both the network subnet and the masquerade ip rule. So when this commit was implemented we had
```
// When a network is not advertised, an example of the rules
// we set for it is:
// 2000:	from all fwmark 0x1001 lookup 1007
// 2000:	from all to 169.254.0.12 lookup 1007
//
// When a network is advertised, an example of the rules
// we set for it is:
// 2000:	from all fwmark 0x1001 lookup 1007
// 2000:	from all to 10.132.0.0/14 lookup 1007
```
but then it changed to
```
// When a network is not advertised, an example of the rules
// we set for it is:
// 2000:	from all fwmark 0x1001 lookup 1007
// 2000:	from all to 169.254.0.12 lookup 1007
//
// When a network is advertised, an example of the rules
// we set for it is:
// 2000:	from all fwmark 0x1001 lookup 1007
// 2000:	from all to 169.254.0.12 lookup 1007
// 2000:	from all to 10.132.0.0/14 lookup 1007
```
making this commit for VRF-Lite a special case of the first scenario, making things a bit more nuanced:
```
// When a network is not advertised on the default VRF, an example of the rules
// we set for it is:
// 2000:	from all fwmark 0x1001 lookup 1007
// 2000:	from all to 169.254.0.12 lookup 1007
//
// When a network is advertised on the default VRF, an example of the rules
// we set for it is:
// 2000:	from all fwmark 0x1001 lookup 1007
// 2000:	from all to 10.132.0.0/14 lookup 1007
// 2000:	from all to 169.254.0.12 lookup 1007
```

## Summary by CodeRabbit

* **New Features**
  * UDN egress SNAT now uses address-set–based destination matching for more accurate outbound NAT.

* **Refactor**
  * Centralized, cached network-advertisement state for consistent reconciliation and VRF‑aware decisions.
  * VRF‑Lite aware default-route handling to avoid incorrect routes when advertised to non‑default VRFs.
  * Isolation rule management split into explicit add/delete flows for clearer rule lifecycle.

* **Tests**
  * Expanded dual‑stack coverage, added default vs non‑default VRF tests, and updated CIDR/mask expectations.

### DIFF
--- a/go-controller/pkg/node/gateway_udn.go
+++ b/go-controller/pkg/node/gateway_udn.go
@@ -89,6 +89,10 @@ type UserDefinedNetworkGateway struct {
 
 	// gwInterfaceIndex holds the link index of gateway interface
 	gwInterfaceIndex int
+
+	// save BGP state at the start of reconciliation loop run to handle it consistently throughout the run
+	isNetworkAdvertisedToDefaultVRF bool
+	isNetworkAdvertised             bool
 }
 
 func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeLister listers.NodeLister,
@@ -225,18 +229,18 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
 		return fmt.Errorf("could not add VRF %s routes for network %s, err: %v", vrfDeviceName, udng.GetNetworkName(), err)
 	}
 
-	isNetworkAdvertised := util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)
+	udng.updateAdvertisementStatus()
 
 	// create the iprules for this network
-	if err = udng.updateUDNVRFIPRules(isNetworkAdvertised); err != nil {
+	if err = udng.updateUDNVRFIPRules(); err != nil {
 		return fmt.Errorf("failed to update IP rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
-	if err = udng.updateAdvertisedUDNIsolationRules(isNetworkAdvertised); err != nil {
+	if err = udng.updateAdvertisedUDNIsolationRules(); err != nil {
 		return fmt.Errorf("failed to update isolation rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
-	if err := udng.updateUDNVRFIPRoute(isNetworkAdvertised); err != nil {
+	if err := udng.updateUDNVRFIPRoute(); err != nil {
 		return fmt.Errorf("failed to update ip routes for network %s: %w", udng.GetNetworkName(), err)
 	}
 
@@ -314,18 +318,16 @@ func (udng *UserDefinedNetworkGateway) DelNetwork() error {
 		}
 	}
 
-	if util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name) {
-		err := udng.updateAdvertisedUDNIsolationRules(false)
-		if err != nil {
-			return fmt.Errorf("failed to remove advertised UDN isolation rules for network %s: %w", udng.GetNetworkName(), err)
-		}
+	err := udng.deleteAdvertisedUDNIsolationRules()
+	if err != nil {
+		return fmt.Errorf("failed to remove advertised UDN isolation rules for network %s: %w", udng.GetNetworkName(), err)
 	}
 
 	if err := udng.delMarkChain(); err != nil {
 		return err
 	}
 	// delete the management port interface for this network
-	err := udng.deleteUDNManagementPort()
+	err = udng.deleteUDNManagementPort()
 	if err != nil {
 		return err
 	}
@@ -483,8 +485,7 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link) 
 
 	// Route2: Add default route: default via 172.18.0.1 dev breth0 mtu 1400
 	// necessary for UDN CNI and host-networked pods default traffic to go to node's gatewayIP
-	isNetworkAdvertised := util.IsPodNetworkAdvertisedAtNode(udng.NetInfo, udng.node.Name)
-	defaultRoute, err := udng.getDefaultRoute(isNetworkAdvertised)
+	defaultRoute, err := udng.getDefaultRouteExceptIfVRFLite()
 	if err != nil {
 		return nil, fmt.Errorf("unable to add default route for network %s, err: %v", udng.GetNetworkName(), err)
 	}
@@ -585,15 +586,7 @@ func (udng *UserDefinedNetworkGateway) computeRoutesForUDN(mpLink netlink.Link) 
 	return retVal, nil
 }
 
-func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool) ([]netlink.Route, error) {
-	vrfs := udng.GetPodNetworkAdvertisedOnNodeVRFs(udng.node.Name)
-	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
-	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
-	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
-	if isNetworkAdvertised && !slices.Contains(vrfs, types.DefaultNetworkName) {
-		return nil, nil
-	}
-
+func (udng *UserDefinedNetworkGateway) getDefaultRoute() ([]netlink.Route, error) {
 	networkMTU := udng.NetInfo.MTU()
 	if networkMTU == 0 {
 		networkMTU = config.Default.MTU
@@ -616,6 +609,16 @@ func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool)
 		})
 	}
 	return retVal, nil
+}
+
+func (udng *UserDefinedNetworkGateway) getDefaultRouteExceptIfVRFLite() ([]netlink.Route, error) {
+	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
+	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
+	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
+	if udng.isNetworkAdvertised && !udng.isNetworkAdvertisedToDefaultVRF {
+		return nil, nil
+	}
+	return udng.getDefaultRoute()
 }
 
 // getV4MasqueradeIP returns the V4 management port masqueradeIP for this network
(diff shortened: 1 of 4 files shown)


go-controller/pkg/node/gateway_udn.go
```diff
@@ -848,30 +851,40 @@ func (udng *UserDefinedNetworkGateway) updateUDNVRFIPRules(isNetworkAdvertised b
...
 		}
-		if err = udng.vrfManager.AddVRFRoutes(util.GetNetworkVRFName(udng.NetInfo), defaultRoute); err != nil {
-			return fmt.Errorf("error while adding default route to VRF %s corresponding to network %s, err: %v",
-				util.GetNetworkVRFName(udng.NetInfo), udng.GetNetworkName(), err)
+
+		if err = udng.vrfManager.AddVRFRoutes(vrfName, defaultRoute); err != nil {
+			return fmt.Errorf("failed to add default route to VRF %s for network %s: %v",
+				vrfName, udng.GetNetworkName(), err)
 		}
 	}
```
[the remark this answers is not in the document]
jcaamano: `vrfManager.DeleteVRFRoutes` ends up calling `route_manager.go` `netlinkDelRoute` which already tolerates "no such process"

go-controller/pkg/node/gateway_udn.go
```diff
@@ -618,6 +611,16 @@ func (udng *UserDefinedNetworkGateway) getDefaultRoute(isNetworkAdvertised bool)
...
 
+func (udng *UserDefinedNetworkGateway) getDefaultRouteExceptIfVRFLite() ([]netlink.Route, error) {
+	// If the network is advertised on a non default VRF then we should only consider routes received from external BGP
+	// device and not send any traffic based on default route similar to one present in default VRF. This is more important
+	// for VRF-Lite usecase where we need traffic to leave from vlan device instead of default gateway interface.
+	if udng.isNetworkAdvertised && !udng.isNetworkAdvertisedToDefaultVRF {
+		return nil, nil
+	}
+	return udng.getDefaultRoute()
+}
```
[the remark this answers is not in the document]
jcaamano: No one should be indexing a slice without having some information about its length.

go-controller/pkg/node/gateway_udn.go
```diff
@@ -89,6 +89,10 @@ type UserDefinedNetworkGateway struct {
 
 	// gwInterfaceIndex holds the link index of gateway interface
 	gwInterfaceIndex int
+
+	// save BGP state at the start of reconciliation loop run to handle it consistently throughout the run
+	isNetworkAdvertisedToDefaultVRF bool
+	isNetworkAdvertised             bool
 }
```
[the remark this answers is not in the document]
jcaamano: The whole `AddNetwork` function has an (implicit) contract of only being called once.

ricky-rav (dismissed):
I'm not a super expert in this part of the code base, but the changes do make sense to me.
/LGTM

jcaamano:
Missed the commit sign off.

tssurya:
Do we have E2E tests for:

1. pod advertised using VRFLite accessing kapi/dns in default network?
2. overlapping UDNs with /30 range beind advertised via default VRF where there are overlapping podIPs and then doing curl to kapi/dns?

jcaamano:
> 1. pod advertised using VRFLite accessing kapi/dns in default network?

[Yes](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/8f7fc5a1c8e493e2da8452204daf808ecfd79f50/test/e2e/route_advertisements.go#L1456)

> 2. overlapping UDNs with /30 range beind advertised via default VRF where there are overlapping podIPs and then doing curl to kapi/dns?

No, our current infra provider implementation of docker isolated networks does not allow for overlap as its isolation is based on iptables. To be able to configure overlaps we will need to implement something else, for example macvlan networks. Out of the scope of this PR.


However I have [this DS PR](https://github.com/openshift/ovn-kubernetes/pull/2705) for which the BGP LGW job passes whereas before it would fail with Peng's PR that I am re-reverting here. We discussed in the past how that would be our test.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -822,12 +824,41 @@ func (bsnc *BaseSecondaryNetworkController) buildUDNEgressSNAT(localPodSubnets [
...
+		// networks towards the default network service cluster IPs
+		// accessible from UDNs: we want the reply traffic to hit the
+		// masquerade IP rule rather than the UDN subnet ip rule to allow
+		// for overlaps in VRF-Lite configurations
+		svcIPsASIDs := udnenabledsvc.GetAddressSetDBIDs()
+		svcIPsAS, err = bsnc.addressSetFactory.GetAddressSet(svcIPsASIDs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get address set with IDs %v: %w", svcIPsASIDs, err)
+		}
+	}
```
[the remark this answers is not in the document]
jcaamano: When the address sets don't exist, currently `GetAddressSet` returns an error.

go-controller/pkg/ovn/base_network_controller_secondary.go
```diff
@@ -867,14 +897,25 @@ func getMasqueradeManagementIPSNATMatch(dstMac string) string {
...
+	for _, as := range addressSets {
+		asIPv4, asIPv6 := as.GetASHashNames()
+		switch {
+		case ipFamily == utilnet.IPv4 && asIPv4 != "":
+			asMatches = append(asMatches, fmt.Sprintf("ip4.dst == $%s", asIPv4))
+		case ipFamily == utilnet.IPv6 && asIPv6 != "":
+			asMatches = append(asMatches, fmt.Sprintf("ip6.dst == $%s", asIPv6))
+		}
+	}
+	switch len(asMatches) {
```
tssurya, line 913: this is a bit too overengineered for my taste, It took me five minutes to understand this logic - but maybe that's me being slow today
I am not sure why the refactor around moving 

ipv4ClusterNodeIPAS, ipv6ClusterNodeIPAS := addrSet.GetASHashNames()
to
asIPv4, asIPv6 := as.GetASHashNames()
within the function and then looping through two address sets and then looping at the len of the match - all to return a match string in the end was done.
but I won't nit pick here, because each person's coding style is differerent...I'm sure my if,else might be confusing for you

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya, line 1: so overall, for non-advertised udns, we already SNAT everything to masqueradeIP so we are good, here to kapi/dns we introduce a SNAT same as cSNAT - logic is good and as was discussed in design LGTM

go-controller/pkg/node/gateway_udn.go
```diff
@@ -900,39 +913,22 @@ func (udng *UserDefinedNetworkGateway) removeDefaultRouteFromVRF() error {
 //	   comment "advertised UDNs V4 subnets"
 //	   elements = { 10.10.0.0/16 comment "cluster_udn_l3network" }
 //	}
-func (udng *UserDefinedNetworkGateway) updateAdvertisedUDNIsolationRules(isNetworkAdvertised bool) error {
+func (udng *UserDefinedNetworkGateway) updateAdvertisedUDNIsolationRules() error {
+	switch {
```
tssurya, line 917: hehe I see you are a fan of switch-ing :)
but in this case I do like it - its cleaner separation for add v/s delete, nice

go-controller/pkg/node/gateway_udn_test.go
```diff
@@ -1774,6 +1786,169 @@ func TestConstructUDNVRFIPRulesPodNetworkAdvertised(t *testing.T) {
...
+				g.Expect(rule.Priority).To(Equal(test.deleteRules[i].priority))
+				g.Expect(rule.Table).To(Equal(test.deleteRules[i].table))
+				g.Expect(rule.Family).To(Equal(test.deleteRules[i].family))
+				g.Expect(*rule.Dst).To(Equal(test.deleteRules[i].dst))
+			}
+		})
+	}
+}
+
+func TestConstructUDNVRFIPRulesPodNetworkAdvertisedToNoneDefaultNetwork(t *testing.T) {
```
tssurya: nit: `TestConstructUDNVRFIPRulesPodNetworkAdvertisedToNonDefaultVRF`?
its very confusing to see pod network advertised to non default network?

go-controller/pkg/node/gateway_udn_test.go
```diff
@@ -1678,7 +1678,7 @@ func TestConstructUDNVRFIPRules(t *testing.T) {
 	}
 }
 
-func TestConstructUDNVRFIPRulesPodNetworkAdvertised(t *testing.T) {
+func TestConstructUDNVRFIPRulesPodNetworkAdvertisedToTheDefaultNetwork(t *testing.T) {
```
tssurya: nit: `TestConstructUDNVRFIPRulesPodNetworkAdvertisedToDefaultVRF` ?

tssurya (dismissed):
@jcaamano LGTM
regarding the two nits on unit test name did you intentionally make them default network instead of default VRF?

go-controller/pkg/ovn/base_network_controller_secondary.go
tssurya: resolving this. nothing to do

jcaamano: It all stems from the fact that I wanted `getClusterNodesDestinationBasedSNATMatch` to handle a variable number of address sets, which includes the possibility of having 0, 1, 2 or more. Today the method is only called with 1 or 2 (introduced in this PR) address sets, but if I am writing a method I just want it to work well to be usable in the future and I didn't see that as an over-engineering problem. The code just tries to build a valid match item regardless of the number of address sets provided.

jcaamano:
> regarding the two nits on unit test name did you intentionally make them default network instead of default VRF?

I just picked this up from peng's PR, they do sound strange. Let me rename them.

jcaamano:
> > regarding the two nits on unit test name did you intentionally make them default network instead of default VRF?
> 
> I just picked this up from peng's PR, they do sound strange. Let me rename them.

Done. I just changed the test names and rebased.

tssurya (dismissed):
thanks @jcaamano

jcaamano:
> thanks @jcaamano

Sorry, couldn't avoid addressing the comment by coderabbitai [above](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5463#discussion_r2276512096).

PR #5466, opened 2025-08-08
Default kind and kind-helm scripts to be interconnect

trozet:
- Users can still deploy legacy by passing new arguments "-ce" or "--enable-central"
- Users will get a new message at the end of script execution to inform them that they no longer need to pass "-ic" or "--enable-interconnect" if they pass these deprecated args
- Fixes bugs/incorrect bash in kind-helm script
- CI should be OK as it relies on the value of the environment variable OVN_ENABLE_INTERCONNECT for jobs

## Summary by CodeRabbit

* **New Features**
  * Interconnect mode is now enabled by default for deployments.
  * Added a new command-line option to explicitly select central (legacy) deployment mode.
  * Informational message provided when interconnect argument is used, indicating it is now the default.

* **Improvements**
  * Enhanced command-line argument handling for deployment mode selection with mutual exclusion enforcement.
  * Automatic installation of kubectl if missing (in one deployment script).
  * Improved error messages and help output for better clarity.

* **Bug Fixes**
  * Corrected numeric comparison for node count configuration.

* **Chores**
  * Removed redundant local kubectl setup in one deployment script.

### DIFF
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -845,3 +845,9 @@ EOF
     done
   fi
 }
+
+interconnect_arg_check() {
+  if [ "${IC_ARG_PROVIDED:-}" = "true" ]; then
+    echo "INFO: Interconnect mode is now the default mode, you do not need to use pass -ic or --enable-interconnect anymore"
+  fi
+}
--- a/contrib/kind-common
+++ b/contrib/kind-common
@@ -55,6 +55,37 @@ run_kubectl() {
   done
 }
 
+setup_kubectl_bin() {
+    ###########################################################################
+    # Description:                                                            #
+    # setup kubectl for querying the cluster                                  #
+    #                                                                         #
+    # Arguments:                                                              #
+    #   $1 - error message if not provided, it will just exit                 #
+    ###########################################################################
+    if [ ! -d "./bin" ]
+    then
+        mkdir -p ./bin
+        if_error_exit "Failed to create bin dir!"
+    fi
+
+    if [[ "$OSTYPE" == "linux-gnu" ]]; then
+        OS_TYPE="linux"
+    elif [[ "$OSTYPE" == "darwin"* ]]; then
+        OS_TYPE="darwin"
+    fi
+
+    pushd ./bin
+       if [ ! -f ./kubectl ]; then
+           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/${OS_TYPE}/${ARCH}/kubectl"
+           if_error_exit "Failed to download kubectl failed!"
+       fi
+    popd
+
+    chmod +x ./bin/kubectl
+    export PATH=${PATH}:$(pwd)/bin
+}
+
 command_exists() {
   cmd="$1"
   command -v ${cmd} >/dev/null 2>&1
--- a/contrib/kind-helm.sh
+++ b/contrib/kind-helm.sh
@@ -62,7 +62,7 @@ set_default_params() {
     KIND_NUM_MASTER=3
   fi
 
-  OVN_ENABLE_INTERCONNECT=${OVN_ENABLE_INTERCONNECT:-false}
+  OVN_ENABLE_INTERCONNECT=${OVN_ENABLE_INTERCONNECT:-true}
   if [ "$OVN_COMPACT_MODE" == true ] && [ "$OVN_ENABLE_INTERCONNECT" != false ]; then
      echo "Compact mode cannot be used together with Interconnect"
      exit 1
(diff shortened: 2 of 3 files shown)


contrib/kind-common
```diff
@@ -845,3 +846,9 @@ EOF
     done
   fi
 }
+
+interconnect_arg_check() {
+  if [ "${IC_ARG_PROVIDED:-}" = "true" ]; then
+    echo "INFO: Interconnect mode is now the default mode, you do not need to use pass -ic or --enable-interconnect anymore"
```
tssurya, line 882: love this! so what we do is we don't necessarily stop people from using the -ic flag its just that we give them this information that you don't need to... it's a no-op

contrib/kind-helm.sh
```diff
@@ -62,7 +62,7 @@ set_default_params() {
     KIND_NUM_MASTER=3
   fi
 
-  OVN_ENABLE_INTERCONNECT=${OVN_ENABLE_INTERCONNECT:-false}
+  OVN_ENABLE_INTERCONNECT=${OVN_ENABLE_INTERCONNECT:-true}
```
tssurya, line 65: ack so that's easier, we just always set it to true?

contrib/kind-helm.sh
```diff
@@ -193,7 +193,10 @@ parse_args() {
                                                   ;;
             -dns | --enable-dnsnameresolver )     OVN_ENABLE_DNSNAMERESOLVER=true
                                                   ;;
+            -ce | --enable-central )              OVN_ENABLE_INTERCONNECT=false
```
tssurya, line 196: I wanted to say ce (central-enable) being reverse of enable-central was confusing but i guess same pattern exists for -ic hehe

contrib/kind-helm.sh
```diff
@@ -408,7 +416,7 @@ create_ovn_kubernetes() {
       label_ovn_single_node_zones
       value_file="values-single-node-zone.yaml"
       ovnkube_db_options=""
-    elif [[ $KIND_NUM_NODES_PER_ZONE > 1 ]]; then
+    elif [[ $KIND_NUM_NODES_PER_ZONE -gt 1 ]]; then
```
tssurya, line 425: @coderabbitai: I am not a bash expert. Is ">" is not correct syntax?

tssurya: ah ack thanks for that explanation

tssurya: so this was a bad code haha nice catch @trozet !

tssurya (dismissed):
lgtm
let's see how CI fares...
ah never mind PR needs a rebase first for CI to run

tssurya:
```
+ echo 'INFO: Interconnect mode is now the default mode, you do not need to use pass -ic or --enable-interconnect anymore'
INFO: Interconnect mode is now the default mode, you do not need to use pass -ic or --enable-interconnect anymore
```

contrib/kind.sh
```diff
@@ -347,8 +316,11 @@ parse_args() {
                                                   ;;
             -adv | --advertise-default-network) ADVERTISE_DEFAULT_NETWORK=true
                                                   ;;
-            -ic | --enable-interconnect )       OVN_ENABLE_INTERCONNECT=true
-                                                ;;
+            -ce | --enable-central )              OVN_ENABLE_INTERCONNECT=false
+                                                  ;;
+            -ic | --enable-interconnect )         OVN_ENABLE_INTERCONNECT=true
+                                                  IC_ARG_PROVIDED=true
+                                                  ;;
```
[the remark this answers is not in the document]
tssurya: oh you mean when someone provides both `-ic` and `-ce` together?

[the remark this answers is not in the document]
tssurya: @trozet that might get a bit tricky lol should we add a check for if both are provided together we exit early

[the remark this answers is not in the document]
trozet: sure I will add that

contrib/kind-common
```diff
@@ -55,6 +56,37 @@ run_kubectl() {
...
+    pushd ./bin
+       if [ ! -f ./kubectl ]; then
+           curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/${OS_TYPE}/${ARCH}/kubectl"
+           if_error_exit "Failed to download kubectl failed!"
+       fi
+    popd
+
+    chmod +x ./bin/kubectl
+    export PATH=${PATH}:$(pwd)/bin
+}
```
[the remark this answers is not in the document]
tssurya: this code was not changed in this PR. its just getting moved around into a common place.

tssurya (approved):

PR #5470, opened 2025-08-11
docs: Document the release process

dave-tucker:
## 📑 Description

This document attempts to document a process for planning and delievering release in OVN-Kubernetes.

Fixes #

Still in draft form. Please ignore for now.

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-planning.md
+++ b/.github/ISSUE_TEMPLATE/release-planning.md
@@ -1,0 +1,42 @@
+---
+name: Release Planning
+about: File a release planning issue
+title: '[RELEASE] <title>'
+labels: kind/release
+---
+
+### Version
+<!--
+The version number follows the semantic versioning scheme.
+- Major version is incremented for any backwards incompatible changes.
+- Minor version is incremented for new features, but not backwards incompatible changes.
+- Patch version is incremented for bug fixes.
+-->
+
+### Important Dates
+<!--
+The release date is the date of the release.
+The code freeze date is the date of the code freeze.
+It is typically 2 weeks before the release date.
+-->
+
+**Release Date:**
+
+**Code Freeze Date:**
+
+### Major Changes
+<!--
+Please list the major changes that are planned for this release.
+This should include new features, bug fixes, and any other changes that are planned.
+-->
+
+### Checklist
+
+- [ ] Feature Complete
+- [ ] Code Frozen
+- [ ] (Optional) Release Candidate Tagged
+- [ ] Documentation Complete
+- [ ] Release Notes Complete
+- [ ] Release Tagged
+- [ ] GitHub Release Created
+- [ ] Release posted to mailing list and community slack channels
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,253 @@
+# Release Process
+
+## Introduction
+
+The process for managing enhancements (OKEPs), issues, and pull requests through an OVN-Kubernetes release is complex. This document describes the process in detail.
+
+## How often are releases?
+
+We release approximately once every 3 calendar months.
+
+## How do we decide on the release date?
+
+The release date is decided during the [community meeting] at the beginning of the release cycle.
+It will be approximately 3 calendar months from the previous release. The release date will not land on a Friday, holiday, or weekend.
+
+## Who is responsible for the release?
+
+The release coordinator is elected at the [community meeting] at the beginning of the release cycle.
+The release coordinator may be any volunteer from the community - a call for volunteers will be issued one week before the next community meeting via the community Slack and mailing list.
+
+### Release Coordinator Responsibilities
+
+- The release coordinator is the person who is ultimately responsible for ensuring that the release is completed on time, and with the desired payload.
+- The release coordinator follows the release process outlined in [#what-is-the-release-process].
+- The release coordinator ensures that all community members are following the correct process.
+- While not a maintainer, the release coordinator is an important role in the community and should be afforded the respect and trust of the maintainers.
+- The release coordinator is empowered to make decisions about the release that are in the best interest of the community.
+  - For example, the release coordinator may choose to delay the release if the payload is not ready or if the quality of the release is not high enough.
+- The release coordinator is responsible for ensuring that progress is being made on the release and regularly updates the community on the progress of the release during the [community meeting].
+- If decisions are required that are outside the scope of the release coordinator's responsibilities, the release coordinator will seek the advice of the maintainers.
+- If the release coordinator is unable to fulfill their responsibilities, they may step down and a new release coordinator will be elected.
+
+## What goes into a release?
+
+A release is a collection of changes that are considered stable and ready for production use.
+These changes can be broken down into the following categories:
+
+- **New features** - new features require an approved OKEP.
+- **Minor changes** - small changes to improve the project.
+- **Bug fixes** - fixes for bugs.
+- **Documentation** - improvements to the documentation.
+- **Testing** - improvements to the testing or CI workflows.
+
+At the [community meeting], we will discuss all approved OKEPs and issues that are targeted for the release and select a maximum of **5** OKEPs for the release.
+This is limited to ensure that:
+
+- There is sufficient development time.
+- Developer focus is not spread too thin.
+- Community feedback is solicited and considered during the release cycle.
+- The release is not too large with too many potential breaking changes.
+
+> [!TIP]
+> As a future enhancement to the process we may wish to consider using T-shirt sizes to the OKEPs instead (XS<->XL). Each size can be represented as a number: XS=1, S=2, M=3, L=4, XL=5. Given release capacity of 10, we could select OKEPs whose combined size is less than or equal to 10.
+
+## How long are releases supported?
+
+OVN-Kubernetes is supported until the next major or minor release.
+
+For example, if the current release is `v1.1` it is supported until either `v2.0` is released or `v1.2` is released.
+
+Patch releases will only be created for critical issues that are discovered after the release. In the case a patch release has been created, that release will be supported until the next major or minor release.
+
+For example, if an issue was found in `v1.1.0` it would be fixed in `v1.1.1` and `v1.1.1` would be supported until `v1.2.0` is released. `v1.1.0` would not be supported after `v1.1.1` is released.
+
+## How do we decide on the release number?
+
+The release number is decided by the community in the [community meeting] during the release planning phase.
+
+The release number is a semantic version number.
+
+The release number is incremented by the following rules:
+
+- Major version is incremented for any backwards incompatible changes.
+- Minor version is incremented for new features, but not backwards incompatible changes.
+- Patch version is incremented for bug fixes.
+
+## What is the release process?
+
+The release process is as follows:
+
+### Step 1: Planning
+
+- The release coordinator is elected following the process outlined in [#who-is-responsible-for-the-release].
+- The release date and code freeze date are decided during the [community meeting] following the process outlined in [#how-do-we-decide-on-the-release-date].
+- The payload of the release is decided during the [community meeting], following the process outlined in [#what-goes-into-a-release].
+- The release coordinator will create a tracking issue and GitHub milestone for the release
+- All issues and pull requests that are targeted for the release are added to the GitHub milestone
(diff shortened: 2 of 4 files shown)


.github/ISSUE_TEMPLATE/release-planning.md
```diff
@@ -0,0 +1,42 @@
...
+- Major version is incremented for any backwards incompatible changes.
+- Minor version is incremented for new features, but not backwards incompatible changes.
+- Patch version is incremented for bug fixes.
+-->
+
+### Important Dates
+<!--
+The release date is the date of the release.
+The code freeze date is the date of the code freeze.
+It is typically 2 weeks before the release date.
```
maiqueb, line 20: these are pretty much using themselves in their definition right ? 

Can't we just keep the relationship between one another ?

.github/ISSUE_TEMPLATE/release-planning.md
```diff
@@ -0,0 +1,42 @@
...
+### Checklist
+
+- [ ] Feature Complete
+- [ ] Code Frozen
+- [ ] (Optional) Release Candidate Tagged
+- [ ] Documentation Complete
+- [ ] Release Notes Complete
+- [ ] Release Tagged
+- [ ] GitHub Release Created
+- [ ] Release posted to mailing list and community slack channels
```
maiqueb, line 42: aren't these 4 just "automation" ? 

Like, I just envision the common flow of pushing a tag to github: 
1. triggers release notes (gathers all PR's and their one-liner description)
2. builds imgs
3. tags release
4. pushes release
5. posts update to community via mailing list / slack

.github/ISSUE_TEMPLATE/release-planning.md
```diff
@@ -0,0 +1,42 @@
...
+<!--
+Please list the major changes that are planned for this release.
+This should include new features, bug fixes, and any other changes that are planned.
+-->
+
+### Checklist
+
+- [ ] Feature Complete
+- [ ] Code Frozen
+- [ ] (Optional) Release Candidate Tagged
```
maiqueb, line 37: can this be optional ? 

Shouldn't we actually run some automated tests on RCs to "prove" their stable enough to be considered ?

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+- The release coordinator follows the release process outlined in [#what-is-the-release-process].
+- The release coordinator ensures that all community members are following the correct process.
+- While not a maintainer, the release coordinator is an important role in the community and should be afforded the respect and trust of the maintainers.
+- The release coordinator is empowered to make decisions about the release that are in the best interest of the community.
+  - For example, the release coordinator may choose to delay the release if the payload is not ready or if the quality of the release is not high enough.
+- The release coordinator is responsible for ensuring that progress is being made on the release and regularly updates the community on the progress of the release during the [community meeting].
+- If decisions are required that are outside the scope of the release coordinator's responsibilities, the release coordinator will seek the advice of the maintainers.
+- If the release coordinator is unable to fulfill their responsibilities, they may step down and a new release coordinator will be elected.
```
maiqueb:
```suggestion
- If the release coordinator is unable to fulfill their responsibilities, they must step down and a new release coordinator will be elected.
```

I suggest stronger wording. I mean, after all, the release coordinator person is *unable* to fulfill their responsibilities. IMHO, they *must* step down.

Thinking a bit more about this, I think there also needs to be a process for the release coordinator to be "released" from their position by the community (maintainers, probably) - e.g. person is on long term sick leave, and the process must continue (just an example).

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+A release is a collection of changes that are considered stable and ready for production use.
+These changes can be broken down into the following categories:
+
+- **New features** - new features require an approved OKEP.
+- **Minor changes** - small changes to improve the project.
+- **Bug fixes** - fixes for bugs.
+- **Documentation** - improvements to the documentation.
+- **Testing** - improvements to the testing or CI workflows.
+
+At the [community meeting], we will discuss all approved OKEPs and issues that are targeted for the release and select a maximum of **5** OKEPs for the release.
```
maiqueb, line 44: why 5 ? would it make sense to have <number of maintainers> instead ? Like, it's not the same thing to have 5 OKEPs when you have 10 maintainers vs 5 OKEPs when you have 1 maintainer (totally made up numbers).

Can we have some clarity into what is an approved OKEP ? How does an OKEP become approved ? Who approves it ?

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+This is limited to ensure that:
+
+- There is sufficient development time.
+- Developer focus is not spread too thin.
+- Community feedback is solicited and considered during the release cycle.
+- The release is not too large with too many potential breaking changes.
+
+> [!TIP]
+> As a future enhancement to the process we may wish to consider using T-shirt sizes to the OKEPs instead (XS<->XL). Each size can be represented as a number: XS=1, S=2, M=3, L=4, XL=5. Given release capacity of 10, we could select OKEPs whose combined size is less than or equal to 10.
```
maiqueb, line 53: this makes sense to me.

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+
+## How long are releases supported?
+
+OVN-Kubernetes is supported until the next major or minor release.
```
maiqueb: wait this mean 0 maintainability for products using this. 

... not 0, but only 3 months of maintained code. 

That's too little imho. FWIW, kubernetes does support N-2 releases (they support the most recent three minor releases).

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+  during this phase
+- The normal review process is followed for all changes
+- The release coordinator updates the community on the progress of the release during the [community meeting]
+
+### Step 3: Code Freeze
+
+- The code freeze date is the 2 weeks before the release date
+- During this phase the `master` branch is closed for new changes
+- Any bug fixes or documentation changes must follow the normal review process, but will also require the explicit approval of the release coordinator.
+- OKEPs may be reviewed and merged into the `master` branch during this phase without explicit approval from the release coordinator since they do not affect the release itself.
```
maiqueb, line 114: does it make sense to clarify that these OKEP's **implementation** must be targeting the **next** release ?

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+- The release coordinator updates the tracking issue to mark the release as complete
+- The release coordinator creates a retrospective document to capture lessons learned
+
+## Release Communication
+
+Throughout the release process, the release coordinator should maintain clear communication with:
+
+- **Maintainers**: For technical decisions and code review coordination
+- **Community**: For progress updates and feedback solicitation
+- **Stakeholders**: For timeline and feature delivery expectations
```
maiqueb, line 154: does it make sense to "define" these roles in the beginning of this doc ? 
I don't have a strong opinion here.

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+
+- Features to be delivered across multiple releases.
+- Features to be disabled in downstream projects.
+
+### Approved -> Implemented
+
+Once an approved OKEP has been implemented and all code has been merged into the `master` branch, the following steps should be followed:
+
+- The tracking issue label `status/approved` is removed.
+- The tracking issue label `status/implemented` is added.
```
maiqueb, line 231: I wonder about this concrete status, and why it is important. 

Could we just have the "approved" and "released" status ? 

Who would consume this intermediate state ? How would they typically react to it ?

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+### Implemented -> Released
+
+Once the OKEP is available in a release, the following steps should be followed:
+
+- The tracking issue label `status/implemented` is removed.
+- The tracking issue label `status/released` is added.
+
+### Released -> Graduated
+
+After release, once the feature has been used for a number of releases, a discussion should be held to determine if the feature should be graduated. All OKEPs are expected to be graduated eventually, although the time between release and graduation may vary.
```
maiqueb: I think the "number of releases" must be defined. 

Also, we probably should have a list of "customers" using the feature - otherwise, we might end up "graduating" features that have no user base - meaning we'd mistake radio silence for "everything's fine" when in reality it means "nobody cared". 

 Just saying that we should want to have legit user feedback, instead of assuming that silence means compliance.

RELEASE.md
```diff
@@ -0,0 +1,253 @@
...
+## What goes into a release?
+
+A release is a collection of changes that are considered stable and ready for production use.
+These changes can be broken down into the following categories:
+
+- **New features** - new features require an approved OKEP.
+- **Minor changes** - small changes to improve the project.
+- **Bug fixes** - fixes for bugs.
+- **Documentation** - improvements to the documentation.
+- **Testing** - improvements to the testing or CI workflows.
```
maiqueb, line 43: how are these selected ? 

In my past experience in this community, any minor change, documentation, or testing improvement have low odds of getting attention / traction. Would it make sense to try to define a process for getting some predictability into this type of work ?

RELEASE.md
maiqueb, line 44: IMHO we also have to discuss how are issues targeted for the release, since that's the initial step into getting feature work into the release.

maiqueb: can we shed some light into the decision process ? 

Some questions on my mind: 
- how are issues targeted for the release ? Who "votes" ? Are all votes equal ? Is this a democracy ? 
- priorities: is there a sense of priorities between the OKEPs ? I think we need to have it. Who chooses those ? 
- if one maintainer is tasked with reviewing / approving two OKEP implementations, it sounds its 100% up to them to choose where to focus, without any regard for the community's interests. This might be OK, but it might not be fair (or not, not sure).

**NOTE:** I'm assuming OKEPs are assigned between maintainers. If one OKEP sits idle without attention, what happens ? 

Plus, what happens when there's interest from the community but the OKEP author doesn't update ?

RELEASE.md
dave-tucker: yes it would.
but i think we need to be mindful of how much process we add at once.
maybe start with the bare minimum, and then if we feel that work isn't getting prioritized, then adjust accordingly?

RELEASE.md
dave-tucker: The OKEP process and stages are documented further down (and in the dedicated OKEP process doc).

The process I had in mind was "rough consensus" - we can refine later if that doesn't work.

Priorities? They're implied because of the limit of how many OKEPs can be implemented in a single release - didn't make it in? then wasn't high enough of a priority. I'm not against having explicit priorites, but again, trying to keep things light in the first revision.

The OKEP author is usually the implementor. Once the OKEP is approved and planned, normal code review rules apply - any maintainer can bless it. I don't want to dismiss your concerns around the handling of OKEPs, but I'd rather defer them until we learn what works and what doesn't.

RELEASE.md
dave-tucker: no i think that's implied. generally okep authors want their okeps implemented asap ;)
asap however is not guaranteed to be the next release anymore.

RELEASE.md
dave-tucker: it's not released until it's actually been shipped in a release.
i felt this distinction was important since a user should be able to tell from this status whether they should be using a `master` build to test out a feature (status/implemented) or whether it's been shipped in a release already `status/released`.

i also considered a situation where we may implement a feature, then decide that it's not stable enough to release... in which case `status/released` seems weird as the end state.

.github/ISSUE_TEMPLATE/release-planning.md
dave-tucker: afaik, we don't currently create RCs.
i'm not against doing it eventually, but wanted to keep things lightweight at first.

.github/ISSUE_TEMPLATE/release-planning.md
dave-tucker: looks like it's all being done manually for now, so I tried to capture it here.
automation would make sense though

RELEASE.md
dave-tucker: > why 5 ?

because there are 5 `owner` entries in the `OWNERS` file... and I was hoping to be controversial and spark a conversation 😆 
i'm sure 5 will morph into something more meaningful over time.

dave-tucker: any suggestions?

PR #5474, opened 2025-08-11
NetworkQoS improvements

jxiaobin:
various improvements to NetworkQoS controller
- use networkqos ns&name as key in the event queue instead of obj pointer
- make networkSelector a mandatory field
- improve networkqos error handling
- fix validation rule in networkqos crd

## Summary by CodeRabbit

* **New Features**
  * NetworkQoS now requires networkSelectors and adds stricter validation for networkAttachment selectors and immutability messaging.

* **Bug Fixes**
  * Improved reconciliation error handling with retries, clearer failure logs and status updates when reconciliation ultimately fails.

* **Performance**
  * Reduced unnecessary reconciliations via namespace and pod label/IP caching.

* **Refactor**
  * Unified key-based event processing for namespaces and pods for simpler, more reliable handling.

* **Tests**
  * E2E specs updated to include required networkSelectors.


go-controller/pkg/crd/networkqos/v1alpha1/types.go
```diff
@@ -47,10 +47,9 @@ type NetworkQoS struct {
...
 type Spec struct {
 	// networkSelector selects the networks on which the pod IPs need to be added to the source address set.
-	// NetworkQoS controller currently supports `NetworkAttachmentDefinitions` type only.
-	// +optional
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="networkSelector is immutable"
-	// +kubebuilder:validation:XValidation:rule="self.all(sel, sel.networkSelectionType == 'ClusterUserDefinedNetworks' || sel.networkSelectionType == 'NetworkAttachmentDefinitions')", message="Unsupported network selection type"
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="networkSelectors are immutable"
+	// +kubebuilder:validation:XValidation:rule="self.filter(sel, sel.networkSelectionType == 'NetworkAttachmentDefinitions').all(sel, has(sel.networkAttachmentDefinitionSelector) && has(sel.networkAttachmentDefinitionSelector.networkSelector) && (has(sel.networkAttachmentDefinitionSelector.networkSelector.matchExpressions) || has(sel.networkAttachmentDefinitionSelector.networkSelector.matchLabels)))", message="networkAttachmentDefinitionSelector cannot have an empty networkSelector"
```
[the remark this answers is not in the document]
trozet: So there are a few things to unpack here. First in the helm k8s.ovn.org_networkqoses.yaml chart, this code already exists. So our helm chart CRD was out of sync with the CRD definition here.

Second, the addition of this check looks like it is to prevent a user from having an empty LabelSelector ({}) or nil selector. Normally in Kubernetes, empty selector means match everything. So if someone had a NAD selector with a namespace and and empty networkSelector, then we should match all NADs in that namespace. I guess we are trying to restrict that so that we only pick IPs off of a single network?

Third, why would this not also apply to the other ClusterUserDefinedNetworkSelector type?

Fourth, we would need to make sure this new tightening doesn't break existing users. I guess for our upstream users it would not, since the CRD in the helm repo was already including this check, but for our RH friends it might.

go-controller/pkg/ovn/controller/network_qos/network_qos_controller.go
```diff
@@ -429,60 +392,41 @@ func (c *Controller) onNQOSNamespaceUpdate(oldObj, newObj interface{}) {
...
-		}
-		ns, ok = tombstone.Obj.(*corev1.Namespace)
-		if !ok {
-			utilruntime.HandleError(fmt.Errorf("tombstone contained object that is not a Namespace: %#v", tombstone.Obj))
-			return
-		}
-	}
-	if ns != nil {
-		c.nqosNamespaceQueue.Add(newEventData(ns, nil))
+	key, err := cache.MetaNamespaceKeyFunc(obj)
```
trozet, line 405: with your change we are losing tombstone handling. I think you need to restore that.

go-controller/pkg/ovn/controller/network_qos/network_qos_controller.go
```diff
@@ -518,22 +469,13 @@ func (c *Controller) podNetworkResolver() func(nadKey string) string {
...
-		pod, ok = tombstone.Obj.(*corev1.Pod)
-		if !ok {
-			utilruntime.HandleError(fmt.Errorf("tombstone contained object that is not a Pod: %#v", tombstone.Obj))
-			return
-		}
-	}
-	if pod != nil {
-		c.nqosPodQueue.Add(newEventData(pod, nil))
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
```
trozet, line 473: same tombstone concern here.

go-controller/pkg/ovn/controller/network_qos/network_qos_pod.go
```diff
@@ -113,14 +118,36 @@ func reconcilePodForDestinations(nqosState *networkQoSState, podNs *corev1.Names
...
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get pod %s: %v", podKey, err)
+		}
+		// fake dead pod with meta only
+		pod = &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:         podNsStr,
+				Name:              podName,
+				Labels:            getCachedPodLabels(c.podLabelsCache, podKey),
```
trozet, line 137: so with moving to using the string keys instead of objects (which is good) it opens up a new problem where upon start up your caches are empty. So when it starts handling events cache lookups will fail for older objects. The way to solve it is to have an initial sync before the workers start that rebuilds all your caches.

This network qos controller is built using its own level driven implementation. It might be better to move to the pkg/controller generic level-driven framework. It handles starting multiple controllers at once and running an initialSync. 

Either way you need an initialSync to avoid race conditions on start up with non-primed caches.

go-controller/pkg/ovn/controller/network_qos/network_qos_controller.go
```diff
@@ -494,19 +438,26 @@ func (c *Controller) onNQOSPodUpdate(oldObj, newObj interface{}) {
 	oldPodLabels := labels.Set(oldPod.Labels)
 	newPodLabels := labels.Set(newPod.Labels)
 	resolver := c.podNetworkResolver()
-	oldPodIPs, _ := util.GetPodIPsOfNetwork(oldPod, c.NetInfo, resolver)
-	newPodIPs, _ := util.GetPodIPsOfNetwork(newPod, c.NetInfo, resolver)
-	oldPodCompleted := util.PodCompleted(oldPod)
-	newPodCompleted := util.PodCompleted(newPod)
+	oldPodIPs, err := util.GetPodIPsOfNetwork(oldPod, c.NetInfo, resolver)
+	if err != nil {
+		klog.Errorf("Failed to get IPs from old version of pod %s/%s: %v", oldPod.Namespace, oldPod.Name, err)
```
trozet, line 443: why is this an error? the pod may not have an IP yet, that is expected until the CDN/UDN controller assigns it?

go-controller/pkg/ovn/controller/network_qos/network_qos_controller.go
```diff
@@ -494,19 +438,26 @@ func (c *Controller) onNQOSPodUpdate(oldObj, newObj interface{}) {
...
+	newPodIPs, err := util.GetPodIPsOfNetwork(newPod, c.NetInfo, resolver)
+	if err != nil {
+		klog.Errorf("Failed to get IPs from new version of pod %s/%s: %v", newPod.Namespace, newPod.Name, err)
+	}
 	if labels.Equals(oldPodLabels, newPodLabels) &&
 		// check for podIP changes (in case we allocate and deallocate) or for dualstack conversion
 		// it will also catch the pod update that will come when LSPAdd and IPAM allocation are done
-		len(oldPodIPs) == len(newPodIPs) &&
-		oldPodCompleted == newPodCompleted {
+		oldPodIPs != nil && newPodIPs != nil && util.IPsAreEqual(oldPodIPs, newPodIPs) {
```
trozet, line 452: how come you removed the checking if a pod is completed or not?

go-controller/pkg/crd/networkqos/v1alpha1/types.go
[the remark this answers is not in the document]
trozet: @tssurya @kyrtapz are you using NetworkQoS in OCP? Trying to get a sense of if tightening the API would break you.

PR #5481, opened 2025-08-13
Fix mgmt port allocation with VFs provided as resource for DPU

wizhaoredhat:
## 📑 Description
This PR fixes several issues when allocating the management port with a SR-IOV resource instead of hard coding the SR-IOV netdev name.
    
Such issues come from the fact that the gateway and management port interfaces need to be present during Init. Otherwise other parts of the code will not have the proper interface.

Also Fix Node Admission Webhook for OVNNodePrimaryDPUHostAddr annotation.

## How to verify it
Run ovn-kubernetes with the NVIDIA BF3

## Summary by CodeRabbit

* **New Features**
  * Automatically derives the gateway interface from the management port when running in DPU host mode (when configured).
  * Allows the OVN node primary DPU host address annotation to be set.
* **Refactor**
  * Resolves management port device and gateway interface during initialization for earlier, consistent setup.
  * Adds clearer logs for explicit overrides and resource-derived configuration.

### DIFF
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -607,6 +607,59 @@ func handleNetdevResources(resourceName string) (string, error) {
 	return netdevice, nil
 }
 
+// configureMgmtPortNetdevFromResource uses device plugin resources to determine and set
+// the management port netdevice if a DP resource name is provided via configuration.
+func configureMgmtPortNetdevFromResource() error {
+	if config.OvnKubeNode.MgmtPortDPResourceName == "" {
+		return nil
+	}
+	if err := handleDevicePluginResources(); err != nil {
+		return err
+	}
+	netdevice, err := handleNetdevResources(config.OvnKubeNode.MgmtPortDPResourceName)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


aserdean:
lgtm, thanks for fixing this @wizhaoredhat 


/retest-failed

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -607,6 +607,59 @@ func handleNetdevResources(resourceName string) (string, error) {
...
+	netdevice, err := handleNetdevResources(config.OvnKubeNode.MgmtPortDPResourceName)
+	if err != nil {
+		return err
+	}
+	if config.OvnKubeNode.MgmtPortNetdev != "" {
+		klog.Warningf("MgmtPortNetdev is set explicitly (%s), overriding with resource...",
+			config.OvnKubeNode.MgmtPortNetdev)
+	}
+	config.OvnKubeNode.MgmtPortNetdev = netdevice
+	klog.V(5).Infof("Using MgmtPortNetdev (Netdev %s) passed via resource %s",
```
trozet, line 628: nit: this could probably be at regular Info level, seems like important information that is only set one time

trozet (approved):

PR #5489, opened 2025-08-15
E2E: add EIP TCP disruption test where client pushes externally while we restart CP

martinkennelly:
Help give us confidence for restarting our control plane when eip is active for CDN,UDN (v4 & v6).
A few bugs recently could have been caught with something like this.

Client sets up tcp test with external server and performs an iperf test while we restart control plane.
Client contineously pushes data externally.

Fails on dataloss.
Logs retransmissions.
Restarts ovnkube node on egress node and client pod node.

This test cannot work currently with the gRPC
implementation of EIP Healthcheck because the EIP is prone to move to a different Node and therefore that
will interrupt the tcp connection.
The gRPC server which helps implement the Egress IP healthcheck is removed when ovnkube node is shutdown / restarted, and there's a possibility the node is no longer assigned as an egress node by cluster manager, and today that will break the TCP connection if the eip is unassigned or reassigned to another node switch back to gRPC when [1] is merged.

[1] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4286

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

- New Features
  - None.
- Tests
  - Added e2e tests validating TCP connectivity during ovnkube-node restarts (iperf3-based), including long-run throughput and packet-loss checks (note: test duplicated in diff).
  - Added an iperf3 JSON parser test suite covering normal, edge and invalid inputs.
  - Added iperf3 parsing utility to produce consistent send/receive/retransmit metrics.
- Chores
  - Improved test helpers to better wait for pod readiness/success and aggregate container termination errors for clearer diagnostics.

### DIFF
--- a/test/e2e/egressip.go
+++ b/test/e2e/egressip.go
@@ -25,10 +25,12 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider"
 	infraapi "github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider/api"
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/ipalloc"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/iperf3helper"
 
 	nadclient "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/typed/k8s.cni.cncf.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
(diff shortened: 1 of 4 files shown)


martinkennelly:
Follow on work needed to add an additional test where client (CDN) sets up the tcp connection but external pushes to client.

martinkennelly:
Interesting failure for LGW mode test lanes.. taking a look

test/e2e/egressip.go
```diff
@@ -1059,6 +1086,222 @@ spec:
...
+		framework.Logf("Get baseline network performance with no disruption")
+		clientPodName := "iperfclient-baseline-" + f.Namespace.Name
+		clientPod, err := createPod(f, clientPodName, unusedEIPNode, f.Namespace.Name, []string{"iperf3",
+			"--client", externalContainerIP,
+			"--port", externalServerContainerPortStr,
+			"--time", "2",
+			"--interval", ".1",
+			"--bitrate", "50M",
+			"--parallel", "1",
+			"--json",
```
huiran0826, line 1193: Shall we include 'server push' in this case as I understand that the difference is only adding additional parameter '--reverse' in this command?

martinkennelly:
@huiran0826 i assumed someone else was taking over this PR from the eip improvement epic. I ran out of time to look into why the LGW job is failing. I can give push rights to whomever takes it over.

PR #5490, opened 2025-08-16
Change networks to relatively rare CIDR ranges

martinkennelly:
10.128.x.x is used intensively as providers network or CDN CIDR.

Real fix is we provide tools for the tests to get
CIDRs for UDNs that dont conflict with CDN,
provider network at test run time instead of this static config..

## Summary by CodeRabbit

* **Tests**
  * Updated IPv4 CIDR ranges in end-to-end tests from 10.128.x.x to 172.31.x.x (e.g., /16, /24, /29) across KubeVirt, multihoming, and network-segmentation suites; inline note added that 172.16.0.0/12 is RFC1918.
  * Adjusted corresponding static IPs, gateways and CIDR-derived test data; IPv6 ranges, test logic and control flow remain unchanged.
  * No changes to public APIs or exported declarations.

### DIFF
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -1548,9 +1548,9 @@ fi
 			cudn       *udnv1.ClusterUserDefinedNetwork
 			vm         *kubevirtv1.VirtualMachine
 			vmi        *kubevirtv1.VirtualMachineInstance
-			cidrIPv4   = "10.128.0.0/24"
+			cidrIPv4   = "172.31.0.0/24" // subnet in private range 172.16.0.0/12 (rfc1918)
 			cidrIPv6   = "2010:100:200::0/60"
-			staticIPv4 = "10.128.0.101"
+			staticIPv4 = "172.31.0.101"
 			staticIPv6 = "2010:100:200::101"
 			staticMAC  = "02:00:00:00:00:01"
 			restart    = testCommand{
@@ -2094,7 +2094,7 @@ ip route add %[3]s via %[4]s
 	Context("with kubevirt VM using layer2 UDPN", Ordered, func() {
 		var (
 			podName                 = "virt-launcher-vm1"
-			cidrIPv4                = "10.128.0.0/24"
+			cidrIPv4                = "172.31.0.0/24"
 			cidrIPv6                = "2010:100:200::/60"
 			primaryUDNNetworkStatus nadapi.NetworkStatus
 			virtLauncherCommand     = func(command string) (string, error) {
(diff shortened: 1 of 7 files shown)


pliurh:
In the commit message, should `128.x.x.x` be replaced with `10.128.x.x`?

pliurh:
11.128.0.0/16 is a valid public IP. Should we instead use the subnet 192.0.2.0/24, which is designated for testing and documentation?

martinkennelly:
> 11.128.0.0/16 is a valid public IP. Should we instead use the subnet 192.0.2.0/24, which is designated for testing and documentation?

100% - thanks.. will change

/hold

martinkennelly:
/unhold

Waiting on CI

pliurh:
LGTM

martinkennelly:
Unhold

@trozet  any interest in this ?

 It may help us downstream as well as other orgs that want to run our e2es downstream. Thanks.

trozet (changes requested):
@martinkennelly is the issue that the UDN IP collides with the provider network IP in downstream testing? The 192.0.0.0/16 network includes a bunch of reserved ranges. Like:

192.0.0.0/24 - This block is reserved for IETF protocol assignments.
   At the time of writing this document, there are no current
   assignments.  Allocation policy for future assignments is given in
   [[RFC5736](https://datatracker.ietf.org/doc/html/rfc5736)].

Can we just make it an environment variable to set these and overwrite as needed?

martinkennelly:
Yep, it collides with provider network but may also collide with CDN.

> use env var

Fair enough. I'll do that.

martinkennelly:
Ive changed my mind after reading `RFC 1918` and also because of needless complexity using an env var. Ive switched to using the last /16 subnet in the private range `172.16.0.0/12` as defined in the aforementioned RFC.  Good enough for now.

martinkennelly:
Waiting on CI before annoying Tim again.

martinkennelly:
@trozet wdyt ?

martinkennelly:
@pliurh I changed the IPs I was using because i wasnt able to find enough subnet using the documentation subnet.. PTAL.. thanks.

test/e2e/kubevirt.go
```diff
@@ -1548,9 +1548,9 @@ fi
 			cudn       *udnv1.ClusterUserDefinedNetwork
 			vm         *kubevirtv1.VirtualMachine
 			vmi        *kubevirtv1.VirtualMachineInstance
-			cidrIPv4   = "10.128.0.0/24"
+			cidrIPv4   = "172.31.0.0/24" // subnet in private range 172.16.0.0/12 (rfc1918)
```
tssurya, line 1551: hmm isn't this service cidr default value in ocp.. is this tested its working d/s? i think its gonna conflict with cluster ip range?

martinkennelly: its possible but in ocp ive only seen `172.30.0.0/16`

martinkennelly: o/origin:
```
$grep -Rni "ServiceNetwork" | grep "172"
pkg/clioptions/clusterdiscovery/provider_test.go:121:	ServiceNetwork: []string{"172.30.0.0/16"},
pkg/clioptions/clusterdiscovery/provider_test.go:131:	ServiceNetwork: []string{"172.30.0.0/16"},
pkg/monitortests/kubeapiserver/apiunreachablefromclientmetrics/monitortest_test.go:28:				serviceNetworkIP: "172.30.0.1",
test/extended/util/configv1shim_test.go:75:			ServiceNetwork: []string{"172.30.0.0/16"},
test/extended/util/configv1shim_test.go:86:			ServiceNetwork:    []string{"172.30.0.0/16"},
```

martinkennelly: Thank you for your time reviewing this :bow:

PR #5493, opened 2025-08-18
OVN EIP: improve startup syncer for EIP

martinkennelly:
## Summary by CodeRabbit

- **Bug Fixes**
  - Nexthop generation now limited to nodes assigned to an EgressIP; stale/invalid nexthops on logical router ports are removed and reconciled.
  - Policies with missing or empty node assignments are logged and removed.
  - Management-port handling tolerates missing ports (skips absent ports) and requires at least one address when present; IPv4/IPv6 addresses parsed per family.
  - Logging for stale nexthops and removals improved.

- **Tests**
  - Added IPv4/IPv6 table-driven tests verifying removal of invalid nexthops and correct reconciliation for multi-node EgressIP assignments.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -1700,13 +1700,13 @@ func (e *EgressIPController) syncStaleEgressReroutePolicy(cache egressIPCache) e
 			// Update Logical Router Policies that have stale nexthops. Notice that we must do this separately
 			// because logicalRouterPolicyStaleNexthops must be populated first
 			for _, staleNextHopLogicalRouterPolicy := range logicalRouterPolicyStaleNexthops {
-				if staleNextHopLogicalRouterPolicy.Nexthop == nil {
-					continue
-				}
-				klog.Infof("syncStaleEgressReroutePolicy will remove stale nexthops for LRP %q for network %s: %s",
-					staleNextHopLogicalRouterPolicy.UUID, networkName, *staleNextHopLogicalRouterPolicy.Nexthop)
+				klog.Infof("syncStaleEgressReroutePolicy will remove stale nexthops for LRP %q for network %s: %v",
+					staleNextHopLogicalRouterPolicy.UUID, networkName, staleNextHopLogicalRouterPolicy.Nexthops)
 			}
-
+			// nothing to do if there's no stale next hops
+			if len(logicalRouterPolicyStaleNexthops) == 0 {
+				continue
+			}
 			err = libovsdbops.DeleteNextHopsFromLogicalRouterPolicies(e.nbClient, cache.networkToRouter[networkName], logicalRouterPolicyStaleNexthops...)
 			if err != nil {
 				return fmt.Errorf("unable to remove stale next hops from logical router policies for network %s: %v", networkName, err)


martinkennelly:
@arghosh93 any eta for review?

arghosh93:
> @arghosh93 any eta for review?

@martinkennelly, I will address this PR before end of this week.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1603,6 +1603,23 @@ func (e *EgressIPController) syncPodAssignmentCache(egressIPCache egressIPCache)
...
+		egressNodeNames, ok := cache.egressIPNameToAssignedNodes[eipName]
+		if !ok {
+			return nil, fmt.Errorf("failed to find egress nodes in cache for EgressIP %q", eipName)
+		}
+		filteredEgressNodesRedirectsCache := make(map[string]redirectIPs)
+		for _, egressNode := range egressNodeNames {
+			nodeRedirect, ok := nodeRedirectCache[egressNode]
+			if !ok {
+				return nil, fmt.Errorf("failed to find egress node %q in node redirect cache", egressNode)
+			}
```
arkadeepsen: Do you think the validation of whether the egress nodes corresponding to an egress IP with that of `nodeRedirectCache` should be done in the `generateCacheForEgressIP()` function itself? If it fails here, the corresponding LRP will be removed.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1866,28 +1894,37 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
 			r := redirectIPs{}
 			mgmtPort := &nbdb.LogicalSwitchPort{Name: ni.GetNetworkScopedK8sMgmtIntfName(node.Name)}
 			mgmtPort, err := libovsdbops.GetLogicalSwitchPort(e.nbClient, mgmtPort)
-			if err != nil {
-				// if switch port isnt created, we can assume theres nothing to sync
-				if errors.Is(err, libovsdbclient.ErrNotFound) {
-					continue
-				}
+			// if the error is not the switch port not found, then return error to allow retry
```
arkadeepsen:
```suggestion
			// if the error is not nil and the switch port not found, then return error to allow retry
```
nit

go-controller/pkg/ovn/egressip.go
```diff
@@ -1866,28 +1894,37 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
...
-			// assuming only one IP per IP family
-			for _, mgmtPortAddress := range mgmtPortAddresses {
-				mgmtPortAddressesStr := strings.Fields(mgmtPortAddress)
-				mgmtPortIP := net.ParseIP(mgmtPortAddressesStr[1])
-				if utilnet.IsIPv6(mgmtPortIP) {
-					if ip := mgmtPortIP.To16(); ip != nil {
-						r.v6MgtPort = ip.String()
+			// if management port is available, gather the data. If it's not available, OVN constructs that depend on a deleted
+			// management port IP will fail and be cleaned up in sync LRPs func.
+			if mgmtPort != nil {
```
arkadeepsen, line 1894:
```suggestion
			if err == nil {
```
WDYT? Do you expect this codeblock to run even when `err != nil`?

go-controller/pkg/ovn/egressip.go
```diff
@@ -1866,28 +1894,37 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
...
+						}
+						if utilnet.IsIPv6(ip) {
+							if r.v6MgtPort == "" && ip.To16() != nil {
+								r.v6MgtPort = ip.String()
+							}
+						} else {
+							if r.v4MgtPort == "" && ip.To4() != nil {
+								r.v4MgtPort = ip.String()
+							}
+						}
```
arkadeepsen, line 1918: Any specific reason to change the way `mgmtPort` was being parsed earlier?

arkadeepsen (commented):
1st pass. I'll go through the unit test in the next pass.
Can there be an e2e test added to cover this scenario?

go-controller/pkg/ovn/egressip.go
```diff
@@ -1894,28 +1894,37 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
...
+				// Extract at most one IP per family; entries are "MAC IP [IP ...]"
+				for _, macPlusIPs := range mgmtPortAddresses {
+					parts := strings.Fields(macPlusIPs)
+					if len(parts) < 2 {
+						continue // no IPs
 					}
-				} else {
-					if ip := mgmtPortIP.To4(); ip != nil {
-						r.v4MgtPort = ip.String()
+					for _, ipStr := range parts[1:] {
```
arghosh93, line 1905: Nice catch, I think earlier we were not considering IPv6 IP in case of a dual stack cluster.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1894,28 +1894,37 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
...
+			// management port IP will fail and be cleaned up in sync LRPs func.
+			if mgmtPort != nil {
+				mgmtPortAddresses := mgmtPort.GetAddresses()
+				if len(mgmtPortAddresses) == 0 {
+					return cache, fmt.Errorf("management switch port %s for node %s does not contain any addresses", ni.GetNetworkScopedK8sMgmtIntfName(node.Name), node.Name)
+				}
+				// Extract at most one IP per family; entries are "MAC IP [IP ...]"
+				for _, macPlusIPs := range mgmtPortAddresses {
+					parts := strings.Fields(macPlusIPs)
+					if len(parts) < 2 {
```
arghosh93, line 1902: This is highly unlikely, right? If it still happens should we log something(v5) instead of ignoring quietly?
However I agree that if it happens, we will also catch elsewhere(during creation of LSP mostly) and log there.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1611,11 +1628,22 @@ func (e *EgressIPController) syncStaleEgressReroutePolicy(cache egressIPCache) e
...
 				}
+				networkNodeRedirectCache, err := limitToValidEgressNodes(eipName, allNetworkNodeRedirectCache)
+				if err != nil {
+					klog.Errorf("syncStaleEgressReroutePolicy unable to find at least one egress node for network %s and egress ip %s,"+
+						" removing logical router policy (UUID: %s)", networkName, eipName, item.UUID)
+					return true
+				}
+				if len(networkNodeRedirectCache) == 0 {
+					klog.Errorf("syncStaleEgressReroutePolicy could not find at least one node in the redirects "+
+						"cache after filtering for egress ip %q, UUID %q, programming error, removing logical router policy", eipName, item.UUID)
```
arghosh93: Should we say 'programming error' in a log message? :D

go-controller/pkg/ovn/egressip.go
```diff
@@ -1611,11 +1628,22 @@ func (e *EgressIPController) syncStaleEgressReroutePolicy(cache egressIPCache) e
...
 					return false
 				}
-				networkNodeRedirectCache, ok := cache.egressNodeRedirectsCache.cache[networkName]
-				if !ok || len(networkNodeRedirectCache) == 0 {
+				allNetworkNodeRedirectCache, ok := cache.egressNodeRedirectsCache.cache[networkName]
+				if !ok || len(allNetworkNodeRedirectCache) == 0 {
 					klog.Infof("syncStaleEgressReroutePolicy found invalid logical router policy (UUID: %s) because no assigned Nodes for EgressIP %s", item.UUID, eipName)
 					return true
 				}
+				networkNodeRedirectCache, err := limitToValidEgressNodes(eipName, allNetworkNodeRedirectCache)
```
arghosh93: Please correct me if below statement is wrong:
Earlier we used to remove stale LRPs by checking nexthop corresponding to every node, but now we are limiting it to egressIPNodes. 
If it's true, I have following questions:
- If egress IP gets moved from a local zone node to a remote zone node and we dont remove stale LRPs from the node where POD is running, wouldn't that create any problem?
- Are we doing this to reduce processing time or was there an issue?

go-controller/pkg/ovn/egressip.go
martinkennelly: true - it should but id still need this check here as i wouldnt be able to ignore a missing value in the redirects cache.

go-controller/pkg/ovn/egressip.go
martinkennelly: > Earlier we used to remove stale LRPs by checking nexthop corresponding to every node, but now we are limiting it to egressIPNodes.

Yes, before, we limited it to egress nodes but now we are limiting it to the set of valid egress nodes for a specific EIP.

> If egress IP gets moved from a local zone node to a remote zone node and we dont remove stale LRPs from the node where POD is running, wouldn't that create any problem?

if the pod is still selected for that EIP we would expect the LRP to still exist but it different next hops.

> Are we doing this to reduce processing time or was there an issue?

issue - because previously, we only checked if the next hops is a valid next hop because its an egress node but it wasnt also restricted to the EIP itself.

go-controller/pkg/ovn/egressip.go
martinkennelly: yep! i think its good to highlight that we should never get into this scenario.

go-controller/pkg/ovn/egressip.go
martinkennelly: its unlikely yes but possible with user intervention or programming error.

go-controller/pkg/ovn/egressip.go
martinkennelly: there could be an error though and mgmtPort isnt set, hence why i check on that var.

go-controller/pkg/ovn/egressip.go
martinkennelly: yep! it wasnt getting caught.

go-controller/pkg/ovn/egressip.go
martinkennelly: Dualstack clusters were broken because we have mac + IP + IP so what would happen is upon sync, we would remove the ipv6 next hop, but later during sync it was readded so for a brief time,  interrupted service.

martinkennelly: Another reason we really need disruption tests for EIP. I have a PR.

martinkennelly:
@arkadeepsen @arghosh93 thank you so much for the review - added some comments.

martinkennelly:
Hitting the ran out of space issue

go-controller/pkg/ovn/egressip.go
arkadeepsen: If you incorporate this change (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5493#discussion_r2322084128), then I think we can avoid the issue of removing unrelated LRPs.

go-controller/pkg/ovn/egressip.go
arkadeepsen: If an error occurs do we want to proceed with the execution of the following code? It means that there was some issue in fetching the details. Shouldn't we only proceed when the fetch was successful (`err == nil`)?

go-controller/pkg/ovn/egressip.go
arkadeepsen: I think we can incorporate the nitpick comment by coderabbit to exit early from the loop: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5493#pullrequestreview-3185192655

martinkennelly: done

martinkennelly:
Theres no downstream bugs for this yet - i havent seen anyone talk about it - i noticed it while testing a fix for another EIP bug.

go-controller/pkg/ovn/egressip.go
martinkennelly: done

go-controller/pkg/ovn/egressip.go
martinkennelly: we cannot return error when building the map of data required to sync because the mgnt port could be gone and we dont want to block the sync back on that.  Above we do return error if we fail to get the management port with an error other than not found:

```
// return if error is anything other than not found to allow retry
			if err != nil && !errors.Is(err, libovsdbclient.ErrNotFound) {
				return cache, fmt.Errorf("failed to find management port for node %s: %v", node.Name, err)
			}
```

martinkennelly: If you agree - please resolve this convo - happy to discuss further cause i appreciate your eyes here..

martinkennelly:
@arkadeepsen @arghosh93 Thanks for reviews! Looking forward to another round. I appreciate the time here.

@trozet any interest in this?

go-controller/pkg/ovn/egressip.go
arkadeepsen: Hey @martinkennelly, this change is still missing.

martinkennelly: hmm - i looked at this comment earlier and i did one change that was recommened. Could you paste in the change youre talking about here? Thanks. @arkadeepsen

arkadeepsen: Here's the one I was looking at:

`1909-1929`: Mgmt-port parsing robustness looks good; add a tiny early-exit

The multi-IP parsing per address entry is correct and fixes dual-stack issues. You can save a few cycles by breaking once both families are found.
```diff
-					for _, ipStr := range parts[1:] {
+					for _, ipStr := range parts[1:] {
 						ip := net.ParseIP(ipStr)
 						if ip == nil {
 							continue
 						}
 						if utilnet.IsIPv6(ip) {
 							if r.v6MgtPort == "" && ip.To16() != nil {
 								r.v6MgtPort = ip.String()
 							}
 						} else {
 							if r.v4MgtPort == "" && ip.To4() != nil {
 								r.v4MgtPort = ip.String()
 							}
 						}
+						if r.v4MgtPort != "" && r.v6MgtPort != "" {
+							break
+						}
 					}
```

martinkennelly: can we ignore this suggestion? I want to get the main fixes in.

arkadeepsen: Sure.

arkadeepsen (approved):
LGTM

arghosh93 (approved):
LGTM

martinkennelly:
@tssurya ive no downstream bug for this but i noticed it during testing of the garp issue. I think its important.

pperiyasamy:
this bug is seen from @huiran0826's testing today, thank you martin for the headsup about this PR.

martinkennelly:
@pperiyasamy I am giving you push rights to my repo for any additional work.

go-controller/pkg/ovn/egressip.go
```diff
@@ -1866,28 +1890,37 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
...
-			// assuming only one IP per IP family
-			for _, mgmtPortAddress := range mgmtPortAddresses {
-				mgmtPortAddressesStr := strings.Fields(mgmtPortAddress)
-				mgmtPortIP := net.ParseIP(mgmtPortAddressesStr[1])
-				if utilnet.IsIPv6(mgmtPortIP) {
-					if ip := mgmtPortIP.To16(); ip != nil {
-						r.v6MgtPort = ip.String()
+			// if management port is available, gather the data. If it's not available, OVN constructs that depend on a deleted
+			// management port IP will fail and be cleaned up in sync LRPs func.
+			if mgmtPort != nil {
```
jcaamano, line 1894: I don't see how this is any different to what was previously done in regards to "Management-port handling tolerates missing ports (skips absent ports)" or EIP: do not skip generating potential valid next hops if LSP not found". This will be nil on `libovsdbclient.ErrNotFound` so it's effectively equivalent to the `continue`we had before?

go-controller/pkg/ovn/egressip.go
```diff
@@ -1681,13 +1702,16 @@ func (e *EgressIPController) syncStaleEgressReroutePolicy(cache egressIPCache) e
 			// Update Logical Router Policies that have stale nexthops. Notice that we must do this separately
 			// because logicalRouterPolicyStaleNexthops must be populated first
 			for _, staleNextHopLogicalRouterPolicy := range logicalRouterPolicyStaleNexthops {
-				if staleNextHopLogicalRouterPolicy.Nexthop == nil {
+				if len(staleNextHopLogicalRouterPolicy.Nexthops) == 0 {
```
jcaamano: How can this happen? We are only adding policies with next hops to `logicalRouterPolicyStaleNexthops`. If it can happen, why wouldn't we remove the policy from logicalRouterPolicyStaleNexthops? What effect does it have to call `DeleteNextHopsFromLogicalRouterPolicies` passing a policy with no next hops? I am guessing none so what's the point of having them?

go-controller/pkg/ovn/egressip.go
```diff
@@ -1626,6 +1636,17 @@ func (e *EgressIPController) syncStaleEgressReroutePolicy(cache egressIPCache) e
...
+				if !ok || len(allNetworkNodeRedirectCache) == 0 {
+					klog.Infof("syncStaleEgressReroutePolicy found invalid logical router policy (UUID: %s) because no assigned Nodes for EgressIP %s", item.UUID, eipName)
+					return true
+				}
+				networkNodeRedirectCache := limitToValidEgressNodes(eipName, allNetworkNodeRedirectCache)
+				if len(networkNodeRedirectCache) == 0 {
+					klog.Errorf("syncStaleEgressReroutePolicy could not find at least one node in the redirects "+
+						"cache after filtering for egress ip %q, UUID %q, programming error, removing logical router policy", eipName, item.UUID)
+					return true
+				}
```
jcaamano: I think this would be better if it was just something like
```
networkNodeRedirectCache := limitToValidEgressNodes(eipName, cache.egressNodeRedirectsCache.cache[networkName])
if len(allNetworkNodeRedirectCache) == 0 {
    klog.Infof("syncStaleEgressReroutePolicy deleting invalid logical router policy %q because there are no existing nodes assigned to its EgressIP %s", item.UUID, eipName)
    return true
}
```

jcaamano:
Also, assuming the zone controller can move an EIP non-disruptively (i.e removing the old LRP and adding the new one on the same transaction, or just updating the next hop in a transaction), wouldn't this change represent a disruptive way to do it because we are splitting the removal from the addition to different transactions? What do we gain from it? Wouldn't the normal EIP reconciliation handle it?
Maybe we should only consider it stale just when the EIP has no nodes assigned at all.

jcaamano:
> Also, assuming the zone controller can move an EIP non-disruptively (i.e removing the old LRP and adding the new one on the same transaction, or just updating the next hop in a transaction), wouldn't this change represent a disruptive way to do it because we are splitting the removal from the addition to different transactions? What do we gain from it? Wouldn't the normal EIP reconciliation handle it? Maybe we should only consider it stale just when the EIP has no nodes assigned at all.

It seems like the normal EIP reconciliation is handled with two separate transactions to remove and to add next hops as well so this is a moot point.

go-controller/pkg/ovn/egressip.go
pperiyasamy: yes, w.r,t `libovsdbclient.ErrNotFound` check, it's all same. but below change also parsing IPv6 mgmt port address in a dual stack cluster. previously it wasn't there.

go-controller/pkg/ovn/egressip.go
pperiyasamy: done.

go-controller/pkg/ovn/egressip.go
pperiyasamy: it was checking `Nexthop` field before and that will be always nil, so no logging for stale nexthop.
but with `Nexthops` field now, there will be an item present. so this check is not required. removed it.

pperiyasamy:
/retest-failed

The E2E test `Services of type NodePort [It] should work on secondary node interfaces for ETP=local and ETP=cluster when backend pods are also served by EgressIP [Feature:Service]` is flaked in https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18688181823/job/53292097779?pr=5493 while connecting secondary provider network into cluster node. not related to changes in PR.

```
2025-10-21T17:26:40.7039317Z [0mServices [38;5;243mof type NodePort [0m[1mshould work on secondary node interfaces for ETP=local and ETP=cluster when backend pods are also served by EgressIP[0m [38;5;204m[Feature:Service][0m
2025-10-21T17:26:40.7042216Z [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/service.go:974[0m
2025-10-21T17:26:40.7043313Z   [1mSTEP:[0m Creating a kubernetes client [38;5;243m@ 10/21/25 17:26:40.702[0m
2025-10-21T17:26:40.7044188Z   I1021 17:26:40.702946 48899 util.go:453] >>> kubeConfig: /home/runner/ovn.conf
2025-10-21T17:26:40.7045375Z   [1mSTEP:[0m Building a namespace api object, basename services [38;5;243m@ 10/21/25 17:26:40.703[0m
2025-10-21T17:26:40.7259963Z   [1mSTEP:[0m Waiting for a default service account to be provisioned in namespace [38;5;243m@ 10/21/25 17:26:40.725[0m
2025-10-21T17:26:40.7313910Z   [1mSTEP:[0m Waiting for kube-root-ca.crt to be provisioned in namespace [38;5;243m@ 10/21/25 17:26:40.731[0m
2025-10-21T17:26:40.7418583Z   [1mSTEP:[0m Creating the endpoints pod, one for each worker [38;5;243m@ 10/21/25 17:26:40.741[0m
2025-10-21T17:26:40.7551126Z   I1021 17:26:40.753686   48899 warnings.go:110] "Warning: would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"ovn-control-plane-ep-container\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container \"ovn-control-plane-ep-container\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"ovn-control-plane-ep-container\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"ovn-control-plane-ep-container\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
2025-10-21T17:26:44.7823969Z   I1021 17:26:44.781240   48899 warnings.go:110] "Warning: would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"ovn-worker-ep-container\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container \"ovn-worker-ep-container\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"ovn-worker-ep-container\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"ovn-worker-ep-container\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
2025-10-21T17:26:48.8352374Z   I1021 17:26:48.834504   48899 warnings.go:110] "Warning: would violate PodSecurity \"restricted:latest\": allowPrivilegeEscalation != false (container \"ovn-worker2-ep-container\" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container \"ovn-worker2-ep-container\" must set securityContext.capabilities.drop=[\"ALL\"]), runAsNonRoot != true (pod or container \"ovn-worker2-ep-container\" must set securityContext.runAsNonRoot=true), seccompProfile (pod or container \"ovn-worker2-ep-container\" must set securityContext.seccompProfile.type to \"RuntimeDefault\" or \"Localhost\")"
2025-10-21T17:26:52.8528850Z   [1mSTEP:[0m Choosing egressIP pod [38;5;243m@ 10/21/25 17:26:52.852[0m
2025-10-21T17:26:52.8529787Z   I1021 17:26:52.852602 48899 service.go:1008] EgressIP pod is services-8210/ovn-control-plane-ep
2025-10-21T17:26:52.8530881Z   [1mSTEP:[0m Label egress node create external container to send egress traffic to via secondary MultiNIC EIP [38;5;243m@ 10/21/25 17:26:52.852[0m
2025-10-21T17:26:52.8607897Z   [1mSTEP:[0m creating secondary network and attaching to all Nodes and external container [38;5;243m@ 10/21/25 17:26:52.858[0m
2025-10-21T17:26:53.1642098Z   I1021 17:26:53.163697 48899 service.go:1025] Unexpected error: network %!s(func() string=0x271fac0) must attach to node ovn-control-plane: 
2025-10-21T17:26:53.1643217Z       <*errors.errorString | 0xc000493710>: 
2025-10-21T17:26:53.1646625Z       failed to attach network to container ovn-control-plane: exit status 1 (Error response from daemon: failed to add interface vethad93f61 to sandbox: error setting interface "vethad93f61" IP to 10.10.10.2/24: cannot program address 10.10.10.2/24 in sandbox interface because it conflicts with existing route {Ifindex: 20 Dst: 10.10.0.0/16 Src: 10.10.0.2 Gw: <nil> Flags: [] Table: 254 Realm: 0}
2025-10-21T17:26:53.1648721Z       )
2025-10-21T17:26:53.1649014Z       {
2025-10-21T17:26:53.1655336Z           s: "failed to attach network to container ovn-control-plane: exit status 1 (Error response from daemon: failed to add interface vethad93f61 to sandbox: error setting interface \"vethad93f61\" IP to 10.10.10.2/24: cannot program address 10.10.10.2/24 in sandbox interface because it conflicts with existing route {Ifindex: 20 Dst: 10.10.0.0/16 Src: 10.10.0.2 Gw: <nil> Flags: [] Table: 254 Realm: 0}\n)",
2025-10-21T17:26:53.1658876Z       }
```

go-controller/pkg/ovn/egressip.go
```diff
@@ -1630,6 +1640,16 @@ func (e *EgressIPController) syncStaleEgressReroutePolicy(cache egressIPCache) e
...
+				allNetworkNodeRedirectCache, ok := cache.egressNodeRedirectsCache.cache[networkName]
+				if !ok || len(allNetworkNodeRedirectCache) == 0 {
+					klog.Infof("syncStaleEgressReroutePolicy deleting invalid logical router policy %q because there are no existing nodes assigned to its EgressIP %s", item.UUID, eipName)
+					return true
+				}
+				networkNodeRedirectCache := limitToValidEgressNodes(eipName, allNetworkNodeRedirectCache)
+				if len(networkNodeRedirectCache) == 0 {
+					klog.Infof("syncStaleEgressReroutePolicy deleting invalid logical router policy %q because there are no existing nodes assigned to its EgressIP %s", item.UUID, eipName)
+					return true
+				}
```
jcaamano: any reason not to apply my suggestion if the error message is going to be the same?

go-controller/pkg/ovn/egressip.go
jcaamano: I see the difference now, not equivalent to the continue we had before, we don't add the mgmt IPs but we do add the gateway and transit switch IPs which would otherwise not be added before.

go-controller/pkg/ovn/egressip.go
pperiyasamy: ah! ok, i thought it only about changing log message, but it also suggests to invoke `limitToValidEgressNodes` with `cache.egressNodeRedirectsCache.cache[networkName]`, missed that, will update.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -13880,6 +13881,247 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+						return fmt.Sprintf("\"%s\":\"%s\"", key, value)
+					}
+					nodeAnnotations := map[string]string{
+						"k8s.ovn.org/l3-gateway-config":               fmt.Sprintf(`{"default":{"mode":"shared","mac-address":"7e:57:f8:f0:3c:49", "ip-address":"%s", "next-hop":"192.168.126.1"}}`, n.primaryINFIP+mask),
+						"k8s.ovn.org/node-primary-ifaddr":             fmt.Sprintf("{%s}", addValueToIPFamilyKey(n.primaryINFIP+mask)),
+						"k8s.ovn.org/node-subnets":                    fmt.Sprintf("{\"default\":[\"%s\"]}", n.podSubnet),
+						"k8s.ovn.org/node-transit-switch-port-ifaddr": fmt.Sprintf("{%s}", addValueToIPFamilyKey(n.transitSWIP+mask)),
+						"k8s.ovn.org/zone-name":                       n.nodeName,
+						util.OVNNodeHostCIDRs:                         fmt.Sprintf("[\"%s\"]", n.primaryINFIP+mask),
+					}
```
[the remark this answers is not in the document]
pperiyasamy: it's okay to use same mask for the unit test. so this change is not needed.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -13880,6 +13881,247 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+					},
+					&egressipv1.EgressIPList{
+						Items: []egressipv1.EgressIP{eIPObj},
+					},
+				)
+				i, podIPv4Net, _ := net.ParseCIDR(podIP + mask)
+				podIPv4Net.IP = i
+				fakeOvn.controller.logicalPortCache.add(egressPod, "", types.DefaultNetworkName, "",
+					nil, []*net.IPNet{podIPv4Net})
+
```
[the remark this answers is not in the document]
pperiyasamy: same comment as before.

jcaamano (approved):

PR #5501, opened 2025-08-19
Improve ACL logging file size monitor

martinkennelly:
Downstream, its logging a lot that the file
doesn't exist. If file doesn't exist, set
file size metric to zero. Do not emit error logs.
My understanding is we dont always expect this file to exist. 
Downstream, we rotate the acl log file and its expectd
that it may not exist.

I see it in the logs contineously and its an easy fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Log file size metric now reports 0 when the file is missing, instead of failing and emitting errors.
  * Reduces noisy logs by ignoring expected “not found” cases during checks.
  * Adds early validation: skips starting the watcher when the log path is empty or the directory doesn’t exist, preventing unnecessary errors and improving stability of metric updates.

### DIFF
--- a/go-controller/pkg/metrics/metrics.go
+++ b/go-controller/pkg/metrics/metrics.go
@@ -159,16 +159,32 @@ func ovnKubeLogFileSizeMetricsUpdater(ovnKubeLogFileMetric *prometheus.GaugeVec,
 	defer ticker.Stop()
 
 	logfile := config.Logging.File
+	// only start the file watcher if the log file directory is valid
+	if logfile == "" {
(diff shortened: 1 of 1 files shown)


martinkennelly:
@trozet any interest in this?

martinkennelly:
@girishmg PTAL

martinkennelly:
@girishmg youre the assignee for this PR so I am going to follow up with you. LMK if you cannot review this.

trozet:
thanks @martinkennelly this has bothered me in the logs for a while

PR #5506, opened 2025-08-20
Fix getHostNamespaceAddressesForNode error wrapping

jcaamano:
Incorrect wrapping of errors in getHostNamespaceAddressesForNode prevented callers to assert the type of error returned. In this specific case the caller delIPFromHostNetworkNamespaceAddrSet could not appropriately check that the error returned was AnnotationNotSetError and was failing a deletion flow that should be noop instead.

This was triggered by [1] changes that cause the deletion flow to depend on annotations set by different controllers instead of annotations set atomically by a single controller, making it possible that some annotations were set but not others and opening up a new error handling flow that wasn't possible before. The scenario is a node that is added and deleted quickly enough for some controller in cluster manager to be able to annotate the node and some not.

[1] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5396

## Summary by CodeRabbit

- **Refactor**
  - Improved internal error handling and diagnostics when determining node gateway/router addresses, including clearer wrapped error reporting while preserving existing public behavior and APIs.

- **Tests**
  - Added table-driven controller tests to cover node-delete retry behavior across multiple node annotation scenarios, increasing coverage and reducing risk of regressions.

### DIFF
--- a/go-controller/pkg/ovn/namespace.go
+++ b/go-controller/pkg/ovn/namespace.go
@@ -376,16 +376,17 @@ func (oc *DefaultNetworkController) getHostNamespaceAddressesForNode(node *corev
 	}
 	// for shared gateway mode we will use LRP IPs to SNAT host network traffic
 	// so add these to the address set.
-	lrpIPs, err := udn.GetGWRouterIPs(node, oc.GetNetInfo())
-	if err != nil {
-		if util.IsAnnotationNotSetError(err) {
-			// FIXME(tssurya): This is present for backwards compatibility
-			// Remove me a few months from now
-			var err1 error
-			lrpIPs, err1 = util.ParseNodeGatewayRouterLRPAddrs(node)
-			if err1 != nil {
-				return nil, fmt.Errorf("failed to get join switch port IP address for node %s: %v/%v", node.Name, err, err1)
-			}
+	lrpIPs, gwIPsErr := udn.GetGWRouterIPs(node, oc.GetNetInfo())
+	if gwIPsErr != nil {
+		if !util.IsAnnotationNotSetError(gwIPsErr) {
+			return nil, gwIPsErr
+		}
+		// FIXME(tssurya): This is present for backwards compatibility
+		// Remove me a few months from now
+		var lrpAddrsErr error
+		lrpIPs, lrpAddrsErr = util.ParseNodeGatewayRouterLRPAddrs(node)
+		if lrpAddrsErr != nil {
+			return nil, fmt.Errorf("failed to fallback to annotations after error %q: %w", gwIPsErr, lrpAddrsErr)
 		}
 	}
 
--- a/go-controller/pkg/ovn/master_test.go
+++ b/go-controller/pkg/ovn/master_test.go
@@ -1566,45 +1566,62 @@ var _ = ginkgo.Describe("Default network controller operations", func() {
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	})
 
-	ginkgo.It("doesn't retry deleting a node that doesn't have a node-subnet annotation", func() {
-		app.Action = func(ctx *cli.Context) error {
-			_, err := config.InitConfig(ctx, nil, nil)
-			gomega.Expect(err).NotTo(gomega.HaveOccurred())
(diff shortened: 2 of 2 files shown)


jcaamano:
> **Actionable comments posted: 0**
> 
> 🧹 Nitpick comments (1)
> > go-controller/pkg/ovn/namespace.go (1)> `385-388`: **Optional: consider stdlib errors.Join if toolchain permits**
> > > If the project is on Go 1.20+ and there’s no repo-wide constraint to use utilerrors.Join, consider using errors.Join for consistency with stdlib semantics. Current change is fine as-is.
> > > ```diff
> > > - return nil, fmt.Errorf("failed to get join switch port IP address for node %s: %w", node.Name, utilerrors.Join(err, err1))
> > > + return nil, fmt.Errorf("failed to get join switch port IP address for node %s: %w", node.Name, errors.Join(err, err1))
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > (Remember to import "errors" if you go this route.)
> 
> 📜 Review details

utilerrors.Join provides single line formatting which is more convenient to us. stdlib errors.Join does multi-line formatting which we don't like.

go-controller/pkg/ovn/namespace.go
```diff
@@ -377,17 +377,18 @@ func (oc *DefaultNetworkController) getHostNamespaceAddressesForNode(node *corev
...
+		// FIXME(tssurya): This is present for backwards compatibility
+		// Remove me a few months from now
+		var parseErr error
+		lrpIPs, parseErr = util.ParseNodeGatewayRouterLRPAddrs(node)
+		if parseErr != nil {
+			err = fmt.Errorf("failed to fallback to annotations after error %q: %w", err, parseErr)
 		}
 	}
+	if err != nil {
+		return nil, err
```
kyrtapz: if `util.IsAnnotationNotSetError(err)` is true and `util.ParseNodeGatewayRouterLRPAddrs(node)` returns a nil error we will still exit here. This is different from the previous behavior and seems wrong.

jcaamano: ooops

go-controller/pkg/ovn/namespace.go
```diff
@@ -376,16 +376,17 @@ func (oc *DefaultNetworkController) getHostNamespaceAddressesForNode(node *corev
...
-			// Remove me a few months from now
-			var err1 error
-			lrpIPs, err1 = util.ParseNodeGatewayRouterLRPAddrs(node)
-			if err1 != nil {
-				return nil, fmt.Errorf("failed to get join switch port IP address for node %s: %v/%v", node.Name, err, err1)
-			}
+	lrpIPs, gwIPsErr := udn.GetGWRouterIPs(node, oc.GetNetInfo())
+	if gwIPsErr != nil {
+		if !util.IsAnnotationNotSetError(gwIPsErr) {
+			return nil, gwIPsErr
```
kyrtapz, line 382: NOTE: While I believe it is correct the additional change in behavior here is that we are now returning for every error that is not `IsAnnotationNotSetError` while previously we just ignored those and returned `nil, nil`.

kyrtapz (dismissed):
LGTM
@npinaeva FYI

kyrtapz:
I think this flow could use a test(successful delete of a node with incomplete annotations) it could even be a unit-test if possible. @jcaamano let me know if it is worth a followup we can ask the rabbit to create an issue.

jcaamano:
> I think this flow could use a test(successful delete of a node with incomplete annotations) it could even be a unit-test if possible. @jcaamano let me know if it is worth a followup we can ask the rabbit to create an issue.

Added! Just copied the one we had for missing subnet annotations but now with missing node-id annotation.

jcaamano:
> `1609-1649`: **Reduce duplication with prior test that handles missing node-subnets**
> 
> This new test largely mirrors the earlier “no node-subnet annotation” test. Consider a small table-driven DescribeTable with the differing annotations as inputs to keep the suite DRY and faster to scan.

Ah heck, I came back because I knew this was coming and couldn't sleep with myself.

PR #5507, opened 2025-08-20
Ensure that UDN update events trigger updates for NAD annotations correctly

trozet:
With b05875b1d671edb696455ef517d9a4ba1e156b9a we introduced copying the UDN annotations to the NAD, except for internal annotations (k8s.ovn.org). However, this patch missed doing this on update events.

Props to @jcaamano for finding this issue!

## Summary by CodeRabbit

* **Bug Fixes**
  * NetworkAttachmentDefinition updates now preserve platform/internal annotations and merge them with user annotations; controller skips updates unless config, labels, or annotations actually change.

* **Tests**
  * Added reconciliation tests confirming annotation changes propagate to NADs across namespaces while preserving internal annotations and keeping network configuration stable.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	netv1 "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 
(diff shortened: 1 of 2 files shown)


cathy-zhou:
lgtm. thanks for catching this.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -76,12 +77,24 @@ func (c *Controller) updateNAD(obj client.Object, namespace string) (*netv1.Netw
...
-	if reflect.DeepEqual(nadCopy.Spec.Config, desiredNAD.Spec.Config) && reflect.DeepEqual(nadCopy.ObjectMeta.Labels, desiredNAD.ObjectMeta.Labels) {
+	// NAD update path, need to merge internal (k8s.ovn.org) current annotations with desired
+	for k, v := range nadCopy.Annotations {
+		if strings.HasPrefix(k, types.OvnK8sPrefix) {
+			if desiredNAD.Annotations == nil {
+				desiredNAD.Annotations = make(map[string]string)
+			}
+			desiredNAD.Annotations[k] = v
+		}
+	}
```
jcaamano, line 88: So what we are saying is that the UDN owns all annotations on the NAD except the OVNK annotations, instead of just owning the annotations that are set on the UDN.

Lets say UDN has annotations A,B,C. Let's say NAD has annotations A,B,D. Let's say none of these annotations are OVNK annotations. The end result would be NAD with annotations A,B,C instead of NAD with annotations A,B,C,D. 

So nobody except OVNK can set and upkeep annotations on NADs owned by UDNs. 

Not seeing a specific problem here just want to make sure this is on purpose.

trozet: yeah that is correct. This behavior is only for NAD created by UDN. My thinking is that NAD is not exposed to a user. A user interfaces with a UDN, so they should never modify the NAD dervied from teh UDN directly. That was created by ovnk.

jcaamano (approved):

martinkennelly:
We need docs update to say we own the nads annotations only when nads are used to support udn config? We may consider Nads generated as internal but user can query. Can be done as follow up if you think its good idea.

jcaamano:
> We need docs update to say we own the nads annotations only when nads are used to support udn config? We may consider Nads generated as internal but user can query. Can be done as follow up if you think its good idea.

@trozet would you consider updating the docs in a follow up commit?

PR #5510, opened 2025-08-21
fix: drop prometheus operator CRDs and make CRs configurable

### LINKED ISSUE #5509: Helm chart installs out-of-date CRDs
### What happened?

Installed the OVN Kubernetes helm chart on a cluster which already had the Prometheus Operator installed. The helm chart installed the ServiceMonitor and PrometheusRule CRDs overwriting those deployed by the Prometheus Operator. In my case there was no downside except confusion but this could potentially cause very bad outcomes.

The current helm chart has an implicit strict dependency on a specific version of the Prometheus Operator. 

### What did you expect to happen?

OVN Kubernetes should not install CRDs that come from other projects without strict versioning and documentation.

### How can we reproduce it (as minimally and precisely as possible)?

Install onv-kubernetes using the helm chart and see the CRDs are installed.

The templates are here:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/crds/monitoring.coreos.com_prometheusrule.yaml
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/crds/monitoring.coreos.com_servicemonitor.yaml

These should probably be removed from the helm chart. The related CRs could be put behind a flag - I believe there aren't essential to OVN-Kubernetes monitoring and are instead an opinionated configuration based on the Prometheus Operator.

### OVN-Kubernetes version

main

### Kubernetes version

--

### OVN version

--

### OVS version

--

### Platform

All

### OS version

All

### Install tools

<details>
helm
</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

killianmuldoon:
## 📑 Description

Remove the `monitoring.coreos.io` CRD  installation from the helm chart and add flags to allow installation on systems that dont' use the Prometheus Operator.

Fixes #5509 

## ✅ Checks

- [ ] My code requires tests


helm/ovn-kubernetes/values-no-ic.yaml
```diff
@@ -181,3 +181,8 @@ monitoring:
   # in defining that.
   commonServiceMonitorSelectorLabels:
     release: kube-prometheus-stack
+  # -- deploy ServiceMonitors for service discovery using the Prometheus Operator
+  enableServiceMonitor: true
```
kyrtapz: Having these as true by default won't work now that you removed the prometheus CRDs.

killianmuldoon: I can default them to false if you prefer.

In the previous case these could be applied but were not functional unless the prometheus operator was installed.
Today installation will fail if this is true and the prometheus operator is not installed.

I could do a more direct check for the presence of the Operator if that's better? That pattern is already in a chart here: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/56d14a3074ba7df00a6b68dfe4ac60aff7fc624a/helm/sdn-dashboard/templates/is-prometheus-installed.yaml#L2

kyrtapz: > I could do a more direct check for the presence of the Operator if that's better? That pattern is already in a chart here

That would be great! We could move defaults to false and ensure prometheus is installed if they were set to true.
As a side note we are rebasing our PR branches on master instead of merging master to it.

killianmuldoon: > As a side note we are rebasing our PR branches on master instead of merging master to it.

Looks like the merge is the automatic behaviour of the GitHub "Update branch" button? First time I've used it I think.

I think the current version is what you're asking for. You can test it with something like:

```bash
helm template ovk helm/ovn-kubernetes \
  --values=helm/ovn-kubernetes/values-no-ic.yaml \
  --set monitoring.enablePrometheusRule=true \
  --set monitoring.enableServiceMonitor=true \
  --api-versions=monitoring.coreos.com/v1/ServiceMonitor \
  --api-versions=monitoring.coreos.com/v1/PrometheusRule
```

and turning on or off the values and available apiVersions.

helm/ovn-kubernetes/templates/ovnkube-alerts.yaml
```diff
@@ -387,4 +392,4 @@ spec:
       annotations:
        message: |
          {{`ovnkube clustermanager allocated IPv6 host subnets value for {{ $labels.network_name }} network should be same as the number of K8s nodes`}}
-
+{{ end }}
```
kyrtapz: NIT: missing new line

helm/ovn-kubernetes/templates/ovnkube-monitor.yaml
```diff
@@ -25,58 +29,85 @@ spec:
...
     matchNames:
-    - ovn-kubernetes
+      - ovn-kubernetes
   selector:
     matchLabels:
-      k8s-app: ovnkube-node
+      k8s-app: ovnkube-cluster-manager
+{{ end }}
+
+---
```
kyrtapz, line 88: Why is that service always applied even if `enableServiceMonitor` is false?
My guess would be that we want to guard everything in this file.

killianmuldoon: The ServiceMonitor is specific to the Prometheus Operator implementation - but this service could be used to collect the underlying metrics even if the Prometheus Operator isn't installed e.g. using builtin Prometheus scrapeConfiguration, or using some other metrics collector.

killianmuldoon: I think you're right that it's better to keep the whole set of objects inside the gate here though - people using other metrics collection approaches can deploy services etc. themselves.

helm/ovn-kubernetes/values-no-ic.yaml
```diff
@@ -181,3 +181,8 @@ monitoring:
   # in defining that.
   commonServiceMonitorSelectorLabels:
     release: kube-prometheus-stack
+  # -- deploy ServiceMonitors for service discovery using the Prometheus Operator
+  enableServiceMonitor: false
+  # -- deploy PrometheusRules for specific metric collection using the Prometheus Operator
+  enablePrometheusRule: false
+
```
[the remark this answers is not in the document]
killianmuldoon: They don't - they're on the same level as `commonServiceMonitorSelectorLabels`

helm/ovn-kubernetes/values-single-node-zone.yaml
```diff
@@ -167,3 +167,8 @@ monitoring:
   # in defining that.
   commonServiceMonitorSelectorLabels:
     release: kube-prometheus-stack
+  # -- deploy ServiceMonitors for service discovery using the Prometheus Operator
+  enableServiceMonitor: false
+  # -- deploy PrometheusRules for specific metric collection using the Prometheus Operator
+  enablePrometheusRule: false
+
```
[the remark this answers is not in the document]
killianmuldoon: They don't - they're on the same level as commonServiceMonitorSelectorLabels

npinaeva:
I can't figure out if anyone is using these rules at all and why they are different for kind.sh vs help install, added this item to the next meeting agenda, hope to get more clarity

killianmuldoon:
>I can't figure out if anyone is using these rules at all and why they are different for kind.sh vs help install, added this item to the next meeting agenda, hope to get more clarity

I'll do my best to join the next meeting.

l8huang:
Should these ServiceMonitor and PromethuesRules resources be moved from `helm/ovn-kubernetes` to `helm/sdn-dashboard`? They are not required for running ovn-k8s.

FYI: we are working on rename`helm/sdn-dashboard` to `helm/sdn-monitoring` and move all observability(prometheus, grafana, alert manager) related manifests into it, so alert and dashboard changes can be deployed separately.

l8huang:
> I can't figure out if anyone is using these rules at all 

BTW: how does Red Hat manage promethues config, alert rules and dashboards? do you use Prometheus operator and Grafana operator?

helm/ovn-kubernetes/values-multi-node-zone.yaml
```diff
@@ -166,3 +166,7 @@ monitoring:
   # in defining that.
   commonServiceMonitorSelectorLabels:
     release: kube-prometheus-stack
+  # -- deploy ServiceMonitors for service discovery using the Prometheus Operator
+  enableServiceMonitor: false
+  # -- deploy PrometheusRules for specific metric collection using the Prometheus Operator
+  enablePrometheusRule: false
```
l8huang, line 172: IMO: these values and the related manifests can be removed by this PR, please see https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5510#issuecomment-3347832347

killianmuldoon: Happy to do this - but I think we need direction from @girishmg on what the desired state is here as discussed on this week's community meeting.

killianmuldoon:
The issue for this one was closed as stale due to inactivity - there seems to be no consensus on how to deal with these CRDs.

kyrtapz:
> The issue for this one was closed as stale due to inactivity - there seems to be no consensus on how to deal with these CRDs.

@npinaeva @girishmg I think this is a valuable change that fixes an issue in the helm charts we provide, can you ptal?

killianmuldoon:
@npinaeva could we consider merging this one? If not I can close it as stale / not desired.

npinaeva:
sorry for such a long wait on our side. Since we don't seem to get much feedback on this PR, I will use my own judgement and merge this. I think any further improvements can be done as a follow up if needed.
Thanks @killianmuldoon  and @kyrtapz for your work and patience!

PR #5523, opened 2025-08-26
OKEP for running DPU and Host OVNK Instances

### LINKED ISSUE #5521: Enabling DPU and Host OVN-Kubernetes in a single cluster

When we first designed DPU support in OVN-Kubernetes, we designed it in a way that all traffic leaving the host would traverse the DPU. In other words, as pods come up on the host, OVNK on the DPU (serving the host) would then wire the pod's VF representors into OVS and handle configuring OVN. This design was short sighted, as there exists use cases where I may want some of my traffic to go via the DPU and some of the traffic to go via other NICs on my host.

To enable this architecture, we need full OVNK running on the host itself to handle the interfaces on the host, with its own OVS, as well as OVNK running on the DPU (serving the host). We also still need the DPU Host component that runs in the host on behalf of the DPU. Both of these OVNK are part of the same K8S cluster, and they both listen to events from kapi server.

### Why is this needed?

We want to be able to segment traffic so that some goes via DPU while some goes by interfaces on the host. Naturally it makes sense to do this on a per UDN basis.

trozet:
Fixes: #5521

## Summary by CodeRabbit

* Documentation
  * Added an enhancement proposal describing Single Cluster Host+DPU segmentation with dual OVN-Kubernetes instances per node, identity-based routing, UDN/CDN handling, a shared CNI shim strategy, and multi-network pod workflows. Includes architecture overview, goals/non-goals, user stories, and placeholders for API, testing, and docs.
  * Updated documentation navigation to include the new proposal under Enhancement Proposals.

### DIFF
--- a/docs/okeps/okep-5521-single-cluster-host-dpu.md
+++ b/docs/okeps/okep-5521-single-cluster-host-dpu.md
@@ -1,0 +1,308 @@
+# OKEP-5521: Single Cluster Host+DPU
+
+* Issue: [#5521](https://github.com/ovn-org/ovn-kubernetes/issues/5521)
+
+## Problem Statement
+
+When we first designed DPU support in OVN-Kubernetes, we designed it in a way that all traffic leaving the host would
+traverse the DPU. In other words, as pods come up on the host, OVNK on the DPU (serving the host) would then wire the
+pod's VF representors into OVS and handle configuring OVN. This design was short-sighted, as there exists use cases
+where I may want some of my traffic to go via the DPU and some of the traffic to go via other NICs on my host.
+
+To enable this architecture, we need full OVNK running on the host itself to handle the interfaces on the host, with
+its own OVS, as well as OVNK running on the DPU (serving the host). We also still need the DPU Host component that runs
+in the host on behalf of the DPU. Both of these OVNK are part of the same K8S cluster, and they both listen to events
+from kapi server.
+
+## Goals
+
+* Allow two OVNK instances to serve a single node. One for DPU and one for the Host.
+* Allow segmenting UDNs via the host or DPU.
+* Avoid each instance doing unnecessary work if the event only affects the other OVNK instance.
+
+## Non-Goals
+
+* Addressing shared OVN, launching pods on the DPU itself, SFC, or any other part of the original 2 cluster DPU design.
+* How to run OVNK on the DPU without kubelet on the DPU.
+
+## Introduction
+
+DPUs offer hardware offloading with OVN-Kubernetes. See [dpu-support](../features/hardware-offload/dpu-support.md) for more details.
+The DPU runs most of OVN-Kubernetes down in the DPU itself, while on the host a small DPU-Host OVNK pod runs:
+
+```mermaid
+flowchart TB
+
+  subgraph HOST[Host]
+    direction TB
+    H1["ovn-kubernetes (DPU-Host mode)"]
+  end
+
+  subgraph DPU[DPU]
+    direction TB
+    D1["ovn-kubernetes (full mode)"]
+    D2[OVS]
+    D3[OVN]
+    D1 --> D2
+    D1 --> D3
+  end
+
+  HOST <--> DPU
+```
+
+When pods on the host are launched, DPU-Host figures out the SRIOV interface for the pod and then annotates device
+information to the pod, so that OVNK on the DPU can then wire the VF Representor into OVS and configure OVN. Now, let's
+consider a use case where we have additional interfaces on the host that we want to leverage:
+
+```mermaid
+flowchart TB
+
+  subgraph HOST[Host]
+    direction TB
+    H1["ovn-kubernetes (DPU-Host mode)"]
+    H2["ovn-kubernetes (Full mode)"]
+    H3[OVS]
+    H4[OVN]
+    H2-->H3
+    H2-->H4
+    H5[NIC]
+    H3-->H5
+  
+  end
+
+  subgraph DPU[DPU]
+    direction TB
+    D1["ovn-kubernetes (Full mode)"]
+    D2[OVS]
+    D3[OVN]
+    D4[NIC]
+    D1 --> D2
+    D1 --> D3
+    D2 --> D4
+  end
+
+  HOST <--> DPU
+```
+
+In this design, a pod may have some interfaces served via the VFs on the DPU, while other interfaces are connected via
+the OVNK running on the host. From this point forward, the OVNK running on behalf of the host will be known as "OVNK-Full Host",
+while the OVNK running on the DPU will be called "OVNK-Full DPU". The DPU-Host component will use its regular naming convention.
+
+
+## User-Stories/Use-Cases
+
+Story 1: Segment UDNs via different NICs
+
+As a cluster admin, I want certain workloads to utilize the DPU for primary UDN/cluster default network access, while
+I want to leverage my host based offloadable NICs for secondary UDNs. This will allow me to use my DPU for networks that
+have north/south capabilities, while only using my host based NICs for east/west acceleration.
+
+## Proposed Solution
+
+### Running multiple OVNK instances on the same node
+
+The first thing that must be addressed is how to actually run two OVN-Kubernetes instances that reference the same
+Kubernetes node object. OVN-Kubernetes leverages annotations on the node to indicate key information about underlay network
+and OVN configuration for the node. Some of these that are unique to the instance of OVN itself include:
+
+* k8s.ovn.org/node-encap-ips
+* k8s.ovn.org/node-chassis-id
+* k8s.ovn.org/l3-gateway-config
+
+These annotations may be different between OVNK-Full Host and OVNK-Full DPU processes. Therefore, one solution is to
+duplicate these annotations and make them specific to the DPU, for example, "dpu-encap-ips". This solution is short-sighted
+and will not accommodate for potentially in the future, using multiple DPUs. Therefore, the only other viable solution is
+to store this information in its own CR. This solution involves two potential paths:
+
+1. Use the existing node CRD and create DPU specific CRs
+2. Create a new DPUNode CRD type that mirrors the format of the Node CRD.
+
+For simplicity, it seems like going with option 1 should be the first choice, unless we hit a complication that forces us
+to use option 2.
+
+#### Creating extra node CRs for DPU instances
+
+The duplicate node CR will be created by OVNK-Full DPU instance that will realize on start up it is running with a different
+ovnk-identity. Let's assume the ovnk-identity is "dpu1-ovnk". OVNK-Full DPU will be started with a new configuration option
+--ovnk-identity "dpu1-ovnk". After start up, OVNK-Full DPU will start a controller in its node controller, which will
+watch for its node and mirror/mutate the information into a new Node CR. The name of this node CR will be
+<original node name>-<ovnk-identity>. For example, given a Node:
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  annotations:
+    k8s.ovn.org/host-cidrs: '["172.18.0.3/16","172.19.0.2/16"]'
+    k8s.ovn.org/l3-gateway-config: '{"default":{"mode":"shared","bridge-id":"breth0","interface-id":"breth0_ovn-worker","mac-address":"c6:a8:ac:29:8c:d0","ip-addresses":["172.18.0.3/16"],"ip-address":"172.18.0.3/16","next-hops":["172.18.0.1"],"next-hop":"172.18.0.1","node-port-enable":"true","vlan-id":"0"}}'
+    k8s.ovn.org/node-chassis-id: b4142d71-22dd-4ed9-b8d9-4dd187e9b0b0
+    k8s.ovn.org/node-encap-ips: '["172.18.0.3"]'
+    k8s.ovn.org/node-id: "3"
+    k8s.ovn.org/node-masquerade-subnet: '{"ipv4":"169.254.0.0/17","ipv6":"fd69::/112"}'
+    k8s.ovn.org/node-primary-ifaddr: '{"ipv4":"172.18.0.3/16"}'
+    k8s.ovn.org/node-subnets: '{"default":["10.244.2.0/24"]}'
+    k8s.ovn.org/node-transit-switch-port-ifaddr: '{"ipv4":"100.88.0.3/16"}'
+    k8s.ovn.org/remote-zone-migrated: ovn-worker
+    k8s.ovn.org/zone-name: ovn-worker
+    kubeadm.alpha.kubernetes.io/cri-socket: unix:///run/containerd/containerd.sock
+    node.alpha.kubernetes.io/ttl: "0"
+    volumes.kubernetes.io/controller-managed-attach-detach: "true"
+  creationTimestamp: "2025-08-25T22:12:56Z"
+  labels:
+    beta.kubernetes.io/arch: amd64
+    beta.kubernetes.io/os: linux
+    k8s.ovn.org/zone-name: ovn-worker
+    kubernetes.io/arch: amd64
+    kubernetes.io/hostname: ovn-worker
+    kubernetes.io/os: linux
+  name: ovn-worker
+  resourceVersion: "102658"
+  uid: 48ea7c29-ee34-48c1-b8a9-43e3db34c007
+spec:
+  podCIDR: 10.244.2.0/24
+  podCIDRs:
+  - 10.244.2.0/24
+```
+
+OVNK-Full DPU will create a new Node and update as needed:
+
+```yaml
+apiVersion: v1
+kind: Node
+metadata:
+  annotations:
+    k8s.ovn.org/identity: dpu1-ovnk
+    k8s.ovn.org/host-cidrs: '["172.18.0.3/16","172.19.0.2/16"]'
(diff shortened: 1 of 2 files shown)


docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+where I may want some of my traffic to go via the DPU and some of the traffic to go via other NICs on my host.
+
+To enable this architecture, we need full OVNK running on the host itself to handle the interfaces on the host, with
+its own OVS, as well as OVNK running on the DPU (serving the host). We also still need the DPU Host component that runs
+in the host on behalf of the DPU. Both of these OVNK are part of the same K8S cluster, and they both listen to events
+from kapi server.
+
+## Goals
+
+* Allow two OVNK instances to serve a single node. One for DPU and one for the Host.
```
booxter, line 19: While at it, consider scenario where we have more than two OVNKs, e.g. with multiple DPUs?

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+For simplicity, it seems like going with option 1 should be the first choice, unless we hit a complication that forces us
+to use option 2.
+
+#### Creating extra node CRs for DPU instances
+
+The duplicate node CR will be created by OVNK-Full DPU instance that will realize on start up it is running with a different
+ovnk-identity. Let's assume the ovnk-identity is "dpu1-ovnk". OVNK-Full DPU will be started with a new configuration option
+--ovnk-identity "dpu1-ovnk". After start up, OVNK-Full DPU will start a controller in its node controller, which will
+watch for its node and mirror/mutate the information into a new Node CR. The name of this node CR will be
+<original node name>-<ovnk-identity>. For example, given a Node:
```
booxter, line 129: it doesn't render properly in github for me. Use back ticks?

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+    H2-->H3
+    H2-->H4
+    H5[NIC]
+    H3-->H5
+  
+  end
+
+  subgraph DPU[DPU]
+    direction TB
+    D1["ovn-kubernetes (Full mode)"]
```
crnithya, line 75: On the DPU it will be DPU-mode

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+flowchart TB
+
+  subgraph HOST[Host]
+    direction TB
+    H1["ovn-kubernetes (DPU-Host mode)"]
+  end
+
+  subgraph DPU[DPU]
+    direction TB
+    D1["ovn-kubernetes (full mode)"]
```
crnithya, line 43: On the DPU it will be DPU-mode

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+spec:
+  podCIDR: 11.244.2.0/24
+  podCIDRs:
+  - 11.244.2.0/24
+```
+
+In the new CR, a new annotation is added "k8s.ovn.org/identity", this will be used by OVNK-Full DPU controllers
+to filter only on these node events. OVNK-Full Host will ignore nodes that have this identity field populated.
+
+TODO: figure out which values in the above would actually change.
```
booxter, line 209: (nit) Once you have the list, it would be helpful to the reader to see a diff specifically.

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+* k8s.ovn.org/node-encap-ips
+* k8s.ovn.org/node-chassis-id
+* k8s.ovn.org/l3-gateway-config
+
+These annotations may be different between OVNK-Full Host and OVNK-Full DPU processes. Therefore, one solution is to
+duplicate these annotations and make them specific to the DPU, for example, "dpu-encap-ips". This solution is short-sighted
+and will not accommodate for potentially in the future, using multiple DPUs. Therefore, the only other viable solution is
+to store this information in its own CR. This solution involves two potential paths:
+
+1. Use the existing node CRD and create DPU specific CRs
```
booxter, line 117: There's a third option that won't involve duplicating `Node`s and dealing with any fallout from always-`DOWN` `Node`s: reuse the existing `Node` object but introduce a scheme to encode OVN-K instance specific information. You mentioned `dpu-encap-ips` and that it would be shortsighted for multi-DPU - I agree. But the shortcoming is only due to the fact that this annotation would be hardcoded.

Instead of hardcoding it, one could make the OVNK instance "identity" part of the annotation key, for example:

```
Kind: Node
metadata:
  annotations:
    <dpu1-identity>-encap-ips: ...
    <dpu2-identity>-encap-ips: ...
    <dpu3-identity>-encap-ips: ...
    <host-identity>-encap-ips: ...
    encap-ips: ...
```

Then, depending on which annotations are present, OVNK instances either fall back to the "default" (`encap-ips`) annotation or rely on their identity-specific annotations (`dpu1-encap-ips`).

---

There's a precedent for this in "OVN World". [This is how](https://github.com/ovn-org/ovn/blob/0935bd0fa6afa24ce013c08a2796758cd26d196b/controller/ovn-controller.8.xml#L413-L435) `ovn-controller` distinguishes between configurations that belong to multiple chassis served from the same host. (Very similar scenario to what we are dealing with here. Albeit more on logical plane - sharing configuration "database" - than on physical level (DPU OVNK is *not* colocated with host's).

(I'll quote below for convenience.)

```
    <p>
        Most of configuration options listed above can also be set for a
        particular chassis name (see <code>external_ids:system-id </code>
        for more information). This can be achieved by setting
        <code>external_ids:option-[chassis]</code> instead of
        <code>external_ids:option</code>. For example, set
        <code>external_ids:ovn-encap-ip-otherhv</code> to use a particular
        IP address for the controller instance named <code>otherhv</code>.
        Name specific configuration options always override any global options
        set in the database.
    </p>

    <p>
        Chassis-specific configuration options in the database plus the ability
        to configure the chassis name to use via the
        <code>system-id-override</code> file or command line allows to run
        multiple <code>ovn-controller</code> instances with unique chassis
        names on the same host using the same <code>vswitchd</code> instance.
        This may be useful when running a hybrid setup with more than one CMS
        managing ports on the host, or to use different datapath types on the
        same host. Also note that this ability is highly experimental and
        has known limitations (for example, stateful ACLs are not supported).
        Use at your own risk.
    </p>

```

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+  podCIDRs:
+  - 11.244.2.0/24
+```
+
+In the new CR, a new annotation is added "k8s.ovn.org/identity", this will be used by OVNK-Full DPU controllers
+to filter only on these node events. OVNK-Full Host will ignore nodes that have this identity field populated.
+
+TODO: figure out which values in the above would actually change.
+
+The node status will remain down, as kubelet will not be running in the DPU. But this status should be ignored by OVNK-Full DPU.
```
booxter, line 211: I am concerned this can spook existing monitoring - or require its adjustment to special-handle these fake "Nodes". I wonder if instead of claiming the Node down we could make it resource-less (no memory or cpu capacity) - would it then make the scheduler smart enough to not attempt to put pods on these "Nodes"?

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+2. Create a new DPUNode CRD type that mirrors the format of the Node CRD.
+
+For simplicity, it seems like going with option 1 should be the first choice, unless we hit a complication that forces us
+to use option 2.
+
+#### Creating extra node CRs for DPU instances
+
+The duplicate node CR will be created by OVNK-Full DPU instance that will realize on start up it is running with a different
+ovnk-identity. Let's assume the ovnk-identity is "dpu1-ovnk". OVNK-Full DPU will be started with a new configuration option
+--ovnk-identity "dpu1-ovnk". After start up, OVNK-Full DPU will start a controller in its node controller, which will
```
booxter, line 127: (Observation) This introduction of "OVNK identity" is **the** key missing piece we were struggling without in #5376. Once OVNKs are self-aware, the exact shape how this identity is mapped onto particular instance-specific configuration becomes more situational / accidental.

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+1. Use the existing node CRD and create DPU specific CRs
+2. Create a new DPUNode CRD type that mirrors the format of the Node CRD.
+
+For simplicity, it seems like going with option 1 should be the first choice, unless we hit a complication that forces us
+to use option 2.
+
+#### Creating extra node CRs for DPU instances
+
+The duplicate node CR will be created by OVNK-Full DPU instance that will realize on start up it is running with a different
+ovnk-identity. Let's assume the ovnk-identity is "dpu1-ovnk". OVNK-Full DPU will be started with a new configuration option
```
booxter, line 126: Some similar configuration is passed via environment variables, namely `K8S_NODE_DPU` and `K8S_NODE`. How is this identity different from these variables? Can they be reused?

docs/okeps/okep-5521-single-cluster-host-dpu.md
trozet: yeah I suppose this is a viable option

docs/okeps/okep-5521-single-cluster-host-dpu.md
trozet: I think this is used to set the node name for DPU mode to be the host node name. So in a DPU cluster, the DPU will look like another k8s node, but you want this DPU instance to serve the host, so you want to set its node name to actually be the name of the host it is serving, rather than the name of the DPU hostname. In contrast to this, you might have another OVNK instance running on the DPU that is serving the DPU itself, which would use the hostname of the DPU.

Not sure how this would interact with k8s identity yet.

docs/okeps/okep-5521-single-cluster-host-dpu.md
trozet: yeah there are a few ways we can just mark the nodes so no pods can ever be scheduled

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
+# OKEP-5521: Single Cluster Host+DPU
+
+* Issue: [#5521](https://github.com/ovn-org/ovn-kubernetes/issues/5521)
+
+## Problem Statement
+
+When we first designed DPU support in OVN-Kubernetes, we designed it in a way that all traffic leaving the host would
```
dave-tucker, line 7: Add a link to the original design here. I took a pass at including relevant information in that do in #5591

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+    D3[OVN]
+    D1 --> D2
+    D1 --> D3
+  end
+
+  HOST <--> DPU
+```
+
+When pods on the host are launched, DPU-Host figures out the SRIOV interface for the pod and then annotates device
+information to the pod, so that OVNK on the DPU can then wire the VF Representor into OVS and configure OVN. Now, let's
```
dave-tucker, line 54: When pods are launched, a VF device is allocated via SR-IOV Device Plugin and the OVN-K CNI in "dpu-host" moves the interface to the correct network namespace, annotates the Pod with the VF details, and then waits for a Pod annotation to appear confirming that the DPU networking is set up. On the OVN-K running on the DPU, it is also watching the same Pod events and will wire the corresponding VF representor device into OVS and configure OVN.

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+
+To enable this architecture, we need full OVNK running on the host itself to handle the interfaces on the host, with
+its own OVS, as well as OVNK running on the DPU (serving the host). We also still need the DPU Host component that runs
+in the host on behalf of the DPU. Both of these OVNK are part of the same K8S cluster, and they both listen to events
+from kapi server.
+
+## Goals
+
+* Allow two OVNK instances to serve a single node. One for DPU and one for the Host.
+* Allow segmenting UDNs via the host or DPU.
```
dave-tucker, line 20: To clarify, we're saying that a UDN (or CDN) should be either:
a. Entirely on the DPU
b. Entirely on the host

We're not expecting scenarios where we have offloaded and non-offloaded Pods that are part of the same UDN?

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+  subgraph DPU[DPU]
+    direction TB
+    D1["ovn-kubernetes (full mode)"]
+    D2[OVS]
+    D3[OVN]
+    D1 --> D2
+    D1 --> D3
+  end
+
+  HOST <--> DPU
```
dave-tucker, line 50: I think it would be helpful to show a Pod, the VF device, and VF representors in this diagram.
Then it's easier to compare and contrast to 👇

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+```
+
+When pods on the host are launched, DPU-Host figures out the SRIOV interface for the pod and then annotates device
+information to the pod, so that OVNK on the DPU can then wire the VF Representor into OVS and configure OVN. Now, let's
+consider a use case where we have additional interfaces on the host that we want to leverage:
+
+```mermaid
+flowchart TB
```
dave-tucker, line 58: Per above comment, also include a Pod with VF + one or more `veth` devices.
That would more clearly show a multi-homed pod that has a DPU interface + one or more non-DPU interfaces.

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+
+As a cluster admin, I want certain workloads to utilize the DPU for primary UDN/cluster default network access, while
+I want to leverage my host based offloadable NICs for secondary UDNs. This will allow me to use my DPU for networks that
+have north/south capabilities, while only using my host based NICs for east/west acceleration.
+
+## Proposed Solution
+
+### Running multiple OVNK instances on the same node
+
+The first thing that must be addressed is how to actually run two OVN-Kubernetes instances that reference the same
```
dave-tucker, line 104: ~~I'm not sure that putting these nodes into the same K8s cluster is advisable.~~

~~To me, the fact that the DPU runs it's own K8s cluster is an implementation detail and irrelevant to the offload solution we have in OVN-K8s. Essentially, ovn-k8s in "dpu" mode need only be pointed at the correct kube-apiserver and it should work.~~

~~One concern is consider the cluster scalability - if we know that we can scale comfortably to 500 nodes, but if a node with a DPU now counts as 2 nodes, our limit is halved to 250.~~

~~Another is that, the DPU isn't really a traditional "Node".~~

~~Finally, I'm not sure if we want to the DPU and the main cluster sharing the same control plane.~~

EDIT: I got confused here. I forgot this proposal was about 2 x OVN-K8s on the host... and they would absolutely be in the same K8s cluster 🤦

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+## Goals
+
+* Allow two OVNK instances to serve a single node. One for DPU and one for the Host.
+* Allow segmenting UDNs via the host or DPU.
+* Avoid each instance doing unnecessary work if the event only affects the other OVNK instance.
+
+## Non-Goals
+
+* Addressing shared OVN, launching pods on the DPU itself, SFC, or any other part of the original 2 cluster DPU design.
+* How to run OVNK on the DPU without kubelet on the DPU.
```
dave-tucker, line 26: While this might be a non-goal it might influence the design if this is a desired outcome in future.

docs/okeps/okep-5521-single-cluster-host-dpu.md
```diff
@@ -0,0 +1,308 @@
...
+
+
+## User-Stories/Use-Cases
+
+Story 1: Segment UDNs via different NICs
+
+As a cluster admin, I want certain workloads to utilize the DPU for primary UDN/cluster default network access, while
+I want to leverage my host based offloadable NICs for secondary UDNs. This will allow me to use my DPU for networks that
+have north/south capabilities, while only using my host based NICs for east/west acceleration.
+
```
dave-tucker, line 99: Should we include a story here for Multiple DPUs?

Given we're limited to number of VFs (I believe 127 per port) I assume either multiple DPUs OR using host-based NICs is the only way to achieve higher pod density.

dave-tucker (commented):
Left some comments, but generally I think I'm 👎 on this solution.
~~I agree with the goals but I do firmly believe that a DPU shouldn't be part of the main K8s cluster... and as an option, it might not be running K8s at all.~~

I think it would be interesting to consider whether it's possible to implement this by combining full mode and dpu-host mode into a single process.

PR #2829, opened 2022-02-23
Delete stale conntrack entries for services

tssurya:
**- What this PR does and why is it needed**

Currently when we delete a clusterIP service we don't cleanup the conntrack entries on the node. This PR adds the logic to do this + we add the tcp protocol to the conntrack filtering process (not sure if we had any reasons for not including TCP?)
    
This PR does two things:
1) cleanup conntrack entries irrespective of protocol stack for service deletion
2) cleanup conntrack entries only for UDP during endpoint deletion. See https://github.com/kubernetes/kubernetes/issues/108523#issuecomment-1074044415 for details.

**- Special notes for reviewers**
Debug logs capturing conntrack deletes (done for testing purposes):
```
I0223 20:48:13.136499  431210 gateway_shared_intf.go:575] SURYA 172.30.151.15, 30102, SCTP
I0223 20:48:13.136519  431210 helper_linux.go:154] SURYA 172.30.151.15, 30102, SCTP
I0223 20:48:13.136531  431210 net_linux.go:411] SURYA 172.30.151.15, 30102, SCTP, &{map[4:0xc000abca20] map[6:30102] 132}
I0223 20:48:13.136545  431210 conntrack_linux.go:93] SURYA &{map[4:0xc000abca20] map[6:30102] 132}
I0223 20:48:13.141386  431210 conntrack_linux.go:102] SURYA 	132 src=169.254.169.2 dst=172.30.151.15 sport=47541 dport=30102 packets=0 bytes=0	src=10.131.0.27 dst=169.254.169.2 sport=30102 dport=47541 packets=0 bytes=0 mark=0
I0223 20:48:13.141397  431210 conntrack_linux.go:454] SURYA 	132 src=169.254.169.2 dst=172.30.151.15 sport=47541 dport=30102 packets=0 bytes=0	src=10.131.0.27 dst=169.254.169.2 sport=30102 dport=47541 packets=0 bytes=0 mark=0
I0223 20:48:13.141405  431210 conntrack_linux.go:472] SURYA false


I0223 20:48:13.142452  431210 conntrack_linux.go:102] SURYA 	132 src=172.31.249.96 dst=172.30.151.15 sport=47541 dport=30102 packets=0 bytes=0	src=172.30.151.15 dst=169.254.169.2 sport=30102 dport=47541 packets=0 bytes=0 mark=0
[...]

PR #5526, opened 2025-08-27
Clear stale conntrack UDP entries for nodePort services

pliurh:
When an EndpointSlice for a UDP NodePort service is updated, stale conntrack entries for removed endpoints must be flushed. The existing logic failed to do this correctly if the backend pod was on a different node than the NodePort service. This patch fixes the issue by flushing conntrack entries by filtering the nodePort when the node is not hosting the backend pod.

In case that the backend pod was on the same node as the NodePort service, this issue won't happen. Since all old pod entries are removed from the node by the function deletePodConntrack when the pod is deleted.

## 📑 Description

It fixs the flaky test "[sig-network] Conntrack should be able to preserve UDP traffic when server pod cycles for a NodePort service [Suite:openshift/conformance/parallel] [Suite:k8s]"

## ✅ Checks

- [ ] My code requires tests


go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1450,6 +1451,32 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
 					netlink.ConntrackReplyAnyIP, nil); err != nil {
 					klog.Errorf("Failed to delete conntrack entry for %s: %v", oldIPStr, err)
 				}
+
+				// Delete conntrack entries for NodePort when the node is on a remote node, using nodePort for --orig-port-src match
+				// For example: udp      17 116 src=172.18.0.2 dst=172.18.0.4 sport=35254 dport=31832 src=10.244.0.8 dst=172.18.0.2 sport=8080 dport=35254 [ASSURED] mark=10 secctx=system_u:object_r:unlabeled_t:s0 zone=16 use=1
+				// TODO: Once vishvananda/netlink support ConntrackFilterType '--reply-port-src', we can use oldPort.Port directly.
+				if svc.Spec.Type == corev1.ServiceTypeNodePort {
```
tssurya: was this an issue that always existed?
I remember fixing this during the olden days for UDN services https://github.com/ovn-kubernetes/ovn-kubernetes/pull/2829/files
was it that we didn't do it for nodeports? OR did we regress somewhere?

tssurya (changes requested):
We need unit and e2e tests for this change.
Is the k8s test not running in the conformance lanes? If so is it flaking u/s as well?

go-controller/pkg/node/default_node_network_controller.go
pliurh: > was this an issue that always existed?

I think so. I don't think this is a regression. The existing code doesn't consider the nodePort service case.

pliurh:
> We need unit and e2e tests for this change. Is the k8s test not running in the conformance lanes? If so is it flaking u/s as well?

We run "[sig-network] Conntrack should be able to preserve UDP traffic when server pod cycles for a NodePort service [Suite:openshift/conformance/parallel] [Suite:k8s]" in the conformance lanes, but with 5 retries. I think it's also flaking in u/s CI. But I don't know how to check it in the CI.

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -1743,4 +1744,117 @@ add element inet ovn-kubernetes remote-node-ips-v6 { 2002:db8:1::4 }
...
+					Type: corev1.ServiceTypeNodePort,
+					Ports: []corev1.ServicePort{{
+						Port:     port,
+						Protocol: udpProtocol,
+						NodePort: nodePort,
+					}},
+				},
+			}),
+		)
+	})
```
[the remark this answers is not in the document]
pliurh: We don't need to care if the backend is local or not. The conntrack entries are the same in both cases. For local pod, all pod IP related conntrack entities by the function `deletePodConntrack`. That's the reason why this issue only happens if the backend is on remote nodes.

pliurh:
@tssurya the existing e2e test shall cover this change. I've added UTs. PTAL.

go-controller/pkg/node/default_node_network_controller_test.go
```diff
@@ -1743,4 +1744,117 @@ add element inet ovn-kubernetes remote-node-ips-v6 { 2002:db8:1::4 }
...
+				},
+				Ports: []discovery.EndpointPort{{Port: &port, Protocol: &udpProtocol}},
+				Endpoints: []discovery.Endpoint{{
+					Addresses: []string{newEndpointIP},
+				}},
+			}
+
+			filters := []ctFilterDesc{{oldEndpointIP, int(port), udpProtocol, netlink.ConntrackReplyAnyIP}}
+			if svc.Spec.Type == corev1.ServiceTypeNodePort {
+				filters = append(filters, ctFilterDesc{oldEndpointIP, int(nodePort), udpProtocol, netlink.ConntrackReplyAnyIP})
```
arghosh93: is there any reason for appending here instead of setting a single ctFilterDesc? Is it to simulate that the conntrack entry is on a remote node?

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1454,6 +1458,20 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
...
 					klog.Errorf("Failed to delete conntrack entry for %s: %v", oldIPStr, err)
 				}
+
+				// Delete conntrack entries for NodePort when the node is on a remote node, using nodePort for --orig-port-src match
+				// In the following example, the first entry will be matched/cleared by the previous DeleteConntrackServicePort call, but not the second one.
+				//  udp      17 88 src=172.18.0.2 dst=10.244.0.8 sport=50713 dport=8080 src=10.244.0.8 dst=100.64.0.4 sport=8080 dport=50713 [ASSURED] mark=0 secctx=system_u:object_r:unlabeled_t:s0 use=1
+				//  udp      17 88 src=172.18.0.2 dst=172.18.0.4 sport=50713 dport=31832 src=10.244.0.8 dst=172.18.0.2 sport=8080 dport=50713 [ASSURED] mark=10 secctx=system_u:object_r:unlabeled_t:s0 zone=16 use=1
+				// TODO: Once vishvananda/netlink support ConntrackFilterType '--reply-port-src', we can use oldPort.Port directly.
+				if svc.Spec.Type == corev1.ServiceTypeNodePort {
+					for _, port := range svc.Spec.Ports {
```
arghosh93: Do we care about internal or external trafficPolicy here? I guess not as in that case conntrack entries will be created on the node hosting the POD and POD deletion would take care of conntrack entry deletion.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1454,6 +1458,20 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
...
+						if err := util.DeleteConntrackServicePort(oldIPStr, port.NodePort, *oldPort.Protocol,
```
arghosh93: should not we use port.Protocol here?

go-controller/pkg/node/default_node_network_controller.go
pliurh: You're right. We don't need to care.

go-controller/pkg/node/default_node_network_controller.go
pliurh: They shall have the same value. However, I will include a condition check in the loop. Thanks for the heads-up.

go-controller/pkg/node/default_node_network_controller_test.go
pliurh: The UT covers cases for both ClusterIP and NodePort services, the second filter will only be used for the NodePort service case.

arghosh93 (dismissed):
LGTM

pliurh:
Just rebase to the latest

pliurh:
Add an e2e test in the control-plane test suite.

PR #5480, opened 2025-08-13
OCPBUGS-59657: allow default network -> localnet on the same node for any localnet subnet

ricky-rav:
Previous work (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5041) allowed default network -> localnet on the same node when host network and localnet shared the same subnet.

Let's extend this to allow the localnet to use any subnet.

The localnet being on a different network, we need an external router to connect host network and localnet:
```
       ┌────────────────────────────────────────┐
       │ external router                        │
       │                                        │
       │                                        │
       ┌──────────────┐                         │
       │ eth0.100     │                         │
       │ 172.20.0.5   │                         │
       │ eth0         │  net.ipv4.ip_forward=1  │
       │ 172.18.0.5   │                         │
       └──────────────┼─────────────────────────┘
           │
           │
           │
┌──────┌────────────┐──────────────────────────────────────────────────────────┐
│      │ eth0       │                                                          │
│      └───┬────────┘                                                          │
│          │                                                                   │
│    ┌────────────┐     ┌───────┐       ┌────────────────────────┐             │
│    │            │─────┤       ┼───────┼  pod in default network│             │
│    │  br-ex     │     │ br-int│       │  eth0                  │             │
│    │  172.18.0.4│     │       │       │  10.244.0.4            │             │
│    │            ┼─────┤       │       └────────────────────────┘             │
│    └────────────┘     └────┬──┘                                              │
│                            │                                                 │
│                      ┌─────┴────────────┐                                    │
│                      │ net1             │                                    │
│                      │ 172.20.0.4       │                                    │
│                      │                  │     nad: "vlanID": 100             │
│                      │ localnet pod     │                                    │
│                      │                  │                                    │
│                      │ ip route add 172.18.0.0/16 via 172.20.0.5 dev net1    │
│                      └──────────────────┘                                    │
└──────────────────────────────────────────────────────────────────────────────┘
```
The traffic path for a packet from the default network pod to the localnet pod is:
simple_diagram_with_arrow
And the reverse is true for the reverse path. 

The reply traffic, from the localnet pod to the default network, is challenging, because it incorrectly matches the flows at prio 100 table=1 that output reply traffic directly to OVN or the host.

We overcome that by making sure that the flow in table=0 at priority=50 that sends traffic to conntrack & table=1 is only applied to traffic whose next hop is OVN/host. To do so, we add a match against `dl_dst=$breth0`:
```
[...]

PR #5529, opened 2025-08-27
Fixes regression from 1448d5ab14337b647f3e3034ea4ffc077431979a

trozet:
The previous commit dropped matching on in_port so that localnet ports would also use table 1. This allows reply packets from a localnet pod towards the shared OVN/LOCAL IP to be sent to the correct port.

However, a regression was introduced where traffic coming from these localnet ports to any destination would be sent to table 1. Egress traffic from the localnet ports is not committed to conntrack, so by sending to table=1 via CT we were getting a miss.

This is especially bad for hardware offload where a localnet port is being used as the Geneve encap port. In this case all geneve traffic misses in CT lookup and is not offloaded.

Table 1 is intended to be for handling IP traffic destined to the shared Gateway IP/MAC that both the Host and OVN use. It is also used to handle reply traffic for Egress IP. To fix this problem, we can add dl_dst match criteria to this flow, ensuring that only traffic destined to the Host/OVN goes to table 1.

Furthermore, after fixing this problem there still exists the issue that localnet -> host/OVN egress traffic will still enter table 1 and CT miss. Potentially this can be fixed with always committing egress traffic, but it might have performance penalty, so deferring that fix to a later date.

### DIFF
--- a/go-controller/pkg/node/bridgeconfig/bridgeflows.go
+++ b/go-controller/pkg/node/bridgeconfig/bridgeflows.go
@@ -632,12 +632,12 @@ func (b *BridgeConfiguration) commonFlows(hostSubnets []*net.IPNet) ([]string, e
 		}
 
 		if ofPortPhys != "" {
-			// table 0, packets coming from external or other localnet ports. Send it through conntrack and
-			// resubmit to table 1 to know the state and mark of the connection.
+			// table 0, packets coming from external or other localnet ports and destined to OVN or LOCAL.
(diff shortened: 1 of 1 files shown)


tssurya:
@trozet let's see how to also factor in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5480 into these changes? cc @ricky-rav we want to make sure this change doesn't break the fix for 5480

trozet:
closing in favor of #5480 which now includes this fix as well as others for ipv6 and more tests

PR #5533, opened 2025-08-28
[WIP] CORENET-59657: allow default network -> localnet on the same node for any localnet subnet... with Tim's fix

ricky-rav:


trozet:
closing as real PR is updated here: #5480

PR #5540, opened 2025-09-01
[WIP] Add `VirtualPrivateNetworkConnect` API

### LINKED ISSUE #5503: Implement the API/controller for ClusterNetworkConnect
### What is this issue about?

Add the API types.go files and implementation of this - also needed for thorough API review as part of the OKEP itself. So might as well do a sample draft of how that would look like with all the correct CEL validations added.

tssurya:
## 📑 Description

This PR adds the API plumbing for `VPNC`
Depends on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5246 merging first
Will be followed by the VPNC controller implementation.

Fixes #5503

## ✅ Checks

- [ ] My code requires tests


go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// ClusterNetworkConnect enables connecting multiple User Defined Networks
+// or Cluster User Defined Networks together.
+//
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:resource:path=clusternetworkconnects,scope=Cluster,shortName=cnc,singular=clusternetworkconnect
```
npinaeva, line 32: I am not sure if we need a short name. We had a discussion around this with the OCP api team for udns, and figured that our resources are not that popular on a cluster level to have it.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,121 @@
...
+	// +required
+	NetworkSelectors types.NetworkSelectors `json:"networkSelectors"`
+
+	// connectSubnets specifies the subnets used for interconnecting the selected networks.
+	// This creates a shared subnet space that connected networks can use to communicate.
+	// Can have at most 1 CIDR for each IP family (IPv4 and IPv6).
+	//
+	// +kubebuilder:validation:Required
+	// +required
+	ConnectSubnets DualStackCIDRs `json:"connectSubnets"`
```
npinaeva: oh btw do we only allow interconnect for same-ipfamily networks? like dualstack and ipv4 is not allowed? Or what happens if I set dualstack here, but all selected networks are ipv4? that would work, but also could be a misconfiguration

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// featuresEnabled specifies which features should be enabled for the connected networks.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x == y))"
```
npinaeva: I think this needs a message

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: haha yea they always say that if we ask,
for `RouteAdvertisement` I see we did `ra` which is nice cause I am always annoyed to type out `clusteruserdefinednetwork` with no typo...(I wish we had tab completion..)

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: good question!
dualstack connect subnet for single stack udn subnets should be accepted just like single stack udns are allowed on dualstack clusters
but I would think its easiest to keep consistency and say we connect only networks of same family types?
I will need to call this out in the enhancement..

@trozet suggested to add defaults here; https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5246#discussion_r2402436018 so that might simplify things? example I'll add a v4 and v6 default value if user doesn't set it, and based on udn defaults we can fill the missing one and controller uses only what it needs to use?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: done

tssurya:
```
Summarizing 1 Failure:
  [FAIL] Gateway EgressIP update EgressIP [It] reconfigures from an old to a new IP
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/egressip/gateway_egressip_test.go:203
```
some unrelated flake.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// +kubebuilder:validation:Required
+	// +required
+	Spec ClusterNetworkConnectSpec `json:"spec"`
+
+	// +optional
+	Status ClusterNetworkConnectStatus `json:"status,omitempty"`
+}
+
+// ClusterNetworkConnectSpec defines the desired state of ClusterNetworkConnect.
+// +kubebuilder:validation:XValidation:rule="!self.networkSelectors.exists(i, i.networkSelectionType != 'ClusterUserDefinedNetworks' && i.networkSelectionType != 'PrimaryUserDefinedNetworks')",message="Only ClusterUserDefinedNetworks or PrimaryUserDefinedNetworks can be selected"
```
npinaeva, line 50: are we going to check that selected CUDNs are all primary and report a status?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+type ClusterNetworkConnectStatus struct {
+	// status is a concise indication of whether the ClusterNetworkConnect
+	// resource is applied with success.
+	// +kubebuilder:validation:Optional
+	Status StatusType `json:"status,omitempty"`
+
+	// conditions is an array of condition objects indicating details about
+	// status of ClusterNetworkConnect object.
+	// +patchMergeKey=type
+	// +patchStrategy=merge
```
npinaeva, line 174: I think these 2 are not needed https://github.com/kubernetes-sigs/gateway-api/issues/342#issuecomment-703324451

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/types"
+)
+
+// ClusterNetworkConnect enables connecting multiple User Defined Networks
+// or Cluster User Defined Networks together.
```
npinaeva: for now sounds like only UDNs can be interconnected OR CUDNs, but not the mix?
Maybe replacing "or" with "and" will be better?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// +kubebuilder:validation:XValidation:rule="size(self) != 2 || !isCIDR(self[0].cidr) || !isCIDR(self[1].cidr) || cidr(self[0].cidr).ip().family() != cidr(self[1].cidr).ip().family()", message="When 2 CIDRs are set, they must be from different IP families"
+	ConnectSubnets []ConnectSubnet `json:"connectSubnets"`
+
+	// connectivityEnabled specifies which connectivity types should be enabled for the connected networks.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=2
+	// +kubebuilder:validation:XValidation:rule="self.all(x, self.exists_one(y, x == y))",message="connectivityEnabled cannot contain duplicate values"
+	ConnectivityEnabled []ConnectivityType `json:"connectivityEnabled"`
```
npinaeva: not a strong opinion, but when I was writing the CRs myself, I realized this field sounds like a bool. Should we try `ConnectivityTypes`?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+
+// +kubebuilder:validation:XValidation:rule="isCIDR(self) && cidr(self) == cidr(self).masked()", message="CIDR must be a valid network address"
+// +kubebuilder:validation:MaxLength=43
+type CIDR string
+
+// +kubebuilder:validation:XValidation:rule="!has(self.networkPrefix) || !isCIDR(self.cidr) || self.networkPrefix > cidr(self.cidr).prefixLength()", message="NetworkPrefix must be smaller than CIDR subnet"
+// +kubebuilder:validation:XValidation:rule="!has(self.networkPrefix) || !isCIDR(self.cidr) || (cidr(self.cidr).ip().family() != 4 || self.networkPrefix < 32)", message="NetworkPrefix must < 32 for ipv4 CIDR"
+type ConnectSubnet struct {
+	// CIDR specifies ConnectSubnet, which is split into smaller subnets for every connected network.
+	// This CIDR should be containing 2*((Number of L3 networks*Max Number of Nodes)+Number of L2 networks) IPs.
```
npinaeva, line 99: should we add here, that the largest supported subnet mask is `/16` due to the tunnel key limitation.
To recap, we have 2^15 (32K) tunnel key, so we can only have the same amount of `/31` subnets => 2^16 is the largest number of ips we can support

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// Example: cidr= "192.168.0.0/16", networkPrefix= 24 means that you can connect 256 Layer3 networks OR
+	// 255 Layer3 networks and 128 Layer2 networks
```
npinaeva: should this example also say how many nodes we consider?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// CIDR also restricts the maximum number of networks that can be connected together
+	// based on what CIDR range is picked. So choosing a large enough CIDR for future use cases
+	// is important.
+	// +required
+	CIDR CIDR `json:"cidr"`
+
+	// NetworkPrefix specifies the prefix length for every connected network.
+	// This prefix length should be equal to or longer than the length of the CIDR prefix.
+	//
+	// For example, if the CIDR is 10.0.0.0/8 and the networkPrefix is 24,
```
npinaeva: Probably better use /16 as the largest network that makes sense

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// then the connect subnet for each connected layer3 network will be 10.0.0.0/24, 10.0.1.0/24, 10.0.2.0/24 etc.
+	// For each connected layer2 network we will allocate a /networkPrefix range
+	// that is then used to allocate /31 and /127 slices for each layer2 network.
+	// A good practice is to set this to a value that ensures it contains more
+	// than twice the number of maximum nodes planned to be deployed in the cluster
+	// and how frequently nodes are created and destroyed.
```
npinaeva: not sure I follow how this line relates to the rest of the sentence, considering we won't be able to explain how nodeID works anyway, should we skip this?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// is important.
+	// +required
+	CIDR CIDR `json:"cidr"`
+
+	// NetworkPrefix specifies the prefix length for every connected network.
+	// This prefix length should be equal to or longer than the length of the CIDR prefix.
+	//
+	// For example, if the CIDR is 10.0.0.0/8 and the networkPrefix is 24,
+	// then the connect subnet for each connected layer3 network will be 10.0.0.0/24, 10.0.1.0/24, 10.0.2.0/24 etc.
+	// For each connected layer2 network we will allocate a /networkPrefix range
```
npinaeva: not for each right? should we say
for layer2 networks we will allocate the next available /networkPrefix range that is then split into /31 or /127 slices for each layer2 network

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// that is then used to allocate /31 and /127 slices for each layer2 network.
+	// A good practice is to set this to a value that ensures it contains more
+	// than twice the number of maximum nodes planned to be deployed in the cluster
+	// and how frequently nodes are created and destroyed.
+	// To be safe, better to set this to a value that's 4 times the maximum nodes planned
```
npinaeva: To better link it to the previous sentence, I would say
Each node gets a /31 subnet for the layer3 networks, hence networkPrefix should contain enough IPs for 4 times the maximum nodes planned?

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,189 @@
...
+	// to be deployed in the cluster.
+	// Example - required values;
```
npinaeva: I think maybe we should skip the minimum part (not to encourage this behaviour :D) and say something like
Example - recommended values; 
<and here goes the "safe" part>

npinaeva (commented):
nice CEL, all works!

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: yea good point

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: as of now in code I ignore the non-primary ones (I'm not doing a special status but that should be possible to do as well if we think that's better feedback to user) - I think even in RA code I see same behaviour
@jcaamano do you know if RA reports an error status if user selected secondary networks instead of primary via CUDN selectors?

I would say we should unify this for all networkSelector features that are not supported on secondary networks.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: yea I don't mind changing this to `ConnectivityTypes`.. @trozet any thoughts?

the reason I had connectivityEnabled: was for expressing "what types of connectivity we have enabled".. not sure if connectivityTypes conveys the same? I think it does so I don't mind changing this
Will keep this comment open to see if Tim is ok with this since we are deviating from enhancement here. so want to make sure all approvers are OK with that.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: hmmm interesting
for most parts I actually just took this always from https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: so that's a good point. I'll add this information,
I guess self note, I should add a validation for the parsing and checking this in code since its hard to do it via the cel validations

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
npinaeva: I don't think they make anything worse, so we can also keep them

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
jcaamano: > @jcaamano do you know if RA reports an error status if user selected secondary networks instead of primary via CUDN selectors?

it does

tssurya: thanks for confirming @jcaamano !
@npinaeva then let's do the same for this as well. Will make a note of that in the status PR. BTW looks like its covered in the enhancement: https://ovn-kubernetes.io/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns/#statusconditions

For different status types and error checks I will do a last PR after our main implementation is done for user experience betterness.

Resolving this as its covered in OKEP and that way I know I won't forget to do this

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: done added 128 nodes to example, hope I got that math right!

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: done.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: yea that sounds a bit confusing, will change this to your suggestion.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: ack leaving it as is for now but yea good to know they are a bit useless.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: ack I can stop at "cluster."
what I meant was like you said too much frequent node create and delete causing nodeIDs to be sporadic haha but yea we can skip that here and add them in the docs in more detail if needed.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: done.

tssurya:
> nice CEL, all works!

thanks a lot for the reviews and testing out the validations!
I think I have addressed all your comments except waiting for tim's response on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5540#discussion_r2484795041

the last push was just rebase so check the push diff before that.

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
```diff
@@ -0,0 +1,188 @@
...
+// +kubebuilder:validation:MaxLength=43
+type CIDR string
+
+// +kubebuilder:validation:XValidation:rule="!has(self.networkPrefix) || !isCIDR(self.cidr) || self.networkPrefix > cidr(self.cidr).prefixLength()", message="NetworkPrefix must be smaller than CIDR subnet"
+// +kubebuilder:validation:XValidation:rule="!has(self.networkPrefix) || !isCIDR(self.cidr) || (cidr(self.cidr).ip().family() != 4 || self.networkPrefix < 32)", message="NetworkPrefix must < 32 for ipv4 CIDR"
+type ConnectSubnet struct {
+	// CIDR specifies ConnectSubnet, which is split into smaller subnets for every connected network.
+	// This CIDR should be containing 2*((Number of L3 networks*Max Number of Nodes)+Number of L2 networks) IPs.
+	// Example: cidr= "192.168.0.0/16", networkPrefix= 24 means that you can connect 256 Layer3 networks
+	// if the cluster has 128 nodes and 0 layer2 networks OR 255 Layer3 networks and 128 Layer2 networks
```
npinaeva: is this sentence a bit broken? should it be:
`if the cluster has 128 nodes you can connect 256 Layer3 networks and 0 layer2 networks OR 255 Layer3 networks and 128 Layer2 networks?`

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
trozet: What about just "Connectivity" or "NetworkConnectivity"?

npinaeva: "Connectivity" probably works better, because Network is already in the CRD name?

tssurya: going with Connectivity in that case!

go-controller/pkg/crd/clusternetworkconnect/v1/types.go
tssurya: hmmm maybe I feel like I changed it during one of the review iterations.. so let me double check

tssurya: I've fixed it

tssurya:
```
Summarizing 1 Failure:
  [FAIL] Node Operations on add and delete [It] manages iptables rules with ExternalIP through retry logic
```
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/19432691212/job/55595396407?pr=5540

not sure if something recently was merged that caused this? seems unrelated to this PR

PR #5543, opened 2025-09-01
fix: Skip node-encap-ips annotation in DPU host mode

aserdean:
The node-encap-ips annotation should not be set when running in DPU host mode since the encap IP is not available on the DPU host. This prevents errors during node initialization when the encap IP cannot be determined.

- Add conditional check for NodeModeDPUHost before setting encap IP annotation
- Maintain existing behavior for non-DPU host modes
- Fixes initialization errors in DPU host mode deployments

## Summary by CodeRabbit

* **Bug Fixes**
  * Nodes in DPU Host mode no longer publish encap IP annotations, preventing incorrect metadata and networking confusion.
  * Non‑DPU‑Host nodes continue to publish encap IP annotations as before, preserving accurate cluster networking.
  * Reduces misconfiguration risk in mixed deployments and simplifies troubleshooting by clarifying annotation behavior.

### DIFF
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -1008,10 +1008,14 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
 		return fmt.Errorf("failed to set node zone annotation for node %s: %w", nc.name, err)
 	}
 
-	encapIPList := sets.New[string]()
-	encapIPList.Insert(strings.Split(config.Default.EffectiveEncapIP, ",")...)
-	if err := util.SetNodeEncapIPs(nodeAnnotator, encapIPList); err != nil {
-		return fmt.Errorf("failed to set node-encap-ips annotation for node %s: %w", nc.name, err)
+	// Set the node-encap-ips annotation with the configured encap IP.
+	// This encap IP is unavailable on the DPU host mode, so we don't need to set it there.
+	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
+		encapIPList := sets.New[string]()
+		encapIPList.Insert(strings.Split(config.Default.EffectiveEncapIP, ",")...)
+		if err := util.SetNodeEncapIPs(nodeAnnotator, encapIPList); err != nil {
+			return fmt.Errorf("failed to set node-encap-ips annotation for node %s: %w", nc.name, err)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1008,10 +1008,14 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
 		return fmt.Errorf("failed to set node zone annotation for node %s: %w", nc.name, err)
 	}
 
-	encapIPList := sets.New[string]()
-	encapIPList.Insert(strings.Split(config.Default.EffectiveEncapIP, ",")...)
-	if err := util.SetNodeEncapIPs(nodeAnnotator, encapIPList); err != nil {
-		return fmt.Errorf("failed to set node-encap-ips annotation for node %s: %w", nc.name, err)
+	// Set the node-encap-ips annotation with the configured encap IP.
+	// This encap IP is unavailable on the DPU host mode, so we need to set it there.
```
arghosh93: nit:
```suggestion
	// This encap IP is unavailable on the DPU host mode, so we don't need to set it there.
```

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1008,10 +1008,14 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
...
+	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
```
arghosh93, line 1013: Dont you need this check also at node IP addressManager?
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/node_ip_handler_linux.go#L246

aserdean: fair point, I forgot that primary addr can change, thanks for pointing that out, will change in an incremental

aserdean:
@girishmg @npinaeva @trozet 
can you PTAL

trozet (approved):
@aserdean this is fine for now, but keep in mind there is going to be some changes around potentially launching DPU Host and another ovnk instance on the same node:

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5523

just raising for you awareness

aserdean:
@trozet : thank you for letting me know, will go through the OKEP as well to make sure we iron out the details since we also are chasing running the reverse an ovnk full mode instace alongside ovnk dpu mode instance.

PR #5485, opened 2025-08-15
enhancement: Improve layer2 topology

npinaeva:
## 📑 Description

We want to use transit router in the layer2 topology.
This takes over from https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5097

Tracking issue: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5545

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## Summary by CodeRabbit

* **Documentation**
  * Added an Enhancement Proposal for a Layer2 transit‑router topology: motivations, architecture, IPv4/IPv6 addressing and per‑node/shared transit subnets, diagrams, API annotations, rollout and rolling‑upgrade plan, NAT/routing relocation, direct router‑to‑router peering, testing/operational guidance, risk mitigations, and a new site navigation entry for the proposal.

PR #5550, opened 2025-09-02
OKEP-5535: Support Multi-NIC and Multi-VTEP in OVN-Kubernetes IC Mode

### LINKED ISSUE #5535: Support multi-NIC and multi-VTEP in IC mode

This is for adding multi-VTEP support to IC mode.

# Node annotation changes

When a node has multiple VTEPs, its encap IPs are listed in the `k8s.ovn.org/node-encap-ips annotation`. To support multi-VTEP in IC mode,  need to introduces additional node annotations:

* k8s.ovn.org/node-transit-switch-port-tunnel-ids:\
  Stores tunnel IDs for all Transit Switch ports. Its length and order must match k8s.ovn.org/node-encap-ips.
* k8s.ovn.org/node-transit-switch-port-ifaddrs:\
  Stores the IP addresses of the Transit Switch ports, ordered to match k8s.ovn.org/node-encap-ips.

Example annotations:
```
k8s.ovn.org/node-encap-ips: '["10.2.0.41","10.2.0.43","10.2.0.45"]'
k8s.ovn.org/node-transit-switch-port-tunnel-ids: '[3,4,5]'
k8s.ovn.org/node-transit-switch-port-ifaddrs: '[{"ipv4":"100.88.0.3/16"},{"ipv4":"100.88.0.4/16"},{"ipv4":"100.88.0.5/16"}]'
```

# Pod annotation changes

Added `encap_ip` to Pod annotation `k8s.ovn.org/pod-networks`.
Example:
```
k8s.ovn.org/pod-networks: '{
  "default": {
    "ip_addresses": [ "10.192.18.35/26" ], 
    "mac_address": "0a:58:0a:c0:12:23",
    "gateway_ips": [ "10.192.18.1" ], 
    "encap_ip": "10.2.0.151", 
    ...
  },
```

The `encap_ip` is used as below:
* Layer 2: Sets the remote LSP’s encap IP based on encap_ip.
* Layer 3: 
  * if remote node has multiple VTEPs, its remote LSPs will be created on-demand
    * when the 1st Pod on remote node is created, the remote LSP corresponds to the Pod's encap_ip will be created, and a route with node subnet to the LSP ifaddr will be added; if 2nd Pod on remote node has same encap_ip, it uses the same remote LSP and route created for the 1st Pod, otherwise, new remote LSP and a dest /32 route will be created
    * if remote node only has one VTEP, it will be same as today.  

More details will be covered by OKEP

### Why is this needed?

In IC mode, OVN-Kubernetes does not have a centralized Southbound DB, so nodes are unaware of remote VIFs’ encap-ip. As a result, for both L2 and L3 networks:

* Traffic may be encapsulated using the wrong VTEP IP.
* Hardware offloading via SR-IOV can break if traffic is sent through the wrong PF.

In L3 networks, a Transit Switch is used for encapsulation, but each node is currently represented by a single LSP, and the encap field is unset—this limits encapsulation to only one VTEP per node.

In L2 networks, the lack of remote encap-ip awareness makes it impossible to correctly encapsulate traffic for remote VIFs.

By enabling ovnkube-controller to set the encap field on remote LSPs' Port_binding, this enhancement would allow IC mode to fully support multi-VTEP deployments, enable correct SR-IOV offloading, and align IC mode behavior more closely with Central mode.

l8huang:
Fixes #5535

## Summary by CodeRabbit

* **Documentation**
  * Added an enhancement proposal detailing support for multi‑NIC and multi‑VTEP in IC mode: goals, Layer‑2/Layer‑3 behaviors, per‑pod and per‑node encapsulation annotations, transit port naming and on‑demand port/route behavior, and unit/E2E testing guidance.
  * Updated site navigation to include the new enhancement proposal.

### DIFF
--- a/docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
+++ b/docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
@@ -1,0 +1,542 @@
+# OKEP-5535: Support Multi-NIC and Multi-VTEP in OVN-Kubernetes IC Mode
+
+* Issue: [#5535](https://github.com/ovn-org/ovn-kubernetes/issues/5535)
+
+## Problem Statement
+
+OVN-Kubernetes Interconnect (IC) mode currently cannot reliably support multi-VTEP setups.
+Nodes with multiple encapsulation endpoints cannot consistently select the correct VTEP per pod/network,
+so traffic may use the wrong tunnel endpoint. This breaks hardware-offload-oriented setups and can
+cause incorrect routing behavior.
+
+## Goals
+
+* Primarily support multiple VTEPs in IC mode for nodes with multiple SR-IOV NICs
+* Ensure correct overlay traffic encapsulation per pod interface
+* Introduce VTEP CRD (`Unmanaged` mode) + NAD `k8s.ovn.org/vtep` as the preferred way to select VTEP per network and
+  propagate selected `encap_ip` into pod network annotation
+* Keep existing SR-IOV PF-mapping based behavior as fallback for compatibility during migration
+* Support veth-based setups as well, primarily for E2E validation
+
+## Non-Goals
+
+* Add non-GENEVE tunnel type support as part of this OKEP
+* Immediately remove legacy SR-IOV PF-mapping based behavior in this phase (it will be deprecated in the future)
+
+## Introduction
+
+This OKEP primarily targets multi-SR-IOV-NIC deployments in IC mode, where a node has multiple
+encapsulation endpoints and traffic must be steered to the correct VTEP per pod/network.
+At the same time, the new control-plane steering model is generic enough to support veth-based
+setups (mainly for E2E validation).
+
+The preferred mechanism is:
+- define unmanaged VTEPs (named groups of encap IP ranges),
+- select a VTEP per network via NAD annotation (`k8s.ovn.org/vtep`),
+- resolve and publish `encap_ip` in pod annotation at allocation time.
+
+For nodes that have only one SR-IOV NIC (or otherwise only one VTEP interface), explicit VTEP
+selection is not required. Those nodes do not need unmanaged VTEP objects, and NADs used there do not
+need `k8s.ovn.org/vtep`; pod annotation can use the node's single encap IP directly.
+This is expected to be the common deployment model for clusters that are not using multi-VTEP.
+
+During migration, the existing SR-IOV PF-mapping behavior remains as fallback.
+In that legacy model, each SR-IOV PF can serve as a separate VTEP interface, and the PF-to-encap-IP
+mapping is configured via `external_ids:ovn-pf-encap-ip-mapping`.
+
+For example, assume a node has 3 SR-IOV NICs with PFs `enp1s0f0`, `enp2s0f0`, and `enp3s0f0`.
+Each PF has an associated bridge interface (`brenp*`) with an IP address used as encap IP:
+
+```bash
+$ ip -br addr
+brenp1s0f0           UNKNOWN        10.0.0.2/16
+brenp2s0f0           UNKNOWN        10.0.0.3/16
+brenp3s0f0           UNKNOWN        10.0.0.4/16
+
+$ ovs-vsctl get open . external_ids:ovn-encap-ip
+"10.0.0.2,10.0.0.3,10.0.0.4"
+
+$ ovs-vsctl get open . external_ids:ovn-pf-encap-ip-mapping
+"enp1s0f0:10.0.0.2,enp2s0f0:10.0.0.3,enp3s0f0:10.0.0.4"
+```
+
+In **Central mode**, when a pod's VIF is backed by a VF, the VIF's encap IP can be derived from
+`external_ids:ovn-pf-encap-ip-mapping`. `ovnkube-node` sets this value on the OVS interface's
+`external_ids:encap-ip`, and `ovn-controller` updates the corresponding `Port_Binding.encap` field
+in the centralized Southbound database. This allows each VIF's encap IP to be distributed to all
+nodes in the cluster. For details, see the
+[Multi-VTEP Documentation](https://ovn-kubernetes.io/features/multiple-networks/multi-vtep).
+
+
+In the current **IC mode** implementation, multi-VTEP steering does not work correctly:
+
+- **Layer 3 topology**: Each Transit Switch has only a **single remote LSP per remote node**,
+  representing just one of the node's VTEPs. A static route directs all traffic destined for the
+  remote node's subnet to this single LSP. As a result, **all traffic goes through one VTEP**,
+  regardless of which VTEP should have been used.
+
+- **Layer 2 topology**: Each remote VIF has its own remote LSP, but because IC mode doesn't have a
+  centralized Southbound database, nodes cannot see the `encap_ip` values of remote VIFs.
+
+Multi-VTEP support is critical for clusters with multiple SR-IOV NICs. VFs are tied to specific PFs,
+and hardware offloading requires packets to be routed through the correct PF (i.e., the correct VTEP).
+This proposal standardizes `encap_ip` publication for IC mode and adds the OVN logical network
+elements needed for per-network VTEP steering.
+
+## User-Stories/Use-Cases
+
+* Story 1: Hardware Offload with Multi-SR-IOV-NIC and Multi-VTEP in IC Mode
+
+  As a network administrator, I want each pod’s traffic to be encapsulated using the VTEP IP of the
+  PF that owns its associated VF, so that hardware offload works reliably and efficiently in
+  multi-NIC environments.
+
+* Story 2: Minimize LSP and Route Bloat in Large-Scale Multi-VTEP Deployments
+
+  As a network administrator, I want the system to create remote LSPs and
+  static routes on demand, instead of pre-provisioning them for every possible VTEP, so that the
+  control plane remains scalable and efficient in large multi-VTEP clusters.
+
+* Story 3: Per-Network VTEP Steering via NAD
+
+  As a network administrator, I want to select a VTEP per NAD using
+  `k8s.ovn.org/vtep`, so that pods attached to different networks on the same
+  node can use different encapsulation endpoints through a consistent
+  control-plane mechanism.
+
+
+## Proposed Solution
+
+### API Details
+
+No new CRD schema is introduced by this OKEP.
+This design reuses the existing VTEP CR in `Unmanaged` mode (introduced by OKEP-5088: EVPN Support)
+as an optional input to map node encap IPs to named VTEPs.
+
+The API-level behavior added by this OKEP is annotation-based:
+- `k8s.ovn.org/vteps` on Node (per-node discovered VTEPs and allocated tunnel IDs)
+- `k8s.ovn.org/vtep` on NAD (per-network VTEP selection)
+
+If unmanaged VTEPs or NAD VTEP selection are not used, the current SR-IOV PF-based behavior remains
+the fallback in this phase.
+
+### Implementation Details
+
+In the current OVN-Kubernetes implementation, each node exposes its encap IPs via the
+`k8s.ovn.org/node-encap-ips` annotation and its L3 network subnets via `k8s.ovn.org/node-subnets`.
+For example, `node-a` might have:
+
+```yaml
+k8s.ovn.org/node-id: "6"
+k8s.ovn.org/node-encap-ips: '["10.0.0.2","10.0.0.3","10.0.0.4"]'
+k8s.ovn.org/node-subnets: '{
+  "default":["10.1.7.0/24"],
+  "net-red":["10.2.13.0/24"],
+  "net-blue":["10.3.17.0/24"]}'
+```
+
+The PF-to-encap-IP mapping in OVS DB is:
+```bash
+$ ovs-vsctl get open . external_ids:ovn-pf-encap-ip-mapping
+"enp1s0f0:10.0.0.2,enp2s0f0:10.0.0.3,enp3s0f0:10.0.0.4"
+```
+
+Optionally, administrators can define unmanaged VTEP objects that group node encap IP ranges by VTEP name.
+For example, for `node-a` with `k8s.ovn.org/node-encap-ips: '["10.0.0.2","10.0.0.3","10.0.0.4"]'`:
+
+```yaml
+apiVersion: k8s.ovn.org/v1
+kind: VTEP
+metadata:
+  name: vtep0
+spec:
+  mode: Unmanaged
+  cidrs:
+    - "10.0.0.2/32"
+---
+apiVersion: k8s.ovn.org/v1
+kind: VTEP
+metadata:
+  name: vtep1
+spec:
+  mode: Unmanaged
+  cidrs:
+    - "10.0.0.3/32"
+---
+apiVersion: k8s.ovn.org/v1
+kind: VTEP
+metadata:
+  name: vtep2
+spec:
+  mode: Unmanaged
+  cidrs:
+    - "10.0.0.4/32"
+```
+
+After these VTEP objects are defined and processed, `node-a` would carry:
+
+```yaml
+k8s.ovn.org/node-id: "6"
+k8s.ovn.org/node-encap-ips: '["10.0.0.2","10.0.0.3","10.0.0.4"]'
+k8s.ovn.org/vteps: '{"vtep0":{"ips":["10.0.0.2"],"tunnel-id":6},"vtep1":{"ips":["10.0.0.3"],"tunnel-id":7},"vtep2":{"ips":["10.0.0.4"],"tunnel-id":8}}'
+```
+
+Networks defined via NAD are mapped to PFs as below:
+
+| Network  | Cluster Subnet | PF       | Encap IP |
+| -------- | -------------- | -------- | -------- |
+| default  | 10.1.0.0/16/24 | enp1s0f0 | 10.0.0.2 |
+| net-red  | 10.2.0.0/16/24 | enp2s0f0 | 10.0.0.3 |
+| net-blue | 10.3.0.0/16/24 | enp3s0f0 | 10.0.0.4 |
+
+In Central mode, a pod VIF's encap IP is propagated across the cluster via the centralized
+Southbound database. In IC mode, however, there is no centralized database, so the VIF's encap IP
+is not visible to remote nodes. This prevents correct encapsulation in multi-VTEP setups.
+
+To address this, each pod must expose its interface's encap IP in its `k8s.ovn.org/pod-networks`
+annotation.
+
+For NAD-attached networks, users can optionally set `k8s.ovn.org/vtep: <vtep-name>` on the NAD.
+When present, `encap_ip` is resolved from the scheduled node's `k8s.ovn.org/vteps` entry for that VTEP
+and written into pod annotation at allocation time:
+- for Layer 2 IC networks, this is done by ovnkube-cluster-manager during pod IP allocation;
+- for Layer 3 networks, this is done by ovnkube-controller during pod annotation allocation
+  (with CNI as fallback if `encap_ip` is still missing).
+
+If the NAD does not have a `k8s.ovn.org/vtep` annotation, the default behavior is:
+- for single-VTEP nodes, `ovnkube-controller` / `ovnkube-cluster-manager` can use the node's only
+  encap IP as `encap_ip` in pod annotation allocation;
+- for multi-VTEP nodes, if NAD does not specify `k8s.ovn.org/vtep`, `encap_ip` may remain unset in
+  pod annotation and CNI can derive it from PF mapping
+  (`external_ids:ovn-pf-encap-ip-mapping`) when applicable.
+
+For example, assume the NADs are annotated to select specific VTEPs:
+
+```yaml
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  name: net-red
+  namespace: default
+  annotations:
+    k8s.ovn.org/vtep: vtep1
+---
+apiVersion: k8s.cni.cncf.io/v1
+kind: NetworkAttachmentDefinition
+metadata:
+  name: net-blue
+  namespace: default
+  annotations:
+    k8s.ovn.org/vtep: vtep2
+```
+
+With those NAD annotations, Pod-a on node-a would have:
+
+```yaml
+k8s.ovn.org/pod-networks: '{
+  "default": {
+    "ip_addresses": ["10.1.7.35/24"],
+    "encap_ip": "10.0.0.2",
+    ...
+  },
+  "default/net-red": {
+    "ip_addresses": ["10.2.13.4/24"],
+    "encap_ip": "10.0.0.3",
+    ...
+  },
+  "default/net-blue": {
+    "ip_addresses": ["10.3.17.5/24"],
+    "encap_ip": "10.0.0.4",
+    ...
+  }
+}'
+```
+
+#### Layer 2 topology solution
+
+For Layer 2 networks, each remote pod's VIF is represented by a remote LSP in the `layer2_ovn_layer2_switch`.
+To enable correct encapsulation, the local `ovnkube-controller` must set the
+`Port_Binding.encap` field for each remote LSP based on the pod's `encap_ip` annotation.
+
+For example, assume `net-blue` is a Layer 2 network and Pod-a is scheduled on node-a. On node-b,
+Pod-a appears as a remote pod with the following remote LSP:
+
+```yaml
+ switch f99af2b3-8586-460b-940d-98c8adc3b2d2 (ovn.blue.layer2_ovn_layer2_switch)
+    port default.ovn.blue_test_pod-a
+        type: remote
+        addresses: ["0a:58:0a:03:11:05 10.3.17.5"]
+```
+
+When `ovnkube-controller` on node-b processes this remote pod, it reads the `encap_ip` (`10.0.0.4`)
+from Pod-a's `k8s.ovn.org/pod-networks` annotation and updates the Port_Binding's `encap` field
+to point to the matching Encap entry for node-a:
+
+```bash
+$ ovn-sbctl show
+Chassis node-a
+    hostname: node-a
+    Encap geneve
+        ip: "10.0.0.2"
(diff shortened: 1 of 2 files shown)


docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,353 @@
...
+
+In clusters where nodes are equipped with multiple SR-IOV NICs and VTEP interfaces, the OVS
+database uses `external_ids:ovn-pf-encap-ip-mapping` to map each SR-IOV PF to its corresponding
+encap IP.
+
+In **Central mode**, when a pod is created and its VIF is backed by a VF, the pod’s encap IP is
+derived from `external_ids:ovn-pf-encap-ip-mapping`. `ovnkube-node` sets this value on the OVS
+interface's `external_ids:ovn-encap-ip`, and `ovn-controller` then updates the corresponding
+`Port_Binding.encap` field in the centralized Southbound database, allowing the VIF’s encap IP
+to be distributed to all other nodes in the cluster.
```
trozet: does this work for destination encap IP selection? So lets say pod A on node 1 has external_ids:ovn-encap-ip=10.0.0.5, and pod B on node 2 has external_ids:ovn-encap-ip=10.0.0.6. 

When pod A goes to send to pod B, ovn-controller it will choose source encap ip of 10.0.0.5, then ip rules will send it out the right vtep interface, but does it also choose 10.0.0.6 for the destination encap IP based on the external ID on the pod B logical port?

That part is vague in:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/features/multiple-networks/multi-vtep.md

Where it simply says "This encap IP will be utilized as the source IP in the tunnel header." and does not seem to talk about destination encap IP.

You should also reference https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/features/multiple-networks/multi-vtep.md in your introduction section to give readers more background.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,353 @@
...
+```
+ switch f99af2b3-8586-460b-940d-98c8adc3b2d2 (ovn.blue.layer2_ovn_layer2_switch)
+    port default.ovn.blue.layer2.test_pod-a
+        type: remote
+        addresses: ["0a:58:0a:03:11:05 10.3.17.5"]
+```
+
+On peer nodes of node-a, the `ovnkube-controller` should populate the corresponding Port_Binding's
+`encap` field for this remote port using the `encap_ip` (`10.0.0.3`) from the pod's
+`k8s.ovn.org/pod-networks` annotation.
```
trozet: can you show what the port will look like with your change (show the encap ip set)

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,353 @@
...
+          type: remote
+          addresses: ["0a:58:64:58:00:08 100.88.0.8/16"]
+
+  ```
+  And it adds a static route with 10.1.7.5/32 to `ovn_cluster_router`:
+  ```
+  IPv4 Routes
+  Route Table <main>:
+        10.1.7.5/32         100.88.0.8 dst-ip
+        10.1.7.0/24         100.88.0.7 dst-ip
```
trozet: nice explanation in this section

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,353 @@
...
+
+For example, node-a should have following annotations:
+```
+k8s.ovn.org/node-id: "6"
+k8s.ovn.org/node-encap-ips: '["10.0.0.1","10.0.0.2","10.0.0.3"]'
+k8s.ovn.org/node-transit-switch-port-tunnel-ids: [6, 7, 8]
+k8s.ovn.org/node-transit-switch-port-ifaddrs: '[
+    {"ipv4":"100.88.0.6/16"},
+    {"ipv4":"100.88.0.7/16"},
+    {"ipv4":"100.88.0.8/16"}]'
```
trozet: who is going to allocate node-transit-switch-port-tunnel-ids? is it done by the node id allocator to avoid conflicts with other node ids?

who configures k8s.ovn.org/node-transit-switch-port-ifaddrs? How does it know it needs 3? based on length of node-encap-ips?

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: yes, this section is about how the node 1 gets the pod B's encap IP and tunnel ID(Port_Binding) from the centralized OVN Southbound DB.

When pod A sends packets to pod B, the source encap ip is chosen when the packet hits the in_port on br-int, the destination encap ip is decided when  OpenFlow `output` port is chosen. In the OpenFlow programmed by ovn-controller, the source and destination encap IP are chosen indirectly, the implementation details behind OVN DB schema are not covered here.

A reference to https://ovn-kubernetes.io/features/multiple-networks/multi-vtep is added, it will be updated after this feature is implemented.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+Networks defined via NAD are mapped to PFs as:
+
+| Network  | Cluster Subnet | PF       | Encap IP |
+| -------- | -------------- | -------- | -------- |
+| default  | 10.1.0.0/16/24 | enp1s0f0 | 10.0.0.1 |
+| net-red  | 10.2.0.0/16/24 | enp2s0f0 | 10.0.0.2 |
+| net-blue | 10.3.0.0/16/24 | enp3s0f0 | 10.0.0.3 |
+
```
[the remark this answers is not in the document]
l8huang: this is expected.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: done

l8huang (commented):
Thanks for comments, updated according.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: updated in doc

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,387 @@
...
+- `k8s.ovn.org/node-transit-switch-port-ifaddr` as the port's address
+
+This setup supports only a single VTEP per node.
+
+To support multiple Transit Switch ports per node -- each associated with a specific encap IP, two new annotations
+must be introduced:
+- `k8s.ovn.org/node-transit-switch-port-tunnel-ids`:  
+  Stores the tunnel IDs allocated for each Transit Switch port.
+- `k8s.ovn.org/node-transit-switch-port-ifaddrs`:  
+  Stores the IP addresses assigned to each Transit Switch port.
```
trozet: we need to figure out if this one is really needed, and how it interacts with:

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5485#issuecomment-3276485851

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,387 @@
...
+
+In the current implementation, a remote Transit Switch port representing a node uses:
+- `k8s.ovn.org/node-id` as the tunnel key
+- `k8s.ovn.org/node-transit-switch-port-ifaddr` as the port's address
+
+This setup supports only a single VTEP per node.
+
+To support multiple Transit Switch ports per node -- each associated with a specific encap IP, two new annotations
+must be introduced:
+- `k8s.ovn.org/node-transit-switch-port-tunnel-ids`:
```
trozet: @npinaeva fyi

trozet: so @l8huang these tunnel keys are for router remote ports with layer 3 networks right? what allocator are you going to use? For layer 2 ports we have:

layer 2 ports -> ncc.tunnelIDAllocator = id.NewIDAllocator(ncc.GetNetworkName(), 32767)

l8huang: yes, these tunnel keys are for layer 3 networks.

I plan to use the same ID allocator (zoneClusterController.nodeIDAllocator) used for k8s.ovn.org/node-id to allocate the additional tunnel keys. This may result in node IDs no longer being consecutive. If that becomes a concern, I can introduce a separate ID allocator specifically for the additional tunnel keys. Let me know which option you prefer.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: Thanks for heads up, I will review https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5485 to check for any potential conflicts.

l8huang (commented):
Thanks for your comments.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
trozet: I don't really care about sequential allocation, but I think you will be restricted to using 32k max keys. That means if you have 8 rails per node, your maximum cluster size is limited to 4k. Are we OK with this limitation?

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
trozet: So I think we determined the current k8s.ovn.org/node-transit-switch-port-ifaddr is not needed. If you know the transit- subnet and you know the tunnel keys, cant you automatically derive the transit-switch-port-ifaddr?

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,387 @@
...
+
+To support multiple Transit Switch ports per node -- each associated with a specific encap IP, two new annotations
+must be introduced:
+- `k8s.ovn.org/node-transit-switch-port-tunnel-ids`:  
+  Stores the tunnel IDs allocated for each Transit Switch port.
+- `k8s.ovn.org/node-transit-switch-port-ifaddrs`:  
+  Stores the IP addresses assigned to each Transit Switch port.
+
+These annotations are added by the ovnkube-cluster-manager when a node joins the cluster. The number of tunnel IDs
+and addresses must match the number of encap IPs configured on the node.
```
npinaeva: IIUC, only a subset of VTEPs is configured for every NAD. So transit IPs and tunnel keys don't need to be unique across all VTEPs, but only within the same NAD?
So if you have 8 VTEPs, and you allocate 3 of them for NAD1, then you don't need annotations for all 8 of them right?

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
npinaeva: also nodeID is its own thing used for more than just tunnel keys allocation, but I also think that ID allocation could go to the NAD annotation and not node annotations, i added a questions below to clarify that

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: sure, I will  remove `k8s.ovn.org/node-transit-switch-port-ifaddrs`.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: >  your maximum cluster size is limited to 4k.

this limitation is posted by OVN/GENEVE, right? 4k nodes one cluster should be sufficient. CC @girishmg WDYT?

> nodeID is its own thing used for more than just tunnel keys allocation, but I also think that ID allocation could go to the NAD annotation and not node annotations

nodeID is used as tunnel key in current implementation -- single vtep case; in multi-vtep case, each encap IP needs a tunnel key, this is node-level resource. The encap IP used by a NAD could be specified indirectly by `k8s.v1.cni.cncf.io/resourceName`, no need to add an tunnel key  annotation to NAD.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: From my understanding, VTEPs and tunnel keys are node-level resources. The NAD selects which one to use via the k8s.v1.cni.cncf.io/resourceName annotation — which is the main reason multi-VTEP is only supported with SR-IOV NICs.  

I expected NAD or UDN to provide a way to explicitly configure which VTEP to use, but this OKEP takes the approach of creating Transit Switch ports on demand. Even so, there’s no need to annotate the NAD with tunnel keys.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
npinaeva: 32K limitation is a max tunnel key for OVN LSP.
VTEPs are node-level resources, but tunnel keys are NAD-level resources, because they only make sense within a transit switch. Also if a given NAD on a given node only uses 3 out of 8 VTEPs, there is no point in creating 8 remote ports. Same applies to the nodes not being used by some NADs at all, which could be useful to improve the scalability, but also could be done later if needed.
If it is too complicated to figure out which VTEPs are used by which NAD, and allocating resources for all of them is a kind of a workaround, it should be mentioned in the enhancement.

npinaeva: I would also vote for using a different allocator from the node-id, because we are not allocating node-ids, but vtep-keys, so it is better to keep them independent

l8huang: > ..., but tunnel keys are NAD-level resources ... Also if a given NAD on a given node only uses 3 out of 8 VTEPs

I'm confused by this. From my understating, the current implement only handles the single-VTEP case. Tunnel keys can also be node-level resourcek, with a 1:1 mapping to VTEPs. For example, assume a node has below mapping:

```markdown
VTEP     Tunnel Key
-------  ---------
vtep0      10
vtep1      11
vtep2      12
```
If NAD-a uses vtep1 and NAD-b uses vtep2, they can directly use tunnel keys 11 and 12, without needing to allocate tunnel keys separately for NAD-a and NAD-b.  @trozet WDYT?


> Also if a given NAD on a given node only uses 3 out of 8 VTEPs, there is no point in creating 8 remote ports.
Agree, the remote port will be created on-demand as described in this OKEP. 

>  If it is too complicated to figure out which VTEPs are used by which NAD, and allocating resources for all of them is a kind of a workaround, 

VTEP selection by NAD isn’t supported, so this isn’t a workaround — it’s just the current design :) Even if VTEP selection by NAD is added later, tunnel keys can still remain a node-level resource.

npinaeva: > If NAD-a uses vtep1 and NAD-b uses vtep2, they can directly use tunnel keys 11 and 12, without needing to allocate tunnel keys separately for NAD-a and NAD-b

Correct, current solution works, but what I am saying in this case tunnel keys don't have to be different. So if vtep1 is only used by NAD1 and vtep2 is only used by NAD2, then NAD1 doesn't need to have a port for vtep2 (because no traffic will ever be sent there), same for NAD2 and vtep1.
SO I guess current design adds all VTEPs possible to every NAD, even though it can't actually be used by all NADs.
So the more optimal solution would be to only create ports for VTEPs that are used by a given NAD.
So if we did that, we could re-use tunnel keys for vtep1 and vtep2 if they are never used in the same NAD.

Current solution where we just connect everything always also works, but it is sub-optimal.
As you said explicit VTEP selection by NAD is not supported, this can only be figured out by looking as which VTEP got assigned to which pods under which NAD. So implicit NAD to VTEP allocation exists via pods.

So I would like to see this kind of explanation in the enhancement if possible

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,426 @@
...
+When a multi-VTEP node joins the cluster, `ovnkube-cluster-manager` allocates tunnel IDs and sets
+this annotation. The number of tunnel IDs matches the number of encap IPs, ensuring a one-to-one
+mapping between each encap IP and its tunnel ID.
+
+For example, node-a would have the following annotations:
+```yaml
+k8s.ovn.org/node-id: "6"
+k8s.ovn.org/node-encap-ips: '["10.0.0.2","10.0.0.3","10.0.0.4"]'
+k8s.ovn.org/node-transit-switch-port-tunnel-ids: '[6,7,8]'
+```
```
l8huang: @trozet @girishmg  could you please comment on this? just want to confirm if it's okay two maintain the relationship of  encap ip and TSP tunnel id in two separate annotations.

l8huang: changed to use new annotations:
```yaml
k8s.ovn.org/vteps: '{"vtep0":{"ips":["10.0.0.2"],"tunnel-id":6},"vtep1":{"ips":["10.0.0.3"],"tunnel-id":7},"vtep2":{"ips":["10.0.0.4"],"tunnel-id":8}}'
```

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,624 @@
...
+
+`ovnkube-controller` therefore filters by matching `resource_name` and uses the first
+`device_ids` entry -- here, `0000:04:01.0` -- to resolve `encap_ip` for **both** attachments. See
+[Risks, Known Limitations and Mitigations](#risks-known-limitations-and-mitigations) for the
+limitation this "first device" heuristic introduces.
+
+When it writes `encap_ip` into the pod's network annotation depends on the network's topology:
+
+- **Default network and Layer 3 UDNs**: `ovnkube-controller` itself allocates the pod's IP and
+  writes the pod's network annotation when handling the pod's create/update event, so it resolves
```
npinaeva, line 258: why resolving encap ip during CNI ADD is not mentioned as an alternative?
Also this needs more detail about what happens when the resources are not yer allocated on pod add and we are supposed to retry it

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,624 @@
...
+  are created, and existing single-VTEP Transit Switch port naming (`tstor-<node-name>`) is
+  unchanged.
+* This design reuses the existing VTEP CRD (in `Unmanaged` mode, from OKEP-5088) without any
+  schema changes, so no CRD migration is required.
+
+## Alternatives
+
+### Use a Dedicated Encap-IP-to-Tunnel-ID Annotation
+
+The current VTEP-based approach does not use the VTEP name. Transit Switch ports are named based
```
npinaeva, line 623: I find current VTEP integration pretty strange, is it only used for tunnel-id generation? We now store the per-node `external_ids:ovn-pf-encap-ip-mapping` but can we make it closer to the existing VTEP, so that it is discoverable from each node, and in addition to "ip" and "tunnel-id" maybe adds "pfName" to the `vteps` annotation? then each node will use vteps annotation to match pfName to its encap IP instead of ovs external-ids?
also how are the encap IPs assigned to PFs and can we make it more transparent in the API?

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
```diff
@@ -0,0 +1,624 @@
...
+node-a's default encap IP (`external_ids:ovn-encap-ip-default`), so publishing it would be
+redundant -- remote nodes already route to `10.0.0.2` by default. `net-red` and `net-blue` do get
+`encap_ip` set, since their PFs (`enp2s0f0`, `enp3s0f0`) resolve to non-default encap IPs.
+
+#### Layer 2 topology solution
+
+For Layer 2 networks, each remote pod's VIF is represented by a remote LSP in the
+`layer2_ovn_layer2_switch`. To enable correct encapsulation, the local `ovnkube-controller` sets
+`options:requested-encap-ip` on the remote LSP based on `encap_ip` in the remote pod's network
+annotation. If the annotation does not contain `encap_ip`, the option is omitted and OVN uses the
```
npinaeva, line 313: can it be added after the LSP is created? in case encap ip annotation comes after the networks

npinaeva (commented):
I think we should add compatibility constraints with the EVPN VTEP feature and how they can/can not work together

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: > I find current VTEP integration pretty strange, is it only used for tunnel-id generation?

If you are talking about below annotation:
```yaml
k8s.ovn.org/node-encap-ip-configs: '[
  {"encap-ip":"10.0.0.2","tunnel-id":6},
  {"encap-ip":"10.0.0.3","tunnel-id":7},
  {"encap-ip":"10.0.0.4","tunnel-id":8}]'
```
yes, it's only for storing encap ip to tunnel-id mapping. `ovnkube-node` exposes `k8s.ovn.org/node-encap-ips` (as in the existing implementation), and `ovnkube-cluster-manager` allocates tunnel IDs for them and add `k8s.ovn.org/node-encap-ip-configs` to node.

> We now store the per-node external_ids:ovn-pf-encap-ip-mapping but can we make it closer to the existing VTEP

Do you mean the VTEP CRD above? pfName is only meaningful on the local node. A pod gets a VF PCI ID, which can’t be mapped to a pfName on another node. Also, different SKUs may use different pfNames for the same vtep.

> how are the encap IPs assigned to PFs and can we make it more transparent in the API?

This is statically configured when the node is provisioned. Exposing it through the API doesn’t seem to provide any additional value or required by any use case.

> We now store the per-node external_ids:ovn-pf-encap-ip-mapping but can we make it closer to the existing VTEP

My suggestion is to avoid involving the VTEP CRD. The goal of this OKEP is to minimize required changes and avoid interference with other features.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: > why resolving encap ip during CNI ADD is not mentioned as an alternative?

That’s what we explicitly want to avoid, since it requires additional Pod patch requests. From scalability perspective, it’s not a solid alternative.

> Also this needs more detail about what happens when the resources are not yer allocated on pod add and we are supposed to retry it

That’s an implementation detail. I think it’s clear that retries are required when waiting for asynchronously allocated resources, so no need to call it out explicitly, keeping the doc concise.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: yes, it can be done, some changes are required to update LSP's `option:requested-encap-ip` when `encap_ip` comes up in annotation.

l8huang:
@npinaeva thanks for looking at it.

> I think we should add compatibility constraints with the EVPN VTEP feature and how they can/can not work together

The scope of this OKEP has been narrowed and no longer involves any NAD/UDN-level mapping to VTEP resources. So I think it’s better to exclude VTEP CRD entirely from the design.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
npinaeva: well I would argue it is not obvious that one way is better than the other, so I would like both to be mentioned with a clear statement why one is better. Also we need to mention the potential effect on pod creation latency, because of those retries

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
npinaeva: I agree that we probably shouldn't use vtep for this feature. Even though both are VTEPs, what we are doing now involves cross-node connection via OVN with tunnel ID allocation and so on, while existing EVPN VTEP is just a vxlan. I also wouldn't change the existing `k8s.ovn.org/vteps` annotation and just come up with another one instead, like `k8s.ovn.org/ovn-vteps` (could use a better name) and put ips with tunnel-ids there.
This way we keep VTEP as a CRD separate from the ovn-based multi-vtep setup.

l8huang: Great, we are aligned. Tim was involved in the discussion of the VTEP CRD and `k8s.ovn.org/vteps`, I need one more confirm from him, then I will update the design accordingly.

docs/okeps/okep-5535-support-multi-nic-and-multi-vtep-in-ic-mode.md
l8huang: got it, I updated the doc and added "Resolve `encap_ip` During CNI ADD CMD" section to Alternatives.

Issue #5552, opened 2025-09-02
Allow specifying UDNs on specific groups of nodes


trozet:
### What would you like to be added?

Would like the ability to select certain nodes that a UDN should be available on.

### Why is this needed?

Scale and efficiency. Today creating many UDNs results in creating a bunch of OVN topologies on all nodes with IC. The control plane overhead here is heavy, and in some use cases I may only want to allocate a few nodes for a tenant. In that case, there is no reason to render their UDN on all nodes.

PR #5553, opened 2025-09-02
OKEP-5552: Add support for dynamic UDN allocation per node

trozet:
Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5552

## Summary by CodeRabbit

* **Documentation**
  * Added an Enhancement Proposal: "Dynamic UDN Node Allocation" under Enhancement Proposals.
  * Navigation updated to include the new proposal for easier discovery; no other nav items changed.
  * Proposal outlines problem, goals, design (per-node allocation), status reporting, testing and rollout guidance, operational considerations, risks, and confirms no public API changes.

### DIFF
--- a/docs/okeps/okep-5552-dynamic-udn-node-allocation.md
+++ b/docs/okeps/okep-5552-dynamic-udn-node-allocation.md
@@ -1,0 +1,205 @@
+# OKEP-5552: Dynamic UDN Node Allocation
+
+* Issue: [#5552](https://github.com/ovn-org/ovn-kubernetes/issues/5552)
+
+## Problem Statement
+
+When scaling UDNs, the control-plane cost of rendering a topology is high. This is the core limiting factor to
+being able to scale to 1000s of UDNs. While there are plans to also improve network controller performance with UDNs,
+there is still valuable savings to be had by not rendering UDNs on nodes where they are not needed.
+
+An example use case where this makes sense is when a Kubernetes cluster has its node resources segmented per tenant. In
+this case, it only makes sense to run the tenant network (UDN) on the nodes where a tenant is allowed to run pods. This
+allows for horizontal scaling to much higher number of overall UDNs running in a cluster.
+
+## Goals
+
+ * To dynamically allow the network to only be rendered on specific nodes.
+ * To increase overall scalability of the number UDNs in a Kubernetes cluster with this solution.
+ * To increase the efficiency of ovnkube operations on nodes where a UDN exists, but is not needed.
+
+## Non-Goals
+
+ * To fully solve control plane performance issues with UDNs. There will be several other fixes done to address that
+   outside of this enhancement.
+ * To provide any type of network security guarantee about exposing UDNs to limited subset of nodes.
+
+## Future Goals
+
+ * Potentially enabling this feature on a per UDN basis, rather than globally.
+
+## Introduction
+
+The purpose of this feature is to add a configuration knob that users can turn on which will only render UDNs on nodes
+where pods exist on that UDN. This feature will allow for higher overall UDN scale and less per-node control plane resource usage
+under conditions where clusters do not have pods on every node, with connections to all UDNs. For example, if I have
+1000 UDNs and 500 nodes, if a particular node only has pods connected to say 200 of those UDNs, then my node is only
+responsible for rendering 200 UDNs instead of 1000 UDNs as it does today.
+
+This can provide significant control plane savings, but comes at a cost. Using the previous example, if a pod is now
+launched in UDN 201, the node will have to render UDN 201 before the pod can be wired. In other words, this introduces
+a one time larger pod latency cost for the first pod wired to the UDN. Additionally, there are more tradeoffs with other
+feature limitations outlined later in this document.
+
+## User-Stories/Use-Cases
+
+Story 1: Segment groups of nodes per tenant
+
+As a cluster admin, I plan to dedicate groups of nodes to either a single tenant or small group of tenants. I plan
+to create a CUDN per tenant, which means my network will only really need to exist on this group of nodes. I would
+like to be able to limit this network to only be rendered on that subset of nodes.
+This way I will be able to have less resource overhead from OVN-Kubernetes on each node,
+and be able to scale to a higher number of UDNs in my cluster.
+
+## Proposed Solution
+
+The proposed solution is to add a configuration knob to OVN-Kubernetes, "--dynamic-udn-allocation", which will enable
+this feature. Once enabled, NADs derived from CUDNs and UDNs will only be rendered on nodes where there is a pod
+scheduled in that respective network. Additionally, if the node is scheduled as an Egress IP Node for a UDN, this node
+will also render the UDN.
+
+When the last pod on the network is deleted from a node, OVNK will not immediately tear down the UDN.
+Instead, OVNK will rely on a dead timer to expire to conclude that this UDN is no longer in use and
+may be removed. This timer will also be configurable in OVN-Kubernetes as "--udn-deletion-grace-period".
+
+### API Details
+
+There will be no API changes. There will be new status conditions introduced in the section below.
+
+### Implementation Details
+
+In OVN-Kubernetes we have three main controllers that handle rendering of networking features for UDNs. They exist as
+ - Cluster Manager - runs on the control-plane, handles cluster-wide allocation, rendering of CUDN/UDNs
+ - Controller Manager - runs on a per-zone basis, handles configuring OVN for all networking features
+ - Node Controller Manager - runs on a per-node basis, handles configuring node specific things like nftables, VRFs, etc.
+
+With this change, Cluster Manger will be largely untouched, while Controller Manager and Node Controller Manager will be
+modified in a few places to filter out rendering UDNs when a pod doesn't exist.
+
+#### Internal Controller Details
+
+In OVN-Kubernetes we have many controllers that handle features for different networks, encompassed under three
+controller manager containers. The breakdown of how these will be modified is outlined below:
+
+* Cluster Manager
+  * UDN Controller — No change
+  * Route Advertisements Controller — No change
+  * Egress Service Cluster — Doesn't support UDN
+  * Endpoint Mirror Controller — No change
+  * EgressIP Controller — No change
+  * Unidling Controller — No change
+  * DNS Resolver — No change
+  * Network Cluster Controller — Modified to report status and exclude nodes not serving the UDN
+* Controller Manager (ovnkube-controller)
+  * Default Network — No change
+  * NAD Controller — Ignore NADs for UDNs that are not active on this node (no pods for the UDN and not an EIP node)
+* Node Controller Manager
+  * Default Network — No change
+  * NAD Controller — Ignore NADs for UDNs that are not active on this node (no pods for the UDN and not an EIP node)
+
+The resulting NAD Controller change will filter out NADs that do not apply to this node, stopping NAD keys from being
+enqueued to the Controller Manager/Node Controller Manager's Network Manager. Those Controller Managers will not need
+to create or run any sub-controllers for nodes that do not have the network. To do this cleanly, NAD Controller will be
+modified to hold a filterFunc field, which the respective controller manager can set in order to filter out NADs. For
+Cluster Manager, this function will not apply, but for Controller Manager and Node Controller Manager it will be a function
+that filters based on if the UDN is serving pods on this node.
+
+#### New Pod/EgressIP Tracker Controller
+
+In order to know whether the Managers should filter out a UDN, a pod controller and egress IP controller will be used
+in the Managers to track this information in memory. The pod controller will be a new level driven controller for
+each manager. For Egress IP, another new controller will be introduced that watches EgressIPs, Namespaces, and NADs in
+order to track which NAD maps to a node serving Egress IP.
+
+When Managers are created, they will start these Pod/EgressIP Tracker Controllers, and set a filterFunc on NAD Controller.
+The filterFunc will query the aforementioned controllers to determine if the NAD being synced matches the local node. If
+not, then NADController will not create the UDN controller for that network.
+
+Additionally, the Pod/EgressIP Tracker Controllers will expose a callback function, called "onNetworkRefChange". When
+the first pod is detected as coming up on a node + NAD combination, or the node activates as an Egress IP node for the
+first time, onNetworkRefChange will be triggered, which allows a callback mechanism to be leveraged for events. The
+Controller Manager and Node Controller Manager will leverage this callback, so that they can trigger NAD Controller to
+reconcile the NAD for these events. This is important as it provides a way to signal that NADController should remove
+a UDN controller if it is no longer active, or alternatively, force the NAD Controller to reconcile a UDN Controller if for example,
+a new remote node has activated.
+
+#### Other Controller Changes
+
+The Layer3 network controller will need to filter out nodes where the UDN is not rendered. Upon receiving events,
+they will query a Manager function called NodeHasNAD. Managers will export a Tracker interface, that only contains this
+method for UDN Controllers to query. The implementation of NodeHasNAD will rely on the Manager querying their pod and
+egress IP trackers.
+
+Upon UDN activation of a remote node, these controllers will need to receive events in order to reconcile the new remote node. 
+To do this, the corresponding tracker will trigger its callback, "onNetworkRefChange". That will trigger the Manager
+to ask NAD Controller to reconcile the UDN controller belonging to this NAD. Once that Layer 3 UDN controller reconciles,
+it will walk nodes and determine what needs to be added or removed. It will take the applicable nodes, set their
+syncZoneICFailed status, then immediately queue the objects to the retry framework with no backoff. This will allow
+the Zone IC (ZIC) controller to properly configure the transit switch with the remote peers, or tear it down, if necessary.
+
+#### Status Condition and Metric Changes
+
+A new status condition will be added to CUDN/UDN that will indicate how many nodes are selected for a network:
+```yaml
+status:
+  conditions:
+    - type: NodesSelected
+      status: "True"
+      reason: DynamicAllocation
+      message: "5 nodes rendered with network"
+      lastTransitionTime: 2025-09-22T20:10:00Z
+```
+
+If the status is "False", then no nodes are currently allocated for the network - no pods or egress IPs assigned.
+
+Cluster Manager will leverage instances of EgressIP and Pod Trackers in order to use that data for updating this status.
+The nodes serving a network are defined as a node with at least one OVN networked pod or having an Egress IP assigned to
+it on a NAD that maps to a UDN or CUDN.
+
+Additionally, events will be posted to the corresponding UDN/CUDN when nodes have become active or inactive for
+a node. This was chosen instead of doing per node status events, as that can lead to scale issues. Using events provides
+the audit trail, without those scale implications. The one drawback of this approach pertains to UDN deactivation. There
+is an "udn-deletion-grace-period" timer used to delay deactivation of a UDN on a node. This is to prevent churn if a pod
+is deleted, then almost immediately re-added. Without storing the timestamp in the API, we are relying internally on in
+memory data. While this is fine for normal operation, if OVN-Kube pod restarts, we lose that context. However, this should
+be fine as when we restart we have to walk and start all network controllers anyway, so we are not really creating a lot of
+extra work for OVN-Kube here.
+
+A metric will also be exposed which allows the user to track over time how many nodes were active for a particular
+network.
+
+### Testing Details
+
+* Unit Tests will be added to ensure the behavior works as expected, including checking that
+OVN switches/routers are not created there is no pod/egress IP active on the node, etc.
+* E2E Tests will be added to create a CUDN/UDN with the feature enabled and ensure pod traffic works correctly between nodes.
+* Benchmark/Scale testing will be done to show the resource savings of 1000s of nodes with 1000s of UDNs.
+
+### Documentation Details
+
+* User-Defined Network feature documentation will be updated with a user guide for this new feature.
+
+## Risks, Known Limitations and Mitigations
+
+Risks:
+ * Additional first-pod cold start latency per UDN/node. Could impact pod readiness SLOs.
+ * Burst reconcile load on large rollouts of pods on inactive nodes.
+
+Limitations:
+ * No OVN central support.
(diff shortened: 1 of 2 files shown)


martinkennelly:
Nice. For EIP we have a set of nodes as egressable and assume all networks are present locally.  Can you document how its going to behave ? Its a shame we cannot report errors today on the EIP status where we could say we are not serving selected pods on network x because network not on that node or that some egress Nodes can't possible serve network x because network not on that node. Can you describe what happens when we have multiple nodes as egressable but only a subset have the network.

npinaeva:
well written, but I think it requires some more info on what happens with existing feature that expect every node to have UDN topo, e.g. for L3 transit switch and L2 switch/transit router all nodes need to know now which nodes are a part of the network to create remote ports, so they will need to watch nodeSelector and all node labels, the same applies e.g. to Connecting UDNs.  For every NAD this should be simple enough to watch only one nodeSelector, but for multi-NAD features (like connecting UDNs) you will need to check node's label changes against all involved UDNs.
Also, will nodeSelector go into the NAD spec or how it will be handled by the network controllers?

trozet:
@npinaeva good points

For UDN Connect it seems natural that the UDN that you want to connect to should also have selected the same nodes? Alternatively, you could infer that the UDN should be on this node if a UDN with a nodeSelector connects to it, but that seems like a bad path to go down.


> Also, will nodeSelector go into the NAD spec or how it will be handled by the network controllers?

I was not planning on putting it in the NAD spec. I was going to do a reverse lookup to find the UDN in the lister and see what its nodeSelector is inside of NAD driven controllers.

docs/okeps/okep-5552-udn-node-selector.md
```diff
@@ -0,0 +1,200 @@
...
+No known risks. This is not a substitute for overall performance improvements for UDN, but rather an optimization for
+clusters where UDNs are only needed on a subset of nodes.
+
+## OVN Kubernetes Version Skew
+
+Targeted for release 1.2.
+
+## Alternatives
+
+Dynamically detecting if a UDN is needed on a node. This could be done by not starting network controllers until a pod is
```
npinaeva: I think this option could completely replace the nodeSelector when scalability is important.
So you will get some extra pod startup delay, but you don't need to label your nodes for every UDN, just schedule your pods where you want them and that's it.
For the last pod deletion, we could add some timeout before removing the UDN topology to avoid churn (like 5 minutes).
It will also potentially scale better, because you don't need to watch node labels, and no need to handle these error cases where pods are scheduled on the nodes that weren't selected by the UDN. So this should be a simpler API.

npinaeva:
> For UDN Connect it seems natural that the UDN that you want to connect to should also have selected the same nodes? 

How so? another UDN can be scheduled wherever I think

docs/okeps/okep-5552-udn-node-selector.md
tssurya: A few thoughts (after speaking with Tim in DM):
Between an API change with node selectors v/s config knob to turn off controllers dynamically based on pods I prefer the non-API change as well. i.e turn off controllers based on pod count. 
Problem with node selectors for me is I don't see a tenant use case. Nodes are priviledged, I don't see how the tenant can know what label to set when they create their UDNs. For CUDNs, for admins its fine. So for end user tenant its hard to know which label to match with. I haven't seen non-admin APIs have nodeselectors before..

This has the side effect of the first pod getting created going to have a longer delay for the networks to be setup.
During upgrades, when the node is drained ovnk is going to abruptly shutdown w/o cleaning up controllers - it comes back up realises no pods now starts deleting and then we will also start recreating all pods now on this node and then we need to wait for all network's tear down to finish - so the bulk pods are waiting for initial add - but if its behind a config knob that is fine

1. Nodeport services won't work if UDNs are turned off on some nodes - I assume the "nodeport service off flag will have to be used in conjunction to this? but not sure how the dynamic nature will work?
2. Load balancers of etp=cluster will have same problem?
3. EgressIPs - being central controller - depending on whether there are 0 pods or 1 pods, that node can or cannot become egress node - I don't like that design much. We will flap too much - so egress node stops becoming egress nodes when pods go to 0 and that is causing unnecessary reassignment.. need to think if there are better alternatives else the config won't work for egressips

trozet: Thanks for the feedback @npinaeva @tssurya. I'll update proposal to be dynamic rather than expose nodeSelector.

@tssurya as for the egress IPs we could make it so that if a node is selected as egress IP it automatically also gets the network on it.

for services on nodes without the UDN...that is a problem. nodeport wont work. loadbalancer/external IP service with etp=cluster is another thing that won't work. Perhaps we can just document this and not implement a workaround for it right now.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+
+As a cluster admin, I plan to dedicate groups of nodes to either a single tenant or small group of tenants. I plan
+to create a CUDN per tenant, which means my network will only really need to exist on this group of nodes. I would
+like to be able to limit this network to only be rendered on that subset of nodes.
+This way I will be able to have less resource overhead from OVN-Kubernetes on each node,
+and be able to scale to a higher number of UDNs in my cluster.
+
+## Proposed Solution
+
+The proposed solution is to add a configuration knob to OVN-Kubernetes, "--dynamic-udn-allocation", which will enable
```
npinaeva, line 56: I am not sure if this should be a cluster-wide config or a per-UDN config, do you think it is safe to assume that all or none UDNs should be dynamic? I guess that also depends on a better understanding of dynamic allocation side-effects, because if at least one of your UDNs can't afford these side-effects, you can't use this feature at all.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+   - Route Advertisements Controller - No change
+   - Egress Service Cluster - Doesn't support UDN
+   - Endpoint Mirror Controller - No change
+   - EgressIP Controller - No change
+   - Unidling Controller - No change
+   - DNS Resolver - No change
+   - Network Cluster Controller - Modified to report status correctly and filter out nodes not serving the UDN
+- Controller Manager (ovnkube-controller)
+   - Default Network - No change
+   - NAD Controller - Modified to ignore NADs that belong to UDNs that do not match nodeSelector
```
npinaeva: there is no nodeSelector anymore?

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+- Node Controller Manager
+   - Default Network - No change
+   - NAD Controller - Modified to ignore NADs that belong to UDNs that do not match nodeSelector
+
+The resulting NAD Controller change will filter out NADs that do not apply to this node, stopping NAD keys from being
+enqueued to the Controller Manager/Node Controller Manager's Network Manager. Those Controller Managers will not need
+to create or run any sub-controllers for nodes that do not have the network. To do this cleanly, NAD Controller will be
+modified to hold a filterFunc field, which the respective controller manager can set in order to filter out NADs. For
```
npinaeva, line 103: I am not quite following how filterFunc should work from this description, could you add a function args and return values?

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+#### New Pod Controller
+
+In order to know whether the Managers should filter out a UDN, a pod controller and egress IP controller will be used
+in the Managers to track this information in memory. The pod controller will be a new level driven controller for
+each manager. For Egress IP, Controller Manager already has an EIP controller that can be leveraged to hold this information.
+For Node Controller Manager, a new EIP controller will need to be added.
+
+#### Other Controller Changes
+
+The network controllers: Layer3, Layer2, Localnet will need to filter out their node handlers for nodes where the UDN
```
npinaeva: do you mean remote nodes here? does it mean that every filter function know all active nodes for every UDN?

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+is not rendered. Upon receiving events, they will also call the filter function from the NAD Controller to determine if
+they need to ignore resources or not.
+
+Upon UDN activation of a node, these controllers will need to receive events in order to reconcile the new node. The reconcile
+function for network controller will be leveraged to find the new resources and queue them to the retry framework
+with no backoff.
+
+#### Status Condition Changes
+
+A new status condition will be added to CUDN/UDN that will indicate how many nodes are serving a network:
```
npinaeva: We need to explain which nodes are included, I think it has both nodes with pods (and nodes waiting for a timeout after the last pod is deleted) and egressIP nodes?

npinaeva (commented):
I would also like to see a full list of side-effects, like nodePorts not working on unused nodes (I feel like there should be something else)? And if possible some estimation of how much longer the first-pod-one-node creation time will be (at least with no other load), I assume we may need to optimize the pod retry mechanism to avoid extra retries during UDN initialization

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+#### Other Controller Changes
+
+The network controllers: Layer3, Layer2, Localnet will need to filter out their node handlers for nodes where the UDN
+is not rendered. Upon receiving events, they will also call the filter function from the NAD Controller to determine if
+they need to ignore resources or not.
+
+Upon UDN activation of a node, these controllers will need to receive events in order to reconcile the new node. The reconcile
+function for network controller will be leveraged to find the new resources and queue them to the retry framework
+with no backoff.
+
```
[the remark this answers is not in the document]
trozet: @coderabbitai this is out of scope for this enhancement. The behavior today of ovn-kubernetes retry framework is to add objects with no backoff when we need to signal kubernetes resources need to be eventually processed.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+A new status condition will be added to CUDN/UDN that will indicate how many nodes are serving a network:
+```yaml
+status:
+  conditions:
+    - type: NodesSelected
+      status: "True"
+      reason: DynamicAllocation
+      message: "5 nodes rendered with network"
+```
+
```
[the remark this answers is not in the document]
trozet: @coderabbitai I don't want to include per node details because if we have 1000 nodes and all are selected it will be giant list

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,185 @@
...
+status:
+  conditions:
+  - type: NodeUnused
+    status: "True"
+    reason: LastPodRemoved
+    message: "The last pod on node worker-1 using this UDN has been removed"
+    lastTransitionTime: "2025-09-15T14:00:00Z"
+    nodeName: worker-1
+```
+
```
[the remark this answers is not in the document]
trozet: I was thinking there will not be a ton of these conditions. The condition only happens when a network was rendered on a node, then torn down because of lack of entities requiring the network on the node. Then if the network comes back up on the node later, I would clear the condition. There wont be a status for every single node.

trozet:
> I would also like to see a full list of side-effects, like nodePorts not working on unused nodes (I feel like there should be something else)? And if possible some estimation of how much longer the first-pod-one-node creation time will be (at least with no other load), I assume we may need to optimize the pod retry mechanism to avoid extra retries during UDN initialization

I listed the limitations in the OKEP. I'm not sure what other side-effects you are looking for me to document. I'm just starting to test it out now so will update if I find anymore issues.

It's really hard to quantify how much longer the first pod on node creation will take. That depends on a variety of factors, like how many UDNs are currently rendered on the node. I can do some testing and provide the results once I have the code fully working.

I don't think we need to mess with the pod retry framework. The UDN controllers each have their own retry framework. All of them will ignore these pods, except for default network controller, which should be allocate everything OK. Once the new UDN controller spins up and starts its WatchPods then the pod will be handled.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: hmm fair point. It would be easy to make it a UDN CRD field instead of a global configuration. I don't think there would be any issues by doing it per UDN either. @tssurya wdyt?

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: I'll add some more detail here, but if you want to see the POC code:
https://github.com/trozet/ovn-kubernetes/commit/c23c4473bdc5ad6046b4df067d9e23d084169311#diff-5edfea6278834d4a7b2cffe3557ff477f54e060d7d75935ebce36a427114b08fR561

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: yeah remote nodes. Will update that wording. I don't think it makes sense to pass the filter function down to those controllers. The filter function is used by nad controller to ignore ensuring networks for nads that are have no resources on the local node. I will probably end up passing the podTracker itself down to the controller, then those controllers can query the podTracker to see if a remote node has any resources/needs to be ignored.

There is a callback I added in podTracker that allows controllerManager to register a function that should be called when the network should be reconciled:
```
func (cm *ControllerManager) OnNetworkRefChange(node, name string, active bool) {
	klog.V(5).Infof("Network change triggered by pod/egress IP events on node: %s, NAD: %s, active: %t", name, node, active)
	cm.networkManager.Interface().Reconcile(name)
}
```

This way we can ask NADController to process the relevant NAD again. Why I'm telling you this is because by calling syncNAD, if the controller already exists, it will call the sub-controller (like layer3) reconcile, which will today sets nodes as needing update:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/layer3_user_defined_network_controller.go#L634

which eventually calls:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/base_network_controller.go#L241

so we can leverage this to trigger a reconcile and request remote nodes to be retried when we detect a network is now live on a remote node.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: right a combination of if node has pod || node is egressIP

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
npinaeva: not sure if this is the most stable mechanism to utilize, but it is more of an implementation details, we can discuss it later

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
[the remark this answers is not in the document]
trozet: we dont need per zone information. With Interconnect we only support 1 node per zone. So it is the equivalent of 1 status per node.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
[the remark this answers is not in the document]
trozet: I'm thinking actually I do not need per node status. I plan to move to just posting an event on the UDN when a node is going inactive or active. That way it will scale better and I no longer need to hold that timestamp in the kapi. I plan to just persist the timestamp in memory, and if OVNK is rebooted, it wont have that context, but that is OK as it will be re-rendering all networks anyway and will determine that it should immediately clean up the network.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: I thought about this some more, and I was concerned about allowing per UDN configuration. The reason is UDN is not an admin API, so it puts the power of this type of internal network scheduling in a tenant/user's hands. I think that could be risky, as an Admin may want their entire cluster to use dynamic allocation, and then a tenant could just disable it for their network. Perhaps for now the safest approach is to just use an all or nothing config flag, and in the future if we have a use case we can always add the api field?

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: fwiw I ended up abstracting out a Tracker Interface that contains a single method. This Interface is implemented by the Controller Manager and passed down to the network controllers, so that they have something to query. TrackerController is implemented by Pod and EgressIP Trackers.

```
type Tracker interface {
	// NodeHasNAD returns true if the given node has at least one pod using the NAD.
	NodeHasNAD(node, nad string) bool
}

// TrackerController is the minimal interface ControllerManager needs.
type TrackerController interface {
	Tracker
	Start() error
	Stop()
}
```

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
npinaeva: sure, I think it would be nice to write it down in the enhancement though and list it as a potential future change

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,195 @@
...
+As a cluster admin, I plan to dedicate groups of nodes to either a single tenant or small group of tenants. I plan
+to create a CUDN per tenant, which means my network will only really need to exist on this group of nodes. I would
+like to be able to limit this network to only be rendered on that subset of nodes.
+This way I will be able to have less resource overhead from OVN-Kubernetes on each node,
+and be able to scale to a higher number of UDNs in my cluster.
+
+## Proposed Solution
+
+The proposed solution is to add a configuration knob to OVN-Kubernetes, "--dynamic-udn-allocation", which will enable
```
dave-tucker, line 56: Another configuration knob 😆 I know it's not part of the OKEP process, but we should probably have a conversation in a community meeting at some point about graduation criteria. When does something become not a knob, and is instead the default behaviour.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,195 @@
...
+Upon UDN activation of a remote node, these controllers will need to receive events in order to reconcile the new remote node. 
+To do this, the corresponding tracker will trigger its callback, "onNetworkRefChange". That will trigger the Manager
+to ask NAD Controller to reconcile the UDN controller belonging to this NAD. Once that Layer 3 UDN controller reconciles,
+it will walk nodes and determine what needs to be added or removed. It will take the applicable nodes, set their
+syncZoneICFailed status, then immediately queue the objects to the retry framework with no backoff. This will allow
+the Zone IC (ZIC) controller to properly configure the transit switch with the remote peers, or tear it down, if necessary.
+
+#### Status Condition Changes
+
+A new status condition will be added to CUDN/UDN that will indicate how many nodes are serving a network:
```
dave-tucker: This seems like a useful thing to expose as a metric - would be pretty useful to see in a time-series format.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,195 @@
...
+```yaml
+status:
+  conditions:
+    - type: NodesSelected
+      status: "True"
+      reason: DynamicAllocation
+      message: "5 nodes rendered with network"
```
dave-tucker, line 149: `LastTransitionTime` might be useful here too.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -0,0 +1,195 @@
...
+the Zone IC (ZIC) controller to properly configure the transit switch with the remote peers, or tear it down, if necessary.
+
+#### Status Condition Changes
+
+A new status condition will be added to CUDN/UDN that will indicate how many nodes are serving a network:
+```yaml
+status:
+  conditions:
+    - type: NodesSelected
+      status: "True"
```
dave-tucker, line 147: Is "False" as a failure condition? Should I be concerned?
Or is it normal when you have a UDN that currently has no scheduled pods.

dave-tucker:
> Some minor comments but overall, LGTM ✅

Having slept on this, I have one final reservation 😆 , but not enough to remove the LGTM.

> While there are plans to also improve network controller performance with UDNs,
there is still valuable savings to be had by not rendering UDNs on nodes where they are not needed.

@trozet let's say we do the UDN controller performance work and get phenomenal results. is there a future where we are so pleased with the UDN controller performance that we would then deprecate this? or do you still see this being something useful longer term?

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: great point. I will add that as well.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: Yeah. If the mode didn't have the limitations that it does, then I would have proposed we turn it on by default if it met the performance/scale bar. Not sure what else we can do here.

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
trozet: False is when there are no nodes selected for the UDN. Normal when there are no scheduled pods. Maybe I should change the type from NodesSelected to "NodesProvisioned" or "NodesActive" or something.

trozet:
> > Some minor comments but overall, LGTM ✅
> 
> Having slept on this, I have one final reservation 😆 , but not enough to remove the LGTM.
> 
> > While there are plans to also improve network controller performance with UDNs,
> > there is still valuable savings to be had by not rendering UDNs on nodes where they are not needed.
> 
> @trozet let's say we do the UDN controller performance work and get phenomenal results. is there a future where we are so pleased with the UDN controller performance that we would then deprecate this? or do you still see this being something useful longer term?

I think it is really hard to predict. If we end up making phenomenal performance/scale gains to the point where there is no benefit to having this feature, then I'm all for deprecating it. We need something now though as a stop gap until we can fix the rest.

trozet:
last push was just rebase

trozet:
thanks @dave-tucker and @npinaeva

PR #5554, opened 2025-09-02
Upstream udn fix

cathy-zhou:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -409,10 +409,15 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 			// sample flow: cookie=0xdeff105, duration=2319.685s, table=2, n_packets=496, n_bytes=67111, priority=300,
 			//              ip,nw_dst=10.96.0.1 actions=mod_dl_dst:02:42:ac:12:00:03,output:"patch-breth0_ov"
 			// This flow is used for UDNs and advertised UDNs to be able to reach kapi and dns services alone on default network
+			mod_vlan_id := ""
```
trozet: I would use a pattern like this to avoid the awkward "%soutput:%s":

```
outputActions := fmt.Sprintf("output:%s", defaultNetConfig.OfPortPatch)
if config.GatewayVLANID != 0 {
  outputActions = fmt.Sprintf("mod_vlan_vid:%d,%s", config.GatewayVLANID, outputActions)
}

flows := []string{fmt.Sprintf("cookie=%s, priority=300, table=2, %s, %s_dst=%s, "+
			"actions=set_field:%s->eth_dst,%s",
            nodetypes.DefaultOpenFlowCookie, ipPrefix, ipPrefix, service.Spec.ClusterIP,
            npw.ofm.getDefaultBridgeMAC().String(), outputActions)}

```

go-controller/pkg/node/gateway_udn.go
```diff
@@ -214,81 +225,92 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
...
+	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
+		nodeSubnets, err := udng.getLocalSubnets()
+		var mgmtIPs []*net.IPNet
+		for _, subnet := range nodeSubnets {
+			mgmtIPs = append(mgmtIPs, udng.GetNodeManagementIP(subnet))
+		}
+		if err != nil {
+			return fmt.Errorf("failed to get node subnets for network %s: %w", udng.GetNetworkName(), err)
+		}
+		if err = udng.openflowManager.addNetwork(udng.NetInfo, nodeSubnets, mgmtIPs, udng.masqCTMark, udng.pktMark, udng.v6MasqIPs, udng.v4MasqIPs); err != nil {
```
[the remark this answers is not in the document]
cathy-zhou: fixed

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1008,10 +1008,12 @@ func (nc *DefaultNodeNetworkController) Init(ctx context.Context) error {
 		return fmt.Errorf("failed to set node zone annotation for node %s: %w", nc.name, err)
 	}
 
-	encapIPList := sets.New[string]()
-	encapIPList.Insert(strings.Split(config.Default.EffectiveEncapIP, ",")...)
-	if err := util.SetNodeEncapIPs(nodeAnnotator, encapIPList); err != nil {
-		return fmt.Errorf("failed to set node-encap-ips annotation for node %s: %w", nc.name, err)
+	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
```
trozet, line 1018: this was just merged: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5543

cathy-zhou:
@girishmg @trozet Please take a look. Thanks!

go-controller/pkg/node/gateway_udn.go
```diff
@@ -125,13 +125,24 @@ func NewUserDefinedNetworkGateway(netInfo util.NetInfo, node *corev1.Node, nodeL
...
+	var intfName string
+	if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
+		if gw.openflowManager == nil {
+			return nil, fmt.Errorf("openflow manager has not been provided for network: %s", netInfo.GetNetworkName())
+		}
+		intfName = gw.openflowManager.defaultBridge.GetGatewayIface()
+	} else {
+		var nexthops []net.IP
+		// dpu-host mode does not have nexthop/gateway interface set, set it here
+		nexthops, intfName, err = getGatewayNextHops()
```
trozet: note in initGatewayDPUHost we modify the config.GatewayInterface:
```
	// If no gateway interface is explicitly configured, use the detected interface
	if config.Gateway.Interface == "" {
		config.Gateway.Interface = kubeIntf
	}
```

that happens during DefaultNodeNetworkController.Start(), but NetworkManager is started before this:

```
func (ncm *NodeControllerManager) Start(ctx context.Context) (err error) {
...
	if ncm.networkManager != nil {
		err = ncm.networkManager.Start()
		if err != nil {
			return fmt.Errorf("failed to start NAD controller: %w", err)
		}
	}

	err = ncm.defaultNodeNetworkController.Start(ctx)
	if err != nil {
		return fmt.Errorf("failed to start default node network controller: %v", err)
	}
```

which means this is a race where you could doing getGatewayNextHops before the config variables are properly set.


perhaps it makes sense to move initGatewayDPUHost to NodeControllerManager.Start, before networkManager.Start.

@aserdean FYI

go-controller/pkg/node/gateway_shared_intf.go
cathy-zhou: this is follow the current convention. see https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/bridgeconfig/bridgeflows.go#L361

go-controller/pkg/node/default_node_network_controller.go
cathy-zhou: ok

go-controller/pkg/node/gateway_udn.go
cathy-zhou: @trozet @aserdean I decided to split initGatewayDPUHost() into two parts, just like initGateway(). initGatewayDPUHostPreStart() would initialize gateway interface/nexthops, related routes etc. I believe that should work. Please take a look.

go-controller/pkg/node/gateway_shared_intf.go
cathy-zhou: fixed

go-controller/pkg/node/gateway_udn.go
```diff
@@ -214,81 +214,92 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
...
-		return fmt.Errorf("failed to get node subnets for network %s: %w", udng.GetNetworkName(), err)
-	}
-	if err = udng.openflowManager.addNetwork(udng.NetInfo, nodeSubnets, mgmtIPs, udng.masqCTMark, udng.pktMark, udng.v6MasqIPs, udng.v4MasqIPs); err != nil {
-		return fmt.Errorf("could not add network %s: %v", udng.GetNetworkName(), err)
+		// add loose mode for rp filter on management port
+		mgmtPortName := util.GetNetworkScopedK8sMgmtHostIntfName(uint(udng.GetNetworkID()))
+		if err := addRPFilterLooseModeForManagementPort(mgmtPortName); err != nil {
+			return fmt.Errorf("could not set loose mode for reverse path filtering on management port %s: %v", mgmtPortName, err)
+		}
 	}
```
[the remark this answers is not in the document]
cathy-zhou: @coderabbitai vrf/routes/ip rules should be programed on dpu-host

go-controller/pkg/node/gateway.go
```diff
@@ -495,14 +515,20 @@ func (g *gateway) SetDefaultBridgeGARPDropFlows(isDropped bool) {
...
-	if err != nil {
-		return err
+	// TBD updateSNATRules() gets node host-cidr by accessing gateway.nodeIPManager, which does not
+	// exist in dpu-host mode
+	if config.Gateway.Mode != types.NodeModeFull {
+		err := g.updateSNATRules()
+		if err != nil {
+			return err
+		}
 	}
```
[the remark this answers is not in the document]
cathy-zhou: this is tbd, as node hostcidr should be retrived by another way and not to rely on nodeIPManager

go-controller/pkg/node/gateway.go
```diff
@@ -495,14 +515,20 @@ func (g *gateway) SetDefaultBridgeGARPDropFlows(isDropped bool) {
...
+		g.openflowManager.requestFlowSync()
 	}
-	// let's sync these flows immediately
-	g.openflowManager.requestFlowSync()
-	err := g.updateSNATRules()
-	if err != nil {
-		return err
+	// TBD updateSNATRules() gets node host-cidr by accessing gateway.nodeIPManager, which does not
+	// exist in dpu-host mode
+	if config.Gateway.Mode != types.NodeModeFull {
```
trozet: why would this be != NodeModeFull, shouldn't it be == ? until we fix dpu-host mode? then it should be != DPUMode?

go-controller/pkg/node/gateway.go
trozet, line 536: right under there is a block to queue all services to retry framework.

this triggers, the service framework to do:
```
func (g *gateway) AddService(svc *corev1.Service) error {
	var err error
	var errors []error

	if g.portClaimWatcher != nil {
		if err = g.portClaimWatcher.AddService(svc); err != nil {
			errors = append(errors, err)
		}
	}
	if g.loadBalancerHealthChecker != nil {
		if err = g.loadBalancerHealthChecker.AddService(svc); err != nil {
			errors = append(errors, err)
		}
	}
	if g.nodePortWatcher != nil {
		if err = g.nodePortWatcher.AddService(svc); err != nil {
			errors = append(errors, err)
		}
	}
	if g.nodePortWatcherIptables != nil {
		if err = g.nodePortWatcherIptables.AddService(svc); err != nil {
			errors = append(errors, err)
		}
	}
}

```

where some of these belong in DPU Host and some belong in DPU. Do we need to also have conditionals around these?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -214,81 +214,92 @@ func (udng *UserDefinedNetworkGateway) AddNetwork() error {
...
-		return fmt.Errorf("could not add VRF %d for network %s, err: %v", udng.vrfTableId, udng.GetNetworkName(), err)
-	}
-	if err = udng.addUDNManagementPortIPs(mplink); err != nil {
-		return fmt.Errorf("unable to add management port IP(s) for link %s, for network %s: %w", mplink.Attrs().Name, udng.GetNetworkName(), err)
-	}
-	if err = udng.vrfManager.AddVRFRoutes(vrfDeviceName, routes); err != nil {
-		return fmt.Errorf("could not add VRF %s routes for network %s, err: %v", vrfDeviceName, udng.GetNetworkName(), err)
-	}
 
 	udng.updateAdvertisementStatus()
```
trozet: shouldn't this not happen in NodeModeDPU? it configures ntfables

go-controller/pkg/node/gateway.go
cathy-zhou: you are right. typo when I port my fixes from downstream.

go-controller/pkg/node/gateway_udn.go
cathy-zhou: it also update udng.isNetworkAdvertised, which can be stored in netConfig.Advertised. from what I can see, netConfig.Advertised also affect openflow rules (used in commonFlows() function)

go-controller/pkg/node/gateway.go
cathy-zhou: in function that involves nodePortWatcher, it already checks nodePortWatcher.dpuMode and does things differently there.

go-controller/pkg/node/gateway_udn.go
trozet: good catch

trozet:
re-running bgp failures

trozet:
@cathy-zhou it looks like something is broken with local gw mode. This BGP test keeps failing:

BGP: Pod to external server when default podNetwork is advertised when a client ovnk pod targeting an external server is created [It] tests are run towards the external agnhost echo server [Feature:RouteAdvertisements]

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18014423707/job/51331455894?pr=5554

go-controller/pkg/node/gateway.go
```diff
@@ -495,14 +517,20 @@ func (g *gateway) SetDefaultBridgeGARPDropFlows(isDropped bool) {
...
+		g.openflowManager.requestFlowSync()
 	}
-	// let's sync these flows immediately
-	g.openflowManager.requestFlowSync()
-	err := g.updateSNATRules()
-	if err != nil {
-		return err
+	// TBD updateSNATRules() gets node host-cidr by accessing gateway.nodeIPManager, which does not
+	// exist in dpu-host mode
+	if config.Gateway.Mode == types.NodeModeFull {
```
trozet: this is wrong, comparing gateway mode to NodeModeFull :)

cathy-zhou: fixed

trozet (approved):

PR #5556, opened 2025-09-03
fix: swap quay.io golang image provider

killianmuldoon:
Use the golang image from quay.io/projectquay/golang due to issues pulling the image at quay.io/lib/golang

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/Makefile
+++ b/go-controller/Makefile
@@ -12,14 +12,14 @@ GOPATH ?= $(shell go env GOPATH)
 TEST_REPORT_DIR?=$(CURDIR)/_artifacts
 export TEST_REPORT_DIR
 GO_VERSION ?= 1.24
-GO_DOCKER_IMG = quay.io/lib/golang:${GO_VERSION}
+GO_DOCKER_IMG = quay.io/projectquay/golang:${GO_VERSION}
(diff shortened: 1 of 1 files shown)


go-controller/Makefile
```diff
@@ -12,14 +12,14 @@ GOPATH ?= $(shell go env GOPATH)
 TEST_REPORT_DIR?=$(CURDIR)/_artifacts
 export TEST_REPORT_DIR
 GO_VERSION ?= 1.24
-GO_DOCKER_IMG = quay.io/lib/golang:${GO_VERSION}
+GO_DOCKER_IMG = quay.io/projectquay/golang:${GO_VERSION}
```
killianmuldoon, line 15: I'm assuming this image is well maintained given it's from projectquay - but I'm not really sure.

tssurya (approved):
I'd like to understand why quay.io/lib/golang:${GO_VERSION} is having issues, but given the full CI is red right now, I am going to get this PR in to unblock us.
Would be good to see if there are known issues and something that auto-fixes itself for us to revert back...
quay.io/lib seems to be more of an official place for hosting all kinds of standard images? https://quay.io/organization/lib
@killianmuldoon can we check what other repos are doing? example like kubernetes? maybe adopting something more standard is better?

PR #5559, opened 2025-09-04
feat: get ovs/ovn file path from environment variable

### LINKED ISSUE #5513: [Refactor requirement] get ovs/ovn file paths from environment variables
### What is this issue about?

The reason of this refactor request is that we want to build a general OVN exporter based on the **metrics** package.

In our experience, these file paths do not always match the hard-coded values, and Kubernetes is not involved. Our goal is to provide a non-breaking refactor to ovn-kubernetes to support this purpose.

The scope should cover all the stable release(1.0, 1.1) & master.

The expected scope:

| Environment Variable | Default Value | Description |
|---------------------|---------------|-------------|
| `OVS_RUNDIR` | "/var/run/openvswitch/" | OVS run directory |
| `OVN_RUNDIR` | "/var/run/ovn/" | OVN run directory |
| `OVS_VSWITCHD_PID` | "/var/run/openvswitch/ovs-vswitchd.pid" | OVS vSwitchd PID file path |
| `OVSDB_SERVER_PID` | "/var/run/openvswitch/ovsdb-server.pid" | OVSDB server PID file path |
| `OVN_NBDB_LOCATION` | "/var/lib/openvswitch/ovnnb_db.db" | OVN northbound database location |
| `OVN_SBDB_LOCATION` | "/var/lib/openvswitch/ovnsb_db.db" | OVN southbound database location |

*notes*: The default value will be the same as the current hard code value, user can base on their needs to configure the environment variables.

jneo8:
## 📑 Description

Add support for configuring OVS and OVN file paths through environment variables to improve deployment flexibility:

- Add OVS_RUNDIR and OVN_RUNDIR for runtime directories
- Add OVN_NBDB_LOCATION and OVN_SBDB_LOCATION for database files
- Add OVS_VSWITCHD_PID and OVSDB_SERVER_PID for process monitoring
- Refactor metrics registration to use dependency injection pattern

Partially Fixes #5513

## ✅ Checks

- [x] My code requires tests

## How to verify it

```
# Unit tests
cd ./go-controller
go test ./pkg/util -run "TestGet.*Env" -v
go test ./pkg/metrics -ginkgo.focus="environment" -v
go test ./pkg/metrics -ginkgo.focus="OVS environment variable functions" -v
go test ./pkg/metrics -ginkgo.focus="RegisterOvnDBMetrics dependency injection integration" -v
go test ./pkg/metrics -v
```


go-controller/pkg/util/ovs.go
```diff
@@ -48,19 +49,69 @@ const (
...
+	ovsRunDir       string = getOVSRunDirFromEnv()
+	ovnRunDir       string = getOVNRunDirFromEnv()
+	OvnNbdbLocation string = getOVNNbdbLocationFromEnv()
+	OvnSbdbLocation string = getOVNSbdbLocationFromEnv()
+
+	savedOVSRunDir       = ovsRunDir
+	savedOVNRunDir       = ovnRunDir
+	savedOvnNbdbLocation = OvnNbdbLocation
+	savedOvnSbdbLocation = OvnSbdbLocation
 )
```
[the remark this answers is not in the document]
jneo8: The suggestion is not very correct, we won't real time overrides the env.

go-controller/pkg/metrics/ovn_db.go
```diff
@@ -322,58 +320,104 @@ var (
...
+		ovnRunDir = "/var/run/ovn/"
+	}
+
+	// Ensure path ends with trailing slash
+	if !strings.HasSuffix(ovnRunDir, "/") {
+		ovnRunDir += "/"
+	}
+
+	return ovnRunDir
+}
```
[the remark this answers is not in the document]
jneo8: There functions are similar but have different approach, the `GetOvnRunDir` return the value from *runner* & but `getOVNRunDir` & `getOVSRunDir` only return from environment variable.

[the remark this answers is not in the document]
jneo8: No, runner.ovnRunDir may be either be ovsRunDir or ovnRunDir depends on the environment.

go-controller/pkg/metrics/metrics.go
```diff
@@ -569,9 +569,41 @@ func startMetricsServer(bindAddress, certFile, keyFile string, handler http.Hand
...
+						"Not registering OVN DB Metrics on this Node.", k8sNodeName, err)
+				} else {
+					klog.Infof("Not registering OVN DB Metrics on this Node since OVN DBs are not running on this node.")
+				}
+				return false
+			}
+			return true
+		},
+		stopChan,
+	)
```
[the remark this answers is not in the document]
jneo8: The code is move from `RegisterOvnDBMetrics`, we don't want to change the current implementation.

jneo8:
Hi @kyrtapz 

This should be ready for review. Can I please ask your help to review it?

kyrtapz:
Thanks for the pr @jneo8!
I think it is a good idea to make the paths configurable, have you considered expanding the config instead of doing that through environment variables? This is what we usually do for all other settings of ovn-k and we already have some wiring for OVN settings here: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/402b187d7a9a2ab0a5286430361aa1853c330d98/go-controller/pkg/config/config.go#L517, it should be straightforward to build on top of that.

As a side note, we usually rebase on the main branch instead of merging it into the PR, if you prefer to do that through GUI you need to select it:
image

jneo8:
Hi @kyrtapz

Thanks for the feedback. I did a quick check over the configuration practices and have a couple of questions to clarify before moving forward:

- OvnAuthConfig holds client authentication, but the metrics pkg usually reads from local files, so auth isn’t required to collect metrics.

- Changing OvnAuthConfig would imply changes to CLI flags (e.g., defaults would live there).

- There’s no OVS-specific configuration now.

- Defaults are initialized in `config.initConfigWithPath`, which bundles all configuration. For #5513, external users (like me) don’t need the entire set; initialization should be decoupled so paths can be configured independently.

Given the above, I’d like to propose introducing a separate, path-only configuration:

```
type OvnPathConfig struct {
    OvsRunDir       string
    OvnRunDir       string
    OvsVswitchdPid  string
    OvsdbServerPid  string
    OvnNbDbLocation string
    OvnSbDbLocation string
}

var PathConfig OvnPathConfig
```

So we can:

- Replace hard-coded paths with values from PathConfig.
- Decouple this change from CLI flag updates (we can add flags later without churn).
- Allow external users to override paths by assigning to PathConfig to fit their environments.

Do you think this approach make sense for you?

kyrtapz:
> Hi @kyrtapz
> 
> Thanks for the feedback. I did a quick check over the configuration practices and have a couple of questions to clarify before moving forward:
> 
>     * OvnAuthConfig holds client authentication, but the metrics pkg usually reads from local files, so auth isn’t required to collect metrics.

In the old days we were connecting to a remote DB that was encrypted. Today we use a scoket file, I think we can adapt the struct name to better reflect what it holds and reuse it here. 
 
>     * Changing OvnAuthConfig would imply changes to CLI flags (e.g., defaults would live there).

Not sure what do you mean, adding fields to it or renaming it shouldn't affect the existing user interface of the CLI or the config file. The changes should be backwards compatible. 

>     * There’s no OVS-specific configuration now.
> 
>     * Defaults are initialized in `config.initConfigWithPath`, which bundles all configuration. For [[Refactor requirement] get ovs/ovn file paths from environment variables #5513](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5513), external users (like me) don’t need the entire set; initialization should be decoupled so paths can be configured independently.
> 
> 
> Given the above, I’d like to propose introducing a separate, path-only configuration:
> 
> ```
> type OvnPathConfig struct {
>     OvsRunDir       string
>     OvnRunDir       string
>     OvsVswitchdPid  string
>     OvsdbServerPid  string
>     OvnNbDbLocation string
>     OvnSbDbLocation string
> }
> 
> var PathConfig OvnPathConfig
> ```
> 
> So we can:
> 
>     * Replace hard-coded paths with values from PathConfig.
> 
>     * Decouple this change from CLI flag updates (we can add flags later without churn).
> 
>     * Allow external users to override paths by assigning to PathConfig to fit their environments.
> 
> 
> Do you think this approach make sense for you?

I am not sure I follow how does this proposal differ from extending the existing struct.
Environment variables can be used to render the appropriate config. Can you give me an example of how the proposed solution limits what you need in https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5513?

jneo8:
@kyrtapz thanks for the reply—responses inline:

> In the old days we were connecting to a remote DB that was encrypted. Today we use a scoket file, I think we can adapt the struct name to better reflect what it holds and reuse it here.

Agreed. We can fold the OVN-related values into `OvnAuthConfig` and read them via `OvnNorth` / `OvnSouth`, replacing the current hard-coded values. Concretely, move these in:

- `OVN_RUNDIR`
- `OVN_NBDB_LOCATION`
- `OVN_SBDB_LOCATION`

`OVN_NBDB_LOCATION` and `OVN_SBDB_LOCATION` are DB endpoints, so naming them `DbLocation` (NB/SB) would be clearer. We may still want a shared `OVN_RUNDIR`; duplication across NB/SB is acceptable for now.

> I am not sure I follow how does this proposal differ from extending the existing struct. Environment variables can be used to render the appropriate config.

Here’s the gap I’m addressing: we have three **OVS** paths hard-coded and not represented in any config struct today:

- `OVS_RUNDIR` (e.g., `/var/run/openvswitch/`)
- `OVS_VSWITCHD_PID` (e.g., `/var/run/openvswitch/ovs-vswitchd.pid`)
- `OVSDB_SERVER_PID` (e.g., `/var/run/openvswitch/ovsdb-server.pid`)

These aren’t covered by existing structures, so we should add a small OVS-specific config (or extend an appropriate one). For example:

```go
type OvsPathConfig struct {
    RunDir         string // e.g., /var/run/openvswitch/
    VswitchdPID    string // e.g., /var/run/openvswitch/ovs-vswitchd.pid
    OvsdbServerPID string // e.g., /var/run/openvswitch/ovsdb-server.pid
}
```

This keeps OVN auth/endpoint settings in `OvnAuthConfig`, while OVS filesystem paths live in `OvsPathConfig`. 

---

Does this make sense? If not, happy to do a quick breakout after the next meeting.

kyrtapz:
> Does this make sense? If not, happy to do a quick breakout after the next meeting.

That sounds good yeah!

jneo8:
Hi @kyrtapz 

Hope this message find you well. The newest change should implement our conclusion before.
Can I please ask for your second round review? Thanks.

go-controller/pkg/util/ovs.go
```diff
@@ -743,7 +740,7 @@ type queryResult struct {
 }
 
 func GetOVNDBServerInfo(timeout int, direction, database string) (*OVNDBServerStatus, error) {
-	sockPath := fmt.Sprintf("unix:/var/run/openvswitch/ovn%s_db.sock", direction)
+	sockPath := fmt.Sprintf("unix:%sovn%s_db.sock", config.OvsPaths.RunDir, direction)
 	transact := fmt.Sprintf(`["_Server", {"op":"select", "table":"Database", "where":[["name", "==", "%s"]], `+
```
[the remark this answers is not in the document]
jneo8: The value are the same

jneo8:
rebased.

go-controller/pkg/config/config.go
```diff
@@ -524,10 +537,24 @@ type OvnAuthConfig struct {
 	Scheme         OvnDBScheme
 	ElectionTimer  uint `gcfg:"election-timer"`
 	northbound     bool
+	// RunDir is OVN run directory.
+	RunDir string
```
kyrtapz: Missing gcfg tags for `RunDir` and `DbLocation`.
I would love to see some configfile tests ensuring the new fields work as expected.

go-controller/pkg/config/config.go
```diff
@@ -2602,6 +2624,13 @@ func buildOvnAuth(exec kexec.Interface, northbound bool, cliAuth, confAuth *OvnA
 		defaultAuth = &savedOvnSouth
 	}
 
+	auth := &OvnAuthConfig{
+		northbound: northbound,
+		exec:       exec,
+		RunDir:     defaultAuth.RunDir,
```
kyrtapz, line 2710: Why is this change required? I would assume that `overrideFields` would take care of it.

go-controller/pkg/config/config_test.go
```diff
@@ -2062,4 +2062,54 @@ udn-allowed-default-services= ns/svc, ns1/svc1
 			gomega.Expect(err).ToNot(gomega.HaveOccurred())
 		})
 	})
+
+	Context("OvsPathConfig and OvnAuthConfig path configurations", func() {
```
kyrtapz: The following tests are not testing anything, especially the last two.

go-controller/pkg/metrics/metrics.go
```diff
@@ -571,7 +571,37 @@ func startMetricsServer(bindAddress, certFile, keyFile string, handler http.Hand
 
 func RegisterOvnMetrics(clientset kubernetes.Interface, k8sNodeName string, ovsDBClient libovsdbclient.Client,
 	metricsScrapeInterval int, stopChan <-chan struct{}) {
-	go RegisterOvnDBMetrics(clientset, k8sNodeName, stopChan)
+	go RegisterOvnDBMetrics(
```
kyrtapz, line 574: Can you elaborate on why was this change needed? Maybe split it out into a separate commit and add the reasoning there.

go-controller/pkg/util/ovs.go
```diff
@@ -597,7 +594,7 @@ func RunOvsVswitchdAppCtl(args ...string) (string, string, error) {
 
 // GetOvsVSwitchdPID retrieves the Process IDentifier for ovs-vswitchd daemon.
 func GetOvsVSwitchdPID() (string, error) {
-	pid, err := afero.ReadFile(AppFs, savedOVSRunDir+"ovs-vswitchd.pid")
+	pid, err := afero.ReadFile(AppFs, config.OvsPaths.RunDir+"ovs-vswitchd.pid")
```
kyrtapz: While we are here, we should use `filepath.Join` instead of this simple add. Is this something you could help with?
This applies to all file path creation in the PR. If you could do that as a separate commit that would be awesome.

go-controller/pkg/metrics/ovn_db.go
```diff
@@ -323,7 +321,7 @@ var (
 )
 
 func getNBDBSockPath() (string, error) {
-	paths := []string{"/var/run/openvswitch/", "/var/run/ovn/"}
+	paths := []string{config.OvsPaths.RunDir, config.OvnSouth.RunDir}
```
kyrtapz:
```suggestion
	paths := []string{config.OvsPaths.RunDir, config.OvnNorth.RunDir}
```

go-controller/pkg/config/config.go
```diff
@@ -138,10 +138,23 @@ var (
 	}
 
 	// OvnNorth holds northbound OVN database client and server authentication and location details
-	OvnNorth OvnAuthConfig
+	OvnNorth = OvnAuthConfig{
+		RunDir:     "/var/run/ovn/",
+		DbLocation: "/var/lib/openvswitch/ovnnb_db.db",
```
kyrtapz: Why did the DB location change from:
```
OvnNbdbLocation = "/etc/ovn/ovnnb_db.db"
OvnSbdbLocation = "/etc/ovn/ovnsb_db.db"
```

go-controller/pkg/util/ovs.go
```diff
@@ -743,7 +740,7 @@ type queryResult struct {
 }
 
 func GetOVNDBServerInfo(timeout int, direction, database string) (*OVNDBServerStatus, error) {
-	sockPath := fmt.Sprintf("unix:/var/run/openvswitch/ovn%s_db.sock", direction)
+	sockPath := fmt.Sprintf("unix:%sovn%s_db.sock", config.OvsPaths.RunDir, direction)
```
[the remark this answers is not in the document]
kyrtapz: While this is a valid concern this change is in line with the previous value.
This code only seems to be used for non-IC which we deprecated and don't have much coverage for so I would refrain from changing it to `ovnRunDir`.

go-controller/pkg/config/config.go
```diff
@@ -524,10 +537,24 @@ type OvnAuthConfig struct {
...
+	// RunDir is OVN run directory.
+	RunDir string
+	// DbLocation is OVN northbound/southbound database location.
+	DbLocation string
 
 	exec kexec.Interface
 }
 
+// OvsPathConfig holds OVS location details.
+type OvsPathConfig struct {
```
kyrtapz, line 573: missing gcfg tages.
If we add an instance of it to the `config` struct we should be able to modify the values through the config file.

go-controller/pkg/config/config.go
jneo8: I keep the change because:
The buildOvnAuth function creates an empty OvnAuthConfig without initializing RunDir and DbLocation, when neither CLI nor config file provides these values, they remain empty strings instead of using the defaults from savedOvnNorth/savedOvnSouth.

jneo8:
Hi @kyrtapz 

Can I please ask for next round review? Thanks.

go-controller/pkg/metrics/ovn_db.go
```diff
@@ -397,13 +387,13 @@ func RegisterOvnDBMetrics(clientset kubernetes.Interface, k8sNodeName string, st
 		func() float64 { return 1 },
 	))
 	var dbProperties []*util.OvsDbProperties
-	nbdbProps, err := util.GetOvsDbProperties(util.OvnNbdbLocation)
```
kyrtapz, line 400: `OvnNbdbLocation/OvnSbdbLocation` variables should be removed from `github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/ovs.go`

go-controller/pkg/config/config.go
kyrtapz: This should use the config run-dir instead of a hard-coded value.

go-controller/pkg/config/config.go
kyrtapz: Is it still required with the recent changes? I don't see why `overrideFields` wouldn't use the default values if cfg/cli were not provided.

go-controller/pkg/config/config.go
```diff
@@ -138,10 +138,23 @@ var (
...
-	OvnSouth OvnAuthConfig
+	OvnSouth = OvnAuthConfig{
+		RunDir:     "/var/run/ovn/",
+		DbLocation: "/etc/ovn/ovnsb_db.db",
+	}
+
+	// OvsPaths holds OVS location details
+	OvsPaths = OvsPathConfig{
+		RunDir:         "/var/run/openvswitch/",
+		VswitchdPid:    "/var/run/openvswitch/ovs-vswitchd.pid",
```
kyrtapz: Do wee need the two PIDs as separate config options?
These make the following functions invalid:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/52f16477709894c201438890d85a027c4d1fd283/go-controller/pkg/util/ovs.go#L596-L614

go-controller/pkg/config/config.go
jneo8: Yes we need the change, removing this will crush all the e2e tests.

The overrideFields function should handle this automatically:
- If confAuth.RunDir or cliAuth.RunDir is set and different from default -> it gets copied
- If they're empty or equal to default -> they get skipped, keeping the zero value

In this case, both confAuth and cliAuth have empty/default values that match defaultAuth, the fields won't get copied.

go-controller/pkg/metrics/ovn_db.go
jneo8: I am confuse, the OvnNbdbLocation/OvnSbdbLocation are not exist on this PR already.

go-controller/pkg/config/config.go
jneo8: I guess it will not make sense to have these two pid files outside of `RunDir(/var/run/openvswitch)`. How about we change the default to file name only?

```go
OvsPaths = OvsPathConfig{
    RunDir: "/var/run/openvswitch/",
    VswitchdPid: "ovs-vswitchd.pid",
    VswitchdPid: "ovsdb-server.pid",
}
```

kyrtapz: Do you think there is value in making the pid file name configurable? If not lets just have the rundir configurable.

go-controller/pkg/metrics/ovn_db.go
kyrtapz: Sorry, my bad. You removed it here https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5559/files#diff-be5eeb18ce3e11bf80db0aa85ee6b61492f80a2a31dd7a0b8241d17aa894aba3L51-L52
Resolving.

go-controller/pkg/config/config.go
jneo8: Hi @kyrtapz , would like to get some response from you on this one also. Thanks.

go-controller/pkg/config/config.go
jneo8: In my use case it will be equal to default value, so should be fine. But it really depends on if we want to make everything configurable. I can remove it first and optional to have another PR to enable this if anyone want it.

kyrtapz: I am in favor of starting small and adding new config options as needed.

go-controller/pkg/config/config.go
kyrtapz: I am still missing something here, `defaultAuth` is not empty? For OVN it should be equal to:
```
	OvnNorth = OvnAuthConfig{
		RunDir:     "/var/run/ovn/",
		DbLocation: "/etc/ovn/ovnnb_db.db",
	}
```

jneo8: The key is in the `overrideFields` function at lines:

```go
if dv.IsValid() && reflect.DeepEqual(dv.Interface(), srcField.Interface()) {
        continue  // Skip setting if src equals default
}
dstField.Set(srcField)
```

This logic says: "If the source value equals the default value, skip updating the destination". This assumes the destination (dst) is already pre-initialized with the correct default values before calling overrideFields.

Trace through the execution, without give default value to auth object:

1. auth is created with zero values: auth.RunDir = ""
2. confAuth.RunDir = "/var/run/ovn/" (from cfg.OvnNorth initialized at line 2417)
3. defaultAuth.RunDir = "/var/run/ovn/" (from savedOvnNorth)
4. In overrideFields(auth, confAuth, defaultAuth):
  - Check: confAuth.RunDir == defaultAuth.RunDir? YES (both "/var/run/ovn/")
  - Action: continue (skip setting)
  - Result: auth.RunDir remains "" -> cause the error

kyrtapz: Thanks for the explanation!
Maybe it would be better to copy the whole defaultauth instead of just some fields but I think either one will do.

kyrtapz (approved):
LGTM
Thanks @jneo8 🎉

jneo8:
rebased.

go-controller/pkg/metrics/metrics.go
```diff
@@ -571,7 +571,37 @@ func startMetricsServer(bindAddress, certFile, keyFile string, handler http.Hand
...
+			return true
+		},
+		stopChan,
+	)
 	go RegisterOvnControllerMetrics(ovsDBClient, metricsScrapeInterval, stopChan)
-	go RegisterOvnNorthdMetrics(clientset, k8sNodeName, stopChan)
+	go RegisterOvnNorthdMetrics(
+		func() bool {
+			err := utilwait.PollUntilContextTimeout(context.Background(), 1*time.Second, 300*time.Second, true, func(_ context.Context) (bool, error) {
+				return checkPodRunsOnGivenNode(clientset, []string{"app=ovnkube-master", "name=ovnkube-master"}, k8sNodeName, true)
```
trozet: you changed "ovn-db-pod=true" -> "app=ovnkube-master"

was that intentional? In Interconnect mode (IC) we do not have ovnkube-master, the ovnkube-node pod has these labels:

  labels:
    app: ovnkube-node
    component: network
    controller-revision-hash: 66577b8948
    kubernetes.io/os: linux
    name: ovnkube-node
    ovn-db-pod: "true"
    pod-template-generation: "1"
    type: infra


for central we have:

  labels:
    component: network
    kubernetes.io/os: linux
    name: ovnkube-master
    pod-template-hash: 79d947c57c
    type: infra

jneo8: Must be a mistake, will update it.

trozet:
thanks @jneo8 !

jneo8:
Hi @kyrtapz & @trozet , thanks for your kindly support on the review.
I guess we are good to merge this PR now?
Please let me know if any blocker to merge this one.

PR #5565, opened 2025-09-05
Fixes for flaky lb service test

pperiyasamy:
This PR contains required changes to fix flakiness with LB service E2E tests.

## Summary by CodeRabbit

* **Tests**
  * Switched end-to-end service readiness checks to use EndpointSlices instead of Endpoints, aligning with newer Kubernetes discovery APIs.
  * Improved readiness accuracy by counting only unique, ready, non-terminating endpoint addresses across slices and deduplicating duplicates.
  * Added a reusable polling helper with interval/timeout controls to wait for an expected number of ready endpoints.
  * Simplified an external-container test utility by removing an unused timing parameter.

### DIFF
--- a/test/e2e/service.go
+++ b/test/e2e/service.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/test/e2e/ipalloc"
 
 	v1 "k8s.io/api/core/v1"
+	discoveryv1 "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -33,7 +34,6 @@ import (
 	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
-	e2eoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
 	e2epodoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
 	e2erc "k8s.io/kubernetes/test/e2e/framework/rc"
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
@@ -200,7 +200,7 @@ var _ = ginkgo.Describe("Services", feature.Service, func() {
 
 		ensureStickySession := func() string {
 			hosts := getServiceBackendsFromPod(execPod, svc.Spec.ClusterIP, int(svc.Spec.Ports[0].Port))
-			uniqHosts := sets.New[string](hosts...)
+			uniqHosts := sets.New(hosts...)
 			gomega.Expect(uniqHosts.Len()).To(gomega.Equal(1), fmt.Sprintf("expected the same backend for every connection with session-affinity set, got %v", uniqHosts))
 			backendPod, _ := uniqHosts.PopAny()
 			return backendPod
@@ -1411,7 +1411,7 @@ func getServiceBackendsFromPod(execPod *v1.Pod, serviceIP string, servicePort in
 	curl := fmt.Sprintf(`curl -q -s --connect-timeout 2 http://%s/`, serviceIPPort)
 	cmd := fmt.Sprintf("for i in $(seq 1 %d); do echo; %s ; done", connectionAttempts, curl)
 
-	stdout, err := e2eoutput.RunHostCmd(execPod.Namespace, execPod.Name, cmd)
+	stdout, err := e2epodoutput.RunHostCmd(execPod.Namespace, execPod.Name, cmd)
 	if err != nil {
 		framework.Logf("Failed to get response from %s. Retry until timeout", serviceIPPort)
 		return nil
@@ -1686,7 +1686,7 @@ metadata:
 	})
 
 	ginkgo.It("Should ensure connectivity works on an external service when mtu changes in intermediate node", func() {
-		err := framework.WaitForServiceEndpointsNum(context.TODO(), f.ClientSet, namespaceName, svcName, 4, time.Second, time.Second*180)
+		err := WaitForServingAndReadyServiceEndpointsNum(context.TODO(), f.ClientSet, namespaceName, svcName, 4, time.Second, time.Second*180)
 		framework.ExpectNoError(err, fmt.Sprintf("service: %s never had an endpoint, err: %v", svcName, err))
 
 		time.Sleep(time.Second * 5) // buffer to ensure all rules are created correctly
@@ -1727,11 +1727,9 @@ metadata:
 		// Ensure service connectivity works from external client with default settings.
 		// Use Eventually because IPv6 takes a while to finish its network configuration
 		// with network namespaces.
-		// TODO: Figure out why keeping this at 5seconds is causing CI flakes after K8s 1.33 rebase
-		// See: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5455
 		gomega.Eventually(func() error {
 			return buildAndRunCommand(fmt.Sprintf("sudo ip netns exec client curl %s:%d/big.iso -o big.iso", svcIPforCurl, endpointHTTPPort))
-		}, 10*time.Second).Should(gomega.BeNil(), "failed to connect with external load balancer service")
+		}, 5*time.Second).Should(gomega.BeNil(), "failed to connect with external load balancer service")
 
 		// Change MTU size of vmtobridge veth pair and verify service connectivity still works.
 		// Set the value >=1280 so that it works for IPv6 as well.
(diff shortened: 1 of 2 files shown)


jcaamano:
@pperiyasamy a couple of questions:

In regards to the first commit, isn't existing retries on the traffic test itself enough to cover up for the time pods take to become ready? Are we hitting unexpected delays there? Or maybe the traffic test itself did not have retries?

In regards to the second commit, I don't understand it very well. We are using BGP so we should not be adding routes manually. Why would we need a route now where we did not need it before?

pperiyasamy:
> 
> In regards to the first commit, isn't existing retries on the traffic test itself enough to cover up for the time pods take to become ready? Are we hitting unexpected delays there? Or maybe the traffic test itself did not have retries?

yes @jcaamano, retry wasn't enough at least for `Should ensure connectivity works on an external service when mtu changes in intermediate node` test and moreover `framework.WaitForServiceEndpointsNum` is not actually waiting until endpoints are ready and serving, instead it returns success as soon as service has reached expected endpoints (only with checking its ip addresses). so had to change it to waiting for endpoints to be ready and serving.

> 
> In regards to the second commit, I don't understand it very well. We are using BGP so we should not be adding routes manually. Why would we need a route now where we did not need it before?

This route is added on the host and to fix connectivity issue on my local cluster, it may not be needed on the CI and default route might steer the traffic to the node there. BGP would adding service routes only on frr container, isn't it ?

pperiyasamy:
> > In regards to the second commit, I don't understand it very well. We are using BGP so we should not be adding routes manually. Why would we need a route now where we did not need it before?
> 
> This route is added on the host and to fix connectivity issue on my local cluster, it may not be needed on the CI and default route might steer the traffic to the node there. BGP would adding service routes only on frr container, isn't it ?

checked this again, yes, you are right, we don't need explicit route for `LbSvCIP` and connectivity is established after `LbSvcIP` is advertised, so this connectivity towards `lbclient (clientnet) -> frr > kind`  service works, but reply traffic is dropped on the host after reaching kind network gateway and not getting into lbclient. but after adding this route, it started working.

jcaamano:
> checked this again, yes, you are right, we don't need explicit route for `LbSvCIP` and connectivity is established after `LbSvcIP` is advertised, so this connectivity towards `lbclient (clientnet) -> frr > kind` service works, but reply traffic is dropped on the host after reaching kind network gateway and not getting into lbclient. but after adding this route, it started working.

But that doesn't look like the correct solution or addressing the core of the problem, right? No routing should happen on the host, routing should happen on the FRR container.

I think we need to understand exactly why the packet is dropped. And if routing is happening on the host, why?

Is it me misunderstanding the test? 

This makes me think that FRR does not have a route for the VIP with the nodes as next hop, and instead we might be relying on a default route pointing to the host.

pperiyasamy:
> But that doesn't look like the correct solution or addressing the core of the problem, right? No routing should happen on the host, routing should happen on the FRR container.
> 
> I think we need to understand exactly why the packet is dropped. And if routing is happening on the host, why?
> 
> Is it me misunderstanding the test?
> 
> This makes me think that FRR does not have a route for the VIP with the nodes as next hop, and instead we might be relying on a default route pointing to the host.

yes @jcaamano , frr should do the routing for incoming packet towards the service and reply packet from the service should just land from host to lbclient via asymmetric path. as you suggested, tried to troubleshoot it with retis, can see syn is sent to lb service ip, syn-ack from service is seen on kind bridge interface, but not reaching to lbclient.

```
19661309887380 (4) [ksoftirqd/4] 45 [tp] skb:kfree_skb #11e1c163b5bcffffa03a460dd900 (skb ffffa03cdc7e9800) drop (reason NETFILTER_DROP)
  192.168.10.0.80 > 172.22.0.3.23456 tos 0x0 ttl 62 id 0 off 0 [DF] len 60 proto TCP (6) flags [S.] seq 3808199748 ack 353483120 win 64704 [mss 1360,sackOK,TS val 4064829574 ecr 296873058,nop,wscale 7]
  ns 0x1/4026531840 if 8 (br-cc1f97e27dc9) rxif 8

^CReceived signal, terminating...
369 event(s) processed
root@vbox:~/retis# sudo tcpdump -ni br-cc1f97e27dc9 tcp port 23456
dropped privs to tcpdump
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on br-cc1f97e27dc9, link-type EN10MB (Ethernet), snapshot length 262144 bytes
14:17:17.767480 IP 192.168.10.0.http > 172.22.0.3.aequus: Flags [S.], seq 3808199748, ack 353483120, win 64704, options [mss 1360,sackOK,TS val 4064844102 ecr 296873058,nop,wscale 7], length 0
14:17:17.767529 IP 172.22.0.3.aequus > 192.168.10.0.http: Flags [S], seq 353483119, win 64240, options [mss 1460,sackOK,TS val 296892711 ecr 0,nop,wscale 7], length 0
14:17:17.767579 IP 192.168.10.0.http > 172.22.0.3.aequus: Flags [S.], seq 3808199748, ack 353483120, win 64704, options [mss 1360,sackOK,TS val 4064844102 ecr 296873058,nop,wscale 7], length 0
^C
3 packets captured
3 packets received by filter
0 packets dropped by kernel

root@vbox:~/retis# sudo conntrack -L | grep 23456
conntrack v1.4.7 (conntrack-tools): 8 flow entries have been shown.
```
Here are the iptable entries present on the node.

```
# iptables -S
-P INPUT ACCEPT
-P FORWARD DROP
-P OUTPUT ACCEPT
-N DOCKER
-N DOCKER-BRIDGE
-N DOCKER-CT
-N DOCKER-FORWARD
-N DOCKER-ISOLATION-STAGE-1
-N DOCKER-ISOLATION-STAGE-2
-N DOCKER-USER
-A FORWARD -d 172.22.0.3/32 -i br-cc1f97e27dc9 -o br-cc1f97e27dc9 -p tcp -m tcp --dport 23456 -j ACCEPT
-A FORWARD -j DOCKER-USER
-A FORWARD -j DOCKER-FORWARD
-A DOCKER -d 172.22.0.3/32 -p tcp -m tcp --dport 23456 -j ACCEPT
-A DOCKER -d 172.18.0.3/32 ! -i br-cc1f97e27dc9 -o br-cc1f97e27dc9 -p tcp -m tcp --dport 6443 -j ACCEPT
-A DOCKER ! -i docker0 -o docker0 -j DROP
-A DOCKER ! -i br-a40680211fe9 -o br-a40680211fe9 -j DROP
-A DOCKER-BRIDGE -o docker0 -j DOCKER
-A DOCKER-BRIDGE -o br-cc1f97e27dc9 -j DOCKER
-A DOCKER-BRIDGE -o br-a40680211fe9 -j DOCKER
-A DOCKER-CT -o docker0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A DOCKER-CT -o br-cc1f97e27dc9 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A DOCKER-CT -o br-a40680211fe9 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A DOCKER-FORWARD -j DOCKER-CT
-A DOCKER-FORWARD -j DOCKER-ISOLATION-STAGE-1
-A DOCKER-FORWARD -j DOCKER-BRIDGE
-A DOCKER-FORWARD -i docker0 -j ACCEPT
-A DOCKER-FORWARD -i br-cc1f97e27dc9 -j ACCEPT
-A DOCKER-FORWARD -i br-a40680211fe9 -j ACCEPT
-A DOCKER-ISOLATION-STAGE-1 -i docker0 ! -o docker0 -j DOCKER-ISOLATION-STAGE-2
-A DOCKER-ISOLATION-STAGE-1 -i br-cc1f97e27dc9 ! -o br-cc1f97e27dc9 -j DOCKER-ISOLATION-STAGE-2
-A DOCKER-ISOLATION-STAGE-1 -i br-a40680211fe9 ! -o br-a40680211fe9 -j DOCKER-ISOLATION-STAGE-2
-A DOCKER-ISOLATION-STAGE-2 -o br-a40680211fe9 -j DROP
-A DOCKER-ISOLATION-STAGE-2 -o br-cc1f97e27dc9 -j DROP
-A DOCKER-ISOLATION-STAGE-2 -o docker0 -j DROP
```
Tried deleting `DOCKER-ISOLATION-STAGE-1` and `DOCKER-ISOLATION-STAGE-2` rules didn't help.

node is with below configuration:

```
# sysctl net.ipv4.conf.all.rp_filter
net.ipv4.conf.all.rp_filter = 0
# sysctl net.bridge.bridge-nf-call-iptables
net.bridge.bridge-nf-call-iptables = 0
```

This setup is on a fedora VM running inside a Virtual Box, so I think this issue may be isolated only to my setup, in that case i can just remove the 2nd commit and go head with only 1st commit.

pperiyasamy:
> Tried deleting `DOCKER-ISOLATION-STAGE-1` and `DOCKER-ISOLATION-STAGE-2` rules didn't help.
> 
> node is with below configuration:
> 
> ```
> # sysctl net.ipv4.conf.all.rp_filter
> net.ipv4.conf.all.rp_filter = 0
> # sysctl net.bridge.bridge-nf-call-iptables
> net.bridge.bridge-nf-call-iptables = 0
> ```
> 
> This setup is on a fedora VM running inside a Virtual Box, so I think this issue may be isolated only to my setup, in that case i can just remove the 2nd commit and go head with only 1st commit.

We figured it's firewalld's nftable rules dropping the reply traffic,.
```
table inet firewalld {
...
table ip raw {
	chain PREROUTING {
		type filter hook prerouting priority raw; policy accept;
		ip daddr 172.18.0.2 iifname != "br-57ac0835adfd" counter packets 0 bytes 0 drop
		ip daddr 172.18.0.3 iifname != "br-57ac0835adfd" counter packets 0 bytes 0 drop
		ip daddr 172.18.0.4 iifname != "br-57ac0835adfd" counter packets 0 bytes 0 drop
		ip daddr 127.0.0.1 iifname != "lo" tcp dport 46765 counter packets 0 bytes 0 drop
		ip daddr 172.18.0.5 iifname != "br-57ac0835adfd" counter packets 0 bytes 0 drop
		ip daddr 172.22.0.2 iifname != "br-28add1607bee" counter packets 0 bytes 0 drop
		ip daddr 172.22.0.3 iifname != "br-28add1607bee" counter packets 705 bytes 42124 drop
	}
}
...
}
```
after stopping firewalld service and flushing these rules, traffic recovers, so this is specific to my setup, hence dropping 2nd commit and will do few runs to check how lb test behaves.
thanks @jcaamano for helping to troubleshoot this.

jcaamano (approved):

PR #5573, opened 2025-09-09
Node controller shutdown: do not ref gateway if not set

martinkennelly:
PR 5373 to drop the GARP flows didnt consider that we set the default network controller and later we set the gateway obj. In-between this period, ovnkube node may receive a stop signal and we do not guard against accessing the gateway if its not yet set.

There is nothing to reconcile if the gateway is not set.

This code will be removed when ovn has native support for silencing garps on shutdown next march.


martinkennelly:
@crnithya ptal

trozet (approved):

PR #5575, opened 2025-09-10
Brings factory handling inline with level driven controllers

trozet:
Our watchFactory addHandler would take informer lock during sync. However, real level driven controllers do not. In level driven controllers the process is:
1. lock informer
2. handler is added to informer
3. initial add list is queued
4. informer is unlocked
5. informer now continues to enqueue events
6. handler re-lists existing objects, runs sync method
7. handler now starts workers to process enqueued events

Our watch factory today:
1. lock informer
2. informer still enqueing events, but workers are paused
3. run sync
4. initial adds are processed
5. handler added to informer
6. informer unlocked
7. workers now resume

The core difference here is in our watch factory, we share workers (queuemap) across all handlers/controllers, where as level driven creates worker threads specific to its handler/controller. Therefore the level-driven controller can afford to continously allow events to be enqueued by the informer, and eventually starting workers will start processing them. In our case, we pause workers, which stops handling for all controllers while we sync+process initial add for the handler. This is really bad for performance.

This patch modifies watch factory behavior to act the same way that level-driven controllers do. Since we are bound to sharing workers across all controllers, a new mechanism is used to avoid pausing workers when handlers are added. It works like this:

1. lock informer
2. informer still enqueuing events, but workers are paused
3. get initial add list
4. initial add now enqueues items to the handler's buffer
5. handler is added to informer, but in a disabled state
6. informer is unlocked, workers resume, but this handler is disabled, so events for this handler are enqueued to the buffer
7. handler re-lists existing objects, runs sync method
8. handler is now enabled, which causes handler to process all events in the buffer, while not blocking informer

This allows watch factory to do its sync and initial add events without any disruption to other handlers that are using the same informers and workers.

## Summary by CodeRabbit

- New Features
  - Added buffered event handling with explicit enable step to ensure smooth activation of new handlers.
  - Introduced informative logging when event handlers are added.

- Bug Fixes
  - Prevented missed events during startup and restarts by buffering and then enabling handlers after syncing existing items.
  - Reduced race conditions and improved error handling with retries and safe rollback on failures.

- Refactor
  - Centralized event dispatch logic and streamlined handler lifecycle for more reliable and predictable behavior.

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -1287,20 +1287,39 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 	intInf := inf.internalInformers[wf.internalInformerIndex]
 
 	intInf.Lock()
-	defer intInf.Unlock()
 
 	// we are going to add a handler, we need to update the atomic signal that handlers exist now
 	// so that we do not miss events after we list current items.
 	// We need to do this after we get internal informer lock, to preserve that we can be the only one updating
 	// the atomic and preserve known state of the atomic while the handler is going to be added in the future
-	hadZeroHandlers := atomic.CompareAndSwapUint32(&intInf.hasHandlers, hasNoHandler, hasHandler)
+	atomic.CompareAndSwapUint32(&intInf.hasHandlers, hasNoHandler, hasHandler)
 
-	items := make([]interface{}, 0)
+	// get list of items to use for initial add
+	infItems := inf.inf.GetStore().List()
+	items := make([]interface{}, 0, len(infItems))
+	for _, obj := range infItems {
+		if filterFunc(obj) {
+			items = append(items, obj)
+		}
+	}
+
+	// add handler with initial add, keeping informer locked
+	handlerID := atomic.AddUint64(&wf.handlerCounter.counter, 1)
+	handler := inf.addHandler(wf.internalInformerIndex, handlerID, priority, filterFunc, funcs, items)
+	klog.V(5).Infof("Added %v event handler %d", objType, handler.id)
+
+	// unlock informer, the workers (queuemap goroutines) can now resume processing. The handler we added
+	// is disabled, so it will be buffering events for now
+	intInf.Unlock()
+
+	// re-list items during informer lock to get accurate information to sync with
+	items = make([]interface{}, 0, len(items))
 	for _, obj := range inf.inf.GetStore().List() {
 		if filterFunc(obj) {
 			items = append(items, obj)
 		}
 	}
+
 	if processExisting != nil {
 		// Process existing items as a set so the caller can clean up
 		// after a restart or whatever. We will wrap it with retries to ensure it succeeds.
@@ -1313,17 +1332,14 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 			return true, nil
 		})
 		if err != nil {
-			// handler is not going to be added, restore previous value if needed
-			if hadZeroHandlers {
-				atomic.StoreUint32(&intInf.hasHandlers, hasNoHandler)
-			}
+			wf.removeHandler(objType, handler)
 			return nil, err
 		}
 	}
 
-	handlerID := atomic.AddUint64(&wf.handlerCounter.counter, 1)
-	handler := inf.addHandler(wf.internalInformerIndex, handlerID, priority, filterFunc, funcs, items)
-	klog.V(5).Infof("Added %v event handler %d", objType, handler.id)
+	// sync is done, turn on the handler
+	handler.Enable()
+
 	return handler, nil
 }
 
--- a/go-controller/pkg/factory/handler.go
+++ b/go-controller/pkg/factory/handler.go
@@ -56,6 +56,13 @@ type Handler struct {
 	// indicates which informer.internalInformers index to use
 	// clients are distributed between internal informers
 	internalInformerIndex int
+
+	// enabled if this handler is ready to process events
+	enabled atomic.Bool
+	// buffer is used to store events that cannot be processed yet since handler is disabled
+	buffer []func()
+	// mu is used to protect buffer access
+	mu sync.Mutex
 }
 
 func (h *Handler) OnAdd(obj interface{}, isInInitialList bool) {
@@ -82,6 +89,55 @@ func (h *Handler) FilterFunc(obj interface{}) bool {
 
 func (h *Handler) kill() bool {
 	return atomic.CompareAndSwapUint32(&h.tombstone, handlerAlive, handlerDead)
+}
+
+// handleEvent checks if this handler is enabled, if not it buffers the event for processing later.
+func (h *Handler) handleEvent(processEvent func()) {
+	if atomic.LoadUint32(&h.tombstone) != handlerAlive {
+		// handler is dead
+		return
+	}
+	// Fast path: handler is enabled, no buffering
+	if h.enabled.Load() {
+		processEvent()
+		return
+	}
+	// Slow path: still disabled, might need to buffer
+	h.mu.Lock()
+	if !h.enabled.Load() {
+		h.buffer = append(h.buffer, processEvent)
+		h.mu.Unlock()
+		return
+	}
+	h.mu.Unlock()
+	// Lost race: handler became enabled while we waited for the lock
+	processEvent()
+}
+
+// Enable the handler, process all buffered events
+func (h *Handler) Enable() {
+	for {
+		if atomic.LoadUint32(&h.tombstone) != handlerAlive {
+			// handler is dead
+			return
+		}
+		var pending []func()
+		h.mu.Lock()
+		if len(h.buffer) == 0 {
+			if atomic.LoadUint32(&h.tombstone) == handlerAlive {
+				h.enabled.Store(true)
+			}
+			h.mu.Unlock()
+			return
+		}
+		pending = h.buffer
+		h.buffer = nil
+		h.mu.Unlock()
+
+		for _, processEvent := range pending {
+			processEvent()
+		}
+	}
 }
 
 type event struct {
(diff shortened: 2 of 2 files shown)


trozet:
@tssurya @jcaamano @kyrtapz fyi

go-controller/pkg/factory/handler.go
```diff
@@ -546,26 +604,13 @@ func newQueuedInformer(queueSize uint32, oType reflect.Type, sharedInformer cach
...
+			objCopy := obj
+			h.buffer = append(h.buffer, func() {
+				h.OnAdd(objCopy, false)
 			})
 		}
-
-		// Wait until all the object additions have been processed
-		addsMap.shutdown()
-		addsWg.Wait()
 	}
```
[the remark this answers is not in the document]
trozet: @coderabbitai isInIniitialList is when the informer does its initial sync against api server. Here we are emulating initial adds long after the informer was already started. Therefore we should pass false, which is the same as the previous code did.

go-controller/pkg/factory/factory.go
```diff
@@ -1313,17 +1332,14 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
...
+			wf.removeHandler(objType, handler)
 			return nil, err
 		}
 	}
 
-	handlerID := atomic.AddUint64(&wf.handlerCounter.counter, 1)
-	handler := inf.addHandler(wf.internalInformerIndex, handlerID, priority, filterFunc, funcs, items)
-	klog.V(5).Infof("Added %v event handler %d", objType, handler.id)
+	// sync is done, turn on the handler
+	handler.Enable()
```
trozet, line 1341: note, eventually i want to run this in a goroutine so adding handlers is non blocking and UDN controllers start faster. I tried doing this but it ended in test failures, probably due to the fact that we rely on some ordering between resource types like node/namespaces/pods during start up. I'll look into it as a follow up to this PR.

trozet: testing with goroutine revealed this race bug in EIP:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5577

Will keep testing and looking for other bugs, but not sure if this should block this PR or could be handled later.

go-controller/pkg/factory/handler.go
```diff
@@ -546,26 +604,13 @@ func newQueuedInformer(queueSize uint32, oType reflect.Type, sharedInformer cach
...
-		addsMap := newQueueMap(queueSize, numEventQueues, addsWg, stopChan)
-		addsMap.start()
-
-		// Distribute the existing items into the handler-specific
-		// channel array.
+		// buffer initial adds, they will execute when the handler is enabled
 		for _, obj := range items {
-			addsMap.enqueueEvent(nil, obj, informer.oType, false, func(e *event) {
-				h.OnAdd(e.obj, false)
+			objCopy := obj
```
trozet, line 609: note this is not a real copy of the object, its a copy of a pointer. I could update this to make a real copy of the object, I'm not sure if it is necessary or not. I don't think we pass a copy of object in the retry framework add handling, and we depend on the controllers to not update the informer object.

go-controller/pkg/factory/handler.go
```diff
@@ -546,26 +604,13 @@ func newQueuedInformer(queueSize uint32, oType reflect.Type, sharedInformer cach
...
-		// is added, only that handler should receive events for all
-		// existing objects.
-		addsWg := &sync.WaitGroup{}
-
-		addsMap := newQueueMap(queueSize, numEventQueues, addsWg, stopChan)
-		addsMap.start()
-
-		// Distribute the existing items into the handler-specific
-		// channel array.
+		// buffer initial adds, they will execute when the handler is enabled
```
npinaeva, line 607: I think `initialAddFunc` doesn't have to be dynamically created anymore, you can just copy this code to the place where it is called

go-controller/pkg/factory/factory.go
```diff
@@ -1287,20 +1287,39 @@ func (wf *WatchFactory) addHandler(objType reflect.Type, namespace string, sel l
 	intInf := inf.internalInformers[wf.internalInformerIndex]
 
 	intInf.Lock()
```
npinaeva, line 1289: factory shouldn't touch internal fields of informer, so a part of this function (at least till intInf.Unlock) probably should be moved to `informer.addHandler()`

go-controller/pkg/factory/handler.go
```diff
@@ -165,11 +221,16 @@ func (i *informer) addHandler(internalInformerIndex int, id uint64, priority int
 		handlerAlive,
 		priority,
 		internalInformerIndex,
+		atomic.Bool{},
+		make([]func(), 0, len(existingItems)+200),
```
npinaeva, line 225: 200 should be a constant, also it would be nice to explain that optimization only works for the first 200 events, so if you get more than that same full lock will happen

go-controller/pkg/factory/handler.go
```diff
@@ -84,6 +91,55 @@ func (h *Handler) kill() bool {
...
+		return
+	}
+	h.mu.Unlock()
+	// Lost race: handler became enabled while we waited for the lock
+	processEvent()
+}
+
+// Enable the handler, process all buffered events
+func (h *Handler) Enable() {
+	for {
```
npinaeva, line 119: this function could potentially never return if the events keep coming?
you could call it as a goroutine, but that is still not ideal

go-controller/pkg/factory/handler.go
```diff
@@ -84,6 +91,55 @@ func (h *Handler) kill() bool {
...
+	}
+	// Fast path: handler is enabled, no buffering
+	if h.enabled.Load() {
+		processEvent()
+		return
+	}
+	// Slow path: still disabled, might need to buffer
+	h.mu.Lock()
+	if !h.enabled.Load() {
+		h.buffer = append(h.buffer, processEvent)
```
npinaeva, line 108: I think this breaks `forEachQueuedHandler` protocol: it is supposed to sequentially run handlers based on their priority, but now if you have disabled handlers they will be skipped, and then they will handle the same event later, so the priorities of handlers don't work anymore. Especially considering the handler could forever (or for a long time) stay disabled?

npinaeva: Another part of this is adding handlers for the same object type but with different priorities: e.g. we add podHandler first, then we start netpol handlers watching pod labels. Before this change you know that when the `AddPodHandler` returns, the handler is added (and enabled) and when you call `AddNetpolPodHandler` after that, you know that every event will be handled by the PodHandler first, and only then passed to the netpol handler. This contract also won't work anymore, and I can imagine it being a problem.

go-controller/pkg/factory/factory.go
trozet: After putting some more thought into this, I realized blocking on Enable is unavoidable. Even in the level driven controller, we block on initial sync. This is important because other handlers that are added later rely on the cache being hot. For example, namespace handler might rely on node handler being synced. Therefore an alternative solution to this problem is to simply increase the number of workers in network controller. This will allow multiple network controllers and add handlers to be added at the same time, without blocking each other. I'll do some testing with this.

go-controller/pkg/factory/factory.go
trozet: I didn't change this code

go-controller/pkg/factory/handler.go
trozet: For your first point, the answer is technically yes. But keep in mind, the current way we addHandler is:
1. lock + pause the workers, informers are still feeding events
2. list all items
3. run sync function
4. run all initial adds
5. unpause workers

That means during this time a new add can come, that is sent to be processed as "forEachQueuedHandler", but that worker is paused. Meanwhile the new add is processed by the addHandler function, without any regard for priority of other handlers. In other words, it will be processed out of order. When step 5 finishes, sure the object may be reprocessed as an add again, and actually follow the priority of forEachQueuedHandler. I'm not sure if this saves us from any bugs or not.

The issue with my commit is it will keep deferring new objects to the buffer until the new handler catches up. So there could be more events processed out of priority order. I'm not sure if it matters or not.

For the second comment, it still works that way. handler.Enable() is blocking, so it will not return from adding the handler until it finishes the sync and catches up with finish off all the buffered objects. So if you AddPodHandler, it will sync+initial items, then return, now you AddNetpolPodHandler.

go-controller/pkg/factory/handler.go
trozet: Right. The hypothesis is that it eventually converges, because the time it takes to process events for one handler should be < than the time it takes to process multiple handlers OR the rate of events to the handler < the rate of processing events. 

I can't call it as a goroutine, due to this:

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5575/files#r2370298563

go-controller/pkg/factory/handler.go
trozet: yeah I just randomly chose 200, I can make it a constant, and maybe expose a metric and we can come up with a better value here.

go-controller/pkg/factory/handler.go
npinaeva: currently the new add events that will be handled out of order are the events that come between steps 1 and 2 plus whatever was already in the queue, AND as you said the same event will be handled with the right handlers priority later on.
This change makes both of these points worse, so I am concerned about the problems/races it can bring up. This change looks equivalently dangerous to just getting rid of the priorities handling, so that could be another solution to consider.
Also the second comment was based on your comment about running addHandler in a goroutine, which I thought was the plan after you fix egressIP bug?

npinaeva: we only have 2 features that require prioritized handling: egressIP and netpol, I could work on removing the netpol dependency, and if we can find help for egressIP, the world would be a better place :)

trozet: thanks for identifying that. Yes I think it would be good anyway to decouple those dependencies. I'll work on egress IP. I'll put a hold on this PR until we fix that first.

trozet:
we decided to just move everything into level driven and get rid of the old framework

PR #5576, opened 2025-09-15
Link manager: collect errors and return when syncing

martinkennelly:
syncLink fn got an error return a year ago and
I think its correct to collect the errors while
processing and return them instead of logging.
This allows callers to handle the errors also.

Also, we would emit a log of success even if
adding an address failed. Stop that.

Continue only logging GARPs.

Renamed addressWanted to wantedAddress because
this follows the common adjective-noun pattern.


martinkennelly:
No bug reported ds - but code cleanup after reviewing issue recently.

PR #5577, opened 2025-09-15
Enhances EIP status validation

trozet:
Adds check to ensure that the status on an EIP matches the IP we have allocated in our cache. If it doesn't consider the status invalid and attempt reallocation.

There is a race exposed by the unit test "should update invalid assignments on duplicated node", where an EIP has 2 IPs assigned to node1, and a bogus IP assigned to node2. The test expects that EIP should rebalance the EIPs with one on node 1 and one on node 2. However a race can happen where EIP node allocator cache becomes corrupt due to lack of validation of status. This can happen:

1. EgressIP has node1: EIP1, EIP2
2. Test starts WatchEgressIP
3. EIP->Sync->Annotates 50000 mark on EIP
4. This triggers an update event with 50000 mark, node1: EIP1, EIP2
5. EIP initial add starts with original event of node1: EIP1, EIP2
6. EIP rebalances one EIP per node, patches EIP object, creates update event with 50000 mark, node1: EIP1, node2: EIP2
7. Next update event processed -> RACE happens, retry framework will grab "latest" object. If the informer was lagging, informer cache has object from step 4
8. EIP validation logic considers this a pass, because cache has the both nodes, each with 1 egress IP. No patch is done to the EIP object, but when validStatus is achieved the eNode cache is updated
9. enode cache is updated so now node1 has 2 allocations for EIP
10. Update event from 6 is processed. Validation state fails because node1 has 2 EIPs. During rebalancing it wont be able to find a suitable node for the EIP due to cache corruption, so nothing will be updated.


go-controller/pkg/clustermanager/egressip_controller.go
```diff
@@ -865,6 +874,7 @@ func (eIPC *egressIPClusterController) addAllocatorEgressIPAssignments(name stri
 	defer eIPC.nodeAllocator.Unlock()
 	for _, status := range statusAssignments {
 		if eNode, exists := eIPC.nodeAllocator.cache[status.Node]; exists {
+			klog.V(5).Infof("Setting egress IP node allocation - node: %s, IP: %s, EIP name: %s", eNode.name, name, status.EgressIP)
```
martinkennelly: your last two logs args are wrong

martinkennelly (commented):
I thought we bundled the mark update + status update into a single patch update for the eip obj. Is that not the case?

trozet:
> I thought we bundled the mark update + status update into a single patch update for the eip obj. Is that not the case?

@martinkennelly we do, but not the initial sync. The initial sync only updates the mark.

martinkennelly:
Will review shortly. Give me two days.

martinkennelly (approved):
/lgtm 

Nice tim! I had to draw out the scenario as its so complex. This fixes a unit test issue but we should get into this scenario on a production cluster unless theres a programming error.

PR #5583, opened 2025-09-19
OCPBUGS-59552: Referencing pod named ports within a service results in bad DNAT rules containing tcp/0 target port

andreaskaris:
With: `externalTrafficPolicy: Local`,  `allocateLoadBalancerNodePorts: false`

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/0a11e5c9f625425dd54be53a3d96f77b81112910/go-controller/pkg/node/gateway_iptables.go#L204 tries to parse named target ports as integer values, yielding 0. Unfortunately, the existing code throws away the error from strconv.Atoi which is the reason why this hasn't been discovered, yet.

The issue can be reproduced with upstream kind via:
```
./contrib/kind.sh -i6 --install-metallb -gm local
```

```
# cat deployment.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: my-dep
  name: my-dep
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 3
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: my-dep
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: my-dep
    spec:
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: nginx
        ports:
        - containerPort: 80
          name: diameterstack
          protocol: TCP
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
```

```
# cat svc.yaml 
apiVersion: v1
kind: Service
metadata:
  labels:
    app: my-dep
  name: my-dep
  namespace: default
spec:
  allocateLoadBalancerNodePorts: false
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: RequireDualStack
  externalTrafficPolicy: Local
  ports:
  - name: diameterstack
    port: 8080
    protocol: TCP
    targetPort: diameterstack
  selector:
    app: my-dep
  type: LoadBalancer
```

Given that the target port `diameterstack` cannot be converted to int via IntValue():
```
// IntValue returns the IntVal if type Int, or if
// it is a String, will attempt a conversion to int,
// returning 0 if a parsing error occurs.
func (intstr *IntOrString) IntValue() int {
	if intstr.Type == String {
		i, _ := strconv.Atoi(intstr.StrVal)
		return i
	}
	return int(intstr.IntVal)
}
```

This will yield something like:
```
#  docker exec -it ovn-worker /bin/bash
# iptables -t nat -L OVN-KUBE-ETP -nv
Chain OVN-KUBE-ETP (1 references)
 pkts bytes target     prot opt in     out     source               destination         
    2   120 DNAT       6    --  *      *       0.0.0.0/0            192.168.10.0         tcp dpt:8080 statistic mode random probability 0.25000000000 to:10.244.1.3:0    # <---- port is 0
```

### ovnkubernetes assumes that named ports can only have one target port

ovnkubernetes maps serviceport -> slice port -> target port all one to one. however, with named ports, a single service port can match multiple pods with the same targetPort name, but different target ports. Should be taken care of in a follow-up.

### br-ex bad flows in gateway mode shared

The very same parsing issue is present for shared GM:
```
./contrib/kind.sh -i6 --install-metallb -gm shared
```

host networked pods with nodeport service:
```
[root@ovnkubernetes main-host-networked]# cat deployment.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: my-dep
  name: my-dep
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 3
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: my-dep
  strategy:
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: my-dep
    spec:
      hostNetwork: true
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: nginx
        ports:
        - containerPort: 18888
          name: diameterstack
          protocol: TCP
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
[root@ovnkubernetes main-host-networked]# cat svc.yaml 
apiVersion: v1
kind: Service
metadata:
  labels:
    app: my-dep
  name: my-dep
  namespace: default
spec:
  # allocateLoadBalancerNodePorts: false
  ipFamilies:
  - IPv4
  # - IPv6
  # ipFamilyPolicy: RequireDualStack
  externalTrafficPolicy: Local
  ports:
  - name: diameterstack-port
    port: 8080
    protocol: TCP
    targetPort: diameterstack
  selector:
    app: my-dep
  type: LoadBalancer
```

In this case, this leads to the attempt to inject invalid flows into br-ex:
```
E0925 15:20:37.228390    1087 openflow_manager.go:131] Failed to add flows, error: exit status 1, stderr, ovs-ofctl: -:5: invalid nat range "172.20.0.3:diameterstack"
```

code rabbit hid this gem in the "nitpicks" - turns out there is more code using TargetPort verbatim:

go-controller/pkg/node/gateway_shared_intf.go (2)

    300-327: Named targetPort in OVS NAT flows is not resolved to a number

    In updateServiceFlowCache and createLbAndExternalSvcFlows, NAT flows use svcPort.TargetPort.String() directly. For named ports this yields the literal name (e.g., "diameterstack"), which is invalid in OVS matches/NAT actions and will break ETP=Local host-network scenarios.

    Resolve to a numeric target port (preferably from localEndpoints per servicePortKey) and use that integer in flow strings. This likely requires threading localEndpoints into updateServiceFlowCache/createLbAndExternalSvcFlows.

    Can you confirm if named targetPorts under ETP=Local + host-network backends are in scope for this PR? If not, I suggest tracking this as a follow-up to avoid silent misprogramming of flows.

    Also applies to: 507-528

    1306-1314: Nil vs empty map may cause unnecessary flow churn

    localEndpointsHaveChanged uses reflect.DeepEqual and can treat nil vs empty map as different, causing a delete+add on first update even if no endpoints exist.

    Consider normalizing to empty maps on cache write/read to avoid unnecessary churn. Not blocking.

### Verify the above also in gw mode local

While testing, I think that flows weren't working for the above scenario for gm local, either. Verify

## 📑 Description

Fixes #https://issues.redhat.com/browse/OCPBUGS-59552

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

- **Bug Fixes**
  - LoadBalancer NAT: per-port, per-endpoint DNAT targets, IPv4/IPv6-aware endpoint selection, and safe empty-endpoint handling.

- **Refactor**
  - Unified endpoint models and APIs: new per-port/node endpoint representations and utilities; service and load‑balancer flows updated to use them; simplified config generation.

- **Tests**
  - Expanded unit and E2E coverage: named ports, protocol-aware paths, readiness sequencing, port-allocation retries, dual-stack flows, and firewall/rule verifications.

- **New Helpers**
  - Added test helpers for unassigned endpoints and improved iptables/nftables inspection, parsing, and matching utilities.

### DIFF
--- a/test/e2e/service.go
+++ b/test/e2e/service.go
@@ -2006,6 +2006,196 @@ spec:
 		// we don't... Is this check really needed?
 	})
 
+	ginkgo.It("Should ensure load balancer service works with 0 node ports when named targetPorts are used and ETP=local", func() {
+		err := WaitForServingAndReadyServiceEndpointsNum(context.TODO(), f.ClientSet, namespaceName, svcName, 4, time.Second, time.Second*180)
+		framework.ExpectNoError(err, fmt.Sprintf("service: %s never had an endpoint, err: %v", svcName, err))
+
+		svcLoadBalancerIP, err := getServiceLoadBalancerIP(f.ClientSet, namespaceName, svcName)
+		framework.ExpectNoError(err, fmt.Sprintf("failed to get service lb ip: %s, err: %v", svcName, err))
+
+		time.Sleep(time.Second * 5) // buffer to ensure all rules are created correctly
+
+		checkExactETPRules := func(noSNATServicesSet string) {
+			svc, err := f.ClientSet.CoreV1().Services(namespaceName).Get(context.TODO(), svcName, metav1.GetOptions{})
+			framework.ExpectNoError(err)
+
+			// Retrieve the loadbalancer's IP
+			var lbIP string
+			if len(svc.Status.LoadBalancer.Ingress) > 0 {
+				lbIP = svc.Status.LoadBalancer.Ingress[0].IP
+			}
+
+			discoveryClient := f.ClientSet.DiscoveryV1()
+			endpointSlices, err := discoveryClient.EndpointSlices(namespaceName).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: fmt.Sprintf("kubernetes.io/service-name=%s", svcName),
+			})
+			framework.ExpectNoError(err)
+
+			// Retrieve unique addresses from all endpointslices
+			uniqueAddresses := sets.New[string]()
+			for _, es := range endpointSlices.Items {
+				uniqueAddresses = uniqueAddresses.Union(getServingAndReadyEndpointSliceAddresses(es))
+			}
+
+			mask := 32
+			if utilnet.IsIPv6String(lbIP) {
+				mask = 128
+			}
+
+			// Build regex patterns for iptables rules using lbIP and uniqueAddresses
+			var patterns []string
+			var sets [][]string
+			for address := range uniqueAddresses {
+				tcpPattern := fmt.Sprintf(".*-A OVN-KUBE-ETP -d %s/%d -p tcp -m tcp --dport 80 -m statistic "+
+					"--mode random --probability .* -j DNAT --to-destination %s$",
+					regexp.QuoteMeta(lbIP), mask, regexp.QuoteMeta(net.JoinHostPort(address, "80")))
+				patterns = append(patterns, tcpPattern)
+				udpPattern := fmt.Sprintf(".*-A OVN-KUBE-ETP -d %s/%d -p udp -m udp --dport 10001 -m statistic "+
+					"--mode random --probability .* -j DNAT --to-destination %s$",
+					regexp.QuoteMeta(lbIP), mask, regexp.QuoteMeta(net.JoinHostPort(address, "10001")))
+				patterns = append(patterns, udpPattern)
+
+				sets = append(sets, []string{address, "tcp", "80"})
+				sets = append(sets, []string{address, "udp", "10001"})
+			}
+			err = wait.PollImmediate(retryInterval, retryTimeout, checkIPTablesRulesPresent(backendNodeName, patterns))
+			framework.ExpectNoError(err, "Couldn't fetch the correct iptables rules, expected to find: %v, err: %v", patterns, err)
+			err = wait.PollImmediate(retryInterval, retryTimeout, checkNFTElementsPresent(backendNodeName, noSNATServicesSet, sets))
+			framework.ExpectNoError(err, "Couldn't fetch the correct nft elements, expected to find: %v, err: %v", sets, err)
+		}
+
+		noSNATServicesSet := "mgmtport-no-snat-services-v4"
+		if utilnet.IsIPv6String(svcLoadBalancerIP) {
+			noSNATServicesSet = "mgmtport-no-snat-services-v6"
+		}
+
+		// Initial sanity check.
+		ginkgo.By("checking number of firewall rules for baseline")
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfETPRules(backendNodeName, 2, "OVN-KUBE-ETP"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of iptable rules, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfETPRules(backendNodeName, 5, "OVN-KUBE-EXTERNALIP"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of iptable rules, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 0, noSNATServicesSet))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of nftables elements, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 0, "mgmtport-no-snat-nodeports"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of nftables elements, err: %v", err)
+
+		ginkgo.By("by sending a TCP packet to service " + svcName + " with type=LoadBalancer in namespace " + namespaceName + " with backend pod " + backendName)
+		externalContainer := infraapi.ExternalContainer{Name: externalClientContainerName}
+		_, err = wgetInExternalContainer(externalContainer, svcLoadBalancerIP, endpointHTTPPort, "big.iso")
+		framework.ExpectNoError(err, "failed to curl load balancer service")
+
+		// Patch the service to use named ports.
+		ginkgo.By("patching service " + svcName + " to named ports")
+		err = patchServiceStringValue(f.ClientSet, svcName, "default", "/spec/ports/0/targetPort", "http")
+		framework.ExpectNoError(err)
+		err = patchServiceStringValue(f.ClientSet, svcName, "default", "/spec/ports/1/targetPort", "udp")
+		framework.ExpectNoError(err)
+		output := e2ekubectl.RunKubectlOrDie("default", "get", "svc", svcName, "-o=jsonpath='{.spec.ports[0].targetPort}'")
+		gomega.Expect(output).To(gomega.Equal("'http'"))
+		output = e2ekubectl.RunKubectlOrDie("default", "get", "svc", svcName, "-o=jsonpath='{.spec.ports[1].targetPort}'")
+		gomega.Expect(output).To(gomega.Equal("'udp'"))
+		time.Sleep(time.Second * 5) // buffer to ensure all rules are created correctly
+		ginkgo.By("checking number of firewall rules for named ports")
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfETPRules(backendNodeName, 2, "OVN-KUBE-ETP"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of iptable rules, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfETPRules(backendNodeName, 5, "OVN-KUBE-EXTERNALIP"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of iptable rules, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 0, noSNATServicesSet))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of nftables elements, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 0, "mgmtport-no-snat-nodeports"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of nftables elements, err: %v", err)
+		ginkgo.By("by sending a TCP packet to service " + svcName + " with type=LoadBalancer in namespace " + namespaceName + " with backend pod " + backendName)
+		externalContainer = infraapi.ExternalContainer{Name: externalClientContainerName}
+		_, err = wgetInExternalContainer(externalContainer, svcLoadBalancerIP, endpointHTTPPort, "big.iso")
+		framework.ExpectNoError(err, "failed to curl load balancer service")
+
+		// Patch the service to use allocateLoadBalancerNodeProts=false and externalTrafficPolicy=local.
+		ginkgo.By("patching service " + svcName + " to allocateLoadBalancerNodePorts=false and externalTrafficPolicy=local")
+
+		err = patchServiceBoolValue(f.ClientSet, svcName, "default", "/spec/allocateLoadBalancerNodePorts", false)
+		framework.ExpectNoError(err)
+
+		output = e2ekubectl.RunKubectlOrDie("default", "get", "svc", svcName, "-o=jsonpath='{.spec.allocateLoadBalancerNodePorts}'")
+		gomega.Expect(output).To(gomega.Equal("'false'"))
+
+		err = patchServiceStringValue(f.ClientSet, svcName, "default", "/spec/externalTrafficPolicy", "Local")
+		framework.ExpectNoError(err)
+
+		output = e2ekubectl.RunKubectlOrDie("default", "get", "svc", svcName, "-o=jsonpath='{.spec.externalTrafficPolicy}'")
+		gomega.Expect(output).To(gomega.Equal("'Local'"))
+
+		time.Sleep(time.Second * 5) // buffer to ensure all rules are created correctly
+
+		ginkgo.By("checking number of firewall rules for allocateLoadBalancerNodePorts=false and externalTrafficPolicy=local")
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfETPRules(backendNodeName, 10, "OVN-KUBE-ETP"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of iptable rules, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 8, noSNATServicesSet))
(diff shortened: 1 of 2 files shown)


andreaskaris:
While working on this, I noticed that OVNK currently has a lack of support for a mixture of Pods in the Service.

https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
```
Port definitions

Port definitions in Pods have names, and you can reference these names in the targetPort attribute of a Service. For example, we can bind the targetPort of the Service to the Pod port in the following way:

apiVersion: v1
kind: Pod
metadata:
  name: nginx
  labels:
    app.kubernetes.io/name: proxy
spec:
  containers:
  - name: nginx
    image: nginx:stable
    ports:
      - containerPort: 80
        name: http-web-svc

---
apiVersion: v1
kind: Service
metadata:
  name: nginx-service
spec:
  selector:
    app.kubernetes.io/name: proxy
  ports:
  - name: name-of-service-port
    protocol: TCP
    port: 80
    targetPort: http-web-svc

This works even if there is a mixture of Pods in the Service using a single configured name, with the same network protocol available via different port numbers. This offers a lot of flexibility for deploying and evolving your Services. For example, you can change the port numbers that Pods expose in the next version of your backend software, without breaking clients.

The default protocol for Services is TCP; you can also use any other supported protocol.

Because many Services need to expose more than one port, Kubernetes supports multiple port definitions for a single Service. Each port definition can have the same protocol, or a different one.
```

The current code doesn't respect this: "This works even if there is a mixture of Pods in the Service using a single configured name, with the same network protocol available via different port numbers. This offers a lot of flexibility for deploying and evolving your Services. For example, you can change the port numbers that Pods expose in the next version of your backend software, without breaking clients."

Example:
```
[root@ovnkubernetes test]# cat deployment.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: my-dep
  name: my-dep
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: my-dep
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: my-dep
        svc: my-svc
    spec:
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: nginx
        ports:
        - containerPort: 80
          name: diameterstack
          protocol: TCP
[root@ovnkubernetes test]# cat deployment2.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: my-dep2
  name: my-dep2
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: my-dep2
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: my-dep2
        svc: my-svc
    spec:
      containers:
      - name: webserver
        image: python:3.11-alpine
        command: ["python", "-m", "http.server", "8081"]
        ports:
        - containerPort: 8081
          name: diameterstack
          protocol: TCP
[root@ovnkubernetes test]# cat svc.yaml 
apiVersion: v1
kind: Service
metadata:
  labels:
    app: my-svc
  name: my-svc
  namespace: default
spec:
  allocateLoadBalancerNodePorts: false
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: RequireDualStack
  externalTrafficPolicy: Local
  ports:
  - name: diameterstack
    port: 8080
    protocol: TCP
    targetPort: diameterstack
  selector:
    svc: my-svc
  type: LoadBalancer
```

This doesn't work with OVNKubernetes, at the moment. I'll defer this to a follow-up, right here I want to fix the actual bug that the end user ran into. Maybe the lack of support for a mixture of Pods in the Service is a feature and not a bug in OVNK?

andreaskaris:
@coderabbitai review

andreaskaris:
code rabbit hid this gem in the "nitpicks" - turns out there is more code using TargetPort verbatim:
```
go-controller/pkg/node/gateway_shared_intf.go (2)

    300-327: Named targetPort in OVS NAT flows is not resolved to a number

    In updateServiceFlowCache and createLbAndExternalSvcFlows, NAT flows use svcPort.TargetPort.String() directly. For named ports this yields the literal name (e.g., "diameterstack"), which is invalid in OVS matches/NAT actions and will break ETP=Local host-network scenarios.

    Resolve to a numeric target port (preferably from localEndpoints per servicePortKey) and use that integer in flow strings. This likely requires threading localEndpoints into updateServiceFlowCache/createLbAndExternalSvcFlows.

    Can you confirm if named targetPorts under ETP=Local + host-network backends are in scope for this PR? If not, I suggest tracking this as a follow-up to avoid silent misprogramming of flows.

    Also applies to: 507-528

    1306-1314: Nil vs empty map may cause unnecessary flow churn

    localEndpointsHaveChanged uses reflect.DeepEqual and can treat nil vs empty map as different, causing a delete+add on first update even if no endpoints exist.

    Consider normalizing to empty maps on cache write/read to avoid unnecessary churn. Not blocking.
```

andreaskaris:
@coderabbitai review

andreaskaris:
@coderabbitai review

andreaskaris:
@coderabbitai review

andreaskaris:
@coderabbitai review

andreaskaris:
@coderabbit review

andreaskaris:
@trozet friendly reminder

cgoncalves:
LGTM

go-controller/pkg/util/util.go
```diff
@@ -679,3 +681,392 @@ func MustParseCIDR(cidr string) *net.IPNet {
...
+	r := make(PortToLBEndpoints)
+	for port, nodeToLBEndpoints := range p {
+		if lbe, ok := nodeToLBEndpoints[node]; ok {
+			r[port] = lbe
+		}
+	}
+	return r
+}
+
+// GetEndpointsForService extracts load balancer endpoints from EndpointSlices for a given Service.
```
ricky-rav: nit: I would remove "load balancer" and just say `service endpoints` or simply `endpoints`

go-controller/pkg/util/util.go
```diff
@@ -679,3 +681,392 @@ func MustParseCIDR(cidr string) *net.IPNet {
...
+
+// getPortName safely extracts the port name from a pointer, returning empty string if nil.
+// This handles the case where EndpointSlice ports may have unnamed ports.
+//
+// Parameters:
+//   - name: Pointer to port name (may be nil)
+//
+// Returns:
+//   - string: Port name or empty string if nil
+func getPortName(name *string) string {
```
ricky-rav, line 1066: I had some doubts about whether we could have more than one endpoint port with an empty name, but according to the documentation we should be good:
```
type EndpointPort struct {
	// name represents the name of this port. All ports in an EndpointSlice must have a unique name.
	// If the EndpointSlice is derived from a Kubernetes service, this corresponds to the Service.ports[].name.
	// Name must either be an empty string or pass DNS_LABEL validation:
	// * must be no more than 63 characters long.
	// * must consist of lower case alphanumeric characters or '-'.
	// * must start and end with an alphanumeric character.
	// Default is empty string.
	Name *string `json:"name,omitempty" protobuf:"bytes,1,name=name"`

```
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/vendor/k8s.io/api/discovery/v1/types.go#L187-L197

go-controller/pkg/ovn/controller/services/lb_config.go
```diff
@@ -137,7 +131,7 @@ var protos = []corev1.Protocol{
 //   - services with NodePort set but *without* ExternalTrafficPolicy=Local or
 //     affinity timeout set.
 func buildServiceLBConfigs(service *corev1.Service, endpointSlices []*discovery.EndpointSlice, nodeInfos []nodeInfo,
-	useLBGroup, useTemplates bool, networkName string) (perNodeConfigs, templateConfigs, clusterConfigs []lbConfig) {
+	useLBGroup, useTemplates bool) (perNodeConfigs, templateConfigs, clusterConfigs []lbConfig) {
```
ricky-rav, line 134: I was going to ask you why you're dropping `networkName`, then I saw that it was only used for logging. All good.

go-controller/pkg/util/util.go
```diff
@@ -679,3 +681,392 @@ func MustParseCIDR(cidr string) *net.IPNet {
...
+			continue
+		}
+		nodeLBEndpoints[node] = lbe
+	}
+
+	var err error = nil
+	if len(nodeLBEndpoints) == 0 {
+		return nodeLBEndpoints, fmt.Errorf("empty node lb endpoints")
+	}
+	return nodeLBEndpoints, err
```
ricky-rav: Here we're defining `err`, but we're not really assigning it. We can remove `var err error = nil` and just do `return nodeLBEndpoints, nil` at the end, right?

go-controller/pkg/util/util.go
```diff
@@ -679,3 +681,392 @@ func MustParseCIDR(cidr string) *net.IPNet {
...
+//     service may be nil!
+//   - port: The target port number for the endpoints
+//   - endpoints: List of discovery endpoints to process
+//
+// Returns:
+//   - LBEndpoints: Structure containing IPv4/IPv6 addresses and port number
+func buildLBEndpoints(service *corev1.Service, port int32, endpoints []discoveryv1.Endpoint) (LBEndpoints, error) {
+	addresses := GetEligibleEndpointAddresses(endpoints, service)
+	v4IPs, _ := MatchAllIPStringFamily(false, addresses)
+	v6IPs, _ := MatchAllIPStringFamily(true, addresses)
```
ricky-rav: super nit: do we want to ignore the error from `MatchAllIPStringFamily`? I see that it returns `ErrorNoIP` when it doesn't find any IP addresses in the provided `addresses`. Here below you're repeating the same logic by checking again the size of the two slices. Perhaps I would replace that by just checking if there was an error in both function calls above.

test/e2e/service.go
```diff
@@ -2333,6 +2340,41 @@ func getServiceLoadBalancerIP(c clientset.Interface, namespace, serviceName stri
...
+	})
+}
+
+func waitForServiceEndpointsNumAndDeploymentReady(clientSet clientset.Interface, namespaceName, serviceName,
+	deploymentName string, expectNum int, interval, retryTimeout time.Duration) error {
+	if err := framework.WaitForServiceEndpointsNum(context.TODO(), clientSet, namespaceName, serviceName, expectNum,
+		interval, retryTimeout); err != nil {
+		return fmt.Errorf("service: %s never had an endpoint, err: %q", serviceName, err)
+	}
+	// If pods are in image pull backoff or initializing (for example), the endpoints will be marked as ready.
```
ricky-rav: Are you sure that `ImagePullBackOff` shows this issue? The container isn't even there, right?

test/e2e/service.go
```diff
@@ -2136,8 +2142,9 @@ spec:
 		// Now we test ETP=local works as expected without EIP re-routes messing with the reply traffic:
 		// lbclient->FRR router->ovn-worker->br-ex->GR_ovn-worker->join->cluster-router->ovn-worker-switch->pod and response goes back
 		// same way without it getting re-routed to egressNode ovn-worker2
-		err := framework.WaitForServiceEndpointsNum(context.TODO(), f.ClientSet, namespaceName, svcName, 4, time.Second, time.Second*120)
```
ricky-rav: This function has been modified recently in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5565 .
You were both trying to solve the same problem, right? :) 

Here are you addressing the case in which no readiness probe is defined and kubelet marks the pod as ready when... the pod is just running? (if I got it right)

If so, we can combine the two approaches.

andreaskaris:
ok I addressed your comments. I'll wait for CI and then have another look myself tomorrow, just to make sure that I didn't mess anything up while being tired

andreaskaris:
@ricky-rav ready for another round, thanks :)

ricky-rav:
@andreaskaris before I take a look again at your PR, I wanted to comment on the other TODO items that you correctly identified.

### (1)
> **ovnkubernetes assumes that named ports can only have one target port**
> ovnkubernetes maps serviceport -> slice port -> target port all one to one. however, with named ports, a single service port can match multiple pods with the same targetPort name, but different target ports. Should be taken care of in a follow-up.

This we can maybe postpone to a later stage, but I have a feeling that it's the exact use case a customer wants for a named target port :-/ 
I have one question though. Isn't this exactly what you described in detail in a separate message in this PR?
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5583#issuecomment-3312797804

### (2)
> **br-ex bad flows in gateway mode shared**
> The very same parsing issue is present for shared GM:
> [...]

I was looking at all the instances of `TargetPort` in our source code and indeed I found that we do `svcPort.TargetPort.String()` in `updateServiceFlowCache` and  `createLbAndExternalSvcFlows`, ignoring that `TargetPort` can be a number or a named port :( 
This happens for ETP=local services for local endpoints that are host-networked:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_shared_intf.go#L302-L324
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/gateway_shared_intf.go#L507-L528

I will open bugs on jira to correctly track these two followup tasks...

andreaskaris:
> I have one question though. Isn't this exactly what you described in detail in a separate message in this PR?

Yes. Sorry, I created the reply first, then modified the PR description

andreaskaris:
> I will open bugs on jira to correctly track these two followup tasks...

Sounds good. I thought about doing those in this PR, but it was a lot of work and I felt that it'd bloat the scope of this here + make review more complex

test/e2e/service.go
```diff
@@ -2006,6 +2006,196 @@ spec:
...
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 0, noSNATServicesSet))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of nftables elements, err: %v", err)
+		err = wait.PollImmediate(retryInterval, retryTimeout, checkNumberOfNFTElements(backendNodeName, 0, "mgmtport-no-snat-nodeports"))
+		framework.ExpectNoError(err, "Couldn't fetch the correct number of nftables elements, err: %v", err)
+		ginkgo.By("by sending a TCP packet to service " + svcName + " with type=LoadBalancer in namespace " + namespaceName + " with backend pod " + backendName)
+		externalContainer = infraapi.ExternalContainer{Name: externalClientContainerName}
+		_, err = wgetInExternalContainer(externalContainer, svcLoadBalancerIP, endpointHTTPPort, "big.iso")
+		framework.ExpectNoError(err, "failed to curl load balancer service")
+
+		// Patch the service to use allocateLoadBalancerNodeProts=false and externalTrafficPolicy=local.
```
ricky-rav, line 2113: nit: `allocateLoadBalancerNodePorts`. I'm surprised code rabbit didn't spot this :)

ricky-rav:
> > I will open bugs on jira to correctly track these two followup tasks...
> 
> Sounds good. I thought about doing those in this PR, but it was a lot of work and I felt that it'd bloat the scope of this here + make review more complex

Yeah, the PR is already big enough :)

PR #5584, opened 2025-09-19
Fix retry logic for endpoint slices deletion

abhat:
In cases where the namespace is already deleted, retry logic should skip endpoint slice object deletion and return success.

Signed-off-by: Aniket Bhat <anbhat@redhat.com>
Assisted-by: Cursor.ai

## Summary by CodeRabbit

* **Bug Fixes**
  * Skip unnecessary retries when removing endpoint data if the related namespace has already been deleted, improving stability and reducing retry noise.
* **Logging**
  * Log namespace-not-found at low verbosity and emit warnings only for unexpected errors, reducing spurious messages and aiding troubleshooting.

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -10,6 +10,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	discovery "k8s.io/api/discovery/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/klog/v2"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
@@ -216,6 +217,17 @@ func (g *gateway) UpdateEndpointSlice(oldEpSlice, newEpSlice *discovery.Endpoint
 func (g *gateway) DeleteEndpointSlice(epSlice *discovery.EndpointSlice) error {
 	var err error
 	var errors []error
+
+	// Check if the namespace still exists before proceeding
+	_, err = g.watchFactory.GetNamespace(epSlice.Namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Namespace %s not found, skipping endpointslice %s deletion (namespace was deleted)", epSlice.Namespace, epSlice.Name)
+			return nil // Return success to skip retry
+		}
+		// For other errors, continue with the deletion process
+		klog.Warningf("Failed to check namespace %s existence: %v, continuing with endpointslice deletion", epSlice.Namespace, err)
+	}
 
 	if g.loadBalancerHealthChecker != nil {
 		// Filter out objects without the default serviceName label to exclude mirrored EndpointSlices
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1169,6 +1169,17 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 	var errors []error
 	var hasLocalHostNetworkEp = false
 
+	// Check if the namespace still exists before proceeding
+	_, err = npw.watchFactory.GetNamespace(epSlice.Namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Namespace %s not found, skipping endpointslice %s deletion (namespace was deleted)", epSlice.Namespace, epSlice.Name)
+			return nil // Return success to skip retry
+		}
+		// For other errors, continue with the deletion process
+		klog.Warningf("Failed to check namespace %s existence: %v, continuing with endpointslice deletion", epSlice.Namespace, err)
+	}
+
 	networkName := types.DefaultNetworkName
 	if util.IsNetworkSegmentationSupportEnabled() {
 		if netName, ok := epSlice.Annotations[types.UserDefinedNetworkEndpointSliceAnnotation]; ok {


go-controller/pkg/node/gateway.go
```diff
@@ -217,6 +218,17 @@ func (g *gateway) DeleteEndpointSlice(epSlice *discovery.EndpointSlice) error {
...
 
+	// Check if the namespace still exists before proceeding
+	_, err = g.watchFactory.GetNamespace(epSlice.Namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Namespace %s not found, skipping endpointslice %s deletion (namespace was deleted)", epSlice.Namespace, epSlice.Name)
+			return nil // Return success to skip retry
+		}
+		// For other errors, continue with the deletion process
+		klog.Warningf("Failed to check namespace %s existence: %v, continuing with endpointslice deletion", epSlice.Namespace, err)
```
pliurh, line 229: Instead of continuing here, should we return an error to retry?

abhat: My logic was if the namespace error is not because the namespace was deleted, then the EndPointSlice deletion's retry count shouldn't be affected negatively. In some ways, we get the alert firing because we exceed the retry count. So this will make the situation worse?

go-controller/pkg/node/gateway.go
```diff
@@ -217,6 +218,17 @@ func (g *gateway) DeleteEndpointSlice(epSlice *discovery.EndpointSlice) error {
 	var err error
 	var errors []error
 
+	// Check if the namespace still exists before proceeding
+	_, err = g.watchFactory.GetNamespace(epSlice.Namespace)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Namespace %s not found, skipping endpointslice %s deletion (namespace was deleted)", epSlice.Namespace, epSlice.Name)
```
abhat, line 225: The bigger question is if we will leak any ovn objects if we don't continue with retrying the EndPointSlice deletion logic.

pliurh: Yeah, it could happen. We may need a periodic resync mechanism to clean up the orphan objects.

pliurh:
The patch needs a rebase. Apart from that, it looks good to me.

go-controller/pkg/node/gateway.go
tssurya: given the code is being executed on pkg/node there is nothing ovn object specific in this code path.
This new code path return prevents two things from happening:

1. given we don't call load balancer healthcheck delete endpointslice code anymoe, that blocks the call towards SyncEndpoints and since we are not level driven in this part of the code, that will have undesired consequences.. so I don't think we should skip this 
2. next we don't call nodeport watcher for deletion - that will also leave us with stale iptable rules for EPS/SVC for that namespace

So here the logic seems wrong to me that given the endpoint is not yet deleted we shouldn't be short circuiting and we must wait for the endpoint delete event to clean up things. 

The expectation is namespace delete will, thanks to the finalizer always wait for the pods i.e endpointslices as well to be deleted before we clean that up - so natural ordering would be to wait for that.

One more thing I don't get is the location of the fix. We have done it in two spots:

1. func (g *gateway) DeleteEndpointSlice which calls g.loadBalancerHealthChecker.DeleteEndpointSlice and g.nodePortWatcher.DeleteEndpointSlice but inside lbhealthchecker we seem to be using the `services` and `endpoints` caches - I'm not sure at which spot we failed in CI and if it was inside this func? in CI I see it only failing inside DeleteEndpointSlice of the nodeportwatcher..
2. func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice

a few options to move forward:

1. we use caches for eps and services during delete and don't depend on actual objects existing in api/informer - so that means the serviceInfo would need to keep a copy of the namespace obj
2. use periodic cleanup - icky
3. move to level driven and eventual consistency - long term
4. This seems more easier to do - If we look carefully at the code around `DeleteEndpointSlice` inside nodePortWatcher which is where the CI said we hit the issue:

```
Failed to delete *factory.endpointSliceForGateway e2e-services-1023/clusterip-service-rt78b, error: error getting active network for service e2e-services-1023/clusterip-service: failed to get namespace "e2e-services-1023": namespace "e2e-services-1023" not found
```

```
if svcConfig, exists := npw.updateServiceInfo(*namespacedName, nil, &hasLocalHostNetworkEp, localEndpoints); exists {
		netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(namespacedName.Namespace)
        // tssurya: This is where we fail
		if err != nil {
			return fmt.Errorf("error getting active network for service %s/%s: %w", namespacedName.Namespace, namespacedName.Name, err)
		}

		// Lock the cache mutex here so we don't miss a service delete during an endpoint delete
		// we have to do this because deleting and adding iptables rules is slow.
		npw.serviceInfoLock.Lock()
		defer npw.serviceInfoLock.Unlock()
        // tssurya: we don't use netinfo in delete!
		if err = delServiceRules(svcConfig.service, sets.List(svcConfig.localEndpoints), npw); err != nil {
			errors = append(errors, err)
		}
        // tssurya:GetActiveNetworkForNamespace can be moved here or even inside addServiceRules
       // and before we call the add we just need to check if namespace exists or not.. 
		if err = addServiceRules(svcConfig.service, netInfo, sets.List(localEndpoints), hasLocalHostNetworkEp, npw); err != nil {
			errors = append(errors, err)
		}
```
this is the part that fails and this was added during UDN time frame - for default network this should ideally be no-op but:
```
func (c *nadController) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
		return &util.DefaultNetInfo{}, nil
	}

	// check if required UDN label is on namespace
	ns, err := c.namespaceLister.Get(namespace)
	if err != nil {
		return nil, fmt.Errorf("failed to get namespace %q: %w", namespace, err)
	}
	if _, exists := ns.Labels[types.RequiredUDNNamespaceLabel]; !exists {
		// UDN required label not set on namespace, assume default network
		return &util.DefaultNetInfo{}, nil
	}
```
but unfortunately we depend on the new annotation on the namespace to call out if its a UDN network or not..

However notice that `delServiceRules` don't depend on netInfo! only the `addServiceRules` depend on it - so IMHO we should be cleaning them up for the case where namespace is gone and services/eps is gone or is being deleted but we don't care about the addServiceRules succeeding here if the entire namespace is already gone (cause both services and EPS are namespace scoped - so we shouldn't need to attempt to re-add any of that setup). So we should move that GetActiveNetworkForNamespace to a spot after the delete and then we should catch the namespace not existing as a apierrors.IsNotFound in there. Also given netInfo is used only inside `addServiceRules` might make more sense to just move GetActiveNetworkForNamespace into that function as well and keep the namespace exists or not check in the DeleteEndpointSlice function

Another aspect to consider is that same problem should happen for services as well which are namespace scoped objects, but we don't see that happening - wondering why this is only an issue for EPS but that's for another PR if we hit that I guess.

tssurya (changes requested):
left an inline question/suggestion
also this fix needs a unit test

pliurh:
I created a new PR according to @tssurya 's comments https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669

Issue #5456, opened 2025-08-07
Kubevirt Virtual Machines with user defined networks with ipamless localnet topology should maintain tcp connection with minimal downtime [It] after failed live migration


npinaeva:
### Which jobs are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/16776444857/job/47520697197?pr=5396

### Which tests are flaking?

Kubevirt Virtual Machines with user defined networks with ipamless localnet topology should maintain tcp connection with minimal downtime [It] after failed live migration

### Since when has it been flaking?

7-aug-2025

### Reason for failure (if possible)

```
[FAILED] Timed out after 2.350s.
  worker-s8sjr: Restart iperf traffic after forcing a vm failure: failed checking iperf3 traffic at file /tmp/testpod-ovn-worker3_10.128.0.2_iperf3.log
  Expected
      <string>: 
  to contain substring
      <string>:  sec 
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:2335 @ 08/06/25 16:18:48.726

```

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/17318222412/job/49167512561?pr=5516
```
Unexpected error:
      <*errors.errorString | 0xc0018bb920>: 
      failed to run [killall iperf3] at VMI worker-r9mm7, error: killall iperf3 failed
      {
          s: "failed to run [killall iperf3] at VMI worker-r9mm7, error: killall iperf3 failed",
      }
  occurred
```

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/17491474378/job/49684222056?pr=5561
```
  worker-mb82h: Restart iperf traffic after forcing a vm failure: failed checking iperf3 traffic at file /tmp/testpod-ovn-worker2_10.128.0.1_iperf3.log
  Expected
      <string>: 
  to contain substring
      <string>:  sec 
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:2340
```

martinkennelly:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/17499204439/job/49710067581?pr=5493

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5493

qinqon:
[...]

PR #5588, opened 2025-09-23
api, udn: subnets must be masked

maiqueb:
## 📑 Description

The OVN-K API was allowing unmasked CIDRs to be used in the UDN API. This led to some misleading error messages, for instance:

Imagine the user would provision this C-UDN manifest:
```yaml
cat <<EOF | kubectl apply -f -
apiVersion: k8s.ovn.org/v1
metadata:
  name: l2-network
spec:
  namespaceSelector:
    matchLabels:
      cudn-group: l2-net
  network:
    topology: Layer2
    layer2:
      role: Primary
      subnets: ["192.168.40.0/24"]
      reservedSubnets: ["192.168.40.4/24"]
      defaultGatewayIPs:
        - "192.168.40.3"
EOF
```
The C-UDN would be provisioned, but, when a workload was created, attaching to this UDN, it would fail with the following error msg:
> 14:19:07.462812 3361 controller.go:258] Controller [zone-nad-controller network controller]: error found while processing cluster_udn_l2-network: [zone-nad-controller network controller]: failed to ensure network cluster_udn_l2-network: failed to start network cluster_udn_l2-network: failed to exclude subnet 192.168.40.0/24 for cluster_udn_l2.network_ovn_layer2_switch: failed to reserve IP 192.168.40.1: provided IP is already allocated

The reason is the API would then mask the CIDR on the `reservedSubnets` attribute - i.e. 192.168.40.0/24 - , and would 
**reserve** the entire subnet. The controller would then fail to reserve any IP address within the subnet.

After this patch, OVN-Kubernetes fails when provisioning said C-UDN, with the following error:
```sh
cat <<EOF | kubectl apply -f -
apiVersion: k8s.ovn.org/v1
metadata:
  name: l2-network
spec:
  namespaceSelector:
    matchLabels:
      cudn-group: l2-net
  network:
    topology: Layer2
    layer2:
      role: Primary
      subnets: ["192.168.40.0/24"] 
      reservedSubnets: ["192.168.40.4/24"]         
      defaultGatewayIPs:
        - "192.168.40.3"
EOF
The ClusterUserDefinedNetwork "l2-network" is invalid: spec.network.layer2.reservedSubnets[0]: Invalid value: "string": CIDR is invalid
```

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced that Layer 2 infrastructure and reserved subnets must be masked network addresses (no host bits) and surface clear validation errors when unmasked CIDRs are provided.

* **Tests**
  * Updated Layer 2 tests and added negative cases to confirm unmasked infrastructure/reserved subnets are rejected with the expected error messages.

### DIFF
--- a/test/e2e/network_segmentation_preconfigured_layer2.go
+++ b/test/e2e/network_segmentation_preconfigured_layer2.go
@@ -121,7 +121,7 @@ var _ = Describe("Network Segmentation: Preconfigured Layer2 UDN", feature.Netwo
 				role:                "primary",
 				defaultGatewayIPs:   joinStrings("172.31.0.10", "2014:100:200::100"),
 				reservedCIDRs:       joinStrings("172.31.1.0/24", "2014:100:200::/122"),
-				infrastructureCIDRs: joinStrings("172.31.0.10/30", "2014:100:200::100/122"),
+				infrastructureCIDRs: joinStrings("172.31.0.8/30", "2014:100:200::100/122"),
 			},
 			expectedGatewayIPs: []string{"172.31.0.10", "2014:100:200::100"},
 		}),
--- a/test/e2e/network_segmentation_preconfigured_layer2.go
+++ b/test/e2e/network_segmentation_preconfigured_layer2.go
@@ -137,6 +137,77 @@ var _ = Describe("Network Segmentation: Preconfigured Layer2 UDN", feature.Netwo
 		}),
 	)
 
+	type invalidAPITestConfig struct {
+		netConfig     *networkAttachmentConfigParams
+		expectedError interface{}
+	}
+	DescribeTable("unmasked reserved / infrastructure subnets are not allowed",
+		func(config invalidAPITestConfig) {
+			podIPs := filterCIDRs(f.ClientSet, config.netConfig.cidr)
+			if len(podIPs) == 0 {
+				Skip("IP family not supported in this environment")
+			}
+
+			By("creating the L2 network")
+			netConfig := config.netConfig
+
+			netConfig.namespace = f.Namespace.Name
+
+			udnManifest := generateUserDefinedNetworkManifest(netConfig, f.ClientSet)
+			cleanup, err := createManifest(netConfig.namespace, udnManifest)
+			Expect(err).To(MatchError(config.expectedError))
+			DeferCleanup(cleanup)
+		},
+		Entry("Layer2 with unmasked IPv4 reserved subnets", invalidAPITestConfig{
+			netConfig: &networkAttachmentConfigParams{
+				name:          "invalid-l2-net-reserved-subnets",
+				topology:      "layer2",
+				cidr:          "172.31.0.0/16",
+				role:          "primary",
+				reservedCIDRs: "172.31.0.10/30",
+			},
+			expectedError: ContainSubstring(
+				"Invalid value: \"object\": reservedSubnets must be a masked network address (no host bits set)",
+			),
+		}),
+		Entry("Layer2 with unmasked IPv6 reserved subnets", invalidAPITestConfig{
+			netConfig: &networkAttachmentConfigParams{
+				name:          "invalid-l2-net-reserved-subnets",
+				topology:      "layer2",
+				cidr:          "2014:100:200::0/60",
+				role:          "primary",
+				reservedCIDRs: "2014:100:200::88/122",
+			},
+			expectedError: ContainSubstring(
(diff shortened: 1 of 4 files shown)


npinaeva:
So the idea is that the desired config was something like `reservedSubnets: ["192.168.40.4/30"]`?
I mean if I typed this and got an error that my masked subnet is wrong, I would just update it to `reservedSubnets: ["192.168.40.0/24"]` which will end up with the same error as now?

maiqueb:
> So the idea is that the desired config was something like `reservedSubnets: ["192.168.40.4/30"]`? I mean if I types this and got an error that my masked subnet is wrong, I would just update it to `reservedSubnets: ["192.168.40.0/24"]` which will end up with the same error as now?

That is a good point. I'm not 100% sure on what we should do - honestly, I was just thinking that the user would notice that something's up with their config. At the end of the day, if they're reserving the full range for this, they ought to know that there won't be an address left for the GW, and so forth. 

The argument for providing yet another error could be made, but the same thing might happen for - let's say hybrid overlay, or the management port. I see these as more "complex" since they depend on the cluster deployment mode ...

What do you think @npinaeva ? Which direction do you think this should go ?

tssurya:
headsup that we are having that API discussion on right behaviour of `isCIDR` upstream: 
https://github.com/kubernetes/kubernetes/issues/134224

thanks to @danwinship and @JoelSpeed for the workaround: https://github.com/kubernetes/kubernetes/issues/134224#issuecomment-3324206437

credit to @asood-rh for discovering this during her testing

and we need tests for this PR to catch the behaviour.

maiqueb:
> headsup that we are having that API discussion on right behaviour of `isCIDR` upstream: [kubernetes/kubernetes#134224](https://github.com/kubernetes/kubernetes/issues/134224)
> 
> thanks to @danwinship and @JoelSpeed for the workaround: [kubernetes/kubernetes#134224 (comment)](https://github.com/kubernetes/kubernetes/issues/134224#issuecomment-3324206437)
> 
> credit to @asood-rh for discovering this during her testing
> 
> and we need tests for this PR to catch the behaviour.

We need to figure out a way to make this work with L3 subnets, which, of course, I neglected since I never think about them.

trozet:
imo we should detect a CIDR with extra bits set after the mask length as an error

trozet (dismissed):

trozet:
in general I agree with the fix, although tests are failing...are we providing bad values in our tests ^_^

maiqueb:
Updated this PR to ensure that it preserves backwards compatibility. 

It only focuses now on the recently added attributes - `infrastructureSubnets` and `reservedSubnets`. 
One e2e test was updated to comply with the rule. 

@kyrtapz you've added the patched e2e test. Can you confirm this is now the expected value ? Can you LGTM this PR ?

maiqueb:
Yeah, this is in a lot better shape now. 

@kyrtapz pushed a new commits w/ the tests you've requested.

kyrtapz (dismissed):
Thanks, LGTM.

maiqueb:
The failure of https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18126949473/job/51587079196?pr=5588 is another instance of flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5456

test/e2e/network_segmentation_preconfigured_layer2.go
```diff
@@ -137,6 +137,77 @@ var _ = Describe("Network Segmentation: Preconfigured Layer2 UDN", feature.Netwo
...
+	type invalidAPITestConfig struct {
+		netConfig     *networkAttachmentConfigParams
+		expectedError interface{}
+	}
+	DescribeTable("unmasked reserved / infrastructure subnets are not allowed",
+		func(config invalidAPITestConfig) {
+			podIPs := filterCIDRs(f.ClientSet, config.netConfig.cidr)
+			if len(podIPs) == 0 {
+				Skip("IP family not supported in this environment")
+			}
```
maiqueb, line 149: I don't have a better way to achieve this, and it seems this was the solution found for the "duplicate IP validation with primary UDN layer 2 pods" context. 

Far from good though.

maiqueb:
The failure of https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18137379156/job/51622264571?pr=5588  is another instance of flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5456

@qinqon I know you're busy, but please take a peek at this whenever you can.

PR #5590, opened 2025-09-24
Add e2e test for default network with no-overlay mode

zhaozhanqi:
e2e test coverage for no-overlay mode to include connectivity tests and OVN routing verification:

   - Add NoOverlay feature flag for test categorization
   - Add pod2service and host2service connectivity tests
   - Add OVN route verification to ensure pod subnet routes exist in GR
     but not in ovn_cluster_router (as routing is handled by BGP)
   - Test connectivity persistence after OVN pod restart
   - Add helper functions: isNoOverlayEnabled, isOutboundSNATEnabled,
     isManagedRoutingEnabled, getNodePodSubnets, and checkOVNLogicalRouterRoutes

   These tests ensure that no-overlay mode correctly routes traffic without
   using Geneve overlay and maintains connectivity after component restarts.

Signed-off-by: zhaozhanqi <zzhao@redhat.com>

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end scenario validating BGP no-overlay connectivity across multiple nodes, covering pod-to-pod and host-to-pod flows and asserting no overlay traffic while confirming application responses succeed.
  * Automates node scheduling, pod readiness, traffic capture on the overlay interface, and log collection for assertions to strengthen multi-node networking coverage.

* **Documentation**
  * Added explanatory comments to CI workflow clarifying that no-overlay runs require dedicated jobs and cannot run alongside other tests.

### DIFF
--- a/test/scripts/e2e-cp.sh
+++ b/test/scripts/e2e-cp.sh
@@ -165,6 +165,10 @@ if [ "$ENABLE_EVPN" != true ]; then
   skip_label "Feature:EVPN"
 fi
 
+if [ "$ENABLE_NO_OVERLAY" != true ]; then
+  skip_label "Feature:NoOverlay"
+fi
+
 if [ "$ENABLE_ROUTE_ADVERTISEMENTS" != true ]; then
   skip_label "Feature:RouteAdvertisements"
 else


zhaozhanqi:
/retitle Add e2e test for default network with no-overlay mode

zhaozhanqi:
/retitle Add e2e test for default network with no-overlay mode

test/e2e/util.go
```diff
@@ -2065,3 +2066,66 @@ func waitForNodeReadyState(f *framework.Framework, nodeName string, timeout time
 		return false
 	}, timeout, 10*time.Second).Should(gomega.BeTrue(), expectationMessage)
 }
+
+// isNoOverlayEnabled checks if no-overlay mode is enabled for the default network
+func isNoOverlayEnabled() bool {
+	ovnKubeNamespace := deploymentconfig.Get().OVNKubernetesNamespace()
+	val := getTemplateContainerEnv(ovnKubeNamespace, "daemonset/ovnkube-node", getNodeContainerName(), "OVN_NO_OVERLAY_ENABLE")
```
pliurh: You can just check the environment variable `ENABLE_NO_OVERLAY`. It will be set by CI.

test/e2e/util.go
```diff
@@ -2065,3 +2066,66 @@ func waitForNodeReadyState(f *framework.Framework, nodeName string, timeout time
...
+	args := []string{"get", "configmap", "ovnkube-config", "-o=jsonpath={.data.ovnkube\\.conf}"}
+	conf := e2ekubectl.RunKubectlOrDie(ovnKubeNamespace, args...)
+
+	// Simplistic check for routing = managed
+	if strings.Contains(conf, "routing = managed") || strings.Contains(conf, "routing=managed") {
+		framework.Logf("Managed routing is enabled in ovnkube-config")
+		return true
+	}
+	return false
+}
```
pliurh: These two functions are not used by code in this PR.

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,378 @@
...
+		clientPod.Namespace,
+		clientPod.Name,
+		curlCmd,
+		framework.Poll,
+		10*time.Second)
+	framework.ExpectNoError(curlErr, "Curl server from client failed")
+	framework.Logf("curl output:\n%s", curlOutput)
+
+	collectCmd := "sh -lc 'kill -INT $(cat /tmp/tcpdump.pid) >/dev/null 2>&1 || true; sleep 1; cat /tmp/tcpdump.log'"
+	tcpdumpOut, _ := e2epodoutput.RunHostCmdWithRetries(tcpdumpPod.Namespace, tcpdumpPod.Name, collectCmd, framework.Poll, 10*time.Second)
```
pliurh: shall we ignore the returned error here?

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,378 @@
...
+	ginkgo.By("start tcpdump on genev_sys_6081 interface to capture traffic")
+	startCmd := fmt.Sprintf("sh -lc 'rm -f /tmp/tcpdump.log /tmp/tcpdump.pid; tcpdump -ni genev_sys_6081 tcp and host %s -n -vv -s 0 -l > /tmp/tcpdump.log 2>&1 & echo $! > /tmp/tcpdump.pid'", serverPodIP)
+	_, tcpdumpErr := e2epodoutput.RunHostCmdWithRetries(
+		tcpdumpPod.Namespace,
+		tcpdumpPod.Name, startCmd,
+		framework.Poll,
+		10*time.Second)
+	framework.ExpectNoError(tcpdumpErr, "run tcpdump failed on genev_sys_6081 interface")
+	// Wait 2 seconds to let the tcpdump ready for capturing traffic
+	time.Sleep(2 * time.Second)
```
pliurh: Can we use `gomega.Eventually` instead of a hard coded wait?

zhaozhanqi:
/retitle Add e2e test for default network with no-overlay mode

test/e2e/no_overlay.go
zhaozhanqi: Thanks, updated this to 
```
	gomega.Eventually(func() bool {
		checkCmd := "sh -lc 'test -f /tmp/tcpdump.pid && kill -0 $(cat /tmp/tcpdump.pid) 2>/dev/null'"
		_, err := e2epodoutput.RunHostCmd(tcpdumpPod.Namespace, tcpdumpPod.Name, checkCmd)
		if err != nil {
			framework.Logf("tcpdump not ready yet: %v", err)
			return false
		}
		return true
	}, 5*time.Second, 200*time.Millisecond).Should(gomega.BeTrue(), "tcpdump should be ready within 5 seconds")
```

test/e2e/no_overlay.go
zhaozhanqi: add step to check the err
`framework.ExpectNoError(err, "Failed to collect tcpdump output")`

test/e2e/util.go
zhaozhanqi: yeah, I removed those two functions and will add into another PR which used

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,139 @@
...
+			// Verify pods have IP addresses
+			gomega.Expect(len(serverPod.Status.PodIPs)).To(gomega.BeNumerically(">", 0), "Server pod should have at least one IP address")
+			gomega.Expect(len(clientPod.Status.PodIPs)).To(gomega.BeNumerically(">", 0), "Client pod should have at least one IP address")
+
+			ginkgo.By("Creating tcpdump pod")
+			tcpdumpPod, err = createPod(f, tcpdumpPodName, nodes.Items[1].Name, f.Namespace.Name, []string{"bash", "-c", "apk update; apk add tcpdump; sleep 20000"}, map[string]string{}, func(p *v1.Pod) {
+				p.Spec.HostNetwork = true
+			})
+			framework.ExpectNoError(err)
+
```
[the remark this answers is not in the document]
zhaozhanqi: updated, thanks.

zhaozhanqi:
@pliurh I updated codes according to your comments,  Could you help take another look when you have time, thanks.

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,397 @@
...
+		})
+
+	})
+
+})
+
+// getTcpdumpOnGenevSys starts tcpdump on genev_sys_6081 interface, runs a curl command, and returns the tcpdump output and curl output.
+func getTcpdumpOnGenevSys(tcpdumpPod *corev1.Pod, clientPod *corev1.Pod, curlCmd string, serverPodIP string) (string, string) {
+	ginkgo.By("start tcpdump on genev_sys_6081 interface to capture traffic")
+	startCmd := fmt.Sprintf("sh -lc 'rm -f /tmp/tcpdump.log /tmp/tcpdump.pid; tcpdump -ni genev_sys_6081 tcp and host %s -n -vv -s 0 -l > /tmp/tcpdump.log 2>&1 & echo $! > /tmp/tcpdump.pid'", serverPodIP)
```
pliurh: The `&` in the shell command backgrounds tcpdump, then `echo $!` always succeeds. So tcpdumpErr is always nil even if tcpdump immediately crashes

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,397 @@
...
+		)
+
+		ginkgo.BeforeEach(func() {
+			if !isNoOverlayEnabled() {
+				ginkgo.Skip("Test requires no-overlay mode to be enabled (OVN_NO_OVERLAY_ENABLE=true)")
+			}
+
+			var err error
+			ginkgo.By("Selecting nodes")
+			nodes, err = f.ClientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
```
pliurh: This includes control-plane nodes. Use `e2enode.GetReadySchedulableNodes()` to find worker nodes only.

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,397 @@
...
+			ginkgo.By("Verifying tcpdump is installed successfully")
+			gomega.Eventually(func() bool {
+				checkCmd := "which tcpdump"
+				output, err := e2epodoutput.RunHostCmd(tcpdumpPod.Namespace, tcpdumpPod.Name, checkCmd)
+				if err != nil {
+					framework.Logf("tcpdump check failed: %v", err)
+					return false
+				}
+				framework.Logf("tcpdump check output: %s", output)
+				return output != "" && err == nil
```
pliurh: `&& err == nil` is redundant here.

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,397 @@
...
+		for _, pip := range serverPod.Status.PodIPs {
+			if utilnet.IsIPv6String(pip.IP) == isIPv6 {
+				serverIP = pip.IP
+				break
+			}
+		}
+		gomega.Expect(serverIP).NotTo(gomega.BeEmpty(), "Could not find matching server pod IP family")
+
+		// Use getTcpdumpOnGenevSys to verify no overlay traffic (filtering by backend server IP)
+		tcpdumpOut, output := getTcpdumpOnGenevSys(tcpdumpPod, clientPod, curlCmd, serverIP)
```
pliurh: Instead of verifying no overlay traffic on `genev_sys_6081`, can we verify that the pod IPs can be capture on breth0 without encapsulation?

zhaozhanqi: hmm, for E-W traffic, podIP can be captured on physic interface but not on breth0/br-ex 
```
/ # tcpdump -i any -nn host 10.128.3.208
tcpdump: data link type LINUX_SLL2
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on any, link-type LINUX_SLL2 (Linux cooked v2), snapshot length 262144 bytes
10:03:53.231552 c7ac4f22d275a25 P   IP 10.131.0.123.49084 > 10.128.3.208.8080: Flags [S], seq 3824685628, win 64240, options [mss 1460,sackOK,TS val 4059084599 ecr 0,nop,wscale 7], length 0
10:03:53.232215 enp2s0 Out IP 10.131.0.123.49084 > 10.128.3.208.8080: Flags [S], seq 3824685628, win 64240, options [mss 1460,sackOK,TS val 4059084599 ecr 0,nop,wscale 7], length 0
10:03:53.233434 enp2s0 In  IP 10.128.3.208.8080 > 10.131.0.123.49084: Flags [S.], seq 1192655212, ack 3824685629, win 65160, options [mss 1460,sackOK,TS val 1392374174 ecr 4059084599,nop,wscale 7], length 0
10:03:53.233873 c7ac4f22d275a25 Out IP 10.128.3.208.8080 > 10.131.0.123.49084: Flags [S.], seq 1192655212, ack 3824685629, win 65160, options [mss 1460,sackOK,TS val 1392374174 ecr 4059084599,nop,wscale 7], length 0
10:03:53.233900 c7ac4f22d275a25 P   IP 10.131.0.123.49084 > 10.128.3.208.8080: Flags [.], ack 1, win 502, options [nop,nop,TS val 4059084602 ecr 1392374174], length 0
10:03:53.234176 enp2s0 Out IP 10.131.0.123.49084 > 10.128.3.208.8080: Flags [.], ack 1, win 502, options [nop,nop,TS val 4059084602 ecr 1392374174], length 0
10:03:53.234237 c7ac4f22d275a25 P   IP 10.131.0.123.49084 > 10.128.3.208.8080: Flags [P.], seq 1:82, ack 1, win 502, options [nop,nop,TS val 4059084602 ecr 1392374174], length 81: HTTP: GET / HTTP/1.1
10:03:53.234250 enp2s0 Out IP 10.131.0.123.49084 > 10.128.3.208.8080: Flags [P.], seq 1:82, ack 1, win 502, options [nop,nop,TS val 4059084602 ecr 1392374174], length 81: HTTP: GET / HTTP/1.1
10:03:53.234753 enp2s0 In  IP 10.128.3.208.8080 > 10.131.0.123.49084: Flags [.], ack 82, win 509, options [nop,nop,TS val 1392374176 ecr 4059084602], length 0
10:03:53.234767 c7ac4f22d275a25 Out IP 10.128.3.208.8080 > 10.131.0.123.49084: Flags [.], ack 82, win 509, options [nop,nop,TS val 1392374176 ecr 4059084602], length 0
10:03:53.234975 enp2s0 In  IP 10.128.3.208.8080 > 10.131.0.123.49084: Flags [P.], seq 1:157, ack 82, win 509, options [nop,nop,TS val 1392374176 ecr 4059084602], length 156: HTTP: HTTP/1.1 200 OK
10:03:53.234979 c7ac4f22d275a25 Out IP 10.128.3.208.8080 > 10.131.0.123.49084: Flags [P.], seq 1:157, ack 82, win 509, options [nop,nop,TS val 1392374176 ecr 4059084602], length 156: HTTP: HTTP/1.1 200 OK

```

pliurh: right, we should capture on the physic interface. It can work for both LGW and SGW.

test/e2e/no_overlay.go
zhaozhanqi: Thanks for catching this

I updated with `kill -0 $PID` to check it's ready

test/e2e/no_overlay.go
zhaozhanqi: @pliurh  Updated tcpdump to use physical interface.

zhaozhanqi: and also updated other place according to your comment. Would like to take a look?

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,400 @@
...
+					IPFamilyPolicy: &familyPolicy,
+				},
+			}
+			serverService, err = f.ClientSet.CoreV1().Services(f.Namespace.Name).Create(context.TODO(), serverService, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "Failed to create server service")
+			framework.Logf("Created service %s with ClusterIPs %v", serverService.Name, serverService.Spec.ClusterIPs)
+
+			ginkgo.By("Creating tcpdump pod")
+			tcpdumpPod, err = createPod(f, tcpdumpPodName, nodes.Items[1].Name, f.Namespace.Name,
+				[]string{"sh", "-c", "apk add --no-cache tcpdump || (apt-get update && apt-get install -y tcpdump) || true; sleep 20000"},
```
pliurh: nit: Can we use the `nicolaka\netshoot` image which already includes the tcpdump command?

zhaozhanqi: ok, updated the image to 'ghcr.io/nicolaka/netshoot:v0.13'

pliurh (dismissed):

test/e2e/images/images.go
```diff
@@ -15,7 +15,8 @@ var (
 	// pre-approve new images.
 	agnHost = image.GetE2EImage(image.Agnhost)
 	// FIXME: iperf3 image should not be retrieved from a users repo and should not have latest tag
-	iperf3 = "quay.io/sronanrh/iperf:latest"
+	iperf3   = "quay.io/sronanrh/iperf:latest"
+	netshoot = "ghcr.io/nicolaka/netshoot:v0.13"
```
tssurya: what specifically in netshoot is it that's required which isn't present in agnhost or iperf?

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+		if !isNoOverlayEnabled() {
+			ginkgo.Skip("Test requires no-overlay mode to be enabled")
+		}
+
+		ginkgo.By("Getting all nodes and their pod subnets")
+		nodes, err := f.ClientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		framework.ExpectNoError(err, "Failed to list nodes")
+		gomega.Expect(len(nodes.Items)).To(gomega.BeNumerically(">=", 2), "Test requires at least 2 nodes")
+
+		ginkgo.By("Verifying pod subnet routes are in GR but not in ovn_cluster_router for each node")
```
tssurya: this is not an e2e test, please make this a unit test - I'm surprised if there isn't a unit test already for this.
all "does things exist in ovn db" can be achieved using libovsdb test harness framework
an end to end test is something that doesn't rely on implementation details but checks out a traffic flow
if there are things you want to wait to be programmed etc we can talk about that but unsure if this is that.

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	e2epodoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
+	utilnet "k8s.io/utils/net"
+)
+
+var _ = ginkgo.Describe("No-Overlay: Default network is enabled with no-overlay", feature.NoOverlay, func() {
+	f := wrappedTestFramework("no-overlay-default-network")
+
+	ginkgo.It("should have pod subnet routes in GR but not in ovn_cluster_router for no-overlay", func() {
```
tssurya: what is the end to end goal of this test really?

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+		var nodes *corev1.NodeList
+
+		const (
+			tcpdumpPodName = "tcpdump-pod-no-overlay"
+			serverPodName  = "server-pod-no-overlay"
+			clientPodName  = "client-pod-no-overlay"
+		)
+
+		ginkgo.BeforeEach(func() {
+			if !isNoOverlayEnabled() {
```
tssurya: so in today's no-overlay lanes what are all the tests running @pliurh ?

1. all control plane tests except what's skipped in https://github.com/ovn-kubernetes/ovn-kubernetes/commit/5db3201f88b78730fd5942d6b8443fa2a33cdb0f#diff-2103f6de0817ab13a12e0d43d074691db2726cd51219315dcf56f835a901063c
2. all BGP tests ?

that's too much duplication and two full lanes for that? did we consider running no-overlay mode tests on BGP lanes itself? what's the bottleneck to doing that if we did enable some of the flags we need for no-overlay on RA i.e not all of the BGP we have close to 4 BGP lanes - just thinking why we didn't repurpose some of them into no-overlay lanes and for a single PR presubmits we keep adding all new lanes per feature (note for evpn we had similar conversations - at first we wanted to run them in bgp lanes w/o new lane and then the lgw lane of bgp was too big and then we wanted to run evpn tests in parallel mode so we stuck to one lane)
trying to understand why we did new lanes and didn't consolidate in existing lanes cc @trozet - maybe this is a topic for u/s meeting since we are now close to 60 lanes aren't we? and ofc unrelated to this PR cause lanes already merged..

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+				ginkgo.Skip("Test requires no-overlay mode to be enabled (OVN_NO_OVERLAY_ENABLE=true)")
```
tssurya: shouldn't feature label be leveraged via e2e-cp.sh to do a trigger for no-overlay and skip it elsewhere? I think that's what other features do - if I run this locally its going to depend on me setting the env var locally - having said that this is a NIT since there are tests where we do skip lgw v/s sgw in similar fashion but that's on a specific issue / caseby case basis- its just weird to do it for a full feature because then all no-overlay describe contexts will need this skip v/s doing it from one place like skip file is better

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+						otherNode.Name, otherPodCIDR, node.Name)
+					// - Routes to remote pod subnets SHOULD exist in each node's GR (imported from BGP)
+					// - Routes to remote pod subnets should NOT exist in ovn_cluster_router (BGP handles routing)
+					err := checkOVNLogicalRouterRoutes(otherPodCIDR, node.Name, true, false)
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				}
+			}
+		}
+	})
+	ginkgo.When("connectivity tests", func() {
```
tssurya, line 28: I hope when the UDN support is added - its not new context but just a ginkgo.Entry to same tests - let's not duplicate code over

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+			framework.Logf("Server pod IPs: %v", serverPod.Status.PodIPs)
+			framework.Logf("Client pod IPs: %v", clientPod.Status.PodIPs)
+
+			// Verify pods have IP addresses
+			gomega.Expect(len(serverPod.Status.PodIPs)).To(gomega.BeNumerically(">", 0), "Server pod should have at least one IP address")
+			gomega.Expect(len(clientPod.Status.PodIPs)).To(gomega.BeNumerically(">", 0), "Client pod should have at least one IP address")
+
+			ginkgo.By("Creating service to select server pod")
+			familyPolicy := corev1.IPFamilyPolicyPreferDualStack
+			serverService = &corev1.Service{
```
tssurya: why are we not using the existing e2e utils for things like sevice create?

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+					},
+					Type:           corev1.ServiceTypeClusterIP,
+					IPFamilyPolicy: &familyPolicy,
+				},
+			}
+			serverService, err = f.ClientSet.CoreV1().Services(f.Namespace.Name).Create(context.TODO(), serverService, metav1.CreateOptions{})
+			framework.ExpectNoError(err, "Failed to create server service")
+			framework.Logf("Created service %s with ClusterIPs %v", serverService.Name, serverService.Spec.ClusterIPs)
+
+			ginkgo.By("Creating tcpdump pod")
```
tssurya, line 93: why do we need a tcpdump pod? that's not a normal way of writing an e2e - its an implementation detail that you are testing?
please add a comment for why these tests need a tcpdump pod

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+						Capabilities: &corev1.Capabilities{
+							Add: []corev1.Capability{"NET_RAW", "NET_ADMIN"},
+						},
+					}
+				})
+			framework.ExpectNoError(err)
+			framework.Logf("tcpdumpPod pod IPs: %v", tcpdumpPod.Status.PodIPs)
+		})
+
+		ginkgo.It("should maintain pod2pod/pod2service/host2pod/host2service connectivity without overlay before and after ovnpod restarted", func() {
```
tssurya: what's "ovnpod" in this context?
ovn pods?
or ovnkube-node pod?
please don't refer OVN pod as ovnkube pod.. they are two separate projects - we mentioned in u/s meeting about the importance of consistency for naming...

tssurya:
> /assign @jcaamano PTAL, thanks

I think this one is on me @zhaozhanqi not Jaime..
I'm reviewing it.

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+			ovnPodList, err := f.ClientSet.CoreV1().Pods(ovnNamespace).List(context.TODO(), metav1.ListOptions{
+				LabelSelector: "app=ovnkube-node",
+				FieldSelector: fmt.Sprintf("spec.nodeName=%s", nodes.Items[0].Name),
+			})
+			framework.ExpectNoError(err, "Failed to list ovnkube-node pods")
+			gomega.Expect(ovnPodList.Items).NotTo(gomega.BeEmpty(), "Should find ovnkube-node pod")
+			ovnPod := &ovnPodList.Items[0]
+			framework.Logf("Found ovnkube-node pod: %s on node %s", ovnPod.Name, nodes.Items[0].Name)
+
+			ginkgo.By("Deleting ovnkube-node pod to trigger restart")
```
tssurya, line 137: do we have basic tests already for non-restart scenario? is that covered by the control plane tests running?
if so please update your describe ginkgo.When("connectivity tests" to ginkgo.When("connectivity tests for restart scenario"

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,390 @@
...
+		})
+
+		ginkgo.It("should maintain pod2pod/pod2service/host2pod/host2service connectivity without overlay before and after ovnpod restarted", func() {
+			// test traffic for pod2pod, host2pod, pod2service, host2service and verify no overlay traffic is captured by tcpdump
+			ginkgo.By("Testing pod2pod connectivity without overlay on different node before OVN pod restart")
+			checkPod2PodConnectivityWithoutOverlay(serverPod, clientPod, tcpdumpPod)
+
+			ginkgo.By("Testing host2pod connectivity without overlay on different node before OVN pod restart")
+			// here use tcpdumpPod as the client (since it's host networked)
+			checkPod2PodConnectivityWithoutOverlay(serverPod, tcpdumpPod, tcpdumpPod)
```
tssurya: I assume for same node case the traffic flow is the same for nooverlay v/s overlay right?
just that different node case takes the underlay for nooverlay v/s overlay for normal case?

test/e2e/util.go
```diff
@@ -2071,3 +2072,32 @@ func waitForNodeReadyState(f *framework.Framework, nodeName string, timeout time
 		return false
 	}, timeout, 10*time.Second).Should(gomega.BeTrue(), expectationMessage)
 }
+
+// isNoOverlayEnabled checks if no-overlay mode is enabled for the default network
+func isNoOverlayEnabled() bool {
```
tssurya: nit: put this closer to the utils for other features that check ENV var

test/e2e/images/images.go
zhaozhanqi: both agnhost and iperf don't have `tcpdump`

test/e2e/no_overlay.go
zhaozhanqi: 1 and 2 yes. 

We cannot run no-overlay cases in BGP lanes because we don't support enable no-overlay in day2 for default network

test/e2e/no_overlay.go
zhaozhanqi: I know there is a way to skip feature test in e2e-cp.sh. However we still need this check for D/S test in future when OTE is ready.

test/e2e/no_overlay.go
zhaozhanqi: Strictly speaking, it's not a e2e case. 
However it's one easy way to prove that all nodes are in a normal state for no-overlay mode without needing to create a pod on each node and test the traffic.  
and also this case is from @pliurh 's suggestion. Maybe Peng had more thought?

test/e2e/no_overlay.go
zhaozhanqi: will add comments as below
```
			// Create tcpdump pod as host networked pod to capture traffic on the physical interface
			// In no-overlay mode pod IPs are routed directly and appear unencapsulated on the physical NIC.
```

test/e2e/no_overlay.go
zhaozhanqi: > do we have basic tests already for non-restart scenario? is that covered by the control plane tests running? if so please update your describe ginkgo.When("connectivity tests" to ginkgo.When("connectivity tests for restart scenario"

I checked this before ovnkube pod restart, so make both two check ( before and after restart ovnkube pod) into this one case
```
ginkgo.By("Testing pod2service connectivity without overlay via service IPs before OVN pod restart")
			checkPod2ServiceConnectivityWithoutOverlay(serverService, serverPod, clientPod, tcpdumpPod)
```

test/e2e/no_overlay.go
zhaozhanqi: yes

test/e2e/no_overlay.go
zhaozhanqi: updated to ovnkube-node pod

test/e2e/no_overlay.go
zhaozhanqi: updated it, thanks

test/e2e/no_overlay.go
pliurh: We can enable no-overlay in day2. But it requires rebooting ovn-k pods. We cannot turn on no-overlay for the default network while other tests are running.

> we keep adding all new lanes per feature

I think it's unavoidable in some way. We need the CI coverage for feature combinations where implementations are not strictly orthogonal and may interact.

test/e2e/no_overlay.go
pliurh: The GR assertion (routes learned by FRR and imported into OVN GR) can only be tested E2E since it requires a real BGP session and FRR running. We also need to verify the expected data path, not just connectivity.

You're right that the "not in ovn_cluster_router" assertion is OVN controller behavior and should be a unit test. We don't have the UT coverage for that yet; I will add it.

test/e2e/util.go
zhaozhanqi: Moved that.

test/e2e/no_overlay.go
tssurya: > The GR assertion (routes learned by FRR and imported into OVN GR) can only be tested E2E since it requires a real BGP session and FRR running. We also need to verify the expected data path, not just connectivity.

for testing import works - i.e here its just pure FRR no ovnkube magic except copying routes from vrf/host to GR in sgw - the key is connectivity works i.e if the north-south is working then its possible it works only if the routers are present in GR right?
adding implementation detail level testing seems a bit odd to me and I am trying to assess it there is something the connectivity check to the external server won't catch here? (because only if the import worked would the traffic flow) - in some sense we seem to be testing FRR here, the part around "routes getting copied into GR from the host hopefully already has UTs, if not they should get added as well"

test/e2e/no_overlay.go
tssurya: let's add a comment with these details to the test.yml file for why we did new lanes

zhaozhanqi: add the comments in new commit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5590/changes/de09234cb6e2667c38f29c2c9f4f6f1e8b050c32

test/e2e/no_overlay.go
tssurya: no downstream details shouldn't cloud the upstream work - See how network segmentation and other features do it - use the same format for skipping

test/e2e/no_overlay.go
tssurya: I still think this test is not an E2E. The import is implicitly tested..please remove this commit unless I am mistaken about something..

test/e2e/no_overlay.go
tssurya: no that's what I was about to mean  so if the shard conformance tests already cover the po2pod etc then are we duplicating it is what I meant
even if you do check the connectivity before restart I think we do it more for completion.. but ok let's just leave things as is. all good.

test/e2e/util.go
```diff
@@ -1462,6 +1463,22 @@ func isPreConfiguredUdnAddressesEnabled() bool {
...
+	ovnKubeNamespace := deploymentconfig.Get().OVNKubernetesNamespace()
+	configData, err := e2ekubectl.RunKubectl(ovnKubeNamespace,
+		"get", "configmap/ovnkube-config", "-o=jsonpath={.data.ovnkube\\.conf}")
+	if err != nil {
+		framework.Logf("isNoOverlayEnabled: failed to read ovnkube-config: %v", err)
+		return false
+	}
+	// Match: transport = no-overlay for upstream and also
+	// matches: transport="no-overlay" for openshift downstreams
```
tssurya: please revert this util to how it was before..
don't bring downstream logic to upstream , use the ENV_VAR like before.. I'm not sure if there was some miscommunication in my review comment

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/images"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
+	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+	e2epodoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
+	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
```
tssurya, line 21: this import should be in the next commit...

test/e2e/util.go
zhaozhanqi: > I'm not sure if there was some miscommunication in my review comment

no, I updated this because the ENV_VAR is not setup in downstream when I used this case testd on downstream cluster.  So here I updated to check ovnkube-config configMap.  

> don't bring downstream logic to upstream

Checking configMap if no-overlay is enabled for default network can work on both upstream and downstream

test/e2e/no_overlay.go
zhaozhanqi: hmm, I saw BGP cases had that
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/route_advertisements.go#L86-L88

test/e2e/no_overlay.go
pliurh: > routes getting copied into GR from the host hopefully already has UTs

Yes

Ok, let's remove the test from e2e.

zhaozhanqi: Thanks, I removed it

zhaozhanqi:
@tssurya I fixed the comments, could you help review again? thanks.

test/e2e/no_overlay.go
tssurya: cc @asood-rh

test/e2e/no_overlay.go
tssurya: hmm but its for a subset of the feature flag not the full isRAEnabled?- that's because BGP has multiple lanes with default network advertised v/s not.... normally for the full feature which is what you are doing - the path is to enable the no overlay only in the lanes where target is nooverlay..does that make sense? so there is no need to skip on test level because its only enabled in the lanes its supposed to be enabled..check out the invocation path for no-overlay tests...

test/e2e/util.go
tssurya: So the thing is it strictly required for upstream or not?
my point is we can't bring in logic just because "something doesn't work downstream so let's make a common util"
since we use ENV VAR for upstream  and all other utils do that we should do the same .. - also see my comment around how the skip is being done: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5590#discussion_r2947270095

for downstream the method to skip/unskip can be done using the feature labels itself..but let's keep that seperate from upstream logic. I don't think we need this util right? because no overlay tests are not enabled in other lanes upstream anyways?

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,285 @@
...
+	collectCmd := "sh -c 'kill -INT $(cat /tmp/tcpdump.pid) >/dev/null 2>&1 || true; sleep 1; cat /tmp/tcpdump.log'"
+	tcpdumpOut, err := e2epodoutput.RunHostCmdWithRetries(tcpdumpPod.Namespace, tcpdumpPod.Name, collectCmd, framework.Poll, 10*time.Second)
+	framework.ExpectNoError(err, "Failed to collect tcpdump output")
+	framework.Logf("tcpdump output:\n%s", tcpdumpOut)
+	return tcpdumpOut, curlOutput
+}
+
+// checkPod2PodConnectivityWithoutOverlay checks that the client pod can connect to the server pod without overlay
+// by verifying the pod IP is visible unencapsulated on the wire.
+func checkPod2PodConnectivityWithoutOverlay(serverPod, clientPod, tcpdumpPod *corev1.Pod) {
```
tssurya: is there a reason to have two seperate utils or can we just pass in a set of IPs (doesn't matter if its podIPs or serviceIPs ) into a single util? i.e can we combine these two utils to avoid code duplication?

test/e2e/util.go
zhaozhanqi: hmm, IMO, as we are trying to enable upstream cases to work in D/S by OTE, we should consider the compatibility, right?

I'm not saying bring downstream logic to upstream.  just checking configmap is the also correct way for upstream, I'm not sure if we have to use ENV VAR to check here? 

and for downstream, I don't know what's you mean the method to skip/unskip using the feature label?   do you mean add it in rules.go as below:

add no-overlay feature, but it will skip all cluster even on no-overlay cluster, that's not expected 

```
  LabelToLabelMaps = map[string][]string{                                                                                                                                                                                                                                                  
      "[Disabled:Unimplemented]": {                                                                                                                                                                                                                                                        
          `[Feature:Service]`,                                                                                                                                                                                                                                                             
          `[Feature:NetworkPolicy]`,                                                                                                                                                                                                                                                       
          // ... other features ...                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
      },                                                                                                                                                                                                                                                                                   
  }                                       
```
correct me if I'm wrong, I will slack you for future discusstion, thanks.

test/e2e/no_overlay.go
zhaozhanqi: for upstream yes, we can easy skip by e2e-cp.sh?  I added this because considering this case will be tested to d/s, I will slack you for more discusstion, thanks

test/e2e/no_overlay.go
zhaozhanqi: I updated it and combine with one function, thanks

test/e2e/util.go
tssurya: > hmm, IMO, as we are trying to enable upstream cases to work in D/S by OTE, we should consider the compatibility, right?

no because upstream has no dependencies for d/s work right? how does OTE work belong in upstream repo?

> I'm not saying bring downstream logic to upstream. just checking configmap is the also correct way for upstream, I'm not sure if we have to use ENV VAR to check here?

the pattern so far has been to check the ENV VAR right? - isn't that how its done upstream?
but more so my point is we don't even need this util and can use the e2e-cp.sh file to trigger no-overlay tests when no-overlay is enabled in CI. use the existing patterns of RA/UDN in how they start the tests in those lanes

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// checkConnectivityWithoutOverlay verifies no-overlay connectivity from clientPod and asserts
+// that traffic is visible unencapsulated on the physical NIC via tcpdump.
+//
+// serverPodIPs is always the real server pod IPs; it is used both as the curl destination
+// (pod-to-pod) and as the tcpdump filter (the actual endpoint seen on the wire).
+// When serviceClusterIPs is non-nil the curl destination switches to those VIPs instead
+// (pod-to-service), while tcpdump still filters on the server pod IP backing the VIP.
+func checkConnectivityWithoutOverlay(serverPodIPs []corev1.PodIP, serviceClusterIPs []string, clientPod, tcpdumpPod *corev1.Pod) {
```
tssurya, line 236: hmm why do we need to distinguish between serviceCIPs and serverpodIPs here? both end up being "destination IPs right?"

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		gomega.Expect(filterIP).NotTo(gomega.BeEmpty(), "Could not find server pod IP matching family of %s", destIP)
+
+		ginkgo.By(fmt.Sprintf("curl %s", destIP))
+		curlCmd := fmt.Sprintf("curl -s -m 2 %s/clientip", net.JoinHostPort(destIP, fmt.Sprint(netexecPort)))
+		// In no-overlay mode pod IPs are routed directly; capturing on the physical NIC with a
+		// pod IP filter proves the traffic is unencapsulated (overlay would hide it inside Geneve).
+		tcpdumpOut, _, curlErr := getTcpdumpOnPhysicalIface(tcpdumpPod, clientPod, curlCmd, filterIP)
+		framework.ExpectNoError(curlErr, "curl to %s failed", destIP)
+		gomega.Expect(tcpdumpOut).To(gomega.MatchRegexp(`(?m)^[1-9][0-9]* packets captured`),
```
tssurya, line 265: I want to say this method of E2E testing with capturing traffic is also not very E2E to me - but I won't hold the PR for this reason since I understand you guys are trying to ensure the traffic is really using no overlay with no encap and there isn't a neat way to replicate what tcpdump can provider with unit tests.
regardless there are chances of breakage when we do down to specifics of how we test things like this in future.
If we were really just testing E2E - we shouldn't be concerned with the implementation details,, but for now let's move on and I don't want to block the PR for that.

test/e2e/no_overlay.go
zhaozhanqi: For pod to service, except curl service_ip, we still need to capture the backend ip ( serverPodIPs) of service for tcpdump

test/e2e/util.go
zhaozhanqi: OK, I already removed this skip the cases and will check how to make this work in D/S, thanks.

zhaozhanqi:
@tssurya Hi, Could you kindly help check this PR again in your spare time, thanks

tssurya (approved):
I need to rebase on master since CI is 5 days old..
will merge this when the new run completes

tssurya:
```
2026-03-23T23:34:24.0302876Z [38;5;10m• [0.158 seconds][0m
2026-03-23T23:34:24.0303385Z [38;5;243m------------------------------[0m
2026-03-23T23:34:24.0305048Z [0m[1m[ReportAfterSuite] Autogenerated ReportAfterSuite for --junit-report[0m
2026-03-23T23:34:24.0306168Z [38;5;243mautogenerated by Ginkgo[0m
2026-03-23T23:34:24.3741332Z [38;5;10m[ReportAfterSuite] PASSED [0.343 seconds][0m
2026-03-23T23:34:24.3741998Z [38;5;243m------------------------------[0m
2026-03-23T23:34:24.3742304Z 
2026-03-23T23:34:24.3742557Z [38;5;9m[1mSummarizing 1 Failure:[0m
2026-03-23T23:34:24.3744658Z   [38;5;9m[FAIL][0m [0mOVN NetworkPolicy Operations [38;5;243mduring execution [38;5;9m[1m[It] correctly creates, updates and deletes network policies[0m
2026-03-23T23:34:24.3746063Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/ovn_test.go:262[0m
2026-03-23T23:34:24.3746533Z 
2026-03-23T23:34:24.3770569Z [38;5;9m[1mRan 552 of 555 Specs in 917.515 seconds[0m
2026-03-23T23:34:24.3774443Z [38;5;9m[1mFAIL![0m -- [38;5;10m[1m551 Passed[0m | [38;5;9m[1m1 Failed[0m | [38;5;11m[1m3 Pending[0m | [38;5;14m[1m0 Skipped[0m
2026-03-23T23:34:24.3777140Z --- FAIL: TestClusterNode (917.88s)
2026-03-23T23:34:24.3778182Z FAIL
2026-03-23T23:34:24.4085134Z coverage: 70.0% of statements
2026-03-23T23:34:25.4158398Z FAIL	github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn	921.998s
2026-03-23T23:34:25.4403267Z FAIL
2026-03-23T23:34:25.6161803Z make: *** [Makefile:65: check] Error 1
2026-03-23T23:34:25.6185344Z ##[error]Process completed with exit code 2.
2026-03-23T23:34:25.6320973Z Post job cleanup.
2026-03-23T23:34:25.7955580Z [command]/usr/bin/git version
2026-03-23T23:34:25.8004373Z git version 2.53.0
2026-03-23T23:34:25.8058226Z Temporarily overriding HOME='/home/runner/work/_temp/b2ac
```
failure seems unrelated to this PR because we only change e2e tests.. I will open a flake for this

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6156

test/e2e/util.go
```diff
@@ -2080,3 +2080,4 @@ func waitForNodeReadyState(f *framework.Framework, nodeName string, timeout time
 		return false
 	}, timeout, 10*time.Second).Should(gomega.BeTrue(), expectationMessage)
 }
+
```
tssurya, line 2083: NIT:  hmmm what's this new line for? please try not to add unwanted new lines next time

test/e2e/no_overlay.go
```diff
@@ -0,0 +1,268 @@
...
+				// Check if all containers are ready
+				for _, condition := range newOvnkubeNodePod.Status.Conditions {
+					if condition.Type == corev1.PodReady && condition.Status == corev1.ConditionTrue {
+						framework.Logf("New ovnkube-node pod %s is ready", newOvnkubeNodePod.Name)
+						return true
+					}
+				}
+				framework.Logf("New ovnkube-node pod %s is not ready yet", newOvnkubeNodePod.Name)
+				return false
+			}, 120*time.Second, 2*time.Second).Should(gomega.BeTrue(), "New ovnkube-node pod should be ready within 120 seconds")
```
tssurya, line 170: NIT: I guess its 2mins for total timeout... I wonder if thats too much wait..

test/e2e/util.go
zhaozhanqi: maybe typo, will remove in next PR. thanks.

test/e2e/no_overlay.go
zhaozhanqi: yeah,  I tested in my local and averaging around 20 seconds, so I can update to 60s in next.

test/e2e/no_overlay.go
asood-rh: @tssurya Missed this comment, yes not duplicating the code for CUDN, just new ginkgo.Entry

PR #5595, opened 2025-09-25
Allow specifying custom ClusterIP for Services

trozet:
Introduces new behavior, where if a Headless Service is created with k8s.ovn.org/cluster-ip annotated with an custom IP address, then a service will be created using that custom IP address.

The custom IP address may not be in the range of the Kubernetes Service CIDR.

Additionally, the load balancer is not placed on the Gateway Router (GR). This is intentional to avoid exposing this Custom IP address anywhere other than inside the SDN. The load balancer is only configured on the worker switches.

**Why would you want this?**

With VMs (kubevirt) there exists a case where a VM may need to cloud-init to get information to bootstrap a VM. Usually it is recommended in kubevirt to use a noseed approach, where a volume is mounted for cloud-init to inspect and find the data:
https://github.com/kubevirt/kubevirt/blob/9c2c41e64a0979cc3a5920729f9ae8ee03affeb5/examples/vmi-masquerade.yaml#L38

However, there may be cases where a VM needs to access a metadata server, similar to the standard convention AWS and OpenStack use of web access to 169.254.169.254. For this use case, we can create a metadata service to handle this. However, we cannot specify 169.254.169.254 for LoadBalancer IP or External IP, Kubernetes forbids it:

```
trozet@fedora:~$ oc create -f ~/externalip_service_udn.yaml The Service "metadata-service" is invalid: spec.externalIPs[0]: Invalid value: "169.254.169.254": may not be in the link-local range (169.254.0.0/16, fe80::/10)
```

So with this commit, we can leverage it as a cluster IP, and use ITP=Local so that VMs access a metadata service on their own node:

```
trozet@fedora:~/go/src/github.com/ovn-org/ovn-kubernetes/go-controller$ cat ~/metadata_service_udn.yaml 
apiVersion: v1
kind: Service
metadata:
  name: metadata-service
  namespace: udn-test
  annotations:
    k8s.ovn.org/cluster-ip: 169.254.169.254 
spec:
  clusterIP: None
  selector:
      pod-name: metadata-server      
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
      name: blah2  
  type: ClusterIP
  internalTrafficPolicy: Local
```

@tssurya @npinaeva @jcaamano looking for feedback on this idea. It probably needs a small okep and and e2e test as well.

### DIFF
--- a/go-controller/pkg/factory/factory.go
+++ b/go-controller/pkg/factory/factory.go
@@ -1814,22 +1814,6 @@ func withServiceNameAndNoHeadlessServiceSelector() func(options *metav1.ListOpti
 	}
 }
 
-// noHeadlessServiceSelector returns a LabelSelector (added to the
-// watcher for EndpointSlices) that will only choose EndpointSlices without "service.kubernetes.io/headless"
-// label.
-func noHeadlessServiceSelector() func(options *metav1.ListOptions) {
-	// headless service label must not be there
-	noHeadlessService, err := labels.NewRequirement(corev1.IsHeadlessService, selection.DoesNotExist, nil)
-	if err != nil {
-		// cannot occur
-		panic(err)
-	}
-
-	return func(options *metav1.ListOptions) {
-		options.LabelSelector = noHeadlessService.String()
-	}
-}
-
 // noAlternateProxySelector is a LabelSelector added to the watch for
 // services that excludes services with a well-known label indicating
 // proxying is via an alternate proxy.
@@ -1897,11 +1881,10 @@ func waitForCacheSyncWithTimeout(factory waitForCacheSyncer, stopCh <-chan struc
 // When network segmentation is enabled it returns a selector that ignores EndpointSlices for headless services.
 // Otherwise, it returns a selector that excludes EndpointSlices a with missing default service name too.
 func getEndpointSliceSelector() func(options *metav1.ListOptions) {
-	endpointSliceSelector := withServiceNameAndNoHeadlessServiceSelector()
 	if util.IsNetworkSegmentationSupportEnabled() {
 		// When network segmentation is enabled we need to watch for mirrored EndpointSlices that do not contain the
 		// default service name.
-		endpointSliceSelector = noHeadlessServiceSelector()
+		return nil
 	}
-	return endpointSliceSelector
+	return withServiceNameAndNoHeadlessServiceSelector()
 }
--- a/go-controller/pkg/ovn/controller/services/lb_config.go
+++ b/go-controller/pkg/ovn/controller/services/lb_config.go
@@ -270,7 +270,10 @@ func buildClusterLBs(service *corev1.Service, configs []lbConfig, nodeInfos []no
 			// The node may not have a gateway router - it might be waiting initialization, or
 			// might have disabled GWR creation via the k8s.ovn.org/l3-gateway-config annotation
 			if node.gatewayRouterName != "" {
-				nodeRouters = append(nodeRouters, node.gatewayRouterName)
+				// Do not set routerTargets for a custom service IP
+				if customIP := util.GetServiceCustomIP(service); customIP == "" {
+					nodeRouters = append(nodeRouters, node.gatewayRouterName)
+				}
 			}
 		}
 	}
@@ -527,16 +530,19 @@ func buildTemplateLBs(service *corev1.Service, configs []lbConfig, nodes []nodeI
 					Templates:   getTemplatesFromRulesTargets(switchV4Rules),
 				})
 			}
-			if len(routerV4Rules) > 0 {
-				out = append(out, LB{
-					Name:        makeLBNameForNetwork(service, proto, "node_router_template_IPv4", netInfo),
-					Protocol:    string(proto),
-					ExternalIDs: eids,
-					Opts:        optsV4,
-					Groups:      []string{netInfo.GetNetworkScopedLoadBalancerGroupName(types.ClusterRouterLBGroupName)},
-					Rules:       routerV4Rules,
-					Templates:   getTemplatesFromRulesTargets(routerV4Rules),
-				})
+			// Do not set routerTargets for a custom service IP
+			if customIP := util.GetServiceCustomIP(service); customIP == "" {
+				if len(routerV4Rules) > 0 {
+					out = append(out, LB{
+						Name:        makeLBNameForNetwork(service, proto, "node_router_template_IPv4", netInfo),
+						Protocol:    string(proto),
+						ExternalIDs: eids,
+						Opts:        optsV4,
+						Groups:      []string{netInfo.GetNetworkScopedLoadBalancerGroupName(types.ClusterRouterLBGroupName)},
+						Rules:       routerV4Rules,
+						Templates:   getTemplatesFromRulesTargets(routerV4Rules),
+					})
+				}
 			}
 		}
 
@@ -553,15 +559,18 @@ func buildTemplateLBs(service *corev1.Service, configs []lbConfig, nodes []nodeI
 				})
 			}
 			if len(routerV6Rules) > 0 {
-				out = append(out, LB{
-					Name:        makeLBNameForNetwork(service, proto, "node_router_template_IPv6", netInfo),
-					Protocol:    string(proto),
-					ExternalIDs: eids,
-					Opts:        optsV6,
-					Groups:      []string{netInfo.GetNetworkScopedLoadBalancerGroupName(types.ClusterRouterLBGroupName)},
-					Rules:       routerV6Rules,
-					Templates:   getTemplatesFromRulesTargets(routerV6Rules),
-				})
+				// Do not set routerTargets for a custom service IP
+				if customIP := util.GetServiceCustomIP(service); customIP == "" {
+					out = append(out, LB{
+						Name:        makeLBNameForNetwork(service, proto, "node_router_template_IPv6", netInfo),
+						Protocol:    string(proto),
+						ExternalIDs: eids,
+						Opts:        optsV6,
+						Groups:      []string{netInfo.GetNetworkScopedLoadBalancerGroupName(types.ClusterRouterLBGroupName)},
+						Rules:       routerV6Rules,
+						Templates:   getTemplatesFromRulesTargets(routerV6Rules),
+					})
+				}
 			}
 		}
 	}
@@ -665,7 +674,7 @@ func buildPerNodeLBs(service *corev1.Service, configs []lbConfig, nodes []nodeIn
 							Targets: targetsETP,
 						})
 					}
-					if cfg.internalTrafficLocal && util.IsClusterIP(vip) { // ITP only applicable to CIP
+					if cfg.internalTrafficLocal && util.IsClusterIP(vip, service) { // ITP only applicable to CIP
 						targetsITP := joinHostsPort(switchV4TargetIPs, cfg.clusterEndpoints.Port)
 						if isv6 {
 							targetsITP = joinHostsPort(switchV6TargetIPs, cfg.clusterEndpoints.Port)
@@ -698,7 +707,10 @@ func buildPerNodeLBs(service *corev1.Service, configs []lbConfig, nodes []nodeIn
 					if cfg.externalTrafficLocal && len(targets) > 0 {
 						noSNATRouterRules = append(noSNATRouterRules, rule)
 					} else {
-						routerRules = append(routerRules, rule)
+						// Do not set routerTargets for a custom service IP
+						if customIP := util.GetServiceCustomIP(service); customIP == "" {
+							routerRules = append(routerRules, rule)
+						}
 					}
 				}
 			}
--- a/go-controller/pkg/ovn/controller/services/lb_config_test.go
+++ b/go-controller/pkg/ovn/controller/services/lb_config_test.go
@@ -123,6 +123,7 @@ func Test_buildServiceLBConfigs(t *testing.T) {
 	inport1 := int32(81)
 	outport1 := int32(8081)
 	outportstr := intstr.FromInt(int(outport))
+	internalTrafficPolicy := corev1.ServiceInternalTrafficPolicyLocal
 
 	// make slices
 	// nil slice = don't use this family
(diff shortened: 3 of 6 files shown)


jcaamano:
maybe kubevirst is solving the use case somehow already? @maiqueb do you know?

tssurya:
> Introduces new behavior, where if a Headless Service is created with k8s.ovn.org/cluster-ip annotated with an custom IP address, then a service will be created using that custom IP address.
> 
> The custom IP address may not be in the range of the Kubernetes Service CIDR.

hmm https://kubernetes.io/docs/concepts/services-networking/service/#headless-services a headless service is something with no clusterIP, at least we might want to not call that a cluster-ip on the annotation and something more vague like a virtualip if we ever go down this route, I would only do annotation hacks if we really don't have another way out.

> 
> Additionally, the load balancer is not placed on the Gateway Router (GR). This is intentional to avoid exposing this Custom IP address anywhere other than inside the SDN. The load balancer is only configured on the worker switches.
> 
> **Why would you want this?**
> 
> With VMs (kubevirt) there exists a case where a VM may need to cloud-init to get information to bootstrap a VM. Usually it is recommended in kubevirt to use a noseed approach, where a volume is mounted for cloud-init to inspect and find the data: https://github.com/kubevirt/kubevirt/blob/9c2c41e64a0979cc3a5920729f9ae8ee03affeb5/examples/vmi-masquerade.yaml#L38
> 
> However, there may be cases where a VM needs to access a metadata server, similar to the standard convention AWS and OpenStack use of web access to 169.254.169.254. For this use case, we can create a metadata service to handle this. However, we cannot specify 169.254.169.254 for LoadBalancer IP or External IP, Kubernetes forbids it:
> 
> ```
> trozet@fedora:~$ oc create -f ~/externalip_service_udn.yaml The Service "metadata-service" is invalid: spec.externalIPs[0]: Invalid value: "169.254.169.254": may not be in the link-local range (169.254.0.0/16, fe80::/10)
> ```

hmm sounds like a common problem that kubevirt should solve?

tssurya:
cc @EdDev

maiqueb:
FWIW, while I see value in adding this as a network feature, I do agree that this is more a virtualization problem to solve. 

I think this metadata server belongs in KubeVirt. FWIW, I think it could be possible to just deploy https://github.com/jalseth/nocloud-metadata-server as a pod / service in the K8S cluster, and see it if works for this use case.

jcaamano:
I am not familiar with cloud init but are we somehow forced to serve it at 169.254.169.254 (or other link local)? I understand the openstack convention but is the configuration with which some VMs have already been built so strict? Or is there something else at play here?

trozet:
The 169.254.169.254 is a convention used by more than OpenStack. It's also used by every major cloud like AWS, GCP, Azure for accessing metadata. I think we should allow for this consistency as well.

 I don't see how the suggestion by @maiqueb solves this problem. To access a metadata service as a pod, you would need to be able to talk to 169.254.169.254, which you cannot do on the SDN without this fix. Other alternatives would be:

 1. Do something like OpenStack does with OVN, where the packet is DNAT'ed (same thing we are doing here) to an ovn-metadata-agent listening on the host. I think this doesn't work with UDNs, because the source IP will be SNAT'ed. Need to look into this a bit more to rule it out completely.
 
 https://docs.openstack.org/developer/networking-ovn/design/metadata_api.html#neutron-and-metadata-with-ovn
 
 2. Run a sidecar with the VM, managed by kubevirt, that acts as the metadata service. Then when the VM queries 169.254.169.254 it acts as the agent and proxies the request. If that proxy uses kube api to get the information it needs, then that is an issue, because for our use case we do not want the VM to have access to the kapi service.

So with this proposal even though the PUDN would talk to a metadata agent pod on the same PUDN, we can apply network policy to only allow the metadata agent to reach kube-api service, and deny it for all other VMs.

girishmg:
@jcaamano @maiqueb @tssurya the ask is simple

Once the VM boots, the user inside the VM doesn't know anything about how the VM came about to be.  At runtime, the user wants to know some metadata associated with the VM. It cannot be at the VM firstboot through cloud-init, because things/artifacts expires (think tokens, certificates, and so on). 

Furthermore, there could be multiple VMs on the same K8s node, and each of this VM can ask for that metadata information through the well-known IP 169.254.169.254.  This IP is a standard and like Tim mentions it is used by all CMS. In the picture below, Node-100 has 10 VMs and curl to 169.254.169.254 has to be routed to the Metadata-service Pod on the same node.

```
+---------------------------------------------------------------------------------+                 
|                                    Node-100                                     |                 
|                                                                                 |                 
|                                                                                 |                 
|                                         +----------+              +------------+|   +------------+
|                                         | MetaData |   +-------+  |            ||   |            |
|                                         | Service  |   |       |  | Service-X  ||   | Service-X  |
|  +----+  +----+         +--------+      |   Pod    +---+AF_UNIX+--+>  DS pOd   ||   |Manager Pod |
|  |VM-1|  |VM-2| ....... | VM-10  |      |          |   |       |  |            ||   |            |
|  +--+-+  +--+-+         +---+----+      +-----+----+   +-------+  +-----+------+|   +----+-------+
|     |       |               |                 |                         |       |        |        
|     |       |               |                 |                         |       |        |        
| +---v-------v---------------v-----------------v---+                +----v-------+--------v--+     
| |                    BLUE PUDN                    |                |          CDN           |     
| +-------------------------------------------------+                +------------+-----------+     
|                                                                                 |                 
+---------------------------------------------------------------------------------+                 
```

It's not possible to run metadata service as a side-car because of kubevirt networking. See below:

```
+----------------------------------------------------------------------+
|virt-laucnher Pod (with cluster-default network, primary-UDN,         |
|secondary-UDN)                                                        |
|                                  +------------------------------+    |
|                                  |              VM              |    |
|                                  |                              |    |
|                                  |       PUDN-IP                |    |
|                                  |   +-----------+    SUDN-IP   |    |
|                                  |   |   eth0    |    +-----+   |    |
|                                  +---+(virtio-FE)+----+ VF0 +---+    |
|                                      +-------+---+    +---+-+        |
|                                              |            |          |
|                                      +-------+----+       |          |
|                                      |   vnet0    |       |          |
|                        +-------------+(virtio-BE) +-+     |          |
|        CUDN-IP         |             +------------+ |     |          |
|     +-----------+      |            k6t-ovn-udn1 (L2|     |          |
|     | eth0 (VF) |      | +---------+   Linux bridge)|     |          |
|     +-----+-----+      | |ovn-udn1-|                |     |          |
|           |            +-+nic (VF) +----------------+     |          |
|           |              +----+----+                      |          |
+-----------+-------------------+---------------------------+----------+
            |                   |                           |           
    +-------v-------+     +-----+---------+          +------+--------+  
+---+VF-Representor +-----+VF-Representor +----------+VF-Representor +-+
|   +---------------+     +---------------+          +---------------+ |
|                                br-int                                |
+----------------------------------------------------------------------+

```

The PUDN interface inside the namespace is bridged and we can't intercept packets to 169.254.169.254 and send it to a local side car. Also, running metadata-service for every Kubevirt VM means that some mutating webhook as to inject the side car .. it is very expensive...


The simplest way to get this working in ovn-k8s CNI is what @trozet posts here. Other way to do is to use AdminPBR policy that will re-route the packets destined to 169.254.169.254 to a local pod running on the same node (attached to same PUDN). The AdminPBR should have to implement ITP=local. So, instead of duplicating things again in AdminPBR, K8s service with that annotation fits the bill.

maiqueb:
> @jcaamano @maiqueb @tssurya the ask is simple
> 
> Once the VM boots, the user inside the VM doesn't know anything about how the VM came about to be. At runtime, the user wants to know some metadata associated with the VM. It cannot be at the VM firstboot through cloud-init, because things/artifacts expires (think tokens, certificates, and so on).
> 
> Furthermore, there could be multiple VMs on the same K8s node, and each of this VM can ask for that metadata information through the well-known IP 169.254.169.254. This IP is a standard and like Tim mentions it is used by all CMS. In the picture below, Node-100 has 10 VMs and curl to 169.254.169.254 has to be routed to the Metadata-service Pod on the same node.
> 
> ```
> +---------------------------------------------------------------------------------+                 
> |                                    Node-100                                     |                 
> |                                                                                 |                 
> |                                                                                 |                 
> |                                         +----------+              +------------+|   +------------+
> |                                         | MetaData |   +-------+  |            ||   |            |
> |                                         | Service  |   |       |  | Service-X  ||   | Service-X  |
> |  +----+  +----+         +--------+      |   Pod    +---+AF_UNIX+--+>  DS pOd   ||   |Manager Pod |
> |  |VM-1|  |VM-2| ....... | VM-10  |      |          |   |       |  |            ||   |            |
> |  +--+-+  +--+-+         +---+----+      +-----+----+   +-------+  +-----+------+|   +----+-------+
> |     |       |               |                 |                         |       |        |        
> |     |       |               |                 |                         |       |        |        
> | +---v-------v---------------v-----------------v---+                +----v-------+--------v--+     
> | |                    BLUE PUDN                    |                |          CDN           |     
> | +-------------------------------------------------+                +------------+-----------+     
> |                                                                                 |                 
> +---------------------------------------------------------------------------------+                 
> ```
> 
> It's not possible to run metadata service as a side-car because of kubevirt networking. See below:
> 
> ```
> +----------------------------------------------------------------------+
> |virt-laucnher Pod (with cluster-default network, primary-UDN,         |
> |secondary-UDN)                                                        |
> |                                  +------------------------------+    |
> |                                  |              VM              |    |
> |                                  |                              |    |
> |                                  |       PUDN-IP                |    |
> |                                  |   +-----------+    SUDN-IP   |    |
> |                                  |   |   eth0    |    +-----+   |    |
> |                                  +---+(virtio-FE)+----+ VF0 +---+    |
> |                                      +-------+---+    +---+-+        |
> |                                              |            |          |
> |                                      +-------+----+       |          |
> |                                      |   vnet0    |       |          |
> |                        +-------------+(virtio-BE) +-+     |          |
> |        CUDN-IP         |             +------------+ |     |          |
> |     +-----------+      |            k6t-ovn-udn1 (L2|     |          |
> |     | eth0 (VF) |      | +---------+   Linux bridge)|     |          |
> |     +-----+-----+      | |ovn-udn1-|                |     |          |
> |           |            +-+nic (VF) +----------------+     |          |
> |           |              +----+----+                      |          |
> +-----------+-------------------+---------------------------+----------+
>             |                   |                           |           
>     +-------v-------+     +-----+---------+          +------+--------+  
> +---+VF-Representor +-----+VF-Representor +----------+VF-Representor +-+
> |   +---------------+     +---------------+          +---------------+ |
> |                                br-int                                |
> +----------------------------------------------------------------------+
> ```
> 
> The PUDN interface inside the namespace is bridged and we can't intercept packets to 169.254.169.254 and send it to a local side car. Also, running metadata-service for every Kubevirt VM means that some mutating webhook as to inject the side car .. it is very expensive...

Using bridge is a user choice. You are not required to use it. For instance, you can use **passt** to connect to primary UDN, which would remove this limitation (yes, bridge network bindings + sidecars doesn't work well. Or at all :sweat_smile: )

I recommend checking out this integration, which has high chances of not requiring any production code changes. Then, you can implement the metadata server in a sidecar, which is an orthogonal solution, upon customer's needs.

> 
> The simplest way to get this working in ovn-k8s CNI is what @trozet posts here. Other way to do is to use AdminPBR policy that will re-route the packets destined to 169.254.169.254 to a local pod running on the same node (attached to same PUDN). The AdminPBR should have to implement ITP=local. So, instead of duplicating things again in AdminPBR, K8s service with that annotation fits the bill.

maiqueb:
> > @jcaamano @maiqueb @tssurya the ask is simple
> > Once the VM boots, the user inside the VM doesn't know anything about how the VM came about to be. At runtime, the user wants to know some metadata associated with the VM. It cannot be at the VM firstboot through cloud-init, because things/artifacts expires (think tokens, certificates, and so on).
> > Furthermore, there could be multiple VMs on the same K8s node, and each of this VM can ask for that metadata information through the well-known IP 169.254.169.254. This IP is a standard and like Tim mentions it is used by all CMS. In the picture below, Node-100 has 10 VMs and curl to 169.254.169.254 has to be routed to the Metadata-service Pod on the same node.
> > ```
> > +---------------------------------------------------------------------------------+                 
> > |                                    Node-100                                     |                 
> > |                                                                                 |                 
> > |                                                                                 |                 
> > |                                         +----------+              +------------+|   +------------+
> > |                                         | MetaData |   +-------+  |            ||   |            |
> > |                                         | Service  |   |       |  | Service-X  ||   | Service-X  |
> > |  +----+  +----+         +--------+      |   Pod    +---+AF_UNIX+--+>  DS pOd   ||   |Manager Pod |
> > |  |VM-1|  |VM-2| ....... | VM-10  |      |          |   |       |  |            ||   |            |
> > |  +--+-+  +--+-+         +---+----+      +-----+----+   +-------+  +-----+------+|   +----+-------+
> > |     |       |               |                 |                         |       |        |        
> > |     |       |               |                 |                         |       |        |        
> > | +---v-------v---------------v-----------------v---+                +----v-------+--------v--+     
> > | |                    BLUE PUDN                    |                |          CDN           |     
> > | +-------------------------------------------------+                +------------+-----------+     
> > |                                                                                 |                 
> > +---------------------------------------------------------------------------------+                 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It's not possible to run metadata service as a side-car because of kubevirt networking. See below:
> > ```
> > +----------------------------------------------------------------------+
> > |virt-laucnher Pod (with cluster-default network, primary-UDN,         |
> > |secondary-UDN)                                                        |
> > |                                  +------------------------------+    |
> > |                                  |              VM              |    |
> > |                                  |                              |    |
> > |                                  |       PUDN-IP                |    |
> > |                                  |   +-----------+    SUDN-IP   |    |
> > |                                  |   |   eth0    |    +-----+   |    |
> > |                                  +---+(virtio-FE)+----+ VF0 +---+    |
> > |                                      +-------+---+    +---+-+        |
> > |                                              |            |          |
> > |                                      +-------+----+       |          |
> > |                                      |   vnet0    |       |          |
> > |                        +-------------+(virtio-BE) +-+     |          |
> > |        CUDN-IP         |             +------------+ |     |          |
> > |     +-----------+      |            k6t-ovn-udn1 (L2|     |          |
> > |     | eth0 (VF) |      | +---------+   Linux bridge)|     |          |
> > |     +-----+-----+      | |ovn-udn1-|                |     |          |
> > |           |            +-+nic (VF) +----------------+     |          |
> > |           |              +----+----+                      |          |
> > +-----------+-------------------+---------------------------+----------+
> >             |                   |                           |           
> >     +-------v-------+     +-----+---------+          +------+--------+  
> > +---+VF-Representor +-----+VF-Representor +----------+VF-Representor +-+
> > |   +---------------+     +---------------+          +---------------+ |
> > |                                br-int                                |
> > +----------------------------------------------------------------------+
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The PUDN interface inside the namespace is bridged and we can't intercept packets to 169.254.169.254 and send it to a local side car. Also, running metadata-service for every Kubevirt VM means that some mutating webhook as to inject the side car .. it is very expensive...
> 
> Using bridge is a user choice. You are not required to use it. For instance, you can use **passt** to connect to primary UDN, which would remove this limitation (yes, bridge network bindings + sidecars doesn't work well. Or at all 😅 )
> 
> I recommend checking out this integration, which has high chances of not requiring any production code changes. Then, you can implement the metadata server in a sidecar, which is an orthogonal solution, upon customer's needs.
> 
> > The simplest way to get this working in ovn-k8s CNI is what @trozet posts here. Other way to do is to use AdminPBR policy that will re-route the packets destined to 169.254.169.254 to a local pod running on the same node (attached to same PUDN). The AdminPBR should have to implement ITP=local. So, instead of duplicating things again in AdminPBR, K8s service with that annotation fits the bill.

Another alternative, which would address this outside OVN-K is to deploy a secondary network w/ IPAM, have its subnet 169.254.169.0/24, deploy a pod w/ the metadata service IP, and implement the service IP there. This pod would have access to the cluster default network via its cluster default network attachment, and would be subject to multi-network policies in case you need to limit it further. 

It wouldn't be able to do HA though, and the blast radius of the solution would be big (metadata service pod goes down, no metadata gets served in the cluster until another instance comes back up).

PR #5606, opened 2025-09-30
Fix stale EIP assignments during failover and controller restart

pperiyasamy:
When ovnkube-controller restart and an EgressIP (EIP) failover occur at the same time, this causes a race condition between event handling and informer cache update for EgressIP status. so while handling pod add event, controller sees older EIP status it fails to remove the SNAT and LRP configuration for the previously assigned node.

```
Nodes: node-1, node-2 and node-3
Egress IPs: EIP-1
Pods: pod1 (placed on node-1), pod2 (placed on node-3)
Egress Assignable Nodes: node-1 and node-2

Scenario (seen from the reproducer setup)
-------------------------------------------

EIP-1 assigned to node-1. node-1 and node-2 rebooted at the sametime.  
This made EIP failover and ovnkube-controller container restart happened almost at the same time.

1. EIP-1 is reassigned to node-2 by cluster manager.
2. EIP controller synchronizes EIP1 object with new object, but it cleans up SNATs and LRPs referring to node-1 due to stale pod IP addresses (due to pod recreation).
3. At the same time, pod1 and pod2 add events are triggered, but EIP controller's watch factory seeing older EIP status from the informer cache, so SNATs and LRPs are created referring to node-1 for the new pod IPs.
4. EIP-1 add event is triggered with new EIP status, EIP controller adds new SNAT entries and updates LRP nexthop for node-2.
5. stale SNATs, LRPs having with stale nexthops for node-1.

```

This PR leverages the `egressStatuses` stored in the `podAssignment` cache to reconcile and remove those stale entries correctly. Have also added an unit test to replicate above mentioned scenario.

## Summary by CodeRabbit

- **New Features**
  - Per-EIP/per-IP assignment tracking to improve visibility and reconciliation after topology changes.
  - Seeded "pending" status markers during sync to preserve assignment info while reconciliation runs.

- **Bug Fixes**
  - Cleanup of stale per-pod and per-EIP status entries to prevent misreporting.
  - More reliable EgressIP failover so SNAT and router nexthops update correctly during simultaneous failover and controller restart.

- **Tests**
  - Added large multi-node end-to-end failover test simulating controller restart; updated assertions and removed ad-hoc cache resets.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -789,25 +789,7 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 	if len(statusAssignments) == 0 {
 		return nil
 	}
-	// We need to proceed with add only under two conditions
-	// 1) egressNode present in at least one status is local to this zone
-	// (NOTE: The relation between egressIPName and nodeName is 1:1 i.e in the same object the given node will be present only in one status)
-	// 2) the pod being added is local to this zone
-	proceed := false
-	for _, status := range statusAssignments {
-		e.nodeZoneState.LockKey(status.Node)
-		isLocalZoneEgressNode, loadedEgressNode := e.nodeZoneState.Load(status.Node)
-		if loadedEgressNode && isLocalZoneEgressNode {
-			proceed = true
-			e.nodeZoneState.UnlockKey(status.Node)
-			break
-		}
-		e.nodeZoneState.UnlockKey(status.Node)
-	}
-	if !proceed && !e.isPodScheduledinLocalZone(pod) {
-		return nil // nothing to do if none of the status nodes are local to this master and pod is also remote
-	}
-	var remainingAssignments []egressipv1.EgressIPStatusItem
+	var remainingAssignments, staleAssignments []egressipv1.EgressIPStatusItem
 	nadName := ni.GetNetworkName()
 	if ni.IsUserDefinedNetwork() {
 		nadNames := ni.GetNADs()
@@ -843,8 +825,15 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 		// podState.egressIPName can be empty if no re-routes were found in
 		// syncPodAssignmentCache for the existing pod, we will treat this case as a new add
 		for _, status := range statusAssignments {
-			if exists := podState.egressStatuses.contains(status); !exists {
+			// Add the status if it's not already in the cache, or if it exists but is in pending state
+			// (meaning it was populated during EIP sync and needs to be processed for the pod).
+			if value, exists := podState.egressStatuses.statusMap[status]; !exists || value == egressStatusStatePending {
 				remainingAssignments = append(remainingAssignments, status)
+			}
+			// Detect stale EIP status entries (same EgressIP reassigned to a different node)
+			// and queue the outdated entry for cleanup.
+			if staleStatus := podState.egressStatuses.hasStaleEIPStatus(status); staleStatus != nil {
+				staleAssignments = append(staleAssignments, *staleStatus)
 			}
 		}
 		podState.podIPs = podIPs
@@ -866,6 +855,32 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 		)
 		podState.standbyEgressIPNames.Insert(name)
 		return nil
+	}
+	for _, staleStatus := range staleAssignments {
+		klog.V(2).Infof("Deleting stale pod egress IP status: %v for EgressIP: %s and pod: %s/%s/%v", staleStatus, name, pod.Namespace, pod.Name, podIPNets)
+		err = e.deletePodEgressIPAssignments(ni, name, []egressipv1.EgressIPStatusItem{staleStatus}, pod)
+		if err != nil {
+			klog.Warningf("Failed to delete stale EgressIP status %s/%v for pod %s: %v", name, staleStatus, podKey, err)
+		}
+		delete(podState.egressStatuses.statusMap, staleStatus)
+	}
+	// We need to proceed with add only under two conditions
+	// 1) egressNode present in at least one status is local to this zone
+	// (NOTE: The relation between egressIPName and nodeName is 1:1 i.e in the same object the given node will be present only in one status)
+	// 2) the pod being added is local to this zone
+	proceed := false
+	for _, status := range statusAssignments {
+		e.nodeZoneState.LockKey(status.Node)
+		isLocalZoneEgressNode, loadedEgressNode := e.nodeZoneState.Load(status.Node)
+		if loadedEgressNode && isLocalZoneEgressNode {
+			proceed = true
+			e.nodeZoneState.UnlockKey(status.Node)
+			break
+		}
+		e.nodeZoneState.UnlockKey(status.Node)
+	}
+	if !proceed && !e.isPodScheduledinLocalZone(pod) {
+		return nil // nothing to do if none of the status nodes are local to this master and pod is also remote
 	}
 	for _, status := range remainingAssignments {
 		klog.V(2).Infof("Adding pod egress IP status: %v for EgressIP: %s and pod: %s/%s/%v", status, name, pod.Namespace, pod.Name, podIPNets)
@@ -1156,6 +1171,8 @@ type egressIPCache struct {
 	egressLocalNodesCache sets.Set[string]
 	// egressIP IP -> assigned node name
 	egressIPIPToNodeCache map[string]string
+	// egressIP name -> egress IP -> assigned node name
+	egressIPToAssignedNodes map[string]map[string]string
 	// node name -> network name -> redirect IPs
 	egressNodeRedirectsCache nodeNetworkRedirects
 	// network name -> OVN cluster router name
@@ -1590,6 +1607,14 @@ func (e *EgressIPController) syncPodAssignmentCache(egressIPCache egressIPCache)
 						}
 					}
 
+					// populate podState.egressStatuses with assigned node for active egressIP IPs.
+					if podState.egressIPName == egressIPName {
+						for egressIPIP, nodeName := range egressIPCache.egressIPToAssignedNodes[egressIPName] {
+							podState.egressStatuses.statusMap[egressipv1.EgressIPStatusItem{
+								EgressIP: egressIPIP, Node: nodeName}] = egressStatusStatePending
+						}
+					}
+
 					e.podAssignment.Store(podKey, podState)
 					return nil
 				}); err != nil {
@@ -1971,6 +1996,9 @@ func (e *EgressIPController) generateCacheForEgressIP() (egressIPCache, error) {
 	// egressIP IP -> node name. Assigned node for EIP.
 	egressIPIPNodeCache := make(map[string]string, 0)
 	cache.egressIPIPToNodeCache = egressIPIPNodeCache
+	// egressIP name -> egressIP IP -> node name.
+	egressIPToAssignedNodes := make(map[string]map[string]string, 0)
+	cache.egressIPToAssignedNodes = egressIPToAssignedNodes
 	cache.markCache = make(map[string]string)
 	egressIPs, err := e.watchFactory.GetEgressIPs()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


go-controller/pkg/ovn/egressip_test.go
```diff
@@ -11154,6 +11145,299 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+					&nbdb.LogicalRouter{
+						Name:     types.OVNClusterRouter,
+						UUID:     types.OVNClusterRouter + "-UUID",
+						Policies: []string{"no-reroute-UUID", "no-reroute-service-UUID", "default-no-reroute-node-UUID", "default-no-reroute-reply-traffic", "reroute-UUID1"},
+					},
+					&nbdb.LogicalRouter{
+						Name:  types.GWRouterPrefix + node1.Name,
+						UUID:  types.GWRouterPrefix + node1.Name + "-UUID",
+						Ports: []string{types.GWRouterToJoinSwitchPrefix + types.GWRouterPrefix + node1.Name + "-UUID"},
+						Nat:   []string{"pod-node-nat-UUID1"},
```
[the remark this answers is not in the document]
pperiyasamy: this suggestion is not for changes in the PR.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -11154,6 +11145,361 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+						ExternalIDs: getEgressIPLRPNoReRoutePodToJoinDbIDs(IPFamilyValueV4, types.DefaultNetworkName, fakeOvn.controller.eIPC.controllerName).GetExternalIDs(),
+					},
+					&nbdb.LogicalRouterPolicy{
+						Priority:    types.EgressIPReroutePriority,
+						Match:       fmt.Sprintf("ip4.src == %s", egressPod.Status.PodIP),
+						Action:      nbdb.LogicalRouterPolicyActionReroute,
+						Nexthops:    []string{node2IPv4TranSwitchIP},
+						ExternalIDs: getEgressIPLRPReRouteDbIDs(eIP.Name, egressPod.Namespace, egressPod.Name, IPFamilyValueV4, types.DefaultNetworkName, fakeOvn.controller.eIPC.controllerName).GetExternalIDs(),
+						UUID:        "reroute-UUID1",
+					},
```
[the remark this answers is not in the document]
pperiyasamy: The `egressPod2` LRP belongs to remote zone node `node3`, so this entry can't be present in global zone ovn database.

go-controller/pkg/ovn/egressip.go
```diff
@@ -921,6 +934,9 @@ func (e *EgressIPController) deleteEgressIPAssignments(name string, statusesToRe
 					return nil
 				}
 				if podStatus.egressIPName != name {
+					if ok := podStatus.egressStatuses.contains(statusToRemove); ok {
```
kyrtapz: Why is this change required? We seem to be removing it  in line 972 too.

kyrtapz:
I am confused with the reproducer:

> EIP-1 assigned to node-1. node-1 and node-2 rebooted at the sametime.  
This made EIP failover and ovnkube-controller container restart happened almost at the same time.
> 1. EIP-1 is reassigned to node-2 by cluster manager.
> 2. EIP controller synchronizes existing EIP objects with older EIP status, so it cleans up SNATs and LRPs referring to node-1 due to stale pod IP addresses.
> 3. At the same time, pod1 and pod2 add events are triggered, but EIP controller's watch factory seeing older EIP status from the informer cache, so SNATs and LRPs are created referring to node-1 for the new pod IPs.
> 4. EIP-1 add event is triggered with new EIP status, EIP controller adds new SNAT entries and updates LRP nexthop for node-2.
> 5. stale SNATs, LRPs having with stale nexthops for node-1.

Why is step 4 an EIP add? It should be an update? in which case we are supposed to cleanup the old status from the configured pods an not leave any stale entries behind.

pperiyasamy:
> Why is step 4 an EIP add? It should be an update? in which case we are supposed to cleanup the old status from the configured pods an not leave any stale entries behind.

@kyrtapz this is at the time of `ovnkube-node` pod restart (due to node reboot), so EIP add is received. isn't that expected ?

kyrtapz:
> > Why is step 4 an EIP add? It should be an update? in which case we are supposed to cleanup the old status from the configured pods an not leave any stale entries behind.
> 
> @kyrtapz this is at the time of `ovnkube-node` pod restart (due to node reboot), so EIP add is received. isn't that expected ?

So you are saying that we ran the sync with the EIP using the old object but then got the ADD with the new one? That sounds weird as I would expect the synthetic ADDs and the initial syncs to use the same objects to avoid exactly this type of issues.

pperiyasamy:
> > > Why is step 4 an EIP add? It should be an update? in which case we are supposed to cleanup the old status from the configured pods an not leave any stale entries behind.
> > 
> > 
> > @kyrtapz this is at the time of `ovnkube-node` pod restart (due to node reboot), so EIP add is received. isn't that expected ?
> 
> So you are saying that we ran the sync with the EIP using the old object but then got the ADD with the new one? That sounds weird as I would expect the synthetic ADDs and the initial syncs to use the same objects to avoid exactly this type of issues.

> So you are saying that we ran the sync with the EIP using the old object but then got the ADD with the new one? 

That's right.

> I would expect the synthetic ADDs and the initial syncs to use the same objects to avoid exactly this type of issues.

Okay, but that hasn’t always been the case. The initial sync sometimes used an outdated EIP object, possibly because the watch factory informer cache wasn’t fully up to date when syncing existing objects. This issue occurs intermittently when EIP failover attempts combined with ovnkube-controller restarts.

pperiyasamy:
> > > > Why is step 4 an EIP add? It should be an update? in which case we are supposed to cleanup the old status from the configured pods an not leave any stale entries behind.
> > > 
> > > 
> > > @kyrtapz this is at the time of `ovnkube-node` pod restart (due to node reboot), so EIP add is received. isn't that expected ?
> > 
> > 
> > So you are saying that we ran the sync with the EIP using the old object but then got the ADD with the new one? That sounds weird as I would expect the synthetic ADDs and the initial syncs to use the same objects to avoid exactly this type of issues.
> 
> > So you are saying that we ran the sync with the EIP using the old object but then got the ADD with the new one?
> 
> That's right.
> 
> > I would expect the synthetic ADDs and the initial syncs to use the same objects to avoid exactly this type of issues.
> 
> Okay, but that hasn’t always been the case. The initial sync sometimes used an outdated EIP object, possibly because the watch factory informer cache wasn’t fully up to date when syncing existing objects. This issue occurs intermittently when EIP failover attempts combined with ovnkube-controller restarts.

@kyrtapz Thanks for the offline discussion, you correctly pointed out EIP initial sync and add happens with new EIP status, only pod add event saw older EIP status from the informer cache which led to stale SNATs/LRP nexthops. so adjusted the PR only for this scenario. PTAL.

go-controller/pkg/ovn/egressip.go
```diff
@@ -846,6 +828,11 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 			if exists := podState.egressStatuses.contains(status); !exists {
 				remainingAssignments = append(remainingAssignments, status)
 			}
+			// Detect stale EIP status entries (same EgressIP reassigned to a different node)
+			// and queue the outdated entry for cleanup.
+			if staleStatus := podState.egressStatuses.hasStaleEIPStatus(status); staleStatus != nil {
```
kyrtapz, line 835: Can we add something to the function title/comments mentioning that this is addressing stale entries that are already in cache?

go-controller/pkg/ovn/egressip.go
```diff
@@ -867,6 +854,38 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
...
+	for _, staleStatus := range staleAssignments {
+		klog.V(2).Infof("Deleting stale pod egress IP status: %v for EgressIP: %s and pod: %s/%s/%v", staleStatus, name, pod.Namespace, pod.Name, podIPNets)
+		err = e.deletePodEgressIPAssignments(ni, name, []egressipv1.EgressIPStatusItem{staleStatus}, pod)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to delete stale EgressIP status %s/%v for pod %s: %w", name, staleStatus, podKey, err))
+		}
+		delete(podState.egressStatuses.statusMap, staleStatus)
+	}
+	if err := utilerrors.Join(errs...); err != nil {
+		klog.Warningf("Failed to cleanup stale EgressIP statuses for EgressIP object %s: %v", name, err)
```
kyrtapz: If you want to just log it there is no point of aggregating the error, just print them directly in the for loop above.

kyrtapz:
After talking to @pperiyasamy I think there still might be a possibility for a race:
1. syncEgressIPs is called from `WatchEgressIPNamespaces` with old EgressIP status.
2. EgressIP status gets updated.
3. Synthetic EgressIPNamespace ADD is called and populates the cache with the new value so there is no chance for figuring out the stale one.

Maybe we need to populate the eipStatus in `syncEgressIPs`.

go-controller/pkg/ovn/egressip.go
pperiyasamy: sure, done.

go-controller/pkg/ovn/egressip.go
pperiyasamy: correct, logging it in above loop itself.

pperiyasamy:
> After talking to @pperiyasamy I think there still might be a possibility for a race:
> 
> 1. syncEgressIPs is called from `WatchEgressIPNamespaces` with old EgressIP status.
> 2. EgressIP status gets updated.
> 3. Synthetic EgressIPNamespace ADD is called and populates the cache with the new value so there is no chance for figuring out the stale one.
> 
> Maybe we need to populate the eipStatus in `syncEgressIPs`.

yes @kyrtapz , restored `syncEgressIPs` changes, PTAL.

go-controller/pkg/ovn/egressip.go
```diff
@@ -843,9 +825,16 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 		// podState.egressIPName can be empty if no re-routes were found in
 		// syncPodAssignmentCache for the existing pod, we will treat this case as a new add
 		for _, status := range statusAssignments {
-			if exists := podState.egressStatuses.contains(status); !exists {
+			// Add the status if it's not already in the cache, or if it exists but is in synced state
+			// (meaning it was populated during EIP sync and needs to be processed for the pod).
+			if value, exists := podState.egressStatuses.statusMap[status]; !exists || value == egressStatusStateSynced {
```
kyrtapz: So using the value for the map is caused by the fact that we need to apply the status on startup at least once to avoid a scenario where the status was updated when ovn-k was out, right?

While I am ok with this approach I don't understand why are you using the notion of `synced`, it is almost the opposite?
Maybe it is just me but this is a bit confusing.

go-controller/pkg/ovn/egressip.go
```diff
@@ -922,6 +937,9 @@ func (e *EgressIPController) deleteEgressIPAssignments(name string, statusesToRe
 					return nil
 				}
 				if podStatus.egressIPName != name {
+					if ok := podStatus.egressStatuses.contains(statusToRemove); ok {
+						podStatus.egressStatuses.delete(statusToRemove)
```
kyrtapz: Why is this required?
We seem to be doing the same already in line 975.

kyrtapz (dismissed):
LGTM

Left some non-blocking comments.

go-controller/pkg/ovn/egressip.go
pperiyasamy: ok, i agree. this is bit confusing, will change the variable like below:

```
// egressStatusStatePending marks entries populated during EIP sync and
// indicates they must be reconciled again for the pod.
const egressStatusStatePending = "pending"
```

go-controller/pkg/ovn/egressip.go
pperiyasamy: good catch, the sync function populates `egressStatuses` unconditionally even for standby EIPs, we need to populate it only for active EIP object. once that is fixed, can remove this change. will update the PR.

pperiyasamy:
saw a flake with EgressFirewall E2E test, unrelated to this PR changes. created an issue https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5693.

PR #5610, opened 2025-09-30
rfc: Getting rid of ovnkube.sh - ovnkube-controller

booxter:
This is a strawman PR that only contains some cliff notes in ovnkube.sh explaining how one could get rid of ovnkube.sh (and bash) in the execution path of ovnkube services, which would allow us to run ovnkube in a "distroless" container.


booxter:
No longer a priority for me.

PR #5616, opened 2025-09-30
fix: Mark --*file options as working with files

booxter:
This should help some shells to improve autocomplete behavior with the
cli (e.g. fish).

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


kyrtapz:
@booxter nice!
There are a bunch more cli flags that take a file path, can we add those too?
For example: `k8s-kubeconfig`, `bootstrap-kubeconfig`, `cert-dir`, `k8s-cacert` and many more. 

Also, are you running the ovn-k binary manually and not as part of the daemonset/deployment? That's interesting!

booxter:
@kyrtapz Last week I started looking at our CLI implementation, which is the cause of a number of small-ish PRs I [sent](https://github.com/ovn-kubernetes/ovn-kubernetes/pulls?q=is%3Aopen+is%3Apr+author%3Abooxter+created%3A%3C2025-10-02).

The idea to mark the `file`-typed options came from Rabbit bot, and while I don't myself use `fish` (which is the only shell that currently considers this flag when generating autocomplete scripts), I found it compelling enough to send a patch to oblige its recommendation - in a separate PR; hence this PR here.

Thanks for pointing out there are more options of the same nature that I should capture. Will do.

---

More broadly speaking...

My ideal [end-goal](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5643) is being able to run OVN-K (and, later, OVN) services via the daemonset without a dependency on bash, to reduce attack vector for a compromised container, reduce size of images etc. But I am going to target a less lofty goal first - allowing to run OVN-K without `ovnkube.sh` at least for a subset of scenarios (even if `ov[ns]-ctl` scripts - and shell dependency that comes with them - are still in place).

I understand that removing bash dependency is a major change to the service startup path and should gather some positive feedback from the wider community before I plunge on the task. So I'm going to raise this topic during our next community meeting. (I will report also a RFC issue capturing rationale this week prior to the meeting, so that the discussion can be more grounded.)

PR #5617, opened 2025-09-30
add lint target to run golanci natively

jluhrsen:
new target lint-run-natively will now pull the golangci-lint binary based on the version in lint.sh locally and run the lint check. This is useful in the case that some system does not have a container runtime available.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


go-controller/Makefile
```diff
@@ -105,6 +105,9 @@ else
 	echo "linter can only be run within a container since it needs a specific golangci-lint version"; exit 1
 endif
 
+lint-run-natively:
```
kyrtapz: Why can't we just move this under the else in the following blocks:
```
lint:
ifeq ($(CONTAINER_RUNNABLE), 0)
	@GOPATH=${GOPATH} ./hack/lint.sh $(CONTAINER_RUNTIME) || { echo "lint failed! Try running 'make lint-fix'"; exit 1; }
else
	echo "linter can only be run within a container since it needs a specific golangci-lint version"; exit 1
endif
lint-fix:
ifeq ($(CONTAINER_RUNNABLE), 0)
	@GOPATH=${GOPATH} ./hack/lint.sh $(CONTAINER_RUNTIME) fix || { echo "ERROR: lint fix failed! There is a bug that changes file ownership to root \
	when this happens. To fix it, simply run 'chown -R <user>:<group> *' from the repo root."; exit 1; }
else
	echo "linter can only be run within a container since it needs a specific golangci-lint version"; exit 1
endif
```
This would make it functional in environments without a container runtime without the need for any new way of calling lint.

I think we can go as far as modifying the above to match what we do with `gofmt` - for consistency.

go-controller/hack/lint.sh
```diff
@@ -1,18 +1,28 @@
 #!/usr/bin/env bash
 VERSION=v1.64.8
-extra_flags=""
+: "${GOLANGCI_LINT_VERSION:=$VERSION}"
+extra_flags=(--verbose --print-resources-usage --modules-download-mode=vendor --timeout=15m0s)
```
kyrtapz, line 4: What is the reason behind adding these extra flags?

go-controller/hack/lint.sh
```diff
@@ -1,18 +1,28 @@
...
 if [ "$#" -ne 1 ]; then
   if [ "$#" -eq 2 ] && [ "$2" == "fix" ]; then
-    extra_flags="--fix"
+    extra_flags+=(--fix)
   else
-    echo "Expected command line argument - container runtime (docker/podman) got $# arguments: $@"
+    echo "Expected command line argument - container runtime (docker/podman) or 'run-natively'; got $# arguments: $*"
```
kyrtapz, line 9: NIT: What's the reason to use `$*` instead of `$@`?

kyrtapz (commented):
Thanks @jluhrsen, overall I think this is a good approach!
I've left some suggestions to make it more consistent with the other tools we are running.

go-controller/Makefile
jluhrsen: I can do that. I was trying not to break existing functionality where 'make lint' would still fail if there is no container runtime. figured that was done for a reason.

please LMK if you want me to change it or leave it as is. I do not care one single bit which way :)

go-controller/hack/lint.sh
jluhrsen: just do de-dup them in the two below golangci-lint commands

go-controller/hack/lint.sh
jluhrsen: coderabbit made me do it :)

go-controller/Makefile
jluhrsen: @kyrtapz , I did this. can you PTAL again?

go-controller/Makefile
```diff
@@ -102,7 +102,8 @@ ifeq ($(CONTAINER_RUNNABLE), 0)
 	@GOPATH=${GOPATH} ./hack/lint.sh $(CONTAINER_RUNTIME) fix || { echo "ERROR: lint fix failed! There is a bug that changes file ownership to root \
 	when this happens. To fix it, simply run 'chown -R <user>:<group> *' from the repo root."; exit 1; }
 else
-	echo "linter can only be run within a container since it needs a specific golangci-lint version"; exit 1
+	echo "no container runtime. attempting to run natively";
```
kyrtapz: You only modified lint-fix and it is running linter instead of trying to run it with fix. 
We need to modify `lint` too and fix `lint-fix`.

jluhrsen: ugh... my bad. working too fast.

it's updated. changed the error message in the run-natively "fix" path as well.

jluhrsen:
@kyrtapz , ok I think this is right now. Unless you want me to listen to coderabbit about being extra careful about the curl to grab the golangci-lint binary. [see here](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5617#discussion_r2414288518) above

I used your trick to test this d/s with [this PR](https://github.com/openshift/ovn-kubernetes/pull/2791). this job ran and here is the [lint build log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2791/pull-ci-openshift-ovn-kubernetes-master-lint/1975961832480837632/artifacts/test/build-log.txt).

go-controller/Makefile
```diff
@@ -94,15 +94,18 @@ lint:
...
 endif
 
 lint-fix:
 ifeq ($(CONTAINER_RUNNABLE), 0)
 	@GOPATH=${GOPATH} ./hack/lint.sh $(CONTAINER_RUNTIME) fix || { echo "ERROR: lint fix failed! There is a bug that changes file ownership to root \
 	when this happens. To fix it, simply run 'chown -R <user>:<group> *' from the repo root."; exit 1; }
 else
-	echo "linter can only be run within a container since it needs a specific golangci-lint version"; exit 1
+	echo "no container runtime. attempting to run natively w/ -fix flag";
+	@GOPATH=${GOPATH} ./hack/lint.sh run-natively || { echo "running lint fix locally failed! There is a bug that changes file ownership to root \
```
kyrtapz: This is wrong, it should be:
```suggestion
	@GOPATH=${GOPATH} ./hack/lint.sh run-natively fix || { echo "running lint fix locally failed! There is a bug that changes file ownership to root \
```

jluhrsen: thank you
done

go-controller/Makefile
```diff
@@ -94,15 +94,18 @@ lint:
 ifeq ($(CONTAINER_RUNNABLE), 0)
 	@GOPATH=${GOPATH} ./hack/lint.sh $(CONTAINER_RUNTIME) || { echo "lint failed! Try running 'make lint-fix'"; exit 1; }
 else
-	echo "linter can only be run within a container since it needs a specific golangci-lint version"; exit 1
+	echo "no container runtime. attempting to run natively";
```
tssurya, line 97: non-actionable-super nit: i guess it doesn't only mean no container runtime? cause that seems to be CONTAINER_RUNTIME var itself?
`$(CONTAINER_RUNTIME) -v`
might also be if RUNTIME isn't running or path not found or other things

go-controller/hack/lint.sh
```diff
@@ -1,18 +1,28 @@
...
-$1 run --security-opt label=disable --rm \
-  -v  ${HOME}/.cache/golangci-lint:/cache -e GOLANGCI_LINT_CACHE=/cache \
-  -v $(pwd):/app -w /app -e GO111MODULE=on docker.io/golangci/golangci-lint:${VERSION} \
-	golangci-lint run --verbose --print-resources-usage \
-	--modules-download-mode=vendor --timeout=15m0s ${extra_flags} && \
-	echo "lint OK!"
\ No newline at end of file
+if [ "$1" = "run-natively" ]; then
+  mkdir -p /tmp/local/bin/
+  curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /tmp/local/bin/ "${GOLANGCI_LINT_VERSION}"
```
tssurya: the -f is fail silently on http errors? do we want that here with https:// url?

go-controller/hack/lint.sh
```diff
@@ -1,18 +1,28 @@
...
+  mkdir -p /tmp/golangci-cache
+  export GOLANGCI_LINT_CACHE=/tmp/golangci-cache
+  /tmp/local/bin/golangci-lint run "${extra_flags[@]}" && \
+  echo "lint OK!"
```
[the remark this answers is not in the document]
tssurya: this comment sounds legit @jluhrsen can you check?
is there a possibility here that curl fails but sh gets no input here and so returns 0?
do we need to maybe include set -o pipefail?

go-controller/Makefile
jluhrsen: isn't this what we are doing if $CONTAINER_RUNNABLE is 0, which I *think* happens if podman or docker is not available.

I can update the message here though if I am misunderstanding.

go-controller/hack/lint.sh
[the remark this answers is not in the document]
jluhrsen: oh, it's definitely a legit comment. It's not a ton of code, but was keeping this as simple as possible since it's not gonna cause too much pain in the off chance we have trouble.

Sounds like you would like it though, so I just updated it to include it.

go-controller/hack/lint.sh
jluhrsen: updated to remove the -f

muscle memory to keep output minimal

go-controller/hack/lint.sh
```diff
@@ -1,18 +1,39 @@
...
+  if ! curl -sSL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh -o "${tmp_installer}"; then
+    echo "failed to download golangci-lint installer" >&2
+    exit 1
+  fi
+  chmod +x "${tmp_installer}"
+  if ! "${tmp_installer}" -b /tmp/local/bin/ "${GOLANGCI_LINT_VERSION}"; then
+    echo "failed to install golangci-lint ${GOLANGCI_LINT_VERSION}" >&2
+    exit 1
+  fi
+  rm -f "${tmp_installer}"  mkdir -p /tmp/golangci-cache
```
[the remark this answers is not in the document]
jluhrsen: thank you! good catch

jluhrsen:
@tssurya  addressed your comments. PTAL

tssurya:
> @jluhrsen lgtm, thanks for fixing this
> 
> but given I don't trust my bash skills nearly enough, can you run your latest as well against downstream - check it works and if you circle back I can merge this.

@jluhrsen PTAL. any updates?

jluhrsen:
> > @jluhrsen lgtm, thanks for fixing this
> > but given I don't trust my bash skills nearly enough, can you run your latest as well against downstream - check it works and if you circle back I can merge this.
> 
> @jluhrsen PTAL. any updates?

[d/s PR to test this](https://github.com/openshift/ovn-kubernetes/pull/2806)
[job that ran](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2806/pull-ci-openshift-ovn-kubernetes-master-lint/1978543947223404544)
[log of the bash running lint](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_ovn-kubernetes/2806/pull-ci-openshift-ovn-kubernetes-master-lint/1978543947223404544/artifacts/test/build-log.txt)

jcaamano (approved):

PR #5622, opened 2025-10-01
fix: Skip node-encap-ips annotation in DPU host mode

aserdean:
The node-encap-ips annotation should not be set when running in DPU host mode since the encap IP is not available on the DPU host. This prevents errors during node initialization when the encap IP cannot be determined.

- Add conditional check for NodeModeDPUHost before setting encap IP annotation
- Maintain existing behavior for non-DPU host modes
- Fixes initialization errors in DPU host mode deployments

--

Cherry-pick from master to release-1.1


npinaeva:
@aserdean I have added CI to this branch, so you could rebase and fix the commit message, then we can merge it

aserdean:
> @aserdean I have added CI to this branch, so you could rebase and fix the commit message, then we can merge it

sorry I missed this, since we already have a new release I am dropping this MR in

PR #5624, opened 2025-10-02
ovs-node: Delete transient ports on startup

### LINKED ISSUE #5596: OVN Kubernetes nodes are unable to connect to network on k3s after reboot
### What would you like to know?

I have a dev cluster where we're testing OVN-Kubernetes for CNI on k3s. 
We're running k3s v1.32.3 and ovn-kubernetes release-1.1 on ubuntu 24.04

Image

Image

Image

Image

Image

Short of removing k3s from the node I have no clue how to restore network connectivity

_No response_

booxter:
## 📑 Description

Since 22d2b24b60, ovn-k modifies /etc/default/ configuration for systemd
services included with rhel and ubuntu so that transient ports are
deleted on ovs service startup. This allowed the node to restore
communication through the NIC after reboot.

Once ovnkube starts, it will rewire the NIC back through the ovs bridge,
and move IP configuration (addresses, routes).

The issue is that 22d2b24b60 was assuming that ovs daemons are running
as systemd services, which is not the case with ovs-node daemonset. For
the latter, it's ovnkube.sh that calls ovs-ctl start. This patch extends
the ovnkube.sh script so that we always call ovs-ctl with
--delete-transient-ports when starting ovs services.

Note: in daemonset scenario, ovnkube should probably not attempt to
modify systemd /etc/ files for ovs services. That said, it's probably
harmless that it does so, since errors on missing /etc/ file are
ignored.

Fixes #5596

## ✅ Checks

- [ ] My code requires tests


booxter:
@trozet @tssurya FYI this should solve the lost IP address on node reboot when using `ovs-node` daemonset.

@wouterhummelink if you could confirm this solves your problem, it would help a lot. Thanks!

booxter:
Drafted for now to add a test case.

test/e2e/util.go
```diff
@@ -1749,3 +1749,105 @@ func findOVNDBLeaderPod(f *framework.Framework, cs clientset.Interface, namespac
...
+		framework.Logf("OVN-Kubernetes cluster is healthy: %d nodes, %d ovnkube-node pods, %d running master pods",
+			len(nodes.Items), len(ovnNodePods.Items), runningMasterPods)
+		return true, nil
+	})
+}
+
+// checkBrEthIPAddress verifies that breth0 interface on the specified node has an IP address
+func checkBrEthIPAddress(nodeName string) error {
+	return wait.PollImmediate(2*time.Second, 60*time.Second, func() (bool, error) {
+		stdout, err := infraprovider.Get().ExecK8NodeCommand(nodeName, []string{"ip", "addr", "show", "breth0"})
```
booxter: (thinking aloud) the name may be different on non-kind clusters. How does one extract bridge name in runtime?

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -518,15 +580,11 @@ func isNetworkAttachedToContainer(networkName, containerName string) bool {
...
 	if err != nil {
-		panic(fmt.Sprintf("failed to check if external container (%s) exists: %v (%s)", name, err, stdOut))
-	}
-	if string(stdOut) == "" {
-		return false
+		panic(fmt.Sprintf("failed to check if external container (%s) exists: %v", name, err))
 	}
-	return true
+	return state != ""
 }
```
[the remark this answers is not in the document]
booxter: Valid point but this should really be done in a separate PR since the existing function that we only refactored here already panics. I've sent: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5635

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -119,6 +119,68 @@ func (k *kind) GetK8HostPort() uint16 {
 	return k.hostPort.Allocate()
 }
 
+// getContainerState returns the state of a container by name
+// Returns empty string if container doesn't exist
+func getContainerState(containerName string) (string, error) {
+	stdOut, err := exec.Command(containerengine.Get().String(), "ps", "-a", "-f", fmt.Sprintf("name=^%s$", containerName), "--format", "{{.State}}").CombinedOutput()
```
booxter, line 139: not great; ideally we'd switch everything to use native libraries to talk to docker / podman. I will take a look at it at some point later not to increase the scope of the PR even further.

booxter:
The new test case failed in at least one of the CI runs (node was not able to restore IP address on breth0). Will have to take a look. Drafting.

test/e2e/node_shutdown_startup.go
```diff
@@ -0,0 +1,259 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+var _ = ginkgo.Describe("Node Shutdown and Startup", ginkgo.Serial, func() {
```
tssurya, line 21: @martinkennelly and @huiran0826 this is similar to what you both are planning to doing for EIP disruption testing upstream. Reviewing this will be good.

test/e2e/node_shutdown_startup.go
```diff
@@ -0,0 +1,259 @@
...
+		initialIPFamilies, err := getBridgeIPAddressFamilies(testNodeName)
+		framework.ExpectNoError(err, "Should be able to get breth0 IP address families before shutdown")
+		framework.Logf("Node %s breth0 initial IP families: IPv4=%v, IPv6=%v", testNodeName, initialIPFamilies.hasIPv4, initialIPFamilies.hasIPv6)
+
+		ginkgo.By("Shut down the node")
+		framework.Logf("Shutting down node %s", testNodeName)
+		err = infraprovider.Get().ShutdownNode(testNodeName)
+		framework.ExpectNoError(err, "Failed to shutdown node %s", testNodeName)
+
+		// Ensure node is started back up regardless of test failure
```
huiran0826, line 81: do we ensure the node back is just started or in 'Ready' status?

test/e2e/node_shutdown_startup.go
```diff
@@ -0,0 +1,259 @@
...
+			}
+
+			framework.Logf("Ensuring node %s is started (cleanup)", testNodeName)
+			if startErr := infraprovider.Get().StartNode(testNodeName); startErr != nil {
+				framework.Logf("Failed to start node %s during cleanup: %v", testNodeName, startErr)
+			}
+		}()
+
+		// Wait for the node to be marked as NotReady
+		ginkgo.By("Waiting for node to be marked as NotReady")
```
huiran0826, line 100: Can we put this node NotReady/Ready status checking to a function that others can reuse it as well?

test/e2e/node_shutdown_startup.go
huiran0826: @booxter Thanks for this node stop/start functions, we have the similar scenarios for EIP testing, and this would be helpful.

test/e2e/node_shutdown_startup.go
booxter: Updated to wait for Ready now.

test/e2e/node_shutdown_startup.go
booxter: Moved the code to a helper function and reused it above when waiting for Node to become ready on deferred cleanup.

test/scripts/e2e-cp.sh
```diff
@@ -141,6 +148,12 @@ if [[ "${WHAT}" != "${NETWORK_SEGMENTATION_TESTS}"* ]]; then
   skip $NETWORK_SEGMENTATION_TESTS
 fi
 
+SERIAL_LABEL="Serial"
+if [[ "${WHAT}" = "$SERIAL_LABEL" ]]; then
+  require_label "$SERIAL_LABEL"
+  shift
+fi
```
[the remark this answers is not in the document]
trozet: @booxter I don't think we need the shift here. We normally pass WHAT as a single string with wildcards for regexp matching. I don't think we handle a test name broken up into multiple arguments.

trozet (changes requested):
thanks @booxter! one minor thing to address

test/e2e/node_shutdown_startup.go
```diff
@@ -0,0 +1,233 @@
...
+		ginkgo.By("Confirm that ovn-k cluster is back to healthy after all services are settled")
+		err = waitOVNKubernetesHealthy(f)
+		framework.ExpectNoError(err, "OVN-Kubernetes cluster should be healthy after node restart")
+
+		ginkgo.By("Confirm that breth0 on the node has IP addresses of expected families (that were moved from eth0)")
+		err = checkBridgeIPAddressFamilies(testNodeName, initialIPFamilies)
+		framework.ExpectNoError(err, "breth0 should have IP addresses of the same families as before restart")
+
+		framework.Logf("Node shutdown/startup test completed successfully for node %s", testNodeName)
+	})
```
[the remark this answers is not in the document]
booxter: I don't think it's reasonable - after OVN-K is up, it's enough time for it to create a bridge (which happens at the start of the service main).

test/e2e/node_shutdown_startup.go
```diff
@@ -0,0 +1,233 @@
...
+		framework.Logf("Failed to parse containers JSON from node %s: %v", nodeName, err)
+		// Fallback: try to extract container IDs using crictl ps without JSON
+		simpleOutput, err := infraprovider.Get().ExecK8NodeCommand(nodeName, []string{"crictl", "ps", "-a"})
+		if err != nil {
+			framework.Logf("Failed to list containers (simple format) on node %s: %v", nodeName, err)
+			return
+		}
+		framework.Logf("Container list (simple format) from node %s:\n%s", nodeName, simpleOutput)
+		return
+	}
```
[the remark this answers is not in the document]
booxter: Indeed.

```
      "metadata": {
        "name": "kindnet-cni",
        "attempt": 0
      },
```

test/scripts/e2e-cp.sh
```diff
@@ -40,12 +40,19 @@ skip() {
...
-  SKIPPED_LABELED_TESTS+="!($*)"
+  LABELED_TESTS+="!($*)"
+}
+
+require_label() {
+  if [ "$LABELED_TESTS" != "" ]; then
+  	LABELED_TESTS+=" && "
+  fi
+  LABELED_TESTS+="$*"
 }
```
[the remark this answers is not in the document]
booxter: The caller is responsible for arguments to make sense. Clearly Serial *and* PARALLEL are wrong.

test/scripts/e2e-cp.sh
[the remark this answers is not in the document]
booxter: @trozet actually I now remember why we need `shift` here - we filter out Serial tests with native ginkgo label, not by test name (tests may and do not contain `Serial` string) - so we need to `shift` otherwise `ginkgo` will be asked to `ginkgo.focus Serial` which results in zero tests executed.

PR #5626, opened 2025-10-03
kubevirt: fix bad release of IPs of live migratable pods

jcaamano:
There's two circumstances when IPs were being released incorrectly:

* when a live migratable pod completed with no migration ongoing it was
  not being released due to IsMigratedSourcePodStale outright assuming a
  completed pod was stale.
* when a live migratable pod completed on a different node than the VM's
  original as part of a migration it was being released when it
  shouldn't, we were simply not checking if it was a migration.

It also improves the tests:

* add checks for IP release
* use per pod SNATs which is the more complicated scenario
* check OVN DB when source pod (or failed migrated pod) is completed

Worth noting is that we are leaving stale SNATs behind as VMs migrate.
Unsure if this is on purpose. Registered https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5627

Other assorted fixes:

* avoid warning when attempting to release IPs of a completed live
  migratable pod out of the VM's original node zone
* fix search of colliding pods for migrated pods

## Summary by CodeRabbit

* **Bug Fixes**
  * Gate IP deallocation until all VM pods complete to avoid premature release during live migration.
  * Skip IP release for live‑migratable pods whose IPs aren’t local to the current zone.
  * Add subnet‑ownership aware node resolution and a fallback all‑nodes lookup; delegate IP-release decisions to a conditional release path when appropriate.

* **Tests**
  * Expanded migration tests to cover SNAT/multi‑GW and NAT cleanup; updated test helpers and expected NBDB/NAT expectations.
  * e2e VM node labeling updated to use "true" for live‑migration selection.

### DIFF
--- a/go-controller/pkg/ovn/kubevirt_test.go
+++ b/go-controller/pkg/ovn/kubevirt_test.go
@@ -108,6 +108,8 @@ var _ = Describe("OVN Kubevirt Operations", func() {
 		subnetIPv6            string
 		transitSwitchPortIPv4 string
 		transitSwitchPortIPv6 string
+		addressIPv4           string
+		addressIPv6           string
 	}
 
 	type testVM struct {
@@ -128,6 +130,8 @@ var _ = Describe("OVN Kubevirt Operations", func() {
 				lrpNetworkIPv6:        "fd98::4/64",
 				transitSwitchPortIPv4: "100.65.0.4/24",
 				transitSwitchPortIPv6: "fd13::4/64",
+				addressIPv4:           "10.89.0.1/24",
+				addressIPv6:           "fc00:f853:ccd:e793::1/64",
 			},
 			node2: {
 				nodeID:                "5",
@@ -137,6 +141,8 @@ var _ = Describe("OVN Kubevirt Operations", func() {
 				lrpNetworkIPv6:        "fd98::5/64",
 				transitSwitchPortIPv4: "100.65.0.5/24",
 				transitSwitchPortIPv6: "fd13::5/64",
+				addressIPv4:           "10.89.0.2/24",
+				addressIPv6:           "fc00:f853:ccd:e793::2/64",
 			},
 			node3: {
 				nodeID:                "6",
@@ -146,6 +152,8 @@ var _ = Describe("OVN Kubevirt Operations", func() {
 				lrpNetworkIPv6:        "fd98::6/64",
 				transitSwitchPortIPv4: "100.65.0.6/24",
 				transitSwitchPortIPv6: "fd13::6/64",
+				addressIPv4:           "10.89.0.3/24",
+				addressIPv6:           "fc00:f853:ccd:e793::3/64",
 			},
 		}
 		vmByName = map[string]testVM{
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/pods.go
```diff
@@ -211,6 +211,20 @@ func (oc *DefaultNetworkController) deleteLogicalPort(pod *corev1.Pod, portInfo
...
+	if kubevirt.IsPodLiveMigratable(pod) {
+		_, hasLocalIPs := oc.lsManager.GetSubnetName(pInfo.ips)
+		if !hasLocalIPs {
+			klog.V(5).Infof("Inhibiting release of live migratable pod %s/%s IPs %s not managed by this zone",
+				pod.Namespace, pod.Name,
+				util.JoinIPNetIPs(pInfo.ips, " "),
+			)
+			return nil
+		}
+	}
```
qinqon, line 229: this has to be 
```golang
if kubevirt.IsPodLiveMigratable(pod) && !allVMPodsCompleted {
```

So in the case where the complted virt-launcher living at the zone owning the IP disappear before all the vm pods are completed we will not de allocate the ip

go-controller/pkg/ovn/kubevirt_test.go
```diff
@@ -41,6 +42,8 @@ var _ = Describe("OVN Kubevirt Operations", func() {
 		dnsServiceIPv6    = "fd7b:6b4d:7b25:d22f::3"
 		clusterCIDRIPv4   = "10.128.0.0/16"
 		clusterCIDRIPv6   = "fe00::/64"
+		subnetSuffixIPv4  = "/24"
+		subnetSuffixIPv6  = "/64"
```
qinqon, line 47: The ipv6 cluster cider is `"fe00::/64"` so subnet ipv6 suffix should be smaller than 64 same as we do for ipv4.

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1059,17 +1055,18 @@ func (bnc *BaseNetworkController) shouldReleaseDeletedPod(pod *corev1.Pod, switc
...
-		}
+	// If this pod applies to live migration it could have migrated within the
+	// zone so get the correct node name corresponding with the subnet. If the
+	// subnet is not tracked within the zone, subnetName will be empty which
+	// will force a lookup for all nodes.
+	subnetName := switchName
+	if !bnc.IsUserDefinedNetwork() && kubevirt.IsPodLiveMigratable(pod) {
+		subnetName, _ = bnc.lsManager.GetSubnetName(podIfAddrs)
+	}
+	lookupAllNodes := subnetName == ""
```
qinqon: Looks like we don't use `lookupAllNodes` ?

qinqon (changes requested):
First pass, not nits.

go-controller/pkg/ovn/pods.go
jcaamano: This has already been checked and decided in `deletePodLogicalPort->shouldReleaseDeletedPod`

go-controller/pkg/ovn/kubevirt_test.go
jcaamano: That issue was already there since the test is already allocating `/64` subnets for nodes
```
			node1: {
				nodeID:                "4",
				subnetIPv4:            "10.128.1.0/24",
				subnetIPv6:            "fd11::/64",
				lrpNetworkIPv4:        "100.64.0.4/24",
				lrpNetworkIPv6:        "fd98::4/64",
				transitSwitchPortIPv4: "100.65.0.4/24",
				transitSwitchPortIPv6: "fd13::4/64",
				addressIPv4:           "10.89.0.1/24",
				addressIPv6:           "fc00:f853:ccd:e793::1/64",
			},
...
```

It doesn't seem to have an impact so not fixing.

go-controller/pkg/ovn/base_network_controller_pods.go
jcaamano: yes, line 1082
```
if !bnc.allocatesPodAnnotation() || lookupAllNodes {
```

qinqon (approved):
LGTM

thanks @jcaamano !

PR #5632, opened 2025-10-06
[IC][EIP] OCPBUGS-42303: Fix race between SB DB and OVN-Controller at startup

martinkennelly:
Ensure ovn-controller has processed the SB DB updates before removing the GARP drop flows by utilizing the hv_cfg field in NB_Global [1]

OVNKube controller increments the nb_cfg value post sync, which is copied to SB DB by northd. OVN-Controllers copy this nb_cfg value from SB DB and write it to their chassis_private tables nb_cfg field after they have processed the SB DB changes. Northd will then look at all the chassis_private tables nb_cfg value and set the NB DBs Nb_global hv_cfg value to the min integer found.

Since IC currently only supports one node per zone, we can be sure ovn-controller is running locally and therefore its ok to block removing the drop GARP flows.

[1] https://man7.org/linux/man-pages/man5/ovn-nb.5.html

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

- Bug Fixes
  - Improves startup reliability by waiting for northbound changes to be processed and resulting flows to be installed, reducing transient connectivity issues.

- Refactor
  - Simplifies synchronization to rely on a single readiness signal from the northbound processing flow rather than cross-checking multiple DB sources.

- Chores
  - Clarifies startup logs and error messages to reflect waiting for northbound changes to be processed by the controller.

### DIFF
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -528,12 +528,12 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 			metrics.MetricOVNKubeControllerReadyDuration.Set(time.Since(startTime).Seconds())
 
 			if isOVNKubeControllerSyncd != nil {
-				klog.Infof("Waiting for OVN northbound database changes to sync to OVN Southbound database")
-				if err = libovsdbutil.WaitUntilNorthdSyncOnce(ctx, libovsdbOvnNBClient, libovsdbOvnSBClient); err != nil {
-					controllerErr = fmt.Errorf("failed waiting for northd to sync OVN Northbound DB to Southbound: %v", err)
+				klog.Infof("Waiting for OVN northbound database changes to be processed by ovn-controller")
+				if err = libovsdbutil.WaitUntilFlowsInstalled(ctx, libovsdbOvnNBClient); err != nil {
+					controllerErr = fmt.Errorf("failed waiting for OVN northbound database changes to be processed by ovn-controller: %v", err)
 					return
 				} else {
-					klog.Infof("OVN northbound database changes synced to OVN Southbound database")
+					klog.Infof("Finished waiting for OVN northbound database changes to be processed by ovn-controller")
 					isOVNKubeControllerSyncd.Store(true)
 				}
 			}
--- a/go-controller/pkg/libovsdb/util/northd_sync.go
+++ b/go-controller/pkg/libovsdb/util/northd_sync.go
@@ -14,26 +14,26 @@ import (
 
 	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/nbdb"
-	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/sbdb"
 )
 
-// WaitUntilNorthdSyncOnce ensures northd has sync'd at least once by increments nb_cfg value in NB DB and waiting
-// for northd to copy it to SB DB. Poll SB DB until context is cancelled.
-// The expectation is that the data you wish to be sync'd to SB DB has already been written to NB DB so when we get the initial
-// nb_cfg value, we know that if we increment that by one and see that value or greater in SB DB, then the data has sync'd.
-// All other processes interacting with nb_cfg increment it. This function depends on other processes respecting that.
-// No guarantee of any changes in SB DB made after this func.
-func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Client) error {
+// WaitUntilFlowsInstalled ensures that ovn-controller has sync'd at least once by incrementing nb_cfg value in NB DB
+// and waiting for northd to write back a value equal or greater to the hv_cfg field in NB_Global.
+// See https://www.ovn.org/support/dist-docs/ovn-nb.5.html for more info regarding nb_cfg / hv_cfg fields.
+// The expectation is that the data you wish to be sync'd is already written to NB DB.
+// Note: if the ovn-controller is down, this will block until it comes back up, therefore this func should only
+// be used with IC mode and one node per zone.
+func WaitUntilFlowsInstalled(ctx context.Context, nbClient client.Client) error {
 	// 1. Get value of nb_cfg
 	// 2. Increment value of nb_cfg
-	// 3. Wait until value appears in SB DB after northd copies it.
+	// 3. Wait until value appears in hv_cfg field thus ensuring ovn-controller has processed the changes
 	nbGlobal := &nbdb.NBGlobal{}
 	nbGlobal, err := libovsdbops.GetNBGlobal(nbClient, nbGlobal)
 	if err != nil {
 		return fmt.Errorf("failed to find OVN Northbound NB_Global table"+
 			" entry: %w", err)
 	}
-	// increment nb_cfg value by 1. When northd consumes updates from NB DB, it will copy this value to SB DBs SB_Global table nb_cfg field.
+	// increment nb_cfg value by 1. When northd consumes updates from NB DB, it will copy this value to SB DBs SB_Global
+	// table nb_cfg field.
 	ops, err := nbClient.Where(nbGlobal).Mutate(nbGlobal, model.Mutation{
 		Field:   &nbGlobal.NbCfg,
 		Mutator: ovsdb.MutateOperationAdd,
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
 	if err != nil {
 		return fmt.Errorf("failed to generate ops to mutate nb_cfg: %w", err)
 	}
-	expectedNbCfgValue := nbGlobal.NbCfg + 1
 	if _, err = libovsdbops.TransactAndCheck(nbClient, ops); err != nil {
 		return fmt.Errorf("failed to transact to increment nb_cfg: %w", err)
 	}
-	sbGlobal := &sbdb.SBGlobal{}
-	// poll until we see the expected value in SB DB every 5 milliseconds until context is cancelled.
-	err = wait.PollUntilContextCancel(ctx, time.Millisecond*5, true, func(_ context.Context) (done bool, err error) {
-		if sbGlobal, err = libovsdbops.GetSBGlobal(sbClient, sbGlobal); err != nil {
+	expectedHvCfgValue := nbGlobal.NbCfg + 1
+	if expectedHvCfgValue < 0 { // handle overflow
+		expectedHvCfgValue = 0
+	}
+	nbGlobal = &nbdb.NBGlobal{}
+	// ovn-northd sets hv_cfg to the lowest int value found for all chassis in the system (IC mode,
+	// we support a single chassis per zone) as reported in the Chassis_Private table in the southbound database.
+	// Thus, hv_cfg equals nb_cfg for the single chassis once it is caught up with NB DB we want.
+	// poll until we see the expected value in NB DB every 5 milliseconds until context is cancelled.
+	err = wait.PollUntilContextCancel(ctx, time.Millisecond*5, true, func(_ context.Context) (done bool, err2 error) {
+		if nbGlobal, err2 = libovsdbops.GetNBGlobal(nbClient, nbGlobal); err2 != nil {
 			// northd hasn't added an entry yet
-			if errors.Is(err, client.ErrNotFound) {
+			if errors.Is(err2, client.ErrNotFound) {
 				return false, nil
 			}
-			return false, fmt.Errorf("failed to get sb_global table entry from SB DB: %w", err)
+			return false, fmt.Errorf("failed to get nb_global table entry from NB DB: %w", err2)
 		}
-		return sbGlobal.NbCfg >= expectedNbCfgValue, nil // we only need to ensure it is greater than or equal to the expected value
+		return nbGlobal.HvCfg >= expectedHvCfgValue, nil // we only need to ensure it is greater than or equal to the expected value
 	})
 	if err != nil {
-		return fmt.Errorf("failed while waiting for nb_cfg value greater than or equal %d in sb db sb_global table: %w", expectedNbCfgValue, err)
+		return fmt.Errorf("failed while waiting for hv_cfg value greater than or equal %d in NB DB nb_global table: %w",
+			expectedHvCfgValue, err)
 	}
 	return nil
 }


go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -14,26 +14,26 @@ import (
...
+	// 3. Wait until value appears in hv_cfg field thus ensuring all ovn-controllers have processed the changes.
 	nbGlobal := &nbdb.NBGlobal{}
 	nbGlobal, err := libovsdbops.GetNBGlobal(nbClient, nbGlobal)
 	if err != nil {
 		return fmt.Errorf("failed to find OVN Northbound NB_Global table"+
 			" entry: %w", err)
 	}
-	// increment nb_cfg value by 1. When northd consumes updates from NB DB, it will copy this value to SB DBs SB_Global table nb_cfg field.
+	// increment nb_cfg value by 1. When northd consumes updates from NB DB, it will copy this value to SB DBs SB_Global
+	// table nb_cfg field.
```
almusil, line 36: One note, the `nb_cfg` is 64 bit signed interger, go seems to wrap into negative value. However that's not the same way we handle that in OVN, when we see `nb_cfg` being the int64 max value we will set it to 0 then increment. The current code doesn't seem to do it (maybe it's hidden in the libovsdb), but it would be a good idea IMO to align those two.

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,27 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
...
+		if nbGlobal, err = libovsdbops.GetNBGlobal(nbClient, nbGlobal); err != nil {
 			// northd hasn't added an entry yet
 			if errors.Is(err, client.ErrNotFound) {
 				return false, nil
 			}
-			return false, fmt.Errorf("failed to get sb_global table entry from SB DB: %w", err)
+			return false, fmt.Errorf("failed to get nb_global table entry from NB DB: %w", err)
 		}
-		return sbGlobal.NbCfg >= expectedNbCfgValue, nil // we only need to ensure it is greater than or equal to the expected value
+		return nbGlobal.HvCfg >= expectedHvCfgValue, nil // we only need to ensure it is greater than or equal to the expected value
```
almusil: And I forgot to add also here, during the wrap being it negative or 0 as mentioned above, this will always be true. So it will return prematurely. For that we do extra check in form of `nbGlobal.HvCfg - expectedHvCfgValue < int32.maxValue`, so the whole condition would be `return nbGlobal.HvCfg >= expectedHvCfgValue && nbGlobal.HvCfg - expectedHvCfgValue < int32.maxValue, nil`.

go-controller/pkg/libovsdb/util/northd_sync.go
martinkennelly: Thanks for this. I didnt consider that here.

We fetch the nb_cfg value here and if its negative, ill add a check to ensure its set to 0. Thatll handle my immediate need without considering libovsdb. Golang has defined behavior for overflow that allows this.

go-controller/pkg/libovsdb/util/northd_sync.go
martinkennelly: I see the value to 0 if its negative so we can avoid this.

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -528,12 +528,12 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
 			metrics.MetricOVNKubeControllerReadyDuration.Set(time.Since(startTime).Seconds())
 
 			if isOVNKubeControllerSyncd != nil {
-				klog.Infof("Waiting for OVN northbound database changes to sync to OVN Southbound database")
-				if err = libovsdbutil.WaitUntilNorthdSyncOnce(ctx, libovsdbOvnNBClient, libovsdbOvnSBClient); err != nil {
-					controllerErr = fmt.Errorf("failed waiting for northd to sync OVN Northbound DB to Southbound: %v", err)
+				klog.Infof("Waiting for OVN northbound database changes to sync to all nodes")
```
tssurya: this log seems a bit confusing and the one on L536
nbdb is per node (at least in IC) -> so waiting for it to sync to all nodes..... means?

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -528,12 +528,12 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
+				if err = libovsdbutil.WaitUntilFlowsInstalled(ctx, libovsdbOvnNBClient); err != nil {
+					controllerErr = fmt.Errorf("failed waiting for northd to sync OVN Northbound DB to all nodes: %v", err)
```
tssurya: nit: I think its good to add we also wait for ovn-controller to sync and update sbdb right not just northd

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -14,26 +14,26 @@ import (
...
 )
 
-// WaitUntilNorthdSyncOnce ensures northd has sync'd at least once by increments nb_cfg value in NB DB and waiting
-// for northd to copy it to SB DB. Poll SB DB until context is cancelled.
-// The expectation is that the data you wish to be sync'd to SB DB has already been written to NB DB so when we get the initial
-// nb_cfg value, we know that if we increment that by one and see that value or greater in SB DB, then the data has sync'd.
-// All other processes interacting with nb_cfg increment it. This function depends on other processes respecting that.
-// No guarantee of any changes in SB DB made after this func.
-func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Client) error {
+// WaitUntilFlowsInstalled ensures that all ovn-controllers have sync'd at least once by increments nb_cfg value in NB DB
```
tssurya: isn't it just 1 ovn-controller? (let's forget icDisabled since its not default anymore..)

tssurya:
```suggestion
// WaitUntilFlowsInstalled ensures that all ovn-controllers have sync'd at least once by incrementing nb_cfg value in NB DB
```

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -14,26 +14,26 @@ import (
...
+// and waiting for northd to write back a value equal or greater to the hv_cfg field.
+// See https://www.ovn.org/support/dist-docs/ovn-nb.5.html for more info regarding nb_cfg / hv_cfg fields.
+// The expectation is that the data you wish to be sync'd and installed by all ovn-controllers has already been written to NB DB.
+// Note: if any ovn-controllers are down, this will block until they come back up, therefore this func should only
+// be used in IC mode and one node per zone.
+func WaitUntilFlowsInstalled(ctx context.Context, nbClient client.Client) error {
 	// 1. Get value of nb_cfg
 	// 2. Increment value of nb_cfg
-	// 3. Wait until value appears in SB DB after northd copies it.
+	// 3. Wait until value appears in hv_cfg field thus ensuring all ovn-controllers have processed the changes.
```
tssurya: again.. just 1 ovn-controller right? specially since you have a comment that says don't use this in nonIC?

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
 	if err != nil {
 		return fmt.Errorf("failed to generate ops to mutate nb_cfg: %w", err)
 	}
-	expectedNbCfgValue := nbGlobal.NbCfg + 1
 	if _, err = libovsdbops.TransactAndCheck(nbClient, ops); err != nil {
 		return fmt.Errorf("failed to transact to increment nb_cfg: %w", err)
 	}
-	sbGlobal := &sbdb.SBGlobal{}
-	// poll until we see the expected value in SB DB every 5 milliseconds until context is cancelled.
+	expectedNbCfgValue := nbGlobal.NbCfg + 1
```
tssurya: I shouldn't have suggested this on the DM
facepalm
now that I have carefully read the code I admit I was wrong
this should indeed be `expectedHvCfgValue`
```
Martin Kennelly
 hey, i dont think its good actually to rename expectedHvCfgValue to expectedNbCfgValue like we suggested cause it gets more confusing imo. We are trying to compare that value to hvCfg field.
```
you are right! :) my head was elsewhere.. sorry.

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
...
+	if expectedNbCfgValue < 0 { // handle overflow
+		expectedNbCfgValue = 0
+	}
+	nbGlobal = &nbdb.NBGlobal{}
+	// ovn-northd sets hv_cfg to the smallest sequence number of all the chassis in the system,
+	// as reported in the Chassis_Private table in the southbound database. Thus, hv_cfg
```
tssurya: and for IC this is just 1 chassis per node/zone right?

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
...
 		return fmt.Errorf("failed to generate ops to mutate nb_cfg: %w", err)
 	}
-	expectedNbCfgValue := nbGlobal.NbCfg + 1
 	if _, err = libovsdbops.TransactAndCheck(nbClient, ops); err != nil {
 		return fmt.Errorf("failed to transact to increment nb_cfg: %w", err)
 	}
-	sbGlobal := &sbdb.SBGlobal{}
-	// poll until we see the expected value in SB DB every 5 milliseconds until context is cancelled.
+	expectedNbCfgValue := nbGlobal.NbCfg + 1
+	if expectedNbCfgValue < 0 { // handle overflow
```
tssurya: ack I see so in case NbCfg increment mutation above overflowed within the transactandcheck, then ovn anyways overflows and sets it to 0 like ales said? so we follow same approach here.. sgtm

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
...
+		expectedNbCfgValue = 0
+	}
+	nbGlobal = &nbdb.NBGlobal{}
+	// ovn-northd sets hv_cfg to the smallest sequence number of all the chassis in the system,
+	// as reported in the Chassis_Private table in the southbound database. Thus, hv_cfg
+	// equals nb_cfg if all chassis are caught up with NB DB.
+	// poll until we see the expected value in NB DB every 5 milliseconds until context is cancelled.
 	err = wait.PollUntilContextCancel(ctx, time.Millisecond*5, true, func(_ context.Context) (done bool, err error) {
-		if sbGlobal, err = libovsdbops.GetSBGlobal(sbClient, sbGlobal); err != nil {
+		if nbGlobal, err = libovsdbops.GetNBGlobal(nbClient, nbGlobal); err != nil {
```
tssurya: super nit: hmm I think its better to have err and err1 for when poll unit context cancel fails v/s the internal errors for that to avoid variable shadowing

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
...
 			// northd hasn't added an entry yet
 			if errors.Is(err, client.ErrNotFound) {
 				return false, nil
 			}
-			return false, fmt.Errorf("failed to get sb_global table entry from SB DB: %w", err)
+			return false, fmt.Errorf("failed to get nb_global table entry from NB DB: %w", err)
```
tssurya: don't we want to poll for the get again? do we want to immediately fail here if we didn't find global?

tssurya (commented):
A few questions otherwise lgtm!
thanks @almusil and @martinkennelly for this

for the future me: 

1. ovnkube-node on start up adds the drop flows to br-ex
2. controllerManager.Start does the syncs in ovnkube-controller side
3. before we pass the isOVNKubeControllerSyncd down to ovnkube-node.... 
4. we call WaitUntilFlowsInstalled which does the following
     - increments the nbcfg in nb-global
     - waits for northd to copy that value to nb_cfg in sb-global
     - northd sets to the value of nb_cfg in nb-global after it finishes applying the corresponding configuration changes to the OVN_Southbound database
     - ovn-controller sets the value of nb_cfg in chassis_private table in SBDB after it programs the flows corresponding to what CMS has freshly passed down i.e latest data and this also ensures ovn-controller has removed the stale garp requests in its pinctrl queue - even if pinctrl is another thread ovn-controller holds the lock - so we know that inc-idl engine and all the other ctrl threads have finished running correctly - there is no race there (there is small corner case race where maybe ovs has not yet received these updates correctly)
     - northd then again sets the hv_cfg value as the min of all chassis-private nb_cfg values which probably in IC is just 1 chassis value
     - once all that is done we are sure ovn-controller and northd have done everything correctly
5. ovnkube now checks hvcfg value written by northd is equal to the nbcfg value we expect through the increment we did (ofc there could be other things altering the values so we only care about equal or greater)
6. at this point we set isOVNKubeControllerSyncd to true
7. using this ovnkube-node removes the drop flows for GARP

This is all temporary and expectation is for us to use https://patchwork.ozlabs.org/project/ovn/patch/67a5c2813f77c67f6069a3766edaf28982e2f81a.1757022110.git.lorenzo.bianconi@redhat.com/ in the future when it gets released

go-controller/cmd/ovnkube/ovnkube.go
martinkennelly: its IC only and we only support one Node but its possible folks configure more than one Node per zone. We dont actually enforce one node per zone so I used the plural for Node.

martinkennelly: "Waiting for OVN northbound database changes to be processed by ovn-controller"

I wont get any more indepth here.

go-controller/pkg/libovsdb/util/northd_sync.go
martinkennelly: was thinking of more than one node per zone.. but ya, i agree. Ill fix this.

go-controller/pkg/libovsdb/util/northd_sync.go
martinkennelly: yep - if we assume 1 node per zone. Ill change it to that.

go-controller/pkg/libovsdb/util/northd_sync.go
martinkennelly: yep

go-controller/pkg/libovsdb/util/northd_sync.go
martinkennelly: if we hit this err, then its an err other than failing to find a NB Global table entry. 
I think its appropriate to return and start again. We would only hit this if the cache is broken i think.

go-controller/pkg/libovsdb/util/northd_sync.go
```diff
@@ -42,24 +42,31 @@ func WaitUntilNorthdSyncOnce(ctx context.Context, nbClient, sbClient client.Clie
...
 		return fmt.Errorf("failed to transact to increment nb_cfg: %w", err)
 	}
-	sbGlobal := &sbdb.SBGlobal{}
-	// poll until we see the expected value in SB DB every 5 milliseconds until context is cancelled.
-	err = wait.PollUntilContextCancel(ctx, time.Millisecond*5, true, func(_ context.Context) (done bool, err error) {
-		if sbGlobal, err = libovsdbops.GetSBGlobal(sbClient, sbGlobal); err != nil {
+	expectedHvCfgValue := nbGlobal.NbCfg + 1
+	if expectedHvCfgValue < 0 { // handle overflow
+		expectedHvCfgValue = 0
+	}
```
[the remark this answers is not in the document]
tssurya: @martinkennelly I think both of these are legit right?
there is a corner cases here where we increment it to int64 on our side and on OVN side something else also changes it - its already overflowed and now it will be 1 >= intmax situation? its far fetched but could happen

and the race condition point is also valid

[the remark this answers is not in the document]
almusil: Speaking about the race condition in general, it would be nice to use the IDL infrastructure to do the increment atomically for you.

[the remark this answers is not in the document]
tssurya: @almusil could you elaborate that a bit more - any links to the IDL infra you mention?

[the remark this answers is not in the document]
almusil: The C idl (no idea about the go one) has a function called `ovsdb_idl_txn_increment()` that will take care of incrementing certain column and failing the transaction if it was incremented by someone else in the meantime.
The value after transaction can be retrieved by `ovsdb_idl_txn_get_increment_new_value()`. So what OVN does the following:
1) If the `nb_cfg` value == int64 max, set it to 0.
2) Call the increment (so it should end up as 1 after wrap).
3) After the txn is complete call `ovsdb_idl_txn_get_increment_new_value` to get the value and compare that to `hv_cfg`.

Does that make sense?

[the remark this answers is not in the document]
tssurya: yeah I see - that makes sense! more tighter to do that but im not sure if there is a libovsdb ops library for this..

In go we seem to do 
ops, err := nbClient.Where(nbGlobal).Mutate(nbGlobal, model.Mutation{
		Field:   &nbGlobal.NbCfg,
		Mutator: ovsdb.MutateOperationAdd,
		Value:   1,
	})
an atomic mutate so its gonna happen no matter what - but we don't seem to fetch that value which ensures its actually our client's update and not something else.
But if we do fetch the latest from the database instead of relying on our local value that will help. - although there is always that added risk of something else updating it in the interim i.e
```
// 1. Our increment makes nb_cfg = INT64_MAX
expectedNbCfgValue = INT64_MAX  // We fetch this value

// 2. Another process increments nb_cfg 
// OVN: INT64_MAX + 1 → wraps to 0, then increments to 1
// Database now has nb_cfg = 1, and hv_cfg will eventually reach 1

// 3. Our polling condition
return nbGlobal.HvCfg >= expectedNbCfgValue  // 1 >= INT64_MAX = FALSE ❌
// This will NEVER be true!
```
seems icky to handle the if maxInt64 the expected value or close to it v/s things have gotten reset to 0 and onwards then we treat it as true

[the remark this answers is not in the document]
tssurya: cc @dave-tucker if you have better ideas with libovsdb calls

go-controller/pkg/libovsdb/util/northd_sync.go
tssurya: or some network connecitivity issues or glitches where we might want to try again..

go-controller/pkg/libovsdb/util/northd_sync.go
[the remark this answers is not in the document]
dave-tucker: From [what I can see of the C IDL](https://github.com/openvswitch/ovs/blob/0f8202e60845743b00b5fc6991b96f6827937c4c/lib/ovsdb-idl.c#L3442-L3471) the process is simply to:
- Use a Mutate operation to increment the field by 1
- Use a Select operation to fetch the value of the field after mutation

The good news is that libovsdb just got support for Select operations. I can do a release + bump in ovn-k8s for you to pick this up if you want.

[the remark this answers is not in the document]
martinkennelly: > Race condition: If another process increments nb_cfg after line 30 but before the mutation at line 45 completes, the local nbGlobal.NbCfg is stale. For example, if we read nb_cfg = 100, another actor increments it to 101, our mutation increments it to 102, but we set expectedHvCfgValue = 101. Then hv_cfg might reach 101 (from the other actor's change) before processing our increment to 102, causing premature return.

I dont believe there is a problem here. 

This func tries to accomplish that the data written in NB DB written before this func is called is applied - basically we want to ensure our sync funcs have finished and the data is consumed by ovn-controller. We use the ever increasing nb_cfg int as a versioning mechanism. We dont mind if other processes increment this value. We just want nb_cfg+1 (at the time we fetched it) or greater to be there and we can be sure the data that was written before this func has propagated.
Its fine to prematurely return if we see > nb_cfg value there because we know the changes before this func have been propagated down to the ovn-controller.

> Overflow gap: The current check only handles Go arithmetic overflow (wraps to negative), but not the case where nb_cfg becomes int64 max in the DB. According to the OVN schema, when nb_cfg reaches int64 max, OVN wraps it to 0 then increments to 1. However, if our mutation results in nb_cfg = int64 max, we compute expectedHvCfgValue = int64 max (not negative, so no reset), but hv_cfg will be 1 after OVN's wrap. The comparison nbGlobal.HvCfg >= expectedHvCfgValue will never be true.


I think this is valid.
Lets say there are two processes working on this field. Lets say we fetch nb_cfg as int64 max minus 1. Lets say another process increments nb_cfg to int64 max in NB DB. Now, we increment and expect int64 max but the ovn dbs have wrapped around to 0, so our condition never becomes true.

However, the field isnt used today in producition and its only incremented on startup. It would take a long time for this to overflow. ~Can I fix this in follow up?~ we will remove this when the ovn fix lands. No need to fix it.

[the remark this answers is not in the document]
tssurya: Spoke with @martinkennelly offline
Overflow case:
given this is a super super corner case (9,223,372,036,854,775,807), let's get this fix in and we know that in the next month or so we want to get https://patchwork.ozlabs.org/project/ovn/patch/67a5c2813f77c67f6069a3766edaf28982e2f81a.1757022110.git.lorenzo.bianconi@redhat.com/ in as the proper path forward that OVN did.

Race condition case: tl;dr would have been good to do the > fetched value instead of doing >= fetched+1 value which are the same things - cause that's what causes the confusion on - oh wait we don't seem to be comparing with the mutated value actually - but regardless the version being higher is what we are after compared to what we fetched,

thanks @dave-tucker ! good to know we got select support!

PR #5633, opened 2025-10-06
fix: delete ovs-node pods on reinstall

booxter:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * KIND-based OVN deployments now include the ovs-node pod by default, providing a more complete networking stack out of the box.

* **Chores**
  * Updated KIND setup defaults so the ovs-node pod is provisioned automatically during OVN deployments.

### DIFF
--- a/contrib/kind.sh
+++ b/contrib/kind.sh
@@ -631,7 +631,7 @@ set_default_params() {
   OVN_HOST_NETWORK_NAMESPACE=${OVN_HOST_NETWORK_NAMESPACE:-ovn-host-network}
   OVN_EGRESSIP_HEALTHCHECK_PORT=${OVN_EGRESSIP_HEALTHCHECK_PORT:-9107}
   OCI_BIN=${KIND_EXPERIMENTAL_PROVIDER:-docker}
-  OVN_DEPLOY_PODS=${OVN_DEPLOY_PODS:-"ovnkube-identity ovnkube-zone-controller ovnkube-control-plane ovnkube-master ovnkube-node"}
+  OVN_DEPLOY_PODS=${OVN_DEPLOY_PODS:-"ovnkube-identity ovnkube-zone-controller ovnkube-control-plane ovnkube-master ovnkube-node ovs-node"}
   OVN_METRICS_SCALE_ENABLE=${OVN_METRICS_SCALE_ENABLE:-false}
   OVN_ISOLATED=${OVN_ISOLATED:-false}
   OVN_GATEWAY_OPTS=${OVN_GATEWAY_OPTS:-""}


contrib/kind.sh
```diff
@@ -631,7 +631,7 @@ set_default_params() {
   OVN_HOST_NETWORK_NAMESPACE=${OVN_HOST_NETWORK_NAMESPACE:-ovn-host-network}
   OVN_EGRESSIP_HEALTHCHECK_PORT=${OVN_EGRESSIP_HEALTHCHECK_PORT:-9107}
   OCI_BIN=${KIND_EXPERIMENTAL_PROVIDER:-docker}
-  OVN_DEPLOY_PODS=${OVN_DEPLOY_PODS:-"ovnkube-identity ovnkube-zone-controller ovnkube-control-plane ovnkube-master ovnkube-node"}
+  OVN_DEPLOY_PODS=${OVN_DEPLOY_PODS:-"ovnkube-identity ovnkube-zone-controller ovnkube-control-plane ovnkube-master ovnkube-node ovs-node"}
```
trozet, line 634: @booxter later in the kind script: 

```
  # When using internal registry force pod reload just the ones with 
  # non OVS containers, restarting OVS pods breaks the cluster.
  if [ "${KIND_CREATE}" == false ] && [ "${KIND_LOCAL_REGISTRY}" == false ] ; then
    for pod in ${OVN_DEPLOY_PODS}; do
        run_kubectl delete pod -n ovn-kubernetes -l name=$pod
    done
  fi
```

it looks like OVS was intentionally not touched.

booxter: This code comes from the following change:

```diff
-  # Force pod reload just the ones with golang containers
-  if [ "${KIND_CREATE}" == false ]; then
+  # When using internal registry force pod reload just the ones with
+  # non OVS containers, restarting OVS pods breaks the cluster.
+  if [ "${KIND_CREATE}" == false ] && [ "${KIND_LOCAL_REGISTRY}" == false ] ; then
```

Before the change, we were deleting all (non-OVS) pods. After the change we are deleting all (non-OVS) pods *if* local registry is not used (off by default).  I am not sure what the issue this code is trying to solve for local registry case, but the list of pods to restart is from before the local registry support was added.

In my local testing with this change, there seemed to be no issue with cluster connectivity when using (default) kind mechanism to push images.

---

Maybe @qinqon could clarify why OVS pod is special here (as the comment suggests). Or maybe what was meant is that *OVN(-K)* pods should not be touched if local registry is used (because the OVN-K pods maintain connectivity to the local registry)?

booxter:
@koolzz noticed that this patch got into a conflict due to helm migration. Does the helm path handle `kind.sh --deploy` correctly by triggering an update to pods after new image is loaded into kind nodes? Doesn't look like it.

koolzz:
> @koolzz noticed that this patch got into a conflict due to helm migration. Does the helm path handle `kind.sh --deploy` correctly by triggering an update to pods after new image is loaded into kind nodes? Doesn't look like it.

Yeah this is a good find, I think I missed this during the migration because the CI tests change tag name.

Regarding this PR though, it breaks for me (ovs-node addition specifically). I posted a fix to bring back the pod del [here](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6408)

booxter:
Mostly covered in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6142 and not sure if there's more needed here.

PR #5634, opened 2025-10-06
metrics: move OVN/OVS metrics update from periodic interval to request-based

### LINKED ISSUE #5585: Move OVS metrics update from periodic interval to request-based
### What is this issue about?

Currently OVS metrics are updated on a fixed interval(default: 30s):

```
func registerOvsMetrics(ovsDBClient libovsdbclient.Client, metricsScrapeInterval int, registry prometheus.Registerer, stopChan <-chan struct{}) {
	registerOvsMetricsOnce.Do(func() {
                ... 

		// OVS datapath metrics updater
		go ovsDatapathMetricsUpdater(util.RunOVSAppctl, metricsScrapeInterval, stopChan)
		// OVS bridge metrics updater
		go ovsBridgeMetricsUpdater(ovsDBClient, util.RunOVSOfctl, metricsScrapeInterval, stopChan)
		// OVS interface metrics updater
		go ovsInterfaceMetricsUpdater(ovsDBClient, metricsScrapeInterval, stopChan)
		// OVS memory metrics updater
		go ovsMemoryMetricsUpdater(util.RunOvsVswitchdAppCtl, metricsScrapeInterval, stopChan)
		// OVS hw Offload metrics updater
		go ovsHwOffloadMetricsUpdater(ovsDBClient, metricsScrapeInterval, stopChan)
		// OVS coverage/show metrics updater.
		go coverageShowMetricsUpdater(ovsVswitchd, stopChan)
	})
}
```

This means that:
- CPU cycles are consumed to update metrics even if no Prometheus server is scraping them
-  Changing the update interval requires redeployment.

It's better to update OVS metrics only when handling a /metrics request.

l8huang:
## 📑 Description

Currently OVN/OVS metrics are updated on a fixed interval(default: 30s), so that:
- CPU cycles are consumed to update metrics even if no Prometheus server is scraping them.
- Changing the update interval requires redeployment.

This change collects OVS metrics only when handling a /metrics request.
 
Fixes #5585

## ✅ Checks

- [x] My code requires tests

## How to verify it
Covered by unit tests.

## Summary by CodeRabbit

* **Refactor**
  * Centralized metrics server with on-demand metric updates, registry-driven registration, and coordinated graceful shutdown.

* **New Features**
  * Runtime toggling of component metrics; dynamic enabling when master pod is detected.

* **Chores**
  * Removed per-interval metrics flag; metrics configured via server options (bind/TLS/per-component).
  * Made command-runner hook public to improve testability.

* **Tests**
  * Added comprehensive metrics server tests and improved test state cleanup.

### DIFF
--- a/go-controller/pkg/metrics/server.go
+++ b/go-controller/pkg/metrics/server.go
@@ -5,13 +5,11 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
-	"io"
 	"net/http"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"github.com/prometheus/common/expfmt"
 
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	utilwait "k8s.io/apimachinery/pkg/util/wait"
@@ -80,12 +78,22 @@ func NewMetricServer(opts MetricServerOptions, ovsDBClient libovsdbclient.Client
 	}
 
 	server.mux = http.NewServeMux()
-	server.mux.Handle("/metrics", promhttp.InstrumentMetricHandler(server.ovnRegistry, http.HandlerFunc(server.handleMetrics)))
+	metricsHandler := promhttp.HandlerForTransactional(
+		prometheus.ToTransactionalGatherer(server.ovnRegistry),
+		promhttp.HandlerOpts{},
+	)
+	server.mux.Handle("/metrics", promhttp.InstrumentMetricHandler(
+		server.ovnRegistry,
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Update metrics in the registry before emitting them.
+			server.handleMetrics(r)
+			// Emit the updated metrics using the transactional handler.
+			metricsHandler.ServeHTTP(w, r)
+		}),
+	))
 
 	return server
 }
-
-const FmtText = `text/plain; version=` + expfmt.TextVersion + `; charset=utf-8`
 
 // registerMetrics registers the metrics to the OVN registry
 func (s *MetricServer) registerMetrics() {
@@ -117,21 +125,6 @@ func (s *MetricServer) EnableOVNDBMetrics() {
 	s.opts.EnableOVNDBMetrics = true
 	klog.Infof("MetricServer registers OVN DB metrics")
 	s.ovsDbProperties, s.opts.dbIsClustered, s.opts.dbFoundViaPath = RegisterOvnDBMetrics(s.ovnRegistry)
-}
-
-// writeRegisteredMetrics writes the registered metrics to the /metrics response.
-func (s *MetricServer) writeRegisteredMetrics(w io.Writer) error {
-	mfs, err := s.ovnRegistry.Gather()
-	if err != nil {
-		return err
-	}
-	enc := expfmt.NewEncoder(w, FmtText)
-	for _, mf := range mfs {
-		if err := enc.Encode(mf); err != nil {
-			return err
-		}
-	}
-	return nil
 }
 
 // updateOvsMetrics updates the OVS metrics
@@ -192,7 +185,7 @@ func (s *MetricServer) updateOvnDBMetrics() {
 }
 
 // handleMetrics handles the /metrics request
-func (s *MetricServer) handleMetrics(w http.ResponseWriter, r *http.Request) {
+func (s *MetricServer) handleMetrics(r *http.Request) {
 	klog.V(5).Infof("MetricServer starts to handle metrics request from %s", r.RemoteAddr)
 
 	if s.opts.EnableOVSMetrics {
@@ -206,13 +199,6 @@ func (s *MetricServer) handleMetrics(w http.ResponseWriter, r *http.Request) {
 	}
 	if s.opts.EnableOVNNorthdMetrics {
 		s.updateOvnNorthdMetrics()
-	}
-
-	// write out the registered metrics
-	w.Header().Set("Content-Type", FmtText)
-	if err := s.writeRegisteredMetrics(io.Writer(w)); err != nil {
-		klog.Errorf("Failed to write registered metrics: %v", err)
-		return
 	}
 }
 
--- a/go-controller/pkg/metrics/server_test.go
+++ b/go-controller/pkg/metrics/server_test.go
@@ -827,7 +827,7 @@ func TestHandleMetrics(t *testing.T) {
 			// Test the /metrics endpoint
 			rec := httptest.NewRecorder()
 			req := httptest.NewRequest("GET", "/metrics", nil)
-			server.handleMetrics(rec, req)
+			server.mux.ServeHTTP(rec, req)
 			if rec.Code != http.StatusOK {
 				t.Errorf("Expected status 200, got %d", rec.Code)
 			}
--- a/go-controller/go.mod
+++ b/go-controller/go.mod
@@ -41,7 +41,6 @@ require (
 	github.com/ovn-kubernetes/libovsdb v0.8.1
 	github.com/prometheus/client_golang v1.22.0
 	github.com/prometheus/client_model v0.6.1
-	github.com/prometheus/common v0.62.0
 	github.com/safchain/ethtool v0.3.1-0.20231027162144-83e5e0097c91
 	github.com/spf13/afero v1.14.0
 	github.com/stretchr/testify v1.10.0
@@ -120,6 +119,7 @@ require (
 	github.com/pborman/uuid v1.2.0 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
+	github.com/prometheus/common v0.62.0 // indirect
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/sirupsen/logrus v1.9.3 // indirect


go-controller/pkg/metrics/metrics.go
```diff
@@ -520,16 +503,29 @@ func StartMetricsServer(bindAddress string, enablePprof bool, certFile string, k
...
+
+	metricsServer.registerMetrics()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		klog.Infof("OVN Metrics Server starts to run ...")
+		if err := metricsServer.Run(stopChan); err != nil {
+			klog.Errorf("Failed to run OVN Metrics Server: %v", err)
+		}
```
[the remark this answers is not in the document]
l8huang: metricsServer.registerMetrics() doesn't return error

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -608,13 +608,37 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
+							runMode.identity)
+						opts.EnableOVNNorthdMetrics = false
+						opts.EnableOVNDBMetrics = false
+					} else {
+						klog.Info("Found OVNKube Master Pod running on this node, registering OVN Northd and DB Metrics")
+					}
+				}
+
+				metrics.StartOVNMetricsServer(opts, ovsClient, ovnClientset.KubeClient, ctx.Done(), wg)
+			}()
```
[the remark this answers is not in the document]
l8huang: this is a issue for central mode,  I will update this PR to fix it

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,283 @@
...
+	s.opts.EnableOVNNorthdMetrics = true
+	klog.Infof("MetricServer registers OVN Northd metrics")
+	RegisterOvnNorthdMetrics(s.ovnRegistry)
+}
+
+func (s *MetricServer) EnableOVNDBMetrics() {
+	s.opts.EnableOVNDBMetrics = true
+	klog.Infof("MetricServer registers OVN DB metrics")
+	RegisterOvnDBMetrics(s.ovnRegistry)
+}
```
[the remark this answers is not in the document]
l8huang: good catch

npinaeva:
that is a fairly large change, with lots of moving pieces, could you please split this into commits that would be easier to review? Especially if code just moved from one place to another, that would help a lot

l8huang:
> that is a fairly large change, with lots of moving pieces, could you please split this into commits that would be easier to review? Especially if code just moved from one place to another, that would help a lot

will do

go-controller/pkg/metrics/server_test.go
```diff
@@ -0,0 +1,824 @@
...
+		{FileName: "/var/run/openvswitch/ovnnb_db.sock", Permissions: 0o755, Content: []byte("")},
+		{FileName: "/var/run/openvswitch/ovnsb_db.sock", Permissions: 0o755, Content: []byte("")},
+		{FileName: "/etc/ovn/ovnnb_db.db", Permissions: 0o755, Content: []byte("abcdefgh")},
+		{FileName: "/etc/ovn/ovnsb_db.db", Permissions: 0o755, Content: []byte("xyz")},
+	}
+
+	for _, file := range files {
+		if err := afero.WriteFile(util.AppFs, file.FileName, file.Content, file.Permissions); err != nil {
+			t.Fatalf("failed to afero.WriteFile: %v", err)
+		}
```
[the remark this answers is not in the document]
l8huang: fixed

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -608,13 +608,50 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
+							klog.Infof("Not registering OVN Northd and DB Metrics, because OVNKube Master Pod is not running on this node(%s)",
+								runMode.identity)
+						} else {
+							klog.Info("Found OVNKube Master Pod running on this node, registering OVN Northd and DB Metrics")
+							metricsServer.EnableOVNNorthdMetrics()
+							metricsServer.EnableOVNDBMetrics()
+						}
+					}()
+				}
+			}()
```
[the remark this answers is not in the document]
l8huang: fixed.

go-controller/pkg/metrics/server_test.go
```diff
@@ -0,0 +1,824 @@
...
+	}
+}
```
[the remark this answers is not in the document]
l8huang: fixed.

l8huang:
split to 3 commits -- very little code was moved between files
@npinaeva please take a look

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,283 @@
...
+	klog.Infof("MetricServer registers OVN Northd metrics")
+	RegisterOvnNorthdMetrics(s.ovnRegistry)
+}
+
+func (s *MetricServer) EnableOVNDBMetrics() {
+	s.opts.EnableOVNDBMetrics = true
+	klog.Infof("MetricServer registers OVN DB metrics")
+	s.ovsDbProperties, s.opts.dbIsClustered, s.opts.dbFoundViaPath = RegisterOvnDBMetrics(s.ovnRegistry)
+}
+
```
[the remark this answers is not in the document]
l8huang: the double double-registration would not happen, see go-controller/cmd/ovnkube/ovnkube.go lines 624-629, if EnableInterconnect is false, opts.EnableOVNNorthdMetrics and opts.EnableOVNDBMetrics will be set to false, so registerMetrics() doesn't register them.

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,283 @@
...
+		if s.opts.dbIsClustered {
+			ovnDBClusterStatusMetricsUpdater(dbProperty)
+		}
+		if s.opts.dbFoundViaPath {
+			updateOvnDBSizeMetrics(dbProperty)
+		}
+		updateOvnDBMemoryMetrics(dbProperty)
+	}
+}
+
```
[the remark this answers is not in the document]
l8huang: this would not be a issue, EnableOVNDBMetrics() is only called when MetricServer start, while updateOvnDBMetrics() is called when handling HTTP request.

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,283 @@
...
+		s.updateOvnNorthdMetrics()
+	}
+
+	// write out the registered metrics
+	w.Header().Set("Content-Type", FmtText)
+	if err := s.writeRegisteredMetrics(io.Writer(w)); err != nil {
+		klog.Errorf("Failed to write registered metrics: %v", err)
+		return
+	}
+}
```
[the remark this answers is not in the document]
l8huang: not a issue, similar as previous

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -608,13 +608,51 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
+						if err != nil {
+							klog.Infof("Not registering OVN Northd and DB Metrics, because OVNKube Master Pod is not running on this node(%s)",
+								runMode.identity)
+						} else {
+							klog.Info("Found OVNKube Master Pod running on this node, registering OVN Northd and DB Metrics")
+							metricsServer.EnableOVNNorthdMetrics()
+							metricsServer.EnableOVNDBMetrics()
+						}
+					}()
+				}
```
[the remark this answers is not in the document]
l8huang: the inner go routine is eliminated

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -608,13 +608,49 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
+					if err != nil {
+						klog.Infof("Not registering OVN Northd and DB Metrics, because OVNKube Master Pod is not running on this node(%s)",
+							runMode.identity)
+					} else {
+						klog.Info("Found OVNKube Master Pod running on this node, registering OVN Northd and DB Metrics")
+						metricsServer.EnableOVNNorthdMetrics()
+						metricsServer.EnableOVNDBMetrics()
+					}
+				}
+			}()
```
npinaeva: why is this function not a part of `StartOVNMetricsServer`?

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,283 @@
...
+		return nil, fmt.Errorf("failed to setup routes: %w", err)
+	}
+
+	return server, nil
+}
+
+// setupRoutes configures all HTTP routes
+func (s *MetricServer) setupRoutes() error {
+	// Metrics endpoints
+	s.mux.HandleFunc("/metrics", s.handleMetrics)
```
npinaeva: I would just inline this into the `NewMetricServer`, but also you can remove the returned error throughout the call stack, because it never returns an error?

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -608,13 +608,49 @@ func runOvnKube(ctx context.Context, runMode *ovnkubeRunMode, ovnClientset *util
...
+					// and if it is, the OVN Northd and DB Metrics will be registered.
+					err := wait.PollUntilContextTimeout(ctx, 5*time.Second, 300*time.Second, true, func(_ context.Context) (bool, error) {
+						return metrics.CheckPodRunsOnGivenNode(ovnClientset.KubeClient, []string{"ovn-db-pod=true"}, runMode.identity, true)
+					})
+					if err != nil {
+						klog.Infof("Not registering OVN Northd and DB Metrics, because OVNKube Master Pod is not running on this node(%s)",
+							runMode.identity)
+					} else {
+						klog.Info("Found OVNKube Master Pod running on this node, registering OVN Northd and DB Metrics")
+						metricsServer.EnableOVNNorthdMetrics()
```
npinaeva, line 644: we could put into the config before calling StartOVNMetricsServer?

npinaeva:
I didn't realize this approach requires writing our own handler. Digging a bit more into this, it seems like scraping metrics periodically instead of on request is considered a better choice: prometheus may call `/metrics` more often than our current scarping interval, which will cause even higher load on the cluster, and the time to scrape metrics should generally be as small as possible to avoid client timeouts. Some more details https://prometheus.io/docs/instrumenting/writing_exporters/#collectors
So I am wondering a bit more about the use case: why would we even enable metrics if we know this cluster doesn't have prometheus running? Or in which case the scrape interval may need to change? Maybe we can add some options for the specific use cases?

l8huang:
Thanks for the thoughtful comments.

> prometheus may call /metrics more often than our current scarping interval, which will cause even higher load on the cluster

With periodic collection, if Prometheus scrapes faster than the interval, both scrapes return identical data because metrics are only refreshed on the ticker. This causes unnecessary scrape load with no new information and creates misleading assumptions about data freshness.

This is actually a strong reason for request-based collection:
- Prometheus gets fresh metrics on every scrape, regardless of scrape interval.
- There is no wasted work collecting data that will never be used.
- If Prometheus scrapes slowly, we collect less. If it scrapes faster, we collect more — but only when needed.


> the time to scrape metrics should generally be as small as possible to avoid client timeouts.

Absolutely — but the expensive part is not the metric collection itself. When scale is large, the most of cpu and request time are consumed by rendering metrics in Prometheus text format and transmit the https response body to Prometheus.

Even today, the periodic system can produce a very large metrics payload, and long render times already require Prometheus operators to choose appropriate scrape interval + timeout settings.

If scrape-time ever becomes a real concern, we can improve the /metrics request handling by leverage Prometheus’s HTTP header `X-Prometheus-Scrape-Timeout-Seconds`:
- The handler can abort scraping early if the deadline is too close.
- This avoids wasting CPU after the client has already timed out.
- Prometheus will reattempt automatically.


> why would we even enable metrics if we know this cluster doesn't have prometheus running?

Metrics is critical for observability, most production should have it enabled. Some time metrics could be useful for troubleshooting as well. With the request-based metrics collection, if there is no scraper, there is zero CPU cost.

> in which case the scrape interval may need to change?

I can think of are two major factors:
- Scrape cost (payload size): 
  Large clusters generate large metrics, which take longer to render and transfer.
- Prometheus storage retention: 
  Higher scrape frequency = more samples = more TSDB usage. 

Scrape intervals depend on cluster scale and Prometheus storage limits. These tuning decisions are deployment-specific, not something the exporter should assume for the user.

go-controller/pkg/metrics/server.go
npinaeva: I don't think replacing ` promhttp.InstrumentMetricHandler(ovnRegistry,
		promhttp.HandlerFor(ovnRegistry, promhttp.HandlerOpts{}))` with our own handler is a good idea, because it misses lots of features implemented by `promhttp`
I think the "proper" way to get your own collection logic is to implement `Collector` as described in https://prometheus.io/docs/instrumenting/writing_exporters/ but it is a bit more complicated. I think you could also try to wrap the default `Gather()` method to update the metrics first.

go-controller/cmd/ovnkube/ovnkube.go
l8huang: In central mode, only a few nodes have ovn-northd and the NB/SB databases deployed as separate Pods.
Instead of waiting 300 seconds before calling StartOVNMetricsServer(), it’s better to let /metrics start working as soon as possible, so that ovn-controller and OVS metrics can be scraped by Prometheus.
This helps avoid gaps in the metrics database during rolling restarts of ovnkube-node Pods.

go-controller/cmd/ovnkube/ovnkube.go
l8huang: My view is that StartOVNMetricsServer() should focus solely on starting the /metrics endpoint.
The logic for determining whether ovn-northd and database metrics are enabled should stay outside of it.

go-controller/pkg/metrics/server.go
l8huang: make sense, inlined setupRoutes() into NewMetricServer().

l8huang: Wrapped the handler with `promhttp.InstrumentMetricHandler()` to add back the prom instrument metrics.

Before this change, promhttp.Handler() was using the default HandlerOpts, which essentially disables all optional features — including timeout handling. The current timeout logic also has several downsides, as described in its API doc:

>	// If handling a request takes longer than Timeout, it is responded to
	// with 503 ServiceUnavailable and a suitable Message. No timeout is
	// applied if Timeout is 0 or negative. Note that with the current
	// implementation, **reaching the timeout simply ends the HTTP requests** as
	// described above (and even that **only if sending of the body hasn't
	// started yet**), while **the bulk work of gathering all the metrics keeps
	// running in the background** (with the eventual result to be thrown
	// away). Until the implementation is improved, **it is recommended to
	// implement a separate timeout** in potentially slow Collectors.
	Timeout time.Duration

For large-scale clusters, it would be better to implement a custom handler that can cancel in-progress work when a timeout is reached.

go-controller/cmd/ovnkube/ovnkube.go
npinaeva: I don't think it makes using the metrics package any easier. You could add a new flag to the opts like `enableDBMetricsDynamic` and then start this goroutine as a part of `StartOVNMetricsServer`
i also think it should be run as a goroutine, otherwise it will delay ovnkube startup

go-controller/pkg/metrics/server.go
npinaeva: `NewMetricServer` also never returns an error, can you clean that up too?

npinaeva: I also meant the features of `promhttp.HandlerFor`, like compression or content type negotiation (instead of hard-coding content type), can get that back too?

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,267 @@
...
+		s.ovsDbProperties, s.opts.dbIsClustered, s.opts.dbFoundViaPath = RegisterOvnDBMetrics(s.ovnRegistry)
+	}
+	if s.opts.EnableOVNControllerMetrics {
+		klog.Infof("MetricServer registers OVN Controller metrics")
+		RegisterOvnControllerMetrics(s.ovsDBClient, s.ovnRegistry)
+	}
+	if s.opts.EnableOVNNorthdMetrics {
+		klog.Infof("MetricServer registers OVN Northd metrics")
+		RegisterOvnNorthdMetrics(s.ovnRegistry)
+	}
```
npinaeva, line 115: all registrations used to happen in their own goroutines, is it fast enough to make it sequential?

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,267 @@
...
+	// write out the registered metrics
+	w.Header().Set("Content-Type", FmtText)
+	if err := s.writeRegisteredMetrics(io.Writer(w)); err != nil {
+		klog.Errorf("Failed to write registered metrics: %v", err)
+		return
+	}
+}
+
+// Run runs the metrics server and blocks until graceful shutdown
+func (s *MetricServer) Run(stopChan <-chan struct{}) error {
```
npinaeva: I find it confusing that there is a function called `startMetricsServer` doing very similar things, but the code is not reused and it is not explained anywhere when to use which

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,267 @@
...
+
+// GetOVNRegistry returns the OVN registry for external use
+func (s *MetricServer) GetOVNRegistry() *prometheus.Registry {
+	return s.ovnRegistry
+}
+
+// SetOVNRegistry sets the OVN registry (for integration with existing code)
+func (s *MetricServer) SetOVNRegistry(registry *prometheus.Registry) {
+	s.ovnRegistry = registry
+}
```
npinaeva: looks like these 2 are only used in tests, should we get rid of them and just directly access local variables?

go-controller/pkg/metrics/ovn.go
```diff
@@ -351,55 +334,26 @@ func getPortCount(ovsDBClient libovsdbclient.Client, portType string) float64 {
...
-		}
-	}
-}
-
-func updateSBDBConnectionMetric(ovsAppctl ovsClient, retry int, retrySleep time.Duration) {
+// updateSBDBConnectionMetric updates the connection status with southbound database
+func updateSBDBConnectionMetric(ovsAppctl ovsClient) {
+	// NOTE: This metric had a retry logic, which is removed because metrics should reflect the reality.
+	// Instead, alert rules should be configured with appropriate thresholds (e.g., "for: 2m") to handle
+	// transient connection issues and only fire alerts for sustained problems.
```
npinaeva, line 341: this may dangerous, as it may cause alerts to fire. We should at least inform the community about this change

go-controller/pkg/metrics/ovs_test.go
```diff
@@ -46,12 +46,50 @@ const (
 	ovsAppctlDumpAggregateSampleOutput = "NXST_AGGREGATE reply (xid=0x4): packet_count=856244 byte_count=3464651294 flow_count=30"
 )
 
-var _ = ginkgo.Describe("OVS metrics", func() {
+var _ = ginkgo.Describe("OVS metrics", ginkgo.Ordered, func() {
```
npinaeva: can you please clarify why ginkgo.Ordered is needed (especially now as compared to before)?

go-controller/pkg/metrics/server_test.go
```diff
@@ -0,0 +1,829 @@
...
+
+var (
+	doctlShowOutput = `system@ovs-system:
```
npinaeva: do you mean dpctl?

go-controller/pkg/metrics/server_test.go
```diff
@@ -0,0 +1,829 @@
...
+			gotMetrics := []string{}
+			for _, line := range strings.Split(got, "\n") {
+				if strings.HasPrefix(line, "# TYPE ") {
+					m := strings.Split(line, " ")[2]
+					gotMetrics = append(gotMetrics, m)
+				}
+			}
+
+			if diff := cmp.Diff(gotMetrics, tc.expectedMetrics, cmpopts.SortSlices(func(x, y string) bool {
+				return x < y
```
npinaeva, line 849: do we want to check metrics values too?

go-controller/cmd/ovnkube/ovnkube.go
l8huang: > i also think it should be run as a goroutine, otherwise it will delay ovnkube startup

`metricsServer, err := metrics.StartOVNMetricsServer(opts, ovsClient, ovnClientset.KubeClient, ctx.Done(), wg)` is already in a goroutine(see line 612), it doesn't block ovnkube start.


`metricsServer.EnableOVNNorthdMetrics()` and  `metricsServer.EnableOVNDBMetrics()` are only called here for Central mode, moving then into `StartOVNMetricsServer()` wouldn’t add much value and could make the control flow less explicit.

go-controller/pkg/metrics/ovn.go
l8huang: Which document should I update for this change in the upcoming release?

go-controller/pkg/metrics/ovs_test.go
l8huang: `ginkgo.Ordered` removed.

It was add for the save/restore some metrics vars in BeforeAll() and AfterAll(), that's for avoiding conflicts with other test cases.

go-controller/pkg/metrics/server.go
l8huang: metrics registration is very fast, checked log, it took less than 100ms

go-controller/pkg/metrics/server.go
l8huang: about  the one in go-controller/pkg/metrics/metrics.go:
```
func startMetricsServer(bindAddress, certFile, keyFile string, handler http.Handler, stopChan <-chan struct{}, wg *sync.WaitGroup) {
```
that exposes the ovnkube-node's  metrics on endpoint config.Metrics.BindAddress.
Here this function exposes ovn and ovs metrics on endpoint config.Metrics.OVNMetricsBindAddress.

Because this commit is mainly for OVN/OVS metrics, I'd like to keep the scope of change small. 

BTW: any reason use two different endpoints for exposing metrics from the same process? Our downstream uses single endpoint to expose all metrics. A option can disable OVN/OVS metrics emitting if they are not wanted. I'm willing to push another PR to combine two endpoints into one if it's okay to upstream.

go-controller/pkg/metrics/server.go
l8huang: yes, they are not in use actually, removed.

go-controller/pkg/metrics/server.go
l8huang: Currently the ovs metrics are aggregated, the amount of emitted metrics(response body) is small, compression is not mandatory now. Is it okay to address the compression issue later by another PR?

l8huang: done

go-controller/pkg/metrics/server_test.go
l8huang: fixed typo

go-controller/pkg/metrics/server_test.go
l8huang: this PR has no change on how value is parsed, the test is mainly for ensure metrics are registered as expected.

l8huang:
Hi @npinaeva, thanks for your comments, PR updated, PTAL, thanks.

go-controller/pkg/metrics/server.go
l8huang: I added one more commit to this PR to add the functionalities provided by promhttp.HandlerFor(), compression is supported now.

l8huang:
@npinaeva promhttp handler's functionalities(compression, etc.) are added back in a new commit, PTAL.

go-controller/pkg/metrics/server.go
npinaeva: I am not quite sure if we need 2 different endpoints for the metrics, maybe we can ask others.
yeah let;s not touch it for now, but maybe rename `MetricServer` to `OVNMetricServer` so that the split is more obvious, also adding a comment saying that this metriics server only handles ovn and ovs metrics may be nice

go-controller/pkg/metrics/ovn.go
npinaeva: I would just update the PR descriptiont with `Release notes: <>` and then send a message to the slack channel with a warning

go-controller/cmd/ovnkube/ovnkube.go
npinaeva: I mean `err := wait.PollUntilContextTimeout(...return metrics.CheckPodRunsOnGivenNode ... )` may (and I think will) be stuck for 5 min if a given node doesn't have db pods?

npinaeva:
@l8huang thanks for the update! I dropped some more questions in the threads, let's finalize those discussions and I don't have any more comments

go-controller/cmd/ovnkube/ovnkube.go
l8huang: `err := wait.PollUntilContextTimeout(...return metrics.CheckPodRunsOnGivenNode ... )` doesn't block the ovnkube start procedure and metrics server,  it won't cause any delay.

go-controller/pkg/metrics/ovn.go
l8huang: PR description updated.

go-controller/pkg/metrics/server.go
l8huang: My plan is to consolidate the two endpoints into a single endpoint if there are no concerns. If that's not feasible,  I will submit another PR to make both endpoints use `MetricServer` and add a feature flag to support a unified metrics endpoint, so our upstream and downstream can align.

Let keep `MetricServer` as it is for now, I will start to work on above after this PR is merged.

go-controller/pkg/metrics/server.go
```diff
@@ -0,0 +1,243 @@
...
+	if s.opts.EnableOVNControllerMetrics {
+		s.updateOvnControllerMetrics()
+	}
+	if s.opts.EnableOVNNorthdMetrics {
+		s.updateOvnNorthdMetrics()
+	}
+}
+
+// Run runs the metrics server and blocks until graceful shutdown
+func (s *MetricServer) Run(stopChan <-chan struct{}) error {
```
npinaeva: this function never returns error, then StartOVNMetricsServer will also never return error, then ovnkube.go also needs a cleanup

go-controller/cmd/ovnkube/ovnkube.go
npinaeva: oh I forgot that the whole block is in a goroutine, but the only actually blocking part is `PollUntilContextTimeout`, so do you mind moving the goroutine under `if !config.OVNKubernetesFeature.EnableInterconnect {`, then it won't create a goroutine in an IC mode at all, and we can just remove tho whole block when we get rid on the central mode upstream

go-controller/pkg/metrics/server.go
l8huang: Good point, the return error is removed.
Also added a `OnFatalError` in `MetricServerOptions`, so ovs-exporter can fast fail if it can't bind to address.

go-controller/cmd/ovnkube/ovnkube.go
l8huang: make sense, code updated.

l8huang:
@npinaeva Thanks for approving this PR. Is there anything I need to do to get it merged?

npinaeva:
> @npinaeva Thanks for approving this PR. Is there anything I need to do to get it merged?

I was just waiting for the CI to pass, merged now!

PR #5639, opened 2025-10-09
udn: Skip join subnet assignment for secondary Layer2 networks

qinqon:
## 📑 Description
Secondary Layer2 networks don't require join subnets, but were incorrectly being assigned the default UDN join subnet (100.65.0.0/16). This caused failures when the cluster join subnet matched the UDN default, as overlap validation would incorrectly flag a conflict.

Fix by only assigning join subnets to primary networks in the Layer2 topology case. Add regression test verifying secondary Layer2 networks can be created successfully even when cluster join subnet matches UDN default join subnet.

N/A

## ✅ Checks

- [x] My code requires tests

## How to verify it
It already include unit tests 

## Summary by CodeRabbit

* **Bug Fixes**
  * Secondary Layer2 user-defined networks no longer receive a join subnet, preventing subnet overlap errors and improving network stability when multiple networks share defaults.

* **Tests**
  * Added a regression test to verify secondary Layer2 networks omit join subnets when cluster join subnets match defaults.
  * Test setup now initializes IPv4/IPv6 cluster join subnets to avoid conflicts across UDN scenarios and ensure reliable test isolation.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
@@ -174,12 +174,14 @@ func renderCNINetworkConfig(networkName, nadName string, spec SpecGetter) (map[s
 			netConfSpec.DefaultGatewayIPs = ipString(cfg.DefaultGatewayIPs)
 		}
 		netConfSpec.JoinSubnet = cidrString(renderJoinSubnets(cfg.Role, cfg.JoinSubnets))
-		// now generate transit subnet for layer2 topology
 		if cfg.Role == userdefinednetworkv1.NetworkRolePrimary {
+			// now generate transit subnet for layer2 topology
 			err := util.SetTransitSubnets(netConfSpec)
 			if err != nil {
 				return nil, err
 			}
+			// Only set join subnet for primary networks; secondary Layer2 networks don't need join subnets
+			netConfSpec.JoinSubnet = cidrString(renderJoinSubnets(cfg.Role, cfg.JoinSubnets))
 		}
 	case userdefinednetworkv1.NetworkTopologyLocalnet:
 		cfg := spec.GetLocalnet()
--- a/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
@@ -25,6 +25,9 @@ var _ = Describe("NetAttachDefTemplate", func() {
 	BeforeEach(func() {
 		config.IPv4Mode = true
 		config.IPv6Mode = true
+		// Set cluster join subnets to avoid conflicts with UDN tests
+		config.Gateway.V4JoinSubnet = "100.64.0.0/16"
+		config.Gateway.V6JoinSubnet = "fd98::/64"
 	})
 
 	DescribeTable("should fail to render NAD spec given",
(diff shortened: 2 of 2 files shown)


go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
```diff
@@ -291,6 +294,38 @@ var _ = Describe("NetAttachDefTemplate", func() {
...
+					Subnets: udnv1.DualStackCIDRs{"203.203.0.0/16"},
+					MTU:     1500,
+				},
+			},
+		}
+
+		// This should NOT fail - secondary Layer2 networks don't get join subnets
+		nad, err := RenderNetAttachDefManifest(testUdn, "test-ns")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(nad).NotTo(BeNil())
```
maiqueb: can we match against the expected NAD instead ?

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
```diff
@@ -291,6 +294,38 @@ var _ = Describe("NetAttachDefTemplate", func() {
...
+
+		// Verify the NAD config does not contain a joinSubnet field
+		Expect(nad.Spec.Config).NotTo(ContainSubstring("joinSubnet"))
```
maiqueb: if we match the expected NAD, we don't need this.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
```diff
@@ -291,6 +294,38 @@ var _ = Describe("NetAttachDefTemplate", func() {
...
+				Layer2: &udnv1.Layer2Config{
+					Role:    udnv1.NetworkRoleSecondary,
+					Subnets: udnv1.DualStackCIDRs{"203.203.0.0/16"},
+					MTU:     1500,
+				},
+			},
+		}
+
+		// This should NOT fail - secondary Layer2 networks don't get join subnets
+		nad, err := RenderNetAttachDefManifest(testUdn, "test-ns")
```
maiqueb: if you match against the NAD's contents, you can have everything as a one liner.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
qinqon: There is no expected nad we want to just check that the generated one is not configuring join subnet tanymore.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
qinqon: We don't need a full expected nad just a check on one field that should not be there

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
qinqon: What do you mean ?

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
maiqueb: I mean something like 

```suggestion
		Expect(RenderNetAttachDefManifest(testUdn, "test-ns")).To(WithTransform(
		    func (nad NAD) string {
		        return nad.Spec.Config
		    },
		    Equal(<expectedNAD>),
		))
```

maiqueb (approved):
my comments are opinionated - IMHO it doesn't mean the PR should be held on those.

Let's get the ball moving forward.

maiqueb:
Hm, you need to rebase, plus fix a failing unit test:
> 
Summarizing 1 Failure:
  [FAIL] NetAttachDefTemplate [It] should succeed for secondary Layer2 network when cluster join subnet matches UDN default
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go:322

/lgtm cancel

maiqueb (changes requested):
UTs are failing; must be rebased

qinqon:
@maiqueb rebased, let's see the unittest now.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
```diff
@@ -174,12 +174,14 @@ func renderCNINetworkConfig(networkName, nadName string, spec SpecGetter) (map[s
 			netConfSpec.DefaultGatewayIPs = ipString(cfg.DefaultGatewayIPs)
 		}
 		netConfSpec.JoinSubnet = cidrString(renderJoinSubnets(cfg.Role, cfg.JoinSubnets))
```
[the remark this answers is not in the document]
npinaeva: @qinqon did you mean to remove this?

npinaeva:
The change makes sense, but I am confused about the actual conflict use case. We set both default join subnet for the default network (100.64.0.0/16) and UDN (100.65.0.0/16) how do we end up with a conflict?

PR #4305, opened 2024-04-23
check disk usage before and after 'Run Tests' step

jluhrsen:
ISSUE: https://github.com/ovn-org/ovn-kubernetes/issues/4287
JIRA: https://issues.redhat.com/browse/SDN-4786

**- What this PR does and why is it needed**

**- Special notes for reviewers**

**- How to verify it**

**- Description for the changelog**

PR #5640, opened 2025-10-09
ci: Remove Runner Diagnostics step

booxter:
- **Revert "check disk usage before and after test steps (#4305)"**

## 📑 Description

It's important to have quick feedback from CI.

This diagnostics step may be valuable when debugging disk usage issues, but it
takes 10 mins+ to run `du` over the entire filesystem, which comprises 1/4 of
total test time for some jobs, for example see:

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18327721132/job/52197975313?pr=5635

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Chores**
  * Streamlined CI by removing the runner diagnostics workflow and related steps across test jobs.
  * Reduces log noise and shortens pipeline execution time, improving reliability and throughput of test and release cycles.
  * No changes to application behavior or public APIs; existing tests and release steps remain unchanged.
  * No action required from users.


booxter:
@jluhrsen let me know what you think of it. Thanks!

jluhrsen:
> @jluhrsen let me know what you think of it. Thanks!

The issue is that the disk usage issues we hit end up being flakes and rare, so we want the diags to show up when those problems show up.

looks like the diag step sometimes takes 10+ minutes and other times finishes much more quickly. 30s, 1m, etc. I'm not sure how to explain that.

I would vote to keep the diags running, but if the team agrees to remove, then I won't fight it :)

booxter:
As far as I saw, 10mins is the first time it runs. Later runs are much shorter. I suspect the file system state is cached in memory.

A reasonable compromise could be dumping disk info on failure only. Though I wonder - what's the value of `du` results that `df` wouldn't cover? @jluhrsen

jluhrsen:
> As far as I saw, 10mins is the first time it runs. Later runs are much shorter. I suspect the file system state is cached in memory.
> 
> A reasonable compromise could be dumping disk info on failure only. Though I wonder - what's the value of `du` results that `df` wouldn't cover? @jluhrsen

the problem is when there is a failure, we will want to know what the disk usage looked like at the different points in the job to know which step introduced whatever new issue came about.

booxter:
@jluhrsen let's see... What if:

- we always do `docker system df` and `df` (these are immediate)
- we dump `du -hs` on failure only?

So you'd see general trend but also details on final failure state.

On a relevant note, I wonder if there's a reason why we implement action to free disk space in-house instead of using a common action used elsewhere like https://github.com/marketplace/actions/free-disk-space-ubuntu ?

jluhrsen:
> @jluhrsen let's see... What if:
> 
> * we always do `docker system df` and `df` (these are immediate)
> * we dump `du -hs` on failure only?

but if we see the output and there is some large file(s) in `du -hs` we may not know if there were already there or something the most recent steps has created.

> 
> So you'd see general trend but also details on final failure state.
> 
> On a relevant note, I wonder if there's a reason why we implement action to free disk space in-house instead of using a common action used elsewhere like https://github.com/marketplace/actions/free-disk-space-ubuntu ?

great question. I would look in to that if I had time ( i don't :( ) but I think probably the current state of disk space
management came a long time ago and maybe these common public actions.

trozet (approved):

PR #5642, opened 2025-10-09
add nft rule for service in the dpu-host mode

cathy-zhou:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

- New Features
  - Improved support for network segmentation in gateway rules, applying per‑network packet marks and ensuring rules only apply on the primary network when segmentation is enabled.

- Bug Fixes
  - Enhanced robustness when the active network cannot be determined, gracefully skipping rule updates instead of failing.
  - More reliable add/remove of service rules for NodePort and ExternalIPs under segmented networks.

- Refactor
  - Streamlined NFT rule generation to reduce coupling and improve maintainability, aligning rule application across watchers and code paths.

### DIFF
--- a/go-controller/pkg/node/gateway_nftables.go
+++ b/go-controller/pkg/node/gateway_nftables.go
@@ -14,7 +14,6 @@ import (
 	"sigs.k8s.io/knftables"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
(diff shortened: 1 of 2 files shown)


cathy-zhou:
@kyrtapz @trozet can you please take a look. The problem is that when addServiceRules() is called in dpu-host mode, the `npw` argument is nil, so we are not able to get the bridgeConfiguration of the primary udn network and are not able to update the related nft rules.

PR #5645, opened 2025-10-10
OKEP-5631: Consolidate ovn-k8s full mode, dpu-mode and dpu-host mode into a single mode of operation

### LINKED ISSUE #5631: Consolidating OVN-K8s DPU mode into OVN-K8s full

Issue #5523 shows that there is a use-case for running OVN-K8s in full mode on the host, alongside DPU-Host mode OVN-K8s.
Similarly, there is a use-case for running OVN-K8s in full mode on the DPU (to act as the CNI for any workloads scheduled on the DPU) alongside OVN-K8s running in DPU Mode.
Not only that, a comment in #5523 suggests a potential future where we have multi DPUs to consider.

If we continue along this trajectory we will end up with  2(N+1) instances of OVN-K8s running, where N is the number of DPUs.
At the simplest case, this is 2(1+1) = 4 OVN-K8s instances.

I propose that we consolidate the OVN-K8s functionality into a single mode of operation (OVN-K8s-full+dpu).
This would reduce the total number of OVN-Kubernetes instances required to 2.
It would also provide a good foundation to move to a shared OVN model, where we don't require 2 OVN deployments.

### Why is this needed?

If we continue to add OVN-K8s instances we are going to have:

1. Challenges performing upgrades: sequencing, handling mixed versions etc...
2. Wasted resources on both the host and DPU
3. A testing nightmare - we've had regressions in DPU-Host mode recently since there is no CI for it.

dave-tucker:
Fixes: #5631

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests


dave-tucker:
FYI This is not ready for review, hence marking as draft.
I've still got plenty of missing details to add... only opening the PR here to get some early comments from collaborators as I firm up some of those missing details.

PR #5646, opened 2025-10-11
Fix management port cleanup

trozet:
Changes:
1.  Ensure that if OVS/OVSDB is in a corrupted state where it leaves the mpx port hanging in the kernel, it still gets cleaned up.
2. Ensure that when gateway Cleanup is called, it attempts to clean everything, and not return on first error.
3. On ovnkube start, make sure we clean up any stale management ports

@girishmg @tssurya FYI

## Summary by CodeRabbit

- New Features
  - Automatic cleanup of stale management ports for user-defined networks.
  - Public IP rule-manager interface to support interchangeable implementations and fakes.

- Bug Fixes
  - Ensures stale network interfaces (MPX/VRF) are removed while valid interfaces remain.
  - Gateway deletion is idempotent so cleanup proceeds despite intermediate errors.

- Refactor
  - Error handling now aggregates multiple failures into a combined error.

- Tests
  - Added tests for stale-interface cleanup, idempotent gateway deletion, and a fake rule-manager error scenario.

### DIFF
--- a/go-controller/pkg/controllermanager/node_controller_manager_test.go
+++ b/go-controller/pkg/controllermanager/node_controller_manager_test.go
@@ -1,12 +1,14 @@
 package controllermanager
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containernetworking/plugins/pkg/testutils"
 	"github.com/stretchr/testify/mock"
+	"github.com/vishvananda/netlink"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


go-controller/pkg/node/gateway_udn.go
```diff
@@ -468,21 +478,79 @@ func (udng *UserDefinedNetworkGateway) addUDNManagementPortIPs(mpLink netlink.Li
...
+	}
+	return nil
+}
+
+// CleanupManagementPorts removes all stale UDN ovn-k8s-mpx ports
+func CleanupManagementPorts(ovsClient client.Client, validNetworks ...util.NetInfo) error {
+	// build valid set of mpx interfaces
+	validMpx := sets.New[string]()
+	for _, netInfo := range validNetworks {
+		validMpx.Insert(util.GetNetworkScopedK8sMgmtHostIntfName(uint(netInfo.GetNetworkID())))
```
cathy-zhou: should only add to valid mpx interfaces if the network is primary network?

trozet: good point. I'll fix it

cathy-zhou: lgtm

go-controller/pkg/node/gateway_udn_test.go
```diff
@@ -910,9 +911,9 @@ var _ = Describe("UserDefinedNetworkGateway", func() {
 			}
 			Expect(udnFlows).To(Equal(0))
 			Expect(udnGateway.openflowManager.defaultBridge.GetNetConfigLen()).To(Equal(1)) // only default network
-
-			By("Deleting the gateway network")
-			Expect(udnGateway.DelNetwork()).To(Succeed())
+			By("Deleting the gateway network with injected error")
+			Expect(udnGateway.DelNetwork()).ToNot(Succeed())
```
npinaeva: do you want to check error message?

go-controller/pkg/node/gateway_udn.go
```diff
@@ -479,17 +484,16 @@ func (udng *UserDefinedNetworkGateway) deleteUDNManagementPort() error {
 		"--", "--if-exists", "del-port", "br-int", interfaceName,
 	)
 	if err != nil {
-		return fmt.Errorf("failed to delete port from br-int for network %s, stdout: %q, stderr: %q, error: %v",
-			udng.GetNetworkName(), stdout, stderr, err)
+		return fmt.Errorf("failed to delete port from br-int - stdout: %q, stderr: %q, error: %w",
```
npinaeva: maybe add port name?

go-controller/pkg/controllermanager/node_controller_manager_test.go
```diff
@@ -279,5 +279,76 @@ var _ = Describe("Healthcheck tests", func() {
...
+
+				mutableNetInfo.SetNetworkID(int(netID))
+
+				By("Cleaning up stale networks")
+				err = ncm.CleanupStaleNetworks(mutableNetInfo)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Stale mpx interface should have been removed")
+				_, err = util.GetNetLinkOps().LinkByName(staleMgtPort)
+				Expect(err).To(HaveOccurred())
```
npinaeva: also check error message?

trozet:
just rebased, will push fixes in just a minute

PR #5649, opened 2025-10-13
Handle failed CloudPrivateIPConfig assignments

kyrtapz:
When a CPIC is created and the cloud controller externally sets its status to Failed, ovn-kubernetes would not clean up or retry the assignment.

Reproducer:
1. CPIC assignment fails
2. The EgressIP status never reflects it as it ignores non-sucessfull assignments
3. Later reconciliation attempts trying to create the same CPIC again
4. Creation failures due to "already exists" errors

To fix the issue remove failed CPICs during CPIC add reconciliation and actually requeue EIPs that are referenced by failed CPICs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of cloud IP assignment: stale cloud IP entries are now cleaned up when their previously assigned nodes no longer exist, triggering retries and clearer event logging.
  * Ensured consistent reconciliation by refining resync and pending-operation cleanup so only affected IP records are retried and system state stays in sync.

### DIFF
--- a/go-controller/pkg/clustermanager/egressip_controller.go
+++ b/go-controller/pkg/clustermanager/egressip_controller.go
@@ -18,6 +18,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/record"
@@ -36,7 +37,6 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 	utilerrors "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


go-controller/pkg/clustermanager/egressip_controller.go
```diff
@@ -211,6 +212,31 @@ func (eIPC *egressIPClusterController) executeCloudPrivateIPConfigOps(egressIPNa
 				if cloudPrivateIPConfig.GetDeletionTimestamp() != nil && !cloudPrivateIPConfig.GetDeletionTimestamp().IsZero() {
 					return fmt.Errorf("cloud update request failed, CloudPrivateIPConfig: %s is being deleted", cloudPrivateIPConfigName)
 				}
+
+				// Handle a scenario in which the object exists in a failed state by removing it if the node it was assigned to no longer exists
+				assignedCondition := meta.FindStatusCondition(cloudPrivateIPConfig.Status.Conditions, string(ocpcloudnetworkapi.Assigned))
+				if cloudPrivateIPConfig.Status.Node != "" && assignedCondition != nil && assignedCondition.Status == metav1.ConditionFalse {
```
arghosh93, line 218: Should we also check what status is being reported to egressIP object? If a node is deleted then that gets reflected into egressIP status also.

kyrtapz: I am not sure I fully get the question but the EgressIP status is only updated when the cloudprivateipconfig is successfully assigned:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/clustermanager/egressip_controller.go#L1501-L1504

arghosh93: Got it. Thanks for the explanation.

PR #5654, opened 2025-10-14
Fix GARP sending 0.0.0.0 due to incorrect IPv4 byte extraction

kyrtapz:
`net.ParseIP()` returns 16-byte IPv4-mapped IPv6 format where IPv4 bytes are at the END, not beginning.
`[4]byte(garp.IP)` in `BroadcastGARP` took wrong bytes for 16 bytes addresses(EgressIP).
For example:
```
12:12:42.518073 00:2c:a5:d2:c4:21 > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 60: Request who-has 0.0.0.0 tell 0.0.0.0, length 46
12:12:42.518127 00:2c:a5:d2:c4:21 > ff:ff:ff:ff:ff:ff, ethertype ARP (0x0806), length 60: Reply 0.0.0.0 is-at 00:2c:a5:d2:c4:21, length 46
```

Fixed by calling To4() before, forcing validated creation via NewGARP(). Interface prevents bypassing extracting the correct IPv4 address.

The second commit adds Unsolicited Neighbor Advertisement support for IPv6 EgressIPs.

Added an E2E that checks that the neighbor learned the EIP MAC from GARP.

## Summary by CodeRabbit

* **New Features**
  * Added IPv6 neighbor-advertisement support for unsolicited address announcements on secondary host networks.

* **Tests**
  * Added e2e tests verifying address advertisement (ARP/NA) behavior for EgressIP on secondary hosts, including neighbor table and GARP verification.
  * Fixed neighbor JSON parsing tags used by tests.

* **Refactor**
  * Consolidated address-advertisement handling behind validated constructors/interfaces with improved error handling and clearer logging.

### DIFF
--- a/go-controller/pkg/node/linkmanager/link_network_manager.go
+++ b/go-controller/pkg/node/linkmanager/link_network_manager.go
@@ -11,6 +11,7 @@ import (
 	utilnet "k8s.io/utils/net"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/ndp"
 )
 
 // Gather all suitable interface address + network mask and offer this as a service.
@@ -203,6 +204,18 @@ func (c *Controller) syncLink(link netlink.Link) error {
 			}
 			if err = util.BroadcastGARP(linkName, garp); err != nil {
 				klog.Errorf("Link manager: failed to send GARP for IP %s over interface %s: %v", addressWanted.IP.String(),
+					linkName, err)
+			}
+		} else if addressWanted.IP.To16() != nil {
+			// For IPv6, send an unsolicited neighbor advertisement to update neighbor caches, in case this IP was
+			// previously active on another node
+			na, err := ndp.NewNeighborAdvertisement(addressWanted.IP, nil)
+			if err != nil {
+				klog.Errorf("Link manager: failed to create NeighborAdvertisement for IP %s: %v", addressWanted.IP.String(), err)
+				continue
+			}
+			if err = ndp.SendUnsolicitedNeighborAdvertisement(linkName, na); err != nil {
+				klog.Errorf("Link manager: failed to send an unsolicited neighbor advertisement for IP %s over interface %s: %v", addressWanted.IP.String(),
 					linkName, err)
 			}
 		}
--- a/go-controller/pkg/util/ndp/na.go
+++ b/go-controller/pkg/util/ndp/na.go
@@ -1,0 +1,104 @@
+package ndp
+
+import (
+	"fmt"
+	"net"
+	"net/netip"
+
+	"github.com/mdlayher/ndp"
+	"golang.org/x/net/ipv6"
+
+	"k8s.io/klog/v2"
+)
+
+// NeighborAdvertisement represents a Neighbor Advertisement for an IPv6 address.
+type NeighborAdvertisement interface {
+	// IP returns the IPv6 address
+	IP() net.IP
+	// MAC returns the MAC address to advertise (nil means use interface MAC)
+	MAC() *net.HardwareAddr
+}
+
+type neighborAdvertisement struct {
+	ip  net.IP
+	mac *net.HardwareAddr
+}
+
+// NewNeighborAdvertisement creates a new Unsolicited Neighbor Advertisement with validation that the IP is IPv6.
+func NewNeighborAdvertisement(ip net.IP, mac *net.HardwareAddr) (NeighborAdvertisement, error) {
+	if ip.To4() != nil {
+		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got IPv4 %s", ip.String())
+	}
+	if ip.To16() == nil {
+		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got %s", ip.String())
+	}
+	if ip.IsMulticast() || ip.IsUnspecified() {
+		return nil, fmt.Errorf("invalid IPv6 NA target address: %s", ip.String())
+	}
+
+	return &neighborAdvertisement{
+		ip:  ip.To16(),
+		mac: mac,
+	}, nil
+}
+
+// IP returns the IPv6 address
+func (u *neighborAdvertisement) IP() net.IP {
+	return u.ip
+}
+
+// MAC returns the MAC address to advertise
+func (u *neighborAdvertisement) MAC() *net.HardwareAddr {
+	return u.mac
+}
+
+// SendUnsolicitedNeighborAdvertisement sends an unsolicited neighbor advertisement for the given IPv6 address.
+// If the mac address is not provided it will use the one from the interface.
+// The advertisement is sent to the all-nodes multicast address (ff02::1).
+// https://datatracker.ietf.org/doc/html/rfc4861#section-4.4
+func SendUnsolicitedNeighborAdvertisement(interfaceName string, na NeighborAdvertisement) error {
+	iface, err := net.InterfaceByName(interfaceName)
+	if err != nil {
+		return fmt.Errorf("failed finding interface %s: %v", interfaceName, err)
+	}
+
+	targetIP := na.IP()
+	mac := na.MAC()
+	if mac == nil {
+		mac = &iface.HardwareAddr
+	}
+
+	targetAddr, ok := netip.AddrFromSlice(targetIP)
+	if !ok {
+		return fmt.Errorf("failed to convert IP %s to netip.Addr", targetIP.String())
+	}
+
+	// Use unspecified address to handle cases where the advertised IP is not assigned to the interface.
+	c, _, err := ndp.Listen(iface, ndp.Unspecified)
+	if err != nil {
+		return fmt.Errorf("failed to create NDP connection on %s: %w", interfaceName, err)
+	}
+	defer c.Close()
+
+	// Unsolicited neighbor advertisement from a host, should override any existing cache entries
+	una := &ndp.NeighborAdvertisement{
+		Router:        false,
+		Solicited:     false,
+		Override:      true,
+		TargetAddress: targetAddr,
+		Options: []ndp.Option{
+			&ndp.LinkLayerAddress{
+				Direction: ndp.Target,
+				Addr:      *mac,
+			},
+		},
+	}
+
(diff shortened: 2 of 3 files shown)


go-controller/pkg/util/ndp/na.go
```diff
@@ -0,0 +1,100 @@
...
+
+	allNodesMulticast := netip.MustParseAddr("ff02::1")
+
+	// Use the target address as the source for the NA
+	// This is required for gratuitous NAs to properly update neighbor caches
+	c, _, err := ndp.Listen(iface, ndp.Addr(na.IP().String()))
+	if err != nil {
+		return fmt.Errorf("failed to create NDP connection on %s: %w", interfaceName, err)
+	}
+	defer c.Close()
```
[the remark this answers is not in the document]
kyrtapz: RFC states:
```
Source Address: An address assigned to the interface from which the advertisement is sent.
```
We only send the NA AFTER we've assigned the IP to the interface so I don't see how what I have is wrong.

go-controller/pkg/util/arp.go
```diff
@@ -6,29 +6,69 @@ import (
...
+func NewGARP(ip net.IP, mac *net.HardwareAddr) (GARP, error) {
+	ip4 := ip.To4()
+	if ip4 == nil {
+		return nil, fmt.Errorf("GARP only supports IPv4 addresses, got %s (len=%d bytes)", ip.String(), len(ip))
+	}
+	return &garp{
+		ip:  [4]byte(ip4),
+		mac: mac,
+	}, nil
+}
```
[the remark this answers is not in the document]
kyrtapz: This compiles and runs well...

test/e2e/egressip.go
```diff
@@ -2957,6 +2957,115 @@ spec:
...
+					neighborMAC = n.Lladdr
+					framework.Logf("Neighbor entry found for %s -> MAC %s", egressIPSecondaryHost, neighborMAC)
+					return true
+				}
+			}
+			return false
+		}, 30*time.Second, 2*time.Second).Should(gomega.BeTrue(),
+			"Neighbor entry should appear via GARP")
+		gomega.Expect(neighborMAC).Should(gomega.Equal(inf.MAC), "neighbor entry should have the correct MAC address")
+	})
```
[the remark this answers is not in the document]
kyrtapz: `IpNeighbor` has lladdr tag already.

martinkennelly:
DM me when its ready for review pat.. thank you <3

kyrtapz:
> DM me when its ready for review pat.. thank you <3

Should be ready to review! I'll sort out any e2e failures if they appear but last run didn't have any that would seem related to the changes introduced.

test/e2e/egressip.go
```diff
@@ -2957,6 +2957,115 @@ spec:
...
 	})
 
+	ginkgo.It("[secondary-host-eip] should send address advertisements for EgressIP", func() {
+		if isUserDefinedNetwork(netConfigParams) {
+			ginkgo.Skip("Unsupported for UDNs")
+		}
+
+		egressIPSecondaryHost := "10.10.10.220"
+		isV6Node := utilnet.IsIPv6(net.ParseIP(egress1Node.nodeIP))
+		if isV6Node {
```
cgoncalves, line 2967: The test does not cover dual-stack IP. This is a limitation in the test suite itself, correct?

go-controller/pkg/util/ndp/na.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got %s", ip.String())
+	}
+	if ip.IsMulticast() || ip.IsUnspecified() {
+		return nil, fmt.Errorf("invalid IPv6 NA target address: %s", ip.String())
+	}
+
+	return &neighborAdvertisement{
+		ip:  ip.To16(),
+		mac: mac,
+	}, nil
```
qinqon, line 42:
```suggestion
	if ip.To4() != nil {
		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got IPv4 %s", ip.String())
	}
	if ip.IsMulticast() || ip.IsUnspecified() {
		return nil, fmt.Errorf("invalid IPv6 NA target address: %s", ip.String())
	}
	ip6 := ip.To16
	if ip6 == nil {
		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got %s", ip.String())
	}
	return &neighborAdvertisement{
		ip: ip6,
		mac: mac,
	}, nil
```

go-controller/pkg/util/ndp/na.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// MAC returns the MAC address to advertise
+func (u *neighborAdvertisement) MAC() *net.HardwareAddr {
+	return u.mac
+}
+
+// SendUnsolicitedNeighborAdvertisement sends an unsolicited neighbor advertisement for the given IPv6 address.
+// If the mac address is not provided it will use the one from the interface.
+// https://datatracker.ietf.org/doc/html/rfc4861#section-4.4
```
qinqon, line 58: Comment that we are using a multicast all nodes as destination.

go-controller/pkg/util/ndp/na.go
```diff
@@ -0,0 +1,102 @@
...
+		TargetAddress: targetAddr,
+		Options: []ndp.Option{
+			&ndp.LinkLayerAddress{
+				Direction: ndp.Target,
+				Addr:      *mac,
+			},
+		},
+	}
+
+	// rfc4861 - hop Limit 255 as per RFC, send to all-nodes multicast address
```
qinqon:
```suggestion
	// rfc4861 - hop Limit 255 for unsolicited as per RFC, send to all-nodes multicast address
```

go-controller/pkg/node/linkmanager/link_network_manager.go
```diff
@@ -196,8 +197,25 @@ func (c *Controller) syncLink(link netlink.Link) error {
...
+			garp, err := util.NewGARP(addressWanted.IP, nil)
+			if err != nil {
+				klog.Errorf("Link manager: failed to create GARP for IP %s: %v", addressWanted.IP.String(), err)
+				continue
+			}
+			if err = util.BroadcastGARP(linkName, garp); err != nil {
+				klog.Errorf("Link manager: failed to send GARP for IP %s over interface %s: %v", addressWanted.IP.String(),
+					linkName, err)
+			}
+		} else if addressWanted.IP.To16() != nil {
```
qinqon, line 209: Isn't else enough ?

```suggestion
		} else {
```

qinqon:
Hey @kyrtapz I facing the following when sending a NA for a pod
```
failed to reconcile neighbours at kubevirt cluster default network local zone: failed sending Unsolicited NA: failed to create NDP connection on ovn-k8s-mp0: ndp: address 
fd00:10:244:4::7 not found on interface ovn-k8s-mp0
```

Looks like we should no tspecify the ip address as listening

go-controller/pkg/util/ndp/na.go
```diff
@@ -0,0 +1,102 @@
...
+	if mac == nil {
+		mac = &iface.HardwareAddr
+	}
+
+	targetAddr, ok := netip.AddrFromSlice(targetIP)
+	if !ok {
+		return fmt.Errorf("failed to convert IP %s to netip.Addr", targetIP.String())
+	}
+
+	c, _, err := ndp.Listen(iface, ndp.Addr(na.IP().String()))
```
qinqon: This will fail if the address is not present at the interface, we should use the "unspecified" value
```suggestion
	c, _, err := ndp.Listen(iface, ndp.Unspecified)
```

go-controller/pkg/util/ndp/na.go
kyrtapz: I don't see a need, To16 on a v6 address does not do much.

go-controller/pkg/node/linkmanager/link_network_manager.go
kyrtapz: I know it is a bit pedantic but in theory the IP can be nil, I would keep it as it doesn't hurt.

go-controller/pkg/util/ndp/na.go
kyrtapz: If there is a need for us to send NAs with IPs that are not assigned then that suggestion makes sense.
Done!

qinqon: > If there is a need for us to send NAs with IPs that are not assigned then that suggestion makes sense. Done!

I introduce that use case at the PR that depends on this.

qinqon:
LGTM

test/e2e/egressip.go
```diff
@@ -2957,6 +2957,105 @@ spec:
...
+				if n.Lladdr != "" {
+					neighborMAC = n.Lladdr
+					framework.Logf("Neighbor entry found for %s -> MAC %s", egressIPSecondaryHost, neighborMAC)
+					return true
+				}
+			}
+			return false
+		}, 30*time.Second, 2*time.Second).Should(gomega.BeTrue(),
+			"Neighbor entry should appear")
+		gomega.Expect(neighborMAC).Should(gomega.Equal(inf.MAC), "neighbor entry should have the correct MAC address")
```
trozet, line 3066: nice test

go-controller/pkg/util/ndp/na.go
```diff
@@ -0,0 +1,104 @@
...
+
+// NewNeighborAdvertisement creates a new Unsolicited Neighbor Advertisement with validation that the IP is IPv6.
+func NewNeighborAdvertisement(ip net.IP, mac *net.HardwareAddr) (NeighborAdvertisement, error) {
+	if ip.To4() != nil {
+		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got IPv4 %s", ip.String())
+	}
+	if ip.To16() == nil {
+		return nil, fmt.Errorf("only IPv6 addresses can be used for NeighborAdvertisement, got %s", ip.String())
+	}
+	if ip.IsMulticast() || ip.IsUnspecified() {
```
trozet, line 35: is this check for multicast correct? i thought neighbor advertisement uses multicast

kyrtapz: It uses multicast as the destination, not the address we send the NA for.

PR #5655, opened 2025-10-14
chore: Remove SetTaintOnNode

dave-tucker:
## 📑 Description

Node-taints for too-small MTU were removed in #3004. Taints for NoSchedule were removed in #2459.
In general, it's not the CNI plugins responsibility to set node taints. This is for the kubelet/container runtime to figure out.

Therefore, it's safe to remove this unused code since it won't be required in future.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Refactor**
  * Removed node taint management operations from the public API, simplifying the interface and reducing surface area.
* **Tests**
  * Removed taint-related test suites and scenarios to align with the updated capabilities.
* **Chores**
  * Updated mocks to reflect the removal of taint operations and ensure consistency across tests and integrations.

These changes streamline the codebase and eliminate unsupported taint operations, with no impact on remaining functionality.

### DIFF
--- a/go-controller/pkg/kube/kube.go
+++ b/go-controller/pkg/kube/kube.go
@@ -55,8 +55,6 @@ type Interface interface {
 	SetAnnotationsOnService(namespace, serviceName string, annotations map[string]interface{}) error
 	SetAnnotationsOnNode(nodeName string, annotations map[string]interface{}) error
 	SetAnnotationsOnNamespace(namespaceName string, annotations map[string]interface{}) error
-	SetTaintOnNode(nodeName string, taint *corev1.Taint) error
-	RemoveTaintFromNode(nodeName string, taint *corev1.Taint) error
 	SetLabelsOnNode(nodeName string, labels map[string]interface{}) error
 	PatchNode(old, new *corev1.Node) error
 	UpdateNodeStatus(node *corev1.Node) error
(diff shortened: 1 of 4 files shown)


dave-tucker:
> then cluster-manager will need to taint the node.

@trozet please don't taint the node! that would stop the scheduling of host-networked pods - which is why #2459 was reverted.... (see linked bz on that PR for discussion) additionally, what if ovn-k8s isn't the only CNI running, should we block scheduling of pods on the node?

The better way to handle this would be to either:
a) Fail fast in the CNI - i.e don't wait forever for annotations to appear if we know the DPU isn't ready, return from ADD immediately to propagate the error.
b) Implement the CNI STATUS command, where if we know we can't respond to ADD, we return a negative error code and let that propagate up the stack (potentially via multus, up to the CRI). This leaves the decision to things higher up the stack as to whether the node should be marked notready.

trozet:
> 't taint the node! that would stop the scheduling of host-networked pods - which is why #2459 was reverted.... (see linked bz on that PR for discussion) additionally, what if ovn-k8s isn't the only CNI running, should we block scheduling of pods on the node?

@dave-tucker  ah! you jogged my memory! I remembered a long time ago we had some problem with tainting the node and we reverted it but I couldn't remember what it was. Thanks! We need to discuss this more in slack

trozet (approved):

PR #5658, opened 2025-10-14
Reintroduce completed pod check in shouldReleaseDeletedPod

jcaamano:
I accidentally removed the check in recent PR [1] which could have performance consequences as checking against other pods has a cost. Reintroduce the check with a hopefully useful comment to prevent it form happening again.

[1] https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5626

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod deletion handling improved: IPs for pods deleted before completion are now released immediately, reducing stale allocations and scheduling delays.
  * Lowers risk of IP exhaustion during high churn and speeds network cleanup for better cluster stability and responsiveness.

* **Tests**
  * Added unit tests covering the pod deletion IP-release behavior to ensure correctness.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_pods.go
+++ b/go-controller/pkg/ovn/base_network_controller_pods.go
@@ -1045,6 +1045,12 @@ func (bnc *BaseNetworkController) shouldReleaseDeletedPod(pod *corev1.Pod, switc
 		}
 	}
 
+	// this pod is being deleted before completion so there is no risk (and no
+	// need to check) that its IPs are being used by other pod
+	if !util.PodCompleted(pod) {
+		return true, nil
+	}
+
 	if bnc.wasPodReleasedBeforeStartup(string(pod.UID), nad) {
 		klog.Infof("Completed pod %s/%s was already released for nad %s before startup",
 			pod.Namespace,
--- a/go-controller/pkg/ovn/base_network_controller_pods_test.go
+++ b/go-controller/pkg/ovn/base_network_controller_pods_test.go
@@ -234,3 +234,41 @@ func TestBaseNetworkController_trackPodsReleasedBeforeStartup(t *testing.T) {
 		})
 	}
 }
+
+func TestBaseNetworkController_shouldReleaseDeletedPod(t *testing.T) {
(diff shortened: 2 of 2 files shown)


go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1045,6 +1045,12 @@ func (bnc *BaseNetworkController) shouldReleaseDeletedPod(pod *corev1.Pod, switc
 		}
 	}
 
+	// this pod is being deleted before completion so there is no risk (and no
+	// need to check) that its IPs are being used by other pod
+	if !util.PodCompleted(pod) {
```
qinqon, line 1050: Is there a possiblity to unit test this ?

qinqon (changes requested):
make sense to me

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1045,6 +1045,12 @@ func (bnc *BaseNetworkController) shouldReleaseDeletedPod(pod *corev1.Pod, switc
 		}
 	}
 
+	// this pod is being deleted before completion so there is no risk (and no
+	// need to check) that its IPs are being used by other pod
```
pperiyasamy, line 1049: for my understanding: can you explain how this pod IP is being used by other pod even before it's released ? because `shouldRelease` flag generated from this method decides whether to release the IP or not. so don't understand how this IP is used by other pod at this point of time.

jcaamano: Like this:

Pod A completes, its IPs are released
Pod B starts, gets assigned the same IPs
We handle again an event for pod A, we check pod B is already using the IPs so we don't release the IP as we handle the event for pod A

We filter out event of pods that were already completed, but this does not work across restarts

go-controller/pkg/ovn/base_network_controller_pods.go
jcaamano: Added (pretty dumb) test.

PR #5668, opened 2025-10-18
test: improve error checks in Zone Interconnect Operations test case

l8huang:
Previously, the error checks only verified that an error occurred, but did not confirm it matched the expected error, causing them to become outdated and inconsistent with the current code.

This change updates the checks to compare the actual error string against the expected one, improving test accuracy.
  

## Summary by CodeRabbit

* **Tests**
  * Improved zone interconnect handler tests with a clearer, isolated setup for error scenarios to reduce inter-test coupling and flakiness.
  * Broadened error assertions to use flexible substring matching, making failure detection more robust and less brittle.
  * Standardized initial test state before error-path runs to improve reliability and clarity of test outcomes.

### DIFF
--- a/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
+++ b/go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
@@ -869,6 +869,11 @@ var _ = ginkgo.Describe("Zone Interconnect Operations", func() {
 	})
 
 	ginkgo.Context("Error scenarios", func() {
+		ginkgo.BeforeEach(func() {
+			initialNBDB = []libovsdbtest.TestData{}
+			initialSBDB = []libovsdbtest.TestData{}
+		})
+
 		ginkgo.It("Missing annotations and error scenarios for local node", func() {
 			app.Action = func(ctx *cli.Context) error {
 				dbSetup := libovsdbtest.TestSetup{
@@ -903,22 +908,17 @@ var _ = ginkgo.Describe("Zone Interconnect Operations", func() {
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to get node id for node - node4")
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("failed to get node id for node - node4")))
 
 				// Set the node id
 				testNode4.Annotations = map[string]string{ovnNodeIDAnnotaton: "5"}
 				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to get the node transit switch port ips for node node4")
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("failed to get the node transit switch port ips for node node4")))
 
 				// Set the node transit switch port ips
 				testNode4.Annotations[ovnTransitSwitchPortAddrAnnotation] = "{\"ipv4\":\"100.88.0.5/16\"}"
 				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to get the network id for the network default on node node4")
-
-				// Set the network id for default network
-				testNode4.Annotations[ovnNodeNetworkIDsAnnotation] = "{\"default\":\"0\"}"
-				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to create/update cluster router ovn_cluster_router to add transit switch port rtots-node4 for the node node4")
+				gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("failed to create/update cluster router ovn_cluster_router to add transit switch port rtots-node4 for the node node4")))
 
 				// Create the cluster router
 				r := newOVNClusterRouter(types.DefaultNetworkName)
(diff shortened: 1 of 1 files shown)


go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
```diff
@@ -903,30 +908,25 @@ var _ = ginkgo.Describe("Zone Interconnect Operations", func() {
...
 
 				// Set the node transit switch port ips
 				testNode4.Annotations[ovnTransitSwitchPortAddrAnnotation] = "{\"ipv4\":\"100.88.0.5/16\"}"
 				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to get the network id for the network default on node node4")
-
-				// Set the network id for default network
-				testNode4.Annotations[ovnNodeNetworkIDsAnnotation] = "{\"default\":\"0\"}"
-				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to create/update cluster router ovn_cluster_router to add transit switch port rtots-node4 for the node node4")
```
ricky-rav, line 921: Is there a reason why you removed all these lines above?

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
```diff
@@ -989,41 +989,22 @@ var _ = ginkgo.Describe("Zone Interconnect Operations", func() {
...
-				err = libovsdbops.CreateOrUpdateChassis(libovsdbOvnSBClient, node4Chassis, encap)
-				gomega.Expect(err).NotTo(gomega.HaveOccurred())
-
-				err = zoneICHandler.AddRemoteZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to parse node node4 subnets annotation")
-
-				// Set node subnet annotation
-				testNode4.Annotations[ovnNodeSubnetsAnnotation] = "{\"default\":[\"10.244.5.0/24\"]}"
-				err = zoneICHandler.AddRemoteZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "unable to create static routes")
```
ricky-rav, line 1026: Also here you're removing a lot of lines.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
l8huang: AddLocalZoneNode() doesn't return error "failed to get the network id for the network default on node node4", so these lines are not required.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
l8huang: similar as above, the test case only check if a error occurred, but doesn't check if the error is expected, that case the test case does match the functional code. This change doesn't make any change on function code, and the test case still has the check at  line 1012 and 1013, which assert no error occured:
```
				err = zoneICHandler.AddRemoteZoneNode(&testNode4)
```
That means the removed lines are not needed anymore -- if there are any error, the case will fail.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
```diff
@@ -989,41 +989,22 @@ var _ = ginkgo.Describe("Zone Interconnect Operations", func() {
...
 
 				// Set the node transit switch port ips
 				testNode4.Annotations[ovnTransitSwitchPortAddrAnnotation] = "{\"ipv4\":\"100.88.0.5/16\"}"
 				err = zoneICHandler.AddRemoteZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to get the network id for the network default on node node4")
-
-				// Set the network id for default network
-				testNode4.Annotations[ovnNodeNetworkIDsAnnotation] = "{\"default\":\"0\"}"
-				err = zoneICHandler.AddLocalZoneNode(&testNode4)
-				gomega.Expect(err).To(gomega.HaveOccurred(), "failed to update chassis node4 for remote port tstor-node4")
```
ricky-rav, line 1012: So this error message doesn't exist anymore, ok.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
ricky-rav, line 1026: this error message is gone as well, OK.

ricky-rav: I see. In both cases we're eventually converging to a successful `AddLocalZoneNode` on one side and `AddRemoteZoneNode` on the other side. OK.

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
```diff
@@ -989,41 +989,22 @@ var _ = ginkgo.Describe("Zone Interconnect Operations", func() {
...
-
-				// Create remote chassis
-				node4Chassis := &sbdb.Chassis{Name: "cb9ec8fa-b409-4ef3-9f42-d9283c47aac9", Hostname: "node4", UUID: "cb9ec8fa-b409-4ef3-9f42-d9283c47aac9"}
-				encap := &sbdb.Encap{ChassisName: node4Chassis.Name, IP: "10.0.0.12"}
-				err = libovsdbops.CreateOrUpdateChassis(libovsdbOvnSBClient, node4Chassis, encap)
-				gomega.Expect(err).NotTo(gomega.HaveOccurred())
```
ricky-rav, line 1018: so here I suppose that your point is that this section that you're removing (about creating a remote chassis) isn't needed anymore, if eventually (later on) `AddRemoteZoneNode` works.

l8huang: yes

l8huang:
@ricky-rav thanks for approve this PR, is anything else needed to merge this PR?

ricky-rav:
> @ricky-rav thanks for approve this PR, is anything else needed to merge this PR?

You need a review from a maintainer. 
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/MAINTAINERS.md

PR #5669, opened 2025-10-20
node: fix DeleteEndpointSlice namespace deletion race condition

pliurh:
When processing an EndpointSlice delete event, if the namespace is
deleted before the event is processed, the old code would fail early
with a "namespace not found" error from GetActiveNetworkForNamespace
before executing delServiceRules. This left stale iptables/nftables
rules and OpenFlow rules in place, causing a resource leak.

Move the GetActiveNetworkForNamespace call to after delServiceRules
and before addServiceRules in DeleteEndpointSlice. This ensures old
service rules are always cleaned up, even if the namespace or UDN is
deleted during event processing.

Add graceful error handling for:
- Namespace NotFound: Skip adding new rules (namespace deleted)
- InvalidPrimaryNetworkError: Skip adding new rules (UDN deleted)
- Other errors: Propagate as before

Unit tests added to verify:
- delServiceRules executes before network lookup
- Graceful handling of namespace/UDN deletion scenarios
- Proper error propagation for non-graceful errors

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensure gateway cleanup runs before fetching network context; skip re-adding rules if the namespace was removed or primary network is invalid, and surface other lookup errors.

* **Refactor**
  * Reordered service-rule flow so deletions occur prior to active-network lookup to reduce race conditions and clarify error handling.

* **Tests**
  * Added Linux-only tests with local mocks covering namespace-missing, invalid-primary-network, generic lookup errors, and service-cache edge cases.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1405,11 +1405,6 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 	}
 	localEndpoints := npw.GetLocalEligibleEndpointAddresses(epSlices, svc)
 	if svcConfig, exists := npw.updateServiceInfo(*namespacedName, nil, &hasLocalHostNetworkEp, localEndpoints); exists {
-		netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(namespacedName.Namespace)
-		if err != nil {
-			return fmt.Errorf("error getting active network for service %s/%s: %w", namespacedName.Namespace, namespacedName.Name, err)
-		}
-
 		// Lock the cache mutex here so we don't miss a service delete during an endpoint delete
 		// we have to do this because deleting and adding iptables rules is slow.
 		npw.serviceInfoLock.Lock()
@@ -1418,6 +1413,28 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 		if err = delServiceRules(svcConfig.service, svcConfig.localEndpoints, npw); err != nil {
 			errors = append(errors, err)
 		}
+
+		// Get network info after deleting old rules, before adding new ones.
+		// This ensures old rules are cleaned up even if namespace/network is deleted,
+		// and allows graceful handling of deletion race conditions.
+		netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(namespacedName.Namespace)
+		if err != nil {
+			// If the namespace was deleted, skip adding new service rules
+			if apierrors.IsNotFound(err) {
+				klog.V(5).Infof("Namespace not found for service %s/%s during endpoint slice delete, skipping adding service rules",
+					namespacedName.Namespace, namespacedName.Name)
+				return utilerrors.Join(errors...)
+			}
+			// If the UDN was deleted, skip adding new service rules
+			if util.IsInvalidPrimaryNetworkError(err) {
+				klog.V(5).Infof("Skipping addServiceRules for %s/%s during endpoint slice delete: primary network invalid: %v",
+					namespacedName.Namespace, namespacedName.Name, err)
+				return utilerrors.Join(errors...)
+			}
+			errors = append(errors, fmt.Errorf("error getting active network for service %s/%s: %w", namespacedName.Namespace, namespacedName.Name, err))
+			return utilerrors.Join(errors...)
+		}
+
 		if err = addServiceRules(svcConfig.service, netInfo, localEndpoints, hasLocalHostNetworkEp, npw); err != nil {
 			errors = append(errors, err)
 		}
--- a/go-controller/pkg/node/gateway_shared_intf_test.go
+++ b/go-controller/pkg/node/gateway_shared_intf_test.go
@@ -1,0 +1,258 @@
+//go:build linux
+// +build linux
+
+package node
+
+import (
+	"fmt"
+
+	nadfake "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/fake"
+
+	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes/fake"
+
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
+	adminpolicybasedrouteclient "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/adminpolicybasedroute/v1/apis/clientset/versioned/fake"
+	udnfakeclient "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/clientset/versioned/fake"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/factory"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/kube"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// Note: Local mocks are used instead of FakeNetworkManager to test specific error conditions
+// (NotFound, InvalidPrimaryNetworkError) from GetActiveNetworkForNamespace. FakeNetworkManager
+// doesn't support error injection. And the tests here are not dependent on the methods that
+// FakeNetworkManager implements. If more node tests need this, we will enhance FakeNetworkManager.
+
+// mockNetworkManagerWithNamespaceNotFoundError simulates namespace deletion race condition
+type mockNetworkManagerWithNamespaceNotFoundError struct {
+	networkmanager.Interface
+}
+
+func (m *mockNetworkManagerWithNamespaceNotFoundError) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	notFoundErr := apierrors.NewNotFound(schema.GroupResource{Resource: "namespaces"}, namespace)
+	return nil, fmt.Errorf("failed to get namespace %q: %w", namespace, notFoundErr)
+}
+
+// mockNetworkManagerWithInvalidPrimaryNetworkError simulates UDN deletion scenario
+type mockNetworkManagerWithInvalidPrimaryNetworkError struct {
+	networkmanager.Interface
+}
+
+func (m *mockNetworkManagerWithInvalidPrimaryNetworkError) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	return nil, util.NewInvalidPrimaryNetworkError(namespace)
+}
+
+// mockNetworkManagerWithError tests that non-graceful errors are properly propagated
+type mockNetworkManagerWithError struct {
+	networkmanager.Interface
+}
+
+func (m *mockNetworkManagerWithError) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	return nil, fmt.Errorf("network lookup failed for namespace %q", namespace)
+}
+
+// verifyIPTablesRule checks if an iptables rule exists and asserts the expected state
+func verifyIPTablesRule(ipt util.IPTablesHelper, serviceIP string, servicePort, nodePort int32, shouldExist bool, message string) {
+	exists, err := ipt.Exists("nat", "OVN-KUBE-NODEPORT",
+		"-p", "TCP", "-m", "addrtype", "--dst-type", "LOCAL",
+		"--dport", fmt.Sprintf("%d", nodePort), "-j", "DNAT",
+		"--to-destination", fmt.Sprintf("%s:%d", serviceIP, servicePort))
+	Expect(err).NotTo(HaveOccurred())
+	if shouldExist {
+		Expect(exists).To(BeTrue(), message)
+	} else {
+		Expect(exists).To(BeFalse(), message)
+	}
+}
+
+// setupServiceAndEndpointSliceWithRules creates a service and endpoint slice, adds them to npw,
+// and verifies iptables rules are created. Returns the created endpoint slice.
+func setupServiceAndEndpointSliceWithRules(npw *nodePortWatcher, ipt util.IPTablesHelper, svcName, namespace, serviceIP, endpointIP string, servicePort, nodePort int32, annotations map[string]string) *discovery.EndpointSlice {
+	// Create service
+	service := newService(svcName, namespace, serviceIP,
+		[]corev1.ServicePort{{
+			Name:       "http",
+			Protocol:   corev1.ProtocolTCP,
+			Port:       servicePort,
+			TargetPort: intstr.FromInt(int(servicePort) + 8000), // e.g., 80 -> 8080
+			NodePort:   nodePort,
+		}},
+		corev1.ServiceTypeNodePort, nil, corev1.ServiceStatus{}, false, false)
+
+	// Create endpoint slice with endpoints
+	epPortName := "http"
+	epPortValue := servicePort + 8000 // Match targetPort
+	epPortProtocol := corev1.ProtocolTCP
+	epSlice := newEndpointSlice(
+		svcName,
+		namespace,
+		[]discovery.Endpoint{
+			{
+				Addresses: []string{endpointIP},
+			},
+		},
+		[]discovery.EndpointPort{
+			{
+				Name:     &epPortName,
+				Protocol: &epPortProtocol,
+				Port:     &epPortValue,
+			},
+		},
+	)
+
+	// Apply annotations if provided
+	if len(annotations) > 0 {
+		if epSlice.Annotations == nil {
+			epSlice.Annotations = make(map[string]string)
+		}
+		for k, v := range annotations {
+			epSlice.Annotations[k] = v
+		}
+	}
+
+	// Add service and endpoint slice
+	err := npw.AddService(service)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = npw.AddEndpointSlice(epSlice)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Verify iptables rules were created
+	verifyIPTablesRule(ipt, serviceIP, servicePort, nodePort, true, "iptables rule should exist before deletion")
+
+	return epSlice
+}
+
+var _ = Describe("DeleteEndpointSlice", func() {
+	var (
+		fakeClient *util.OVNNodeClientset
+		watcher    *factory.WatchFactory
+		npw        *nodePortWatcher
+		iptV4      util.IPTablesHelper
+		iptV6      util.IPTablesHelper
+	)
+
+	const (
+		nodeName      = "test-node"
+		testNamespace = "test-namespace"
+		testService   = "test-service"
+	)
+
+	BeforeEach(func() {
+		var err error
+		// Restore global default values before each test
+		Expect(config.PrepareTestConfig()).To(Succeed())
+		config.Gateway.Mode = config.GatewayModeLocal
+		config.IPv4Mode = true
+		config.IPv6Mode = false
+
+		fakeClient = &util.OVNNodeClientset{
+			KubeClient: fake.NewSimpleClientset(),
+		}
+		fakeClient.AdminPolicyRouteClient = adminpolicybasedrouteclient.NewSimpleClientset()
+		fakeClient.NetworkAttchDefClient = nadfake.NewSimpleClientset()
+		fakeClient.UserDefinedNetworkClient = udnfakeclient.NewSimpleClientset()
+
+		watcher, err = factory.NewNodeWatchFactory(fakeClient, nodeName)
+		Expect(err).NotTo(HaveOccurred())
+		err = watcher.Start()
+		Expect(err).NotTo(HaveOccurred())
+
+		// Initialize nodePortWatcher with default network manager
+		iptV4, iptV6 = util.SetFakeIPTablesHelpers()
+		npw = initFakeNodePortWatcher(iptV4, iptV6)
+		npw.watchFactory = watcher
+		npw.networkManager = networkmanager.Default().Interface()
+
+		// Initialize nodeIPManager (required for GetLocalEligibleEndpointAddresses)
+		k := &kube.Kube{KClient: fakeClient.KubeClient}
+		npw.nodeIPManager = newAddressManagerInternal(nodeName, k, nil, watcher, nil, false)
+	})
+
+	AfterEach(func() {
+		watcher.Shutdown()
+	})
+
+	Context("when UDN is deleted before processing endpoint slice", func() {
+		It("should execute delServiceRules and gracefully skip addServiceRules", func() {
+			// Setup service and endpoint slice with iptables rules
+			// Add UDN annotation to simulate a mirrored UDN EndpointSlice
+			epSlice := setupServiceAndEndpointSliceWithRules(npw, iptV4, testService, testNamespace, "10.96.0.2", "10.244.0.2", 80, 30081,
+				map[string]string{types.UserDefinedNetworkEndpointSliceAnnotation: "test-udn"})
+
+			// Replace network manager with one that returns InvalidPrimaryNetworkError
+			// This simulates UDN deletion scenario
+			npw.networkManager = &mockNetworkManagerWithInvalidPrimaryNetworkError{}
+
+			// Call DeleteEndpointSlice - should not return error
+			err := npw.DeleteEndpointSlice(epSlice)
+
+			// Should gracefully handle UDN deletion (no error)
+			Expect(err).NotTo(HaveOccurred())
+
+			// iptables rules should be deleted even when UDN is deleted
+			verifyIPTablesRule(iptV4, "10.96.0.2", 80, 30081, false, "iptables rule should be deleted even when UDN is deleted")
+		})
+	})
+
+	Context("when network lookup returns other errors", func() {
+		It("should execute delServiceRules but return error from network lookup", func() {
+			// Setup service and endpoint slice with iptables rules
+			epSlice := setupServiceAndEndpointSliceWithRules(npw, iptV4, testService, testNamespace, "10.96.0.3", "10.244.0.3", 80, 30082, nil)
+
+			// Replace network manager with one that returns a generic error
+			npw.networkManager = &mockNetworkManagerWithError{}
+
+			// Call DeleteEndpointSlice - should return error
+			err := npw.DeleteEndpointSlice(epSlice)
+
+			// Should return error for other types of failures
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("error getting active network"))
+			Expect(err.Error()).To(ContainSubstring(testNamespace))
+			Expect(err.Error()).To(ContainSubstring(testService))
+
+			// iptables rules should still be deleted even when error is returned
+			verifyIPTablesRule(iptV4, "10.96.0.3", 80, 30082, false, "iptables rule should be deleted even when error occurs")
+		})
+	})
+
+	Context("when service does not exist in cache", func() {
+		It("should return nil without error", func() {
+			// Create endpoint slice (but no service in cache)
+			epSlice := newEndpointSlice(testService, testNamespace, nil, nil)
+
+			// Call DeleteEndpointSlice when service not in cache
+			err := npw.DeleteEndpointSlice(epSlice)
+
+			// Should return nil (no-op when not in cache)
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
+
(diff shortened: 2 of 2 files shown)


weliang1:
The testing failed more frequently recently in 4.21 payloads

Slack threads from TRT team about the test failure:
https://redhat-internal.slack.com/archives/C02K89U2EV8/p1762940723576079
https://redhat-internal.slack.com/archives/C01CQA76KMX/p1762978795354319

Detailed Failure Analysis Report from Claude:
https://issues.redhat.com/browse/OCPBUGS-61864?focusedId=28465002&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28465002

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -853,7 +865,7 @@ func (npw *nodePortWatcher) AddService(service *corev1.Service) error {
 	if exists := npw.addOrSetServiceInfo(name, service, hasLocalHostNetworkEp, localEndpoints); !exists {
 		klog.V(5).Infof("Service Add %s event in namespace %s came before endpoint event setting svcConfig",
 			service.Name, service.Namespace)
-		if err := addServiceRules(service, netInfo, localEndpoints, hasLocalHostNetworkEp, npw); err != nil {
+		if err := addServiceRules(service, localEndpoints, hasLocalHostNetworkEp, npw, npw.networkManager); err != nil {
```
ricky-rav: Wait, so here we already have `netInfo`, but we're passing `npw.networkManager` as input so that `addServiceRules` can obtain `netInfo` once again? Would it be a bad idea to have `addServiceRules` take an optional `netInfo` that we can pass here?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1139,7 +1146,7 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) e
 	out, exists := npw.getServiceInfo(*svcNamespacedName)
 	if !exists {
 		klog.V(5).Infof("Endpointslice %s ADD event in namespace %s is creating rules", epSlice.Name, epSlice.Namespace)
-		if err = addServiceRules(svc, netInfo, localEndpoints, hasLocalHostNetworkEp, npw); err != nil {
+		if err = addServiceRules(svc, localEndpoints, hasLocalHostNetworkEp, npw, npw.networkManager); err != nil {
```
ricky-rav: uhm also here we already have `netInfo`. Are we interested in reducing the number of calls to `GetActiveNetworkForNamespace` or do we just want to clean up the logic in the code? What was the main reason behind this PR? Or actually, what was the bug behind this PR? :)

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1213,7 +1215,11 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 		if err = delServiceRules(svcConfig.service, svcConfig.localEndpoints, npw); err != nil {
 			errors = append(errors, err)
 		}
-		if err = addServiceRules(svcConfig.service, netInfo, localEndpoints, hasLocalHostNetworkEp, npw); err != nil {
+		if err = addServiceRules(svcConfig.service, localEndpoints, hasLocalHostNetworkEp, npw, npw.networkManager); err != nil {
+			// namespace not found, skip to add service rules
+			if strings.Contains(err.Error(), fmt.Sprintf("namespace \"%s\" not found", namespacedName.Namespace)) {
```
ricky-rav: would it be better to check a specific error type here?

go-controller/pkg/node/gateway_shared_intf.go
pliurh: The bug is https://issues.redhat.com/browse/OCPBUGS-61864. In `DeleteEndpointSlice`, the `netInfo` returned by `GetActiveNetworkForNamespace` is only used by `addServiceRules` but not `delServiceRules`. So this PR moves the logic into `addServiceRules`, `delServiceRules` will no longer be skipped when the namespace is not found. For `addServiceRules`, if the namespace is not found, it should be a no-op without returning an error. As the namespace has been removed, the service must have been removed, and the service's rules should be cleaned by the service controller.

go-controller/pkg/node/gateway_shared_intf.go
pliurh: I should move this check into `addServiceRules`. If the namespace is not found, there is no service to be handled; `addServiceRules` shall be a no-op.

go-controller/pkg/node/gateway_shared_intf.go
pliurh: `netInfo` is returned by `GetActiveNetworkForNamespace`, this PR moves `GetActiveNetworkForNamespace` call into `addServiceRules`, so we don't need to pass `netInfo` to it.

ricky-rav: My concern was that we're calling `GetActiveNetworkForNamespace` twice in this execution path (on line 849 above and in `addServiceRules`), but I see that   `GetActiveNetworkForNamespace`  in the end only inspects some maps. OK.

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: oh I see, thanks for the pointer.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -672,9 +672,21 @@ func (npw *nodePortWatcher) updateServiceInfo(index ktypes.NamespacedName, servi
...
-	// For dpu or Full mode
+func addServiceRules(service *corev1.Service, localEndpoints util.PortToLBEndpoints, svcHasLocalHostNetEndPnt bool, npw *nodePortWatcher, networkManager networkmanager.Interface) error {
+	// Fetch network info
+	var netInfo util.NetInfo
 	var err error
+	if networkManager != nil {
+		netInfo, err = networkManager.GetActiveNetworkForNamespace(service.Namespace)
+		if err != nil {
+			return fmt.Errorf("error getting active network for service %s/%s: %w", service.Namespace, service.Name, err)
+		}
```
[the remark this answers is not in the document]
ricky-rav: @pliurh should we return nil also when `GetActiveNetworkForNamespace` throws an `InvalidPrimaryNetworkError`?

ricky-rav (dismissed):
/LGTM
Thanks, @pliurh !

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1344,7 +1362,7 @@ func (npw *nodePortWatcher) AddEndpointSlice(epSlice *discovery.EndpointSlice) e
 	out, exists := npw.getServiceInfo(*svcNamespacedName)
 	if !exists {
 		klog.V(5).Infof("Endpointslice %s ADD event in namespace %s is creating rules", epSlice.Name, epSlice.Namespace)
-		if err = addServiceRules(svc, netInfo, localEndpoints, hasLocalHostNetworkEp, npw); err != nil {
```
tssurya, line 1347: hmm sounds like for AddEndpointSlice we weren't checking for active namespace? seems like this was done only for services and now we are introducing the check also for EPS? was that intentional?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1418,7 +1431,7 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
 		if err = delServiceRules(svcConfig.service, svcConfig.localEndpoints, npw); err != nil {
 			errors = append(errors, err)
 		}
```
tssurya, line 1415: I think one of the easiest fixes I had suggested was just moving the getactivenetworkfornamespace after delServiceRules and before calling addServiceRules.. but yea I don't mind the approach taken here to move it into addServiceRules itself so that the caller doesn't have burden of checking that.. what bothers me is it would be good to understand if EPS and other spots that call addServiceRules really need that check introduced or not.

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,594 @@
...
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// mockNetworkManagerWithError is a mock implementation of networkmanager.Interface
+// that returns a non-NotFound error from GetActiveNetworkForNamespace
+type mockNetworkManagerWithError struct {
```
tssurya: can you use type FakeNetworkManager struct { instead?
I think other tests are using this at least in controller and cm side.. maybe node can use it too?

and we can simulate the namespace delete right? didn't the fakeclient delete namespace work? is that why we need this?

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,594 @@
...
+						},
+					},
+				},
+			}
+
+			err := addServiceRules(service, nil, false, nil, networkManager)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should handle SCTP protocol", func() {
```
tssurya: hmm were these tests AI generated? what's special about TCP/SCTP/UDP in this PR for the logic you added? We didn't change any of the conntrack stuff right? If it assisted by AI please amend commit msg with assisted-by message..

Are all of these tests related to this delete PR? Could you please provide more info on test coverage in the commit msg so its easier to understand what are the new tests you added and what each test is aiming to test from test design perspective.

cause the deleting namespace causes the glitch bug you are fixing shouldn't really matter on type of service or protocols right? not sure if  I am missing something.

go-controller/pkg/node/gateway_shared_intf.go
pliurh: Yes, we were. The `netInfo` is returned by `GetActiveNetworkForNamespace`. Although there is a slight redundancy in AddService and AddEndpointSlice where GetActiveNetworkForNamespace is called once in the caller for other usage before addServiceRules.

go-controller/pkg/node/gateway_shared_intf.go
pliurh: GetActiveNetworkForNamespace was invoked each time addServiceRules is called. I didn't remove the `GetActiveNetworkForNamespace` from some callers, as it's required by other code there. It introduces a slight redundancy in the "Add" paths. But I think it's an acceptable performance tradeoff for code clarity.

If you don't like this idea, I can go with the easy fix you suggested.

go-controller/pkg/node/gateway_shared_intf_test.go
pliurh: The FakeNetworkManager doesn't use fakeclient. It returns `InvalidPrimaryNetworkError` when the namespace does not exist. For tests here, as they don't use the methods that FakeNetworkManager implemented; it is just more convenient to mock the errors that `GetActiveNetworkForNamespace` may return with a local implementation.

go-controller/pkg/node/gateway_shared_intf_test.go
pliurh: Yeah, the test code is AI-generated. The PR is not just about deletion. It's a refactor of `addServiceRules`. I added those tests as we didn't have UT for this function at all.

tssurya: ack please add Assissted-by: Name of the model in the commit message for AI generated tests.

pliurh: done.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -734,9 +734,32 @@ func (npw *nodePortWatcher) updateServiceInfo(index ktypes.NamespacedName, servi
...
+	if networkManager != nil {
+		netInfo, err = networkManager.GetActiveNetworkForNamespace(service.Namespace)
+		if err != nil {
+			// If the namespace was deleted, there's no need to process service rules
+			if apierrors.IsNotFound(err) {
+				klog.Warningf("Namespace %s not found for service %s/%s, skipping adding service rules",
+					service.Namespace, service.Namespace, service.Name)
+				return nil
+			}
+			// The InvalidPrimaryNetworkError is returned when the UDN is not found because it has already been deleted.
```
tssurya: it could also be due to UDN not being created yet but the namespace has a label..doesn't necessarily mean deleted..should we be retrying for this or will the UDN creation event trigger another reconcile for add service or update service?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -734,9 +734,32 @@ func (npw *nodePortWatcher) updateServiceInfo(index ktypes.NamespacedName, servi
...
+			if util.IsInvalidPrimaryNetworkError(err) {
```
tssurya: Should we handle `IsUnprocessedActiveNetworkError` as well or do we want to retry?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -734,9 +734,32 @@ func (npw *nodePortWatcher) updateServiceInfo(index ktypes.NamespacedName, servi
...
+				klog.V(5).Infof("Skipping addServiceRules for %s/%s: primary network invalid: %v", service.Namespace, service.Name, err)
+				return nil
+			}
+			return fmt.Errorf("error getting active network for service %s/%s: %w", service.Namespace, service.Name, err)
+		}
+	} else {
+		// This should never happen in practice, but handle defensively
```
tssurya: this is strange.. was this part AI assisted as well? I've seen this pattern that AI does too defensive coding with such comments but there is no explanation on how this could be nil cause it also says it should never be the case..is it for some dpu mode or something? would be good if we can get rid of the if else cause the old code wasn't being defensive before calling getactivenetwork..

go-controller/pkg/node/gateway_shared_intf.go
tssurya: hmm so now we see this call to `GetActiveNetworkForNamespace` like you mentioned being done twice if coming from the `AddService` and `AddEndpointSlice` paths..(https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669/changes#r2621599304)

looks like for `UpdateService` and `DeleteEndpointSlice` calls its done only once in addServiceRules... so IIUC only Add calls have it redundant is that right?

Redundancy is not great because `GetActiveNetworkForNamespace` is not exactly performant efficient much since it seems to be listing UDNs, making a copy of the networkInfo etc..so better to avoid that being done twice if we can. wdyt? We can refactor later if necessary...

The other bit is given `GetActiveNetworkForNamespace` is being called earlier on it self in `AddService` and `AddEndpointSlice`  there is still that issue of namespace being deleted and us getting an add event - though that's almost very rare but still something to handle....it would have been good if we could avoid the calls there but seems like not and we might have more cases like that for the caller needing the netInfo before calling addServiceRules..

So let's go with the simple fix for now to fix the bug in question...- I guess that means before the `GetActiveNetworkForNamespace` we need to check for IsNotFound and IsInvalidPrimaryNetworkError? (although here it could also mean the UDN was not created yet.. so in that case I hope we will get another service add event once the UDN/NAD is created or will be not retry that?)

go-controller/pkg/node/gateway_shared_intf_test.go
tssurya: I see, please add a comment to that effect on why FakeNetworkManager was not reused here and why these tests are not following the same pattern as other tests in ovnkube-controller and cluster-manager side..

Since these are the first set of tests here on services side and if we do a mocked version, I worry in future we will start growing this interface and then we have one fake and one mocked version both doing different things..

I understand the convinience part totally, but perhaps at least best to leave a comment for paper trail so that its clear we thought about it and deliberately chose this since FakeNetworkManager isn't returning IsNotFound and that its not possible to add it there due to reason X?

tssurya (changes requested):
I think let's go with the simple fix for the redundancy reasoning we have.. and we might have more such cases in future where caller needs the netInfo much earlier on than inside addServiceRules..at least for now it doesn't seem like we are ready for doing that..I definitely think this part of code needs bit more structure and cleanup but that's outside the scope of this PR

Also could you please update your PR description with the CI error and the actual bug around endpointslice delete and why we are doing this fix. I totally get the what's done part in the describe, but its missing the why part..

pliurh:
Refactored according to @tssurya 's comments.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1418,6 +1413,28 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
...
+		// and allows graceful handling of deletion race conditions.
+		netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(namespacedName.Namespace)
+		if err != nil {
+			// If the namespace was deleted, skip adding new service rules
+			if apierrors.IsNotFound(err) {
+				klog.V(5).Infof("Namespace not found for service %s/%s during endpoint slice delete, skipping adding service rules",
+					namespacedName.Namespace, namespacedName.Name)
+				return utilerrors.Join(errors...)
+			}
+			// If the UDN was deleted, skip adding new service rules
```
tssurya, line 1428: PTAL https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669#discussion_r2653465338

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1418,6 +1413,28 @@ func (npw *nodePortWatcher) DeleteEndpointSlice(epSlice *discovery.EndpointSlice
...
+			if util.IsInvalidPrimaryNetworkError(err) {
+				klog.V(5).Infof("Skipping addServiceRules for %s/%s during endpoint slice delete: primary network invalid: %v",
+					namespacedName.Namespace, namespacedName.Name, err)
+				return utilerrors.Join(errors...)
+			}
+			errors = append(errors, fmt.Errorf("error getting active network for service %s/%s: %w", namespacedName.Namespace, namespacedName.Name, err))
```
tssurya, line 1434: PTAL https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669#discussion_r2653470897

I think what you have here is perhaps correct - but what we need to confirm is if NAD does get created layer would network add go program all services again or not..

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
+//go:build linux
```
tssurya, line 1: is there something specific in the tests that this should only be built on linux distro?

go-controller/pkg/node/gateway_shared_intf_test.go
tssurya: this was addressed. resolving this.

go-controller/pkg/node/gateway_shared_intf.go
tssurya: this was addressed. resolving this.

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+			npw.networkManager = &mockNetworkManagerWithInvalidPrimaryNetworkError{}
+
+			// Call DeleteEndpointSlice - should not return error
+			err := npw.DeleteEndpointSlice(epSlice)
+
+			// Should gracefully handle UDN deletion (no error)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Service should still be in cache (delServiceRules executed)
+			_, exists := npw.serviceInfo[svcKey]
```
tssurya: isn't it actually better to check for iptable rules gone or flows gone to be sure that delete service rules was called? so check rules exist at startup and then check they are gone?
If addepSlice wasn't called then none of the rule cleanup we are testing for delServiceRules is being called right?
hm I am not sure how service still in cache is a guarantee for delServiceRules was actually called

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+			// Create endpoint slice
+			epSlice := &discovery.EndpointSlice{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      testService + "-xyz",
+					Namespace: testNamespace,
+					Labels: map[string]string{
+						discovery.LabelServiceName: testService,
+					},
+					Annotations: map[string]string{
+						types.UserDefinedNetworkEndpointSliceAnnotation: types.DefaultNetworkName,
```
tssurya: but this is not a UDN endpoint is it? does the default network endpointslices have that annotation?

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+			}
+
+			// Replace network manager with one that returns InvalidPrimaryNetworkError
+			// This simulates UDN deletion scenario
+			npw.networkManager = &mockNetworkManagerWithInvalidPrimaryNetworkError{}
+
+			// Call DeleteEndpointSlice - should not return error
+			err := npw.DeleteEndpointSlice(epSlice)
+
+			// Should gracefully handle UDN deletion (no error)
```
tssurya, line 200: so which scenario is this? the `IsInvalidPrimaryNetworkError` ?

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+			// Replace network manager with one that returns a generic error
+			npw.networkManager = &mockNetworkManagerWithError{}
+
+			// Call DeleteEndpointSlice - should return error
+			err := npw.DeleteEndpointSlice(epSlice)
+
+			// Should return error for other types of failures
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("error getting active network"))
```
tssurya, line 221: this is nice test

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+			// Add service to npw cache
+			svcKey := ktypes.NamespacedName{Namespace: testNamespace, Name: testService}
+			npw.serviceInfo[svcKey] = &serviceConfig{
+				service:               service,
+				hasLocalHostNetworkEp: false,
+				localEndpoints:        util.PortToLBEndpoints{},
+			}
+
+			// Create endpoint slice
```
tssurya: are we really creating the ep slice and the namespace somewhere in these tests on startup?

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+
+	Context("when namespace is deleted before processing endpoint slice", func() {
+		It("should clean up old rules even when namespace is gone", func() {
+			// This test verifies the core fix: delServiceRules runs BEFORE network lookup
+			// So even if namespace is deleted, old rules are cleaned up
+
+			// Create service with some endpoints using helper
+			service := createTestService(testNamespace, testService, "10.96.0.10", 30090)
+
+			// Add service to cache with local endpoints
```
tssurya: this code is used mostly in localgateway, maybe better to change the mode to local though codewise we setup things for local and shared so not a big deal

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+					V4IPs: []string{"10.244.0.5"},
+				},
+			}
+			npw.serviceInfo[svcKey] = &serviceConfig{
+				service:               service,
+				hasLocalHostNetworkEp: false,
+				localEndpoints:        localEndpoints,
+			}
+
+			// Create endpoint slice
```
tssurya: time for a util to return this like you have for service? unless there is already some helpers in `gateway_localnet_linux_test.go`  i.e `newEndpointSlice`

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+type mockNetworkManagerWithError struct {
+	networkmanager.Interface
+}
+
+func (m *mockNetworkManagerWithError) GetActiveNetworkForNamespace(namespace string) (util.NetInfo, error) {
+	return nil, fmt.Errorf("network lookup failed for namespace %q", namespace)
+}
+
+// Helper function to create test services with different configurations
+func createTestService(namespace, name, clusterIP string, nodePort int32) *corev1.Service {
```
tssurya: can you reuse `newService` in gateway_localnet_linux_test.go so that each file doesn't end up with a helper util and its just one util per package? if there is something special you need here then its fine

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,306 @@
...
+						types.UserDefinedNetworkEndpointSliceAnnotation: types.DefaultNetworkName,
+					},
+				},
+				AddressType: discovery.AddressTypeIPv4,
+			}
+
+			// Simulate namespace deletion - network lookup will fail
+			npw.networkManager = &mockNetworkManagerWithNamespaceNotFoundError{}
+
+			// Before fix: delServiceRules would NOT run (early return on network lookup error)
```
tssurya: comments like this - with before fix and after fix start loosing meaning once PR is merged.. - I think its another downfall of AI

tssurya (changes requested):
code wise fix looks good. - would be good to also do a d/s verification against CI where we caught this.
just a few questions there and on the testing side..(I've reviewed full PR including tests, so future iterations should be simpler and smaller)

thanks for the new version, looks much better.

go-controller/pkg/node/gateway_shared_intf.go
pliurh: I don't understand your concern. I don't think that is an issue `DeleteEndpointSlice()` needs to care about.

go-controller/pkg/node/gateway_shared_intf_test.go
pliurh: Yes.

go-controller/pkg/node/gateway_shared_intf_test.go
pliurh: No. We don't need to. We don't need t get ep slice or namespace with the fake client.

go-controller/pkg/node/gateway_shared_intf.go
tssurya: sorry I wasn't clear in the comment. What I mean is during endpointslice delete when we want to reprogramme the rules for other endpointslices for this service - i.e when we call addServiceRules, and before that we call GetActiveNetworkForNamespace here... that function returns IsUnprocessedActiveNetworkError type error. This means the NAD was not yet created successfully for this namespace i.e CM's udn controller hasn't processed the NAD here. In that case should we retry the error OR is there a path in the node pkg side where upon NAD creation i.e network add we would get another event for processing service rules or not - I was just thinking out loud but then concluded that what you have here is correct and that retrying for that scenario is perhaps the right way.

go-controller/pkg/node/gateway_shared_intf_test.go
pliurh: This test is for UDN endpoint. I should not use `DefaultNetworkName` though.

go-controller/pkg/node/gateway_shared_intf_test.go
pliurh: The test uses `newService` and `newEndpointSlice` which are defined in `gateway_localnet_linux_test.go`.

pliurh:
Update the tests to address @tssurya 's comments.

go-controller/pkg/node/gateway_shared_intf_test.go
```diff
@@ -0,0 +1,252 @@
...
+	if shouldExist {
+		Expect(exists).To(BeTrue(), message)
+	} else {
+		Expect(exists).To(BeFalse(), message)
+	}
+}
+
+// setupServiceAndEndpointSliceWithRules creates a service and endpoint slice, adds them to npw,
+// and verifies iptables rules are created. Returns the created endpoint slice.
+func setupServiceAndEndpointSliceWithRules(npw *nodePortWatcher, ipt util.IPTablesHelper, svcName, namespace, serviceIP, endpointIP string, servicePort, nodePort int32) *discovery.EndpointSlice {
```
tssurya: nice!

tssurya:
> Update the tests to address @tssurya 's comments.

thanks a lot, the tests look good now.

just two more comments:

1. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669#discussion_r2660479930 was missed
2. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669#discussion_r2661585874 from coderabbit if you want to consider that

go-controller/pkg/node/gateway_shared_intf.go
pliurh: `DeleteEndpointSlice` cannot be invoked when the namespace is labeled but the UDN/CUDN CR was never created. Since this function can only process services that are already in cache. For the service to be in the cache, `AddService` must have succeeded once. And `AddService` also invokes `GetActiveNetworkForNamespace`. So here`GetActiveNetworkForNamespace` could only return `InvalidPrimaryNetworkError` when the UDN was created and removed.

pliurh:
update to resolve the coderabbit comment https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5669#discussion_r2661585874

tssurya (approved):

Issue #5672, opened 2025-10-20
OKEP: Support for metadata service with UDNs


trozet:
### What would you like to be added?

VM cloud platforms typically offer metadata service at 169.254.169.254. For kubevirt VMs running on ovn-kubernetes, we would like to offer the same kind of support. How that metadata service operates as far as API, etc is beyond the scope of this work. This is simply to enable someone (maybe in the future kubevirt) to bring their own metadata service and make it accessible to pods via 169.254.169.254.

### Why is this needed?

169.254.169.254 is a local IP, and kubernetes does not allow you to specify an external IP with this IP, as it does not work with kube-proxy. Our SDN works differently though and can support it. Having this support is useful for VMs who need to access metadata service that either:
1. do not want to use noseed cloud init
2. need to access metadata post-cloud init with dynamic information

PR #5673, opened 2025-10-20
OKEP-5672: Metadata Service

trozet:
Closes: #5672

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design documentation for the Metadata Service feature, detailing API specifications, network configuration extensions, implementation approach, testing strategies, known limitations, and future integration considerations. Documentation is now available for reference.

### DIFF
--- a/docs/okeps/okep-5672-metadata-service.md
+++ b/docs/okeps/okep-5672-metadata-service.md
@@ -1,0 +1,245 @@
+# OKEP-5672: Metadata Service
+
+* Issue: [#5672](https://github.com/ovn-org/ovn-kubernetes/issues/5672)
+
+## Problem Statement
+
+VM cloud platforms typically offer metadata service at 169.254.169.254.
+For Kubevirt VMs running on ovn-kubernetes, we would like to offer the same kind of support.
+
+## Goals
+
+* Allow a user to configure metadata service access with their UDN that enables access to 169.254.169.254 with Internal Traffic Policy (ITP) local
+* Limit access to this IP within the SDN (ovn-kubernetes networked pods)
+
+## Non-Goals
+
+* Standardizing any kind of REST API, or metadata pod backend for the service
+* IPv6 Support
+
+## Introduction
+
+Most VM cloud platform providers expose a metadata service to VMs. This service is used by the VM to get cloud-init
+data at boot time, as well as to query for metadata post-boot. The providers all consistently use 169.254.169.254 as
+the IP to access metadata. However, each platform uses different REST API to query this metadata. Kubevirt offers
+cloud-init via a [NoCloud](https://cloudinit.readthedocs.io/en/latest/reference/datasources/nocloud.html) option, but
+this does not provide a dynamic source that can be queried post-boot.
+
+The goal of this feature is to provide a way that a user can specify to OVN-Kubernetes that they would like to use
+a metadata service provided at 169.254.169.254. OVN-Kubernetes will handle creating an OVN Load Balancer that will use
+169.254.169.254 as the VIP, and then the user can specify backends. How these backends are implemented is beyond the
+scope of this enhancement or OVN-Kubernetes. In the future it may be standardized by a higher layer, like Kubevirt.
+
+Note there is limited support for metadata service on VM cloud platforms for IPv6. There is also no consistent IPv6
+address used. For that reason, IPv6 support will be deferred from this enhancement.
+
+## User-Stories/Use-Cases
+
+Story 1: Provide Dynamic Metadata to a VM
+
+As a tenant using Kubevirt, I want to enable my VMs to be able to query a metadata service I manage as pods
+so that my VM can get updated token information during runtime. The VM should be able to query 169.254.169.254
+to get dynamically updated information.
+
+## Proposed Solution
+
+This proposal includes specifying new API within the UDN CRD spec to allow a user to specify the metadata service
+name in their UDN. Once identified, OVN-Kubernetes will add an OVN Load Balancer with the 169.254.169.254 VIP for
+this service.
+
+### API Details
+
+CUDN and UDN API will be changed to allow a new field to reference a metadata service name:
+
+```yaml
+apiVersion: k8s.cni.cncf.io/v1
+kind: ClusterUserDefinedNetwork
+metadata:
+  name: shared-primary
+spec:
+  namespaceSelector:
+    matchLabels:
+      cudn-access: shared
+  network:
+    topology: Layer3
+    layer3:
+      role: Primary
+      subnets:
+      - cidr: 10.20.0.0/16
+    metadataServiceRef:
+      name: shared-metadata
+      namespace: infra-services
+```
+
+For CUDN, the namespace is required to indicate which namespace contains the service. For UDN, it is implied:
+
+```yaml
+apiVersion: k8s.cni.cncf.io/v1
+kind: UserDefinedNetwork
+metadata:
+  name: tenant-a-primary
+  namespace: tenant-a
+spec:
+  topology: Layer3
+  layer3:
+    role: Primary
+    subnets:
+    - cidr: 10.10.0.0/16
+  metadataServiceRef:
+    name: tenant-a-metadata
+```
+
+Additionally, the metadataServiceRef is only valid for Primary networks, since Secondary networks do not support
+services. Full CRD golang changes:
+
+```go
+// MetadataServiceRef specifies a metadata service to be exposed via the special VIP 169.254.169.254.
+type MetadataServiceRef struct {
+	// Name is the name of the Service that backs the metadata endpoint.
+	//
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +required
+	Name string `json:"name"`
+
+	// Namespace is the namespace of the Service.
+	//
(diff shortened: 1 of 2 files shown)


docs/okeps/okep-5672-metadata-service.md
```diff
@@ -0,0 +1,245 @@
...
+* Issue: [#5672](https://github.com/ovn-org/ovn-kubernetes/issues/5672)
+
+## Problem Statement
+
+VM cloud platforms typically offer metadata service at 169.254.169.254.
+For Kubevirt VMs running on ovn-kubernetes, we would like to offer the same kind of support.
+
+## Goals
+
+* Allow a user to configure metadata service access with their UDN that enables access to 169.254.169.254 with Internal Traffic Policy (ITP) local
```
dave-tucker, line 12: ITP local is an implementation detail. Might be worth calling it out more generically within the goals section.

docs/okeps/okep-5672-metadata-service.md
```diff
@@ -0,0 +1,245 @@
...
+* Limit access to this IP within the SDN (ovn-kubernetes networked pods)
```
dave-tucker, line 13: Can you explain the type of limits you're expecting? i.e is it that only Pods should be able to access it? Or only certain pods?

docs/okeps/okep-5672-metadata-service.md
```diff
@@ -0,0 +1,245 @@
...
+* No host network endpoint support.
+* No inter-UDN support (like with Network Connect).
+* No support for any other service type other than internal.
+* No IPv6 support.
+
+## OVN-Kubernetes Version Skew
+
+Targeted for release 1.2.
+
+## Alternatives
```
dave-tucker, line 232: Possibly worth calling out the other alternative(s) that were discussed.

1. Making it a ClusterIP service, that's ITP local, and adding an annotation to specify an additional VIP. This was discounted because we can't do any validation of the IP address range requested in the annotation to avoid invalid or overlapping ranges. Therefore the service would be deployed, but potentially inaccessible over the custom VIP.
2. Making a Headless Service. This made less sense, since ITP-local only applies to services that use kube-proxy (headless services don't do that). Equally headless services don't have VIPs, so adding an annotation that assigns a VIP (just to avoid allocating a ClusterIP) seems very weird.

docs/okeps/okep-5672-metadata-service.md
```diff
@@ -0,0 +1,245 @@
...
+
+As a tenant using Kubevirt, I want to enable my VMs to be able to query a metadata service I manage as pods
+so that my VM can get updated token information during runtime. The VM should be able to query 169.254.169.254
+to get dynamically updated information.
+
+## Proposed Solution
+
+This proposal includes specifying new API within the UDN CRD spec to allow a user to specify the metadata service
+name in their UDN. Once identified, OVN-Kubernetes will add an OVN Load Balancer with the 169.254.169.254 VIP for
+this service.
```
dave-tucker, line 48: When we chatted on Slack, I thought we were proposing using https://ovn-kubernetes.io/okeps/okep-5233-preconfigured-udn-addresses/ and extending those concepts from L2 UDNs to apply to L3 UDNs too, given we think that it could be required for EVPN.

So the UX would be:

1. Add `reservedSubnets: ["169.254.169.254/32"]` to your UDN
2. When deploying your metadata service pod add this annotation: `v1.multus-cni.io/default-network: |
      {"name": "default", "namespace": "ovn-kubernetes", "ips": ["169.254.169.254"], "mac": "00:1A:2B:3C:4D:5E", "ipam-claim-reference": "my-claim"}`

kyrtapz:
@trozet a wild idea but have you considered addressing this need by using the new `ServiceCIDR` resource in k8s? We are not supporting that today but maybe there is a chance to kill two birds with one stone :wink:

docs/okeps/okep-5672-metadata-service.md
trozet: Yeah its more of a use case thing, I should move it there

docs/okeps/okep-5672-metadata-service.md
trozet: only ovn-networked pods within the UDN. Host networked pods, or the host cannot get to this metadata service.(The service is only rendered on the OVN worker switch)

docs/okeps/okep-5672-metadata-service.md
trozet: Yeah then I realized it wont work cause we need a pod per node attached to the UDN, and they cant all have the same IP, especially on a layer 2 network. So i went back to the service approach, but made it a singleton in the API so I thought it would be OK.

trozet:
> @trozet a wild idea but have you considered addressing this need by using the new `ServiceCIDR` resource in k8s? We are not supporting that today but maybe there is a chance to kill two birds with one stone 😉

@kyrtapz you mean this thing? https://kubernetes.io/docs/tasks/network/extend-service-ip-ranges/

I didn't know about it, but it looks like it just allows you to add more service ranges right? I cant choose which service IP I want to use?

kyrtapz:
> > @trozet a wild idea but have you considered addressing this need by using the new `ServiceCIDR` resource in k8s? We are not supporting that today but maybe there is a chance to kill two birds with one stone 😉
> 
> @kyrtapz you mean this thing? https://kubernetes.io/docs/tasks/network/extend-service-ip-ranges/
> 
> I didn't know about it, but it looks like it just allows you to add more service ranges right? I cant choose which service IP I want to use?

I was thinking you could try to add a range like `169.254.169.254/32` and create a service with that as the ClusterIP. Not sure if it would work but I was curious if you considered it.

dave-tucker:
> @kyrtapz you mean this thing? https://kubernetes.io/docs/tasks/network/extend-service-ip-ranges/

> I didn't know about it, but it looks like it just allows you to add more service ranges right? I cant choose which service IP I want to use?

Correct, you can't specify which IP you want, you just get allocated one from any of the available ranges.
The issue then is if you were to add 169.254.169.254/32 it could only be claimed by one service.
The requirement is to have one of these services per UDN.

We basically need `internalIPs` in the Service spec as opposed to [externalIPs](https://kubernetes.io/docs/concepts/services-networking/service/#external-ips) 😆

kyrtapz:
> Correct, you can't specify which IP you want, you just get allocated one from any of the available ranges.

You can, this works (k8s doesn't deny it):
```
apiVersion: networking.k8s.io/v1
kind: ServiceCIDR
metadata:
  name: metadata-cidr
spec:
  cidrs:
  - 169.254.169.252/30
---
apiVersion: v1
kind: Service
metadata:
  name: metadata-service
  namespace: default
spec:
  clusterIP: 169.254.169.254
  ports:
  - port: 80
    protocol: TCP
    targetPort: 80
    name: http
  type: ClusterIP
```
NOTE: `/30` was required because K8s doesn't handle `/32` and `/31` CIDRs correctly: https://github.com/kubernetes/kubernetes/blob/497ed03ae77411ed2d030c5aaec9ec1155392549/pkg/api/servicecidr/servicecidr.go#L95-L113

> The issue then is if you were to add 169.254.169.254/32 it could only be claimed by one service. The requirement is to have one of these services per UDN.

This seems like a separate issue where we want to have a service accessible from multiple UDNs.
If we want to have the same "service" to be resolved to different endpoints on every UDNs it is kinda confusing with how we are handling k8s services :D We now need a new UDNService CRD :rofl: 
If you both agree this is not the way to go I will stop derailing this conversation :smile:

dave-tucker:
@kyrtapz that's really cool! thanks for correcting me - i've learned something today.
tbh i'm not thrilled or convinced with any of the options we have on hand so please derail away 😆

docs/okeps/okep-5672-metadata-service.md
qinqon: > Yeah then I realized it wont work cause we need a pod per node attached to the UDN, and they cant all have the same IP, especially on a layer 2 network

@trozet do we need a pod per node for HA purposes ? or the metadata that you need to configure is different per node ? 

Aso take into account that metadata servers can be access from VMs before the global ip is configured, so the src ip will be a LLA too, but as long as LLA at VM are generated from MAC addresses OVN accept them and port security is not an issue.

qinqon:
@trozet one test we should do (I think we don't even need VMs for it) is to attach a pair of pods to a layer2 network and configure a ipv4 LLA generated from mac address on both, if you are able to connect between them it's clear that the OVN's port security is not an issue as far as the LLA is generated from mac (don't know if this is a hard requirement for your use cases, it has some security concerns)

trozet:
@kyrtapz 
>This seems like a separate issue where we want to have a service accessible from multiple UDNs.
If we want to have the same "service" to be resolved to different endpoints on every UDNs it is kinda confusing with how we are handling k8s services :D We now need a new UDNService CRD 🤣
If you both agree this is not the way to go I will stop derailing this conversation 😄

We don't want a global metadata service that anyone can query. That would mean that tenants can query instance information about other tenants. OpenStack implements a metadata server per tenant network. We want a service per UDN. Services are already namespace scoped, and per UDN. For services on a CUDN they are reachable by anyone on that CUDN.

trozet:
we decided this doesn't really fit as a UDN API or a k8s service modification, so the plan is to implement in admin internal route CRD, which is downstream only now but will be upstreamed later

Issue #5674, opened 2025-10-20
Support for detecting ovnkube dpu failure


trozet:
### What would you like to be added?

With the DPU (trusted) architecture, we have ovnkube running on the host (dpu-host) mode, and ovnkube running in the DPU (dpu) mode. Kubelet talks on the host to dpu-host mode, which annotates the VF rep information so that dpu ovnkube can plug the port into OVS and handle all of the OVN configuration. The dpu component is managed outside of the host kubernetes cluster.

The purpose of this addition is to be able to detect when the dpu side goes down, and update the CNI status as not ready on from the host.

### Why is this needed?

If the dpu component goes down, there is nothing to signal that there is a CNI failure on the host. The dpu-host will just keep waiting for the dpu side to configure until CNI ADD timeout occurs.

trozet:
@aserdean FYI

PR #5675, opened 2025-10-20
OKEP 5674: DPU Healthcheck

trozet:
Closes: #5674

## Summary by CodeRabbit

* **Documentation**
  * Added DPU Healthcheck proposal (OKEP-5674) describing DPU-host/DPU-mode architecture, health-check objectives for detecting DPU down states, impacts on networking and node readiness, node-lease-based health mechanism, configuration knobs, testing and rollout plans, risks and alternatives.
  * Updated site navigation to include "DPU Healthcheck" under Enhancement Proposals.

### DIFF
--- a/docs/okeps/okep-5674-dpu-healthcheck.md
+++ b/docs/okeps/okep-5674-dpu-healthcheck.md
@@ -1,0 +1,169 @@
+# OKEP-5674: DPU Healthcheck
+
+* Issue: [#5674](https://github.com/ovn-org/ovn-kubernetes/issues/5674)
+
+## Problem Statement
+
+With the DPU (trusted) architecture, we have ovnkube running on the host (DPU-Host mode), and ovnkube running in the
+DPU (DPU mode). Kubelet talks on the host to dpu-host mode, which annotates the VF rep information so that DPU ovnkube
+can plug the port into OVS and handle all the OVN configuration. The DPU component is managed outside the host
+kubernetes cluster.
+
+The purpose of this addition is to be able to detect when the DPU side goes down, and indicate that the CNI is no longer
+ready on the node.
+
+## Goals
+
+* Decrease CNI failure time by detecting when the DPU is unhealthy
+* Avoid Kubernetes scheduling new pods via to nodes where the DPU is unhealthy
+
+## Non-Goals
+
+* Tainting the node
+* Distinguishing between different types of failure in the DPU
+
+## Introduction
+
+As a refresher, the architecture of the DPU architecture in "trusted" mode is shown below:
+
+```mermaid
+flowchart TB
+
+  subgraph HOST[Host]
+    direction TB
+    H1["Kubelet"]
+    H2["ovn-kubernetes (DPU-Host mode)"]
+  end
+
+  subgraph DPU[DPU]
+    direction TB
+    D1["ovn-kubernetes (DPU mode)"]
+    D2[OVS]
+    D3[OVN]
+    D1 --> D2
+    D1 --> D3
+  end
+
+  HOST <--> DPU
+```
+
+Note, "trusted" mode here means the DPU side has the kubeconfig of the host cluster, and runs OVNK in the DPU. There
+is an "untrusted" mode, but it is outside the scope of this enhancement and will not use this feature.
+
+As can be seen in the diagram, the ovn-kubernetes DPU mode is running down in the DPU, managed outside the purview of
+the host Kubernetes cluster. Therefore, the Host Kubernetes cluster has no observability or failure detection of this pod.
+When a DPU-accelerated pod is created, the following happens:
+
+1. Kubelet invokes Multus which checks that the UDN/NAD has a resourceName field
+2. SR-IOV plugin is called and reserves a VF, passes the PCI address in the CNI ADD
+3. OVN-Kubernetes DPU-Host receives the CNI ADD, and annotates the pod with the connection details.
+4. OVN-Kubernetes DPU waits for the annotation, plugs the VF representor into OVS, and annotates the pod with typical
+k8s.ovn.org annotations.
+5. OVN-Kubernetes DPU-Host sees annotations, and then moves the VF into the pod and configures it
+6. DPU-Host sends CNI ADD reply
+
+Now, if the DPU side goes down, DPU-Host will keep waiting on the CNI ADD for 2 minutes. All CNI ADDs will keep failing.
+From a user's perspective on the Host Kubernetes cluster, they may have no insight into what is running in the DPU from
+KAPI. We can make this better by:
+
+1. Detecting when the DPU side has gone down.
+2. Failing CNI operations fast as we know when the DPU is down.
+3. Triggering Kubelet to report the node as not ready to avoid future pods from being scheduled to this node.
+
+## User-Stories/Use-Cases
+
+Story 1: Make DPU operations more robust
+
+As a cluster admin, I want to leverage DPUs for hardware offload. However, I do not want to compromise my
+visibility into the cluster and be stuck with components that are failing outside my view. I'd like to have DPU
+issues propagate up the stack so that I can see when a node has failed. Additionally, I would like Kubelet to no longer
+schedule pods to this node until it has come back online.
+
+## Proposed Solution
+
+Kubernetes uses node leases as a mechanism to maintain health within a cluster. This solution proposes that a custom
+node lease is created and used by OVN-Kubernetes to detect when a DPU is no longer functioning. The DPU OVN-Kubernetes
+component will update this node lease periodically. DPU-Host will monitor this node lease, and when it expires it will:
+
+1. Trigger CNI operations to fail fast, CNI is down.
+2. Implement the CNI STATUS verb to signal CNI issues to the container runtime.
+
+The container runtime such as CRI-O or containerd will call CNI STATUS. When CNI STATUS returns an error, the runtime
+will report to Kubelet that `NetworkReady=false`, and then Kubelet will set the Kubernetes node status as `NotReady`.
+Kubernetes will stop scheduling new pods to this node, and after a grace period (default 300 seconds), Kube Node Controller
+will taint the node, triggering eviction for pods that do not tolerate the taint.
+
+### API Details
+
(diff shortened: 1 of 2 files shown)


docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+
+1. Kubelet invokes Multus which checks that the UDN/NAD has a resourceName field
+2. SR-IOV plugin is called and reserves a VF, passes the PCI adddress in the CNI ADD
+3. OVN-Kubernetes DPU-Host receives the CNI ADD, and annotates the pod with the connection details.
+4. OVN-Kubernetes DPU waits for the annotation, plugs the VF representor into OVS, and annotates the pod with typical
+k8s.ovn.org annotations.
+5. OVN-Kubernetes DPU-Host sees annotations, and then moves the VF into the pod and configures it
+6. DPU-Host sends CNI ADD reply
+
+Now, if the DPU side goes down, DPU-Host will keep waiting on the CNI ADD for 2 minutes. All CNI ADDs will keep failing.]
```
dave-tucker: stray `]`

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+can plug the port into OVS and handle all the OVN configuration. The DPU component is managed outside the host
+kubernetes cluster.
+
+The purpose of this addition is to be able to detect when the DPU side goes down, and indicate that the CNI is no longer
+ready on the node.
+
+## Goals
+
+* Decrease CNI failure time by detecting when the DPU is unhealthy
+* Report NetworkUnavailable node condition so that Kubelet will stop scheduling pods to this node
```
dave-tucker: Actually, we don't report that at all. It's out of our hands - basically it's up to the CRI and Kubelet as I understand it.

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+schedule pods to this node until it has come back online.
+
+## Proposed Solution
+
+Kubernetes uses node leases as a mechanism to maintain health within a cluster. This solution proposes that a custom
+node lease is created and used by OVN-Kubernetes to detect when a DPU is no longer functioning. The DPU OVN-Kubernetes
+component will update this node lease periodically. DPU-Host will monitor this node lease, and when it expires it will:
+
+1. Trigger CNI operations to fail fast, CNI is down.
+2. Update the node condition with NetworkUnavailable.
```
dave-tucker:
```suggestion
2. Implement the CNI STATUS verb to signal CNI issues to the container runtime
```

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+
+Kubelet will see the NetworkUnavailable and stop scheduling pods to this node. This will not prevent OVN-Kubernetes pods
```
dave-tucker: AIUI the flow is as follows:

- CNI STATUS is called by CRI
- CNI STATUS returns an error
- CRI propagates that error back to Kubelet
- Kubelet will mark the node as `NotReady`

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+The default setting will be 10 seconds. Set this value to 0 to disable.
+In addition, another config knob "--dpu-node-lease-duration" will be exposed that takes a time value in seconds. It will
+be used as the max time before the DPU is considered dead. The default time will be 40 seconds.
+These values are similar to what Kubernetes uses for Kubelet node leases.
+
+### Implementation Details
+
+OVN-Kubernetes default node network controller (DNNC) will be updated to update the node lease when the mode is configured
+to be DPU mode, and the --dpu-node-lease-renew-interval is > 0. OVN-Kubernetes CNI server will be updated to track
+this lease, and when the DPU node lease expires it will trigger CNI to fail and update the node with a condition like:
```
dave-tucker: Per previous comment, we can only guarantee that CNI ADD and STATUS will fail with a decent error message. The propagation of that error to kubelet is not up to us.

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+    status: "False"
+    reason: DPUHealthy
+    message: "OVN-Kubernetes DPU has resumed lease updates and datapath connectivity"
+    lastHeartbeatTime: "2025-10-20T15:34:00Z"
+    lastTransitionTime: "2025-10-20T15:34:00Z"
+```
+
+### Testing Details
+
+* Unit tests will be added to make sure node conditions get updated as expected
```
dave-tucker: We're not changing node conditions. Unit tests should focus on CNI commands

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+
+### API Details
+
+A configuration knob will be introduced in OVN-Kubernetes to set the "--dpu-node-lease-renew-interval" time in seconds.
+The default setting will be 10 seconds. Set this value to 0 to disable.
+In addition, another config knob "--dpu-node-lease-duration" will be exposed that takes a time value in seconds. It will
+be used as the max time before the DPU is considered dead. The default time will be 40 seconds.
+These values are similar to what Kubernetes uses for Kubelet node leases.
+
+### Implementation Details
```
dave-tucker, line 104: I think we should cover something here on any RBAC changes required.
Do we currently write to the apiServer of the tenant k8s from ovn-k8s dpu-mode or any reason?
If we're introducing writes (that seems risky) what RBAC controls do we have to mitigate this?

There needs to be more implementation details on how the Node Lease mechanism is supposed to work too.
I know how this works for kubelet, but details as to which namespace these leases will live in are important.
For example, what's the lifecycle? When is the lease created, updated, and removed.

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+    reason: DPUHealthy
+    message: "OVN-Kubernetes DPU has resumed lease updates and datapath connectivity"
+    lastHeartbeatTime: "2025-10-20T15:34:00Z"
+    lastTransitionTime: "2025-10-20T15:34:00Z"
+```
+
+### Testing Details
+
+* Unit tests will be added to make sure node conditions get updated as expected
+* E2E tests will be added where there is no DPU mode, and DPU-Host mode is just launched. Checking that
```
dave-tucker, line 144: Not sure how best to test this E2E, as the actual error propagation is part of CRIO/containerd/kubelet and not up to us. Tests are likely to fail if we make assertions on the state of nodes since we're testing kubelet at that point not ovn-k8s.

docs/okeps/okep-5674-dpu-healthcheck.md
trozet: I went  a different route because:
"Status is purely informational. A plugin MUST NOT rely on STATUS being called."
"Plugins should always expect other CNI operations (like ADD, DEL, etc) even if STATUS returns an error. STATUS does not prevent other runtime requests."

https://www.cni.dev/docs/spec/#status-check-plugin-status

danwinship: When it says "A plugin MUST NOT rely on STATUS being called", it really means "this is all brand new stuff, and who knows how long it will take for runtimes to support it". If you know this feature will only be used in certain environments, and those environments use CRI implementations that implement STATUS in a way that you're happy with, then you're all set.

(The second part still applies: you don't know _when_ the runtime is going to call STATUS, so you still have to deal with returning errors correctly from other calls.)

docs/okeps/okep-5674-dpu-healthcheck.md
```diff
@@ -0,0 +1,159 @@
...
+6. DPU-Host sends CNI ADD reply
+
+Now, if the DPU side goes down, DPU-Host will keep waiting on the CNI ADD for 2 minutes. All CNI ADDs will keep failing.]
+From a user's perspective on the Host Kubernetes cluster, they may have no insight into what is running in the DPU from
+KAPI. We can make this better by:
+
+1. Detecting when the DPU side has gone down.
+2. Failing CNI operations fast as we know when the DPU is down.
+3. Set the node as NetworkUnavailable so that Kubelet will stop scheduling pods here.
+
```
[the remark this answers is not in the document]
trozet: will be fixed in next push

docs/okeps/okep-5674-dpu-healthcheck.md
trozet: I'll update this to be more generic and reinforce the goal of stopping kubernetes from scheduling pods to this node.

docs/okeps/okep-5674-dpu-healthcheck.md
trozet: Yes the dpu mode already writes to the kapi server. It needs to add the k8s.ovn.org annotations to the pod, and probably some other annotations to the node.

There needs to be RBAC added to allow ovn-kubernetes to CRUD the lease, and it should be namespace scoped to just the ovn-kubernetes namespace (least permissive). I'll add some details.

docs/okeps/okep-5674-dpu-healthcheck.md
trozet: well e2e is to test the entire system, so if we report error and kubernetes does not put the node in not ready state, that is a systemic failure and we should figure out the issue and file/fix the bug upstream.

trozet:
next push is just rebase

trozet:
Implementation: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5777

dave-tucker (dismissed):
Thanks @trozet for going through multiple revisions on this. Design LGTM 🎉

trozet:
just fixed rebase for mkdocs.yaml didn't change anything else

npinaeva (approved):
makes sense!

PR #5676, opened 2025-10-20
EIP: Change capacity storage to pointers

tssurya:
We need to differentiate between
the field not being set and field
being set to 0 which means no
available capacity on that node.

## 📑 Description

See https://github.com/openshift/cloud-network-config-controller/pull/183 for details.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node capacity handling in egress IP allocation to better support edge cases where capacity metrics are unavailable or incomplete, enhancing system stability and preventing potential allocation failures during node selection.


kyrtapz (approved):
LGTM

PR #5680, opened 2025-10-21
Reconcile neighbors after the vm live migration on default cluster network.

### LINKED ISSUE #3986: [FAIL] Kubevirt Virtual Machines when live migrated [It] with pre-copy should keep connectivity
Seeing failures with this test still:

```
Kubevirt Virtual Machines when live migrated [It] with pre-copy should keep connectivity
/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:656

  [FAILED] Oct 25 23:45:54.083: Timed out after 60.010s.
  worker1: after live migrate, migration #1: Check tcp connection is not broken
      <*errors.errorString | 0xc0003bb4a0>: 
      http connection to virtual machine was broken
      {
          s: "http connection to virtual machine was broken",
      }
  In [It] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:296 @ 10/25/23 23:45:54.084

  There were additional failures detected.  To view them in detail run ginkgo -vv
```
https://github.com/ovn-org/ovn-kubernetes/actions/runs/6646915605/job/18062050736?pr=3979

qinqon:
## Description

After VM live migration to a different node, neighbor cache entries (ARP/NDP)
become stale and must be updated to ensure proper traffic routing. This PR adds
neighbor reconciliation logic that sends GARPs and unsolicited NAs to update
ARP/NDP caches with the arp_proxy MAC address when VMs migrate between nodes
with different subnet ownership.

### Gateway IP neighbor fix

After PR#5773 changed DHCP to advertise the actual subnet gateway IP (e.g.,
10.244.0.1) instead of the link-local 169.254.1.1 address, VMs resolve the
gateway to the node's real LRP MAC on the subnet-owning switch (the LRP's ARP
responder has higher priority than arp_proxy). After migration, this LRP MAC
doesn't exist on the destination switch, so we must also GARP the gateway IP to
update the VM's neighbor entry to the arp_proxy MAC. This issue was masked in
e2e tests by PR#4755 which changed the connectivity test checks.

References:
- Bug introduced by: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5773
- Masked in tests by: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4755

### Reconciliation details

The reconciliation handles two scenarios:
- **At migration target (local zone):** notify the VM that subnet peers and the
  gateway IP should use the arp_proxy MAC via GARPs/unsolicited NAs.
- **At subnet-owning node (remote zone):** delete stale source pod LSPs to
  prevent wrong ARP responses, and notify remaining pods to use the arp_proxy
  MAC for the migrated VM's IP.

### E2E test refactor

This also refactors the kubevirt e2e tests related to default pod network
migration so they use iperf3 for north-south and east-west connectivity,
ensuring downtime measurement at east-west and removing the custom TCP server
that was previously used.

Fixes #3986

## Checks
- [x] My code requires tests

## How to verify it
It includes e2e tests

## Summary by CodeRabbit

* **Documentation**
  * Expanded live-migration guide with detailed neighbor advertising workflow, GARP handling, and multi-zone neighbor table updates.

* **New Features**
  * Enhanced live migration support with zone-aware pod routing and neighbor reconciliation for improved network consistency across migrations.

* **Tests**
  * Updated live migration tests to validate multiple migration modes and dynamic network configurations.

* **Chores**
  * Removed legacy test infrastructure components no longer in use.


qinqon:
Broke unit test with change at 
```
git diff 81db7b84327da943e3448f757db0539e61da6ed6..bccc0996072986fd320792894488fde359e4ffef
```

docs/features/live-migration.md
```diff
@@ -428,6 +429,80 @@ The point to point routing cleanup (remove of static routes and policies) will b
 - VM is live migrated back to the node that owns its IP, all the routing related to the VM is removed at all the ovn zones.
 - ovn-kubernetes controllers are restarted, stale routing is removed.
 
+**Advertising neighbors:**
+To improve the connectivity after live migration  ovn-k sends GARPs and unsolicited NAs to pods at node owning the subnet and
```
kyrtapz: This sentence is a little confusing to read, did you mean:
```
To improve connectivity after live migration, ovn-k sends GARPs and unsolicited NAs to pods on the node owning the subnet and to the VM on the migration target node.
```

docs/features/live-migration.md
```diff
@@ -428,6 +429,80 @@ The point to point routing cleanup (remove of static routes and policies) will b
...
+└─────────────────────────────────┘
+```
+
+Now after live migration from node1 to node2 the situation is the following:
+
+
+```text
+    ┌────────────────────────┐ ┌────────────────────────┐
+    │logical switch node2    │ │logic switch node1      │ 
+    │   (10.244.1.0/24)      │ │   (10.244.0.0/24)      │
```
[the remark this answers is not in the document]
kyrtapz: nit but makes sense. Maybe add a space too?

go-controller/pkg/kubevirt/router.go
```diff
@@ -166,33 +179,73 @@ func EnsureLocalZonePodAddressesToNodeRoute(watchFactory *factory.WatchFactory,
...
 }
 
-// EnsureRemoteZonePodAddressesToNodeRoute will add static routes when live
-// migrated pod belongs to remote zone to send traffic over transwitch switch
-// port of the node where the pod is running:
-//   - A dst-ip with live migrated pod ip as prefix and nexthop the pod's
-//     current node transit switch port.
-func EnsureRemoteZonePodAddressesToNodeRoute(watchFactory *factory.WatchFactory, nbClient libovsdbclient.Client, pod *corev1.Pod, nadName string) error {
+func reconcileClusterDefaultNetworkLocalZoneNeighbours(watchFactory *factory.WatchFactory, pod *corev1.Pod, podAnnotation *util.PodAnnotation, nadName string) error {
+	// TODO: Implement neighbours update for non IC
```
kyrtapz, line 197: Is this something you are planning to do? Is there any upstream issue tracking it that we can link?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -683,3 +711,63 @@ func newRouterAdvertisementWithPrefixInfos(ip net.IP, destinationMAC net.Hardwar
 		PrefixInfos:    prefixInfos,
 	}
 }
+func findRunningPodsIPsFromPodSubnet(watchFactory *factory.WatchFactory, podAnnotation *util.PodAnnotation, nadName string) ([]*net.IPNet, error) {
+	nodeOwningSubnet, err := findNodeOwningSubnet(watchFactory, podAnnotation, nadName)
+	if err != nil {
+		return nil, err
+	}
+	if nodeOwningSubnet == nil {
+		return nil, nil
```
kyrtapz: Shouldn't this be an error?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -683,3 +711,63 @@ func newRouterAdvertisementWithPrefixInfos(ip net.IP, destinationMAC net.Hardwar
...
+	}
+
+	pods, err := watchFactory.GetAllPods()
```
kyrtapz: Why are you looking through **ALL** pods, shouldn't it be possible to limit it to namespaces/networks?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -683,3 +711,63 @@ func newRouterAdvertisementWithPrefixInfos(ip net.IP, destinationMAC net.Hardwar
...
+		if err != nil {
+			klog.Errorf("Failed unmarshaling pod ovn for pod %s/%s annotations before GARP: %v", podToNotify.Namespace, podToNotify.Name, err)
+			continue
+		}
+
+		if util.PodCompleted(podToNotify) {
+			continue
+		}
+
+		//FIXME: Is comparing MAC enough to discard pod's from same VM
```
kyrtapz: Doesn't kubevirt store the VM identifier in a label or something?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -683,3 +711,63 @@ func newRouterAdvertisementWithPrefixInfos(ip net.IP, destinationMAC net.Hardwar
 		PrefixInfos:    prefixInfos,
 	}
 }
+func findRunningPodsIPsFromPodSubnet(watchFactory *factory.WatchFactory, podAnnotation *util.PodAnnotation, nadName string) ([]*net.IPNet, error) {
```
kyrtapz: I think this function could use a comment on what it does.

go-controller/pkg/kubevirt/router.go
```diff
@@ -166,33 +179,73 @@ func EnsureLocalZonePodAddressesToNodeRoute(watchFactory *factory.WatchFactory,
...
+	if err != nil {
+		return err
+	}
+
+	vmRunningAtCurrentNode := pod.Spec.NodeName == currentNode
+
+	// Now that the VM is running at a node not owning his subnet, using pod's mac
+	// will not work since it's a different switch so overwrite those neighbors with
+	// proxy mac from gatway
+	if !currentNodeOwnsSubnet && vmRunningAtCurrentNode {
```
kyrtapz: So on a node that does not own the subnet but has the VM migrated to we send GARPs for all IPs that are running on the original node?

go-controller/pkg/ovn/ovn.go
```diff
@@ -208,7 +208,7 @@ func (oc *DefaultNetworkController) ensureLocalZonePod(oldPod, pod *corev1.Pod,
 
 	if kubevirt.IsPodLiveMigratable(pod) {
 		v4Subnets, v6Subnets := util.GetClusterSubnetsWithHostPrefix()
-		return kubevirt.EnsureLocalZonePodAddressesToNodeRoute(oc.watchFactory, oc.nbClient, oc.lsManager, pod, ovntypes.DefaultNetworkName, append(v4Subnets, v6Subnets...))
+		return kubevirt.ReconcileClusterDefaultNetworkLocalZone(oc.watchFactory, oc.nbClient, oc.lsManager, pod, ovntypes.DefaultNetworkName, append(v4Subnets, v6Subnets...))
```
kyrtapz: What is the point of passing the nad name to a function that is default network specific?

go-controller/pkg/kubevirt/switch.go
```diff
@@ -53,3 +66,58 @@ func ComposeARPProxyLSPOption() string {
...
+		}
+		lsp := &nbdb.LogicalSwitchPort{
+			Name: util.GetLogicalPortName(vmPod.Namespace, vmPod.Name),
+		}
+		lsp, err := libovsdbops.GetLogicalSwitchPort(nbClient, lsp)
+		if err != nil {
+			continue
+		}
+		logicalSwitch := &nbdb.LogicalSwitch{
+			Name: vmPod.Spec.NodeName,
```
kyrtapz, line 147: What if the "related pod" is not running on the same node?

go-controller/pkg/kubevirt/router.go
```diff
@@ -243,29 +296,38 @@ func EnsureRemoteZonePodAddressesToNodeRoute(watchFactory *factory.WatchFactory,
...
-	// that target VM pod is ready to receive traffic so we can route
-	// taffic to it.
-	targetReadyTimestamp := pod.Annotations[kubevirtv1.MigrationTargetReadyTimestamp]
+	currentNode := os.Getenv("K8S_NODE")
+	currentNodeOwnsSubnet, err := nodeContainsPodSubnet(watchFactory, currentNode, podAnnotation, nadName)
+	if err != nil {
+		return err
+	}
+	if !vmRunningAtNodeOwningSubnet && currentNodeOwnsSubnet {
+		// If we don't do this ARPs will be answer again with wrong mac
```
kyrtapz: Why does that happen? Shouldn't the old pod get removed?

kyrtapz (commented):
Thanks @qinqon! I have some comments.
Just for my knowledge, live-migration is only supported for l2 and default networks is it?

qinqon:
> Thanks @qinqon! I have some comments. Just for my knowledge, live-migration is only supported for l2 and default networks is it?

@kyrtapz yes, we just support default pod network for special annotated VMs (the ones used at hypershift), and normal layer2.

go-controller/pkg/kubevirt/pod.go
qinqon: For example if we migrate a VM, and then that node goes down this will be the case and since we don't have nothing to GARP for we just return empty here.

go-controller/pkg/kubevirt/pod.go
qinqon: Right, we should filter pods from node and using the default pod network, let me check if there are some infra to do this kind of filtering.

qinqon: @kyrtapz there is no label that denotate the node so we cannot filter the the pods by node from the informer also there is no label to denotate default pod network so we cannot filter either by that.

go-controller/pkg/kubevirt/pod.go
qinqon: I will pass the pod and check the virtual machine name instead.

go-controller/pkg/kubevirt/router.go
qinqon: We are planning to not support this for non IC, I don't think there are any live migration for default pod network non IC users.

go-controller/pkg/kubevirt/router.go
qinqon: Right, since VM is at node that need to do routing to access IPs at the VM's subnet we need to update VM's neighbor cache so those IPs access over the gateway since they are not part of the node's switch.

If we don't do so, the traffic will be down until the ip neighors at VM invalidates and send a normal ARP that will be responded with the proxy ARP and traffic will work, but that can take at to 10 seconds or so from what I have see, and depends on guest VM network configuration.

go-controller/pkg/kubevirt/router.go
qinqon: There is some time window between libvirt domain ready and virt-launcher pod completed, during that time the original LSP is still up.

go-controller/pkg/kubevirt/switch.go
qinqon: What do you mean ? we are removing the LSP from running pods that are not the live migration target pods, that's the whole purpose of this.

go-controller/pkg/ovn/ovn.go
qinqon: Agree, will fix that, I doubt we will ever implement this for UDN layer3.

qinqon:
@kyrtapz PR rebased, ready for review

go-controller/pkg/kubevirt/pod.go
```diff
@@ -237,7 +238,30 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+		for _, subnet := range podAnnotation.IPs {
+			for _, nodeHostSubNet := range nodeHostSubNets {
+				if nodeHostSubNet.Contains(subnet.IP) {
+					return node, nil
+				}
+			}
+		}
+	}
+	return nil, nil
+}
```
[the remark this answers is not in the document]
qinqon: Make sense I will change it.

go-controller/pkg/kubevirt/router.go
kyrtapz: I guess there is no need for the TODO then? Maybe it would be better to just clearly state it in the function description(after adding it 😅)?

go-controller/pkg/kubevirt/router.go
```diff
@@ -243,29 +296,38 @@ func EnsureRemoteZonePodAddressesToNodeRoute(watchFactory *factory.WatchFactory,
...
-	// When a virtual machine start up this
-	// label is the signal from KubeVirt to notify that the VM is
-	// ready to receive traffic.
-	targetNode := pod.Labels[kubevirtv1.NodeNameLabel]
-
-	// This annotation only appears on live migration scenarios and it signals
-	// that target VM pod is ready to receive traffic so we can route
-	// taffic to it.
-	targetReadyTimestamp := pod.Annotations[kubevirtv1.MigrationTargetReadyTimestamp]
+	currentNode := os.Getenv("K8S_NODE")
```
kyrtapz: This seems to be the first time we introduce direct dependency on `K8S_NODE` in the code, it means with this change all deployment types need to ensure it is set.
I think an alternative might be to depend on the `zone` name but I don't like it all that much either so I am ok to keep this one as is. 
Let's just check it is set, for sanity.

go-controller/pkg/kubevirt/switch.go
```diff
@@ -53,3 +66,58 @@ func ComposeARPProxyLSPOption() string {
...
+	for _, vmPod := range vmPods {
+		if util.PodCompleted(vmPod) {
+			continue
+		}
+		lsp := &nbdb.LogicalSwitchPort{
+			Name: util.GetLogicalPortName(vmPod.Namespace, vmPod.Name),
+		}
+		lsp, err := libovsdbops.GetLogicalSwitchPort(nbClient, lsp)
+		if err != nil {
+			continue
```
kyrtapz, line 144: NIT: Let's log the error?

kyrtapz (commented):
Looks good to me, ready for a second pair of eyes. 
One concern is that this will be quite heavy when there is a lot of pods and VMs spread out across the cluster, I don't see a better solution though. 
Few minor comments.

qinqon:
> Looks good to me, ready for a second pair of eyes. One concern is that this will be quite heavy when there is a lot of pods and VMs spread out across the cluster, I don't see a better solution though. Few minor comments.

Although this is true, the scenario is very niche, since this is done for default pod network VMs with the migration annotation, so more or less hypershift.

I don't know if implementing some kind of parallelization would help here.

qinqon:
I introduced a bug, here it is the fix https://github.com/ovn-kubernetes/ovn-kubernetes/compare/04a58fba9d7ce372ca7fa78ed3f748dff869b99d..22e61d412b05922f17a2b3df46e114b8ae552f92

@kyrtapz @npinaeva ^

docs/features/live-migration.md
```diff
@@ -428,6 +429,80 @@ The point to point routing cleanup (remove of static routes and policies) will b
...
+To improve the situation ovn-k sends the following GARPs and unsolicited NA taking
+into account that the arp proxy mac is `0a:58:0a:f3:00:00`:
+
+At node1 broadcast vm's IP should use arp proxy mac
+- `GARP(10.244.0.8 -> 0a:58:0a:f3:00:00, broadcast mac)`
+- one per migrated VM
+
+At node2 advertise to VM that pod's from the same subnet should go over proxy_arp:
+- `GARP(10.244.0.9 -> 0a:58:0a:f3:00:00, vm mac(0a:58:0a:f4:00:01))`
+- one per pod on the same subnet (in this example, only for pod1).
```
npinaeva: So now after the migration is done, node2 may have some other pods (pod2 10.244.1.2), which will populate the new neighbour entry for a vm. Now that vm migrates again (to a new node3 or back to node1) how do we make sure the neighbours tables for pod2 are updated?

qinqon: > So now after the migration is done, node2 may have some other pods (pod2 10.244.1.2), which will populate the new neighbour entry for a vm. Now that vm migrates again (to a new node3 or back to node1) how do we make sure the neighbours tables for pod2 are updated?

Since pod2 is not at the VM's subnet, it will always go from the VM over the default gw, so there is no problem there, the main issue is for pods on the same subnet.

qinqon:
/retest 

Not related
```
Summarizing 1 Failure:
  [FAIL] e2e egress IP validation Cluster Default Network [It] [secondary-host-eip] should send address advertisements for EgressIP [Feature:EgressIP]
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/egressip.go:3064
```

qinqon:
Something got merged and east/west iperf connectins get broken at default pod network during live migration

qinqon:
> Something got merged and east/west iperf connectins get broken at default pod network during live migration

@npinaeva so the issue appears after the following PR got merged
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5773

And also looks like the test here were not really checking if we break the tcp connection after the following got merged long long time ago:
- https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4755

So we got the perfect storm, will ping you when PR is ready.

qinqon:
@coderabbitai review

PR #5683, opened 2025-10-23
Enhance ipamclaim status update support

RamLavi:
## 📑 Description
This PR adds ipamClaim status support for conditions and ownerPod field.
The conditions are generally introduced on this [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/network/requesting-staticips-for-vms-being-migrated-into-primary-l2-udns-using-MTV.md#ipamclaim-crd).

*Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
_If PR is about `failing-tests or flakes`, please post the related issues/tests in a comment and do not use `Fixes`_*
-->
Fixes #[CORENET-6132](https://issues.redhat.com/browse/CORENET-6132)

## ✅ Checks

- [ ] My code requires tests

### script to manually check successful allocation and conflict allocation:
```
cat <<EOF | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: default-network-annotation
  labels:
    k8s.ovn.org/primary-user-defined-network: ""
    e2e-framework: default-network-annotation
EOF

cat <<EOF | kubectl apply -f -
apiVersion: k8s.ovn.org/v1
kind: UserDefinedNetwork
metadata:
  name: l2network
  namespace: default-network-annotation
spec:
  topology: Layer2
  layer2:
    role: Primary
    subnets:
      - 103.0.0.0/16
      - 2014:100:200::0/60
    ipam:
      mode: Enabled
      lifecycle: Persistent
EOF

# should succeed
kubectl apply -f - <<EOF
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: worker1
  namespace: default-network-annotation
spec:
  running: true
  template:
    metadata:
      annotations:
        network.kubevirt.io/addresses: "{\"overlay\": [\"103.0.0.3\", \"2014:100:200::20\"]}"
    spec:
      architecture: amd64
      domain:
        devices:
          disks:
          - disk:
              bus: virtio
            name: containerdisk
          - disk:
              bus:
              bus: virtio
            name: cloudinitdisk
          interfaces:
          - name: overlay
            binding:
              name: l2bridge
          rng: {}
        machine:
          type: q35
        resources:
          requests:
            memory: 512Mi
      networks:
      - name: overlay
        pod: {}
      terminationGracePeriodSeconds: 5
      volumes:
      - containerDisk:
          image: quay.io/kubevirt/fedora-with-test-tooling-container-disk:20241009_c8d5e5cc45
        name: containerdisk
      - cloudInitNoCloud:
          userData: |
            #cloud-config
            hostname: worker1
            manage_etc_hosts: true
            users:
              - name: core
                plain_text_passwd: core
                lock_passwd: false
                sudo: ALL=(ALL) NOPASSWD:ALL
        name: cloudinitdisk
EOF

kubectl wait --for=condition=AgentConnected vmi -n default-network-annotation worker1 --timeout 5m

# should fail
kubectl apply -f - <<EOF
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  name: worker2
  namespace: default-network-annotation
spec:
  running: true
  template:
    metadata:
      annotations:
        network.kubevirt.io/addresses: "{\"overlay\": [\"103.0.0.3\", \"2014:100:200::30\"]}"
    spec:
      architecture: amd64
      domain:
        devices:
          disks:
          - disk:
              bus: virtio
            name: containerdisk
          - disk:
              bus:
              bus: virtio
            name: cloudinitdisk
          interfaces:
          - name: overlay
            binding:
              name: l2bridge
          rng: {}
        machine:
          type: q35
        resources:
          requests:
            memory: 512Mi
      networks:
      - name: overlay
        pod: {}
      terminationGracePeriodSeconds: 5
      volumes:
      - containerDisk:
          image: quay.io/kubevirt/fedora-with-test-tooling-container-disk:20241009_c8d5e5cc45
        name: containerdisk
      - cloudInitNoCloud:
          userData: |
            #cloud-config
            hostname: worker1
            manage_etc_hosts: true
            users:
              - name: core
                plain_text_passwd: core
                lock_passwd: false
                sudo: ALL=(ALL) NOPASSWD:ALL
        name: cloudinitdisk
EOF
```
expected results:
```
# kubectl get ipamclaim -n default-network-annotation worker1.overlay -ojson | jq .status
{
  "conditions": [
    {
      "lastTransitionTime": "2025-10-23T08:32:14Z",
      "message": "IP addresses successfully allocated",
      "observedGeneration": 1,
      "reason": "SuccessfulAllocation",
      "status": "True",
      "type": "IPsAllocated"
    }
  ],
  "ips": [
    "103.0.0.3/16",
    "2014:100:200::20/60"
  ],
  "ownerPod": {
    "name": "virt-launcher-worker1-kd72w"
  }
}

# kubectl get ipamclaim -n default-network-annotation worker2.overlay -ojson | jq .status
{
  "conditions": [
    {
      "lastTransitionTime": "2025-10-23T08:33:55Z",
      "message": "failed to update pod default-network-annotation/virt-launcher-worker2-lqddr: failed to ensure requested or annotated IPs [103.0.0.3/16 2014:100:200::30/60] for default-network-annotation/l2network/default-network-annotation/virt-launcher-worker2-lqddr: provided IP is already allocated",
      "observedGeneration": 1,
      "reason": "IPAddressConflict",
      "status": "False",
      "type": "IPsAllocated"
    }
  ],
  "ips": [],
  "ownerPod": {
    "name": "virt-launcher-worker2-lqddr"
  }
}

```

## Summary by CodeRabbit

* **New Features**
  * Added a helper to detect full-allocation errors and a public API to update IPAM claim status based on allocation outcomes.

* **Bug Fixes**
  * Claim reconciliation now runs at operation completion; reconciliation errors are merged into operation results and claim status is set consistently for success, subnet exhaustion, and IP/MAC conflicts.

* **Tests**
  * Expanded unit and e2e tests with event-based checks for duplicate IPs, MAC conflicts, and subnet exhaustion.

### DIFF
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -32,6 +32,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -2408,7 +2409,60 @@ chpasswd: { expire: False }
 		)
 	})
 
+	getIPAMClaimName := func(vmName, netName string) string {
+		return fmt.Sprintf("%s.%s", vmName, netName)
+	}
+
+	verifyIPAMClaimStatusSuccess := func(ipamClaimName string) {
+		Eventually(func(g Gomega) {
+			ipamClaim := &ipamclaimsv1alpha1.IPAMClaim{}
+			g.Expect(crClient.Get(context.Background(), crclient.ObjectKey{
+				Namespace: namespace,
+				Name:      ipamClaimName,
+			}, ipamClaim)).To(Succeed(), "Should get IPAMClaim")
+
+			g.Expect(ipamClaim.Status.OwnerPod).NotTo(BeNil(), "OwnerPod should be set")
+			g.Expect(ipamClaim.Status.OwnerPod.Name).To(HavePrefix("virt-launcher-"), "OwnerPod should be the virt-launcher pod")
+
+			g.Expect(ipamClaim.Status.Conditions).NotTo(BeEmpty(), "Conditions should be set")
+			condition := meta.FindStatusCondition(ipamClaim.Status.Conditions, "IPsAllocated")
+			g.Expect(condition).NotTo(BeNil(), "IPsAllocated condition should exist")
+			g.Expect(condition.Status).To(Equal(metav1.ConditionTrue), "Condition status should be True")
+			g.Expect(condition.Reason).To(Equal("SuccessfulAllocation"), "Condition reason should be SuccessfulAllocation")
+
+			g.Expect(ipamClaim.Status.IPs).NotTo(BeEmpty(), "IPs should be set on successful allocation")
+		}).
+			WithTimeout(30*time.Second).
+			WithPolling(2*time.Second).
+			Should(Succeed(), fmt.Sprintf("IPAMClaim %s should have expected successful status", ipamClaimName))
+	}
+
+	verifyIPAMClaimStatusFailure := func(ipamClaimName string, expectedReason string) {
+		Eventually(func(g Gomega) {
+			ipamClaim := &ipamclaimsv1alpha1.IPAMClaim{}
+			g.Expect(crClient.Get(context.Background(), crclient.ObjectKey{
+				Namespace: namespace,
+				Name:      ipamClaimName,
+			}, ipamClaim)).To(Succeed(), "Should get IPAMClaim")
+
+			g.Expect(ipamClaim.Status.OwnerPod).NotTo(BeNil(), "OwnerPod should be set")
+			g.Expect(ipamClaim.Status.OwnerPod.Name).To(HavePrefix("virt-launcher-"), "OwnerPod should be the virt-launcher pod")
+
+			g.Expect(ipamClaim.Status.Conditions).NotTo(BeEmpty(), "Conditions should be set")
+			condition := meta.FindStatusCondition(ipamClaim.Status.Conditions, "IPsAllocated")
+			g.Expect(condition).NotTo(BeNil(), "IPsAllocated condition should exist")
+			g.Expect(condition.Status).To(Equal(metav1.ConditionFalse), "Condition status should be False")
+			g.Expect(condition.Reason).To(Equal(expectedReason), "Condition reason should match")
+
+			g.Expect(ipamClaim.Status.IPs).To(BeEmpty(), "IPs should not be set on failed allocation")
+		}).
+			WithTimeout(30*time.Second).
+			WithPolling(2*time.Second).
+			Should(Succeed(), fmt.Sprintf("IPAMClaim %s should have expected failure status", ipamClaimName))
+	}
+
 	Context("duplicate addresses validation", func() {
+		const networkName = "net1"
 		var (
 			cudn          *udnv1.ClusterUserDefinedNetwork
 			duplicateIPv4 = "10.128.0.200" // Static IP that will be used by both VMs
@@ -2432,7 +2486,7 @@ chpasswd: { expire: False }
 			namespace = fr.Namespace.Name
 
 			dualCIDRs := filterDualStackCIDRs(fr.ClientSet, []udnv1.CIDR{udnv1.CIDR(cidrIPv4), udnv1.CIDR(cidrIPv6)})
-			cudn, _ = kubevirt.GenerateCUDN(namespace, "net1", udnv1.NetworkTopologyLayer2, udnv1.NetworkRolePrimary, dualCIDRs)
+			cudn, _ = kubevirt.GenerateCUDN(namespace, networkName, udnv1.NetworkTopologyLayer2, udnv1.NetworkRolePrimary, dualCIDRs)
 			createCUDN(cudn)
 		})
 
@@ -2459,12 +2513,18 @@ chpasswd: { expire: False }
 			createVirtualMachine(vm1)
 			waitForVMReadinessAndVerifyIPs(vm1.Name, staticIPs)
 
+			By("Verifying first VM IPAMClaim has successful status")
+			verifyIPAMClaimStatusSuccess(getIPAMClaimName(vm1.Name, networkName))
+
 			By("Creating second VM with duplicate static IP - should fail")
 			vm2 := createVMWithStaticIP("test-vm-2", staticIPs)
 			createVirtualMachine(vm2)
 
 			By("Verifying pod fails with duplicate IP allocation error")
 			waitForVMPodErrorEvent(vm2.Name, "provided IP is already allocated")
+
+			By("Verifying second VM IPAMClaim has failure status with IPAddressConflict")
+			verifyIPAMClaimStatusFailure(getIPAMClaimName(vm2.Name, networkName), "IPAddressConflict")
 
 			By("Verifying first VM is still running normally")
 			waitForVMReadinessAndVerifyIPs(vm1.Name, staticIPs)
@@ -2495,6 +2555,9 @@ chpasswd: { expire: False }
 			Expect(crClient.Get(context.Background(), crclient.ObjectKeyFromObject(vm1), vmi1)).To(Succeed())
 			Expect(vmi1.Status.Interfaces[0].MAC).To(Equal(testMAC), "vmi status should report the requested mac")
 
+			By("Verifying first VM IPAMClaim has successful status")
+			verifyIPAMClaimStatusSuccess(getIPAMClaimName(vm1.Name, networkName))
+
 			By("Create second VM requesting the same MAC address")
 			vmi2 := newVMIWithPrimaryIfaceMAC(testMAC)
 			vm2 := generateVM(vmi2)
@@ -2514,6 +2577,9 @@ chpasswd: { expire: False }
 				HaveField("Reason", "ErrorAllocatingPod"),
 				HaveField("Message", ContainSubstring("MAC address already in use")),
 			)))
+
+			By("Verifying second VM IPAMClaim has failure status with MACAddressConflict")
+			verifyIPAMClaimStatusFailure(getIPAMClaimName(vm2.Name, networkName), "MACAddressConflict")
 
 			By("Assert second VM not running")
 			Expect(crClient.Get(context.Background(), crclient.ObjectKeyFromObject(vm2), vmi2)).To(Succeed())
@@ -2640,6 +2706,64 @@ chpasswd: { expire: False }
 			// Verify it got the requested IPs
 			actualAddresses := virtualMachineAddressesFromStatus(vmi, len(staticIPs))
 			Expect(actualAddresses).To(ConsistOf(staticIPs), "VM should get the requested static IPs")
+		})
+	})
+
+	Context("ipv4 subnet exhaustion", func() {
+		const networkName = "net1"
+		var (
+			cudn     *udnv1.ClusterUserDefinedNetwork
+			cidrIPv4 = "10.130.0.0/30" // subnet with no usable IPs
+			cidrIPv6 = "2011:100:200::0/120"
+		)
+
+		BeforeEach(func() {
+			l := map[string]string{
+				"e2e-framework":           fr.BaseName,
+				RequiredUDNNamespaceLabel: "",
+			}
+			ns, err := fr.CreateNamespace(context.Background(), fr.BaseName, l)
+			Expect(err).NotTo(HaveOccurred())
+			fr.Namespace = ns
+			namespace = fr.Namespace.Name
+
+			dualCIDRs := filterDualStackCIDRs(fr.ClientSet, []udnv1.CIDR{udnv1.CIDR(cidrIPv4), udnv1.CIDR(cidrIPv6)})
+			cudn, _ = kubevirt.GenerateCUDN(namespace, networkName, udnv1.NetworkTopologyLayer2, udnv1.NetworkRolePrimary, dualCIDRs)
+			createCUDN(cudn)
+		})
+
+		It("should fail when subnet is exhausted", func() {
+			By("Creating VM that should fail due to subnet exhaustion")
+			exhaustedVMName := "exhausted-vm"
+			vm := fedoraWithTestToolingVM(
+				nil, // labels
+				nil, // no static IP annotations
+				nil, // nodeSelector
+				kubevirtv1.NetworkSource{
+					Pod: &kubevirtv1.PodNetwork{},
+				},
+				`#cloud-config
+password: fedora
+chpasswd: { expire: False }
+`,
+				`version: 2
+ethernets:
+  eth0:
+    dhcp4: true
+    dhcp6: true
+    ipv6-address-generation: eui64`,
+			)
+			vm.Name = exhaustedVMName
+			vm.Namespace = namespace
+			vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].Bridge = nil
+			vm.Spec.Template.Spec.Domain.Devices.Interfaces[0].Binding = &kubevirtv1.PluginBinding{Name: "l2bridge"}
+			createVirtualMachine(vm)
+
+			By("Verifying pod fails with subnet exhaustion error")
+			waitForVMPodErrorEvent(exhaustedVMName, "subnet address pool exhausted")
+
+			By("Verifying VM IPAMClaim has failure status with SubnetExhausted")
+			verifyIPAMClaimStatusFailure(getIPAMClaimName(exhaustedVMName, networkName), "SubnetExhausted")
 		})
 	})
 })


RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/187e42daed75eb64122c1e9ad5ac068e622d30fe..d9cfeac156020847569bece1db60e039602f24a4): Rebase ( no conflicts)

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/d9cfeac156020847569bece1db60e039602f24a4..92fc3a403ad8c6c47fa9b58fedc2924aa733b33e): fix lint issues

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/5234499fa8b9ea3c553134085a1e283a54c392b7..6e53e12f0ea2036e90de5309aa61701654b4ec80): Rebase (no conflicts)

Moved to Ready in order to see the e2e results

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/6e53e12f0ea2036e90de5309aa61701654b4ec80..911ae7eba5f4ae3d93c500ad0da2b6cab7d085ec):
- Add a DNM commit pinning to https://github.com/k8snetworkplumbingwg/ipamclaims/pull/11 in order to show that ipamclaim CRD fix is indeed working
- Add successful allocation, ip conflict, and ip exhausted ipamclaim condition tests

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -244,10 +252,16 @@ func allocatePodAnnotationWithTunnelID(
...
-	if ipamClaim != nil && err == nil {
-		newIPAMClaim := ipamClaim.DeepCopy()
-		newIPAMClaim.Status.IPs = util.StringSlice(podAnnotation.IPs)
-		err = claimsReconciler.Reconcile(ipamClaim, newIPAMClaim, ipAllocator)
+	if ipamClaim != nil {
+		updatedIPAMClaim := updateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
+		allocationErr := err
+		reconcileErr := claimsReconciler.Reconcile(ipamClaim, updatedIPAMClaim, ipAllocator)
+		// Reconcile error takes precedence
+		if reconcileErr != nil {
```
RamLavi: perhaps we only want to log in case of error? would appreciate opinions here..

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/911ae7eba5f4ae3d93c500ad0da2b6cab7d085ec..48c2d59a8ee255ea0d379a2ce67a1f4b27a2ab9f): Rebase (no conflicts)

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/48c2d59a8ee255ea0d379a2ce67a1f4b27a2ab9f..eba8dd4f5a6a1923de8c850d9fb5409c6addb99f): Removed the DNM commit about ipamClaim CRD change. 
Conditions of "error IP allocation" will be added with empty `IPs` slice

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -225,13 +251,116 @@ func allocatePodAnnotationWithTunnelID(
...
+
+	if hasIPAMClaim && !hasPersistentIPs {
+		klog.Errorf(
+			"Pod %s/%s referencing an IPAMClaim on network %q which does not honor it",
+			pod.GetNamespace(),
+			pod.GetName(),
+			netInfo.GetNetworkName(),
+		)
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
maiqueb: this seems to make sense, but I think we wanted to allow the network to be reconcilable. 

Not sure where we stand today, given UDNs are immutable (but NADs are not); I recommend interviewing a maintainer (publicly, on the ovn-k slack) about this.

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -1086,7 +1055,8 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
 			if tt.netInfo == nil {
 				tt.netInfo = &util.DefaultNetInfo{}
 				tt.nadName = types.DefaultNetworkName
-				if !tt.ipam || tt.idAllocation || tt.persistentIPAllocation || tt.args.ipamClaim != nil {
+				shouldHaveIPAMClaimRef := tt.args.network != nil && tt.args.network.IPAMClaimReference != ""
+				if !tt.ipam || tt.idAllocation || tt.persistentIPAllocation || shouldHaveIPAMClaimRef {
```
maiqueb: would you take the opportunity of moving this to a properly named helper ? 

Would simplify the reader's life.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -146,10 +148,16 @@ func allocatePodAnnotation(
 		allocateToPodWithRollback,
 	)
 
-	if ipamClaim != nil && err == nil {
-		newIPAMClaim := ipamClaim.DeepCopy()
-		newIPAMClaim.Status.IPs = util.StringSlice(podAnnotation.IPs)
-		err = claimsReconciler.Reconcile(ipamClaim, newIPAMClaim, ipAllocator)
+	if ipamClaim != nil {
+		updatedIPAMClaim := updateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
+		allocationErr := err
```
maiqueb: IMHO passing the err down the stack is a mistake. You can check in the caller if it was successful, and invoke the appropriate handler method.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -290,6 +304,64 @@ func fetchIPAMClaim(
 	return nil, nil
 }
 
+// updateIPAMClaimStatus updates the IPAM claim status with conditions and IPs.
+// Returns an updated copy of the claim with:
+// - Conditions set based on allocation success/failure
+// - IPs set only on successful allocation
+// - Owner pod set on successful allocation
+func updateIPAMClaimStatus(
```
maiqueb: you have one big function that does IMHO a bunch of things, including figuring out if this was a success or failure from the existing error. 

I don't think it should be like this; I would expect you to check that in the caller, and then invoke the proper method - one for success, the other for failure.

go-controller/pkg/allocator/pod/pod_annotation.go
maiqueb: do we want to crash loop the pod ? if not, only log. 

Is this reconcilable ? I think so - i.e. the workload that previously had the "reserved" IP was removed.

test/e2e/kubevirt.go
```diff
@@ -2521,4 +2563,62 @@ ethernets:
...
+				`#cloud-config
+password: fedora
+chpasswd: { expire: False }
+`,
+				`version: 2
+ethernets:
+  eth0:
+    dhcp4: true
+    dhcp6: true
+    ipv6-address-generation: eui64`,
```
maiqueb, line 2754: isn't this the default ?

test/e2e/kubevirt.go
```diff
@@ -2351,7 +2352,68 @@ chpasswd: { expire: False }
 		)
 	})
 
+	getIPAMClaimName := func(vmName, netName string) string {
+		return fmt.Sprintf("%s.%s", vmName, netName)
+	}
+
+	verifyIPAMClaimStatus := func(ipamClaimName string, expectedConditionStatus metav1.ConditionStatus, expectedReason string) {
```
maiqueb: how about we have two of these ? one for expected successes, another for expected failures ?

You can move the common part to a function which you'd call from the functions I've referred above.

test/e2e/kubevirt.go
```diff
@@ -2351,7 +2352,68 @@ chpasswd: { expire: False }
...
+
+		Eventually(func(g Gomega) {
+			ipamClaim := &ipamclaimsv1alpha1.IPAMClaim{}
+			err := crClient.Get(context.TODO(), crclient.ObjectKey{
+				Namespace: namespace,
+				Name:      ipamClaimName,
+			}, ipamClaim)
+			g.Expect(err).NotTo(HaveOccurred(), "Should get IPAMClaim")
```
maiqueb: nit: maybe inline these ? are you re-using the `err` for something else ?

test/e2e/kubevirt.go
```diff
@@ -2351,7 +2352,68 @@ chpasswd: { expire: False }
...
+
+			g.Expect(ipamClaim.Status.OwnerPod).NotTo(BeNil(), "OwnerPod should be set")
+			g.Expect(ipamClaim.Status.OwnerPod.Name).To(HavePrefix("virt-launcher-"), "OwnerPod should be the virt-launcher pod")
+
+			g.Expect(ipamClaim.Status.Conditions).NotTo(BeEmpty(), "Conditions should be set")
+			condition := meta.FindStatusCondition(ipamClaim.Status.Conditions, "IPsAllocated")
```
maiqueb, line 2452: remind me how this works ?... so we'll find IPsAllocated always ?... Does this status condition make sense if we're requesting a dedicated address ?

test/e2e/kubevirt.go
```diff
@@ -2351,7 +2352,68 @@ chpasswd: { expire: False }
...
+			g.Expect(condition).NotTo(BeNil(), "IPsAllocated condition should exist")
+			g.Expect(condition.Status).To(Equal(expectedConditionStatus), "Condition status should match")
+			g.Expect(condition.Reason).To(Equal(expectedReason), "Condition reason should match")
+
+			if expectedConditionStatus == metav1.ConditionTrue {
+				g.Expect(ipamClaim.Status.IPs).NotTo(BeEmpty(), "IPs should be set on successful allocation")
+			} else {
+				g.Expect(ipamClaim.Status.IPs).To(BeEmpty(), "IPs should not be set on failed allocation")
+			}
```
maiqueb: imho we should ditch this if else.

test/e2e/kubevirt.go
```diff
@@ -2351,7 +2352,68 @@ chpasswd: { expire: False }
...
 
+	getIPAMClaimName := func(vmName, netName string) string {
+		return fmt.Sprintf("%s.%s", vmName, netName)
+	}
+
+	verifyIPAMClaimStatus := func(ipamClaimName string, expectedConditionStatus metav1.ConditionStatus, expectedReason string) {
+
+		Eventually(func(g Gomega) {
+			ipamClaim := &ipamclaimsv1alpha1.IPAMClaim{}
+			err := crClient.Get(context.TODO(), crclient.ObjectKey{
```
maiqueb: nit: I'd use `Background` instead of `TODO`. Or do you really plan on updating this in the future ?

test/e2e/kubevirt.go
```diff
@@ -2351,7 +2352,68 @@ chpasswd: { expire: False }
...
+			}
+		}).
+			WithTimeout(30*time.Second).
+			WithPolling(2*time.Second).
+			Should(Succeed(), fmt.Sprintf("IPAMClaim %s should have correct status", ipamClaimName))
+	}
+
+	waitForVMIPodEventFailure := func(vmName, expectedMessageSubstring, errorDescription string) {
+		Eventually(func() []corev1.Event {
+			podList, err := fr.ClientSet.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
```
maiqueb: ditto; s/TODO/Background/g

test/e2e/kubevirt.go
```diff
@@ -2521,4 +2563,62 @@ ethernets:
 			)), "second VM should not be ready due to MAC conflict")
 		})
 	})
+
+	Context("subnet exhaustion", func() {
+		const networkName = "net1"
+		var (
+			cudn     *udnv1.ClusterUserDefinedNetwork
+			cidrIPv4 = "10.130.0.0/30" // no usable IPs (.1 and .2, .0 is network infra owned, .3 is broadcast)
```
maiqueb: nit

IMHO you should also add that .1 and .2 are not usable; skimming through this, I thought that .1 and .2 were available, but .0 and .3 were not. 

So I would either:
- simplify the comment - just say "no usable IPs"
- be more specific in the comment (specify why none of the 4 IPs in that subnet is not usable).

Your call.

test/e2e/kubevirt.go
```diff
@@ -2521,4 +2563,62 @@ ethernets:
...
+				},
+				`#cloud-config
+password: fedora
+chpasswd: { expire: False }
+`,
+				`version: 2
+ethernets:
+  eth0:
+    dhcp4: true
+    dhcp6: true
```
maiqueb, line 2753: I don't think this will work; I think that you need to template this properly depending on the results from `filterDualStackCIDRs` . i.e. if the cluster is:
- single stack ipv4: only enable dhcp4
- single stack ipv6: only enable dhcp6
- dual stack: enable both, as you have now

@qinqon to keep me honest. I recall some issues when you enabled the "wrong" family and there was no way to get an IP from it.

test/e2e/kubevirt.go
```diff
@@ -2521,4 +2563,62 @@ ethernets:
 			)), "second VM should not be ready due to MAC conflict")
 		})
 	})
+
+	Context("subnet exhaustion", func() {
+		const networkName = "net1"
+		var (
+			cudn     *udnv1.ClusterUserDefinedNetwork
+			cidrIPv4 = "10.130.0.0/30" // no usable IPs (.1 and .2, .0 is network infra owned, .3 is broadcast)
+			cidrIPv6 = "2011:100:200::0/120"
```
maiqueb, line 2717: /120 ? 

Doesn't that subnet have 256 assignable IPs ?

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: Agree, we should not fail if the issue is updatng the IPAMClaim status, important thing is allocation not reporting at the status, still it mertis a proper error log.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -140,6 +148,18 @@ func allocatePodAnnotation(
...
+		updatedIPAMClaim := updateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
+		allocationErr := err
+		reconcileErr := claimsReconciler.Reconcile(ipamClaim, updatedIPAMClaim, ipAllocator)
+		// Reconcile error takes precedence
+		if reconcileErr != nil {
+			err = reconcileErr
+		} else {
+			err = allocationErr
+		}
+	}
```
qinqon: We need the minimal code here and just log inside updateIPAMClaimStatus is there were an error
```suggestion
	if ipamClaim != nil {
		updatedIPAMClaim := claimsReconciler.UpdateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
	}
```

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -225,13 +251,117 @@ func allocatePodAnnotationWithTunnelID(
...
+
+	if hasIPAMClaim && !hasPersistentIPs {
+		klog.Errorf(
+			"Pod %s/%s referencing an IPAMClaim on network %q which does not honor it",
+			pod.GetNamespace(),
+			pod.GetName(),
+			netInfo.GetNetworkName(),
+		)
+		return nil, nil
+	}
```
qinqon: Simplify all this

```suggestion
	if !hasIPAMClaim {
	  return nil, nil
	}

	if !hasPersistentIPs {
		klog.Errorf(
			"Pod %s/%s referencing an IPAMClaim on network %q which does not honor it",
			pod.GetNamespace(),
			pod.GetName(),
			netInfo.GetNetworkName(),
		)
		return nil, nil
	}
	ipamClaim, err := claimsReconciler.FindIPAMClaim(network.IPAMClaimReference, network.Namespace)
	if err != nil {
			return nil, fmt.Errorf("error retrieving IPAMClaim for pod %s/%s: %w", pod.GetNamespace(), pod.GetName(), err)
	}
	return ipamClaim, nil
```

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: I don't hink this as bad since we have multiple callers, we better compute the IPAMClaim status at one place, but is true that inside of it we can factor it a little with little functions.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -225,13 +251,117 @@ func allocatePodAnnotationWithTunnelID(
...
+		meta.SetStatusCondition(&updatedClaim.Status.Conditions, metav1.Condition{
+			Type:               conditionType,
+			Status:             metav1.ConditionTrue,
+			Reason:             "SuccessfulAllocation",
+			Message:            "IP addresses successfully allocated",
+			ObservedGeneration: updatedClaim.Generation,
+			LastTransitionTime: now,
+		})
+	} else {
+		var reason string
```
qinqon: Maybe better reset here the IPs in case we are updating something and we have IPs before.
```suggestion
		var reason string
		updatedClaim.Status.IPs = []string{}
```

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -225,13 +251,117 @@ func allocatePodAnnotationWithTunnelID(
...
+	podAnnotation *util.PodAnnotation,
+	podName string,
+	allocationErr error,
+) *ipamclaimsapi.IPAMClaim {
+	updatedClaim := ipamClaim.DeepCopy()
+
+	updatedClaim.Status.OwnerPod = &ipamclaimsapi.OwnerPod{Name: podName}
+	updatedClaim.Status.IPs = []string{}
+
+	if allocationErr == nil &&
```
qinqon: I will factor the logic here into somethin like 
```suggestion
	if allocationErr != nil {
	    return updateIPAMClimaStatusOnError(allocationErr...)
	}
	updateIPAMClimaStatusOnSuccess(...)
```

test/e2e/kubevirt.go
RamLavi: I don't know really, this is [taken](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d93369dc7627667e578d8c7f6cbe7cfb5af13f5c/test/e2e/kubevirt.go#L2402) from the functions above (createVMWithStaticIP).

test/e2e/kubevirt.go
RamLavi: We do something similar in createVMWithStaticIP. 
It seems to run and pass both on the [dual-stack live-migration lane](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18904041775/job/53960856516?pr=5683#step:16:17819) and the [single stack ipv4 lane](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/18904041775/job/53960856484?pr=5683#step:16:6105).
but @qinqon please double check me.

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: DONE

go-controller/pkg/allocator/pod/pod_annotation_test.go
RamLavi: something like this?
```
func requiresCustomNetworkSetup(ipam bool, idAllocation bool, persistentIPAllocation bool, network *nadapi.NetworkSelectionElement) bool {
	hasIPAMClaimRef := network != nil && network.IPAMClaimReference != ""
	return !ipam || idAllocation || persistentIPAllocation || hasIPAMClaimRef
}
```

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: ACK you're absolutely right,  will only log in case of ipamClaim reconcile error

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: ACK, will only log in case of reconcileErr. thanks

go-controller/pkg/allocator/pod/pod_annotation_test.go
maiqueb: no, please no. 

Make it a `tt` method, or receive tt as an argument.

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: both success and fail conditions set the same condition type `IPsAllocated` - I think we don't really need to split them.
I moved the Conditions handling to a helper. @maiqueb @qinqon wdyt?

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: ACK

go-controller/pkg/allocator/pod/pod_annotation.go
maiqueb: that `err` attribute is pretty much an artifact representing a boolean... 

I would avoid it, and compute in the caller which type of condition I want to set. IMHO it more natural, but I can compromise. Let's see what the next iteration will be.

test/e2e/kubevirt.go
RamLavi: DONE

test/e2e/kubevirt.go
RamLavi: DONE

test/e2e/kubevirt.go
RamLavi: DONE

test/e2e/kubevirt.go
RamLavi: >remind me how this works ?... so we'll find IPsAllocated always

yes

>remind me how this works ?... so we'll find IPsAllocated always 

by dedicated do you mean the .1,.2,.3 reserved IPs? in that case - yes, the result would be IsErrAllocated=true and you need to tell the user that this IP is taken.
and if by dedicated you mean a specific IP that the user wants - then the answer is also yes - cause you want to tell them if the allocation succeeded.

test/e2e/kubevirt.go
RamLavi: DONE (by splitting to 2 functions as you requested)

test/e2e/kubevirt.go
RamLavi: DONE

test/e2e/kubevirt.go
RamLavi: chose simplistic.

test/e2e/kubevirt.go
RamLavi: You are correct. I am only checking ipamClaim condition logic, and to do that I'm **using** the case of IPV4 subnet exhaustion.

test/e2e/kubevirt.go
maiqueb: I meant the latter. 

But yes, got it. Makes sense. Thank you.

test/e2e/kubevirt.go
maiqueb: hm. OK, I guess not that bad. 

Could you add the ipv4 keyword to the test description (or to the context) ? That way, it shows it is intentional.

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/eba8dd4f5a6a1923de8c850d9fb5409c6addb99f..dc13e5990f3496f8af41fadc1c793854440db113): Address @qinqon @maiqueb reviews

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -140,6 +148,13 @@ func allocatePodAnnotation(
 		allocateToPodWithRollback,
 	)
 
+	if ipamClaim != nil {
+		updatedIPAMClaim := updateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
+		if reconcileErr := claimsReconciler.Reconcile(ipamClaim, updatedIPAMClaim, ipAllocator); reconcileErr != nil {
+			klog.Errorf("Failed to reconcile IPAM claim %s/%s: %v", ipamClaim.Namespace, ipamClaim.Name, reconcileErr)
+		}
+	}
+
```
qinqon:
```suggestion
	if ipamClaim != nil {
		claimReconciler.UpdateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, ipAllocator, err)
     }
```

Let''s do the klog.Errorf inside of the reconciler, so we simplify the caller and make it clear that we don't handle the errors there.

qinqon (changes requested):

go-controller/pkg/allocator/pod/pod_annotation.go
qinqon: I don't agree we have multiple callers, that should not care about ipam claim status handling, the error here is part of the computation, the fact that is a boolean is just an implementation detail, the logic at caller should kind of a hook like "now try to update the ipam claim status if possible"

test/e2e/kubevirt.go
qinqon: We latest NetworkManager the may-fail for IPv6 is true if we set dhcp6: true, this mean that the VM will not propertly start up if we don't have dual stack, so yes, we should use different cloud init depending on ip family.

test/e2e/kubevirt.go
qinqon: Nope, the latest NetworkManager default is generate the IPv6 LLA at random and ignore the mac addres, this do not work with ovn-kubernetes port security things.

go-controller/pkg/allocator/pod/pod_annotation.go
maiqueb: I don't follow what's the issue. We can do that, but not pass the err down the stack, which is just an artifact for a boolean. 

Just compute the IPs you want to set in the status in the caller, and pass it to this `updateStatus` function as a parameter. 

You can also build the condition (or just the reason, which, afaiu, is the only thing you need the `err` variable for) in the caller and pass it down fwiw.

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: DONE

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/dc13e5990f3496f8af41fadc1c793854440db113..6e8a639773dc4eabfc7090f4a3fcc5c1dbc68e44): Address @qinqon [review](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683#discussion_r2493993302)

qinqon (changes requested):
So  factoring thins out a little and making sure that we cover internal errors too.

UPDATED: Refactor proposal got lost.

go-controller/pkg/allocator/pod/pod_annotation_test.go
RamLavi: I don't follow, `tt` is a test  struct object used for this specific test table, it's not an interface:
```
tests := []struct {
		name                            string
		args                            args
		netInfo                         util.NetInfo
		nadName                         string
		ipam                            bool
		idAllocation                    bool
		persistentIPAllocation          bool
		enablePreconfiguredUDNAddresses bool
		role                            string
		podAnnotation                   *util.PodAnnotation
		invalidNetworkAnnotation        bool
		wantUpdatedPod                  bool
		wantGeneratedMac                bool
		wantPodAnnotation               *util.PodAnnotation
		wantReleasedIPs                 []*net.IPNet
		wantReleasedIPsOnRollback       []*net.IPNet
		wantReservedMAC                 net.HardwareAddr
		wantReleaseMACOnRollback        net.HardwareAddr
		wantReleaseID                   bool
		wantRelasedIDOnRollback         bool
		wantErr                         bool
		isSingleStackIPv4               bool
		isSingleStackIPv6               bool
		multiNetworkDisabled            bool
	}
```

So you prefer that I add it as a function and not a variable?

test/e2e/kubevirt.go
RamLavi: DONE

test/e2e/kubevirt.go
RamLavi: So why don't we do it on any of the current tests?

RamLavi:
> So factoring thins out a little and making sure that we cover internal errors too.
```
	if ipam.IsErrFull(allocationErr) {
		reason = "SubnetExhausted"
	} else if ipam.IsErrAllocated(allocationErr) {
		reason = "IPAddressConflict"
	} else {
		reason = "InternalError"
	}
```

I don't think we should set InternalError in this case. Let me explain.
- it's not mentioned in the [enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/network/requesting-staticips-for-vms-being-migrated-into-primary-l2-udns-using-MTV.md#ipamclaim-crd).
- Not every `allocatePodAnnotation()` error should reflect to the ipamClaim (for example - retryOnConflict). IMO we should keep it focused on adding condition when something certain happened (ip already exists, ip pool exhaused). On other cases - we should not set a condition at all.

But the debate is very much valid, I will add the change but keep this part commented ([link](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683/files#diff-b45ef2d3bef0b2a17f93c8980065f71bd1b0889c851cc44207a470a2e8b594fbR249)), and ask maintainer to give us their opinion. I'm good with what ever they will decide.

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/6e8a639773dc4eabfc7090f4a3fcc5c1dbc68e44..4eb688f13bb1cc09d93a5173c7cfd77b558e45ce): Added suggested nits to updateStatus helper functions

qinqon (dismissed):
LGTM
Let's see if InternalError make sense for maintainers.

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +201,67 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
 	klog.V(5).Infof("Released IPs: %+v", ips)
 	return nil
 }
+
+// UpdateIPAMClaimStatus updates the IPAM claim status and reconciles it.
+// This method handles status updates and error logging internally.
+func (icr *IPAMClaimReconciler) UpdateIPAMClaimStatus(
```
kyrtapz, line 208: This is a little confusing, `UpdateIPAMClaimStatus` modifies the object which is then applied in `Reconcile`. Why can't this be handled by `Reconcile` directly?

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +201,67 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
+	updatedClaim := ipamClaim.DeepCopy()
+	updatedClaim.Status.OwnerPod = &ipamclaimsapi.OwnerPod{Name: podName}
+	if allocationErr != nil {
+		updateIPAMClaimAllocationErrorStatus(updatedClaim, allocationErr)
+	} else {
+		updateIPAMClaimAllocationSuccessStatus(updatedClaim, podAnnotation)
+	}
+
+	if reconcileErr := icr.Reconcile(ipamClaim, updatedClaim, ipReleaser); reconcileErr != nil {
+		klog.Errorf("Failed to reconcile IPAM claim %s/%s: %v", ipamClaim.Namespace, ipamClaim.Name, reconcileErr)
```
kyrtapz: Why are errors only logged? This doesn't look right.
In the previous implementation if the `Reconcile` failed the IP allocation would be rolled back and we would likely retry, here we just ignore it?

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +201,67 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
+	allocationErr error,
+) {
+	updatedClaim.Status.IPs = []string{}
+	var reason string
+
+	if ipam.IsErrFull(allocationErr) {
+		reason = "SubnetExhausted"
+	} else if ipam.IsErrAllocated(allocationErr) {
+		reason = "IPAddressConflict"
+		// } else {
```
kyrtapz: I think we should set the error condition here, we cannot just hope it was transient and it will eventually go away. What would be the reasoning for not doing that?
As a side note, you were not clearing the condition so potentially the ipamclaim would be stuck with the previous value.

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +201,67 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
+	ipamClaim *ipamclaimsapi.IPAMClaim,
+	podAnnotation *util.PodAnnotation,
+	podName string,
+	ipReleaser IPReleaser,
+	allocationErr error,
+) {
+	updatedClaim := ipamClaim.DeepCopy()
+	updatedClaim.Status.OwnerPod = &ipamclaimsapi.OwnerPod{Name: podName}
+	if allocationErr != nil {
+		updateIPAMClaimAllocationErrorStatus(updatedClaim, allocationErr)
```
kyrtapz, line 217: NIT: Do you think these two helpers make sense? Personally I think I would find it easier to follow if we put it all here or make one helper to get the status.  Right now we set the owner here but the rest in helpers.

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +201,67 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
+	if reconcileErr := icr.Reconcile(ipamClaim, updatedClaim, ipReleaser); reconcileErr != nil {
+		klog.Errorf("Failed to reconcile IPAM claim %s/%s: %v", ipamClaim.Namespace, ipamClaim.Name, reconcileErr)
+	}
+}
+
+func updateIPAMClaimAllocationSuccessStatus(
+	updatedClaim *ipamclaimsapi.IPAMClaim,
+	podAnnotation *util.PodAnnotation,
+) {
+	if podAnnotation != nil && len(podAnnotation.IPs) > 0 {
```
kyrtapz, line 228: Isn't this an error scenario if we set the condition without also setting the IPs?

go-controller/pkg/persistentips/allocator.go
RamLavi: I agree, UpdateIPAMClaimStatus should not reconcile.
Moved it outside, keeping the code structure as before (update status, then reconcile)

go-controller/pkg/persistentips/allocator.go
RamLavi: I initially did it in the same helper as you suggest, but the reviewers thought it would be more readable separated. 
I personally can find merit in both options. 
So if you strongly feel I should change it back - just say so.

go-controller/pkg/persistentips/allocator.go
RamLavi: It is. if there are no ips then the error should not be nil accordingly (and therefor be in the "bad-status" helper).
this condition is merely a defensive protection so that we don't panic when setting the IPs.

go-controller/pkg/persistentips/allocator.go
RamLavi: >set the error condition

by error condition you mean the condition `message` right?

well it is actually very much related to a [discussion](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683#issuecomment-3496170903) with @qinqon on the PR.
We would love you opinion on that.

why it is related? because if we want to return err on every allocationErr type then indeed we must set the condition.message as well (to also fit the general case message).
My intention however was to only set "failure condition" in cases where we deterministically know what failed (i.e. `IsErrFull` and `IsErrAllocated`), and by that avoiding transient conditions. and in that case we can use the allocationErr.Error() as the message, though I don't mind being explicit anyways, as you ask)

Once I know your opinion on the matter then I will go with whatever you say

go-controller/pkg/persistentips/allocator.go
RamLavi: Before this change, IPAMClaim Reconcile would only be done when there was no err, so there was no err shadowing in case the reconcile failed:
```
	if ipamClaim != nil && err == nil {
		newIPAMClaim := ipamClaim.DeepCopy()
		newIPAMClaim.Status.IPs = util.StringSlice(podAnnotation.IPs)
		err = claimsReconciler.Reconcile(ipamClaim, newIPAMClaim, ipAllocator)
        }
```

But now we also update the IPAMClaim if there is an error (to set the fail conditions), so if there is both allocationErr AND reconcileErr then we need to decide what to return. 
You know what - I'll wrap them together.

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/4eb688f13bb1cc09d93a5173c7cfd77b558e45ce..facdac0ffe2d63e0a5f3ce554a0079c3fd80aebf): Addressed @kyrtapz reviews. PTAL

RamLavi:
All updated test [pass](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/19470198733/job/55719116797?pr=5683#step:16:17072)

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +201,63 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
+		updatedClaim.Status.IPs = util.StringSlice(podAnnotation.IPs)
+	}
+
+	setIPClaimIPsAllocatedStatusCondition(updatedClaim, metav1.ConditionTrue, "SuccessfulAllocation", "IP addresses successfully allocated")
```
kyrtapz: Is there a point to update the status condition if there were no IPs added? This seems like a buggy status.

go-controller/pkg/persistentips/allocator.go
kyrtapz: I think we need the status to be set on every error to avoid a scenario where we encounter an issue and we don't report it to the user. Note that we depend on the IPs from the status too so that can be affected if we ever want to change the status.
Are there any transient issues that are always happening that made you concerned about it?

go-controller/pkg/persistentips/allocator.go
kyrtapz: I think you are still changing behavior here. Imagine a scenario similar to this:
```
  1. Pod requests an IPAMClaim with 10.0.0.5 already allocated
  2. allocatePodAnnotationWithRollback allocates new IP 10.0.0.10 successfully
  3. Pod annotation update succeeds, UpdatePodWithRetryOrRollback returns nil error
  4. UpdateIPAMClaimStatus is called with err=nil, sets IPs to [10.0.0.10]
  5. Reconcile() is called with oldClaim.Status.IPs=[10.0.0.5] and newClaim.Status.IPs=[10.0.0.10]
  6. Reconcile() detects IP mismatch at line 126-133 and returns error
  7. Error is joined but 10.0.0.10 is never released, ipamclaim reconciliation will never succeed
```
I am not confident if this is something that can happen in the wild but I think there was a discussion in the past of whether we protect ipamclaims IP from changing.
@qinqon do you remember?

go-controller/pkg/persistentips/allocator.go
kyrtapz: We can keep it as is.

go-controller/pkg/persistentips/allocator.go
RamLavi: well naturally retry on conflicts transient errors come to mind. Maybe there are more. 
The question is - do we want to "soil" the conditions with transient conditions...

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/facdac0ffe2d63e0a5f3ce554a0079c3fd80aebf..59b9bceae93db6f1cde2c4620aa00499dc16af2a): Rebase + fix review no.2: removed the extraction of ipamclaim handling outside of allocate function (will attempt in separate PR)

RamLavi:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/59b9bceae93db6f1cde2c4620aa00499dc16af2a..465ecd81dce4a3e6e9208b4db312448072546bcc: Rebase (no conflicts)

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -400,6 +400,15 @@ func allocatePodAnnotationWithRollback(
...
 	}
 
+	defer func() {
+		if ipamClaim != nil && claimsReconciler != nil {
+			updatedClaim := claimsReconciler.UpdateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
+			if reconcileErr := claimsReconciler.Reconcile(ipamClaim, updatedClaim, ipAllocator); reconcileErr != nil {
+				err = errors.Join(err, fmt.Errorf("failed to reconcile IPAM claim %s/%s: %w", ipamClaim.Namespace, ipamClaim.Name, reconcileErr))
+			}
+		}
+	}()
```
qinqon, line 456: Let's just reduce identation, so we don't do if statements nesting.
```suggestion
	defer func() {
		if ipamClaim == nil || claimsReconciler == nil {
		    return 
		}
		updatedClaim := claimsReconciler.UpdateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
		if reconcileErr := claimsReconciler.Reconcile(ipamClaim, updatedClaim, ipAllocator); reconcileErr != nil {
			err = errors.Join(err, fmt.Errorf("failed to reconcile IPAM claim %s/%s: %w", ipamClaim.Namespace, ipamClaim.Name, reconcileErr))
		}
	}()
```

qinqon (changes requested):
Is this legit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683/files#r2575883147?

RamLavi:
> Is this legit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683/files#r2575883147?

mr.s rabbit's suggestion is over defensive in my opinion.
We only get to this function if there was no allocation error, both of their options make no sense, considering the way the code currently works.

go-controller/pkg/persistentips/allocator.go
RamLavi: I understand.
I'll remove the first commit that extracts ipam-claim logic from the allocation function. That's what's causing the change of behavior. I'll try to introduce it in a separate PR so that we could discuss it appropriately.
Now - failing to reconcile ipamClaim will result with rollback.

go-controller/pkg/persistentips/allocator.go
RamLavi: I added another case that can result with rolback - MAC conflict.
With the returning of ipam-claim logic to the allocation function - there is no worry from getting retry on conflicts errors.

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: [DONE](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/465ecd81dce4a3e6e9208b4db312448072546bcc..187192896a278a6b601c5eb3371978c6c0ff1ac5)

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/465ecd81dce4a3e6e9208b4db312448072546bcc..187192896a278a6b601c5eb3371978c6c0ff1ac5): Address @qinqon 's review

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +202,63 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
+
+	if ipam.IsErrFull(allocationErr) {
+		reason = "SubnetExhausted"
+	} else if ipam.IsErrAllocated(allocationErr) {
+		reason = "IPAddressConflict"
+	} else if errors.Is(allocationErr, mac.ErrReserveMACConflict) {
+		reason = "MACAddressConflict"
+	} else {
+		reason = "InternalError"
+	}
```
maiqueb, line 250: switch case ? 

and maybe computing the reason outside ? I shiver everytime I see an error being input to something ...

go-controller/pkg/persistentips/allocator.go
```diff
@@ -197,3 +202,63 @@ func (icr *IPAMClaimReconciler) releaseIPs(ipamClaim *ipamclaimsapi.IPAMClaim, i
...
 	return nil
 }
+
+// UpdateIPAMClaimStatus updates the IPAM claim status.
+// This method handles status updates and error logging internally.
+func (icr *IPAMClaimReconciler) UpdateIPAMClaimStatus(
+	ipamClaim *ipamclaimsapi.IPAMClaim,
+	podAnnotation *util.PodAnnotation,
+	podName string,
+	allocationErr error,
```
maiqueb, line 212: I'd be so much happier if we didn't have to pass errors along helpers to figure out which scenario we're on. 

This is opinionated.

qinqon:
> The current changes look good to me, thanks @RamLavi! @qinqon, @maiqueb are you good with this pr?

LGTM

qinqon (approved):
LGTM

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -400,6 +400,16 @@ func allocatePodAnnotationWithRollback(
 		hasIPAMClaim = ipamClaim != nil && len(ipamClaim.Status.IPs) > 0
 	}
 
+	defer func() {
+		if ipamClaim == nil || claimsReconciler == nil {
+			return
+		}
+		updatedClaim := claimsReconciler.UpdateIPAMClaimStatus(ipamClaim, podAnnotation, pod.Name, err)
```
kyrtapz, line 452: Something I just thought of...
With this change any pod that tries to miss-behave will be able to wipe the ipamclaims status and move it to failed state, is that correct?
If there is an IPAMClaims that exists already and I create a random pod that references it to mess with it will we loose the status? I am not sure if we can do much about it but let's get a common understanding on what would happen and if it is a real issue.

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -400,6 +400,16 @@ func allocatePodAnnotationWithRollback(
...
+		if reconcileErr := claimsReconciler.Reconcile(ipamClaim, updatedClaim, ipAllocator); reconcileErr != nil {
```
kyrtapz, line 453: One more thing....
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9ea5b14aa3b2dcf2e6ebf36bbf3330337cf262e1/go-controller/pkg/persistentips/allocator.go#L90-L104
Isn't the reconcile only updating the object when an IP was not already set?

go-controller/pkg/allocator/pod/pod_annotation.go
kyrtapz: Maybe it is not a problem with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683/files#r2584738662?

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: yes.
Once an IPAMClaim has IPs allocated, Reconcile will never update it via the API. It only validates that any subsequent reconciliation attempts are trying to use the same IPs.
This means:
- A malicious pod cannot wipe an existing IPAMClaim's IPs through your new code
- The only time IPs get written is when the claim is initially empty
- After that, Reconcile is read-only (validation-only)

go-controller/pkg/allocator/pod/pod_annotation.go
RamLavi: >If there is an IPAMClaims that exists already and I create a random pod that references it to mess with it will we loose the status?
Maybe it is not a problem with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5683/files#r2584738662?

Indeed.  we can imagine this Scenario: pod A created, and IPAMClaim exists with Status.IPs = ["10.0.0.5"], malicious Pod B references it, allocation fails.
In the defer block:
1. `UpdateIPAMClaimStatus` is called with the allocation error
2. It creates `updatedClaim` with `Status.IPs = [] `
3. `Reconcile(ipamClaim, updatedClaim, ipAllocator)` is called
Now in Reconcile:
```
mustUpdateIPAMClaim := (oldIPAMClaim == nil ||         // false (it exists)
    len(oldIPAMClaim.Status.IPs) == 0) &&             // false (has IPs)
    newIPAMClaim != nil                                // true
// Result: false
```
- `oldIPAMClaim = the original ipamClaim ` with `Status.IPs = ["10.0.0.5"]`
- `newIPAMClaim = updatedClaim `with `Status.IPs = []`
- ---> `mustUpdateIPAMClaim` = false

we then find the original&update ipamclaims are different `(["10.0.0.5"] != [])` and return err..

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/187192896a278a6b601c5eb3371978c6c0ff1ac5..627915c79a899a7bdde82c8987e4ef6be085f710): rebase

RamLavi:
[Change](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/627915c79a899a7bdde82c8987e4ef6be085f710..8832169b9f206ceeced68bffa9930d239af423bc): Remove redundant function.

RamLavi:
> @RamLavi there is a conflict :(, we are good to merge once it is resolved.

@kyrtapz PTAL

PR #4388, opened 2024-05-23
Make egress firewall node controller level-driven

npinaeva:
Beginning of the story is in https://github.com/ovn-org/ovn-kubernetes/pull/4386
Switch to level-driven controller to avoid
dependency between node handler and ef node handler.
If EF handler is ever deadlocked (and it can be as node handler iterates
all egress firewalls, which in turn, depend on dns resolvers), node
handler should not be affected.
See a new test for more idea.

Issue #3431, opened 2023-02-18
pod creation blocks network policy creation


freedge:
with the use of gatekeeper, it is possible to have admission rules that slow down the creation of the pods. These rules are called by OVNK while patching pods with annotations for IP and MAC addresses.
During that time, no network policy can be created. This is a problem when restarting the ovnkube-master on a system with a lot of network policies, as each pod creation further delays the completion of the watchers.
After a restart it can take hours for all the network policies to be installed, causing new pods to be started without network connectivity for hours.
Contention seems to come from a [WatchFactory.informers](https://github.com/ovn-org/ovn-kubernetes/blob/4bb783f48b9e7579f4e6a132986bb4c054832725/go-controller/pkg/factory/factory.go#L553) mutex.

This can be reproduced on ovn-kubernetes main branch:
1: installing gatekeeper policies
```kubectl apply -f https://raw.githubusercontent.com/open-policy-agent/gatekeeper/master/deploy/gatekeeper.yaml```

here the slow admission policies is simulated using a fake service running on 10.224.123.1:8443 (```nc -l -p 8443 -k``` on that server) as I did not find a "sleep" function in OPA doc...

```yaml:
apiVersion: externaldata.gatekeeper.sh/v1alpha1
kind: Provider
metadata:
  name: dummy-provider
spec:
  url: https://10.224.123.1:8443/validate
  timeout: 10
[...]

PR #5688, opened 2025-10-24
Egress firewall refactor - Level driven + UDN support

trozet:
- Breaks EgressFirewall into its own level driven controller
- Logic modified to not remove firewall rules and then add them (creating a gap of time where no firewall is enforced)
- Adds support for UDN
- Various other fixes to handling errors that were being ignored before with egress firewall, and never retried (like status update)
- Adds UDN unit tests
- Adds UDN e2e test for egress firewall

## Summary by CodeRabbit

* **New Features**
  * Per-namespace egress firewall controller and NAD handler registration.
  * E2E tests extended to cover User-Defined Network (UDN) scenarios.

* **Refactor**
  * Controller-centric egress firewall lifecycle and reconciliation introduced.
  * DNS-based destination validation simplified; legacy OVN egress-firewall implementation removed.
  * Network manager test paths consolidated to use the public network manager.

* **Tests**
  * Added comprehensive egress firewall unit tests and updated test harnesses.

### DIFF
--- a/go-controller/pkg/networkmanager/fake.go
+++ b/go-controller/pkg/networkmanager/fake.go
@@ -2,6 +2,9 @@ package networkmanager
 
 import (
 	"context"
+	"sync"
+
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util/errors"
@@ -44,13 +47,28 @@ func (fcm *FakeControllerManager) Reconcile(_ string, _, _ util.NetInfo) error {
 }
 
 type FakeNetworkManager struct {
+	sync.Mutex
 	// namespace -> netInfo
 	// if netInfo is nil, it represents a namespace which contains the required UDN label but with no valid network. It will return invalid network error.
 	PrimaryNetworks map[string]util.NetInfo
+	HandlerFuncs    []handlerFunc
+	// UDNNamespaces are a list of namespaces that require UDN for primary network
+	UDNNamespaces sets.Set[string]
 }
 
-func (fnm *FakeNetworkManager) RegisterNADHandler(_ handlerFunc) error {
+func (fnm *FakeNetworkManager) RegisterNADHandler(h handlerFunc) error {
+	fnm.Lock()
+	defer fnm.Unlock()
+	fnm.HandlerFuncs = append(fnm.HandlerFuncs, h)
 	return nil
+}
+
+func (fnm *FakeNetworkManager) TriggerHandlers(nadName string, info util.NetInfo, removed bool) {
+	fnm.Lock()
+	defer fnm.Unlock()
+	for _, h := range fnm.HandlerFuncs {
+		h(nadName, info, removed)
+	}
 }
 
 func (fnm *FakeNetworkManager) Interface() Interface {
@@ -70,8 +88,13 @@ func (fnm *FakeNetworkManager) GetActiveNetworkForNamespace(namespace string) (u
 }
 
 func (fnm *FakeNetworkManager) GetActiveNetworkForNamespaceFast(namespace string) util.NetInfo {
+	fnm.Lock()
+	defer fnm.Unlock()
 	if primaryNetworks, ok := fnm.PrimaryNetworks[namespace]; ok {
 		return primaryNetworks
+	}
+	if fnm.UDNNamespaces != nil && fnm.UDNNamespaces.Has(namespace) {
+		return nil
 	}
 	return &util.DefaultNetInfo{}
 }
--- a/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
+++ b/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
@@ -60,6 +60,7 @@ const (
 	// This is due to legacy reasons and left alone for now to avoid upgrade issues. May be updated in the future.
 	defaultControllerOwner         = "default-network-controller"
 	EgressFirewallAppliedCorrectly = "EgressFirewall Rules applied"
+	egressFirewallName             = "default"
 )
 
 const (
@@ -109,13 +110,16 @@ type matchKind int
 type cacheEntry struct {
 	pgName          string
 	hasNodeSelector bool
+	stale           bool
 }
 
 type EFController struct {
 	name        string
 	zone        string
 	ruleCounter sync.Map
-	cache       *syncmap.SyncMap[*cacheEntry] // cache stores a mapping of rendered firewalls -> nodes + networks
+	// cache stores a mapping of rendered firewall namespace -> nodes + networks
+	// only one firewall (named "default") can exist per namespace
+	cache *syncmap.SyncMap[*cacheEntry]
 
 	// libovsdb northbound client interface
 	nbClient libovsdbclient.Client
@@ -313,7 +317,10 @@ func (oc *EFController) initialSync() error {
 
 func (oc *EFController) Start() error {
 	klog.Infof("Starting EgressFirewall controller")
-	return controller.StartWithInitialSync(oc.initialSync, oc.controller, oc.nodeController)
+	if err := controller.StartWithInitialSync(oc.initialSync, oc.controller, oc.nodeController); err != nil {
+		return err
+	}
+	return oc.networkManager.RegisterNADHandler(oc.handleNetworkEvent)
 }
 
 func (oc *EFController) Stop() {
@@ -321,26 +328,82 @@ func (oc *EFController) Stop() {
 	controller.Stop(oc.nodeController, oc.controller)
 }
 
+func (oc *EFController) handleNetworkEvent(nadName string, info util.NetInfo, removed bool) {
+	if info != nil && !info.IsPrimaryNetwork() { // egressFirewall only supported for primary network
+		return
+	}
+	if removed { // delete case
+		namespace, _, err := cache.SplitMetaNamespaceKey(nadName)
+		if err != nil {
+			klog.Errorf("%s: failed splitting key %s: %v", oc.name, nadName, err)
+			return
+		}
+		oc.cache.LockKey(namespace)
+		entry, ok := oc.cache.Load(namespace)
+		if !ok {
+			oc.cache.UnlockKey(namespace)
+			return // no cache entry exists, so nothing to remove
+		}
+		entry.stale = true
+		oc.cache.UnlockKey(namespace)
+		klog.V(3).Infof("NAD removed for egress firewall in namespace: %q. Will sync.", namespace)
+		oc.controller.Reconcile(fmt.Sprintf("%s/%s", namespace, egressFirewallName))
+		return
+	}
+	// add/update case
+	namespace, _, err := cache.SplitMetaNamespaceKey(nadName)
+	if err != nil {
+		klog.Errorf("%s: failed splitting key %s: %v", oc.name, nadName, err)
+		return
+	}
+	ef, err := oc.efLister.EgressFirewalls(namespace).Get(egressFirewallName)
+	if err != nil || ef == nil {
+		return
+	}
+	key, err := cache.MetaNamespaceKeyFunc(ef)
+	if err != nil {
+		return
+	}
+	klog.V(3).Infof("NAD add/update for egress firewall in namespace: %q. Will sync.", namespace)
+	oc.controller.Reconcile(key)
+}
+
 func (oc *EFController) sync(key string) (updateErr error) {
-	oc.cache.LockKey(key)
-	defer oc.cache.UnlockKey(key)
-	klog.V(5).Info("Syncing EgressFirewall", key)
+	klog.V(5).Infof("Syncing EgressFirewall %s", key)
 	namespace, efName, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
 		return fmt.Errorf("invalid resource key for egress firewall: %s", key)
 	}
+	shouldRemove := false
+	shouldAddUpdate := false
+
 	ef, err := oc.efLister.EgressFirewalls(namespace).Get(efName)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
 			return err
 		}
+		shouldRemove = true
+	} else {
+		shouldAddUpdate = true
+	}
+
+	oc.cache.LockKey(namespace)
+	defer oc.cache.UnlockKey(namespace)
+
+	if entry, ok := oc.cache.Load(namespace); ok {
+		if entry.stale {
+			shouldRemove = true
+		}
+	}
+
+	if shouldRemove {
 		// delete case
-		cacheEntry, ok := oc.cache.Load(key)
+		entry, ok := oc.cache.Load(namespace)
 		if !ok {
 			return nil
 		}
 		klog.Infof("Removing egress firewall %s", key)
-		pgName := cacheEntry.pgName
+		pgName := entry.pgName
 
 		p := libovsdbops.GetPredicate[*nbdb.ACL](oc.GetEgressFirewallACLDbIDsNoRule(namespace), nil)
 		invalidACLs, err := libovsdbops.FindACLsWithPredicate(oc.nbClient, p)
@@ -350,6 +413,7 @@ func (oc *EFController) sync(key string) (updateErr error) {
 		if err := libovsdbops.DeleteACLsFromPortGroups(oc.nbClient, []string{pgName}, invalidACLs...); err != nil {
 			return fmt.Errorf("error deleting stale ACLs for egress firewall %s: %w", key, err)
 		}
+		oc.cache.Delete(namespace)
 		if err := oc.dnsNameResolver.Delete(namespace); err != nil {
 			return err
 		}
@@ -361,6 +425,9 @@ func (oc *EFController) sync(key string) (updateErr error) {
 			klog.Errorf("Unable to decrement egress firewall rule count, cache miss for key: %s", key)
 		}
 		metrics.DecrementEgressFirewallCount()
+	}
+
+	if !shouldAddUpdate {
 		return nil
 	}
 
@@ -385,7 +452,7 @@ func (oc *EFController) sync(key string) (updateErr error) {
 	if updateErr != nil {
 		return
 	}
-	oc.cache.Store(key, entry)
+	oc.cache.Store(namespace, entry)
 	// remove stale ACLs
 	p := libovsdbops.GetPredicate[*nbdb.ACL](oc.GetEgressFirewallACLDbIDsNoRule(namespace),
 		func(acl *nbdb.ACL) bool {
@@ -561,12 +628,23 @@ func efNodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
 func (oc *EFController) updateEgressFirewallForNode(nodeName string) error {
 	klog.V(3).Infof("Syncing node %q for egress firewall", nodeName)
 
-	efKeys := oc.cache.GetKeys()
-	for _, key := range efKeys {
-		if err := oc.cache.DoWithLock(key, func(key string) error {
-			if cacheEntry, ok := oc.cache.Load(key); ok && cacheEntry.hasNodeSelector {
-				klog.Infof("Syncing egress firewall %s due to node %q change", key, nodeName)
-				oc.controller.Reconcile(key)
+	efNamespaces := oc.cache.GetKeys()
+	for _, namespace := range efNamespaces {
+		if err := oc.cache.DoWithLock(namespace, func(key string) error {
+			if entry, ok := oc.cache.Load(key); ok && entry.hasNodeSelector {
+				ef, err := oc.efLister.EgressFirewalls(namespace).Get(egressFirewallName)
+				if err != nil {
+					if !apierrors.IsNotFound(err) {
+						return err
+					}
+					return nil
+				}
+				efKey, err := cache.MetaNamespaceKeyFunc(ef)
+				if err != nil {
+					return fmt.Errorf("couldn't get key for egress firewall object %+v: %v", ef, err)
+				}
+				klog.Infof("Syncing egress firewall %s due to node %q change", efKey, nodeName)
+				oc.controller.Reconcile(efKey)
 			}
 			return nil
 		}); err != nil {
--- a/go-controller/pkg/ovn/egressfirewall_test.go
+++ b/go-controller/pkg/ovn/egressfirewall_test.go
@@ -18,6 +18,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	utilnet "k8s.io/utils/net"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
@@ -762,6 +763,77 @@ var _ = ginkgo.Describe("OVN EgressFirewall Operations", func() {
 
 					expectedDatabaseState = getEFExpectedDbAfterDelete(expectedDatabaseState)
 					gomega.Eventually(fakeOVN.nbClient).Should(libovsdb.HaveData(expectedDatabaseState))
+
+					return nil
+				}
+
+				err := app.Run([]string{app.Name})
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			})
+			ginkgo.It(fmt.Sprintf("correctly handles egressFirewall with UDN change, gateway mode %s", gwMode), func() {
+				config.Gateway.Mode = gwMode
+				config.OVNKubernetesFeature.EnableMultiNetwork = true
+				config.OVNKubernetesFeature.EnableNetworkSegmentation = true
+				app.Action = func(*cli.Context) error {
+					namespace1 := *newUDNNamespace("namespace1")
+					netconf := dummyLayer2PrimaryUserDefinedNetwork("192.168.0.0/16")
+					networkConfig, err := util.NewNetInfo(netconf.netconf())
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+					fakeNM := &networkmanager.FakeNetworkManager{
+						PrimaryNetworks: map[string]util.NetInfo{
+							namespace1.Name: networkConfig,
+						},
+						UDNNamespaces: sets.New[string](namespace1.Name),
+					}
+					fakeOVN.networkManager = fakeNM
+					// add UDN namespaced port group
+					ownerController := networkConfig.GetNetworkName() + "-network-controller"
+					pgIDs := getNamespacePortGroupDbIDs(namespace1.Name, ownerController)
+					namespacePortGroup := libovsdbutil.BuildPortGroup(pgIDs, nil, nil)
+					dbSetup.NBData = append(dbSetup.NBData, namespacePortGroup)
+					egressFirewall := newEgressFirewallObject("default", namespace1.Name, []egressfirewallapi.EgressFirewallRule{
+						{
+							Type: "Allow",
+							Ports: []egressfirewallapi.EgressFirewallPort{
+								{
+									Protocol: "TCP",
+									Port:     100,
+								},
+							},
+							To: egressfirewallapi.EgressFirewallDestination{
+								CIDRSelector: "1.2.3.5/23",
+							},
+						},
+					})
+
+					startOvn(dbSetup, []corev1.Namespace{namespace1}, []egressfirewallapi.EgressFirewall{*egressFirewall}, true)
+
+					expectedDatabaseState := getEFExpectedDbUDN(initialData, fakeOVN, namespace1.Name,
+						"(ip4.dst == 1.2.3.5/23)", "((tcp && ( tcp.dst == 100 )))", nbdb.ACLActionAllow, networkConfig.GetNetworkName())
+					gomega.Eventually(fakeOVN.nbClient).Should(libovsdb.HaveData(expectedDatabaseState))
+
+					ginkgo.By("triggering fake network manager to simulate NAD deletion, OVN EF config should be removed")
+					// remove primary NAD to simulate network being totally deleted
+					fakeNM.Lock()
+					delete(fakeNM.PrimaryNetworks, namespace1.Name)
+					fakeNM.Unlock()
+
+					fakeNM.TriggerHandlers(netconf.nadName, nil, true)
+
+					expectedDatabaseState = getEFExpectedDbAfterDelete(expectedDatabaseState)
+					gomega.Eventually(fakeOVN.nbClient).Should(libovsdb.HaveData(expectedDatabaseState))
+
+					ginkgo.By("creating new UDN for namespace, EF should get configured")
+					netconf2 := dummyLayer2PrimaryUserDefinedNetwork("128.168.0.0/16")
+					networkConfig2, err := util.NewNetInfo(netconf2.netconf())
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+					fakeNM.Lock()
+					fakeNM.PrimaryNetworks[namespace1.Name] = networkConfig2
+					fakeNM.Unlock()
+					fakeNM.TriggerHandlers(netconf.nadName, networkConfig2, false)
+					expectedDatabaseState2 := getEFExpectedDbUDN(initialData, fakeOVN, namespace1.Name,
+						"(ip4.dst == 1.2.3.5/23)", "((tcp && ( tcp.dst == 100 )))", nbdb.ACLActionAllow, networkConfig2.GetNetworkName())
+					gomega.Eventually(fakeOVN.nbClient).Should(libovsdb.HaveData(expectedDatabaseState2))
 
 					return nil
 				}


go-controller/pkg/ovn/egressfirewall_test.go
```diff
@@ -591,6 +638,50 @@ var _ = ginkgo.Describe("OVN EgressFirewall Operations", func() {
...
+					gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+					expectedDatabaseState := getEFExpectedDbUDN(initialData, fakeOVN, namespace1.Name,
+						"(ip4.dst == 1.2.3.4/23)", "((udp && ( udp.dst == 100 )))", nbdb.ACLActionDrop, networkConfig.GetNetworkName())
+					gomega.Eventually(fakeOVN.nbClient).Should(libovsdb.HaveData(expectedDatabaseState))
+					return nil
+				}
+				err := app.Run([]string{app.Name})
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			})
```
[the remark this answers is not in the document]
trozet: ns is defined as a constant in the package in layer3_user_defined_network_controller_test.go. But i'll just use a different variable that is within the same file

test/e2e/network_segmentation_egress_firewall.go
```diff
@@ -0,0 +1,178 @@
...
+
+	waitForEFApplied := func(namespace string) {
+		Eventually(func() bool {
+			output, err := kubectl.RunKubectl(namespace, "get", "egressfirewall", "default")
+			if err != nil {
+				framework.Failf("could not get the egressfirewall default in namespace: %s", namespace)
+			}
+			return strings.Contains(output, "EgressFirewall Rules applied")
+		}, 10*time.Second).Should(BeTrue(), fmt.Sprintf("expected egress firewall in namespace %s to be successfully applied", namespace))
+	}
```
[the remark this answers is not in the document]
trozet: the firewall should be created by the time we run this command, so it should never fail to get the firewall. This is the same code as in egress_firewall.go

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -164,14 +238,41 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
+					return fmt.Errorf("failed to search for port groups during egress firewall ACL sync: %w", err)
+				}
+				if len(foundPGs) > 1 {
+					pgNames := make([]string, len(foundPGs))
+					for _, pg := range foundPGs {
+						pgNames = append(pgNames, pg.Name)
+					}
+					klog.Warningf("Found multiple port groups associated with the same ACL %q: %s",
+						*acl.Name, strings.Join(pgNames, ","))
+				}
```
trozet, line 292: how does it yield empty values? it cycles through every element in foundPGs and appends it?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -512,76 +817,151 @@ func (oc *DefaultNetworkController) deleteEgressFirewallRule(namespace, pgName s
...
+		Force:        true,
+		FieldManager: oc.zone,
+	}
+
+	applyObj := egressfirewallapply.EgressFirewall(egressFirewall.Name, egressFirewall.Namespace).
+		WithStatus(egressfirewallapply.EgressFirewallStatus().
+			WithMessages(newMsg))
+	_, err := oc.kube.EgressFirewallClient.K8sV1().EgressFirewalls(egressFirewall.Namespace).ApplyStatus(context.TODO(), applyObj, applyOptions)
+
+	return err
```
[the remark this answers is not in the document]
trozet: you forgot to consider numRemovedRules, which is how many rules were removed during the update. You assume the delta is always higher newCount than previous, but it coudl be there is an update that reduces number of rules

test/e2e/egress_firewall.go
```diff
@@ -525,206 +587,206 @@ spec:
...
+						})
+						if err != nil && !strings.Contains(err.Error(), "Address already assigned") {
+							framework.Failf("failed to add new IP address %s to node %s: %v", ip, nodeName, err)
+						}
+						ginkgo.DeferCleanup(func() error {
+							_, err = infraprovider.Get().ExecK8NodeCommand(nodeName, []string{"ip", "addr", "delete", ip, "dev", deploymentconfig.Get().PrimaryInterfaceName()})
+							return err
+						})
+						toCurlSecondaryNodeIPAddresses.Insert(ip)
 					}
```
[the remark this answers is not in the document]
trozet: that's not true. If you add an ip address without mask it assumes full mask in Linux.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -16,38 +18,50 @@ import (
...
 	// transaction time to delete 80K ACLs from 2 port groups is ~4.5 sec.
 	// transaction time to delete 80K acls from one port group and add them to another
 	// is ~3 sec
 	// Therefore this limit is safe enough to not exceed 10 sec transaction timeout.
-	aclChangePGBatchSize = 80000
+	aclChangePGBatchSize           = 80000
+	matchKindV4CIDR      matchKind = iota
+	matchKindV6CIDR
+	matchKindV4AddressSet
+	matchKindV6AddressSet
```
npinaeva: may be better to move this into its own `const ()` to avoid accidentally adding new values

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -16,38 +18,50 @@ import (
...
+	defaultControllerOwner = "default-network-controller"
```
npinaeva: do you want to add a comment on how this is used?

go-controller/pkg/ovn/egressfirewall.go
```diff
@@ -16,38 +18,50 @@ import (
...
+	aclChangePGBatchSize           = 80000
+	matchKindV4CIDR      matchKind = iota
+	matchKindV6CIDR
+	matchKindV4AddressSet
+	matchKindV6AddressSet
+	defaultControllerOwner = "default-network-controller"
 )
 
 type egressFirewall struct {
-	sync.Mutex
```
npinaeva, line 50: chef's kiss

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -164,14 +238,41 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
+					if len(item.ACLs) == 0 {
+						return false
+					}
+					for _, aclUUID := range item.ACLs {
+						if acl.UUID == aclUUID {
+							return true
+						}
+					}
+					return false
+				}
```
npinaeva, line 276: can't we infer the port group name from the ACL ids?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -72,75 +86,135 @@ type destination struct {
...
-// newEgressFirewallRule creates a new egressFirewallRule. For the logging level, it will pick either of
-// aclLoggingAllow or aclLoggingDeny depending if this is an allow or deny rule.
-func (oc *DefaultNetworkController) newEgressFirewallRule(rawEgressFirewallRule egressfirewallapi.EgressFirewallRule, id int) (*egressFirewallRule, error) {
-	efr := &egressFirewallRule{
-		id:     id,
-		access: rawEgressFirewallRule.Type,
+type matchKind int
+
+type EFController struct {
+	sync.RWMutex
```
npinaeva: any comments on how to use the lock and which structs it guards?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -512,76 +817,161 @@ func (oc *DefaultNetworkController) deleteEgressFirewallRule(namespace, pgName s
...
+}
+
+func getNamespacePortGroupDbIDs(ns string, controller string) *libovsdbops.DbObjectIDs {
+	return libovsdbops.NewDbObjectIDs(libovsdbops.PortGroupNamespace, controller,
+		map[libovsdbops.ExternalIDKey]string{
+			libovsdbops.ObjectNameKey: ns,
+		})
+}
+
+func (oc *EFController) getNamespacePortGroupNameUnknownOwner(namespace string) (string, error) {
```
npinaeva: maybe better call it `getNamespacePrimaryPortGroupName`? I fine the unknown owner confusing

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -450,50 +586,219 @@ func (oc *DefaultNetworkController) addEgressFirewallRules(ef *egressFirewall, p
...
 
-func (oc *DefaultNetworkController) deleteEgressFirewallRule(namespace, pgName string, ruleIdx int) error {
+func (oc *EFController) GetEgressFirewallACLDbIDsNoRule(namespace string) *libovsdbops.DbObjectIDs {
+	return libovsdbops.NewDbObjectIDs(libovsdbops.ACLEgressFirewall, defaultControllerOwner,
+		map[libovsdbops.ExternalIDKey]string{
+			libovsdbops.ObjectNameKey: namespace,
+		})
+}
+
+func (oc *EFController) GetEgressFirewallACLDbIDs(namespace string, ruleIdx int) *libovsdbops.DbObjectIDs {
```
npinaeva, line 908: mayeb add a comment here that we currently reuse the previous default controller as an owner, but it may be upgraded at some point to properly reference the egressfirewall controller?

npinaeva: it should be a pretty simple upgrade, maybe as a follow up?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +289,83 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
-	})
+	// add/update case
+
+	// find old ACLs, we are going to track them, then remove them after we add
+	// Egress Firewall dynamically allocates priority depending on the rule position in the spec, which
+	// complicates tracking stale vs valid rules
+	// Therefore we wholesale add all rules then delete old rules, ensuring no gaps where there is
+	// no firewall applied
+	p := libovsdbops.GetPredicate[*nbdb.ACL](oc.GetEgressFirewallACLDbIDsNoRule(namespace), nil)
+	invalidACLs, err := libovsdbops.FindACLsWithPredicate(oc.nbClient, p)
```
npinaeva: Wouldn't it be easier to just run a similar predicate search, but with extra condition
`acl.ExternalIDs[libovsdbops.RuleIndex] > len(egressfirewall.Spec.Egress)` after you have created all the ACLs?
and delete whatever matches?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +289,83 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
-	staleNamespaces := map[string][]*nbdb.ACL{}
-	for _, acl := range efACLs {
-		namespace := acl.ExternalIDs[libovsdbops.ObjectNameKey.String()]
-		// no key will be treated as an empty namespace and ACLs will just be deleted
-		if staleUUIDs.Has(acl.UUID) {
-			staleNamespaces[namespace] = append(staleNamespaces[namespace], acl)
+	err = oc.addEgressFirewall(egressFirewall, &invalidACLs)
+	if err == nil {
+		// remove invalidACLs
+		if removalErr := libovsdbops.DeleteACLsFromAllPortGroups(oc.nbClient, invalidACLs...); removalErr != nil {
```
npinaeva: it would be more efficient to just delete from the right port group, we know its name right?

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -29,6 +27,7 @@ import (
 	addressset "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/address_set"
 	anpcontroller "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/controller/admin_network_policy"
 	apbroutecontroller "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/controller/apbroute"
+	egressfirewall2 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/controller/egressfirewall"
```
npinaeva: efcontroller?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +289,83 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
+			err = fmt.Errorf("error deleting stale ACLs for egress firewall %s: %v", egressFirewall.Name, err)
 		}
 	}
-	if len(staleNamespaces) == 0 {
-		return nil
+	if statusErr := oc.setEgressFirewallStatus(egressFirewall, err); statusErr != nil {
+		return fmt.Errorf("failed to update egress firewall status %s, error: %w",
```
npinaeva: I think this error should only be returned if `err == nil` otherwise it may override the more serious problem, which prevented the egress firewall from being configured

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -72,75 +86,135 @@ type destination struct {
...
+	// libovsdb northbound client interface
+	nbClient libovsdbclient.Client
+	kube     *kube.KubeOVN
+
+	nodeLister      corelisters.NodeLister
+	namespaceLister corelisters.NamespaceLister
+	efLister        v2.EgressFirewallLister
+
+	// egressFirewalls is a map of namespaces and the egressFirewall attached to it
+	egressFirewalls sync.Map
```
npinaeva: is it used?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -72,75 +86,135 @@ type destination struct {
...
+	oc.controller.Reconcile(key)
 }
 
 // syncEgressFirewall deletes stale db entries for previous versions of Egress Firewall implementation and removes
 // stale db entries for Egress Firewalls that don't exist anymore.
 // Egress firewall implementation had many versions, the latest one makes no difference for gateway modes, and creates
 // ACLs on namespaced port groups.
-func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interface{}) error {
-	err := oc.deleteStaleACLs()
+func (oc *EFController) syncEgressFirewall() error {
```
npinaeva: can we rename it to the initialSync?

go-controller/pkg/ovn/namespace.go
```diff
@@ -285,14 +287,23 @@ func (oc *DefaultNetworkController) updateNamespace(old, newer *corev1.Namespace
...
+			for _, fw := range egressFirewalls {
+				fwKey, err := cache.MetaNamespaceKeyFunc(fw)
+				if err != nil {
+					klog.Errorf("Failed to get key for EgressFirewall %s/%s, will not update ACL logging: %v", old.Name, fwKey, err)
+					continue
+				}
+				klog.Infof("Namespace %s: EgressFirewall ACL logging setting updating to deny=%s allow=%s",
+					old.Name, nsInfo.aclLogging.Deny, nsInfo.aclLogging.Allow)
+				oc.efController.Reconcile(fwKey)
+			}
```
npinaeva, line 306: nice!

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -320,60 +384,131 @@ func (oc *DefaultNetworkController) addEgressFirewall(egressFirewall *egressfire
...
+	klog.V(3).Infof("Syncing node %q for egress firewall", nodeName)
+
+	// cycle through egress firewalls and check if any match this node's labels
+	egressFirewalls, err := oc.efLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list all egress firewalls: %w", err)
+	}
+
+	for _, egressFirewall := range egressFirewalls {
+		ef, err := oc.buildEgressFirewallConstruct(egressFirewall)
```
npinaeva: I don't think you need to build, you can just use original EF fields directly

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -320,60 +384,131 @@ func (oc *DefaultNetworkController) addEgressFirewall(egressFirewall *egressfire
...
+		if err != nil {
+			return fmt.Errorf("failed to create EgressFirewall object for egress firewall %s/%s: %w",
+				egressFirewall.Namespace, egressFirewall.Name, err)
+		}
+
+		for _, rule := range ef.egressRules {
```
npinaeva: I think this doesn't update egress firewalls that don't match a nodeSelector anymore, so cleanup case?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -512,76 +817,161 @@ func (oc *DefaultNetworkController) deleteEgressFirewallRule(namespace, pgName s
...
+		if err != nil {
+			return err
+		}
+		newCount := uint32(len(egressFirewall.Spec.Egress))
+		if prev, ok := oc.ruleCounter.Load(key); ok {
+			// update: adjust rule count by delta (can be positive, negative, or zero)
+			prevCount := prev.(uint32)
+			delta := int(newCount) - int(prevCount)
+			if delta != 0 {
+				metrics.UpdateEgressFirewallRuleCount(float64(delta))
```
npinaeva, line 953: we can just replace `UpdateEgressFirewallRuleCount` with `SetEgressFirewallRuleCount` and then we don't need to calculate anything, just put current `len(egressFirewall.Spec.Egress)` and on delete 0?

go-controller/pkg/types/resource_status.go
```diff
@@ -7,10 +7,11 @@ import (
 
 // this file defines error messages that are used to figure out if a resource reconciliation failed
 const (
-	APBRouteErrorMsg       = "failed to apply policy"
-	EgressFirewallErrorMsg = "EgressFirewall Rules not correctly applied"
-	EgressQoSErrorMsg      = "EgressQoS Rules not correctly applied"
-	NetworkQoSErrorMsg     = "NetworkQoS Destinations not correctly applied"
+	APBRouteErrorMsg               = "failed to apply policy"
+	EgressFirewallErrorMsg         = "EgressFirewall Rules not correctly applied"
+	EgressFirewallAppliedCorrectly = "EgressFirewall Rules applied"
```
npinaeva: I am not sure this is the right place, because this module only cares about errors (see comment)
maybe just move it to egressfirewall?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -320,60 +385,180 @@ func (oc *DefaultNetworkController) addEgressFirewall(egressFirewall *egressfire
...
+		subnets := config.Default.ClusterSubnets
+		if !netInfo.IsDefault() {
+			subnets = netInfo.Subnets()
+		}
+		for _, clusterSubnet := range subnets {
+			if clusterSubnet.CIDR.Contains(ipNet.IP) || ipNet.Contains(clusterSubnet.CIDR.IP) {
+				clusterSubnetIntersection = append(clusterSubnetIntersection, clusterSubnet.CIDR)
+				break
+			}
+		}
```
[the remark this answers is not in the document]
trozet: cluster cidrs should be non-overlapping and validated in config.go.  Therefore the rule cidr should never be contained within more than one cluster cidr.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: I think we could also remove `oc.moveACLsToNamespacedPortGroups` completely (because the previous release has already synced then), but do as you wish

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -119,14 +119,26 @@ func (c *Controller) deleteNAD(obj client.Object, namespace string) error {
...
+		return fmt.Errorf("failed to list pods at target namespace %q: %w", nadCopy.Namespace, err)
 	}
 	// This is best-effort check no pod using the subject NAD,
-	// noting prevent a from being pod creation right after this check.
+	// nothing prevents pod creation right after this check.
 	if err := NetAttachDefNotInUse(nadCopy, pods); err != nil {
 		return &networkInUseError{err: err}
 	}
 
+	// Check if an egress firewall still exists in this namespace
```
npinaeva: I am not sure if this is better than just reconciling egressFirewalls on ~namespace~ UDN create. We already have that functionality on namespace log level change?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -414,9 +415,58 @@ func (oc *EFController) addEgressFirewall(egressFirewall *egressfirewallapi.Egre
...
+		cidrSelector = egressFirewallDestination.CIDRSelector
+		netInfo, err := oc.networkManager.GetActiveNetworkForNamespace(namespace)
+		if err != nil {
+			return "", "", nil, nil,
+				fmt.Errorf("failed to validate egress firewall destination: %w", err)
+		}
+		subnets := config.Default.ClusterSubnets
+		if !netInfo.IsDefault() {
+			subnets = netInfo.Subnets()
+		}
```
npinaeva: you can just do `netInfo.SUbnets() ` for default network too?

go-controller/pkg/util/egressfirewall.go
```diff
@@ -21,48 +19,25 @@ const (
 	dnsRegex = `^(\*\.)?([a-zA-Z0-9]([-a-zA-Z0-9]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z0-9]([-a-zA-Z0-9]{0,61}[a-zA-Z0-9])?\.?$`
 )
 
-// ValidateAndGetEgressFirewallDestination validates an egress firewall rule destination and returns
+// ValidateAndGetEgressFirewallDNSDestination validates an egress firewall rule destination and returns
```
npinaeva, line 22: why not move the whole function to egressfirewall?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -414,9 +415,58 @@ func (oc *EFController) addEgressFirewall(egressFirewall *egressfirewallapi.Egre
...
+		for _, clusterSubnet := range subnets {
+			if clusterSubnet.CIDR.Contains(ipNet.IP) || ipNet.Contains(clusterSubnet.CIDR.IP) {
+				clusterSubnetIntersection = append(clusterSubnetIntersection, clusterSubnet.CIDR)
```
npinaeva, line 560: did you mean to assign this value to the matchTarget?

go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
```diff
@@ -0,0 +1,715 @@
...
+					Type: egressfirewallapi.EgressFirewallRuleAllow,
+					To:   egressfirewallapi.EgressFirewallDestination{NodeSelector: &metav1.LabelSelector{MatchLabels: nodeLabel}},
+				},
+				id:  1,
+				err: false,
+				output: egressFirewallRule{
+					id:     1,
+					access: egressfirewallapi.EgressFirewallRuleAllow,
+					to:     destination{nodeAddrs: map[string][]string{node1Name: {node1Addr}}, nodeSelector: &metav1.LabelSelector{MatchLabels: nodeLabel}},
+				},
```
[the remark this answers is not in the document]
trozet: it does compile, because in go you can use shorthand composite literals when the type can be inferred cause the type is defined before it

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: no, I ended up not needing it. Will remove.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: Why does it matter which error gets propagated? Either way it means we need to retry. I can join them together for better logging if you want.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: will rename it. I thought about removing moveACLs... but decided not to because I'm not 100% confident downstream will upgrade in order and we might miss this in the upgrade path. Doesn't hurt to leave it for now and then remove it later (paranoia)

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: i don't want to assume

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: this was the code before, i'm not changing anything here

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: i added a comment on the constant

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: that would happen in the egress firewall controller...or maybe I'm not understanding your question

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: this isn't a per egress firewall metric right? i think its global for all egress firewall rules

go-controller/pkg/types/resource_status.go
trozet: that's kind of weird to have a file called resource_status.go and then define only error statuses in it. I'd rather have all the status messages defined in the same place. I'd rather just remove the comment.

trozet: also weird these other functions are defined in here if its only for error messages

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: it's actually not needed anymore. I'll remove it.

trozet: actually it is used by a unit test "OVN EgressFirewall Operations during execution egress firewall with node selector doesn't affect node handler"

will add comment and change the code to only RLock in case number of workers is ever increased

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
trozet: This doesn't have anything to do with namespace reconciliation. It's for UDN reconcilation.

go-controller/pkg/util/egressfirewall.go
trozet: cause its used somewhere else

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: yeah

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: the only case where we can get the wrong port group name by the namespace name is if the primary UDN in the namespace has changed, but in that case the old port group with all its ACLs will be (or I think should have already been) deleted on network sync anyway?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: I think that test should have been removed with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4388
Now node handler for EF is independent from the retry framework, and can't block other node events

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +289,104 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
-				pgName := oc.getNamespacePortGroupName(namespace)
-				// re-attach from ClusterPortGroupNameBase to namespaced port group.
-				// port group should exist, because namespace handler will create it.
-				ops, err = libovsdbops.AddACLsToPortGroupOps(oc.nbClient, ops, pgName, acls...)
-				if err != nil {
-					return fmt.Errorf("failed to build cleanup ops: %w", err)
-				}
+	// remove stale ACLs
+	p := libovsdbops.GetPredicate[*nbdb.ACL](oc.GetEgressFirewallACLDbIDsNoRule(namespace), nil)
+	fullPredicate := func(acl *nbdb.ACL) bool {
```
npinaeva: this function should go as the second argument to `GetPredicate` where you pass nil now, then you don't need an extra `if p(acl)` check

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +289,104 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
+		}
+		// delete case
+		klog.V(3).Infof("Removing egress firewall %s", key)
+		p := libovsdbops.GetPredicate[*nbdb.ACL](oc.GetEgressFirewallACLDbIDsNoRule(namespace), nil)
+		invalidACLs, err := libovsdbops.FindACLsWithPredicate(oc.nbClient, p)
 		if err != nil {
-			return fmt.Errorf("failed to remove egress firewall acls from node logical switches: %v", err)
+			return fmt.Errorf("error finding ACLs for egress firewall %s: %w", key, err)
+		}
+		if err := libovsdbops.DeleteACLsFromAllPortGroups(oc.nbClient, invalidACLs...); err != nil {
```
npinaeva: do you want to use the port group name here too?

go-controller/pkg/ovn/default_network_controller.go
npinaeva: can you do the same for all `egressfirewall2` imports?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: the previous code used internal structs that were already parsed, and you are now fetching it from the cache, and parsing again, which is not really needed, you can just change like this
``` 
for _, rule := range ef.egressRules
v
for _, rule := range egressFirewall.Spec.Egress
```
and like
```
rule.to.cidrSelector
v
rule.To.CIDRSelector
```

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: oh I missed that you sync on every node label change, not only matching selector. That is a bit worse for perf than before, but maybe fine for level-driven controller

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: lol that makes more sense then

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +289,104 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
+			return err
 		}
+		numRules, ok := oc.ruleCounter.Load(key)
+		if ok {
+			metrics.UpdateEgressFirewallRuleCount(float64(-numRules.(uint32)))
+			oc.ruleCounter.Delete(key)
+		} else {
+			klog.Errorf("Unable to decrement egress firewall rule count, cache miss for key: %s", key)
+		}
+		metrics.DecrementEgressFirewallCount()
```
npinaeva, line 427: I find it confusing that metrics are handled here for delete, but in `setEgressFirewallStatus` for update. Maybe we can update `setEgressFirewallStatus` to handle both and just tell it if was update or delete?

go-controller/pkg/types/resource_status.go
npinaeva: that file is used by the status manager, that needs to distinguish failure vs success.
This being a separate file shows that there is a cross-component dependency on the error messages and changing them requires an upgrade planning.
Success messages don't have this dependency and hence can be defined locally.
If you think this needs to be changed, I would postpone this to its own PR re-organizing all messages for al components and not just one

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
npinaeva: yep, I meant UDN create. When old UDN is deleted, the whole port group with all ACLs will be gone, so on new UDN create when namespace handler creates the port group, we could just call reconcile for egress firewall in this namespace?

go-controller/pkg/util/egressfirewall.go
npinaeva: so why don't you just pass the netInfo to the function as an argument and leave it here?
Splitting this function is somewhat confusing

test/e2e/egress_firewall.go
```diff
@@ -738,30 +809,45 @@ spec:
...
-			e2ekubectl.RunKubectlOrDie(f.Namespace.Name, "delete", "egressfirewall", "default")
+				framework.Logf("Deleting EgressFirewall in namespace %s", f.Namespace.Name)
+				e2ekubectl.RunKubectlOrDie(f.Namespace.Name, "delete", "egressfirewall", "default")
+			})
 		})
 	})
-})
+}
+
+func init() {
```
npinaeva, line 843: why don't you use `ginkgo.DescribeTable` for this?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: this is only run once at start up, it doesnt hurt anything, and provides broader coverage

go-controller/pkg/types/resource_status.go
trozet: i think it does need to be changed, but i'll postpone it

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
trozet: seems problematic. The UDN gets torn down and the egress firewall still shows status applied. We protect the UDN from being torn down when pods are still on it, I don't see why we wouldn't extend it to other namespace scoped objects. Who knows if even services or other features withstand this same scenario. I don't think we have proper testing. Support for this could always be added later, I'd rather play it safe for now.

go-controller/pkg/util/egressfirewall.go
trozet: It's not more confusing, it's better encapsulation. Returning 4-5 arguments and throwing most of them away + passing a netInfo when you dont need it cause you are only interested in parsing DNS is not a good approach. This way the DNS parsing is done in this function and then it can also be called by validateAndGetEgressFirewallDestination to do DNS parsing when parsing the entire egress firewall.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: it hurts startup time

go-controller/pkg/util/egressfirewall.go
npinaeva: the idea of running full EF validation for DNS names was to avoid creating DNS resources if the EF is not valid, because it won't be actually created by the EF controller.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: its like nanoseconds... i dont want to change this

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: we don't know the controller of the port group, so best i could do is build a predicate and delete from port group if libovsdops.ObjectNameKey matches the namespace name

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: setEgressFirewallStatus doesn't run on delete

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: done

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: i dont want to refactor this, the code does more than what you suggest. It is validating and setting up values. You can do it as a follow up if you want

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: im not sure what you mean. If a node label changes, we have to walk the egress firewalls to see if any had node selectors then ask for them to be reconciled. I could hold a cache map of ef -> nodes, but I dont want to do that.

If an egress firweall is updated to not match on node selector, then the egress firewall controller will handle detecting that change and updating it.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: tried this by adding delete with predicate:

```
func DeleteACLsFromPortGroupWithPredicateOps(nbClient libovsdbclient.Client, ops []ovsdb.Operation,
	p portGroupPredicate, acls ...*nbdb.ACL) ([]ovsdb.Operation, error) {
	if len(acls) == 0 {
		return ops, nil
	}

	deleted := []nbdb.PortGroup{}
	pg := nbdb.PortGroup{
		ACLs: make([]string, 0, len(acls)),
	}

	for _, acl := range acls {
		pg.ACLs = append(pg.ACLs, acl.UUID)
	}

	opModel := operationModel{
		ModelPredicate:   p,
		OnModelMutations: []interface{}{&pg.ACLs},
		ExistingResult:   &deleted,
		ErrNotFound:      false,
		BulkOp:           true,
	}

	m := newModelClient(nbClient)
	return m.DeleteOps(ops, opModel)
}

```

turns out it doesn't work because in libovsdb it doesn't respect the existingResults for onmodelMutations and only works on Model. Probably should be fixed, but not doing that now. The only other option is to predicate search and then delete as 2 separate operations so will do it that way for now

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
trozet: i think a good opportunity to make this better later will be with the dynamic UDN stuff. We can leverage the NAD tracker ->notifier there to execute notifications to independent controllers there so that all the controllers dont need to watch NADs too. Just something to think about for later

test/e2e/egress_firewall.go
trozet: cause the tests need to start with Network Segmentation, and putting it in a table will append the names

go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
```diff
@@ -0,0 +1,715 @@
...
+
+	config.Default.ClusterSubnets = []config.CIDRNetworkEntry{{CIDR: clusterSubnet}}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			if tc.dnsNameResolverEnabled {
+				config.OVNKubernetesFeature.EnableDNSNameResolver = true
+			}
+			netInfo, err := util.NewNetInfo(&ovncnitypes.NetConf{
+				NetConf:  cnitypes.NetConf{Name: validUDNName},
```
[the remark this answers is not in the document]
trozet: will fix

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: done

go-controller/pkg/util/egressfirewall.go
trozet: i don't think it makes sense to have DNS controller doing EF validation as well as EF controller doing it. If we want to block DNS controller from creating DNS entries then it should check if the egress firewall status is green.

trozet:
next push will be just to rebase

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: removed it

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +290,118 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
-			return fmt.Errorf("failed to remove egress firewall acls from node logical switches: %v", err)
+			return fmt.Errorf("error finding port groups for egress firewall namespace %s: %w", namespace, err)
+		}
+		pgNames := make([]string, len(pgs))
+		for _, pg := range pgs {
+			pgNames = append(pgNames, pg.Name)
+		}
+		if err := libovsdbops.DeleteACLsFromPortGroups(oc.nbClient, pgNames, invalidACLs...); err != nil {
+			return fmt.Errorf("error deleting stale ACLs for egress firewall %s: %w", key, err)
+		}
```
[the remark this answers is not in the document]
trozet: fixing

go-controller/pkg/util/egressfirewall_test.go
```diff
@@ -112,83 +105,26 @@ func TestValidateAndGetEgressFirewallDestination(t *testing.T) {
...
 	for _, tc := range testcases {
 		t.Run(tc.name, func(t *testing.T) {
 
 			if tc.dnsNameResolverEnabled {
 				config.OVNKubernetesFeature.EnableDNSNameResolver = true
 			}
 
-			cidrSelector, dnsName, clusterSubnetIntersection, nodeSelector, err :=
-				ValidateAndGetEgressFirewallDestination(tc.egressFirewallDestination)
+			dnsName, err :=
```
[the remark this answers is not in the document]
trozet: fixed

go-controller/pkg/ovn/namespace.go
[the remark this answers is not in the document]
trozet: it doesn't return an error?

// Reconcile queues the key to the egress firewall controller
func (oc *EFController) Reconcile(key string) {
	oc.controller.Reconcile(key)
}

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
ormergi: I think we should avoid such checks on deletion UDN deletion flow because they are racefull, as  EgressFirewall CR can be deleted on millisecond earlier before this check, or created one millisecond after this check. Nothing prevent this to happen. 

The same goes for the existing check for Pods, in retrospect it should have been checked by some webhook instead.

Please consider removing this check.

ormergi (commented):
Hi, I had one concern about UDN support, related to  extending UDN deletion flow, PTAL

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: > You can do it as a follow up if you want

I don't think this is a reasonable answer to a review comment

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: > we don't know the controller of the port group

port group is owned (and cleaned up) by the network controller. If the primary UDN ever changes, which is the only way for the target port group to change, old primary UDN will be cleaned up together with the port group and its acls, so it is safe to use current active network for the port group ID. 
Performance-wise there is not point to remove acls from the port group that will be deleted anyway, I am not sure what is the scenario in which this code helps?

npinaeva:
I think we need to decide what is the desired behaviour of this and future primary-network-controllers, should they all use finalizers and prevent deletion (which is racy and was only done for pods because pods can't be reasonably handled after a NAD deletion) or should we just reconcile on primary NAD create/delete.
Thinking about the expected status for EF in the namespace with no primary UDN (there also should be no pods in that case) I could argue that both "successfully applied" (no pods are selected, so doing nothing is kinda expected) vs "error: this namespace doesn't have primary network" (less sure about that, because having a namespace with no primary networks is a valid intermediate state) make sense.
Also if we will use finalizer, what do we do if the egress firewall is created before the primary UDN? Currently it will just fail to get a pg owner and fail, but then shouldn't it be the same in case of PUDN delete?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: we had cases where initial sync took up to 10 min due to the long predicate search, so that is not as fast as you think.
It is also related to https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5688#discussion_r2486329318, maybe we can figure out if this search can be replaced with pg name

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
trozet: I'm going to change it to track the NADs and i'll remove the commit

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -300,14 +301,63 @@ func (oc *EFController) initialSync() error {
...
+				oc.cache.UnlockKey(cachedFirewallName)
+				oc.controller.Reconcile(cachedFirewallName)
+			}
+		}
+		return
+	}
+	if info == nil {
+		return
+	}
+	for _, ns := range info.GetNADNamespaces() {
```
npinaeva: why do you need to iterate over all network's NADs? shouldn't `RegisterNADHandler` be about just 1 NAD?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -300,14 +301,63 @@ func (oc *EFController) initialSync() error {
...
+	if info != nil && !info.IsPrimaryNetwork() { // egressFirewall only supported for primary network
+		return
+	}
+	if len(removedNADName) > 0 { // delete case
+		namespace, _, err := cache.SplitMetaNamespaceKey(removedNADName)
+		if err != nil {
+			klog.Errorf("%s: failed splitting key %s: %v", oc.name, removedNADName, err)
+			return
+		}
+		cachedFirewallNames := oc.cache.GetKeys()
```
npinaeva: there is only 1 egress firewall named "default" in every namespace, so you could use namespace as a cache key, and then you don't need to iterate over the whole cache

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -72,75 +92,130 @@ type destination struct {
...
-	if err != nil {
-		return efr, err
+	controllerConfig := &controller.ControllerConfig[egressfirewallapi.EgressFirewall]{
+		RateLimiter: workqueue.DefaultTypedControllerRateLimiter[string](),
+		Informer:    efInformer.Informer(),
+		Lister:      efInformer.Lister().List,
+		MaxAttempts: controller.InfiniteAttempts,
+		Reconcile:   c.sync,
+		// Node changes can require reconcile, need to always reconcile. Should not be many updates to EF.
+		ObjNeedsUpdate: func(_, _ *egressfirewallapi.EgressFirewall) bool {
```
npinaeva: this one should be changed to ` old == nil || new == nil || !reflect.DeepEqual(old.Spec, new.Spec)` to avoid a full reconcile on every node's status condition update

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -188,120 +290,120 @@ func (oc *DefaultNetworkController) syncEgressFirewall(egressFirewalls []interfa
...
-	})
+	// add/update case
+	defer func() {
+		if statusErr := oc.setEgressFirewallStatus(ef, updateErr); statusErr != nil {
+			updateErr = utilerrors.Join(updateErr, fmt.Errorf("failed to update egress firewall status %s, error: %w",
+				GetEgressFirewallNamespacedName(ef), statusErr))
+		}
+	}()
+
+	pgName, err := oc.getNamespacePortGroupName(ef.Namespace)
```
npinaeva, line 442: do you want to handle `util.IsInvalidPrimaryNetworkError(err)` separately to stop retrying, since it won't be fixed until NAD update happens?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: I'm not sure there is any evidence to suggest this will take 10 minutes, even with 100s of port groups. Regardless, I cannot infer the port group name as it is generated by the primary ID key, which includes the network controller:
```
"k8s.ovn.org/id"="udn-test_l3-primary-network-controller:Namespace:udn-test"
```

I can however make the predicate search more narrow, by filtering based on the name and owner external id properties.

go-controller/pkg/networkmanager/fake.go
```diff
@@ -45,9 +47,32 @@ func (fcm *FakeControllerManager) Reconcile(_ string, _, _ util.NetInfo) error {
...
+	return nil
+}
+
+func (fnm *FakeNetworkManager) TriggerHandlers(removedNADName string, info util.NetInfo) {
+	fnm.Lock()
+	defer fnm.Unlock()
+	for _, h := range fnm.HandlerFuncs {
+		h(removedNADName, info)
+	}
+}
```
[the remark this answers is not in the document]
trozet: The callback functions should not try to access networkManager. NetworkManager has to call the handlers under lock. The callback function is fed the current network information for the affected NAD, so there should be no need to query the networkManager.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: done

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: because a subscriber might need to know about NADs that dont exist yet. For example in the egress firewall case, if a PUDN is deleted, then a new one is created, we wouldn't know about the new NAD and know to update egress firewall.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: I worry a bit about that since the callback does not retry upon error, even though the error probability of being able to list in the informer cache is pretty much 0. The rate limiter should throttle back retries on the object so I don't think it is worth changing this.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: done

trozet:
last push was just rebase to fix minor conflict

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: I thought NADHandler will be executed on every NAD create/update, so when a PUDN is deleted, and then a new one is created we should get 2 NAD syncs, one for delete and one for create? and both will call NADHandlers?

trozet: oh I read your question wrong. I thought you were saying why don't I just do a RegisterNADHandler per NAD. I don't really understand the question still, because this is iterating over nad namespaces. A CUDN can have many namespaces associated with it, and an egress firewall might be in each one.

npinaeva: I think i expected `handleNetworkEvent` to take NadName and event type like deleted/created, but instead you pass NadName for delete and netInfo for create. It looks like it is called on every NAD event, even if NADs are a part of the same CUDN you will get an event for each one of them?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: yeah that works
took me some time to dig up the evidence, here https://github.com/ovn-kubernetes/ovn-kubernetes/issues/3431#issuecomment-1440473091 there were more examples in the d/s Jira

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: I'll modify the callback to always pass NADName and event type, so then egress firewall can just reconcile with that. But in the future, other controllers that register for the callback will need the netInfo, so will keep that as well.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -0,0 +1,1238 @@
...
+			klog.Errorf("%s: failed splitting key %s: %v", oc.name, removedNADName, err)
+			return
+		}
+		oc.cache.LockKey(namespace)
+		entry, ok := oc.cache.Load(namespace)
+		if !ok {
+			return // no cache entry exists, so nothing to remove
+		}
+		entry.stale = true
+		oc.cache.UnlockKey(namespace)
```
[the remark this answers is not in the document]
trozet: fixed

PR #5691, opened 2025-10-27
OCPBUGS-63160: Fix named port handling in ETP=local services with host-networked pods

andreaskaris:
When a service has a named targetPort, ExternalTrafficPolicy=local, and host-networked backend pods, ovn-kubernetes incorrectly uses the named port string directly in OpenFlow rules instead of resolving it to the numeric port. This causes ovs-ofctl to fail with errors like:
  "ovs-ofctl: -:5: invalid nat range "172.20.0.3:diameterstack""

This is a follow-up to commit "Fix named port handling in externalTrafficPolicy Local services" which fixed the same issue for iptables rules when allocateLoadBalancerNodePorts=false.

The fix passes localEndpoints to updateServiceFlowCache() and createLbAndExternalSvcFlows() to properly resolve named ports to their numeric values before constructing OpenFlow rules. The resolved port is retrieved from localEndpoints[svcPortKey].Port instead of using svcPort.TargetPort.String() which returns the unresolved name.

Also adds validation to ensure the gateway IPs exist in localEndpoints before constructing the NAT rules.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for LoadBalancer with ExternalTrafficPolicy=Local including host‑networked pods, named ports, external IPs, NAT/iptables/OpenFlow validation.
  * Expanded e2e service matrix to exercise named vs numeric ports, ExternalTrafficPolicy (Cluster/Local), hostNetwork, TCP/UDP, dual‑stack, NAT and MTU scenarios.
  * Updated test expectations to reflect revised NodePort return‑flow identifiers.

* **Improvements**
  * Endpoint‑aware flow generation and a new safe endpoint‑lookup API for correct per‑flow target‑port resolution and return‑traffic handling.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -323,22 +323,26 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
 					klog.V(5).Infof("Adding flows on breth0 for Nodeport Service %s with targetPort %d in "+
 						"Namespace: %s since ExternalTrafficPolicy=local", service.Name, targetPort, service.Namespace)
 					// table 0, This rule matches on all traffic with dst port == NodePort, DNAT's the nodePort to the svc targetPort
-					// If ipv6 make sure to choose the ipv6 node address for rule
+					// If IPv6 make sure to choose the IPv6 node address for rule, otherwise send to the IPv4 node address.
+					gatewayAddress := npw.gatewayIPv4
 					if strings.Contains(flowProtocol, "6") {
-						// Regardless of the IP addresses in the local endpoints list, we always send to npw.gatewayIPv6.
-						nodeportFlows = append(nodeportFlows,
-							fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, tp_dst=%d, "+
-								"actions=ct(commit,zone=%d,nat(dst=[%s]:%d),table=6)",
-								cookie, npw.ofportPhys, flowProtocol, svcPort.NodePort,
-								config.Default.HostNodePortConntrackZone, npw.gatewayIPv6, targetPort))
-					} else {
-						// Regardless of the IP addresses in the local endpoints list, we always send to npw.gatewayIPv4.
-						nodeportFlows = append(nodeportFlows,
-							fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, tp_dst=%d, "+
-								"actions=ct(commit,zone=%d,nat(dst=%s:%d),table=6)",
-								cookie, npw.ofportPhys, flowProtocol, svcPort.NodePort,
-								config.Default.HostNodePortConntrackZone, npw.gatewayIPv4, targetPort))
+						gatewayAddress = fmt.Sprintf("[%s]", npw.gatewayIPv6)
 					}
+					// This is just an additional safety check - we should never get into this situation.
+					// If gatewayAddress is empty, we cannot install the following flow and we will continue with the
+					// next flowProtocol.
+					if gatewayAddress == "" || gatewayAddress == "[]" {
+						err := fmt.Errorf("gateway IP not available for protocol %s", flowProtocol)
+						errors = append(errors, err)
+						continue
+					}
+
+					nodeportFlows = append(nodeportFlows,
+						fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, tp_dst=%d, "+
+							"actions=ct(commit,zone=%d,nat(dst=%s:%d),table=6)",
+							cookie, npw.ofportPhys, flowProtocol, svcPort.NodePort,
+							config.Default.HostNodePortConntrackZone, gatewayAddress, targetPort))
+
 					nodeportFlows = append(nodeportFlows,
 						// table 6, Sends the packet to the host. Note that the constant etp svc cookie is used since this flow would be
 						// same for all such services.
@@ -554,22 +558,26 @@ func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service,
 			// case1 (see function description for details)
 			klog.V(5).Infof("Adding flows on breth0 for %s Service %s in Namespace: %s since ExternalTrafficPolicy=local", ipType, service.Name, service.Namespace)
 			// table 0, This rule matches on all traffic with dst ip == LoadbalancerIP / externalIP, DNAT's the nodePort to the svc targetPort
-			// If ipv6 make sure to choose the ipv6 node address for rule
+			// If IPv6 make sure to choose the IPv6 node address for rule, otherwise send to the IPv4 node address.
+			gatewayAddress := npw.gatewayIPv4
 			if strings.Contains(flowProtocol, "6") {
-				// Regardless of the IP addresses in the local endpoints list, we always send to npw.gatewayIPv6.
-				externalIPFlows = append(externalIPFlows,
-					fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, %s=%s, tp_dst=%d, "+
-						"actions=ct(commit,zone=%d,nat(dst=[%s]:%d),table=6)",
-						cookie, npw.ofportPhys, flowProtocol, nwDst, externalIPOrLBIngressIP, svcPort.Port,
-						config.Default.HostNodePortConntrackZone, npw.gatewayIPv6, targetPort))
-			} else {
-				// Regardless of the IP addresses in the local endpoints list, we always send to npw.gatewayIPv4.
-				externalIPFlows = append(externalIPFlows,
-					fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, %s=%s, tp_dst=%d, "+
-						"actions=ct(commit,zone=%d,nat(dst=%s:%d),table=6)",
-						cookie, npw.ofportPhys, flowProtocol, nwDst, externalIPOrLBIngressIP, svcPort.Port,
-						config.Default.HostNodePortConntrackZone, npw.gatewayIPv4, targetPort))
+				gatewayAddress = fmt.Sprintf("[%s]", npw.gatewayIPv6)
 			}
+			// This is just an additional safety check - we should never get into this situation.
+			// If gatewayAddress is empty, we cannot install the following flow and we will continue with the
+			// next flowProtocol.
+			if gatewayAddress == "" || gatewayAddress == "[]" {
+				err := fmt.Errorf("gateway IP not available for protocol %s", flowProtocol)
+				errors = append(errors, err)
+				continue
+			}
+
+			externalIPFlows = append(externalIPFlows,
+				fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, %s=%s, tp_dst=%d, "+
+					"actions=ct(commit,zone=%d,nat(dst=%s:%d),table=6)",
+					cookie, npw.ofportPhys, flowProtocol, nwDst, externalIPOrLBIngressIP, svcPort.Port,
+					config.Default.HostNodePortConntrackZone, gatewayAddress, targetPort))
+
 			externalIPFlows = append(externalIPFlows,
 				// table 6, Sends the packet to Host. Note that the constant etp svc cookie is used since this flow would be
 				// same for all such services.


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -505,27 +517,36 @@ func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service,
...
 			}
 			externalIPFlows = append(externalIPFlows,
 				// table 6, Sends the packet to Host. Note that the constant etp svc cookie is used since this flow would be
 				// same for all such services.
 				fmt.Sprintf("cookie=%s, priority=110, table=6, actions=output:LOCAL",
 					etpSvcOpenFlowCookie),
 				// table 0, Matches on return traffic, i.e traffic coming from the host networked pod's port, and unDNATs
-				fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%s, actions=ct(commit,zone=%d nat,table=7)",
-					cookie, flowProtocol, svcPort.TargetPort.String(), config.Default.HostNodePortConntrackZone),
+				fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%d, actions=ct(commit,zone=%d nat,table=7)",
```
andreaskaris: Note that there always was unpredictable behavior here as the same flow will be added with different cookies, yielding:
```
sh-5.2# ovs-ofctl dump-flows breth0 | grep 0xbe3a856277f3fbfc
 cookie=0xbe3a856277f3fbfc, duration=6.126s, table=0, n_packets=0, n_bytes=0, idle_age=7101, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(commit,table=7,zone=64003,nat)
 cookie=0xbe3a856277f3fbfc, duration=7101.164s, table=0, n_packets=0, n_bytes=0, idle_age=7101, priority=110,arp,in_port=1,arp_tpa=192.168.10.0,arp_op=1 actions=LOCAL
 cookie=0xbe3a856277f3fbfc, duration=7101.164s, table=0, n_packets=0, n_bytes=0, idle_age=7101, priority=110,tcp,in_port=1,nw_dst=192.168.10.0,tp_dst=8080 actions=ct(commit,table=6,zone=64003,nat(dst=172.18.0.3:18888))
sh-5.2#  ovs-ofctl dump-flows breth0 --names | grep 0xc3e675a338eae452
 cookie=0xc3e675a338eae452, duration=7102.120s, table=0, n_packets=0, n_bytes=0, priority=110,arp,in_port=eth0,arp_tpa=1.1.1.1,arp_op=1 actions=LOCAL
 cookie=0xc3e675a338eae452, duration=7102.120s, table=0, n_packets=0, n_bytes=0, priority=110,tcp,in_port=eth0,nw_dst=1.1.1.1,tp_dst=8080 actions=ct(commit,table=6,zone=64003,nat(dst=172.18.0.3:18888))
sh-5.2#  ovs-ofctl dump-flows breth0 --names | grep 0xc3e675a338eae452
 cookie=0xc3e675a338eae452, duration=15.053s, table=0, n_packets=0, n_bytes=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(commit,table=7,zone=64003,nat)
 cookie=0xc3e675a338eae452, duration=7365.087s, table=0, n_packets=0, n_bytes=0, priority=110,arp,in_port=eth0,arp_tpa=1.1.1.1,arp_op=1 actions=LOCAL
 cookie=0xc3e675a338eae452, duration=7365.087s, table=0, n_packets=0, n_bytes=0, priority=110,tcp,in_port=eth0,nw_dst=1.1.1.1,tp_dst=8080 actions=ct(commit,table=6,zone=64003,nat(dst=172.18.0.3:18888))
```

andreaskaris: ~~I'll try to address this here; otherwise I'll create a bug for it - either way it's low prio~~

Fixed

andreaskaris: Even with the current fix in place, there's another issue. However, any resolution to this issue will actually have an impact on the logic:
```
[root@ovnkubernetes ~]# for i in {1..20}; do   echo "Loop $i";   kubectl exec -n ovn-kubernetes ovs-node-bp6xw -- ovs-ofctl dump-flows breth0 | awk '{gsub(/duration=[^,]*, /, "");gsub(/n_packets=[^,]*, /, "");gsub(/n_bytes=[^,]*, /, "");gsub(/idle_age=[^,]*, /, "");gsub(/hard_age=[^,]*, /, "");print}' > /tmp/a;   sleep 60;   kubectl exec -n ovn-kubernetes ovs-node-bp6xw -- ovs-ofctl dump-flows breth0 | awk '{gsub(/duration=[^,]*, /, "");gsub(/n_packets=[^,]*, /, "");gsub(/n_bytes=[^,]*, /, "");gsub(/idle_age=[^,]*, /, "");gsub(/hard_age=[^,]*, /, "");print}' > /tmp/b;   if ! diff /tmp/a /tmp/b; then     break;   fi; done
Loop 1
Loop 2
Loop 3
Loop 4
Loop 5
14c14
<  cookie=0x3a230dd087c740b7, table=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(commit,table=7,zone=64003,nat)
---
>  cookie=0xf32eff67f946f651, table=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(table=7,zone=64003,nat)
[root@ovnkubernetes ~]# 
```

https://man7.org/linux/man-pages/man7/ovs-actions.7.html
```
The ct action
       Syntax:
              ct([argument]...)
              ct(commit[,argument]...)

       The action has two modes of operation, distinguished by whether
       commit is present.  The following arguments may be present in
       either mode:

          zone=value
                 A zone is a 16-bit id that isolates connections into
                 separate domains, allowing overlapping network addresses
                 in different zones.  If a zone is not provided, then the
                 default is 0. The value may be specified either as a
                 16-bit integer literal or a field or subfield in the
                 syntax described under Field Specifications above.

       Without commit, this action sends the packet through the
       connection tracker.  The connection tracker keeps track of the
       state of TCP connections for packets passed through it.  For each
       packet through a connection, it checks that it satisfies TCP
       invariants and signals the connection state to later actions using
       the ct_state metadata field, which is documented in [ovs-fields(7)](https://man7.org/linux/man-pages/man7/ovs-fields.7.html).
```

andreaskaris: The issue is with the same lines, though. From upstream/master:
```
func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netInfo util.NetInfo, add, hasLocalHostNetworkEp bool) error {
(...)
						// table 0, Matches on return traffic, i.e traffic coming from the host networked pod's port, and unDNATs
						fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%s, actions=ct(zone=%d nat,table=7)",
							cookie, flowProtocol, svcPort.TargetPort.String(), config.Default.HostNodePortConntrackZone),
```

```
func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service, netConfig *bridgeconfig.BridgeUDNConfiguration, svcPort *corev1.ServicePort, add bool,
(...)
				// table 0, Matches on return traffic, i.e traffic coming from the host networked pod's port, and unDNATs
				fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%s, actions=ct(commit,zone=%d nat,table=7)",
					cookie, flowProtocol, svcPort.TargetPort.String(), config.Default.HostNodePortConntrackZone),
```

I guess that ovn / openflow only accept one of w/ or w/o commit.

andreaskaris: For example, with upstream/master, let's use this deployment:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: my-dep2
  name: my-dep2
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 3
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: my-dep2
  strategy:
  template:
    metadata:
      creationTimestamp: null
      labels:
        app: my-dep2
    spec:
      hostNetwork: true
      containers:
      - image: nginx
        imagePullPolicy: Always
        name: nginx
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: my-dep2
  name: my-dep2
  namespace: default
spec:
  # allocateLoadBalancerNodePorts: false
  ipFamilies:
  - IPv4
  # - IPv6
  # ipFamilyPolicy: RequireDualStack
  externalTrafficPolicy: Local
  externalIPs:
    - 1.1.1.1
  ports:
  - name: diameterstack-port
    port: 8080
    protocol: TCP
    targetPort: 18888
  selector:
    app: my-dep2
  type: LoadBalancer
```

Then, let's run:
```
for i in {1..20}; do   echo "Loop $i";   kubectl exec -n ovn-kubernetes ovs-node-t8jwx -- ovs-ofctl dump-flows breth0 | awk '{gsub(/duration=[^,]*, /, "");gsub(/n_packets=[^,]*, /, "");gsub(/n_bytes=[^,]*, /, "");gsub(/idle_age=[^,]*, /, "");gsub(/hard_age=[^,]*, /, "");print}' > /tmp/a$i;   sleep 60;   kubectl exec -n ovn-kubernetes ovs-node-t8jwx -- ovs-ofctl dump-flows breth0 | awk '{gsub(/duration=[^,]*, /, "");gsub(/n_packets=[^,]*, /, "");gsub(/n_bytes=[^,]*, /, "");gsub(/idle_age=[^,]*, /, "");gsub(/hard_age=[^,]*, /, "");print}' > /tmp/b$i;   diff /tmp/a$i /tmp/b$i; done
```

We see 2 issues:
a) Fixed here in the PR, the conflict between the ExternalIP and the LB status IP:
```
Loop 1
14c14
<  cookie=0xd849e0d1ccf7cf5e, table=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(commit,table=7,zone=64003,nat)
---
>  cookie=0x3f4efe56ff0c2158, table=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(commit,table=7,zone=64003,nat)
```

b) Conflict between flow from updateServiceFlowCache without commit and  createLbAndExternalSvcFlows with commit.
```
Loop 3
14c14
<  cookie=0x3f4efe56ff0c2158, table=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(commit,table=7,zone=64003,nat)
---
>  cookie=0x8c1f56e1e4501579, table=0, priority=110,tcp,in_port=LOCAL,tp_src=18888 actions=ct(table=7,zone=64003,nat)
```

For b), I'd need to know if the flow needs to be `commit`, or without it ... @ricky-rav any idea?

test/e2e/service.go
```diff
@@ -231,322 +232,360 @@ var _ = ginkgo.Describe("Services", feature.Service, func() {
...
+									for _, ovnKubeNodePod := range ovnKubeNodePods.Items {
+										framework.Logf("Flushing the ip route cache on %s", ovnKubeNodePod.Name)
+										containerName := "ovnkube-node"
+										if isInterconnectEnabled() {
+											containerName = "ovnkube-controller"
+										}
+										_, err := e2ekubectl.RunKubectl(ovnKubernetesNamespace, "exec", ovnKubeNodePod.Name, "--container", containerName, "--",
+											"ip", "route", "flush", "cache")
+										framework.ExpectNoError(err, "Flushing the ip route cache failed")
 									}
```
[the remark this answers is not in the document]
andreaskaris: false positive

test/e2e/service.go
```diff
@@ -231,322 +232,360 @@ var _ = ginkgo.Describe("Services", feature.Service, func() {
...
-					if !hostNetwork || isLocalGWModeEnabled() {
-						packetSizes = []string{"small"}
-					}
+						e2epod.NewPodClient(f).CreateSync(context.TODO(), clientPod)
+
+						ginkgo.By(fmt.Sprintf("Creating the server pod with hostNetwork:%t", hostNetwork))
+						// Create the server pod.
+						// Wait for 1 minute and if the pod does not come up, select a different port and try again.
+						// Wait for a max of 5 minutes.
+						serverPodPortTCP := infraprovider.Get().GetK8HostPort() // maybe a host net or cluster net pod but select host port anyway
```
andreaskaris: So I didn't want to change existing code too much, but this here is a bug, right? It's setting the port _outside_ of the Eventually ... I'm inclined to move these 2 lines _into_ the eventually in order to fix this ..

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -300,28 +301,44 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
...
+
 					// case1 (see function description for details)
 					var nodeportFlows []string
-					klog.V(5).Infof("Adding flows on breth0 for Nodeport Service %s in Namespace: %s since ExternalTrafficPolicy=local", service.Name, service.Namespace)
+					klog.V(5).Infof("Adding flows on breth0 for Nodeport Service %s with targetPort %d in "+
+						"Namespace: %s since ExternalTrafficPolicy=local", service.Name, targetPort, service.Namespace)
 					// table 0, This rule matches on all traffic with dst port == NodePort, DNAT's the nodePort to the svc targetPort
 					// If ipv6 make sure to choose the ipv6 node address for rule
 					if strings.Contains(flowProtocol, "6") {
+						// Regardless of the IP addresses in the local endpoints list, we always send to npw.gatewayIPv6.
```
ricky-rav: This comment alarmed me quite a bit. Can we rephrase it? :-D   
Service ports are orthogonal to the IP family in use and localEndpoints just includes the port information for our practical purposes. Is this what you wanted to convey? :)

test/e2e/service.go
ricky-rav: Ah yes, you're right. Service creation might fail if the port is already in use, so it seems to be a good idea to retry this within the Eventually block

```
	// GetK8HostPort returns a Node port. Requesting a port that maybe exposed in tests to avoid multiple parallel
	// tests utilizing conflicting ports. It also allows infra provider implementations to set Nodes
	// allowed port range and therefore comply with cloud provider firewall rules.
	GetK8HostPort() uint16 // supported K8 host ports
```

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -532,6 +532,12 @@ func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service,
 				continue
 			}
 			targetPort := lbe.Port
+			targetPortCookie, err := svcToCookie(service.Namespace, service.Name, "", targetPort)
```
ricky-rav: Let's add a comment here to specify why we're doing this. It took me a while to understand what was going on :)

go-controller/pkg/node/gateway_shared_intf.go
ricky-rav: You opened a pandora's box, @andreaskaris :-D
So return traffic should not be committed to conntrack, it must have been overlooked until now :) 

How about we remove `commit` in that flow in `createLbAndExternalSvcFlows` and we use `targetPortCookie` in `updateServiceFlowCache` as well?

go-controller/pkg/node/gateway_shared_intf.go
andreaskaris: done

go-controller/pkg/node/gateway_shared_intf.go
andreaskaris: done

test/e2e/service.go
andreaskaris: done

go-controller/pkg/node/gateway_shared_intf.go
andreaskaris: I removed it. I also got rid of the duplication with:
```
					gatewayAddress := npw.gatewayIPv4
					if strings.Contains(flowProtocol, "6") {
						gatewayAddress = fmt.Sprintf("[%s]", npw.gatewayIPv6)
					}
					nodeportFlows = append(nodeportFlows,
						fmt.Sprintf("cookie=%s, priority=110, in_port=%s, %s, tp_dst=%d, "+
							"actions=ct(commit,zone=%d,nat(dst=%s:%d),table=6)",
							cookie, npw.ofportPhys, flowProtocol, svcPort.NodePort,
							config.Default.HostNodePortConntrackZone, gatewayAddress, targetPort))
```

test/e2e/service.go
```diff
@@ -231,322 +232,361 @@ var _ = ginkgo.Describe("Services", feature.Service, func() {
...
-						// bit is not set. However, the decision on setting DF bit is left up to the kernel, and
-						// is unpredictable. If the DF bit is set, the iptables rule that DNATs nodeport -> cluster IP
-						// will then attempt to route the packet, and hit our 1400 byte MTU route. This will cause:
-						// 172.18.0.2:37755->10.96.141.254:9881(udp) sk_skb_reason_drop(SKB_DROP_REASON_PKT_TOO_BIG)
-						packetSizes = []string{"small"}
+		for _, namedPort := range []bool{true, false} {
+			for _, etp := range []v1.ServiceExternalTrafficPolicy{v1.ServiceExternalTrafficPolicyCluster, v1.ServiceExternalTrafficPolicyLocal} {
+				hostNetwork := hostNetwork
+				ginkgo.When(fmt.Sprintf("a nodePort service targeting a pod with hostNetwork:%t, namedPort:%t, ETP:%s is created",
+					hostNetwork, namedPort, etp), func() {
```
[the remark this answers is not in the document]
andreaskaris: I removed the hostNetwork := hostNetwork rebind, this isn't necessary any more in newer go versions

go-controller/pkg/node/gateway_shared_intf.go
andreaskaris: I retested manually with:
```
# pod=$(oc get pods -n ovn-kubernetes -o wide | grep ovs-node | grep ovn-control-plane | awk '{print $1}'); for i in {1..20}; do   echo "Loop $i";   kubectl exec -n ovn-kubernetes $pod -- ovs-ofctl dump-flows breth0 | awk '{gsub(/duration=[^,]*, /, "");gsub(/n_packets=[^,]*, /, "");gsub(/n_bytes=[^,]*, /, "");gsub(/idle_age=[^,]*, /, "");gsub(/hard_age=[^,]*, /, "");print}' > /tmp/a$i;   sleep 60;   kubectl exec -n ovn-kubernetes $pod -- ovs-ofctl dump-flows breth0 | awk '{gsub(/duration=[^,]*, /, "");gsub(/n_packets=[^,]*, /, "");gsub(/n_bytes=[^,]*, /, "");gsub(/idle_age=[^,]*, /, "");gsub(/hard_age=[^,]*, /, "");print}' > /tmp/b$i;   diff /tmp/a$i /tmp/b$i; done
Loop 1
Loop 2
Loop 3
Loop 4
Loop 5
Loop 6
Loop 7
Loop 8
Loop 9
Loop 10
Loop 11
Loop 12
Loop 13
Loop 14
Loop 15
Loop 16
Loop 17
Loop 18
Loop 19
Loop 20
```

And it's stable, now

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -300,28 +301,58 @@ func (npw *nodePortWatcher) updateServiceFlowCache(service *corev1.Service, netI
...
+
 					nodeportFlows = append(nodeportFlows,
 						// table 6, Sends the packet to the host. Note that the constant etp svc cookie is used since this flow would be
 						// same for all such services.
 						fmt.Sprintf("cookie=%s, priority=110, table=6, actions=output:LOCAL",
 							etpSvcOpenFlowCookie),
 						// table 0, Matches on return traffic, i.e traffic coming from the host networked pod's port, and unDNATs
-						fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%s, actions=ct(zone=%d nat,table=7)",
-							cookie, flowProtocol, svcPort.TargetPort.String(), config.Default.HostNodePortConntrackZone),
+						// Return traffic should not be committed to conntrack.
```
ricky-rav: Here you've correctly removed `commit` and added a comment about it, but once it's gone, we don't know why you're talking about committing to conntrack :)
I think you can do it in a separate commit and have the commit message say `Return traffic should not be committed to conntrack.` .

andreaskaris: done

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -577,7 +577,7 @@ func (npw *nodePortWatcher) createLbAndExternalSvcFlows(service *corev1.Service,
 					etpSvcOpenFlowCookie),
 				// table 0, Matches on return traffic, i.e traffic coming from the host networked pod's port, and unDNATs.
 				// Use targetPortCookie, as the flow will be the same for each ExternalIP / LB status IP.
-				fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%d, actions=ct(commit,zone=%d nat,table=7)",
+				fmt.Sprintf("cookie=%s, priority=110, in_port=LOCAL, %s, tp_src=%d, actions=ct(zone=%d nat,table=7)",
```
trozet, line 588: just a note that by doing this, if the host network pod initiated egress traffic using the port as the source port, it would not be committed, then return traffic would be attempted to be committed and be ct invalid, but I guess the traffic would still forward to LOCAL so guess this is OK.

trozet (approved):

PR #5694, opened 2025-10-27
no nft/ipt operation on dpu node

cathy-zhou:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Refactor**
  * Optimized network setup and cleanup operations by conditionally skipping certain infrastructure tasks in Data Processing Unit (DPU) mode.
  * Strengthened defensive checks and early returns in gateway initialization routines.
  * Consolidated NodePort forwarding rule management logic to ensure consistent handling of service networks.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1618,20 +1618,20 @@ func newNodePortWatcher(
 				return nil, fmt.Errorf("unable to configure UDN isolation nftables: %w", err)
 			}
 		}
-	}
 
-	var subnets []*net.IPNet
-	for _, subnet := range config.Default.ClusterSubnets {
-		subnets = append(subnets, subnet.CIDR)
-	}
-	subnets = append(subnets, config.Kubernetes.ServiceCIDRs...)
-	if config.Gateway.DisableForwarding {
-		if err := initExternalBridgeServiceForwardingRules(subnets); err != nil {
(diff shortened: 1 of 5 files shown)


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1618,20 +1618,20 @@ func newNodePortWatcher(
...
 
-	var subnets []*net.IPNet
-	for _, subnet := range config.Default.ClusterSubnets {
-		subnets = append(subnets, subnet.CIDR)
-	}
-	subnets = append(subnets, config.Kubernetes.ServiceCIDRs...)
-	if config.Gateway.DisableForwarding {
-		if err := initExternalBridgeServiceForwardingRules(subnets); err != nil {
-			return nil, fmt.Errorf("failed to add accept rules in forwarding table for bridge %s: err %v", gwBridge.GetGatewayIface(), err)
+		var subnets []*net.IPNet
```
trozet, line 1622: shouldn't this and all the stuff above it be conditional on NodeModeFull or DPUHost? Don't we also need those iptables/nftables rules for dpu host mode?

cathy-zhou: newNodePortWatcher() is only call in full mode or dpu mode. In dpu-host mode,  newNodePortWatcherIptables() is called instead

PR #5699, opened 2025-11-02
Fix RouteAdvertisements targetVRF default matching

### LINKED ISSUE #5580: Importing routes using FRR-k8s to CUDN doesn't work as expcted
### What happened?

The following example in BGP docs doesnt work https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/features/bgp-integration/route-advertisements.md#import-routes-into-a-cudn

It seems the FRR wont work as expected result in no connectivity from a pod on the CUDN to the service running in the external network.

The next example for importing routes to default VRF and leaking them to the CUDN VRF work, and pod on CUDN can reach services on the external network.

Although we see the traffic originating from the CUDN pod is SNAT'ed instead of being routed utilizing BGP infra:
```
root@ovn-worker2:/# tcpdump -i eth0 -eennvv dst 172.26.0.3 or src 172.26.0.3
tcpdump: listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
13:15:03.285424 96:26:e3:a1:2f:cf > ba:a1:f6:f0:84:a1, ethertype IPv4 (0x0800), length 98: (tos 0x0, ttl 62, id 24226, offset 0, flags [DF], proto ICMP (1), length 84)
    192.89.0.2 > 172.26.0.3: ICMP echo request, id 51584, seq 0, length 64
13:15:04.285003 96:26:e3:a1:2f:cf > ba:a1:f6:f0:84:a1, ethertype IPv4 (0x0800), length 98: (tos 0x0, ttl 62, id 24247, offset 0, flags [DF], proto ICMP (1), length 84)
    192.89.0.2 > 172.26.0.3: ICMP echo request, id 51584, seq 1, length 64
...
```
`192.89.0.2` - the SNAT'ed CUDN pod IP, the orginal pod IP was `22.100.2.3`
`172.26.0.3` - service running on external network 

This doesnt align with docs saying routes would be utilized for egress traffic.

### What did you expect to happen?

The mentioned example to work,having pods on CUDN reach service running on the external network over BGP.

### How can we reproduce it (as minimally and precisely as possible)?

See bellow reproduced for the mentioned [example](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/features/bgp-integration/route-advertisements.md#import-routes-into-a-cudn)
<details>

```bash
#! /bin/bash 

ns_list=(blue red)

frr_ip="192.89.0.6"
external_cidr="172.26.0.0/16"
external_svc_ip="172.26.0.3"

for n in ${ns_list[@]}; do
  echo "creating namespaces.."
   cat <<EOF | kubectl create -f -
---
apiVersion: v1
kind: Namespace
metadata:
  name: $n
  labels:
    k8s.ovn.org/primary-user-defined-network: ""
    network: extranet
EOF
done

echo "creating cudn.."
cat <<EOF | kubectl create -f -
---
apiVersion: k8s.ovn.org/v1
metadata:
  name: extranet
  labels:
    advertise: "true"
spec:
  namespaceSelector:
    matchLabels:
      network: extranet
  network:
    topology: Layer3
    layer3:
      role: Primary
      subnets:
      - cidr: "22.100.0.0/16"
        hostSubnet: 24
EOF

echo "creating frr-config.."
cat <<EOF | kubectl create -n frr-k8s-system -f -
apiVersion: frrk8s.metallb.io/v1beta1
kind: FRRConfiguration
metadata:
  labels:
    use-for-advertisements: extranet
  name: receive-filtered-extranet
  namespace: frr-k8s-system
spec:
  nodeSelector: {}
  bgp:
    routers:
    - asn: 64512
      neighbors:
      - address: ${frr_ip}
        asn: 64512
        disableMP: true
        toReceive:
          allowed:
            mode: filtered
            prefixes:
            - prefix: ${external_cidr}
      vrf: extranet
EOF

echo "creating pod.."
first_ns=${ns_list[0]}
cat <<EOF | kubectl -n ${first_ns} create  -f -
apiVersion: v1
kind: Pod
metadata:
  name:  p1
spec:
  containers:
  - name: test
    args: [pause]
    image: registry.k8s.io/e2e-test-images/agnhost:2.52
  terminationGracePeriodSeconds: 0
EOF
  
echo "waiting for pod readiness.."
kubectl -n ${first_ns} wait pod/p1 --for condition=Ready

echo "conn check to external service outside cluster network.."
kubectl -n ${first_ns} exec p1 -- ping -w 7 ${external_svc_ip}

[ "$?" != 0 ] && echo "Connectivity check from CUDN pod to external service failed" && exit 1
```

</details>

_No response_

### OVN-Kubernetes version

<details>

```console
$ ovnkube --version
# paste output here
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### OVN version

<details>

```console
$ oc rsh -n ovn-kubernetes ovnkube-node-xxxxx (pick any ovnkube-node pod on your cluster)
$ rpm -q ovn
# paste output here
```

</details>

### OVS version

<details>

```console
$ oc rsh -n ovn-kubernetes ovs-node-xxxxx (pick any ovs pod on your cluster)
$ rpm -q openvswitch
# paste output here
```

</details>

### Platform

<details>

Is it baremetal? GCP? AWS? Azure?
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

yboaron:
## 📑 Description

Fixes RouteAdvertisements not matching FRRConfiguration when `targetVRF` is set to `"default"`.

  When `targetVRF` is set to `"default"`, it should match FRRConfiguration routers with empty VRF string (`""`),
  which is how FRR represents the default VRF. Previously, the comparison `"default" != ""` would fail, causing
  the router to be skipped and preventing route leaking between the default VRF and user-defined network VRFs.

Fixes #5580

## ✅ Checks

- [X] My code requires tests
- [X] if so, I have added and/or updated the tests as required

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed RouteAdvertisement handling to consistently treat "default" VRF the same as empty VRF, ensuring proper FRR router configuration and prefix advertisement in default virtual routing scenarios.

### DIFF
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller_test.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller_test.go
@@ -943,6 +943,43 @@ func TestController_reconcile(t *testing.T) {
 			reconcile:            "ra",
 			expectAcceptedStatus: metav1.ConditionFalse,
 		},
+		{
+			name: "reconciles pod RouteAdvertisement with targetVRF 'default' matching router with empty VRF string",
+			ra:   &testRA{Name: "ra", TargetVRF: "default", AdvertisePods: true, NetworkSelector: map[string]string{"selected": "true"}},
+			frrConfigs: []*testFRRConfig{
+				{
+					Name:      "frrConfig",
+					Namespace: frrNamespace,
+					Routers: []*testRouter{
+						// Router with empty VRF (default VRF in FRR)
+						{ASN: 1, Prefixes: []string{"1.1.1.0/24"}, Neighbors: []*testNeighbor{
+							{ASN: 1, Address: "1.0.0.100"},
+						}},
+					},
+				},
+			},
+			nads: []*testNAD{
+				{Name: "red", Namespace: "red", Network: util.GenerateCUDNNetworkName("red"), Topology: "layer3", Subnet: "1.2.0.0/16", Labels: map[string]string{"selected": "true"}},
+			},
+			nodes:                []*testNode{{Name: "node", SubnetsAnnotation: "{\"default\":\"1.1.0.0/24\", \"cluster_udn_red\":\"1.2.0.0/24\"}"}},
+			reconcile:            "ra",
+			expectAcceptedStatus: metav1.ConditionTrue,
+			expectFRRConfigs: []*testFRRConfig{
+				{
+					Labels:       map[string]string{types.OvnRouteAdvertisementsKey: "ra"},
+					Annotations:  map[string]string{types.OvnRouteAdvertisementsKey: "ra/frrConfig/node"},
+					NodeSelector: map[string]string{"kubernetes.io/hostname": "node"},
+					Routers: []*testRouter{
+						// Router in default VRF (empty string) with imports from red VRF
+						{ASN: 1, Prefixes: []string{"1.2.0.0/24"}, Imports: []string{"red"}, Neighbors: []*testNeighbor{
+							{ASN: 1, Address: "1.0.0.100", Advertise: []string{"1.2.0.0/24"}},
+						}},
+						// Router in red VRF with imports from default VRF
+						{ASN: 1, VRF: "red", Imports: []string{"default"}},
+					}},
+			},
+			expectNADAnnotations: map[string]map[string]string{"red": {types.OvnRouteAdvertisementsKey: "[\"ra\"]"}},
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {


yboaron:
cc @ormergi

jcaamano:
@yboaron 

I think that the appropriate fix is for the documentation to not use `targetVRF: default` and omit targetVRF instead. 

Issue #5580 reports section https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/features/bgp-integration/route-advertisements.md#import-routes-into-a-cudn not working but it is for a different reason: we don't support establishing a BGP session over a CUDN VRF with no external interface attached. This is because the source IP address of the session will be the only IP address associated with the VRF, which is the management port IP address, and the external BGP peer would not know how to reach to that IP address back.

To support that there are different options but we would need to use FRR configuration that is not available in FRR-K8s API other than through its raw configuration stanza which is something we don't want to promote. So I guess the best course of action is to remove that section of the documentation as well.

Let me know if you can help with any of this. Thanks!

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -616,8 +616,10 @@ func (c *Controller) generateFRRConfiguration(
 			matchedVRF = router.VRF
 			matchedNetwork = selectedNetworks.networkVRFs[matchedVRF]
 			advertisePrefixes = selectedNetworks.hostNetworkSubnets[matchedNetwork]
-		case targetVRF == "":
+		case targetVRF == "" || targetVRF == "default":
```
pliurh, line 619: Shall we also treat `router.VRF` in the same way? Otherwise, we will hit the issue when `router.VRF="default"` and `target=""`.

yboaron:
Apologies for the delayed reply. 

Thanks for providing the detailed context, @jcaamano;  since you've already pushed [#5774](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5774), I'll go ahead and close this PR

PR #5704, opened 2025-11-05
Fix EIP reassignment on Namespace/Pod/EgressIP label updates

pperiyasamy:
When a namespace/pod/EgreessIP label update causes it to move from one EIP to another,
the EIP controller may process the associated EIPs in an order that leads
to incorrect assignment behavior.

Example Scenario:

1. Two EIPs exist:
   * eip1 matches namespace label test: qe
   * eip2 matches namespace label test: dev
2. Namespace ns1 initially has label test: dev and is served by eip2.
3. The label on ns1 is updated from test: dev to test: qe.
4. The EIP controller processes the Namespace update event:

   Step 1: eip1 is processed first but skips assignment since the pod is already served by eip2.
    -    In reconcileEgressIPNamespace, eip1 is processed first and matches the new Namespace object.
    -    It invokes addNamespaceEgressIPAssignments → addPodEgressIPAssignments for the pod, detects that eip2 (not yet processed) is still serving the pod, adds eip1 to  podState.standbyEgressIPNames, and returns.

   Step 2: eip2 is processed next, matches the old Namespace object, and deletes the pod from the assignment cache.
    -    In deleteNamespaceEgressIPAssignment → deletePodEgressIPAssignments, it cleans up OVN entries (LRP, SNAT and address sets) and removes the pod’s status entry from podAssignment cache.

As a result, ns1 is no longer assigned to eip1

Fix:
When eip2 is processed in the deletePodEgressIPAssignments method, promote eip1 from standby EgressIP to active.
The same issue might also occur during Pod label updates or EgressIP selector label updates (Namespace or Pod selector).
Added unit tests to cover Namespace, Pod, and EgressIP label update scenarios to reproduce the issue and verify the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable EgressIP cleanup with best-effort standby promotion when active assignments drop; promotion suppressed during stale-network/stale-path cleanup.

* **Refactor**
  * Deletion and reconciliation flows now accept a control flag to govern standby promotion, ensuring consistent per-pod removal and safer promotion behavior.

* **Tests**
  * Expanded end-to-end tests and new test scaffolding for EgressIP reassignment, standby promotion, cleanup paths, and cache/resilience scenarios.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -767,7 +767,7 @@ func (e *EgressIPController) addNamespaceEgressIPAssignments(ni util.NetInfo, na
 func (e *EgressIPController) addPodEgressIPAssignmentsWithLock(ni util.NetInfo, name string, statusAssignments []egressipv1.EgressIPStatusItem, mark util.EgressIPMark, pod *corev1.Pod) error {
 	e.podAssignment.LockKey(getPodKey(pod))
 	defer e.podAssignment.UnlockKey(getPodKey(pod))
-	e.deletePreviousNetworkPodEgressIPAssignments(ni, name, statusAssignments, pod)
+	e.deletePreviousNetworkPodEgressIPAssignments(ni, name, statusAssignments, pod, false)
 	return e.addPodEgressIPAssignments(ni, name, statusAssignments, mark, pod)
 }
 
@@ -857,7 +857,13 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 	}
 	for _, staleStatus := range staleAssignments {
 		klog.V(2).Infof("Deleting stale pod egress IP status: %v for EgressIP: %s and pod: %s/%s/%v", staleStatus, name, pod.Namespace, pod.Name, podIPNets)
-		err = e.deletePodEgressIPAssignments(ni, name, []egressipv1.EgressIPStatusItem{staleStatus}, pod)
+		// When deleting stale status during assignment, assignStandby=true is passed, but standby promotion
+		// will never occur because standbyEgressIPNames is empty (or doesn't contain this EgressIP) at this point:
+		// - If podState is new, standbyEgressIPNames is initialized empty
+		// - If this EIP is already active, it was removed from standbyEgressIPNames
+		// - If another EIP is active, we return early and never reach this code
+		// Additionally, even if statusMap becomes temporarily empty, new statuses are added immediately after.
+		err = e.deletePodEgressIPAssignments(ni, name, []egressipv1.EgressIPStatusItem{staleStatus}, pod, true)
 		if err != nil {
 			klog.Warningf("Failed to delete stale EgressIP status %s/%v for pod %s: %v", name, staleStatus, podKey, err)
 		}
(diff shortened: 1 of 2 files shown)


jcaamano:
> * eip1 is processed first but skips assignment since the pod is already served by eip2

Can you explain better how this happens? When EIP1 is processed, pod should be reassigned regardless of whether it was previously assigned given that only one EIP can select a pod at any point in time.

> Fix:
> Process EIP assignment deletion before addition when handling namespace update event to ensure consistent reassignment.

You should add a comment explaining why you are doing this. Otherwise, anyone can reverse your change in the future completely unaware of the consequences.

pperiyasamy:
> > * eip1 is processed first but skips assignment since the pod is already served by eip2
> 
> Can you explain better how this happens? When EIP1 is processed, pod should be reassigned regardless of whether it was previously assigned given that only one EIP can select a pod at any point in time.
> 
> > Fix:
> > Process EIP assignment deletion before addition when handling namespace update event to ensure consistent reassignment.
> 
> You should add a comment explaining why you are doing this. Otherwise, anyone can reverse your change in the future completely unaware of the consequences.

@jcaamano As we discussed offline, I'm moving the fix into `deletePodEgressIPAssignments` method itself so that it can handle EIP reassignment for label updates on namespace, pod and egressip objects. adding unit tests to cover all these scenarios. Will correct messaging in the next PR update.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -11146,6 +11146,472 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 			),
 		)
 
+		buildEIPReassignmentState := func(egressPod *corev1.Pod, eipName, eIPIP, podIP, nodeIP, nodeName string, withAssignment bool) []libovsdbtest.TestData {
```
jcaamano: Since these new tests come with its own set of utility functions that could, but are not, used by other tests, would it make sense to contain them in a subcontext for code clarity and better organization? like `ginkgo.Context("on label updates")`

go-controller/pkg/ovn/egressip.go
```diff
@@ -857,7 +857,11 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
 	}
 	for _, staleStatus := range staleAssignments {
 		klog.V(2).Infof("Deleting stale pod egress IP status: %v for EgressIP: %s and pod: %s/%s/%v", staleStatus, name, pod.Namespace, pod.Name, podIPNets)
-		err = e.deletePodEgressIPAssignments(ni, name, []egressipv1.EgressIPStatusItem{staleStatus}, pod)
+		// When deleting stale status during assignment, assignStandby=true is passed, but standby promotion
+		// won't actually occur because either:
+		// The pod still has other statuses for this EgressIP (statusMap not empty), OR
+		// New statuses will be added immediately later in this method.
+		err = e.deletePodEgressIPAssignments(ni, name, []egressipv1.EgressIPStatusItem{staleStatus}, pod, true)
 		if err != nil {
```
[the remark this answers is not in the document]
pperiyasamy: This `podState.standbyEgressIPNames` is also empty for this scenario, so moving standby EIP into active never happen. will update the comment.

go-controller/pkg/ovn/egressip_test.go
pperiyasamy: done.

PR #5709, opened 2025-11-06
cleanup primary pUDN connection details annotaton in cmdDel

cathy-zhou:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pod cleanup in DPU-host mode: handling now distinguishes default vs non-default networks. For the default network, the full DPU connection-details annotation is purged; for non-default networks, the prior per-network removal behavior is preserved.
  * Added explicit error reporting for failures during DPU connection-details cleanup, treating missing entries as non-errors while surfacing other cleanup failures.

### DIFF
--- a/go-controller/pkg/cni/cni.go
+++ b/go-controller/pkg/cni/cni.go
@@ -10,6 +10,7 @@ import (
 	current "github.com/containernetworking/cni/pkg/types/100"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


go-controller/pkg/cni/cni.go
```diff
@@ -311,9 +311,26 @@ func (pr *PodRequest) cmdDel(clientset *ClientSet) (*Response, error) {
...
 			netdevName = dpuCD.VfNetdevName
+			if pr.netName == types.DefaultNetworkName {
+				// if this is the default network name, remove the whole DPU connection-details annotation,
+				// including the primary UDN connection-details if any
+				updatePodAnnotationNoRollback := func(pod *corev1.Pod) (*corev1.Pod, func(), error) {
+					delete(pod.Annotations, util.DPUConnectionDetailsAnnot)
+					return pod, nil, nil
+				}
+
+				_ = util.UpdatePodWithRetryOrRollback(
```
arghosh93: Should we catch the error and print it, unless we want to return at this point?

go-controller/pkg/cni/cni.go
```diff
@@ -311,9 +311,26 @@ func (pr *PodRequest) cmdDel(clientset *ClientSet) (*Response, error) {
...
+					clientset.podLister,
+					&kube.Kube{KClient: clientset.kclient},
+					pod,
+					updatePodAnnotationNoRollback,
+				)
+			} else {
+				// Delete the DPU connection-details annotation for this NAD
+				_ = pr.updatePodDPUConnDetailsWithRetry(&kube.Kube{KClient: clientset.kclient}, clientset.podLister, nil)
```
arghosh93: Why don't we care about the error here?

cathy-zhou: same as above.

go-controller/pkg/cni/cni.go
cathy-zhou: I am following the old behavior. but I agree with you, it is better to return error here.  fixed

go-controller/pkg/cni/cni.go
```diff
@@ -311,9 +311,28 @@ func (pr *PodRequest) cmdDel(clientset *ClientSet) (*Response, error) {
...
+					&kube.Kube{KClient: clientset.kclient},
+					pod,
+					updatePodAnnotationNoRollback,
+				)
+			} else {
+				// Delete the DPU connection-details annotation for this NAD
+				err = pr.updatePodDPUConnDetailsWithRetry(&kube.Kube{KClient: clientset.kclient}, clientset.podLister, nil)
+			}
+			if err != nil {
+				return nil, fmt.Errorf("failed to cleanup the DPU connection details annotation for NAD %s: %v", pr.nadName, err)
```
trozet, line 336: this seems problematic because on CNI DEL the pod might be gone, then the CNI DEL will fail. I think you need to check for if apierrors.IsNotFound(err)

go-controller/pkg/cni/cni.go
```diff
@@ -311,9 +312,29 @@ func (pr *PodRequest) cmdDel(clientset *ClientSet) (*Response, error) {
...
+					updatePodAnnotationNoRollback,
+				)
+			} else {
+				// Delete the DPU connection-details annotation for this NAD
+				err = pr.updatePodDPUConnDetailsWithRetry(&kube.Kube{KClient: clientset.kclient}, clientset.podLister, nil)
+			}
+			// not an error if pod has already been deleted
+			if err != nil && !apierrors.IsNotFound(err) {
+				return nil, fmt.Errorf("failed to cleanup the DPU connection details annotation for NAD %s: %v", pr.nadName, err)
+			}
```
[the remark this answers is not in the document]
cathy-zhou: cni del is called by reverse order, starts from the last interface, and the default interface will always be the last to call

PR #5712, opened 2025-11-06
node: fix conntrack deletion to use service port instead of endpoint port

pliurh:
When reconciling conntrack entries for endpoint changes, we need to use
the service port (the externally exposed port) rather than the endpoint
port (the target port on the pod). This ensures conntrack entries are
properly flushed for the actual service port that clients connect to.

Changes:
- Add FindServicePortForEndpointSlicePort helper to map endpoint ports to service ports
- Update reconcileConntrackUponEndpointSliceEvents to use service port for conntrack deletion
- Handle service not found error explicitly (return early, service deletion is handled separately)

All tests pass successfully.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Added utilities and types to improve service-port resolution and load‑balancer endpoint construction.

* **Bug Fixes**
  * Treat missing/deleted Services referenced by EndpointSlices as non-fatal to avoid reconciliation failures.
  * Resolve endpoint-slice port names to concrete service ports before conntrack deletion; preserve aggregated error reporting.
  * Emit finer-grained per-event diagnostic logs for conntrack operations.

* **Tests**
  * Expanded table-driven and unit tests covering IPv4/IPv6, dual-stack, named/unnamed ports, multi-endpoint and edge cases.

### DIFF
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -1340,9 +1340,15 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
 		return fmt.Errorf("cannot reconcile conntrack: %v", err)
 	}
 	svc, err := nc.watchFactory.GetService(namespacedName.Namespace, namespacedName.Name)
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Service %s/%s not found (might have been deleted) when reconciling conntrack for endpointslice %s",
+				namespacedName.Namespace, namespacedName.Name, oldEndpointSlice.Name)
+			// service is not found, likely deleted, flushing service conntrack entries will be handled at service reconciliation. No-op here.
+			return nil
+		}
 		return fmt.Errorf("error while retrieving service for endpointslice %s/%s when reconciling conntrack: %v",
-			newEndpointSlice.Namespace, newEndpointSlice.Name, err)
+			oldEndpointSlice.Namespace, oldEndpointSlice.Name, err)
 	}
 	for _, oldPort := range oldEndpointSlice.Ports {
 		if *oldPort.Protocol != corev1.ProtocolUDP { // flush conntrack only for UDP
@@ -1356,10 +1362,21 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
 				if newEndpointSlice != nil && util.DoesEndpointSliceContainEligibleEndpoint(newEndpointSlice, oldIPStr, *oldPort.Port, *oldPort.Protocol, svc) {
 					continue
 				}
+				portName := ""
+				if oldPort.Name != nil {
+					portName = *oldPort.Name
+				}
+				servicePort, err := util.FindServicePortForEndpointSlicePort(svc, portName, *oldPort.Protocol)
+				if err != nil {
+					klog.Errorf("Failed to get service port for endpoint %s: %v", oldIPStr, err)
+					continue
+				}
 				// upon update and delete events, flush conntrack only for UDP
-				if err := util.DeleteConntrackServicePort(oldIPStr, *oldPort.Port, *oldPort.Protocol,
+				klog.V(5).Infof("Deleting conntrack entry for endpoint %s, port %d, protocol %s", oldIPStr, servicePort.Port, *oldPort.Protocol)
+				if err := util.DeleteConntrackServicePort(oldIPStr, servicePort.Port, *oldPort.Protocol,
 					netlink.ConntrackReplyAnyIP, nil); err != nil {
-					klog.Errorf("Failed to delete conntrack entry for %s: %v", oldIPStr, err)
+					klog.Errorf("Failed to delete conntrack entry for %s port %d: %v", oldIPStr, servicePort.Port, err)
+					errors = append(errors, err)
 				}
 			}
 		}
--- a/go-controller/pkg/util/util.go
+++ b/go-controller/pkg/util/util.go
@@ -888,6 +888,23 @@ func GetEndpointsForService(endpointSlices []*discoveryv1.EndpointSlice, service
 	return globalEndpoints, localEndpoints, errors.Join(validationErrors...)
 }
 
+// FindServicePortForEndpointSlicePort returns the ServicePort that corresponds to an EndpointSlice port
+// by matching the port name and protocol. This is the canonical way to map EndpointSlice ports to
+// Service ports, as Kubernetes guarantees that ServicePort.Name matches EndpointPort.Name.
+func FindServicePortForEndpointSlicePort(service *corev1.Service, endpointslicePortName string, endpointslicePortProtocol corev1.Protocol) (*corev1.ServicePort, error) {
+	if service == nil {
+		return nil, fmt.Errorf("unable to resolve port for endpointslice %q/%q: service is nil",
+			endpointslicePortName, endpointslicePortProtocol)
+	}
+	for _, servicePort := range service.Spec.Ports {
+		if servicePort.Name == endpointslicePortName && servicePort.Protocol == endpointslicePortProtocol {
+			return &servicePort, nil
+		}
+	}
+	return nil, fmt.Errorf("service %s/%s has no port with name %q and protocol %s",
+		service.Namespace, service.Name, endpointslicePortName, endpointslicePortProtocol)
+}
+
 // groupEndpointsByNode organizes a list of endpoints by their associated node names.
 // Endpoints without a NodeName are skipped, as they cannot be assigned to specific nodes.
 // This is used for building per-node endpoint mappings for local traffic policies.
--- a/go-controller/pkg/node/default_node_network_controller_test.go
+++ b/go-controller/pkg/node/default_node_network_controller_test.go
@@ -6,16 +6,20 @@ import (
 	"net"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/containernetworking/plugins/pkg/ns"
 	"github.com/containernetworking/plugins/pkg/testutils"
 	nadfake "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/clientset/versioned/fake"
+	"github.com/stretchr/testify/mock"
 	"github.com/urfave/cli/v2"
 	"github.com/vishvananda/netlink"
 
 	corev1 "k8s.io/api/core/v1"
+	discovery "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes/fake"
 
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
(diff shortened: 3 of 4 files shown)


go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1353,13 +1354,35 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
...
+				if svc != nil {
+					for _, p := range svc.Spec.Ports {
+						if p.Protocol != *oldPort.Protocol {
+							continue
+						}
+						if int32(p.TargetPort.IntValue()) == *oldPort.Port {
+							port = p.Port
+							break
+						}
+					}
```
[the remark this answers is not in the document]
ricky-rav: @pliurh Yes, let's pay attention to this. There's an ongoing effort to support named target ports throughout our code base: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5583

go-controller/pkg/util/net_linux.go
```diff
@@ -637,6 +641,24 @@ func DeleteConntrack(ip string, port int32, protocol corev1.Protocol, ipFilterTy
 	return nil
 }
 
+// deleteConntrackFunc is the function variable that can be mocked in tests
+var deleteConntrackFunc DeleteConntrackFunc = deleteConntrackImpl
+
+// DeleteConntrack deletes conntrack entries matching the given parameters
+func DeleteConntrack(ip string, port int32, protocol corev1.Protocol, ipFilterType netlink.ConntrackFilterType, labels [][]byte) error {
+	return deleteConntrackFunc(ip, port, protocol, ipFilterType, labels)
+}
```
ricky-rav: I'm not a big fan of having functions whose implementation depends on a package-level variable :)

go-controller/pkg/node/default_node_network_controller.go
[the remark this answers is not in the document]
ricky-rav: So when `targetPort` is a string, `IntValue()` returns 0, so the comparison always fails and the code will always fall back to port=0, which deletes all conntrack entries for the IP in question.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1353,13 +1354,35 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
...
+
+				// find the service port that maps to the endpoint port
+				if svc != nil {
+					for _, p := range svc.Spec.Ports {
+						if p.Protocol != *oldPort.Protocol {
+							continue
+						}
+						if int32(p.TargetPort.IntValue()) == *oldPort.Port {
+							port = p.Port
+							break
```
ricky-rav: So here the `break` statement lets us find the first matching target port. If multiple service ports map to the same endpoint port, only the first service port will have its conntrack entries cleaned up. 
```
  Service:
    [...]
    ports:
    - port: 80  # <--- conntrack entry cleaned up
      targetPort: 8080  <--- will match
    - port: 8080  # <--- will be skipped
      targetPort: 8080  
```
I would keep a list of service ports for which we want to delete the corresponding conntrack entries.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1353,13 +1354,35 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
 				oldIPStr := utilnet.ParseIPSloppy(oldIP).String()
 				// upon an update event, remove conntrack entries for IP addresses that are no longer
 				// in the endpointslice, skip otherwise
-				if newEndpointSlice != nil && util.DoesEndpointSliceContainEligibleEndpoint(newEndpointSlice, oldIPStr, *oldPort.Port, *oldPort.Protocol, svc) {
+				if newEndpointSlice != nil && svc != nil && util.DoesEndpointSliceContainEligibleEndpoint(newEndpointSlice, oldIPStr, *oldPort.Port, *oldPort.Protocol, svc) {
 					continue
 				}
+				port = 0 // reset to default (will match all ports if no service port found)
```
ricky-rav: Uhm... is it necessary to wipe out everything as fallback? You could easily have a pod with two containers, each one listening on a different port and being the backend of a different service. If one service is not found (got deleted?) is it correct to wipe out the conntrack entry for the other service, which might still be expected to be functional?

```
  Pod:
    containers:
    - name: app
      ports:
      - containerPort: 8080
        name: "http"
        protocol: UDP
      - containerPort: 9090
        name: "metrics"
        protocol: UDP

  Service A:
    ports:
    - port: 80
      targetPort: 8080

  Service B:
    ports:
    - port: 9090
      targetPort: 9090
```

go-controller/pkg/util/net_linux.go
pliurh: This is for mocking the function in UT

go-controller/pkg/node/default_node_network_controller.go
pliurh: I didn't know this was a valid configuration. I'll fix it.

pliurh: changed to use the port name to match.

go-controller/pkg/node/default_node_network_controller.go
pliurh: fixed

go-controller/pkg/util/util.go
```diff
@@ -888,6 +888,19 @@ func GetEndpointsForService(endpointSlices []*discoveryv1.EndpointSlice, service
...
+// FindServicePortForEndpointSlicePort returns the ServicePort that corresponds to an EndpointSlice port
+// by matching the port name and protocol. This is the canonical way to map EndpointSlice ports to
+// Service ports, as Kubernetes guarantees that ServicePort.Name matches EndpointPort.Name.
+func FindServicePortForEndpointSlicePort(service *corev1.Service, endpointslicePortName string, endpointslicePortProtocol corev1.Protocol) (corev1.ServicePort, error) {
+	for _, servicePort := range service.Spec.Ports {
+		if servicePort.Name == endpointslicePortName && servicePort.Protocol == endpointslicePortProtocol {
+			return servicePort, nil
+		}
+	}
+	return corev1.ServicePort{}, fmt.Errorf("service %s/%s has no port with name %s and protocol %s",
```
ricky-rav: NIT, in case the port name is an empty string:
```suggestion
	return corev1.ServicePort{}, fmt.Errorf("service %s/%s has no port with name %q and protocol %s",
```

go-controller/pkg/node/default_node_network_controller.go
ricky-rav: Awesome. With port names, there's a 1:1 mapping between endpoint ports and service ports.

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1340,7 +1340,13 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
...
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Service %s/%s not found (might have been deleted) when reconciling conntrack for endpointslice %s",
+				namespacedName.Namespace, namespacedName.Name, oldEndpointSlice.Name)
+			// TODO: service is not found, likely deleted, flushing service conntrack entries shall be handled at service reconciliation
+			return nil
+		}
 		return fmt.Errorf("error while retrieving service for endpointslice %s/%s when reconciling conntrack: %v",
 			newEndpointSlice.Namespace, newEndpointSlice.Name, err)
```
ricky-rav: This is actually my code, but since we're here already, could you fix the following? 
```suggestion
        return fmt.Errorf("error while retrieving service for endpointslice %s/%s when reconciling conntrack: %v",
            oldEndpointSlice.Namespace, oldEndpointSlice.Name, err)
```
Upon endpoint slice delete, we call `h.nc.reconcileConntrackUponEndpointSliceEvents(endpointslice, nil)` and if an error happens in `nc.watchFactory.GetService` above that is not `IsNotFound` , here we panic :-S

go-controller/pkg/util/util.go
```diff
@@ -888,6 +888,19 @@ func GetEndpointsForService(endpointSlices []*discoveryv1.EndpointSlice, service
...
+// Service ports, as Kubernetes guarantees that ServicePort.Name matches EndpointPort.Name.
+func FindServicePortForEndpointSlicePort(service *corev1.Service, endpointslicePortName string, endpointslicePortProtocol corev1.Protocol) (corev1.ServicePort, error) {
+	for _, servicePort := range service.Spec.Ports {
+		if servicePort.Name == endpointslicePortName && servicePort.Protocol == endpointslicePortProtocol {
+			return servicePort, nil
+		}
+	}
+	return corev1.ServicePort{}, fmt.Errorf("service %s/%s has no port with name %q and protocol %s",
+		service.Namespace, service.Name, endpointslicePortName, endpointslicePortProtocol)
+}
```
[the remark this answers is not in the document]
ricky-rav: @pliurh let's take care of this and for me we're good :) Thanks!

jcaamano:
> * Use port=0 (match all ports) when service is not found or port doesn't match
> * Switch from DeleteConntrackServicePort to DeleteConntrack to allow port=0

@pliurh I don't see those changes so I am confused. You probably need to remove that? What have we decided to do in this case?

@arkadeepsen is there any intersection on these changes to the other bug you fixed about conntrack that we were being unable to verify successfully?

jcaamano:
Don't we have any existing e2e coverage for this?

pliurh:
> I don't see those changes so I am confused. You probably need to remove that? What have we decided to do in this case?

I refactored the PR but forgot to update the description.

> Don't we have any existing e2e coverage for this?

No. The sig-network test uses the same port for both service and endpoint. This issue only happens when these ports are different.

arkadeepsen:
> @arkadeepsen is there any intersection on these changes to the other bug you fixed about conntrack that we were being unable to verify successfully?

The bug I have created the PR (#5715) for, handles the changes in the service spec and will trigger the deletion of stale UDP conntrack entries, if needed. The deletion is done using the original direction which refers to the service IP and port.

This PR is fixing the issue of deletion of stale UDP conntrack entries on update of endpointSlices (particularly change of endpoints). Here the deletion is done using the reply direction referring to the endpoint IP and port.

pliurh:
This PR addresses a different issue than #5715. 5717 is to address the case of changing the target UDP port of a nodePort service. This PR fixes the issue that the backend pod of a clusterIP service changes.

jcaamano:
@pliurh @arkadeepsen I was talking about https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5300 but upon review I see that is related to external gateways, not services.

jcaamano:
> No. The sig-network test uses the same port for both service and endpoint. This issue only happens when these ports are different.

Should we contribute upstream to provide this coverage, or rather add our own e2e here? Do the upstream tests exercise the conntrack cleanup or would they fail if it doesn't happen?

pliurh:
Yeah, we can change this [test](https://github.com/kubernetes/kubernetes/blob/5a26d2af610d1b9d3aba01ebd608282d06b193fc/test/e2e/network/conntrack.go#L223) to use different port number for service and target port

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1340,9 +1340,15 @@ func (nc *DefaultNodeNetworkController) reconcileConntrackUponEndpointSliceEvent
 		return fmt.Errorf("cannot reconcile conntrack: %v", err)
 	}
 	svc, err := nc.watchFactory.GetService(namespacedName.Namespace, namespacedName.Name)
-	if err != nil && !apierrors.IsNotFound(err) {
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.V(5).Infof("Service %s/%s not found (might have been deleted) when reconciling conntrack for endpointslice %s",
+				namespacedName.Namespace, namespacedName.Name, oldEndpointSlice.Name)
+			// TODO: service is not found, likely deleted, flushing service conntrack entries shall be handled at service reconciliation
```
jcaamano: nit: This should not be a `TODO` right? it should be already happening?

pliurh: Yeah, I just found the logic in `nodePortWatcher`. I will remove this comment. However, I'm still wondering why we didn't do it here in `DefaultNodeNetworkController`.

pliurh:
I proposed a [PR](https://github.com/kubernetes/kubernetes/pull/135264) for the u/s test. It will test the case addressed by this PR.

PR #5714, opened 2025-11-06
support pod with multiple same secondary UDN

cathy-zhou:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Indexed NAD keys to support multiple attachments to the same secondary network.
  * ECMP routing and per-interface table routing for pods with multiple interfaces to the same network.

* **Improvements**
  * Consistent NAD-key based handling across annotations, readiness gates, DPU/OVS flows, port tracking and logging.
  * Per-interface route metrics and better interface discovery/cleanup.

* **Tests**
  * New e2e coverage validating multihoming redundancy, failover and cross-interface connectivity.

### DIFF
--- a/test/e2e/multihoming.go
+++ b/test/e2e/multihoming.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"strings"
 	"time"
 
@@ -16,6 +17,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
+	e2ekubectl "k8s.io/kubernetes/test/e2e/framework/kubectl"
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 
 	mnpapi "github.com/k8snetworkplumbingwg/multi-networkpolicy/pkg/apis/k8s.cni.cncf.io/v1beta1"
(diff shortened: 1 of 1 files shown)


go-controller/pkg/cni/cni.go
```diff
@@ -317,21 +362,28 @@ func (pr *PodRequest) cmdDel(clientset *ClientSet) (*Response, error) {
...
+				// the pod was added before "external_ids:pod-if-name" was introduced, fall back to the old way to find
+				// out the OVS interface associated with this CNIDel request
+				condString = []string{"external-ids:sandbox=" + pr.SandboxID}
+				if pr.netName != types.DefaultNetworkName {
+					condString = append(condString, fmt.Sprintf("external_ids:%s=%s", types.NADExternalID, pr.nadName))
+				} else if pod != nil {
+					condString = append(condString, fmt.Sprintf("external_ids:%s{=}[]", types.NADExternalID))
+				}
+				ovsIfNames, err = ovsFind("Interface", "name", condString...)
+			}
```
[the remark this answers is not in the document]
cathy-zhou: based on ovs-vsctl man page, "{=} {!=}: Test for set equality or inequality, respectively.", it means this value is empty

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -538,19 +538,17 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
...
-			return false, err
+		on, networkMap, err := util.GetPodNADToNetworkMapping(pod, bsnc)
+		if !on || err != nil {
+			if err != nil {
+				return false, fmt.Errorf("pod %s/%s is not scheduled on network %s: %v",
+					pod.Namespace, pod.Name, bsnc.GetNetworkName(), err)
+			}
+			return false, fmt.Errorf("pod %s/%s is not scheduled on network %s",
+				pod.Namespace, pod.Name, bsnc.GetNetworkName())
 		}
```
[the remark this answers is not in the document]
cathy-zhou: when it comes here, it means this pod is plumbed over this network.  so on should be true. I will log an error message instead.

cathy-zhou:
@girishmg @trozet please take a look. thanks!

go-controller/pkg/allocator/pod/pod_annotation_test.go
```diff
@@ -873,21 +873,6 @@ func Test_allocatePodAnnotationWithRollback(t *testing.T) {
...
-				idAllocator: &idAllocatorStub{
-					reserveIDError: errors.New("ID allocation error"),
-				},
-			},
-			podAnnotation: &util.PodAnnotation{
-				MAC:      randomMac,
-				TunnelID: 200,
-			},
-			wantErr: true,
-		},
```
trozet, line 995: why is this test removed?

go-controller/pkg/util/multi_network.go
```diff
@@ -1233,6 +1245,35 @@ func GetNADName(namespace, name string) string {
...
+// the resulted nadKey example is like "ns/nad", "ns/nad/1", "ns/nad/2" etc...
+func GetIndexedNADKey(nadName string, n int) string {
+	if n == 0 {
+		return nadName
+	}
+	return fmt.Sprintf("%s/%d", nadName, n)
+}
+
+// GetNadFromIndexedNADKey returns NAD name part from a nadName key (with or without index)
+func GetNadFromIndexedNADKey(nadNameWithIndex string) (string, int, error) {
```
trozet: nit: parseNADKey

go-controller/pkg/util/multi_network.go
```diff
@@ -413,6 +414,17 @@ func (nInfo *mutableNetInfo) EqualNADs(nads ...string) bool {
...
+// to check if the network needs to be plumbed over
+func (nInfo *mutableNetInfo) HasNADKey(nadKey string) bool {
+	nadName, _, err := GetNadFromIndexedNADKey(nadKey)
+	if err != nil {
+		return false
+	}
+
+	return nInfo.HasNAD(nadName)
+}
+
```
trozet, line 427: this function is missing nInfo.Rlock(),  defer nInfo.RUnlock()

go-controller/pkg/util/multi_network.go
```diff
@@ -1480,11 +1521,13 @@ func SubnetOverlapCheck(netconf *ovncnitypes.NetConf) (*net.IPNet, *net.IPNet, e
...
 //
-//	bool: if this Pod is on this Network; true or false
-//	map[string]*nettypes.NetworkSelectionElement: all NetworkSelectionElement that pod is requested
-//	    for the specified network, key is NADName. Note multiple NADs of the same network are allowed
-//	    on one pod, as long as they are of different NADName.
-//	error:  error in case of failure
+//		bool: if this Pod is on this Network; true or false
+//		map[string]*nettypes.NetworkSelectionElement: all NetworkSelectionElement that the pod requests
+//		    for the specified network, keyed by NAD key. NAD keys are of the form "namespace/name"
+//	        for the first attachment of a NAD, and "namespace/name/<idx>" for additional attachments
```
trozet: specify where index starts. At 1 right?

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -449,7 +449,7 @@ func (bsnc *BaseUserDefinedNetworkController) removePodForUserDefinedNetwork(pod
 
 	var alreadyProcessed bool
 	for nadName, podAnnotation := range podNetworks {
-		if !bsnc.HasNAD(nadName) {
+		if !bsnc.HasNADKey(nadName) {
```
trozet: this is a bit confusing that it says HasNADKey(nadName), we should rename nadName ->nadKey where we read it off pod annotations

go-controller/pkg/cni/utils.go
```diff
@@ -152,6 +144,36 @@ func PodAnnotation2PodInfo(podAnnotation map[string]string, podNADAnnotation *ut
...
+	for idx, network := range networks {
+		nad := util.GetNADName(network.Namespace, network.Name)
+		// for multiple NetworkSelectionElements of the same NAD, set its nadName to indexed nadName
+		cnt := nNADs[nad]
+		nNADs[nad] = cnt + 1
+		if network.InterfaceRequest != "" {
+			if network.InterfaceRequest != ifName {
+				continue
+			}
+		} else if fmt.Sprintf("net%d", idx+1) != ifName {
```
trozet: maybe a comment here on how this value is determined. It's multus right?

go-controller/pkg/cni/utils.go
```diff
@@ -152,6 +144,36 @@ func PodAnnotation2PodInfo(podAnnotation map[string]string, podNADAnnotation *ut
 	return podInterfaceInfo, nil
 }
 
+// GetCNINADKey gets the pod's nadKey (nadName with index in case there are multiple same NADs in the pod)
```
trozet: comment here about what ifName does as part of the filter/search criteria

go-controller/pkg/cni/utils.go
```diff
@@ -152,6 +144,36 @@ func PodAnnotation2PodInfo(podAnnotation map[string]string, podNADAnnotation *ut
...
+				continue
+			}
+		} else if fmt.Sprintf("net%d", idx+1) != ifName {
+			continue
+		}
+		nadKey := util.GetIndexedNADKey(nad, cnt)
+		// check if the derived nadKey is for the given nadName
+		if nadKey != nadName && !strings.HasPrefix(nadKey, nadName+"/") {
+			return "", fmt.Errorf("unexpected NAD key %s associated with CNI request with ifName %s, expected for NAD %s", nadKey, ifName, nadName)
+		}
```
trozet: this seems wrong to me because you are iterating through networks, nad if you didn't find a nad that matches, then you return error during the loop. Shouldn't you just change line 156 to check if nad == nadName, then continue if its not the same nad?

go-controller/pkg/cni/cni.go
```diff
@@ -148,13 +150,41 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(
...
+				} else {
+					return false, nil
+				}
+			} else {
+				return true, nil
+			}
+		})
+	if pod == nil {
+		return nil, fmt.Errorf("failed to get pod %s/%s: %v", pr.PodNamespace, pr.PodName, err)
+	}
```
trozet: can you use GetPodWithAnnotations instead and just use a dummy podAnnotWaitCond ?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -494,6 +494,11 @@ func ConfigureOVS(ctx context.Context, namespace, podName, hostIfaceName string,
 		fmt.Sprintf("external_ids:sandbox=%s", sandboxID),
 	}
 
+	// pod interface name, used to identify CNI request with the same NAD
+	if podIfName != "" {
+		ovsArgs = append(ovsArgs, fmt.Sprintf("external_ids:pod-if-name=%s", podIfName))
```
trozet, line 603: would be good to include a commit message talking about this new external ID

go-controller/pkg/cni/cni.go
```diff
@@ -148,13 +150,41 @@ func (pr *PodRequest) cmdAddWithGetCNIResultFunc(
...
+
+	if pr.netName != types.DefaultNetworkName {
+		nadKey, err := GetCNINADKey(pod, pr.IfName, pr.nadName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get NAD key for CNI Add request %v: %v", pr, err)
+		}
+		pr.nadName = nadKey
```
trozet: so here you set the pod request nad name to the nadKey, the nadName is then passed around like:

isDPUReady(annotCondFn, pr.nadName)

^ can this function hahndle the nadKey or does it really expect the nadName? It can get confusing when we are overloading the pr.nadName with the key and the name. Can we keep it separate? like not overwrite pr.nadName an leave it as the name, then store the key somewhere else if it needs to be stored, or just pass the key to the functions that need it?

go-controller/pkg/allocator/pod/pod_annotation.go
```diff
@@ -565,12 +565,18 @@ func AddRoutesGatewayIP(
...
 	pod *corev1.Pod,
 	podAnnotation *util.PodAnnotation,
+	nadName string,
 	network *nadapi.NetworkSelectionElement) error {
 
 	// generate the nodeSubnets from the allocated IPs
 	nodeSubnets := util.IPsToNetworkIPs(podAnnotation.IPs...)
 
 	if netinfo.IsUserDefinedNetwork() {
+		_, index, err := util.GetNadFromIndexedNADKey(nadName)
```
trozet: this commit needs a commit message explaining why this commit is needed. I don't see why you need to add routes with different metrics. Each attachment should have a directly connected route for layer 2, and layer 3 I could see you needing to add a route for each, but I don't see why they would be different metrics.

go-controller/pkg/ovn/base_network_controller_user_defined.go
```diff
@@ -539,19 +538,17 @@ func (bsnc *BaseUserDefinedNetworkController) hasIPAMClaim(pod *corev1.Pod, nadN
...
+		}
+		if !on {
+			klog.Warningf("Pod %s/%s is not scheduled on network %s", pod.Namespace, pod.Name, bsnc.GetNetworkName())
+			return false, nil
 		}
-		for _, network := range allNetworks {
-			if network.Namespace == nadNamespace && network.Name == nadName {
-				// found network selection element, check if it has IPAM
+		for nadName, network := range networkMap {
+			if nadName == nadKey {
```
trozet: networkMap no longer has a nadName, it's a nadKey. So you should call it something else and not compare nadName to nadKey.

go-controller/pkg/allocator/pod/pod_annotation.go
cathy-zhou: for layer3 network, say 172.31.0.0/16/24, for each layer3 interface, with current code it will try to add :
172.31.0.0/16 via 172.31.2.1 dev net1
172.31.0.0/16 via 172.31.2.1 dev net2
the second route add operation would fail with EEXIST. I brought up this with you and @girishmg and I thought we agreed to add metrics based routes.

go-controller/pkg/allocator/pod/pod_annotation_test.go
cathy-zhou: this is duplicated test as the previous one. codeRabbit found this and asked me to remove one of them

go-controller/pkg/util/multi_network.go
cathy-zhou: it calls into nInfo.HasNAD() which calls RLock() and RUnlock().

go-controller/pkg/util/multi_network.go
cathy-zhou: ok

go-controller/pkg/util/multi_network.go
cathy-zhou: yes

go-controller/pkg/ovn/base_network_controller_user_defined.go
cathy-zhou: will fix

go-controller/pkg/cni/helper_linux.go
cathy-zhou: will do

go-controller/pkg/cni/cni.go
cathy-zhou: will do

go-controller/pkg/ovn/base_network_controller_user_defined.go
cathy-zhou: will do

go-controller/pkg/cni/cni.go
cathy-zhou: sure

cathy-zhou:
@trozet addressed your comments, please take a look.

go-controller/pkg/allocator/pod/pod_annotation.go
trozet: I think you would rather have ECMP rather than doing metrics. The correct syntax for that is:

```
root@ovn-worker:/# ip route add 172.19.0.0/16 dev eth0
RTNETLINK answers: File exists
root@ovn-worker:/#  ip route replace 172.19.0.0/16 \
  nexthop dev eth0 \
  nexthop dev eth1
root@ovn-worker:/# ip route show
default via 172.18.0.1 dev breth0 
10.96.0.0/16 via 169.254.0.4 dev breth0 src 169.254.0.2 mtu 1400 
10.244.0.0/24 dev ovn-k8s-mp0 proto kernel scope link src 10.244.0.2 
10.244.0.0/16 via 10.244.0.1 dev ovn-k8s-mp0 
169.254.0.0/17 dev breth0 proto kernel scope link src 169.254.0.2 
169.254.0.1 dev breth0 src 172.18.0.3 
169.254.0.3 via 10.244.0.1 dev ovn-k8s-mp0 
172.18.0.0/16 dev breth0 proto kernel scope link src 172.18.0.3 
172.19.0.0/16 
	nexthop dev eth0 weight 1 
	nexthop dev eth1 weight 1 

```

go-controller/pkg/ovn/egressip.go
```diff
@@ -790,20 +790,20 @@ func (e *EgressIPController) addPodEgressIPAssignments(ni util.NetInfo, name str
...
+		nadKey = nadNames[0] // there should only be one active network
 	}
-	podIPNets, err := e.getPodIPs(ni, pod, nadName)
+	podIPNets, err := e.getPodIPs(ni, pod, nadKey)
 	if err != nil {
 		return fmt.Errorf("failed to get pod %s/%s IPs: %v", pod.Namespace, pod.Name, err)
 	}
 	if len(podIPNets) == 0 {
-		return fmt.Errorf("failed to get pod ips for pod %s on network %s with NAD name %s", podKey, ni.GetNetworkName(), nadName)
+		return fmt.Errorf("failed to get pod ips for pod %s on network %s with NAD key %s", podKey, ni.GetNetworkName(), nadKey)
```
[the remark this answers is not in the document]
cathy-zhou: this is not new after my fix. even prior to my changes, there could be multiple NADs (with different NAD names) and multiple pod IPs in the same network. So I didn't change the current behavior.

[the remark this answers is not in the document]
trozet: @coderabbitai egress IP is only supported on the primary UDN, which cannot have multiple attachments to the same network. For secondary networks (which support multiple attachments to the same NAD), egress IP is unsupported

test/e2e/multihoming.go
```diff
@@ -2382,6 +2383,288 @@ ip a add %[4]s/24 dev %[2]s
...
+
+				By("verifying both interfaces have unique interface names")
+				interfaceNames := make([]string, 2)
+				for i := 0; i < 2; i++ {
+					Expect(netStatus[i].Interface).NotTo(BeEmpty(), fmt.Sprintf("Interface %d should have a name", i))
+					interfaceNames[i] = netStatus[i].Interface
+				}
+				Expect(interfaceNames[0]).NotTo(Equal(interfaceNames[1]), "First and second interface names should be different")
+
+				By(fmt.Sprintf("Successfully validated two interfaces with IPs: %s, %s", ips[0], ips[1]))
```
trozet, line 2477: I think it would be worth adding a check to make sure the correct route exists in the pod with ecmp the correct nexthops. wdyt?

cathy-zhou: Added. please take a look. @trozet

go-controller/pkg/cni/helper_linux.go
```diff
@@ -210,31 +211,98 @@ func setupNetwork(link netlink.Link, ifInfo *PodInterfaceInfo) error {
...
+		if err := cniPluginLibOps.AddRoute(nil, gw, link, ifInfo.RoutableMTU, 0); err != nil {
 			return fmt.Errorf("failed to add gateway route to link '%s': %v", link.Attrs().Name, err)
 		}
 	}
+
+	// all pod links of the same secondary UDN up to the current one
+	var links []netlink.Link
+	var iptableNum int
+
+	if len(ifInfo.PodIfNamesOfSameNAD) != 0 && len(ifInfo.Routes) != 0 {
```
trozet, line 223: shouldn't this be just be

```
if len(ifInfo.PodIfNamesOfSameNAD) > 0 {
```

wouldn't you have at least 1 (like your check in cni.go on line 241-243?

also, you dont need the len(ifInfo.Routes) cause you will need to do this even for just the link local routes for layer 2 right?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -210,31 +211,98 @@ func setupNetwork(link netlink.Link, ifInfo *PodInterfaceInfo) error {
...
+			if err != nil {
+				return fmt.Errorf("failed to lookup pod interface %s when setup route for link %s: %v", ifName, link.Attrs().Name, err)
+			}
+			links = append(links, ifLink)
+		}
+
+		// for ECMP routes, need to configure the following sysctl configuration, stop ARP flux and stop RP filter drop from happening
+		//   sysctl -w net.ipv4.conf.<linkName>.arp_ignore=1 # only reply to ARP requests if its sent to the IP on this interface
+		//   sysctl -w net.ipv4.conf.<linkName>.arp_announce=2 # when sending ARP use the IP that belongs to this interface
+		//   sysctl -w net.ipv4.conf.<linkName>.rp_filter=2 # allow reverse path filter check to pass if there is any route to the source
```
trozet, line 248: nice comments!

go-controller/pkg/cni/helper_linux.go
```diff
@@ -210,31 +211,98 @@ func setupNetwork(link netlink.Link, ifInfo *PodInterfaceInfo) error {
...
+		if err := setSysctl(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/arp_announce", linkName), 2); err != nil {
+			return fmt.Errorf("failed to set arp_announce for %s: %v", linkName, err)
+		}
+		if err := setSysctl(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/rp_filter", linkName), 2); err != nil {
+			return fmt.Errorf("failed to set rp_filter for %s: %v", linkName, err)
+		}
+
+		// then add ip rules/routes to the specific table
+		// configure IP rule for this interface
+		// 	 ip rule add from <IP_on_this_interface> table <iptableNum>
```
trozet: maybe add a comment here saying:
"ip rules are needed to force traffic from an ip to egress the correct interface via a route table for that interface"

go-controller/pkg/cni/helper_linux.go
```diff
@@ -210,31 +211,98 @@ func setupNetwork(link netlink.Link, ifInfo *PodInterfaceInfo) error {
...
+		for _, ip := range ifInfo.IPs {
+			rule := netlink.NewRule()
```
trozet, line 263: do we need to set priority at all here?

go-controller/pkg/cni/helper_linux.go
```diff
@@ -210,31 +211,98 @@ func setupNetwork(link netlink.Link, ifInfo *PodInterfaceInfo) error {
...
+				if err := cniPluginLibOps.AddRoute(route.Dest, route.NextHop, link, ifInfo.RoutableMTU, 0); err != nil {
+					return fmt.Errorf("failed to add pod route %v via %v: %v", route.Dest, route.NextHop, err)
 				}
-				ifLink, err := util.GetNetLinkOps().LinkByName(ifName)
-				if err != nil {
-					return fmt.Errorf("failed to lookup pod interface %s when setup route for link %s: %v", ifName, link.Attrs().Name, err)
+			} else {
+				// otherwise replace it with ECMP routes
+				if err := cniPluginLibOps.ReplaceRouteECMP(route.Dest, route.NextHop, links, ifInfo.RoutableMTU); err != nil {
+					return fmt.Errorf("failed to replace pod route %v via %v through links %v: %v", route.Dest, route.NextHop, links, err)
```
trozet, line 298: you dont need the ecmp routes anymore, because the ip rule and routes in the table take care of it

go-controller/pkg/cni/helper_linux.go
trozet: oh i see at the end of  changes you have:
// Only set for when there are more than one pod interface with the same UDN

this is a little confusing. Can we just set it to the the number of overall interfaces on the NAD and use if greater than 1?

cathy-zhou: >  Only set for when there are more than one pod interface with the same UDN

I thought about it but that would complicate the the unit testing cases, as we then have to set ifInfo.PodIfNamesOfSameNAD for the most common test case that has only 1 PodIfNamesOfSameNAD.

cathy-zhou: > also, you dont need the len(ifInfo.Routes) cause you will need to do this even for just the link local routes for layer 2 right?

do we? I don't see any test failures when in the layer2 traffic tests.

go-controller/pkg/cni/helper_linux.go
cathy-zhou: I tested on my setup. with the following ip rules, delete ecmp routes won't work.

```
# ip rule
0:      from all lookup local
32764:  from 172.31.1.7 lookup 102
32765:  from 172.31.1.3 lookup 101
32766:  from all lookup main
32767:  from all lookup default

```

go-controller/pkg/util/net_linux.go
```diff
@@ -50,6 +50,7 @@ type NetLinkOps interface {
 	RouteReplace(route *netlink.Route) error
 	RouteListFiltered(family int, filter *netlink.Route, filterMask uint64) ([]netlink.Route, error)
 	RuleListFiltered(family int, filter *netlink.Rule, filterMask uint64) ([]netlink.Rule, error)
+	RuleAdd(rule *netlink.Rule) error
```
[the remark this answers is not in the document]
cathy-zhou: it is in go-controller/pkg/util/mocks/NetLinkOps.go

go-controller/pkg/cni/helper_linux.go
cathy-zhou: no. the result looks like below:

```
0:      from all lookup local
32764:  from 172.31.3.5 lookup 104
32765:  from 172.31.3.6 lookup 103
32766:  from all lookup main
32767:  from all lookup default

```

go-controller/pkg/cni/helper_linux.go
cathy-zhou: done

go-controller/pkg/cni/helper_linux.go
```diff
@@ -192,13 +211,99 @@ func setupNetwork(link netlink.Link, ifInfo *PodInterfaceInfo) error {
...
 	}
+
+	// all pod links of the same secondary UDN up to the current one
+	var links []netlink.Link
+	var iptableNum int
+
+	if len(ifInfo.PodIfNamesOfSameNAD) != 0 && len(ifInfo.Routes) != 0 {
+		// needs ip table for each Pod interface of the same secondary UDN, table number start from 100
+		iptableNum = 100 + link.Attrs().Index
+
```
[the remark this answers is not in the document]
trozet: interfaces cant be deleted and recreated within kubernetes. The pod CNI adds and deletes all interfaces at once.

go-controller/pkg/cni/helper_linux.go
trozet: you are right we still need the route otherwise dest route lookup for finding initial source interface to use in linux wont work

go-controller/pkg/cni/helper_linux.go
trozet: seems fine for now

go-controller/pkg/cni/helper_linux.go
trozet: In the layer 2 case it looks like we dont need the policy rules/routing. Since we have two link local routes to the same subnet and no ecmp route:
172.31.1.0/24 dev net1 proto kernel scope link src 172.31.1.3                                         
172.31.1.0/24 dev net2 proto kernel scope link src 172.31.1.7

linux allows tie breaking for link routes so that coupled with the syctl arp changes makes this work

trozet (approved):

PR #5678, opened 2025-10-21
Bump to k8s v1.34.1

meinali566:
## 📑 Description

```
go-controller rebase 1.34
> go-controller
 go get k8s.io/api@v0.34.1 
 go get k8s.io/client-go@v0.34.1
 go get k8s.io/component-helpers@v0.34.1
 go get k8s.io/kubernetes@v1.34.1
 go get k8s.io/apiextensions-apiserver@v0.34.1
 go get sigs.k8s.io/controller-runtime@v0.22.1
 go get k8s.io/controller-manager@v0.34.1
 go get sigs.k8s.io/structured-merge-diff/v4@v4.7.0
 go get sigs.k8s.io/knftables@v0.0.19
 go get k8s.io/kubelet@v0.34.1
 go get sigs.k8s.io/apiserver-network-proxy/konnectivity-client@v0.33.0
 go get k8s.io/kube-openapi
 go get sigs.k8s.io/json
 go get k8s.io/utils
 go get github.com/openshift/api
 go get github.com/openshift/client-go
 go mod tidy
 go mod vendor

> Update
   1.Update GO version in go.mod and Makefile
   2.Update hack/lint.sh to v2.5.0, which uses “golangci-lint migrate” [Refer to: https://golangci-lint.run/docs/product/migration-guide/ ] .
   3.“typecheck” isn’t a linter and it runs automatically. So we don’t need it anymore in golangci-lint.yml [Refer to: https://golangci-lint.run/docs/welcome/faq/#why-do-you-have-typecheck-errors ]
   4.after golangci-lint version 2, staticcheck has a new feature: quickfix(QF) [https://staticcheck.dev/docs/checks#QF ]. It will affect lint results, so disable all QF checks first.

> Fix
   1.Fix TypeConverter undefined errors in pkg/crd/…/v1/apislyconfiguration/utils.go:
      k8s.io/client-go/testing package has removed the type converter, which uses apimachinery package instead. Therefore, update applyconfigurations’ dependency. Refer to:https://github.com/kubernetes/client-go/commit/571b9f39e05ac6d7722ec2a88874d2c89b36c098#diff-291ee9f54cd81b510e6064f8f19ad09b74ed4ea311f35413553a57efad745e67 
    2.Fix staticcheck and testifylint errors:
       11 issues:
       * staticcheck: 7
       * testifylint: 4

test/e2e rebase 1.34
> test/e2e
   go get k8s.io/kubernetes@v1.34.1
   go get k8s.io/api@v0.34.1
[...]

PR #5715, opened 2025-11-07
UDP Conntrack entry cleanup for service updates

arkadeepsen:
## 📑 Description

A new test is added in kubernetes which checks the UDP conntrack deletion for service update as part of k8s v1.34.0:
- https://github.com/kubernetes/kubernetes/pull/132708

This PR adds the implementation to delete stale UDP conntrack entries when a service is updated. This ensures that old conntrack entries do not interfere with packet routing after service changes. The function checks for changes in service's ip:port:targetPort combination and performs cleanup accordingly, improving the handling of connectionless UDP traffic.

Fixes #

The k8s v1.34.1 bump PR (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5678) has a shard conformance test failure due to the new test: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/19126923783/job/54660669978?pr=5678

## ✅ Checks

- [x] My code requires tests

## How to verify it

Added unit tests to check correct UDP conntrack entry deletion. Additionally, was able to successfully run the upstream kubernetes test in a local kind cluster.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UDP conntrack cleanup on service updates — precise per-port cleanup and coverage for NodePort, ExternalIP, LoadBalancer IP and ClusterIP changes; TCP updates remain unaffected.
  * Conntrack deletion routines now return matched-deletion counts alongside error results for clearer outcomes.

* **Tests**
  * Expanded unit tests for UDP conntrack scenarios (target-port changes, multi-port services, External/LB IP changes) with exact deletion-count verification; re-enabled related e2e test.

* **Chores**
  * Tightened deletion logging and aggregated error/result handling across conntrack paths.

### DIFF
--- a/test/scripts/e2e-kind.sh
+++ b/test/scripts/e2e-kind.sh
@@ -69,9 +69,6 @@ validates that there is no conflict between pods with same hostPort but differen
 # https://issues.redhat.com/browse/OCPBUGS-61380
 should support named targetPorts that resolve to different ports on different endpoints
 
-# https://issues.redhat.com/browse/OCPBUGS-61378
-should be able to cleanup conntrack entries when UDP service target port changes for a NodePort service
-
 # https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5119
 \[sig-network\] Services should implement NodePort and HealthCheckNodePort correctly when ExternalTrafficPolicy changes
 


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +924,94 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+
+	// Get the new IP:port:targetPort combinations for External and LB IP, ClusterIP and NodePort
+	newExternalIPs := util.GetExternalAndLBIPs(new)
+	newClusterIPs := util.GetClusterIPs(new)
+	newExternalIPPortTargetPorts := sets.New[string]()
+	newClusterIPPortTargetPorts := sets.New[string]()
+	newNodeIPPortTargetPorts := sets.New[string]()
+	var newNodeIPs []net.IP
+	if util.ServiceTypeHasNodePort(new) {
+		newNodeIPs, _ = npw.nodeIPManager.ListAddresses()
```
pliurh: Why is the returned error ignored here?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +924,94 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+		for _, newExternalIP := range newExternalIPs {
+
+			newExternalIPPortTargetPorts.Insert(fmt.Sprintf("%s:%d:%s", newExternalIP, newPort.Port, newPort.TargetPort.String()))
+		}
+		for _, newClusterIP := range newClusterIPs {
+			newClusterIPPortTargetPorts.Insert(fmt.Sprintf("%s:%d:%s", newClusterIP, newPort.Port, newPort.TargetPort.String()))
+		}
+		for _, newNodeIP := range newNodeIPs {
+			newNodeIPPortTargetPorts.Insert(fmt.Sprintf("%s:%d:%s", newNodeIP.String(), newPort.NodePort, newPort.TargetPort.String()))
+		}
```
pliurh: The same pattern is repeated 3 times in the function (for externalIPs, clusterIPs, nodeIPs). This could be extracted into a helper function.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +924,94 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+		for _, oldNodeIP := range oldNodeIPs {
+			if !newNodeIPPortTargetPorts.Has(fmt.Sprintf("%s:%d:%s", oldNodeIP.String(), oldPort.NodePort, oldPort.TargetPort.String())) {
+				if err := util.DeleteConntrackServicePort(oldNodeIP.String(), oldPort.NodePort, oldPort.Protocol,
+					netlink.ConntrackOrigDstIP, nil); err != nil {
+					errs = append(errs, fmt.Errorf("failed to delete conntrack entry for service %s/%s with nodeIP %s, nodePort %d, protocol %s: %v",
+						old.Namespace, old.Name, oldNodeIP.String(), oldPort.NodePort, oldPort.Protocol, err))
+				} else {
+					deletedCount++
+				}
+			}
```
pliurh: The same pattern is repeated 3 times in the function (for externalIPs, clusterIPs, nodeIPs). This could be extracted into a helper function.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -895,6 +895,14 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
 		if err = delServiceRules(old, svcConfig.localEndpoints, npw); err != nil {
 			errors = append(errors, err)
 		}
+
+		// Delete conntrack entries for UDP ports of the service if they have changed.
+		// UDP is connectionless, so stale conntrack entries can cause packets to be routed to
+		// old target ports even after service updates. Unlike TCP (which is stateful and manages
+		// its own connection lifecycle), UDP conntrack entries must be explicitly cleaned up.
+		if err = npw.deleteConntrackForUpdatedUDPServicePorts(old, new); err != nil {
```
pliurh: Can we check if the service ip/port changed and just invoke `deleteConntrackForService(old)` instead? Even if some conntrack entries are removed unintentionally, the NEW packet will match the same rule as the original packet did. The impact on traffic should be minimal in the worst case.

pliurh:
@tssurya @jcaamano  A dumb question, why do we handle service conntrack cleanup at `nodePortWatcher`. Shouldn't it be just for nodePort service? We have function `reconcileConntrackUponEndpointSliceEvents` in `DefaultNodeNetworkController`. Why don't we handle service event in `DefaultNodeNetworkController` too?

go-controller/pkg/node/gateway_shared_intf.go
arkadeepsen: I had initially thought of this approach but upon analyzing the code complexity, I went ahead with this approach.

If we are deleting all UDP conntrack entries upon service ip/port change, we have to go through each port (say n) and service IP (say n), then the overall complexity is O(n^2). However, we do call the `DeleteConntrackServicePort` function for each of these combinations. Which also adds to the overall time taken.

With the current approach, we traverse the port and IP combination for both old and new service but since they are done in separate loops the overall complexity still remains at O(n^2). Additionally, we are also skipping the call to `DeleteConntrackServicePort` function when the combination is found in both the services. With this approach we are also able to delete the UDP conntrack entries which needs to be deleted and the other entries are not impacted.

I can add an additional check in the function itself to first check if there are any changes in the service ip/port itself and skip the function if there are no changes.

go-controller/pkg/node/gateway_shared_intf.go
arkadeepsen: The second value returned is of type `[]*net.IPNet` and not of error type.

arkadeepsen:
Last push was a rebase on the master branch

pliurh:
LGTM

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +925,131 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+		if oldPort.Protocol != corev1.ProtocolUDP {
+			continue
+		}
+		for _, oldIP := range oldIPs {
+			port := oldPort.Port
+			ipType := "svcVIP"
+			if portType == "nodePort" {
+				port = oldPort.NodePort
+				ipType = "nodeIP"
+			}
```
jcaamano: Why do we have this handling for `oldServicePorts` but not for `newServicePorts` which we are comparing against?

go-controller/pkg/node/gateway_localnet_linux_test.go
```diff
@@ -1767,6 +1769,268 @@ var _ = Describe("Node Operations", func() {
...
+					return true
+				}, "1s", "100ms").Should(BeTrue())
+
+				return nil
+			}
+			err := app.Run([]string{app.Name})
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("deletes conntrack entries for UDP when ExternalIP changes", func() {
```
jcaamano, line 1975: do we need coverage for LoadBalancer IP changes as well?

go-controller/pkg/node/gateway_localnet_linux_test.go
```diff
@@ -1767,6 +1769,268 @@ var _ = Describe("Node Operations", func() {
 		})
 	})
 
+	Context("on update", func() {
+		It("deletes conntrack entries for UDP ports when target port changes", func() {
```
jcaamano, line 1773: Do we need coverage when the port changes?

go-controller/pkg/node/gateway_shared_intf.go
arkadeepsen: Thanks for catching that. I missed it when refactoring the code into this helper function.

arkadeepsen:
> I was wondering, without the use of labels, how can we be sure we are not deleting valid entries in case we are, for example, reacting late to an update that is now somehow stale.

How should I use labels to guard against stale update events?

arkadeepsen:
Last push was a rebase

jcaamano:
> > I was wondering, without the use of labels, how can we be sure we are not deleting valid entries in case we are, for example, reacting late to an update that is now somehow stale.
> 
> How should I use labels to guard against stale update events?

I would expect OVN to let us configure these labels in the LB, or that OVN somehow generates labels from some LB fields we have control of, so that we can correlate what service the conntrack entries are for. So that then we can avoid deleting conntrack entries for a different service if we are processing a stale entry.

In any case, our first point of agreement should be: is my concern valid? Or am I missing something?

arkadeepsen:
> > > I was wondering, without the use of labels, how can we be sure we are not deleting valid entries in case we are, for example, reacting late to an update that is now somehow stale.
> > 
> > 
> > How should I use labels to guard against stale update events?
> 
> I would expect OVN to let us configure these labels in the LB, or that OVN somehow generates labels from some LB fields we have control of, so that we can correlate what service the conntrack entries are for. So that then we can avoid deleting conntrack entries for a different service if we are processing a stale entry.
> 
> In any case, our first point of agreement should be: is my concern valid? Or am I missing something?

As we are only deleting UDP conntrack entries, it should not impact the connectivity. Even for valid UDP conntrack entry deletion, the packet will consult the rules installed by OVNK and the traffic should continue with minimal impact in the worst case.
@pliurh had a suggestion for deletion of all the UDP conntrack entries for the service which also was based on this idea: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5715#discussion_r2514223129

jcaamano:
> > > > I was wondering, without the use of labels, how can we be sure we are not deleting valid entries in case we are, for example, reacting late to an update that is now somehow stale.
> > > 
> > > 
> > > How should I use labels to guard against stale update events?
> > 
> > 
> > I would expect OVN to let us configure these labels in the LB, or that OVN somehow generates labels from some LB fields we have control of, so that we can correlate what service the conntrack entries are for. So that then we can avoid deleting conntrack entries for a different service if we are processing a stale entry.
> > In any case, our first point of agreement should be: is my concern valid? Or am I missing something?
> 
> As we are only deleting UDP conntrack entries, it should not impact the connectivity. Even for valid UDP conntrack entry deletion, the packet will consult the rules installed by OVNK and the traffic should continue with minimal impact in the worst case. @pliurh had a suggestion for deletion of all the UDP conntrack entries for the service which also was based on this idea: [#5715 (comment)](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5715#discussion_r2514223129)

But we might load balance again selecting a ~~new~~ different endpoint and disrupting the connection from the application layer perspective.

jcaamano:
You might need to unskip this now
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/5118b213a0c5f0772867808f037a05519a5d4499/test/scripts/e2e-kind.sh#L73

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +925,139 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+		ipType := "svcVIP"
+		if portType == "nodePort" {
+			port = oldPort.NodePort
+			ipType = "nodeIP"
+		}
+		for _, oldIP := range oldIPs {
+			matchString := util.JoinServiceIPPortTargetPort(ipToString(oldIP), oldPort.Port, oldPort.TargetPort)
+			if portType == "nodePort" {
+				matchString = fmt.Sprintf("%s:%d", matchString, oldPort.NodePort)
+			}
```
jcaamano: ~~Not using IPs here?~~ This is confusing, the way you have a specific method `JoinServiceIPPortTargetPort` and all the comments referencing `IP:port:targetPort` but then having specific cases were this is not what you are actually doing. why not resolve it within `JoinServiceIPPortTargetPort`?

edit: realized this is `ip:port:targetPort:nodePort`? But I guess it is confusing anyway.

arkadeepsen: I updated the code based on this comment: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5715#discussion_r2536384316

The `matchString` already has the `IP:port:targetPort` details. For node port the `matchString` will also contain the node port as well. This case is for node port only which checks the `IP:port:targetPort:NodePort` so that even if no changes are there specific for node IP, node port and target port, if there is  changes in the port, then we should delete the conntrack entry for node port.

arkadeepsen: I'll update the code comments accordingly to reflect this scenario.

arkadeepsen:
> > > > > I was wondering, without the use of labels, how can we be sure we are not deleting valid entries in case we are, for example, reacting late to an update that is now somehow stale.
> > > > 
> > > > 
> > > > How should I use labels to guard against stale update events?
> > > 
> > > 
> > > I would expect OVN to let us configure these labels in the LB, or that OVN somehow generates labels from some LB fields we have control of, so that we can correlate what service the conntrack entries are for. So that then we can avoid deleting conntrack entries for a different service if we are processing a stale entry.
> > > In any case, our first point of agreement should be: is my concern valid? Or am I missing something?
> > 
> > 
> > As we are only deleting UDP conntrack entries, it should not impact the connectivity. Even for valid UDP conntrack entry deletion, the packet will consult the rules installed by OVNK and the traffic should continue with minimal impact in the worst case. @pliurh had a suggestion for deletion of all the UDP conntrack entries for the service which also was based on this idea: [#5715 (comment)](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5715#discussion_r2514223129)
> 
> But we might load balance again selecting a ~new~ different endpoint and disrupting the connection from the application layer perspective.

On the original point of deletion of valid entries, this would mean that the combination of `ip:port:targetPort` transiently changed then again came back to the same value. Now, there should be at least 2 update events:
- first one which transiently changed the combination
- the last one got the combination back to the original value

The cache might receive the late update events at the same time or the receipt of the last update event might be delayed even further.
In the first scenario, when both the events are received at the same time:
- The `WatchResourceFiltered` should ensure that we use the latest service object from the cache:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/5118b213a0c5f0772867808f037a05519a5d4499/go-controller/pkg/retry/obj_retry.go#L593
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/5118b213a0c5f0772867808f037a05519a5d4499/go-controller/pkg/retry/obj_retry.go#L680
- This will ensure that we use the latest service object and the UDP conntrack entry deletion of the valid entry will not be performed.

In the second scenario, when the last event is received with further delay:
- During the transient change in the `ip:port:targetPort` combination, the UDP connection will fail as the event wasn't received in time. However, the connection should start working when the `ip:port:targetPort` combination goes back to the original value.
- When the first update event is processed, it will result in the deletion of the valid UDP conntrack entry and as you mentioned the the backend for the lb might change due to this.

The second scenario might be a rare one, and since this is specific for UDP services, the connection should ideally continue with a minimal disruption.

jcaamano:
> The second scenario might be a rare one, and since this is specific for UDP services, the connection should ideally continue with a minimal disruption.

Just want to highlight this might not be a minimal disruption. UDP means no connection and sate on the transport layer, but there can still be on the higher layers.

Anyway, bear with me, need a bit more time to understand if there is nothing better we can do.

go-controller/pkg/node/gateway_shared_intf.go
arkadeepsen: Updated it here: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/9ce64b2bed41c5667bff5e14e624aba86eaaea05..418ff82666a354858b685d461f60efe6f2ef4dc2

jcaamano:
> Anyway, bear with me, need a bit more time to understand if there is nothing better we can do.

I guess we are doing more of the same thing as what we do in the `DeleteService` or `UpdateEndpointSlices` flow, so not really introducing anything new in this regard.

I wonder why we are not using the `ct_flush` LB option, there must be some history about it. I will ask.

Another thing that comes to mind is, the delete code paths that delete conntrack don't really care it is TCP or UDP while update code paths only care about UDP. I wonder why, I might or I would be missing an explanation somewhere about it.

go-controller/pkg/util/kube.go
```diff
@@ -953,3 +953,13 @@ func IsLastUpdatedByManager(manager string, managedFields []metav1.ManagedFields
 	}
 	return lastUpdateOurs.After(lastUpdateTheirs)
 }
+
+// JoinIPServicePort returns the IP:port:targetPort:nodePort combination of a service port for
+// a NodePort type service, otherwise returns IP:port:targetPort combination of the service port.
+func JoinIPServicePort(ip string, servicePort corev1.ServicePort, isNodePort bool) string {
```
jcaamano: I think I would rather have this method in `gateway_shared_intf.go` as it is pretty tailored to what you are doing there instead of generally useful?

jcaamano:
> I wonder why we are not using the `ct_flush` LB option, there must be some history about it. I will ask.
> 
> Another thing that comes to mind is, the delete code paths that delete conntrack don't really care it is TCP or UDP while update code paths only care about UDP. I wonder why, I might or I would be missing an explanation somewhere about it.

Tracked a decision being made [here](https://github.com/kubernetes/kubernetes/issues/108523#issuecomment-1074044415)
The motivation for UDP specific handling is [here](https://github.com/kubernetes/kubernetes/issues/108523#issuecomment-1059392164). It's regarding endpoint deletion only and makes sense to extend it to service updates although there is the open question of what handling we should have for TCP on service updates. I potentially see it as a case of delete/add service and thus TCP should break as well otherwise we could potentially not be handling other services correctly. In any case, since this seems to be still largely implementation defined and we are only breaking UDP conformance, we can address this later.

I guess handling TCP differently than UDP makes using `ct_flush` not the best approach.

A great follow up would be to doc this and link to doc from comments close to this code.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -895,6 +896,16 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+
+		// Delete conntrack entries for UDP ports of the service if they have changed.
+		// UDP is connectionless, so stale conntrack entries can cause packets to be routed to
+		// old target ports even after service updates. Unlike TCP (which is stateful and manages
+		// its own connection lifecycle), UDP conntrack entries must be explicitly cleaned up.
+		// This behavior is extending the decision taken regarding enpoint deletion of UDP services.
+		// For more details, see https://github.com/kubernetes/kubernetes/issues/108523#issuecomment-1074044415.
+		if err = npw.deleteConntrackForUpdatedUDPServiceIPPorts(old, new); err != nil {
+			errors = append(errors, fmt.Errorf("failed to delete conntrack entries for UDP ports of service %v: %v", name, err))
+		}
```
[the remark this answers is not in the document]
arkadeepsen: Updated the comment

arkadeepsen:
> A great follow up would be to doc this and link to doc from comments close to this code.

I have added a link to https://github.com/kubernetes/kubernetes/issues/108523#issuecomment-1074044415 in the comments block of the `UpdateService` where the conntrack delete function is being called.

I'll create a follow up PR for documenting this and let you know when it's ready.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +927,144 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+
+	// Node IPs
+	var newNodeIPs []net.IP
+	if util.ServiceTypeHasNodePort(new) {
+		newNodeIPs, _ = npw.nodeIPManager.ListAddresses()
+	}
+	var oldNodeIPs []net.IP
+	if util.ServiceTypeHasNodePort(old) {
+		oldNodeIPs, _ = npw.nodeIPManager.ListAddresses()
+	}
```
jcaamano, line 974: Sorry I did not notice this before and it might be a nit, but wouldn't it be better if we made a single consistent call to `npw.nodeIPManager.ListAddresses()` since we don't want to handle updates of node IPs?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +927,144 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+		ipType := "svcVIP"
+		portType := "svcPort"
+		if isNodePort {
+			port = oldPort.NodePort
+			ipType = "nodeIP"
+			portType = "nodePort"
+		}
+		for _, oldIP := range oldIPs {
+			if !newIPPortTargetPorts.Has(JoinIPServicePort(ipToString(oldIP), oldPort, isNodePort)) {
+				if err := util.DeleteConntrackServicePort(ipToString(oldIP), port, oldPort.Protocol,
```
jcaamano: We have some inconsistent logging when calling `DeleteConntrackServicePort`. [On some places we log](https://github.com/arkadeepsen/ovn-kubernetes/blob/b43a212c4f467c91595917146be582153e26b998/go-controller/pkg/node/default_node_network_controller.go#L1375), here we don't.

go-controller/pkg/node/gateway_shared_intf.go
arkadeepsen: Updated the code to call `npw.nodeIPManager.ListAddresses()`. The latest change should do it correctly.

go-controller/pkg/node/gateway_shared_intf.go
arkadeepsen: The logging is now done only inside `util.DeleteConntrack` function. Removed logging in other places.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -916,6 +927,148 @@ func (npw *nodePortWatcher) UpdateService(old, new *corev1.Service) error {
...
+	nodePortDeletedCount, nodePortErrs := deleteStaleUDPConntrackForUDPPortsAndIPs(new.Spec.Ports, newNodeIPs, old.Spec.Ports, oldNodeIPs, old.Namespace, old.Name, true)
+
+	deletedCount = extIPsDeletedCount + clusterIPsDeletedCount + nodePortDeletedCount
+	errs = append(errs, extIPsErrs...)
+	errs = append(errs, clusterIPsErrs...)
+	errs = append(errs, nodePortErrs...)
+
+	if deletedCount > 0 {
+		klog.Infof("Deleted %d UDP conntrack entries for service %s/%s", deletedCount, old.Namespace, old.Name)
+	}
```
jcaamano, line 984: nit alert

I think this might be misleading because it doesn't really mean conntrack entries have been deleted. 

If you want this log, what I would do is have `DeleteConntrack` return the number of entries deleted (`ConntrackDeleteFilters` returns this info) and have the caller log it.

Other option is to keep it as `V(5)` so that you can log it with the log you already added to `DeleteConntrack`.

arkadeepsen: Updated the code to return the number of entries deleted and then using that value to log the final deletion count.

jcaamano:
@arkadeepsen aren't you intrigued by the two last @coderrabitai nitpicks? Additionally I would consider not exporting `IPStringer` nor `JoinIPServicePort` since they are not to be used outside of the package. Anyway, let me know if you want to act on it or not. Starting to feel like the nits are never-ending :stuck_out_tongue:

arkadeepsen:
> @arkadeepsen aren't you intrigued by the two last @coderrabitai nitpicks? Additionally I would consider not exporting `IPStringer` nor `JoinIPServicePort` since they are not to be used outside of the package. Anyway, let me know if you want to act on it or not. Starting to feel like the nits are never-ending 😛

Addressed the nitpick comments here: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/471b6b4e249b3a1c290487a709dc0714fcdb7afc..cf71b3a9d849f4f8655f90694087459f8aad116e

Hopefully @coderabbitai doesn't come up with any new one :smiling_face_with_tear:

arkadeepsen:
Last push was a rebase

tssurya:
So did I miss something in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/2829#issuecomment-1075100886 ? where we did the clean up based on the upstream issue? I guess there were scenarios we didn't consider?

arkadeepsen:
The CI lanes with `enable-dns-name-resolver` are failing. Created a fix for it here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5761

jcaamano (approved):

PR #5716, opened 2025-11-07
 Remove --subresource=status from ovnkube.sh get_node_zone

koolzz:
## 📑 Description

Error observed in ovnkube-master. log

```
Error from server (Forbidden): nodes "np-bommr-bks-04.np-bommr.nvmetal.net" is forbidden: User "system:serviceaccount:ovn-kubernetes:ovn-master" cannot get resource "nodes/status" in API group "" at the cluster scope 
```
From [ovkube.sh](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/images/ovnkube.sh#L861)

The ovn-master cluster role needs access to nodes/status subresource, but the ovkube.sh get_node_zone is used by a few different identities so instead of modifying multiple rbac files, just remove the `--subresource=status` from the kubectl cmd.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Chores**
  * Updated RBAC permissions for OVN Kubernetes node controller to include list and watch operations on nodes in addition to read access.
  * Renamed role binding for improved naming consistency.

### DIFF
--- a/dist/templates/rbac-ovnkube-node.yaml.j2
+++ b/dist/templates/rbac-ovnkube-node.yaml.j2
@@ -35,9 +35,9 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-    name: node-reader
+    name: ovnkube-node-reader
 roleRef:
-    name: node-reader
+    name: ovnkube-node-reader
     kind: ClusterRole
     apiGroup: rbac.authorization.k8s.io
 subjects:
@@ -114,7 +114,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-    name: node-reader
+    name: ovnkube-node-reader
 rules:
     - apiGroups: [""]
       resources:
--- a/helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
+++ b/helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
@@ -35,9 +35,9 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-    name: ovnkube-node-status-reader
+    name: ovnkube-node-reader
 roleRef:
-    name: ovnkube-node-status-reader
+    name: ovnkube-node-reader
     kind: ClusterRole
     apiGroup: rbac.authorization.k8s.io
 subjects:
@@ -114,12 +114,12 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-    name: ovnkube-node-status-reader
+    name: ovnkube-node-reader
 rules:
     - apiGroups: [""]
       resources:
-          - nodes/status
-      verbs: [ "get" ]
+          - nodes
+      verbs: [ "get", "list", "watch" ]
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1


kyrtapz:
LGTM
CI failures look unrelated - this change is for non-ic only.

kyrtapz (dismissed):

kyrtapz:
Looking a bit more into it having the `--subresource=status` here:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e3d9429b5edd7eb6ed06291b71d5e7a9ed91cc1e/dist/images/ovnkube.sh#L861
doesn't do much as we are going to get the full object anyway.
So an alternative solution would be to modify https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L34-L46
to allow the SA to read the node object without the status subresource.
This would mean the ovnkube-master doesn't need the extra RBAC but it doesn't really matter which way we go so I still fine with the proposed apporach.

koolzz:
> Looking a bit more into it having the `--subresource=status` here:
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e3d9429b5edd7eb6ed06291b71d5e7a9ed91cc1e/dist/images/ovnkube.sh#L861
> 
> 
> doesn't do much as we are going to get the full object anyway.
> So an alternative solution would be to modify https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L34-L46
> to allow the SA to read the node object without the status subresource.
> This would mean the ovnkube-master doesn't need the extra RBAC but it doesn't really matter which way we go so I still fine with the proposed apporach.

I've actually just updated it to just nuke the `--subresource=status`, lmk :)

Also seems like the nodes is already specified here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L133

kyrtapz:
> > Looking a bit more into it having the `--subresource=status` here:
> > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e3d9429b5edd7eb6ed06291b71d5e7a9ed91cc1e/dist/images/ovnkube.sh#L861
> > 
> > doesn't do much as we are going to get the full object anyway.
> > So an alternative solution would be to modify https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L34-L46
> > to allow the SA to read the node object without the status subresource.
> > This would mean the ovnkube-master doesn't need the extra RBAC but it doesn't really matter which way we go so I still fine with the proposed apporach.
> 
> I've actually just updated it to just nuke the `--subresource=status`, lmk :)

I would expect only removing it will cause ovnkube-node to fail on startup.

koolzz:
> > > Looking a bit more into it having the `--subresource=status` here:
> > > https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e3d9429b5edd7eb6ed06291b71d5e7a9ed91cc1e/dist/images/ovnkube.sh#L861
> > > 
> > > doesn't do much as we are going to get the full object anyway.
> > > So an alternative solution would be to modify https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L34-L46
> > > to allow the SA to read the node object without the status subresource.
> > > This would mean the ovnkube-master doesn't need the extra RBAC but it doesn't really matter which way we go so I still fine with the proposed apporach.
> > 
> > 
> > I've actually just updated it to just nuke the `--subresource=status`, lmk :)
> 
> I would expect only removing it will cause ovnkube-node to fail on startup.

Ah I see what you mean, I added node access to `ovnkube-node-status-reader`

kyrtapz:
> Also seems like the nodes is already specified here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L133

It is more complex unfortunately. The `ovnkube-node` role is only applied to this group https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L20-L22.
It represents the certificate user that is created and handled in the go binary. 
When the webhook is enabled the ovnkube-node service account could ONLY read the nodes/status: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/dist/templates/rbac-ovnkube-node.yaml.j2#L30-L46.
When I tried to only remove the subresource switch I did see the error log:
```
ovn_metrics_bind_address=172.18.0.3:9476
metrics_bind_address=172.18.0.3:9410
ovnkube_metrics_tls_opts=
ovnkube_config_duration_enable_flag: --metrics-enable-config-duration
Error from server (Forbidden): nodes "ovn-worker2" is forbidden: User "system:serviceaccount:ovn-kubernetes:ovnkube-node" cannot get resource "nodes" in API group "" at the cluster scope
```

So we need to adapt that role too.

dist/templates/rbac-ovnkube-node.yaml.j2
```diff
@@ -118,8 +118,8 @@ metadata:
 rules:
     - apiGroups: [""]
       resources:
-          - nodes/status
-      verbs: [ "get" ]
+          - nodes
```
kyrtapz, line 121: Let's rename the ClusterRole to `node-reader` or similar? I would hate to go with `ovnkube-node-node-reader` :D

kyrtapz:
Another small improvement which we could(but don't have to) do is to modify this function https://github.com/ovn-kubernetes/ovn-kubernetes/blob/e3d9429b5edd7eb6ed06291b71d5e7a9ed91cc1e/dist/images/ovnkube.sh#L860-L871 to return an error if it was not able to get the node instead of quietly ignoring all errors.

This is sort of a separate issue so feel free to ignore it but if we got it right the first time we would get a better signal.

dist/templates/rbac-ovnkube-node.yaml.j2
koolzz: Hey sorry, to clarify this, do you wan to drop the ovnkube-node prefix? Woudn't that make it confusing to identify the clusterrole? 

Also, do I need to manually change the `helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml` file, or is there some automation to copy the changes?

koolzz:
> etly ignoring all errors.
> This is sort of a separate issue so feel free to ignore it but if we

In case of error would we want to terminate the script via `exit 1`?

dist/templates/rbac-ovnkube-node.yaml.j2
kyrtapz: I am fine with what is there now, I am also good with `ovnkube-node-reader` if that makes it more obvious it is a ovn-k role.
As for the helm charts, we need to modify those manually.

kyrtapz:
> > etly ignoring all errors.
> > This is sort of a separate issue so feel free to ignore it but if we
> 
> In case of error would we want to terminate the script via `exit 1`?

yeah, that should be enough.

koolzz:
> > > etly ignoring all errors.
> > > This is sort of a separate issue so feel free to ignore it but if we
> > 
> > 
> > In case of error would we want to terminate the script via `exit 1`?
> 
> yeah, that should be enough.

let me add that as separate pr, for now updated the helm rbac file, lmk if looks good. ty

kyrtapz:
Looks good, thanks @koolzz

trozet (approved):
will wait for green CI then merge

PR #5721, opened 2025-11-13
add eip e2e cases on ns/pod labels update

huiran0826:
## 📑 Description

This is one e2e test coverage for end user bug https://issues.redhat.com/browse/OCPBUGS-61742
The scenarios is:

1. Two EIPs exist:

- eip1 matches namespace label test: qe
- eip2 matches namespace label test: dev

2. There is a pod in test namespace
3.  change the label from the namespace from test=qe to test=dev or from test=dev to test=qe for a couple of times
4. Verify the source IP, the reported issue was that srcIP was nodeIP instead of egressIP.

And this case covers below scenarios: 
should handle EIP reassignment correctly on namespace label update 
should handle EIP reassignment correctly on pod label update 
should handle EIP reassignment correctly on EgressIP object update

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Tests**
  * Added an end-to-end test validating EgressIP reassignment across repeated namespace, pod, and EgressIP selector updates. Verifies correct IP assignment, SNAT behavior, connectivity to external endpoints and node IPs, status length checks after changes, cleanup of CRD configuration, and deterministic reassignment across multiple iterations.

### DIFF
--- a/test/e2e/egressip.go
+++ b/test/e2e/egressip.go
@@ -225,6 +225,40 @@ func isSupportedAgnhostForEIP(externalContainer infraapi.ExternalContainer) bool
 	return true
 }
 
+// Create EgressIP Manifest
+func createEIPManifest(name string, podLabel, namespaceLabel map[string]string, egressIPs ...string) string {
+	var ipsYAML string
+	for _, ip := range egressIPs {
+		ipsYAML += fmt.Sprintf("\n    - %s", ip)
+	}
+
+	var podLabelYaml string
+	for k, v := range podLabel {
+		podLabelYaml = fmt.Sprintf("%s: %s", k, v)
+	}
+
+	var namespaceLabelYaml string
+	for k, v := range namespaceLabel {
+		namespaceLabelYaml = fmt.Sprintf("%s: %s", k, v)
+	}
+	egressIPConfig := fmt.Sprintf(`apiVersion: k8s.ovn.org/v1
+kind: EgressIP
+metadata:
+    name: %s
+spec:
+    egressIPs:%s
+    podSelector:
+        matchLabels:
+            %s
+    namespaceSelector:
+        matchLabels:
+            %s
+`, name, ipsYAML, podLabelYaml, namespaceLabelYaml)
+
+	return egressIPConfig
+
+}
+
 // targetHostNetworkContainerAndTest targets the internal host network test container from
 // our test pods, collects its logs and verifies that the logs have traces
 // of the `verifyIPs` provided. We need to target the test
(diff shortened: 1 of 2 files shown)


huiran0826:
cc @pperiyasamy

test/e2e/egressip.go
```diff
@@ -691,6 +691,46 @@ var _ = ginkgo.DescribeTableSubtree("e2e egress IP validation", feature.EgressIP
...
+	verifyLRPRule := func(srcPodIP net.IP, podOVNPod string, dbContainerName string, ovnKubernetesNamespace string) bool {
+		lrpCmd := "ovn-nbctl lr-policy-list ovn_cluster_router | grep -v inport | grep reroute"
+		lrps, err := e2ekubectl.RunKubectl(ovnKubernetesNamespace, "exec", podOVNPod, "-c", dbContainerName, "--", "sh", "-c", lrpCmd)
+		if err != nil {
+			framework.Logf("Error: Check the OVN DB to ensure LRP is added, err: %v", err)
+			return false
+		}
+		return strings.Contains(lrps, srcPodIP.String()) && strings.Count(lrps, "100 ") == 1
+	}
+
```
pperiyasamy: checking SNAT and LRP rules not needed for E2Es. traffic test would cover these. look at other EIP tests for reference.

test/e2e/egressip.go
```diff
@@ -3073,6 +3113,440 @@ spec:
...
+		var egressIP1, egressIP2 net.IP
+		var err error
+		if utilnet.IsIPv6String(egress1Node.nodeIP) {
+			egressIP1, err = ipalloc.NewPrimaryIPv6()
+			egressIP2, err = ipalloc.NewPrimaryIPv6()
+		} else {
+			egressIP1, err = ipalloc.NewPrimaryIPv4()
+			egressIP2, err = ipalloc.NewPrimaryIPv4()
+		}
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new Node IP")
```
pperiyasamy: it's not Node IP, must be Egress IP. 
this check must be done for every IP allocation.

test/e2e/egressip.go
```diff
@@ -3073,6 +3113,440 @@ spec:
...
+
+		podNamespace := f.Namespace
+		nsEIPLabels := map[string]string{
+			"wants": "egress",
+		}
+		nsEIPLabels2 := map[string]string{
+			"wants": "egress2",
+		}
+
+		var egressIPConfig = `apiVersion: k8s.ovn.org/v1
```
pperiyasamy: all three tests having some duplicate code for things like creating EgressIPs, pod creation etc, it can be moved to helper methods.

test/e2e/egressip.go
```diff
@@ -3073,6 +3113,440 @@ spec:
...
+				return !verifySNATRule(egressIPName, egressIP1, srcPodIP, ovnPodEIP1, dbContainerName, ovnKubernetesNamespace)
+			}, retryTimeout).Should(gomega.BeTrue(), fmt.Sprintf("snat rule is not correct for EgressIP %s or stale rules exist for %s (iteration %d)", egressIP2.String(), egressIP1.String(), i))
+			gomega.Eventually(func() bool {
+				return verifyLRPRule(srcPodIP, podOVNPod, dbContainerName, ovnKubernetesNamespace)
+			}, retryTimeout).Should(gomega.BeTrue(), fmt.Sprintf("lrp rule is not correct (iteration %d)", i))
+		}
+
+	})
+
+	ginkgo.It("should handle EIP reassignment correctly on pod label update", func() {
```
pperiyasamy: This test must not be about having one EIP object and assigning it between two pods. It should be having two EIP objects, update on those EIPs should cause pod moving from one EIP to another.

test/e2e/egressip.go
huiran0826: sure, added traffic check instead of snat/lrp

test/e2e/egressip.go
huiran0826: Combined 3 cases into one

test/e2e/egressip.go
huiran0826: Updated it.

test/e2e/egressip.go
```diff
@@ -3072,6 +3072,171 @@ spec:
...
+		}
+
+		ginkgo.By("9. Update both EgressIP objects that namespace having same labels, and different pod selectors labels")
+		e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
+		e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName2, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress2\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
+
+		ginkgo.By("10. Check that status of both EgressIP objects is of length one")
+		verifySpecificEgressIPStatusLengthEquals(egressIPName, 1, nil)
+		verifySpecificEgressIPStatusLengthEquals(egressIPName2, 1, nil)
+		podNamespace = getNamespace(f, podNamespace.Name)
```
pperiyasamy, line 3365: so at this time, pod is not managed by any EIPs ? can you add a traffic test to confirm it (i.e. external container much receive the pod traffic SNATed with only pod hosted node IP) ?

test/e2e/egressip.go
```diff
@@ -3072,6 +3072,171 @@ spec:
...
+		framework.ExpectNoError(err, "failed to create pod %s/%s", f.Namespace.Name, pod1Name)
+		framework.Logf("Created pod %s on node %s", pod1Name, pod1Node.name)
+
+		// Run namespace label updates multiple times to ensure EIP reassignment works well
+		for i := 1; i <= 5; i++ {
+			ginkgo.By(fmt.Sprintf("5.%d. Update namespace labels match egressIP object1%s selectors (iteration %d)", i, egressIPName, i))
+			podNamespace = getNamespace(f, podNamespace.Name)
+			updateNamespaceLabels(f, podNamespace, egressLabels)
+			ginkgo.By(fmt.Sprintf("6.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from object1 %s (iteration %d)", i, egressIPName, i))
+			err := wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP1.String()}).WithContext())
```
pperiyasamy: can you add a check external container not receiving traffic with node IP ? this will confirm no stale LRP nexthops.

test/e2e/egressip.go
```diff
@@ -3072,6 +3072,171 @@ spec:
...
+			framework.ExpectNoError(err, "6.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from object1 %s (iteration %d): %v", i, egressIPName, i, err)
+
+			ginkgo.By(fmt.Sprintf("7.%d. Update namespace labels to match egressIP object2 %s selectors (iteration %d)", i, egressIPName2, i))
+			podNamespace = getNamespace(f, podNamespace.Name)
+			updateNamespaceLabels(f, podNamespace, egressLabels2)
+			ginkgo.By(fmt.Sprintf("8.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from object2 %s (iteration %d)", i, egressIPName2, i))
+			err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP2.String()}).WithContext())
```
pperiyasamy: same here. also for pod/eip update cases as well.

test/e2e/egressip.go
```diff
@@ -3072,6 +3072,171 @@ spec:
...
+		ginkgo.By("9. Update both EgressIP objects that namespace having same labels, and different pod selectors labels")
+		e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
+		e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName2, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress2\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
+
+		ginkgo.By("10. Check that status of both EgressIP objects is of length one")
+		verifySpecificEgressIPStatusLengthEquals(egressIPName, 1, nil)
+		verifySpecificEgressIPStatusLengthEquals(egressIPName2, 1, nil)
+		podNamespace = getNamespace(f, podNamespace.Name)
+		updateNamespaceLabels(f, podNamespace, egressLabels)
+
```
pperiyasamy, line 3372: now it should be served by `egressIPName`, can you add traffic test ?

test/e2e/egressip.go
```diff
@@ -3072,6 +3072,171 @@ spec:
...
+			pod1.Labels = egressLabels2
+			updatePod(f, pod1)
+			ginkgo.By(fmt.Sprintf("14.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from object2 %s (iteration %d)", i, egressIPName2, i))
+			err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP2.String()}).WithContext())
+			framework.ExpectNoError(err, "14.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from object2 %s (iteration %d): %v", i, egressIPName2, i, err)
+		}
+
+		// Run EIP object updates multiple times to ensure EIP reassignment works well
+		for i := 1; i <= 5; i++ {
+			ginkgo.By(fmt.Sprintf("15.%d. Update EgressIP object1 %s selectors to match pod labels and EgressIP object2 %s not matching pod labels,(iteration %d)", i, egressIPName, egressIPName2, i))
```
pperiyasamy: you may avoid using `object1` and `object2` since EIP name already coveys that.

test/e2e/egressip.go
```diff
@@ -3072,6 +3072,171 @@ spec:
...
+    namespaceSelector:
+        matchLabels:
+            wants: egress
+`
+		if err := os.WriteFile(egressIPYaml, []byte(egressIPConfig), 0644); err != nil {
+			framework.Failf("Unable to write CRD config to disk: %v", err)
+		}
+		defer func() {
+			if err := os.Remove(egressIPYaml); err != nil {
+				framework.Logf("Unable to remove the CRD config from disk: %v", err)
```
pperiyasamy: don't you need to delete EIP object at the end of the test ? it will ensure it won't conflict with other tests.

test/e2e/egressip.go
pperiyasamy: thanks.

test/e2e/egressip.go
huiran0826: This defer function will delete the EIP. I checked the env after the case run, EIPs were deleted successfully.

test/e2e/egressip.go
huiran0826: Updated it.

test/e2e/egressip.go
huiran0826: Added it.

test/e2e/egressip.go
huiran0826: yes, it served as node IP, added the check.

test/e2e/egressip.go
huiran0826: I think current way is to verify the egress IP is present, which implicitly means the node IP is not used. 
Or you want me to add something  like this to explicitly verify it? 
```
ginkgo.By("Verify source IP is NOT the node IP")
err = wait.PollImmediate(retryInterval, retryTimeout, targetExternalContainerAndTest(
    primaryTargetExternalContainer, 
    podNamespace.Name, 
    pod1Name, 
    true, 
    []string{pod1Node.nodeIP}))  

// If node IP is NOT found, verifyIPs will never become empty,
// so the function keeps retrying and eventually times out
// This means the wait will return an error, which is what we want!
framework.ExpectError(err, "Node IP should NOT be used as source IP - connection should succeed but node IP should not be found")
```

test/e2e/egressip.go
pperiyasamy: ok, it actually got deleted in `AfterEach` function https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/egressip.go#L845-L846. the defer function only deletes file.

test/e2e/egressip.go
pperiyasamy: yes, it's good to check this as well to confirm LRP not having any stale nexthop taking nodeIP path.

huiran0826: Added the check

test/e2e/egressip.go
huiran0826: added the check

pperiyasamy:
LGTM

Nice work @huiran0826 !

huiran0826:
@tssurya

huiran0826:
@tssurya would you please take a look?  Thanks!

tssurya:
> @tssurya would you please take a look? Thanks!

Hey @huiran0826 thanks for the reminder, sorry I didn't pay attention to this.. I'm on it this week.

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
+	/*
+		This test does the following:
+		0. Add the "k8s.ovn.org/egress-assignable" label to two nodes
+		1. Create an EgressIP object with one egress IP defined
+		2. Create an EgressIP object with one egress IP defined
+		3. Check that status of both EgressIP objects is of length one
+		4. Create one pod matching the EgressIP
+		5. Run namespace label updates multiple times to ensure EIP reassignment works well
```
tssurya: I'm not sure I see the why do we need this PR in the PR description / commit message.
Was there some bug that was not caught cause we missed coverage for this?

Reason I ask is all label updates do have unit test coverage that ensures EIP setup is done correctly. While I do see value in us checking for the connectivity aspects and having e2e coverage, I feel like I am missing some point in why its multiple times update scenario etc..

tssurya (changes requested):
misses some motivation behind the change if any before I start reviewing the test case code and it would be good to call out the benefit of this test v/s the namespace and pod label update unit tests we have that ensures the setUP is done properly
If there was some background bug let's capture that here.

test/e2e/egressip.go
huiran0826: As we are adding e2e test coverage for EIP customer bugs as requested and this is for https://issues.redhat.com/browse/OCPBUGS-61742. 

The implementation closely mirrors the customer's original scenario, which involved repeatedly updating a label to trigger the reported issue.

A benefit I think is that these e2e tests can be brought downstream in the future to ensure regression coverage and quality for our downstream product.

tssurya: @huiran0826 since this is upstream project repo with no downstream ties, could you please update the PR description with the bug scenario description on what the issue was OR else link the PR https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5704 to your current PR as adding tests for that fix.

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
 		err = wait.PollImmediate(retryInterval, retryTimeout, targetExternalContainerAndTest(primaryTargetExternalContainer, pod2OtherNetworkNamespace, pod2Name, true, []string{egressIP1.String()}))
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
+	/*
+		This test does the following:
+		0. Add the "k8s.ovn.org/egress-assignable" label to two nodes
+		1. Create an EgressIP object with one egress IP defined
```
tssurya:
```suggestion
		1. Create an EgressIP-A object with one egress IPA defined
```

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		2. Create an EgressIP object with one egress IP defined
```
tssurya:
```suggestion
		2. Create an EgressIP-B object with one egress IPB defined
```

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		3. Check that status of both EgressIP objects is of length one
+		4. Create one pod matching the EgressIP
```
tssurya: matches which one? EgressIP-A or B?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
 		err = wait.PollImmediate(retryInterval, retryTimeout, targetExternalContainerAndTest(primaryTargetExternalContainer, pod2OtherNetworkNamespace, pod2Name, true, []string{egressIP1.String()}))
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
+	/*
+		This test does the following:
+		0. Add the "k8s.ovn.org/egress-assignable" label to two nodes
```
tssurya: do we really need two nodes or will it also be fine with 1 node since its two different objects?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		1. Create an EgressIP object with one egress IP defined
+		2. Create an EgressIP object with one egress IP defined
+		3. Check that status of both EgressIP objects is of length one
+		4. Create one pod matching the EgressIP
+		5. Run namespace label updates multiple times to ensure EIP reassignment works well
+		6. Run pod label updates multiple times to ensure EIP reassignment works well
+		7. Run EIP object updates multiple times to ensure EIP reassignment works well
```
tssurya: step7 is doing what updates to the EIP object?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		}
+
+		ginkgo.By("0. Add the \"k8s.ovn.org/egress-assignable\" label to two nodes")
+		e2enode.AddOrUpdateLabelOnNode(f.ClientSet, egress1Node.name, "k8s.ovn.org/egress-assignable", "dummy")
+		e2enode.AddOrUpdateLabelOnNode(f.ClientSet, egress2Node.name, "k8s.ovn.org/egress-assignable", "dummy")
+
+		ginkgo.By("1. Create an EgressIP object with one egress IP defined")
+		var egressIP1, egressIP2 net.IP
+		var err error
+		var retryTimeout2 = 2 * retryInterval
```
tssurya, line 3258: is this 2 seconds? why are we defining another timeout variable like this and not reusing the global ones like other tests?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		if utilnet.IsIPv6String(egress1Node.nodeIP) {
+			egressIP1, err = ipalloc.NewPrimaryIPv6()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new EgressIP")
+			egressIP2, err = ipalloc.NewPrimaryIPv6()
```
tssurya: nit: move egressIP2 closer to the step where you create the second egressIP object

test/e2e/egressip.go
tssurya: or use object1 and object2 - otherwise both steps 1 and 2 look same

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new EgressIP")
+			egressIP2, err = ipalloc.NewPrimaryIPv4()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new EgressIP")
+		}
+
+		podNamespace := f.Namespace
+		egressLabels := map[string]string{
+			"wants": "egress",
+		}
+		egressLabels2 := map[string]string{
```
tssurya: move this closer to the second object

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+    egressIPs:
+    - ` + egressIP2.String() + `
+    podSelector:
+        matchLabels:
+            wants: egress
+    namespaceSelector:
+        matchLabels:
+            wants: egress2
+`
+		if err := os.WriteFile(egressIPYaml, []byte(egressIPConfig2), 0644); err != nil {
```
tssurya: consider having a small util for EIP creation that can be reused to avoid the code duplication

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+			updateNamespaceLabels(f, podNamespace, egressLabels)
+			ginkgo.By(fmt.Sprintf("6.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from %s (iteration %d)", i, egressIPName, i))
+			err := wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP1.String()}).WithContext())
+			framework.ExpectNoError(err, "6.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from %s (iteration %d): %v", i, egressIPName, i, err)
+			ginkgo.By(fmt.Sprintf("7.%d. Verify source IP is NOT the node IP (iteration %d)", i, i))
+			err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout2, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{pod1Node.nodeIP}).WithContext())
+			gomega.Expect(err).To(gomega.HaveOccurred(), "Node IP should NOT be used as source IP - connection should succeed but node IP should not be found")
```
tssurya, line 3327: this step is weird, if the src IP is confirmed to be EIP then that means its not nodeIP right? what am I missing? at this stage the pod is only matching one EIP object right?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+			err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP2.String()}).WithContext())
+			framework.ExpectNoError(err, "9.%d. Check connectivity from pod to an external container and verify that the srcIP is the expected egressIP from %s (iteration %d): %v", i, egressIPName2, i, err)
+			ginkgo.By(fmt.Sprintf("10.%d. Verify source IP is NOT the node IP (iteration %d)", i, i))
+			err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout2, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{pod1Node.nodeIP}).WithContext())
+			gomega.Expect(err).To(gomega.HaveOccurred(), "Node IP should NOT be used as source IP - connection should succeed but node IP should not be found")
+		}
+
+		ginkgo.By("11. Update both EgressIP objects that namespace having same labels, and different pod selectors labels")
```
tssurya: what does "Update both EgressIP objects that namespace having same labels" mean? do you mean update both egressIP objects such that they have same namespace selector but different pod selector?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
+		e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName2, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress2\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
+
+		ginkgo.By("12. Check that status of both EgressIP objects is of length one")
+		verifySpecificEgressIPStatusLengthEquals(egressIPName, 1, nil)
+		verifySpecificEgressIPStatusLengthEquals(egressIPName2, 1, nil)
+		ginkgo.By("13. Check connectivity from that one to an external \"node\" and verify that the IP is the node IP.")
```
tssurya, line 3359: don't we have other tests for checking node connectivity? isn't this duplicating that?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{pod1Node.nodeIP}).WithContext())
+		framework.ExpectNoError(err, "Step 13. Check connectivity from that one to an external \"node\" and verify that the IP is the node IP, failed, err: %v", err)
+		podNamespace = getNamespace(f, podNamespace.Name)
+		updateNamespaceLabels(f, podNamespace, egressLabels)
```
tssurya, line 3366: this update doesn't have a ginkgo.By block?

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+		This test does the following:
+		0. Add the "k8s.ovn.org/egress-assignable" label to two nodes
+		1. Create an EgressIP object with one egress IP defined
+		2. Create an EgressIP object with one egress IP defined
+		3. Check that status of both EgressIP objects is of length one
+		4. Create one pod matching the EgressIP
+		5. Run namespace label updates multiple times to ensure EIP reassignment works well
+		6. Run pod label updates multiple times to ensure EIP reassignment works well
+		7. Run EIP object updates multiple times to ensure EIP reassignment works well
+	*/
```
tssurya, line 3246: the test has soo many more steps that's not in this description

test/e2e/egressip.go
```diff
@@ -3181,6 +3181,199 @@ spec:
...
+			ginkgo.By(fmt.Sprintf("20.%d. Verify source IP is NOT the node IP (iteration %d)", i, i))
+			err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout2, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{pod1Node.nodeIP}).WithContext())
+			gomega.Expect(err).To(gomega.HaveOccurred(), "Node IP should NOT be used as source IP - connection should succeed but node IP should not be found")
+		}
+
+		// Run EIP object updates multiple times to ensure EIP reassignment works well
+		for i := 1; i <= 5; i++ {
+			ginkgo.By(fmt.Sprintf("21.%d. Update EgressIP %s selectors to match pod labels and EgressIP %s not matching pod labels,(iteration %d)", i, egressIPName, egressIPName2, i))
+			e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName, "-p", "{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress2\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}", "--type=merge")
```
tssurya: use string variab;e for the `"{\"spec\":{\"podSelector\":{\"matchLabels\":{\"wants\": \"egress2\"}},\"namespaceSelector\": {\"matchLabels\":{\"wants\": \"egress\"}}}}"` coupled with the `egressLabels` and `egressLabels2` to make it more readable.

test/scripts/e2e-cp.sh
```diff
@@ -188,6 +188,7 @@ else
     skip "e2e egress IP validation Cluster Default Network Should validate egress IP logic when one pod is managed by more than one egressIP object"
     skip "e2e egress IP validation Cluster Default Network Should re-assign egress IPs when node readiness / reachability goes down/up"
     skip "Pod to external server PMTUD when a client ovnk pod targeting an external server is created when tests are run towards the agnhost echo server queries to the hostNetworked server pod on another node shall work for UDP"
+    skip "e2e egress IP validation Cluster Default Network Should handle EIP reassignment correctly on namespace and pod label updates, and EIP object updates"
```
tssurya, line 204: BGP doesn't work with EIP pods trying to reach other nodes? did it fail in CI?

test/e2e/egressip.go
huiran0826: Thanks, updated it!

test/e2e/egressip.go
huiran0826: updated it.

test/e2e/egressip.go
huiran0826: done

test/e2e/egressip.go
huiran0826: updated it

test/e2e/egressip.go
huiran0826: use one node instead

test/e2e/egressip.go
huiran0826: Updated description to include more detailed steps

test/e2e/egressip.go
huiran0826: @tssurya Regarding your comment `is this 2 seconds? why are we defining another timeout variable like this and not reusing the global ones like other tests?`, because we added this check, and need to wait timeout. 
the default timeout value is 2m that is much longer, so  defining a smaller timeout variable and only used it for this` node IP is not used` check.

test/e2e/egressip.go
huiran0826: updated it

test/e2e/egressip.go
huiran0826: updated the description with more detailed steps

test/e2e/egressip.go
huiran0826: This is check added from previous review comments.

test/scripts/e2e-cp.sh
huiran0826: The step `verify that the IP is the node IP` would fail in BGP CI just like other skipping cases.

test/e2e/egressip.go
huiran0826: added

test/e2e/egressip.go
huiran0826: right

test/e2e/egressip.go
huiran0826: right, please see the previous thread that @pperiyasamy 's comment for this step. I `@`  you there.

test/scripts/e2e-cp.sh
tssurya: do we remember why? I thought for BGP we SNAT to nodeIP if the destination is other nodes

huiran0826: if I remember it correctly, it showed pod IP from log, not node IP.

test/e2e/egressip.go
tssurya: I guess you mean this comment: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2568377539

test/e2e/egressip.go
tssurya: I think this is answered here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2667362505

tssurya: @huiran0826 so is it really 2 seconds though? because isn't that too short a time interval to wait for to ensure nodeIP is not present?

test/e2e/egressip.go
tssurya: could you please then update that statement to what I mentioned cause the way it is now "that namespace" is not clear

test/e2e/egressip.go
tssurya: I meant:

```
specString := fmt.Sprintf("{\"spec\":{\"podSelector\":{\"matchLabels\":%s},\"namespaceSelector\": {\"matchLabels\":%s}}}", string(egressLabels2JSON), string(egressLabelsJSON))
e2ekubectl.RunKubectlOrDie("default", "patch", "EgressIP/"+egressIPName, "-p", specString, "--type=merge")
```
here and elsewhere, please try to curtail really long lines into readable length.

another example:
```
err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout2,true,
              targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name,
               true, []string{pod1Node.nodeIP}).WithContext())
```
is way more readable than a single line we need to scroll

most of the lines here are super long throughout, its a nit but would be good if you could adhere to that standard

tssurya:
thanks @huiran0826 for the updates,

a few more comments that are remaining - some of them you probably missed?:

1. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2669557332
2. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2666083470
3. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2666057654
4. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2666055113
5. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2666051052
6. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2666046571

test/e2e/egressip.go
huiran0826: got it, updated the codes.

test/e2e/egressip.go
huiran0826: updated it.

test/e2e/egressip.go
huiran0826: done

test/e2e/egressip.go
huiran0826: done

test/e2e/egressip.go
huiran0826: done

test/e2e/egressip.go
huiran0826: I think 2 seconds  with 2 rounds checking is fine and before this check, it also has the expected EgressIP check after label update which meant that changing already took effect.

huiran0826:
@tssurya Thanks for taking time to review this, much appreciated!  I addressed all the comments, if I still missed anyone, let me know!

test/e2e/egressip.go
```diff
@@ -3181,6 +3225,215 @@ spec:
...
+		egressLabels := map[string]string{
+			"wants": "egress",
+		}
+		createEgressIP(egressIPName, egressIPYaml, egressLabels, egressLabels, egressIP1.String())
+
+		ginkgo.By("2. Create second EgressIP object with one egress IP2 defined")
+		var egressIP2 net.IP
+		if utilnet.IsIPv6String(egress1Node.nodeIP) {
+			egressIP2, err = ipalloc.NewPrimaryIPv6()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new EgressIP")
```
tssurya: the error assertion is common and can be done outside the if else

test/e2e/egressip.go
```diff
@@ -3181,6 +3225,215 @@ spec:
...
+		ginkgo.By("0. Add the \"k8s.ovn.org/egress-assignable\" label to one node")
+		e2enode.AddOrUpdateLabelOnNode(f.ClientSet, egress1Node.name, "k8s.ovn.org/egress-assignable", "dummy")
+
+		ginkgo.By("1. Create an EgressIP object with one egress IP1 defined")
+		var egressIP1 net.IP
+		var err error
+		var retryTimeout2 = 2 * retryInterval
+		if utilnet.IsIPv6String(egress1Node.nodeIP) {
+			egressIP1, err = ipalloc.NewPrimaryIPv6()
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new EgressIP")
```
tssurya: the error assertion is common and can be done outside the if else, cuts down duplication

test/e2e/egressip.go
```diff
@@ -225,6 +225,50 @@ func isSupportedAgnhostForEIP(externalContainer infraapi.ExternalContainer) bool
...
+    namespaceSelector:
+        matchLabels:
+            %s
+`, name, ipsYAML, podLabelYaml, namespaceLabelYaml)
+
+	if err := os.WriteFile(yamlPath, []byte(egressIPConfig), 0644); err != nil {
+		framework.Failf("Unable to write CRD config to disk: %v", err)
+	}
+
+	ginkgo.DeferCleanup(func() {
```
tssurya: thanks for the util! but note that defer blocks like this can't be within the helper function...
I would say make your `createEIPManifest` util return the egressIPConfig as a param and then 

the actual os.WriteFile, creation of the EIP can be within the test itself, that way defer stays inside the test.

test/e2e/egressip.go
huiran0826: this is updated from coderabbitai [comment](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5721#discussion_r2554991913) , do you want me to revert it?

tssurya: code rabbit was correct based on your old code where you were defining both eip 1 and 2 together under same if else
since we moved eip2 to be closer to the actual step where its created
you dont have two err anymore

test/e2e/egressip.go
huiran0826: updated it.

test/e2e/egressip.go
huiran0826: done

test/e2e/egressip.go
huiran0826: done

PR #5725, opened 2025-11-17
Fixes regression with ecabc89

trozet:
With multiple subnets for a network, only the first one was being used for cluster subnet exclusion.

## Summary by CodeRabbit

* **Tests**
  * Expanded egress firewall test coverage to validate CIDR selector with multiple cluster subnets
  * Added test case verifying proper handling of multiple cluster subnets intersection
  * Updated test setup to include additional cluster subnet scenarios

* **Bug Fixes**
  * Fixed egress firewall destination logic to evaluate all matching cluster subnets during CIDR intersection analysis, previously only processing the first match

### DIFF
--- a/go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
+++ b/go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
@@ -508,6 +508,8 @@ type output struct {
 func TestValidateAndGetEgressFirewallDestination(t *testing.T) {
 	clusterSubnetStr := "10.1.0.0/16"
 	_, clusterSubnet, _ := net.ParseCIDR(clusterSubnetStr)
+	extraClusterSubnetStr := "10.2.0.0/16"
+	_, extraClusterSubnet, _ := net.ParseCIDR(extraClusterSubnetStr)
 	udnClusterSubnetStr := "9.0.0.0/16"
 	_, udnClusterSubnet, _ := net.ParseCIDR(udnClusterSubnetStr)
 	validUDNName := "udn-test"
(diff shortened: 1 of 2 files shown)


cathy-zhou:
lgtm

go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
```diff
@@ -683,7 +696,7 @@ func TestValidateAndGetEgressFirewallDestination(t *testing.T) {
 		t.Fatalf("failed to PrepareTestConfig: %v", err)
 	}
 
-	config.Default.ClusterSubnets = []config.CIDRNetworkEntry{{CIDR: clusterSubnet}}
+	config.Default.ClusterSubnets = []config.CIDRNetworkEntry{{CIDR: clusterSubnet}, {CIDR: extraClusterSubnet}}
```
npinaeva, line 699: I have one concern about this: I don't think we are planning to support multi-subnet default cluster network right only UDNs? So do you think it is better to move this test to the `ovn/egressfirewall_test.go` where UDN-based tests live?

npinaeva:
The change is ok, but let me clarify:
it is not a regression, because it only breaks with UDNs (which were not supported before at all), And only with UDNs that have multiple subnet of the same IP family, which we still don't support upstream. So you are kinda expanding the EF functionality to support future multi-subnet UDNs right?

go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
trozet: It looks like we do support this, but we are lacking CI testing for it:

https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5729

I think we should fix it as with the multiple cluster subnets OKEP for UDN we want to allow it there anyway, this would make everything consistent.

npinaeva (approved):

PR #5726, opened 2025-11-18
Add ValidatingAdmissionPolicy for EgressIP

arghosh93:
This commit is to add couple of ValidatingAdmissionPolicy to take care of following conditions:
- k8s.ovn.org/egressip-mark annotation should not be added while creating an EgressIP.
- A regular user should not be able to add k8s.ovn.org/egressip-mark annotation. Only a system user is allowed to do so.

## 📑 Description

Fixes #

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests verifying creation or annotation of EgressIP is rejected when a protected system annotation is present; tests assert exact error messages, skip in Helm-enabled runs, and are duplicated across variants.

* **Behavior Change**
  * Reconciliation no longer automatically restores a removed system-managed EgressIP mark; tests now validate cache/state instead of restoration.

* **Chores**
  * Added a Helm-enabled check to conditionally skip Helm-specific test paths.

### DIFF
--- a/dist/templates/rbac-ovnkube-master.yaml.j2
+++ b/dist/templates/rbac-ovnkube-master.yaml.j2
@@ -231,3 +231,44 @@ spec:
         operations:  ["UPDATE"]
         resources:   ["pods"]
 {%- endif %}
+
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: egressip-mark-annotation-validation
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups: ["k8s.ovn.org"]
+      apiVersions: ["v1"]
+      operations: ["CREATE", "UPDATE"]
+      resources: ["egressips"]
+  validations:
+  # Prevent creating EgressIP with the annotation
+  - expression: 'request.operation != "CREATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations)'
+    message: 'EgressIP resources cannot be created with the "k8s.ovn.org/egressip-mark" annotation. This annotation is managed by the system.'
+    reason: Invalid
+  # Prevent modifying or removing the annotation once set
+  - expression: 'request.operation != "UPDATE" || !has(oldObject.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations) ||
+      (has(object.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in object.metadata.annotations) && oldObject.metadata.annotations["k8s.ovn.org/egressip-mark"] == object.metadata.annotations["k8s.ovn.org/egressip-mark"])'
+    message: 'The "k8s.ovn.org/egressip-mark" annotation cannot be modified or removed once set. This annotation is managed by the system.'
+    reason: Invalid
+  # Only OVN controller service account can add the annotation
+  {% if ovn_enable_interconnect == "true" -%}
+  - expression: 'request.operation != "UPDATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations) || (has(oldObject.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations)) || request.userInfo.username == "system:serviceaccount:ovn-kubernetes:ovnkube-cluster-manager"'
+  {% else %}
+  - expression: 'request.operation != "UPDATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations) || (has(oldObject.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations)) || request.userInfo.username == "system:serviceaccount:ovn-kubernetes:ovnkube-master"'
+  {%- endif %}
+    message: 'A regular user must not add "k8s.ovn.org/egressip-mark" annotation to an EgressIP custom resource.'
+    reason: Invalid
+
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicyBinding
+metadata:
+  name: egressip-mark-annotation-validation-binding
+spec:
+  policyName: egressip-mark-annotation-validation
+  validationActions: [Deny]
--- a/test/e2e/egressip.go
+++ b/test/e2e/egressip.go
@@ -36,6 +36,7 @@ import (
 	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
 	"k8s.io/kubernetes/test/e2e/framework/pod"
 	e2epodoutput "k8s.io/kubernetes/test/e2e/framework/pod/output"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
 	utilnet "k8s.io/utils/net"
 )
 
@@ -3441,6 +3442,130 @@ spec:
 					podNamespace.Name, pod1Name, true, []string{pod1Node.nodeIP}).WithContext())
 			gomega.Expect(err).To(gomega.HaveOccurred(), "Node IP should NOT be used as source IP - connection should succeed but node IP should not be found")
 		}
+	})
+
+	ginkgo.It("Should fail if egressip-mark annotation is present during EgressIP creation", func() {
+		// This check can be removed when https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5879 is addressed
+		if isHelmEnabled() {
+			e2eskipper.Skipf("Skipping this test for HELM environment as we dont create required Validatingadmissionpolicy in a HELM environment")
+		}
+
+		ginkgo.By("1. Create an EgressIP object with one egress IP defined")
+		var egressIP1 net.IP
+		var err error
+		if utilnet.IsIPv6String(egress1Node.nodeIP) {
+			egressIP1, err = ipalloc.NewPrimaryIPv6()
+		} else {
+			egressIP1, err = ipalloc.NewPrimaryIPv4()
+		}
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new Node IP")
+
+		var egressIPConfig = `apiVersion: k8s.ovn.org/v1
+kind: EgressIP
+metadata:
+    name: ` + egressIPName + `
+    annotations:
+      ` + util.EgressIPMarkAnnotation + `: "50000"
+spec:
+    egressIPs:
+    - ` + egressIP1.String() + `
+    namespaceSelector:
+        matchLabels:
+            name: ` + f.Namespace.Name + `
+`
+		if err := os.WriteFile(egressIPYaml, []byte(egressIPConfig), 0644); err != nil {
+			framework.Failf("Unable to write CRD config to disk: %v", err)
+		}
+		defer func() {
+			if err := os.Remove(egressIPYaml); err != nil {
+				framework.Logf("Unable to remove the CRD config from disk: %v", err)
+			}
+		}()
+
+		ginkgo.By("2. Create an EgressIP with k8s.ovn.org/egressip-mark annotation defined")
+		_, err = e2ekubectl.RunKubectl("default", "create", "-f", egressIPYaml)
+		gomega.Expect(err).To(gomega.HaveOccurred(), "Should fail if k8s.ovn.org/egressip-mark annotation is present during creation")
+		gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("EgressIP resources cannot be created with the \"k8s.ovn.org/egressip-mark\" annotation. This annotation is managed by the system.")))
+	})
+
+	ginkgo.It("Should fail if egressip-mark annotation is being added by a regular user", func() {
+		// This check can be removed when https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5879 is addressed
+		if isHelmEnabled() {
+			e2eskipper.Skipf("Skipping this test for HELM environment as we dont create required Validatingadmissionpolicy in a HELM environment")
+		}
+
+		ginkgo.By("1. Add the \"k8s.ovn.org/egress-assignable\" label to egress1Node node")
+		egressNodeAvailabilityHandler := egressNodeAvailabilityHandlerViaLabel{f}
+		egressNodeAvailabilityHandler.Enable(egress1Node.name)
+		defer egressNodeAvailabilityHandler.Restore(egress1Node.name)
+
+		podNamespace := f.Namespace
+		labels := map[string]string{
+			"name": f.Namespace.Name,
+		}
+		updateNamespaceLabels(f, podNamespace, labels)
+
+		ginkgo.By("2. Create an EgressIP object with one egress IP defined")
+		var egressIP1 net.IP
+		var err error
+		if utilnet.IsIPv6String(egress1Node.nodeIP) {
+			egressIP1, err = ipalloc.NewPrimaryIPv6()
+		} else {
+			egressIP1, err = ipalloc.NewPrimaryIPv4()
+		}
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "must allocate new Node IP")
+
+		var egressIPConfig = `apiVersion: k8s.ovn.org/v1
+kind: EgressIP
+metadata:
+    name: ` + egressIPName + `
+spec:
+    egressIPs:
+    - ` + egressIP1.String() + `
+    namespaceSelector:
+        matchLabels:
+            name: ` + f.Namespace.Name + `
+`
+		if err := os.WriteFile(egressIPYaml, []byte(egressIPConfig), 0644); err != nil {
+			framework.Failf("Unable to write CRD config to disk: %v", err)
+		}
+		defer func() {
+			if err := os.Remove(egressIPYaml); err != nil {
+				framework.Logf("Unable to remove the CRD config from disk: %v", err)
+			}
+		}()
+
+		framework.Logf("Create the EgressIP configuration")
+		e2ekubectl.RunKubectlOrDie("default", "create", "-f", egressIPYaml)
+
+		ginkgo.By("3. Check that the status is of length one and that it is assigned to egress1Node")
+		statuses := verifyEgressIPStatusLengthEquals(1, nil)
+		if statuses[0].Node != egress1Node.name {
+			framework.Failf("Step 3. Check that the status is of length one and that it is assigned to egress1Node, failed")
+		}
+
+		ginkgo.By("4. Try updating k8s.ovn.org/egressip-mark annotation")
+		// Get the current annotation value to ensure we try to overwrite with a different value
+		annotationsJSON, err := e2ekubectl.RunKubectl("", "get", "egressip", egressIPName, "-o", "jsonpath={.metadata.annotations}")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(), "Failed to get annotations")
+		var annotations map[string]string
+		err = json.Unmarshal([]byte(annotationsJSON), &annotations)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(), "Failed to unmarshal annotations JSON")
+		currentValue := annotations[util.EgressIPMarkAnnotation]
+
+		newValue := 50000
+		if currentValue == "50000" {
+			newValue = 50001
+		}
+
+		_, err = e2ekubectl.RunKubectl("", "annotate", "--overwrite", "egressip", egressIPName, fmt.Sprintf("%s=%d", util.EgressIPMarkAnnotation, newValue))
+		gomega.Expect(err).To(gomega.HaveOccurred(), "Should fail if k8s.ovn.org/egressip-mark is being updated")
+		gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("The \"k8s.ovn.org/egressip-mark\" annotation cannot be modified or removed once set. This annotation is managed by the system.")))
+
+		ginkgo.By("5. Try removing k8s.ovn.org/egressip-mark annotation")
+		_, err = e2ekubectl.RunKubectl("", "annotate", "--overwrite", "egressip", egressIPName, fmt.Sprintf("%s-", util.EgressIPMarkAnnotation))
+		gomega.Expect(err).To(gomega.HaveOccurred(), "Should fail if k8s.ovn.org/egressip-mark is being removed")
+		gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("The \"k8s.ovn.org/egressip-mark\" annotation cannot be modified or removed once set. This annotation is managed by the system.")))
 	})
 
 	ginkgo.DescribeTable("[OVN network] multiple namespaces with different primary networks", func(otherNetworkAttachParms networkAttachmentConfigParams) {
--- a/dist/images/daemonset.sh
+++ b/dist/images/daemonset.sh
@@ -1161,6 +1161,7 @@ ovn_enable_dnsnameresolver=${ovn_enable_dnsnameresolver} \
 ovn_route_advertisements_enable=${ovn_route_advertisements_enable} \
 ovn_pre_conf_udn_addr_enable=${ovn_pre_conf_udn_addr_enable} \
 ovn_advertised_udn_isolation_mode=${ovn_advertised_udn_isolation_mode} \
+ovn_enable_interconnect=${ovn_enable_interconnect} \
   jinjanate ../templates/rbac-ovnkube-master.yaml.j2 -o ${output_dir}/rbac-ovnkube-master.yaml
 
 cp ../templates/rbac-ovnkube-identity.yaml.j2 ${output_dir}/rbac-ovnkube-identity.yaml
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -1445,6 +1445,11 @@ func isLocalGWModeEnabled() bool {
 	return present && val == "local"
 }
 
+func isHelmEnabled() bool {
+	val, present := os.LookupEnv("USE_HELM")
+	return present && val == "true"
+}
+
 func isPreConfiguredUdnAddressesEnabled() bool {
 	ovnKubeNamespace := deploymentconfig.Get().OVNKubernetesNamespace()
 	val := getTemplateContainerEnv(ovnKubeNamespace, "daemonset/ovnkube-node", getNodeContainerName(), "OVN_PRE_CONF_UDN_ADDR_ENABLE")


test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
 
+	ginkgo.It("Should fail if egressip-mark annotation is present during EgressIP creation", func() {
+		egressIPValidatingadmissionpolicy, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicy", "egressip-create-validation")
+		egressIPValidatingadmissionpolicyBinding, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicybinding", "egressip-create-validation-binding")
+		if egressIPValidatingadmissionpolicy == "" || egressIPValidatingadmissionpolicyBinding == "" {
+			ginkgo.Skip("Required validatingadmissionpolicy and validatingadmissionpolicy are missing, hence skipping this test")
```
pliurh: The message is not aligned with the condition check above.

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
 
+	ginkgo.It("Should fail if egressip-mark annotation is present during EgressIP creation", func() {
+		egressIPValidatingadmissionpolicy, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicy", "egressip-create-validation")
+		egressIPValidatingadmissionpolicyBinding, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicybinding", "egressip-create-validation-binding")
+		if egressIPValidatingadmissionpolicy == "" || egressIPValidatingadmissionpolicyBinding == "" {
```
pliurh: In which CI lane shall we skip this test?

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
...
+				framework.Logf("Unable to remove the CRD config from disk: %v", err)
+			}
+		}()
+
+		ginkgo.By("2. Create an EgressIP with k8s.ovn.org/egressip-mark annotation defined")
+		_, err = e2ekubectl.RunKubectl("default", "create", "-f", egressIPYaml)
+		gomega.Expect(err).To(gomega.HaveOccurred(), "Should fail if k8s.ovn.org/egressip-mark annotation is present during creation")
+		gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("EgressIP resources cannot be created with the \"k8s.ovn.org/egressip-mark\" annotation. This annotation is managed by the system.")))
+	})
+
```
pliurh, line 3490: Can you also add test for updating?

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -230,3 +230,65 @@ spec:
...
+
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: egressip-update-validation
+spec:
+  failurePolicy: Fail
+  matchConditions:
+  - expression: '("k8s.ovn.org/egressip-mark" in object.metadata.annotations) && (!has(oldObject.metadata.annotations) ||
```
pliurh: This condition only catch the case that the `k8s.ovn.org/egressip-mark` annotation is added to the EIP. We shall also prevent the `k8s.ovn.org/egressip-mark` from being modified.

test/e2e/egressip.go
arghosh93: oops, Thanks!

test/e2e/egressip.go
arghosh93: While installing with Helm, we don't create validatingAdmissionPolicies. So, in that case, all egressIP E2E test fails, but egressIP tests are important for that lane also, it seems. So I have added a check to make sure the required validatingAdmissionPolicy and validatingAdmissionPolicyBinding are present.

test/e2e/egressip.go
arghosh93: The Update validatingAdmissionPolicy makes sure that only ovnkube-cluster-manager(ic)/ovnkube-master(non-ic) Service account can update egressip-mark annotation. 
EgressIp assignment would not complete if cluster-manager cant add egressip-mark annotation and all E2Es will eventually fail. Passing of all other EgressIP E2E test implies that validatingAdmissionPolicy for Update operation is working properly.

dist/templates/rbac-ovnkube-master.yaml.j2
arghosh93: Egressip-mark annotation gets restored on update.
[github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/clustermanager/egressip_controller_test.go](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/clustermanager/egressip_controller_test.go#L3291)

test/e2e/egressip.go
pliurh: Can we also add validatingAdmissionPolicies to the Helm chart?

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
 
+	ginkgo.It("Should fail if egressip-mark annotation is present during EgressIP creation", func() {
+		egressIPValidatingadmissionpolicy, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicy", "egressip-create-validation")
+		egressIPValidatingadmissionpolicyBinding, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicybinding", "egressip-create-validation-binding")
+		if egressIPValidatingadmissionpolicy == "" || egressIPValidatingadmissionpolicyBinding == "" {
+			ginkgo.Skip("Required validatingadmissionpolicy and validatingadmissionpolicyBinding are missing, hence skipping this test")
```
pliurh: should be `validatingadmissionpolicy or validatingadmissionpolicyBinding`

test/e2e/egressip.go
arghosh93: This will need a separate PR.  It seems we dont configure much while creating helm environment.

dist/templates/rbac-ovnkube-master.yaml.j2
tssurya: this is actually a good point since I also wanted to ask the same thing..
Since we are doing a VAP why not keep the VAP and remove the code for ensureMark that's restoring the mark on update ? - the current code seems to be reactive to the update and restores mark right? It's not actually blocking it v/s what we want is a block ?

```
// ensureMark ensures that if a mark was remove or changed value, then restore the mark.
func (eIPC *egressIPClusterController) ensureMark(old, new *egressipv1.EgressIP) {
	// Adding the mark to annotations is bundled with status update in-order to minimise updates, cover the case where there is no update to status
	// and mark annotation has been modified / removed. This should only occur for an update and the mark was previous set.
	if old != nil && new != nil {
		if util.IsEgressIPMarkSet(old.Annotations) && util.EgressIPMarkAnnotationChanged(old.Annotations, new.Annotations) {
			mark, _, err := eIPC.getOrAllocMark(new.Name)
			if err != nil {
				klog.Errorf("Failed to restore EgressIP %s mark because unable to retrieve mark: %v", new.Name, err)
			} else if err = eIPC.patchEgressIP(new.Name, generateMarkPatchOp(mark)); err != nil {
				klog.Errorf("Failed to restore EgressIP %s mark because patching failed: %v", new.Name, err)
			}
		}
	}
}
```
here IIUC once a mark is added or set it can't be modified.. can it? I think its safer to totally block the mark update, than try to restore it to the old value once the user did change it?

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -230,3 +230,65 @@ spec:
...
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: egressip-create-validation
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups: ["k8s.ovn.org"]
+      apiVersions: ["v1"]
+      operations: ["CREATE"]
```
tssurya: ok so for CREATION we block everyone.. makes sense

I think maybe we can combine both of these into same VAP? but not a big deal cause keeping them separate at least makes the messaging more clear.

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
 		framework.ExpectNoError(err, "Step 11. Check connectivity from other pod and verify that the srcIP is the expected egressIP and verify that the srcIP is the expected nodeIP, failed: %v", err)
 	})
 
+	ginkgo.It("Should fail if egressip-mark annotation is present during EgressIP creation", func() {
+		egressIPValidatingadmissionpolicy, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicy", "egressip-create-validation")
```
tssurya: shouldn't you check the err is not nil here i.e ExpectNoError ?

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
...
+		egressIPValidatingadmissionpolicyBinding, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicybinding", "egressip-create-validation-binding")
+		if egressIPValidatingadmissionpolicy == "" || egressIPValidatingadmissionpolicyBinding == "" {
+			ginkgo.Skip("Required validatingadmissionpolicy or validatingadmissionpolicyBinding are missing, hence skipping this test")
+		}
+		ginkgo.By("0. Add the \"k8s.ovn.org/egress-assignable\" label to egress1Node node")
```
tssurya: do we need the node label step for this test?

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,55 @@ spec:
...
+				framework.Logf("Unable to remove the CRD config from disk: %v", err)
+			}
+		}()
+
+		ginkgo.By("2. Create an EgressIP with k8s.ovn.org/egressip-mark annotation defined")
+		_, err = e2ekubectl.RunKubectl("default", "create", "-f", egressIPYaml)
+		gomega.Expect(err).To(gomega.HaveOccurred(), "Should fail if k8s.ovn.org/egressip-mark annotation is present during creation")
+		gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("EgressIP resources cannot be created with the \"k8s.ovn.org/egressip-mark\" annotation. This annotation is managed by the system.")))
+	})
+
```
tssurya, line 3490: what about a test for the `UPDATE` EIP action ? - for both adding annotation and updating it? can you expand this same test but with the case where a regular user tries the update and it doesn't work? - maybe see some kubernetes tests for how they simulate some other user serviceaccount create to make it seem like end user is touching it. Cause that's the main bug you are trying to fix which lacks coverage today.

tssurya (changes requested):

dist/templates/rbac-ovnkube-master.yaml.j2
arghosh93: Yes Surya, makes sense. I am going to add a new VAP to block annotation change after it is added.

test/e2e/egressip.go
arghosh93: During installation using Helm, we don't create these VAPs. If we fail due to an error, then lanes with Helm installation will also fail. I tried fixing the Helm installation, but it seems it requires a separate PR.

test/e2e/egressip.go
arghosh93: not required. will remove it in next push.

test/e2e/egressip.go
arghosh93: I am going to add a new test where, at first, we create an egress Ip then try to update the annotation.

test/e2e/egressip.go
```diff
@@ -3182,6 +3182,107 @@ spec:
...
+			}
+		}()
+
+		framework.Logf("Create the EgressIP configuration")
+		e2ekubectl.RunKubectlOrDie("default", "create", "-f", egressIPYaml)
+
+		ginkgo.By("3. Check that the status is of length one and that it is assigned to egress1Node")
+		statuses := verifyEgressIPStatusLengthEquals(1, nil)
+		if statuses[0].Node != egress1Node.name {
+			framework.Failf("Step 3. Check that the status is of length one and that it is assigned to egress1Node, failed")
```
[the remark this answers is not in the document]
arghosh93: done

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -230,3 +230,94 @@ spec:
...
+  {% if ovn_enable_interconnect == "true" -%}
+  - expression: '(request.userInfo.username == "system:serviceaccount:ovn-kubernetes:ovnkube-cluster-manager")'
+  {% else %}
+  - expression: '(request.userInfo.username == "system:serviceaccount:ovn-kubernetes:ovnkube-master")'
+  {%- endif %}
+    message: 'A regular user must not add "k8s.ovn.org/egressip-mark" annotation to an EgressIP custom resource.'
+    reason: Invalid
+
+---
+apiVersion: admissionregistration.k8s.io/v1
```
tssurya, line 236: I think this one and the one below can just be one VAP? (so total 2 VAPs and not 3)

```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: egressip-annotation-validation
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups: ["k8s.ovn.org"]
      apiVersions: ["v1"]
      operations: ["CREATE", "UPDATE"]
      resources: ["egressips"]
  validations:
  # Block creation with the annotation
  - expression: 'request.operation != "CREATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations)'
    message: 'EgressIP resources cannot be created with the "k8s.ovn.org/egressip-mark" annotation. This annotation is managed by the system.'
    reason: Invalid
  # Block modification of the annotation value on update
  - expression: 'request.operation != "UPDATE" || 
      !has(oldObject.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations) ||
      !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations) ||
      oldObject.metadata.annotations["k8s.ovn.org/egressip-mark"] == object.metadata.annotations["k8s.ovn.org/egressip-mark"]'
    message: 'The "k8s.ovn.org/egressip-mark" annotation value cannot be modified once set. This annotation is managed by the system.'
    reason: Invalid
```
something like that?

test/e2e/egressip.go
tssurya: @coderabbitai: Open a good first issue after double checking we don't have equivalent of VAPs in the helm installation lane?

tssurya: @arghosh93 what you should do instead is assert the error if `USE_HELM: true` is not set which means its non=helm lane?

so we can just skip these tests in helm lane by doing:
```
if isHelmEnabled() {
			e2eskipper.Skipf("...")
		}
```
and
```
func isHelmEnabled() bool {
	val, present := os.LookupEnv("USE_HELM")
	return present && val == "true"
}
```
right at the beginning of this test before even trying to fetch the VAPs

I think that's more robust than relying on the empty policy cause in the non-helm lanes its possible something went wrong - so we should be asserting the error properly

test/e2e/egressip.go
```diff
@@ -3275,172 +3220,73 @@ spec:
...
+		egressIPValidatingadmissionpolicyBinding, _ := e2ekubectl.RunKubectl("default", "get", "--no-headers", "validatingadmissionpolicybinding", "egressip-annotation-modify-validation-binding")
+		if egressIPValidatingadmissionpolicy == "" || egressIPValidatingadmissionpolicyBinding == "" {
+			ginkgo.Skip("Required validatingadmissionpolicy or validatingadmissionpolicyBinding are missing, hence skipping this test")
+		}
+		ginkgo.By("1. Add the \"k8s.ovn.org/egress-assignable\" label to egress1Node node")
+		egressNodeAvailabilityHandler := egressNodeAvailabilityHandlerViaLabel{f}
+		egressNodeAvailabilityHandler.Enable(egress1Node.name)
+		defer egressNodeAvailabilityHandler.Restore(egress1Node.name)
+
+		podNamespace := f.Namespace
```
tssurya, line 3502: what's this labeling done for?

test/e2e/egressip.go
```diff
@@ -3275,172 +3220,73 @@ spec:
...
-					podNamespace.Name, pod1Name, true, []string{pod1Node.nodeIP}).WithContext())
-			gomega.Expect(err).To(gomega.HaveOccurred(), "Node IP should NOT be used as source IP - connection should succeed but node IP should not be found")
+		ginkgo.By("3. Check that the status is of length one and that it is assigned to egress1Node")
+		statuses := verifyEgressIPStatusLengthEquals(1, nil)
+		if statuses[0].Node != egress1Node.name {
+			framework.Failf("Step 3. Check that the status is of length one and that it is assigned to egress1Node, failed")
 		}
+
+		ginkgo.By("4. Try updating k8s.ovn.org/egressip-mark annotation")
+		_, err = e2ekubectl.RunKubectl("", "annotate", "--overwrite", "egressip", egressIPName, fmt.Sprintf("%s=%d", util.EgressIPMarkAnnotation, 50000))
```
tssurya: is this done by a regular user or admin or ovnk? - title of the `It` block says its a regular user... I guess its admin and non-ovnk role basically?

arghosh93: yes, it should be admin but definitely not any ovnk SA.

test/e2e/egressip.go
arghosh93: This is just to make sure that we have a namespace with the label mentioned under namespaceSelector of the egressip CR.

dist/templates/rbac-ovnkube-master.yaml.j2
arghosh93: Ack

test/e2e/egressip.go
tssurya: I get that, my question was more for is there any correlation to that v/s VAP testing? i.e do we need to be labelling a node+adding namespace etc for testing VAP?
I don't mind if it stays but was curious if it was to just check the lifecycle?

test/e2e/egressip.go
```diff
@@ -3275,172 +3220,73 @@ spec:
...
 
-		ginkgo.By("2. Create second EgressIP object with one egress IP2 defined")
-		var egressIP2 net.IP
+		ginkgo.By("2. Create an EgressIP with k8s.ovn.org/egressip-mark annotation defined")
+		_, err = e2ekubectl.RunKubectl("default", "create", "-f", egressIPYaml)
+		gomega.Expect(err).To(gomega.HaveOccurred(), "Should fail if k8s.ovn.org/egressip-mark annotation is present during creation")
+		gomega.Expect(err).To(gomega.MatchError(gomega.ContainSubstring("EgressIP resources cannot be created with the \"k8s.ovn.org/egressip-mark\" annotation. This annotation is managed by the system.")))
+	})
+
+	ginkgo.It("Should fail if egressip-mark annotation is being added by a regular user", func() {
```
tssurya, line 3491: NIT
```suggestion
	ginkgo.It("Should fail if egressip-mark annotation is being added by a non-ovnk-system user", func() {
```

test/e2e/egressip.go
tssurya: ah it came from https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5726#discussion_r2715784043 ok nvm all good

go-controller/pkg/clustermanager/egressip_controller_test.go
```diff
@@ -3288,24 +3287,7 @@ var _ = ginkgo.Describe("OVN cluster-manager EgressIP Operations", func() {
...
-				assignedMarkStr, err = getEgressIPAnnotationValue(eIP.Name)()
-				gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "failed to get egress IP mark from annotations")
-				assignedMarkAfterUpdate, err := strconv.Atoi(assignedMarkStr)
-				gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "failed to convert mark to string")
-				gomega.Expect(assignedMark).Should(gomega.Equal(assignedMarkAfterUpdate), "Mark should be identical if annotation is cleared")
-				ginkgo.By("confirm cache is unchanged")
+				ginkgo.By("confirm cache is set correctly")
 				cachedMark, _, err := fakeClusterManagerOVN.eIPC.getOrAllocMark(eIP.Name)
 				gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 				gomega.Expect(cachedMark).Should(gomega.Equal(assignedMark), "EIP annotation and cache mark integer must be the same")
```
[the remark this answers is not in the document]
arghosh93: Line number 3284 to 3288 make sure that egressip-mark annotation exists on the egressip object. On the other hand, syncEgressIPMarkAllocator() called by egressIPClusterControllerEventHandler during execution of SyncFunc() makes sure to exhibit an error if cache or annotation update fails.

test/e2e/egressip.go
tssurya: haha you've pushed so many times you could have addressed the NIT as well but ok :d

test/e2e/egressip.go
```diff
@@ -3443,6 +3444,117 @@ spec:
...
+
+		framework.Logf("Create the EgressIP configuration")
+		e2ekubectl.RunKubectlOrDie("default", "create", "-f", egressIPYaml)
+
+		ginkgo.By("3. Check that the status is of length one and that it is assigned to egress1Node")
+		statuses := verifyEgressIPStatusLengthEquals(1, nil)
+		if statuses[0].Node != egress1Node.name {
+			framework.Failf("Step 3. Check that the status is of length one and that it is assigned to egress1Node, failed")
+		}
+
```
tssurya, line 3546: I think having a step here where you print the value from EgressIP annotation mark is important? what if its 50000? :D

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -230,3 +230,71 @@ spec:
...
```
tssurya, line 236: nit: I still think we can merge both these VAPs into one, is there something around matchconditions+expressions that makes this hard?
I think its possible but see what you think

tssurya (commented):
last three comments to consider

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -230,3 +230,71 @@ spec:
...
+      apiVersions: ["v1"]
+      operations: ["CREATE", "UPDATE"]
+      resources: ["egressips"]
+  validations:
+  # Prevent creating EgressIP with the annotation
+  - expression: 'request.operation != "CREATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations)'
+    message: 'EgressIP resources cannot be created with the "k8s.ovn.org/egressip-mark" annotation. This annotation is managed by the system.'
+    reason: Invalid
+  # Prevent modifying or removing the annotation once set
+  - expression: 'request.operation != "UPDATE" || !has(oldObject.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations) ||
```
tssurya, line 254: one half negation and other half positive is making it harder to read but maybe its just me

dist/templates/rbac-ovnkube-master.yaml.j2
tssurya:
```
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicy
metadata:
  name: egressip-mark-annotation-validation
spec:
  failurePolicy: Fail
  matchConstraints:
    resourceRules:
    - apiGroups: ["k8s.ovn.org"]
      apiVersions: ["v1"]
      operations: ["CREATE", "UPDATE"]
      resources: ["egressips"]
  validations:
  # Rule 1: Prevent creating EgressIP with the annotation
  - expression: 'request.operation != "CREATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations)'
    message: 'EgressIP resources cannot be created with the "k8s.ovn.org/egressip-mark" annotation. This annotation is managed by the system.'
    reason: Invalid

  # Rule 2: Prevent modifying or removing the annotation once set
  - expression: 'request.operation != "UPDATE" || !has(oldObject.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations) || (has(object.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in object.metadata.annotations) && oldObject.metadata.annotations["k8s.ovn.org/egressip-mark"] == object.metadata.annotations["k8s.ovn.org/egressip-mark"])'
    message: 'The "k8s.ovn.org/egressip-mark" annotation cannot be modified or removed once set. This annotation is managed by the system.'
    reason: Invalid

  # Rule 3: Only OVN controller service account can add the annotation
  {% if ovn_enable_interconnect == "true" -%}
  - expression: 'request.operation != "UPDATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations) || (has(oldObject.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations)) || request.userInfo.username == "system:serviceaccount:ovn-kubernetes:ovnkube-cluster-manager"'
  {% else %}
  - expression: 'request.operation != "UPDATE" || !has(object.metadata.annotations) || !("k8s.ovn.org/egressip-mark" in object.metadata.annotations) || (has(oldObject.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in oldObject.metadata.annotations)) || request.userInfo.username == "system:serviceaccount:ovn-kubernetes:ovnkube-master"'
  {%- endif %}
    message: 'Only the OVN controller can add the "k8s.ovn.org/egressip-mark" annotation to an EgressIP resource.'
    reason: Invalid
---
apiVersion: admissionregistration.k8s.io/v1
kind: ValidatingAdmissionPolicyBinding
metadata:
  name: egressip-mark-annotation-validation-binding
spec:
  policyName: egressip-mark-annotation-validation
  validationActions:
  - Deny
```
something like that does it work? don't use this exact one just took it from AI as example.. but I think whatever rules you have can be combined into 1 VAP?

arghosh93: Yes Surya, matchConditions expression in egressip-update-validation VAP makes it harder. 
But looking at your suggestion, it seems that it is possible and keeps the functionality intact after combining both. Let me try it out and see if E2Es are passing.

test/e2e/egressip.go
arghosh93: I think the egressip-mark value does not matter much here. If the annotation overwrite is allowed, then the test would fail. It does not check egressip status to validate its state.

arghosh93: i get your point. if mark value is same then it does not exhibit an error.

arghosh93: Thanks Surya :)

test/e2e/egressip.go
```diff
@@ -3443,6 +3444,130 @@ spec:
...
+		ginkgo.By("4. Try updating k8s.ovn.org/egressip-mark annotation")
+		// Get the current annotation value to ensure we try to overwrite with a different value
+		annotationsJSON, err := e2ekubectl.RunKubectl("", "get", "egressip", egressIPName, "-o", "jsonpath={.metadata.annotations}")
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(), "Failed to get annotations")
+		var annotations map[string]string
+		err = json.Unmarshal([]byte(annotationsJSON), &annotations)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(), "Failed to unmarshal annotations JSON")
+		currentValue := annotations[util.EgressIPMarkAnnotation]
+
```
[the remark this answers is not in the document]
arghosh93: test/e2e/cluster_network_connect.go:L331-L352 uses same method. How is this approach different?

dist/templates/rbac-ovnkube-master.yaml.j2
```diff
@@ -230,3 +230,44 @@ spec:
...
+      (has(object.metadata.annotations) && ("k8s.ovn.org/egressip-mark" in object.metadata.annotations) && oldObject.metadata.annotations["k8s.ovn.org/egressip-mark"] == object.metadata.annotations["k8s.ovn.org/egressip-mark"])'
+    message: 'The "k8s.ovn.org/egressip-mark" annotation cannot be modified or removed once set. This annotation is managed by the system.'
+    reason: Invalid
+  # Only OVN controller service account can add the annotation
```
tssurya, line 258:
```suggestion
  # Only ovnkube controller service account can add the annotation as an update to the EgressIP object
```

tssurya:
unrelated timed out error:
```
Summarizing 1 Failure:
  [TIMEDOUT] OVS CPU affinity pinning [It] can be enabled on specific nodes by creating enable_dynamic_cpu_affinity file [Feature:OVSCPUPin]
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/ovspinning.go:21

Ran 298 of 911 Specs in 10832.349 seconds
FAIL! - Suite Timeout Elapsed -- 297 Passed | 1 Failed | 2 Pending | 611 Skipped
--- FAIL: TestE2E (10832.54s)
```
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21665685817/job/62461406383?pr=5726
this is the first time I am seeing it timeout.. I don't know if this lane is totally cuckoo on time
unrelated to this PR so merging this.

https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5913

PR #5728, opened 2025-11-19
egressip: skip bridge configuration for secondary network IPs

pliurh:
Skip configuring EgressIPs on the bridge when they don't belong to the OVN (primary) network. This prevents conflicts when EgressIPs are assigned to secondary host networks.

Uses util.IsOVNNetwork() to check network membership and adds a helper function parseAndValidateEIP() to consolidate validation logic and fix missing packet mark validation.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## How to verify it
1. Deploy a cluster with nodes that have multiple interfaces.
2. Create an EgreseIP for a node's secondary interface.
3. Check if the egress IP is only assigned to the secondary interface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Egress IP handling now skips secondary (non-primary) host-network IPs during add/update/delete/sync to prevent incorrect bridge/annotation changes and noisy logs.
  * Mistaken secondary-network Egress IPs are cleaned from annotations, bridge, and cache during updates/sync.
  * Address removal treats "address not available" (EADDRNOTAVAIL) as non-fatal to avoid spurious errors.

* **Refactor**
  * Centralized Egress IP parsing/validation and OVN/primary-network detection for consistent behavior and skip semantics.

* **Tests**
  * Added tests for secondary-network Egress IP scenarios, cleanup flows, host CIDR/annotation helpers.

### DIFF
--- a/go-controller/pkg/node/egressip/gateway_egressip.go
+++ b/go-controller/pkg/node/egressip/gateway_egressip.go
@@ -209,115 +209,94 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
 }
 
 func (g *BridgeEIPAddrManager) AddEgressIP(eip *egressipv1.EgressIP) (bool, error) {
-	var isUpdated bool
-	if !util.IsEgressIPMarkSet(eip.Annotations) {
-		return isUpdated, nil
+	ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip)
+	if err != nil {
+		return false, fmt.Errorf("failed to add EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
 	}
-	for _, status := range eip.Status.Items {
-		if status.Node != g.nodeName {
-			continue
-		}
-		ip, pktMark, err := parseEIPMarkIP(eip.Annotations, status.EgressIP)
-		if err != nil {
-			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to extract config from EgressIP obj: %v", err)
-		}
-		// must always add to cache before adding IP because we want to inform node ip handler that this is not a valid node IP
-		g.cache.insertMarkIP(pktMark, ip)
-		if err = g.addIPToAnnotation(ip); err != nil {
-			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to add EgressIP IP to Node annotation: %v", err)
-		}
-		if err = g.addIPBridge(ip); err != nil {
-			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because failed to add address to link: %v", err)
-		}
-		isUpdated = true
-		break // no need to continue as only one EIP IP is assigned to a node
+	if shouldSkip {
+		return false, nil
 	}
-	return isUpdated, nil
+	// must always add to cache before adding IP because we want to inform node ip handler that this is not a valid node IP
+	g.cache.insertMarkIP(pktMark, ip)
+	if err = g.addIPToAnnotation(ip); err != nil {
+		return false, fmt.Errorf("failed to add EgressIP gateway config because unable to add EgressIP IP to Node annotation: %v", err)
+	}
+	if err = g.addIPBridge(ip); err != nil {
+		return false, fmt.Errorf("failed to add EgressIP gateway config because failed to add address to link: %v", err)
+	}
+	return true, nil
 }
 
 func (g *BridgeEIPAddrManager) UpdateEgressIP(oldEIP, newEIP *egressipv1.EgressIP) (bool, error) {
+	// Parse and validate old EgressIP
+	oldIP, oldMark, oldSkip, err := g.parseAndValidateEIP(oldEIP)
+	if err != nil {
+		return false, fmt.Errorf("unable to parse and validate old EgressIP: %v", err)
+	}
+
+	// Parse and validate new EgressIP
+	newIP, newMark, newSkip, err := g.parseAndValidateEIP(newEIP)
+	if err != nil {
+		return false, fmt.Errorf("unable to parse and validate new EgressIP: %v", err)
+	}
+
+	// Both should be skipped - no change
+	if oldSkip && newSkip {
+		return false, nil
+	}
+
+	// Both exist and are the same - no change
+	if !oldSkip && !newSkip && oldIP.Equal(newIP) {
+		return false, nil
+	}
+
 	var isUpdated bool
-	// at most, one status item for this node will be found.
-	for _, oldStatus := range oldEIP.Status.Items {
-		if oldStatus.Node != g.nodeName {
-			continue
-		}
-		if !util.IsEgressIPMarkSet(oldEIP.Annotations) {
-			// this scenario may occur during upgrade from when ovn-k didn't apply marks to EIP objs
-			break
-		}
-		if util.IsItemInSlice(newEIP.Status.Items, oldStatus) {
-			// if one status entry exists in both status items, then nothing needs to be done because no status update.
-			// also, because at most only one status item can be assigned to a node, we can return early.
-			return isUpdated, nil
-		}
-		ip, pktMark, err := parseEIPMarkIP(oldEIP.Annotations, oldStatus.EgressIP)
-		if err != nil {
-			return isUpdated, fmt.Errorf("failed to update EgressIP SNAT for ext bridge cache because unable to extract config from old EgressIP obj: %v", err)
-		}
-		if err = g.deleteIPBridge(ip); err != nil {
+
+	// Delete old if it exists and is different from new
+	if !oldSkip {
+		if err = g.deleteIPBridge(oldIP); err != nil {
 			return isUpdated, fmt.Errorf("failed to update EgressIP gateway config because failed to delete address from link: %v", err)
 		}
-		g.cache.deleteMarkIP(pktMark, ip)
-		if err = g.deleteIPsFromAnnotation(ip); err != nil {
+		g.cache.deleteMarkIP(oldMark, oldIP)
+		if err = g.deleteIPsFromAnnotation(oldIP); err != nil {
 			return isUpdated, fmt.Errorf("failed to update EgressIP gateway config because unable to delete EgressIP IP from Node annotation: %v", err)
 		}
 		isUpdated = true
-		break
 	}
-	for _, newStatus := range newEIP.Status.Items {
-		if newStatus.Node != g.nodeName {
-			continue
-		}
-		if !util.IsEgressIPMarkSet(newEIP.Annotations) {
-			// this scenario may occur during upgrade from when ovn-k didn't apply marks to EIP objs
-			return isUpdated, nil
-		}
-		ip, pktMark, err := parseEIPMarkIP(newEIP.Annotations, newStatus.EgressIP)
-		if err != nil {
-			return isUpdated, fmt.Errorf("failed to update EgressIP gateway config because unable to extract config from EgressIP obj: %v", err)
-		}
-		// must always add to OF cache before adding IP because we want to inform node ip handler that this is not a valid node IP
-		g.cache.insertMarkIP(pktMark, ip)
-		if err = g.addIPToAnnotation(ip); err != nil {
+
+	// Add new if it exists
+	if !newSkip {
+		// must always add to cache before adding IP because we want to inform node ip handler that this is not a valid node IP
+		g.cache.insertMarkIP(newMark, newIP)
+		if err = g.addIPToAnnotation(newIP); err != nil {
 			return isUpdated, fmt.Errorf("failed to update EgressIP gateway config because unable to add EgressIP IP to Node annotation: %v", err)
 		}
-		if err = g.addIPBridge(ip); err != nil {
+		if err = g.addIPBridge(newIP); err != nil {
 			return isUpdated, fmt.Errorf("failed to update EgressIP gateway config because failed to add address to link: %v", err)
 		}
 		isUpdated = true
-		break
 	}
+
 	return isUpdated, nil
 }
 
 func (g *BridgeEIPAddrManager) DeleteEgressIP(eip *egressipv1.EgressIP) (bool, error) {
-	var isUpdated bool
-	if !util.IsEgressIPMarkSet(eip.Annotations) {
-		return isUpdated, nil
+	ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip)
+	if err != nil {
+		return false, fmt.Errorf("failed to delete EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
 	}
-	for _, status := range eip.Status.Items {
-		if status.Node != g.nodeName {
-			continue
-		}
-		if !util.IsEgressIPMarkSet(eip.Annotations) {
-			continue
-		}
-		ip, pktMark, err := parseEIPMarkIP(eip.Annotations, status.EgressIP)
-		if err != nil {
-			return isUpdated, fmt.Errorf("failed to delete EgressIP gateway config because unable to extract config from EgressIP obj: %v", err)
-		}
-		if err = g.deleteIPBridge(ip); err != nil {
-			return isUpdated, fmt.Errorf("failed to delete EgressIP gateway config because failed to delete address from link: %v", err)
-		}
-		g.cache.deleteMarkIP(pktMark, ip)
-		if err = g.deleteIPsFromAnnotation(ip); err != nil {
-			return isUpdated, fmt.Errorf("failed to delete EgressIP gateway config because failed to delete EgressIP IP from Node annotation: %v", err)
-		}
-		isUpdated = true
-		break // no need to continue as only one EIP IP is assigned per node
+	// Skip secondary network IPs. Cleanup of stale secondary IPs from old code is handled in SyncEgressIP.
+	if shouldSkip {
+		return false, nil
 	}
-	return isUpdated, nil
+	if err = g.deleteIPBridge(ip); err != nil {
+		return false, fmt.Errorf("failed to delete EgressIP gateway config because failed to delete address from link: %v", err)
+	}
+	g.cache.deleteMarkIP(pktMark, ip)
+	if err = g.deleteIPsFromAnnotation(ip); err != nil {
+		return false, fmt.Errorf("failed to delete EgressIP gateway config because failed to delete EgressIP IP from Node annotation: %v", err)
+	}
+	return true, nil
 }
 
 func (g *BridgeEIPAddrManager) SyncEgressIP(objs []interface{}) error {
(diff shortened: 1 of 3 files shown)


cgoncalves:
Tested this PR, seems to fix the issue with the EgressIP being added to both OVN primary network interface and secondary host network interface.

Created EgressIP got assigned to expected worker `ovn-worker2`:

```
$ kubectl get eip 
NAME                   EGRESSIPS        ASSIGNED NODE   ASSIGNED EGRESSIPS
egressip-dual-vlan94   192.168.118.30   ovn-worker2     192.168.118.30
```

EgressIP is assigned to the expected `dummy0` interface only:
```
root@ovn-worker2:/# ip a | grep 192.168.118 
    inet 192.168.118.1/24 scope global dummy0
    inet 192.168.118.30/32 scope global dummy
```

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,6 +208,63 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
+	return util.IsOVNNetwork(eIPConfig, ip), nil
+}
+
+// parseAndValidateEIP parses and validates an EgressIP for this node
+// Returns:
+// - ip: the parsed IP address
+// - pktMark: the parsed packet mark
+// - shouldSkip: true if this EgressIP should be skipped (e.g., belongs to secondary network)
+// - error: any error encountered during parsing/validation
+func (g *BridgeEIPAddrManager) parseAndValidateEIP(eip *egressipv1.EgressIP, eipAddr string) (net.IP, util.EgressIPMark, bool, error) {
```
arghosh93: Since we have the entire egressIP object available in this function, can we extract the eipAddr from eip and eliminate passing duplicate data?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,6 +208,63 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
 }
 
+// isOVNNetworkIP checks if the given IP belongs to the OVN (primary) network
+// Returns true if the IP is on the OVN network, false if it's on a secondary network
+func (g *BridgeEIPAddrManager) isOVNNetworkIP(ip net.IP) (bool, error) {
+	node, err := g.nodeLister.Get(g.nodeName)
+	if err != nil {
+		return false, fmt.Errorf("failed to get node %s: %v", g.nodeName, err)
+	}
+	eIPConfig, err := util.ParseNodePrimaryIfAddr(node)
```
arghosh93: Either the variable name or the name of the struct this function returns seems misleading to me. But this is a super NIT and may be ignored.

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,6 +208,63 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
+	// Parse the IP address
+	ip := net.ParseIP(eipAddr)
+	if ip == nil {
+		return nil, pktMark, false, fmt.Errorf("failed to parse EgressIP %s", eipAddr)
+	}
+
+	// Check if this IP belongs to the OVN (primary) network
+	isOVN, err := g.isOVNNetworkIP(ip)
+	if err != nil {
+		return nil, pktMark, false, err
```
arghosh93: Should we pass some error message here to later identify the stage from where it is being emitted?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -217,9 +274,12 @@ func (g *BridgeEIPAddrManager) AddEgressIP(eip *egressipv1.EgressIP) (bool, erro
 		if status.Node != g.nodeName {
 			continue
 		}
-		ip, pktMark, err := parseEIPMarkIP(eip.Annotations, status.EgressIP)
+		ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip, status.EgressIP)
 		if err != nil {
-			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to extract config from EgressIP obj: %v", err)
+			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
+		}
+		if shouldSkip {
```
arghosh93: If we return from here, then neither the combination of egress IP and pkt mark would be inserted into cache, nor the egress IP dedicated for the secondary interface would be added to the node's annotation. I think it's better to persist old configs apart from the addition of IP belonging to the secondary interface subnet to the primary interface.

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -341,8 +410,12 @@ func (g *BridgeEIPAddrManager) SyncEgressIP(objs []interface{}) error {
...
 				continue
 			}
-			if ip, pktMark, err := parseEIPMarkIP(eip.Annotations, status.EgressIP); err != nil {
-				klog.Errorf("Failed to sync EgressIP %s gateway config because unable to extract config from EIP obj: %v", eip.Name, err)
+			ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip, status.EgressIP)
+			if err != nil {
+				klog.Errorf("Failed to sync EgressIP %s gateway config because unable to parse and validate EgressIP: %v", eip.Name, err)
+			} else if shouldSkip {
+				// Skip IPs not on OVN network (i.e., secondary network IPs)
+				klog.V(5).Infof("Skipping EgressIP %s during sync because it does not belong to the OVN network", ip.String())
```
arghosh93: Have you checked whether old egress IPs already added to br-ex interface are being removed or not. I think it would still be there even after applying this patch. Do we have any plan to take care of this?

ricky-rav (approved):
/LGMT
But let's also wait for somebody with more experience than me in egressip code ;)

go-controller/pkg/node/egressip/gateway_egressip.go
ricky-rav: I think we can count on the fact that after a reboot the IP in question won't be on breth0 anymore.

pliurh: The annotation won't be cleaned up after a reboot. I added cleanup logic to `UpdateEgressIP` ~and `DeleteEgressIP`~.

huiran0826:
@jcaamano

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,86 +208,164 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
-			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because failed to add address to link: %v", err)
-		}
-		isUpdated = true
-		break // no need to continue as only one EIP IP is assigned to a node
+	ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip)
+	if err != nil {
+		return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
 	}
+	if shouldSkip {
+		return isUpdated, nil
```
arghosh93: I think we are still returning early in case of secondary egress IP. I think we still need to update mark cache and append the node annotation even if the egress IP is for a secondary network. Are you doing this intentionally?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,86 +208,164 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
+	}
+	// must always add to cache before adding IP because we want to inform node ip handler that this is not a valid node IP
+	g.cache.insertMarkIP(pktMark, ip)
+	if err = g.addIPToAnnotation(ip); err != nil {
+		return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to add EgressIP IP to Node annotation: %v", err)
+	}
+	if err = g.addIPBridge(ip); err != nil {
+		return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because failed to add address to link: %v", err)
+	}
+	isUpdated = true
```
arghosh93: We are not changing the state of this boolean mostly. Can we just return true/false instead?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,86 +208,164 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
-		if err = g.addIPBridge(ip); err != nil {
-			return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because failed to add address to link: %v", err)
-		}
-		isUpdated = true
-		break // no need to continue as only one EIP IP is assigned to a node
+	ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip)
+	if err != nil {
+		return isUpdated, fmt.Errorf("failed to add EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
 	}
+	if shouldSkip {
```
arghosh93, line 216: Do we want to skip adding the egress IP mark to the cache and the egress IP to the node annotation for egress IP on a secondary interface?  I see the same implementation in the updateEgressIP() and syncEgressIP() functions. 
We may avoid adding it to cache as this cant be node IP but I am not sure about the annotation.
Can you please explain a bit if this is intended?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -208,86 +208,164 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
+		} else {
+			// Check if this is a secondary network IP that was mistakenly configured
+			ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(oldEIP)
+			if err != nil {
+				return isUpdated, fmt.Errorf("failed to update EgressIP gateway config because unable to parse and validate old EgressIP: %v", err)
+			}
+
+			// If it's a secondary network IP and in annotation, it needs cleanup even if status didn't change.
+			// This scenario may occur when old code added secondary network IP to the bridge.
+			if shouldSkip && g.isIPInAnnotation(ip) {
```
arghosh93: We are executing deleteEgressIPConfig() function regardless of shouldSkip being true or false. Do we need both conditional checks?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -296,27 +374,18 @@ func (g *BridgeEIPAddrManager) DeleteEgressIP(eip *egressipv1.EgressIP) (bool, e
...
+		return isUpdated, fmt.Errorf("failed to delete EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
 	}
+	// Skip secondary network IPs. Cleanup of stale secondary IPs from old code is handled in UpdateEgressIP.
+	if shouldSkip {
+		return isUpdated, nil
+	}
+	if err = g.deleteEgressIPConfig(ip, pktMark); err != nil {
+		return isUpdated, fmt.Errorf("failed to delete EgressIP gateway config because %v", err)
+	}
+	isUpdated = true
```
arghosh93: we are not setting isUpdated mostly. Can we return either true or false?

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: Do you mean the annotation `k8s.ovn.org/bridge-egress-ips`? As the name suggests, we shall only put the EIP for the bridge interface to this annotation. There is another annotation for the EIP on the secondary interfaces `k8s.ovn.org/secondary-host-egress-ips`.

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: I refactored the function to make it clearer.

go-controller/pkg/node/egressip/gateway_egressip.go
arghosh93: okay. Thanks Peng!

go-controller/pkg/node/egressip/gateway_egressip.go
arghosh93: Looks nice

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -209,115 +209,149 @@ func (g *BridgeEIPAddrManager) GetCache() *MarkIPsCache {
...
-			return isUpdated, fmt.Errorf("failed to delete EgressIP gateway config because failed to delete EgressIP IP from Node annotation: %v", err)
-		}
-		isUpdated = true
-		break // no need to continue as only one EIP IP is assigned per node
+	ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(eip)
+	if err != nil {
+		return false, fmt.Errorf("failed to delete EgressIP gateway config because unable to parse and validate EgressIP: %v", err)
+	}
+	// Skip secondary network IPs. Cleanup of stale secondary IPs from old code is handled in UpdateEgressIP.
+	if shouldSkip {
```
jcaamano, line 289: Why do we handle here the equivalent of `// Scenario 1: Cleanup mistakenly configured secondary network IP`. What does `Cleanup` mean in that sentence?

go-controller/pkg/node/linkmanager/link_network_manager.go
```diff
@@ -152,7 +154,7 @@ func (c *Controller) DelAddress(address netlink.Addr) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	if err := util.GetNetLinkOps().AddrDel(link, &address); err != nil {
-		if !util.GetNetLinkOps().IsLinkNotFoundError(err) {
+		if !util.GetNetLinkOps().IsLinkNotFoundError(err) && !errors.Is(err, unix.EADDRNOTAVAIL) {
```
jcaamano, line 157: Why this change?

pliurh: If the address doesn't exist, we will get `unix.EADDRNOTAVAIL`. It shall not be considered a failure.

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: This is for cleaning up the mistaken configuration that exists before this fix.

jcaamano: Why only do it on update, and not on delete?

go-controller/pkg/node/linkmanager/link_network_manager.go
jcaamano: Right but is this related with the other changes in this PR, or is it an independent issue that you are fixing along with the others? When would this happen?

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: The cleanup will only happen during an upgrade. So the unexpected IP should have already been removed before deleting the EgressIP.

go-controller/pkg/node/linkmanager/link_network_manager.go
pliurh: This function will be invoked when removing the unexpected EIP from the breth0 interface. There is a chance that the EIP only exists in the node annotation `k8s.ovn.org/bridge-egress-ips`, but has not been added to the interface. In that case, the `AddrDel` will fail with `unix.EADDRNOTAVAIL`.

jcaamano: I think this could happen with any EIP, not just unexpected EIPs. That's why I am asking. I think it is a good overall fix.

go-controller/pkg/node/egressip/gateway_egressip.go
jcaamano: Yeah, that's the normal case. But an EIP could also be deleted after OVNK upgrade but before node rollout? Are we exposed to that?

go-controller/pkg/node/linkmanager/link_network_manager.go
pliurh: Yeah, I just found this issue while testing this fix.

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: `SyncEgressIP()` can handle this case. It will be invoked at bootup and clean all stale configurations for each node. So the cleanup in `UpdateEgressIP()` is also unnecessary.

pliurh: `SyncEgressIP()` can handle this case. It will be invoked at bootup and clean all stale configurations for each node. So the cleanup in `UpdateEgressIP()` is also unnecessary.

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -251,9 +246,13 @@ func (g *BridgeEIPAddrManager) UpdateEgressIP(oldEIP, newEIP *egressipv1.EgressI
 			// also, because at most only one status item can be assigned to a node, we can return early.
 			return isUpdated, nil
 		}
-		ip, pktMark, err := parseEIPMarkIP(oldEIP.Annotations, oldStatus.EgressIP)
+		ip, pktMark, shouldSkip, err := g.parseAndValidateEIP(oldEIP)
 		if err != nil {
-			return isUpdated, fmt.Errorf("failed to update EgressIP SNAT for ext bridge cache because unable to extract config from old EgressIP obj: %v", err)
+			return isUpdated, fmt.Errorf("unable to parse and validate old EgressIP: %v", err)
```
jcaamano: Looking into consolidating a bit more...

I think that `IsEgressIPMarkSet` could be considered within `parseAndValidateEIP` returning `shouldSkip` set to `true`? Then we wouldn't have to additionally iterate status items because of it.

We would have to iterate `oldEIP.Status.Items` twice just to do `util.IsItemInSlice(newEIP.Status.Items, oldStatus)`, I guess that is acceptable, but do it separate, not nested iteration: so once here and once within `parseAndValidateEIP`, but not within `parseAndValidateEIP` for every iteration here.

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -486,23 +481,74 @@ func (g *BridgeEIPAddrManager) getAnnotationIPs() ([]net.IP, error) {
...
+func (g *BridgeEIPAddrManager) isOVNNetworkIP(ip net.IP) (bool, error) {
+	node, err := g.nodeLister.Get(g.nodeName)
+	if err != nil {
+		return false, fmt.Errorf("failed to get node %s: %v", g.nodeName, err)
+	}
+	nodePrimaryIPs, err := util.ParseNodePrimaryIfAddr(node)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse node primary interface address for node %s: %v", g.nodeName, err)
+	}
+	return util.IsOVNNetwork(nodePrimaryIPs, ip), nil
```
jcaamano, line 473: Could this rely on `util.IsOVNNetwork` which is what we use elsewhere?

pliurh: I'm not sure I understand your point. Do you mean we shall use `GetNodeEIPConfig()` instead of `ParseNodePrimaryIfAddr()` like we use `util.IsOVNNetwork` elsewhere?

This function is to check if the given IP belongs to the bridge interface. So we shall compare it with the bridge interface subnet range instead of the allowed EIP range.

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: I moved `IsEgressIPMarkSet` into `parseAndValidateEIP` and refactored the `UpdateEgressIP` to reduce the iteration.

go-controller/pkg/node/egressip/gateway_egressip.go
jcaamano: Yeah, sorry, I guess that is what I was trying to say.

I looked into what `GetNodeEIPConfig` does and in all cases it relies in `GetNodeIfAddrAnnotation` which `ParseNodePrimaryIfAddr` also does. So I guess it is fine to use `ParseNodePrimaryIfAddr` and save the time parsing all the other EIP data that `GetNodeEIPConfig` does. 

So what you had before. Sorry again.

jcaamano:
The only other question is: should we reconcile if the node IPs change? However unlikely, I think we technically should. But I don't see the other EIP controller `go-controller/pkg/node/controllers/egressip/egressip.go` concerned about it so I wouldn't do it in this PR and rather a follow up.

jcaamano:
Worth looking at https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5728#pullrequestreview-3607200359

pliurh:
Yeah. I agree that we shall handle the node IPs change case for EIP. And it should be in another PR.

go-controller/pkg/node/egressip/gateway_egressip.go
pliurh: I used `ParseNodePrimaryIfAddr()`. I changed it back.

jcaamano (approved):

PR #5759, opened 2025-11-26
Run tests with Multi-subnet default network

### LINKED ISSUE #5729: Adding coverage for multiple cluster subnets with default networks
### What would you like to see added/updated/deleted in our end-to-end tests?

According to our documentation we imply support for multiple subnets of the same IP family for the ClusterSubnets value:

		Usage: "A comma separated set of IP subnets and the associated " +
			"hostsubnet prefix lengths to use for the cluster (eg, \"10.128.0.0/14/23,10.0.0.0/14/23\"). " +
			"Each entry is given in the form [IP address/prefix-length/hostsubnet-prefix-length] " +
			"and cannot overlap with other entries. The hostsubnet-prefix-length " +
			"defines how large a subnet is given to each node and may be different " +
			"for each entry. For IPv6 subnets, it must be 64 (and does not need to " +
			"be explicitly specified). For IPv4 subnets an explicit " +
			"hostsubnet-prefix should be specified, but for backward compatibility " +
			"it defaults to 24 if unspecified."

It implies multiple subnets of the same family may be used and that they might not even have the same host prefix size.

For OpenShift, the API says:

ClusterNetworkEntry is a subnet from which to allocate PodIPs. A network of size HostPrefix (in CIDR notation) will be allocated when nodes join the cluster. If the HostPrefix field is not used by the plugin, it can be left unset. Not all network providers support multiple ClusterNetworks

It also looks like from the OpenShift API and CNO code that it does not restrict to only a single subnet per IP family.

Since we allow this configuration we should add a way to deploy with with helm and have a CI lane using multiple cluster subnets (along with different host prefix lengths). We should also add documentation to indicate how it may be used.

### Why is this needed?

Looks like this was implicitly supported and some consumers rely on this behavior as discovered in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5725

npinaeva:
## 📑 Description

Add a requirement to have the same hostNetwork length for all networks. See commit message for details. This probably should be fixed in the future.

Enabling multi-subnet for all CI lanes (is there a need to run single-subnet lanes at all?)

Fixes #5729

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Added multi-pod-subnet flag enabling clusters to use multiple pod subnets simultaneously.
  * Extended egress and network isolation policies to generate rules for all subnet combinations.

* **Improvements**
  * Added validation ensuring IPv4 cluster subnets share consistent host subnet lengths.

* **Tests**
  * Expanded test coverage for multi-subnet configurations and multi-destination egress policies.

### DIFF
--- a/go-controller/pkg/ovn/gateway_test.go
+++ b/go-controller/pkg/ovn/gateway_test.go
@@ -41,7 +41,7 @@ func generateAdvertisedUDNIsolationExpectedNB(testData []libovsdbtest.TestData, 
 		if utilnet.IsIPv6CIDR(subnet) {
 			ipPrefix = "ip6"
 		}
-		passMatches = append(passMatches, fmt.Sprintf("(%s.src == %s && %s.dst == %s)", ipPrefix, subnet, ipPrefix, subnet))
+		passMatches = append(passMatches, fmt.Sprintf("((%s.src == %s) && (%s.dst == %s))", ipPrefix, subnet, ipPrefix, subnet))
 
 	}
 	passACL := libovsdbutil.BuildACLWithDefaultTier(
--- a/go-controller/pkg/ovn/udn_isolation.go
+++ b/go-controller/pkg/ovn/udn_isolation.go
@@ -326,17 +326,33 @@ func (bnc *BaseNetworkController) addAdvertisedNetworkIsolation(nodeName string)
 	if err != nil {
 		return fmt.Errorf("failed to get advertised subnets addresset %s for network %s: %w", GetAdvertisedNetworkSubnetsAddressSetDBIDs(), bnc.GetNetworkName(), err)
 	}
-
+	var ipv4Subnets, ipv6Subnets []*net.IPNet
 	for _, subnet := range bnc.Subnets() {
-		ipPrefix := "ip4"
 		if utilnet.IsIPv6CIDR(subnet.CIDR) {
-			ipPrefix = "ip6"
+			ipv6Subnets = append(ipv6Subnets, subnet.CIDR)
+		} else {
+			ipv4Subnets = append(ipv4Subnets, subnet.CIDR)
 		}
-		passMatches = append(passMatches, fmt.Sprintf("(%s.src == %s && %s.dst == %s)", ipPrefix, subnet.CIDR, ipPrefix, subnet.CIDR))
 		cidrs = append(cidrs, subnet.CIDR.String())
-
 	}
-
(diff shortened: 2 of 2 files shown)


go-controller/pkg/ovn/egressip_test.go
```diff
@@ -13593,11 +13593,25 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+						UUID:        "no-reroute-ipv4-3-UUID",
+						ExternalIDs: getEgressIPLRPNoReRoutePodToPodDbIDs(IPFamilyValueV4, types.DefaultNetworkName, fakeOvn.controller.eIPC.controllerName).GetExternalIDs(),
+					},
+					&nbdb.LogicalRouterPolicy{
+						Priority:    types.DefaultNoRereoutePriority,
+						Match:       "ip4.src == 10.132.0.0/14 && ip4.dst == 10.132.0.0/14",
+						Action:      nbdb.LogicalRouterPolicyActionAllow,
+						UUID:        "no-reroute-ipv4-4-UUID",
+						ExternalIDs: getEgressIPLRPNoReRoutePodToPodDbIDs(IPFamilyValueV4, types.DefaultNetworkName, fakeOvn.controller.eIPC.controllerName).GetExternalIDs(),
+					},
```
trozet, line 13614: nice find!

go-controller/pkg/config/utils.go
```diff
@@ -103,6 +105,16 @@ func ParseClusterSubnetEntriesWithDefaults(clusterSubnetCmd string, ipv4HostLeng
...
+			if !ipv6 {
+				if ipv4HostSubnet == 0 {
+					// this is the first ipv4 entry we are processing, record its host subnet length
+					ipv4HostSubnet = parsedClusterEntry.HostSubnetLength
+				} else if parsedClusterEntry.HostSubnetLength != ipv4HostSubnet {
+					return nil, fmt.Errorf("all IPv4 cluster subnet entries must have the same host subnet length; found %d and %d",
+						ipv4HostSubnet, parsedClusterEntry.HostSubnetLength)
+				}
+			}
+
```
trozet, line 117: like you said we should look into fixing this in the future. I'm not sure why we even use src-ip routes in ovn_cluster_router. There was a proposal to modify OVN behavior to be able to prioritize src-ip over dst-ip, which I think is what we should do. But if we can avoid using src-ip routes at all that would be better. I'll follow up with you later.

PR #5770, opened 2025-12-03
Fix coderabbit config

kyrtapz:
Fix the problem with excludes in coderabbit.
Explicitly set the values that are the current defaults as per: https://docs.coderabbit.ai/configuration/yaml-validator.
I think we can consider making the rabbit less chatty in the future but this should be a good start. 

cc: @tssurya @jcaamano

### DIFF
--- a/.coderabbit.yml
+++ b/.coderabbit.yml
@@ -1,2 +1,17 @@
-paths_ignore:
-  - "**/vendor/**"
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: "en-US"
+early_access: false
+reviews:
+  profile: "chill"
+  request_changes_workflow: false
+  high_level_summary: true
+  poem: true
+  review_status: true
+  collapse_walkthrough: false
+  auto_review:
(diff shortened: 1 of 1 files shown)


jcaamano:
> Explicitly set the values that are the current defaults as per: https://docs.coderabbit.ai/configuration/yaml-validator.

Not really sure how that link represents defaults, when I look at the schema I see a lot more properties defaulting to non native default values that we would be missing here. 

Not a problem in my view though, this is as good start as any.

kyrtapz:
> > Explicitly set the values that are the current defaults as per: https://docs.coderabbit.ai/configuration/yaml-validator.
> 
> Not really sure how that link represents defaults, when I look at the schema I see a lot more properties defaulting to non native default values that we would be missing here.
> 
> Not a problem in my view though, this is as good start as any.

You are right, I would still like to keep those as it makes it clear that we are today enabling the chatty options. It also makes it a bit easier to turn those off as a next step.

jcaamano:
how is that @coderabbitai is skipping the review? :laughing:

PR #5773, opened 2025-12-03
fix(kubevirt): Use pod network subnet gateway IP for DHCP router option

qinqon:
## 📑 Description
Replace the hardcoded ARP proxy IP with the actual subnet gateway IP when configuring DHCP options for migratable KubeVirt pods. This ensures VMs receive the correct default gateway matching their pod's subnet configuration.

On scenarios where the primary interfaces has multiple addresses the and there are some nftables masquerade rules, the address to masquerade with is choosen from the routing, so if the default gw is pointing to a link local address then the link local address of the of the interfaces will be choose as src ip, while the one that should be choose is the global address assigned as pod subnet on VM.

Fixes #https://issues.redhat.com/browse/OCPBUGS-65688

This is related to hypershift kubevirt provider where the VMs are worker nodes for a openshift cluster with local gateway mode so masquerading is done with nftables.

## ✅ Checks

- [x] My code requires tests

## How to verify it
The tests are already verifying it

## Summary by CodeRabbit

* **Documentation**
  * Updated live-migration docs with revised gateway examples, clarified DHCP replies advertising the subnet gateway, and refined ARP proxy behavior and routing explanations.

* **Bug Fixes / Behavior**
  * DHCP now uses per-node gateway addresses and collects cluster DNS servers; DHCP failures include pod namespace/name for clearer troubleshooting.
  * Removed one legacy DHCP configuration path for user-defined networks.

* **Tests**
  * Test coverage extended for per-node gateway scenarios (IPv4, IPv6, dual-stack).

### DIFF
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -381,12 +381,12 @@ func (oc *DefaultNetworkController) addLogicalPort(pod *corev1.Pod) (err error) 
 	_ = oc.logicalPortCache.add(pod, switchName, types.DefaultNetworkName, lsp.UUID, podAnnotation.MAC, podAnnotation.IPs)
 
 	if kubevirt.IsPodLiveMigratable(pod) {
-		if err := kubevirt.EnsureDHCPOptionsForMigratablePod(oc.controllerName, oc.nbClient, oc.watchFactory, pod, podAnnotation.IPs, lsp); err != nil {
-			return err
+		if err := oc.ensureDHCP(pod, podAnnotation, lsp); err != nil {
+			return fmt.Errorf("failed configuring DHCP for default network at pod %s/%s: %w", pod.Namespace, pod.Name, err)
 		}
 	}
 
-	//observe the pod creation latency metric for newly created pods only
+	// observe the pod creation latency metric for newly created pods only
 	if newlyCreatedPort {
 		metrics.RecordPodCreated(pod, oc.GetNetInfo())
 	}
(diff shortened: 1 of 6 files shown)


go-controller/pkg/ovn/pods.go
```diff
@@ -381,12 +381,24 @@ func (oc *DefaultNetworkController) addLogicalPort(pod *corev1.Pod) (err error)
...
+		}
+		opts := []kubevirt.DHCPConfigsOpt{
+			kubevirt.WithIPv4DNSServer(dnsServerIPv4),
+			kubevirt.WithIPv6DNSServer(dnsServerIPv6),
+		}
+		ipv4Gateway, _ := util.MatchFirstIPFamily(false /*ipv4*/, podAnnotation.Gateways)
+		if ipv4Gateway != nil {
+			opts = append(opts, kubevirt.WithIPv4Router(ipv4Gateway.String()))
+		}
+		if err := kubevirt.EnsureDHCPOptionsForLSP(oc.controllerName, oc.nbClient, pod, podAnnotation.IPs, lsp, opts...); err != nil {
```
npinaeva: this looks like almost an exact copy of https://github.com/ovn-kubernetes/ovn-kubernetes/blob/575f5cc04674a6f90beda620621697a320eeedb4/go-controller/pkg/ovn/base_network_controller_user_defined.go#L923 (except for the MTU option)? Should they be the same function?

qinqon:
@npinaeva also looks like our hypershift testing confirm that this fix our issues.

go-controller/pkg/ovn/pods.go
qinqon: So later on I found that MTU is configured at QEMU level, so we don't really need to pass it over DHCP, but is not bad to do so, I will unify the functions at the base controller for both default pod and udns.

qinqon:
/retest
unit test failure not related

PR #5777, opened 2025-12-04
Implements DPU Health Check + adds CNI version 1.1.0 support

trozet:
From OKEP #5674

## Summary by CodeRabbit

* **Documentation**
  * Added DPU health monitoring docs describing lease-based health tracking and failure behaviors.

* **New Features**
  * DPU health gating for CNI Add/Status; unhealthy DPUs cause ADD to fail and STATUS to report not-ready.
  * Plugin exposes GC and STATUS CNI operations.

* **Configuration**
  * New flags to control DPU lease renew interval and lease duration.
  * CNI spec/version updated to 1.1.0.

* **Chores**
  * RBAC rules added to allow lease operations.

* **Tests**
  * Added tests for DPU lease lifecycle, STATUS not-ready paths, GC, and updated version expectations.

### DIFF
--- a/go-controller/pkg/cni/cniserver_test.go
+++ b/go-controller/pkg/cni/cniserver_test.go
@@ -167,7 +167,7 @@ func TestCNIServer(t *testing.T) {
 				Config: []byte(cniConfig),
 				DeviceInfo: nadapi.DeviceInfo{
 					Type:    "vdpa",
-					Version: "1.0.0",
+					Version: "1.1.0",
 					Vdpa: &nadapi.VdpaDevice{
 						ParentDevice: "vdpa:0000:65:00.3",
 						Driver:       "vhost",
--- a/go-controller/pkg/cni/types.go
+++ b/go-controller/pkg/cni/types.go
@@ -84,6 +84,9 @@ const CNICheck command = "CHECK"
 // CNIStatus is the command representing a plugin readiness check
 const CNIStatus command = "STATUS"
 
+// CNIGC is the command representing CNI runtime garbage collection
+const CNIGC command = "GC"
+
 // Request sent to the Server by the OVN CNI plugin
 type Request struct {
 	// CNI environment variables, like CNI_COMMAND and CNI_NETNS
--- a/go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
+++ b/go-controller/cmd/ovn-k8s-cni-overlay/ovn-k8s-cni-overlay.go
@@ -25,6 +25,7 @@ func main() {
 				Add:    p.CmdAdd,
 				Check:  p.CmdCheck,
 				Del:    p.CmdDel,
+				GC:     p.CmdGC,
 				Status: p.CmdStatus,
 			},
 			version.All,
--- a/go-controller/pkg/clustermanager/clustermanager_test.go
+++ b/go-controller/pkg/clustermanager/clustermanager_test.go
@@ -1649,7 +1649,7 @@ var _ = ginkgo.Describe("Cluster Manager", func() {
 						)
 					gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
-					const expectedNADContents = `{"cniVersion": "0.4.0", "name": "ovn-kubernetes", "type": "ovn-k8s-cni-overlay"}`
+					const expectedNADContents = `{"cniVersion": "1.1.0", "name": "ovn-kubernetes", "type": "ovn-k8s-cni-overlay"}`
 					gomega.Expect(nad.Spec.Config).To(gomega.Equal(expectedNADContents))
 
 					return nil
--- a/go-controller/pkg/clustermanager/networkconnect/controller_components_test.go
+++ b/go-controller/pkg/clustermanager/networkconnect/controller_components_test.go
@@ -1259,11 +1259,11 @@ func TestNADNeedsUpdate(t *testing.T) {
 	udnOwner := makeUDNOwnerRef("test-udn")
 
 	makePrimaryNADConfig := func(name string) string {
-		return fmt.Sprintf(`{"cniVersion": "0.4.0", "name": "%s", "type": "ovn-k8s-cni-overlay", "topology": "layer3", "role": "primary", "netAttachDefName": "test/%s"}`, name, name)
+		return fmt.Sprintf(`{"cniVersion": "1.1.0", "name": "%s", "type": "ovn-k8s-cni-overlay", "topology": "layer3", "role": "primary", "netAttachDefName": "test/%s"}`, name, name)
 	}
 
 	makeSecondaryNADConfig := func(name string) string {
-		return fmt.Sprintf(`{"cniVersion": "0.4.0", "name": "%s", "type": "ovn-k8s-cni-overlay", "topology": "layer3", "netAttachDefName": "test/%s"}`, name, name)
+		return fmt.Sprintf(`{"cniVersion": "1.1.0", "name": "%s", "type": "ovn-k8s-cni-overlay", "topology": "layer3", "netAttachDefName": "test/%s"}`, name, name)
 	}
 
 	tests := []struct {
--- a/go-controller/pkg/clustermanager/networkconnect/controller_test.go
+++ b/go-controller/pkg/clustermanager/networkconnect/controller_test.go
@@ -73,7 +73,7 @@ func newTestUDNNAD(name, namespace, network string, networkID string) *nadv1.Net
 		},
 		Spec: nadv1.NetworkAttachmentDefinitionSpec{
 			Config: fmt.Sprintf(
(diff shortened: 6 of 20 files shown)


trozet:
Note this now works, but for using with multus it requires  https://github.com/k8snetworkplumbingwg/multus-cni/pull/1470

It also requires a containerd version v2 or later, that includes a go-cni version of v1.1.11+

aserdean:
> Note this now works, but for using with multus it requires [k8snetworkplumbingwg/multus-cni#1470](https://github.com/k8snetworkplumbingwg/multus-cni/pull/1470)
> 
> It also requires a containerd version v2 or later, that includes a go-cni version of v1.1.11+

And as indicated by you it also requires setting the cni version to 1.1.0.

docs/features/hardware-offload/dpu-support.md
```diff
@@ -55,3 +55,15 @@ For detailed configuration of gateway interfaces in DPU host mode, see [DPU Gate
...
+## DPU health monitoring
+
+OVN-Kubernetes uses a custom Kubernetes `Lease` in the `ovn-kubernetes` namespace to track the health of the DPU side of a trusted deployment.
+The DPU host creates the lease and sets an owner reference to the Kubernetes `Node`, while ovnkube running on the DPU renews the lease on a regular interval.
+
+Two ovnkube-node options control this behavior:
+- `--dpu-node-lease-renew-interval` (seconds, default 10). Set to `0` to disable the health check.
+- `--dpu-node-lease-duration` (seconds, default 40).
+
+If the lease expires, the DPU host CNI server fails `ADD` requests immediately with `DPU Not Ready` and the `STATUS` command returns a CNI error with code `50`.
```
aserdean: s/with code `50`/with code `50` (The plugin is not available)

go-controller/pkg/cni/cniserver.go
```diff
@@ -141,7 +154,23 @@ func cniRequestToPodRequest(cr *Request) (*PodRequest, error) {
 	}
 
 	req := &PodRequest{
-		Command: command(cmd),
+		Command:   command(cmd),
+		timestamp: time.Now(),
+	}
+
+	// Match the Kubelet default CRI operation timeout of 2m
+	req.ctx, req.cancel = context.WithTimeout(context.Background(), 2*time.Minute)
```
aserdean: maybe worth moving 2*time.Minute as a const and giving the variable a good name

go-controller/pkg/cni/types.go
```diff
@@ -210,4 +221,5 @@ type Server struct {
 	kubeAuth             *KubeAPIAuth
 	networkManager       networkmanager.Interface
 	ovsClient            client.Client
+	dpuHealth            DPUStatusProvider
```
aserdean, line 224: I'm wondering if this should be a generic name in the off chance we might need a health check for the full mode at some point.

go-controller/pkg/config/config.go
```diff
@@ -236,7 +236,9 @@ var (
 
 	// OvnKubeNode holds ovnkube-node parsed config file parameters and command-line overrides
 	OvnKubeNode = OvnKubeNodeConfig{
-		Mode: types.NodeModeFull,
+		Mode:                      types.NodeModeFull,
+		DPUNodeLeaseRenewInterval: 10,
```
aserdean, line 240: should this be a const in types as well?

go-controller/pkg/config/config_test.go
```diff
@@ -2026,39 +2034,88 @@ udn-allowed-default-services= ns/svc, ns1/svc1
...
+			err := buildOvnKubeNodeConfig(&cliConfig, &config{OvnKubeNode: OvnKubeNode})
+			gomega.Expect(err).To(gomega.HaveOccurred())
+			gomega.Expect(err.Error()).To(gomega.ContainSubstring("dpu-node-lease-duration"))
+		})
+
+		It("Fails if DPU node lease duration is less than or equal to renew interval", func() {
+			cliConfig := config{
+				OvnKubeNode: OvnKubeNodeConfig{
+					Mode:                      types.NodeModeFull,
+					DPUNodeLeaseRenewInterval: 10,
```
aserdean, line 2080: Do we want to add a testcase in which `DPUNodeLeaseRenewInterval` is `0`?

helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
```diff
@@ -235,3 +235,33 @@ rules:
...
+kind: RoleBinding
+metadata:
+    name: ovnkube-node-dpu-leases
+    namespace: ovn-kubernetes
+roleRef:
+    name: ovnkube-node-dpu-leases
+    kind: Role
+    apiGroup: rbac.authorization.k8s.io
+subjects:
+    - kind: ServiceAccount
```
aserdean, line 257: I know that ovnkubeidentity is not enabled on DPU / DPUhosts atm but I believe it would be best if we could gate this similar to other roles, in the off chance this feature will be worked on/enabled, i.e.:
```
subjects:
    {{- if eq (hasKey .Values.global "enableOvnKubeIdentity" | ternary .Values.global.enableOvnKubeIdentity true) true }}
    - kind: Group
      name: system:ovn-nodes
      apiGroup: rbac.authorization.k8s.io
    {{- else }}
    - kind: ServiceAccount
      name: ovnkube-node
      namespace: ovn-kubernetes
    {{- end }}
```

dist/templates/rbac-ovnkube-node.yaml.j2
```diff
@@ -235,3 +235,30 @@ rules:
...
+kind: RoleBinding
+metadata:
+    name: ovnkube-node-dpu-leases
+    namespace: ovn-kubernetes
+roleRef:
+    name: ovnkube-node-dpu-leases
+    kind: Role
+    apiGroup: rbac.authorization.k8s.io
+subjects:
+    - kind: ServiceAccount
```
aserdean, line 255: same ask as here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5777/changes#r2826402162

helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
```diff
@@ -235,3 +235,33 @@ rules:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+    name: ovnkube-node-dpu-leases
+    namespace: ovn-kubernetes
+rules:
+    - apiGroups: ["coordination.k8s.io"]
+      resources:
+          - leases
+      verbs: [ "get", "create", "update" ]
```
aserdean, line 272: this are different verbs than from the jinja template: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5777/changes#diff-4daa87967cc1e7e6e9b9a025289ea7576881ce9d8e3a85750a8e0db0ee65512bR264

do we need: List, Watch, Patch and Delete?

go-controller/pkg/cni/cniserver_test.go
```diff
@@ -287,3 +299,108 @@ func TestCNIServer(t *testing.T) {
...
+			if !strings.Contains(cniErr.Msg, dpuNotReadyMsg) {
+				t.Fatalf("[%s] expected error to mention DPU not ready, got %q", tc.name, cniErr.Msg)
+			}
+			if !strings.Contains(cniErr.Msg, "lease expired") {
+				t.Fatalf("[%s] expected error to include lease reason, got %q", tc.name, cniErr.Msg)
+			}
+		} else if len(body) != 0 {
+			t.Fatalf("[%s] expected empty body for success, got %q", tc.name, string(body))
+		}
+	}
```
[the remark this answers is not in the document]
trozet: all tests are expected to use err = config.PrepareTestConfig(), so this is not really a valid comment. We do not need to restore configs manually after each test.

go-controller/pkg/cni/types.go
trozet: I think we  should for an actual second provider to be proposed. Once that happens we can refactor to be a generic interface like you are thinking.

go-controller/pkg/config/config.go
trozet: I think this is OK for now.

helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
trozet: good point and yes we do want identity to work.

helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
trozet: no it looks like we do not...will remove those from the the j2 path.

aserdean:
LGTM thanks for working through the comments

PR #5787, opened 2025-12-08
Add drop iptables rules for pod subnet

pperiyasamy:
When EgressIP is enabled for the cluster, it may be configured on a secondary interface. Because of this, there can be a delay in programming the IP rule and SNAT iptables rule for a pod that is served by an EgressIP. During this window, if the pod initiates egress traffic, its IP may leak onto the external network through the egress gateway interface (breth0 or br-ex).

This PR adds iptables rules to drop such un-SNATed traffic. Dropping the initial packets is acceptable because they will be retried once the EgressIP rules for the pod are fully programmed.

## Summary by CodeRabbit

* **New Features**
  * Improved gateway rules to better control egress traffic with per-subnet handling and protocol-aware filtering for IPv4/IPv6.
  * Automatic drop rules applied for cluster subnets across forward and output paths to strengthen isolation.

* **Bug Fixes**
  * Fixed ordering of rule application and cleanup to ensure NAT/iptables rules are applied and removed reliably before related routing entries are changed.

### DIFF
--- a/go-controller/pkg/node/controllers/egressip/egressip.go
+++ b/go-controller/pkg/node/controllers/egressip/egressip.go
@@ -815,9 +815,6 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
 }
 
 func (c *Controller) deleteIPConfig(podIPConfigToDelete *podIPConfig) error {
-	if err := c.ruleManager.Delete(podIPConfigToDelete.ipRule); err != nil {
-		return err
-	}
 	if podIPConfigToDelete.v6 {
 		if err := c.iptablesManager.DeleteRule(utiliptables.TableNAT, iptChainName, utiliptables.ProtocolIPv6,
 			podIPConfigToDelete.ipTableRule); err != nil {
@@ -828,6 +825,9 @@ func (c *Controller) deleteIPConfig(podIPConfigToDelete *podIPConfig) error {
 			podIPConfigToDelete.ipTableRule); err != nil {
 			return err
 		}
+	}
+	if err := c.ruleManager.Delete(podIPConfigToDelete.ipRule); err != nil {
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pperiyasamy:
@jcaamano @tssurya can we take this approach to solve pod IP leaking to external network when EIP configured on secondary networks ? Please let me know, Based on the suggestion, need to decide if it worth to investigate the failures with E2E tests (looks like it's introduced after adding drop rules for pod subnet) and updating EIP E2Es.

tssurya:
> @jcaamano @tssurya can we take this approach to solve pod IP leaking to external network when EIP configured on secondary networks ? Please let me know, Based on the suggestion, need to decide if it worth to investigate the failures with E2E tests (looks like it's introduced after adding drop rules for pod subnet) and updating EIP E2Es.

@arghosh93 tells me there are some flaws in the design and we can't do this? why is that?

tssurya:
> @jcaamano @tssurya can we take this approach to solve pod IP leaking to external network when EIP configured on secondary networks ? Please let me know, Based on the suggestion, need to decide if it worth to investigate the failures with E2E tests (looks like it's introduced after adding drop rules for pod subnet) and updating EIP E2Es.

@arghosh93 tells me there are some flaws in the design and we can't do this? why is that? (what I was hoping is nft equivalent of what our drop flows do in br-ex , sorry I haven't checked or reviewed this PR yet)

pperiyasamy:
> > @jcaamano @tssurya can we take this approach to solve pod IP leaking to external network when EIP configured on secondary networks ? Please let me know, Based on the suggestion, need to decide if it worth to investigate the failures with E2E tests (looks like it's introduced after adding drop rules for pod subnet) and updating EIP E2Es.
> 
> @arghosh93 tells me there are some flaws in the design and we can't do this? why is that? (what I was hoping is nft equivalent of what our drop flows do in br-ex , sorry I haven't checked or reviewed this PR yet)

@tssurya I don't think we can use this solution as it is because it drops pod traffic that is not SNATed. so it would break ICNI and BGP usecases.

tssurya:
> > > @jcaamano @tssurya can we take this approach to solve pod IP leaking to external network when EIP configured on secondary networks ? Please let me know, Based on the suggestion, need to decide if it worth to investigate the failures with E2E tests (looks like it's introduced after adding drop rules for pod subnet) and updating EIP E2Es.
> > 
> > 
> > @arghosh93 tells me there are some flaws in the design and we can't do this? why is that? (what I was hoping is nft equivalent of what our drop flows do in br-ex , sorry I haven't checked or reviewed this PR yet)
> 
> @tssurya I don't think we can use this solution as it is because it drops pod traffic that is not SNATed. so it would break ICNI and BGP usecases.

but we should do it only for the pods that are selected as Egress IP pods. (some sort of address-set mechanism? marking mechanism?)

PR #5789, opened 2025-12-09
Fix E2E for metal platform

pperiyasamy:
This PR fixes network segmentation E2E tests to run successfully on metal platforms and makes the test framework more flexible for downstream customization.

## Release Notes

* **Tests**
  * Enhanced end-to-end test reliability with timeout-aware pod startup handling
  * Refactored test infrastructure configuration for improved setup flow
  * Updated test network parameters and IPv4 subnet ranges for better test coverage
  * Added timeout context support to test infrastructure providers

### DIFF
--- a/test/e2e/network_segmentation.go
+++ b/test/e2e/network_segmentation.go
@@ -62,13 +62,13 @@ var _ = Describe("Network Segmentation", feature.NetworkSegmentation, func() {
 		nodeHostnameKey                     = "kubernetes.io/hostname"
 		podClusterNetPort            uint16 = 9000
 		podClusterNetDefaultPort     uint16 = 8080
-		userDefinedNetworkIPv4Subnet        = "172.31.0.0/16" // last subnet in private range 172.16.0.0/12 (rfc1918)
+		userDefinedNetworkIPv4Subnet        = "172.16.0.0/16" // first subnet in private range 172.16.0.0/12 (rfc1918)
 		userDefinedNetworkIPv6Subnet        = "2014:100:200::0/60"
-		customL2IPv4Gateway                 = "172.31.0.3"
+		customL2IPv4Gateway                 = "172.16.0.3"
 		customL2IPv6Gateway                 = "2014:100:200::3"
-		customL2IPv4ReservedCIDR            = "172.31.1.0/24"
+		customL2IPv4ReservedCIDR            = "172.16.1.0/24"
 		customL2IPv6ReservedCIDR            = "2014:100:200::100/120"
-		customL2IPv4InfraCIDR               = "172.31.0.0/30"
+		customL2IPv4InfraCIDR               = "172.16.0.0/30"
 		customL2IPv6InfraCIDR               = "2014:100:200::/122"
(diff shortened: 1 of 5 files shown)


test/e2e/network_segmentation.go
```diff
@@ -2453,7 +2453,7 @@ func runUDNPod(cs clientset.Interface, namespace string, serverPodConfig podConf
 			return v1.PodFailed
 		}
 		return updatedPod.Status.Phase
-	}, 2*time.Minute, 6*time.Second).Should(Equal(v1.PodRunning))
+	}, infraprovider.Get().GetPodReadyPollTimeout(), infraprovider.Get().GetPodReadyPollInterval()).Should(Equal(v1.PodRunning))
```
jcaamano: I would define this as deployment config actually.

I would refactor this to use as much as possible helpers already available in framework pod package: WaitTimeoutForPodRunningInNamespace, WaitTimeoutForPodRunningReadyInNamespace, ...

I don't think we need to define an interval, if we use the framework helpers it will default to TestContext.timeouts.Poll which is 2s, that should be good enough?

It might be usefull to align our timeouts to framework ones:
```
var defaultTimeouts = TimeoutContext{
	Poll:                      2 * time.Second, // from the former e2e/framework/pod poll interval
	PodStart:                  5 * time.Minute,
	PodStartShort:             2 * time.Minute,
	PodStartSlow:              15 * time.Minute,
```

pperiyasamy: yes, right, using `WaitForPodRunningInNamespaceSlow` helper now which is 15 min timeout which is more than enough for dowstream. 2s poll interval should be fine, i guess.

I see `defaultTimeouts` aligns with pod framework timeouts already, but this value is overridden with custom timeout in E2Es. so i think we could just leverage default values for `Eventually` block going forward.

test/e2e/network_segmentation.go
```diff
@@ -2446,14 +2447,13 @@ func runUDNPod(cs clientset.Interface, namespace string, serverPodConfig podConf
...
-	Eventually(func() v1.PodPhase {
-		updatedPod, err = cs.CoreV1().Pods(namespace).Get(context.Background(), serverPod.GetName(), metav1.GetOptions{})
-		if err != nil {
-			return v1.PodFailed
-		}
-		return updatedPod.Status.Phase
-	}, 2*time.Minute, 6*time.Second).Should(Equal(v1.PodRunning))
+	// Retrieve and use pod start timeout value from deployment config.
+	timeouts := deploymentconfig.Get().GetDefaultTimeoutContext()
+	Expect(timeouts).NotTo(BeNil())
```
jcaamano: I wouldn't really be concerned about this being nil, it is not the SUT and we can guarantee by contract it is not. Other existing uses of `framework.NewTimeoutContext()` don't really check for nil.
We can then be more concise when getting a timeout value.

pperiyasamy: agreed, removed `nil` check.

pperiyasamy:
/hold

we may still need adjust UDN network subnets to work with few d/s cluster.

PR #5790, opened 2025-12-09
fix(e2e, kubevirt): Wait for IP after VM restart in ipamless localnet tests

qinqon:
## 📑 Description
Enhance the ipamless localnet e2e test to verify that static IP addresses are
properly configured after a VM restart.
    
This addresses a race condition where the VM may restart but cloud-init hasn't
finished configuring the static IPs yet.

Fixes #https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5456

## ✅ Checks

- [ ] My code requires tests

## How to verify it
Is a e2e check

## Summary by CodeRabbit

* **Bug Fixes**
  * Wait for cloud-init to finish after VM restart and verify static IPs to ensure IP consistency
  * Improved IP address parsing and filtering to skip link-local addresses and better validate guest addresses

* **Tests**
  * Made test steps and reporting more explicit for clearer step-tracking
  * Explicitly set DHCP flags in test network data to clarify IPAM behavior and expected outcomes

### DIFF
--- a/test/e2e/kubevirt.go
+++ b/test/e2e/kubevirt.go
@@ -339,7 +339,7 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 				}
 				// Fail fast
 				Expect(iperfLog).NotTo(ContainSubstring("iperf3: error"), stage+": "+iperfLogFile)
-				//Remove last carriage return to propertly split by new line.
+				// Remove last carriage return to properly split by new line.
 				iperfLog = strings.TrimSuffix(iperfLog, "\n")
 				iperfLogLines := strings.Split(iperfLog, "\n")
 				if len(iperfLogLines) == 0 {
@@ -744,7 +744,6 @@ var _ = Describe("Kubevirt Virtual Machines", feature.VirtualMachineSupport, fun
 				}
 				return familyFn(*iface), nil
 			}
-
 		}
 
 		addressesFromStatus = func(vmi *kubevirtv1.VirtualMachineInstance) func() ([]string, error) {
@@ -1186,7 +1185,6 @@ passwd:
 				}
 				liveMigrateAndCheck(vm.Name, td.mode, endpoints, "after live migration to node owning the subnet")
 			}
-
 		}
 
 		checkPodHasIPAtStatus = func(g Gomega, pod *corev1.Pod) {
@@ -1248,7 +1246,7 @@ fi
 				if err != nil {
 					return nil, err
 				}
-				for _ = range idx {
+				for range idx {
 					ip = iputils.NextIP(ip)
 				}
 				ipNet.IP = ip
@@ -1277,7 +1275,6 @@ fi
 					}
 					pod.Spec.Containers[0].Image = images.IPerf3()
 					pod.Spec.Containers[0].Args = []string{iperfServerScript + "\n sleep infinity"}
-
 				})
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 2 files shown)


test/e2e/kubevirt.go
```diff
@@ -2328,7 +2324,8 @@ chpasswd: { expire: False }
 			iperfServerTestPods, err = createIperfServerPods(selectedNodes, cudn.Name, cudn.Spec.Network.Localnet.Role, filterCIDRs(fr.ClientSet, ipv4CIDR, ipv6CIDR))
 			Expect(err).NotTo(HaveOccurred())
 
-			networkData, err := staticIPsNetworkData(filterCIDRs(fr.ClientSet, vmiIPv4, vmiIPv6))
+			filteredCIDRs := filterCIDRs(fr.ClientSet, vmiIPv4, vmiIPv6)
+			networkData, err := staticIPsNetworkData(filteredCIDRs)
```
maiqueb: I don't understand this change. Isn't this the exact same thing ?

test/e2e/kubevirt.go
```diff
@@ -2384,16 +2381,31 @@ chpasswd: { expire: False }
...
 			waitVirtualMachineInstanceReadiness(vmi)
-			Expect(crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)).To(Succeed())
+			Expect(crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)).To(Succeed(), step)
 
 			step = by(vmi.Name, "Login to virtual machine after virtual machine instance force killed")
 			Expect(virtClient.LoginToFedora(vmi, "fedora", "fedora")).To(Succeed(), step)
 
+			step = by(vmi.Name, "Wait for cloud init to finish after vm restart")
+			output, err = virtClient.RunCommand(vmi, "cloud-init status --wait", time.Minute)
+			Expect(err).NotTo(HaveOccurred(), step+": "+output)
```
maiqueb, line 2396: OK, this is the relevant bit to address the race condition.

test/e2e/kubevirt.go
```diff
@@ -2384,16 +2381,31 @@ chpasswd: { expire: False }
...
+
+			step = by(vmi.Name, "Verify static IP is configured after vm restart")
+			filteredIPs := []string{}
+			for _, filteredCIDR := range filteredCIDRs {
+				filteredIPs = append(filteredIPs, strings.Split(filteredCIDR, "/")[0])
+			}
```
maiqueb, line 2402: maybe move this into an helper ? you could then inline the call in the eventually block below. 

100% opinionated.

test/e2e/kubevirt/ip.go
```diff
@@ -32,8 +33,11 @@ func RetrieveAllGlobalAddressesFromGuest(cli *Client, vmi *v1.VirtualMachineInst
 			continue
 		}
 		for _, address := range iface.Addresses {
-			// Skip non DHCPv6 address
-			if address.Family == "inet6" && address.PrefixLen != 128 {
+			ip := net.ParseIP(address.Local)
+			if ip == nil {
+				return nil, fmt.Errorf("invalid ip address %q", address.Local)
+			}
+			if ip.IsLinkLocalUnicast() {
```
maiqueb, line 40: We're now storing all IPs that are not on the loopback interface, and that are not link local.
Before, we were exclusively looking for ALL IPv4 and ALL non DHCPv6 provides IPv6 addresses. 

AFAIU, you're changing the behavior of this function because you were missing some statically defined IPv6 addresses. Right ?

qinqon: yep, that's it.

test/e2e/kubevirt.go
qinqon: We are using this at just one place, not sure it merits a function

test/e2e/kubevirt.go
maiqueb: To be clear: this was sort of a note to self kind of thing - nothing to do on your end @qinqon

maiqueb:
> I see there's still a flake in the e2e tests, and another test failed.
> 
> Both of them regard live-migration - while this PR focuses on fixing restart.
> 
> Are we sure this fix is required ?

/lgtm

OK I now realize the failed test is for primary UDN: 
> after live migration of VirtualMachine with interface binding for UDN with Primary/Layer2

The fix provided in this PR is for IPAM **less** setups, hence, it does not affect the failed test.

npinaeva:
known flake, added a link to the issue

PR #5551, opened 2025-09-02
fix: multiarch fedora build

killianmuldoon:
enable proper multi-arch fedora builds by building ovn-kubernetes inside the container build

## 📑 Description

Build the ovn-kubernetes binaries inside the Fedora image build. This wil build truly multi-arch containers as part of the Fedora build and release. With the current github workflow the resulting arm64 images contain amd64 binaries. 

Part of #5504 

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## Summary by CodeRabbit

- New Features
  - Streamlined image build via a single Make target, making local builds faster and simpler.
- Refactor
  - Simplified the OVN image build process by removing manual binary collection and metadata steps.
- Chores
  - Updated CI workflow to adjust Docker build context for Fedora images; no impact on functionality or release artifacts.

PR #5794, opened 2025-12-10
ci: Allow build-go.sh to work in git worktrees inside containers

pliurh:
When building container images (e.g., fedora-image), the source directory is copied into the container build context. In git worktrees, the .git file is only a reference to the actual git directory, so git commands fail inside the container when the referenced path is not available.

Add error suppression and fallback to "unknown" for GIT_COMMIT and GIT_BRANCH to allow builds to succeed. Also allow environment variable overrides for CI/CD systems.

## 📑 Description

Fixes the issue that `make fedora-image` doesn't work in a git worktree folder anymore after https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5551

## ✅ Checks

- [ ] My code requires tests

## How to verify it

Create a git worktree, run `make -C dist/images fedora-image` in that directory.

## Summary by CodeRabbit

* **Chores**
  * Improved build process resilience by adding fallback values for version and branch information when unavailable in non-standard contexts.

### DIFF
--- a/go-controller/hack/build-go.sh
+++ b/go-controller/hack/build-go.sh
@@ -16,8 +16,8 @@ build_binaries() {
 
     # Add a buildid to the executable - needed by rpmbuild
     BUILDID=${BUILDID:-0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \n')}
-    GIT_COMMIT=$(git rev-parse HEAD)
-    GIT_BRANCH=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD)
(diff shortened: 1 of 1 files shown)


tssurya:
cc @killianmuldoon

killianmuldoon:
/lgtm

This actually could fix another issue we have with caching the build - including .git in the context means there's ~always changes in the context after copying over the directory. This should allow reducing that by choosing to override the variables and not copying in the .git folder.

pliurh:
@tssurya PTAL

PR #5796, opened 2025-12-10
[retry] Make an exception for pods and retry them infinitely.

npinaeva:
Pods creation relies on many things, including resource availability. The problem I faced is related to a NAD having only 2 IPs (because they are public aka very precious), for 2 pods running in a namespace. This means that a new pod can't be started (e.g. during an upgrade) until one of the old pods is terminated (and have released the IP). If previous pod shutdown goes wrong for whatever reason, the new pod will be crashlooping until the IP is successfully released. If the amount of retry attempts is exceeded in the meantime, the new pod will never come up by itself and will require a manual re-create.

## 📑 Description

Fixes #

I decided to only so this for pods as the most important source of retry problems, but that could potentially be done for all resources.
I am not sure if we still should limit delete retries (considering the pod may be deleted from kapi, but stay in the retry queue forever. On the other hand, releasing resources is important enough to keep retrying, and failures on delete don't happen as often.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Pods now support indefinite retry behavior so pod operations continue retrying until successful rather than being dropped after a fixed number of failures.

* **Tests**
  * Updated pod retry tests to reflect the new retry semantics, including expectations for immediate retry/reset behavior and OVN state assertions after retries.

### DIFF
--- a/go-controller/pkg/retry/obj_retry.go
+++ b/go-controller/pkg/retry/obj_retry.go
@@ -41,6 +41,9 @@ type retryObjEntry struct {
 	backoff   time.Duration
 	// number of times this object has been unsuccessfully added/updated/deleted
 	failedAttempts uint8
+	// infiniteRetry indicates whether this object should be retried indefinitely, regardless of the number of failed attempts
+	// Used for pods only right now
+	infiniteRetry bool
 }
 
 type EventHandler interface {
@@ -146,6 +149,10 @@ func (r *RetryFramework) initRetryObjWithAddBackoff(obj interface{}, lockedKey s
 	entry, _ := r.retryEntries.LoadOrStore(lockedKey, &retryObjEntry{backoff: backoff})
 	entry.timeStamp = time.Now()
 	entry.newObj = obj
+	if _, isPod := obj.(*corev1.Pod); isPod {
+		// for pods we want to retry indefinitely
+		entry.infiniteRetry = true
+	}
 	entry.failedAttempts = 0
 	entry.backoff = backoff
 	return entry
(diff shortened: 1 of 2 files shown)


npinaeva:
> Shall we add an `InfiniteRetry` flag to the `ResourceHandler` or `RetryFramework` struct to make the option configurable instead of binding it to pod only?

Yeah that could be useful too, I am looking for feedback on pod-only vs all resources and whether it should be enabled/disabled by default

go-controller/pkg/retry/obj_retry.go
```diff
@@ -325,7 +340,7 @@ func (r *RetryFramework) resourceRetry(objKey string, now time.Time) {
 			} else if err := r.ResourceHandler.UpdateResource(entry.config, entry.newObj, true); err != nil {
 				entry.timeStamp = time.Now()
 				entry.failedAttempts++
-				if entry.failedAttempts >= MaxFailedAttempts {
+				if entry.failedAttempts >= MaxFailedAttempts && !entry.infiniteRetry {
```
trozet, line 346: could end up overflowing entry.failedAttempts when a entry has infiniteRetry. We have logs that print out  this value like:

klog.Infof("Removing old object: %s %s (failed: %v)", r.ResourceHandler.ObjType, objKey, entry.failedAttempts)

klog.Infof("Retry successful for %s %s after %d failed attempt(s)", r.ResourceHandler.ObjType, objKey, entry.failedAttempts)

probably need to cap incrementing to 255?

trozet:
> > Shall we add an `InfiniteRetry` flag to the `ResourceHandler` or `RetryFramework` struct to make the option configurable instead of binding it to pod only?
> 
> Yeah that could be useful too, I am looking for feedback on pod-only vs all resources and whether it should be enabled/disabled by default

I think setting it to all resources makes sense, but I'm not sure if it is worth it right now. As we transition away from retry framework it becomes moot. So maybe we only fix on failure? If we see this again for another resource type then we just turn it on for all? I'm OK going either direction

go-controller/pkg/retry/obj_retry.go
npinaeva: yeah I think it makes more sense to cap at the max value vs increase the field size

npinaeva:
> I think setting it to all resources makes sense, but I'm not sure if it is worth it right now. As we transition away from retry framework it becomes moot. So maybe we only fix on failure? If we see this again for another resource type then we just turn it on for all? I'm OK going either direction

Yeah I wanted to be more conservative here, but also other types of resources have less dependencies on other things like cluster resources and hence have a smaller chance to succeed with infinite retry. I like you suggested plan

PR #5798, opened 2025-12-11
nodeallocator: fix subnet leak when hybrid overlay is enabled

aswinsuryan:
## 📑 Description

When the hybrid overlay feature is enabled (specifically when hybrid overlay cluster subnets are configured), the HandleDeleteNode function would return early after releasing the hybrid overlay subnet. This caused the regular cluster subnets allocated to the node to never be released, leading to a subnet leak that eventually exhausts the cluster CIDR pool.

This commit fixes the issue by removing the early return, ensuring that both the hybrid overlay subnets and the standard node subnets are properly released upon node deletion.When the hybrid overlay feature is enabled (specifically when hybrid overlay cluster subnets are configured), the HandleDeleteNode function would return early after releasing the hybrid overlay subnet. This caused the regular cluster subnets allocated to the node to never be released, leading to a subnet leak that eventually exhausts the cluster CIDR pool.

This commit fixes the issue by removing the early return, ensuring that both the hybrid overlay subnets and the standard node subnets are properly released upon node deletion.

A new test case TestNodeAllocator_HandleDeleteNode is added to verify that both types of subnets are correctly released.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Node deletion now completes full cleanup: hybrid overlay and cluster subnet allocations—including unmanaged hybrid overlay cases—are properly evaluated and released, preventing lingering allocations and stale state.

* **Tests**
  * Added test coverage confirming that both standard and hybrid overlay subnet allocations are released and usage counts return to zero after node deletion.

### DIFF
--- a/go-controller/pkg/clustermanager/node/node_allocator.go
+++ b/go-controller/pkg/clustermanager/node/node_allocator.go
@@ -343,7 +343,6 @@ func (na *NodeAllocator) syncNodeNetworkAnnotations(node *corev1.Node) error {
 func (na *NodeAllocator) HandleDeleteNode(node *corev1.Node) error {
 	if na.hasHybridOverlayAllocation() {
 		na.releaseHybridOverlayNodeSubnet(node.Name)
-		return nil
 	}
 
 	if na.hasNodeSubnetAllocation() || na.hasHybridOverlayAllocationUnmanaged() {
--- a/go-controller/pkg/clustermanager/node/node_allocator_test.go
+++ b/go-controller/pkg/clustermanager/node/node_allocator_test.go
@@ -12,7 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
-	"k8s.io/client-go/listers/core/v1"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/tools/cache"
 
 	ovncnitypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/cni/types"
@@ -400,12 +400,12 @@ func TestController_allocateNodeSubnets_ReleaseOnError(t *testing.T) {
 	}
 }
 
-func newFakeNodeLister(nodes []*corev1.Node) v1.NodeLister {
+func newFakeNodeLister(nodes []*corev1.Node) listersv1.NodeLister {
 	indexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	for _, node := range nodes {
 		_ = indexer.Add(node)
 	}
-	return v1.NewNodeLister(indexer)
+	return listersv1.NewNodeLister(indexer)
 }
 
 func TestController_CleanupStaleAnnotation(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pliurh (commented):
Only the hybrid overlay nodes should have hybrid overlay subnet allocation. Why are regular cluster subnets allocated?

aswinsuryan:
> Only the hybrid overlay nodes should have hybrid overlay subnet allocation. Why are regular cluster subnets allocated?

In this configuration, when tested, regular cluster nodes are allocated both a standard cluster subnet and a Hybrid Overlay subnet.

pliurh:
According to [code](https://github.com/aswinsuryan/ovn-kubernetes/blob/2e5ec0d1a8a26322fce53f24a5c892af8ba64d4a/go-controller/pkg/clustermanager/node/node_allocator.go#L219-L220) here, only the hybrid overlay nodes should have hybrid overlay subnet allocation. If regular cluster nodes can be allocated hybrid overlay subnet, we shall fix that.

aswinsuryan:
> id overlay nodes should have hybrid overlay subnet allocation. If regular cluster nodes can be allocated hybrid overlay subnet, we shall fix that.

@pliurh Regular nodes have regulat subnet ip allocated , and if the hybrid-overlay is enabled regular subnet ip is not cleared when node is deleted. It is not a case where regular node is allocated hybrid subnet.

pliurh:
I see the issue. But I think we shall fix it by checking if the deleted node is an HO node, instead of removing the early return.

aswinsuryan:
> . But I think we shall fix it by checking if the deleted node is an HO node, instead of removing the early return.



> According to [code](https://github.com/aswinsuryan/ovn-kubernetes/blob/2e5ec0d1a8a26322fce53f24a5c892af8ba64d4a/go-controller/pkg/clustermanager/node/node_allocator.go#L219-L220) here, only the hybrid overlay nodes should have hybrid overlay subnet allocation. If regular cluster nodes can be allocated hybrid overlay subnet, we shall fix that.

@pliurh In testing, I observed that HO nodes did have a primary subnet allocated as well. Shouldn't we always be clearing it to avoid leaks?

pliurh:
We shouldn't allocatie primary subnets to HO nodes when hybrid overlay cluster subnets are configured.

aswinsuryan:
> We shouldn't allocatie primary subnets to HO nodes when hybrid overlay cluster subnets are configured.

@pliurh Oh sorry ,  I was wrong in my  statement , i mean to say with HO enabled the nodes that are not HO will get primary subnet. 

I got what you meant , I restored early return and added a check to see if the node has HO subnet. But it seems code rabbit suggest to remove the early return.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -342,8 +342,10 @@ func (na *NodeAllocator) syncNodeNetworkAnnotations(node *corev1.Node) error {
 // HandleDeleteNode handles the delete node event
 func (na *NodeAllocator) HandleDeleteNode(node *corev1.Node) error {
 	if na.hasHybridOverlayAllocation() {
-		na.releaseHybridOverlayNodeSubnet(node.Name)
-		return nil
+		if _, ok := node.Annotations[hotypes.HybridOverlayNodeSubnet]; ok {
```
pliurh: Can you put these check in one line?

go-controller/pkg/clustermanager/node/node_allocator_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes/fake"
-	"k8s.io/client-go/listers/core/v1"
+	v1 "k8s.io/client-go/listers/core/v1"
```
pliurh: Is this necessary?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -342,8 +342,10 @@ func (na *NodeAllocator) syncNodeNetworkAnnotations(node *corev1.Node) error {
...
+			na.releaseHybridOverlayNodeSubnet(node.Name)
+			return nil
+		}
```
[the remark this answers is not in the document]
pliurh: Nodes are intended to be either hybrid or standard, it is unlikely a node can has both allocation. But cleanup code should be defensive, and `ReleaseAllNetworks` is idempotent. So we can remove skip the return here.

go-controller/pkg/clustermanager/node/node_allocator.go
aswinsuryan: Since we are removing early return we don't need an additional check, so removed it

go-controller/pkg/clustermanager/node/node_allocator_test.go
aswinsuryan: for better readability i made it listerv1

PR #5799, opened 2025-12-12
fix(localnet, ipamless): Prevent LSP deletion on sync

qinqon:
## 📑 Description
When ovnkube-node restarts, it runs syncPodsForUserDefinedNetwork which calls
allocatePodIPs. For IPAM-less localnet networks (switches with no subnets),
IsNonHostSubnetSwitch returns true, causing allocatePodIPs to return empty string.
This prevents the pod from being added to expectedLogicalPorts map, causing
deleteStaleLogicalSwitchPorts to delete the LSP.

This change add an explicit flag at the subnet allocator to denote that
that allocator was created as part of a no host subnet switch, this way
code explicitly diferenciates between localnet ipamless and no host
subnet since both do no have a subnet but no host subnet do not even
have LSPs

Fixes #https://issues.redhat.com/browse/OCPBUGS-66994

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * IP allocation now only runs when allocation is required, avoiding unnecessary attempts and errors.
  * Pods on non-host-subnet (IPAM-less) switches are consistently skipped, improving stability during network processing.

* **Infrastructure**
  * Underlay setup/teardown reworked to centrally discover and manage OVS pods and per-node bridge mappings with a consolidated cleanup flow.

* **Tests**
  * New end-to-end test ensures IPAM-less localnet pod ports and network annotations persist across node restarts.

### DIFF
--- a/go-controller/pkg/ovn/base_network_controller_pods.go
+++ b/go-controller/pkg/ovn/base_network_controller_pods.go
@@ -61,12 +61,21 @@ func (bnc *BaseNetworkController) allocatePodIPsOnSwitch(pod *corev1.Pod,
 	if !util.PodScheduled(pod) || util.PodWantsHostNetwork(pod) {
 		return "", nil
 	}
-	// skip nodes that are not running ovnk (inferred from host subnets)
-	if bnc.lsManager.IsNonHostSubnetSwitch(switchName) {
+
+	// skip nodes that are not running ovnk (inferred from host subnets and
+	// topology type)
+	if bnc.isNonHostSubnetSwitch(switchName) {
 		return "", nil
 	}
 
 	expectedLogicalPortName = bnc.GetLogicalPortName(pod, nadName)
+
+	// For IPAM-less networks (e.g., localnet without subnet ), skip IP allocation.
+	// The logical port should still be tracked to prevent deletion during sync.
+	if !bnc.doesNetworkRequireIPAM() {
+		return expectedLogicalPortName, nil
+	}
+
 	// it is possible to try to add a pod here that has no node. For example if a pod was deleted with
 	// a finalizer, and then the node was removed. In this case the pod will still exist in a running state.
 	// Terminating pods should still have network connectivity for pre-stop hooks or termination grace period
@@ -116,7 +125,7 @@ func (bnc *BaseNetworkController) deleteStaleLogicalSwitchPorts(expectedLogicalP
 		for _, n := range nodes {
 			// skip nodes that are not running ovnk (inferred from host subnets)
 			switchName := bnc.GetNetworkScopedSwitchName(n.Name)
-			if bnc.lsManager.IsNonHostSubnetSwitch(switchName) {
+			if bnc.isNonHostSubnetSwitch(switchName) {
 				continue
 			}
 			switchNames = append(switchNames, switchName)
@@ -397,8 +406,7 @@ func (bnc *BaseNetworkController) podExpectedInLogicalCache(pod *corev1.Pod) boo
 		return false
 	}
 	return !util.PodWantsHostNetwork(pod) &&
-		!(bnc.lsManager.IsNonHostSubnetSwitch(switchName) &&
-			bnc.doesNetworkRequireIPAM()) &&
+		!bnc.isNonHostSubnetSwitch(switchName) &&
 		!util.PodCompleted(pod)
 }
 
@@ -1208,4 +1216,8 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
 		return false
 	}
 	return bnc.releasedPodsBeforeStartup[nad].Has(uid)
+}
+
+func (bnc *BaseNetworkController) isNonHostSubnetSwitch(switchName string) bool {
+	return bnc.doesNetworkRequireIPAM() && bnc.lsManager.IsNonHostSubnetSwitch(switchName)
 }
--- a/go-controller/pkg/ovn/base_network_controller_user_defined.go
+++ b/go-controller/pkg/ovn/base_network_controller_user_defined.go
@@ -282,7 +282,7 @@ func (bsnc *BaseUserDefinedNetworkController) ensurePodForUserDefinedNetwork(pod
 		return nil
 	}
 
-	if bsnc.doesNetworkRequireIPAM() && bsnc.lsManager.IsNonHostSubnetSwitch(switchName) {
+	if bsnc.isNonHostSubnetSwitch(switchName) {
 		klog.V(5).Infof(
 			"Pod %s/%s requires IPAM but does not have an assigned IP address", pod.Namespace, pod.Name)
 		return nil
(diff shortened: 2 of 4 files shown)


jcaamano:
So you are basically saying there can't be local stale ports to clean up because you are ~~not~~ setting `expectedLogicalPorts` for ~~any~~ all local ports. That can't be right? The code is there for a reason?

/cc @cathy-zhou

jcaamano:
@qinqon can you explain a bit better the scenario? Why are we starting up with a pod that is not annotated but the LSP for it already exists?

qinqon:
> @qinqon can you explain a bit better the scenario? Why are we starting up with a pod that is not annotated but the LSP for it already exists?

Somehow during upgrade (VMs get live migrated on pod eviction) the new virt-launcher receiving the transaction get created so first it creates the LSP and then it tries to annotate it (that's not happening, there can be a failure, or similar), then the ovn kubernetes controller get restarted and the pod has LSP but no annotation so the LSP is deleted so VM connection is broken.

jcaamano:
> first it creates the LSP and then it tries to annotate it

how can this happen? I don't see a code path where the LSP gets created without the annotation

qinqon:
> > first it creates the LSP and then it tries to annotate it
> 
> how can this happen? I don't see a code path where the LSP gets created without the annotation

Me either, but the OVS event clearly show that the LSP is deleted and the only place that I found was this, we haven't test it yet since QE cannot reproduce it.

jcaamano:
> > > first it creates the LSP and then it tries to annotate it
> > 
> > 
> > how can this happen? I don't see a code path where the LSP gets created without the annotation
> 
> Me either, but the OVS event clearly show that the LSP is deleted and the only place that I found was this, we haven't test it yet since QE cannot reproduce it.

I am not sure how we can progress with this PR without knowing that combined with my first question: You are no longer handling stale LSPs since you are automatically assuming there are none.

qinqon:
> > > > first it creates the LSP and then it tries to annotate it
> > > 
> > > 
> > > how can this happen? I don't see a code path where the LSP gets created without the annotation
> > 
> > 
> > Me either, but the OVS event clearly show that the LSP is deleted and the only place that I found was this, we haven't test it yet since QE cannot reproduce it.
> 
> I am not sure how we can progress with this PR without knowing that combined with my first question: You are no longer handling stale LSPs since you are automatically assuming there are none.

After some more digging I have reproduce the LSP deletion issue and the root cause is different this is the reproducer
- Create localnet ipamless with pods
- Kill the ovnkube-node pod where the pod is attached
- The LSP get deleted at sync 

The issue is that this functiono here returns "true" https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/ovn/base_network_controller_pods.go#L65

So we end up returning an empty LSP name, so we don't register as existing LSP and we delete the LSP.

I will prepare a fix for this with a proper e2e so it's all clear.

qinqon:
@jcaamano I have re-purpose the PR since the root caus was different, now the test include unit test and e2e, I have verified upstream that sync deleted LSP for localnet ipamless scenario, and this is what happend during upgrades.

jcaamano:
Ideally I would keep the allocator out of it. The allocator is just that, a general purpose allocator. Why does it have to be in charge of keeping track which nodes have been configured in no-overlay (HO)? It made sense while it was equivalent to not having subnets, but if it is no longer equivalent, it doesn't mean we have to continue using it for a non-fit purpose.

It doesn't look like we support HO for localnet (or L2) networks anyway, because we have a single switch, not a switch per node. So if we ever supported it, the allocator wouldn't be an appropriate place to keep track of it anyway.

So for the time being, why not just qualify the call to `IsNonHostSubnetSwitch` to return true only for the network topologies that support it?

qinqon:
> Ideally I would keep the allocator out of it. The allocator is just that, a general purpose allocator. Why does it have to be in charge of keeping track which nodes have been configured in no-overlay (HO)? It made sense while it was equivalent to not having subnets, but if it is no longer equivalent, it doesn't mean we have to continue using it for a non-fit purpose.
> 
> It doesn't look like we support HO for localnet (or L2) networks anyway, because we have a single switch, not a switch per node. So if we ever supported it, the allocator wouldn't be an appropriate place to keep track of it anyway.
> 
> So for the time being, why not just qualify the call to `IsNonHostSubnetSwitch` to return true only for the network topologies that support it?

You mean keeping the subnet check with an extra "no localnet" check ? 

I am ok with that, non having new flag at the allocator make sense.

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1209,3 +1217,8 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
 	}
 	return bnc.releasedPodsBeforeStartup[nad].Has(uid)
 }
+
+func (bnc *BaseNetworkController) isNonOvnkSwitch(switchName string) bool {
```
jcaamano: nit: I would stick with existing terminology `isNonHostSubnetSwitch` (NonHostSubnet being the important part being used widely in the project) until we change it for something better overall.

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -1209,3 +1217,8 @@ func (bnc *BaseNetworkController) wasPodReleasedBeforeStartup(uid, nad string) b
...
+	return (!bnc.IsUserDefinedNetwork() || bnc.TopologyType() == ovntypes.Layer3Topology) &&
+		bnc.lsManager.IsNonHostSubnetSwitch(switchName)
```
jcaamano: Here in "podExpectedInLogicalCache" and in "ensurePodForUserDefinedNetwork" it is being checked like:
```
bsnc.doesNetworkRequireIPAM() && bsnc.lsManager.IsNonHostSubnetSwitch(switchName)
```
Align and reuse?

qinqon: I will re-implement the function with 
```golang
bsnc.doesNetworkRequireIPAM() && bsnc.lsManager.IsNonHostSubnetSwitch(switchName)
```

qinqon:
Looks like after killing ovnkube-node pod, the ovnkube-controller is crashing and make it restart again so it takes more than 3 minutes to recover, I am investigating why ovnkue-controller is timeing out after restart on watching the ipamless nad.

qinqon:
Added a change to skip ip allocation on sync for ipamless, that was not happening before and makes the ovnkube-node to fail after restart.

qinqon:
/retest

CI error not related
```
Summarizing 1 Failure:
  [FAIL] e2e ingress traffic validation Validating ingress traffic to manually added node IPs [BeforeEach] Should be allowed by externalip services to a new node ip
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:1709
```

PR #5623, opened 2025-10-01
Dynamic UDN Allocation Feature Implementation

trozet:
Implementation of this OKEP: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5553

## Summary by CodeRabbit

* **New Features**
  * Dynamic UDN allocation with pod and egress tracking and per-node active-network notifications.

* **Configuration**
  * New CLI/Helm/environment flags to enable dynamic UDN allocation and configure a grace period; exposed at runtime.

* **Testing**
  * CI expanded to include network-segmentation-dynamic; many new unit/integration/e2e tests for dynamic UDN and segmentation scenarios.

* **Metrics**
  * New metric reporting number of nodes rendering dynamic UDNs.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #5800, opened 2025-12-13
Fix EF RegisterNADHandler race + Add developer docs for level-driven, network-aware controllers

trozet:
Refactors NAD Reconcilation for egress firewall controller, and future controllers.

OVN-Kubernetes has scale issues with network controllers today. We spin
up 1 network controller per UDN, which incurs heavy cost when handling
resource object events. The cost for example of unmarshaling a node
annotation can be O(n) where n is the number of UDN controllers that are
parsing the object.

To fix this scale problem, the project has started to move to single
controller instances that are aware of all UDNs. Therefore, handling and
parsing of resource objects are done once. A few controllers have
already moved in this direction, and the remaining components that are
within a UDN controller (such as pod, node, namespace) will be migrated
incrementally.

Furthermore, as we move to per-resource, multi-network aware
controllers, each controller type needs to be able to get network
information. One obvious way to do this is to add a level-driven controller
for NADs in each main controller.
However, this too has a performance cost, because upon each NAD event,
each controller will need to parse the NAD. Since we already have
networkManager (NAD Controller), it is parsing and updating its cache
with the NAD. In order to solve this problem, NetworkManager has been
extended with a "RegisterNADReconciler" function, which is a callback that
controllers can register with NAD Controller to be informed when a NAD
event happens. Controllers can then query NAD Controller to access its
cache as the source of truth.

For example there is commonly used GetActiveNetworkForNamespace, and a
new API, GetPrimaryNADForNamespace is added in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5623, as well as
GetNetInfoForNADKey.

Controllers should be using the pkg/controller Reocnciler framework to
implement their level-driven controllers, which are fed keys externally
from NAD Controller.

@npinaeva @tssurya @jcaamano @kyrtapz @girishmg 

## Summary by CodeRabbit

* **New Features**
  * NAD-based reconciliation: network-attachment changes now drive egress-firewall reconciliation via a reconciler registration mechanism.

* **Improvements**
  * More precise subnet/port-group caching and log hashing for stability.
  * Hardened start/stop and initial sync flows with safer pruning and improved error handling and deferred notifications.

* **Tests**
  * Added NAD reconciliation and egress-firewall sync tests; tests assert reconciliation and resource-version behaviors.

* **Documentation**
  * Added "Level-Driven Controllers" guidance.

### DIFF
--- a/docs/developer-guide/developer.md
+++ b/docs/developer-guide/developer.md
@@ -39,3 +39,52 @@ the `types.go` has been created according to sig-apimachinery docs, the develope
 `make codegen` to be able to generate all the clientgen, listers and informers for the new
 CRD along with the deep-copy methods and actual yaml files which get created in `_output/crd`
 folder and are copied over to `dist/templates` to then be used when creating a KIND cluster.
+
+## Level-Driven Controllers
+
+### Background
+
+OVN-Kubernetes has scale issues with network controllers today. We spin
+up 1 network controller per UDN, which incurs heavy cost when handling
+resource object events. The cost for example of unmarshaling a node
+annotation can be O(n) where n is the number of UDN controllers that are
+parsing the object.
+
+To fix this scale problem, the project has started to move to single
+controller instances that are aware of all UDNs. Therefore, handling and
+parsing of resource objects are done once. A few controllers have
+already moved in this direction, and the remaining components that are
+within a UDN controller (such as pod, node, namespace) will be migrated
+incrementally.
+    
+Furthermore, as we move to per-resource, multi-network aware
+controllers, each controller type needs to be able to get network
+information. One obvious way to do this is to add a level-driven controller
+for NADs in each main controller.
+However, this too has a performance cost, because upon each NAD event,
+each controller will need to parse the NAD. Since we already have
+networkManager (NAD Controller), it is parsing and updating its cache
+with the NAD. In order to solve this problem, NetworkManager has been
+extended with a "RegisterNADReconciler" function, which is a callback that
+controllers can register with NAD Controller to be informed when a NAD
+event happens. Controllers can then query NAD Controller to access its
+cache as the source of truth.
+    
+For example there is commonly used GetActiveNetworkForNamespace, and a
+new API, GetPrimaryNADForNamespace is added in #5623, as well as
+GetNetInfoForNADKey.
+    
+Controllers should be using the pkg/controller Reconciler framework to
+implement their level-driven controllers, which are fed keys externally
+from NAD Controller.
+
+### Guidelines
+
+We prefer level-driven controllers built with `pkg/controller/controller.go`. When adding a new controller:
+
+- Use the shared controller framework rather than bespoke loops or per-network controllers.
+- Design controllers to be User Defined Network (UDN) aware: a single controller instance should reconcile objects across all networks instead of spinning up one instance per network.
+- If the controller is network-aware, do **not** create a separate NAD Controller, use a Reconciler.
+- The network manager is the source of truth for NADs; register a lightweight, non-blocking handler via `RegisterNADReconciler` that will queue keys to the Reconciler.
+- RegisterNADReconciler  **before** starting controller workers to avoid missing events during startup.
+- For a concrete example, see `go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go`.


go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
tssurya, line 1: is there a test gap due to which this race didn't get caught?

npinaeva:
I think there are some more areas to cover if we are considering turning this into a framework.
We are kinds replacing NAD controllers with this callback, but the callback currently has its own filter on which events are even processed https://github.com/ovn-kubernetes/ovn-kubernetes/blob/43b6dd31a339db04f8f6f4c66a7da02933666c6c/go-controller/pkg/networkmanager/nad_controller.go#L448-L451
Other controllers may need to watch other changes, e.g. for CNC we need to resync on label change. Should we provide controller-level functionality and let every callback also register a filter? Or do we just remove the filter completely and react to all events from all controllers?
initialSync is another missing feature compared to normal controllers: even if we call `RegisterNADHandler` first, the callback may be called before the intiialSync is complete, so this part here https://github.com/ovn-kubernetes/ovn-kubernetes/blob/43b6dd31a339db04f8f6f4c66a7da02933666c6c/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go#L341 will be broken, we probably need to put extra conditions on what is allowed to happen in this callback, probably just adding the key to the worker queue should be allowed

trozet:
@npinaeva @tssurya the framework should be more comprehensive now

- Refactored to simplify egress firewall to just reconcile the key in its handler
- Added ability for RegisterNADHandler to take a needsUpdate function, so now we can filter per handler
- Added ability to DeRegisterNADHandler to remove when controller no longer needs it

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -445,10 +489,26 @@ func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmen
...
 
 	// also reconcile the network in case its route advertisements changed
-	return !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec) ||
+	interesting := !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec) ||
 		oldNAD.Annotations[types.OvnRouteAdvertisementsKey] != newNAD.Annotations[types.OvnRouteAdvertisementsKey] ||
 		oldNAD.Annotations[types.OvnNetworkIDAnnotation] != newNAD.Annotations[types.OvnNetworkIDAnnotation] ||
 		oldNAD.Annotations[types.OvnNetworkNameAnnotation] != newNAD.Annotations[types.OvnNetworkNameAnnotation]
+	if !interesting {
+		return false
+	}
```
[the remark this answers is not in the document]
trozet: I'll go with option 3

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -445,10 +489,26 @@ func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmen
...
+		selected := make([]NADHandlerID, 0, len(c.handlers))
+		for id, h := range c.handlers {
+			if h.needsUpdate == nil || h.needsUpdate(oldNAD, newNAD) {
+				selected = append(selected, id)
+			}
+		}
+		c.Lock()
+		c.handlerUpdateDecisions[key] = selected
+		c.Unlock()
+	}
```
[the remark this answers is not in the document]
trozet: will fix

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -225,6 +224,21 @@ func (oc *EFController) initialSync() error {
 		existingEFNamespaces[ef.Namespace] = true
 	}
 
+	// Seed the cache with current namespaces/port-groups so early NAD events
+	// (queued before workers start) can still be cleaned up correctly.
+	for _, ef := range egressFirewalls {
+		if _, ok := oc.cache.Load(ef.Namespace); ok {
+			continue
+		}
+		pgName, err := oc.getNamespacePortGroupName(ef.Namespace)
```
npinaeva: this will give you current pg name (with the current primary network), but stale acls may be applied to a different one?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -110,7 +110,6 @@ type matchKind int
 type cacheEntry struct {
 	pgName          string
 	hasNodeSelector bool
-	stale           bool
```
npinaeva, line 113: awesome!

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,19 +188,24 @@ func (c *nadController) Stop() {
...
-	c.handlerFuncs = append(c.handlerFuncs, handler)
+	c.handlers = append(c.handlers, nadHandler{handler: handler, needsUpdate: needsUpdate})
 	return nil
 }
 
-// executeHandlers should always be done under lock
 func (c *nadController) executeHandlers(nadName string, info util.NetInfo, removed bool) {
-	for _, handler := range c.handlerFuncs {
-		handler(nadName, info, removed)
+	handlers, ok := c.handlerUpdateDecisions[nadName]
```
npinaeva: let's add a comment that this function must be called with lock

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -445,10 +460,26 @@ func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmen
...
 	// also reconcile the network in case its route advertisements changed
-	return !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec) ||
+	interesting := !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec) ||
 		oldNAD.Annotations[types.OvnRouteAdvertisementsKey] != newNAD.Annotations[types.OvnRouteAdvertisementsKey] ||
 		oldNAD.Annotations[types.OvnNetworkIDAnnotation] != newNAD.Annotations[types.OvnNetworkIDAnnotation] ||
 		oldNAD.Annotations[types.OvnNetworkNameAnnotation] != newNAD.Annotations[types.OvnNetworkNameAnnotation]
+	if !interesting {
+		return false
+	}
+	if len(c.handlers) > 0 {
```
npinaeva: I think we need to lock before this check, otherwise you are reading protected field without a lock

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -192,19 +190,50 @@ func (c *nadController) Stop() {
 // usage of this function should be restricted to lightweight, non-blocking operations.
 // An optional needsUpdate predicate can be provided to filter update events for this handler;
 // if nil, all updates are delivered.
-func (c *nadController) RegisterNADHandler(handler handlerFunc, needsUpdate func(old, new *nettypes.NetworkAttachmentDefinition) bool) error {
+// Returns an opaque handler ID that can be used to deregister the handler.
+func (c *nadController) RegisterNADHandler(handler handlerFunc, needsUpdate func(old, new *nettypes.NetworkAttachmentDefinition) bool) (NADHandlerID, error) {
+	c.Lock()
+	defer c.Unlock()
+	c.nextHandlerID++
```
npinaeva: do we hope this will never get out of bound?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -445,10 +489,28 @@ func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmen
 	}
 
 	// also reconcile the network in case its route advertisements changed
-	return !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec) ||
+	interesting := !reflect.DeepEqual(oldNAD.Spec, newNAD.Spec) ||
 		oldNAD.Annotations[types.OvnRouteAdvertisementsKey] != newNAD.Annotations[types.OvnRouteAdvertisementsKey] ||
 		oldNAD.Annotations[types.OvnNetworkIDAnnotation] != newNAD.Annotations[types.OvnNetworkIDAnnotation] ||
 		oldNAD.Annotations[types.OvnNetworkNameAnnotation] != newNAD.Annotations[types.OvnNetworkNameAnnotation]
+	if !interesting {
```
npinaeva: so what if subscriber has a different filter, that selects more events as interesting? the example I had with label changes would be one case

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -325,6 +367,8 @@ func (c *nadController) syncNAD(key string, nad *nettypes.NetworkAttachmentDefin
 
 	c.Lock()
 	defer c.Unlock()
+	// ensure we clear stored update decisions for this NAD once this sync completes
+	defer delete(c.handlerUpdateDecisions, key)
```
npinaeva: shouldn't this be done in `execute_handlers`?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -445,10 +489,28 @@ func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmen
...
+		selected := make([]NADHandlerID, 0, len(c.handlers))
+		c.RLock()
+		for id, h := range c.handlers {
+			if h.needsUpdate == nil || h.needsUpdate(oldNAD, newNAD) {
+				selected = append(selected, id)
+			}
+		}
+		c.RUnlock()
+		c.Lock()
+		c.handlerUpdateDecisions[key] = selected
```
npinaeva: I think this is racy: `nadNeedsUpdate` may be called multiple times before the controller actually runs `syncNAD` , so `handlerUpdateDecisions` will be overridden with the latest event results instead of accumulating all required updates

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,19 +186,55 @@ func (c *nadController) Stop() {
...
+	delete(c.handlers, id)
+	// remove from pending decisions
+	for key, ids := range c.handlerUpdateDecisions {
+		filtered := make([]NADHandlerID, 0, len(ids))
+		for _, hid := range ids {
+			if hid != id {
+				filtered = append(filtered, hid)
+			}
+		}
+		c.handlerUpdateDecisions[key] = filtered
```
npinaeva: considering handler IDs don't ever repeat, I think we can skip this cleanup and just ignore non-existing ids in `executeHandlers` (which is already done?)

go-controller/pkg/networkmanager/nad_controller.go
trozet: From codex: 
NADHandlerID is a uint64 and we just increment nextHandlerID on each registration. In practice that gives ~1.8e19 distinct IDs before wrap. Even if you registered/deregistered a handler every nanosecond, it would take hundreds of years to overflow. So for real-world usage it’s effectively inexhaustible. If
  you ever wanted to be fully defensive, we could add a guard to panic/return an error on wrap, but it’s not realistically needed.

go-controller/pkg/networkmanager/nad_controller.go
trozet: then they should include that match criteria here for now. My thinking is this is temporary and in the future all of this will go away once we get rid of the UDN controllers

go-controller/pkg/networkmanager/nad_controller.go
trozet: no because execute_hanlders is called twice in sync, so we dont want to reset the decisions until sync is done

go-controller/pkg/networkmanager/nad_controller.go
trozet: isn't that more "level driven"? When the workqueue finally processes key, it will get the latest version, lock, then evaluate the handlers.

go-controller/pkg/networkmanager/nad_controller.go
trozet: yeah we can drop the clean up here

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: `handlerUpdateDecisions` is actually event-driven, so let's say we get the first update that handler1 considers important, now we get a second update that handler1 filters out. on the second update we will overwrite `handlerUpdateDecisions[key]` with an empty list, now the actual syncNAD happens and handler1 won't receive its callback

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: can you elaborate on "all of this" that will go away?
I think if we are making this a generic handler, we should maybe change this filter to also check all annotations and labels, I am not sure though how it will affect nad_controller performance..

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: let's add a comment maybe on the field?

go-controller/pkg/networkmanager/nad_controller.go
trozet: i mean we wont filter anymore within nad controller (return true always), and we will leave it up to handlers to register their own needsUpdate. We cant change this now as it would hurt UDN controllers too much.

npinaeva: do you think we can decouple existing handling from the callbacks, so that we can still use callbacks with a different filter, and if nad_controller also find the event interesting, callbacks will be called after nad_controller has handled it?

go-controller/pkg/networkmanager/nad_controller.go
trozet: fixed

go-controller/pkg/networkmanager/nad_controller.go
trozet: added last commit for that

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -244,6 +244,36 @@ func (oc *EFController) initialSync() error {
...
+				}
+			}
+			return false
+		}
+		foundPGs, err := libovsdbops.FindPortGroupsWithPredicate(oc.nbClient, p)
+		if err != nil {
+			return fmt.Errorf("failed to search for port groups during egress firewall cache seed: %w", err)
+		}
+		if len(foundPGs) > 0 {
+			oc.cache.Store(namespace, &cacheEntry{pgName: foundPGs[0].Name})
```
npinaeva, line 321: you can also simplify the cleanup of non-existing namespaces now, because you already know the parent port group?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -317,55 +347,39 @@ func (oc *EFController) initialSync() error {
...
-	return oc.networkManager.RegisterNADHandler(oc.handleNetworkEvent)
+	oc.nadHandlerID = id
+	return controller.StartWithInitialSync(oc.initialSync, oc.controller, oc.nodeController)
 }
 
 func (oc *EFController) Stop() {
 	klog.Infof("%s: shutting down", oc.name)
+	if oc.nadHandlerID != 0 {
+		if err := oc.networkManager.DeRegisterNADHandler(oc.nadHandlerID); err != nil {
+			klog.Warningf("%s: failed to deregister NAD handler: %v", oc.name, err)
```
npinaeva: do you want to do `oc.nadHandlerID = 0` on successful derigestration?

docs/developer-guide/developer.md
```diff
@@ -39,3 +39,14 @@ the `types.go` has been created according to sig-apimachinery docs, the develope
 `make codegen` to be able to generate all the clientgen, listers and informers for the new
 CRD along with the deep-copy methods and actual yaml files which get created in `_output/crd`
 folder and are copied over to `dist/templates` to then be used when creating a KIND cluster.
+
+## Level-Driven Controllers
+
+We prefer level-driven controllers built with `pkg/controller/controller.go`. When adding a new controller:
```
tssurya, line 83: could you also please add the nice benefits on why someone should do this here so that the motivation is clear?

RE:
```
Furthermore, as we move to per-resource, multi-network aware
controllers, each controller type needs to be able to get network
information. One obvious way to do this is to add a reconciler for NADs.
However, this too has a performance cost, because upon each NAD event,
each controller will need to parse the NAD. Since we already have
networkManager (NAD Controller), it is parsing and updating its cache
with the NAD. In order to solve this problem, NetworkManager has been
extended with a "RegisterNADHandler" function, which is a callback that
controllers can register with NAD Controller to be informed when a NAD
event happens. Controllers can then query NAD Controller to access its
cache as the source of truth.
```

Do we know how much gain we get ? any numbers or measurements done already while EFW started using this?

Had asked for the benefits in slack: https://cloud-native.slack.com/archives/C08452HR8V6/p1765810995690419?thread_ts=1765553582.417829&cid=C08452HR8V6 but cross posting here since you didn't reply there.

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -429,7 +495,34 @@ func isOwnUpdate(manager string, managedFields []metav1.ManagedFieldsEntry) bool
...
+
 func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmentDefinition) bool {
+	if oldNAD == nil && newNAD != nil {
+		key := fmt.Sprintf("%s/%s", newNAD.Namespace, newNAD.Name)
+		c.addDecision(key, NetworkControllerUpdateID)
+		return true
+	}
+
 	if oldNAD == nil || newNAD == nil {
 		return true
```
npinaeva, line 456: just returning without updating decisions won't work because there may be another event coming before sync is called, that will update decisions with just one handler and using `handlerUpdateDecisions == nil` to run all handlers won't work anymore

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,20 +188,75 @@ func (c *nadController) Stop() {
...
+	if _, ok := c.handlers[id]; !ok {
+		return fmt.Errorf("handler %d not found", id)
+	}
+	delete(c.handlers, id)
 	return nil
 }
 
-// executeHandlers should always be done under lock
+// executeHandlers runs handlers that were added by RegisterNADHandler.
+// Must be called with nadController locked.
```
npinaeva: I think we need more different locks, otherwise we are preventing adding/deleting handlers while syncNAD is running

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -418,7 +482,9 @@ func (c *nadController) syncNAD(key string, nad *nettypes.NetworkAttachmentDefin
 	}
 
 	// reconcile the network
-	c.networkController.EnsureNetwork(ensureNetwork)
+	if c.nadNeedsNetworkControllerAddUpdate(key) {
```
npinaeva: I don't think leaving `executeHandlers` inside `syncNAD` is reasonable since we have decoupled their filters? you will run the whole syncNAD when it is not needed just to end up calling `executeHandlers` in the end?

npinaeva (commented):
there are still some problems, I have tried simplifying the locking and separating syncNAD from other handlers https://github.com/npinaeva/ovn-kubernetes/commit/daa7bfd4cb7eff7786ae52aef7a2204d92f7b3ab but it has its side effects (like callbacks not getting netInfo anymore)
not sure which way makes more sense, could be something in between..

go-controller/pkg/networkmanager/nad_controller.go
trozet: this if statement really shouldn't be here anymore. I'll remove it

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,20 +188,75 @@ func (c *nadController) Stop() {
 	c.networkController.Stop()
 }
 
-// RegisterNADHandler adds functions to be executed during NAD delete/update/add calls
-// usage of this function should be restricted to lightweight, non-blocking operations
-func (c *nadController) RegisterNADHandler(handler handlerFunc) error {
+// RegisterNADHandler adds functions to be executed during NAD delete/update/add calls.
```
tssurya: what if a controller wants something only on startup? like a list of NADs?
so if a controller wants to at initial sync time fetch/list all unmarshalled/parsed NADs from NM is that possible?

asking cause in evpn we were discussing about udncontroller needing the information about vlanid stored somewhere so that on startup this can be parsed and synced correctly - and this will be stored on nad.spec.conf but parsing all of the nads on startup can be expensive so if NM already does that wondering if there is potential to expand the framework to not register on all events, but have a one time list thing exposed? (doesn't have to be part of this PR - just wanted to explore the possibility of if that's doable)

go-controller/pkg/networkmanager/nad_controller.go
trozet: it is needed, because when controllers register they need to know the netInfo, and they will access things like GetActiveNetworkForNamespace. That is the value of the syncNAD is that it is centrally parsing the information for all the controllers.

npinaeva: you don't need netinfo to call `GetActiveNetworkForNamespace` right? the handler can also get latest netinfo by nadName if needed, but calling the same callback twice on delete and on ensure with different netInfos is also not very level-driven

go-controller/pkg/networkmanager/nad_controller.go
trozet: I'm not sure how much this matters right now. The nadController is single threaded. The only place this could be a problem is really other controllers that add informer handlers for NADs, which the point of this is to not need that. However there may be some that are doing it now already and they could be impacted. I'll see if I can reduce contention.

docs/developer-guide/developer.md
tssurya: clarifying it here as well :) what I mean was what I pasted above which is in your very nice commit message - to put that into the docs. I enjoyed reading the commit/PR message and felt it would be good to have those in the doc as well

go-controller/pkg/networkmanager/nad_controller.go
trozet: Actually thinking this a bit more I'm less convinced it makes sense to add this complexity. The reason is when we syncNAD even if we ensure network controller, it just updates cache and adds a key to the network manager workqueue. So there is hardly any blocking or work being done other than parsing the NAD.

go-controller/pkg/networkmanager/nad_controller.go
trozet: yeah good point. Since NADController starts first, you should be OK. So like what is done in this pr you would do:
1. Start NAD Controller
2. RegisterNADHandler for EVPN (starts queuing events, so you dont miss any)
3. Initial sync - we add a new API GetAllNetworks or GetAllPrimaryNetworks, etc. Then you get all the parsed netinfos/nads and do your thing.

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,20 +188,75 @@ func (c *nadController) Stop() {
...
+		}
+		// reset handlers after notifying all
+		delete(c.handlerUpdateDecisions, nadName)
+		return
+	}
+	// add/update with no filtering, send all
+	for _, h := range c.handlers {
+		h.handler(nadName, info, removed)
+	}
+}
```
npinaeva: why no nad name means update all? shouldn't it mean nothing to update?

go-controller/pkg/networkmanager/nad_controller.go
tssurya: cc @jcaamano / @mattedallo as well for awareness.

jcaamano:
Can we document when `RegisterNADHandler` should be used vs `ControllerManager.Reconcile`/`ReconcilableNetworkController.Reconcile` or maybe deprecate the latter or document the  strong preference of the former vs the latter so that unnecessary new uses of them are not introduced?

go-controller/pkg/networkmanager/nad_controller.go
trozet: it was a fallback, because syncAll does not addDecisions. I'll fix the comment

trozet:
@jcaamano @tssurya 

@npinaeva brought up a point that she would rather see this be more level driven. Instead of integrating so closely with nadController and sending events to callbacks we discussed a different design where:

1. each controller creates a NAD Reconciler (Reconciler is a basic level-driven controller that is fed externally of items)
2. each controller registers with NadController, and NADController queues nad keys to this per-controller NAD Reconciler
3. the controller NAD reconciler worker handles sync'ing the NAD, by querying NADController for latest nad/netInfo, then compares that to its cache to decide what to do next

I think this is probably cleaner. My next push will be just fixes to the comments that exist now, then I'll add another commit  that refactors into level driven and we can see what that looks like.

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,19 +176,31 @@ func (c *nadController) Stop() {
...
+	}
+	return fmt.Errorf("reconciler not found")
+}
+
+// notifyReconcilers enqueues the NAD key to all registered reconcilers
+func (c *nadController) notifyReconcilers(key string) {
+	for _, r := range c.reconcilers {
+		r.Reconcile(key)
 	}
 }
```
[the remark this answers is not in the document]
trozet: I don't think there is a race. notifyReconcilers is deferred under lock in the two places it is called.

trozet:
@tssurya looks like a new flake?
Summarizing 1 Failure:
  [ABORTED] ClusterNetworkConnect ClusterManagerController [JustAfterEach] when CNC is created before networks single network created after CNC: annotations are updated L3 P-UDN [Feature:NetworkConnect]

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20345318211/job/58456766815?pr=5800

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,19 +176,31 @@ func (c *nadController) Stop() {
...
-// executeHandlers should always be done under lock
-func (c *nadController) executeHandlers(nadName string, info util.NetInfo, removed bool) {
-	for _, handler := range c.handlerFuncs {
-		handler(nadName, info, removed)
+// DeRegisterNADReconciler removes a previously registered reconciler.
+func (c *nadController) DeRegisterNADReconciler(r NADReconciler) error {
+	c.Lock()
+	defer c.Unlock()
+	for i, rec := range c.reconcilers {
+		if rec == r {
```
npinaeva: comparing interfaces like this is dangerous: you need to make sure every type of reconciler is actually comparable with all its fields. I think the previous id-based version is safer and faster

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -180,19 +176,31 @@ func (c *nadController) Stop() {
...
+			c.reconcilers = append(c.reconcilers[:i], c.reconcilers[i+1:]...)
+			return nil
+		}
+	}
+	return fmt.Errorf("reconciler not found")
+}
+
+// notifyReconcilers enqueues the NAD key to all registered reconcilers
+func (c *nadController) notifyReconcilers(key string) {
```
npinaeva, line 212: must be called with lock?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -444,6 +455,26 @@ func (c *nadController) nadNeedsUpdate(oldNAD, newNAD *nettypes.NetworkAttachmen
 		return false
 	}
 
+	defer func() {
```
npinaeva, line 471: I had to think a bit about whether we want to call this before or after `isOwnUpdate` let's add a comment saying that since all reconciles are called after the nad_controller has finished its handling, all updates made by the nad_controller would be available for reconcilers on the first try.

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -552,6 +583,24 @@ func (c *nadController) GetNetwork(name string) util.NetInfo {
...
+		return nil
+	}
+	network := c.networkController.getNetwork(networkName)
+	if network == nil && networkName == types.DefaultNetworkName {
+		return &util.DefaultNetInfo{}
+	}
+	if network == nil {
+		return nil
+	}
+	// Return a copy so callers can safely read fields without depending on controller locks.
```
npinaeva, line 613: I think this should go to the interface and function docs too? We have a strange combo, where `GetNetwork`, `GetActiveNetworkForNamespaceFast` and `GetActiveNetwork` return a direct reference to the netInfo and `GetNetInfoForNADKey` and `GetActiveNetworkForNamespace` return a copy. Some of the docs mention that and some don't

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -197,6 +199,19 @@ func NewEFController(
 		nodeControllerConfig,
 	)
 
+	// this controller does not feed from an informer, networks are manually
```
npinaeva: maybe instead of manually say added by the nad_controller?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -376,7 +411,10 @@ func (oc *EFController) sync(key string) (updateErr error) {
 	}
 	shouldRemove := false
 	shouldAddUpdate := false
+	var currentPgName string
+	var currentSubnetsKey string
 
+	efExists := false
```
npinaeva: I find following all the new flags a bit complicated, I suggest instead we just build existing vs desired cache entry and then decide what to do based on the difference, e.g.
```
    existingEntry, _ := oc.cache.Load(namespace)
	var newEntry *cacheEntry

	ef, err := oc.efLister.EgressFirewalls(namespace).Get(efName)
	if err != nil {
		if !apierrors.IsNotFound(err) {
			return err
		}
		// leave newEntry as nil
	} else {
		activeNetwork, err := oc.networkManager.GetActiveNetworkForNamespace(namespace)
		if err != nil {
			if util.IsInvalidPrimaryNetworkError(err) {
				// Namespace requires P-UDN, but it does not exist. Remove EF config and surface error in status.
				updateErr = err
			} else {
				return fmt.Errorf("failed to get active network for egress firewall %s/%s namespace: %w",
					namespace, efName, err)
			}
		} else {
			ownerController := activeNetwork.GetNetworkName() + "-network-controller"
			newEntry = &cacheEntry{
				pgName:            libovsdbutil.GetPortGroupName(getNamespacePortGroupDbIDs(namespace, ownerController)),
				subnetsKey:        subnetsKeyForNetInfo(activeNetwork),
				efResourceVersion: ef.ResourceVersion,
			}
		}
	}
```


then you could do
```
if existingEntry == nil && newEntry == nil ||
		existingEntry != nil && newEntry != nil && newEntry.Equals(existingEntry) {
		// no change
		return nil
	}
```

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -385,32 +423,73 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
+		cachedEntry.subnetsKey == currentSubnetsKey {
+		shouldAddUpdate = false
+	}
+
 	if shouldRemove {
 		// delete case
-		entry, ok := oc.cache.Load(namespace)
-		if !ok {
-			return nil
+		if cachedPGName == "" {
```
npinaeva: reading the remove sequence again, don't we skip cleaning up old dnsNameResolver entries on EF update? we only do extra ACL cleanup I think

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: then you would probably need cleanup on `newEntry == nil` or on port group name change (and maybe a dns cleanup on EF version change like I mentioned in the next comment) and update would be also done in `newEntry != nil`

tssurya:
> @tssurya looks like a new flake?
> Summarizing 1 Failure:
>   [ABORTED] ClusterNetworkConnect ClusterManagerController [JustAfterEach] when CNC is created before networks single network created after CNC: annotations are updated L3 P-UDN [Feature:NetworkConnect]
> 
> https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20345318211/job/58456766815?pr=5800

Let me check ill open an issue for this
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5817

tssurya:
> > @tssurya looks like a new flake?
> > Summarizing 1 Failure:
> > [ABORTED] ClusterNetworkConnect ClusterManagerController [JustAfterEach] when CNC is created before networks single network created after CNC: annotations are updated L3 P-UDN [Feature:NetworkConnect]
> > https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20345318211/job/58456766815?pr=5800
> 
> Let me check ill open an issue for this #5817

thankfully this is not a new flake.. it seems related to some core dump found in bfdd process in frr that leads to test abort. not related to CNC feature at all..

```
udn-w6znx --wait --timeout=60s --ignore-not-found'
2025-12-18T19:07:23.3493589Z   I1218 19:07:23.348904 422782 builder.go:156] stderr: ""
2025-12-18T19:07:23.3495110Z   I1218 19:07:23.348958 422782 builder.go:157] stdout: "userdefinednetwork.k8s.ovn.org \"test-udn-w6znx\" deleted from test-dyn-udn-layer3-9tdx5 namespace\n"
2025-12-18T19:07:23.3652387Z   [1mSTEP:[0m dump namespace information after failure [38;5;243m@ 12/18/25 19:07:23.36[0m
2025-12-18T19:07:23.3654126Z [38;5;204m• [ABORTED] [2.945 seconds][0m
2025-12-18T19:07:23.3656046Z [38;5;204m[1mClusterNetworkConnect ClusterManagerController [JustAfterEach] [38;5;243mwhen CNC is created before networks [0msingle network created after CNC: annotations are updated [38;5;243mL3 P-UDN[0m [38;5;204m[Feature:NetworkConnect][0m
2025-12-18T19:07:23.3658385Z   [38;5;204m[JustAfterEach][0m [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1141[0m
2025-12-18T19:07:23.3660443Z   [38;5;243m[It] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/cluster_network_connect.go:691[0m
2025-12-18T19:07:23.3662269Z 
2025-12-18T19:07:23.3662982Z   [38;5;204m[ABORTED] Coredumps found during test execution: core.426704.bfdd.ovn-control-plane.6[0m
2025-12-18T19:07:23.3664751Z   [38;5;204mIn [1m[JustAfterEach][0m[38;5;204m at: [1m/home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1212[0m [38;5;243m@ 12/18/25 19:07:23.097[0m
2025-12-18T19:07:23.3665924Z [38;5;243m------------------------------[0m
```

`2025-12-18T19:07:23.3662982Z   [38;5;204m[ABORTED] Coredumps found during test execution: core.426704.bfdd.ovn-control-plane.6[0m`

go-controller/pkg/networkmanager/nad_controller.go
trozet: I think those are mistakes. We should always return a copy I think. I'll take a look and add a commit to fix that if so.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -374,44 +397,73 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
 
-	if shouldRemove {
-		// delete case
-		entry, ok := oc.cache.Load(namespace)
-		if !ok {
-			return nil
+	// Delete existing state if there is no desired state.
+	if newEntry == nil {
+		pgName := ""
+		if existingEntry != nil {
```
npinaeva: if existingEntry == nil, there is nothing to cleanup?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -374,44 +397,73 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
+		return updateErr
 	}
 
-	if shouldRemove {
-		// delete case
-		entry, ok := oc.cache.Load(namespace)
-		if !ok {
-			return nil
+	// Delete existing state if there is no desired state.
+	if newEntry == nil {
```
npinaeva, line 453: should we also do this if port gorup name has changed? then you could remove the extra cleanup in the end of this function

npinaeva: maybe skip dnsNameresolved cleanup if newEntry != nil?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -422,37 +474,18 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
-			ef.Namespace, ef.Name, err)
-	}
-
-	entry := &cacheEntry{
-		pgName: pgName,
-	}
-	updateErr = oc.addEgressFirewall(ef, entry)
-	if updateErr != nil {
-		return
+	// Apply desired state.
```
npinaeva, line 482: if newEntry == nil {return}?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: also you would've already returned if both are nil at this point

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: ah it already does that

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -374,44 +398,75 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
+			if logErr != nil {
+				return fmt.Errorf("failed to get acl logging levels for egress firewall %s/%s: %w",
+					namespace, efName, logErr)
+			}
+			ownerController := activeNetwork.GetNetworkName() + "-network-controller"
+			newEntry = &cacheEntry{
+				pgName:            libovsdbutil.GetPortGroupName(getNamespacePortGroupDbIDs(namespace, ownerController)),
+				subnetsKey:        subnetsKeyForNetInfo(activeNetwork),
+				efResourceVersion: ef.ResourceVersion,
+				hasNodeSelector:   efHasNodeSelector(ef),
```
npinaeva: you don't need to set it here though, because we can ignore hasNodeSelector in comparison: hasNodeSelector can't change without efResourceVersion also changing.
So you only need to know if existingEntry has it, and then set it for newEntry in reconcile?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -374,44 +398,75 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
 
-		p := libovsdbops.GetPredicate[*nbdb.ACL](oc.GetEgressFirewallACLDbIDsNoRule(namespace), nil)
-		invalidACLs, err := libovsdbops.FindACLsWithPredicate(oc.nbClient, p)
-		if err != nil {
-			return fmt.Errorf("error finding ACLs for egress firewall %s: %w", key, err)
+	// Delete existing state if there is no desired state.
+	if newEntry == nil {
+		pgName := existingEntry.pgName
+		if pgName == "" {
+			pgName, _ = oc.getNamespacePortGroupName(namespace)
```
npinaeva: this can't happen though? we never add entry to the cache with empty port group

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -374,44 +398,75 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
-			return nil
-		}
-		klog.Infof("Removing egress firewall %s", key)
-		pgName := entry.pgName
+	// If nothing relevant changed since last apply, skip update.
+	if existingEntry != nil && existingEntry.hasNodeSelector {
+		// Node selector rules depend on node state; always recalculate.
+	} else if entriesEqual(existingEntry, newEntry) {
+		return updateErr
+	}
```
npinaeva, line 450: also may be equivalent, but easier to understand for me would be
```
// If nothing relevant changed since last apply, skip update.
	if entriesEqual(existingEntry, newEntry) {
		if existingEntry != nil && existingEntry.hasNodeSelector {
			// Node selector rules depend on node state; always recalculate.
		} else {
			return updateErr
		}
	}

```

trozet: i dont think that is any clearer and it runs entriesEqual which isnt necessary

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: i think this was from when we were handling deletes where the cache may not have an entry or something. I'll remove

npinaeva (approved):
lets pray for green CI

trozet:
@tssurya this one looks like a network connect flake without a coredump:

```
 [38;5;9m[FAIL][0m [0mNetworkConnect ClusterManager Controller Integration Tests [38;5;243mClusterNetworkConnect ClusterManager Annotation Tests [38;5;9m[1m[It] 12. CNC updated when namespace label changes to stop matching Primary UDN selector[0m
2025-12-19T20:32:48.6140620Z   [38;5;243m/home/runner/work/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/clustermanager/networkconnect/controller_test.go:942[0m
```

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20381732091/job/58573609537?pr=5800

PR #5803, opened 2025-12-15
Adding Perf Test GH Action

jtaleric:
## 📑 Description

WIP Adding Performance job 

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Added a CI performance testing pipeline that provisions kind clusters, builds/loads images, runs kube-burner workloads, collects profiles/metrics, uploads artifacts, and posts performance reports.
  * Optional infra nodes with automated Prometheus install and node-targeted metrics collection; new CLI options to configure infra count and metrics endpoint.
  * New performance tooling: standardized metrics, local/remote metric endpoints, workload templates, and a report generator that can save or post reports.

* **Documentation**
  * Added performance tooling README and usage guidance.

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -2,6 +2,8 @@ name: performance-test
 on:
   pull_request:
     branches: [ master ]
+  issue_comment:
+    types: [created]
   schedule:
     - cron: '0 6 * * *'  # Daily at 6 AM UTC
   workflow_dispatch:
@@ -311,7 +313,7 @@ jobs:
         sleep 30
 
         cp ./contrib/perf/metric-endpoint.yml kube-burner/examples/workloads/kubelet-density-cni
-        yes | cp ./contrib/perf/workloads/kubelet-density-cni.yml kube-burner/examples/workloads/kubelet-density-cni/
+        cp -f ./contrib/perf/workloads/kubelet-density-cni.yml kube-burner/examples/workloads/kubelet-density-cni/
         cp ./contrib/perf/performance-meta.yml kube-burner/examples/workloads/kubelet-density-cni
         cp ./contrib/perf/metric-endpoint-local.yml kube-burner/examples/workloads/kubelet-density-cni
         cp ./contrib/perf/metrics.yml kube-burner/examples/workloads/kubelet-density-cni
@@ -326,7 +328,9 @@ jobs:
         else
           kube-burner init --config kubelet-density-cni.yml -e metric-endpoint.yml --user-metadata perf-meta.yml
         fi
-
+        
+        mkdir /tmp/pprof-data
+        cp pprof-data/* /tmp/pprof-data
 
     - name: Export kube-burner data
       if: always()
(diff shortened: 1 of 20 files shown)


.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,276 @@
+name: performance-test
+on:
+  pull_request:
+    branches: [ master ]
+
+env:
+  K8S_VERSION: v1.34.3
```
npinaeva: I think this version doesn't exist, try using 1.34.0 as the other jobs?

contrib/perf/workloads/kubelet-density-cni.yml
```diff
@@ -0,0 +1,30 @@
...
+    objects:
+
+      - objectTemplate: templates/webserver-deployment.yml
+        replicas: 1
+
+      - objectTemplate: templates/webserver-service.yml
+        replicas: 1
+
+      - objectTemplate: templates/curl-deployment.yml
+        replicas: 1
```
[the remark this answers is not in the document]
jtaleric: Template files are in the git clone of kube-burner.

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,515 @@
...
+      KIND_INSTALL_PROMETHEUS: "true"
+      KIND_PROMETHEUS_INFRA_ONLY: "true"
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Setup tmate session
+      id: tmate
+      uses: mxschmitt/action-tmate@v3
+      with:
+        detached: true
```
trozet: can we get rid of this?

dist/images/ovnkube.sh
```diff
@@ -1319,8 +1319,7 @@ ovn-master() {
   fi
   echo "egressservice_enabled_flag=${egressservice_enabled_flag}"
 
-  ovnkube_master_metrics_bind_address="${metrics_endpoint_ip}:9409"
-  ovnkube_master_metrics_bind_address="${metrics_endpoint_ip}:${metrics_master_port}"
+  ovnkube_master_metrics_bind_address="127.0.0.1:${metrics_master_port}"
```
trozet: we cant change it here. Earlier in the file there is:
```
metrics_endpoint_ip=${K8S_NODE_IP:-0.0.0.0}
```

we should change this to something like:
```
metrics_endpoint_ip="${METRICS_IP:-${K8S_NODE_IP:-0.0.0.0}}"
```

then we can propagate the value from kind.sh -> templates -> ovnkube.sh. This is a hassle but it is the way things work today. So in kind.sh and kind-helm.sh you can expose a new argument like --metrics-ip, which sets a var METRICS_IP. Then when ./daemonset.sh is called you pass --metrics-ip=$METRICS_IP, then in the template files like dist/templates/ovnkube-single-node-zone.yaml.j2 you edit this file to render the env variable METRICS_IP, and finally when ovnkube.sh runs it will use this value.

.github/workflows/performance-test.yml
```diff
@@ -0,0 +1,526 @@
...
+
+jobs:
+  performance-job:
+    name: performance
+    runs-on: [oracle-vm-32cpu-128gb-x86-64]
+    timeout-minutes: 240
+    needs: [build-pr]
+    permissions:
+      contents: read
+      pull-requests: write
```
trozet, line 38: i think you need
issues: write

might not need pull-requests: write, but i guess its ok for now

trozet (changes requested):

trozet (approved):

PR #5801, opened 2025-12-15
CUDN controller: EVPN configuration translation and VID allocation

mattedallo:
## 📑 Description

Depends on #5779

This PR extends the UDN (User Defined Network) controller to support EVPN networks by:
1. Translating EVPN configuration from CUDN API to NAD (NetworkAttachmentDefinition) CNI config
2. Implementing cluster-wide VID (VLAN ID) allocation for EVPN VRFs
3. Validating VTEP existence for EVPN CUDNs with reactive reconciliation

VIDs are required in the Single VXLAN Device (SVD) architecture to allow the Linux bridge to map traffic to the correct VNI.

### Commit 1: EVPN configuration translation to NAD
- Extends `SpecGetter` interface with `GetTransport()` and `GetEVPN()` methods
- Adds `renderEVPNConfig()` to translate EVPN config from CUDN API to CNI NetConf format
- Adds transport type constants

### Commit 2: Cluster-wide VID allocation
- Adds `vidAllocator` to the UDN controller for cluster-wide VID allocation (range 1-4094)
- Allocates one VID per VRF (MAC-VRF and IP-VRF can have separate VIDs)
- Adds VID field to `VRFConfig` in CNI types
- Implements VID recovery on controller restart from existing NADs
- Releases VIDs when CUDN is deleted
- Exposes VID via `NetInfo` interface (`EVPNMACVRFVID`, `EVPNIPVRFVID`)

### Commit 3: VTEP validation for EVPN CUDNs
- Validates VTEP existence before creating NADs for EVPN networks
- Reports VTEPNotFound status condition with actionable error message
- Adds VTEPNotifier to watch VTEP CRs and trigger CUDN reconciliation
- Enables eventual consistency: CUDNs and VTEPs can be created in any order

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [x] My code requires tests
- [x] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

## How to verify it

Unit tests have been added:
- `controller_test.go`: Integration tests for EVPN NAD creation with VID, VTEP validation, and reactive reconciliation 
- `net-attach-def-template_test.go`: Tests for EVPN config rendering
- `multi_network_test.go`: Tests for VID exposure via NetInfo
- `notifier/vtep_test.go`: Tests for VTEPNotifier create/delete notifications

## Summary by CodeRabbit

* **New Features**
[...]

PR #5807, opened 2025-12-17
Add EVPN support to RouteAdvertisements controller 

kyrtapz:
Based on: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5801

Generate FRR configuration for networks with EVPN configured:
- Activate EVPN on global router with advertise-all-vni
- MAC-VRF: configure route targets if specified
- IP-VRF: create VRF-to-VNI mappings and per-VRF BGP routers with
  unicast advertisement and route targets
- Create routers for IP-VRF networks if not already selected
Uses raw FRR config until frr-k8s provides typed EVPN API.

Note: I've opted not to explicitly set the defaults for RD/RT and let FRR handle it. This ensures we rely on the default behavior. 

Remaining bit is to figure out if we are ok with the possible repetition/duplication when multiple frrconfigs are selected by an RA. 

## Summary by CodeRabbit

* **New Features**
  * Added EVPN support for route advertisements, including MAC‑VRF and IP‑VRF vrf/VNI handling and generated EVPN FRR configuration.

* **Improvements**
  * Deterministic ordering and stronger validation for EVPN networks (router requirements, consistency checks).
  * Improved diagnostic message when EVPN transport is requested but the feature is disabled.

* **Tests**
  * Added comprehensive tests covering MAC‑VRF, IP‑VRF, dual‑stack, and combined EVPN scenarios.

### DIFF
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller.go
@@ -527,9 +527,13 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements) 
 		}
 	}
 
+	// Validate EVPN configuration requirements
+	hasEVPNConfig := len(selectedNetworks.macVRFConfigs) > 0 || len(selectedNetworks.ipVRFConfigs) > 0
+	if hasEVPNConfig && !util.IsEVPNEnabled() {
+		return nil, nil, fmt.Errorf("%w: EVPN networks selected but EVPN feature is not enabled", errConfig)
+	}
 	// Require a router with default VRF for any EVPN configuration, since the
 	// global EVPN section with advertise-all-vni is required for EVPN to work properly.
-	hasEVPNConfig := len(selectedNetworks.macVRFConfigs) > 0 || len(selectedNetworks.ipVRFConfigs) > 0
 	if hasEVPNConfig && !frrRouterVRFs.Has("") {
 		return nil, nil, fmt.Errorf("%w: EVPN requires a router with default VRF but none were found in selected FRRConfigurations", errConfig)
 	}
--- a/go-controller/pkg/clustermanager/routeadvertisements/controller_test.go
+++ b/go-controller/pkg/clustermanager/routeadvertisements/controller_test.go
@@ -1331,6 +1331,7 @@ exit
 			config.OVNKubernetesFeature.EnableMultiNetwork = true
 			config.OVNKubernetesFeature.EnableRouteAdvertisements = true
 			config.OVNKubernetesFeature.EnableEgressIP = true
+			config.OVNKubernetesFeature.EnableEVPN = true
 
 			fakeClientset := util.GetOVNClientset().GetClusterManagerClientset()
 			addGenerateNameReactor[*frrfake.Clientset](fakeClientset.FRRClient)
@@ -1683,6 +1684,7 @@ func TestUpdates(t *testing.T) {
 			config.OVNKubernetesFeature.EnableMultiNetwork = true
 			config.OVNKubernetesFeature.EnableRouteAdvertisements = true
 			config.OVNKubernetesFeature.EnableEgressIP = true
+			config.OVNKubernetesFeature.EnableEVPN = true
 
 			fakeClientset := util.GetOVNClientset().GetClusterManagerClientset()
 
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -1016,7 +1016,7 @@ func (c *Controller) validateEVPNVTEP(cudn *userdefinednetworkv1.ClusterUserDefi
 	}
 
 	if !util.IsEVPNEnabled() {
-		return fmt.Errorf("EVPN transport requested but enable-evpn flag is not set")
+		return fmt.Errorf("EVPN transport requested but EVPN feature is not enabled")
 	}
 
 	// CEL validation ensures EVPN is set when transport is EVPN.
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
@@ -165,7 +165,7 @@ func (c *Controller) allocateEVPNVIDsIfNeeded(obj client.Object) ([]template.Ren
 
 	// EVPN transport is requested - ensure the feature is enabled.
 	if !util.IsEVPNEnabled() {
-		return nil, fmt.Errorf("EVPN transport requested but enable-evpn flag is not set")
+		return nil, fmt.Errorf("EVPN transport requested but EVPN feature is not enabled")
 	}
 
 	evpnCfg := spec.GetEVPN()
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
@@ -1288,7 +1288,7 @@ var _ = Describe("User Defined Network Controller", func() {
 					Type:    "NetworkCreated",
 					Status:  "False",
 					Reason:  "NetworkAttachmentDefinitionSyncError",
-					Message: "EVPN transport requested but enable-evpn flag is not set",
+					Message: "EVPN transport requested but EVPN feature is not enabled",
 				}}), "should report error when EVPN flag is disabled")
 
 				// NAD should not be created when EVPN is disabled
--- a/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
@@ -200,7 +200,7 @@ func renderCNINetworkConfig(networkName, nadName string, spec SpecGetter, opts *
 
 	if spec.GetTransport() == userdefinednetworkv1.TransportOptionEVPN {
 		if !util.IsEVPNEnabled() {
-			return nil, fmt.Errorf("EVPN transport requested but enable-evpn flag is not set")
+			return nil, fmt.Errorf("EVPN transport requested but EVPN feature is not enabled")
 		}
 		netConfSpec.EVPN = renderEVPNConfig(spec, opts)
 	}
--- a/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template_test.go
@@ -969,7 +969,7 @@ var _ = Describe("NetAttachDefTemplate", func() {
 
 			_, err := RenderNetAttachDefManifest(cudn, "test-ns")
 			Expect(err).To(HaveOccurred())
-			Expect(err.Error()).To(ContainSubstring("EVPN transport requested but enable-evpn flag is not set"))
+			Expect(err.Error()).To(ContainSubstring("EVPN transport requested but EVPN feature is not enabled"))
 		})
 	})
 


go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +782,49 @@ func (c *Controller) generateFRRConfiguration(
...
+			if router.VRF == "" { // default VRF
+				globalRouterASN = router.ASN
+				for _, neighbor := range router.Neighbors {
+					neighbors = append(neighbors, neighbor.Address)
+				}
+				break
+			}
+		}
+
+		if globalRouterASN == 0 {
```
kyrtapz: This check enforces that every frrconfig matched by an RA for an EVPN network must contain a global router. I think this is acceptable—it keeps the logic simple, and the potential repetition shouldn't be a problem for frr-k8s. I’d like to hear what others think.

If we were to loosen this constraint, we would need to perform the check outside of generateFRRConfiguration and only configure EVPN bits for frrconfigs that actually contain a global router.

go-controller/pkg/clustermanager/routeadvertisements/evpn_rawconfig.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	buf.WriteString("  advertise-all-vni\n")
+
+	for _, cfg := range macVRFs {
+		if cfg.RouteTarget == "" {
+			continue
+		}
+		fmt.Fprintf(&buf, "  vni %d\n", cfg.VNI)
+		fmt.Fprintf(&buf, "   route-target import %s\n", cfg.RouteTarget)
+		fmt.Fprintf(&buf, "   route-target export %s\n", cfg.RouteTarget)
```
jcaamano, line 99: Any idea why we wouldn't need `exit-vni` statement here?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -48,8 +48,9 @@ import (
 )
 
 const (
-	generateName = "ovnk-generated-"
-	fieldManager = "clustermanager-routeadvertisements-controller"
+	generateName          = "ovnk-generated-"
+	fieldManager          = "clustermanager-routeadvertisements-controller"
+	evpnRawConfigPriority = 10
```
jcaamano, line 54: Any basis on choosing this priority? vs max int lol

go-controller/pkg/clustermanager/routeadvertisements/evpn_rawconfig.go
```diff
@@ -0,0 +1,134 @@
+package routeadvertisements
+
+// TODO: Delete this file when frr-k8s adds typed API support for EVPN.
```
jcaamano: TODO makes more sense to me where `generateEVPNRawConfig` is called

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -385,6 +411,40 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
...
+				},
+				NetworkName: networkName,
+				VRFName:     vrf,
+				HasIPv4:     hasIPv4,
+				HasIPv6:     hasIPv6,
+			})
+		}
+		if (len(selectedNetworks.macVRFConfigs) > 0 || len(selectedNetworks.ipVRFConfigs) > 0) && ra.Spec.TargetVRF == "" {
+			return nil, nil, fmt.Errorf("%w: advertising networks with EVPN configuration to default target VRF is not supported", errConfig)
+		}
```
jcaamano: I would be more restrictive with this check. I would check that TargetVRF is either `auto` or equal to the CUDN vrf. Otherwise you end up in a situation were you are configuring prefixes in the unicast address family on one router but configuring the IPVRF in a different router (since you are not using the target VRF in line 440).

go-controller/pkg/clustermanager/routeadvertisements/controller.go
jcaamano: I would loosen the constraint, not to avoid repetition (even if it could be avoided when using raw configuration, we will definitely have it when we move to the formal API) but rather to be aligned with our current approach of being able to find any router in any of the selected configs. 

Keeping things as they are now, `generateEVPNRawConfig` could potentially just silently ignore that part when asn|neighbors are not provided. The only other additional requirement would be to eventually know if the default VRF was found or not once all `generateFRRConfiguration` has been done just as we currently do with other VRFs, and report an error otherwise of which overall does not seem a like a great deal of extra complication because it would just be returning that piece of information back from `generateFRRConfiguration` as we currently do with `matchNetworks` which you might want to reuse for this adding `types.DefaultNetwork` or some other reserved keyword to it.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +782,49 @@ func (c *Controller) generateFRRConfiguration(
...
+		if slices.ContainsFunc(routers, func(r frrtypes.Router) bool { return r.VRF == cfg.VRFName }) {
+			continue
+		}
+		klog.Infof("Creating router for EVPN network %q VRF %q with ASN=%d, prefixes=%v",
+			cfg.NetworkName, cfg.VRFName, globalRouterASN, selectedNetworks.hostNetworkSubnets[cfg.NetworkName])
+		routers = append(routers, frrtypes.Router{
+			ASN:      globalRouterASN,
+			VRF:      cfg.VRFName,
+			Prefixes: selectedNetworks.hostNetworkSubnets[cfg.NetworkName],
+		})
```
jcaamano: I guess I have a similar comment as before, to be able to find the router defined in any of the configs. With raw config this is not a problem, but we will need this with the formal API.

You could just look into `matchedNetworks` at the end of the process and generate an extra configuration for the unmatched ones.

I guess I am fine with the current approach but we will need to do something else later on.

BTW, aren't we triggering this error in this scenario?
```
		// check that we matched all the selected networks on 'auto'
		if ra.Spec.TargetVRF == "auto" && !matchedNetworks.HasAll(selectedNetworks.networks...) {
			return nil, nil, fmt.Errorf("%w: selected FRRConfigurations for node %q don't match all selected networks with target VRF 'auto'", errConfig, nodeName)
		}
```

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +782,49 @@ func (c *Controller) generateFRRConfiguration(
...
+	// If a router doesn't exist for an EVPN network's VRF, create it here since
+	// EVPN routers don't need neighbors (they use the global router's neighbors).
+	for _, cfg := range selectedNetworks.ipVRFConfigs {
+		if slices.ContainsFunc(routers, func(r frrtypes.Router) bool { return r.VRF == cfg.VRFName }) {
+			continue
+		}
+		klog.Infof("Creating router for EVPN network %q VRF %q with ASN=%d, prefixes=%v",
+			cfg.NetworkName, cfg.VRFName, globalRouterASN, selectedNetworks.hostNetworkSubnets[cfg.NetworkName])
+		routers = append(routers, frrtypes.Router{
+			ASN:      globalRouterASN,
```
jcaamano: hmm I didn't consider this ASN when I thought about generating these routers. While using the underlay ASN is one of the alternatives that makes sense, I am not sure if it is too much of an assumption or if it would be better to require these to be explicitly configured or even have ASN as a CUDN IPVRF field.

well, worst case the flexibility is there

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +782,49 @@ func (c *Controller) generateFRRConfiguration(
...
+			VRF:      cfg.VRFName,
+			Prefixes: selectedNetworks.hostNetworkSubnets[cfg.NetworkName],
+		})
 	}
 
+	if len(routers) == 0 && len(selectedNetworks.macVRFConfigs) == 0 {
+		// we ended up with no routers and there is no macVRF to configure, bail out
+		return nil, nil
+	}
```
jcaamano: wouldn't we prematurely bail out in an scenario where we just want to apply the global evpn configuration?

go-controller/pkg/clustermanager/routeadvertisements/evpn_rawconfig.go
kyrtapz: Good spot! not sure how I missed that, added it now.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: I wanted to pick something that allows the user to override it if required, added a comment.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: Done.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: Done, I followed your idea of adding a special key to `matchNetworks`.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: Yeah, I think it is a reasonable default and if a user provides the vrf router we should use it's asn instead.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: We wouldn't, it's an and condition.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: I modified the logic to find the router if any frrcfg, should be good now.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -385,6 +417,41 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
...
+					VNI:         ipVNI,
+					RouteTarget: network.EVPNIPVRFRouteTarget(),
+				},
+				NetworkName: networkName,
+				VRFName:     vrf,
+				HasIPv4:     hasIPv4,
+				HasIPv6:     hasIPv6,
+			})
+		}
+		if (len(selectedNetworks.macVRFConfigs) > 0 || len(selectedNetworks.ipVRFConfigs) > 0) && (ra.Spec.TargetVRF != "auto" && ra.Spec.TargetVRF != vrf) {
```
jcaamano: I usually like to break out this long conditions to some more readable form like
```
hasEVPNConfig := len(selectedNetworks.macVRFConfigs) > 0 || len(selectedNetworks.ipVRFConfigs) > 0

if hasEVPNConfig && ra.Spec.TargetVRF != "auto" && ra.Spec.TargetVRF != vrf {
```

Subjective style but sharing with you in case you will take it

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -577,6 +661,10 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
 		if ra.Spec.TargetVRF == "auto" && !matchedNetworks.HasAll(selectedNetworks.networks...) {
 			return nil, nil, fmt.Errorf("%w: selected FRRConfigurations for node %q don't match all selected networks with target VRF 'auto'", errConfig, nodeName)
 		}
+		// check that we found a global router if MAC-VRF EVPN is configured.
+		if len(selectedNetworks.macVRFConfigs) > 0 && !matchedNetworks.Has(evpnGlobalRouterKey) {
```
jcaamano: What is the practical difference of `matchedNetworks.Has(evpnGlobalRouterKey)` vs `frrRouterVRFs.Has("")`? If there is none, we could consolidate to one of the approaches. If there is any, then are we sure we need two different interpretations of what the global router is?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +809,65 @@ func (c *Controller) generateFRRConfiguration(
...
 	}
 
+	// For IP-VRF: Find or create routers for each EVPN network's VRF.
+	// IP-VRF routers don't need neighbors for EVPN (they use the global router's neighbors).
+	for _, cfg := range selectedNetworks.ipVRFConfigs {
+		if frrRouterVRFs.Has(cfg.VRFName) {
+			// VRF router exists somewhere - check if it's in the current source
+			for _, router := range source.Spec.BGP.Routers {
+				if router.VRF == cfg.VRFName {
+					matchedNetworks.Insert(cfg.NetworkName)
```
jcaamano: Did you actually need this `matchedNetworks.Insert(cfg.NetworkName)`? Wouldn't that happen already somewhere in line 725? Not an issue, but I am curious.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +809,65 @@ func (c *Controller) generateFRRConfiguration(
...
+					}
+					break
+				}
+			}
+			// If not in current source, another source will handle it
+		} else if globalRouterASN > 0 {
+			// VRF router doesn't exist anywhere - create with global ASN
+			klog.Infof("Creating router for EVPN network %q VRF %q with ASN=%d, prefixes=%v",
+				cfg.NetworkName, cfg.VRFName, globalRouterASN, selectedNetworks.hostNetworkSubnets[cfg.NetworkName])
+			matchedNetworks.Insert(cfg.NetworkName)
```
jcaamano, line 853: I guess this is how you avoided the last concern of this [comment](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5807#discussion_r2676277645)

go-controller/pkg/clustermanager/routeadvertisements/controller.go
jcaamano: I understand that but this condition can still be true if:
1. Among all the selected network we are only handling IPVRFs, and no MACVRFs
2. All IPVRFs are matched on a different FRRConfiguration
3. This is the only FRRConfiguration with a global router

can't it?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: Yeah, I this splitting it out is more readable.
Done.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: Exactly.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: You are right, updated the logic to account for that.

kyrtapz:
Rebase: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/3b4ba3b4bdf107f76954bccf6886ac98af1348bf..ebb0fb8814393be92313e6daa69d942fa183056e

kyrtapz:
Addressing review comments: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/ebb0fb8814393be92313e6daa69d942fa183056e..6c89de7b90972aeada2470f35344494c47ac3ec5

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: I was sure there was a scenario that required it but I cannot figure out what it was...
I removed it as I cannot think of any scenario where the earlier insert would be enough.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
kyrtapz: You are right! I removed the `evpnGlobalRouterKey` approach and moved the check closer together for better readability.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -455,6 +522,22 @@ func (c *Controller) generateFRRConfigurations(ra *ratypes.RouteAdvertisements)
...
+			frrRouterVRFs.Insert(router.VRF)
+		}
+	}
+
+	// Validate IP-VRF networks: each needs either an existing VRF router or
+	// a global router to create one from.
+	for _, cfg := range selectedNetworks.ipVRFConfigs {
+		// The global router is represented by an empty string in the set.
+		if !frrRouterVRFs.Has(cfg.VRFName) && !frrRouterVRFs.Has("") {
+			return nil, nil, fmt.Errorf("%w: IP-VRF EVPN network %q requires a router with VRF %q or a global router, but none were found in selected FRRConfigurations", errConfig, cfg.NetworkName, cfg.VRFName)
```
jcaamano: nit: user facing error and not convinced with the `global router` term...`a router with default VRF`?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -720,11 +805,66 @@ func (c *Controller) generateFRRConfiguration(
...
+	}
+
+	// For IP-VRF: Find or create routers for each EVPN network's VRF.
+	// IP-VRF routers don't need neighbors for EVPN (they use the global router's neighbors).
+	for _, cfg := range selectedNetworks.ipVRFConfigs {
+		if frrRouterVRFs.Has(cfg.VRFName) {
+			// VRF router exists somewhere - check if it's in the current source
+			for _, router := range source.Spec.BGP.Routers {
+				if router.VRF == cfg.VRFName {
+
```
jcaamano: nit: weird looking blank line

jcaamano (commented):
I have one remaining concern.

There is an scenario where we might be selecting a network with just an IP-VRF and FRRConfiguration with a router on that VRF but no router on default VRF. The result is no underlay EVPN configuration generated, and if that is also not generated through a different RA, we end up with a configuration that doesn't work.

So I am guessing if we need to actually require a router on the default VRF being selected. While there are other approaches this might be the simplest

kyrtapz:
hit a known flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4221 
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21512455093/job/61983553800?pr=5807

kyrtapz:
PR is approved and CI is green, merging it. 
Thanks @jcaamano !

PR #5461, opened 2025-08-07
add new default network advertisement toggle tests (pod to external & pod to another node)

jechen0648:
… test

## 📑 Description

Add new default network advertisement toggle test, here is the sequence of events/tests:
1.  When default network is advertised, curl external server from client pod, podIP should be used 
2. Create another internal server pod on another node, curl the internal server pod from client, nodeIP should be used 
3. Delete default network RA, repeat 1, nodeIP should be used
4. While default network unadveristed, repeat 2, nodeIP should be used 
5. Apply default network RA back to restore original test setup

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [x ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## How to verify it

I have a local run test log:  https://privatebin.corp.redhat.com/?c5c1ebdfd4a9ace9#3QdLF1eK6qTuf1EjZgCQnezbcRuhBfSBn9VwR6SL7iVg

## Summary by CodeRabbit

* **Tests**
  * Expanded e2e coverage for route advertisements to validate connectivity via pod IP (no-SNAT) and host-node IP (SNAT) across multiple nodes.
  * Added a host-networked pod scenario on a second node with a randomized high port and collection of relevant node IPs; dual-target checks against an external server and a second cluster node (IPv4/IPv6 where supported).
  * Added polling for RouteAdvertisement acceptance, automated restoration with abort on failure, and repeated connectivity verification.
  * Introduced a shared netexec port constant and clarified BGP test descriptions.

PR #5821, opened 2025-12-22
Refactor BGP Setup for Default Network BGP Session

jechen0648:
## 📑 Description

This PR refactors the BGP cluster bootstrap setup for the **default network BGP session** by replacing the previous dependency on an external `demo.sh` script and a git-patched repository with a self-contained Go-based toolchain that shares its core FRR configuration logic with the existing e2e route advertisement tests.

### Problem

The previous shell-based BGP setup in `kind-common.sh` had several fragility issues:

- Relied on cloning an external repository and applying a git patch at runtime
- Duplicated FRR template rendering that already existed inline in `route_advertisements.go`
- `go build -trimpath` (used in CI) strips `runtime.Caller()` paths, breaking template discovery at runtime
- The FRR image version was hardcoded in shell and Go code independently, creating drift

### Module Hierarchy and Dependency

Three components are involved, arranged in a clear dependency hierarchy:

```
test/e2e/infraprovider/frr/frr.go        ← shared library (no external dependencies)
        ↑                    ↑
        |                    |
test/e2e/cmd/bgp-setup/    test/e2e/route_advertisements.go
  main.go                   (e2e test suite)
(standalone binary)
```

**`test/e2e/infraprovider/frr/frr.go` — shared FRR configuration package**

- Owns the `Router` and `FRRKubernetesConfig` types that model FRR BGP peer configuration
- Provides a `TemplateSource` abstraction (embedded templates compiled into the binary via `//go:embed`, or filesystem-based for development)
- Exposes `GenerateFRRDaemonConfig` and `GenerateFRRK8sConfig` / `GenerateFRRK8sConfigToDir` to render FRR and frr-k8s configuration from those types
- Has no dependency on any e2e test framework (`ginkgo`, `k8s.io/kubernetes/test/e2e/framework`)

**`test/e2e/cmd/bgp-setup/main.go` — standalone cluster bootstrap binary**

- Imports `infraprovider/frr` for template rendering — the only test-infrastructure package it uses
- Uses `frr.NewEmbeddedSource()` so templates are baked into the binary (no runtime file lookup)
- Implements its own `os/exec` wrappers (`runCmd`, `runKubectl`) rather than reusing helpers from `test/e2e/executor.go` or `infraprovider/providers/kind/kind.go` because those helpers transitively pull in `k8s.io/kubernetes/test/e2e/framework` and `ginkgo`, both of which require test suite initialization (`RegisterFailHandler`, `RunSpecs`) that never happens in a standalone binary — the same class of problem as the `deploymentconfig.Get()` panic that was observed before the `--agnhost-image` flag was added
- Accepts all image references (`--frr-image`, `--agnhost-image`) and other configuration as command-line flags so it is self-contained and requires no e2e framework initialization

**`test/e2e/route_advertisements.go` — e2e test suite**

- Also imports `infraprovider/frr` for template rendering — the same package `bgp-setup` uses
- Runs inside a full Ginkgo test suite where framework initialization has already occurred, so it can use `images.FRR()` and other framework helpers freely
- Does **not** call `bgp-setup`; it uses the `infraprovider/frr` types directly to generate configs that it applies to the already-running FRR container that `bgp-setup` deployed during cluster setup

**Execution flow — cluster setup (kind.sh)**

```
kind.sh
  └─ run_bgp_setup deploy-frr          # Phase 1a, before OVN
  └─ run_bgp_setup deploy-bgp-server   # Phase 1b, before OVN
  └─ (install OVN / ovn-kubernetes)
  └─ run_bgp_setup install-frr-k8s     # Phase 2, after OVN + CNI ready
```

`bgp-setup` is built once per invocation of `kind.sh` (lazy rebuild via mtime check) and executed as a subprocess. The FRR container it deploys in Phase 1a is the same container that the e2e tests will later configure through `infraprovider/frr`.

**Execution flow — e2e tests (route_advertisements.go)**

```
Ginkgo suite
  └─ BeforeSuite / test setup
       └─ route_advertisements.go calls frr.GenerateFRRDaemonConfig(...)
                                    and frr.GenerateFRRK8sConfig(...)
          using the already-running FRR container (deployed by bgp-setup)
```

The e2e tests do not re-deploy infrastructure; they reconfigure the existing FRR container by writing updated configs and reloading.

### Solution

**4 logical commits:**

1. **`test/e2e/infraprovider`: extract shared FRR configuration package**
   Pulls the inline FRR template rendering from `route_advertisements.go` into a reusable `infraprovider/frr` package. Uses `//go:embed` so templates are compiled into binaries. `route_advertisements.go` is updated to use `frr.NewEmbeddedSource()`.

2. **`test/e2e/cmd/bgp-setup`: add standalone Go tool for BGP cluster bootstrap**
   Introduces the `bgp-setup` binary with phase-based execution. Fixes a panic caused by `images.AgnHost()` calling `deploymentconfig.Get()` (uninitialized outside a test suite) by adding an `--agnhost-image` flag. Adds `--frr-image` flag so the image version is sourced from `${FRR_DEPLOYED_IMAGE}` in the shell environment rather than being hardcoded.

3. **`contrib`: delegate BGP cluster provisioning to bgp-setup (phased execution)**
   Removes ~360 lines of shell FRR/BGP functions from `kind-common.sh` and `kind-helm.sh`, replacing them with a `run_bgp_setup()` function that builds and invokes `bgp-setup` at the correct phases:

   | Phase | Timing | What |
   |---|---|---|
   | `deploy-frr` | Before OVN | FRR container ready |
   | `deploy-bgp-server` | Before OVN | BGP server ready |
   | `install-frr-k8s` | After OVN | CNI available for frr-k8s |

   Cluster destroy now cleans up stale Kind network endpoints and the bgpserver container.

4. **`test/e2e, contrib`: add EVPN and managed routing support to bgp-setup**
   Ports the remaining feature areas from the original shell implementation:
   - **EVPN** (`--enable-evpn`): configures l2vpn evpn address-family, activates all BGP neighbors as route-reflector-clients, and sets `advertise-all-vni`. Waits for `"Neighbor"` in `show bgp summary` before applying config to avoid races with bgpd startup. Sets `net.ipv6.conf.all.keep_addr_on_down=1` for IPv6 VRF enslavement.
   - **Managed routing** (`--bgp-port`): patches frr-k8s bgpd to listen on port 179 before OVN starts; splits `install-frr-k8s` (apply manifest only) from `wait-frr-k8s` (readiness wait post-OVN) because frr-k8s pods cannot become ready until the CNI is available.

### Testing

Validated by building a BGP Kind cluster using `contrib/kind.sh` with the refactored scripts, verifying:
- Standard BGP cluster
- EVPN cluster (`--enable-evpn`)
- Managed routing cluster (`ENABLE_NO_OVERLAY_MANAGED_ROUTING=true`)

Assisted-By: claude-4.5-opus-high

Fixes #

See the "Module Hierarchy and Dependency" section above for a detailed explanation of how `infraprovider/frr`, `cmd/bgp-setup/main.go`, and `route_advertisements.go` relate to each other and the two execution flows (cluster setup vs. e2e).

## ✅ Checks

- [ ] My code requires tests

## How to verify it

Build a BGP-enabled KinD cluster:
```bash
KIND_INSTALL_INGRESS=false OVN_HA=false ENABLE_MULTI_NET=false \
  ENABLE_ROUTE_ADVERTISEMENTS=true \
  ./contrib/kind.sh
```

Run route advertisement e2e tests:
```bash
./test/scripts/e2e-kind.sh -focus "route-advertisement"
```

### DIFF
--- a/contrib/kind-common.sh
+++ b/contrib/kind-common.sh
@@ -1264,7 +1264,7 @@ run_bgp_setup() {
     --ipv6="$ipv6_flag" \
     --bgp-server-subnet-ipv4="${BGP_SERVER_NET_SUBNET_IPV4}" \
     --bgp-server-subnet-ipv6="${BGP_SERVER_NET_SUBNET_IPV6}" \
-    --frr-k8s-version="${FRR_K8S_VERSION:-v0.0.21}" \
+    --frr-k8s-version="${FRR_K8S_GIT_REF:-${FRR_K8S_VERSION:-v0.0.21}}" \
     --frr-image="${FRR_DEPLOYED_IMAGE}" \
     --network-name="${NETWORK_NAME:-default}" \
     --cluster-name="${KIND_CLUSTER_NAME:-ovn}" \
--- a/contrib/kind-common.sh
+++ b/contrib/kind-common.sh
@@ -1220,229 +1220,6 @@ readonly FRR_DEPLOYED_IMAGE=quay.io/frrouting/frr:10.6.0
 # without changing the pinned frr-k8s release.
 FRR_K8S_FRR_IMAGE=${FRR_K8S_FRR_IMAGE:-${FRR_DEPLOYED_IMAGE}}
 readonly FRR_EXTERNAL_DEMO_IMAGE=${FRR_DEPLOYED_IMAGE}
-readonly FRR_TMP_DIR=$(mktemp -d -u)
-
-clone_frr() {
-  [ -d "$FRR_TMP_DIR" ] || {
-    mkdir -p "$FRR_TMP_DIR" && trap 'rm -rf $FRR_TMP_DIR' EXIT
-    pushd "$FRR_TMP_DIR" || exit 1
-    git clone --no-tags --single-branch --branch main https://github.com/metallb/frr-k8s
-    pushd frr-k8s
-    git checkout --detach "$FRR_K8S_GIT_REF"
-    popd
-
-    # Download the patches
-    curl -Ls https://github.com/jcaamano/frr-k8s/archive/refs/heads/ovnk-bgp-v0.0.21.tar.gz | tar xzvf - frr-k8s-ovnk-bgp-v0.0.21/patches --strip-components 1
-
-    # Change into the cloned repo directory before applying patches
-    pushd frr-k8s
-    # The OVN-K demo patch was authored before upstream bumped the demo
-    # image. Normalize that context before applying the patch; the image is
-    # bumped to FRR_EXTERNAL_DEMO_IMAGE below.
-    sed -i 's|quay.io/frrouting/frr:10.4.3|quay.io/frrouting/frr:9.1.0|g' hack/demo/demo.sh
-    git apply ../patches/*
-
-    # The upstream frr-k8s demo.sh hardcodes quay.io/frrouting/frr:9.1.0,
-    # which crashes on musl libc (Alpine) due to a race condition in
-    # pthread_setname_np during BGP keepalive thread startup
-    # (https://github.com/FRRouting/frr/issues/15699, fixed in FRR 10.1 by
-    # https://github.com/FRRouting/frr/pull/15714).
-    #
-    # Bump to 10.4.1 for upstream demo was posted here: https://github.com/metallb/frr-k8s/pull/404
-    # We bump further to 10.6.0 to include additional fixes for EVPN and coredumps:
-    # https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5874#issuecomment-3907335193
-    # https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5874#issuecomment-3898408592
-    # https://github.com/FRRouting/frr/pull/20496
-    #
-    # Note: 10.6.0 carries the bfdd coredump fix:
-    # https://github.com/FRRouting/frr/pull/19822
-    # https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6299
-    replace_in_file_or_exit \
-      hack/demo/demo.sh \
-      "${FRR_K8S_UPSTREAM_FRR_IMAGE}" \
-      "${FRR_EXTERNAL_DEMO_IMAGE}"
-
-    popd
-
-    popd || exit 1
-  }
-}
-
-deploy_frr_external_container() {
-  echo "Deploying FRR external container ..."
-  clone_frr
- 
-  pushd "$FRR_TMP_DIR" || exit 1
-  run_kubectl apply -f frr-k8s/charts/frr-k8s/charts/crds/templates/
-  popd || exit 1
- 
-  # apply the demo which will deploy an external FRR container that the cluster
-  # can peer with acting as BGP (reflector) external gateway
-  pushd "${FRR_TMP_DIR}"/frr-k8s/hack/demo || exit 1
-  # modify config template to configure neighbors as route reflector clients
-  # First check if IPv4 network already exists
-  grep -q 'network '"${BGP_SERVER_NET_SUBNET_IPV4}" frr/frr.conf.tmpl || \
-    sed -i '/address-family ipv4 unicast/a \ \ network '"${BGP_SERVER_NET_SUBNET_IPV4}"'' frr/frr.conf.tmpl
-
-  # Add route reflector client config
-  sed -i '/remote-as 64512/a \ neighbor {{ . }} route-reflector-client' frr/frr.conf.tmpl
-
-  if [ "$PLATFORM_IPV6_SUPPORT" == true ]; then
-    # Check if IPv6 address-family section exists
-    if ! grep -q 'address-family ipv6 unicast' frr/frr.conf.tmpl; then
-      # Add IPv6 address-family section if it doesn't exist
-      sed -i '/exit-address-family/a \ \
-  address-family ipv6 unicast\
-    network '"${BGP_SERVER_NET_SUBNET_IPV6}"'\
-  exit-address-family' frr/frr.conf.tmpl
-    else
-      # Add network to existing IPv6 section
-      sed -i '/address-family ipv6 unicast/a \ \ network '"${BGP_SERVER_NET_SUBNET_IPV6}"'' frr/frr.conf.tmpl
-    fi
-
-    # Add route-reflector-client for IPv6 neighbors
-    sed -i '/neighbor fc00.*remote-as 64512/a \ neighbor {{ . }} route-reflector-client' frr/frr.conf.tmpl
-  fi
-  if [ "${OCI_BIN}" == "podman" ]; then
-    # frr-k8s' demo script prefers docker when both docker and podman are
-    # installed. Force its podman path, and avoid its host-network fallback
-    # because podman cannot later attach a host-network container to bgpnet.
-    replace_in_file_or_exit \
-      ./demo.sh \
-      'CLI=docker' \
-      'CLI=podman'
-    replace_in_file_or_exit \
-      ./demo.sh \
-      'CLI_BR_NET_BY_SUBNET_FN="docker_get_br_net_by_subnet"' \
-      'CLI_BR_NET_BY_SUBNET_FN="podman_get_br_net_by_subnet"'
-    sed -i '/^pushd \.\/frr\/ && {/i\
-if [ "$CLI" = "podman" ]; then\
-    NETWORK=${FRR_K8S_DEMO_NETWORK:-kind}\
-fi\
-' ./demo.sh
-  fi
-  ./demo.sh
-  popd || exit 1
-  if frr_k8s_remote_enabled && [ -n "${DPU_SIM_GATEWAY_NETWORK:-}" ]; then
-    configure_dpu_sim_frr_gateway_peers
-  fi
-  if  [ "$PLATFORM_IPV6_SUPPORT" == true ]; then
-    # Enable IPv6 forwarding in FRR
-    $OCI_BIN exec frr sysctl -w net.ipv6.conf.all.forwarding=1
-    # Enable keep_addr_on_down to preserve IPv6 addresses during VRF enslavement.
-    # Without this, IPv6 global addresses are removed when interfaces are moved to a VRF,
-    # causing FRR/zebra to fail creating FIB nexthop groups ("no fib nhg" bug).
-    # See: https://docs.kernel.org/networking/vrf.html (section 4: Enslave L3 interfaces)
-    #      https://github.com/FRRouting/frr/issues/1666
-    $OCI_BIN exec frr sysctl -w net.ipv6.conf.all.keep_addr_on_down=1
-  fi
-
-  if [ "$ENABLE_EVPN" == true ]; then
-    echo "Configuring global EVPN BGP on external FRR (advertise-all-vni + neighbor activation)..."
-    # Enable l2vpn evpn address-family, activate all neighbors, and advertise-all-vni.
-    # Neighbors are already configured by demo.sh; extract them from the running config.
-    # This is cluster-level infrastructure shared across all EVPN tests; configured once
-    # at install time so individual tests don't need to manage it.
-    # Wait for FRR daemons to be ready ("Not all daemons are up, cannot write config").
-    local attempts=0 daemon_status
-    while ! daemon_status=$($OCI_BIN exec frr vtysh -c "show daemons" 2>&1); do
-      if (( ++attempts > 30 )); then
-        echo "error: FRR daemons did not become ready after 30 attempts"
-        echo "last daemon status: $daemon_status"
-        exit 1
-      fi
-      sleep 1
-    done
-    local bgp_neighbors vtysh_cmds
-    bgp_neighbors=$($OCI_BIN exec frr vtysh -c "show running-config" | grep "^ neighbor.*remote-as" | awk '{print $2}')
-    vtysh_cmds=(-c "configure terminal" -c "router bgp 64512" -c "address-family l2vpn evpn")
-    for neighbor in $bgp_neighbors; do
-      vtysh_cmds+=(-c "neighbor $neighbor activate")
-      vtysh_cmds+=(-c "neighbor $neighbor route-reflector-client")
-    done
-    vtysh_cmds+=(-c "advertise-all-vni" -c "exit-address-family" -c "end" -c "write memory")
-    $OCI_BIN exec frr vtysh "${vtysh_cmds[@]}"
-    echo "Global EVPN BGP config complete on external FRR"
-  fi
-}
-
-deploy_bgp_external_server() {
-  # We create an external docker container that acts as the server (or client) outside the cluster
-  # in the e2e tests that levergae router advertisements.
-  # This container will be connected to the frr container deployed above to simulate a realistic
-  # network topology
-  # -----------------               ------------------                         ---------------------
-  # |               | 172.26.0.0/16 |                |       172.18.0.0/16     | ovn-control-plane |
-  # |   external    |<------------- |   FRR router   |<------ KIND cluster --  ---------------------
-  # |    server     |               |                |                         |    ovn-worker     |   (client pod advertised
-  # -----------------               ------------------                         ---------------------    using RouteAdvertisements
-  #                                                                            |    ovn-worker2    |    from default pod network)
-  #                                                                            ---------------------
-  local ip_family ipv6_network
-  if [ "$PLATFORM_IPV4_SUPPORT" == true ] && [ "$PLATFORM_IPV6_SUPPORT" == true ]; then
-    ip_family="dual"
-    ipv6_network="--ipv6 --subnet=${BGP_SERVER_NET_SUBNET_IPV6}"
-  elif  [ "$PLATFORM_IPV6_SUPPORT" == true ]; then
-    ip_family="ipv6"
-    ipv6_network="--ipv6 --subnet=${BGP_SERVER_NET_SUBNET_IPV6}"
-  else
-    ip_family="ipv4"
-    ipv6_network=""
-  fi
-  $OCI_BIN rm -f bgpserver
-  $OCI_BIN network rm -f bgpnet
-  $OCI_BIN network create --subnet="${BGP_SERVER_NET_SUBNET_IPV4}" ${ipv6_network} --driver bridge bgpnet
-  $OCI_BIN network connect bgpnet frr
-  $OCI_BIN run  --cap-add NET_ADMIN --user 0  -d --network bgpnet  --rm  --name bgpserver -p "${BGP_SERVER_HOST_PORT}:8080"  registry.k8s.io/e2e-test-images/agnhost:2.45 netexec
-  # let's make the bgp external server have its default route towards FRR router so that we don't need to add routes during tests back to the pods in the
-  # cluster for return traffic
-  local bgp_network_frr_v4 bgp_network_frr_v6 kind_network_frr_v4 kind_network_frr_v6
-  bgp_network_frr_v4=$($OCI_BIN inspect -f '{{.NetworkSettings.Networks.bgpnet.IPAddress}}' frr)
-  echo "FRR bgp network IPv4: ${bgp_network_frr_v4}"
-  $OCI_BIN exec bgpserver ip route replace default via "$bgp_network_frr_v4"
-  if  [ "$PLATFORM_IPV6_SUPPORT" == true ] ; then
-    bgp_network_frr_v6=$($OCI_BIN inspect -f '{{.NetworkSettings.Networks.bgpnet.GlobalIPv6Address}}' frr)
-    echo "FRR bgp network IPv6: ${bgp_network_frr_v6}"
-    $OCI_BIN exec bgpserver ip -6 route replace default via "$bgp_network_frr_v6"
-  fi
-  if [ "$ADVERTISED_UDN_ISOLATION_MODE" == "loose" ]; then
-    kind_network_frr_v4=$($OCI_BIN inspect -f '{{.NetworkSettings.Networks.kind.IPAddress}}' frr)
-    echo "FRR kind network IPv4: ${kind_network_frr_v4}"
-    # If UDN isolation is in loose disabled, we need to set the default gateway for the nodes in the cluster
-    # to the FRR router so that cross-UDN traffic can be routed back to the pods in the cluster in the loose mode.
-    echo "Setting default gateway for nodes in the cluster to FRR router IPv4: ${kind_network_frr_v4}"
-    set_nodes_default_gw "$kind_network_frr_v4"
-    if  [ "$PLATFORM_IPV6_SUPPORT" == true ] ; then
-      kind_network_frr_v6=$($OCI_BIN inspect -f '{{.NetworkSettings.Networks.kind.GlobalIPv6Address}}' frr)
-      echo "FRR kind network IPv6: ${kind_network_frr_v6}"
-      set_nodes_default_gw "$kind_network_frr_v6"
-    fi
-  else
-    # disable the default route to make sure the container only routes accross
-    # directly connected or learnt networks (doing this at the very end since
-    # docker changes the routing table when a new network is connected)
-    $OCI_BIN exec frr ip route delete default
-    $OCI_BIN exec frr ip route
-    $OCI_BIN exec frr ip -6 route delete default
-    $OCI_BIN exec frr ip -6 route
-  fi
-}
-
-set_nodes_default_gw() {
-  local gw="$1"
-  local ip_cmd="ip"
-  local route_cmd="route replace default via"
-
-  # Check if $gw is IPv6 (contains ':')
-  if [[ "$gw" == *:* ]]; then
-    ip_cmd="ip -6"
-  fi
-
-  KIND_NODES=$(kind_get_nodes)
-  for node in $KIND_NODES; do
-    $OCI_BIN exec "$node" $ip_cmd $route_cmd "$gw"
-  done
-}
 
 destroy_bgp() {
   if $OCI_BIN ps --format '{{.Names}}' | grep -Eq '^bgpserver$'; then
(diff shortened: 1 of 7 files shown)


jechen0648:
/assign @jcaamano  
Please help review the PR, thanks!

jcaamano:
Thanks for the effort! Some general comments before digging in deep:

- Has this PR been assisted by AI? If so, I think it would be good to specify it in the PR description. See `Assisted-By` example in this [PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5641)
- Commits have no commit messages, please add them. See [commit guidelines](https://ovn-kubernetes.io/governance/CONTRIBUTING/#commit-message-guidelines).
- There are a lot of changes that seem completely unrelated and are also not explained anywhere. Please make sure those changes are on their own commits with appropriate commit messages explaining their reason to be and the relationship with the main purpose of the PR. For example, why would adding proper security context for pods be related to this PR?
- Don't just refer to "coderabbitai comments" in your descriptions, we are not sure what comments you are referring to. Describe what you are actually fixing.
- Why can't we reuse existing templates in `test/e2e/testdata/routeadvertisements` instead of defining new ones?

jechen0648:
> Thanks for the effort! Some general comments before digging in deep:
> 
> * Has this PR been assisted by AI? If so, I think it would be good to specify it in the PR description. See `Assisted-By` example in this [PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5641)
> * Commits have no commit messages, please add them. See [commit guidelines](https://ovn-kubernetes.io/governance/CONTRIBUTING/#commit-message-guidelines).
> * There are a lot of changes that seem completely unrelated and are also not explained anywhere. Please make sure those changes are on their own commits with appropriate commit messages explaining their reason to be and the relationship with the main purpose of the PR. For example, why would adding proper security context for pods be related to this PR?
> * Don't just refer to "coderabbitai comments" in your descriptions, we are not sure what comments you are referring to. Describe what you are actually fixing.
> * Why can't we reuse existing templates in `test/e2e/testdata/routeadvertisements` instead of defining new ones?

Thanks for these general comments:
1. added Assisted by: claude-4.5-opus-high
2. all commits have commit messages, the third commit can have some more information, just added more details to third commit's message
3.  It is true that the second and forth commits were fixes to failures in CI , they are not related to this PR, I just felt like to see all CI lanes pass :-).   
4. security context for pods was added in order to avoid warning messages being thrown out during test execution, also in order to save test case execution time, as time was wasted on checking security and printing warning messages 
5.  I will change to use existing templates in test/e2e/testdata/routeadvertisements tomorrow

Thanks!

test/scripts/install-kind.sh
```diff
@@ -10,9 +10,25 @@ esac
...
+    fi
+    echo "Failed to fetch KIND checksum (attempt ${retry}/${KIND_DOWNLOAD_RETRIES}), retrying..."
+    sleep 5
+done
+
+if [ -z "${KIND_SHA}" ]; then
+    echo "ERROR: Could not fetch KIND SHA256 checksum from ${KIND_SHA_URL}"
+    exit 1
+fi
+
```
jcaamano: Just add this to the retry loop within `install_kind`. 

Also what is the difference between having a loop, and the usage of `--retry` and `--retry-delay` options? It seems like we would use the former or the latter, but not both.

.github/workflows/test.yml
```diff
@@ -479,7 +479,7 @@ jobs:
...
           - {"target": "external-gateway",  "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv6",      "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "ipv4",      "disable-snat-multiple-gws": "SnatGW",   "second-bridge": "1br", "ic": "ic-disabled", "num-workers": "3", "network-segmentation": "enable-network-segmentation"}
-          - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "num-workers": "3", "network-segmentation": "enable-network-segmentation", "routeadvertisements": "true"}
+          - {"target": "kv-live-migration", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "num-workers": "3", "network-segmentation": "enable-network-segmentation", "routeadvertisements": "advertise-default"}
```
jcaamano: How is this change related to the commit that contains it?

test/e2e/cmd/bgp-setup/main.go
```diff
@@ -1082,26 +1083,16 @@ func addPodNetworkRoutes(cfg *Config, clientset *kubernetes.Clientset) error {
...
-	parts := strings.Split(arrStr, ",")
-	for _, part := range parts {
-		part = strings.TrimSpace(part)
-		part = strings.Trim(part, "\"")
-		if part != "" {
-			subnets = append(subnets, part)
-		}
+	// Collect all subnets from all networks
+	for _, nets := range networkSubnets {
+		subnets = append(subnets, nets...)
```
jcaamano, line 1229: ~~I think this was done the way it was as a performance optimization. Since you are not explaining why you are making this change, I would just leave it as it was.~~

contrib/kind-common
```diff
@@ -350,12 +350,37 @@ kubectl_wait_pods() {
...
+    fi
+    
+    if [ $attempt -lt $max_retries ]; then
+      echo "Some pods not ready yet, waiting ${retry_delay}s before retry..."
+      kubectl get pods -A -o wide || true
+      sleep ${retry_delay}
+    fi
+    
+    attempt=$((attempt + 1))
+  done
```
jcaamano: How is adding a retry loop any different that just increasing `timeout`, what problem are you fixing here?

test/e2e/kubevirt.go
```diff
@@ -1841,6 +1841,8 @@ write_files:
 								},
 							},
 						}},
+						NodeSelector:             metav1.LabelSelector{},
+						FRRConfigurationSelector: metav1.LabelSelector{},
```
jcaamano: What are you fixing here? The commit message is not providing enough detail.

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -721,27 +720,18 @@ func getNetworkInterface(containerName, networkName string) (api.NetworkInterfac
...
 	getInterfaceNameUsingIP := func(ip string) (string, error) {
-		ipFlag, err := getIPFamilyFlagForIPRoute2(ip)
-		if err != nil {
-			return "", fmt.Errorf("failed to get IP family flag for %s: %w", ip, err)
-		}
-		allInfAddrBytes, err := exec.Command(containerengine.Get().String(), "exec", "-i", containerName, "ip", "-br", ipFlag, "a", "sh").CombinedOutput()
+		// Use 'ip -o addr show' instead of 'ip -br addr show' because -br flag is not supported by BusyBox
+		// which is used by some container images like cloudflare/goflow.
+		// The -o (oneline) flag is supported by both BusyBox and full iproute2.
+		allInfAddrBytes, err := exec.Command(containerengine.Get().String(), "exec", "-i", containerName, "ip", "-o", "addr", "show").CombinedOutput()
```
jcaamano: How is this failing for us? The commit message is not providing enough detail.

test/e2e/util.go
```diff
@@ -1219,6 +1219,7 @@ func wrappedTestFramework(basename string) *framework.Framework {
 func newPrivelegedTestFramework(basename string) *framework.Framework {
 	f := framework.NewDefaultFramework(basename)
 	f.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
+	f.NamespacePodSecurityWarnLevel = admissionapi.LevelPrivileged
```
jcaamano: Why add this? The commit message is not providing enough detail.

jcaamano (commented):
> added Assisted by: claude-4.5-opus-high

In a PR with so many unrelated changes, I would probably do this per commit, adding the tag to the appropriate commit messages. I am not actually sure if some of you fixes nto related to the main purpose of the PR are also AI driven or not.

> all commits have commit messages, the third commit can have some more information, just added more details to third commit's message

Sorry, I guess I was actually referring to the contents of the commit message. The formatting of your messages are unconventional and for practical purposes incorrect. From [git commit doc](https://git-scm.com/docs/git-commit#_discussion):
```
Though not required, it’s a good idea to begin the commit message with a single 
short (no more than 50 characters) line summarizing the change, followed by a blank 
line and then a more thorough description. The text up to the first blank line in a 
commit message is treated as the commit title, and that title is used throughout Git. 
For example, [git-format-patch[1]](https://git-scm.com/docs/git-format-patch) turns 
a commit into email, and it uses the title on the Subject line and the rest of the 
commit in the body.
```
For example
```
1. fix to a case in kubevirt.go that failed in CI
2. fix to comment by coderabbitai regarding security consideration: world readable kubeconfig
3. fix to a typo in README.md
```
The first line `1. fix to a case in kubevirt.go that failed in CI` is not a summary of the changes on the commit. Also, if you are doing an enumeration of seemingly unrelated things, it looks like those three changes should be split into three commits.

> security context for pods was added in order to avoid warning messages being thrown out during test execution, also in order to save test case execution time, as time was wasted on checking security and printing warning messages

Split that into its own commit. Also, you are only doing this for BGP tests cases, but I guess all the test cases might be suffering form the same issue, right?

> It is true that the second and forth commits were fixes to failures in CI , they are not related to this PR, I just felt like to see all CI lanes pass :-).

Please review the contents of commits 3, 4 and 6. Specially for those commits, I can't make sense of how their description relate with their contents, everything is a bit de-organized.

> I will change to use existing templates in test/e2e/testdata/routeadvertisements tomorrow

The intention was to reuse the code as well. Seems there is a lot of overlap of code in
`test/e2e/cmd/bgp-setup/main.go` and `test/e2e/route_advertisements.go`; `deployFRRExternalContainer`,  `generateFRRConfigFiles`, `deployBGPExternalServer` vs `generateFRRConfiguration`, `generateFRRk8sConfiguration`, `runBGPNetworkAndServer`. Can we refactor this to share as much code as possible?

test/e2e/cmd/bgp-setup/main.go
jcaamano: Disregard, I didn't notice this was new code. But why add an extra commit instead of doing it directly in the commit where you introduced the change? It's like you an AI are acting as different personas here :P

test/e2e/kubevirt.go
jechen0648: I have faded memory on this, I remember it failed in a CI lane, and the failure message prompted to this fix.   I need to do better job to document all the reasons of my changes

test/e2e/cmd/bgp-setup/main.go
jechen0648: This is change by commit message "change to use JSON parsing in main.go instead of string manuplation",  it was suggested by coderabbitai https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821#pullrequestreview-3604391828

test/e2e/util.go
jechen0648: this was added to eliminate security warning during test execution, I will look into how to split commits.  maybe it is better to start a new PR to organize better

test/e2e/cmd/bgp-setup/main.go
jcaamano: So the workflow would be:
1. You push the PR
2. You review and fix all coderabbitai comments on their original commits
3. Then a human reviews

Meaning, there is no need to keep coderabbitai fixes on a separate commit before a human was involved. 

Once a human reviewer is involved, then there is two different workflows:
A. Do rework in original commits, avoid mixing rework with rebases
B. There are occasions when the PR is big or complicated or going to be long lived or multiple conflicting rebases need to be involved, or maybe because author or reviewer need clarity or prefer that style of work; when using separate commits for the rework and squashing them last minute before merging might be preferred.

contrib/kind.sh
```diff
@@ -22,6 +22,56 @@ delete() {
   timeout 5 kubectl --kubeconfig "${KUBECONFIG}" delete namespace ovn-kubernetes || true
   sleep 5
   kind delete cluster --name "${KIND_CLUSTER_NAME:-ovn}"
+
+  # Clean up stale endpoints from kind network that kind delete may leave behind
```
tssurya: what endpoints?

contrib/kind.sh
```diff
@@ -22,6 +22,56 @@ delete() {
...
+  if $OCI_BIN network ls --format '{{.Name}}' | grep -q '^kind$'; then
+    echo "Cleaning up stale endpoints from kind network..."
+    for endpoint in ovn-control-plane ovn-worker ovn-worker2 ovn-worker3 ovn-worker4 frr; do
```
tssurya: how does hardcoding values of the node names like this even work? that's not gonna work across all lanes right?

contrib/kind.sh
```diff
@@ -22,6 +22,56 @@ delete() {
...
+      $OCI_BIN network disconnect -f kind "$endpoint" 2>/dev/null || true
+    done
+  fi
+}
+
+# Build and run the bgp-setup tool from test/e2e/cmd/bgp-setup
+# This replaces the shell functions: deploy_frr_external_container, deploy_bgp_external_server, install_frr_k8s
```
tssurya: such comments from AI are not helpful, given you have already deleted these 3 utils from the kind.sh file already - when this comment lands in repo what would it even mean? just becomes stale..
I would strongly advise you to review all code and specially comments generated by AI

contrib/kind.sh
```diff
@@ -22,6 +22,56 @@ delete() {
...
+# Arguments:
+#   $1 - phase: "deploy-frr", "deploy-bgp-server", "deploy-containers", "install-frr-k8s", or "all"
+run_bgp_setup() {
+  local phase="${1:-all}"
+  local bgp_setup_bin="${DIR}/../test/e2e/bgp-setup"
+  
+  # Build bgp-setup if it doesn't exist or if source is newer
+  if [ ! -f "$bgp_setup_bin" ] || [ "${DIR}/../test/e2e/cmd/bgp-setup/main.go" -nt "$bgp_setup_bin" ]; then
+    echo "Building bgp-setup tool..."
+    pushd "${DIR}/../test/e2e" > /dev/null
```
tssurya: do we have plans to add more such setup binaries, example for EVPN ?
cause now its placed under test/e2e/cmd/bgp-setup? so maybe having _output/bin/bgp-setup is better to store the binary than just leaving it under test/e2e with other e2e.go files..

contrib/kind.sh
```diff
@@ -22,6 +22,56 @@ delete() {
...
+  local ipv4_flag="true"
+  local ipv6_flag="false"
+  if [ "$PLATFORM_IPV4_SUPPORT" == true ]; then
+    ipv4_flag="true"
+  else
+    ipv4_flag="false"
+  fi
+  if [ "$PLATFORM_IPV6_SUPPORT" == true ]; then
+    ipv6_flag="true"
+  fi
```
tssurya: just replace with
```
  local ipv4_flag="${PLATFORM_IPV4_SUPPORT:-true}"
  local ipv6_flag="${PLATFORM_IPV6_SUPPORT:-false}"
```
?

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,210 @@
+# BGP Setup Tool
+
+A standalone Go program that sets up BGP infrastructure for route advertisement testing in OVN-Kubernetes KinD clusters.
+
+## Overview
+
+This tool implements the functionality of the following shell functions from `contrib/kind-common`:
```
tssurya: but, by the time this code lands those shell functions are totally gone right? what meaning will this README.md then make? did AI write this?
instead rephrase to write the actual steps

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,210 @@
...
+- `install_frr_k8s()` - Installs frr-k8s operator and configurations
+- Creates FRRConfiguration for the default network
+
+This tool reuses packages from the e2e test framework:
+- `containerengine` - for container runtime abstraction
+- `images` - for container images
+
+## Network Topology
+
+```
```
tssurya: I think the comment above the diagram and some of the details in the diagram that was present isn't being reflected here.

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+Run with default settings (IPv4 only, strict isolation):
+```bash
+./bgp-setup
+```
+
+### With IPv6 Support
+
+```bash
+./bgp-setup --ipv4 --ipv6
```
tssurya, line 82: what if its singlestack IPV6?

.github/workflows/test.yml
tssurya: not sure why we are changing the CI lane to advertise-default here? vm live migration lane has nothing to do with default network BGP advertised right?

contrib/kind.sh
```diff
@@ -22,6 +22,56 @@ delete() {
...
+  "$bgp_setup_bin" \
+    --phase="$phase" \
+    --container-runtime="$OCI_BIN" \
+    --ipv4="$ipv4_flag" \
+    --ipv6="$ipv6_flag" \
+    --bgp-server-subnet-ipv4="${BGP_SERVER_NET_SUBNET_IPV4}" \
+    --bgp-server-subnet-ipv6="${BGP_SERVER_NET_SUBNET_IPV6}" \
+    --isolation-mode="${ADVERTISED_UDN_ISOLATION_MODE:-strict}" \
+    --advertise-default-network="${ADVERTISE_DEFAULT_NETWORK:-true}" \
+    --kubeconfig="${KUBECONFIG}"
```
tssurya: the flags used here seem much lesser than that's written in bgp-setup/README.md. Why so?

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,210 @@
...
+| `--bgp-server-subnet-ipv6` | `fc00:f853:ccd:e796::/64` | IPv6 CIDR for BGP server network |
+| `--ipv4` | `true` | Enable IPv4 support |
+| `--ipv6` | `false` | Enable IPv6 support |
+| `--frr-k8s-version` | `v0.0.21` | Version of frr-k8s to use |
+| `--network-name` | `default` | Name for the BGP network |
+| `--kubeconfig` | `~/.kube/config` | Path to kubeconfig file |
+| `--advertise-default-network` | `true` | Advertise the default network |
+| `--isolation-mode` | `strict` | UDN isolation mode: strict or loose |
+| `--cleanup` | `false` | Only cleanup existing BGP infrastructure |
+| `--use-direct-api` | `false` | Use direct API server address (control plane container IP) instead of kubeconfig server. Only works when Docker bridge network is routable from host. |
```
tssurya, line 134: are we really using all these parameters? or do we really need all these flags?

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+```bash
+./bgp-setup --ipv4 --ipv6
+```
+
+### Custom Network Configuration
+
+```bash
+./bgp-setup \
+  --bgp-peer-subnet-ipv4 172.36.0.0/16 \
```
tssurya: what's this peer subnet range? i didn't see it in the old code.. not sure if I missed it

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,210 @@
...
+The `run_bgp_setup` function in kind.sh automatically builds the tool if needed
+and passes the appropriate configuration flags.
+
+## What It Does
+
+1. **Apply FRR-K8s CRDs** - Installs the FRRConfiguration CRD directly from GitHub (no git clone required)
+2. **Deploy FRR container** - Creates an FRR router container attached to the kind network
+3. **Create bgpnet network** - Creates a separate network for the BGP server
+4. **Deploy bgpserver** - Creates an agnhost container acting as an external server
+5. **Configure routing** - Sets up routing between containers
```
tssurya: between what containers?

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,210 @@
...
+6. **Install frr-k8s** - Deploys the frr-k8s operator directly from GitHub
+7. **Apply FRRConfiguration** - Creates BGP peering configuration for the cluster
+8. **Add pod network routes** - Optionally adds routes for pod networks (requires sudo)
```
tssurya: pod routes? could you please remind me which ones these are?

test/e2e/infraprovider/frr/frr.go
```diff
@@ -0,0 +1,229 @@
...
+type Router struct {
+	VRF           string
+	NeighborsIPv4 []string
+	NeighborsIPv6 []string
+	NetworksIPv4  []string
+	NetworksIPv6  []string
+}
+
+// Config contains the full FRR configuration data for templates
+type Config struct {
```
tssurya: so is it fair to rename this as FRRKubernetesConfig struct?

test/e2e/infraprovider/frr/frr.go
```diff
@@ -0,0 +1,229 @@
+// Package frr provides shared FRR (Free Range Routing) configuration utilities
+// for both e2e tests and the bgp-setup tool.
```
tssurya, line 5: well, where is it used from bgp-setup tool today? I don't think I see anything from this used in bgp-setup tool.

test/e2e/util.go
tssurya: yes a smaller separate PR for this is better, please don't couple the warning fixing with the BGP refactor work - your&AI's first commit is already hard to understand when I try to tie the pieces together, so anything unrelated please put that into a smaller PR to be fixed,
I assume these warnings happening are not related to the changes and even exist today correct?
It's good to fix them but better to do it as individual PR and open an u/s issue tracker for this.

test/e2e/route_advertisements.go
```diff
@@ -105,6 +104,19 @@ var _ = ginkgo.Describe("BGP: When default podNetwork is advertised", feature.Ro
 						"infinity",
 					}
 				}
+				// Add required security context to comply with PodSecurity "restricted" policy
```
tssurya: why is this in this commit? and why is this needed now and not before?

test/e2e/route_advertisements.go
```diff
@@ -988,6 +1031,21 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
 			pod := e2epod.NewAgnhostPod(udnNamespaceA.Name, fmt.Sprintf("pod-1-%s-net-%s", nodes.Items[0].Name, cudnA.Name), nil, nil, []corev1.ContainerPort{{ContainerPort: 8080}}, "netexec")
 			pod.Spec.NodeName = nodes.Items[0].Name
 			pod.Labels = map[string]string{"network": cudnA.Name}
+			// Add required security context to comply with PodSecurity "restricted" policy
```
tssurya: can this be a PR of its own.. is this change related to the refactor changes you are making - I see this being done for all BGP tests

contrib/kind-common.sh
```diff
@@ -318,15 +318,15 @@ delete_metallb_dir() {
   rm -rf "${METALLB_DIR}"
 }
 
-# kubectl_wait_pods will set a total timeout of 300s for IPv4 and 480s for IPv6. It will first wait for all
+# kubectl_wait_pods will set a total timeout of 360s for IPv4 and 540s for IPv6. It will first wait for all
```
tssurya: do you know why we need this extra now when it was not needed before? Is all the code we are adding for setup taking longer?  so are we adding an extra minute with this PR for a setup?

tssurya (changes requested):
Initial set of comments, (I haven't started main.go yet, I only reviewed the README, started reviewing infraprovide/frr and then the kind.sh bits so far)

@jechen0648 could you please update your PR description and commit message with a very detailed module hierarchy/dependency explanation between:

1. infraprovider/frr
2. cmd/bgp-setup/main.go
3. e2e/routeadvertisement.go

example, what's order of execution and level of code is shared? how is the workflow executed for the cluster setup and then also for the e2e setups

Additionally could you please do a self-review of your own PR - i.e the bits assisted by AI and see what you learn? - i.e drop a set of review comments for yourself for the code that AI assisted you with by going through it line by line.

also it seems the goal is refactoring for "Default BGP session" not "default network advertised" cause that means we don't take CUDNs advertisement into consideration when you put it that way...
Since VRF Lite BGP sessions don't use demo.sh we don't need to consider that..

contrib/kind-common.sh
jechen0648: if you click on the 4th commit, I have description for this commit as:
Occasionally, BGP kind cluster installation would exit out after detecting out some kube system pod not in Running state,
regardless pods reached Running state after some extra time.  As a result, frr-k8s was not installed.  Increase the waiting
time to ensure frr-k8s gets installed after all system pods reach Running state

no, I am just increasing total timeout timer, it does not mean it would use up all 360s/540s, it is the maximum time it would wait

contrib/kind.sh
jechen0648: bgp-setup is specifically for BGP testing setup, it is not a production binary.  Keeping it in test/e2e makes more sense to me.
Would it be better to organize it into a new dedicated directory,  something like test/e2e/_output/bin?  EVPN related setup can be added here?

contrib/kind.sh
jechen0648: removed in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

contrib/kind-common.sh
jechen0648: just want to add more explanation: BGP kind cluster installation being exited out prematurely due to kube system pod not reaching Running state happened to me multiple times when I built one locally.

contrib/kind.sh
tssurya: that's what I meant - test/e2e/_output/bin/bgp-setup
I wasn't meaning to put it outside test/e2e - it belongs inside

contrib/kind.sh
jechen0648: fixed in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: this refers to func addPodNetworkRoutes() that routes are added to allow host to reach pod network inside of cluster

.github/workflows/test.yml
jechen0648: reverted in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

contrib/kind.sh
jechen0648: containers still attached to the Docker kind network that need to be disconnected during cleanup, added another comment line next to explain in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

contrib/kind.sh
jechen0648: changed in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

contrib/kind.sh
jechen0648: fixed in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: you are right, updated it in commit:https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: updated in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: fixed in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: not needed, cleaned up in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: did clean up in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397, thanks!

test/e2e/cmd/bgp-setup/README.md
jechen0648: added more details in commit:https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/cmd/bgp-setup/README.md
jechen0648: added details for this line in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/a90d66fb54c6bcc7b0115184fc6e9239b4679397

test/e2e/infraprovider/frr/frr.go
jechen0648: updated in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/c9711c4f56172b3152ceb1e179fe2afa88a0ee0c to make sure frr under infraprovider is used in bgp-setup tool

contrib/kind.sh
jechen0648: placed the setup binaries in test/e2e/_output/bin/bgp-setup in commit: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/c9711c4f56172b3152ceb1e179fe2afa88a0ee0c

test/e2e/infraprovider/frr/frr.go
jechen0648: yes, changed in commit:https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5821/commits/c9711c4f56172b3152ceb1e179fe2afa88a0ee0c

test/e2e/route_advertisements.go
jechen0648: without proper pod security, warning message would be printed 3 times per pod, no one cared about it before, that is why it was not there before.  I had wanted to fix them when I was doing https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5461, but I was told to ignore it.  Now I would prefer eliminating warning messages to save test execution time.

test/e2e/route_advertisements.go
jechen0648: Same reason as above.  This is added to reduce test wasted on printing warning messages, thus improve overall BGP test execution time

test/e2e/route_advertisements.go
tssurya: do it as a separate PR @jechen0648 and drop a link here and I can review that separately. - given it already exists and has nothing to do with this PR I would prefer another PR for fixing all warning related issues.

test/e2e/route_advertisements.go
tssurya: do it as a separate PR @jechen0648 and drop a link here and I can review that separately.

test/e2e/infraprovider/frr/frr.go
tssurya: My question still stands - why was this not considered in PR design? And why was there a comment in the code that was not truthfully speaking about how the code was actually done.
this ties in with my overall comment - I would like for you to self-review your own PR first - the bits written by AI :) let's see what nice things you find

contrib/kind-common.sh
tssurya: are we hitting that issue in CI? can you give me a job link to when it happened in CI? if so next step is understanding the RCA - bumping the timer is not the correct fix IMHO specially when we don't know "why" a problem is happening. i.e
a problem is happening that's causing delay -> the fix here seems to be - let's wait longer..
what I want to see is:
a problem is happening that's causing delay -> what's the reason for that ? is this being hit in CI? are others seeing this outside of this PR? what's the RCA ? how do we fix the problem. keep the same timer but investigate why things are taking longer to get into running state.

test/e2e/cmd/bgp-setup/README.md
tssurya: @jechen0648 : same q here and in all comments - while I really appreciate you fixing the issues I am finding as I review I want to ask a broader question - did AI write this or did you? did you proof read this PR? if not could you please do that so that its easier for the reviewer?
PS: I am a fan of AI-driven development but I am not a fan of using AI as the driver - its an assistant so the author needs to be vetting every line on their own PR if its AI-assisted - otherwise reviewers are going to be over-burdened. Specially the overall design of the PR and README's  - that's totally the author's responsibility.

contrib/kind-common.sh
jechen0648: kube system pods failing to reach Running state happened occasionally to me several times when I built BGP KIND cluster to test my work LOCALLY, I did not see it fail in CI.

test/e2e/route_advertisements.go
jechen0648: created a separate PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5862, cleaned up pod security stuff from this PR.

test/e2e/route_advertisements.go
jechen0648: created a separate PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5862, cleaned up pod security stuff from this PR.

contrib/kind-common.sh
tssurya: > kube system pods failing to reach Running state happened occasionally to me several times when I built BGP KIND cluster to test my work LOCALLY, I did not see it fail in CI.

let's remove this commit since we don't seem to know the "why are kube system pods failing to reach running state" and it seems to only happen in your local env.

tssurya:
@jechen0648 : I see you are still pushing some changes, is this PR now ready for another round?

jechen0648:
> @jechen0648 : I see you are still pushing some changes, is this PR now ready for another round?

yes, I was just about to tag you that it is ready for review now, thank you Surya  @tssurya

jechen0648:
@tssurya   I resolved newer conflicts, it is ready for review again, it has been over 2 months since the PR was submitted, it would be great if we can work toward its merge soon, thank you very much!

tssurya:
Is this PR ready for reviews? Seems like EVPN Lane is failing pretty bad: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/24052074296/job/70150934999?pr=5821

test/e2e/infraprovider/providers/kind/kind.go
jechen0648: no longer make any change to this test/e2e/infraprovider/providers/kind/kind.go file

test/e2e/infraprovider/frr/frr.go
```diff
@@ -0,0 +1,229 @@
...
+	"os"
+	"path"
+	"path/filepath"
+	"text/template"
+
+	utilnet "k8s.io/utils/net"
+)
+
+// Image is the FRR container image used for BGP testing
+const Image = "quay.io/frrouting/frr:10.4.1"
```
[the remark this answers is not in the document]
jechen0648: no longer using hardcoded frr image, using ${FRR_DEPLOYED_IMAGE}

jechen0648:
> Initial set of comments, (I haven't started main.go yet, I only reviewed the README, started reviewing infraprovide/frr and then the kind.sh bits so far)
> 
> @jechen0648 could you please update your PR description and commit message with a very detailed module hierarchy/dependency explanation between:
> 
> 1. infraprovider/frr
> 2. cmd/bgp-setup/main.go
> 3. e2e/routeadvertisement.go
> 
> example, what's order of execution and level of code is shared? how is the workflow executed for the cluster setup and then also for the e2e setups
> 
> Additionally could you please do a self-review of your own PR - i.e the bits assisted by AI and see what you learn? - i.e drop a set of review comments for yourself for the code that AI assisted you with by going through it line by line.
> 
> also it seems the goal is refactoring for "Default BGP session" not "default network advertised" cause that means we don't take CUDNs advertisement into consideration when you put it that way... Since VRF Lite BGP sessions don't use demo.sh we don't need to consider that..

added module hierarchy/dependency explanation in PR description section

Default BGP session is the session we advertise default network, it does not include CUDNs advertisement or VRF-Lite BGP

test/e2e/infraprovider/frr/frr.go
jechen0648: In the earlier version of this PR, bgp-setup had its own inline template rendering and did not import infraprovider/frr at all. That has been fixed as part of later version of this refactor: bgp-setup now delegates all FRR/frr-k8s config generation to infraprovider/frr.  Similarly, route_advertisements.go uses frr.NewEmbeddedSource, frr.GenerateFRRDaemonConfig, and frr.GenerateFRRK8sConfig from the same package to render FRR daemon configs and FRRConfiguration CRDs during e2e test setup — the two consumers share the same template files and config generation logic.   So now the comment line accurately reflects how the package is used.

contrib/kind-common.sh
jechen0648: reverted

test/e2e/cmd/bgp-setup/README.md
```diff
@@ -0,0 +1,253 @@
...
+
+1. **Apply FRR-K8s CRDs** - Installs the FRRConfiguration CRD directly from GitHub (no git clone required)
+2. **Deploy FRR container** - Creates an FRR router container attached to the kind network
+3. **Create bgpnet network** - Creates a separate network for the BGP server
+4. **Deploy bgpserver** - Creates an agnhost container acting as an external server
+5. **Configure routing** - Sets up routing between bgpserver, FRR router, and cluster nodes
+6. **Install frr-k8s** - Deploys the frr-k8s operator directly from GitHub
+7. **Apply FRRConfiguration** - Creates BGP peering configuration for the cluster when route advertisements are enabled
+8. **Add pod network routes** through func addPodNetworkRoutes() - Optionally adds routes for pod networks when `--advertise-default-network=true` (requires sudo)
+
```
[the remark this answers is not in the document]
jechen0648: @coderabbitai this has been resolved

jechen0648:
[kind-dpu-no-overlay CI failure](https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/27376799218/job/80942341369?pr=5821)  is recorded in https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6536

PR #5824, opened 2025-12-24
Fix flaky egress-ip unit test case

l8huang:
Test case `Gateway EgressIP update EgressIP [It] reconfigures from an old to a new IP` occasionally failed with:
```
  [FAILED] Expected
      <[]string | len:0, cap:0>: []
  to consist of
      <[]string | len:1, cap:1>: ["192.168.1.6"]
  the missing elements were
      <[]string | len:1, cap:1>: ["192.168.1.6"]
```

This could be caused by race condition between the Kubernetes API update and the informer cache sync.

Replace the immediate assertion with `gomega.Eventually()` to poll until the informer cache reflects the updated annotation.
  

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of EgressIP configuration tests by adding polling-based assertions to wait for asynchronous annotation propagation before validation. This ensures stable verification of initial assignment and reconfiguration flows, reducing flakiness and making test outcomes more deterministic.

### DIFF
--- a/go-controller/pkg/node/egressip/gateway_egressip_test.go
+++ b/go-controller/pkg/node/egressip/gateway_egressip_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


go-controller/pkg/node/egressip/gateway_egressip_test.go
```diff
@@ -198,9 +199,11 @@ var _ = ginkgo.Describe("Gateway EgressIP", func() {
 			isUpdated, err = addrMgr.UpdateEgressIP(assignedEIP1, assignedEIP2)
 			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "should process a valid EgressIP")
 			gomega.Expect(isUpdated).Should(gomega.BeTrue())
-			node, err := addrMgr.nodeLister.Get(nodeName)
-			gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "node should be present within kapi")
-			gomega.Expect(parseEIPsFromAnnotation(node)).Should(gomega.ConsistOf(ipV4Addr2))
+			gomega.Eventually(func() []string {
+				node, err := addrMgr.nodeLister.Get(nodeName)
+				gomega.Expect(err).ShouldNot(gomega.HaveOccurred(), "node should be present within kapi")
```
pliurh, line 204: `gomega.Expect()` should not be inside the `Eventually` block. Otherwise, there won't be a retry when `err !=nil`.

l8huang: Thanks for viewing.

The purpose of this `Eventually` block is waiting for the node's annotation being updated, if a error occurred while getting the node, no need to retry it anymore.

l8huang: @pliurh I see https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5728 has the same fix, this PR will be closed.

PR #5826, opened 2025-12-28
dist: Fix fedora build

ormergi:
## 📑 Description

The development workflow is broken due to recent changes around code that build the project's fedora based images.

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/dist/images/Dockerfile.fedora
+++ b/dist/images/Dockerfile.fedora
@@ -22,7 +22,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-RUN apt-get update && apt-get install -y -qq make git
+RUN dnf install --best --refresh -y make git
 COPY ${OVN_KUBERNETES_DIR} ovn-kubernetes
 RUN cd ovn-kubernetes/dist/images && \
     CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} make bld
@@ -107,7 +107,7 @@ RUN echo "Running on $BUILDPLATFORM, building for $TARGETPLATFORM"
 
 # Install koji, don't clean dnf cache we will install extra packages at
 # Final stage
-RUN dnf install --best --refresh -y --setopt=tsflags=nodocs koji 
+RUN dnf install --best --refresh -y --setopt=tsflags=nodocs koji
 
 RUN if [ "$TARGETPLATFORM" = "linux/amd64" ] || [ -z "$TARGETPLATFORM" ] ; then koji download-build $ovnver --arch=x86_64  ; \
     else koji download-build $ovnver --arch=aarch64 ; fi


tssurya:
cc @killianmuldoon

dist/images/Dockerfile.fedora
```diff
@@ -22,7 +22,8 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-RUN apt-get update && apt-get install -y -qq make git
+RUN dnf install --best --refresh -y make git && \
+    dnf clean all
```
qinqon: This is a build stage you don't need to clean it up.

dist/images/Dockerfile.fedora
```diff
@@ -107,7 +109,8 @@ RUN echo "Running on $BUILDPLATFORM, building for $TARGETPLATFORM"
 
 # Install koji, don't clean dnf cache we will install extra packages at
 # Final stage
-RUN dnf install --best --refresh -y --setopt=tsflags=nodocs koji 
+RUN dnf install --best --refresh -y --setopt=tsflags=nodocs koji && \
+    dnf clean all
```
qinqon: There is also a explicit comment that instruct not to clean this up, also the fix is just the apt-get stuff.

dist/images/Makefile
```diff
@@ -28,7 +28,7 @@ OVN_FROM := source
 OVN_GITSHA := $(shell git ls-remote "${OVN_REPO}" "${OVN_GITREF}" | sort -k2  -V  |tail -1 | awk '{ print $$1 }')
 endif
 GO_VERSION ?= 1.24
-GO_IMAGE = quay.io/lib/golang:${GO_VERSION}
+GO_IMAGE = quay.io/projectquay/golang:${GO_VERSION}
```
tssurya, line 31: the lib one was retired was it?

dist/images/Dockerfile.fedora
tssurya: yea could you also tell why the dnf clean all is done here and also above in the PR description?

dist/images/Dockerfile.fedora
ormergi: Done

dist/images/Dockerfile.fedora
ormergi: I though its good call to clean up cache for builder layers but looking at this again there is no much benefit from it.

I removed clean-all calls where not needed.

dist/images/Makefile
ormergi: I dont know if retired but in practice it doesnt work, pulling the image fails every time :/ 
This is how the issue resolved recently https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5786

ormergi:
[changes: ](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/0541bbe33cffcda233e8e4b172069e50be6fa726..d9bf37b95eab9c302918a8e152f94b39663fa476) resolved https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5826#pullrequestreview-3615057185

dist/images/Dockerfile.fedora
```diff
@@ -107,8 +107,7 @@ RUN echo "Running on $BUILDPLATFORM, building for $TARGETPLATFORM"
 
 # Install koji, don't clean dnf cache we will install extra packages at
 # Final stage
-RUN dnf install --best --refresh -y --setopt=tsflags=nodocs koji 
-
+RUN dnf install --best --refresh -y --setopt=tsflags=nodocs koji
```
qinqon, line 110: This is a noop change, keep the new line so we don't see this diff.

ormergi: I missed this one, Fixed

dist/images/Makefile
tssurya: oh I remember now, lib is actually retired.. (I had confirmed internally within RH on slack).. I remember merging https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5556 sounds like we missed a spot?

dist/images/Dockerfile.fedora
tssurya: I still see this in the diff.. but its ok will ignore the nit.

tssurya (approved):
we might need to cherry pick this into release-1.2 as well
thanks for the fix!

dist/images/Dockerfile.fedora
```diff
@@ -22,7 +22,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
-RUN apt-get update && apt-get install -y -qq make git
+RUN dnf install --best --refresh -y make git
```
tssurya, line 25: do we need to do dnf update first? (ah refresh is present so ok)

tssurya:
@qinqon are you ok with the updated PR?

dist/images/Dockerfile.fedora
ormergi: Yep, `--refresh` is equivalent to `dnf upgrade`

dist/images/Dockerfile.fedora
ormergi: There was a redundant newline I removed
The diff now is due to redundent space removed (by the IDE reformating).

qinqon:
@tssurya all good!

PR #5541, opened 2025-09-01
Implement Network Connect controller in ovnkube-controller (`PodNetwork` Connectivity)

tssurya:
Depends on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5707

## 📑 Description

See https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5246 for design

This PR adds a network-connect controller in ovn-kube controller process that
watches for CNCs and Nodes and creates per CNC the corresponding objects into NBDB:

1. connectRouter
2. Ports on connect router and network routers connecting the connect router <-> network routers
3. LRPs on each connected network router to reach other networks via connect router
4. static routes on connect router to each node-networks subnet for l3 and the whole network subnet for l2
5. This ensure pod2pod full connectivity across connected networks.

Fixes #5536

Notes for Reviewers:

- Watching for NADs is not necessary here I think because we have all the info we need based on allocateSubnets annotation from CM
- What about watch namespaces? do we really need that? cause all perms/combs of NAD stuff is already taken care of in the CM level, watching for the connectedNetworks is as simple as watching for the annotations in the CNC. - no need to watch for namespaces as well.
- ~findNetworkByID looks expensive, but we need to check how much expensive it is and see if we need to tilt towards caches.~ DONE, moved it to network manager with a small cache that makes this O(1).
- Support for connectivity field will come in next PR along with services support
- Node add/delete calls a sync on all lrps and static routes - but most of them should be idempotent in libovsdb transact. It's hard to maintain per network/node cache in that manner of calculative updates.

TODO:

- Also check what's the allocation starting? is it always from .2? what's blocking 0 and 1 IPs..
- ~Secondly, node-id changes are hitting some referential integrity issues during port delete - need to check what can be done, - need to enhance test coverage here.~ - this is DONE it was discussed in one of the review comments.
- Update commit messages to show the router, ports, policies and static routes properly
- the new e2e that has 24 steps - that takes 30mins - there is some things we can do to improve runUDNPod to not wait till pod is up and to split that out as a separate util to save time, can be a FUP for future.

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [x] My code requires tests
- [x] if so, I have added and/or updated the tests as required
- [x] All the tests have passed in the CI 

#### Create CNC

```
apiVersion: k8s.ovn.org/v1
kind: ClusterNetworkConnect
metadata:
  name: colored-enterprise
spec:
[...]

PR #5828, opened 2025-12-30
Implement Network Connect `ClusterIPServiceNetwork` Connectivity

### LINKED ISSUE #5581: Implement Services connectivity across the connected Networks
### What is this issue about?

This is about adding support for services connectivity.

tssurya:
Depends on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5541

## 📑 Description

This PR adds support for partial service connectivity for NetworkConnect. See https://ovn-kubernetes.io/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns/#services for design details.

Fixes #5581 

TBD:

1. Figure out if we really need `ct.new` match in the ACL. Bug is still reproducible, have opened https://issues.redhat.com/browse/FDP-3124/https://issues.redhat.com/browse/FDP-3131 will work with @dceara and OVN team on `pass-related`. Meanwhile we stick with the workaround. See https://cloud-native.slack.com/archives/C08452HR8V6/p1770720413427739. We decided to go the route of doing two Pass ACLs.. see the slack thread for more information. We will FUP on this piece, tracked via https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6048
2. ~Figure out how to re-design "to-self" block that's happening now - that's why CI is red, A few options:~
    ~2.1 Add a higher priority ACL to allow "to-self" subnets on each network's switch? - that's going to be an extra ACL per network - We will go with this - add a higher priority ACL at 5000 to allow src=dst=same network~ THIS IS DONE
    ~2.2 Do 2.1 globally - do we care if we have ACLs of other connected networks in the given network?~
    ~2.3 Any OVN specific hair pin options?~
3. ~The first commit which is a bug fix is not handling upgrades scenario - we probably at this point need to update the names of the load balancers at startup.~
     * ~So looks like `EnsureLBs` is handling this correctly already. It takes `existingCacheLBs` and `LBs` desired.~
     * ~`toDelete` map initializes it with all existing cached LBs.~
     * ~`toDelete[lb.UUID] = lb` -- every existing LB starts as "to delete"~
     * ~`existingLB := existingByName[lb.Name]` -- matches by name since that's the index in DB~
     * ~`delete(toDelete, existingLB.UUID)` -- only removes from toDelete if name matches~
     * ~Everything still in toDelete gets deleted~
     * ~So: old per-node LB (name `_node_router+switch`) stays in `toDelete` because new desired LB has name `_cluster` (no name match). Old LB gets deleted~
     * I'll test and confirm this as well. THIS IS DONE, I checked it. We don't need to do anything specific for this.
4. ~Add UTs for commits 3&4~ THIS IS DONE
5. ~Check with @dceara if load balancer delete will cleanup references or not in OVN? - Ensure a UT exists for this~ - See https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828#issuecomment-3818734261 , we are good here, THIS IS DONE
6. ~Anything to consider during restart scenario? cache population or sync required?~ THIS IS DONE, Added a repair function
7. ~Reuse existing code for `getNetworkOwnerKeyForNamespace` - see if we can get rid of that util~ - I didn't see any room for reuse, this is DONE
8. **This PR doesn't consider Dynamic UDNs feature - that will need another change on top once this lands.** Tracked via https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5963. A basic level of implementation consideration was done during create case but its not being elaborately fleshed out since dynamic udns landed much after CNC
9. self-note: double-check for error returns properly
10. self-note: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828#discussion_r2730888631
11. Discuss with @npinaeva and @trozet about CIP/EIP/LBVIP point - but after testing it out - perhaps now is the time to rename `ClusterIPServiceNetwork` to `ServiceNetwork` OR we stick with the same name but internal behaviour changes a bit. See https://cloud-native.slack.com/archives/C08452HR8V6/p1770802433305249. Will do this in a new PR, opened a tracker for this https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5960

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Per-CNC service LoadBalancerGroup support, service-aware reconciliation, and startup repair for stale CNC-related objects.
  * EndpointSlice mirroring with optional IP transform and new helpers for load‑balancer group ↔ switch associations.
  * New ACL priorities and naming for network-connect service flows.

* **Bug Fixes**
  * Host-endpoint detection and LB external IDs now respect per-network subnet context.
  * Network cleanup removes orphaned load balancers.

* **Tests**
  * Large expansion of unit, integration, and e2e tests for service connectivity, UDN, repair, and cleanup.

### DIFF
--- a/go-controller/pkg/clustermanager/networkconnect/controller_components_test.go
+++ b/go-controller/pkg/clustermanager/networkconnect/controller_components_test.go
@@ -769,7 +769,9 @@ func TestController_reconcileClusterNetworkConnect(t *testing.T) {
 				for _, n := range nads {
 					mutableNetInfo.AddNADs(fmt.Sprintf("%s/%s", n.Namespace, n.Name))
 				}
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks[namespace] = mutableNetInfo
+				fakeNM.Unlock()
 			}
 			for _, nad := range tt.nads {
 				nadKey := fmt.Sprintf("%s/%s", nad.Namespace, nad.Name)
@@ -1789,7 +1791,9 @@ func TestMustProcessCNCForNAD(t *testing.T) {
 				g.Expect(err).ToNot(gomega.HaveOccurred())
 				mutableNetInfo := util.NewMutableNetInfo(netInfo)
 				mutableNetInfo.SetNADs(tt.nad.Namespace + "/" + tt.nad.Name)
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks[tt.nad.Namespace] = mutableNetInfo
+				fakeNM.Unlock()
 			}
 			if tt.nad != nil {
 				nadKey := fmt.Sprintf("%s/%s", tt.nad.Namespace, tt.nad.Name)
@@ -2499,7 +2503,9 @@ func TestController_reconcileNamespace(t *testing.T) {
 				for _, n := range nads {
 					mutableNetInfo.AddNADs(fmt.Sprintf("%s/%s", n.Namespace, n.Name))
 				}
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks[namespace] = mutableNetInfo
+				fakeNM.Unlock()
 			}
 			for _, nad := range tt.nads {
 				nadKey := fmt.Sprintf("%s/%s", nad.Namespace, nad.Name)
--- a/go-controller/pkg/clustermanager/networkconnect/controller_test.go
+++ b/go-controller/pkg/clustermanager/networkconnect/controller_test.go
@@ -685,8 +685,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo := util.NewMutableNetInfo(netInfo)
 				mutableNetInfo.AddNADs(nsName + "/" + nadName)
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks[nsName] = mutableNetInfo
 				fakeNM.NADNetworks[nsName+"/"+nadName] = netInfo
+				fakeNM.Unlock()
 
 				// Ensure namespace and NAD are visible in informer caches before CNC creation.
 				gomega.Eventually(func() bool {
@@ -697,8 +699,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 					_, err := controller.nadLister.NetworkAttachmentDefinitions(nsName).Get(nadName)
 					return err == nil
 				}).WithTimeout(2 * time.Second).Should(gomega.BeTrue())
+
+				fakeNM.Lock()
 				fakeNM.NADNetworks[nsName+"/"+nadName] = netInfo
-				fakeNM.NADNetworks[nsName+"/"+nadName] = netInfo
+				fakeNM.Unlock()
 
 				// Create CNC with Primary UDN selector
 				cnc := testCNC(cncName, []apitypes.NetworkSelector{
@@ -759,8 +763,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo1 := util.NewMutableNetInfo(netInfo1)
 				mutableNetInfo1.AddNADs("frontend-a/primary-udn")
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks["frontend-a"] = mutableNetInfo1
 				fakeNM.NADNetworks["frontend-a/primary-udn"] = netInfo1
+				fakeNM.Unlock()
 
 				// Create second namespace and UDN
 				ns2 := &corev1.Namespace{
@@ -786,8 +792,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo2 := util.NewMutableNetInfo(netInfo2)
 				mutableNetInfo2.AddNADs("frontend-b/primary-udn")
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks["frontend-b"] = mutableNetInfo2
 				fakeNM.NADNetworks["frontend-b/primary-udn"] = netInfo2
+				fakeNM.Unlock()
 
 				// Create a non-matching namespace
 				ns3 := &corev1.Namespace{
@@ -881,8 +889,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo := util.NewMutableNetInfo(netInfo)
 				mutableNetInfo.AddNADs(nsName + "/" + nadName)
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks[nsName] = mutableNetInfo
 				fakeNM.NADNetworks[nsName+"/"+nadName] = netInfo
+				fakeNM.Unlock()
 
 				// Ensure namespace and NAD are visible in informer caches before CNC creation.
 				gomega.Eventually(func() bool {
@@ -956,8 +966,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo := util.NewMutableNetInfo(netInfo)
 				mutableNetInfo.AddNADs(nsName + "/" + nadName)
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks[nsName] = mutableNetInfo
 				fakeNM.NADNetworks[nsName+"/"+nadName] = netInfo
+				fakeNM.Unlock()
 
 				// Ensure namespace and NAD are visible in informer caches before CNC creation.
 				gomega.Eventually(func() bool {
@@ -1727,8 +1739,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller InitialSync Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo1 := util.NewMutableNetInfo(netInfo1)
 				mutableNetInfo1.AddNADs("pudn1-ns/primary-udn")
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks["pudn1-ns"] = mutableNetInfo1
 				fakeNM.NADNetworks["pudn1-ns/primary-udn"] = netInfo1
+				fakeNM.Unlock()
 
 				ns2 := newTestNamespace("pudn2-ns", map[string]string{
 					"cnc1-pudn":                     "true",
@@ -1748,8 +1762,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller InitialSync Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo2 := util.NewMutableNetInfo(netInfo2)
 				mutableNetInfo2.AddNADs("pudn2-ns/primary-udn")
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks["pudn2-ns"] = mutableNetInfo2
 				fakeNM.NADNetworks["pudn2-ns/primary-udn"] = netInfo2
+				fakeNM.Unlock()
 
 				// ============================================================
 				// Create NADs for CNC2 (1 CUDN + 1 P-UDN)
@@ -1777,8 +1793,10 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller InitialSync Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 				mutableNetInfo3 := util.NewMutableNetInfo(netInfo3)
 				mutableNetInfo3.AddNADs("pudn3-ns/primary-udn")
+				fakeNM.Lock()
 				fakeNM.PrimaryNetworks["pudn3-ns"] = mutableNetInfo3
 				fakeNM.NADNetworks["pudn3-ns/primary-udn"] = netInfo3
+				fakeNM.Unlock()
 
 				// ============================================================
 				// Create CNCs


go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -921,3 +958,608 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+		// Must belong to this network
+		if lb.ExternalIDs[ovntypes.NetworkExternalID] != networkName {
+			return false
+		}
+		// Must be a ClusterIP LB (name contains "_cluster")
+		return strings.Contains(lb.Name, "_cluster")
+	}
+
+	return libovsdbops.FindLoadBalancersWithPredicate(c.nbClient, predicate)
+}
```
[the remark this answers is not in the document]
tssurya: connecting a default network with cudns or udns is not allowed

tssurya:
I'm looking into the UT:
```
controller:ClusterNetworkConnect:cnc1:2:1:v4 k8s.ovn.org/name:cnc1 k8s.ovn.org/owner-controller:ovnkube-network-connect-controller k8s.ovn.org/owner-type:ClusterNetworkConnect network-id:2 node-id:1]} ip_prefix:10.1.0.0/24 nexthop:192.168.1.3] Rows:[] Columns:[] Mutations:[] Timeout:<nil> Where:[] Until: Durable:<nil> Comment:<nil> Lock:<nil> UUID: UUIDName:u4090446381} {Op:mutate Table:Logical_Router Row:map[] Rows:[] Columns:[] Mutations:[{Column:static_routes Mutator:insert Value:{GoSet:[{GoUUID:u4090446381}]}}] Timeout:<nil> Where:[where column _uuid == {c63c7f18-d84f-4015-89f5-bf634bc7a91a}] Until: Durable:<nil> Comment:<nil> Lock:<nil> UUID: UUIDName:}]
2026-01-26T08:16:53.2204360Z I0126 08:16:53.220266   31615 topology.go:234] CNC cnc1: executed 2 create operations for network netB
2026-01-26T08:16:53.2207261Z     topology_test.go:1024: 
2026-01-26T08:16:53.2207988Z         Expected
2026-01-26T08:16:53.2208670Z             <[]*nbdb.LogicalRouterPolicy | len:0, cap:0>: []
2026-01-26T08:16:53.2209086Z         not to be empty
2026-01-26T08:16:53.2270312Z I0126 08:16:53.224889   31615 watch.go:218] "Stopping fake watcher"
2026-01-26T08:16:53.2277877Z I0126 08:16:53.224990   31615 reflector.go:364] "Stopping reflector" type="*v1.Node" resyncPeriod="0s" reflector="k8s.io/client-go/informers/factory.go:160"
2026-01-26T08:16:53.2279178Z I0126 08:16:53.225050   31615 watch.go:218] "Stopping fake watcher"
2026-01-26T08:16:53.2279861Z I0126 08:16:53.225118   31615 watch.go:218] "Stopping fake watcher"
2026-01-26T08:16:53.2280517Z I0126 08:16:53.225179   31615 watch.go:218] "Stopping fake watcher"
2026-01-26T08:16:53.2281445Z I0126 08:16:53.225218   31615 watch.go:218] "Stopping fake watcher"
2026-01-26T08:16:53.2282093Z I0126 08:16:53.225305   31615 watch.go:218] "Stopping fake watcher"
2026-01-26T08:16:53.2283736Z I0126 08:16:53.225377   31615 reflector.go:364] "Stopping reflector" type="*v1.NetworkPolicy" resyncPeriod="0s" reflector="k8s.io/client-go/informers/factory.go:160"
2026-01-26T08:16:53.2285533Z I0126 08:16:53.225441   31615 reflector.go:364] "Stopping reflector" type="*v1.Namespace" resyncPeriod="0s" reflector="k8s.io/client-go/informers/factory.go:160"
2026-01-26T08:16:53.2287358Z I0126 08:16:53.225507   31615 reflector.go:364] "Stopping reflector" type="*v1.Service" resyncPeriod="0s" reflector="k8s.io/client-go/informers/factory.go:160"
2026-01-26T08:16:53.2288998Z I0126 08:16:53.225555   31615 reflector.go:364] "Stopping reflector" type="*v1.Pod" resyncPeriod="0s" reflector="k8s.io/client-go/informers/factory.go:160"
2026-01-26T08:16:53.2324785Z I0126 08:16:53.231719   31615 reflector.go:364] "Stopping reflector" type="*v1.EndpointSlice" resyncPeriod="0s" reflector="k8s.io/client-go/informers/factory.go:160"
2026-01-26T08:16:53.2369234Z I0126 08:16:53.236699   31615 factory.go:668] Stopping watch factory
2026-01-26T08:16:53.2369935Z --- FAIL: TestSyncNetworkConnectionsInactiveNetwork (0.33s)
2026-01-26T08:16:53.2370570Z === RUN   TestCreateRoutingPoliciesOps
2026-01-26T08:16:53.2371826Z === RUN   TestCreateRoutingPoliciesOps/create_single_IPv4_routing_policy
2026-01-26T08:16:53.3944247Z I0126 08:16:53.393581   31615 model_client.go:374] Create operations generated as: [{Op:insert Table:Logical_Router_Policy Row:map[action:reroute external_ids:{GoMap:map[destination-network-id:2 ip-family:v4 k8s.ovn.org/id:ovnkube-network-connect-controller:ClusterNetworkConnect:test-cnc:1:2:v4:network-router-1 k8s.ovn.org/name:test-cnc k8s.ovn.org/owner-controller:ovnkube-network-connect-controller k8s.ovn.org/owner-type:ClusterNetworkConnect router-name:network-router-1 source-network-id:1]} match:inport == "switch-to-router-port" && ip4.dst == 10.200.0.0/24 nexthops:{GoSet:[192.168.0.0]} priority:9001] Rows:[] Columns:[] Mutations:[] Timeout:<nil> Where:[] Until: Durable:<nil> Comment:<nil> Lock:<nil> UUID: UUIDName:u4090446382}]
```
seems unrelated to my PR but something to do with the new code I rebased on ... grrr - let me see if I need to account for something new for inactive networks

tssurya:
oh wow another coredump that's unrelated - I'm looking into it:

```
• [ABORTED] [166.706 seconds]
Network Segmentation [JustAfterEach] Sync perform east/west traffic between nodes following OVN Kube node pod restart L3 [Feature:NetworkSegmentation]
  [JustAfterEach] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1165
  [It] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:1854

  [ABORTED] Coredumps found during test execution: core.16179.ovnkube.ovn-worker2.6
  In [JustAfterEach] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1236 @ 01/26/26 09:07:19.963
```

tssurya:
> oh wow another coredump that's unrelated - I'm looking into it:
> 
> ```
> • [ABORTED] [166.706 seconds]
> Network Segmentation [JustAfterEach] Sync perform east/west traffic between nodes following OVN Kube node pod restart L3 [Feature:NetworkSegmentation]
>   [JustAfterEach] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1165
>   [It] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/network_segmentation.go:1854
> 
>   [ABORTED] Coredumps found during test execution: core.16179.ovnkube.ovn-worker2.6
>   In [JustAfterEach] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1236 @ 01/26/26 09:07:19.963
> ```

unrelated to this PR: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5878

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1000,611 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+
+// findNetworkSwitches finds logical switches for a network for the local zone node.
+// For Layer2, there's a single distributed switch (localZoneNode is ignored).
+// For Layer3, there's a switch per node, and we use direct lookup by constructing
+// the switch name from the node name (avoiding predicate scan over all switches).
+// We only support 1 node per zone for this feature.
+func (c *Controller) findNetworkSwitches(netInfo util.NetInfo) ([]*nbdb.LogicalSwitch, error) {
```
npinaeva: it is always only 1 switch though? would it be easier to just return 1 object instead of a list?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1000,611 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+		}
+
+		// Get remaining network's switches and LBs
+		remainingSwitches, err := c.findNetworkSwitches(remainingNetInfo)
+		if err != nil {
+			klog.Warningf("Failed to find switches for remaining network %s: %v", remainingNetInfo.GetNetworkName(), err)
+			continue
+		}
+
+		remainingLBs, err := c.findClusterIPLoadBalancers(remainingNetInfo)
```
npinaeva: i think this part can be optimizied if we pass a set of remainings network names, and only run a predicate once but matching all remaining networks?

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -332,6 +389,147 @@ func (c *Controller) reconcileNode(key string) error {
...
+	if isConnected {
+		klog.V(5).Infof("Service %s/%s network %s is connected by CNC %s with ClusterIPServiceNetwork",
+			svc.Namespace, svc.Name, networkOwnerKey, cnc.Name)
+	}
+
+	return isConnected
+}
+
+// hasClusterIPServiceNetwork checks if the CNC has ClusterIPServiceNetwork in its connectivity spec.
+func hasClusterIPServiceNetwork(cnc *networkconnectv1.ClusterNetworkConnect) bool {
```
npinaeva: maybe `serviceConnectivityEnabled` is a better name? I expected some network overlap check here :D

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1000,611 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+			allowMatch, nbdb.ACLActionPass, nil, libovsdbutil.LportEgress, 0)
+		acls = append(acls, allowServiceACL)
+	}
+
+	// Build drop-pod ACL match based on IP mode (single ACL)
+	// TODO(tssurya): check if we are dropping same network traffic.
+	var dropMatch string
+	switch {
+	case config.IPv4Mode && config.IPv6Mode:
+		dropMatch = fmt.Sprintf("(ip4.src == $%s && ip4.dst == $%s || ip6.src == $%s && ip6.dst == $%s) && ct.new",
```
npinaeva: I am wondering if we should match on the source here. In case some other source IP comes through the switch, we still don't want it going to the connected network I guess?
maybe there is no such traffic possible, but still

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -280,6 +316,27 @@ func nodeNeedsUpdate(oldObj, newObj *corev1.Node) bool {
...
+// serviceNeedsUpdate determines if a service change requires reconciliation.
+// We only care about service CREATE for ClusterIPServiceNetwork connectivity.
+// Service deletes don't need processing because:
+//   - LogicalSwitch.load_balancer is a weak reference with min=0 and refType=weak
+//   - When the LB is deleted by services controller, OVSDB automatically
+//     garbage-collects the UUID from all switches (including cross-network attachments)
+//
+// Service updates (like endpoint changes) are handled by the services controller.
+// We only need to care about load balancer attachments. LB Updates or deattachments
+// are handled by the services controller.
```
npinaeva: the loadbalancers  themselves never change on service update? what if the service spec changes?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1000,611 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+		dropPodACL := libovsdbutil.BuildACL(dbIDs, ovntypes.NetworkConnectDropPodTrafficPriority,
+			dropMatch, nbdb.ACLActionDrop, nil, libovsdbutil.LportEgress, 0)
+		acls = append(acls, dropPodACL)
+	}
+
+	return acls
+}
+
+// findSwitchesWithCNCACLs returns the names of switches that currently have this CNC's ACLs.
+func (c *Controller) findSwitchesWithCNCACLs(cncName string) (sets.Set[string], error) {
```
npinaeva: can't we use `cncState.connectedNetworks` for this like we do in the `syncNetworkConnections` and get the diff? it will be faster I think than running predicates

go-controller/pkg/types/const.go
```diff
@@ -97,6 +97,11 @@ const (
 	PrimaryUDNAllowPriority = 1001
 	// Default deny acl rule priority
 	PrimaryUDNDenyPriority = 1000
+	// Priority for allowing service traffic to pass through before the drop ACL
+	// for network connect partial service connectivity
+	NetworkConnectAllowServiceTrafficPriority = 5000
+	// Priority for dropping pod-to-pod traffic between connected networks
+	NetworkConnectDropPodTrafficPriority = 4950
```
npinaeva: so do we want them to be applied before the UDN isolation? I don't think there is a real intersection, but just to be sure

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1000,611 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+		return fmt.Errorf("failed to create address set ops for CNC %s: %w", cncName, err)
+	}
+
+	// Get the hashed address set names for ACL matches
+	hashNameV4, hashNameV6 := as.GetASHashNames()
+
+	// Build ACLs once (not per switch - they're identical)
+	acls := c.buildPartialConnectivityACLs(cncName, hashNameV4, hashNameV6)
+
+	// Create/update ACLs - needed to be done only once - during updates this should be no-op
```
npinaeva: I think it will be actually updated every time, but the lookup will be fast based on dbIDs

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1000,611 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+			}
+		}
+
+		// Collect switches for this network
+		switches, err := c.findNetworkSwitches(netInfo)
+		if err != nil {
+			klog.Warningf("Failed to find switches for network %s: %v", netInfo.GetNetworkName(), err)
+			continue
+		}
+		allSwitches = append(allSwitches, switches...)
```
npinaeva: it looks like the only usage for it is to build `desiredSwitches` set and we only need names and not the actual db entires? can we maybe build `desiredSwitches` right here instead without a db lookup?

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: load balancers do change on service update ie if the spec changes between etp=local, etp=cluster, itp=cluster,itp=local, endpointslice deletion/addition - all end up being a load balancer update/mutate taken care of by the services controller but they are never added/recreated again for same service - i.e expectation is the UUID/`Name` that's the dbindex for the LB always stays the same unless service itself is deleted and added again
So for CNC connect we don't need to care about what inside the LB changed, as long as we attach/deattach LB UUIDs to switches properly things should work for us wrt DNAT

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yea :D I realised this as well will change it to single switch

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: ah really? :( ah you mean we would send the ops but on DB side its no-op ? that's what I meant to write but when I read my comment it is mis-leading.. will reword it
TBD check the behaviour to confirm if ACLs are getting updated OR Not

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: Today, its not possible for other srcIP to come from the switch because we set the port sercurity set on LSPs BUT maybe there are cases like traffic from host via mpX into switch that could hack up to other networks..
There is already plan in future to disable port security to allow for mac spoofing so perhaps its not a bad idea to keep only dst here. - so in that case traffic is always dropped on origin switch itself..
I'll remove the source match for now as its not serving any extra purpose.

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yea this is a good point! let me check if I can get rid of the db lookup and save some operations

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I made this `getNetworkSwitchName`

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: > maybe `serviceConnectivityEnabled` is a better name? 

yea I can rename it to `serviceConnectivityEnabled`.

> I expected some network overlap check here :D

haha you mean between the subnets we want to connect? I thought when you do the pod subnet conflict check it would be taken care of that we won't allow for the CNC creation to succeed with overlapping subnets.. OR did you mean between service network and pod subnets - there is a bunch of conflict implementation checks to do for sure perhaps in its own PR

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yea it reduces from double loop over remaining Networks and remainingSwitches/disconnectedSwitches
I hope I understood this comment correctly PTAL the fix

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: this is fixed

go-controller/pkg/types/const.go
tssurya: yeah I don't think we have real intersection but I can change this to

1. 500
2. 475
3. 450

values to make it after the isolation ones

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I think the reason I didn't use that was because we update the caches at the end of syncNetworkConnections for whichever networks succeeded there..
let me see if I can reuse that but then update the cache only when `syncNetworkConnections` and `syncPartialConnectivity` and `syncServiceConnectivity` all complete for a network

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: hmm on second thoughts - the only thing we know for sure is that we don't want traffic from a connected network going to the other connected network for pod-network traffic.. I think let's stick with current match.
for traffic originating from other sources... problem is there could be network policies applied on namespace level and we don't want to drop that intentionally...since the policies are all applied to the common switch and not to transit router of the CNC I think what we have today is accurate for now

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I am going to return successful networks from syncPartialConnectivity && syncNetworkConnections - only the intersection of that will be entered into the cache as successfully connected networks..
there is a small downside to doing this which is there could be a case for a network where syncPartialConnectivity succeeded but syncNetworkConnections failed, so we retry and syncPartialConnectivity would also get recomputed.. but I think that is OK ?

go-controller/pkg/ovn/controller/networkconnect/controller.go
npinaeva: yeah exactly, all these checks are done, but maybe because I have worked on it recently, this function name made me confused :D

go-controller/pkg/ovn/controller/networkconnect/controller.go
npinaeva: huh that is not that obvious, so you are saying any service always has the same amount of loadbalancers regardless of updates. Is there some obvious way for me to check that? Maybe also add a comment in the code that creates LBs saying that we rely on this "feature"?

go-controller/pkg/ovn/controller/networkconnect/topology.go
npinaeva: yeah it is more difficult to store successfully added networks if addition consists of multiple parts, maybe we can just add another set to track successfully partially connected networks? then they can be independent?

go-controller/pkg/ovn/controller/networkconnect/topology.go
npinaeva: yeah this one is not so important, but I just want to clarify the idea here. So for service-only connectivity we don't want networks to be able to talk to each other, but the actual connecting routes still exist in the CNC router, so if the traffic is not dropped at the entrance (aka on the switch)  it will reach another network. That what made me think the this switch should never send traffic to the connected network destination (I would expect users that only request service connectivity to not want anyone to reach their podIPs). I don't fully understand the network policy case: so if the namespace has network policy allowing connected network we will override it with drop, but I would also expect the connected network to not be reachable anyway?

Side-note: we allow 2 networks to be selected by different CNCs as long as they have different connect subnets, but if one CNC has podConnectivity and another doesn't then the traffic will be dropped on the switch by the opartial connectivity ACLs anyway, effectively only allowing serviceConnectivity? Do we need another check for this in the CNC validation?

go-controller/pkg/ovn/controller/networkconnect/topology.go
npinaeva: yeah exactly, I have another nit: it would be nice to reuse code between `findClusterIPLoadBalancers` and `findClusterIPLoadBalancersForNetworks`, e.g. by calling `findClusterIPLoadBalancersForNetworks with a single-object set or using `(networkNames ...string) kind of param and having one function

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: ah I see, yup renamed it.

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yes :D I am trying to get rid of one of them for sure... I didn't mean for both to stay sorry

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: > huh that is not that obvious, so you are saying any service always has the same amount of loadbalancers regardless of updates. Is there some obvious way for me to check that? Maybe also add a comment in the code that creates LBs saying that we rely on this "feature"?

correct! Each ns/servicename has exactly 1 LB for all "clusterIPs, loadbalancers, externalIPs" (well 3 if its tcp,udp,sctp) - no matter how many updates we get the number of loadbalancer is always max 3 - what gets updated is VIP to backend value in the LB.
NodePorts are the only ones with multiple load balancers - i.e cause the backend pods when they are host-networked we have a masquerade logic to counter that. - but we don't care about nodeports for our case.

I'll see if I can poinpoint exact code for this and add a comment.. its been soo long since we did services controller.. let me dig into it

There is a slightly related discussion to be had wrt to should we rename "clusterIPServiceNetwork" to just "serviceNetwork"... but let me have that in a slack thread later once I finish up all the review comments and refactors.

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I'm refactoring serviceconnectivity now to make the reuse of cache more central.. let's see how it comes out
so just going to treat load balancers as another ops after staticroute ops

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: > yeah this one is not so important, but I just want to clarify the idea here. So for service-only connectivity we don't want networks to be able to talk to each other, but the actual connecting routes still exist in the CNC router, so if the traffic is not dropped at the entrance (aka on the switch) it will reach another network. 

yea, agreed

> That what made me think the this switch should never send traffic to the connected network destination (I would expect users that only request service connectivity to not want anyone to reach their podIPs). 

yes!

> I don't fully understand the network policy case: so if the namespace has network policy allowing connected network we will override it with drop, but I would also expect the connected network to not be reachable anyway?

yea what I meant was, let's say networkA(172.31.0.0/24) networkB(172.31.1.0/24) are connected, srcIP is 192.168.0.5 originating from the node or somewhere a srcIP that's not of networkA is trying to reach networkB via switchA (we know such a traffic flow doesn't exist today but being hypothetical here), and if the user does have a policy to allow that srcIP for networkB I was just thinking if we want to block that or not - i.e the SLA we say is "podNetwork is not connected" "service network is" but maybe we can make it stronger removing the srcIP from the picture and always dropping.. and its not that pod network isn't connected, but that its total isolation - and that's fine by me, hmm OK maybe actually keeping only destination is more secure?

> 
> Side-note: we allow 2 networks to be selected by different CNCs as long as they have different connect subnets, but if one CNC has podConnectivity and another doesn't then the traffic will be dropped on the switch by the opartial connectivity ACLs anyway, effectively only allowing serviceConnectivity? Do we need another check for this in the CNC validation?

hmm so 
CNC1: netA,netB -> podNetwork only -> so noACLs -> traffic will be allowed to netB
CNC2: netA,netC -> serviceNetwork only -> yes ACLs - but these will pass service traffic, pass to self-network, drop traffic to destination netC only allow to services on netC on switchA

so that case is OK ^ and its point 1 in OKEP under valid ways of connecting UDNs

I think the scenario that's gonna go wrong which you mean is:

CNC1: netA,netB -> service only - so this drops traffic to netB
CNC2: netA,netB -> pod only - so this expects traffic to netB

this is problematic and its point2 in OKEP under valid ways of connecting UDNs

https://ovn-kubernetes.io/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns/#validation-field-checks

and we said:

A ClusterNetworkConnect could be selecting a subset of networks already selected by another ClusterNetworkConnect API. What is the expectation in this case? Example, ClusterNetworkConnect-A selects blue, green, red and yellow networks and ClusterNetworkConnect-B selects blue and green networks.
This is going to be considered a user-configuration error oversight. This same issue also exists for other APIs like RouteAdvertisement (wonder what we do there?).
We expect the rules to be additively implemented but in this case, we won't optimize for anything since its a hard problem to solve. We will end up create duplicate connect networks unnecessarily and it will be an ECMP routing situation on which connect network will be used to reach the other side.
We will not provide any validation around selectors across all connect APIs to rule out overlap except we will ensure the connectSubnets is non-overlapping across all connectAPIs that select a given network which is same as the first case.

but its true that I didn't think of the type of connectivity.. I would say no need for validation check - unless its easy to do it as part of the other validations we are doing cause at the end of the day - we are honoring the "Secure" case of playing it safe and assuming drop

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I think I've succeeded in removing this util and merging things better - and reusing the connectedNetworks cache instead of building it again from predicates... but let's see what the next round of push tells :D 

resolving this comment

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: done.

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: Ive done a comment above the findLoadBalacers util: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/9156a8259dc7c0a9f2150113f313219c643fa866/go-controller/pkg/ovn/controller/services/lb_config.go#L121 is the algorithm used

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I've removed src match and added a comment capturing the why.

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/236ea6cb960e303b03a30e679fc4be5587a2e6fb..e76401bd0c668997488aa157556a641c2e4881d0 is addressing the review comments + refactoring it a  bit better

next ill just rebase on master that can be ignored

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -332,6 +392,147 @@ func (c *Controller) reconcileNode(key string) error {
...
+			c.cncController.Reconcile(cnc.Name)
+		}
+	}
+
+	return nil
+}
+
+// getNetworkOwnerKeyForNamespace returns the owner key (e.g., "layer3_5") for the primary network
+// of the given namespace. Returns empty string if namespace uses default network.
+// TODO(tssurya): Make this a common utility function and reuse the one in clustermanager/networkconnect/controller.go
```
npinaeva: just tagging this to not forget

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1091,589 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+			v4Subnets = append(v4Subnets, subnet)
+		} else {
+			v6Subnets = append(v6Subnets, subnet)
+		}
+	}
+
+	// Build match for same-network traffic (src and dst both in this network's subnets)
+	var matches []string
+	for _, v4Subnet := range v4Subnets {
+		matches = append(matches, fmt.Sprintf("(ip4.src == %s && ip4.dst == %s)", v4Subnet, v4Subnet))
```
npinaeva: do we want to use src match here considering we dropped it from the deny acl?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1091,589 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+	// This drops any new connections to connected networks' pod subnets.
+	// Same-network traffic is allowed by the per-network allow-same-network ACLs
+	// at higher priority (475). Service traffic is allowed by allow-service ACL (500).
+	// We only match on dst (not src) because:
+	// 1. This ACL is on egress (from-lport), so it only evaluates outbound pod traffic
+	// 2. Matching only dst is more restrictive - blocks traffic regardless of source
+	var dropMatch string
+	switch {
+	case config.IPv4Mode && config.IPv6Mode:
+		dropMatch = fmt.Sprintf("(ip4.dst == $%s || ip6.dst == $%s) && ct.new",
```
npinaeva, line 1688: is the ct.new part still under investigation?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1091,589 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+	// Delete address sets using proper DbObjectIDs predicate
+	asPredicateIDs := libovsdbops.NewDbObjectIDs(libovsdbops.AddressSetClusterNetworkConnect, controllerName,
+		map[libovsdbops.ExternalIDKey]string{
+			libovsdbops.ObjectNameKey: cncName,
+		})
+	asPredicate := libovsdbops.GetPredicate[*nbdb.AddressSet](asPredicateIDs, nil)
+	err = libovsdbops.DeleteAddressSetsWithPredicate(c.nbClient, asPredicate)
+	if err != nil {
+		return fmt.Errorf("failed to delete address sets for CNC %s: %w", cncName, err)
+	}
```
npinaeva: I think we can use `	c.addressSetFactory.DestroyAddressSet(getConnectedUDNSubnetsAddressSetDbIDs(cncName))` instead to avoid a predicate search

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -187,6 +220,19 @@ func (c *Controller) syncNetworkConnections(cnc *networkconnectv1.ClusterNetwork
 		// Build ops per network to keep transaction sizes bounded
 		var createOps []ovsdb.Operation
 
+		// Add partial connectivity ACLs to this network's switch
+		// This creates the ACLs (idempotent) and adds them to the switch
+		if partialConnectivityDesired {
+			switchName, ok := partialConnState.networkSwitches[networkID]
+			if ok {
```
npinaeva: isn't it guaranteed to be present? I was a bit confused that we pass both `partialConnState + networkID` and switch name to the function, even though switch name can be obtained from `partialConnState` (which is what happens here)

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -378,6 +435,38 @@ func (c *Controller) syncNetworkConnections(cnc *networkconnectv1.ClusterNetwork
...
 		}
 
+		// If ClusterIPServiceNetwork is enabled, cleanup LB attachments for this network.
+		// Unlike the add case where bidirectionality happens naturally via the outer loop
+		// (each network attaches its LBs to others, covering all pairs), in delete the
+		// deleted network won't be iterated again. So we must explicitly cleanup both
+		// directions: remove this network's LBs from remaining switches AND remove
+		// remaining networks' LBs from this network's switch.
+		if serviceConnectivityDesired {
+			remainingNetworks := cncState.connectedNetworks.Clone()
```
npinaeva: should we call it previouslyConnectedNetworks? remaining made me think it was a set of still-connected networks, which would mean the cleanup for a set of deleted network won't be complete (same for the parameter name in the `cleanupLoadBalancersOps`)

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yea there is a TBD - 1st point in PR describe....

dceara:
Hi @tssurya!

On:

> Check with @dceara if load balancer delete will cleanup references or not in OVN? - Ensure a UT exists for this

References to Load_Balancer records are all "weak" in the OVN NB schema.  That means that whenever the DB server processes a transaction deleting a load balancer, all references to it from other tables are automatically cleared.  That's the behavior for all "weak" references in general.

So ovnk doesn't have to explicitly cleanup the references.  It can rely on ovsdb-server doing that when processing the LB delete transaction.

Regards,
Dumitru

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -145,6 +151,19 @@ func NewController(
...
+		Informer:       wf.ServiceCoreInformer().Informer(),
+		Lister:         serviceLister.List,
+		Reconcile:      c.reconcileService,
+		ObjNeedsUpdate: serviceNeedsUpdate,
+		Threadiness:    1,
+	}
+	c.serviceController = controllerutil.NewController(
+		"ovnkube-network-connect-service-controller",
+		serviceCfg,
+	)
```
trozet, line 182: I'm a bit hesitant about adding a services controller here. In my opinion I think it would be better to make the service controller reconcile, then in service controller it checks the CNC cache to see which networks are connected to see if it should care about the service or not.

I think this will become even more obvious as the most logical path once the per-UDN services controller is simplified into 1.

tssurya: > I'm a bit hesitant about adding a services controller here. In my opinion I think it would be better to make the service controller reconcile, then in service controller it checks the CNC cache to see which networks are connected to see if it should care about the service or not.

hmm, what CNC cache? do you mean adding some kind of CNCnotifier into the services controller?

> 
> I think this will become even more obvious as the most logical path once the per-UDN services controller is simplified into 1.

but this doesn't exist today.. - the single services controller.. do we want each service controller to "watch CNC objects?" 

I think we spoke about this approach here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5246#discussion_r2402708720 and https://cloud-native.slack.com/archives/C08452HR8V6/p1760004741369849?thread_ts=1760004611.343839&cid=C08452HR8V6

not sure if you are suggesting something different from that solution..

tssurya:
There redness in CI is because the e2e test fails with the current new update:
There is a potential race when service is intact, but we are deleting and recreating the UDN+pods in the namespace - so service update doesn't happen, the load balancer is not cleaned up when old service controller is stopped... and then when new service controller comes up it deletes this stale LB that causes confusion in CNC cause it had just attached that old LB thinking that's the LB. - I am fixing this by adding an expected count when fetching LBs to ensure we are retrying if we don't find LBs...a more neater but elaborate way is to add LBNotifier instead of service controller - we can see if that's needed in case we are missing anything..

tssurya:
> I had to pause the review here, because I realized we may need to fix the LB cross-referencing somehow. So currently if 2 CNCs select the same set of networks with service connect, we will cross-add networks loadbalancers to each others switches, but we never count how many CNC "caused" this addition. So if we delete one of the CNCs, then delete workflow will remove LBs, even though the second CNC is still present

this is a good catch, I'm refactoring to use LBGs (per CNC) to avoid this situation

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: yea there is a TODO in PR describe so that I don't forget.. just waiting for main code to settle a bit

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I guess we could drop the deny acl match yea

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I think we ended up simplifying the code when we use LBG now.. so this should get outdated.. I can squash commits in the end if that helps..but keeping the LBG commit separate on top for now for easier review/bugfix capture

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: I think this comment from @jcaamano https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5911#discussion_r2774970623 also answers this question. I think @npinaeva you were right that we always send an update if the entry is found and that we don't do anything smarter than that at libovsdbops level
cc @dave-tucker as well you were right too

tssurya:
The things I addressed in the recent push were:

1. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828#issuecomment-3867316209 via commit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828/changes/9920ef2bd799fee1f33004d570adf7b3b0f1c6e9
2. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828#issuecomment-3867827435 via commit https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828/changes/0f33f0f52a66d5672cb1aa145225b360a7c478c2

I am working on addressing remaining review comments. I do think we probably need to squash the above two commits - but I didn't wanna make it hard for @npinaeva since I had to rebase so kept new change in commits along with why we redid the design with LBG - but I can also squash them into main commit.

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: done.

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yup you are right, ok check is unnecessary
and switchname is also redundant as the function can just look it up

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -187,6 +244,19 @@ func (c *Controller) syncNetworkConnections(cnc *networkconnectv1.ClusterNetwork
...
+		if partialConnectivityDesired {
+			switchName, ok := partialConnState.networkSwitches[networkID]
+			if ok {
+				createOps, err = c.ensurePartialConnectivityACLsOps(createOps, partialConnState, networkID, switchName)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("CNC %s: failed to ensure partial connectivity ACLs for network %s: %w", cncName, netInfo.GetNetworkName(), err))
+					continue
+				}
+			}
+		}
```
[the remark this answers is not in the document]
tssurya: its fixed in latest push, ptal

test/e2e/cluster_network_connect.go
```diff
@@ -2807,5 +2857,1376 @@ var _ = Describe("ClusterNetworkConnect OVN-Kubernetes Controller", feature.Netw
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("29. Deleting green PUDN network object")
+			deleteUDN(greenNs, greenUDN)
+
+			By("29. Waiting for green pods to be deleted")
+			Eventually(func() bool {
+				_, err := cs.CoreV1().Pods(greenNs).Get(context.Background(), pods["green-pod-0"].Name, metav1.GetOptions{})
+				return err != nil
+			}, 60*time.Second, 1*time.Second).Should(BeTrue())
```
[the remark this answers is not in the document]
tssurya: this is fixed in new push, ptal

test/e2e/cluster_network_connect.go
```diff
@@ -2808,4 +2865,954 @@ var _ = Describe("ClusterNetworkConnect OVN-Kubernetes Controller", feature.Netw
...
+
+			By("27. Deleting blue PUDN network object")
+			deleteUDN(blueNs, blueUDN)
+
+			By("27. Waiting for blue pods to be deleted")
+			Eventually(func() bool {
+				_, err := cs.CoreV1().Pods(blueNs).Get(context.Background(), pods["blue-pod-0"].Name, metav1.GetOptions{})
+				return err != nil
+			}, 60*time.Second, 1*time.Second).Should(BeTrue())
+
```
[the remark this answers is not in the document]
tssurya: this is fixed in new push, ptal

test/e2e/cluster_network_connect.go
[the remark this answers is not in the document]
tssurya: huh what about in L3465?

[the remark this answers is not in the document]
tssurya: and L3434

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -391,14 +498,54 @@ func (c *Controller) syncNetworkConnections(cnc *networkconnectv1.ClusterNetwork
...
+	if partialConnectivityWasEnabled && !partialConnectivityDesired {
+		klog.V(4).Infof("CNC %s: partial connectivity disabled, cleaning up ACLs", cncName)
+		if err := c.cleanupPartialConnectivity(cncName); err != nil {
+			errs = append(errs, fmt.Errorf("CNC %s: failed to cleanup partial connectivity: %w", cncName, err))
+		}
+	}
+
+	// Update state
+	cncState.clusterIPServiceNetworkEnabled = serviceConnectivityDesired
+	cncState.podNetworkConnectEnabled = podConnectivityDesired
```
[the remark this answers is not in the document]
tssurya: this is fixed in latest version ptal

test/e2e/cluster_network_connect.go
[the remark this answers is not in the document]
tssurya: no please check L3465: 
```
Eventually(func() bool {
				_, err := cs.CoreV1().Pods(greenNs).Get(context.Background(), pods["green-pod-1"].Name, metav1.GetOptions{})
				return err != nil
			}, 60*time.Second, 1*time.Second).Should(BeTrue())
```
and L3434
```
Eventually(func() bool {
				_, err := cs.CoreV1().Pods(blueNs).Get(context.Background(), pods["blue-pod-1"].Name, metav1.GetOptions{})
				return err != nil
			}, 60*time.Second, 1*time.Second).Should(BeTrue())
```
are you not able to see these lines?

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: hmm I can't find the overlapping util function in cluster manager either..
for this all I did is replaced code with util `util.ComputeNetworkOwner(primaryNetwork.TopologyType(), primaryNetwork.GetNetworkID())` and resolved the TODO, if we see room for consolidating further we can..

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -29,6 +29,28 @@ func getConnectRouterName(cncName string) string {
...
+// Each CNC gets its own LBG so that overlapping CNCs don't interfere with each other's cleanup.
+func getCNCServiceLBGroupName(cncName string) string {
+	return ovntypes.NetworkConnectServiceLBGroupPrefix + cncName
+}
+
+// findCNCServiceLBGroup looks up the CNC's service LoadBalancerGroup by name and returns it
+// with its UUID populated. Returns nil, nil if the LBG doesn't exist.
+func (c *Controller) findCNCServiceLBGroup(cncName string) (*nbdb.LoadBalancerGroup, error) {
+	lbgName := getCNCServiceLBGroupName(cncName)
+	lbgs, err := libovsdbops.FindLoadBalancerGroupsWithPredicate(c.nbClient, func(lbg *nbdb.LoadBalancerGroup) bool {
```
npinaeva: Name is a server-side index for lb group, so we don't need a predicate

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1119,639 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+			}
+			expectedCount += protocols.Len()
+		}
+	}
+	return expectedCount, nil
+}
+
+// ensureLoadBalancers finds and validates ClusterIP LBs for a single network.
+// Returns the found LBs and an error if the found count doesn't match the expected count.
+// This is a pure find+validate function -- it does NOT build any switch or LBG mutation ops.
```
npinaeva, line 1260: I think the name should be changed then to `findLoadBalancers`, we usually use ensure with create db ops?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -963,3 +1119,639 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+
+	// Count expected LBs based on ClusterIP services in this network's namespaces
+	expectedLBCount, err := c.countExpectedClusterIPLBs(netInfo)
+	if err != nil {
+		return nil, fmt.Errorf("failed to count expected LBs for network %s: %w", netInfo.GetNetworkName(), err)
+	}
+
+	foundLBCount := len(thisnetworkLBs)
+
+	klog.V(5).Infof("Network %s (ID=%d): found %d ClusterIP LBs (expected %d)",
```
npinaeva, line 1281: We need this check in case the service controller hasn't created all the loadbalancers yet?

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -166,6 +202,7 @@ func (c *Controller) Start() error {
 		c.cncController,
 		c.nodeController,
 		c.nadReconciler,
+		c.serviceController,
```
npinaeva, line 209: Aren't we missing initial sync on startup to cleanup non-existent CNCs?

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: yea! exactly.. ideally it would be nice to add a "loadbalancer" create notifier that we can plug in.. but for now this works.. I'll add a comment to clarify that's why we are having to do this.
I hit this race in CI since we have e2e tests that check for "service didn't get any updates, all that happened was UDN got recreated" more details in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828/changes/62f59792edeb0cdb115c9063cce194ee03f87912. But tl;dr was some services in that network could have their lbs created - they get programmed, but when rest get programmed there is no trigger.. so had to add this logic

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: yup we are! point6 in my PR description and time to add a commit for this! on it..

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: good point!

tssurya:
all I did was plain rebase.. now working on addressing comments + remaining items in the PR describe... hold reviews..

tssurya:
> 3. The first commit which is a bug fix is not handling upgrades scenario - we probably at this point need to update the names of the load balancers at startup.
>    
>    * So looks like `EnsureLBs` is handling this correctly already. It takes `existingCacheLBs` and `LBs` desired.
>    * `toDelete` map initializes it with all existing cached LBs.
>    * `toDelete[lb.UUID] = lb` -- every existing LB starts as "to delete"
>    * `existingLB := existingByName[lb.Name]` -- matches by name since that's the index in DB
>    * `delete(toDelete, existingLB.UUID)` -- only removes from toDelete if name matches
>    * Everything still in toDelete gets deleted
>    * So: old per-node LB (name `_node_router+switch`) stays in `toDelete` because new desired LB has name `_cluster` (no name match). Old LB gets deleted
>    * I'll test and confirm this as well. THIS IS DONE, I checked it. We don't need to do anything specific for this.

proof of test:

```
I0211 10:52:56.570576   14254 services_controller.go:498] Services do not match for network=default, existing lbs: []services.LB{services.LB{Name:"green_green.network_Service_green/service-green_TCP_node_router+switch_ovn-control-plane", UUID:"82de3e7c-5df5-43f0-a257-35a0f62582a4", Protocol:"tcp", ExternalIDs:map[string]string{"k8s.ovn.org/kind":"Service", "k8s.ovn.org/network":"green_green-network", "k8s.ovn.org/owner":"green/service-green", "k8s.ovn.org/role":"primary"}, Opts:services.LBOpts{Reject:false, EmptyLBEvents:false, AffinityTimeOut:0, SkipSNAT:false, Template:false, AddressFamily:""}, Rules:[]services.LBRule{}, Templates:services.TemplateMap{}, Switches:[]string{"green_green.network_ovn-control-plane"}, Routers:[]string{"GR_green_green.network_ovn-control-plane"}, Groups:[]string{}}}, built lbs: []services.LB{services.LB{Name:"Service_green/service-green_TCP_cluster", UUID:"", Protocol:"TCP", ExternalIDs:map[string]string{"k8s.ovn.org/kind":"Service", "k8s.ovn.org/owner":"green/service-green"}, Opts:services.LBOpts{Reject:true, EmptyLBEvents:false, AffinityTimeOut:0, SkipSNAT:false, Template:false, AddressFamily:""}, Rules:[]services.LBRule{services.LBRule{Source:services.Addr{IP:"10.96.6.229", Port:80, Template:(*services.Template)(nil)}, Targets:[]services.Addr{services.Addr{IP:"10.244.1.4", Port:8080, Template:(*services.Template)(nil)}, services.Addr{IP:"10.244.2.7", Port:8080, Template:(*services.Template)(nil)}}}}, Templates:services.TemplateMap(nil), Switches:[]string{}, Routers:[]string{}, Groups:[]string{"clusterLBGroup"}}}
```
Before:
```
Defaulted container "nb-ovsdb" out of: nb-ovsdb, sb-ovsdb, ovn-northd, ovnkube-controller, ovn-controller, ovs-metrics-exporter
sh-5.2# ovn-nbctl find Load_Balancer | grep -E "name.*_node_router|name.*_cluster"
name                : "Service_frr-k8s-system/frr-k8s-metrics-service_TCP_node_router_ovn-control-plane"
name                : "Service_default/kubernetes_TCP_node_router_ovn-control-plane"
name                : "blue_blue.network_Service_blue/service-blue_TCP_node_router+switch_ovn-control-plane"
name                : "green_green.network_Service_green/service-green_TCP_node_router+switch_ovn-control-plane"
name                : "Service_kube-system/kube-dns_UDP_cluster"
name                : "Service_kube-system/kube-dns_TCP_cluster"
name                : "Service_frr-k8s-system/frr-k8s-webhook-service_TCP_node_router+switch_ovn-control-plane"
sh-5.2# ovn-nbctl list load-balancer blue_blue.network_Service_blue/service-blue_TCP_node_router+switch_ovn-control-plane
_uuid               : 3f286712-f760-4582-8a7d-1d3ae903beb3
external_ids        : {"k8s.ovn.org/kind"=Service, "k8s.ovn.org/network"=blue_blue-network, "k8s.ovn.org/owner"="blue/service-blue", "k8s.ovn.org/role"=primary}
health_check        : []
ip_port_mappings    : {}
name                : "blue_blue.network_Service_blue/service-blue_TCP_node_router+switch_ovn-control-plane"
options             : {event="false", hairpin_snat_ip="169.254.0.5 fd69::5", neighbor_responder=none, reject="true", skip_snat="false"}
protocol            : tcp
selection_fields    : []
vips                : {"10.96.124.96:80"="103.103.1.5:8080"}
sh-5.2# ovn-nbctl list load-balancer green_green.network_Service_green/service-green_TCP_node_router+switch_ovn-control-plane
_uuid               : 82de3e7c-5df5-43f0-a257-35a0f62582a4
external_ids        : {"k8s.ovn.org/kind"=Service, "k8s.ovn.org/network"=green_green-network, "k8s.ovn.org/owner"="green/service-green", "k8s.ovn.org/role"=primary}
health_check        : []
ip_port_mappings    : {}
name                : "green_green.network_Service_green/service-green_TCP_node_router+switch_ovn-control-plane"
options             : {event="false", hairpin_snat_ip="169.254.0.5 fd69::5", neighbor_responder=none, reject="true", skip_snat="false"}
protocol            : tcp
selection_fields    : []
vips                : {"10.96.6.229:80"="104.104.0.4:8080,104.104.1.4:8080"}
sh-5.2# 
sh-5.2# 
```
After:
```
Defaulted container "nb-ovsdb" out of: nb-ovsdb, sb-ovsdb, ovn-northd, ovnkube-controller, ovn-controller, ovs-metrics-exporter
sh-5.2# ovn-nbctl find Load_Balancer | grep -E "name.*_node_router|name.*_cluster"
name                : "Service_frr-k8s-system/frr-k8s-metrics-service_TCP_node_router_ovn-control-plane"
name                : "Service_default/kubernetes_TCP_node_router_ovn-control-plane"
name                : "Service_kube-system/kube-dns_UDP_cluster"
name                : "Service_kube-system/kube-dns_TCP_cluster"
name                : "Service_frr-k8s-system/frr-k8s-webhook-service_TCP_node_router+switch_ovn-control-plane"
name                : "Service_blue/service-blue_TCP_cluster"
name                : "Service_green/service-green_TCP_cluster"
sh-5.2#  ovn-nbctl list load-balancer Service_blue/service-blue_TCP_cluster
_uuid               : 66e2c7d6-f816-419d-b42d-4261041b772c
external_ids        : {"k8s.ovn.org/kind"=Service, "k8s.ovn.org/owner"="blue/service-blue"}
health_check        : []
ip_port_mappings    : {}
name                : "Service_blue/service-blue_TCP_cluster"
options             : {event="false", hairpin_snat_ip="169.254.169.5 fd69::5", neighbor_responder=none, reject="true", skip_snat="false"}
protocol            : tcp
selection_fields    : []
vips                : {"10.96.124.96:80"="10.244.1.3:8080"}
sh-5.2#  ovn-nbctl list load-balancer Service_green/service-green_TCP_cluster
_uuid               : 75f37727-10bd-40a3-9945-0458adcc5805
external_ids        : {"k8s.ovn.org/kind"=Service, "k8s.ovn.org/owner"="green/service-green"}
health_check        : []
ip_port_mappings    : {}
name                : "Service_green/service-green_TCP_cluster"
options             : {event="false", hairpin_snat_ip="169.254.169.5 fd69::5", neighbor_responder=none, reject="true", skip_snat="false"}
protocol            : tcp
selection_fields    : []
vips                : {"10.96.6.229:80"="10.244.1.4:8080,10.244.2.7:8080"}
```
resolving this.

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: seems like what we want is pass related.. for now we should keep the workaround but we are deciding which is better way to go here: https://cloud-native.slack.com/archives/C08452HR8V6/p1770720413427739

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: agreed, changing..

go-controller/pkg/ovn/controller/networkconnect/topology.go
tssurya: changed, add a getloadbalancergroup util

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: this is done.

go-controller/pkg/ovn/controller/networkconnect/controller.go
tssurya: ok actually you were right!
I figured out a corner case that I had missed - like mentioned above 1-3 loadbalancers can be there but,,, whether its 1 or 2 or 3 is determined by the ports on service spec which can be updated.. so I'll add that one bug as a fix.. I realised this as I added a unit test around this and it wasn't passing.. lol.. I'll also add a e2e test that tests the service spec update change...

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/ca6d7aa5ec7b28eda7a22a90b3f5b33de87baefb..690e7e0cc16a12309881cae66ca46f594e024c87 addresses

1. review comments and
2. adds more unit tests for service connectivity
3. and adds an initialSync
4. and fixes a bug around service update
 - all done in new commits on top, see commit messages for more information

I still need to add unit tests for partialconnectivity + rename things to `ServiceNetwork` - will do those two as new commits as well.

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -332,6 +413,142 @@ func (c *Controller) reconcileNode(key string) error {
...
+	}
+
+	// Check if it's the default network
+	if primaryNetwork.IsDefault() {
+		return "", nil
+	}
+
+	// Build owner key: "layer3_{networkID}" or "layer2_{networkID}"
+	return util.ComputeNetworkOwner(primaryNetwork.TopologyType(), primaryNetwork.GetNetworkID()), nil
+}
```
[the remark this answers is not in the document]
tssurya: good bot! this is a gap. why didn't you point it out in earlier iterations? did something change in the most recent push?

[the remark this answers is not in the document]
tssurya: @coderabbitai: on thinking more.. but still i don't get one thing though.. when syncNAD requeues the CNC - shouldn't that take care of syncing the service as well?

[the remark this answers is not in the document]
tssurya: @coderabbitai: on thinking more.. but still i don't get one thing though.. when syncNAD requeues the CNC - shouldn't that take care of syncing the service as well? I don't think this is a bug is it?

go-controller/pkg/ovn/controller/networkconnect/repair_test.go
```diff
@@ -0,0 +1,411 @@
...
+	nbClient, cleanup, err := libovsdbtest.NewNBTestHarness(libovsdbtest.TestSetup{
+		NBData: initialDB,
+	}, nil)
+	require.NoError(t, err)
+
+	c := &Controller{
+		nbClient:          nbClient,
+		cncLister:         newFakeCNCLister(cncNames...),
+		addressSetFactory: addressset.NewOvnAddressSetFactory(nbClient, config.IPv4Mode, config.IPv6Mode),
+	}
```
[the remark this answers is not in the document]
tssurya: but this test isn't setting any config values is it?
I changed this to passing true,true i think that should be ok instead of passing in config values

tssurya:
> https://github.com/ovn-kubernetes/ovn-kubernetes/compare/ca6d7aa5ec7b28eda7a22a90b3f5b33de87baefb..690e7e0cc16a12309881cae66ca46f594e024c87 addresses
> 
> 1. review comments and
> 2. adds more unit tests for service connectivity
> 3. and adds an initialSync
> 4. and fixes a bug around service update
> 
> * all done in new commits on top, see commit messages for more information
> 
> I still need to add unit tests for partialconnectivity 

https://github.com/ovn-kubernetes/ovn-kubernetes/compare/1c39a8e942cfaca6db2718cf4f0a97fc65edf935..b4b0dc293805c08943320945d68b73d3829da5df
adds remaining UTs and adds basic support for dynamic udns skipping..There is a tracker to consider dynamic UDNs better via https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5963

tssurya:
last push was just rebase on master...
OK so only two FUPs needed: points 8 & 11 in TBD of the PR describe. I'm happy with the rest of the elements here provided CI passes with lots more of the tests I added now...
I think I have addressed everything else except point1, I guess we have a bunch of ideas in https://cloud-native.slack.com/archives/C08452HR8V6/p1770720413427739 just need to pick one of these...

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -332,6 +416,144 @@ func (c *Controller) reconcileNode(key string) error {
...
+	}
+
+	// Get the network owner key for this service's namespace.
+	// This determines which network the service belongs to.
+	networkOwnerKey, err := c.getNetworkOwnerKeyForNamespace(namespace)
+	if err != nil {
+		// Log and skip - namespace might not have a primary UDN
+		klog.V(5).Infof("Could not get network owner key for namespace %s: %v", namespace, err)
+		return nil
+	}
```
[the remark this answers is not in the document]
tssurya: @coderabbitai but when NADSync happens it will automatically sync services right? same as https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828#discussion_r2797953314 right?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -970,3 +1162,631 @@ func (c *Controller) getNodeSubnet(netInfo util.NetInfo, nodeName string) ([]*ne
...
+			cncName, len(disconnectedLBs), disconnectedNetInfo.GetNetworkName(), lbgName)
+	}
+
+	// Remove the CNC's LBG from the disconnected network's switch
+	disconnectedSwitchName, err := c.getNetworkSwitchName(disconnectedNetInfo)
+	if err != nil {
+		klog.Warningf("CNC %s: failed to get switch name for disconnected network %s: %v",
+			cncName, disconnectedNetInfo.GetNetworkName(), err)
+		return ops, nil
+	}
```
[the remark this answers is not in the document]
tssurya: but we do return the ops done so far right?
isn't LBG a weak reference? so LBs from that LBG will be gone if the switch is gone so getNetworkSwitchName fails right? so its just linking an empty LBG?
since switch is totally gone there is nothing to remove it from right? isn't that why the getNetworkSwitchName would fail?

go-controller/pkg/ovn/controller/networkconnect/controller.go
```diff
@@ -332,6 +417,144 @@ func (c *Controller) reconcileNode(key string) error {
...
+	defer func() {
+		klog.V(4).Infof("Reconciling service %s/%s for network connect took %v", namespace, name, time.Since(startTime))
+	}()
+
+	// Get the service
+	svc, err := c.serviceLister.Services(namespace).Get(name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	// Service may be nil if deleted (this shouldn't happen since we only process create events)
```
npinaeva: now I think it's "we don't process delete events"?

tssurya: yea :D oops fixing this

tssurya:
wow CI seems really red and for all unrelated reasons.. I'm gonna ignore this red run.. I will fix point1 i.e we decided to go with two pass ACLs and then address @npinaeva 's comment above and repush and we'll see how CI behaves

tssurya:
last two commits are the only changes compared to previous iteration of this PR, diff is https://github.com/ovn-kubernetes/ovn-kubernetes/compare/b51388125f1e08d873d8286795ce34d3dd9cc010..3545c2dff173c229b18e358182f9bb262f7c991f

tssurya:
@npinaeva I think we are really close here, the last 3 commits are the changes compared to your last review, PTAL
I hope CI is green this time, it was red yesterday due to a buggy code in services controller which is fixed as the top commit..

tssurya:
~omg how is a rebase causing such stupid CI failures.. I'm looking into https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22627600057/job/65568394415?pr=5828~ FIXED

tssurya:
I'll rebase on master since we merged https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5886  just to be in the clear of no failures..
no functional changes. only the last 3 commits are the new ones compared to our last rounds..

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -857,21 +857,10 @@ func (oc *EFController) addEgressFirewallRules(ef *egressFirewall, pgName string
 	return nil
 }
 
-// deleteStaleACLs cleans up 2 previous implementations:
-//   - Cleanup the old implementation (using LRP) in local GW mode
-//     For this it just deletes all LRP setup done for egress firewall
+// deleteStaleACLs cleans up a previous implementation detail:
```
npinaeva: since we are here, I think we can fully remove this function, it's been here since release 1.0, so we can safely remove it (who knows what else it could clean up wrong)

go-controller/pkg/ovn/controller/services/services_controller.go
```diff
@@ -602,10 +602,11 @@ func (c *Controller) skipService(name, namespace string) bool {
...
 		serviceNAD, err := c.networkManager.GetPrimaryNADForNamespace(namespace)
 		if err != nil {
-			// If the namespace requires a UDN that hasn't been processed yet, the default controller
-			// should skip this service; the UDN controller will handle it once ready.
+			// If the namespace's primary NAD state is unknown (e.g., NAD deleted during
+			// network recreation), all controllers must skip. The correct controller
+			// will process the service once the NAD is re-established and triggers a re-sync.
 			if util.IsInvalidPrimaryNetworkError(err) {
-				return c.netInfo.IsDefault()
+				return true
```
npinaeva, line 609: nice catch! and sorry for this hustle (I reviewed the original change, but it is very hard to catch these tiny things :))

tssurya: no problem! all good..

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/d5843bd973aa7b88bc75be9118c85fd113cb87d7..4a1797d30df78e8e706a7aeb0a3efc4efdb557cd

is the diff

I'll also now rebase on master while at it.

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
tssurya: done
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/d5843bd973aa7b88bc75be9118c85fd113cb87d7..4a1797d30df78e8e706a7aeb0a3efc4efdb557cd

tssurya:
diff is in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828/changes/84641f7e1a7ecb8542d2c281b12460ed61d6fd17

PR #5838, opened 2026-01-09
Reset MTU of both VF and VF-Reps when the Pod is gone

qiaofengq:
## 📑 Description

During pod creation, MTU may be set to a new value for VF and VF representor interfaces through the functions:
-  `(*defaultPodRequestInterfaceOps) ConfigureInterface`
- `(bnnc *BaseNodeNetworkController) addRepPort`

However, the corresponding functions for pod deletion do not change the MTU value back:
- `(*defaultPodRequestInterfaceOps) UnconfigureInterface`
- `(bnnc *BaseNodeNetworkController) delRepPort`

This commit adds an option in the config module as the default MTU value for VF. When a pod is removed, the corresponding functions will reset the MTU of both VF and VF representor to the default value.

## ✅ Checks

- [x] My code requires tests

## How to verify it

Minor updates in `go-controller/pkg/node/base_node_network_controller_dpu_test.go` to verify the MTU reset

## Summary by CodeRabbit

* **Bug Fixes**
  * MTU values now reliably reset to the default during SR-IOV/port cleanup and reconfiguration.
  * Improved error handling and logging for host-side network interface operations.

* **Tests**
  * Extended tests to verify MTU reset behavior during cleanup and failure recovery.

### DIFF
--- a/go-controller/pkg/cni/helper_linux.go
+++ b/go-controller/pkg/cni/helper_linux.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/knftables"
 
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
@@ -817,6 +818,10 @@ func (*defaultPodRequestInterfaceOps) UnconfigureInterface(pr *PodRequest, ifInf
 				if err != nil {
 					return fmt.Errorf("failed to bring down container interface %s %s: %v", pr.IfName, podDesc, err)
 				}
+				err = util.GetNetLinkOps().LinkSetMTU(link, config.DefaultVFMTU)
+				if err != nil {
+					return fmt.Errorf("failed to reset MTU on the container interface %s %s: %v", pr.IfName, podDesc, err)
+				}
 				// rename netdevice to make sure it is unique in the host namespace:
 				// if original name of netdevice is empty, sandbox id and a '0' letter prefix is used to make up the unique name.
 				oldName := ifInfo.NetdevName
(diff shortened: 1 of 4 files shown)


go-controller/pkg/node/base_node_network_controller_dpu.go
```diff
@@ -321,8 +321,11 @@ func (bnnc *BaseNodeNetworkController) delRepPort(pod *corev1.Pod, dpuCD *util.D
 	if err != nil {
 		klog.Warningf("Failed to get link device for representor port %s. %v", vfRepName, err)
 	} else {
-		if linkDownErr := util.GetNetLinkOps().LinkSetDown(link); linkDownErr != nil {
-			klog.Warningf("Failed to set link down for representor port %s. %v", vfRepName, linkDownErr)
+		if err = util.GetNetLinkOps().LinkSetDown(link); err != nil {
```
aserdean, line 324: I believe on the representor side we should always default to the OVS setting, in theory there is an option called mtu_request on the interface table of OVSDB.

This would align both userspace OVS and kernel OVS.

go-controller/pkg/cni/helper_linux.go
```diff
@@ -925,11 +930,21 @@ func (pr *PodRequest) deletePort(ifaceName, podNamespace, podName string) {
 		// DEL should be idempotent; don't return an error just log it
 		klog.Warningf("Failed to delete pod %q OVS port %s: %v\n  %q", podDesc, ifaceName, err, string(out))
 	}
-	// skip deleting representor ports
-	if pr.CNIConf.DeviceID == "" {
-		if err = util.LinkDelete(ifaceName); err != nil {
-			klog.Warningf("Failed to delete pod %q interface %s: %v", podDesc, ifaceName, err)
+
+	link, err := util.GetNetLinkOps().LinkByName(ifaceName)
```
aserdean, line 934: check for the err here and after the block from https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5838/changes#diff-01a2ef868473804cd3d9b894b093e5be0735cfb21bf27f375ece3a33cc07edc9R936-R945 to avoid nesting.

go-controller/pkg/node/base_node_network_controller_dpu.go
qiaofengq: Thank you very much for your review! To clarify, do you suggest setting `mtu_request` to `DefaultVFMTU` in OVSDB instead of using `LinkSetMTU()` in this `delRepPort` function? I am uncertain since this function also removes the port and interface from OVSDB right after setting MTU. Would `mtu_request` work in this case?

In addition, how about the function `addRepPort`, which is not changed in this PR? If I understand it correctly, it calls `cni.ConfigureOVS` which sets `mtu_request` only in userspace mode but not kernal mode. Should this function also be changed?

PR #5840, opened 2026-01-10
Bug fix: Skip BridgeEIPAddrManager in DPU-host mode

tsorya:
Bug fix: Skip BridgeEIPAddrManager in DPU-host mode
In DPU-host mode, BridgeEIPAddrManager should not be initialized
because EgressIP is handled by ovnkube running on the DPU where
OVS runs. The DPU-host has no OVS and no openflow manager.

This fix:
- Adds shouldHandleBridgeEgressIP() helper function that checks:
  - Network Segmentation is enabled
  - Interconnect is enabled
  - Gateway mode is not disabled
  - Not running in DPU-host mode
- Uses this helper in all EgressIP handler methods
- Removes unused egressip import from gateway_init.go

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Refactor**
  * Centralized guard for bridge EgressIP processing to consistently skip bridge-specific logic when not applicable, and ensure egress IP add/update/delete/sync operations trigger reconciliation when needed.

* **Bug Fixes**
  * Prevented bridge/OVS-related initialization during DPU-host pre-start so bridge-specific watchers and next-hop setup are not run inapplicable environments.

### DIFF
--- a/go-controller/pkg/node/gateway.go
+++ b/go-controller/pkg/node/gateway.go
@@ -234,8 +234,21 @@ func (g *gateway) DeleteEndpointSlice(epSlice *discovery.EndpointSlice) error {
 	return utilerrors.Join(errors...)
 }
 
+// canHandleBridgeEgressIP returns true if this node should handle EgressIP
+// configuration on the bridge. Returns false if:
+// - Network segmentation (UDN) is not enabled
+// - Interconnect is not enabled
+// - Gateway mode is disabled
+// - Running in DPU-host mode (EgressIP is handled by ovnkube on the DPU where OVS runs)
+func canHandleBridgeEgressIP() bool {
+	return util.IsNetworkSegmentationSupportEnabled() &&
+		config.OVNKubernetesFeature.EnableInterconnect &&
+		config.Gateway.Mode != config.GatewayModeDisabled &&
+		config.OvnKubeNode.Mode != types.NodeModeDPUHost
+}
+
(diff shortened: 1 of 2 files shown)


go-controller/pkg/node/gateway.go
```diff
@@ -234,8 +234,21 @@ func (g *gateway) DeleteEndpointSlice(epSlice *discovery.EndpointSlice) error {
 	return utilerrors.Join(errors...)
 }
 
+// shouldHandleBridgeEgressIP returns true if this node should handle EgressIP
```
aserdean: s/shouldHandleBridgeEgressIP/canHandleBridgeEgressIP/g

go-controller/pkg/node/gateway_init.go
```diff
@@ -475,6 +475,10 @@ func (nc *DefaultNodeNetworkController) initGatewayDPUHostPreStart(kubeNodeIP ne
 		return err
 	}
 
+	// In DPU-host mode, bridgeEIPAddrManager is not initialized because:
+	// - There's no OVS on the host (it runs on the DPU)
+	// - Traffic is offloaded to the DPU which handles EgressIP configuration
```
aserdean: I would omit the word offload just to not make confusion with hardware offload.

Maybe something like the following will work:
Traffic is handled on the DPU which has the EgressIP configuration?

PR #5844, opened 2026-01-12
Set the value of gateway representor port in DPU mode

qiaofengq:
## 📑 Description

During configuring the bridge ports in DPU mode, assign the host representor port to `gwIfaceRep` so that `GetGatewayIfaceRep()` returns the correct result.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected sequencing so DPU gateway interface representor is populated earlier during bridge port setup, improving DPU network initialization reliability.

* **Tests**
  * Added unit checks to validate the gateway bridge representor is set correctly in both standard and DPU gateway initialization paths.

### DIFF
--- a/go-controller/pkg/node/bridgeconfig/bridgeconfig.go
+++ b/go-controller/pkg/node/bridgeconfig/bridgeconfig.go
@@ -409,7 +409,7 @@ func (b *BridgeConfiguration) ConfigureBridgePorts() error {
 		if err != nil {
 			return err
 		}
-
+		b.gwIfaceRep = hostRep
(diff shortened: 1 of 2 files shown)


aserdean:
Thanks so much for fixing this. 

Do you mind folding in a test that checks: in DPU mode, after gateway init, GetGatewayIfaceRep() returns the expected host representor? You can add a small test helper (e.g. getGatewayBridge()) and assert this in the existing shareGatewayInterfaceDPUTest.

PR #5848, opened 2026-01-15
test: fix race condition in waitForRollout

pliurh:
Add a new parameter updateFunc and capture resource generation
before applying the update. This ensures the function properly waits
for the generation to increase after the update is applied to the API
server, fixing a race condition where the function might check the
resource status before the update takes effect.

Rename waitForRollout to updateAndWaitForRollout to better reflect
that the function both applies an update via the callback and waits
for the rollout to complete.

## 📑 Description

Fixes #

Fix CI flack for tests `Should validate flow data of br-int is sent to an external gateway`.

## ✅ Checks

- [ ] My code requires tests

## How to verify it

CI test `Should validate flow data of br-int is sent to an external gateway` can verify it.

## Summary by CodeRabbit

* **Tests**
  * Improved rollout verification: an update action is executed before polling, and polling now requires the resource generation to advance beyond its prior value before declaring success—reducing false positives.
  * Skips rollout wait when no environment changes are needed and emits clearer, generic diagnostic logging during waits.

### DIFF
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -657,11 +657,14 @@ func waitClusterHealthy(f *framework.Framework, numControlPlanePods int, control
 	})
 }
 
-// waitForRollout waits for the daemon set in a given namespace to be
+// updateAndWaitForRollout waits for the resource in a given namespace to be
 // successfully rolled out following an update.
 //
+// The updateFunc parameter is a callback that performs the update operation
+// (e.g., applying a new configuration).
+//
 // If allowedNotReadyNodes is -1, this method returns immediately without waiting.
-func waitForRollout(c kubernetes.Interface, ns string, resource string, allowedNotReadyNodes int32, timeout time.Duration) error {
+func updateAndWaitForRollout(c kubernetes.Interface, ns string, resource string, allowedNotReadyNodes int32, timeout time.Duration, updateFunc func()) error {
 	if allowedNotReadyNodes == -1 {
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/util.go
```diff
@@ -657,11 +657,11 @@ func waitClusterHealthy(f *framework.Framework, numControlPlanePods int, control
 	})
 }
 
-// waitForRollout waits for the daemon set in a given namespace to be
+// waitForRollout waits for the resource in a given namespace to be
 // successfully rolled out following an update.
 //
 // If allowedNotReadyNodes is -1, this method returns immediately without waiting.
-func waitForRollout(c kubernetes.Interface, ns string, resource string, allowedNotReadyNodes int32, timeout time.Duration) error {
+func waitForRollout(c kubernetes.Interface, ns string, resource string, allowedNotReadyNodes int32, timeout time.Duration, updateFunc func()) error {
```
npinaeva: should we call it updateAndWaitForRollout?
Also a little comment about the updateFunc would be nice

npinaeva:
i guess this is not a full fix for `e2e br-int flow monitoring export validation Should validate flow data of br-int is sent to an external gateway [It] with sflow` considering it just failed?

pliurh:
> i guess this is not a full fix for `e2e br-int flow monitoring export validation Should validate flow data of br-int is sent to an external gateway [It] with sflow` considering it just failed?

No, it's not. The flakes are caused by various reason. This PR only fix one of them.

npinaeva (approved):

Issue #5841, opened 2026-01-10
Fix BUILDER_IMAGE typo and split multi-arch builds across runners to avoid disk space issues


Soli0222:
## Summary

The current `docker.yml` workflow has two issues:
1. Incorrect `BUILDER_IMAGE` reference causing Fedora builds to fail (ref: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20863932033)
2. Multi-platform builds on a single runner causing disk space exhaustion (ref: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20444176734)

## Problem 1: Wrong BUILDER_IMAGE

The workflow currently uses an incorrect Golang image:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d0b56f50fe7e2021096310ba36896a824c71039e/.github/workflows/docker.yml#L17

However, this image is Debian-based. The Fedora Dockerfile expects `dnf` package manager, which results in:
```
/bin/sh: 1: dnf: not found
```

**Expected:** The image should be `quay.io/projectquay/golang:1.24` (Fedora-based), matching what's used in the Makefile:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d0b56f50fe7e2021096310ba36896a824c71039e/dist/images/Makefile#L31

The Makefile correctly passes this to the Dockerfile:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/d0b56f50fe7e2021096310ba36896a824c71039e/dist/images/Makefile#L55

## Problem 2: Disk Space Exhaustion on Single Runner

Building both `linux/amd64` and `linux/arm64` images on a single runner causes disk space issues. GitHub-hosted runners have limited disk space (~14GB free), and building multi-arch images with QEMU emulation consumes significant storage.

## Proposed Solution

1. **Fix the `BUILDER_IMAGE` typo**: Change from `quay.io/lib/golang:1.24` to `quay.io/projectquay/golang:1.24`

2. **Split builds across multiple runners** using the matrix strategy pattern from [Docker's official documentation](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners):
   - Build each platform (amd64, arm64) on separate runners in parallel
   - Upload digests as artifacts
   - Merge into a multi-platform manifest in a final job

### Benefits
- Fixes Fedora build failures
- Avoids disk space exhaustion
- Faster builds (parallel execution)
[...]

PR #5566, opened 2025-09-05
fix: multiarch ubuntu build

killianmuldoon:
fix the Ubuntu arm image build while keeping the existing multi-arch image target and Dockerfiles in place.

## 📑 Description

Part of #5504 

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## Summary by CodeRabbit

* **Chores**
  * Updated Docker image build configuration to use the repository root as the build context, ensuring all required resources are available during image creation.
  * No changes to application functionality; this improves build reliability and reproducibility of images.
  * CI/workflow behavior adjusted only for build context handling; runtime behavior remains unchanged.

PR #5850, opened 2026-01-15
parallelize multi-platform Docker image builds

Soli0222:
## 📑 Description

Fixes #

Improvements to the second issue proposed in #5841

### Problem : Multi-platform builds on a single runner causing disk space exhaustion (ref: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/20444176734)

Building both `linux/amd64` and `linux/arm64` images on a single runner causes disk space issues. GitHub-hosted runners have limited disk space (~14GB free), and building multi-arch images with QEMU emulation consumes significant storage.

### Proposed Solution

**Split builds across multiple runners** using the matrix strategy pattern from [Docker's official documentation](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners):
   - Build each platform (amd64, arm64) on separate runners in parallel
   - Upload digests as artifacts
   - Merge into a multi-platform manifest in a final job

### Proof of Concept

I have already implemented and tested these fixes in my fork:

- **Fixed workflow**: https://github.com/Soli0222/ovn-kubernetes/blob/master/.github/workflows/docker.yml
- **Successful workflow run**: https://github.com/Soli0222/ovn-kubernetes/actions/runs/20880137190
- **Published images**:
  - https://github.com/Soli0222/ovn-kubernetes/pkgs/container/ovn-kubernetes%2Fovn-kube-ubuntu
  - https://github.com/Soli0222/ovn-kubernetes/pkgs/container/ovn-kubernetes%2Fovn-kube-fedora

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Chores**
  * CI now runs separate per-platform Fedora and Ubuntu build jobs, enabling parallel amd64/arm64 image builds.
  * Per-platform caching, digest export/upload, and metadata propagation implemented to produce reliable digest-based multi-arch manifests.
  * Merge steps assemble and inspect multi-arch manifests from per-platform digests; global QEMU setup replaced by per-platform execution.
  * Workflow triggers and PR-run support improved for more consistent, faster publishes.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -21,7 +21,7 @@ jobs:
     name: Build Fedora (${{ matrix.platform }})
     runs-on: ${{ matrix.runner }}
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         include:
         - platform: linux/amd64
@@ -163,7 +163,7 @@ jobs:
     name: Build Ubuntu (${{ matrix.platform }})
     runs-on: ${{ matrix.runner }}
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         include:
         - platform: linux/amd64


tssurya:
@killianmuldoon please see if you can review this one

killianmuldoon:
Hi @Soli0222  - can you enable this workflow on PRs temporarily so we can see it working? We won't be able to push the images due to lack of permissions - but we'll see the build passing.

Soli0222:
Hi @killianmuldoon I've resolved the conflicts and enabled the workflow on PRs. You can see it running here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21136245530/job/60778854911?pr=5850

.github/workflows/docker.yml
```diff
@@ -16,31 +18,32 @@ env:
...
       uses: actions/checkout@v4
 
     - name: Set up Go
       uses: actions/setup-go@v5
       with:
         go-version-file: 'go-controller/go.mod'
-        # Disabling cache to avoid warnings until these two issues are fixed
-        # https://github.com/actions/setup-go/issues/424
-        # https://github.com/actions/setup-go/issues/403
-        # cache-dependency-path: "**/*.sum"
```
killianmuldoon, line 33: Should we remove these comments?

.github/workflows/docker.yml
```diff
@@ -16,31 +18,32 @@ env:
...
         cache: false
       id: go
 
-    - name: Log in to the GH Container registry
```
killianmuldoon, line 37: Why is this being moved around?

.github/workflows/docker.yml
```diff
@@ -64,48 +67,229 @@ jobs:
...
         popd 
-    
+
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v3
       with:
         platforms: all
 
     - name: Set up Docker Buildx
-      id: buildx
```
killianmuldoon, line 74: Why is this ID being removed?

.github/workflows/docker.yml
```diff
@@ -64,48 +67,229 @@ jobs:
...
+
+    - name: Create manifest list and push
+      working-directory: /tmp/digests
+      run: |
+        docker buildx imagetools create $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
+          $(printf '${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.REPOSITORY }}/${{ env.FEDORA_IMAGE_NAME }}@sha256:%s ' *)
+
+    - name: Inspect image
+      run: |
+        docker buildx imagetools inspect ${{ env.REGISTRY }}/${{ env.OWNER }}/${{ env.REPOSITORY }}/${{ env.FEDORA_IMAGE_NAME }}:${{ steps.meta.outputs.version }}
```
killianmuldoon, line 159: What does this inspect step do?

.github/workflows/docker.yml
```diff
@@ -64,48 +67,229 @@ jobs:
...
+
+  # Build Ubuntu image for each platform
+  build-ubuntu:
+    name: Build Ubuntu (${{ matrix.platform }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
killianmuldoon: Would fail fast be preferrable here? Do we want partial releases?

.github/workflows/docker.yml
```diff
@@ -64,48 +67,229 @@ jobs:
...
+      run: |
+        export GOPATH=$(go env GOPATH)
+        echo "GOPATH=$GOPATH" >> $GITHUB_ENV
+        echo "$GOPATH/bin" >> $GITHUB_PATH
+
+    - name: Build ovnkube-binaries copy to context 
+      run: | 
+        pushd go-controller 
+         make 
+        popd
```
killianmuldoon, line 210: In this case the build is always done on the matching architecture, right? So arm64 images are built on arm64 hardware?

killianmuldoon: It would be good to build the binaries inside the image so this isn't so much of the worry - I have a PR here that started that: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5566

killianmuldoon:
Thanks for fixing this - the approach generally looks good. I just have some questions about specific changes in there.

.github/workflows/docker.yml
Soli0222: Apologies, that was unintended. I have reverted the change in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5850/commits/a743199ed08f54731a3546e470fb410a5eb851df

.github/workflows/docker.yml
Soli0222: There was no specific reason for this. I have reverted it. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5850/commits/2e0a076778050f6a96aeb8643e248cfd738c7d71

.github/workflows/docker.yml
Soli0222: Since there are no multiple builders, removing the ID didn't actually cause any issues. However, as the original workflow explicitly specified the ID and builder settings, I have restored it in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5850/commits/c1cdf8c103b63d6c5c346dc4cb28e643e8803a0e .

.github/workflows/docker.yml
Soli0222: It's a sanity check to verify the manifest list. It provides visibility into which platforms are included in the pushed image. I can remove it if you think it's unnecessary.

example
```
Name:      ghcr.io/soli0222/ovn-kubernetes/ovn-kube-fedora:master
MediaType: application/vnd.oci.image.index.v1+json
Digest:    sha256:b6a51f9194383b720f4e90083630c8e20e21a9c5b23575498308b4d85ca6ff6d
           
Manifests: 
  Name:        ghcr.io/soli0222/ovn-kubernetes/ovn-kube-fedora:master@sha256:8526472ccd025155b519662ccbcd649668a2af0ba54fbe3b8a8c6110bb6bd05c
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    linux/arm64
               
  Name:        ghcr.io/soli0222/ovn-kubernetes/ovn-kube-fedora:master@sha256:f5732cc9622d422f913ebd56ff80978a4ab785849114943e1b50de14e2184049
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    unknown/unknown
  Annotations: 
    vnd.docker.reference.digest: sha256:8526472ccd025155b519662ccbcd649668a2af0ba54fbe3b8a8c6110bb6bd05c
    vnd.docker.reference.type:   attestation-manifest
               
  Name:        ghcr.io/soli0222/ovn-kubernetes/ovn-kube-fedora:master@sha256:e0f8b558479018b8a465a02763199d1850b8884b7c6754fb09625898a904829a
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    linux/amd64
               
  Name:        ghcr.io/soli0222/ovn-kubernetes/ovn-kube-fedora:master@sha256:14c8814984f2449380ba9d93b4013e85425461ccee966a7b016a660bd7877118
  MediaType:   application/vnd.oci.image.manifest.v1+json
  Platform:    unknown/unknown
  Annotations: 
    vnd.docker.reference.digest: sha256:e0f8b558479018b8a465a02763199d1850b8884b7c6754fb09625898a904829a
    vnd.docker.reference.type:   attestation-manifest
```

.github/workflows/docker.yml
Soli0222: Agreed, we definitely want to avoid partial releases. I've updated `fail-fast` to `true`. https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5850/commits/a20a530637452d15bf65c34baf6a0252b2b067d4

.github/workflows/docker.yml
Soli0222: Yes, updated to use native runners:

- `linux/amd64` → `ubuntu-latest`
- `linux/arm64` → `ubuntu-24.04-arm`

Each architecture is now built on matching hardware.

```yaml
  build-fedora:
    name: Build Fedora (${{ matrix.platform }})
    runs-on: ${{ matrix.runner }}
    strategy:
      fail-fast: true
      matrix:
        include:
        - platform: linux/amd64
          runner: ubuntu-latest
        - platform: linux/arm64
          runner: ubuntu-24.04-arm
```

Soli0222: Agreed, building binaries inside the image would be a cleaner solution.

Soli0222:
@killianmuldoon 
Thanks for the review and for approving the approach!
I've addressed all of your questions - could you take another look when you get a chance?

Regarding the workflow trigger, I'd like to restore the branch filters just before the final merge:
```yaml
on:
  push:
    # branches: [ master, release-1.0, release-1.1, release-1.2 ]
  pull_request:
    branches: [ master ]
```

.github/workflows/docker.yml
killianmuldoon: Definitely - but out of the critical path here.

killianmuldoon:
/lgtm pending reverting the branch filters

Thanks for working on this!

Soli0222:
Thanks for the review!

tssurya:
@Soli0222 could you please squash the commits in a logical way - example last revert commit should probably just end up removing the original commit where you added it

Soli0222:
@tssurya 
Thanks for the feedback! I've reorganized the commits:
- Squashed the core parallelization changes with their fixes
- Kept ARM64 native runners and fail-fast as separate commits since they're independent improvements
- Removed the PR trigger commit and its revert

.github/workflows/docker.yml
```diff
@@ -39,8 +53,8 @@ jobs:
       with:
         registry: ${{ env.REGISTRY }}
         username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
```
tssurya, line 42: I think this is unintended whitespace? can't see what changed
but we have added a newline as well?

.github/workflows/docker.yml
```diff
@@ -89,23 +99,201 @@ jobs:
...
+
+    - name: Upload digest
+      uses: actions/upload-artifact@v4
+      with:
+        name: digests-fedora-${{ env.PLATFORM_PAIR }}
+        path: /tmp/digests/*
+        if-no-files-found: error
+        retention-days: 1
+
+  # Merge Fedora multi-platform images
```
tssurya, line 122: does anything change here wrt to the packages we ship https://github.com/orgs/ovn-kubernetes/packages?repo_name=ovn-kubernetes for fedora/ubuntu distro wise?

tssurya: ah nvm I see your PR describe clearly
Screenshot from 2026-01-29 13-56-17
wonder what's the unknown/unknown tag but ok it exists the same way in master :) - I didn't know we already added image support for arm/amd but nice improvement here

.github/workflows/docker.yml
```diff
@@ -89,23 +99,201 @@ jobs:
...
+    name: Build Ubuntu (${{ matrix.platform }})
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+        - platform: linux/amd64
+          runner: ubuntu-latest
+        - platform: linux/arm64
+          runner: ubuntu-24.04-arm
```
tssurya, line 172: hmm how did you get these runner details from? I know docker says its from here https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners but hope CNCF org has these runners? I guess we do get these for free since we are public repo..

.github/workflows/docker.yml
Soli0222: The whitespace change is intentional. The original workflow file had trailing whitespace after the `password:` line.
I've cleaned up those unnecessary trailing spaces to align the formatting.

Also, I removed the unnecessary whitespace on the empty line (line 43) as well. No functional changes were made.

.github/workflows/docker.yml
Soli0222: Thanks for understanding from the PR description! 

The unknown/unknown tag you mentioned is actually an attestation manifest that Docker automatically includes when building images. It's used for supply chain security (SLSA provenance) and doesn't affect the actual image distribution. 

As you noted, this behavior already exists in the current master build, so no change there.

.github/workflows/docker.yml
Soli0222: Yes, according to the GitHub Actions documentation, the `ubuntu-24.04-arm` runners are currently available for free for **all public repositories**. Since this repository is public, we should have access to them regardless of the specific CNCF plan.
ref: https://docs.github.com/en/actions/reference/runners/github-hosted-runners

I recall seeing this runner being adopted in other CNCF projects/discussions as well, so it should be safe to use.

Soli0222:
@tssurya 
Thanks for the detailed review! I'm glad the overall logic looks good.
Looking forward to getting this merged.

tssurya:
> overall logic looks ok, thanks for the PR only concern I had is I feel we have duplicated the code a bit too much for the build and merge jobs/steps but its not a big deal

any thoughts on the duplicated code? any chance of consolidating those a bit better through common utils or do we want to look into that only as a FUP because of how the step registry yaml templating works?

Soli0222:
@tssurya 
Good point on the code duplication. There are a few options to consolidate:

1. Matrix consolidation: Merge build-fedora/build-ubuntu into a single job with distro in the matrix
2. Reusable workflows: Extract common logic into a callable workflow

I'd like to experiment with these approaches locally to find the cleanest solution before proposing changes. Since this would require some trial and error, I think it's better to handle as a follow-up PR rather than expanding the scope here.

Would you be okay with landing this as-is and tackling the consolidation in a separate PR?

PR #5858, opened 2026-01-19
Use `nft destroy` to simplify the UDN cleanup code

danwinship:
## 📑 Description
Requires a kernel/userspace that supports `nft destroy`, and then replaces the hacky UDN map cleanup code.

**This makes ovn-kubernetes require kernel 6.3 or later (or RHEL 9) and `nft` 1.0.8 or later; ie, something from July 2023 or newer.**

If that's a problem, I could also make it only use `tx.Destroy` when you're using UDNs (and so make it only require the new-ish stuff when you're using UDNs)?

## ✅ Checks
not sure if we document kernel/userspace requirements here or on the web site?

cc @kyrtapz since you added the "``TODO: Switch to `nft destroy` when supported``" that this PR removes...

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal network gateway rule management by consolidating deletion operations into a unified flow.
  * Optimized network rule deletion to use direct destruction operations instead of intermediate steps, improving efficiency.
  * Enhanced gateway initialization configuration for stricter operational requirements.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -827,38 +827,14 @@ func delServiceRules(service *corev1.Service, localEndpoints util.PortToLBEndpoi
 		}
 		nftElems := getGatewayNFTRules(service, localEndpoints, true)
 		nftElems = append(nftElems, getGatewayNFTRules(service, localEndpoints, false)...)
+		if util.IsNetworkSegmentationSupportEnabled() {
+			nftElems = append(nftElems, getUDNNFTRules(service, nil)...)
+		}
 		if len(nftElems) > 0 {
 			if err := nodenft.DeleteNFTElements(nftElems); err != nil {
 				err = fmt.Errorf("failed to delete nftables rules for service %s/%s: %v",
 					service.Namespace, service.Name, err)
 				errors = append(errors, err)
-			}
-		}
-
-		if util.IsNetworkSegmentationSupportEnabled() {
-			// NOTE: The code below is not using nodenft.DeleteNFTElements because it first adds elements
-			// before removing them, which fails for UDN NFT rules. These rules only have map keys,
-			// not key-value pairs, making it impossible to add.
-			// Attempt to delete the elements directly and handle the IsNotFound error.
-			//
-			// TODO: Switch to `nft destroy` when supported.
-			nftElems = getUDNNFTRules(service, nil)
-			if len(nftElems) > 0 {
-				nft, err := nodenft.GetNFTablesHelper()
-				if err != nil {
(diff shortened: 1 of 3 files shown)


kyrtapz:
The changes themselves look good to me.

> This makes ovn-kubernetes require kernel 6.3 or later (or RHEL 9) and nft 1.0.8 or later; ie, something from July 2023 or newer.
If that's a problem, I could also make it only use tx.Destroy when you're using UDNs (and so make it only require the new-ish stuff when you're using UDNs)?

I think we should agree to a minimum supported nft/kernel version and enforce/document it. wdyt? 
cc: @girishmg @npinaeva @tssurya @trozet

danwinship:
> This makes ovn-kubernetes require kernel 6.3 or later (or RHEL 9) and `nft` 1.0.8 or later; ie, something from July 2023 or newer.
> 
> If that's a problem, I could also make it only use `tx.Destroy` when you're using UDNs (and so make it only require the new-ish stuff when you're using UDNs)?

oh, I meant to add "or we could just not merge this, and leave it as a TODO" if we don't like the stricter version requirements even for UDNs.

npinaeva:
> I think we should agree to a minimum supported nft/kernel version and enforce/document it. wdyt? cc: @girishmg @npinaeva @tssurya @trozet

That makes sense! we already have a required minimal kernel version to make sure host isolation works, but it is not really documented anywhere and won't be a problem if UDNs are not used. I think the fix is f07e28e4c623168f9fa5c00f518bd341d4014aa6 which seems to be introduced in 6.6.53 https://cdn.kernel.org/pub//linux/kernel/v6.x/ChangeLog-6.6.53

danwinship:
Filed https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5873 about documenting required versions.

kyrtapz (approved):
LGTM
@npinaeva are we good here?

npinaeva:
considering UDN isolation won't work with older kernel anyway, I guess yes!
but the docs still need to be updated :D

danwinship:
> considering UDN isolation won't work with older kernel anyway, I guess yes!

Note that the PR as merged makes ovn-k unconditionally require `nft destroy` support. (It checks at startup time.) I had said I could change that, but nobody said I should.

> but the docs still need to be updated :D

#5873. I wasn't sure what to document.

npinaeva:
@danwinship apparently we broke d/s and decided that we generally want to support older kernels.
Current limitation is for UDN isolation only, so if someone's UDN is not using kubelet probes they could still use it.
I am not sure how much we could limit the Destroy support requirement in relation to the enabled features (and if it brings more order or chaos), lmk what you think

Issue #5823, opened 2025-12-23
ovn-kubernetes not ensuring network configuration state


aserdean:
### What happened?

After deploying ovnk8s in dpu-host mode the pod running on the host we get on the host PF the secondary 169 IP:

```
4: enp144s0f0np0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 9c:63:c0:13:f1:14 brd ff:ff:ff:ff:ff:ff
    inet 10.0.120.75/22 metric 100 brd 10.0.123.255 scope global dynamic enp144s0f0np0
       valid_lft 5721sec preferred_lft 5721sec
    inet 169.254.0.2/17 brd 169.254.127.255 scope global enp144s0f0np0
       valid_lft forever preferred_lft forever
```

After running `netplan apply` we loose it and cannot longer access the kubernetes svc ips.

We need to delete/recreate the ovnk8s pod on the host to recover.

### What did you expect to happen?

This is due to the fact is not re-ensuring the state it did during start: ip neighbours, management port (ovn-k8s-mp0), custom routes etc.

### How can we reproduce it (as minimally and precisely as possible)?

Deploy ovnkubernetes in host mode, remove network state done by it.

### OVN-Kubernetes version

all

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### OVN version

<details>

```console
$ oc rsh -n ovn-kubernetes ovnkube-node-xxxxx (pick any ovnkube-node pod on your cluster)
$ rpm -q ovn
# paste output here
```

</details>

### OVS version

<details>

```console
$ oc rsh -n ovn-kubernetes ovs-node-xxxxx (pick any ovs pod on your cluster)
$ rpm -q openvswitch
# paste output here
```

</details>

### Platform

<details>

Is it baremetal? GCP? AWS? Azure?
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Install tools

<details>

</details>

### Container runtime (CRI) and version (if applicable)

<details>

</details>

PR #5859, opened 2026-01-20
node: Recreate management port if deleted externally

tsorya:
When the management port interface (ovn-k8s-mp0) is deleted externally (e.g., by running 'netplan apply' on a DPU-host), the doReconcile() function would fail because the interface no longer exists. Previously, this would leave the node in a broken state requiring a pod restart to recover.

This change modifies the reconciliation loop to call create() when doReconcile() fails, which will recreate the management port interface along with its IP configuration and routes.

The create() functions for all management port types (OVS, netdev, representor) are idempotent - they use '--may-exist' flags for OVS commands and check existing state before making changes, so calling them multiple times is safe.

Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5823

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Management port reconciliation now attempts a safe recreation when reconciliation detects a missing interface; failures are logged and recreation errors are returned, while successful recreation suppresses the reconciliation error.

* **Tests**
  * Added tests covering management port startup and reconciliation flows, including create-on-start, recreate-on-reconcile-failure, and no-recreate-on-success behaviors.

### DIFF
--- a/go-controller/pkg/node/managementport/port.go
+++ b/go-controller/pkg/node/managementport/port.go
@@ -84,7 +84,16 @@ func start(mp managementPort, stopChan <-chan struct{}) (func(), error) {
 							return true
 						}
 					},
-					mp.doReconcile,
+					func() error {
+						if err := mp.doReconcile(); err != nil {
+							// doReconcile may fail if the interface was deleted.
+							// In that case, try to recreate it. create() is idempotent
+							// and safe to call even if the interface already exists.
+							klog.Errorf("Failed to reconcile management port, attempting to recreate: %v", err)
+							return mp.create()
+						}
+						return nil
+					},
 				)
 				if err != nil {
 					klog.Errorf("Failed to reconcile management port: %v", err)
--- a/go-controller/pkg/node/managementport/port_test.go
+++ b/go-controller/pkg/node/managementport/port_test.go
@@ -1,0 +1,153 @@
+package managementport
+
+import (
+	"errors"
+	"sync/atomic"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+// mockManagementPort is a mock implementation of managementPort interface for testing
+type mockManagementPort struct {
+	createFunc      func() error
+	doReconcileFunc func() error
+	period          time.Duration
+
+	createCalls      int32
+	doReconcileCalls int32
+}
+
+func (m *mockManagementPort) create() error {
+	atomic.AddInt32(&m.createCalls, 1)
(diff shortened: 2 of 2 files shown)


go-controller/pkg/node/managementport/port.go
```diff
@@ -84,7 +84,19 @@ func start(mp managementPort, stopChan <-chan struct{}) (func(), error) {
...
 					},
-					mp.doReconcile,
+					func() error {
+						if err := mp.doReconcile(); err != nil {
+							// doReconcile may fail if the interface was deleted.
+							// In that case, try to recreate it. create() is idempotent
+							// and safe to call even if the interface already exists.
+							klog.Warningf("Failed to reconcile management port, attempting to recreate: %v", err)
+							if createErr := mp.create(); createErr != nil {
+								klog.Warningf("Failed to recreate management port: %v", createErr)
```
pliurh: Can we return `createErr` here, instead of always returning the original error?

go-controller/pkg/node/managementport/port.go
```diff
@@ -84,7 +84,19 @@ func start(mp managementPort, stopChan <-chan struct{}) (func(), error) {
...
+							}
+							return err
```
pliurh: Do we need to return the original error when the `mp.create()` success?

go-controller/pkg/node/managementport/port_test.go
```diff
@@ -0,0 +1,154 @@
...
+			Expect(mp.getCreateCalls()).To(Equal(1))
+
+			// Trigger reconciliation
+			reconcile()
+
+			// Wait for doReconcile to be called
+			Eventually(doReconcileCalled, 1*time.Second).Should(Receive())
+
+			// Give some time for any potential create() calls
+			time.Sleep(100 * time.Millisecond)
```
pliurh: why not also use `Eventually()`?

go-controller/pkg/node/managementport/port.go
tsorya: Good question, 
i am returning an error here cause as there is a chance that error we got will not be fixed by create so we actually would like to retry it with backoff? As if create will succeed but error will be still there we will not use wait.Backoff logic at all.
But maybe if we have something that can't be fixed by create, it will fail create will too.
Not sure what's the best here

tsorya: Changed to what was proposed

tsorya:
@pliurh fixed your comments

PR #5861, opened 2026-01-20
Add OKEP for multiple ovn-kubernetes instances on a node

### LINKED ISSUE #5825: cannot run ovn-kubernetes in dpu mode and ovn-kubernetes in full mode on the same host
Creating this issue just to raise the fact that on a DPU we may want to run ovn-kubernetes in full mode besides just the ovn-kubernetes in DPU mode.

What is the usecase for the above:
Imagine we have a standard deployment for a node that has a DPU.

We assume that that deployment will have two Kuberentes clusters.

One for the Control plane and the Host that has a DPU, let's name this cluster A.
Another Kubernetes cluster,, cluster B, that has the DPU node inside and runs ovn-kuberentes in DPU mode on behalf of cluster A.

We may find ourselves in a situation in which we may want the primary CNI in cluster B to be ovn-kubernetes hence we will have two ovn-k stacks that will run side by side on the same node.

aserdean:
Introduce an enhancement proposal for running several ovn-kubernetes instances on
one node, each with its own OVN stack and its own Kubernetes cluster. The initial
use case is a DPU that runs one instance in DPU mode, serving pods of the host
cluster, and one in full mode, serving pods scheduled on the DPU itself.

Each instance owns its own northbound and southbound databases, so no OVN object
needs an instance-specific identifier. Only Open vSwitch resources and names in
the shared host namespaces have to be scoped: the integration bridge, the patch
and management ports, conntrack zones, file and database paths, and the metrics,
health check and readiness endpoints. OVN locates the configuration of a
co-hosted ovn-controller through external_ids suffixed with its chassis name,
which is the mechanism this design builds on and which has been available
upstream since OVN 23.03.

Signed-off-by: Alin Serdean <aserdean@nvidia.com>
Co-authored-by: Igal Tsoiref <itsoiref@redhat.com>
Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5825
POC: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5354

### DIFF
--- a/docs/okeps/okep-5861-two-ovnks.md
+++ b/docs/okeps/okep-5861-two-ovnks.md
@@ -1,0 +1,479 @@
+# OKEP-5861: Multiple OVN-Kubernetes Instances on a Node
+
+* Issue: [#5825](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5825)
+* Enhancement PR: [#5861](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5861),
+  which this OKEP is numbered after
+* POC: [#5354](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5354)
+
+## Problem Statement
+
+ovn-kubernetes cannot run more than one instance on the same node, so a DPU
+cannot simultaneously run an instance in DPU mode (serving pods of the host
+cluster) and an instance in full mode (serving pods running on the DPU itself).
+
+## Goals
+
+* Run multiple independent ovn-kubernetes instances on the same node, each with
+  its own OVN stack and its own Kubernetes cluster. The primary use case is two
+  instances on a DPU: one in DPU mode and one in full mode.
+* Give every instance unambiguous ownership of the host state it creates:
+  configuration, logs, sockets, PID files, OVN database files, OVS resources and
+  interface names.
+* Keep single-instance deployments byte-for-byte compatible: when no instance
+  identifier is configured, every path and name is what it is today.
+* Keep reconciliation and cleanup idempotent, so that an instance never adopts,
+  renames or deletes another instance's resources.
+
+## Non-Goals
+
+* Changing how interfaces are selected for offload. Whether a pod uses an
+  accelerated device or a veth remains a deployment decision, exactly as it is
+  today for a single instance.
+* Redesigning DPU support, or introducing a multi-tenancy API.
+* Providing an operator or any other lifecycle management framework to deploy
+  and upgrade the instances.
+* Migrating an existing instance between instance identifiers. Changing an
+  identifier changes the names of its host resources and its OVN chassis, so it
+  is a remove-and-recreate operation rather than an in-place upgrade.
+* Sharing an OVN stack, Kubernetes cluster, integration bridge or pod network.
+  The local Open vSwitch, host namespaces and physical hardware remain shared
+  and are partitioned as described below.
+
+## Introduction
+
+A DPU runs its own operating system and its own Kubernetes node, separate from
+the host it is plugged into. In the DPU (trusted) architecture described in
+[OKEP-5674](okep-5674-dpu-healthcheck.md), ovn-kubernetes runs in DPU mode on
+the DPU and in DPU-host mode on the host, both serving the *host* cluster.
+
+Some deployments additionally want to schedule pods on the DPU itself, as part
+of a second, DPU-local cluster. That requires a second, full-mode
+ovn-kubernetes instance on the same DPU, with its own OVN northbound and
+southbound databases and its own Kubernetes API server.
+
+Both instances share the local Open vSwitch installation, host namespaces and
+physical hardware. Their OVN stacks and Kubernetes clients are per instance.
+Upstream OVN already supports this topology:
+since OVN 23.03, `ovn-controller` can be co-hosted, meaning several
+`ovn-controller` instances with unique chassis names can use the same
+`ovs-vswitchd`, each driving its own integration bridge. This OKEP describes the
+changes ovn-kubernetes needs in order to use that capability.
+
+Note that nothing in this design is specific to a DPU or to a count of two: it
+generalizes to N instances on any node. DPUs with two instances are simply the
+use case that is being proposed, implemented and tested here.
+
+## User-Stories/Use-Cases
+
+Story 1: Run DPU-local workloads while offloading host workloads
+
+As a cluster admin deploying DPUs, I want to run an ovn-kubernetes instance in
+DPU mode that provides networking for pods on the host, and at the same time run
+an ovn-kubernetes instance in full mode that provides networking for pods
+scheduled on the DPU, so that I can use the DPU both as an offload engine for
+the host cluster and as a compute node of a separate cluster.
+
+## Proposed Solution
+
+Every instance is given an instance identifier. Its host state is derived from
+that identifier, except for the explicitly configured gateway bridge and
+conntrack zone range. Each instance runs its own `ovn-controller` against the
+shared Open vSwitch, using a unique chassis name and a dedicated integration
+bridge.
+
+```mermaid
+flowchart TB
+  subgraph DPU["DPU"]
+    direction TB
+    subgraph I1["instance id: host (DPU mode)"]
+      A1["ovnkube-node"]
+      C1["ovn-controller<br/>chassis: NODE-host"]
+      B1["br-int-host"]
+    end
+    subgraph I2["instance id: dpu (full mode)"]
+      A2["ovnkube-node"]
+      C2["ovn-controller<br/>chassis: NODE-dpu"]
+      B2["br-int-dpu"]
+    end
+    OVS["ovs-vswitchd + ovsdb-server<br/>(shared)"]
+    A1 --> B1
+    A2 --> B2
+    C1 --> B1
+    C2 --> B2
+    B1 --> OVS
+    B2 --> OVS
+  end
+  DB1["OVN NB/SB<br/>host cluster"]
+  DB2["OVN NB/SB<br/>DPU-local cluster"]
+  K1["Kubernetes API<br/>host cluster"]
+  K2["Kubernetes API<br/>DPU-local cluster"]
+  C1 --> DB1
+  C2 --> DB2
+  A1 --> DB1
+  A2 --> DB2
+  A1 --> K1
+  A2 --> K2
+```
+
+Because each instance has its own northbound and southbound databases, no OVN
+logical object needs an instance-specific identifier: an instance is the only
+writer of its own databases. Only resources in the shared Open vSwitch,
+including connection-tracking zone allocation, and names in the shared host
+namespaces (network interfaces, file paths, TCP ports), need to be scoped per
+instance.
+
+### API Details
+
+The instance identifier is the primary knob. The gateway bridge and conntrack
+zone range are configured explicitly because they own shared host resources:
+
+| Knob | Default | Meaning |
+| --- | --- | --- |
+| `--instance-id=STRING` | `""` | Short identifier of this instance. Empty selects today's behaviour exactly. |
+| `--ovn-chassis-name=STRING` | Existing chassis name if the identifier is empty; otherwise `NODE-INSTANCE_ID` | Optional chassis-name override, and therefore the `external_ids` suffix. |
+| `--bridge-name=STRING` | `br-int` when unset, otherwise `br-int-INSTANCE_ID` | Optional integration-bridge override. |
+| `--gateway-bridge=STRING` | Existing gateway bridge if the identifier is empty; otherwise required when gateway mode is enabled | Per-instance gateway bridge. |
+| `--ovn-ct-zone-range=MIN-MAX` | Required when the identifier is set | Non-overlapping OVN connection-tracking zone range. |
+
+The instance identifier must be a valid DNS label and short enough that every
+interface name derived from it fits in `IFNAMSIZ`, which is validated at startup
+(see [Naming constraints](#naming-constraints)). It is the value that appears in
+interface names, directory paths and metrics labels, so it is deliberately kept
+short and human readable, for example `host` for the DPU-mode instance serving
+the host cluster and `dpu` for the full-mode instance serving the DPU-local
+cluster.
+
+The chassis name is a separate, longer identity because it must be unique within
+its OVN southbound database, not merely on the node, and because OVN itself
+requires it as the suffix of the chassis-specific `external_ids`. Deriving it
+from the node name and the instance identifier keeps it unique by construction.
+It is passed to `ovn-controller` with the `-n` command line option, which takes
+precedence over the `system-id-override` file and over `external_ids:system-id`
+in the database. That option is needed because `external_ids:system-id` is the
+one option that cannot be made chassis-specific. In multi-instance mode,
+ovn-kubernetes uses `-n` and never writes this shared key.
+
+`--ovn-chassis-name` and `--bridge-name` already exist in the POC.
+`--instance-id` is the primary input: when the override flags are not set,
+ovn-kubernetes derives the chassis name and integration bridge from it.
+`--gateway-bridge` and `--ovn-ct-zone-range` remain explicit because they
+partition shared host resources.
+
+Startup rejects an effective chassis name, bridge, path, interface, port or
+gateway bridge that is already owned by another local instance.
+
+Admission and cleanup are serialized by a node-wide lock at
+`/var/run/ovn-kubernetes/instance-admission.lock`, acquired before validation
+and held until instance-owned resources are created. A rejected concurrent start
+creates no resources.
+
+### Implementation Details
+
+#### Instance-scoped host state
+
+When the instance identifier is set, it is appended to the directories that the
+instance owns. For an identifier of `host`:
+
+| State | Today | Instance `host` |
+| --- | --- | --- |
+| OVN database files | `/etc/ovn/` (`OVN_ETCDIR`) | `/etc/ovn/host/` |
+| OVN sockets and PID files | `/var/run/ovn/` (`run-dir`) | `/var/run/ovn/host/` |
+| OVN logs | `/var/log/ovn/` (`OVN_LOGDIR`) | `/var/log/ovn/host/` |
+| ovn-kubernetes logs | `/var/log/ovn-kubernetes/` | `/var/log/ovn-kubernetes/host/` |
+| CNI server socket and certificates | `/var/run/ovn-kubernetes/` | `/var/run/ovn-kubernetes/host/` |
+| ovn-kubernetes configuration file | `/etc/openvswitch/ovn_k8s.conf` (`--config-file`) | one file per instance, already selectable |
+| Shared OVS sockets and PID files | `/var/run/openvswitch/` (`OVS_RUNDIR`) | unchanged, shared by every instance |
+
+A separate OVN database directory is required because each instance runs its own
+northbound and southbound databases, and `ovsdb-server` would otherwise reuse
+the same files, sockets and PID files. The database, run and log directories of
+OVN itself are selected by `OVN_ETCDIR`, `OVN_RUNDIR` and `OVN_LOGDIR` in the
+image scripts, and the northbound and southbound run directories are already
+ovn-kubernetes config knobs (`run-dir`), so the sockets need no new API, only a
+per-instance value.
+
+The run directory of the shared Open vSwitch is deliberately not scoped:
+`ovs-vswitchd` and `ovsdb-server` are shared, so every instance keeps talking to
+them through `/var/run/openvswitch/`.
+
+Each instance talks to its own Kubernetes API server, using its own
+kubeconfig or its own service account token, mounted only into its own pods.
+Each instance also installs its own CNI configuration and CNI server socket in
+its own directory; the CNI configuration directory read by a given kubelet
+must only ever contain the configuration of the instance that serves that
+kubelet's cluster.
+
+#### OVS configuration
+
+The instances share one `ovsdb-server`, and therefore one `Open_vSwitch` table
+in the `Open_vSwitch` database. `ovn-controller` reads most of its configuration
+from `external_ids` in that table, and supports a chassis-specific form of most
+such options: `external_ids:<option>-<chassis-name>` overrides
+`external_ids:<option>`. The instance is responsible for writing the
+chassis-specific keys for its own chassis, and must not write the unsuffixed
+ones when an instance identifier is set. With an empty identifier, existing
+unsuffixed keys are preserved. For chassis `NODE-host`:
+
+```text
+external_ids:ovn-remote-NODE-host="ssl:10.0.0.1:9642"
+external_ids:ovn-encap-type-NODE-host="geneve"
+external_ids:ovn-encap-ip-NODE-host="10.0.0.5"
+external_ids:ovn-bridge-NODE-host="br-int-host"
+external_ids:ovn-bridge-datapath-type-NODE-host="netdev"
+```
+
+Note that the suffix is the *chassis name*, not the instance identifier: the
+suffix is how `ovn-controller` finds its own configuration, and it looks it up
+by the chassis name it was started with.
+
+Only an allowlist of OVN-supported chassis-specific keys is suffixed: the five
+keys above, `ovn-bridge-mappings`, the probe and flow-cache options,
+`ovn-is-interconn`, `ovn-monitor-all`, and `ovn-pf-encap-ip-mapping`.
+`external_ids:system-id` is never suffixed or updated in multi-instance mode.
+`ovn-set-local-ip` was removed in OVN 24.09.
+
+`ovn-bridge-mappings` maps physical network names to local bridges and therefore
+decides which bridge each `ovn-controller` patches; it is written unsuffixed
+today by ovn-kubernetes, so two instances would overwrite each other's mappings.
+OVN reads it through the chassis-specific lookup, so the suffixed form works.
+
+Each instance must use its own integration bridge, named after the instance
+(`br-int-host`, `br-int-dpu`). Sharing one integration bridge between
+co-hosted controllers is not supported in upstream OVN.
+
+Not every co-hosting key is an `external_ids` key. `ovn-controller` keeps a
+per-chassis index in `other_config:ovn-chassis-idx-<chassis-name>` of the same
+table, which it allocates itself: the first controller on the host gets an empty
+index, the next gets `0`, and tunnel ports are named `ovn<idx>-<chassis>-<hex>`
+so that co-hosted controllers cannot collide on interface names. ovn-kubernetes
+neither sets nor needs this key, but cleanup must not delete another chassis's
+entry, and anything that matches tunnel port names has to accept the indexed
+form.
+
+Each instance sets `--ovn-ct-zone-range` to a non-overlapping range, so
+controllers cannot dynamically pick the same zone for different logical ports.
+Startup validates the range bounds and checks it against ranges owned by local
+instances. This key is chassis-specific:
+
+```text
+external_ids:ct-zone-range-NODE-host="1-32767"
+external_ids:ct-zone-range-NODE-dpu="32768-65535"
+```
+
+Support for these boundaries was added in OVN 24.09.
+
+#### Gateway bridge and uplink
+
+Each instance needs its own `--gateway-bridge`, because ovnkube-node owns the
+whole flow table of a gateway bridge rather than a subset of it: the OpenFlow
+manager re-synchronizes it with `ovs-ofctl --bundle replace-flows`, and gateway
+cleanup replaces every flow with a single `NORMAL` action. Two instances pointed
+at one bridge would erase each other's flows on every synchronization.
+
+The existing `--gateway-interface` selects the uplink. It must be distinct per
+instance, or be a distinct VLAN subinterface when one physical port is shared.
+Startup validates the resolved gateway bridge and uplink ownership. Each
+instance then publishes its own physical network mapping through its
+chassis-specific `ovn-bridge-mappings`, so that its `ovn-controller` patches the
+right bridge.
+
+Sharing a single gateway bridge would require flow ownership inside the bridge,
+which the OpenFlow manager does not have today, and is out of scope here.
+
+#### Instance-scoped OVS and interface names
+
+Several names are currently hardcoded and have to be derived from the instance
+identifier instead. Each of these is a resource that two instances would
+otherwise create, reconfigure and delete underneath each other:
+
+* Integration bridge: `br-int` today, `br-int-<instance-id>` when an instance
+  identifier is set. Every place that assumes the literal `br-int` must use the
+  configured bridge name instead, including `ovs-appctl` and `ovs-vsctl`
+  invocations, the hybrid overlay, and the sample decoder.
+* Gateway and localnet patch ports: `GetPatchPortName()` builds the name from
+  `types.PatchPortSuffix`, which is the literal `-to-br-int`, and the localnet
+  port lookup makes the same assumption. Both must use the configured
+  integration bridge name, otherwise the two instances generate identical patch
+  port names and each one deletes the other's ports during reconciliation.
+* Management port: `types.K8sMgmtIntfName` is the literal `ovn-k8s-mp0`, and
+  both instances would otherwise create, rename, add routes and rules for, and
+  clean up the same interface. User defined networks add further management
+  interfaces built from `types.K8sMgmtIntfNamePrefix`, so the whole family has
+  to be scoped, and `types.SFlowAgent` hardcodes the same name and must follow.
+  See [Naming constraints](#naming-constraints) for the naming scheme, which is
+  constrained by `IFNAMSIZ`.
+* Any other OVS resource created by an instance must carry an ownership
+  `external_ids` entry naming the owning instance, and every lookup of such a
+  resource must filter on it. The POC does this for pod interfaces with
+  `external_ids:bridge-name`, which is the pattern to generalize.
+
+Reconciliation and cleanup must select resources by that ownership entry.
+Resources belonging to another instance are ignored; resources belonging to no
+instance are reported and left in place rather than adopted, so that starting,
+restarting and removing an instance is idempotent and cannot disrupt a
+co-hosted instance.
+
+#### Health checks and metrics
+
+The following listeners and probes bind to shared resources and must be
+configured per instance:
+
+* Listeners: controller, node, cluster-manager and OVN metrics, the OVS
+  exporter and optional healthz use unique per-instance ports, validated at
+  startup. Metrics carry the instance identifier as a label.
+* EgressIP node reachability: `--egressip-node-healthcheck-port` must differ
+  per instance.
+* Readiness: the `ovn-kube-util readiness-probe` command addresses
+  `ovn-controller` by its bare target name, so with two co-hosted controllers it
+  reaches whichever one owns the default control socket, and its `ovnkube-node`
+  target stats the literal `/etc/cni/net.d/10-ovn-kubernetes.conf`. Both must
+  become instance aware, so that a probe cannot report an instance healthy
+  because a co-hosted instance is running. Its checks against the shared
+  `ovsdb-server` and `ovs-vswitchd` sockets stay as they are, since those
+  daemons are shared.
+
+Logs are written to the instance-specific log directory and include the
+instance identifier, so that a centralized logging stack can separate the
+instances.
+
+#### Naming constraints
+
+Linux limits interface names to 15 characters plus a terminating NUL
+(`IFNAMSIZ` is 16), and this applies to the *values* of the names above, not to
+the `external_ids` keys. The constraint binds quickly. `br-int-host` fits at 11
+characters, but the management port does not: `ovn-k8s-mp0` is already 11
+characters, and appending `-host` needs 16.
+
+The management port family is therefore renamed rather than suffixed when an
+instance identifier is set, replacing the fixed `ovn-k8s-` prefix:
+
+```text
+instance id unset:  ovn-k8s-mp0      ovn-k8s-mp1 ...   (unchanged)
+instance id "host": ovnk-host-mp0    ovnk-host-mp1 ...
+```
+
+This leaves `15 - len("ovnk--mp") - digits` characters for the identifier. With
+up to three digits of user defined network index, that is four characters, so
+`host` and `dpu` fit and `local` does not. The identifier is validated at
+startup against every name template it feeds, and startup fails with a clear
+error naming the template that overflowed, rather than producing a truncated or
+colliding interface name.
+
+### Testing Details
+
+* Unit tests: name generation and validation for every instance-scoped name
+  (integration bridge, patch ports, management ports, paths, chassis name),
+  including the rejection of identifiers that overflow `IFNAMSIZ`, and the
+  empty-identifier case that must reproduce today's names exactly. Validate
+  gateway-bridge, uplink and connection-tracking range collisions, including
+  concurrent admission and cleanup, and verify multi-instance mode does not
+  write `external_ids:system-id`.
+* E2E tests: the existing `kind-dpu-offload` job already builds the two clusters
+  this OKEP needs, since `dpu-simulator` brings up a host cluster and a
+  DPU cluster as separate kind clusters with separate kubeconfigs. Extend it to
+  run the DPU-mode instance and the full-mode instance on the simulated DPU
+  node, and verify pod-to-pod, pod-to-service and egress traffic independently
+  in IPv4-only, IPv6-only and dual-stack clusters. This keeps the motivating
+  use case testable without DPU hardware.
+* Isolation tests: each instance only sees its own configuration, sockets, PID
+  files, database files, CNI configuration, integration bridge, patch ports,
+  management ports and conntrack zones; restarting or deleting one instance
+  leaves the other's resources and datapath untouched; a stale resource from a
+  removed instance is not adopted by the surviving instance. Cover three
+  instances to verify non-overlapping connection-tracking ranges.
+* Cross feature tests: network policy, services, and EgressIP in each instance,
+  since these depend on stateful conntrack behaviour that is affected by
+  co-hosting (see [Risks, Known Limitations and
+  Mitigations](#risks-known-limitations-and-mitigations)).
+* Scale tests: measure per-instance consumption and contention in shared OVS,
+  host and hardware resources to define a supported instance limit.
+
+### Documentation Details
+
+* A new page under the DPU documentation describing the architecture above,
+  with example DaemonSets for the two instances showing the instance
+  identifier, the separate credentials, host paths, integration bridges,
+  metrics ports and readiness probes, plus the OVS commands to inspect the
+  chassis-specific `external_ids`.
+* An entry for this OKEP in `mkdocs.yml`.
+
+## Risks, Known Limitations and Mitigations
+
+* OVN 24.09 documents co-hosting `ovn-controller` instances as experimental
+  and does not support stateful ACLs. OVN 26.09.0 removes the experimental
+  status and supports stateful features when controllers use separate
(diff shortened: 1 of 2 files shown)


docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
+OKEP-5861: Dual OVN-Kubernetes Instances on DPU Nodes
+
+Problem Statement
+
+ovn-kubernetes cannot currently run one instance in DPU mode and another in full mode on the same node.
+
+This blocks use cases where a DPU card needs to run two ovn-kubernetes instances simultaneously:
+ - one in DPU mode to manage networking for pods on the host (which belongs to a separate Kubernetes cluster)
+ - another in full mode for pods running locally on the DPU itself.
```
npinaeva: this looks more lika a part of the design to me, as it doesn't explain the use cases for full and dpu mode at the same time and doesn't mention other parts of this design, e.g. how to select which interfaces should be wired locally and which ones should be offloaded. If there is another okep explaining the larger use case, we should link it, or maybe extend this one to cover the full use case?

aserdean: > as it doesn't explain the use cases for full and dpu mode at the same time 

This behavior is described in detail in the enhancement proposal here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5861/files#diff-48fa1dd6b477df4d3a1167ec7881de661b28ff64aed0636bf15d70a15c800f8cR37-R80

> e.g. how to select which interfaces should be wired locally and which ones should be offloaded

That is not a goal of this proposal, as we are not redesigning the existing support model. If users choose to use accelerated device or veth interfaces on a node, that remains their decision, even when the node is a DPU node.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+
+When `INSTANCE_ID` is unset or empty, paths remain unchanged (e.g., `/etc/ovn-kubernetes/`), preserving backward compatibility with single-instance deployments.
+
+Deploy two DaemonSets on DPU nodes:
+
+ovnkube-node-dpu: OVN_K8S_NODE_MODE=dpu, OVN_K8S_INSTANCE_ID=dpu
+
+ovnkube-node-full: OVN_K8S_NODE_MODE=full, OVN_K8S_INSTANCE_ID=dpu-full
+
+Separate ownership of OVS/OVN objects via naming or external_ids, so each instance reconciles only its own ports and logical ports.
```
dceara: I might have missed it or it's just implied from the goal statement at the top of the document (`Support two independent ovn-kubernetes instances on the same node`) but I'm assuming this enhancement is about running two (well multiple) full ovn-k **stacks**.  That is, we'd have separate OVN NB and SB databases for each instance, right?

In that case, there are no OVN objects that need updated external_ids.  Each ovn-kube manages its own NB database, others won't touch it.  The only things that need to be tagged with an instance specific identifier are OVS objects.

Or am I missing something?

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+​
+
+Avoid conflicts over configuration paths, sockets, and OVS/OVN resources.
+
+Keep single-instance deployments fully compatible.
+
+
+Non-Goals
+
+Running more than two instances per node.
```
dceara: It's a non-goal, but essentially it would also work. :)

docs/okeps/okep-5861-two-ovnks.md
aserdean: >but I'm assuming this enhancement is about running two (well multiple) full ovn-k stacks. That is, we'd have separate OVN NB and SB databases for each instance, right?

correct.

> In that case, there are no OVN objects that need updated external_ids. Each ovn-kube manages its own NB database, others won't touch it. The only things that need to be tagged with an instance specific identifier are OVS objects.
>
>Or am I missing something?

also correct

The phrase ‘OVS/OVN objects’ was a mental slip; the text should refer only to OVS objects.
will remove the OVN bits from there in the next incremental.

docs/okeps/okep-5861-two-ovnks.md
aserdean: indeed :D

docs/okeps/okep-5861-two-ovnks.md
bn222: > If users choose to use accelerated device or veth interfaces on a node, that remains their decision, even when the node is a DPU node.

This is key, and I agree completly. If you muddle the two together, we're back in the same mess we are today: development is cumbersome because you _need_ a DPU (the hardware) to be able to do this work.

docs/okeps/okep-5861-two-ovnks.md
bn222: Can we add a diagram of all the components of the stack: ovn-k, nb, sb, controller, ovs? It would help to see what is shared and what is not when we run two ovn-k instances.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+
+Introduce an instance identifier (e.g., `OVN_K8S_INSTANCE_ID`) to isolate each instance's file paths and avoid conflicts:
+
+Config: /etc/ovn-kubernetes/$INSTANCE_ID
+
+Runtime: /var/run/ovn-kubernetes/$INSTANCE_ID
+
+Logs: /var/log/ovn-kubernetes/$INSTANCE_ID
+
+When `INSTANCE_ID` is unset or empty, paths remain unchanged (e.g., `/etc/ovn-kubernetes/`), preserving backward compatibility with single-instance deployments.
```
wizhaoredhat: Wondering how to make this idempotent, we should make sure to cleanup. Or should we let it become stale?

wizhaoredhat: Same with the external-ids for OvS

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
+OKEP-5861: Dual OVN-Kubernetes Instances on DPU Nodes
```
shettyg: I took this OKEP and the accompanying PR as a baseline to bring up 3 OVN IC stacks on a DPU. I had to make some additional changes and this review is based on my experience doing the POC.

Upstream OVN currently supports running multiple stacks on the same node. This OKEP specifically calls out DPUs and also reduces the scope of the number of stacks to just 2. I think, it may not really be needed for that specific call out and the solution as you are proposing can handle more than 2 stacks and on machines other than DPUs too.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+High-Level Design
+
+Introduce an instance identifier (e.g., `OVN_K8S_INSTANCE_ID`) to isolate each instance's file paths and avoid conflicts:
+
+Config: /etc/ovn-kubernetes/$INSTANCE_ID
+
+Runtime: /var/run/ovn-kubernetes/$INSTANCE_ID
+
+Logs: /var/log/ovn-kubernetes/$INSTANCE_ID
+
```
shettyg: We will also need a separate directory for the OVN databases (assuming multiple OVN IC stacks)

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+If unset, behavior and paths remain as today for backward compatibility.
+
+If set, all file paths, sockets, and PID files are derived from the instance ID.
+​
+
+OVS/OVN 
+
+Shared OVS on the DPU may serve both instances.
+
+The relevant bits for OVN inside the OVSDB are the following external_ids:
```
shettyg: "OVSDB" is a little confusing here. "Open_vSwitch database's Open_vSwitch table" may be more correct.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+
+```
+external_ids:ovn-remote -> external_ids:ovn-remote-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-encap-type -> external_ids:ovn-encap-type-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-encap-ip -> external_ids:ovn-encap-ip-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-bridge -> external_ids:ovn-bridge-<OVN_K8S_INSTANCE_ID>=
+external-ids:ovn-bridge-datapath-type -> external_ids:ovn-bridge-datapath-type-<OVN_K8S_INSTANCE_ID>=
+```
```
shettyg, line 116: The recent version of OVN added the support for "ct-zone-range" to have different conntrack zones for each stack

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+
+Each instance will generate the right values. To note here the ovn-bridge-<OVN_K8S_INSTANCE_ID>=, should have a value of max 15 chars.
+
+The `OVN_K8S_INSTANCE_ID` will also be sent to ovn-controller via the argument:
+`--ovn-controller-system-id=STRING          set chassis name (system-id) for ovn-controller (passed as -n)` once the OVN stack will be brought up.
+
+
```
shettyg: While doing the POC, I had to make a few more changes and I am calling it out here so that you can consider adding them to the OKEP.

* The patch port names generated to connect integration bridge to physical bridge are hardcoded in the code base. So that has to be handled.

The proposed per-instance ovn-bridge external ID is not sufficient by itself. ovn-kubernetes also constructs and discovers gateway/localnet patch-port names containing the integration bridge name. The POC had to replace hard-coded -to-br-int logic in GetPatchPortName() and localnetPortInfo() with the configured integration bridge.

* Distinct management-port names via `OVNKUBE_NODE_MGMT_PORT_NAME_SUFFIX`

When you have multiple stacks, each stack needs to have its own management interface names ("ovn-k8s-mp0"). Both node stacks otherwise create, rename, configure routes/rules for, and clean up `ovn-k8s-mp0`. A suffix per stack may be the right path because UDNs create additional mgmt interfaces

* ovn-kube-util readiness-probe needs changes

Since we have multiple OVN stacks with their sockets in stack specific paths, the readiness-probe needs to know which stack it needs to health check.

* Metrics port collision 

Two ovn-kubernetes stacks will have the same metrics port. Probably same change for egressIP health?

* Accompanying PR changes

The accompanying PR makes changes to ovn-kubernetes to handle different encap-ips, integration bridge etc. I think the OKEP should mention it.

shettyg (commented):
I took this OKEP and the accompanying PR as a baseline to bring up 3 OVN IC stacks on a DPU. I had to make some additional changes and this review is based on my experience doing the POC.

docs/okeps/okep-5861-two-ovnks.md
aserdean: The OKEP now has an Introduction and User-Stories section covering why both modes run at once, and it links OKEP-5674 for the DPU architecture it builds on. Interface selection stays an explicit non-goal: whether a pod uses an accelerated device or a veth is the same deployment decision it is today for a single instance, and coupling the two would mean you need DPU hardware to work on this at all.

docs/okeps/okep-5861-two-ovnks.md
aserdean: The document is now titled for multiple instances on a node and states that the mechanism generalizes to N instances, with two on a DPU as what is implemented and tested first.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Added. The diagram shows per-instance ovnkube-node, ovn-controller, integration bridge, NB/SB databases and Kubernetes API, against the single shared ovs-vswitchd and ovsdb-server, so what is shared and what is not is explicit.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Every OVS resource an instance creates carries an ownership external_ids entry, and every lookup filters on it. Resources owned by another instance are ignored, and unowned resources are reported and left in place rather than adopted, so nothing goes stale silently and no instance deletes another's ports. This generalizes the external_ids:bridge-name tag the POC already sets on pod interfaces, and the isolation tests cover restart and delete of one instance.

docs/okeps/okep-5861-two-ovnks.md
aserdean: added /var/lib/ovn-kubernetes/<instance-id>/ per instance, with the rationale in the doc — every instance runs its own NB and SB, including raft state under interconnect, so ovsdb-server would otherwise reuse the same files, sockets and PID files.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Updated the design doc with the five requested changes:

Use the configured integration bridge when building patch port names.

Scope management ports, readiness probes, metrics, and the EgressIP health-check port per instance.

Link the POC from the document header and references.

One open point is management-port naming. ovn-k8s-mp0 leaves little room within the 15-character interface-name limit, so the doc proposes ovnk-<id>-mp<n>. The POC’s ovn-k8s-mp10 naming is included as an alternative.

Also called out the main risk: co-hosted ovn-controller is still experimental and does not support stateful ACLs, which network policy relies on.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Rescoped exactly as you suggest: retitled "Multiple OVN-Kubernetes Instances on a Node", the "no more than two instances" non-goal is gone, and the design now generalizes to N instances on any node, with two on a DPU as what gets implemented and tested first. Your 3-stack POC is what convinced me to drop the DPU-only framing. (I posted this on the OVN database directory thread by mistake it belongs here.)

docs/okeps/okep-5861-two-ovnks.md
aserdean: fixed it now reads "the Open_vSwitch database's Open_vSwitch table".

docs/okeps/okep-5861-two-ovnks.md
aserdean: Added, with a per-chassis example. I checked controller/ct-zone.c on OVN main: the key is read through get_chassis_external_id_value, so external_ids:ct-zone-range-<chassis> is honoured per stack. The boundaries landed in OVN 24.09, which is now the minimum OVN version in the version skew section.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+external_ids:ovn-encap-ip -> external_ids:ovn-encap-ip-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-bridge -> external_ids:ovn-bridge-<OVN_K8S_INSTANCE_ID>=
+external-ids:ovn-bridge-datapath-type -> external_ids:ovn-bridge-datapath-type-<OVN_K8S_INSTANCE_ID>=
+```
+
+Each instance will generate the right values. To note here the ovn-bridge-<OVN_K8S_INSTANCE_ID>=, should have a value of max 15 chars.
+
+The `OVN_K8S_INSTANCE_ID` will also be sent to ovn-controller via the argument:
+`--ovn-controller-system-id=STRING          set chassis name (system-id) for ovn-controller (passed as -n)` once the OVN stack will be brought up.
+
```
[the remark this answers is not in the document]
aserdean: Neither suggested option is needed: OVN has supported chassis-specific external_ids since 23.03, so no ovn-controller patch is required. The correction is that the suffix must be the chassis name, not an arbitrary instance id, because that is the key ovn-controller looks itself up by; --ovn-controller-system-id sets it via -n. The OKEP is rewritten accordingly and now records the upstream caveat that co-hosted controllers are experimental and do not support stateful ACLs.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+Extend DPU documentation with:
+
+Dual-instance overview.
+
+Example DaemonSets and CNI snippets.
+​
+
+Add CI/e2e coverage for:
+
+A node running two ovn-kubernetes instances side by side (both in full mode), verifying pod connectivity for each instance independently.
```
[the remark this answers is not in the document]
aserdean: All covered in the rewrite: CNI configuration separation, per-instance kubeconfig or service account, per-instance metrics ports with an instance label, instance-scoped logs, readiness probes, and a Risks section listing the shared host resources (nftables/iptables, routing table IDs, IPAM) that must be partitioned before a second instance is admitted.

docs/okeps/okep-5861-two-ovnks.md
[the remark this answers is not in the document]
aserdean: Fixed in the rewrite — formatting and zero-width characters cleaned up, fenced blocks have a language, the typo is gone, the IFNAMSIZ limit is stated as applying to the name values with the arithmetic shown, and the test plan covers DPU mode plus full mode.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
+OKEP-5861: Dual OVN-Kubernetes Instances on DPU Nodes
+
+Problem Statement
+
+ovn-kubernetes cannot currently run one instance in DPU mode and another in full mode on the same node.
+
+This blocks use cases where a DPU card needs to run two ovn-kubernetes instances simultaneously:
+ - one in DPU mode to manage networking for pods on the host (which belongs to a separate Kubernetes cluster)
+ - another in full mode for pods running locally on the DPU itself.
+​
```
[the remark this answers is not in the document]
aserdean: Fixed in the rewrite — formatting and zero-width characters cleaned up, fenced blocks have a language, the typo is gone, the IFNAMSIZ limit is stated as applying to the name values with the arithmetic shown, and the test plan covers DPU mode plus full mode.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,99 @@
...
+
+```
+external_ids:ovn-remote -> external_ids:ovn-remote-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-encap-type -> external_ids:ovn-encap-type-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-encap-ip -> external_ids:ovn-encap-ip-<OVN_K8S_INSTANCE_ID>=
+external_ids:ovn-bridge -> external_ids:ovn-bridge-<OVN_K8S_INSTANCE_ID>=
+external-ids:ovn-bridge-datapath-type -> external_ids:ovn-bridge-datapath-type-<OVN_K8S_INSTANCE_ID>=
+```
+
+Each instance will generate the right values. To note here the ovn-bridge-<OVN_K8S_INSTANCE_ID>=, should have a value of max 15 chars.
```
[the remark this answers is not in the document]
aserdean: Fixed in the rewrite — formatting and zero-width characters cleaned up, fenced blocks have a language, the typo is gone, the IFNAMSIZ limit is stated as applying to the name values with the arithmetic shown, and the test plan covers DPU mode plus full mode

docs/okeps/okep-5861-two-ovnks.md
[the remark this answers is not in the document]
aserdean: Fixed in the rewrite — formatting and zero-width characters cleaned up, fenced blocks have a language, the typo is gone, the IFNAMSIZ limit is stated as applying to the name values with the arithmetic shown, and the test plan covers DPU mode plus full mode.

docs/okeps/okep-5861-two-ovnks.md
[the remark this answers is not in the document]
aserdean: Fixed in the rewrite — formatting and zero-width characters cleaned up, fenced blocks have a language, the typo is gone, the IFNAMSIZ limit is stated as applying to the name values with the arithmetic shown, and the test plan covers DPU mode plus full mode.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,481 @@
...
+* Keep reconciliation and cleanup idempotent, so that an instance never adopts,
+  renames or deletes another instance's resources.
+
+## Non-Goals
+
+* Changing how interfaces are selected for offload. Whether a pod uses an
+  accelerated device or a veth remains a deployment decision, exactly as it is
+  today for a single instance.
+* Redesigning DPU support, or introducing a multi-tenancy API.
+* Providing an operator or any other lifecycle management framework to deploy
```
shettyg, line 33: I wonder a clarification or addition here will help: that support to upgrade between different instance-ids is a non-GOAL.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,481 @@
...
+The chassis name is a separate, longer identity because it must be unique within
+its OVN southbound database, not merely on the node, and because OVN itself
+requires it as the suffix of the chassis-specific `external_ids`. Deriving it
+from the node name and the instance identifier keeps it unique by construction.
+It is passed to `ovn-controller` with the `-n` command line option, which takes
+precedence over the `system-id-override` file and over `external_ids:system-id`
+in the database. That option is needed because `external_ids:system-id` is the
+one option that cannot be made chassis-specific.
+
+`--ovn-chassis-name` and `--bridge-name` already exist in the POC. This OKEP
```
shettyg: I do not quite understand what "on top of" means here. Are you suggesting that "--instance-id" will be passed to ovnkube-* and it will use it to create ovn-chassis-name and bridge-name? Or are those two options in addition to "--instance-id"?

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,481 @@
...
+| OVN database files | `/etc/ovn/` (`OVN_ETCDIR`) | `/etc/ovn/host/` |
+| OVN sockets and PID files | `/var/run/ovn/` (`run-dir`) | `/var/run/ovn/host/` |
+| OVN logs | `/var/log/ovn/` (`OVN_LOGDIR`) | `/var/log/ovn/host/` |
+| ovn-kubernetes logs | `/var/log/ovn-kubernetes/` | `/var/log/ovn-kubernetes/host/` |
+| CNI server socket and certificates | `/var/run/ovn-kubernetes/` | `/var/run/ovn-kubernetes/host/` |
+| ovn-kubernetes configuration file | `/etc/openvswitch/ovn_k8s.conf` (`--config-file`) | one file per instance, already selectable |
+| Shared OVS sockets and PID files | `/var/run/openvswitch/` (`OVS_RUNDIR`) | unchanged, shared by every instance |
+
+A separate OVN database directory is required because each instance runs its own
+northbound and southbound databases (including their raft state when
```
shettyg: I did not understand why we are talking about "raft state when interconnect is used" here.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,481 @@
...
+```
+
+Note that the suffix is the *chassis name*, not the instance identifier: the
+suffix is how `ovn-controller` finds its own configuration, and it looks it up
+by the chassis name it was started with.
+
+This applies to *every* key ovn-kubernetes writes to or reads from the
+`Open_vSwitch` table, not only to the five above. The most consequential one is
+`ovn-bridge-mappings`, which maps physical network names to local bridges and
+therefore decides which bridge each `ovn-controller` patches; it is written
```
shettyg: "it is written unsuffixed today" was a little confusing for me. I think "it is written unsuffixed today by ovn-kubernetes" makes it less confusing.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,481 @@
...
+includes `ovn-remote-probe-interval`, `ovn-bridge-remote-probe-interval`,
+`ovn-is-interconn`, `ovn-monitor-all`, `ovn-ofctrl-wait-before-clear`,
+`ovn-enable-lflow-cache`, `ovn-set-local-ip`, and the ovn-kubernetes specific
+`ovn-pf-encap-ip-mapping`. Rather than suffixing each call site, the suffix is
+applied centrally in the helpers that get and set these keys, so that a key
+added later is scoped automatically. Note that `ovn-set-local-ip` was removed in
+OVN 24.09, where it became unconditionally true.
+
+Each instance must use its own integration bridge, named after the instance
+(`br-int-host`, `br-int-dpu`). Sharing one integration bridge between
```
shettyg, line 241: "Sharing one integration bridge between co-hosted controllers is not supported in upstream OVN" is more explicit

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,481 @@
...
+instance identifier is set, replacing the fixed `ovn-k8s-` prefix:
+
+```text
+instance id unset:  ovn-k8s-mp0      ovn-k8s-mp1 ...   (unchanged)
+instance id "host": ovnk-host-mp0    ovnk-host-mp1 ...
+```
+
+This leaves `15 - len("ovnk--mp") - digits` characters for the identifier. With
+up to three digits of user defined network index, that is four characters, so
+`host` and `dpu` fit and `local` does not. The identifier is validated at
```
shettyg, line 357: When I did the POC, I used a separate mgmt-port-suffix which was a single character. So that strategy would mean that along with instance id, you need this separate id specific to handle mgmt ports. Your design means that we can continue to use instance id to handle this one too (more simpler), but it also means that you are now constrained by how long the string that represents instance-id can be. Just putting out here as a food for thought for other reviewers who likely will have more insight.

shettyg (commented):
This version takes care of all the things that I raised before. I do have some minor comments.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Correct. This design deliberately uses the instance ID for management-port naming, trading identifier length for a single configuration input. Startup validates every derived interface name against IFNAMSIZ; a separate suffix remains an alternative if reviewers prefer more identifier flexibility.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Yes. --instance-id is the primary input. ovn-kubernetes derives --ovn-chassis-name and --bridge-name from it.

The existing flags remain optional overrides for deployments that need explicit values; they are not additional required configuration.

aserdean:
@dceara can you please take another look if you have spare the time?

shettyg:
LGTM from me - the latest version of OKEP satisfies all the additional things I had to do when I did a POC with 3 OVN stacks on a simulated DPU. (It is possible that I did not think about some of the ovn-kubernetes features that needs additional changes)

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+  A1 --> DB1
+  A2 --> DB2
+  A1 --> K1
+  A2 --> K2
+```
+
+Because each instance has its own northbound and southbound databases, no OVN
+logical object needs an instance-specific identifier: an instance is the only
+writer of its own databases. Only resources in the shared Open vSwitch, and
+names in the shared host namespaces (network interfaces, file paths, TCP ports),
```
dceara: Nit: maybe mention conntrack zones here too.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+* An entry for this OKEP in `mkdocs.yml`.
+
+## Risks, Known Limitations and Mitigations
+
+* Co-hosting `ovn-controller` instances is documented upstream as highly
+  experimental, with the explicit limitation that stateful ACLs are not
+  supported. ovn-kubernetes relies on stateful ACLs for network policy, and on
+  conntrack for services and egress traffic. Giving each instance its own
+  integration bridge and a non-overlapping `ct-zone-range` (OVN 24.09) is
+  expected to address the underlying zone conflict, but this must be validated
```
dceara: We now have tests upstream to ensure this works and doesn't break: https://github.com/ovn-org/ovn/commit/b543a88

I posted an OVN patch to remove the experimental tag of the "co-hosting ovn-controller instances" feature (starting with 26.09.0):
https://patchwork.ozlabs.org/project/ovn/patch/20260818123757.2919839-1-dceara@redhat.com/

So maybe we should consider rephrasing this.

dceara (approved):
Two small nits on my side, otherwise this looks good to me, thanks!

docs/okeps/okep-5861-two-ovnks.md
dceara: Update: the OVN patch above has been merged so the feature won't be experimental anymore starting with 26.09.0.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+* Issue: [#5825](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5825)
+* Enhancement PR: [#5861](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5861),
+  which this OKEP is numbered after
+* POC: [#5354](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5354)
+
+## Problem Statement
+
+ovn-kubernetes cannot run more than one instance on the same node, so a DPU
+cannot simultaneously run an instance in DPU mode (serving pods of the host
+cluster) and an instance in full mode (serving pods running on the DPU itself).
```
npinaeva, line 12: I am still voting for up-leveling the problem statement and use case to the actual problem.
AFAIU the problem is that with current model DPU cluster can't create any non-hostnetwork pods of its own. The fact that it requires 2 ovn-ks runnning on the DPu is an implementation detail

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+
+## Goals
+
+* Run multiple independent ovn-kubernetes instances on the same node, each with
```
npinaeva, line 16: here too I don't think running any number of instances is the goal, let's start with the functionality we want to support

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+`ovs-vswitchd`, each driving its own integration bridge. This OKEP describes the
+changes ovn-kubernetes needs in order to use that capability.
+
+Note that nothing in this design is specific to a DPU or to a count of two: it
+generalizes to N instances on any node. DPUs with two instances are simply the
+use case that is being proposed, implemented and tested here.
+
+## User-Stories/Use-Cases
+
+Story 1: Run DPU-local workloads while offloading host workloads
```
npinaeva, line 68: that is a good user story

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+
+As a cluster admin deploying DPUs, I want to run an ovn-kubernetes instance in
+DPU mode that provides networking for pods on the host, and at the same time run
+an ovn-kubernetes instance in full mode that provides networking for pods
+scheduled on the DPU, so that I can use the DPU both as an offload engine for
+the host cluster and as a compute node of a separate cluster.
+
+Story 2: Keep existing deployments untouched
```
npinaeva: that sounds more like a reasonable backwards-compatibility requirement, not a user story?

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+need to be scoped per instance.
+
+### API Details
+
+A single new knob identifies the instance, and everything else is derived from
+it:
+
+| Knob | Default | Meaning |
+| --- | --- | --- |
+| `--instance-id=STRING` | `""` | Short identifier of this instance. Empty selects today's behaviour exactly. |
```
npinaeva, line 132: that sounds very generic as a name, maybe at least ovn-instance-id? or something more specific

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+| `--ovn-chassis-name=STRING` | `NODE-INSTANCE_ID` | Chassis name, and therefore the `external_ids` suffix. Derived from the node name and the instance identifier; settable for deployments that need an explicit chassis name. |
+| `--bridge-name=STRING` | `br-int`, or `br-int-INSTANCE_ID` | Integration bridge of this instance. Derived from the instance identifier, so that two instances never share one bridge by accident. |
```
npinaeva: so why are these 2 flags, but saying they are derived from the first one?

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,487 @@
...
+short value instead of separately choosing a chassis name, a bridge name, a set
+of directories, an interface name suffix and a set of ports, and enforces the
+relationship between those names rather than leaving it to the deployer.
+
+### Implementation Details
+
+#### Instance-scoped host state
+
+When the instance identifier is set, it is appended to the directories that the
+instance owns. For an identifier of `host`:
```
npinaeva, line 175: considering today's supported option is host, and what we are adding here is actually dpu, should we instead keep host directories and make new ones for dpu?

docs/okeps/okep-5861-two-ovnks.md
aserdean: this is conflicting with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5861#discussion_r3831006201 : should this stay a generic multi-instance design with DPU as the first use case, or be limited to DPU mode plus full mode on a DPU?

docs/okeps/okep-5861-two-ovnks.md
aserdean: Agreed. I moved this to Backwards Compatibility and kept only the DPU use case as a user story.

docs/okeps/okep-5861-two-ovnks.md
aserdean: This is intentionally the generic unique identifier for an ovn-kubernetes instance. It is used to derive the instance-owned resources, so --instance-id is sufficient

docs/okeps/okep-5861-two-ovnks.md
aserdean: They are not three required inputs. --instance-id derives the default chassis and bridge names; the existing flags remain optional overrides for deployments that need explicit values. I will clarify this in the table.

docs/okeps/okep-5861-two-ovnks.md
aserdean: We do not need fixed host and dpu directory names. Two co-hosted instances need separate paths, but those paths are derived from their configured unique instance IDs. `host` and `dpu` are only examples for the DPU deployment.

docs/okeps/okep-5861-two-ovnks.md
shettyg: We did have a use case for 3 OVN stacks (2 ovn-kubernetes and 1 OVN) on our DPUs. The OVN stack container was built with ovnkube.sh as the basis. So keeping it generic for 3 stacks should be okay.

npinaeva: not sue I am following the 3 stacks use case, like the OVN one has nothing to do with ovn-k?
The design will be multi-instance and preferably generic, but the goal and use cases should be specific. I think it goes exactly in line with the previous comment? I am suggesting to have goal/use cases talk about the actual DPU cluster setup we want to support and explain the generic multi-instance design as an implementation

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+### API Details
+
+The instance identifier is the primary knob. The gateway bridge and conntrack
+zone range are configured explicitly because they own shared host resources:
+
+| Knob | Default | Meaning |
+| --- | --- | --- |
+| `--instance-id=STRING` | `""` | Short identifier of this instance. Empty selects today's behaviour exactly. |
+| `--ovn-chassis-name=STRING` | Existing chassis name if the identifier is empty; otherwise `NODE-INSTANCE_ID` | Optional chassis-name override, and therefore the `external_ids` suffix. |
+| `--bridge-name=STRING` | `br-int` when unset, otherwise `br-int-INSTANCE_ID` | Optional integration-bridge override. |
```
npinaeva, line 134: do we need the overrides though? I think it would be nice to have predictable names based on the instanceID and less flags to maintain unless there is a specific use case where override is required?

docs/okeps/okep-5861-two-ovnks.md
npinaeva: oh I see, I guess empty ID will keep current directories

docs/okeps/okep-5861-two-ovnks.md
shettyg: >like the OVN one has nothing to do with ovn-k?

Right. We were using ovn-k8s helm charts (with ovnkube.sh) and ovnkube-util for health checks. Except, there was no ovnkube-node.

We also had another use case wherein we did third stack of ovn-kubernetes to support untrusted k8s clusters (nothing in prod). So the point I was trying to convey was that the current OKEP as written  supports multiple ovn-kubernetes stacks for a downstream user.

>The design will be multi-instance and preferably generic, but the goal and use cases should be specific. 

This should be okay. Alin, I am okay if you change the goal to be in line with what Nadia wants. 

What I really want is the implementation to be generic - so that a downstream user need not change any code and  can have many stacks  on a node.

docs/okeps/okep-5861-two-ovnks.md
aserdean: Yes, we do want users to choose.

The instance ID gives us a predictable default, but the override is intentional so users can define whatever configuration fits their setup.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+## Goals
+
+* Run a DPU-mode instance for the host cluster and a full-mode instance for a
+  DPU-local cluster on the same DPU.
+* Give every instance unambiguous ownership of the host state it creates:
+  configuration, logs, sockets, PID files, OVN database files, OVS resources and
+  interface names.
+* Keep single-instance deployments byte-for-byte compatible: when no instance
+  identifier is configured, every path and name is what it is today.
+* Keep reconciliation and cleanup idempotent, so that an instance never adopts,
```
wizhaoredhat, line 24: This seems like the 2 instances DPU-mode + Full mode are independent. There is no plans for them to communicate with each other?

What would be a "bad outcome" is passing both the infrastructure and tenant cluster KUBECONFIGS to both DPU and Full mode instances IMO

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+| Knob | Default | Meaning |
+| --- | --- | --- |
+| `--instance-id=STRING` | `""` | Short identifier of this instance. Empty selects today's behaviour exactly. |
+| `--ovn-chassis-name=STRING` | Existing chassis name if the identifier is empty; otherwise `NODE-INSTANCE_ID` | Optional chassis-name override, and therefore the `external_ids` suffix. |
+| `--bridge-name=STRING` | `br-int` when unset, otherwise `br-int-INSTANCE_ID` | Optional integration-bridge override. |
+| `--gateway-bridge=STRING` | Existing gateway bridge if the identifier is empty; otherwise required when gateway mode is enabled | Per-instance gateway bridge. |
+| `--ovn-ct-zone-range=MIN-MAX` | Required when the identifier is set | Non-overlapping OVN connection-tracking zone range. |
+
+The instance identifier must be a valid DNS label and short enough that every
+interface name derived from it fits in `IFNAMSIZ`, which is validated at startup
```
wizhaoredhat, line 139: It would suck that this is a runtime validation. Crashing because IFNAMSIZE is not fun. Maybe we can already determine the max size for instance-id based on the code.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+(see [Naming constraints](#naming-constraints)). It is the value that appears in
+interface names, directory paths and metrics labels, so it is deliberately kept
+short and human readable, for example `host` for the DPU-mode instance serving
+the host cluster and `dpu` for the full-mode instance serving the DPU-local
+cluster.
+
+The chassis name is a separate, longer identity because it must be unique within
```
wizhaoredhat, line 146: Can chassis name be very long? Do we need size safeguards like with IFNAMESIZE above?

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+`--instance-id` is the primary input: when the override flags are not set,
+ovn-kubernetes derives the chassis name and integration bridge from it.
+`--gateway-bridge` and `--ovn-ct-zone-range` remain explicit because they
+partition shared host resources.
+
+Startup rejects an effective chassis name, bridge, path, interface, port or
+gateway bridge that is already owned by another local instance.
+
+Admission and cleanup are serialized by a node-wide lock at
+`/var/run/ovn-kubernetes/instance-admission.lock`, acquired before validation
```
wizhaoredhat, line 166: Would this file path be override-able? Does it mean this is a change to the current behavior? Requiring even instance-id="" to also look for the lock.

Also.. Wouldn't concurrency starts be ok? Do we need this lock? Because each instance touches it's own files and has its own OVN DB.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+in the `Open_vSwitch` database. `ovn-controller` reads most of its configuration
+from `external_ids` in that table, and supports a chassis-specific form of most
+such options: `external_ids:<option>-<chassis-name>` overrides
+`external_ids:<option>`. The instance is responsible for writing the
+chassis-specific keys for its own chassis, and must not write the unsuffixed
+ones when an instance identifier is set. With an empty identifier, existing
+unsuffixed keys are preserved. For chassis `NODE-host`:
+
+```text
+external_ids:ovn-remote-NODE-host="ssl:10.0.0.1:9642"
```
wizhaoredhat, line 218: Is it possible for the chassis-name to change? If so, wouldn't there be old/stale entries for external_ids + the files mentioned above? If so, any proposed mechanism to clean it up.

docs/okeps/okep-5861-two-ovnks.md
```diff
@@ -0,0 +1,479 @@
...
+Not every co-hosting key is an `external_ids` key. `ovn-controller` keeps a
+per-chassis index in `other_config:ovn-chassis-idx-<chassis-name>` of the same
+table, which it allocates itself: the first controller on the host gets an empty
+index, the next gets `0`, and tunnel ports are named `ovn<idx>-<chassis>-<hex>`
+so that co-hosted controllers cannot collide on interface names. ovn-kubernetes
+neither sets nor needs this key, but cleanup must not delete another chassis's
+entry, and anything that matches tunnel port names has to accept the indexed
+form.
+
+Each instance sets `--ovn-ct-zone-range` to a non-overlapping range, so
```
wizhaoredhat, line 253: Why can't the ct-zone-range be a parameter we pass in. So we don't need to synchronize the range dynamically between different instances. This assumes 2 instances but the design allows for more. The controller wouldn't know how many instances total there are... would it?

wizhaoredhat (commented):
I feel like for a use case like this, an Operator pattern would be very nice to handle starting the OVN-K instances and managing the lifecycle of them on each node (e.g. no need to use lock files to synchronize). But definitely out of scope for this OKEP. Just a observation I made looking at the design.

Thanks Alin! Overall I like this design, just a few questions mostly centered around synchronization between each instance.

docs/okeps/okep-5861-two-ovnks.md
npinaeva: can you give an example of a setup where this is needed? and e.g. can't be solved by picking a different instanceID?

PR #1597, opened 2020-08-07
implement dns names in egressfirewall

JacobTanenbaum:
**- What this PR does and why is it needed**
This PR enhances the egressfirewall by allowing cluster admins to specify dnsNames to allow/block

**- Special notes for reviewers**

Most of the relevant changes are in go-controller/pkg/ovn/egressfirewall_dns.go. The most challenging part was the Sync() to make sure that dns entries stay as up to date as possible  

**- How to verify it**

Manually for now I am working on unit testing 

**- Description for the changelog**

enhance egressfirewall to allow/deny traffic based on dns name

PR #5867, opened 2026-01-21
EgressFirewall: Use exponential backoff to refresh IP addresses for DNS names

arkadeepsen:
## 📑 Description

For the following DNS name lookup scenarios, the TTL of a DNS name was being set at 30 minutes:
- When an error is encountered during DNS name lookup
- TTL value can't be parsed into duration
- When TTL returned is zero

Except for the zero TTL scenario, the rest of the scenarios can be transient in nature in some cases, and the 30 minutes delay can cause a significant synchronization issue for DNS name rules in EgressFirewall. 

For the zero TTL scenario, a TTL of 5 seconds is used to refresh the IP addresses.

In the case of the error scenarios, instead of directly using a conservative TTL of 30 minutes, 5 seconds TTL is used for the first 10 times, then exponential backoff, is used which is capped off at 2 minutes. This can help mitigate the synchronization issues when these transient scenarios are encountered. Additionally, capping off at 2 minutes rather than 30 minutes will let ovnk pods to get the latest IP addresses quickly when the transient problem goes away.

Fixes #

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Adds retry tracking and exponential backoff for DNS lookups with enforced minimum and maximum TTL behavior.

* **Bug Fixes**
  * Better handling of missing/invalid TTLs (including TTL=0), deduplicated IP results, and fallback to previously known IPs when lookups fail.

* **Tests**
  * Expanded coverage for TTL handling, backoff/doubling/capping, retry semantics, IP equality, and update flows.

### DIFF
--- a/go-controller/pkg/util/dns.go
+++ b/go-controller/pkg/util/dns.go
@@ -16,8 +16,12 @@ import (
 )
 
 const (
-	// defaultTTL is used if an invalid or zero TTL is provided.
-	defaultTTL = 30 * time.Minute
+	// defaultMinTTL is the minimum TTL value that will be used for a domain name if an invalid or zero TTL is found
+	defaultMinTTL = 5 * time.Second
+	// defaultMaxTTL is the maximum TTL value that will be used for a domain name if an invalid or zero TTL is found
+	defaultMaxTTL = 2 * time.Minute
+	// maxRetryBeforeBackoff is the maximum number of times to retry a DNS lookup before exponential backoff starts
+	maxRetryBeforeBackoff = 10
 )
 
 type dnsValue struct {
@@ -27,6 +31,8 @@ type dnsValue struct {
 	ttl time.Duration
 	// Holds (last dns lookup time + ttl), tells when to refresh IPs next time
 	nextQueryTime time.Time
+	// Number of times the DNS lookup has been retried before backoff starts
+	retryCount int
 }
 
 type DNS struct {
@@ -105,11 +111,22 @@ func (d *DNS) updateOne(dns string) (bool, error) {
 		return false, fmt.Errorf("DNS value not found in dnsMap for domain: %q", dns)
 	}
 
-	ips, ttl, err := d.getIPsAndMinTTL(dns)
-	if err != nil {
-		res.nextQueryTime = time.Now().Add(defaultTTL)
-		d.dnsMap[dns] = res
-		return false, err
+	ips, ttl, retry, err := d.getIPsAndMinTTL(dns)
+	if retry {
+		// If the DNS lookup has been retried maxRetryCount times, use exponential backoff
+		// by doubling the previous TTL. The TTL is capped at defaultMaxTTL.
+		if res.retryCount >= maxRetryBeforeBackoff {
+			ttl = min(res.ttl*2, defaultMaxTTL)
+		} else {
+			// Increment the retry count
+			res.retryCount++
+		}
+		// If no valid IPs were found, use the previous IPs as fallback.
+		if len(ips) == 0 {
+			ips = res.ips
+		}
+	} else {
+		res.retryCount = 0
 	}
 
 	changed := false
@@ -120,10 +137,10 @@ func (d *DNS) updateOne(dns string) (bool, error) {
 	res.ttl = ttl
 	res.nextQueryTime = time.Now().Add(res.ttl)
 	d.dnsMap[dns] = res
-	return changed, nil
+	return changed, err
 }
 
-func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
+func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, bool, error) {
 	ips := []net.IP{}
 	ttlSet := false
 	var ttlSeconds uint32
@@ -197,19 +214,27 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
 	}
 
 	if !ttlSet || (len(ips) == 0) {
-		return nil, defaultTTL, fmt.Errorf("IPv4 or IPv6 addr not found for domain: %q, nameservers: %v", domain, d.nameservers)
+		return nil, defaultMinTTL, true, fmt.Errorf("IPv4 or IPv6 addr not found for domain: %q, nameservers: %v", domain, d.nameservers)
 	}
+
+	ips = removeDuplicateIPs(ips)
 
 	ttl, err := time.ParseDuration(fmt.Sprintf("%ds", minTTL))
 	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("invalid TTL value for domain: %q, err: %v, defaulting ttl=%s", domain, err, defaultTTL.String()))
-		ttl = defaultTTL
+		utilruntime.HandleError(fmt.Errorf("invalid TTL value for domain: %q, err: %v", domain, err))
+		return ips, defaultMinTTL, true, nil
 	}
 	if ttl == 0 {
-		ttl = defaultTTL
+		// If the TTL is 0, return the default minimum TTL. The retry is set to false as this
+		// is not an error scenario. TTL being 0 is a valid scenario for some DNS servers
+		// and it means that the IP addresses should be refreshed everytime whenever the DNS
+		// name is being used. From the point of view of OVN-Kubernetes, the IP addresses are
+		// refreshed every defaultMinTTL.
+		klog.V(5).Infof("TTL value is 0 for domain: %q, defaulting ttl=%s", domain, defaultMinTTL.String())
+		return ips, defaultMinTTL, false, nil
 	}
 
-	return removeDuplicateIPs(ips), ttl, nil
+	return ips, ttl, false, nil
 }
 
 func (d *DNS) GetNextQueryTime() (time.Time, string, bool) {
--- a/go-controller/pkg/util/dns_test.go
+++ b/go-controller/pkg/util/dns_test.go
@@ -70,13 +70,16 @@ func TestGetIPsAndMinTTL(t *testing.T) {
 	tests := []struct {
 		desc             string
 		errExp           bool
+		retry            bool
 		ipv4Mode         bool
 		ipv6Mode         bool
 		dnsOpsMockHelper []ovntest.TestifyMockHelper
+		expectedTTL      time.Duration
 	}{
 		{
 			desc:     "call to Exchange fails IPv4 only",
 			errExp:   true,
+			retry:    true,
 			ipv4Mode: true,
 			ipv6Mode: false,
 			dnsOpsMockHelper: []ovntest.TestifyMockHelper{
@@ -89,10 +92,12 @@ func TestGetIPsAndMinTTL(t *testing.T) {
 					CallTimes:           1,
 				},
 			},
+			expectedTTL: defaultMinTTL,
 		},
 		{
 			desc:     "Exchange returns correctly but Rcode != RcodeSuccess IPv4 only",
 			errExp:   true,
+			retry:    true,
 			ipv4Mode: true,
 			ipv6Mode: false,
 			dnsOpsMockHelper: []ovntest.TestifyMockHelper{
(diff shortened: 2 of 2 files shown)


igsilya:
@arkadeepsen fwiw, TTL of zero is a completely valid value and it is not a transient situation.  It's set when the server is asking the client to not cache the result of the lookup, or as RFC 1035 says:
```
TTL             a 32 bit signed integer that specifies the time interval
                that the resource record may be cached before the source
                of the information should again be consulted.  Zero
                values are interpreted to mean that the RR can only be
                used for the transaction in progress, and should not be
                cached.  For example, SOA records are always distributed
                with a zero TTL to prohibit caching.  Zero values can
                also be used for extremely volatile data.
```
Since the situation is not transient, the code will eventually back off to 30 min again.  I don't think a valid reply with TTL of zero should be treated the same as error, i.e. there shouldn't be a backoff in this case, but just some sort of a reasonable default value.

arkadeepsen:
> @arkadeepsen fwiw, TTL of zero is a completely valid value and it is not a transient situation. It's set when the server is asking the client to not cache the result of the lookup, or as RFC 1035 says:
> 
> ```
> TTL             a 32 bit signed integer that specifies the time interval
>                 that the resource record may be cached before the source
>                 of the information should again be consulted.  Zero
>                 values are interpreted to mean that the RR can only be
>                 used for the transaction in progress, and should not be
>                 cached.  For example, SOA records are always distributed
>                 with a zero TTL to prohibit caching.  Zero values can
>                 also be used for extremely volatile data.
> ```
> 
> Since the situation is not transient, the code will eventually back off to 30 min again. I don't think a valid reply with TTL of zero should be treated the same as error, i.e. there shouldn't be a backoff in this case, but just some sort of a reasonable default value.

@igsilya makes sense. I treated them as same as currently we are using 30 mins as default for all the scenarios. For the improved DNS resolution using `DNSNameResolver` feature, a minimum TTL of 5 secs is used for 0 TTL scenario: https://github.com/openshift/enhancements/blob/master/enhancements/dns/coredns-egressfirewall-integration.md#dns-lookup-interception

arkadeepsen:
@npinaeva We tested the above PR with a scenario with a LB whose DNS name was being used in the EFW DNS name rule. The LB was recreated with the same DNS name which took some time to re-provision. However, even with the exponential backoff the overall time it took for the underlying address set to get the new IP address of the DNS name was close to 20 mins.

As the default max TTL value is set to 30 mins, we are seeing this longer delay for the EFW controller to reconcile the latest IP addresses when these transient issues occur. I am inclined to reduce the 30 mins max to TTL to a lower value (maybe 1 or 2 mins). Going through the PR (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/1597) which adedd the DNS name feature in EFW, I couldn't find the rationale of why the default TTL value was set to 30 mins. 

I am also inclined to add an inital retry after every 5 secs for a number of times (say 10) and then start the exponential backoff which is capped off by the relatively lower max default TTL. PLMK your thoughts.

igsilya:
> I treated them as same as currently we are using 30 mins as default for all the scenarios.

The current code is incorrect, that's why you're fixing it.  Treating the zero value the same as borken reply doesn't make much sense.  Since it's not a transient issue this PR doesn't solve the problem, it is just delaying it by a few minutes, as the backoff will go up to the max value pretty fast.  There shouldn't be a backoff in the zero TTL case, but a short re-check at a constant time intervals, IMO.

npinaeva:
> it is just delaying it by a few minutes, as the backoff will go up to the max value pretty fast.

yeah +1 on that, zero TTL is a well-known problem because we can't "not cache" the result, so I agree that zero TTL deserves a more frequent lookup, but it should be different form the error cases, where a backoff makes sense. Now choosing a reasonable refresh interval is a tricky task, for consistency with the new DNS resolution we could set it to 5 seconds, but that may also overload the system. There is no data to pick up any better number though, so maybe we can go with that.
In general, this old dns resolution way doesn't work well with TTL=0 by design, and I think it even should be documented somewhere (but we do what we can for our users :))

arkadeepsen:
I have updated the code to handle zero TTL scenario differently. For this scenario, a 5 secs TTL value is used.

I have also updated the handling of the error scenario. For this scenario, the DNS lookup will be retried every 5 secs for 10 times, then exponential backoff will be used if the DNS lookup still ends up with an error. The exponential backoff is capped off at 2 mins.

I'll do a rebase on the master which should fix the failing unit test CI. PTAL.

go-controller/pkg/util/dns.go
```diff
@@ -16,8 +16,12 @@ import (
...
 
 const (
-	// defaultTTL is used if an invalid or zero TTL is provided.
-	defaultTTL = 30 * time.Minute
+	// defaultMinTTL is the minimum TTL value that will be used for a domain name if an invalid or zero TTL is found
+	defaultMinTTL = 5 * time.Second
+	// defaultMaxTTL is the maximum TTL value that will be used for a domain name if an invalid or zero TTL is found
+	defaultMaxTTL = 2 * time.Minute
+	// maxRetryCount is the maximum number of times to retry a DNS lookup before exponential backoff starts
+	maxRetryCount = 10
```
npinaeva: that sounds like we won't retry after this value is reached, maybe better call it `retryBeforeBackoff`?

go-controller/pkg/util/dns.go
```diff
@@ -27,6 +31,8 @@ type dnsValue struct {
 	ttl time.Duration
 	// Holds (last dns lookup time + ttl), tells when to refresh IPs next time
 	nextQueryTime time.Time
+	// Number of times the DNS lookup has been retried
+	retryCount int
```
npinaeva, line 35: I think it is not incremented after the `maxRetryCount` is reached, so would it be more accurate to say "Number of times the DNS lookup has been retried before backoff`?

go-controller/pkg/util/dns.go
```diff
@@ -197,19 +214,21 @@ func (d *DNS) getIPsAndMinTTL(domain string) ([]net.IP, time.Duration, error) {
...
 	ttl, err := time.ParseDuration(fmt.Sprintf("%ds", minTTL))
 	if err != nil {
-		utilruntime.HandleError(fmt.Errorf("invalid TTL value for domain: %q, err: %v, defaulting ttl=%s", domain, err, defaultTTL.String()))
-		ttl = defaultTTL
+		utilruntime.HandleError(fmt.Errorf("invalid TTL value for domain: %q, err: %v", domain, err))
+		return ips, defaultMinTTL, true, nil
 	}
 	if ttl == 0 {
-		ttl = defaultTTL
+		return ips, defaultMinTTL, false, nil
```
npinaeva, line 234: it would be nice to add a comment here saying why we return false for ttl=0

go-controller/pkg/util/dns_test.go
```diff
@@ -319,3 +364,379 @@ func TestAdd(t *testing.T) {
 	}
 
 }
+
+func TestIPsEqual(t *testing.T) {
```
npinaeva, line 372: this function didn't change? why are you adding these tests?

go-controller/pkg/util/dns_test.go
```diff
@@ -319,3 +364,379 @@ func TestAdd(t *testing.T) {
...
+	}
+	for i, tc := range tests {
+		t.Run(fmt.Sprintf("%d:%s", i, tc.desc), func(t *testing.T) {
+			res := ipsEqual(tc.oldips, tc.newips)
+			assert.Equal(t, tc.expEqual, res)
+		})
+	}
+}
+
+func TestUpdateOne(t *testing.T) {
```
npinaeva, line 454: this looks like an AI-generated blob that doesn't even test what the updateOne function is actually doing, I believe it can be safely removed

go-controller/pkg/util/dns_test.go
arkadeepsen: I added the test because for error scenario also this function is being called now with nil input. Added the tests to verify it works correctly for all the scenarios.

go-controller/pkg/util/dns_test.go
arkadeepsen: This test specifically checks the TTL value being used for the different scenarios:

- Normal DNS lookup
- Zero TTL response
- Error scenario with retries and backoff

This does give a better understanding how it'll behave during the different scenarios.

arkadeepsen: I think I can improve the readability of the testcases. Let me update it.

arkadeepsen: I have updated the test by removing the repetitive part from the testcases. Now the testcases look clean and more readable. PLMK whether that's fine.

go-controller/pkg/util/dns.go
arkadeepsen: Hey @npinaeva, do you think we should add a info/warning log message here when zero TTL is encountered? I was testing the existing behavior by changing the IP address behind the DNS record. The TTL was defaulting to 30m but no log messages were being printed why it got changed to 30m. When I tested the same scenario with this PR, the new IP address was fetched without much delay, but still didn't get any log messages. So I wasn't sure whether we were hiting the issue or not when testing with this PR. After adding some log messages for the TTL 0 scenario, I was able to understand that, sometimes the DNS name server returns 0 TTL when it is queried just at the time of the TTL expiration. Adding a log message here will help to debug issues in the future. WDYT?

arkadeepsen: Added an info log message for zero TTL scenario. PLMK your thoughts.

npinaeva: Yeah at v5 I think this message can be useful for debugging

go-controller/pkg/util/dns_test.go
```diff
@@ -319,3 +364,180 @@ func TestAdd(t *testing.T) {
...
+			newips:   nil,
+			expEqual: true,
+		},
+		{
+			desc:     "oldips and newips are both empty",
+			oldips:   []net.IP{},
+			newips:   []net.IP{},
+			expEqual: true,
+		},
+	}
```
npinaeva, line 445: what about nil and empty at the same time?

go-controller/pkg/util/dns_test.go
```diff
@@ -319,3 +364,180 @@ func TestAdd(t *testing.T) {
...
+			numCalls:              1,
+			exchangeOpsMockHelper: exchangeSuccessNoAnswerOpsMockHelper,
+			expTTL:                defaultMinTTL,
+		},
+		{
+			desc:                  "when TTL returned is 0 by Exchange function, defaultMinTTL is used",
+			numCalls:              1,
+			exchangeOpsMockHelper: exchangeSuccessZeroTTLOpsMockHelper,
+			expTTL:                defaultMinTTL,
+		},
```
npinaeva, line 499: I would focus more on the backoff/zero-TTL behaviour than different error types. For example, it would be nice to have 11 retries with zero ttl to make sure we don't backoff in this case

npinaeva (commented):
couple more nits

go-controller/pkg/util/dns_test.go
arkadeepsen: Added both the scenarios here: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/fa2cea9b7e23373b35a62f4a57a162d29d433ffd..bc024ab6e651677ec208b7b71e763371814c788c

go-controller/pkg/util/dns_test.go
arkadeepsen: Added a few more test cases for the backoff/zero-TTL scenario here: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/fa2cea9b7e23373b35a62f4a57a162d29d433ffd..bc024ab6e651677ec208b7b71e763371814c788c

arkadeepsen:
Last push was a rebase

go-controller/pkg/util/dns_test.go
```diff
@@ -128,19 +167,24 @@ func TestGetIPsAndMinTTL(t *testing.T) {
...
 			} else {
 				require.NoError(t, err)
 			}
+			assert.Equal(t, tc.retry, retry, "the exponentialBackoff variable should match the return from dns.getIPsAndMinTTL()")
+			if tc.retry {
+				assert.Equal(t, defaultMinTTL, ttl, "the ttl variable should match the return from dns.getIPsAndMinTTL()")
+			}
 			mockDNSOps.AssertExpectations(t)
 		})
 	}
```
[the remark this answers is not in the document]
arkadeepsen: Addressed here: https://github.com/ovn-kubernetes/ovn-kubernetes/compare/b01e7ddb8f62c23787c40413341c52fbc3566101..1d5ee82ff968cbdefd741130ec49aec63d6086e3

PR #5871, opened 2026-01-22
egressip: set IFA_PROTO on EgressIP addresses to prevent nmstate persistence

cgoncalves:
When nmstate applies configuration changes, it captures the current interface state including all IP addresses. This causes EgressIPs to be persisted to .nmconnection files, which is undesirable as they are dynamically managed by OVN-Kubernetes.

This commit sets IFA_PROTO=84 (matching RTPROT_OVN used for routes) on EgressIP addresses. This allows tools like nmstate to identify and filter out OVN-managed addresses when capturing interface state.

Note: IFA_PROTO requires Linux kernel 5.18+; on older kernels, the attribute is silently ignored.

## Summary by CodeRabbit

* **Tests**
  * Added unit tests for egress IP address handling, validating OVN-marked addresses, IPv4/IPv6 flag behavior, and address construction.

* **Chores**
  * Updated netlink dependency to a newer patch release for improved compatibility.

### DIFF
--- a/go-controller/pkg/types/const.go
+++ b/go-controller/pkg/types/const.go
@@ -155,6 +155,13 @@ const (
 	// LabelUserDefinedServiceName label key used in mirrored EndpointSlices that contains the service name matching the EndpointSlice
 	LabelUserDefinedServiceName = "k8s.ovn.org/service-name"
 
+	// IFAProtOVNK is the IFA_PROTO value used to mark addresses as
+	// OVN-Kubernetes-managed. Value 85 is used instead of RTPROT_OVN (84)
+	// so we do not reuse a protocol identifier already owned by OVN.
+	// IFA_PROTO requires Linux kernel 5.18+; on older kernels the
+	// attribute is silently ignored.
+	IFAProtOVNK = 85
+
 	// Packet marking
 	EgressIPNodeConnectionMark         = "1008"
 	EgressIPReplyTrafficConnectionMark = 42
--- a/go-controller/pkg/util/egressip/net.go
+++ b/go-controller/pkg/util/egressip/net.go
@@ -13,8 +13,17 @@ import (
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
 )
 
+// ifaProtocolOVN is the IFA_PROTO value used to mark addresses as OVN-managed.
+// This matches RTPROT_OVN (84) used for OVN-managed routes.
+const ifaProtocolOVN = 84
+
 // GetNetlinkAddress returns a netlink address configured with specific
-// egress ip parameters
+// egress ip parameters. The address is marked with IFA_PROTO=84 (OVN)
+// to indicate it is managed by OVN-Kubernetes. This allows tools like
+// nmstate to identify and filter out OVN-managed addresses when capturing
+// interface state, preventing them from being persisted to .nmconnection files.
+// Note: IFA_PROTO requires Linux kernel 5.18+; on older kernels, the
+// attribute is silently ignored.
 func GetNetlinkAddress(ip net.IP, ifindex int) *netlink.Addr {
 	return &netlink.Addr{
 		IPNet:     &net.IPNet{IP: ip, Mask: util.GetIPFullMask(ip)},
@@ -22,6 +31,7 @@ func GetNetlinkAddress(ip net.IP, ifindex int) *netlink.Addr {
 		Scope:     int(netlink.SCOPE_UNIVERSE),
 		ValidLft:  getNetlinkAddressValidLft(ip),
 		LinkIndex: ifindex,
+		Protocol:  ifaProtocolOVN, // Mark as OVN-managed for nmstate filtering
 	}
 }
 
--- a/go-controller/pkg/util/egressip/net.go
+++ b/go-controller/pkg/util/egressip/net.go
@@ -10,11 +10,17 @@ import (
 	"github.com/vishvananda/netlink"
 	"golang.org/x/sys/unix"
 
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
 )
 
 // GetNetlinkAddress returns a netlink address configured with specific
-// egress ip parameters
+// egress ip parameters. The address is marked with IFA_PROTO=85 (OVN-K)
+// to indicate it is managed by OVN-Kubernetes. This allows external tools
+// to identify and filter out OVN-Kubernetes-managed addresses when
+// capturing interface state.
+// Note: IFA_PROTO requires Linux kernel 5.18+; on older kernels, the
+// attribute is silently ignored.
 func GetNetlinkAddress(ip net.IP, ifindex int) *netlink.Addr {
 	return &netlink.Addr{
 		IPNet:     &net.IPNet{IP: ip, Mask: util.GetIPFullMask(ip)},
@@ -22,6 +28,7 @@ func GetNetlinkAddress(ip net.IP, ifindex int) *netlink.Addr {
 		Scope:     int(netlink.SCOPE_UNIVERSE),
 		ValidLft:  getNetlinkAddressValidLft(ip),
 		LinkIndex: ifindex,
+		Protocol:  types.IFAProtOVNK, // Mark as OVN-Kubernetes-managed for external tooling
 	}
 }
 
--- a/go-controller/pkg/util/egressip/net_test.go
+++ b/go-controller/pkg/util/egressip/net_test.go
@@ -1,0 +1,89 @@
+// SPDX-FileCopyrightText: Copyright The OVN-Kubernetes Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package egressip
+
+import (
+	"math"
+	"net"
+	"testing"
+
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+)
+
+func TestGetNetlinkAddress(t *testing.T) {
+	tests := []struct {
+		name             string
+		ip               net.IP
+		ifindex          int
+		expectedProto    int
+		expectedScope    int
+		expectedMask     int
+		expectedFlags    int
+		expectedValidLft int
+	}{
+		{
+			name:             "IPv4 address",
+			ip:               net.ParseIP("192.168.1.100"),
+			ifindex:          5,
+			expectedProto:    ifaProtocolOVN,
+			expectedScope:    int(netlink.SCOPE_UNIVERSE),
+			expectedMask:     32,
+			expectedFlags:    0,
+			expectedValidLft: 0,
+		},
+		{
+			name:             "IPv6 address",
+			ip:               net.ParseIP("2001:db8::1"),
+			ifindex:          10,
+			expectedProto:    ifaProtocolOVN,
+			expectedScope:    int(netlink.SCOPE_UNIVERSE),
+			expectedMask:     128,
+			expectedFlags:    unix.IFA_F_NODAD,
+			expectedValidLft: math.MaxUint32,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			addr := GetNetlinkAddress(tt.ip, tt.ifindex)
+
+			// Verify Protocol is set to OVN (84)
+			if addr.Protocol != tt.expectedProto {
+				t.Errorf("Protocol = %d, want %d", addr.Protocol, tt.expectedProto)
+			}
+
+			// Verify Scope
+			if addr.Scope != tt.expectedScope {
+				t.Errorf("Scope = %d, want %d", addr.Scope, tt.expectedScope)
+			}
+
+			// Verify LinkIndex
+			if addr.LinkIndex != tt.ifindex {
+				t.Errorf("LinkIndex = %d, want %d", addr.LinkIndex, tt.ifindex)
+			}
+
+			// Verify IP
+			if !addr.IP.Equal(tt.ip) {
+				t.Errorf("IP = %v, want %v", addr.IP, tt.ip)
+			}
+
+			// Verify mask (should be full mask for EgressIP)
+			ones, _ := addr.Mask.Size()
+			if ones != tt.expectedMask {
+				t.Errorf("Mask prefix length = %d, want %d", ones, tt.expectedMask)
+			}
+
+			// Verify Flags
+			if addr.Flags != tt.expectedFlags {
+				t.Errorf("Flags = %d, want %d", addr.Flags, tt.expectedFlags)
+			}
+
+			// Verify ValidLft
+			if addr.ValidLft != tt.expectedValidLft {
+				t.Errorf("ValidLft = %d, want %d", addr.ValidLft, tt.expectedValidLft)
+			}
+		})
+	}
+}
--- a/go-controller/pkg/util/egressip/net_test.go
+++ b/go-controller/pkg/util/egressip/net_test.go
@@ -1,0 +1,91 @@
+// SPDX-FileCopyrightText: Copyright The OVN-Kubernetes Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package egressip
+
+import (
+	"math"
+	"net"
+	"testing"
+
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
+)
+
+func TestGetNetlinkAddress(t *testing.T) {
+	tests := []struct {
+		name             string
+		ip               net.IP
+		ifindex          int
+		expectedProto    int
+		expectedScope    int
+		expectedMask     int
+		expectedFlags    int
+		expectedValidLft int
+	}{
+		{
+			name:             "IPv4 address",
+			ip:               net.ParseIP("192.168.1.100"),
+			ifindex:          5,
+			expectedProto:    types.IFAProtOVNK,
+			expectedScope:    int(netlink.SCOPE_UNIVERSE),
+			expectedMask:     32,
+			expectedFlags:    0,
+			expectedValidLft: 0,
+		},
+		{
+			name:             "IPv6 address",
+			ip:               net.ParseIP("2001:db8::1"),
+			ifindex:          10,
+			expectedProto:    types.IFAProtOVNK,
+			expectedScope:    int(netlink.SCOPE_UNIVERSE),
+			expectedMask:     128,
+			expectedFlags:    unix.IFA_F_NODAD,
+			expectedValidLft: math.MaxUint32,
+		},
+	}
+
+	for _, tt := range tests {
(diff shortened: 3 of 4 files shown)


arghosh93 (dismissed):
LGTM

arghosh93:
@cgoncalves can you please address failed E2Es?

arghosh93 (dismissed):
LGTM

go-controller/pkg/util/egressip/net_test.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+func TestIfaProtocolOVN_Constant(t *testing.T) {
+	// Verify ifaProtocolOVN is set to 84 (matching RTPROT_OVN)
+	const expectedValue = 84
+	if ifaProtocolOVN != expectedValue {
+		t.Errorf("ifaProtocolOVN = %d, want %d", ifaProtocolOVN, expectedValue)
+	}
+}
```
jcaamano: Isn't this excessive?

go-controller/pkg/util/egressip/net_test.go
```diff
@@ -0,0 +1,104 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			flags := getNetlinkAddressFlag(tt.ip)
+			if flags != tt.expectedFlags {
+				t.Errorf("getNetlinkAddressFlag() = %d, want %d", flags, tt.expectedFlags)
+			}
+		})
+	}
+}
```
jcaamano: Why not check this in the previous test?

go-controller/pkg/util/egressip/net_test.go
```diff
@@ -0,0 +1,104 @@
+package egressip
+
+import (
+	"net"
+	"testing"
+
+	"github.com/vishvananda/netlink"
+)
+
+func TestGetNetlinkAddress(t *testing.T) {
```
jcaamano, line 17: The usefulness of theses tests is limited, since practically they just test constants instead of logic. But no harm done either.

go-controller/pkg/util/egressip/net_test.go
```diff
@@ -0,0 +1,104 @@
...
+		{
+			name:          "IPv4 address should have OVN protocol (84)",
+			ip:            net.ParseIP("192.168.1.100"),
+			ifindex:       5,
+			expectedProto: ifaProtocolOVN,
+			expectedScope: int(netlink.SCOPE_UNIVERSE),
+			expectedMask:  32, // /32 for host address
+		},
+		{
+			name:          "IPv6 address should have OVN protocol (84)",
```
jcaamano: I guess these tests are more generically checking for all properties of the addresses, so test names could better reflect that.

go-controller/pkg/util/egressip/net_test.go
```diff
@@ -0,0 +1,104 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			addr := GetNetlinkAddress(tt.ip, tt.ifindex)
+
+			// Verify Protocol is set to OVN (84)
+			if addr.Protocol != tt.expectedProto {
+				t.Errorf("Protocol = %d, want %d (OVN)", addr.Protocol, tt.expectedProto)
+			}
+
+			// Verify Scope
+			if addr.Scope != tt.expectedScope {
```
jcaamano, line 60: Could check for valid lifetime as well?

jcaamano (commented):
@cgoncalves 

What makes nmstate skip addresses configured with this prototocol?

Have we agreed with OVN team that we are going to use their protocol identifier for this purpose? Wouldn't technically make more sense to register our own?

go-controller/pkg/util/egressip/net_test.go
cgoncalves: Fair point. The flags are already set by `GetNetlinkAddress`, so checking them as part of `TestGetNetlinkAddress` makes more sense.

go-controller/pkg/util/egressip/net_test.go
cgoncalves: Fair. These tests do serve as a regression guard if the function body changes (e.g., someone accidentally drops the Protocol field). I'd keep them.

go-controller/pkg/util/egressip/net_test.go
cgoncalves: OK. I'll rename to something like "IPv4 address" and "IPv6 address" since they're checking all properties.

go-controller/pkg/util/egressip/net_test.go
cgoncalves: I agree, this test is pointless.

go-controller/pkg/util/egressip/net_test.go
cgoncalves: Sure.

cgoncalves:
> What makes nmstate skip addresses configured with this prototocol?

https://github.com/nmstate/nmstate/pull/3129

> Have we agreed with OVN team that we are going to use their protocol identifier for this purpose? Wouldn't technically make more sense to register our own?

RTPROT_OVN (84) is already registered in the IANA routing protocol registry by the OVN project and used by OVN for routes. Using the same identifier for addresses is consistent. OVN-Kubernetes is part of the OVN ecosystem, and these addresses are OVN-managed. Registering a separate protocol ID for ovn-kubernetes specifically would be unnecessary fragmentation.

jcaamano:
> [nmstate/nmstate#3129](https://github.com/nmstate/nmstate/pull/3129)

Can you point me to something more specific? That PR is just a draft that seems to be talking about alt-names, not rtproto specifiers.

> RTPROT_OVN (84) is already registered in the IANA routing protocol registry by the OVN project and used by OVN for routes. Using the same identifier for addresses is consistent. OVN-Kubernetes is part of the OVN ecosystem, and these addresses are OVN-managed. Registering a separate protocol ID for ovn-kubernetes specifically would be unnecessary fragmentation.

When you say `OVN-managed`, you are using OVN/OVNK interchangeably, right? These addresses are managed by OVNK, not OVN. They are the same ecosystem but not the same thing and precisely because they are deployed together, it might be interesting to tell apart which is doing what. In any case, I think we should get an ACK from the OVN team at the very least. Can you reach out to @dceara if he does not see this ping?

cgoncalves:
> > [nmstate/nmstate#3129](https://github.com/nmstate/nmstate/pull/3129)
> 
> Can you point me to something more specific? That PR is just a draft that seems to be talking about alt-names, not rtproto specifiers.

Sorry, I linked the wrong PR. See https://github.com/nmstate/nmstate/pull/3075

> 
> > RTPROT_OVN (84) is already registered in the IANA routing protocol registry by the OVN project and used by OVN for routes. Using the same identifier for addresses is consistent. OVN-Kubernetes is part of the OVN ecosystem, and these addresses are OVN-managed. Registering a separate protocol ID for ovn-kubernetes specifically would be unnecessary fragmentation.
> 
> When you say `OVN-managed`, you are using OVN/OVNK interchangeably, right? These addresses are managed by OVNK, not OVN. They are the same ecosystem but not the same thing and precisely because they are deployed together, it might be interesting to tell apart which is doing what.

RTPROT_OVN (84) has been assigned in the Linux kernel, IProute2 and OVN for IP **routes**.
* https://github.com/torvalds/linux/commit/6002850fdfe0b4343136670a9895b6ba4ee3285b
* https://github.com/iproute2/iproute2/commit/131810fb264fb26b1aba020469447a378cd47b61

In fact, OVN used the arbitrary value `84` for a while until it became assigned in the kernel:
* https://github.com/ovn-org/ovn/commit/189a6ac5e3dc2aba092c78c4d4fd5c61599e1be2

There is nothing other than 4 (values `0` to `3`) assigned for IP **addresses**:
* https://github.com/torvalds/linux/blob/1d5dcaa3bd65f2e8c9baa14a393d3a2dc5db7524/include/uapi/linux/if_addr.h#L73-L77

> In any case, I think we should get an ACK from the OVN team at the very least. Can you reach out to @dceara if he does not see this ping?

Will do.

dceara:
> > [nmstate/nmstate#3129](https://github.com/nmstate/nmstate/pull/3129)
> 
> Can you point me to something more specific? That PR is just a draft that seems to be talking about alt-names, not rtproto specifiers.
> 
> > RTPROT_OVN (84) is already registered in the IANA routing protocol registry by the OVN project and used by OVN for routes. Using the same identifier for addresses is consistent. OVN-Kubernetes is part of the OVN ecosystem, and these addresses are OVN-managed. Registering a separate protocol ID for ovn-kubernetes specifically would be unnecessary fragmentation.
> 
> When you say `OVN-managed`, you are using OVN/OVNK interchangeably, right? These addresses are managed by OVNK, not OVN. They are the same ecosystem but not the same thing and precisely because they are deployed together, it might be interesting to tell apart which is doing what. In any case, I think we should get an ACK from the OVN team at the very least. Can you reach out to @dceara if he does not see this ping?

Thanks @jcaamano for the ping!  You're right, please don't use RTPROT_OVN, it's used by ovn-controller (**not** ovn-kubernetes) to mark routes advertised with ovn's native dynamic-routing support.

While this PR doesn't seem to use it for routes, I'd still suggest you don't use it, it's "ovn specific".

Thanks,
Dumitru

jcaamano:
> Sorry, I linked the wrong PR. See [nmstate/nmstate#3075](https://github.com/nmstate/nmstate/pull/3075)

I see, but that looks like a proposal that wasn't merged?

cgoncalves:
> > When you say `OVN-managed`, you are using OVN/OVNK interchangeably, right? These addresses are managed by OVNK, not OVN. They are the same ecosystem but not the same thing and precisely because they are deployed together, it might be interesting to tell apart which is doing what. In any case, I think we should get an ACK from the OVN team at the very least. Can you reach out to @dceara if he does not see this ping?
> 
> Thanks @jcaamano for the ping! You're right, please don't use RTPROT_OVN, it's used by ovn-controller (**not** ovn-kubernetes) to mark routes advertised with ovn's native dynamic-routing support.
> 
> While this PR doesn't seem to use it for routes, I'd still suggest you don't use it, it's "ovn specific".

Thanks for the quick feedback!

I could either pick an arbitrary number not registered in rtnetlink.h (Linux kernel) like OVN did for a while, or go through the hassle of registering it in there and in IPRoute2. Thoughts?

cgoncalves:
> > Sorry, I linked the wrong PR. See [nmstate/nmstate#3075](https://github.com/nmstate/nmstate/pull/3075)
> 
> I see, but that looks like a proposal that wasn't merged?

@cathay4t's comment from 8 hours ago: "Closing because I heard no feedback from OpenShift team who request this feature/demo."

It has been on hold for two things: this PR and kernel RHEL 9.4 backport (RHEL-146140).

cathay4t:
@cgoncalves Once you get everything ready and nmstate demo tested, feel free to ping nmstate team to ask a reopen. I close that draft PR just for stale.

jcaamano:
> @cathay4t's comment from 8 hours ago: "Closing because I heard no feedback from OpenShift team who request this feature/demo."

This is confusing. By ` OpenShift team`, is it referring to us, OVN-Kubernetes team? It looks like we should have provided feedback but we were never aware of that PR? Or am I missing this request for feedback coming through a different channel? The way this PR described it was as if that functionality was already available in NMState without providing further scope.

Anyway, it looks like a discussion for the overall design needs to happen if it hasn't already. If it did, I missed it.

jcaamano:
> Anyway, it looks like a discussion for the overall design needs to happen if it hasn't already. If it did, I missed it.

So in case I am misunderstanding, if the feedback request is just about setting the RTPROTO on EIP addresses so that other tools can filter them out (regardless of what that tool is, nmstate or any other) then let's bring this up in the community sync. Potentially registering ourselves as a new RTPROTO feels like something we should discuss as a community.

If the feedback is for the overall solution from OpenShift perspective, then I do have doubts (do ignore this if this is not the kid of feedback we are being requested):
* why is this targeting nmstate, which is not always there, if the issue lies with NetworkManager?
* why is it based on configuration that needs to enacted by the user? Looks like it is anybody's problem but the user. Can't this be based on global configuration that CMS can act on?
* how does this work with DHCP?
* have other properties, like labels, that carry less meaningful weight than RTPROTO, been considered?

cathay4t:
> > @cathay4t's comment from 8 hours ago: "Closing because I heard no feedback from OpenShift team who request this feature/demo."
> 
> This is confusing. By ` OpenShift team`, is it referring to us, OVN-Kubernetes team? It looks like we should have provided feedback but we were never aware of that PR? Or am I missing this request for feedback coming through a different channel? The way this PR described it was as if that functionality was already available in NMState without providing further scope.
> 
> Anyway, it looks like a discussion for the overall design needs to happen if it hasn't already. If it did, I missed it.

Nmstate did not move forward that PR from draft to ready for review because we are not sure this actually solve the original problem without test.

I have provided scratch build rpm of nmstate for your team to test and apparently we are blocked by kernel team on support this feature. It has been stale a long time, hence nmstate close it and will revive the code review once confirmed as working solution.

cathay4t:
> > Anyway, it looks like a discussion for the overall design needs to happen if it hasn't already. If it did, I missed it.
> 
> So in case I am misunderstanding, if the feedback request is just about setting the RTPROTO on EIP addresses so that other tools can filter them out (regardless of what that tool is, nmstate or any other) then let's bring this up in the community sync. Potentially registering ourselves as a new RTPROTO feels like something we should discuss as a community.
> 
> If the feedback is for the overall solution from OpenShift perspective, then I do have doubts (do ignore this if this is not the kid of feedback we are being requested):
> 
>     * why is this targeting nmstate, which is not always there, if the issue lies with NetworkManager?

Nmstate will preserve unmentioned configuration, otherwise we will purge a lot of thing if desired state just bring a interface up. Without IP address protocol, nmstate see VIP as static IP address, hence try to preserve it and store it
into NetworkManager configuration.

With IP address protocol introduced, nmstate will ignore static IP with OVS/OVN protocol, hence not causing problem when VIP changed or moved to other node.

> 
>     * why is it based on configuration that needs to enacted by the user? Looks like it is anybody's problem but the user. Can't this be based on global configuration that CMS can act on?

User configuration change is not required, nmstate will ignore any static IP with protocol not 1,2,3(used by kernel).

> 
>     * how does this work with DHCP?

DHCP IP address has `preferred-life-time` and `valid-life-time`, which nmstate can easily filter out from static IP address.

> 
>     * have other properties, like labels, that carry less meaningful weight than RTPROTO, been considered?

The label only exist for IPv4, not IPv6 and kernel team refuse to add label support in IPv6.

cgoncalves:
I got feedback from an upstream kernel uAPI maintainer. He personally would not add OVN-Kubernetes to the IFA_PROT list. The reason was the fact that no other IFAPROT_* is present/added for user-space consumption makes him think this change is possibly not legit (while the OVN route counter-part has a quite rich history of editing).

go-controller/pkg/util/egressip/net.go
```diff
@@ -13,15 +13,27 @@ import (
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
 )
 
+// ifaProtocolOVN is the IFA_PROTO value used to mark addresses as
+// OVN-Kubernetes-managed.
+// Value 85 is used instead of RTPROT_OVN (84) so address protocol does not
+// collide with the route protocol, as recommended by the OVN team.
+const ifaProtocolOVN = 85
```
jcaamano: Can you move this to `types/const.go`  as `IFAProtOVN`?

test/e2e/egressip.go
```diff
@@ -598,6 +598,58 @@ var _ = ginkgo.Describe("e2e egress IP validation", feature.EgressIP, func() {
...
+			gomega.Eventually(func(g gomega.Gomega) {
+				output, err := infraprovider.Get().ExecK8NodeCommand(nodeName,
+					[]string{"ip", "-d", "-j", "addr", "show", "to", eipAddr})
+				g.Expect(err).NotTo(gomega.HaveOccurred(), "failed to get address info on node %s", nodeName)
+
+				var entries []ipAddrEntry
+				g.Expect(json.Unmarshal([]byte(output), &entries)).To(gomega.Succeed(), "failed to parse ip addr JSON")
+
+				found := false
+				for _, entry := range entries {
```
jcaamano: nit: change to only expect a single entry?

cgoncalves: Done

go-controller/pkg/util/egressip/net.go
cgoncalves: Done

go-controller/pkg/types/const.go
```diff
@@ -155,6 +155,13 @@ const (
 	// LabelUserDefinedServiceName label key used in mirrored EndpointSlices that contains the service name matching the EndpointSlice
 	LabelUserDefinedServiceName = "k8s.ovn.org/service-name"
 
+	// IFAProtOVN is the IFA_PROTO value used to mark addresses as
+	// OVN-Kubernetes-managed. Value 85 is used instead of RTPROT_OVN (84)
+	// so address protocol does not collide with the route protocol, as
+	// recommended by the OVN team. IFA_PROTO requires Linux kernel 5.18+;
+	// on older kernels the attribute is silently ignored.
+	IFAProtOVN = 85
```
jcaamano: Actually, sorry I missed this before.

Let's use `IFAProtOVNK`.

And I would amend the comment because I don't think this is about "address protocol does not collide with the route protocol", it's about not using a value that we could foresee as already owned by OVN, and we don't want to use the same values OVN would use.

If OVN eventually sets an interface protocol value on an interface, the would potentially use "84". If OVNK eventually sets an RTPROT on a route, it would potentially be "85". So this is about using a different value than OVN, not about using a different value than what we would use for RTPROT.

test/e2e/egressip.go
```diff
@@ -598,6 +598,57 @@ var _ = ginkgo.Describe("e2e egress IP validation", feature.EgressIP, func() {
...
+				g.Expect(json.Unmarshal([]byte(output), &entries)).To(gomega.Succeed(), "failed to parse ip addr JSON")
+				g.Expect(entries).To(gomega.HaveLen(1), "expected a single interface entry for EgressIP %s on node %s", eipAddr, nodeName)
+
+				var ai *ipAddrInfo
+				for i, info := range entries[0].AddrInfo {
+					if info.Local == eipAddr {
+						ai = &entries[0].AddrInfo[i]
+						break
+					}
+				}
```
jcaamano, line 641: we should also expect only a single addr info, right?

cgoncalves: `ip -j addr show to <eip>` filters which interface is returned, not which addresses are listed on it. The JSON object for that interface still includes every address on the NIC (`addr_info`: node IP, IPv6 link-local, EgressIP, ...).

We already expect a single interface entry (line 633). Requiring `len(addr_info) == 1` is what failed CI in last push (secondary-host NICs and IPv6). We look up the EgressIP in `addr_info` and check `IFA_PROTO` on that match.

go-controller/pkg/types/const.go
cgoncalves: Agree, will rename and reword.

jcaamano (approved):

PR #5881, opened 2026-01-26
EgressIP: Add ecmp_symmetric_reply option to fix connection resets

bpickard22:
When multiple egress IPs are assigned to a namespace, the Logical Router Policy (LRP) has multiple nexthops and uses ECMP load balancing. Without the ecmp_symmetric_reply option, different packets from the same TCP connection can be hashed to different nexthops, resulting in different SNAT IPs and connection resets from the remote server.

This adds the ecmp_symmetric_reply option to ensure all packets from the same connection use the same nexthop, maintaining consistent SNAT IP and preventing connection resets.

This aligns with the existing pattern used in egress gateway (egressgw.go) and APB routes (apbroute/network_client.go).

Fixes: Connection resets with multiple egress IPs and external endpoints

Assisted-by: Claude Sonnet 4.5

## Summary by CodeRabbit

* **Bug Fixes**
  * Egress IP reroute policies now enable ECMP symmetric-reply by default, ensuring consistent bidirectional routing behavior across network types.

* **Tests**
  * Unit tests and fixtures updated to validate symmetric-reply behavior in reroute policies and related NAT entries.

### DIFF
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -2970,6 +2970,9 @@ func (e *EgressIPController) createReroutePolicyOps(ni util.NetInfo, ops []ovsdb
 		}
 		addPktMarkToLRPOptions(options, mark.String())
 	}
+	// Enable connection-tracked ECMP to ensure all packets from the same connection
+	// use the same nexthop, preventing connection resets with multiple egress IPs
+	options["ecmp_symmetric_reply"] = "true"
 	dbIDs := getEgressIPLRPReRouteDbIDs(egressIPName, podNamespace, podName, ipFamily, ni.GetNetworkName(), e.controllerName)
 	p := libovsdbops.GetPredicate[*nbdb.LogicalRouterPolicy](dbIDs, nil)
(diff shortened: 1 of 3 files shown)


bpickard22:
Hey @booxter, do you plan on reviewing this or should I try to get some review in the upstream slack?

booxter:
@bpickard22 sorry I don't regularly do reviews here. The automatic system assigns me to PRs but it really shouldn't. (I think it's because I was added as a member of the org to have access to features like bug assignment.) Feel free to find a better reviewer.

@tssurya @trozet FYI

dceara:
Hi @bpickard22!

OVN's Logical_Router_Policy doesn't support ecmp_symmetric_reply behavior.  That's only supported for Logical_Router_Static_Route records:

https://github.com/ovn-org/ovn/blob/714d1eac0a775338129813961a79654902a11f82/ovn-nb.xml#L4910-L4934

So setting it on a Logical_Router_Policy will have no effect.

Regards,
Dumitru

bpickard22:
will come up with a different approach to handle tcp connection drops with multiple EIPs due to @dceara comment

PR #5884, opened 2026-01-28
 Allow emitting metrics on a single endpoint

### LINKED ISSUE #5869: Consolidate OVN K8s metrics and OVN metrics endpoint in ovnkube-node
### What is this issue about?

Currently ovnkube-node exposes two separate metrics endpoints:
```
	&cli.StringFlag{
		Name:        "metrics-bind-address",
		Usage:       "The IP address and port for the OVN K8s metrics server to serve on (set to 0.0.0.0 for all IPv4 interfaces)",
		Destination: &cliConfig.Metrics.BindAddress,
	},
	&cli.StringFlag{
		Name:        "ovn-metrics-bind-address",
		Usage:       "The IP address and port for the OVN metrics server to serve on (set to 0.0.0.0 for all IPv4 interfaces)",
		Destination: &cliConfig.Metrics.OVNMetricsBindAddress,
	},
```

In our environment, we use a single endpoint that emits both OVN-Kubernetes metrics and OVN metrics.
I would like to understand whether it is acceptable to consolidate these into a single endpoint upstream.

Questions:
1. Is it acceptable to expose both OVN-Kubernetes metrics and OVN metrics on a single endpoint?
2. Alternatively, would it make sense to introduce a feature flag that allows the OVN-Kubernetes metrics endpoint to emit all metrics (i.e., merge both)?

l8huang:
## 📑 Description

Changes:
-  Unify the metrics servers used by ovnkube-node and OVS/OVN metrics
- Allow emitting metrics on a single endpoint

Fixes #5869
 
## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Option to combine OVN/OVS and ovnkube-node metrics into a single configurable endpoint.
  * Ability to reuse the default Prometheus registry when endpoints are combined.
  * Added optional pprof endpoints and a runtime log-level flag endpoint for profiling and live log-level control.

* **Tests**
  * Updated tests to support configurable metric registries and gathering paths.

### DIFF
--- a/go-controller/cmd/ovnkube/ovnkube.go
+++ b/go-controller/cmd/ovnkube/ovnkube.go
@@ -14,6 +14,7 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/urfave/cli/v2"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -261,6 +262,15 @@ func determineOvnkubeRunMode(ctx *cli.Context) (*ovnkubeRunMode, error) {
 	return mode, nil
 }
 
+// Determine if we should serve both ovnkube-node and OVN/OVS metrics on a single endpoint.
+func combineMetricsEndpoints(runMode *ovnkubeRunMode) bool {
+	return runMode != nil &&
+		runMode.node &&
+		config.Metrics.BindAddress != "" &&
+		config.Metrics.BindAddress == config.Metrics.OVNMetricsBindAddress &&
+		config.OvnKubeNode.Mode != types.NodeModeDPUHost
+}
+
 func startOvnKube(ctx *cli.Context, cancel context.CancelFunc) error {
 	pidfile := ctx.String("pidfile")
 	if pidfile != "" {
@@ -311,9 +321,9 @@ func startOvnKube(ctx *cli.Context, cancel context.CancelFunc) error {
 
 	eventRecorder := util.EventRecorder(ovnClientset.KubeClient)
 
-	// Start metric server for master and node. Expose the metrics HTTP endpoint if configured.
+	// Start the general metrics server only when not combined.
 	// Non LE master instances also are required to expose the metrics server.
-	if config.Metrics.BindAddress != "" {
+	if config.Metrics.BindAddress != "" && !combineMetricsEndpoints(runMode) {
 		metrics.StartMetricsServer(config.Metrics.BindAddress, config.Metrics.EnablePprof,
 			config.Metrics.NodeServerCert, config.Metrics.NodeServerPrivKey, ctx.Done(), ovnKubeStartWg)
 	}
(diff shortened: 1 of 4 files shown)


go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -261,6 +262,13 @@ func determineOvnkubeRunMode(ctx *cli.Context) (*ovnkubeRunMode, error) {
 	return mode, nil
 }
 
+// Determine if we should serve both opvnkube-node and OVN/OVS metrics on a single endpoint.
+func combineMetricsEndpoints() bool {
+	return config.Metrics.BindAddress != "" &&
+		config.Metrics.BindAddress == config.Metrics.OVNMetricsBindAddress &&
+		config.OvnKubeNode.Mode != types.NodeModeDPUHost
+}
```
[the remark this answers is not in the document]
l8huang: The modes using leader election (cluster-manager, ovnkube-controller without node) are control-plane components that don't run OVS/OVN locally—they don't have `--ovn-metrics-bind-address` option, socombineMetricsEndpoints() returns false and StartMetricsServer is called normally.

Please re-evaluate your comment.

l8huang:
@npinaeva Could you please review the PR? Thanks.

go-controller/pkg/metrics/server.go
```diff
@@ -63,68 +69,92 @@ type MetricServer struct {
...
+	}
+	gatherer := opts.Gatherer
+	if gatherer == nil {
+		if reg, ok := registerer.(prometheus.Gatherer); ok {
+			gatherer = reg
+		} else {
+			gatherer = prometheus.DefaultGatherer
+		}
+	}
+
```
npinaeva, line 81: I think it would be easier to understand if we simply pass these default values where we pass nil now. Then the test cases will be able to call `gatherer.Gather()` without touching server internal fields, and we probably won't need to store `gatherer` in the server struct at all

go-controller/cmd/ovnkube/ovnkube.go
```diff
@@ -261,6 +262,13 @@ func determineOvnkubeRunMode(ctx *cli.Context) (*ovnkubeRunMode, error) {
 	return mode, nil
 }
 
+// Determine if we should serve both opvnkube-node and OVN/OVS metrics on a single endpoint.
```
npinaeva: typo opvnkube?

go-controller/cmd/ovnkube/ovnkube.go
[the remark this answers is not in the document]
npinaeva: I think it may be useful though to double-check in the code that those modes don't skip server creation. Just if someone sets `ovn-metrics-bind-address` by accident thinking it won't do anything on master anyway, there is no unexpected side effects?

go-controller/cmd/ovnkube/ovnkube.go
l8huang: fixed

go-controller/cmd/ovnkube/ovnkube.go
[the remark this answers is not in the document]
l8huang: updated combineMetricsEndpoints() to take runMode and returns true only for node mode with matching non-empty bind addresses, excluding DPU host -- Central ovnkube-master and IC ovnkube-control-plane always return false.

go-controller/pkg/metrics/server.go
l8huang: Simplified this part to remove gatherer.

npinaeva (approved):
nice cleanup!

Issue #5854, opened 2026-01-16
Reduce setup time for tests in network-segmentation lane


tssurya:
### What is this issue about?

Reduce the network segmentation lane test run time by inspecting where its taking the most time.

PR #5886, opened 2026-01-28
CNC: add status report from the cluster manager.

npinaeva:
This requires moving selected networks discovery first for future cross-network validation.
Add clusternetworkconnects/status patch rights to cluster manager, also update helm CM yaml with missed CNC rights.
Update Patch reactor for unit tests to apply status conditions. Fix some error wrapping for potential future use with errors.As().
Added e2e test for 1 of the more complex errors, hoping we don't need to duplicate every error condition from the unit tests there.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * ClusterNetworkConnect now publishes explicit status conditions with validation feedback and controlled transition timestamps; reconciliation runs in a synchronized flow with deterministic subnet and tunnel-key allocation and cleanup on deletion.

* **Validation**
  * Cross-connection overlap checks, IP-family compatibility checks, and cluster-subnet overlap prevention added.

* **Tests**
  * Expanded unit and e2e tests, new subnet allocator for tests, and status assertion helpers.

* **Chores**
  * RBAC updated and improved error/status handling.

### DIFF
--- a/go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
+++ b/go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"net"
+	"slices"
 	"time"
 
 	nadlisters "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/client/listers/k8s.cni.cncf.io/v1"
@@ -366,7 +368,7 @@ func (c *Controller) crossValidateCNCs(cncState *clusterNetworkConnectState, dis
 }
 
 func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetworkConnect) ([]util.NetInfo, sets.Set[string], error) {
-	discoveredNetworks := []util.NetInfo{}
+	discoveredNetworks := map[string]util.NetInfo{}
 	allMatchingNADKeys := sets.New[string]()
 	var errs []error
 
@@ -406,7 +408,7 @@ func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetwo
 				// This NAD passed all validation checks, so it's selected by this CNC
 				nadKey := nad.Namespace + "/" + nad.Name
 				allMatchingNADKeys.Insert(nadKey)
-				discoveredNetworks = append(discoveredNetworks, network)
+				discoveredNetworks[network.GetNetworkName()] = network
 			}
 		case apitypes.PrimaryUserDefinedNetworks:
 			namespaceSelector, err := metav1.LabelSelectorAsSelector(&selector.PrimaryUserDefinedNetworkSelector.NamespaceSelector)
@@ -430,14 +432,14 @@ func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetwo
 					continue
 				}
 				allMatchingNADKeys.Insert(nadKey)
-				discoveredNetworks = append(discoveredNetworks, namespacePrimaryNetwork)
+				discoveredNetworks[namespacePrimaryNetwork.GetNetworkName()] = namespacePrimaryNetwork
 			}
 		default:
 			errs = append(errs, fmt.Errorf("%w: unsupported network selection type %s", errConfig, selector.NetworkSelectionType))
 		}
 	}
 
-	return discoveredNetworks, allMatchingNADKeys, kerrors.NewAggregate(errs)
+	return slices.Collect(maps.Values(discoveredNetworks)), allMatchingNADKeys, kerrors.NewAggregate(errs)
 }
 
 // allocateSubnets allocates subnets for the given discovered networks


npinaeva:
I fixed some tests that were failing because we used UDNs with overlapping subnets :D Maybe I haven't fixed all of them (havent' checked last failures), but I think it should be ready for review

dist/templates/rbac-ovnkube-cluster-manager.yaml.j2
```diff
@@ -91,6 +91,7 @@ rules:
           - clusteruserdefinednetworks/finalizers
           - routeadvertisements/status
           - clusternetworkconnects # Unlike core kubernetes objects, for CRDs there is strict enforcement of status subresource permissions. Since CM has to annotate CNC we need to provide full object patch permissions.
+          - clusternetworkconnects/status
```
tssurya, line 94: given we provide full object update access, is it necessary to separately also provide it for /status?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -9,19 +10,24 @@ import (
...
 
 	networkconnectv1 "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/clusternetworkconnect/v1"
+	cncapply "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/clusternetworkconnect/v1/apis/applyconfiguration/clusternetworkconnect/v1"
 	apitypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
 	ovntypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
 )
 
+const fieldManager = "clustermanager-clusternetworkconnect-controller"
```
tssurya: nit: `clustermanager-networkconnect-controller` ? I kept the folder name and controller name as `networkconnect` to see if in future we end up having to do `NetworkConnect` namespace scoped CRD for tenants.. we can reuse this same elements/controller/code

npinaeva:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> ⚠️ Outside diff range comments (1)
> 
> > test/e2e/cluster_network_connect.go (1)> `2218-2230`: _⚠️ Potential issue_ | _🟡 Minor_
> > > **Connectivity timeouts may be too tight for CI environments.**
> > > These `Eventually` calls use a 5-second timeout for datapath connectivity checks. Even though CNC annotation convergence is verified beforehand (with 60s waits), OVN flow programming happens asynchronously and may not complete within 5s, especially cross-node. Consider bumping to at least 30s for the `expectSuccess` path and the initial `Eventually` in the `expectFailure` path.
> > > 

@tssurya maybe I shouldn't touch the retries and it can actually take up to 5s to sync? I didn't think it was the case at least for test clusters, but don't want to introduce any flakes because of it either. I was running this test locally and got tired waiting for completion :D

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -81,9 +87,47 @@ func getPrimaryNADForNamespace(networkMgr networkmanager.Interface, namespaceNam
 	return primaryNADKey, namespacePrimaryNetwork, nil
 }
 
+func (c *Controller) updateStatus(cnc *networkconnectv1.ClusterNetworkConnect, e error) {
+	klog.V(5).Infof("Updating status for cluster network connect %s, err=%v", cnc.Name, e)
```
tssurya: should we put this log after we successfully apply the status? cause sometimes we could be just returning at L108 in which case no updating is happening.

tssurya:
> > Caution
> > Some comments are outside the diff and can’t be posted inline due to platform limitations.
> > ⚠️ Outside diff range comments (1)
> > > test/e2e/cluster_network_connect.go (1)> `2218-2230`: _⚠️ Potential issue_ | _🟡 Minor_
> > > > **Connectivity timeouts may be too tight for CI environments.**
> > > > These `Eventually` calls use a 5-second timeout for datapath connectivity checks. Even though CNC annotation convergence is verified beforehand (with 60s waits), OVN flow programming happens asynchronously and may not complete within 5s, especially cross-node. Consider bumping to at least 30s for the `expectSuccess` path and the initial `Eventually` in the `expectFailure` path.
> 
> @tssurya maybe I shouldn't touch the retries and it can actually take up to 5s to sync? I didn't think it was the case at least for test clusters, but don't want to introduce any flakes because of it either. I was running this test locally and got tired waiting for completion :D

I am fixing the CI time here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828/changes/8a00cdcb85ca1c334bc6638c84f84c6631d44df8 see https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5854 so you can leave that part out of this PR yea - it will land with the services changes..
and bad bot.. if it takes OVN 5seconds to sync something I would be super upset as an end-user

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -81,9 +87,47 @@ func getPrimaryNADForNamespace(networkMgr networkmanager.Interface, namespaceNam
...
+	condition := metaapply.Condition().
+		WithType("Accepted").
+		WithStatus("True").
+		WithReason("ValidationSucceeded").
+		WithMessage("All validation checks passed successfully")
+	if e != nil {
+		condition = condition.WithStatus("False").
+			WithReason("ValidationFailed").
```
tssurya: I think the `ValidationFailed` is generic enough for the validate CNC step you are adding
but if something failed during allocation etc would that be considered Validation?
i.e Steps for discovery, subnet allocation, annotation update and tunnelID allocation errors - they won't be considered validationfailed right? ie 

1. either we can make this name more generic for the reason OR
2. we move the `updateStatus` inside `reconcileClusterNetworkConnect` to be two different conditions - one for `Validation` another for `Setup`

offline I was ok with single overarching umbrella validation so option1. not sure if @trozet had other thoughts

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -81,9 +87,47 @@ func getPrimaryNADForNamespace(networkMgr networkmanager.Interface, namespaceNam
...
+		cncapply.ClusterNetworkConnect(cnc.Name).WithStatus(
+			cncapply.ClusterNetworkConnectStatus().WithConditions(condition),
+		),
+		metav1.ApplyOptions{
+			FieldManager: fieldManager,
+			Force:        true,
+		},
+	)
+	if err != nil {
+		klog.Errorf("Failed to update status for cluster network connect %s: %v", cnc.Name, err)
```
tssurya, line 131: nit, do we want to wrap this error with %w? not that we are parsing it anywhere specifically..

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -196,6 +266,102 @@ func (c *Controller) reconcileClusterNetworkConnect(key string) error {
 	return kerrors.NewAggregate(errs)
 }
 
+// validateCNC performs validation checks for the given CNC connectNetworks and discovered networks.
+func validateCNC(cncState *clusterNetworkConnectState, discoveredNetworks []util.NetInfo) error {
+	// Check for overlap with cluster networks.
+	// This includes service CIRD and masquerade CIDR
```
tssurya: nit: `service CIDR`

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -81,9 +87,47 @@ func getPrimaryNADForNamespace(networkMgr networkmanager.Interface, namespaceNam
...
+	klog.V(5).Infof("Updating status for cluster network connect %s, err=%v", cnc.Name, e)
+	condition := metaapply.Condition().
+		WithType("Accepted").
+		WithStatus("True").
+		WithReason("ValidationSucceeded").
+		WithMessage("All validation checks passed successfully")
+	if e != nil {
+		condition = condition.WithStatus("False").
+			WithReason("ValidationFailed").
+			WithMessage(e.Error())
```
tssurya: I forget how we do it for the other features.. do we need to truncate the length of the error ? IIRC the max length for message is ~36000 - unsure if we do it for other features or if its auto taken care of - sometimes with multiple returns not sure if this ends up being a really big message that will end up complaining out of length..

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -196,6 +266,102 @@ func (c *Controller) reconcileClusterNetworkConnect(key string) error {
 	return kerrors.NewAggregate(errs)
 }
 
+// validateCNC performs validation checks for the given CNC connectNetworks and discovered networks.
+func validateCNC(cncState *clusterNetworkConnectState, discoveredNetworks []util.NetInfo) error {
+	// Check for overlap with cluster networks.
+	// This includes service CIRD and masquerade CIDR
+	if err := util.CheckSubnetOverlapWithClusterSubnets(cncState.connectSubnets, "ClusterNetworkConnect.connectSubnets"); err != nil {
```
tssurya, line 278: as of today - we don't allow mutation of subnets in networks and in connectsubnets though we plan to in the future - so until then do we see benefit in doing this only during create? or given the number of networks per CNC is expected to be pretty small, its not very worrying for perf/ when done with each reconcile? i.e given you are storing the connectSubnets in cncCache state - at least for that we know if something changed there and only then we need to revalidate for old!=new cache value? since serviceCIDR/masqueradeCIDR is also rare range not sure if its worth doing this every time in reconcile unless the value of connect subnets changed...
also applicable to the next validation check per UDN

go-controller/pkg/util/multi_network.go
```diff
@@ -2273,3 +2273,76 @@ func getFirstNonOverlappingSubnet(subnet1, subnet2 *net.IPNet, netMask net.IPMas
...
+					si.GetNetworkName(), si.Subnets(), sj.GetNetworkName(), sj.Subnets())
+			}
+		}
+	}
+	return nil
+}
+
+// CheckSubnetOverlapWithNetworks checks whether the given subnets overlap with
+// any of the given networks' subnets, transit subnets, or join subnets.
+// subnetsName is used in the error message to explain where those subnets are coming from.
```
tssurya, line 2291: I think we don't care about overlap of connectsubnet with default network's transit/cluster-subnet/join so its ok we are not checking for those

go-controller/pkg/util/multi_network.go
```diff
@@ -2273,3 +2273,76 @@ func getFirstNonOverlappingSubnet(subnet1, subnet2 *net.IPNet, netMask net.IPMas
...
+
+	if NetworksOverlap(subnets, []*net.IPNet{v4MasqueradeCIDR, v6MasqueradeCIDR}) {
+		return fmt.Errorf("%s with subnets %v and %s with subnets %v",
+			config.ConfigSubnetMasquerade, []*net.IPNet{v4MasqueradeCIDR, v6MasqueradeCIDR},
+			subnetsName, subnets)
+	}
+	return nil
+}
+
+func NetworksOverlap(n1, n2 []*net.IPNet) bool {
```
tssurya: does this belong in util/net.go near `IPNetOverlaps` ? most people won't think of checking here and we don't want more duplicate overlap utils

go-controller/pkg/util/multi_network.go
```diff
@@ -2273,3 +2273,76 @@ func getFirstNonOverlappingSubnet(subnet1, subnet2 *net.IPNet, netMask net.IPMas
 	nextIP := iputils.NextIP(baseSubnetLastIP.IP)
 	return &net.IPNet{IP: nextIP, Mask: netMask}
 }
+
+func cidrNetworkEntriesToNetIPs(cidrEntries []config.CIDRNetworkEntry) []*net.IPNet {
```
tssurya: I think `GetAllClusterSubnetsFromEntries` in kube.go util does the same

go-controller/pkg/util/multi_network.go
```diff
@@ -2273,3 +2273,76 @@ func getFirstNonOverlappingSubnet(subnet1, subnet2 *net.IPNet, netMask net.IPMas
...
+		netIPs = append(netIPs, entry.CIDR)
+	}
+	return netIPs
+}
+
+func CheckNetworksOverlap(subnets []NetInfo) error {
+	for i, si := range subnets {
+		for j := 0; j < i; j++ {
+			sj := subnets[j]
+			if NetworksOverlap(cidrNetworkEntriesToNetIPs(si.Subnets()), cidrNetworkEntriesToNetIPs(sj.Subnets())) {
```
tssurya: will preconverting once in a prior loop help? just thinking if calling cidrNetworkEntriesToNetIPs inloop is doing this N^2 times conversions/allocations v/s perhaps doing it once before is an extra loop but maybe better - hard to tell though

go-controller/pkg/util/multi_network.go
```diff
@@ -2273,3 +2273,76 @@ func getFirstNonOverlappingSubnet(subnet1, subnet2 *net.IPNet, netMask net.IPMas
...
+			config.ConfigSubnetMasquerade, []*net.IPNet{v4MasqueradeCIDR, v6MasqueradeCIDR},
+			subnetsName, subnets)
+	}
+	return nil
+}
+
+func NetworksOverlap(n1, n2 []*net.IPNet) bool {
+	for _, s1 := range n1 {
+		for _, s2 := range n2 {
+			if s1.Contains(s2.IP) || s2.Contains(s1.IP) {
```
tssurya: there is also `IsContainedInAnyCIDR` that's used for the default network side.. I think what you have is more efficient? that checks the network address alone that's sufficient I think?
in general there is some overlap with the utils we have there for default network v/s UDN subnet collisions but those are more hardcoded/confined to the configs so I guess its ok, still there is some room to consolidate all of these into some confict-detection file within utils maybe - but I'm fine if that's not done in this PR

helm/ovn-kubernetes/charts/ovnkube-control-plane/templates/rbac-ovnkube-cluster-manager.yaml
```diff
@@ -78,6 +78,7 @@ rules:
           - clusteruserdefinednetworks
           - routeadvertisements
           - vteps
+          - clusternetworkconnects
```
tssurya, line 81: isn't there a helm lane for net-seg? probably not ... nice catch !

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -126,12 +180,35 @@ func (c *Controller) reconcileClusterNetworkConnect(key string) error {
...
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to discover selected networks for CNC %s: %w", cncName, err))
+	}
+	// STEP2: Validate the CNC
+	if err = validateCNC(cncState, discoveredNetworks); err != nil {
+		return fmt.Errorf("validation failed for CNC %s: %w", cncName, err)
+	}
+	if err = c.crossValidateCNCs(cncState, discoveredNetworks); err != nil {
+		return fmt.Errorf("cross-validation failed for CNC %s: %w", cncName, err)
+	}
```
tssurya, line 214: another one is https://ovn-kubernetes.io/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns/#connecting-udns-with-no-overlay-mode - to check if `Transport` is set to `NoOverlay` - we return validationFailed..

but the EVPN OKEP landed after CNC OKEP, not sure if @trozet considered a section for NetworkConnect there - can't see it in https://ovn-kubernetes.io/okeps/okep-5088-evpn/#feature-compatibility but there given the design is https://ovn-kubernetes.io/okeps/okep-5088-evpn/#using-evpn-as-the-overlay-for-user-defined-networks not going to be via GENEVE I.e not going to be via OVN `encap-type` our feature cannot support cross connecting such features that choose different mode of east-west. For EVPN the truth is we didn't consider the design of how that's gonna look like, but given we aren't using OVN path there at the very least we should just report error status to user if they are selecting networks of `transport` type either no-overlay or EVPN -  we are not supporting day2 changes in network spec for transport

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
tssurya: EDIT: I realised we can select/deselect networks anytime so we can't optimize much of this - only the serviceCIDR or masCIDR or some things can be optimised..

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -196,6 +266,102 @@ func (c *Controller) reconcileClusterNetworkConnect(key string) error {
...
+	// That includes overlaps with pod subnet, transit and join subnets.
+	if err := util.CheckSubnetOverlapWithNetworks(cncState.connectSubnets, "ClusterNetworkConnect.connectSubnets", discoveredNetworks); err != nil {
+		return fmt.Errorf("selected networks overlap with cluster network connect subnets: %w", err)
+	}
+	// Now check for overlaps among the discovered networks themselves.
+	if err := util.CheckNetworksOverlap(discoveredNetworks); err != nil {
+		return fmt.Errorf("can not connect selected networks: selected networks have overlapping subnets: %w", err)
+	}
+	// Also check ipFamily compatibility between discovered networks and connectSubnets
+	// The only incompatible case is when discovered networks contain both ipv4-only and ipv6-only networks.
```
tssurya, line 291: self-note: I should probably add a e2e for connecting a v4 with dualstack and see what happens if i curled the v6 IP from the v4 pod. probably its gonna fail but thats expected..

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -196,6 +266,102 @@ func (c *Controller) reconcileClusterNetworkConnect(key string) error {
...
+	// Step 2: check for connectSubnets overlap with those CNCs
+
+	// Build a set of selected networks for this CNC.
+	// We can't use cncState.selectedNetworks because it is only updated at the end of the sync,
+	// and validation should happen before that.
+	currentCNCSelectedNetworks := sets.New[string]()
+	for _, network := range discoveredNetworks {
+		networkID := network.GetNetworkID()
+		if networkID == ovntypes.NoNetworkID {
+			continue
```
tssurya, line 348: i think we'll retry when we get a network id so this should be ok and get retried in another sync again?

tssurya (changes requested):
overall its super nice, thanks for implementing the status, our users are going to like this!
did a first pass, I didn't get a chance to review tests will come back for that

dist/templates/rbac-ovnkube-cluster-manager.yaml.j2
npinaeva: yeah I get surprised every time, but they are different resources :D we have some examples https://github.com/ovn-kubernetes/ovn-kubernetes/blob/8dffca239141811cb6d3a1a04d98b96f5cb1d6ca/dist/templates/rbac-ovnkube-cluster-manager.yaml.j2#L87-L90

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: oooh that is deep.. yeah I think it is unlikely that we will need to distinguish them anyway because they will be updating different resources, so let's try that

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: I think that is why I had it at the beginning for debugging :D
Maybe I can just remove it, you can see the result hopefully in the CR.status?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: yeah that is exactly what I was thinking: I am not sure if having 2 conditions brings extra value, because you probably can't do much with one of them being OK and another being failed.
So we mainly allocate things like tunnel ID and subnets in the function (after validating), so should we call it `ResourceAllocationSucceeded/Failed`?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: yeah exactly, I think compared to cross-CNC validation this is nothing :D

go-controller/pkg/util/multi_network.go
npinaeva: yeah default network checks are in the `CheckSubnetOverlapWithClusterSubnets`, hopefully I put all the right stuff in there

go-controller/pkg/util/multi_network.go
npinaeva: yeah good point

go-controller/pkg/util/multi_network.go
npinaeva: I knew it should be somewhere, but couldn't find!

go-controller/pkg/util/multi_network.go
npinaeva: yeah I can optimize a bit

go-controller/pkg/util/multi_network.go
npinaeva: I was trying to, but not sure if this is possible, e.g. we have `(cs *ConfigSubnets) CheckForOverlaps()` that does a similar thing, but with a special struct and returns a special error type
I was keeping in mind cross-udn checks for route advertisements, but the PR implementing it also organizes everything into special structs... so considering there isn't much logic a lot depends on data formats that are always different, so i tried to keep some simple functions like this one that myabe will be reused and more specific ones like  `CheckSubnetOverlapWithNetworks`

helm/ovn-kubernetes/charts/ovnkube-control-plane/templates/rbac-ovnkube-cluster-manager.yaml
npinaeva: nope, do we want to add one?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: oh this one slipped away, good catch!
So maybe a good place for this check is in the `discoverSelectedNetworks` to not even put this network to the discovered list and aggregate errors?

helm/ovn-kubernetes/charts/ovnkube-control-plane/templates/rbac-ovnkube-cluster-manager.yaml
tssurya: I wouldn't want to make that part of this PR but if you have time maybe converting the "dynanic udns or one of the other lanes we have for net-seg into helm" is a good idea.. I think tim plans to do this for all features and get rid of non-helm lanes as part of 1.3 release based on what he committed in previous u/s meeting

go-controller/pkg/util/multi_network.go
tssurya: makes sense - I don't know if its worth adding some developer docs or maybe whoever re-dos it for RA will hopefully try to reuse what you have done and not add more specific structs...

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
tssurya: true :D speaking of.. do you think we need to do some scale testing for this or shall we take care of that post merge and think of improvements if we see this becoming a bottleneck?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
tssurya: yup I think that having two different conditions is not going to change anything for the user who shouldn't have to know the implementation details 
I'm fine calling it `ResourceAllocationSucceeded` and `ResourceAllocationFailed`
and should we think of emitting an event in case of failure /use that for a metric? I didn't add any metrics section for this feature I think (my bad) - I don't want to expand the scope of this PR - so if you prefer we track that separately that's fine too

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
tssurya: yea +1 to remove, seeing it in status is clear enough

dist/templates/rbac-ovnkube-cluster-manager.yaml.j2
tssurya: haha ack I see now.. I vaguely remember needing both elsewhere as well.. resolving this.

go-controller/pkg/util/multi_network.go
tssurya: yea I think yes, cause its checking for service cidrs and masquerade which makes sense.. we don't care about default network's pod/join/transit subnet much since that's not connectable to other networks via CNC selector, so we are good here.

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: good point, I think we only do that for ANP, but we really should do it for everything. Let me see if I can come up with some function

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: yeah metrics is a good idea, I think we could use some standardization across the repo, so it would be nice to cover in a separate PR

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: I think compared to all the other heavy-lifting CNC is doing, this cross-object check shouldn't be a bottleneck, especially since we don't do any extra parsing and only use existing info

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: well that one turned out to be correct: this is a log message and not a returned error :D

test/e2e/cluster_network_connect.go
```diff
@@ -123,6 +123,22 @@ func generateNetworkSubnets(cs clientset.Interface, topology, v4Subnet, v6Subnet
...
 }
 
+// newTestNetworkSubnetsAllocator returns a function that yields unique (v4CIDR, v6CIDR) pairs
+// for UDN/CUDN creation within a single test.
+//
+// This prevents multiple networks in the same test from accidentally sharing the same CIDRs.
+func newTestNetworkSubnetsAllocator() func() (string, string) {
+	i := 0
+	return func() (string, string) {
+		i++
```
tssurya, line 133: oo nice, smart way to avoid conflicts..

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -229,10 +400,15 @@ func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetwo
 				if !network.IsPrimaryNetwork() {
 					continue
 				}
+				if network.Transport() == ovntypes.NetworkTransportNoOverlay || network.Transport() == ovntypes.NetworkTransportEVPN {
```
tssurya, line 403: nit: would be good to have a quick unit test but not a biggie

npinaeva: ah good point

go-controller/pkg/clustermanager/networkconnect/controller_test.go
```diff
@@ -1248,6 +1253,334 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
...
+			nad := newTestUDNNADWithSubnets(udnName, namespace, network, networkID, subnets)
+			_, err = fakeClientset.NetworkAttchDefClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespace).Create(
+				context.Background(), nad, metav1.CreateOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			netInfo, err := util.ParseNADInfo(nad)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			mutableNetInfo := util.NewMutableNetInfo(netInfo)
+			mutableNetInfo.AddNADs(namespace + "/" + udnName)
+			fakeNM.PrimaryNetworks[namespace] = mutableNetInfo
```
tssurya, line 1329: btw we should lock this cache change.. I'm hitting the race in my CNC services PR cause I forget to do this in my tests :) so thought of mentioning it here as well
```
					fakeNM.Lock()
					fakeNM.PrimaryNetworks[namespace1.Name] = networkConfig2
					fakeNM.Unlock()
```
seems to be the pattern in other feature testing

tssurya: ah no ignore this - its during startup we are safe.. I had a place where I was modifying it during the tests.

npinaeva:
ah I pushed rebase and changes at same time sorry
the only changes are unit test in `controller_test`: added locking for fakeNM and a new test case for no overlay and evpn

tssurya:
thanks @npinaeva !

Issue #5876, opened 2026-01-26
Coredump found ovnkube pod - network-segmentation lane `RegisterOvnControllerMetrics`


tssurya:
### Which jobs are flaking?

network segmentation, noHA, SGW, dualstack, noSNATGW, 1br, IC, single node, disable...something I can't read-cause-its-too-long

### Which tests are flaking?

```
Summarizing 1 Failure:
  [ABORTED] e2e egress IP validation Network Segmentation: IPv4 L3 role primary [JustAfterEach] [OVN network] Using different methods to disable a node's availability for egress Should validate the egress IP functionality against remote hosts disabling egress nodes with egress-assignable label [Feature:EgressIP]
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/util.go:1236
```

### Since when has it been flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21331694570/job/61397685565?pr=5874

### Reason for failure (if possible)

```
(dlv) bt
 0  0x0000000000483141 in runtime.raise
    at /usr/local/go/src/runtime/sys_linux_amd64.s:154
 1  0x000000000045ae85 in runtime.dieFromSignal
    at /usr/local/go/src/runtime/signal_unix.go:967
 2  0x000000000045b4c6 in runtime.sigfwdgo
    at /usr/local/go/src/runtime/signal_unix.go:1179
 3  0x0000000000459d45 in runtime.sigtrampgo
    at /usr/local/go/src/runtime/signal_unix.go:432
 4  0x0000000000483141 in runtime.raise
    at /usr/local/go/src/runtime/sys_linux_amd64.s:153
 5  0x000000000045ae85 in runtime.dieFromSignal
    at /usr/local/go/src/runtime/signal_unix.go:967
 6  0x000000000044171a in runtime.crash
    at /usr/local/go/src/runtime/signal_unix.go:1056
 7  0x000000000044171a in runtime.fatalthrow.func1
    at /usr/local/go/src/runtime/panic.go:1287
 8  0x000000000047fa6a in runtime.systemstack
    at /usr/local/go/src/runtime/asm_amd64.s:514
 9  0x000000000047fa08 in runtime.systemstack_switch
    at /usr/local/go/src/runtime/asm_amd64.s:479
10  0x0000000000441678 in runtime.fatalthrow
    at /usr/local/go/src/runtime/panic.go:1276
11  0x00000000004412dc in runtime.fatal
[...]

PR #5888, opened 2026-01-28
CUDN: cleanup NADs in terminating namespaces without pods

kyrtapz:
Skip namespaces with deletionTimestamp set when selecting target namespaces, triggering NAD deletion for terminating namespaces.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter handling of deleted/terminating namespaces: terminating namespaces are skipped from normal selection, marked as deleting, and now trigger targeted cleanup reconciliations (with deletion-aware logging and safer nil-handling) to remove stale network attachments.

* **Tests**
  * Added unit and end-to-end tests validating NetworkAttachmentDefinition lifecycle: ensure NADs are removed when target namespaces terminate and created when namespace selection changes.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go
@@ -46,10 +46,11 @@ func NewNamespaceNotifier(nsInformer corev1informer.NamespaceInformer, subscribe
 func (c *NamespaceNotifier) needUpdate(old, new *corev1.Namespace) bool {
 	nsCreated := old == nil && new != nil
 	nsDeleted := old != nil && new == nil
+	nsDeleting := new != nil && !new.DeletionTimestamp.IsZero()
 	nsLabelsChanged := old != nil && new != nil &&
 		!reflect.DeepEqual(old.Labels, new.Labels)
 
-	return nsCreated || nsDeleted || nsLabelsChanged
+	return nsCreated || nsDeleted || nsDeleting || nsLabelsChanged
 }
 
 // reconcile notify subscribers with the request namespace key following namespace events.
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -504,14 +504,14 @@ func (c *Controller) ReconcileNetAttachDef(key string) error {
 // ReconcileNamespace enqueue relevant Cluster UDN CR requests following namespace events.
 func (c *Controller) ReconcileNamespace(key string) error {
 	namespace, err := c.namespaceInformer.Lister().Get(key)
-	if err != nil {
-		// Ignore removed namespaces
-		if apierrors.IsNotFound(err) {
-			return nil
-		}
+	if err != nil && !apierrors.IsNotFound(err) {
 		return fmt.Errorf("failed to get namespace %q from cache: %w", key, err)
 	}
-	namespaceLabels := labels.Set(namespace.Labels)
+
+	var namespaceLabels labels.Set
+	if namespace != nil {
+		namespaceLabels = namespace.Labels
+	}
 
 	c.namespaceTrackerLock.RLock()
 	defer c.namespaceTrackerLock.RUnlock()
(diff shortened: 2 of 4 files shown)


npinaeva (commented):
I think we need one more if in the `ReconcileNamespace` function to force a CUDN reconciliation on namespace delete/deletion timestamp being set, for now we seem to ignore that

kyrtapz:
> I think we need one more if in the `ReconcileNamespace` function to force a CUDN reconciliation on namespace delete/deletion timestamp being set, for now we seem to ignore that

Hmm, do we? I think ReconcileNamespace is called from https://github.com/ovn-kubernetes/ovn-kubernetes/blob/7f5e251d31a1c111bef30f7223efca325028ba6f/go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go#L62 which will be triggered with my changes on NS getting the deletion timestamp set.

trozet (approved):
/lgtm

trozet:
just rebased to pick up unit test fix

trozet:
while you are here it might be good to add a quick e2e test as well

kyrtapz:
> One minor thing. In ReconcileNamespace when there is a delete, we return out early if we cant find the namespace in the informer cache. I think instead the function should continue to check if the namespaceTracker has an entry, and if so it should call c.cudnController.Reconcile(cudnName)

When a namespace doesn't exist anymore there will be nothing to do for the CUDN controller since there will be no NADs to clean. It could potentially cleanup the NS cache though so I don't mind adding it.

kyrtapz:
@coderabbitai create an issue for the crash in https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21470554584/job/61842614481?pr=5888

go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go
```diff
@@ -46,10 +46,11 @@ func NewNamespaceNotifier(nsInformer corev1informer.NamespaceInformer, subscribe
 func (c *NamespaceNotifier) needUpdate(old, new *corev1.Namespace) bool {
 	nsCreated := old == nil && new != nil
 	nsDeleted := old != nil && new == nil
+	nsDeleting := new != nil && new.DeletionTimestamp != nil
```
npinaeva: that is what I meant :)

kyrtapz: not sure I follow 😅 this line was there in the initial version of the pr.

npinaeva:
> When a namespace doesn't exist anymore there will be nothing to do for the CUDN controller since there will be no NADs to clean. It could potentially cleanup the NS cache though so I don't mind adding it.

Are we talking about the `namespaceTracker`? I don't think it should be possible to have no NADs but a cache entry (that would mean a bug in the cache). I would vote to not cause an extra reconcile if we agree that is not possible

go-controller/pkg/clustermanager/userdefinednetwork/notifier/namespace.go
npinaeva: maybe some github issues, I only saw changes in the `controller.go` when I reviewed first

kyrtapz:
> > When a namespace doesn't exist anymore there will be nothing to do for the CUDN controller since there will be no NADs to clean. It could potentially cleanup the NS cache though so I don't mind adding it.
> 
> Are we talking about the `namespaceTracker`? I don't think it should be possible to have no NADs but a cache entry (that would mean a bug in the cache). I would vote to not cause an extra reconcile if we agree that is not possible

I initially thought we won't do CUDN reconcile when a deleted namespace is no longer in our tracker but this is not true as we fallback to this: https://github.com/kyrtapz/ovn-kubernetes/blob/9edf9627a91c7d167d5e2c0533fe2ffee1722fd8/go-controller/pkg/clustermanager/userdefinednetwork/controller.go#L283-L293
So I don't see a reason for reconciling a CUDN on a namespace deletion either, especially considering we are now handling terminating namespaces.

trozet:
> . It could potentially cleanup the NS cache though so I don't mind adding it.

yeah just to be safe

trozet:
> bout the `namespaceTracker`? I don't think it should be possible to have no NADs but a cache entry (that would mean a bug in the cache). I would vote to not cause an extra reconcile if we agree that is not possible

it's possible to miss an event or we don't handle an event correctly in our code. If you don't wnat to reconcile we shoudl at least have a check here and log an error

kyrtapz:
> > bout the `namespaceTracker`? I don't think it should be possible to have no NADs but a cache entry (that would mean a bug in the cache). I would vote to not cause an extra reconcile if we agree that is not possible
> 
> it's possible to miss an event or we don't handle an event correctly in our code. If you don't wnat to reconcile we shoudl at least have a check here and log an error

[Added it](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5888/changes#diff-891f2bd2e1eea7ee0df79d1ed3d55e72df60b9a2bbb60fe64374e8131fb7a269R523-R529) in a way we do not reconcile if the deleted namespace is not in cache.
For the record, this should never trigger in practice since terminating namespaces are now handled before deletion.
I consider it a cheap safety net for a black swan we hope never lands ;)

kyrtapz:
hitting:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5876
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/4221

kyrtapz:
@trozet FYI, will merge it once CI gets back.
This pr slipped through the cracks.

PR #5892, opened 2026-01-29
Remove bits about docker ipv6 support being experimental

danwinship:
Working on porting iptables stuff in the e2e tests and found this hack, which says

    // Remove this func when [docker ipv6 support] is no longer experimental.

It's supported by default as of docker 27.0.0 (https://github.com/moby/moby/pull/47747), and our CI images are using Ubuntu 28.04, which has docker 28.0.4 (https://github.com/actions/runner-images/blob/ubuntu24/20260119.4/images/ubuntu/Ubuntu2404-Readme.md#tools).

So assuming the comment is correct, this hack can just be removed...

cc @martinkennelly 

## Summary by CodeRabbit

* **Tests**
  * Removed KinD-specific IPv6 isolation workaround from end-to-end tests. This simplifies the test setup, reduces special-case failure paths for IPv6 test runs on KinD, and streamlines node listing/attachment steps during e2e execution.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
-	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/containerengine"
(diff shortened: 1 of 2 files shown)


danwinship:
test failures are spurious; the tests affected by the change (the ones that start with "`e2e egress IP validation`", in IPv6 and dual-stack jobs only) passed

danwinship:
@dave-tucker needed a rebase around #5965

Issue #5893, opened 2026-01-29
Add Dynamic UDN support for cluster manager allocation


trozet:
### What would you like to be added?

The recent Dynamic UDN feature implementation only included filtering out inactive UDNs from the node and zone controllers. I would like to add support for also filtering for cluster-manager allocation. This will just be an update to the existing enhancement and completed before the next release.

### Why is this needed?

There was a request to also dynamically allocate subnets to nodes without the UDNs. This is useful in an environment where a UDN will only be on a few nodes out of thousands, and a user wants to use smaller subnets that are sized for only a few nodes.

dave-tucker:
Had a conversation with @trozet today about upgrade considerations for D-UDN.
This needs to go in the feature docs. We said:

- Upgrades are not "tenant" aware, therefore if you are upgrading more than one node at once, you need to mindful not to take down all the nodes that are hosting your dynamic UDN at once.
- If you expect a D-UDN only to be rendered on 4 nodes, you should probably allocate enough IP addresses for 8 nodes (i.e expected node count * 2). This is to cover cases where you may loose multiple nodes at once + the scheduler moves your workload on to new nodes and you need a D-UDN subnet allocating before the grace period for the subnets has expired on the old nodes
- Rolling updates (1 node at a time) should not cause issues, providing the other nodes hosting your D-UDN have sufficient capacity for any workloads that are rescheduled when the candidate node is being drained.
- Rolling updates of > 1 node will require much deeper planning

PR #5894, opened 2026-01-29
Updates Dynamic UDN OKEP with dynamic allocation

trozet:
Closes: #5893

## Summary by CodeRabbit

* **Documentation**
  * Expanded network allocation docs to describe dynamic per-node UDN subnet allocation, active/inactive node behavior, multi-condition rendering, smaller per-UDN subnets for BGP, and status reporting enhancements.
* **New Features**
  * Added configurable deletion grace period and refined per-node allocation/selection semantics with multi-condition rendering.
* **Tests**
  * Added unit tests covering dynamic allocation behavior.
* **Chores**
  * Updated target platform version.

### DIFF
--- a/docs/okeps/okep-5552-dynamic-udn-node-allocation.md
+++ b/docs/okeps/okep-5552-dynamic-udn-node-allocation.md
@@ -63,9 +63,16 @@ advertising this subnet with BGP and want to use the smallest subnet possible.
 ## Proposed Solution
 
 The proposed solution is to add a configuration knob to OVN-Kubernetes, "--dynamic-udn-allocation", which will enable
-this feature. Once enabled, NADs derived from CUDNs and UDNs will only be rendered on nodes where there is a pod
-scheduled in that respective network. Additionally, if the node is scheduled as an Egress IP Node for a UDN, this node
-will also render the UDN.
+this feature. Once enabled, NADs derived from CUDNs and UDNs will only be rendered on nodes where one of the conditions
+below are true:
+
+1. There is a pod scheduled in the respective network.
+2. The node is scheduled as an Egress IP Node for a network.
+3. If Cluster Network Connect (CNC) CR exists which selects a network that meets 1 or 2, then all other networks and NADs
+selected by the CNC will also be included.
+
+If a node meets the conditions above, the node is considered *active*. A node where a network is not rendered and
+does not meet the above criteria is considered *inactive*.
 
 When the last pod on the network is deleted from a node, OVNK will not immediately tear down the UDN.
 Instead, OVNK will rely on a dead timer to expire to conclude that this UDN is no longer in use and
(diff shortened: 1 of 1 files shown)


docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -15,6 +15,7 @@ allows for horizontal scaling to much higher number of overall UDNs running in a
 ## Goals
 
  * To dynamically allow the network to only be rendered on specific nodes.
+ * To dynamically allocate subnets and other per-node+UDN information.
```
dave-tucker: Aren't we already dynamically allocating subnets. i.e if you add a new node to a cluster with a L3 UDN, that node will get allocated a hostSubnet sized range from the UDN CIDR.
What's changing, as I understand it, is that you're only allocating subnets for node where the UDN has been rendered. Would that be better described as:
`
To only allocate subnets and other per-node UDN information on nodes where the UDN is rendered.
`

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
```diff
@@ -41,6 +42,11 @@ launched in UDN 201, the node will have to render UDN 201 before the pod can be
 a one time larger pod latency cost for the first pod wired to the UDN. Additionally, there are more tradeoffs with other
 feature limitations outlined later in this document.
 
+Furthermore, there is a desire to dynamically allocate per-node network assignments, such as per node subnet allocation.
```
dave-tucker: `Furthermore, there is a desire to allocate per-node network assignments, such as per node subnet allocation, only on nodes where a UDN has been rendered.`

docs/okeps/okep-5552-dynamic-udn-node-allocation.md
npinaeva: Considering the whole feature is called dynamic UDN, "dynamic" was supposed to mean depending on the usage (of course anything watching k8s objects is kinda dynamic, but that is not what we meant with this feature name, which I guess is too late to debate :)). Maybe we can include both like
`To dynamically allocate subnets and other per-node UDN information on nodes where the UDN is rendered.`

trozet:
> this makes sense, somewhat related question: do we want to also update the OKEP with CNC + dynamic UDN fixes? 

>> Or maybe that will be another follow up and we should at least say that these features are not compatible?

@npinaeva as much as I would like to ignore CNC I think the only way I will remember to fix it and make it work with D-UDN is to address it now. I'll add a second commit which lists it as compatible and try to remember what changes we need to do :)

trozet:
@npinaeva I tracked down the key comments we had around CNC when the initial implementation for D-UDN. For your convenience linking them here:


https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5623/changes#r2716994159

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5623/changes#r2721773015

npinaeva:
makes sense! I have only one extra optimization note (not sure if want to do this at all and if we need to document it here)
We could skip rendering Layer3 connected networks if `ClusterIPServiceNetwork` is false, because for Layer3 network CNC router will send it to the right node (as opposed to transit router in layer2 case) and we don't need to reference service LBs for that network if `ClusterIPServiceNetwork` is false.

npinaeva (approved):

PR #5904, opened 2026-02-02
Remove limit on ovs-node

### LINKED ISSUE #5903: Remove cpu/mem limit from ovs-node
Following the pattern of our other pods in ovn-kubernetes. 

During perf workloads we were noticing oom kills due to the mem limits we have set on the ovs-node ds.

jtaleric:
## 📑 Description

Removing cpu/mem limit from ovs-node. This is causing issues with the performance workload, and there doesn't seem to be a good reason to limit the ovs-node pods. Most deployments have ovs running outside of a pod/container.

Fixes #5903 

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Chores**
  * Removed CPU and memory resource limits from the OVS node container in Kubernetes deployments.
  * Confirmed no other container configuration (environment variables, probes, mounts, or resource requests) was changed as part of this update.

### DIFF
--- a/dist/templates/ovs-node.yaml.j2
+++ b/dist/templates/ovs-node.yaml.j2
@@ -85,9 +85,6 @@ spec:
           requests:
             cpu: 100m
             memory: 300Mi
-          limits:
-            cpu: 500m
-            memory: 500Mi
         env:
         - name: OVN_DAEMONSET_VERSION
           value: "1.2.0"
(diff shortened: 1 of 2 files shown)


dist/templates/ovs-node.yaml.j2
```diff
@@ -85,9 +85,6 @@ spec:
           requests:
             cpu: 100m
             memory: 300Mi
-          limits:
-            cpu: 500m
```
kyrtapz, line 89: We should also remove it from the helm copy.: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/8571ba40ce9f32a205fa948ccda1707f7f4a47a5/helm/ovn-kubernetes/charts/ovs-node/templates/ovs-node.yaml#L87-L89

jtaleric: ack. I haven't looked at the helm stuff so good catch! 🙇🏽

trozet (approved):

Issue #5899, opened 2026-02-02
Flacking UT: race detected during execution of test --- FAIL: TestUserDefinedNetworkController


pliurh:
### Which jobs are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21591351296/job/62212036359?pr=5780

### Which tests are flaking?

go-controller/pkg/clustermanager/userdefinednetwork.TestUserDefinedNetworkController()

### Since when has it been flaking?

likely caused by 86c71320625aaff1e3a4d9addadead8a8eea52ac

### Reason for failure (if possible)

The data race occurs because the NetworkManager is never stopped between tests, leaving worker goroutines running that continue to read from global config while the next test's BeforeEach writes to it.

PR #5905, opened 2026-02-02
Fixes race with cluster manager UDN test

trozet:
network manager was being created in the test but not shutdown.

Fixes: #5899

## Summary by CodeRabbit

* **Tests**
  * Improved test infrastructure and reliability for network controller testing.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
@@ -43,6 +43,7 @@ var _ = Describe("User Defined Network Controller", func() {
 	var (
 		cs *util.OVNClusterManagerClientset
 		f  *factory.WatchFactory
+		nm networkmanager.Controller
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


trozet:
@mattedallo PTAL

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
mattedallo, line 96: Thank you Tim for fixing this!
The fix looks good.
As an alternative, we could also use [DeferCleanup(nm.Stop)](https://onsi.github.io/ginkgo/#cleaning-up-our-cleanup-code-defercleanup) here instead of the AfterEach pattern, it will avoid the need for suite-level variables and nil checks

mattedallo:
LGTM

Left a minor comment.
Thank you @trozet

go-controller/pkg/clustermanager/userdefinednetwork/controller_test.go
trozet: thanks @mattedallo . I didn't know about that. I don't see that used around the unit test code. I do see it used in our e2es. Maybe we should start using it/refactor our unit tests to use it. I'll try to remember to use that in the future.

PR #5910, opened 2026-02-03
e2e: always wait for NAD network-id before proceeding

booxter:
## 📑 Description

There's a race condition when a pod is created right after a NAD is
created. The cluster manager allocates the network-id annotation
asynchronously, so the NAD can be without network-id for a moment. If
the node's nad_controller processes the NAD during this window, it
returns early without setting up the network and doesn't reconcile
later. This breaks all tests that depend on the NAD, as pods timeout
waiting for OVN annotations that are never set.

Some tests were dealing with this problem by sleeping. Instead of that,
we are going to wait for the annotation before proceeding.

This patch introduces a helper function to create NADs in e2e tests that
will guarantee the annotation. All existing occurrences of NAD creation
in e2e suite are updated to use the helper.

I'm hitting failures in multihoming tests that suggest this race condition
happening. I haven't seen other tests I touched here to do the same, but I
think it's wise to cover all occurrences of a potential race in one place.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed race conditions in end-to-end tests where resources were created but not fully initialized before test execution.

* **Tests**
  * Enhanced test reliability by adding synchronization to ensure network resources reach a ready state before proceeding with test assertions across multiple test suites.

### DIFF
--- a/test/e2e/egress_firewall.go
+++ b/test/e2e/egress_firewall.go
@@ -161,11 +161,8 @@ func egressFirewallPolicyValidationTests(useUDN bool, udnTopology string) {
 
 				netConfig = newNetworkAttachmentConfig(nadCfg)
 				netConfig.namespace = f.Namespace.Name
-				_, err = nadClient.NetworkAttachmentDefinitions(f.Namespace.Name).Create(
-					context.Background(),
-					generateNAD(netConfig, f.ClientSet),
-					metav1.CreateOptions{},
-				)
+				nad := generateNAD(netConfig, f.ClientSet)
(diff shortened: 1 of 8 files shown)


booxter:
Discussed with Tim, we suspect "the controller never catching up" scenario may have been fixed already in recent commits. I'm posting a [separate PR](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6447) with just the removal of now unnecessary `Sleep`s and closing this PR here as not needed.

PR #5839, opened 2026-01-10
DPU IC changes

crnithya:
## 📑 Description
This PR enables DPU/DPU Hosts to operate in IC mode. This works in a two cluster model with DPU hosts being on a separate cluster from DPUs. 
DPUs run the full ovn IC stack while only ovn-node component is run on DPU hosts.  DPUs require credentials to access the DPU hosts cluster and watch on k8s resources and act upon their updates. 
This PR depends on https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5911. Since hostname on DPU's chassis will be different from DPU host's node name, and it's the DPU hosts which are in the K8s cluster that DPU's watch on, this mismatch would be a problem with requested-chassis option referring to chassis hostname. But since DPU already annotates DPU host with its chassis-id, https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5911 will fix this issue.

Fixes #

## ✅ Checks

- [x] My code requires changes to the documentation
- [x] if so, I have updated the documentation as required
- [ ] My code requires tests
- [x] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## Summary by CodeRabbit

* **New Features**
  * Accept Kubernetes CA cert as base64 via env/CLI (K8S_CACERT_DATA / --k8s-cacert-data) with proper precedence.
  * New Helm chart and values for single-node DPU deployments and a DPU DaemonSet; added DPU node/management port, VF resource and network‑segmentation options.
  * Added configMap volume for runtime ovnkube config and new environment flags for interconnect/no‑overlay.

* **Bug Fixes**
  * Improved multi-source auth handling and safer redaction of tokens and CA data in logs.

* **Documentation**
  * New DPU deployment guide for two‑cluster DPU-host and DPU setups.

PR #5911, opened 2026-02-04
Move from chassis hostname -> chassis ID

trozet:
We set options for "requested_chassis" on ports which take either the SBDB chassis.hostname or the SBDB chassis.name. The chassis.name is derived from the node's OVS system-id. It annotated to the node as a chassis-id annotation. chassis.hostname is set locally by OVN-Controller when it creates the local SBDB chassis entry when it starts. The hostname for the local node is derived from an external_id in OVS. In OVN-Kubernetes we add remote chassis (looking for their chassis-id) annotation and then set their hostname as their KAPI node name. The hostname is not used by anything in OVS or OVN, and is a string value, and not the primary index for the chassis.

The end result is inconsistency, where some objects like NBDB GatewayChassis rely on chassis-id, while requested_chassis is using hostname. Using hostname creates further issues where OVS may be running in DPU and not on the KAPI node itself.

This commit moves everything to use chassis-id. Changes include:

 - Simplify syncChassis to track everything by chassis-id, remove checking for hostname.
 - Make requested_chassis usage use chassis-id.
 - Add validation function in libovsdb to make sure no one in the future tries to use hostname.
 - Fix unit tests and create consistent fake chassis id from helper function.
 - When creating a chassis do not set hostname for local nodes. OVN-Controller manages this. Set hostname to be KAPI node name for remote nodes.

Some other notes:
1. During chassis update, we do not modify hostname as part of libovsdb.
2. When we upgrade, there shouldn't be traffic outage as ports are updated for their requested_chassis to be chassis-id, since the chassis-id already exists in the SBDB entry.

## Summary by CodeRabbit

* **New Features**
  * Node-aware port binding now uses per-node chassis identifiers across controllers and tests.

* **Improvements**
  * Validation added to detect invalid chassis identifiers early.
  * Enhanced error handling and suppressed-error behavior when chassis annotations are missing or unparseable.
  * Chassis sync and bookkeeping now operate on node/chassis IDs for greater reliability.

* **Bug Fixes**
  * Logical port operations fail fast on invalid chassis options to prevent incorrect updates.


go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -557,8 +557,15 @@ func (c *Controller) ensureConnectPortsOps(ops []ovsdb.Operation, cnc *networkco
...
+				chassisID, err := util.ParseNodeChassisIDAnnotation(node)
+				if err != nil {
+					if util.IsAnnotationNotSetError(err) {
+						return nil, ovntypes.NewSuppressedError(err)
+					}
+					return nil, fmt.Errorf("failed to parse node chassis-id for node %s: %w", node.Name, err)
+				}
 				ops, err = c.createRouterPortOps(ops, connectRouterName, connectPortName, portPairInfo.connectPortIPs,
-					"", cncName, networkID, nodeID, tunnelKey, node.Name)
+					"", cncName, networkID, nodeID, tunnelKey, chassisID)
```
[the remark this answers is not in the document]
trozet: other errors in the loop trigger immediate error return. The caller also immediately continues on any error and does not even use the ops created by this function.

go-controller/pkg/ovn/master.go
```diff
@@ -443,24 +441,34 @@ func (oc *DefaultNetworkController) syncChassis(localZoneNodeNames, remoteZoneNo
...
+		chassisID, err := util.ParseNodeChassisIDAnnotation(node)
+		if err != nil {
+			if util.IsAnnotationNotSetError(err) {
+				klog.V(5).Infof("Skipping chassis sync for node %s due to missing chassis-id annotation", node.Name)
+				continue
+			}
+			return fmt.Errorf("failed to parse node chassis-id for node %s: %w", node.Name, err)
 		}
+		delete(chassisNameMap, chassisID)
 	}
```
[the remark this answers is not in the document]
trozet: if the node is missing a chassis-id then most certainly the sbdb chassis we have is stale. We should be removing it in that case.

go-controller/pkg/ovn/master.go
```diff
@@ -471,11 +479,11 @@ func (oc *DefaultNetworkController) syncChassis(localZoneNodeNames, remoteZoneNo
...
 
 	if err := libovsdbops.DeleteChassis(oc.sbClient, staleChassis...); err != nil {
```
crnithya, line 478: Can we have a debug log like
if len(staleChassis) > 0 {
		klog.V(5).Infof("syncChassis(): Deleting Stale Chassis: %v", chassisNameMap)
	}
We have hit issues in the past where the chassis were getting deleted and readded and this will help with debugging such cases.

go-controller/pkg/ovn/master.go
```diff
@@ -471,11 +479,11 @@ func (oc *DefaultNetworkController) syncChassis(localZoneNodeNames, remoteZoneNo
...
-		return fmt.Errorf("failed Deleting chassis %v error: %v", chassisHostNameMap, err)
+		return fmt.Errorf("failed Deleting chassis %v error: %v", chassisNameMap, err)
 	}
 
 	if err := libovsdbops.DeleteChassisTemplateVar(oc.nbClient, staleChassisTemplateVars...); err != nil {
-		return fmt.Errorf("failed Deleting chassis template vars %v error: %v", chassisHostNameMap, err)
+		return fmt.Errorf("failed Deleting chassis template vars %v error: %v", chassisNameMap, err)
```
crnithya: It should be templateChassisMap not chassisNameMap in the error message

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -559,6 +550,26 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(pod *corev1.Pod, nadKe
 		return nil, nil, nil, false, nil
 	}
 
+	// Bind the port to the node's chassis; prevents ping-ponging between
+	// chassis if ovnkube-node isn't running correctly and hasn't cleared
+	// out iface-id for an old instance of this pod, and the pod got
+	// rescheduled.
+	var node *corev1.Node
+	if !config.Kubernetes.DisableRequestedChassis {
```
aserdean, line 560: I think with this PR we might not need this flag anymore and it could be safely removed; I’ll run some tests on my side to confirm.
​

I’m also not sure whether deprecating this flag is in scope for this PR or should be handled in a separate one.

trozet: i think a separate PR makes more sense here

crnithya: Yes, as discussed in the other PR, we should remove this flag from IC mode and add a validation in config.go to prevent disabling it for IC mode as a separate PR

crnithya:
LGTM. Thanks for addressing this.

go-controller/pkg/libovsdb/ops/switch.go
```diff
@@ -323,6 +323,9 @@ func createOrUpdateLogicalSwitchPortsOps(nbClient libovsdbclient.Client, ops []o
 	opModels := make([]operationModel, 0, len(lsps)+1)
 
 	for _, lsp := range lsps {
+		if err := validateRequestedChassisOption(lsp.Options); err != nil {
+			return nil, err
```
tssurya, line 327: nit: format wrap error?

go-controller/pkg/libovsdb/ops/switch.go
```diff
@@ -486,6 +489,9 @@ func DeleteLogicalSwitchPortsWithPredicateOps(nbClient libovsdbclient.Client, op
 // existing
 func UpdateLogicalSwitchPortSetOptions(nbClient libovsdbclient.Client, lsp *nbdb.LogicalSwitchPort) error {
 	options := lsp.Options
+	if err := validateRequestedChassisOption(options); err != nil {
+		return err
```
tssurya: nit here as well to format wrap

go-controller/pkg/libovsdb/ops/validation.go
tssurya, line 1: seems specific to chassis options validation? should we just put this in ops/chassis.go itself? OR maybe this is better suited in libovsdb/util package as this is not doing any transact ops stuff, ? maybe libovsdb/util/chassis.go ? or libovsdb/util/validation.go?

go-controller/pkg/libovsdb/ops/switch.go
```diff
@@ -486,6 +489,9 @@ func DeleteLogicalSwitchPortsWithPredicateOps(nbClient libovsdbclient.Client, op
 // existing
 func UpdateLogicalSwitchPortSetOptions(nbClient libovsdbclient.Client, lsp *nbdb.LogicalSwitchPort) error {
```
tssurya: am I missing something or is this function not used anywhere?

go-controller/pkg/ovn/zone_interconnect/chassis_handler.go
```diff
@@ -160,12 +160,14 @@ func (zch *ZoneChassisHandler) createOrUpdateNodeChassis(node *corev1.Node, isRe
...
 	chassis := sbdb.Chassis{
-		Name:     chassisID,
-		Hostname: node.Name,
+		Name: chassisID,
 		OtherConfig: map[string]string{
 			"is-remote": strconv.FormatBool(isRemote),
 		},
 	}
+	if isRemote {
+		chassis.Hostname = node.Name
```
tssurya, line 173: ah so for local nodes ovn controller sets this anyways and so ovnkube technically speaking only needs to set it for the IC remote zone node logic but we were setting it always huh

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
```diff
@@ -460,9 +460,18 @@ func (zic *ZoneInterconnectHandler) createRemoteZoneNodeResources(node *corev1.N
...
+			// remote node may not have the annotation yet, suppress it
+			return types.NewSuppressedError(err)
+		}
+		return fmt.Errorf("failed to parse node chassis-id for node %s: %w", node.Name, err)
+	}
+
 	lspOptions := map[string]string{
 		libovsdbops.RequestedTnlKey:  strconv.Itoa(nodeID),
-		libovsdbops.RequestedChassis: node.Name,
+		libovsdbops.RequestedChassis: chassisID,
```
tssurya, line 474: hmm so during upgrades are we sure here and everywhere this is going to be a seamless update? i.e at startup it just updates ? cause some of these libovsdb update functions are partial NonDefault Value based updates .. not sure what's the dbindex for ports but I hope its an update and doesn't consider this a new port upon upgrade

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler_test.go
```diff
@@ -35,8 +35,8 @@ const (
 	// ovnNodeZoneNameAnnotation is the node annotation name to store the node zone name.
 	ovnNodeZoneNameAnnotation = "k8s.ovn.org/zone-name"
 
-	// ovnNodeChassisIDAnnotatin is the node annotation name to store the node chassis id.
-	ovnNodeChassisIDAnnotatin = "k8s.ovn.org/node-chassis-id"
```
tssurya, line 39: lool - its consistently wrong in spelling :)

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -559,6 +550,26 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(pod *corev1.Pod, nadKe
...
+			return nil, nil, nil, false, err
+		}
+		chassisID, err := util.ParseNodeChassisIDAnnotation(node)
+		if err != nil {
+			if util.IsAnnotationNotSetError(err) {
+				return nil, nil, nil, false, ovntypes.NewSuppressedError(err)
+			}
+			return nil, nil, nil, false, err
+		}
+		lsp.Options[libovsdbops.RequestedChassis] = chassisID
```
tssurya, line 572: so when we upgrade to this version are all pods going to undergo this update - that's gonna be a lot of at once pod add update transactions isn't it?
edit: I realised this function is only called of addPort is true - so is this change only applicable to newly created ports? cause that makes things even more inconsistent where half pods are on nodeName and half are on chassisID?

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -559,6 +550,26 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(pod *corev1.Pod, nadKe
...
+	// chassis if ovnkube-node isn't running correctly and hasn't cleared
+	// out iface-id for an old instance of this pod, and the pod got
+	// rescheduled.
+	var node *corev1.Node
+	if !config.Kubernetes.DisableRequestedChassis {
+		node, err = bnc.watchFactory.GetNode(pod.Spec.NodeName)
+		if err != nil {
+			return nil, nil, nil, false, err
+		}
+		chassisID, err := util.ParseNodeChassisIDAnnotation(node)
```
tssurya, line 565: is this going to bump up the pod add latency ? cause of the extra annotation parsing? and I think we are doing a new dependency here where pod cannot get its port added till chassis-annotation is not set on node - v/s previously that wasn't the case?

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -919,9 +931,11 @@ func (bnc *BaseNetworkController) allocatePodAnnotation(pod *corev1.Pod, existin
 			podDesc, switchName)
 	}
 
-	node, err := bnc.watchFactory.GetNode(pod.Spec.NodeName)
-	if err != nil {
-		return nil, false, err
+	if node == nil {
```
tssurya: ah its just to exclude the refetch from informer if it was already done?

go-controller/pkg/ovn/egressip_udn_l3_test.go
```diff
@@ -41,26 +41,25 @@ var _ = ginkgo.Describe("EgressIP Operations for user defined network with topol
...
-		v4Node1Tsp           = "100.88.0.2"
-		node2Name            = "node2"
-		v4Node2Net1          = "20.129.0.0/16"
-		v4Node2Tsp           = "100.88.0.3"
-		podName4             = "egress-pod4"
-		v4Pod1IPNode2Net1    = "20.129.0.2"
-		v4Pod2IPNode2Net1    = "20.129.0.3"
-		eIP1Mark             = 50000
-		eIP2Mark             = 50001
-		userDefinedNetworkID = "2"
```
tssurya, line 62: was this unused? makes the comment in the next line moot I guess which refers to userDefinedNetworkID

go-controller/pkg/ovn/layer2_user_defined_network_controller.go
```diff
@@ -889,13 +899,23 @@ func (oc *Layer2UserDefinedNetworkController) addRouterSetupForRemoteNodeGR(node
 	if err != nil {
 		return nil
 	}
+
+	chassisID, err := util.ParseNodeChassisIDAnnotation(node)
```
tssurya, line 903: I wonder if we should parse it once and keep it in controller somewhere to avoid annotation parsing for both routersetup and switchsetup

go-controller/pkg/ovn/master.go
```diff
@@ -443,26 +441,37 @@ func (oc *DefaultNetworkController) syncChassis(localZoneNodeNames, remoteZoneNo
 
 	// Delete existing nodes from the chassis map.
 	// Also delete existing templateVars from the template map.
-	for _, nodeName := range localZoneNodeNames {
-		if chassis, ok := chassisHostNameMap[nodeName]; ok {
-			delete(chassisNameMap, chassis.Name)
-			delete(chassisHostNameMap, chassis.Hostname)
-			delete(templateChassisMap, chassis.Name)
+	for _, node := range localZoneNodes {
+		chassisID, err := util.ParseNodeChassisIDAnnotation(node)
```
tssurya, line 445: is there a chance for race condition here where new node has joined cluster, ovn-controller creates the chassis entry, ovnkube-node hasn't annotated yet, syncChassis runs and now we delete the chassis since its considered stale?
should we add a fall back for upgraded nodes to still use nodeName if chassisid not found or are we safe here?

go-controller/pkg/libovsdb/ops/validation.go
trozet: I can put it in ops/chassis.go if you prefer. I don't think we should add a sub util folder

go-controller/pkg/libovsdb/ops/switch.go
trozet: it's not. I debated removing it, but I thought if someone adds it back they might miss adding the validation call :)

I'll just remove it.

go-controller/pkg/libovsdb/ops/switch.go
trozet: you mean you want to wrap "requested-chassis must be a valid UUID, got %q"" in something like Failed to update logical switch port: requested-chassis must be a valid UUID, got %q?

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: for the first part about latency, it's valid. I'll add a cache and fallback. The second dependency part I don't think exists. If the chassis ID doesn't exist then then the topology doesnt exist. If no chassis-id, there is no SBDB entry for that remote node (chassis setup failed for IC) which means the port cant bind anyway and there is no connectivity. 

Nodes are always processed first in ovnk, cause everything depends on that.

go-controller/pkg/ovn/master.go
crnithya: Fallback of using nodename will not work for DPU in IC mode, https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5839, since the nodename will be that of DPU host, and the hostname on the chassis that of DPU. 
How about aborting that run of syncChassis if chassis-id annotation is not found on localzonenodes?  It will run periodically every 5 minutes anyway.

go-controller/pkg/ovn/master.go
```diff
@@ -443,26 +441,37 @@ func (oc *DefaultNetworkController) syncChassis(localZoneNodeNames, remoteZoneNo
...
+		if err != nil {
+			if util.IsAnnotationNotSetError(err) {
+				klog.V(5).Infof("Skipping chassis sync for node %s due to missing chassis-id annotation", node.Name)
```
crnithya: This error message is misleading, since the chassis will be actually deleted and not skipped.

go-controller/pkg/ovn/master.go
```diff
@@ -443,26 +441,37 @@ func (oc *DefaultNetworkController) syncChassis(localZoneNodeNames, remoteZoneNo
...
+				continue
+			}
+			return fmt.Errorf("failed to parse node chassis-id for node %s: %w", node.Name, err)
```
crnithya: Chassisid annotation error for one node, will abort the entire syncchassis operation. Shouldn't we just consider the chassis belonging to that node if it exists to be stale and delete it?

go-controller/pkg/ovn/zone_interconnect/chassis_handler.go
trozet: right, but it also wasnt part of update ops, so it was never getting updated either. Also if you change the value to something else for the local node, ovn-controller will change it back right after.

go-controller/pkg/libovsdb/ops/switch.go
trozet: I'm not sure it is necessary cause we the error will propagate up and in the caller we catch errors with a better message like: klog.Errorf("Failed to add logical port %+v to switch %s: %v", logicalSwitchPort, switchName, err)

go-controller/pkg/ovn/zone_interconnect/zone_ic_handler.go
trozet: Yeah at start up zone ic will add all nodes and update each port. It doesn't consider it a new port. LSP options are not an index.

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: so fun fact about libovsdbops... it looks like we always update. The createOrUpdateOps logic just checks if it finds an entry, then updates it no matter what and generates ops from what I can see. A test here shows it:
https://gist.github.com/trozet/1c4bcd1250f34c253f119c1d34da225f


The only thing that will prevent this is checking the cache first, before deciding to create/update. The spot where you are commenting attempts to be smart by building a customFields list to know only which fields to update...however if there are no customFields, it doesn't skip update like you would think. Instead it hits this code:

```
	if len(customFields) != 0 {
		fieldInterfaces = getFieldsToUpdate(lsp, customFields)
	} else {
		fieldInterfaces = getAllUpdatableFields(lsp)
	}
```

Where then it updates all fields. This is something that should be fixed in general and causes unnecessary txns, but should be in a different PR. An even bigger fix we could look at is comparing fields during libovsdb ops generation and not generating update ops if the fields are equal so that it is then built in.

@jcaamano @kyrtapz fyi

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: yeah I should just remove this because I didn't do this in the UDN path. I think it isn't worth it.

go-controller/pkg/ovn/egressip_udn_l3_test.go
trozet: Looks like @npinaeva added this here https://github.com/ovn-kubernetes/ovn-kubernetes/commit/77124b84a8b0507b1a11511a048ba143995bc8d8#diff-801673a4f293c2d4d780faedb626d2f79463dcfeb2d35368415680cae1c2a77cR61

```
		//tnlKey = zoneinterconnect.BaseTransitSwitchTunnelKey + secondaryNetworkID
		tnlKey = "16711685"
```

The BaseTransitSwitchTunnelKey is BaseTransitSwitchTunnelKey = 16711683

So she added the 2 implicitly. I think the comment and code is fine.

go-controller/pkg/ovn/master.go
trozet: If a new node joins the cluster, it has no OVN topology yet, especially if there is no chassis-id. If syncChassis deletes the local chassis, OVN will immediately recreate it

go-controller/pkg/ovn/master.go
trozet: I can rephrase to say chassis may be deleted for node ....

go-controller/pkg/ovn/master.go
trozet: this code practically is unreachable because ParseNodeChassisIDAnnotation only returns newAnnotationNotSetError for an error. I'll just remove this and on any parsing error continue

go-controller/pkg/ovn/base_network_controller_pods.go
jcaamano: > The createOrUpdateOps logic just checks if it finds an entry, then updates it no matter what and generates ops from what I can see

createOrUpdateOps checks if it finds an entry, if it does, if updates it, if it doesn't, it creates it:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/libovsdb/ops/model_client.go#L224-L251

`doWhenFound` -> `m.update`/`m.mutate`
`doWhenNotFound` -> `m.create`

When creating it creates the object as specified. When updating, it updates the fields specified in `OnModelUpdates`:

```
	// OnModelUpdates specifies the fields from Model that will be used as
	// the update value.
	OnModelUpdates []interface{}
```

If `OnModelUpdates` is not set, nothing will be updated. Usually we default to set `OnModelUpdates` to update all non-zero fields that have been provided. But we do have custom behavior depending on the situation including providing from outside of libovsdbops the fields to be updated like in the LSP example highlighted above.

I am not sure I understand what the intended fix is. But if you don't want to update, why would you call a `createOrUpdate` method instead of just a `create` method (that would leave `OnModelUpdates` unset)? There is no such method because we never needed it but we could just add it if we do need it.

trozet:
> Nice PR describe! Making it consistent makes sense..
> 
> ```
> The hostname is not used by anything in OVS or OVN, and is a string value, and not the primary index for the chassis.
> ```
> 
> Any idea why ovnkube then sets the chassis.HostName for remote nodes? is it required if we are setting the chassisID correctly?

@tssurya 
It's needed for when you set requested_chassis on layer 2 ports or remote ovn_cluster_router ports on transit switch and you use the hostname instead of the chassis-id. Now that things are switched to chassis-id there is no *technical* reason to have it. However, it does help with debugging when you are looking at an OVN DB and you list the chassis to see which KAPI node it belongs to. For that reason, we keep setting hostname for remote nodes.

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: talked with @jcaamano we both agreed we should fix the customFields usage and probably also implement a field checker deeper in model ops. I'll work on those fixes, but they are outside of this PR. In summary, these changes do not introduce anymore update operations than what already exist today.

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: actually there is no reason to add a cache, because there is no json unmarshaling for this annotation we just read it directly as a string

go-controller/pkg/ovn/layer2_user_defined_network_controller.go
trozet: no reason to do this, since we dont do parsing for this really, we just read the annotation valuee as a string

go-controller/pkg/ovn/master.go
crnithya: LGTM

go-controller/pkg/ovn/base_network_controller_pods.go
tssurya: hmmm I'm not sure I got all that, I'll ping you later today for this specific PR context
cause my main question is on upgrades are we going to end up updating the chassisID for all LSPs? - so normally we do send updates but they were being a no-op on the DB size is it? and now it would just update the LSPs? the confusion I had was cause I saw we check for `addPort` and only then do the ops

go-controller/pkg/ovn/base_network_controller_pods.go
tssurya: > for the first part about latency, it's valid. I'll add a cache and fallback. The second dependency part I don't think exists. If the chassis ID doesn't exist then then the topology doesnt exist. If no chassis-id, there is no SBDB entry for that remote node (chassis setup failed for IC) which means the port cant bind anyway and there is no connectivity.
> 
> Nodes are always processed first in ovnk, cause everything depends on that.

gotcha makes sense

go-controller/pkg/ovn/base_network_controller_pods.go
tssurya: > talked with @jcaamano we both agreed we should fix the customFields usage and probably also implement a field checker deeper in model ops. I'll work on those fixes, but they are outside of this PR. In summary, these changes do not introduce anymore update operations than what already exist today.

yea  that would be nice or if we had a way to get smarter in libovsdb layer to know when to update v/s not that would be nice too. I had a similar comment from https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828#discussion_r2730989573 CNC PR that I wanted to check out..

go-controller/pkg/libovsdb/ops/chassis.go
```diff
@@ -171,3 +174,17 @@ func CreateOrUpdateChassis(sbClient libovsdbclient.Client, chassis *sbdb.Chassis
 
 	return nil
 }
+
+func validateRequestedChassisOption(options map[string]string) error {
```
npinaeva, line 180: this is only meant to catch the wrong usages of hostname instead of chassis-id? A comment would be nice

go-controller/pkg/ovn/controller/networkconnect/topology.go
npinaeva, line 568: should we also rename the `createRouterPortOps` argument name from remoteChassisName?

go-controller/pkg/ovn/zone_interconnect/chassis_handler.go
npinaeva, line 173: So why do we even set it then? The message says 'This commit moves everything to use chassis-id" so I expected to never see `Hostname` used in the code..

go-controller/pkg/ovn/base_network_controller_pods.go
```diff
@@ -559,6 +550,26 @@ func (bnc *BaseNetworkController) addLogicalPortToNetwork(pod *corev1.Pod, nadKe
 		return nil, nil, nil, false, nil
 	}
 
+	// Bind the port to the node's chassis; prevents ping-ponging between
+	// chassis if ovnkube-node isn't running correctly and hasn't cleared
+	// out iface-id for an old instance of this pod, and the pod got
+	// rescheduled.
```
npinaeva: have we ever figured out if this is still a problem with IC?

trozet: ping ponging not a problem for IC, but still need to set requested chassi for remote ports, like layer 2 pods

go-controller/pkg/ovn/zone_interconnect/chassis_handler.go
trozet: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5911#issuecomment-3861356450

go-controller/pkg/ovn/controller/networkconnect/topology.go
trozet: well it is still technically correct. The sbdb chassis name field == chassis ID (derived from ovs system id)

go-controller/pkg/libovsdb/ops/chassis.go
trozet: yeah, if i have to push again i will add a comment

go-controller/pkg/ovn/base_network_controller_pods.go
trozet: "my main question is on upgrades are we going to end up updating the chassisID for all LSPs?"

yeah, on start up, it will add all ports (addPort=true), then it will try to get the current LSP, and attempt to only do a txn if the fields have changed. However, when it detects no fields have changed the logic is broken and assumes everything needs to be updated. So that's why I said update will happen either way and I'll fix it  in another PR.

tssurya: ah gotcha, ok makes sense! thanks @trozet !

go-controller/pkg/ovn/zone_interconnect/chassis_handler.go
npinaeva: that kinda tells me this comment should be in the code, but also we still don't have hostname in the update fields, so it is only set on creation, which is also confusing. Another comment saying that is intended could be useful

go-controller/pkg/ovn/base_network_controller_pods.go
npinaeva: should we update this comment then?

npinaeva:
my comment are mostly about adding comments. While not critical, I find it useful to avoid confusion in the future.
Maybe we could have it as a follow up if this part needs to be merged asap

trozet:
@npinaeva pushed again and added comments where requested

PR #5912, opened 2026-02-04
ANP: Skip redundant status updates to reduce API server load

tssurya:
## 📑 Description

When reconciling AdminNetworkPolicy or BaselineAdminNetworkPolicy, the controller updates the status condition on every successful sync, even when the status is already in the desired state. At scale (500-1000 nodes) OR when there is a lot of pod churn, this causes unnecessary API server load spikes since each zone's controller repeatedly patches the same CRD.

This PR adds a doesStatusNeedAnUpdate() utility function that compares the existing condition with the new condition. If Status, Reason, and Message are all unchanged, skip the ApplyStatus call entirely. This optimization applies to both ANP and BANP status update paths.

Credit to @liqcui for working with the end-user in replicating what they were observing in their cluster.
End-user has a cron job where they create and destroy pods across 20 namespaces every 5 mins. End user has around 83 ANPs, each one matches at least 1 namespace but also each namespace could have upto 20ANPs matching on them.. This causes a clear spike in kube-apiserver load exactly in 5mins intervals. Two end users have reported this.
@liqcui will test this fix and drop a comment to see if this fixes the issue

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **Performance**
  * Skips redundant status writes for admin network policies, reducing API server load and improving responsiveness.
* **Bug Fixes**
  * Prevents unnecessary status-change attempts while preserving transition timing when real updates occur.
* **Logging**
  * Reduces noisy logs for no-op status checks and adds concise success logs when status patches are applied.
* **Tests**
  * Adds unit tests to validate idempotent status behavior and API-call suppression when no changes occur.

### DIFF
--- a/go-controller/pkg/ovn/controller/admin_network_policy/status.go
+++ b/go-controller/pkg/ovn/controller/admin_network_policy/status.go
@@ -46,6 +46,22 @@ const (
 	policyNotReadyReason = "SetupFailed"
 )
 
+// doesStatusNeedAnUpdate compares the existing condition with the new condition
+// and returns true if an update is needed, false if the status is already in the desired state.
+// This helps avoid unnecessary API server calls when the status hasn't changed.
+func doesStatusNeedAnUpdate(existingCondition *metav1.Condition, newCondition metav1.Condition) bool {
+	if existingCondition == nil {
+		return true // condition doesn't exist yet, needs to be created
+	}
+	// Check if Status, Reason, and Message are all the same - if so, no update needed
+	if existingCondition.Status == newCondition.Status &&
+		existingCondition.Reason == newCondition.Reason &&
+		existingCondition.Message == newCondition.Message {
+		return false
+	}
+	return true
+}
+
 // updateANPStatusToReady updates the status of the policy to reflect that it is ready
 // Each zone's ovnkube-controller will call this, hence let's update status using server-side-apply
 func (c *Controller) updateANPStatusToReady(anpName string) error {
@@ -59,8 +75,6 @@ func (c *Controller) updateANPStatusToReady(anpName string) error {
 	if err != nil {
 		return fmt.Errorf("unable to update the status of ANP %s, err: %v", anpName, err)
 	}
-	klog.V(5).Infof("Patched the status of ANP %v with condition type %v/%v",
-		anpName, policyReadyStatusType+c.zone, metav1.ConditionTrue)
 	return nil
 }
 
@@ -83,8 +97,6 @@ func (c *Controller) updateANPStatusToNotReady(anpName, message string) error {
 	if err != nil {
 		return fmt.Errorf("unable update the status of ANP %s, err: %v", anpName, err)
 	}
-	klog.V(3).Infof("Patched the status of ANP %v with condition type %v/%v and reason %s/%s",
-		anpName, policyReadyStatusType+c.zone, metav1.ConditionFalse, policyNotReadyReason, message)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


liqcui:
The issue was fixed after applying the fix, the master CPU usage and kube-apiserver pod usage decreased as attached snapshot
**master node cpu usage**
master node cpu usage
**kube-apiserver cpu usage**
kube-apiserver cpu usage

go-controller/pkg/ovn/controller/admin_network_policy/status.go
```diff
@@ -94,6 +110,10 @@ func (c *Controller) updateANPZoneStatusCondition(newCondition metav1.Condition,
 		return err
 	}
 	existingCondition := meta.FindStatusCondition(anp.Status.Conditions, newCondition.Type)
+	if !doesStatusNeedAnUpdate(existingCondition, newCondition) {
```
arkadeepsen, line 109: I think we can utilize existing `SetStatusCondition` from apimachinery's meta package: https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#SetStatusCondition. The existing `updateANPZoneStatusCondition` and `updateBANPZoneStatusCondition` functions can be updated accordingly.

go-controller/pkg/ovn/controller/admin_network_policy/status.go
arkadeepsen: `updateANPZoneStatusCondition` should probably take care of printing the log messages rather than here. Another way would be to return a boolean and print the appropriate message according to the boolean value. Otherwise, this message will be misleading.

arkadeepsen: Same issue with this log message as well.

go-controller/pkg/ovn/controller/admin_network_policy/status.go
tssurya: see https://github.com/ovn-kubernetes/ovn-kubernetes/commit/7a69306756e4ecc3de8f24a857a32a25ee6f3cb3 :) on why we removed it
IIRC it was ok to use it with UpdateStatus+retryconflict but with SSA I need the apply configuration types for the condition owned by the fieldmanager not the entire list of conditions we mutate.. I think that's why almost all features doing SSA are manually having to deal with the LastTransitionTime - SetStatusCondition part

go-controller/pkg/ovn/controller/admin_network_policy/status.go
tssurya: great point will fix this

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/b2e28afe3d7354011db5ec8ffbc9835534135709..0c9e511147d517bf7f72f2463f3a7e92f872dcb8 addresses review comment

I'll now rebase on master while at it.

go-controller/pkg/ovn/controller/admin_network_policy/status.go
arkadeepsen: Makes sense. Do you think we need to add unit test for the newly added `doesStatusNeedAnUpdate` function?

arkadeepsen: Do you think this can be moved to a common utils file instead? The function can be utilized for other objects as well then.

tssurya: > Do you think this can be moved to a common utils file instead? The function can be utilized for other objects as well then.

I think that can be done in a FUP? i.e I am not quiet sure if all features are using the message and reason fields as well like we do for ANP, but its a good suggestion overall if other features are also having the same scale bottleneck..

tssurya: > Makes sense. Do you think we need to add unit test for the newly added `doesStatusNeedAnUpdate` function?

yup we should add some unit tests. PTAL I have added them in latest push

tssurya:
https://github.com/ovn-kubernetes/ovn-kubernetes/compare/f40e92c7091db3b399199639d7f9e624d661148a..9a04af38a587e9ebdc9334902e69f542ada0e6db adds the tests,

I'll now rebase on master

arkadeepsen (approved):
LGTM

PR #5936, opened 2026-02-05
No overlay: Add outboundSNAT support to CDN for both LGW and SGW mode

pliurh:
## 📑 Description

Implements outbound SNAT support for no-overlay mode with selective exemption of pod-to-pod traffic.

Changes
SNAT exemption framework: Added support for exempted_ext_ips field in OVN SNAT rules to exclude specific destinations
Shared gateway mode: Created OVN address set containing cluster CIDRs and node IPs to exempt internal traffic from SNAT
Local gateway mode: Added nftables return rules to bypass SNAT for pod-to-pod and pod-to-local-node traffic
Code organization: Centralized pod SNAT operations into reusable AddPodSNATOps method, moved constants to types package
Testing: Added KIND/Helm deployment support for SNAT mode, enabled in CI, fixed route advertisement e2e tests
When enabled, external traffic uses node IP (SNATed) while pod-to-pod traffic retains original pod IPs.

Fixes #

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Outbound SNAT for no-overlay mode with automatic exemptions for cluster-internal and local traffic (IPv4/IPv6).
  * CLI/installer accepts optional "snat-enabled" for no-overlay to enable outbound SNAT.

* **Configuration**
  * New Helm value global.enableNoOverlaySnat (default: false).
  * Runtime validation prevents enabling outbound SNAT unless no-overlay is active.

* **Tests**
  * Expanded unit and e2e tests covering SNAT, exemptions, and related flows.

### DIFF
--- a/helm/ovn-kubernetes/charts/ovnkube-node-dpu-host/templates/ovnkube-node-dpu-host.yaml
+++ b/helm/ovn-kubernetes/charts/ovnkube-node-dpu-host/templates/ovnkube-node-dpu-host.yaml
@@ -225,6 +225,10 @@ spec:
           value: {{ hasKey .Values.global "enableDynamicUDNAllocation" | ternary .Values.global.enableDynamicUDNAllocation false | quote }}
         - name: OVN_DYNAMIC_UDN_GRACE_PERIOD
           value: {{ default "" .Values.global.dynamicUDNGracePeriod | quote }}
+        - name: OVN_ROUTE_ADVERTISEMENTS_ENABLE
+          value: {{ hasKey .Values.global "enableRouteAdvertisements" | ternary .Values.global.enableRouteAdvertisements false | quote }}
+        - name: OVN_ADVERTISED_UDN_ISOLATION_MODE
+          value: {{ default "strict" .Values.global.advertisedUDNIsolationMode | quote }}
         - name: OVN_NO_OVERLAY_ENABLE
           value: {{ default "false" .Values.global.enableNoOverlay | quote }}
         - name: OVNKUBE_NODE_MGMT_PORT_NETDEV
--- a/helm/ovn-kubernetes/charts/ovnkube-node-dpu/templates/ovnkube-node-dpu.yaml
+++ b/helm/ovn-kubernetes/charts/ovnkube-node-dpu/templates/ovnkube-node-dpu.yaml
@@ -244,6 +244,10 @@ spec:
           value: {{ hasKey .Values.global "enableDynamicUDNAllocation" | ternary .Values.global.enableDynamicUDNAllocation false | quote }}
         - name: OVN_DYNAMIC_UDN_GRACE_PERIOD
           value: {{ default "" .Values.global.dynamicUDNGracePeriod | quote }}
+        - name: OVN_ROUTE_ADVERTISEMENTS_ENABLE
+          value: {{ hasKey .Values.global "enableRouteAdvertisements" | ternary .Values.global.enableRouteAdvertisements false | quote }}
+        - name: OVN_ADVERTISED_UDN_ISOLATION_MODE
+          value: {{ default "strict" .Values.global.advertisedUDNIsolationMode | quote }}
         - name: OVN_NO_OVERLAY_ENABLE
           value: {{ default "false" .Values.global.enableNoOverlay | quote }}
         - name: OVN_HOST_NETWORK_NAMESPACE
--- a/helm/ovn-kubernetes/values-single-node-zone-dpu.yaml
+++ b/helm/ovn-kubernetes/values-single-node-zone-dpu.yaml
@@ -67,6 +67,10 @@ global:
   enableMultiNetwork: false
   # -- Configure to use user defined networks (UDN) feature with ovn-kubernetes
   enableNetworkSegmentation: false
+  # -- Configure to enable no-overlay mode for the default network
+  enableNoOverlay: false
+  # -- Configure to enable SNAT for outbound traffic in no-overlay mode
+  enableNoOverlaySnat: false
   # -- Configure to enable workloads with preconfigured network connect to user defined networks (UDN) with ovn-kubernetes
   enablePreconfiguredUDNAddresses: false
   # -- Configure to enable IPsec


pliurh:
update to address coderabbitai's comments.

test/e2e/route_advertisements.go
```diff
@@ -183,10 +183,34 @@ var _ = ginkgo.Describe("BGP: When default podNetwork is advertised", feature.Ro
 				}
 			}
 
-			ginkgo.By("queries to the external server are not SNATed (uses podIP)")
-			podv4IP, podv6IP, err := podIPsForDefaultNetwork(f.ClientSet, f.Namespace.Name, clientPod.Name)
-			framework.ExpectNoError(err, fmt.Sprintf("Getting podIPs for pod %s failed: %v", clientPod.Name, err))
-			framework.Logf("Client pod IP address v4=%s, v6=%s", podv4IP, podv6IP)
+			// Determine expected source IP based on SNAT configuration
+			var expectedV4IP, expectedV6IP string
+			snatEnabled := os.Getenv("ENABLE_NO_OVERLAY_OUTBOUND_SNAT") == "true"
```
zhaozhanqi: This variable is only available in upstream, it's not existing in D/S.  I suggest check configmap in ovnkube-config.

go-controller/pkg/ovn/address_set/address_set.go
```diff
@@ -55,6 +55,8 @@ type AddressSetFactory interface {
 type AddressSet interface {
 	// GetASHashNames returns the hashed name for v6 and v4 addressSets
 	GetASHashNames() (string, string)
+	// GetASUUID returns the UUIDs for v6 and v4 addressSets
+	GetASUUID() (string, string)
```
trozet, line 59: very nice commit message on this address set commit!

go-controller/pkg/ovn/egressgw.go
```diff
@@ -601,22 +601,24 @@ func (oc *DefaultNetworkController) deletePodSNAT(nodeName string, extIPs, podIP
 	return nil
 }
 
-// buildPodSNAT builds per pod SNAT rules towards the nodeIP that are applied to the GR where the pod resides
-func buildPodSNAT(extIPs, podIPNets []*net.IPNet, match string) ([]*nbdb.NAT, error) {
+// buildPodSNAT builds per pod SNAT rules towards the nodeIP that are applied to the GR where the pod resides.
+// exemptedExtIPs should be an AddressSet name (e.g., "$a123456_cluster_subnets").
```
trozet: or UUID right?

trozet: should we be worried if whatever drives the address set creation deletes/recreates the address set, and now even though it is the same name, the SNAT is now broken because it was referencing a UUID rather than a name?

If we use UUID we will have to be careful that any reconciliation of the address set forces and update of the NAT.

helm/ovn-kubernetes/values-single-node-zone.yaml
```diff
@@ -86,6 +86,8 @@ global:
   advertisedUDNIsolationMode: "strict"
   # -- Configure to enable no-overlay mode for the default network
   enableNoOverlay: false
+  # -- Configure to enable SNAT for outbound traffic in no-overlay mode
```
trozet, line 90: missing DPU helm chart?

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -405,6 +405,15 @@ func (oc *DefaultNetworkController) init() error {
 		}
 	}
 
+	// Cleanup no-overlay SNAT exemption address set if not in no-overlay mode with SNAT enabled.
+	if !util.IsNoOverlaySNATExemptionNeeded(oc) {
+		if err := cleanupNoOverlaySNATExemptionAddressSet(oc.addressSetFactory, oc.GetNetInfo(), oc.controllerName); err != nil {
+			// Cleanup may fail if the address set is still referenced by NAT rules.
+			// But this is harmless and can be retried on next run.
```
trozet: what does this mean? there is no next run of init?

shouldn't this be handled in syncNodes?

go-controller/pkg/ovn/master.go
```diff
@@ -679,6 +679,21 @@ func (oc *DefaultNetworkController) addUpdateLocalNodeEvent(node *corev1.Node, n
...
+	// Sync no-overlay SNAT exemption address set for no-overlay mode with outbound SNAT enabled in SGW mode.
+	// The address set contains cluster CIDRs + local zone node IPs and is used in SNAT
+	// exemption rules to prevent SNATing pod-to-pod and pod-to-local-node traffic.
+	// In LGW mode, nftables sets are used instead.
+	if util.IsNoOverlaySNATExemptionNeeded(oc) && (nSyncs.syncNode || nSyncs.syncGw) {
+		hostAddrs, err := util.GetNodeHostAddrs(node)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to get host addresses for node %s: %w", node.Name, err))
+		} else {
+			if err := syncNoOverlaySNATExemptionAddressSet(oc.addressSetFactory, oc.GetNetInfo(), oc.controllerName, hostAddrs); err != nil {
```
trozet, line 705: dont you also need to handle address set cleanup in the local node delete path? cleanupNodeResources ?

go-controller/pkg/ovn/no_overlay_snat_exemption.go
```diff
@@ -0,0 +1,180 @@
...
+			return fmt.Errorf("failed to set addresses in no-overlay SNAT exemption address set: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// getNoOverlaySNATExemptionAsUUID returns the hashed address set UUIDs for IPv4 and IPv6.
+// These hash names are used to reference the address set in SNAT rules.
+func getNoOverlaySNATExemptionAsUUID(
```
trozet, line 165: note: only unit tested in this commit, I guess it is used in a later commit

go-controller/pkg/ovn/gateway.go
```diff
@@ -873,6 +874,17 @@ func (gw *GatewayManager) updateGWRouterNAT(nodeName string, gwConfig *GatewayCo
 	// DisableSNATMultipleGWs is only applicable to cluster default network and not to user defined networks.
 	// For user defined networks, we always add SNAT rules regardless of whether the network is advertised or not.
 	if !config.Gateway.DisableSNATMultipleGWs || gw.netInfo.IsPrimaryNetwork() {
```
trozet, line 876: just to be 100% sure, outbound snat=true will work with either DisableSNATMultipleGWs true or false, right?

go-controller/pkg/ovn/gateway.go
```diff
@@ -1605,3 +1628,66 @@ func (gw *GatewayManager) oldLayer2TopoCleanup() error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	isNetworkAdvertised := oc.isPodNetworkAdvertisedAtNode(nodeName)
+	gwRouterName := oc.GetNetworkScopedGWRouterName(nodeName)
+
+	if util.IsNoOverlaySNATExemptionNeeded(oc.GetNetInfo()) {
+		// Get the no-overlay SNAT exemption address set UUIDs
+		v4UUID, v6UUID, err = getNoOverlaySNATExemptionAsUUID(oc.addressSetFactory, oc.GetNetInfo(), oc.controllerName)
```
trozet, line 1693: when we are using functions that expect uuids, would it be better to use uuid types instead of string?

go-controller/pkg/ovn/gateway.go
```diff
@@ -893,7 +905,14 @@ func (gw *GatewayManager) updateGWRouterNAT(nodeName string, gwConfig *GatewayCo
...
 
-			nat = libovsdbops.BuildSNATWithMatch(&externalIP[0], entry, "", extIDs, snatMatch)
+			exemptedExtIPs := ""
+			if ipFamily == utilnet.IPv6 {
+				exemptedExtIPs = v6UUID
+			} else {
+				exemptedExtIPs = v4UUID
+			}
+
+			nat = libovsdbops.BuildSNATWithExemptedExtIPs(&externalIP[0], entry, "", extIDs, snatMatch, exemptedExtIPs)
```
trozet, line 915: exempted IPs works with hardware offload? @dceara ?

go-controller/pkg/node/gateway_nftables.go
```diff
@@ -219,7 +219,19 @@ func getUDNNFTRules(service *corev1.Service, netConfig *bridgeconfig.BridgeUDNCo
...
+// 1. A return rule to exempt traffic to cluster CIDRs + local node IPs from SNAT
+// 2. A masquerade rule for all other traffic
+// Example for outboundSNAT=enabled:
+//
+//	chain ovn-kube-pod-subnet-masq {
+//		ip saddr 10.128.0.0/24 ip daddr @no-overlay-snat-exempt-v4 return
+//		ip saddr 10.128.0.0/24 masquerade
+//		ip6 saddr fd00:10:244:1::/64 ip6 daddr @no-overlay-snat-exempt-v6 return
+//		ip6 saddr fd00:10:244:1::/64 masquerade
+//	}
```
trozet, line 233: nice

go-controller/pkg/ovn/egressgw.go
pliurh: My comment is wrong. `exempted_ext_ips` only accept UUID of an address_set, not names. The UUID is fetched fresh from the DB on each `buildPodSNAT` call. And the OVSDB prevents the address set from being deleted while NAT rules reference it, so stale AS UUIDs can't occur either.

helm/ovn-kubernetes/values-single-node-zone.yaml
pliurh: Not sure how DPU works with BGP/NoOverlay. I don't have DPU environment to verify it either. Can we address it in a separated PR?

go-controller/pkg/ovn/default_network_controller.go
pliurh: An address_set cannot be removed if it's still referenced by NAT rules. When toggling outbound SNAT from enabled to disabled, this code may be execuated before all NAT rules get updated by ovnkube-node. In that case, we don't treat this deletion failure as an error. Since the address_set won't be referenced by any NAT rules eventually. And the stale address_set won't do any harm and can be removed in the ovnkube next restart.

pliurh: ~~But I think you're right. I'll move it to `syncNodes()`.~~
I moved the code to `syncNodesPeriodic` to ensure the address set can be removed. Since `syncNodes` is only called once at start up too.

go-controller/pkg/ovn/master.go
pliurh: no-overlay only supports one node per IC zone. If the local node is deleted, the entire local DB goes away with it.

go-controller/pkg/ovn/gateway.go
pliurh: Yes, outboundSNAT=enabled works with both DisableSNATMultipleGWs values. The two settings are orthogonal — outboundSNAT controls whether SNAT happens in no-overlay mode, while DisableSNATMultipleGWs controls SNAT granularity.

go-controller/pkg/ovn/gateway.go
pliurh: The string type is used by `libovsdb` for UUID. We will need a comprehansive refactor to use a new UUID type. I suggest we address it separately.

pliurh:
update to address @trozet and @zhaozhanqi 's comments.

pliurh:
update to resolve rebase conflicts.

go-controller/pkg/ovn/gateway.go
dceara: @trozet Yes, AFAICT, it should work fine with HWOL.  The SNAT zone will be completely skipped if traffic matches the exemption rule.  So we don't do any ct lookup in that zone for such traffic => no potential non-offloadable conntrack matches.

helm/ovn-kubernetes/values-single-node-zone.yaml
trozet: You don't need to validate that it works in DPU environment, but you do need to update the helm charts with support. It should just work, but we will handle testing it for you in our downstream.

Need to add support here:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/values-single-node-zone-dpu.yaml

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/charts/ovnkube-node-dpu/templates/ovnkube-node-dpu.yaml#L248


https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/charts/ovnkube-node-dpu-host/templates/ovnkube-node-dpu-host.yaml#L229

go-controller/pkg/node/node_nftables.go
```diff
@@ -39,6 +40,108 @@ func setupRemoteNodeNFTSets() error {
...
+			tx.Add(&knftables.Element{
+				Set: types.NFTNoOverlaySNATExemptV4,
+				Key: []string{addr},
+			})
+		}
+	}
+
+	// Update IPv6 set if needed
+	if config.IPv6Mode && len(desiredIPv6Addrs) > 0 {
+		// Flush and repopulate the set
```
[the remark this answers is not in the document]
pliurh: The desired entries won't be empty in a running cluster.

go-controller/pkg/ovn/master.go
trozet: I don't understand that one. So if i delete my local kapi node, then recreate it, that doesn't mean my DB got deleted. 

Either way not a common thing, but I guess I'm more concerned about if we restart OVNK and turn off no overlay, things get cleand up on start up. Looks like you moved stuff to syncNodesPeriodic so that is probably good enough.

go-controller/pkg/ovn/gateway.go
```diff
@@ -1054,6 +1075,8 @@ func GetNetworkScopedClusterSubnetSNATMatch(nbClient libovsdbclient.Client, netI
 			return "", nil
 		}
 		return fmt.Sprintf("outport == %q", types.GWRouterToExtSwitchPrefix+netInfo.GetNetworkScopedGWRouterName(nodeName)), nil
+	} else if netInfo.Transport() == types.NetworkTransportNoOverlay && netInfo.OutboundSNAT() == types.NoOverlaySNATEnabled {
+		return "", nil
```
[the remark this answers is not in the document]
pliurh: This function is only used by shared-gateway mode. So the shared-gateway-mode gating is not necessary

go-controller/pkg/ovn/no_overlay_snat_exemption_test.go
```diff
@@ -0,0 +1,222 @@
...
+			// Cleanup
+			err = cleanupNoOverlaySNATExemptionAddressSet(addressSetFactory, netInfo, controllerName)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			// Verify address set was removed - UUIDs should be empty after cleanup
+			as, err = addressSetFactory.GetAddressSet(dbIDs)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			v4UUID, v6UUID = as.GetASUUID()
+			gomega.Expect(v4UUID).To(gomega.BeEmpty())
+			gomega.Expect(v6UUID).To(gomega.BeEmpty())
```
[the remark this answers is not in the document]
pliurh: This test uses NewFakeOVN(false), so it runs against the real ovnAddressSetFactory (backed by a libovsdb test server), not FakeAddressSetFactory. The real GetAddressSet returns an empty address set with no error when no rows match — GetASUUID() correctly returns empty strings.

go-controller/pkg/ovn/gateway.go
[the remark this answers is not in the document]
pliurh: In LGW mode, traffic egresses via mp0, so GR SNAT rules are never used.

pliurh:
update to address coderabbitai's comments.

pliurh:
update to address rebase conflicts.

go-controller/pkg/ovn/gateway.go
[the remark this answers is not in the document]
trozet: @pliurh are we creating GR SNAT with empty match in LGW mode? If so we should remove that.

go-controller/pkg/ovn/master.go
pliurh: Yeah. If the node is deleted, its IPs remain stale in the address set. But when the node returns, `addUpdateLocalNodeEvent(()` will trigger `syncNoOverlaySNATExemptionAddressSet(()` to update the address set with the new node's IPs. So we need to handle the address set cleanup.

pliurh:
Address https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5936#discussion_r2880881466 and add helm DPU support.

pliurh:
Update to resolve merge conflicts.

trozet (approved):

PR #5938, opened 2026-02-05
helm: replace non-functional critical-pod annotation with priorityClassName

booxter:
## 📑 Description

This should address:

```
  I0205 01:12:49.383618 65129 builder.go:156] stderr: "Warning:
  spec.template.metadata.annotations[scheduler.alpha.kubernetes.io/critical-pod]:
  non-functional in v1.16+; use the \"priorityClassName\" field
  instead\n"
```

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Configuration Updates**
  * Pod scheduling priority across relevant system components has been modernized to use Kubernetes PriorityClass ("system-cluster-critical") instead of deprecated scheduling annotations.
  * This update removes legacy annotations and standardizes priority configuration in pod specs for more consistent and supported scheduling behavior.

### DIFF
--- a/dist/templates/ovnkube-node.yaml.j2
+++ b/dist/templates/ovnkube-node.yaml.j2
@@ -27,7 +27,7 @@ spec:
         type: infra
         kubernetes.io/os: "linux"
     spec:
-      priorityClassName: "system-cluster-critical"
+      priorityClassName: "system-node-critical"
       serviceAccountName: ovnkube-node
       hostNetwork: true
       dnsPolicy: Default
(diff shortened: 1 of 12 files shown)


trozet (approved):

trozet (changes requested):
@booxter do you mind changing it for ovnkube-single-node-zone-dpu as well? That one was recently added and want to make sure it doesn't slip through the cracks.

booxter:
There were more hits of the same issue in older files too. Not sure how I managed to miss them. :) I updated all occurrences now.

girishmg (changes requested):
@booxter there are two priority classes -- system-node-critical and system-cluster-critical...

perhaps the ovnkube-node daemonset pod should be system-node-critical and the centralized cluster-manger should be system-cluster-critical.. 

this is what we have in our downstream

PR #5941, opened 2026-02-06
cudn: remove Geneve transport option, use field omission for default

pliurh:
## 📑 Description
Remove "Geneve" from the TransportOption enum in the CUDN API. Omitting
the transport field now signals "use the default OVN overlay transport",
letting ovn-encap-type govern the actual encapsulation. This avoids a
sentinel enum value that merely means "system default".

Drop the internal NetworkTransportDefault constant. Transport() now
returns the stored value directly; empty string is the canonical
representation for default transport throughout the codebase.

Fix CEL validation rules for noOverlay and evpn field guards to use
has(self.transport) before comparing the transport value, preventing
errors when the optional transport field is absent.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Refactor**
  * Network transport default changed from "Geneve" to "Default"
  * Allowed transport options updated to "Default", "NoOverlay", and "EVPN"
  * Configuration defaults, CLI help text, validation messages, and docs updated to reflect the new transport values

* **Tests**
  * Unit and end-to-end tests and test scenarios updated to expect the new default transport and to assert transport metadata where applicable


pliurh:
> Does "ovn" actually have a default overlay transport mechanism? 

It's a required configuration; there is no default in OVN. But for OVN-K, the default is GENEVE is unspecified.

There is a Slack thread discussing this change https://cloud-native.slack.com/archives/C08452HR8V6/p1767884789148339

go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
```diff
@@ -252,7 +253,7 @@ type RoutingOption string
 
 const (
 	TransportOptionNoOverlay TransportOption = "NoOverlay"
-	TransportOptionGeneve    TransportOption = "Geneve"
+	TransportOptionDefault   TransportOption = "Default"
```
tssurya: is there a way to tie this back to `encap-type` ? otherwise is it only me who finds value `default` not telling the user what that means? - should we mention it in the field docs or should we make the default value `ovn-encap-type` or something? 
and side-note - is there precedence for other APIs in Kubernetes to call the value of the "default" as "default"?

jcaamano: Not sure what's wrong with it.

But yeah, we should apidoc that `Default` is the default transport provided by OVN which uses whatever ovn-encap-type is set to.

The previous value, `geneve` does not make sense if ovn-encap-type was set to vxlan.

We can make other suggestions though.

girishmg:
OVN can be deployed with VxLaN as the overlay. We did this in NVIDIA few years ago since some of the NICs didn't offload GENEVE.

Say, we continue to use VxLAN instead of GENEVE. Then if someone creates a UDN with transport set to GENEVE, then OVN will still do VxLAN. 

So, setting the transport to default would mean whatever the deployer has set in ovn-encap-type ... for the most cases it is GENEVE...

pliurh:
Add detailed description for the value `Default`.

pliurh:
rebase the changes.

go-controller/pkg/crd/userdefinednetwork/v1/cudn.go
tssurya: > Not sure what's wrong with it.

not sure what in my comment was not clear,  just having default was not really meaning anything - not sure how that value was conveying the intent back to the user on its own,

> 
> But yeah, we should apidoc that `Default` is the default transport provided by OVN which uses whatever ovn-encap-type is set to.
> 

the docs accompanying change makes this fine, that makes it at least sensible for the user on what default actually means

tssurya:
@pliurh do we really need `default` OR `geneve` really? or shall we just keep `nooverlay`/`evpn` ? Is it really needed to have a default for the sake of it?

pliurh:
> @pliurh do we really need `default` OR `geneve` really? or shall we just keep `nooverlay`/`evpn` ? Is it really needed to have a default for the sake of it?

I think we can just keep `nooverlay`/`evpn`.

pliurh:
@tssurya I refactored the PR. PTAL.

go-controller/pkg/clustermanager/userdefinednetwork/template/net-attach-def-template.go
```diff
@@ -279,18 +279,16 @@ func renderCNINetworkConfig(networkName, nadName string, spec SpecGetter, opts *
...
-		return types.NetworkTransportGeneve
-	case "NoOverlay":
+	case userdefinednetworkv1.TransportOptionNoOverlay:
 		return types.NetworkTransportNoOverlay
-	case "EVPN":
+	case userdefinednetworkv1.TransportOptionEVPN:
 		return types.NetworkTransportEVPN
 	default:
-		return crdTransport // Return as-is for validation to catch
+		return string(crdTransport) // "" for default, or as-is for unknown values
```
tssurya: NIT: I hope we don't take any unknown values cause we have kubebuilder validation right? it can only be NOOVERLAY/EVPN ENUMs or its unset in which case its empty

tssurya (changes requested):
please update the api reference docs on the website as well

pliurh:
update the api reference doc as well

pliurh:
update to address https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5941#discussion_r2857488739

PR #5943, opened 2026-02-06
Fixes missing Dynamic UDN integration, Incorrect logic with GetActiveNetworkForNamespace, adds EgressIP NAD Reconciler

trozet:
Summary Changes-Include:

-  Integrate node -> gateway with Dynamic UDN + do not fail syncServices when a network is not found.
-  Fix Network Policy integration with Dynamic UDN + fix KNP delete path relying on getting active network.
-  Fix Services Controller integration with Dynamic UDN.
-  Fix Egress Firewall initialSync with Dynamic UDN + do not fail initialSync when network is not found.
-  Fix EgressIP integration with Dynamic UDN + upgrade EgressIP to use NAD Reconciler.
-  Adds E2E test to create primary+secondary UDN with a pod, KNP, MNP, egress IP, restart OVNK pods and make sure they come back ready.

## Release Notes

* **New Features**
  * Added NAD reconciler integration for Egress IP pod retry handling on network attachment changes.
  * Introduced per-node gating for network operations to skip processing when networks are unavailable.

* **Bug Fixes**
  * Improved network namespace resolution with refined error handling for invalid primary networks.
  * Enhanced Egress IP and service endpoint processing to handle missing or unavailable networks gracefully.
  * Added nil checks for network availability across multiple network-dependent operations.

* **Tests**
  * Added comprehensive end-to-end test for network segmentation with User-Defined Networks and Egress IP.

### DIFF
--- a/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
+++ b/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
@@ -110,7 +110,7 @@ type matchKind int
 type cacheEntry struct {
 	pgName            string
 	hasNodeSelector   bool
-	subnetsKey        string
+	subnets           []*net.IPNet
 	efResourceVersion string
 	logHash           string
 }
@@ -439,7 +439,7 @@ func (oc *EFController) sync(key string) (updateErr error) {
 			ownerController := activeNetwork.GetNetworkName() + "-network-controller"
 			newEntry = &cacheEntry{
 				pgName:            libovsdbutil.GetPortGroupName(getNamespacePortGroupDbIDs(namespace, ownerController)),
-				subnetsKey:        subnetsKeyForNetInfo(activeNetwork),
+				subnets:           subnetsForNetInfo(activeNetwork),
 				efResourceVersion: ef.ResourceVersion,
 				logHash:           aclLogHash(aclLoggingLevels),
 			}
@@ -535,20 +535,19 @@ func (oc *EFController) sync(key string) (updateErr error) {
 	return
 }
 
-func subnetsKeyForNetInfo(netInfo util.NetInfo) string {
+func subnetsForNetInfo(netInfo util.NetInfo) []*net.IPNet {
 	if netInfo == nil {
-		return ""
+		return nil
 	}
 	subnets := netInfo.Subnets()
-	if len(subnets) == 0 {
-		return ""
+	unsortedSubnets := make([]*net.IPNet, 0, len(subnets))
+	for _, subnet := range subnets {
+		if subnet.CIDR == nil {
+			continue
+		}
+		unsortedSubnets = append(unsortedSubnets, subnet.CIDR)
 	}
-	keys := make([]string, 0, len(subnets))
-	for _, s := range subnets {
-		keys = append(keys, s.String())
-	}
-	slices.Sort(keys)
-	return strings.Join(keys, ",")
+	return util.CopyIPNets(unsortedSubnets)
 }
 
 func entriesEqual(a, b *cacheEntry) bool {
@@ -559,7 +558,7 @@ func entriesEqual(a, b *cacheEntry) bool {
 		return false
 	default:
 		return a.pgName == b.pgName &&
-			a.subnetsKey == b.subnetsKey &&
+			util.IsIPNetsEqual(a.subnets, b.subnets) &&
 			a.efResourceVersion == b.efResourceVersion &&
 			a.logHash == b.logHash
 	}
@@ -619,7 +618,7 @@ func (oc *EFController) addEgressFirewall(egressFirewall *egressfirewallapi.Egre
 
 // validateAndGetEgressFirewallDestination validates an egress firewall rule destination and returns
 // the parsed contents of the destination.
-func (oc *EFController) validateAndGetEgressFirewallDestination(namespace string, egressFirewallDestination egressfirewallapi.EgressFirewallDestination) (
+func (oc *EFController) validateAndGetEgressFirewallDestination(namespace string, egressFirewallDestination egressfirewallapi.EgressFirewallDestination, entry *cacheEntry) (
 	cidrSelector string,
 	dnsName string,
 	clusterSubnetIntersection []*net.IPNet,
@@ -639,18 +638,13 @@ func (oc *EFController) validateAndGetEgressFirewallDestination(namespace string
 			return "", "", nil, nil, err
 		}
 		cidrSelector = egressFirewallDestination.CIDRSelector
-		netInfo, err := oc.networkManager.GetActiveNetworkForNamespace(namespace)
-		if netInfo == nil || err != nil {
-			if err == nil {
-				err = fmt.Errorf("no active network found for namespace %s ", namespace)
-			}
-			return "", "", nil, nil,
-				fmt.Errorf("failed to validate egress firewall destination: %w", err)
+		if entry == nil || entry.subnets == nil {
+			return "", "", nil, nil, fmt.Errorf("failed to "+
+				"validate egress firewall destination: missing cached subnets for namespace %s", namespace)
 		}
-		subnets := netInfo.Subnets()
-		for _, clusterSubnet := range subnets {
-			if clusterSubnet.CIDR.Contains(ipNet.IP) || ipNet.Contains(clusterSubnet.CIDR.IP) {
-				clusterSubnetIntersection = append(clusterSubnetIntersection, clusterSubnet.CIDR)
+		for _, clusterSubnet := range entry.subnets {
+			if clusterSubnet.Contains(ipNet.IP) || ipNet.Contains(clusterSubnet.IP) {
+				clusterSubnetIntersection = append(clusterSubnetIntersection, clusterSubnet)
 			}
 		}
 	} else {
@@ -678,7 +672,7 @@ func (oc *EFController) newEgressFirewallRule(namespace string, rawEgressFirewal
 	// fields of efr.
 	var err error
 	efr.to.cidrSelector, efr.to.dnsName, efr.to.clusterSubnetIntersection, efr.to.nodeSelector, err =
-		oc.validateAndGetEgressFirewallDestination(namespace, rawEgressFirewallRule.To)
+		oc.validateAndGetEgressFirewallDestination(namespace, rawEgressFirewallRule.To, entry)
 	if err != nil {
 		return efr, err
 	}
--- a/go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
+++ b/go-controller/pkg/ovn/controller/egressfirewall/egress_firewall_test.go
@@ -484,7 +484,7 @@ var _ = ginkgo.Describe("OVN test basic functions", func() {
 				subnets = append(subnets, config.CIDRNetworkEntry{CIDR: cidr})
 			}
 			config.Default.ClusterSubnets = subnets
-			entry := &cacheEntry{}
+			entry := &cacheEntry{subnets: subnetsForNetInfo(&util.DefaultNetInfo{})}
 			output, err := efController.newEgressFirewallRule("default", tc.egressFirewallRule, tc.id, entry)
 			if tc.err == true {
 				gomega.Expect(err).To(gomega.HaveOccurred())
@@ -716,8 +716,13 @@ func TestValidateAndGetEgressFirewallDestination(t *testing.T) {
 			if len(tc.udnName) > 0 {
 				network = tc.udnName
 			}
+			entry := &cacheEntry{subnets: subnetsForNetInfo(&util.DefaultNetInfo{})}
+			if len(tc.udnName) > 0 {
+				entry.subnets = subnetsForNetInfo(netInfo)
+			}
+
 			cidrSelector, dnsName, clusterSubnetIntersection, nodeSelector, err :=
-				efController.validateAndGetEgressFirewallDestination(network, tc.egressFirewallDestination)
+				efController.validateAndGetEgressFirewallDestination(network, tc.egressFirewallDestination, entry)
 			if tc.expectedErr {
 				require.Error(t, err)
 			} else {
--- a/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall_sync_test.go
+++ b/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall_sync_test.go
@@ -187,5 +187,5 @@ func TestEFControllerSync_UpdatesOnSubnetChangeAndSkipsWhenUnchanged(t *testing.
 	entry, ok := oc.cache.Load(namespace)
 	require.True(t, ok)
 	require.Equal(t, pgName, entry.pgName)
-	require.Equal(t, subnetsKeyForNetInfo(netInfo2), entry.subnetsKey)
+	require.True(t, util.IsIPNetsEqual(subnetsForNetInfo(netInfo2), entry.subnets))
 }
--- a/go-controller/pkg/util/net.go
+++ b/go-controller/pkg/util/net.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/big"
 	"net"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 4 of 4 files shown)


go-controller/pkg/networkmanager/api.go
```diff
@@ -106,6 +107,51 @@ type Controller interface {
 	Stop()
 }
 
+// ResolveActiveNetworkForNamespaceOnNode returns the active primary network for the namespace only if
+// the network is active on the given node. It uses the NAD cache to resolve the namespace's primary
+// NAD and network name, checks NodeHasNetwork, and then calls GetActiveNetworkForNamespace to fetch
+// the rendered NetInfo. It returns (nil, false, nil) when the network is not active on the node.
+func ResolveActiveNetworkForNamespaceOnNode(nm Interface, nodeName, namespace string) (util.NetInfo, bool, error) {
```
npinaeva: also `resolve` in the name is a bit confusing, why not `GetActive...`?

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -890,13 +893,17 @@ func (npw *nodePortWatcher) AddService(service *corev1.Service) error {
 
 	klog.V(5).Infof("Adding service %s in namespace %s", service.Name, service.Namespace)
 
-	netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(service.Namespace)
```
npinaeva, line 893: I am not following what is the special part about node networkmanager that this can't/doesn't work?

go-controller/pkg/util/util.go
```diff
@@ -366,6 +366,8 @@ func IsClusterIP(svcVIP string) bool {
 	return false
 }
 
+// UnprocessedActiveNetworkError indicates that the primary UDN is required and exists
+// for a namespace, but our network manager has not processed it yet.
```
npinaeva: that's nice

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -1193,15 +1216,6 @@ func (bnc *BaseNetworkController) DeleteResourceCommon(objType reflect.Type, obj
 		if !ok {
 			return fmt.Errorf("could not cast obj of type %T to *knet.NetworkPolicy", obj)
 		}
-		netinfo, err := bnc.networkManager.GetActiveNetworkForNamespace(knp.Namespace)
```
npinaeva, line 1196: "This fixes it by checking the nad controller cache instead, which will
always have the full info to determine if the KNP belongs to CDN."
doesn't `GetActiveNetworkForNamespace` alredy check nad controller cache?

go-controller/pkg/ovn/controller/services/services_controller.go
```diff
@@ -600,22 +600,36 @@ func (c *Controller) RequestFullSync(nodeInfos []nodeInfo) {
...
 		}
 
+		serviceNetworkName := types.DefaultNetworkName
+		isDefaultNetwork := serviceNAD == types.DefaultNetworkName
+		if !isDefaultNetwork {
+			serviceNetworkName = c.networkManager.GetNetworkNameForNADKey(serviceNAD)
+			if serviceNetworkName == "" {
+				return true
+			}
+		}
```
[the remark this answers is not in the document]
trozet: If the UDN controller gets IsUnprocessedActiveNetworkError, then it knows this service is meant for a UDN, but it doesn't know which UDN, so it has to fail to retry. Relying on endpoint slice to trigger requeue doesn't mean that the local UDN controller will have processed the NAD by then. The endpoint slice generation for UDNs is done by cluster manager (centralized component) not the node's UDN controller running as part of the zone/ovnkube-controller.

If we are the default network controller and we get an unprocessed active network error, then we know the namespace the service in requires primary UDN, so as the default network we know we don't care about this service and can skip.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -1168,6 +1170,30 @@ func (bnc *BaseNetworkController) AddResourceCommon(objType reflect.Type, obj in
...
+			}
+		} else {
+			networkName := bnc.networkManager.GetNetworkNameForNADKey(foundNamespaceNAD)
+			if networkName == "" {
+				return fmt.Errorf("no primary network found for namespace %s", np.Namespace)
+			}
+			if bnc.GetNetworkName() != networkName {
+				return nil
+			}
+		}
```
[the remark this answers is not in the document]
trozet: makes sense, if the error is invalid primary network, then this network requires UDN, but none exists, so it cannot be our UDN/CDN controller since our controller is already running.

go-controller/pkg/ovn/base_network_controller.go
trozet: GetActiveNetworkForNamespace not only checks nadController, but also grabs the info from the network controller cache. It's not safe to rely on for dynamic UDN cause we dont add networks to the network controller cache when they are not active.

go-controller/pkg/node/gateway_shared_intf.go
trozet: I think https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5943#discussion_r2783148533 clarifies it

go-controller/pkg/ovn/controller/services/services_controller.go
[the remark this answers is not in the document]
trozet: @coderabbitai did you see this?

go-controller/pkg/ovn/base_network_controller.go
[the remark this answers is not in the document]
trozet: Note, GetPrimaryNADForNamespace today does not check/return InvalidPrimaryNetworkError. We could add that check, but we would need to go through every current caller of that function today and see how they should handle that error.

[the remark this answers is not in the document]
trozet: I'll defer this to another PR where I can just focus on it there.

trozet:
UT failure was not due to this PR and reported and fixed here: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5948

trozet:
Well the new e2e I added found an egress IP crash  that has nothing to do with this PR :) https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5950

go-controller/pkg/ovn/base_network_controller.go
npinaeva: can't you just return a new error type from the `GetActiveNetworkForNamespace` instead?

npinaeva: I think currently it returns `NewUnprocessedActiveNetworkError`, which is also wrong because we have processed the network?

trozet: No, it breaks the entire definition. The keyword is "Active". On nodes where dynamic UDN is considering a node inactive, it has no Active network. 

Therefore from the caller perspective it interprets that as "the NAD hasn't yet been processed on my NAD controller, I should wait for it (by raising error, retrying later, etc)".

But this semantic is wrong, because on an inactive node the network will never become ready, and it will be stuck in an infinite loop waiting on something that will never exist if the node doesn't go active.

The correct semantic is to use the GetPrimaryNADForNamespace, which is dynamic UDN agnostic and will actually return the information when the NAD controller processed this NAD, not when the network became active on this node.

After identifying the NAD controller has processed this NAD, the caller can then proceed to (if it needs to):
1. Now should this node be active? (NodeHasNetwork)
2. Get me the Active Network (GetActiveNetworkForNamespace)

Some callers don't need the additional info, and some do.

go-controller/pkg/networkmanager/api.go
```diff
@@ -44,11 +43,11 @@ type watchFactory interface {
 // information to the rest of the project.
 type Interface interface {
 	// GetActiveNetworkForNamespace returns a copy of the primary network for
-	// the namespace if any or the default network otherwise. If there is a
-	// primary UDN defined but the NAD has not been processed yet, returns
-	// ErrNetworkControllerTopologyNotManaged. Used for controllers that are not
-	// capable of reconciling primary network changes. If unsure, use this one
-	// and not GetActiveNetworkForNamespaceFast.
+	// the namespace if any or the default network otherwise.
+	// If the network is non-existent for a legitimate reason (namespace gone or
```
npinaeva, line 47: I think it is also important to say that it returns no error in this case. People may expect `err == nil` means `netInfo != nil`

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -847,56 +850,37 @@ func (c *nadController) GetActiveNetworkForNamespace(namespace string) (util.Net
...
-				return nil, util.NewUnprocessedActiveNetworkError(namespace, cudn.Name)
-			}
+	// no network exists in the network manager
+	if primaryNAD != "" {
+		if config.OVNKubernetesFeature.EnableDynamicUDNAllocation {
+			// primary NAD exists, no network, and DUDN is enabled, treat this like the network doesn't exist
+			return nil, nil
 		}
+		// primary NAD exists, but missing in network manager. This should never happen.
+		panic(fmt.Sprintf("NAD Controller broken consistency with Network Manager for primary NAD: %s", primaryNAD))
```
npinaeva, line 881: not sure if this is the best way out...

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -567,6 +567,10 @@ func (c *Controller) processEIP(eip *eipv1.EgressIP) (*eIPConfig, sets.Set[strin
 			if err != nil {
 				return nil, selectedNamespaces, selectedPods, selectedNamespacesPodIPs, fmt.Errorf("failed to get active network for namespace %s: %v", namespace.Name, err)
 			}
+			if netInfo == nil {
+				// no active network
+				continue
+			}
```
npinaeva, line 573: isn't that beautiful?!

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1476,10 +1486,14 @@ func (npw *nodePortWatcher) UpdateEndpointSlice(oldEpSlice, newEpSlice *discover
 	var err error
 	var errors []error
 
-	netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(newEpSlice.Namespace)
+	netInfo, err := npw.networkManager.GetActiveNetworkForNamespace(oldEpSlice.Namespace)
```
npinaeva: shoudln't namespace be the same on both?

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
```diff
@@ -422,19 +422,12 @@ func (oc *EFController) sync(key string) (updateErr error) {
...
-			}
-			if util.IsInvalidPrimaryNetworkError(netErr) {
-				// Namespace requires P-UDN, but it does not exist. Remove EF config and surface error in status.
-				updateErr = netErr
-			} else {
-				return fmt.Errorf("failed to get active network for egress firewall %s/%s namespace: %w",
-					namespace, efName, netErr)
+		if activeNetwork == nil || netErr != nil {
+			// Namespace requires P-UDN, but it does not exist. Remove EF config and surface error in status.
+			if netErr == nil {
```
npinaeva: if `activeNetwork == nil` don't we want to just ignore this EF with no error?

go-controller/pkg/ovn/controller/services/services_controller.go
```diff
@@ -600,22 +600,36 @@ func (c *Controller) RequestFullSync(nodeInfos []nodeInfo) {
 // belong to the network that this service controller is responsible for.
 func (c *Controller) skipService(name, namespace string) bool {
 	if util.IsNetworkSegmentationSupportEnabled() {
-		serviceNetwork, err := c.networkManager.GetActiveNetworkForNamespace(namespace)
+		serviceNAD, err := c.networkManager.GetPrimaryNADForNamespace(namespace)
```
npinaeva, line 603: can you clarify the difference between using `GetActiveNetworkForNamespace` and then checking for the same errors + nil  vs this? I think that would be easier to read, but maybe there is some difference

go-controller/pkg/ovn/egressip.go
```diff
@@ -2501,6 +2517,23 @@ func (e egressStatuses) delete(deleteStatus egressipv1.EgressIPStatusItem) {
 	delete(e.statusMap, deleteStatus)
 }
 
+func podIPSliceEqual(oldIPs, newIPs []net.IP) bool {
```
npinaeva, line 2550: nit: there is a similar function in `dns.go` `ipsEqual(oldips, newips []net.IP) bool` if you feel like it, could be nice to unify

go-controller/pkg/networkmanager/nad_controller.go
trozet: it is the same pattern we already use here: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/nad_controller.go#L1045

and here:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/nad_controller.go#L1054

and here:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/nad_controller.go#L956

go-controller/pkg/node/gateway_shared_intf.go
trozet: yeah I just messed that up, I'll revert it

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
trozet: I debated this one a bit. My thought process was if we have no active network, it could be because someone deleted the primary UDN for this namespace. If that happens, we want to remove egress firewall config, and then I think it makes sense to propagate an error to EF saying this thing has no UDN, will not render rules.

go-controller/pkg/ovn/controller/services/services_controller.go
trozet: Similar, but GetPrimaryNADForNamespace is slightly less lock contentious as it doesn't need to query network controller. GetPrimaryNADForNamespace also doesn't filter with dynamic UDN, so controllers that don't want to get back information that is filtered.

I'll update the API comments to reflect his.

go-controller/pkg/ovn/egressip.go
trozet: let me see if it is easy enough to do

trozet: Looked at this and the functions arent really the same. The isEqual in dns.go just checks for membership, while the podIPSliceEqual actually checks for slice equality. For example [1.1.1.1, 1.1.1.1] vs [1.1.1.1, 2.2.2.2] would be false in podIPSliceEqual, but ipsEqual can return true. Not sure if it is intentional or a bug in ipsEqual, but regardless I'd rather not try to address that in this PR.

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: alright if we are not hitting constant crashloops then maybe it works

go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go
npinaeva: the cleanup has to happen, but I am not sure if the error is more helpful or misleading. Especially in the dynamic udn case, we don't want to report status errors from the nodes that have UDN filtered out?

trozet: yeah good point. I'll change it to not update any status if the UDN is filtered out

trozet:
Reported flake here: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5978

Been seeing multiple flakes on this bgp advertise-default dualstack lane in different places. I don't think it is related to this PR.

Issue #5948, opened 2026-02-09
UT Flake: TestNADController NAD added then deleted


trozet:
### Which jobs are flaking?

unit test

### Which tests are flaking?

TestNADController/NAD_added_then_deleted

### Since when has it been flaking?

Just saw it 2/9/2026

### Reason for failure (if possible)

This flake can happen due to a race/improper check in the test where nad controller syncs, then immediately checks if started controllers == expected. In this case there are no expected controllers, so the check can pass immediately before the network actually sync'ed via network controller. The test then tests consistently, at which point the network controller could have finished adding, but has not deleted yet, causing the test to fail.

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21835152046/job/63004016430?pr=5943

2026-02-09T17:58:14.9837755Z === RUN   TestNADController/NAD_added_then_deleted
  2026-02-09T17:58:15.0209167Z I0209 17:58:15.020616   27410 network_controller.go:373] [ network controller]: syncing all networks
  2026-02-09T17:58:15.0210757Z I0209 17:58:15.020654   27410 network_controller.go:384] [ network controller]: finished syncing all networks. Time taken: 45.254µs
  2026-02-09T17:58:15.0211824Z I0209 17:58:15.020676   27410 controller.go:157] Starting controller [ network controller] with 1 workers
  2026-02-09T17:58:15.0220216Z I0209 17:58:15.021774   27410 nad_controller.go:752] : ensuring NAD "test/nad_1" reference for network "networkAPrimary" with id 1
  2026-02-09T17:58:15.0221485Z I0209 17:58:15.021926   27410 network_controller.go:400] [ network controller]: sync network networkAPrimary
  2026-02-09T17:58:15.0222817Z I0209 17:58:15.021920   27410 nad_controller.go:700] : removing NAD "test/nad_1" reference for network "networkAPrimary"
  2026-02-09T17:58:15.0223833Z starting network: networkAPrimary layer2
  2026-02-09T17:58:15.0225657Z I0209 17:58:15.022300   27410 network_controller.go:402] [ network controller]: finished syncing network networkAPrimary, took 376.503µs
  2026-02-09T17:58:15.0227640Z I0209 17:58:15.022373   27410 network_controller.go:400] [ network controller]: sync network networkAPrimary
[...]

PR #5949, opened 2026-02-09
Fix asynchronous race with nad controller UT

trozet:
With the nad controller tests, it syncsNAD then waits for an Eventually clause to be satisified that checks if the expected number of networks started, stopped, etc. Afterwards, it then checks these values are consistent. The fake network controller uses a controller tcm lock to store started, deleted networks on reconcile.

A race exists because network controller is an asynchronous reocncile loop, so although the syncNAD calls in the test can execute, it does not mean network controller has processed anything. This can lead to false positives on pass criteria where this kind of event loop happens:

1. Sync NAD (create network), sync nad (delete network), expect running (0 networks).
2. Before network controller runs, eventually code gets hit and takes tcm lock. Checks that started networks (empty) contains expected running networks (empty). Passes eventually checks.
3. Create network processes now that tcm lock is released. Now tcm has 1 started network, 0 stopped networks.
4. Eventually loop processes. Started networks (1) contains expected running networks (empty). Passes due to contains semantics. Stopped network check expects 1, but fails. Test flake
5. Delete never go ta chance to run.

To fix this, synchronize syncNetwork calls when syncNAD happens. It's the only way to truly ensure the actions are completed by the time we want to test pass/fail criteria.

Fixes: #5948

## Summary by CodeRabbit

* **Tests**
  * Improved test infrastructure for network controller testing with more deterministic reconciliation flow.

### DIFF
--- a/go-controller/pkg/networkmanager/nad_controller_test.go
+++ b/go-controller/pkg/networkmanager/nad_controller_test.go
@@ -178,12 +178,14 @@ func (tnc *testNetworkController) Start(context.Context) error {
 func (tnc *testNetworkController) Stop() {
 	tnc.tcm.Lock()
 	defer tnc.tcm.Unlock()
+	fmt.Printf("stopping network: %s\n", testNetworkKey(tnc))
 	tnc.tcm.stopped = append(tnc.tcm.stopped, testNetworkKey(tnc))
 }
 
 func (tnc *testNetworkController) Cleanup() error {
 	tnc.tcm.Lock()
 	defer tnc.tcm.Unlock()
+	fmt.Printf("cleaning up network: %s\n", testNetworkKey(tnc))
 	tnc.tcm.cleaned = append(tnc.tcm.cleaned, testNetworkKey(tnc))
 	return nil
 }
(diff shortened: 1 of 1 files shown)


jcaamano:
```
Eventually loop processes. Started networks (1) contains expected running networks (empty). Passes due to contains semantics. Stopped network check expects 1, but fails. Test flake
Delete never go ta chance to run.
```
You mean that `Eventually` times out before it has a chance to find the expected conditions. Or is it `Consistently` that makes the test fail? 


Have you thought about tweaking this line:
```
expectStopped := sets.New(tcm.started...).Difference(sets.New(expectRunning...)).UnsortedList()
```
so instead of using `tcm.started` there, we should know what networks need to be started based on test args.

trozet:
> ```
> Eventually loop processes. Started networks (1) contains expected running networks (empty). Passes due to contains semantics. Stopped network check expects 1, but fails. Test flake
> Delete never go ta chance to run.
> ```
> 
> You mean that `Eventually` times out before it has a chance to find the expected conditions. Or is it `Consistently` that makes the test fail?
> 
> Have you thought about tweaking this line:
> 
> ```
> expectStopped := sets.New(tcm.started...).Difference(sets.New(expectRunning...)).UnsortedList()
> ```
> 
> so instead of using `tcm.started` there, we should know what networks need to be started based on test args.

@jcaamano the Eventually passes in my case (false positive), but the Consistently immediately fails. I did think about what you suggested, but that would need more context to each test to know which ones should be created. Like:

1. Eventually g.Expect(tcm.started).To(gomega.ContainElements(expectRunning) - we expect none running, empty slice, so passes, even if tcm.started is empty
2. We add context to the test case to have expectStarted so g.Expect(tcm.started).To(gomega.ContainElements(expectStarted) - problem 1 is solved
3. Next evaluation is g.Expect(tcm.stopped).To(gomega.ConsistOf(expectStopped), so now we to calculate from expectedStarted and expectedRunning, which ones should be stopped. Then assert tcm.stopped contains them.

I think this is worse than forcing nm sync to happen during each nad controller operation.

jcaamano:
> but that would need more context to each test to know which ones should be created

wouldn't it just be the difference between input networks and expected networks?

> I think this is worse than forcing nm sync to happen during each nad controller operation.

I understand this might be the best option if there is no straight forward alternative. So if my suggestion above doesn't really work, and we don't really want to add more args to test input, then agree to go with this alternative.

jcaamano (approved):

Issue #5950, opened 2026-02-09
Coredump with egress IP modifying node informer cache objects


trozet:
### What happened?

Egress IP coredump during newly added e2e test:

 https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21835152046/job/63005106583?pr=5943

BT shows egress IP modifying informer node objects triggering the crash:

```
Sending output to pager...
 0  0x0000000000483141 in runtime.raise
    at /usr/local/go/src/runtime/sys_linux_amd64.s:154
 1  0x000000000045ae85 in runtime.dieFromSignal
    at /usr/local/go/src/runtime/signal_unix.go:967
 2  0x000000000045b4c6 in runtime.sigfwdgo
    at /usr/local/go/src/runtime/signal_unix.go:1179
 3  0x0000000000459d45 in runtime.sigtrampgo
    at /usr/local/go/src/runtime/signal_unix.go:432
 4  0x0000000000483141 in runtime.raise
    at /usr/local/go/src/runtime/sys_linux_amd64.s:153
 5  0x000000000045ae85 in runtime.dieFromSignal
    at /usr/local/go/src/runtime/signal_unix.go:967
 6  0x000000000044171a in runtime.crash
    at /usr/local/go/src/runtime/signal_unix.go:1056
 7  0x000000000044171a in runtime.fatalthrow.func1
    at /usr/local/go/src/runtime/panic.go:1287
 8  0x000000000047fa6a in runtime.systemstack
    at /usr/local/go/src/runtime/asm_amd64.s:514
 9  0x000000000047fa08 in runtime.systemstack_switch
    at /usr/local/go/src/runtime/asm_amd64.s:479
10  0x0000000000441678 in runtime.fatalthrow
    at /usr/local/go/src/runtime/panic.go:1276
11  0x00000000004412dc in runtime.fatal
    at /usr/local/go/src/runtime/panic.go:1123
12  0x0000000000479998 in internal/runtime/maps.fatal
    at /usr/local/go/src/runtime/panic.go:1058
13  0x000000000040c877 in runtime.mapaccess2_faststr
    at /usr/local/go/src/internal/runtime/maps/runtime_faststr_swiss.go:174
14  0x0000000001e5f285 in github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util.GetNodeIfAddrAnnotation
    at /workspace/ovn-kubernetes/go-controller/pkg/util/node_annotations.go:647
[...]

PR #5951, opened 2026-02-09
EgressIP: Fix crash from mutating node informer object

trozet:
Code was modifying the annotations of the informer cache node object. If this was happening while another goroutine was reading the annotation map, it would trigger ovnkube to crash!

Fixes: #5950

## Summary by CodeRabbit

* **Bug Fixes**
  * Node and pod annotation updates now operate on copied objects and initialize missing annotation storage to avoid unintended mutations and improve update reliability.
* **New Features**
  * Added a local annotation cache and an atomic update path to keep bridge egress IP annotations synchronized and consistent.
* **Tests**
  * Tests updated to observe eventual annotation state, improving resilience to asynchronous updates.

### DIFF
--- a/go-controller/pkg/node/egressip/gateway_egressip.go
+++ b/go-controller/pkg/node/egressip/gateway_egressip.go
@@ -175,6 +175,7 @@ type BridgeEIPAddrManager struct {
 	nodeName         string
 	bridgeName       string
 	nodeAnnotationMu sync.Mutex
+	annotationIPs    sets.Set[string]
 	eIPLister        egressiplisters.EgressIPLister
 	eIPInformer      cache.SharedIndexInformer
 	nodeLister       corev1listers.NodeLister
@@ -195,6 +196,7 @@ func NewBridgeEIPAddrManager(nodeName, bridgeName string, linkManager *linkmanag
 		nodeName:         nodeName,     // k8 node name
 		bridgeName:       bridgeName,   // bridge name for which EIP IPs are managed
 		nodeAnnotationMu: sync.Mutex{}, // mu for updating Node annotation
+		annotationIPs:    sets.New[string](),
 		eIPLister:        eIPInformer.Lister(),
 		eIPInformer:      eIPInformer.Informer(),
 		nodeLister:       nodeInformer.Lister(),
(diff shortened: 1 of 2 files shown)


npinaeva:
you know I love to reward good work with more work...
I found some more usages of `UpdateNodeStatus` that don't do a deepcopy in `gateway_egressip` can you please update those too?

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -305,6 +307,9 @@ func (g *BridgeEIPAddrManager) SyncEgressIP(objs []interface{}) error {
 	if err != nil {
 		return fmt.Errorf("failed to sync EgressIP gateway config because unable to get Node annotation: %v", err)
 	}
+	g.nodeAnnotationMu.Lock()
+	g.annotationIPs = sets.New[string](getIPsStr(annotIPs...)...)
+	g.nodeAnnotationMu.Unlock()
```
npinaeva, line 312: I don't think will work with the optimization where we skip the update if the ip doesn't exist. So if we get a stale annotation here and ip1 is not in the annotation from the cache, but it is in the annotation in kapi, then calling delete(ip1) will check that annotation doesn't have it and skip?

trozet: I'm not sure in what case the cache is out of sync with kapi, but I'll just remove the skip and always update.

npinaeva: Oh i think I see the misunderstanding, I thought `SyncEgressIP` is the regular sync function, but it is actually the initial sync. I don't think removing the skip unneeded updates is a good idea, as kapi calls are quite expensive.
let me add another question to the part that confused me

go-controller/pkg/node/egressip/gateway_egressip.go
```diff
@@ -344,35 +349,29 @@ func (g *BridgeEIPAddrManager) SyncEgressIP(objs []interface{}) error {
...
 	}
+	g.nodeAnnotationMu.Lock()
+	g.annotationIPs = sets.New[string]()
+	for _, ipv4 := range configs.v4 {
+		g.annotationIPs.Insert(ipv4)
+	}
+	for _, ipv6 := range configs.v6 {
+		g.annotationIPs.Insert(ipv6)
+	}
+	g.nodeAnnotationMu.Unlock()
```
npinaeva: so why do we need that, shouldn't cache me already correctly updated after the `addIPToAnnotation` and `deleteIPsFromAnnotation`?

trozet: Agree it's not needed. Removing.

npinaeva (approved):
minus 10 races from the egressip code, a couple more to go :)

trozet:
just going to rebase to pick up the CI fix

PR #5954, opened 2026-02-10
Fixes e2e sflow test

trozet:
Changes-Include:

- Fix `kind` network interface lookup to work with BusyBox and iproute2:
  - Switch from `ip -br -4/-6 a sh` to `ip -o -f inet|inet6 addr show` parsing.
  - Keep interface validation via `ip link show`.
  - Correct IPv6 log message text (`IPv6` instead of `IPv4`).

- Make flow collector startup deterministic per protocol:
  - Start `goflow` with protocol-specific flags so only the target collector is enabled.
  - Avoid mixed startup log lines from unrelated collectors.

- Improve sflow test robustness:
  - Wait for `br-int` sflow row to appear per `ovnkube-node` pod before tuning.
  - Tune `sampling=1 polling=1` after row exists (best-effort if set fails).

- Fix flow log assertion flake:
  - Search all collector log lines for expected keyword instead of only the last line.

- Fix stale/incorrect error handling:
  - Use the actual `ExecWithOptions` returned error for ovs-vsctl checks.
  - Fix variable scoping bug in sflow tuning block (`setErr/setStderr`).

Not 100% sure this will solve the issue, but I ran this over 300 times in a row and did not see a failure in my setup.

Fixes: #4221

## Summary by CodeRabbit

* **Tests**
  * Enhanced end-to-end tests for the flow collector: dynamic sFlow setup on nodes and more reliable collector startup arguments.
  * More robust log validation by scanning all collector logs and surfacing the matching line for easier diagnosis.
  * Improved test-side command execution and error reporting, plus sturdier IP/interface detection parsing in local test environments.

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -1948,6 +1948,20 @@ var _ = ginkgo.Describe("e2e br-int flow monitoring export validation", func() {
 		return fmt.Sprintf(collectorContainerTemplate, port)
 	}
 
+	getCollectorArgs := func(protocol flowMonitoringProtocol, port uint16) []string {
+		args := []string{"-kafka=false"}
+		switch protocol {
+		case sflow:
+			// Disable other collectors to avoid non-deterministic startup ordering in logs.
+			args = append(args, "-nf=false", "-nfl=false", "-sflow=true", fmt.Sprintf("-sflow.port=%d", port))
+		case netflow_v5:
+			args = append(args, "-nf=false", "-sflow=false", "-nfl=true", fmt.Sprintf("-nfl.port=%d", port))
+		case ipfix:
+			args = append(args, "-nfl=false", "-sflow=false", "-nf=true", fmt.Sprintf("-nf.port=%d", port))
+		}
+		return args
+	}
+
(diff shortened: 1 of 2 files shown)


test/e2e/e2e.go
```diff
@@ -1997,13 +2063,14 @@ var _ = ginkgo.Describe("e2e br-int flow monitoring export validation", func() {
...
-						framework.Logf("Successfully found string %s in last log line of"+
-							" the collector: %s", keyword, lastLine)
-						return true, nil
+					for _, line := range logLines {
+						if strings.Contains(line, keyword) {
+							framework.Logf("Successfully found string %s in collector logs line: %s", keyword, line)
+							return true, nil
+						}
 					}
+					lastLine := logLines[len(logLines)-1]
```
npinaeva: that log message doesn't make sense anymore? should just say `"%s not found in collector logs`?

npinaeva (commented):
I guess this is reasonable :D

test/e2e/e2e.go
trozet: done

PR #5959, opened 2026-02-11
Minimize ACLs by combining ipBlocks into single ACL

pperiyasamy:
Combine all ipBlocks in a NetworkPolicy rule into single ACL instead of creating one ACL per ipBlock.
This reduces ACL bloat when policy having multiple ipBlocks for ingress/egress rules.

## Summary by CodeRabbit

* **Refactor**
  * Optimized network policy IP block handling by consolidating multiple per-block ACLs into a single unified ACL for improved performance and streamlined management.

### DIFF
--- a/go-controller/pkg/ovn/gress_policy.go
+++ b/go-controller/pkg/ovn/gress_policy.go
@@ -22,6 +22,11 @@ import (
 const (
 	// emptyIdx is used to create ACL for gressPolicy that doesn't have ipBlocks
 	emptyIdx = -1
+	// ipBlockCombinedIdx is used when creating an ACL for a gressPolicy
+	// that contains ipBlocks. Previously, one ACL was created per ipBlock.
+	// This is changed to create a single combined ACL for all ipBlocks,
+	// and this special index value identifies those new ACLs.
+	ipBlockCombinedIdx = -2
 )
 
 type gressPolicy struct {
@@ -167,14 +172,14 @@ func (gp *gressPolicy) allIPsMatch() string {
 	}
 }
 
-func (gp *gressPolicy) getMatchFromIPBlock(lportMatch, l4Match string) []string {
+func (gp *gressPolicy) getMatchFromIPBlock(lportMatch, l4Match string) string {
 	var direction string
 	if gp.policyType == knet.PolicyTypeIngress {
 		direction = "src"
 	} else {
 		direction = "dst"
 	}
-	var matchStrings []string
+	var ipBlockMatches []string
 	var matchStr, ipVersion string
 	for _, ipBlock := range gp.ipBlocks {
 		if utilnet.IsIPv6CIDRString(ipBlock.CIDR) {
@@ -185,17 +190,22 @@ func (gp *gressPolicy) getMatchFromIPBlock(lportMatch, l4Match string) []string 
 		if len(ipBlock.Except) == 0 {
 			matchStr = fmt.Sprintf("%s.%s == %s", ipVersion, direction, ipBlock.CIDR)
 		} else {
-			matchStr = fmt.Sprintf("%s.%s == %s && %s.%s != {%s}", ipVersion, direction, ipBlock.CIDR,
+			matchStr = fmt.Sprintf("(%s.%s == %s && %s.%s != {%s})", ipVersion, direction, ipBlock.CIDR,
 				ipVersion, direction, strings.Join(ipBlock.Except, ", "))
 		}
-		if l4Match == libovsdbutil.UnspecifiedL4Match {
-			matchStr = fmt.Sprintf("%s && %s", matchStr, lportMatch)
-		} else {
-			matchStr = fmt.Sprintf("%s && %s && %s", matchStr, l4Match, lportMatch)
-		}
-		matchStrings = append(matchStrings, matchStr)
+		ipBlockMatches = append(ipBlockMatches, matchStr)
 	}
-	return matchStrings
+	var l3Match string
+	if len(ipBlockMatches) == 1 {
+		l3Match = ipBlockMatches[0]
+	} else {
+		l3Match = fmt.Sprintf("(%s)", strings.Join(ipBlockMatches, " || "))
+	}
+
+	if l4Match == libovsdbutil.UnspecifiedL4Match {
+		return fmt.Sprintf("%s && %s", l3Match, lportMatch)
+	}
+	return fmt.Sprintf("%s && %s && %s", l3Match, l4Match, lportMatch)
 }
 
 // addNamespaceAddressSet adds a namespace address set to the gress policy.
@@ -285,13 +295,11 @@ func (gp *gressPolicy) buildLocalPodACLs(portGroupName string, aclLogging *libov
 	for protocol, l4Match := range libovsdbutil.GetL4MatchesFromNetworkPolicyPorts(gp.portPolicies) {
 		if len(gp.ipBlocks) > 0 {
 			// Add ACL allow rule for IPBlock CIDR
-			ipBlockMatches := gp.getMatchFromIPBlock(lportMatch, l4Match)
-			for ipBlockIdx, ipBlockMatch := range ipBlockMatches {
-				aclIDs := gp.getNetpolACLDbIDs(ipBlockIdx, protocol)
-				acl := libovsdbutil.BuildACLWithDefaultTier(aclIDs, types.DefaultAllowPriority, ipBlockMatch, action,
-					aclLogging, gp.aclPipeline)
-				createdACLs = append(createdACLs, acl)
-			}
+			ipBlockMatch := gp.getMatchFromIPBlock(lportMatch, l4Match)
+			aclIDs := gp.getNetpolACLDbIDs(ipBlockCombinedIdx, protocol)
+			acl := libovsdbutil.BuildACLWithDefaultTier(aclIDs, types.DefaultAllowPriority, ipBlockMatch, action,
+				aclLogging, gp.aclPipeline)
+			createdACLs = append(createdACLs, acl)
 		}
 		// if there are pod/namespace selector, then allow packets from/to that address_set or
 		// if the NetworkPolicyPeer is empty, then allow from all sources or to all destinations.
@@ -334,10 +342,10 @@ func (gp *gressPolicy) getNetpolACLDbIDs(ipBlockIdx int, protocol string) *libov
 			// gress rule index
 			libovsdbops.GressIdxKey: strconv.Itoa(gp.idx),
 			// acls are created for every gp.portPolicies which are grouped by protocol:
-			// - for empty policy (no selectors and no ip blocks) - empty ACL
+			// - for empty policy (no selectors and no ip blocks) - empty ACL with idx=emptyIdx (-1)
 			// OR
-			// - all selector-based peers ACL
-			// - for every IPBlock +1 ACL
+			// - all selector-based peers ACL with idx=emptyIdx (-1)
+			// - all ipBlocks combined into a single ACL with idx=ipBlockCombinedIdx (-2)
 			// Therefore unique id for a given gressPolicy is protocol name + IPBlock idx
 			// (protocol will be "None" if no port policy is defined, and empty policy and all
 			// selector-based peers ACLs will have idx=-1)
--- a/go-controller/pkg/ovn/policy_test.go
+++ b/go-controller/pkg/ovn/policy_test.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"runtime"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
@@ -291,10 +292,24 @@ func getGressACLs(gressIdx int, peers []knet.NetworkPolicyPeer, policyType knet.
 		acl.UUID = dbIDs.String() + "-UUID"
 		acls = append(acls, acl)
 	}
-	for i, ipBlock := range ipBlocks {
-		match := fmt.Sprintf("ip4.%s == %s && %s == @%s", ipDir, ipBlock, portDir, pgName)
+	if len(ipBlocks) > 0 {
+		var ipBlockMatches []string
+		for _, ipBlock := range ipBlocks {
+			ipVersion := "ip4"
+			if utilnet.IsIPv6CIDRString(ipBlock) {
+				ipVersion = "ip6"
+			}
+			ipBlockMatches = append(ipBlockMatches, fmt.Sprintf("%s.%s == %s", ipVersion, ipDir, ipBlock))
+		}
+		var match string
+		if len(ipBlockMatches) == 1 {
+			match = ipBlockMatches[0]
+		} else {
+			match = fmt.Sprintf("(%s)", strings.Join(ipBlockMatches, " || "))
+		}
+		match = fmt.Sprintf("%s && %s == @%s", match, portDir, pgName)
 		action := allowAction(params.statelessNetPol)
-		dbIDs := gp.getNetpolACLDbIDs(i, libovsdbutil.UnspecifiedL4Protocol)
+		dbIDs := gp.getNetpolACLDbIDs(ipBlockCombinedIdx, libovsdbutil.UnspecifiedL4Protocol)
 		acl := libovsdbops.BuildACL(
 			libovsdbutil.GetACLName(dbIDs),
 			direction,
@@ -361,6 +376,17 @@ func getPolicyData(params *netpolDataParams) []libovsdbtest.TestData {
 		acls = append(acls, getGressACLs(i, egress.To, knet.PolicyTypeEgress, params)...)
 	}
 
+	pg := getPolicyPortGroup(params, acls)
+
+	data := []libovsdbtest.TestData{}
+	for _, acl := range acls {
+		data = append(data, acl)
+	}
+	data = append(data, pg)
+	return data
+}
+
+func getPolicyPortGroup(params *netpolDataParams, acls []*nbdb.ACL) *nbdb.PortGroup {
 	lsps := []*nbdb.LogicalSwitchPort{}
 	for _, uuid := range params.localPortUUIDs {
 		lsps = append(lsps, &nbdb.LogicalSwitchPort{UUID: uuid})
@@ -375,12 +401,7 @@ func getPolicyData(params *netpolDataParams) []libovsdbtest.TestData {
 	)
 	pg.UUID = pg.Name + "-UUID"
 
-	data := []libovsdbtest.TestData{}
-	for _, acl := range acls {
-		data = append(data, acl)
-	}
-	data = append(data, pg)
-	return data
+	return pg
 }
 
 func newNetpolDataParams(networkPolicy *knet.NetworkPolicy) *netpolDataParams {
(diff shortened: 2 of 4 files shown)


go-controller/pkg/ovn/gress_policy.go
```diff
@@ -284,14 +353,22 @@ func (gp *gressPolicy) buildLocalPodACLs(portGroupName string, aclLogging *libov
...
+				match = gp.getMatchFromIPBlock(lportMatch, l4Match)
+			}
+			if match == "" {
+				continue
 			}
+			aclIDs := gp.getNetpolACLDbIDs(ipBlockCombinedIdx, protocol)
+			acl := libovsdbutil.BuildACLWithDefaultTier(aclIDs, types.DefaultAllowPriority,
+				match, action, aclLogging, gp.aclPipeline)
+			createdACLs = append(createdACLs, acl)
 		}
```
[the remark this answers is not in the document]
pperiyasamy: good catch! done.

go-controller/pkg/ovn/policy_test.go
```diff
@@ -380,6 +394,49 @@ func getPolicyData(params *netpolDataParams) []libovsdbtest.TestData {
...
+			gp := &gressPolicy{
+				policyNamespace: params.networkPolicy.Namespace,
+				policyName:      params.networkPolicy.Name,
+				policyType:      knet.PolicyTypeEgress,
+				idx:             i,
+				controllerName:  DefaultNetworkControllerName,
+			}
+			ipBlockAS := buildIPBlockAddressSets(gp, allowCIDRs, exceptCIDRs)
+			data = append(data, ipBlockAS...)
+		}
```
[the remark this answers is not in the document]
pperiyasamy: done.

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1601,6 +1646,72 @@ func (bnc *BaseNetworkController) shutdownHandlers(np *networkPolicy) {
...
+			// Cleanup allow set on failure
+			if len(allowCIDRs) > 0 {
+				allowDbIDs := gp.getIPBlockAddrSetDbIDs(ipBlockAllow)
+				_ = bnc.addressSetFactory.DestroyAddressSet(allowDbIDs)
+				delete(np.ipBlockAddressSets, allowDbIDs.String())
+			}
+			return fmt.Errorf("failed to create ipBlock except address set: %v", err)
+		}
+		gp.ipv4BlockExcept, gp.ipv6BlockExcept = exceptAS.GetASHashNames()
+		np.ipBlockAddressSets[exceptDbIDs.String()] = true
```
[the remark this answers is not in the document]
pperiyasamy: isn't this network policy configuration problem ?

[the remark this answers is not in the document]
pperiyasamy: good catch, but
> Gradually expanding allowed ranges

but ovnk doesn't support updating network policy, so do we still need to address it ?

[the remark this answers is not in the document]
pperiyasamy: > * **Fix the implementation**: Keep per-ipBlock address sets or pre-compute effective allow CIDRs for each block separately

Keeping per-ipBlock address set bloat OVN with address sets when policies having lots of ipBlocks. hence this option is ruled out. so we are left with 3 options, 

1. pre-compute effective allow CIDRs (this may be a bit complex logic to derive and less intuitive) for each block separately and use it in the Address Set.
2. use inline CIDRs match in ACL itself instead of using Address Set.
3. Document this if this is a very corner case.

will implement it depending on a feedback from @npinaeva and @arkadeepsen .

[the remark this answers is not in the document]
pperiyasamy: cc @igsilya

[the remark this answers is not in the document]
igsilya: I didn't look at the code deep enough, but it feels like the rabbit is correct.  And since ovn-k doesn't modify existing ALCs, but always re-creates them from scratch, there is no real benefit from having an address set.

It may be something to improve in the future (actually modify the ACLs), but for now it may be better to just inline the CIDRs into the ACL match, since there is no possibility for incremental processing anyways.

go-controller/pkg/ovn/gress_policy.go
```diff
@@ -185,17 +187,27 @@ func (gp *gressPolicy) getMatchFromIPBlock(lportMatch, l4Match string) []string
...
-			matchStr = fmt.Sprintf("%s && %s", matchStr, lportMatch)
-		} else {
-			matchStr = fmt.Sprintf("%s && %s && %s", matchStr, l4Match, lportMatch)
-		}
-		matchStrings = append(matchStrings, matchStr)
+		ipBlockMatches = append(ipBlockMatches, matchStr)
+	}
+
+	if len(ipBlockMatches) == 0 {
+		return ""
```
npinaeva: We only call this function if `len(gp.ipBlocks) > 0`, so I think this will should never happen

go-controller/pkg/ovn/gress_policy.go
```diff
@@ -185,17 +187,27 @@ func (gp *gressPolicy) getMatchFromIPBlock(lportMatch, l4Match string) []string
...
 	}
-	return matchStrings
+
+	var l3Match string
+	if len(ipBlockMatches) == 1 {
+		l3Match = ipBlockMatches[0]
+	} else {
+		l3Match = fmt.Sprintf("(%s)", strings.Join(ipBlockMatches, " || "))
```
npinaeva, line 202: we need to check the max length for an ACL match and make sure we are not exceeding it

go-controller/pkg/ovn/gress_policy.go
```diff
@@ -22,6 +22,8 @@ import (
 const (
 	// emptyIdx is used to create ACL for gressPolicy that doesn't have ipBlocks
 	emptyIdx = -1
+	// ipBlockCombinedIdx is used to create ACL for gressPolicy that have ipBlocks
```
npinaeva: I would add some more details to that, saying that we used to create an ACL per ipBlock, but then switched to creating only 1 ACL for all ipBlocks, hence this special value is used by the new ACLs. We also need an update in the `getNetpolACLDbIDs` comment explaining how `IpBlockIndexKey` works and the same on the `ACLNetworkPolicy` in the `db_object_types.go`

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1067,13 +1067,20 @@ func (bnc *BaseNetworkController) createNetworkPolicy(policy *knet.NetworkPolicy
...
+		ops, err = libovsdbops.CreateOrUpdateACLsOps(bnc.nbClient, ops, bnc.GetSamplingConfig(), createdACLs...)
 		if err != nil {
 			return fmt.Errorf("failed to create ACL ops: %v", err)
 		}
+		// Remove legacy per-ipBlock ACLs from the port group
+		if len(deletedACLs) > 0 {
+			ops, err = libovsdbops.DeleteACLsFromPortGroupOps(bnc.nbClient, ops, np.portGroupName, deletedACLs...)
+			if err != nil {
+				return fmt.Errorf("failed to create ops to remove legacy ACLs from port group: %v", err)
+			}
```
npinaeva: doing that on every update is not the most optimal way. It would be better to run a single predicate search in the `syncNetworkPolicies` to find if there are any ACLs with `IpBlockIndexKey` > 0 and write down the netpolNames of those ACLs to some `staleIPBlockPolicies` set, then in `addNetworkPolicy` check if this policy is in the set, and cleanup stale ACLs after we have commited the updated ACLs. This way we avoid doing any extra db ops after the cleanup is done once.

go-controller/pkg/ovn/gress_policy.go
igsilya: FWIW, there isn't a defined limit.  As long as it fits into `size_t`, it should be fine.  And we'll have much more issues if the ACL size even remotely approaches `SIZE_MAX`. :)

pperiyasamy: ok, i currently set it to 65535 as max length for ACL match, is that ok or not introduce max limit, wdyt ?

go-controller/pkg/ovn/gress_policy.go
pperiyasamy: removed.

go-controller/pkg/ovn/gress_policy.go
pperiyasamy: added some info about Idx.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: @npinaeva good idea! added it, it seems like stale ACLs are removed automatically, noticed this behavior while testing the change with "reconciles existing networkPolicies with equivalent rules" unit test.
so this piece of [code](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5959/changes/555795667083c5d346403cabb09516d9ca9687b7#diff-0b82ad568a45217e5458d7e3153a8ad9550d662bc639137bb265cefb0d440b45R1107-R1116) unnecessarily deleting ACLs for the policy. how does that happen ? are ACLs removed since port group is not referring to old ACLs any more ? or does ovnk delete it somewhere else ?

go-controller/pkg/ovn/gress_policy.go
```diff
@@ -285,12 +303,26 @@ func (gp *gressPolicy) buildLocalPodACLs(portGroupName string, aclLogging *libov
...
+			if queryLegacyIPBlockACLs {
+				// we created one ACL per ipBlock with their indexes
+				// Now we use ipBlockCombinedIdx (-2) for all ipBlocks
+				// so legacy ACLs should be removed
+				for i := range gp.ipBlocks {
+					legacyACLIDs := gp.getNetpolACLDbIDs(i, protocol)
+					legacyACL := libovsdbutil.BuildACLWithDefaultTier(legacyACLIDs, types.DefaultAllowPriority,
+						"", action, aclLogging, gp.aclPipeline)
+					skippedACLs = append(skippedACLs, legacyACL)
+				}
```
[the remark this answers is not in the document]
pperiyasamy: There is no support of updating network policy , so this scenario doesn't occur and skippedACLs contains everything what needs to be deleted.

go-controller/pkg/ovn/gress_policy.go
igsilya:
```python
>>> len("(192.168.0.0/24)")
16
>>> 65536 / 16
4096.0
```
4K ipBlocks sounds like a limit that someone will actually hit.  I'd suggest to not add a limit or at least make it significantly larger.

go-controller/pkg/ovn/base_network_controller_policy.go
npinaeva: ah right, even better. So ACLs are removed by the db when they are not referenced, so after the netpol is updated on restart, pg.ALCS will be updated and we don't need any extra cleanup at all, just maybe one unit test checking transition from the old format to the new

go-controller/pkg/ovn/gress_policy.go
npinaeva: there is actually no limit on how many ipBlocks can be added from the kube-api side, and we technically didn't have any limit in ovn-k either, so introducing it would be an incompatible change, and considering no explicit limit on the ACL match exists in OVN I think we can just not have it and hope for the best

pperiyasamy: agree both of you. removed acl match limit now.

go-controller/pkg/ovn/base_network_controller_policy.go
pperiyasamy: right, i too confirmed it with @dceara, removed legacy ACLs cleanup logic now. checking ACL transition with a new test.

go-controller/pkg/ovn/policy_test.go
```diff
@@ -956,6 +977,149 @@ var _ = ginkgo.Describe("OVN NetworkPolicy Operations", func() {
...
+					"",
+					false,
+					legacyACLID.GetExternalIDs(),
+					nil,
+					types.DefaultACLTier,
+				)
+				legacyACL.UUID = legacyACLID.String() + "-UUID"
+				initialData = append(initialData, legacyACL)
+
+				legacyACLID2 := secondgp2.getNetpolACLDbIDs(0, libovsdbutil.UnspecifiedL4Protocol)
```
npinaeva, line 1086: shouldn't this one have an index=1?

pperiyasamy:
@coderabbitai 

> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > go-controller/pkg/ovn/gress_policy.go (1)> `283-333`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Verify that legacy per-IPBlock ACL cleanup covers indices beyond the current `gp.ipBlocks` count.**
> > > The concern is valid. When a policy is updated (after upgrade from old code) to reduce its ipBlock count, legacy ACLs with higher indices (e.g., previously indices 0–4, now policy has only 2 ipBlocks) may not be cleaned up. The `buildLocalPodACLs` function returns `skippedACLs` only for empty address set cases, not for legacy per-ipBlock indices. There is no predicate-based deletion that removes all ACLs with `IpBlockIndexKey >= 0` for the given policy/gress/protocol combination, leaving stale allow ACLs in OVN.

The explicit ACL cleanup is not needed because OVN itself garbage collect it when not referred by a port group.

> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@go-controller/pkg/ovn/gress_policy.go` around lines 283 - 333,
> > > buildLocalPodACLs currently only creates/skips ACLs for the current gp.ipBlocks
> > > and won't mark legacy per‑IPBlock ACLs (with higher indices than the current
> > > count) for deletion; update buildLocalPodACLs to detect legacy ipBlock indices
> > > and append corresponding ACL placeholders to skippedACLs so they will be
> > > removed: determine the range of legacy indices (either from a stored
> > > gp.prevIPBlockCount field you add to gressPolicy or by scanning existing ACL
> > > external IDs), for each legacy index call
> > > gp.getNetpolACLDbIDs(ipBlockCombinedIdx, protocol) and build a matching ACL
> > > (using gp.getMatchFromIPBlock or a minimal match) and append it to skippedACLs
> > > rather than createdACLs; reference functions/fields: buildLocalPodACLs,
> > > gp.ipBlocks, gp.getNetpolACLDbIDs, gp.getMatchFromIPBlock,
> > > skippedACLs/createdACLs so cleanup covers indices beyond the current gp.ipBlocks
> > > count.
> > > ```
> 
> 🧹 Nitpick comments (2)
> > go-controller/pkg/ovn/gress_policy_test.go (1)> `72-109`: **Duplicate test cases and missing `UnspecifiedL4Match` path.**
> > > Two independent concerns:
> > > 
> > > 1. **Duplicate test cases**: "IPv4 only with except" (lines 72–82) and "IPv4 with IPv4 except" (lines 98–109) have identical `ipBlocks`, `lportMatch`, `l4Match`, and `expected` values. One of them is redundant.
> > > 2. **Missing branch coverage**: Every test case uses `l4Match: "input"`, exercising only the `return fmt.Sprintf("%s && %s && %s", l3Match, l4Match, lportMatch)` path. The `l4Match == libovsdbutil.UnspecifiedL4Match` branch (`return fmt.Sprintf("%s && %s", l3Match, lportMatch)`) has no test coverage here.
> > > 
> > > 
> > > ♻️ Suggested additions / removals
> > > ```diff
> > > -		{
> > > -			desc: "IPv4 with IPv4 except",
> > > -			ipBlocks: []*knet.IPBlock{
> > > -				{
> > > -					CIDR:   "0.0.0.0/0",
> > > -					Except: []string{"10.1.0.0/16"},
> > > -				},
> > > -			},
> > > -			lportMatch: "fake",
> > > -			l4Match:    "input",
> > > -			expected:   "(ip4.src == 0.0.0.0/0 && ip4.src != {10.1.0.0/16}) && input && fake",
> > > -		},
> > > +		{
> > > +			desc: "IPv4 only no except, unspecified l4Match",
> > > +			ipBlocks: []*knet.IPBlock{
> > > +				{
> > > +					CIDR: "0.0.0.0/0",
> > > +				},
> > > +			},
> > > +			lportMatch: "fake",
> > > +			l4Match:    libovsdbutil.UnspecifiedL4Match,
> > > +			expected:   "ip4.src == 0.0.0.0/0 && fake",
> > > +		},
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 

this is not specific to the changes this PR is addressing, so i prefer not to add this test under this PR.

> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@go-controller/pkg/ovn/gress_policy_test.go` around lines 72 - 109, Remove the
> > > redundant test case "IPv4 with IPv4 except" (it duplicates "IPv4 only with
> > > except") and add a new test that sets l4Match to libovsdbutil.UnspecifiedL4Match
> > > (instead of "input") to exercise the UnspecifiedL4Match branch; ensure the new
> > > test uses a distinct desc (e.g., "IPv4 only with except and unspecified l4"),
> > > keeps the same ipBlocks and lportMatch, and expects a string built with
> > > fmt.Sprintf("%s && %s", l3Match, lportMatch) (i.e., no l4 component) so the code
> > > path where l4Match == libovsdbutil.UnspecifiedL4Match is covered.
> > > ```
> > 
> > 
> > 
> > go-controller/pkg/ovn/gress_policy.go (1)> `175-209`: **Operational concern: no guard for very large combined ACL match strings.**
> > > When a `NetworkPolicy` rule contains many ipBlocks, the combined match string `(ip4.src == X || ip4.src == Y || ...)` can grow unboundedly. OVN has internal limits on ACL match string length. Exceeding those limits would cause OVN NB DB transaction failures that are hard to diagnose.
> > > Consider documenting a known-safe upper bound, or adding a check/warning when `len(gp.ipBlocks)` exceeds a reasonable threshold.
> > > Based on learnings: "combining all IPBlock CIDRs into a single ACL provides optimization benefits… However, this approach needs to consider OVN ACL match string length limitations."
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@go-controller/pkg/ovn/gress_policy.go` around lines 175 - 209,
> > > getMatchFromIPBlock can build arbitrarily large ACL match strings from
> > > gp.ipBlocks; add a guard at the top of getMatchFromIPBlock that estimates the
> > > combined length (sum of CIDR lengths plus separators) or simply checks
> > > len(gp.ipBlocks) against a defined constant (e.g. maxIPBlocks or
> > > maxMatchLength), and if exceeded emit a warning (include gp or context info) and
> > > return a safe short match (e.g. fall back to lportMatch only) instead of
> > > building the huge string; modify getMatchFromIPBlock to define the threshold
> > > constant, perform the check before constructing ipBlockMatches, log the
> > > condition (with the offending count/estimated length), and return the short
> > > fallback to avoid creating oversized ACL match strings.
> > > ```
> 

There is no any specific max limit associated with ACL match string, so not restricting with limit on purpose.

> 🤖 Prompt for all review comments with AI agents

pperiyasamy:
> **Question:** Does `buildLocalPodACLs` explicitly remove the legacy per-IPBlock ACLs from `Port_Group.acls`, or does the calling code handle this cleanup via predicates/operations? If not handled, these stale ACLs could allow unintended traffic.

@coderabbitai yes, Port_Group is rebuilt during the upgrade only referring to new ACLs.

PR #5964, opened 2026-02-12
Reduce obj_retry.go log spam

kyrtapz:
These are low hanging fruits that bring no value but contribute significantly to the log output:
- Only log "Retry successful" when there were actual retries (failedAttempts > 0), not on every first-attempt success.
- Move terminal state pod detection messages to V(5).
- Remove duplicate "Update event received" V(5) log that fired before the equality check. There is already a log line for actual updates.

## Summary by CodeRabbit

* **Chores**
  * Reduced logging noise by removing a verbose message about equal updates and trimming an extra trailing newline in update logs.
  * Lowered terminal-state notifications to less prominent verbosity to reduce alert fatigue.
  * Retry success messages now only emit when there was at least one prior failed attempt.

### DIFF
--- a/go-controller/pkg/retry/obj_retry.go
+++ b/go-controller/pkg/retry/obj_retry.go
@@ -415,7 +415,9 @@ func (r *RetryFramework) resourceRetry(objKey string, now time.Time) {
 			}
 		}
 
-		klog.Infof("Retry successful for %s %s after %d failed attempt(s)", r.ResourceHandler.ObjType, objKey, entry.failedAttempts)
+		if entry.failedAttempts > 0 {
+			klog.Infof("Retry successful for %s %s after %d failed attempt(s)", r.ResourceHandler.ObjType, objKey, entry.failedAttempts)
+		}
 		if initObj != nil {
 			r.ResourceHandler.RecordSuccessEvent(initObj)
 		}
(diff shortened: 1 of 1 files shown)


l8huang:
FYI: For jobs failed with error:
```
ERROR: failed to load image: command "docker exec --privileged -i ovn-worker ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1

Command Output: ctr: content digest sha256:0006f95da849809729f05b2e1c63734df8567b18f1ba5a4a6fcb41bbb9201926: not found
```

The error should be fixed by https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5968

go-controller/pkg/retry/obj_retry.go
```diff
@@ -415,7 +415,9 @@ func (r *RetryFramework) resourceRetry(objKey string, now time.Time) {
 			}
 		}
 
-		klog.Infof("Retry successful for %s %s after %d failed attempt(s)", r.ResourceHandler.ObjType, objKey, entry.failedAttempts)
+		if entry.failedAttempts > 0 {
+			klog.Infof("Retry successful for %s %s after %d failed attempt(s)", r.ResourceHandler.ObjType, objKey, entry.failedAttempts)
+		}
```
ricky-rav, line 420: This got me curious about when we see this log being printed, since in the expected execution flow with a failed add/update/delete event and a subsequent successful retry `entry.failedAttempts` is 1.
 
So whenever we "manually" readd an object to the retry queue, we call `AddRetryObjWithAddNoBackoff`, which in turn calls `initRetryObjWithAddBackoff`, where indeed `entry.failedAttempts = 0`.

PR #5591, opened 2025-09-24
docs: Improve DPU Feature Documentation

dave-tucker:
## 📑 Description

When attempting to read how the DPU feature worked, I found that our docs point to a Google document that is not public.
I have reverse engineered the implementation to try and describe the design. Mistakes may have been made.

This attemtps to improve the DPU documentation to describe in detail how this feature works. It explains the 2 cluster requirement and includes a diagram on how pod networking is set up in this design. It also includes assumptions that ovn-k8s makes about the environment.

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## Summary by CodeRabbit

* **Documentation**
  * Expanded DPU support guide into a comprehensive multi-section reference covering architecture, user stories, deployment models (off-host vs on-host), OVN-Kubernetes flows, health/lease mechanics, pod networking workflow, gateway configuration options, troubleshooting, best practices, future items, and known limitations.
  * Consolidated and removed separate derive-from-mgmt-port and gateway-interface docs; their guidance is folded into the expanded DPU guide.

PR #5976, opened 2026-02-13
node: improve management port resilience after interface changes

tsorya:
node: improve management port resilience in DPU host mode
Identify the management port VF by its PCI device ID instead of
interface name. At startup, the device ID is resolved from the current
netdev name and stored in the managementPortNetdev struct. All
subsequent lookups (create, reconcile) use the PCI device ID to find
the VF, making the management port resilient to interface renames.

When reconciliation fails, the code attempts to recreate the management
port by resolving the VF from its device ID. If the PCI device itself
is no longer present — due to a DPU reboot, an administrator destroying
VFs, or a firmware settings change that causes PCI re-enumeration — the
process terminates so the container can restart and the device plugin
can re-allocate the correct VF. Transient errors (route, iptables,
config failures) are returned to the retry loop instead.

Assisted-by: Cursor
Signed-off-by: Igal Tsoiref <itsoiref@redhat.com>

### DIFF
--- a/go-controller/pkg/node/managementport/port_dpu_linux.go
+++ b/go-controller/pkg/node/managementport/port_dpu_linux.go
@@ -4,10 +4,12 @@
 package managementport
 
 import (
+	"errors"
 	"fmt"
 	"net"
 	"time"
 
+	nadapi "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/vishvananda/netlink"
 
 	"k8s.io/klog/v2"
@@ -17,6 +19,8 @@ import (
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
 )
+
+var errMgmtPortDeviceNotFound = errors.New("management port PCI device not found")
 
 type managementPortRepresentor struct {
 	cfg        *managementPortConfig
@@ -104,25 +108,48 @@ func (mp *managementPortRepresentor) doReconcile() error {
 }
 
 type managementPortNetdev struct {
-	ifName        string
-	netdevDevName string
-	cfg           *managementPortConfig
-	routeManager  *routemanager.Controller
+	ifName       string
+	deviceID     string
+	cfg          *managementPortConfig
+	routeManager *routemanager.Controller
 }
 
-// newManagementPortNetdev creates a new managementPortNetdev
-func newManagementPortNetdev(netdevDevName string, cfg *managementPortConfig, routeManager *routemanager.Controller) *managementPortNetdev {
+// newManagementPortNetdev creates a new managementPortNetdev.
+// deviceID is the PCI device ID (e.g., "0000:03:00.2") used to identify the VF.
+func newManagementPortNetdev(deviceID string, cfg *managementPortConfig, routeManager *routemanager.Controller) *managementPortNetdev {
 	return &managementPortNetdev{
-		ifName:        types.K8sMgmtIntfName,
-		netdevDevName: netdevDevName,
-		cfg:           cfg,
-		routeManager:  routeManager,
+		ifName:       types.K8sMgmtIntfName,
+		deviceID:     deviceID,
+		cfg:          cfg,
+		routeManager: routeManager,
 	}
 }
 
+// findNetdevByDeviceID resolves the current interface name from the PCI device ID.
+func (mp *managementPortNetdev) findNetdevByDeviceID() (netlink.Link, error) {
+	if mp.deviceID == "" {
+		return nil, fmt.Errorf("no device ID available")
+	}
+
+	netdevName, err := util.GetNetdevNameFromDeviceId(mp.deviceID, nadapi.DeviceInfo{})
+	if err != nil {
+		return nil, fmt.Errorf("%w: device ID %s lookup failed: %v", errMgmtPortDeviceNotFound, mp.deviceID, err)
+	}
+	if netdevName == "" {
+		return nil, fmt.Errorf("%w: device ID %s resolved to empty netdev name", errMgmtPortDeviceNotFound, mp.deviceID)
+	}
+
+	link, err := util.GetNetLinkOps().LinkByName(netdevName)
+	if err != nil {
+		return nil, fmt.Errorf("device ID %s resolved to %s but LinkByName failed: %w", mp.deviceID, netdevName, err)
+	}
+
+	return link, nil
+}
+
 func (mp *managementPortNetdev) create() error {
-	klog.V(5).Infof("Lookup netdevice link and existing management port using '%v'", mp.netdevDevName)
-	link, err := util.GetNetLinkOps().LinkByName(mp.netdevDevName)
+	klog.Infof("Management port netdev create: deviceID=%s, ifName=%s", mp.deviceID, mp.ifName)
+	link, err := mp.findNetdevByDeviceID()
 	if err != nil {
 		return err
 	}
@@ -136,22 +163,20 @@ func (mp *managementPortNetdev) create() error {
 
 	// configure management port: name, mac, MTU, iptables
 	// mac addr, derived from the first entry in host subnets using the .2 address as mac with a fixed prefix.
-	klog.V(5).Infof("Setup netdevice management port: %s", link.Attrs().Name)
+	klog.V(5).Infof("Setup netdevice management port: %s (deviceID: %s)", link.Attrs().Name, mp.deviceID)
 	mgmtPortMac := util.IPAddrToHWAddr(util.GetNodeManagementIfAddr(mp.cfg.hostSubnets[0]).IP)
 	err = bringupManagementPortLink(types.DefaultNetworkName, link, &mgmtPortMac, mp.ifName, config.Default.MTU)
 	if err != nil {
 		return err
 	}
 
-	if mp.netdevDevName != mp.ifName && config.OvnKubeNode.Mode != types.NodeModeDPUHost {
-		// Store original interface name for later use
+	if link.Attrs().Name != mp.ifName && config.OvnKubeNode.Mode != types.NodeModeDPUHost {
 		if _, stderr, err := util.RunOVSVsctl("set", "Open_vSwitch", ".",
-			"external-ids:ovn-orig-mgmt-port-netdev-name="+mp.netdevDevName); err != nil {
+			"external-ids:ovn-orig-mgmt-port-netdev-name="+link.Attrs().Name); err != nil {
 			return fmt.Errorf("failed to store original mgmt port interface name: %s", stderr)
 		}
 	}
 
-	// Setup Iptable and routes
 	err = createPlatformManagementPort(mp.ifName, mp.cfg, mp.routeManager)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


go-controller/pkg/node/managementport/port_dpu_linux.go
```diff
@@ -108,23 +109,51 @@ type managementPortNetdev struct {
 	netdevDevName string
 	cfg           *managementPortConfig
 	routeManager  *routemanager.Controller
+	// deviceID is the PCI device ID for fallback lookup when interface name changes
+	deviceID string
```
wizhaoredhat: When the DPU reboots, the ovn-kube node container would also be gone. How does the code retain state to keep the deviceID when the container goes away?

Shouldn't management ports not be reference by name at all? Shouldn't it come from the device plugin (which would pass the device ID)?

go-controller/pkg/node/managementport/port_dpu_linux_test.go
```diff
@@ -298,23 +312,30 @@ var _ = Describe("Mananagement port DPU tests", func() {
...
+				"createPlatformManagementPort error"))
 
 			err = mgmtPortDpuHost.create()
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).To(ContainSubstring(
 				"createPlatformManagementPort error"))
+			Expect(execMock.CalledMatchesExpected()).To(BeTrue(), execMock.ErrorDesc)
 		})
 	})
 })
```
aserdean, line 499: fallback mechanism and doesn't test the actual fallback. The reconcile-to-create path has 0% coverage per coveralls.

go-controller/pkg/node/managementport/port.go
```diff
@@ -84,7 +84,16 @@ func start(mp managementPort, stopChan <-chan struct{}) (func(), error) {
 							return true
 						}
 					},
-					mp.doReconcile,
+					func() error {
```
aserdean: It's idempotent so it won't break things, but it's unnecessary OVS churn for what should be a netdev-specific fix. Scoping it to managementPortNetdev would be cleaner.

go-controller/pkg/node/managementport/port_dpu_linux.go
tsorya: When dpu reboots the ovn-kube  in dpu-host mode will continue running as it runs on the host.
Though i agree that we should provide device and not link in the first place

go-controller/pkg/node/managementport/port.go
tsorya: This will happened only on error that should not occur and if did why not to try to fix it by recreating things for all modes?

docs/features/hardware-offload/dpu-support.md
```diff
@@ -45,6 +45,38 @@ These aforementioned parts are expected to be deployed also on two different Kub
...
+At startup, the VF is looked up by its current interface name — which may be the
+original kernel name (e.g., `enp3s0f0v0`) or `ovn-k8s-mp0` if a previous instance
+already renamed it. It is then configured as the management port.
+
+A periodic reconciliation loop monitors the management port and re-applies its
+configuration (routes, addresses, nftables rules). If reconciliation fails — for
+example because the underlying interface has disappeared — the code attempts to
+recreate the management port from scratch.
+
+#### DPU reboot recovery
```
aserdean: not tied to a crash/reboot, this could also happen in a misconfig case, i.e. if a admin destroys vfs.

this needs to be reworded about ownership, maybe we should also have a table of what ovnkube in host mode requires on start.

go-controller/pkg/node/managementport/port_dpu_linux.go
```diff
@@ -164,5 +164,19 @@ func (mp *managementPortNetdev) reconcilePeriod() time.Duration {
 }
 
 func (mp *managementPortNetdev) doReconcile() error {
-	return createPlatformManagementPort(mp.ifName, mp.cfg, mp.routeManager)
+	if err := createPlatformManagementPort(mp.ifName, mp.cfg, mp.routeManager); err != nil {
+		klog.Warningf("Failed to reconcile management port netdev, attempting to recreate: %v", err)
+		if err := mp.create(); err != nil {
+			// The management port VF has disappeared and cannot be found
```
aserdean: the styling of this is very weird limited to 80 char, IMO it should be at least as long as
```
klog.Fatalf("Failed to recreate management port netdev, terminating so device plugin can re-allocate the correct VF on restart: %v", err)
```

go-controller/pkg/node/managementport/port_dpu_linux.go
```diff
@@ -164,5 +164,19 @@ func (mp *managementPortNetdev) reconcilePeriod() time.Duration {
...
+			// by its known name. For example, a DPU reboot while the host
```
aserdean: not sure if such a huge comment is needed in the code. in theory this should be short just stating why we we fail which you also hint in the log message.

In general a good motivation should be included in the commit message, assumptions should be written in the doc and a good log message in the code.

docs/features/hardware-offload/dpu-support.md
```diff
@@ -45,6 +45,41 @@ These aforementioned parts are expected to be deployed also on two different Kub
 
 For detailed configuration of gateway interfaces in DPU host mode, see [DPU Gateway Interface Configuration](dpu-gateway-interface.md).
 
+### Management Port in DPU Host Mode
+
+#### Startup Prerequisites
```
aserdean: not the right place for this. please be also aware about https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6001.

go-controller/pkg/node/managementport/port_linux.go
```diff
@@ -90,7 +90,12 @@ func NewManagementPortController(
 		c.ports[ovsPort] = newManagementPortOVS(cfg, routeManager)
 	}
 	if hasNetdev {
-		c.ports[netdevPort] = newManagementPortNetdev(netdevDevName, cfg, routeManager)
+		deviceID, err := util.GetDeviceIDFromNetdevice(netdevDevName)
```
aserdean: Prefer the node annotation k8s.ovn.org/node-mgmt-port for the management port device ID when it’s set. The controller already has the node, and the annotation is the stored source of truth that survives renames and reboots. Resolve the device ID from the netdev name only when the annotation is missing (e.g. first start).

docs/features/hardware-offload/dpu-support.md
tsorya: Should i remove it ? Can we add more something from here to the doc you are adding?

go-controller/pkg/node/managementport/port_linux.go
tsorya: @aserdean i can add annotation check but with a fallback to `util.GetDeviceIDFromNetdevice(netdevDevName)`

From what i understand netdev can be used in full mode too, like this 
`--ovnkube-node-mode=full
--ovnkube-node-mgmt-port-netdev=enp3s0f0v0`

in this case annotation will not be added to the node object and we will fail. Do you prefer it this way?

aserdean: I think the annotation should have prio above the argument similar like the initialization.

aserdean:
one last ask can you please add a test for:

- Assert the sentinel with errors.Is(err, errMgmtPortDeviceNotFound) where the test claims “errMgmtPortDeviceNotFound”.

- Add negative path for PCI resolution failure in port_linux_test.go.

- Optional: add tests for reconcile→create (recreate success, recreate non-device error, recreate device-loss/fatal).

go-controller/pkg/node/managementport/port_dpu_linux.go
```diff
@@ -104,25 +108,45 @@ func (mp *managementPortRepresentor) doReconcile() error {
...
+	}
+}
+
+// findNetdevByDeviceID resolves the current interface name from the PCI device ID.
+func (mp *managementPortNetdev) findNetdevByDeviceID() (netlink.Link, error) {
+	if mp.deviceID == "" {
+		return nil, fmt.Errorf("no device ID available")
+	}
+
+	netdevName, err := util.GetNetdevNameFromDeviceId(mp.deviceID, nadapi.DeviceInfo{})
```
aserdean, line 134: nit: looking at GetNetdevNameFromDeviceId, there can be a result which is not an error and "", it might be worth to check before using `GetNetLinkOps().LinkByName`

aserdean:
@trozet do you mind triggering the DPU CI for this please and take a look the changes lgtm, just one minor nit that I see

docs/features/hardware-offload/derive-from-mgmt-port.md
```diff
@@ -140,6 +140,42 @@ The feature integrates with existing:
...
 - Network configuration
 
+## Management Port VF Identification and Recovery
+
+### VF Identification
+
+The management port VF is identified by its PCI device ID, not by interface name.
+At startup, the PCI device ID is preferred from the node annotation
+(`k8s.ovn.org/node-mgmt-port`) when available. If the annotation is missing
+(e.g., first start), the device ID is resolved from the netdev name via sysfs.
```
trozet: this isn't true if using ovnkube-node-mgmt-port-dp-resource-name right? It's derived from the SRIOV resource pool and assumed to be the first PCI address there?

docs/features/hardware-offload/derive-from-mgmt-port.md
```diff
@@ -140,6 +140,42 @@ The feature integrates with existing:
...
+All subsequent operations (reconciliation, recreation) locate the VF by its PCI
+device ID, which is stable across interface renames.
+
+### Reconciliation and Recovery
+
+A periodic reconciliation loop monitors the management port and re-applies its
```
trozet: so these sections go into detail about how management port works with offload, but the purpose of this specific document is more related to # From PCI Address Gateway Interface Feature

I think we are overloading this document with more about how management port works with offload. I think we need to probably move in a direction similar to: 

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5591

trozet: note if you dont want to wait on 5591 you we could merge this PR as code change only, then you can pinky promise to update the docs later ;)

go-controller/pkg/node/managementport/port_dpu_linux.go
```diff
@@ -134,24 +161,20 @@ func (mp *managementPortNetdev) create() error {
 		}
 	}
 
-	// configure management port: name, mac, MTU, iptables
-	// mac addr, derived from the first entry in host subnets using the .2 address as mac with a fixed prefix.
```
trozet, line 138: why is this comment removed?

go-controller/pkg/node/managementport/port_linux.go
```diff
@@ -90,7 +90,22 @@ func NewManagementPortController(
...
+		var deviceID string
+		if cfgs, err := util.ParseNodeManagementPortAnnotation(node); err == nil {
+			if devCfg, ok := cfgs[types.DefaultNetworkName]; ok && devCfg.DeviceId != "" {
+				deviceID = devCfg.DeviceId
+				klog.Infof("Management port PCI device ID: %s (from node annotation)", deviceID)
+			}
+		}
+		if deviceID == "" {
+			var err error
+			deviceID, err = util.GetDeviceIDFromNetdevice(netdevDevName)
```
trozet, line 102: so it looks like this netdevDevName is derived from this on the dpu-host:

```
func getMgmtPortAndRepNameModeDPUHost() (string, string, error) {
	netdevName, err := getManagementPortNetDev(config.OvnKubeNode.MgmtPortNetdev)
	if err != nil {
		return "", "", err
	}
	return netdevName, "", nil
}
```
I'm wondering what happens when MgmtPortDPResourceName is provided as config and not the config.OvnKubeNode.MgmtPortNetdev. I'm guessing the DPResoureName path ends up writing the value of config.OvnKubeNode.MgmtPortNetdev when it finds a suitable VF? @aserdean is that correct?

Looks like it does that in initDefaultNodeNetworkController:

```
		config.OvnKubeNode.MgmtPortNetdev = netdevice
		klog.V(5).Infof("Using MgmtPortNetdev (Netdev %s) passed via resource %s",
			config.OvnKubeNode.MgmtPortNetdev, ncm.deviceAllocator.ResourceName())
```

docs/features/hardware-offload/derive-from-mgmt-port.md
tsorya: Yeah, ill remove the doc from here and will update it after Dave's PR will be pushed

go-controller/pkg/node/managementport/port_dpu_linux.go
tsorya: Mistake, missed it somehow.

go-controller/pkg/node/managementport/port_linux.go
aserdean: > I'm wondering what happens when MgmtPortDPResourceName is provided as config and not the config.OvnKubeNode.MgmtPortNetdev. I'm guessing the DPResoureName path ends up writing the value of config.OvnKubeNode.MgmtPortNetdev when it finds a suitable VF? @aserdean is that correct?

correct, at least that would be the expectation

aserdean:
LGTM, doing one more round of validation on my side

aserdean:
LGTM, 

Tested-by: Alin Serdean <aserdean@nvidia.com>

PR #5980, opened 2026-02-15
Perf/Scale: Eliminate always add pod IP to namespace address set

trozet:
During perf/scale testing it was seen that adding every pod IP to the namespace address sets costs a lot of memory churn in ovnk and a lot of OVN NB ovsdb txns:

(*ovnAddressSets).AddAddressesReturnOps in /build/sdn/ovn-kubernetes/go-controller/pkg/ovn/address_set/address_set.go
      19MB   621.96GB (flat, cum) 19.25% of Total

This PR makes the following high level changes:

1. Removes EgressQoS dependency on namespace AS. It now maintains its own address set when needed. (Should we just deprecate EgressQos?)
2. Removes NetworkQoS dependency on namespace address set. It now maintains its own.
3. Reduces memory allocation during hasAddressSet

Further work to only create ns address set will be handled after address set management is convert to a level-driven controller: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5999

## Summary by CodeRabbit

* **Refactor**
  * Added per-namespace reference tracking and namespace-aware gating so address sets are populated and maintained only when needed.
  * Unified address-set handling across QoS, Egress, and policy flows; simplified pod-selection behavior.

* **Bug Fixes**
  * Reconciles and clears namespace address sets during policy sync and lifecycle changes to prevent stale or incorrect memberships.

* **Tests**
  * Updated tests and added benchmarks to reflect empty/reduced per-pod IP population and new reconciliation behavior.

### DIFF
--- a/go-controller/pkg/ovn/address_set/address_set_benchmark_test.go
+++ b/go-controller/pkg/ovn/address_set/address_set_benchmark_test.go
@@ -1,0 +1,94 @@
+package addressset
+
+import (
+	"fmt"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func hasAllAddressesWithSet(existing, wanted []string) bool {
+	if len(existing) == 0 {
+		return false
+	}
+	return sets.NewString(existing...).HasAll(wanted...)
+}
+
+func hasAllAddressesWithScan(existing, wanted []string) bool {
+	if len(existing) == 0 {
+		return false
+	}
+	for _, want := range wanted {
+		found := false
+		for _, have := range existing {
+			if have == want {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	return true
+}
+
+func buildBenchmarkIPs(size int) []string {
+	ips := make([]string, size)
+	for i := 0; i < size; i++ {
+		ips[i] = fmt.Sprintf("10.%d.%d.%d", (i>>16)&0xff, (i>>8)&0xff, i&0xff)
+	}
+	return ips
+}
+
+func buildWantedIPs(existing []string, wantedCount int, includeMissing bool) []string {
+	wanted := make([]string, 0, wantedCount)
+	for i := 0; i < wantedCount; i++ {
+		// Pick addresses from the tail to avoid accidental early hits.
+		wanted = append(wanted, existing[len(existing)-1-i])
+	}
+	if includeMissing {
+		wanted[wantedCount-1] = "192.0.2.250"
+	}
+	return wanted
+}
+
+func benchmarkHasAllAddresses(
+	b *testing.B,
+	implName string,
+	fn func(existing, wanted []string) bool,
+	existingSize, wantedCount int,
+	includeMissing bool,
+) {
+	b.Helper()
+	name := fmt.Sprintf("%s/existing-%d/wanted-%d/missing-%t", implName, existingSize, wantedCount, includeMissing)
+	b.Run(name, func(b *testing.B) {
+		existing := buildBenchmarkIPs(existingSize)
+		wanted := buildWantedIPs(existing, wantedCount, includeMissing)
+		expected := !includeMissing
+		if got := fn(existing, wanted); got != expected {
+			b.Fatalf("unexpected result: got=%t expected=%t", got, expected)
+		}
+
+		b.ReportAllocs()
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			_ = fn(existing, wanted)
+		}
+	})
+}
+
+func BenchmarkHasAllAddressesSetVsScan(b *testing.B) {
+	existingSizes := []int{500, 50000}
+	wantedCounts := []int{1, 10}
+	missingCases := []bool{false, true}
+
+	for _, existingSize := range existingSizes {
+		for _, wantedCount := range wantedCounts {
+			for _, includeMissing := range missingCases {
+				benchmarkHasAllAddresses(b, "set", hasAllAddressesWithSet, existingSize, wantedCount, includeMissing)
+				benchmarkHasAllAddresses(b, "scan", hasAllAddressesWithScan, existingSize, wantedCount, includeMissing)
+			}
+		}
+	}
+}
--- a/go-controller/pkg/ovn/address_set/address_set.go
+++ b/go-controller/pkg/ovn/address_set/address_set.go
@@ -598,7 +598,20 @@ func (as *ovnAddressSet) hasAddresses(addresses ...string) bool {
 		return false
 	}
 
-	return sets.NewString(existingAddresses...).HasAll(addresses...)
+	for _, address := range addresses {
+		found := false
+		for _, existingAddress := range existingAddresses {
+			if existingAddress == address {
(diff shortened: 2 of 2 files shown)


go-controller/pkg/ovn/address_set/address_set_benchmark_test.go
```diff
@@ -0,0 +1,94 @@
...
+
+		b.ReportAllocs()
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			_ = fn(existing, wanted)
+		}
+	})
+}
+
+func BenchmarkHasAllAddressesSetVsScan(b *testing.B) {
```
npinaeva, line 81: that is cool!

go-controller/pkg/ovn/base_network_controller_namespace.go
```diff
@@ -68,6 +68,121 @@ func getNamespaceAddrSetDbIDs(namespaceName, controller string) *libovsdbops.DbO
...
+		}
+	}()
+	if nsInfo.addressSet == nil {
+		return nil
+	}
+	addressSet := nsInfo.addressSet
+	nsUnlock()
+	unlocked = true
+
+	return addressSet.SetAddresses(util.StringSlice(bnc.getAllNamespacePodAddresses(namespace)))
```
npinaeva: is it safe to modify address set after the namespace is unlocked?

go-controller/pkg/ovn/base_network_controller_namespace.go
```diff
@@ -68,6 +68,121 @@ func getNamespaceAddrSetDbIDs(namespaceName, controller string) *libovsdbops.DbO
...
+	nsUnlock()
+	unlocked = true
+	return addressSet.SetAddresses(nil)
+}
+
+func (bnc *BaseNetworkController) shouldPopulateNamespaceAddressSetOnCreate(namespace string, ns *corev1.Namespace) bool {
+	if bnc.isHostNetworkNamespace(namespace) {
+		return true
+	}
+	if bnc.hasNamespaceAddressSetReference(namespace) {
```
npinaeva: wouldn't that always be false on create because namespace controller started before the netpol controller and no one else add references?

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1397,17 +1550,32 @@ func (bnc *BaseNetworkController) handlePeerNamespaceSelectorAdd(np *networkPoli
...
+			namespacesToRef.Insert(namespace.Name)
 		}
 	}
 	np.RUnlock()
+
+	for namespace := range namespacesToRef {
+		bnc.addNamespaceAddressSetReference(namespace)
+	}
+	for namespace := range namespacesToSync {
+		if err := bnc.syncNamespaceAddressSetForPolicy(namespace); err != nil {
```
npinaeva: this technically should only be done if we added a first reference, maybe it makes sense to combine `addNamespaceAddressSetReference` and `syncNamespaceAddressSetForPolicy` so that namespace side decides when to update the address set?

go-controller/pkg/ovn/base_network_controller_policy.go
```diff
@@ -1371,6 +1506,24 @@ func (bnc *BaseNetworkController) cleanupNetworkPolicy(np *networkPolicy) error
...
+				if err := bnc.clearNamespaceAddressSetForPolicy(namespace); err != nil {
+					referenceErrors = append(referenceErrors,
+						fmt.Errorf("failed to clear namespace address set for namespace %s: %w", namespace, err))
+				}
+			}
+		}
+	}
+	if len(referenceErrors) > 0 {
+		return utilerrors.Join(referenceErrors...)
+	}
```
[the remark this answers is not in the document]
npinaeva: unfortunately that is true
we could just say that we won't return an error on `clearNamespaceAddressSetForPolicy` then there will be no retry and no double-decrement, and leaving a filled address set is not too bad?

go-controller/pkg/ovn/base_network_controller_namespace.go
trozet: good catch! No it does not look like it. We rely on RLock to Add/Del IPs and a previous version of myself commented that SetAddresses is not thread safe during runtime. The other mutable operations do it under RLock, so as a cheap fix I think here we can just use WLock.

trozet: write lock will mostly fix it, but there are places (like in the pod add handler) where we release nsinfo R lock after building the address set ops then later execute the libovsdb txn. This could lead to a state where  the libovsdb op from the pod handler executes out of order with whatever we setAddresses to in the network policy namespace handler. To fix/minimize this in the network policy handler we take a write lock, then get the current addresses in the set, and then the current addresses of pods in the cluster, compute the difference and do explicit add/delete operations. That way we don't blindly set and update the address set according to the current state of the cluster. I think that should be OK with the pod handler delayed add/delete operations.

I thought about just only using add addresses, but AI warned me not to do that as the address set could be stale and its best to also try to delete stale addresses that may exist there.

go-controller/pkg/ovn/base_network_controller_namespace.go
trozet: well multiple network policies can reference same ns address set, so it could or could not have already have references

go-controller/pkg/ovn/base_network_controller_policy.go
trozet: I think it is clearer now with the references then the sync address set being two different functions, first one just updates references, second one syncs the address set if needed.

trozet:
next push will just be rebase

go-controller/pkg/ovn/base_network_controller_namespace.go
npinaeva: I am not sure I understand why only setAddresses is not thread-safe, when none of the address set operation are thread-safe, there is no locking inside and the same address set shouldn't be updated from multiple threads using any operation?
Also it is still racy with the pod add operation: after pod add got its ops and released the ns lock, netpol delPeer handler can fully cleanup the address set before pod add transaction is executed?

npinaeva: I think one more race is this:
pod add handler checks if namespace has references, gets false, skips adding podIP to the addr set
netpol adds a namespace reference and initializes the addr set using `getAllNamespacePodAddresses` which gets IPs from the kapi (pod add handler hasn't finished yet and hasn't applied the annotations to the kapi)
pod add handler finishes its work, annotates the IPs
at this point we don't have new pod's IPs in the addr set

trozet:
@npinaeva brought up legit races here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5980#discussion_r2883705231

there is no really good way to solve these races with the code as is. I'm convinced the only way to solve it is to convert address sets to be a level driven controller. I'm going to leave commits 1 and  2 which decouple the a namespace address set a bit and then keep commit 4 which has a perf improvement. Then we will go with Nadia's PR for level driven: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5999

npinaeva (approved):

Issue #5953, opened 2026-02-10
TestEgressIPTrackerControllerWithInformer/multiple_EgressIPs_select_same_namespace_triggers_single_callback


npinaeva:
### Which jobs are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21856956343/job/63075939252?pr=5886

### Which tests are flaking?

UT TestEgressIPTrackerControllerWithInformer/multiple_EgressIPs_select_same_namespace_triggers_single_callback

### Since when has it been flaking?

10 feb 2026

### Reason for failure (if possible)

```
    egressip_tracker_test.go:288: 
        Timed out after 3.001s.
        Expected
            <[]networkmanager.callbackEvent | len:0, cap:0>: []
        to consist of
            <[]networkmanager.callbackEvent | len:1, cap:1>: [
                {node: "node7", nad: "ns7/primary", active: true},
            ]
        the missing elements were
            <[]networkmanager.callbackEvent | len:1, cap:1>: [
                {node: "node7", nad: "ns7/primary", active: true},
            ]

```

trozet:
Same root cause here: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22039191493/job/63677149699?pr=5981

PR #5981, opened 2026-02-15
Perf/Scale: Reduce memory allocation by OpenFlow Manager

trozet:
It was seen during a scale run that there is significant memory
consumption used by OpenFlow Manager:

Showing nodes accounting for 2995262.20MB, 90.52% of 3308836.28MB total
Dropped 2093 nodes (cum <= 16544.18MB)
      flat  flat%   sum%        cum   cum%
         0     0%     0% 1337234.51MB 40.41%  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node.(*openflowManager).Run.func1
289192.79MB  8.74%  8.74% 1335678.47MB 40.37%  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node.(*openflowManager).syncFlows
       6MB 0.00018%  8.74% 1046400.59MB 31.62%  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util.ReplaceOFFlows
         0     0%  8.74% 1041733.16MB 31.48%  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/retry.(*RetryFramework).DoWithLock
  599.04MB 0.018%  8.76% 709890.84MB 21.45%  bytes.(*Buffer).grow
709291.80MB 21.44% 30.19% 709291.80MB 21.44%  bytes.growSlice
         0     0% 30.19% 707386.34MB 21.38%  bytes.(*Buffer).Write

This PR reduces memory consumption per cycle and adds benchmark tests showing the gains. This should reduce overall peak memory usage under heavy churn.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging to include bridge identifiers and flow counts.
  * Safer lock handling for concurrent flow cache operations to reduce race risks.

* **Refactor**
  * Switched flow submission to a streaming approach to lower memory usage and improve performance.

* **Tests**
  * Added unit and benchmark tests to validate streaming behavior and measure performance.

### DIFF
--- a/go-controller/pkg/util/ovs_benchmark_test.go
+++ b/go-controller/pkg/util/ovs_benchmark_test.go
@@ -1,0 +1,82 @@
+package util
+
+import (
+	"bytes"
+	"io"
+	"strings"
+	"testing"
+)
+
+var benchmarkFlowBytesSink int64
+var benchmarkFlowCountSink int
+
+func BenchmarkReplaceOFFlowsInputRendering(b *testing.B) {
+	benchCases := []struct {
+		name      string
+		flowCount int
+	}{
+		{
+			name:      "1k_flows",
+			flowCount: 1000,
+		},
+		{
+			name:      "5k_flows",
+			flowCount: 5000,
+		},
+	}
+
+	for _, tc := range benchCases {
+		flows := makeBenchmarkFlows(tc.flowCount)
+		totalBytes := benchmarkFlowsBytes(flows)
+
+		b.Run(tc.name+"/join_buffer", func(b *testing.B) {
+			b.ReportAllocs()
+			b.SetBytes(totalBytes)
+			for i := 0; i < b.N; i++ {
+				stdin := &bytes.Buffer{}
+				stdin.Write([]byte(strings.Join(flows, "\n")))
+				written, err := io.Copy(io.Discard, stdin)
+				if err != nil {
+					b.Fatalf("failed to drain old flow payload: %v", err)
+				}
+				benchmarkFlowBytesSink = written
+				benchmarkFlowCountSink = stdin.Len()
+			}
+		})
+
+		b.Run(tc.name+"/stream_reader", func(b *testing.B) {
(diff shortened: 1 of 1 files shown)


trozet:
hit UT flake: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5953

go-controller/pkg/util/ovs.go
```diff
@@ -669,11 +670,79 @@ func AddOFFlowWithSpecificAction(bridgeName, action string) (string, string, err
...
+		if r.flowOffset >= len(flow) {
+			// Current flow was fully consumed; advance and schedule delimiter if
+			// there is another flow.
+			r.flowIndex++
+			r.flowOffset = 0
+			if r.flowIndex < len(r.flows) {
+				r.needEOL = true
+			}
+			continue
+		}
```
jcaamano, line 718: nit

do we need this block (lines 711-720) ? Seems it is already handled later (lines 726-733) but not sure if this is covering a corner case I couldn't guess.

go-controller/pkg/util/ovs.go
```diff
@@ -669,11 +670,79 @@ func AddOFFlowWithSpecificAction(bridgeName, action string) (string, string, err
...
+// stream for ovs-ofctl stdin without constructing one large joined string.
+type openFlowStdinReader struct {
+	flows      []string
+	flowIndex  int
+	flowOffset int
+	needEOL    bool
+}
+
+// Read implements io.Reader over r.flows, producing output equivalent to
+// strings.Join(flows, "\n"), but in small chunks to reduce peak allocations.
```
jcaamano, line 683: it's even better right? because we are not allocating anything, just copying from `flows` to `p` directly.

go-controller/pkg/util/ovs.go
trozet: I think we can optimize this a bit

trozet: should be better now

PR #5983, opened 2026-02-16
node: reconcile masquerade resources via link and address events

tsorya:
node: reconcile masquerade resources via link and address events
Masquerade resources (IPs, routes, MAC bindings) on the gateway
interface can be lost when the interface is recreated or addresses
are removed externally. Add ensureMasqueradeResources to restore
them in the correct dependency order (primary IP check, masquerade
IP, routes, MAC bindings), with a fast path that skips when the
interface LinkIndex and masquerade IP are unchanged.

Hook into existing infrastructure for all node modes:
- linkManager: gateway interface link events
- addressManager: new OnMasqueradeIPChanged callback for masquerade
  IP events, periodic sync as fallback
- DPU host mode: minimal addressManager for netlink subscription,
  skipping annotation management

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Adds a masquerade reconciler that runs in host mode and is triggered on gateway startup, gateway MAC updates, link/address events, and masquerade IP changes to ensure configuration is reapplied when needed.

* **Improvements**
  * Consolidates one-time masquerade setup into an idempotent reconciliation flow that restores IPs, routes, and neighbor (ARP/ND) bindings for greater reliability when interfaces or addresses change.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1793,7 +1793,6 @@ func newGateway(
 			}
 			gw.openflowManager.requestFlowSync()
 		}
-
 		if config.Gateway.NodeportEnable {
 			klog.Info("Creating Gateway Node Port Watcher")
 			gw.nodePortWatcher, err = newNodePortWatcher(gwBridge, gw.openflowManager, gw.nodeIPManager, watchFactory, networkManager)
@@ -2097,6 +2096,96 @@ func setNodeMasqueradeIPOnExtBridge(extBridgeName string) error {
 		}
 	}
 
+	return nil
+}
+
+// masqueradeIPConfigured checks if the masquerade IPv4/IPv6 addresses are
+// present on the given link.
+func masqueradeIPConfigured(link netlink.Link) bool {
+	if config.IPv4Mode {
+		_, masqIPNet, _ := net.ParseCIDR(config.Gateway.V4MasqueradeSubnet)
+		masqIPNet.IP = config.Gateway.MasqueradeIPs.V4HostMasqueradeIP
+		if exists, err := util.LinkAddrExist(link, masqIPNet); err != nil || !exists {
+			return false
+		}
+	}
+	if config.IPv6Mode {
+		_, masqIPNet, _ := net.ParseCIDR(config.Gateway.V6MasqueradeSubnet)
+		masqIPNet.IP = config.Gateway.MasqueradeIPs.V6HostMasqueradeIP
+		if exists, err := util.LinkAddrExist(link, masqIPNet); err != nil || !exists {
+			return false
+		}
+	}
+	return true
+}
+
+// masqueradeReconciler ensures that all masquerade-related resources
+// (IP addresses, routes, and MAC bindings) are configured on the gateway interface.
+// It is idempotent and safe to call concurrently from multiple event sources
+// (linkManager, addressManager). A mutex serializes concurrent calls.
+//
+// A fast path checks if the masquerade IP is present and the interface has not changed
+// (same LinkIndex as last successful run). This makes frequent calls cheap (~2 netlink reads).
+// The slow path restores all resources in dependency order: masquerade IP first (routes
+// depend on it for gateway reachability), then routes, then MAC bindings.
+type masqueradeReconciler struct {
+	mu            sync.Mutex
+	lastLinkIndex int
+	routeManager  *routemanager.Controller
+	nodeName      string
+	watchFactory  factory.NodeWatchFactory
+}
+
+func (r *masqueradeReconciler) ensure() error {
+	if r.routeManager == nil || r.watchFactory == nil {
+		return fmt.Errorf("masquerade reconciler not fully initialized (routeManager=%v, watchFactory=%v)", r.routeManager, r.watchFactory)
+	}
+	gwIface := config.Gateway.Interface
+	if gwIface == "" || gwIface == types.DeriveFromMgmtPort {
+		return nil
+	}
+	if !r.mu.TryLock() {
+		return nil
+	}
+	defer r.mu.Unlock()
+
+	link, err := util.GetNetLinkOps().LinkByName(gwIface)
+	if err != nil {
+		return fmt.Errorf("interface %s not found: %w", gwIface, err)
+	}
+
+	// Fast path: same interface (LinkIndex) and masquerade IP present — resources are healthy.
+	if link.Attrs().Index == r.lastLinkIndex && masqueradeIPConfigured(link) {
+		return nil
+	}
+
+	klog.Infof("Ensuring masquerade resources on %s (ifindex=%d, lastSuccessful=%d)",
+		gwIface, link.Attrs().Index, r.lastLinkIndex)
+
+	// Verify interface has a primary (non-link-local, non-masquerade) IP before adding
+	// any resources. Prevents adding masquerade IP before DHCP completes.
+	ifAddrs, err := nodeutil.GetNetworkInterfaceIPAddresses(gwIface)
+	if err != nil {
+		return fmt.Errorf("failed to get IP addresses for interface %s: %w", gwIface, err)
+	}
+	// 1. Masquerade IP first — routes depend on it for gateway reachability.
+	if err := setNodeMasqueradeIPOnExtBridge(gwIface); err != nil {
+		return fmt.Errorf("failed to set masquerade IP: %w", err)
+	}
+	// 2. Routes with the current LinkIndex (interface may have been recreated).
+	if err := addMasqueradeRoute(r.routeManager, gwIface, r.nodeName, ifAddrs, r.watchFactory); err != nil {
+		return fmt.Errorf("failed to add masquerade route: %w", err)
+	}
+	if err := configureSvcRouteViaInterface(r.routeManager, gwIface, DummyNextHopIPs()); err != nil {
+		return fmt.Errorf("failed to configure service route: %w", err)
+	}
+	// 3. ARP/ND entries for masquerade IPs.
+	if err := addHostMACBindings(gwIface); err != nil {
+		return fmt.Errorf("failed to add MAC bindings: %w", err)
+	}
+
+	r.lastLinkIndex = link.Attrs().Index
+	klog.Infof("Masquerade resources ensured on %s (ifindex=%d)", gwIface, r.lastLinkIndex)
 	return nil
 }
 
--- a/go-controller/pkg/node/node_ip_handler_linux.go
+++ b/go-controller/pkg/node/node_ip_handler_linux.go
@@ -41,7 +41,13 @@ type addressManager struct {
 	nodePrimaryAddr net.IP
 	gatewayBridge   *bridgeconfig.BridgeConfiguration
 
-	OnChanged func()
+	OnChanged             func()
+	OnMasqueradeIPChanged func()
+	// gatewayIfIndex caches the link index of config.Gateway.Interface.
+	// Used in DPUHost mode to filter address events to only the gateway
+	// interface. Refreshed in sync() every 30s; all access is from the
+	// runInternal goroutine so no synchronization is needed.
+	gatewayIfIndex int
 	sync.Mutex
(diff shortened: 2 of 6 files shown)


go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2126,6 +2126,39 @@ func setNodeMasqueradeIPOnExtBridge(extBridgeName string) error {
...
+func ensureMasqueradeResources(routeManager *routemanager.Controller, gwIface, nodeName string, watchFactory factory.NodeWatchFactory) error {
+	// 0. Verify interface has a primary (non-link-local, non-masquerade) IP before
+	// adding any resources. Without this check, we'd add the masquerade IP to an
+	// interface that hasn't acquired its primary IP yet (e.g., DHCP pending after
+	// DPU reboot), which can block DHCP from completing.
+	ifAddrs, err := nodeutil.GetNetworkInterfaceIPAddresses(gwIface)
+	if err != nil {
+		return fmt.Errorf("interface %s not ready (no primary IP): %w", gwIface, err)
+	}
+	// 1. Masquerade IP must be first — routes depend on it for gateway reachability
```
wizhaoredhat: Would it be possible to add a check to see if Masquerade IPs have been reset on the DPU Host, before attempting to do all the set/add operations? I think that is more kind to the DPU Host's CPU.

tsorya: Sure but this complicates things a little bit. 
It means we must be sure that all operations worked and not only ip set. 
If we are ok to make it more complex I'll do it.

go-controller/pkg/node/gateway.go
```diff
@@ -348,9 +353,45 @@ func (g *gateway) Start() error {
...
+// ensureMasqueradeResources: each function checks existence before acting, so
+// normal ticks are lightweight no-ops. When the interface is recreated (DPU reboot)
+// or masquerade IPs are removed, the functions detect the difference and restore.
+func (g *gateway) runMasqueradeReconciler(stopChan <-chan struct{}, wg *sync.WaitGroup) {
+	if config.Gateway.Interface == "" {
+		klog.Warning("Gateway interface not configured, skipping masquerade reconciler")
+		return
+	}
+
+	const reconcilePeriod = 30 * time.Second
```
trozet: i think you can potentially hook into the node ip manager and link manager to get netlink events when the gateway interface goes down or loses an IP address. Wouldn't this be something we also want to run for non-dpu setups as well?

tsorya: Let me test events option. 
Sure it will match all the flows, just wanted to minimize the change but ill enable it for all

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2126,6 +2126,39 @@ func setNodeMasqueradeIPOnExtBridge(extBridgeName string) error {
 	return nil
 }
 
+// ensureMasqueradeResources ensures that all masquerade-related resources
+// (IP addresses, routes, and MAC bindings) are configured on the gateway interface.
+// This function is idempotent and can be called safely on every reconcile tick.
+// The masquerade IP must be set first because routes use a gateway IP (e.g., 169.254.0.4)
+// that is only reachable once the masquerade IP (e.g., 169.254.0.2/29) creates the
+// connected route for the masquerade subnet.
+func ensureMasqueradeResources(routeManager *routemanager.Controller, gwIface, nodeName string, watchFactory factory.NodeWatchFactory) error {
```
aserdean: please remember to add a unit test, ideally it should cove:
Test ensureMasqueradeResources is truly idempotent (call twice, second is a no-op)
Test it recovers after interface recreation (new LinkIndex)
Test it skips when no primary IP is present (DHCP pending)

go-controller/pkg/node/gateway.go
```diff
@@ -348,9 +353,45 @@ func (g *gateway) Start() error {
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		ticker := time.NewTicker(reconcilePeriod)
```
aserdean: time.NewTicker fires after the first 30s interval, so there's a window between Start() and the first tick where a DPU reboot could leave masquerade resources missing. Consider restructuring to reconcile immediately on entry.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2126,6 +2126,39 @@ func setNodeMasqueradeIPOnExtBridge(extBridgeName string) error {
...
+	// 0. Verify interface has a primary (non-link-local, non-masquerade) IP before
+	// adding any resources. Without this check, we'd add the masquerade IP to an
+	// interface that hasn't acquired its primary IP yet (e.g., DHCP pending after
+	// DPU reboot), which can block DHCP from completing.
+	ifAddrs, err := nodeutil.GetNetworkInterfaceIPAddresses(gwIface)
```
aserdean, line 2167: This function is called from a goroutine in runMasqueradeReconciler. If routeManager or watchFactory is nil (e.g. partially initialized gateway struct), it will panic inside the goroutine, which is hard to debug. Consider adding nil guards at the top:

> if routeManager == nil || watchFactory == nil {
>     return fmt.Errorf("routeManager or watchFactory not initialized")
> }
>

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2126,6 +2126,39 @@ func setNodeMasqueradeIPOnExtBridge(extBridgeName string) error {
...
+	if err != nil {
+		return fmt.Errorf("interface %s not ready (no primary IP): %w", gwIface, err)
```
aserdean: nit pick:
The error message always says "not ready (no primary IP)" but GetNetworkInterfaceIPAddresses can fail for other reasons too — interface doesn't exist, permission errors, IPv6 subnet lookup failures, etc. This could be misleading when debugging. Consider a more generic message:
 return fmt.Errorf("failed to get IP addresses for interface %s: %w", gwIface, err)

go-controller/pkg/node/gateway.go
tsorya: @trozet from what i see we need to change addressManager logic and enable it for masquerade ip as it is filtered out right now? Is it ok?

go-controller/pkg/node/gateway_shared_intf.go
tsorya: Thought about it, after changes that @trozet asked i don't see a reason to complicates things here, running once in 2 minutes should be ok

go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -169,6 +171,10 @@ func (c *addressManager) runInternal(stopChan <-chan struct{}, subscribe subscri
 				}
 				continue
 			}
+			if util.IsAddressReservedForInternalUse(a.LinkAddress.IP) {
+				c.OnMasqueradeIPChanged()
+				continue
+			}
```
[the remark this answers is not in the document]
tsorya: it can't be nil

go-controller/pkg/node/gateway_shared_intf.go
wizhaoredhat: So blocking on netlink events?

go-controller/pkg/node/gateway.go
trozet: should be OK as long as we are not considering the masq IP as a host address that we annotate.

go-controller/pkg/node/node_ip_handler_linux.go
```diff
@@ -55,14 +61,15 @@ func newAddressManager(nodeName string, k kube.Interface, mgmtPort managementpor
...
-		syncPeriod:    30 * time.Second,
+		nodeName:              nodeName,
+		watchFactory:          watchFactory,
+		cidrs:                 sets.New[string](),
+		mgmtPort:              mgmtPort,
+		gatewayBridge:         gwBridge,
+		OnChanged:             func() {},
+		OnMasqueradeIPChanged: func() {},
+		useNetlink:            useNetlink,
+		syncPeriod:            30 * time.Second,
```
aserdean, line 72: Have you considered subscribing to route and neighbor netlink events (e.g. piggybacking on routeManager's existing RouteSubscribe, and adding a NeighSubscribe filtered to the gateway interface) instead of relying on the periodic sync?

That would make the reconciler fully reactive and remove the blind spot for route/MAC-binding-only deletions.

aserdean:
Consider adding a nil guard in isValidNodeIP.

In normal (non-DPU) node mode, refreshGatewayIfIndex() is called every 30s sync even though gatewayIfIndex is only used for DPUHost event filtering. This wastes a LinkByName netlink call. Consider gating it:

if config.OvnKubeNode.Mode == types.NodeModeDPUHost {
    c.refreshGatewayIfIndex()

tsorya:
> In normal (non-DPU) node mode, refreshGatewayIfIndex() is called every 30s sync even though gatewayIfIndex is only used for DPUHost event filtering. This wastes a LinkByName netlink call. Consider gating it:
> 
> if config.OvnKubeNode.Mode == types.NodeModeDPUHost {     c.refreshGatewayIfIndex()

This change applies to both DPU host mode and full mode; therefore, it should not be gated in either case

go-controller/pkg/node/node_ip_handler_linux.go
tsorya: That’s a great point. Moving to a reactive model using netlink events would definitely be more robust and eliminate those blind spots. However, this PR focuses on leveraging the existing reconciliation logic rather than introducing a new architectural pattern.

aserdean: fair, we could have a follow-up if anyone wants to pursue this

aserdean:
> This change applies to both DPU host mode and full mode; therefore, it should not be gated in either case

Makes sense I thought that case is already handled!
Feel free to ignore the comment.

Just still this small nit if you want include:
Consider adding a nil guard in isValidNodeIP.

tsorya:
> > This change applies to both DPU host mode and full mode; therefore, it should not be gated in either case
> 
> Makes sense I thought that case is already handled! Feel free to ignore the comment.
> 
> Just still this small nit if you want include: Consider adding a nil guard in isValidNodeIP.

You mean this one https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/node_ip_handler_linux.go#L415 ?

aserdean:
> You mean this one https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/node/node_ip_handler_linux.go#L415 ?

yes

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -2121,63 +2121,64 @@ func masqueradeIPConfigured(link netlink.Link) bool {
...
-	if link.Attrs().Index == lastMasqueradeLinkIndex && masqueradeIPConfigured(link) {
+	if link.Attrs().Index == r.lastLinkIndex && masqueradeIPConfigured(link) {
 		return nil
 	}
 
 	klog.Infof("Ensuring masquerade resources on %s (ifindex=%d, lastSuccessful=%d)",
-		gwIface, link.Attrs().Index, lastMasqueradeLinkIndex)
+		r.gwIface, link.Attrs().Index, r.lastLinkIndex)
 
-	// Verify interface has a primary (non-link-local, non-masquerade) IP before adding
```
trozet: comments are lost in this commit

tsorya: added back

aserdean:
LGTM, please remember to squash the commits

aserdean:
I am testing the PR and seeing: 
```
ovnkube.go:147] failed to run ovnkube: failed to start node network controller: failed to init default node network controller: interface derive-from-mgmt-port not found: Link not found
```
working with @tsorya to iron out the order of the functions

aserdean:
LGTM, 

Tested-by: Alin Serdean <aserdean@nvidia.com>

PR #5992, opened 2026-02-18
Trivial E2E egress IP fixes

trozet:
- Changed the MTU test external container to start
  directly with netexec instead of creating it with
  pause and launching netexec in a detached goroutine.

- We need to set the PMTU sysctl before the socket is opened for the
  application.

- Fixed curlErr that was being shadowed and not evaluated properly.

## Summary by CodeRabbit

* **Tests**
  * Replaced a paused-container approach with an active network service to enable realistic MTU-related behavior during egress testing.
  * Simplified connectivity polling and unified error handling for clearer, more accurate test failure reporting.

### DIFF
--- a/test/e2e/egressip.go
+++ b/test/e2e/egressip.go
@@ -2160,35 +2160,24 @@ spec:
 		providerPrimaryNetwork, err := infraprovider.Get().PrimaryNetwork()
 		framework.ExpectNoError(err, "failed to get providers primary network")
 		externalContainerPrimary := infraapi.ExternalContainer{Name: "external-container-for-egressip-mtu-test", Image: images.AgnHost(),
-			Network: providerPrimaryNetwork, CmdArgs: []string{"pause"}, ExtPort: externalContainerPrimaryPort}
+			Network: providerPrimaryNetwork, RuntimeArgs: []string{"--sysctl", "net.ipv4.ip_no_pmtu_disc=2"},
+			CmdArgs: []string{"netexec", httpPort, udpPort}, ExtPort: externalContainerPrimaryPort}
 		externalContainerPrimary, err = providerCtx.CreateExternalContainer(externalContainerPrimary)
 		framework.ExpectNoError(err, "failed to create external container: %s", externalContainerPrimary.String())
 
-		// First disable PMTUD
-		_, err = infraprovider.Get().ExecExternalContainerCommand(externalContainerPrimary, []string{"sysctl", "-w", "net.ipv4.ip_no_pmtu_disc=2"})
-		framework.ExpectNoError(err, "disabling PMTUD in the external kind container failed: %v", err)
-		providerCtx.AddCleanUpFn(func() error {
-			_, err = infraprovider.Get().ExecExternalContainerCommand(externalContainerPrimary, []string{"sysctl", "-w", "net.ipv4.ip_no_pmtu_disc=0"})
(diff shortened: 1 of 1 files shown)


trozet:
rebasing to pick up latest e2e fixes

trozet:
Test failed in CI:

e2e egress IP validation Cluster Default Network [It] of replies to egress IP packets that require fragmentation [LGW][IPv4]

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22157252858/job/64075332006?pr=5992

looking into it.

trozet:
@kyrtapz according to AI it was failing because:

```
  - Before your commit, netexec was started after sysctl (pause container + exec), so its UDP socket inherited the PMTU setting.
  - After your commit, netexec starts before sysctl, so its already-open UDP socket keeps default PMTU behavior; first large reply can be dropped, and this test does not retry.

```

so updated to make the sysctl setting happen before the agent is running a server

test/e2e/egressip.go
```diff
@@ -2160,35 +2160,24 @@ spec:
...
 		ginkgo.By("Checking connectivity to the external kind container and verify that the source IP is the egress IP")
 		var curlErr error
-		_ = wait.PollUntilContextTimeout(
+		err = wait.PollUntilContextTimeout(
 			context.Background(),
 			retryInterval,
 			retryTimeout,
 			true,
 			func(ctx context.Context) (bool, error) {
-				curlErr := curlAgnHostClientIPFromPod(podNamespace.Name, pod1Name, egressIP1.String(), externalContainerPrimary.GetIPv4(), externalContainerPrimary.GetPortStr())
```
npinaeva, line 2187: lol

test/e2e/egressip.go
```diff
@@ -2160,35 +2160,24 @@ spec:
...
+			Network: providerPrimaryNetwork, RuntimeArgs: []string{"--sysctl", "net.ipv4.ip_no_pmtu_disc=2"},
+			CmdArgs: []string{"netexec", httpPort, udpPort}, ExtPort: externalContainerPrimaryPort}
 		externalContainerPrimary, err = providerCtx.CreateExternalContainer(externalContainerPrimary)
 		framework.ExpectNoError(err, "failed to create external container: %s", externalContainerPrimary.String())
 
-		// First disable PMTUD
-		_, err = infraprovider.Get().ExecExternalContainerCommand(externalContainerPrimary, []string{"sysctl", "-w", "net.ipv4.ip_no_pmtu_disc=2"})
-		framework.ExpectNoError(err, "disabling PMTUD in the external kind container failed: %v", err)
-		providerCtx.AddCleanUpFn(func() error {
-			_, err = infraprovider.Get().ExecExternalContainerCommand(externalContainerPrimary, []string{"sysctl", "-w", "net.ipv4.ip_no_pmtu_disc=0"})
```
npinaeva, line 2171: we don't really care about setting it back right?

npinaeva: I think @trozet said we actually do?
let's wait with the merge until confirmed

kyrtapz: isn't this only affecting the containers net namespace?

trozet: yeah on second thought the external container is not host networked, so @kyrtapz is right it should only affect that network namespace

Issue #5993, opened 2026-02-18
E2E Flake: ACL Logging for AdminNetworkPolicy and BaselineAdminNetworkPolicy [BeforeEach] the ANP ACL logs have the expected log level


trozet:
### Which jobs are flaking?

bgp noHA local dualstack advertise-default lane

### Which tests are flaking?

ACL Logging for AdminNetworkPolicy and BaselineAdminNetworkPolicy [BeforeEach] the ANP ACL logs have the expected log level

### Since when has it been flaking?

last week

### Reason for failure (if possible)
```
[Feature:AdminNetworkPolicy, Feature:BaselineNetworkPolicy]
  [BeforeEach] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/acl_logging.go:192
  [It] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/acl_logging.go:237

  [FAILED] Unexpected error:
      <exec.CodeExitError>: 
      error running /usr/local/bin/kubectl --server=https://127.0.0.1:33085/ --kubeconfig=/home/runner/ovn.conf --namespace=default create ns anp-peer-open:
      Command stdout:
      
      stderr:
      Error from server (AlreadyExists): object is being deleted: namespaces "anp-peer-open" already exists
      
      error:
      exit status 1
      {
          Err: <*errors.errorString | 0xc001275320>{
              s: "error running /usr/local/bin/kubectl --server=https://127.0.0.1:33085/ --kubeconfig=/home/runner/ovn.conf --namespace=default create ns anp-peer-open:\nCommand stdout:\n\nstderr:\nError from server (AlreadyExists): object is being deleted: namespaces \"anp-peer-open\" already exists\n\nerror:\nexit status 1",
          },
          Code: 1,
      }
  occurred
  In [BeforeEach] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/acl_logging.go:207 @ 02/14/26 00:37:00.792
```

Flakes because the clean up deletes a namespace, but does not wait for it to be fully destroyed, then the next test tries to re-create the same namespace.

PR #5994, opened 2026-02-18
E2E: Fixes flaking ANP ACL logging

trozet:
The test was re-using the same namespace name, but would not wait for namespace deletion before starting the next test. This would cause the next test to fail because it cannot create the namespace while it is terminating.

Change the test to use random namespace names.

Fixes: #5993


kyrtapz (approved):
I don't like the nsNames array approach and the fact that we don't create the namespaces with the framework but these were not introduced here and the fix makes sense to me.
LGTM!

PR #5999, opened 2026-02-19
Switch to a separate address set manager

npinaeva:
## 📑 Description

It follow the logic and interfaces of pod_selector_address_set
from the ovn package, but uses a single level-driven controller
instead of per-address-set retry frameworks and is separated from
the BNC.
We can drop the previous address set format cleanup, because
it's been present since release-1.0.
2 cleanup unit tests are dropped, because the cleanup logic has
been simplified. Previously we had to stop all handlers and
now we only need to delete address set.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Refactor**
  * Introduced a centralized Address Set Manager and wired it into network controller lifecycles for more consistent address-set handling.

* **Tests**
  * Consolidated and standardized test object factories into a shared testing package; added tests and suites for address-set behavior.

* **Chores**
  * Removed legacy pod-selector-based address-set implementation and related tests; updated controllers to use the new manager.

### DIFF
--- a/go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
+++ b/go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
@@ -43,14 +43,23 @@ type podSelectorAddressSet struct {
 	// selectedNamespaces is a cache for namespaces that were selected by this address set during the last reconciliation
 	// used to optimize events processing.
 	selectedNamespaces sets.Set[string]
+
+	// network-specific fields
+	controllerName string
+	netInfo        util.NetInfo
 }
 
 // AddressSetManager manages shared address sets with pod IPs based on provided pod and namespace selectors.
+// It shared across network controllers.
 type AddressSetManager struct {
 	name     string
 	nbClient libovsdbclient.Client
 
-	addressSetFactory addressset.AddressSetFactory
+	// address set factory ip modes only affect which IPs are getting selected for the operations
+	// different networks may have different setups, so we need all combinations
+	addressSetFactoryV4        addressset.AddressSetFactory
+	addressSetFactoryV6        addressset.AddressSetFactory
+	addressSetFactoryDualstack addressset.AddressSetFactory
 
 	// addressSets stores all currently used address sets.
 	addressSets *syncmap.SyncMap[*podSelectorAddressSet]
@@ -62,30 +71,23 @@ type AddressSetManager struct {
 	nsController         controller.Controller
 	addressSetReconciler controller.Reconciler
 
-	// network-specific fields
-	// Warning: when we switch to the global manager instead of per-network
-	// these attributes will be needed per address set request, but also make sure to update
-	// getPodSelectorKey to include controllerName, fortunately it is already in the dbIDs and they don't need to change.
-	controllerName          string
-	netInfo                 util.NetInfo
+	// All network controllers are getting this function from the same networkmanager, so we can share it
 	getNetworkNameForNADKey func(nadKey string) string
 }
 
 func NewAddressSetManager(podInformer coreinformers.PodInformer, namespaceInformer coreinformers.NamespaceInformer,
-	nbClient libovsdbclient.Client, factory addressset.AddressSetFactory, controllerName string,
-	netInfo util.NetInfo, getNetworkNameForNADKey func(nadKey string) string) *AddressSetManager {
+	nbClient libovsdbclient.Client, getNetworkNameForNADKey func(nadKey string) string) *AddressSetManager {
 	m := &AddressSetManager{
-		name:                    "pod-selector-address-set-manager",
-		nbClient:                nbClient,
-		addressSetFactory:       factory,
-		addressSets:             syncmap.NewSyncMap[*podSelectorAddressSet](),
-		podLister:               podInformer.Lister(),
-		namespaceLister:         namespaceInformer.Lister(),
-		controllerName:          controllerName,
-		netInfo:                 netInfo,
-		getNetworkNameForNADKey: getNetworkNameForNADKey,
+		name:                       "pod-selector-address-set-manager",
+		nbClient:                   nbClient,
+		addressSetFactoryV4:        addressset.NewOvnAddressSetFactory(nbClient, true, false),
+		addressSetFactoryV6:        addressset.NewOvnAddressSetFactory(nbClient, false, true),
+		addressSetFactoryDualstack: addressset.NewOvnAddressSetFactory(nbClient, true, true),
+		addressSets:                syncmap.NewSyncMap[*podSelectorAddressSet](),
+		podLister:                  podInformer.Lister(),
+		namespaceLister:            namespaceInformer.Lister(),
+		getNetworkNameForNADKey:    getNetworkNameForNADKey,
 	}
-
 	podCfg := &controller.ControllerConfig[corev1.Pod]{
 		RateLimiter:    workqueue.DefaultTypedControllerRateLimiter[string](),
 		Reconcile:      m.reconcilePod,
@@ -132,8 +134,7 @@ func (m *AddressSetManager) Stop() {
 }
 
 func (m *AddressSetManager) initialSync() error {
-	predicateIDs := libovsdbops.NewDbObjectIDs(libovsdbops.AddressSetPodSelector, m.controllerName, nil)
-	return libovsdbutil.DeleteAddrSetsWithoutACLRef(predicateIDs, m.nbClient)
+	return libovsdbutil.DeleteAddrSetsWithoutACLRefAnyController(libovsdbops.AddressSetPodSelector, m.nbClient)
 }
 
 // EnsureAddressSet returns address set for requested (podSelector, namespaceSelector, namespace).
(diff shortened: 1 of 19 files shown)


go-controller/pkg/ovn/layer2_user_defined_network_controller.go
```diff
@@ -398,6 +399,9 @@ func NewLayer2UserDefinedNetworkController(
 		eIPController:          eIPController,
 		remoteNodesNoRouter:    sync.Map{},
 	}
+	oc.addressSetManager = addresssetmanager.NewAddressSetManager(oc.watchFactory.PodCoreInformer(),
```
trozet: Need to think about if we want a global address set manager that is network aware, instead of a per UDN address set manager.

npinaeva: I actually started with that, but it was ugly wiring it into the existing UDN controllers. You will have to pass netinfo and `getNetworkNameForNADKey` for every address set request though when we switch (but I assume it will be a common requirement, so we will come up with something)

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -0,0 +1,523 @@
...
+	libovsdbops "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
+	libovsdbutil "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/libovsdb/util"
+	addressset "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/address_set"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/syncmap"
+	ovntypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/types"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/util"
+)
+
+// PodSelectorAddressSet stores address set for modifications and used selectors that define this address set.
+type PodSelectorAddressSet struct {
```
trozet: nit: could we just call this SelectorAddressSet?

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -0,0 +1,523 @@
...
+	// backRefs is a map of objects that use this address set.
+	// keys must be unique for all possible users, e.g. for NetworkPolicy use (np *networkPolicy) getKeyWithKind().
+	// Must only be changed with oc.podSelectorAddressSets Lock.
```
trozet: little odd that you comment about a lock that isn't part of this struct. Really nothing in this struct should be modified by two goroutines at the same time, and that protection is handled by the consumer a sync.Map right? If so I would just remove this comment.

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -0,0 +1,523 @@
...
+			addrSet, found := m.addressSets.Load(addrSetKey)
+			if !found {
+				// nothing to do
+				return nil
+			}
+			// only check address sets that have previously matched pod's namespace to avoid extra reconciliations
+			previouslyMatchedNamespaces := addrSet.selectedNamespaces
+			if previouslyMatchedNamespaces == nil || previouslyMatchedNamespaces.Has(namespace) {
+				m.addressSetReconciler.Reconcile(addrSetKey)
+				return nil
```
trozet, line 291: this code seems weird to me. Why rely on previouslyMatchedNamespaces instead of calculating if the pod's namespace applies to an address set with a common function you use in the namespace handling path?

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -0,0 +1,523 @@
...
+func getPodSelectorKey(podSelector, namespaceSelector *metav1.LabelSelector, namespace string) string {
+	var namespaceKey string
+	if namespaceSelector == nil {
+		// namespace is static
+		namespaceKey = namespace
+	} else {
+		namespaceKey = shortLabelSelectorString(namespaceSelector)
+	}
+	return namespaceKey + "_" + shortLabelSelectorString(podSelector)
+}
```
trozet, line 542: this is all pretty gross, but I can't think of a better way to make a unique key. If we don't use a unique key then on EnsureAddress set we have to loop through every address set compare all values of the struct to find a match.

go-controller/pkg/ovn/base_network_controller.go
```diff
@@ -38,6 +38,7 @@ import (
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/networkmanager"
 	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/observability"
 	addressset "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/address_set"
+	"github.com/ovn-org/ovn-kubernetes/go-controller/pkg/ovn/addresssetmanager"
```
trozet: would it be better to put the addresssetmanager under pkg/ovn/controller? I guess that is where we are storing our level-driven controllers these days.

go-controller/pkg/ovn/layer2_user_defined_network_controller.go
trozet: I think that is OK. It is similar to what I'm doing in the node level driven controller. The benefit is we dont have to make N number of address set managers for every UDN we spawn. I think we should have one address set manager in this PR that is network aware for UDNs. It's ok to have another one for CDN for now.

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -0,0 +1,523 @@
...
+
+func (m *AddressSetManager) Stop() {
+	klog.Infof("Stopping %s controller", m.name)
+	controller.Stop(m.podController, m.nsController, m.addressSetReconciler)
+}
+
+func (m *AddressSetManager) initialSync() error {
+	predicateIDs := libovsdbops.NewDbObjectIDs(libovsdbops.AddressSetPodSelector, m.controllerName, nil)
+	return libovsdbutil.DeleteAddrSetsWithoutACLRef(predicateIDs, m.nbClient)
```
trozet: so I think this addressset manager has to start up before the network policy stuff starts. Which means if an NP was deleted, the ACL wont be removed until after this initial sync was done. That means I guess in network policy we still need to clean up stale address sets as well?

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
trozet: if we are sure this will always be pods then that is fine. I'm just worried that we are missing a generic case where we are not selecting only pod ips for an address set or something I dont know... Maybe we end up selecting another CRD like a VTEP CRD that has encap ips in them, and then we build an address set from that. Just thinking about long term more controllers creating address sets from things other than pod ips.

npinaeva:
that was rebase

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
npinaeva: It would be easier to tell in the future when that functionality is needed, otherwise we can guess the worng name
for now it only manages pod IPs based on selector, if you want to simplify the name, maybe `podsAddressSet` is more accurate? when you only say selector it doesn't explain what is selected

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
npinaeva: more than that, this lock doesn't exist anymore :)

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
npinaeva: yeah that is an optimization, I tried to explain the logic a little higher in the comment.
imagine you have a namespace selector `ns=A, podSelector={}` so select all pods in namespace A, now with every pod event in any of the namespaces, this address set would have to be reconciled, find that namespace doesn't match, then do nothing. Instead, we can always be sure that if the namespace wasn't selected before a pod change can't affect the addr set. This should have heavy impact on performance

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
npinaeva: yeah I don't love it either, but the good news is that selectors probably will never change, so it is kinda safe and we have already been using this for a long time :D

go-controller/pkg/ovn/base_network_controller.go
npinaeva: good question... I see controllers folder more like the per-feature/resource controllers, while this manager is just an internal part for the other controllers to use, so I am not sure

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
npinaeva: this logic didn't change, it actually deletes stale address sets on the next restart only. otherwise there is no good way to cleanup in a race-free way and the hope is that not too many changes can happen while ovn-k restarts, so it is ok to have a bit of stale addr sets in the db

go-controller/pkg/ovn/layer2_user_defined_network_controller.go
npinaeva: that was a struggle, let's see if it works

PR #6004, opened 2026-02-20
Handle stale unprocessed NAD if namespace is already gone

booxter:
## 📑 Description

When a ClusterUserDefinedNetwork creates a NAD and the target namespace
is immediately marked as terminating, there is a race condition between
NAD creation and deletion. The deleteNAD function reads the NAD from the
informer lister cache, which may not have received the watch event for
the just-created NAD yet. When the lister returns NotFound, deleteNAD
treated this as "already deleted" and returned nil. The caller then
removed the namespace from namespaceTracker, permanently losing track of
it. Subsequent syncs never attempted to delete the NAD again, leaving it
orphaned.

Fix this by falling back to an API check when the NAD is not found in
the lister cache. If the NAD exists in the API, return an error so the
caller preserves the namespace in the tracker and the work queue retries
the sync. On the next attempt, the informer cache will have caught up
and the normal deletion path will proceed.

Includes a test case that simulates the scenario.

This issue was hit in:
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22238494590/job/64336007750

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of network attachment definition deletion when the cache is out of sync with the API. The controller now verifies existence through the API and properly manages retry logic, ensuring reliable cleanup during namespace deletion scenarios.

* **Tests**
  * Added test coverage for cache-lag edge cases during namespace termination to ensure proper NAD tracking and retry behavior.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
@@ -118,8 +118,21 @@ func (c *Controller) deleteNAD(obj client.Object, namespace string) error {
 	}
 	nadCopy := nad.DeepCopy()
 
-	if nadCopy == nil ||
-		!metav1.IsControlledBy(nadCopy, obj) ||
+	if nadCopy == nil {
+		// NAD not found in lister cache; verify it's truly gone from the API
+		// before treating as success. The lister cache may lag behind the API,
+		// especially right after the NAD was created.
+		_, apiErr := c.nadClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespace).Get(context.Background(), obj.GetName(), metav1.GetOptions{})
+		if apiErr == nil {
+			return fmt.Errorf("NetworkAttachmentDefinition %s/%s not found in cache but still exists in the API, retrying", namespace, obj.GetName())
+		}
+		if !apierrors.IsNotFound(apiErr) {
+			return fmt.Errorf("failed to get NetworkAttachmentDefinition %s/%s from API: %v", namespace, obj.GetName(), apiErr)
+		}
+		return nil
+	}
+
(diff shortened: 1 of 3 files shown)


go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -118,8 +118,21 @@ func (c *Controller) deleteNAD(obj client.Object, namespace string) error {
 	}
 	nadCopy := nad.DeepCopy()
 
-	if nadCopy == nil ||
-		!metav1.IsControlledBy(nadCopy, obj) ||
+	if nadCopy == nil {
```
trozet: you could use https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/networkmanager/api.go#L91

RegisterNADReconciler to get nad events, then check if this nad says it is owned by a CUDN/UDN + decide what to do then.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -118,8 +118,21 @@ func (c *Controller) deleteNAD(obj client.Object, namespace string) error {
...
+		// NAD not found in lister cache; verify it's truly gone from the API
+		// before treating as success. The lister cache may lag behind the API,
+		// especially right after the NAD was created.
+		_, apiErr := c.nadClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(namespace).Get(context.Background(), obj.GetName(), metav1.GetOptions{})
+		if apiErr == nil {
```
trozet: if you already get the NAD here, why error and retry? why not just use it and delete the thing?

PR #6005, opened 2026-02-23
Fix UDN network controller deadlock due to stopChan nil race

trozet:
In UDNs, goroutines are started for some controllers like NetworkQoS where a waitgroup is used to add a reference to the goroutine, and then stopChan is passed as a mechanism to shutdown the NetworkQoS controller.

If the UDN controller starts, and then shutsdown very quickly, the stopChan is closed and reset to nil. It is set to nil as a pattern we use to guard multiple Stop calls to the UDN controller (Stop may be called multiple times). However, if the NetworkQoS goroutine does not finish starting before the stopChan is closed and reset to nil, then by the time NetworkQos gets to read stopChan, it will hang forever, causing the UDN controller waitgroup to wait forever.

This will deadlock the entire network manager from being able to start/stop anymore UDN controllers!

We can see this behavior in CI here:
```
I0223 04:37:24.677192      77 network_controller.go:415] [zone-nad-controller network controller]: sync network wpnhc_tenant-blue
I0223 04:37:24.677203      77 localnet_user_defined_network_controller.go:311] Stoping controller for UDN wpnhc_tenant-blue
I0223 04:37:24.677209      77 base_secondary_layer2_network_controller.go:39] Stop secondary localnet network controller of network wpnhc_tenant-blue
I0223 04:37:24.677241      77 obj_retry.go:473] Stop channel got triggered: will stop retrying failed objects of type *v1.Namespace
I0223 04:37:24.677250      77 network_qos_controller.go:215] Starting controller wpnhc_tenant-blue-network-controller
I0223 04:37:24.677256      77 network_qos_controller.go:218] Waiting for informer caches (networkqos,namespace,pod,node) to sync
I0223 04:37:24.677263      77 obj_retry.go:473] Stop channel got triggered: will stop retrying failed objects of type *v1beta2.MultiNetworkPolicy
I0223 04:37:24.677270      77 shared_informer.go:349] "Waiting for caches to sync" controller="wpnhc_tenant-blue-network-controller"
I0223 04:37:24.677339      77 shared_informer.go:356] "Caches are synced" controller="wpnhc_tenant-blue-network-controller"
```

There is never a "finished syncing network wpnhc_tenant-blue" log again after this for zone-nad-controller, nor any other networks for that matter after this point in the log. However, there are logs for node-nad-controller as it did not hit this race.

To fix this, pass a copy of the oc.stopChan to the goroutines. Channels are copied as a reference so closing the oc.stopChan still closes the copy, and we can still allow oc.stopChan to be set to nil as a Stop guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of network QoS and network controllers by preventing race conditions in shutdown handling across multiple components.

### DIFF
--- a/go-controller/pkg/node/default_node_network_controller.go
+++ b/go-controller/pkg/node/default_node_network_controller.go
@@ -969,10 +969,10 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
 			return err
 		}
 		nc.wg.Add(1)
-		go func() {
+		go func(stopCh <-chan struct{}) {
 			defer nc.wg.Done()
-			nodeController.Run(nc.stopChan)
-		}()
+			nodeController.Run(stopCh)
+		}(nc.stopChan)
 	} else if config.OvnKubeNode.Mode != types.NodeModeDPUHost {
 		// attempt to cleanup the possibly stale bridge
 		_, stderr, err := util.RunOVSVsctl("--if-exists", "del-br", "br-ext")
@@ -1080,7 +1080,7 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
 	nc.linkManager.Run(nc.stopChan, nc.wg)
 
 	nc.wg.Add(1)
-	go func() {
+	go func(stopCh <-chan struct{}) {
 		defer nc.wg.Done()
 		podResClient, err := podresourcesapi.New()
 		if err != nil {
@@ -1092,8 +1092,8 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
 				klog.V(4).Infof("Error closing PodResourcesAPI client: %v", err)
 			}
 		}()
-		ovspinning.Run(ctx, nc.stopChan, podResClient)
-	}()
+		ovspinning.Run(ctx, stopCh, podResClient)
+	}(nc.stopChan)
 
 	klog.Infof("Default node network controller initialized and ready.")
 	return nil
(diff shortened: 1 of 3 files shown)


npinaeva:
>  However, if the NetworkQoS goroutine does not finish starting before the stopChan is closed and reset to nil, then by the time NetworkQos gets to read stopChan, it will hang forever, causing the UDN controller waitgroup to wait forever. 

Just to clarify: the only way to get nil channel is to run  `oc.nqosController.Run()` after `oc.stopChan` is set to nil. 
Calling Stop() before Start() has returned shouldn't be possible, so the only way to get this race is to use goroutines in the `Start()`. I think it would be more reliable to just add a `if stopCh == nil {return}` to the start methods to avoid this problem for all callers vs making sure stopChan is preserved before each call.
I think we have some more similar cases in the node controller
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6b86da4dd3bdef3702248b0a30786391c745aa7e/go-controller/pkg/node/default_node_network_controller.go#L974
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6b86da4dd3bdef3702248b0a30786391c745aa7e/go-controller/pkg/node/default_node_network_controller.go#L1095
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/6b86da4dd3bdef3702248b0a30786391c745aa7e/go-controller/pkg/node/default_node_network_controller.go#L1140
But I am not sure if this is a problem for the service controller

trozet:
@npinaeva I don't think that pattern you suggested is safe:
1. It can cause a data race if oc.stopChan is being written as nil while the goroutine reads it.
2. Just because it passes oc.stopChan == nil doesnt guarantee the read right after it wont be nil.

There's no synchronicity here between reads and writes.

Default controller it is not likely to ever stop very quickly, but we should probably also fix this there. If you agree with my fix I'll also change that.

npinaeva (approved):
beautiful :)

PR #6009, opened 2026-02-24
Revert "Use `nft destroy` to simplify the UDN cleanup code"

npinaeva:
Reverts ovn-kubernetes/ovn-kubernetes#5858
Apparently that broke d/s and in general a newer kernel version is only needed if the user wants UDN to work, while a general nftables requirement affects all features.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of network segmentation rules cleanup to ensure proper removal of rules even when they consist of map keys only. Enhanced error handling to treat missing rules as non-fatal, improving reliability during network configuration teardown.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -827,14 +827,38 @@ func delServiceRules(service *corev1.Service, localEndpoints util.PortToLBEndpoi
 		}
 		nftElems := getGatewayNFTRules(service, localEndpoints, true)
 		nftElems = append(nftElems, getGatewayNFTRules(service, localEndpoints, false)...)
-		if util.IsNetworkSegmentationSupportEnabled() {
-			nftElems = append(nftElems, getUDNNFTRules(service, nil)...)
-		}
 		if len(nftElems) > 0 {
 			if err := nodenft.DeleteNFTElements(nftElems); err != nil {
 				err = fmt.Errorf("failed to delete nftables rules for service %s/%s: %v",
 					service.Namespace, service.Name, err)
 				errors = append(errors, err)
+			}
+		}
+
+		if util.IsNetworkSegmentationSupportEnabled() {
+			// NOTE: The code below is not using nodenft.DeleteNFTElements because it first adds elements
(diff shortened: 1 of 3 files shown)


danwinship:
I can change it so that it only uses `nft.Destroy` in the UDN case. I just didn't do it that way before, because it means that you if you have a too-old nft/kernel, and you intend to use UDNs, everything will seem to be working fine until you delete a UDN, and _then_ things will fail.

npinaeva:
> I can change it so that it only uses `nft.Destroy` in the UDN case. I just didn't do it that way before, because it means that you if you have a too-old nft/kernel, and you intend to use UDNs, everything will seem to be working fine until you delete a UDN, and _then_ things will fail.

I think we could use it for UDNs only and have a `Destroy` requirement on initialization under `if util.IsNetworkSegmentationSupportEnabled()` (considering host network isolation is a required part of UDN feature and it needs a newer kernel anyway). If we go that way, I should also promote the cgroupv2 check to the same level.

The only problem is that both those changes are kinda breaking, because UDNs used to "work" with the older kernels, just kubelet probes didn't work. It could be considered a bugfix too... Let me get some more info from the d/s CI which was the first victim of this change and also ask in slack if anyone has any objections

npinaeva:
@danwinship looks like we should be fine with the UDN-only requirements, could you post an update PR please?

danwinship:
hm... actually I think that would just make things more confusing because then we'd have to explain "you can call `nodenft.DeleteElements` in case X, but not case Y, unless UDNs are enabled, in which case case Y will also work"... If we can't always rely on having `nft destroy`, then I think reverting is probably nicer. (I can move the special-case code for UDNs out of `delServiceRules` into `DeleteElements` later, but I don't have an up-to-date and tested version of that right now, so if you want to get this fixed now, then let's just merge this PR.)

danwinship:
@npinaeva I filed #6032 with an alternative approach. (It reverts the "destroy" requirement, but keeps the special-case isolated in `DeleteNFTElements` rather than reverting it back to `delServiceRules`.)

npinaeva:
alright then for the sake of time, let me merge this one and then do a review of the new PR (I think rebasing shouldn't be hard)

PR #6011, opened 2026-02-24
Add GARP drop flows for patch ports

bpickard22:
Drop GARP packets from patch ports to prevent duplicate GARPs during EgressIP failover. When an EgressIP moves between nodes, OVN-controller on the old node may have stale SBDB entries causing it to send GARPs that compete with legitimate GARPs from the new node, corrupting ARP caches.

The flow rule targets only patch port traffic (in_port=<ofport>). Legitimate GARPs sent by link manager from physical interfaces bypass OVS and are unaffected.

Workaround until OVN v26.03+ with disable_garp_rarp option is available.

Assisted-by: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network stability during Egress IP failover scenarios by preventing duplicate gratuitous ARP packets on patch ports.

* **Chores**
  * Updated network configuration accessors and gateway initialization logic to support enhanced network segmentation handling.

### DIFF
--- a/go-controller/pkg/node/bridgeconfig/bridgeconfig.go
+++ b/go-controller/pkg/node/bridgeconfig/bridgeconfig.go
@@ -53,6 +53,14 @@ func (netConfig *BridgeUDNConfiguration) IsDefaultNetwork() bool {
 	return netConfig.MasqCTMark == nodetypes.CtMarkOVN
 }
 
+func (netConfig *BridgeUDNConfiguration) GetPatchPort() string {
+	return netConfig.PatchPort
+}
+
+func (netConfig *BridgeUDNConfiguration) GetOfPortPatch() string {
+	return netConfig.OfPortPatch
+}
+
 func (netConfig *BridgeUDNConfiguration) setOfPatchPort() error {
 	ofportPatch, stderr, err := util.GetOVSOfPort("get", "Interface", netConfig.PatchPort, "ofport")
 	if err != nil {
@@ -520,6 +528,17 @@ func (b *BridgeConfiguration) SetEIPMarkIPs(eipMarkIPs *egressip.MarkIPsCache) {
 	b.mutex.Lock()
 	defer b.mutex.Unlock()
 	b.eipMarkIPs = eipMarkIPs
+}
+
+// GetNetConfigs returns the network configurations for this bridge
+func (b *BridgeConfiguration) GetNetConfigs() []*BridgeUDNConfiguration {
+	b.mutex.Lock()
+	defer b.mutex.Unlock()
+	var netConfigs []*BridgeUDNConfiguration
+	for _, netConfig := range b.netConfig {
+		netConfigs = append(netConfigs, netConfig)
+	}
+	return netConfigs
 }
 
 func (b *BridgeConfiguration) SetDropGARP(drop bool) {
--- a/go-controller/pkg/node/gateway_shared_intf.go
+++ b/go-controller/pkg/node/gateway_shared_intf.go
@@ -1742,6 +1742,7 @@ func newGateway(
 				return err
 			}
 		}
+
 		if util.IsNetworkSegmentationSupportEnabled() && config.OVNKubernetesFeature.EnableInterconnect && config.Gateway.Mode != config.GatewayModeDisabled {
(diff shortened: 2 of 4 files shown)


go-controller/pkg/node/bridgeconfig/bridgeconfig.go
```diff
@@ -497,7 +505,41 @@ func (b *BridgeConfiguration) SetNetworkOfPatchPort(netName string) error {
...
 		return fmt.Errorf("failed to find network %s configuration on bridge %s", netName, b.bridgeName)
 	}
-	return netConfig.setOfPatchPort()
+	if err := netConfig.setOfPatchPort(); err != nil {
+		return err
+	}
+
+	// Add GARP drop flow for this network's patch port (workaround for FDP-1537)
+	return b.addGARPDropFlowForNetwork(netConfig)
+}
```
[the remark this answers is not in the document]
bpickard22: imo we should be failing if we are unable to write the drop garp flow here, thats the whole point of this pr, to ensure that we are dropping the EIP Garp sent by the ovn-controller. 
This code will all be removed when we get to ovn 26.03 and can enable the disable_garp_rarp flag on the ovn router, but until then we should always drop the garp

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1742,6 +1785,19 @@ func newGateway(
 				return err
 			}
 		}
+
+		// Add GARP drop flows (workaround for FDP-1537)
```
arghosh93: Adding the GARP drop flow while setting up the gateway is very risky, considering that we are not removing that flow at any stage going forward.

go-controller/pkg/node/gateway_shared_intf.go
```diff
@@ -1742,6 +1785,19 @@ func newGateway(
...
+		if err := addGARPDropFlows(gwBridge); err != nil {
+			klog.Warningf("Failed to add GARP drop flows for bridge %s: %v",
+				gwBridge.GetBridgeName(), err)
+		}
+		if exGwBridge != nil {
+			if err := addGARPDropFlows(exGwBridge); err != nil {
```
arghosh93: I think exGwBridge corresponds to br-ex. But do we expect any GARP coming from br-ex? Ideally, it originates from the ovn-controller and corresponds to an SNAT rule set at the logical router with some specific flag.
Adding such a flow and having it forever can disrupt the networking service, which relies on neighbour learning.

go-controller/pkg/node/bridgeconfig/bridgeconfig.go
```diff
@@ -497,7 +505,41 @@ func (b *BridgeConfiguration) SetNetworkOfPatchPort(netName string) error {
...
+
+// addGARPDropFlowForNetwork adds a GARP drop flow for a single network's patch port.
+// See addGARPDropFlows() for details. Workaround for FDP-1537.
+func (b *BridgeConfiguration) addGARPDropFlowForNetwork(netConfig *BridgeUDNConfiguration) error {
```
arghosh93: It seems we have a similar implementation in place that adds this flow when the graceful shutdown of the ovnkube-controller app is triggered.  We also remove this flow when the node becomes ready after reboot, and nbdb and sbdb sync completes.
https://github.com/openshift/ovn-kubernetes/pull/2750/changes/82fc3bf7aac4f15ff7659f703c03c92947016a49#diff-3241e7827ddac578a8cd8f524a4f1723aaff1fe0633a05bb307de7d976842552R920

arghosh93 (changes requested):
I think it is risky to add GARP drop flows and have them all along. I also think that a similar implementation is in place and done via the below d/s merge.
https://github.com/openshift/ovn-kubernetes/pull/2750/changes/82fc3bf7aac4f15ff7659f703c03c92947016a49#diff-3241e7827ddac578a8cd8f524a4f1723aaff1fe0633a05bb307de7d976842552

bpickard22:
I just dont think this approach will work. We cannot have a perma drop flow to block the illegit garp, and implementing some "smart" block is just too complex and would introduce other races. Our best approach will be to implement https://github.com/openshift/cluster-network-operator/pull/2910, and then when we upgrade to ovn 26.3+ which implements the drop_garp_rarp flag, we can remove the workaround in cno. 

Thanks for review @arghosh93

PR #6016, opened 2026-02-25
kubevirt: detect completed migration when source pod is gone

qinqon:
## 📑 Description
DiscoverLiveMigrationStatus required two pods to detect a migration.
Once the source pod was removed it returned nil, causing
shouldHandleLiveMigration to be false and the target LSP Enabled field
to never be updated to true. All traffic from the target pod was then
unconditionally dropped by port security.

Teach DiscoverLiveMigrationStatus to recognize a completed migration
when only the target pod remains (or the source is completed) and it
carries the MigrationTargetReadyTimestamp annotation. Return
LiveMigrationTargetDomainReady with a nil SourcePod so the target LSP
gets properly enabled. Guard the source pod disable path against nil
SourcePod since there is nothing to disable when the source is already
gone.

Fixes #

## ✅ Checks

- [x] My code requires tests

## How to verify it
The PR includes unit tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer handling during VM live-migration: network port operations now avoid errors if the migrating source pod is absent or completed, and the system recognizes a ready migration target sooner to keep connectivity stable.
* **Tests**
  * Added and updated tests covering scenarios with missing or completed source pods and ready migration targets.

### DIFF
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -477,6 +477,14 @@ func DiscoverLiveMigrationStatus(client *factory.WatchFactory, pod *corev1.Pod) 
 
 	// no migration
 	if len(vmPods) < 2 {
+		// If the only remaining pod has the migration target ready
+		// annotation, the migration completed and the source pod is gone.
+		if len(vmPods) == 1 && isTargetPodReady(vmPods[0]) {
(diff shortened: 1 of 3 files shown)


qinqon:
> /lgtm but @qinqon need to rebase

@trozet rebased

PR #6030, opened 2026-02-26
Fix subnet masquerading SNAT when using local gateway

pacevedom:
## 📑 Description
When using local gateway combined with the node annotation `k8s.ovn.org/node-ingress-snat-exclude-subnets` (introduced in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5113), traffic destined to the excluded subnets is being masqueraded by nftables, breaking source addresses.
The nft chain `ovn-kube-pod-subnet-masq` needs to exclude those subnets to not perform SNAT.

Fixes #

## ✅ Checks

- [x] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Exclude configured subnets from local gateway SNAT processing with an early bypass for matching traffic (IPv4 and optional IPv6 in dual‑stack).

* **Tests**
  * Added end‑to‑end tests validating local gateway SNAT behavior, subnet exclusion via node annotation, nftables rule/set presence, and cleanup for both single and multiple subnets (IPv4 and conditional IPv6).

### DIFF
--- a/go-controller/pkg/node/gateway_nftables.go
+++ b/go-controller/pkg/node/gateway_nftables.go
@@ -207,6 +207,26 @@ func getUDNNFTRules(service *corev1.Service, netConfig *bridgeconfig.BridgeUDNCo
 	return rules
 }
 
+// getNoSNATSubnetsReturnRules returns nftables return rules for the pod subnet
+// masquerade chain that skip SNAT for traffic matching the no-snat subnet sets.
+// This allows preserving the original pod source IP for specific destinations.
+func getNoSNATSubnetsReturnRules() []*knftables.Rule {
+	var rules []*knftables.Rule
+	if config.IPv4Mode {
+		rules = append(rules, &knftables.Rule{
+			Chain: nftablesPodSubnetMasqChain,
+			Rule:  knftables.Concat("ip", "daddr", "@", types.NFTMgmtPortNoSNATSubnetsV4, "return"),
+		})
+	}
+	if config.IPv6Mode {
+		rules = append(rules, &knftables.Rule{
+			Chain: nftablesPodSubnetMasqChain,
+			Rule:  knftables.Concat("ip6", "daddr", "@", types.NFTMgmtPortNoSNATSubnetsV6, "return"),
+		})
+	}
+	return rules
+}
+
 // getLocalGatewayPodSubnetMasqueradeNFTRule creates rules for masquerading traffic from the pod subnet CIDR
 // in local gateway node in a separate chain which is then called from local gateway masquerade chain.
 //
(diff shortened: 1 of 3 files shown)


go-controller/pkg/node/gateway_nftables.go
```diff
@@ -477,6 +485,14 @@ func addOrUpdateLocalGatewayPodSubnetNFTRules(isAdvertisedNetwork bool, cidrs ..
 	// if network toggles between advertised and non-advertised, we need to flush the chain and re-add correct rules
 	tx.Flush(podSubnetChain)
 
+	// Skip masquerade for traffic destined to no-snat subnets
```
yboaron, line 555: minor,  consider extracting duplicate code into helper function

test/e2e/local_gateway_masquerade.go
```diff
@@ -0,0 +1,343 @@
+package e2e
```
yboaron, line 1: Can we add test that validates actual traffic behavior?  , for example verify that source IPs are actually preserved when annotation is set

test/e2e/local_gateway_masquerade.go
```diff
@@ -0,0 +1,343 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2enode "k8s.io/kubernetes/test/e2e/framework/node"
+	e2eskipper "k8s.io/kubernetes/test/e2e/framework/skipper"
+)
+
+const (
+	ovnKubePodSubnetMasqChain = "ovn-kube-pod-subnet-masq"
+	mgmtportNoSNATSubnetsV4   = "mgmtport-no-snat-subnets-v4"
```
yboaron: Maybe you can use existing constants instead of redefining

mgmtportNoSNATSubnetsV4 --> types.NFTMgmtPortNoSNATSubnetsV4
mgmtportNoSNATSubnetsV6 --> types.NFTMgmtPortNoSNATSubnetsV6
ovnNodeDontSNATSubnetsAnnotation --> util.OvnNodeDontSNATSubnets

go-controller/pkg/node/gateway_nftables.go
pacevedom: Done

test/e2e/local_gateway_masquerade.go
pacevedom: Done

test/e2e/local_gateway_masquerade.go
pacevedom: Done

yboaron (approved):
Tested it locally with Submariner deployment between two ovnk (local gateway mode) kind clusters  - looks fine.

yboaron:
Hi @pacevedom, just a heads up — since this PR has been idle for a while with some merge conflicts, I'm planning to raise an updated version based on your work. thanks for your fix

PR #6032, opened 2026-02-26
Handle the UDN cleanup special case without using `nft destroy`

danwinship:
Alternative to #6009 which is more compatible with future iptables→nftables porting plans.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NFTables map element deletion by preloading existing map contents and verifying existence before removal, avoiding unnecessary operations.

* **Tests**
  * Added comprehensive Linux-only tests covering NFTables element updates and deletions, including map elements and non-existent objects.

* **Refactor**
  * Unified UDN rule deletion into the general NFT deletion flow when Network Segmentation is enabled.

### DIFF
--- a/go-controller/pkg/node/nftables/util.go
+++ b/go-controller/pkg/node/nftables/util.go
@@ -26,18 +26,71 @@ func UpdateNFTElements(elements []*knftables.Element) error {
 
 // DeleteNFTElements deletes the given nftables set/map elements. The set or map must
 // exist, but if the elements aren't already in the set/map, no error is returned.
+//
+// To avoid depending on `nft destroy` (which requires kernel 6.3+), this Add()s each
+// element before Delete()ing it, so you won't get an error if the element wasn't already
+// in the set/map, which means that for map elements, the Value field must be set to the
+// correct value. Alternatively, you can leave Value unset, in which case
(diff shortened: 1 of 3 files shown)


danwinship:
@npinaeva rebased

npinaeva:
I think you need to remove the `gateway_shared_intf` hack again?

go-controller/pkg/node/nftables/util.go
```diff
@@ -26,18 +26,71 @@ func UpdateNFTElements(elements []*knftables.Element) error {
 
 // DeleteNFTElements deletes the given nftables set/map elements. The set or map must
 // exist, but if the elements aren't already in the set/map, no error is returned.
+//
+// If you pass map elements without the Value field set then DeleteNFTElements will need to
+// first "list" the map to find its current contents, to ensure that it only tries to
+// delete the elements that are actually present. (With set elements, or map elements with
```
npinaeva: should we also say that passing the wrong map element value will cause an error? (I think add won't change the existing value and delete will then fail?)

danwinship:
> I think you need to remove the `gateway_shared_intf` hack again?

oops

npinaeva (approved):

Issue #5952, opened 2026-02-10
BGP: Pod to external server when CUDN network is advertised Route Advertisements [no-overlay,helm]


npinaeva:
### Which jobs are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21825372127/job/62970028422?pr=5867

### Which tests are flaking?

```

Summarizing 2 Failures:
  [FAIL] BGP: Pod to external server when CUDN network is advertised Route Advertisements [It] layer3 [Feature:RouteAdvertisements]
  /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/route_advertisements.go:775
  [FAIL] BGP: Pod to external server when CUDN network is advertised Route Advertisements [It] layer2 [Feature:RouteAdvertisements]

```

### Since when has it been flaking?

Feb 10 2026

### Reason for failure (if possible)

```
[FAILED] Testing pod to external traffic failed: RunHostCmd still failed after 1m10.596437793s: error running /usr/local/bin/kubectl --server=https://127.0.0.1:46393/ --kubeconfig=/home/runner/ovn.conf --namespace=pod2external-route-advertisements-2131 exec echo-client-pod -- /bin/sh -x -c curl --max-time 10 -g -q -s http://172.26.0.3:8080/clientip:
  Command stdout:

  stderr:
  + curl --max-time 10 -g -q -s http://172.26.0.3:8080/clientip
  command terminated with exit code 28

  error:
  exit status 28: RunHostCmd still failed after 1m10.596437793s: error running /usr/local/bin/kubectl --server=https://127.0.0.1:46393/ --kubeconfig=/home/runner/ovn.conf --namespace=pod2external-route-advertisements-2131 exec echo-client-pod -- /bin/sh -x -c curl --max-time 10 -g -q -s http://172.26.0.3:8080/clientip:
  Command stdout:

  stderr:
  + curl --max-time 10 -g -q -s http://172.26.0.3:8080/clientip
  command terminated with exit code 28
```

tssurya:
I'm hitting the same flake..
@pliurh PTAL OR @zhaozhanqi 

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21921172006/job/63302516174?pr=5828 (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828)

This is specific to no-overlay lane which was added newly last week

npinaeva:
[...]

PR #6034, opened 2026-02-26
E2E: refactor container operations into reusable abstractions

pperiyasamy:
This PR extracts container and network management operations from KIND provider-specific code into shared, reusable abstractions, and adds provider factory pattern with additional E2E infrastructure improvements.                      
                                                                                                                                                                                                                                     
  Currently, container operations are tightly coupled to individual infrastructure providers, leading to:                                                                                                                              
                                                                                                                                                                                                                                     
  - Code duplication across providers
  - Difficult maintenance when fixing bugs or adding features
  - Harder on boarding for new infrastructure providers

  This refactoring addresses these issues by creating a clean abstraction layer.

  ## New Package Structure

  Created `test/e2e/infraprovider/engine/container` package with reusable components:

  - `ContainerOps` - Container lifecycle and network management operations
  - `TestContext` - Base implementation of test context with cleanup management
  - `Provider` - Base provider with common infrastructure functionality

  ## Code Movement

  - Moved Network types from kind to `test/e2e/infraprovider/engine/container/network.go`                                                                                                                                         
  - Consolidated CommandRunner interface into `common.go`                                                                                                                                                                           
  - Moved kind package to `infraprovider/internal/kind`
  - Moved KIND implementation to internal package with provider factory pattern
  - Centralized AgnHost image retrieval in deployment config

  ## Infrastructure Improvements

  - Added provider factory pattern for cleaner provider initialization
  - KIND provider now embeds `container.Provider` and `container.TestContext`, focusing only on KIND-specific logic

  ## Bug Fixes

  - Fixed NetworkCIDRsFmt() to wrap Go templates in single quotes, preventing shell interpretation issues when commands are executed over SSH
  - Go template string Name=^%s$ doesn't work with podman for retrieving container state, so replaced it with name=^%s$
  - Fixed GetIPv6Gateway returning wrong gateway
  - EVPN E2E: Use f.CreateNamespace for proper label propagation
  - EVPN E2E: Exclude default cluster network CIDRs from random subnet generation to prevent conflicts

### DIFF
--- a/test/e2e/infraprovider/api/api.go
+++ b/test/e2e/infraprovider/api/api.go
@@ -10,24 +10,44 @@ import (
 
 // Provider represents the infrastructure provider
 type Provider interface {
-	// Name returns the name of the provider, example 'kind'.
-	Name() string
+	ExternalContainerProvider
+	ClusterProvider
 	// NewTestContext providers a per test sandbox. Dependent infra test constructs are created within each test and automatically cleaned
 	// after each test.
 	NewTestContext() Context
+}
 
+type ClusterProvider interface {
 	// PrimaryNetwork returns OVN-Kubernetes primary infrastructure network information
 	PrimaryNetwork() (Network, error)
+	GetK8NodeNetworkInterface(instance string, network Network) (NetworkInterface, error)
+	// ExecK8NodeCommand executes a command on a K8 Node host network namespace and filesystem
+	ExecK8NodeCommand(nodeName string, cmd []string) (string, error)
+	// GetK8HostPort returns a Node port. Requesting a port that maybe exposed in tests to avoid multiple parallel
+	// tests utilizing conflicting ports. It also allows infra provider implementations to set Nodes
+	// allowed port range and therefore comply with cloud provider firewall rules.
+	GetK8HostPort() uint16 // supported K8 host ports
+	// ShutdownNode shuts down the specified node
+	ShutdownNode(nodeName string) error
+	// StartNode starts the specified node
+	StartNode(nodeName string) error
+	// PreloadImages pulls the given images and loads them into the cluster
+	// so that they are available to pods without a runtime pull. Providers
+	// that do not support preloading may implement this as a no-op.
+	PreloadImages(images []string)
+	// Name returns the name of the cluster provider, example 'kind'.
+	Name() string
+	// Get platform specific timeout values
+	GetDefaultTimeoutContext() *framework.TimeoutContext
+}
+
+type ExternalContainerProvider interface {
 	// ListNetworks returns the names of all networks
 	ListNetworks() ([]string, error)
 	// GetNetwork returns a network
 	GetNetwork(name string) (Network, error)
 	// GetExternalContainerNetworkInterface fetches network interface information from the external container attached to a specific network
 	GetExternalContainerNetworkInterface(container ExternalContainer, network Network) (NetworkInterface, error)
-	GetK8NodeNetworkInterface(instance string, network Network) (NetworkInterface, error)
-
-	// ExecK8NodeCommand executes a command on a K8 Node host network namespace and filesystem
-	ExecK8NodeCommand(nodeName string, cmd []string) (string, error)
 	ExecExternalContainerCommand(container ExternalContainer, cmd []string) (string, error)
 	GetExternalContainerLogs(container ExternalContainer) (string, error)
 	// GetExternalContainerPort returns a port. Requesting a port that maybe exposed in tests to avoid multiple parallel
@@ -35,23 +55,6 @@ type Provider interface {
 	// allowed port range and therefore comply with cloud provider firewall rules.
 	GetExternalContainerPort() uint16
 	ExternalContainerPrimaryInterfaceName() string
-	// GetK8HostPort returns a Node port. Requesting a port that maybe exposed in tests to avoid multiple parallel
-	// tests utilizing conflicting ports. It also allows infra provider implementations to set Nodes
-	// allowed port range and therefore comply with cloud provider firewall rules.
-	GetK8HostPort() uint16 // supported K8 host ports
-
-	// ShutdownNode shuts down the specified node
-	ShutdownNode(nodeName string) error
-	// StartNode starts the specified node
-	StartNode(nodeName string) error
-
-	// Get platform specific timeout values
-	GetDefaultTimeoutContext() *framework.TimeoutContext
-
-	// PreloadImages pulls the given images and loads them into the cluster
-	// so that they are available to pods without a runtime pull. Providers
-	// that do not support preloading may implement this as a no-op.
-	PreloadImages(images []string)
 }
 
 // Underlay represents the configuration for an underlay network.
@@ -70,16 +73,23 @@ type Underlay struct {
 }
 
 type Context interface {
-	CreateExternalContainer(container ExternalContainer) (ExternalContainer, error)
-	DeleteExternalContainer(container ExternalContainer) error
+	ExternalContainerContextProvider
+	ClusterContextProvider
+	AddCleanUpFn(func() error)
+}
 
+type ExternalContainerContextProvider interface {
 	CreateNetwork(name string, subnets ...string) (Network, error)
 	DeleteNetwork(network Network) error
+	CreateExternalContainer(container ExternalContainer) (ExternalContainer, error)
+	DeleteExternalContainer(container ExternalContainer) error
 	AttachNetwork(network Network, instance string) (NetworkInterface, error)
 	DetachNetwork(network Network, instance string) error
-	SetupUnderlay(f *framework.Framework, underlay Underlay) error
+}
 
-	AddCleanUpFn(func() error)
+type ClusterContextProvider interface {
+	// anything done on the cluster that could be cleaned up after test
+	SetupUnderlay(f *framework.Framework, underlay Underlay) error
 }
 
 type Network interface {


test/e2e/containerengine/container_engine.go
```diff
@@ -14,7 +14,7 @@ func (ce ContainerEngine) String() string {
 
 func (ce ContainerEngine) NetworkCIDRsFmt() string {
 	if ce == Podman {
-		return "{{json .Subnets }}"
+		return "'{{json .Subnets }}'"
```
qinqon: This adds literal single-quote characters to the format string. Since `CommandRunner.Run()` uses `exec.Command` (no shell), these quotes are passed verbatim to Podman's `--format` arg. The output will be `'[{"Subnet":...}]'` which will fail `json.Unmarshal` in `GetNetwork()` at `container_ops.go:335`.

The Docker branch doesn't wrap in quotes. Revert to:
```go
return "{{json .Subnets }}"
```

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,567 @@
...
+}
+
+func (o *ContainerOps) IsNetworkAttachedToContainer(networkName, containerName string) bool {
+	// error is returned if failed to find network attached to instance or no IPv4/IPv6 Ips.
+	_, err := o.GetNetworkInterface(containerName, networkName)
+	if err != nil {
+		return false
+	}
+	return true
+}
```
qinqon, line 165: `IsNetworkAttachedToContainer` swallows all errors as "not attached". If `GetNetworkInterface` fails for a transient reason (e.g. command timeout), `ConnectNetwork` will attempt a redundant connect and `DisconnectNetwork` will silently skip the disconnect.

Change signature to `(bool, error)` and propagate errors in callers.

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -44,417 +40,72 @@ func IsProvider() bool {
 }
 
 type kind struct {
-	externalContainerPort *portalloc.PortAllocator
-	hostPort              *portalloc.PortAllocator
+	base.Provider
+	infraprovider.CommandRunner
```
qinqon: Embedding the `CommandRunner` interface here is unnecessary — `kind` already implements `Run()` directly at line 60. The embedded interface field is never assigned and just shadows the concrete method. Remove it.

test/e2e/containerengine/container_engine.go
```diff
@@ -14,7 +14,7 @@ func (ce ContainerEngine) String() string {
...
 	}
```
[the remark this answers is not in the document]
pperiyasamy: this change is intended, this would make `podman network inspect -f ...` command to work event if it runs directly or ssh. it also doesn't produce output with single quoted string (see below).
```
$ sudo podman network inspect -f '{{json .Subnets }}' podman
[{"subnet":"10.88.0.0/16","gateway":"10.88.0.1"}]
```

pperiyasamy:
> **Actionable comments posted: 1**
> 
> ♻️ Duplicate comments (1)
> > test/e2e/containerengine/container_ops.go (1)> `112-133`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Propagate attachment-check errors instead of collapsing them to `false`.**
> > > `ConnectNetwork` / `DisconnectNetwork` currently treat infra failures as “not attached” because `IsNetworkAttachedToContainer` (Lines 141-148) drops all errors. This can trigger incorrect network mutations under transient runner/inspect failures.
> > > 
> > > 💡 Proposed fix
> > > ```diff
> > >  import (
> > >  	"context"
> > > +	"errors"
> > >  	"encoding/json"
> > >  	"fmt"
> > > @@
> > >  func (o *ContainerOps) ConnectNetwork(network api.Network, container string) error {
> > > @@
> > > -	if o.IsNetworkAttachedToContainer(network.Name(), container) {
> > > +	attached, err := o.IsNetworkAttachedToContainer(network.Name(), container)
> > > +	if err != nil {
> > > +		return fmt.Errorf("failed to check network attachment (%s -> %s): %w", network.Name(), container, err)
> > > +	}
> > > +	if attached {
> > >  		return fmt.Errorf("network %s is already attached to container %s", network.Name(), container)
> > >  	}
> > > @@
> > >  func (o *ContainerOps) DisconnectNetwork(network api.Network, container string) error {
> > > @@
> > > -	if !o.IsNetworkAttachedToContainer(network.Name(), container) {
> > > +	attached, err := o.IsNetworkAttachedToContainer(network.Name(), container)
> > > +	if err != nil {
> > > +		return fmt.Errorf("failed to check network attachment (%s -> %s): %w", network.Name(), container, err)
> > > +	}
> > > +	if !attached {
> > >  		return nil
> > >  	}
> > > @@
> > > -func (o *ContainerOps) IsNetworkAttachedToContainer(networkName, containerName string) bool {
> > > +func (o *ContainerOps) IsNetworkAttachedToContainer(networkName, containerName string) (bool, error) {
> > >  	// error is returned if failed to find network attached to instance or no IPv4/IPv6 Ips.
> > >  	_, err := o.GetNetworkInterface(containerName, networkName)
> > >  	if err != nil {
> > > -		return false
> > > +		if errors.Is(err, api.NotFound) {
> > > +			return false, nil
> > > +		}
> > > +		return false, err
> > >  	}
> > > -	return true
> > > +	return true, nil
> > >  }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 

I don't want to make this change, because we also need to ignore the error when container is not set with ip address. it's the same logic what was there eariler and just being moved into `container_ops.go` module.

> > > Also applies to: 141-148
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@test/e2e/containerengine/container_ops.go` around lines 112 - 133,
> > > IsNetworkAttachedToContainer currently swallows errors causing
> > > ConnectNetwork/DisconnectNetwork to assume "not attached" on infra failures;
> > > change IsNetworkAttachedToContainer to return (bool, error) instead of bool,
> > > update its implementation to return the error from the inspect/runner call, and
> > > update callers ConnectNetwork and DisconnectNetwork to check and propagate that
> > > error (e.g. if err != nil return fmt.Errorf("failed to check if network %s is
> > > attached to container %s: %w", network.Name(), container, err)) before deciding
> > > to attach/detach so transient runner/inspect failures don't collapse to false.
> > > ```
> 
> 🧹 Nitpick comments (1)
> > test/e2e/infraprovider/base/provider.go (1)> `9-13`: **Guard allocator access to avoid nil-pointer panics in future provider reuse.**
> > > `GetExternalContainerPort` and `GetK8HostPort` assume allocators are always initialized. That holds in current `kind.New`, but this base type is intended for reuse and can panic if another provider forgets initialization.
> > > 
> > > 💡 Proposed hardening
> > > ```diff
> > >  func (p *Provider) GetExternalContainerPort() uint16 {
> > > +	if p.ExternalContainerPort == nil {
> > > +		panic("ExternalContainerPort allocator is not initialized")
> > > +	}
> > >  	return p.ExternalContainerPort.Allocate()
> > >  }
> > >  
> > >  func (p *Provider) GetK8HostPort() uint16 {
> > > +	if p.HostPort == nil {
> > > +		panic("HostPort allocator is not initialized")
> > > +	}
> > >  	return p.HostPort.Allocate()
> > >  }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > Also applies to: 35-40
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@test/e2e/infraprovider/base/provider.go` around lines 9 - 13, The Provider
> > > methods GetExternalContainerPort and GetK8HostPort currently assume
> > > Provider.ExternalContainerPort and Provider.HostPort are non-nil; add nil guards
> > > in each method (in type Provider) to avoid panics: detect if
> > > ExternalContainerPort/HostPort is nil and return a clear error (or lazily
> > > initialize the allocator if the codebase prefers) instead of dereferencing;
> > > update both GetExternalContainerPort and GetK8HostPort to check the
> > > corresponding field, return a descriptive error referencing
> > > Provider.ExternalContainerPort or Provider.HostPort, and ensure callers handle
> > > the error (also add/update tests to cover the nil-allocator case).
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> Inline comments:
> In `@test/e2e/containerengine/container_engine.go`:
> - Around line 15-18: The Podman format string returned by
> ContainerEngine.NetworkCIDRsFmt() includes literal single quotes, causing Podman
> to return quoted JSON that later breaks json.Unmarshal in GetNetwork; remove the
> surrounding single quotes from the Podman branch (i.e., change the Podman return
> in NetworkCIDRsFmt to "{{json .Subnets }}") so the raw JSON is passed to
> GetNetwork (which expects unquoted JSON); ensure this mirrors the behavior
> handled correctly by GetNetworkInterface and update any related tests if they
> assume quoted output.
> 
> ---
> 
> Duplicate comments:
> In `@test/e2e/containerengine/container_ops.go`:
> - Around line 112-133: IsNetworkAttachedToContainer currently swallows errors
> causing ConnectNetwork/DisconnectNetwork to assume "not attached" on infra
> failures; change IsNetworkAttachedToContainer to return (bool, error) instead of
> bool, update its implementation to return the error from the inspect/runner
> call, and update callers ConnectNetwork and DisconnectNetwork to check and
> propagate that error (e.g. if err != nil return fmt.Errorf("failed to check if
> network %s is attached to container %s: %w", network.Name(), container, err))
> before deciding to attach/detach so transient runner/inspect failures don't
> collapse to false.
> 
> ---
> 
> Nitpick comments:
> In `@test/e2e/infraprovider/base/provider.go`:
> - Around line 9-13: The Provider methods GetExternalContainerPort and
> GetK8HostPort currently assume Provider.ExternalContainerPort and
> Provider.HostPort are non-nil; add nil guards in each method (in type Provider)
> to avoid panics: detect if ExternalContainerPort/HostPort is nil and return a
> clear error (or lazily initialize the allocator if the codebase prefers) instead
> of dereferencing; update both GetExternalContainerPort and GetK8HostPort to
> check the corresponding field, return a descriptive error referencing
> Provider.ExternalContainerPort or Provider.HostPort, and ensure callers handle
> the error (also add/update tests to cover the nil-allocator case).
> ```
> 
> ℹ️ Review info

The `ExternalContainerPort` and `HostPort` are always initialized when provider is created, so don't need for nil check. ignoring this comment.

pperiyasamy:
@coderabbitai responded to your comments. ptal.

test/e2e/containerengine/container_engine.go
tssurya: what's this single quote insertion for?

test/e2e/containerengine/container_network.go
```diff
@@ -0,0 +1,58 @@
+package containerengine
+
+import (
+	"fmt"
+
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider/api"
+	"k8s.io/utils/net"
+)
+
+type Network struct {
```
tssurya: any reason why we didn't just rename `providers/kind/network.go` -> to this file name and changed the struct from `containerEngineNetwork` to `Network` ? the `containerEngineNetwork` seems to fit with the `containerEngine` element inside containerengine folder

test/e2e/infraprovider/engine/container/network/network.go
```diff
@@ -0,0 +1,58 @@
...
+	Gateway string `json:"Gateway"`
+}
+
+func (n Network) Name() string {
+	return n.NetName
+}
+
+func (n Network) IPv4IPv6Subnets() (string, string, error) {
+	if len(n.Configs) == 0 {
+		return "", "", fmt.Errorf("network %s has no CIDR configured", n.NetName)
```
tssurya, line 26: why were panics replaced with returns? (asking out of curiosity)

test/e2e/infraprovider/base/container_ops.go
```diff
@@ -0,0 +1,567 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/images"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider"
```
tssurya: isn't this a bit weird of a dependency we are introducing?
and I guess the infraprovider is also importing the containerengine?

test/e2e/containerengine/container_ops.go
```diff
@@ -0,0 +1,567 @@
...
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/images"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider/api"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kubernetes/test/e2e/framework"
+	utilnet "k8s.io/utils/net"
+)
+
+const (
+	nameFormat                     = "{{.Name}}"
```
tssurya: please add some doc strings - what are these constants going to be used for?

test/e2e/infraprovider/command_runner.go
```diff
@@ -0,0 +1,6 @@
+package infraprovider
+
+// CommandRunner executes docker/podman commands
+type CommandRunner interface {
```
tssurya: hmm do we need a separate file just for this?

test/e2e/containerengine/container_network.go
pperiyasamy: i didn't want other providers to reuse network.go from kind package, that's the reason moved out from here, or shall i move it under `infraprovider/network.go` ?

test/e2e/infraprovider/engine/container/context.go
```diff
@@ -0,0 +1,127 @@
...
+
+// CleanUp must be synchronized by caller
+func (c *TestContext) cleanUp() error {
+	var errs []error
+	// generic cleanup activities
+	for i := len(c.CleanUpFns) - 1; i >= 0; i-- {
+		if err := c.CleanUpFns[i](); err != nil {
+			errs = append(errs, err)
+		}
+	}
```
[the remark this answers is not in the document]
pperiyasamy: this was just copied from kind provider into base test context, so i prefer not to make this change, because E2Es are already passing with it.

test/e2e/infraprovider/internal/kind/kind.go
```diff
@@ -44,417 +40,72 @@ func IsProvider() bool {
...
-		return "", fmt.Errorf("cannot exec into container %q because it doesn't exist: %w", container.Name, api.NotFound)
-	}
-	cmdArgs := append([]string{"exec", container.Name}, cmd...)
-	out, err := exec.Command(containerengine.Get().String(), cmdArgs...).CombinedOutput()
-	if err != nil {
-		return "", fmt.Errorf("failed to exec container command (%s): err: %v, stdout: %q", strings.Join(cmdArgs, " "), err, out)
+		return "", fmt.Errorf("command failed: %w, output: %s", err, out)
 	}
 	return string(out), nil
 }
```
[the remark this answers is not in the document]
pperiyasamy: let E2Es add a timeout guard for command execution, so don't want to do it here.

pperiyasamy:
@coderabbitai are you fine with my response ?

test/e2e/infraprovider/engine/container/context.go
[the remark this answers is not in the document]
pperiyasamy: @coderabbitai ^

test/e2e/infraprovider/internal/kind/kind.go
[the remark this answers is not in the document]
pperiyasamy: @coderabbitai ^

test/e2e/infraprovider/base/container_ops.go
pperiyasamy: you are right, moved it under infraprovider base package, i think that is the right place for this. makes sense ?

test/e2e/infraprovider/engine/container/network/network.go
pperiyasamy: doesn't want E2E process to fail with a panic, let test fail and move onto next test..

test/e2e/infraprovider/command_runner.go
pperiyasamy: yes, not required, moved the interface into `infraprovider/provider.go`.

test/e2e/containerengine/container_ops.go
pperiyasamy: done.

test/e2e/containerengine/container_engine.go
pperiyasamy: added a comment to explain this.

test/e2e/containerengine/container_network.go
tssurya: no its fine being under containerengine,
I think q was why is it called `Network` instead of `ContainerEngineNetwork`

tssurya: I guess overall we want to see what belongs in base provider v/s what belongs in containerengine and stick with that format

pperiyasamy: this network may also represent a host network and that may not be a created via container runtime.

test/e2e/infraprovider/providers/kind/kind.go
```diff
@@ -44,417 +40,72 @@ func IsProvider() bool {
...
-			}
-		}
-		stdOut, err := exec.Command(containerengine.Get().String(), "network", "rm", network.Name()).CombinedOutput()
-		if err != nil {
-			framework.Logf("failed to delete network %s: %s (%s)", network.Name(), err, stdOut)
-			return false, nil
-		}
-		return true, nil
-	})
+func (c *contextKind) GetExternalContainerImage() string {
```
pperiyasamy: self note: this function has to be removed.

test/e2e/infraprovider/provider.go
```diff
@@ -6,6 +6,11 @@ import (
 
 type Name string
 
+// CommandRunner executes docker/podman commands
```
tssurya: nit: i think we should not mention podman or docker explicitly no? i think goal was to support any abstraction but not a big deal..

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,592 @@
...
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/containerengine"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/images"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider"
+	"github.com/ovn-org/ovn-kubernetes/test/e2e/infraprovider/api"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kubernetes/test/e2e/framework"
+	utilnet "k8s.io/utils/net"
+)
+
+const (
```
qinqon, line 21: Let's replace all this docker inspect commands with just one tha returns all the json and you parse it with the struct, you going to also being able to have same thing for docker and podman:

  1. Container inspect — 7 calls → 1                                                                                                                                                                               
                                                                                                                                                                                                                   
  Currently (lines 190-298 of container_ops.go), each field is a separate CLI call:


```golang
  // 7 separate calls like:
  o.CmdRunner.Run("inspect", "-f", "'{{ with index .NetworkSettings.Networks \"mynet\" }}{{ .IPAddress }}{{ end }}'", containerName)
  o.CmdRunner.Run("inspect", "-f", "'{{ with index .NetworkSettings.Networks \"mynet\" }}{{ .Gateway }}{{ end }}'", containerName)
  o.CmdRunner.Run("inspect", "-f", "'{{ with index .NetworkSettings.Networks \"mynet\" }}{{ .MacAddress }}{{ end }}'", containerName)
  // ... 4 more
```

  With structs:
```golang
  // One call, one unmarshal
  type ContainerInspect struct {
      NetworkSettings struct {
          Networks map[string]EndpointSettings `json:"Networks"`
      } `json:"NetworkSettings"`
  }

  type EndpointSettings struct {
      Gateway             string `json:"Gateway"`
      IPAddress           string `json:"IPAddress"`
      IPPrefixLen         int    `json:"IPPrefixLen"`
      IPv6Gateway         string `json:"IPv6Gateway"`
      GlobalIPv6Address   string `json:"GlobalIPv6Address"`
      GlobalIPv6PrefixLen int    `json:"GlobalIPv6PrefixLen"`
      MacAddress          string `json:"MacAddress"`
  }

  func (o *ContainerOps) GetNetworkInterface(containerName, networkName string) (api.NetworkInterface, error) {
      out, err := o.CmdRunner.Run("inspect", containerName)
      // docker inspect returns a JSON array
      var inspectResult []ContainerInspect
      json.Unmarshal([]byte(out), &inspectResult)

      ep := inspectResult[0].NetworkSettings.Networks[networkName]
      ni := api.NetworkInterface{
          IPv4:        ep.IPAddress,
          IPv4Prefix:  strconv.Itoa(ep.IPPrefixLen),
          IPv4Gateway: ep.Gateway,
          IPv6:        ep.GlobalIPv6Address,
          IPv6Prefix:  strconv.Itoa(ep.GlobalIPv6PrefixLen),
          IPv6Gateway: ep.IPv6Gateway,
          MAC:         ep.MacAddress,
      }
      // InfName still needs the `ip -br a` exec (not in docker inspect)
      ...
  }
```
  2. Network inspect — eliminates Docker/Podman template divergence

  Currently GetNetwork() needs NetworkCIDRsFmt() which returns different Go templates per engine. GetContainersAttachedToNetwork() and DeleteNetwork() also use templates.

```golang
  // Works for both Docker and Podman
  type NetworkInspect struct {
      Name       string                          `json:"Name"`
      IPAM       IPAMConfig                      `json:"IPAM"`
      Subnets    []NetworkConfig                 `json:"subnets,omitempty"` // Podman
      Containers map[string]NetworkContainerInfo  `json:"Containers"`
  }

  type IPAMConfig struct {
      Config []NetworkConfig `json:"Config"`
  }

  type NetworkContainerInfo struct {
      Name string `json:"Name"`
  }
```

  Then GetNetwork() becomes:
```golang
  func (o *ContainerOps) GetNetwork(networkName string) (infraprovider.Network, error) {
      out, err := o.CmdRunner.Run("network", "inspect", networkName)
      var result []NetworkInspect
      json.Unmarshal([]byte(out), &result)

      // Normalize Docker vs Podman
      configs := result[0].IPAM.Config
      if len(configs) == 0 {
          configs = result[0].Subnets // Podman
      }
      ...
  }
```
  And GetContainersAttachedToNetwork() / DeleteNetwork() just iterate result[0].Containers instead of parsing template output with string splitting and quote trimming.

test/e2e/containerengine/container_network.go
pperiyasamy: i just realized NetworkConfig is tied to containerengine construct, so moving network module back into containerengine module.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: now doing with single inspect command for container and network.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: done.

test/e2e/infraprovider/providers/kind/kind.go
pperiyasamy: you are right, removed `CommandRunner` interface from `kind` struct.

test/e2e/containerengine/container_engine.go
pperiyasamy: this is not needed anymore with new `nework inspect` command implmentation, removed it.
thanks for the great feedback.

tssurya:
@qinqon great review comments! :D PR is looking much better now, I'll also take a pass now..
the containerinspect idea is something i wanted to suggest as well you got there before me!

test/e2e/containerengine/container_network.go
tssurya: FTR I wanted it to be under container_network.go under containerengine but I was just curious why it didn't get moved as is and why the struct became generic as `Network`

> this network may also represent a host network and that may not be a created via container runtime.

can you shed some light on what this means? you mean could be a node or a VM? if so please leave a comment because given its under the containerengine folder its better suited to represent it as a containerNetworkEngine struct. are we using it today to present a host network? - do you mean the equivalent of the OCP node?

test/e2e/containerengine/container_network.go
```diff
@@ -0,0 +1,75 @@
...
+	Subnets    []NetworkConfig                 `json:"subnets,omitempty"` // Podman
+	Containers map[string]NetworkContainerInfo `json:"Containers"`
+}
+
+type IPAMConfig struct {
+	Config []NetworkConfig `json:"Config"`
+}
+
+type NetworkContainerInfo struct {
+	Name string `json:"Name"`
```
tssurya: are we planning to add more things here? why isn't this just represented as inline value for now till we have the need to wanting a struct?

test/e2e/containerengine/container_engine.go
```diff
@@ -6,22 +6,30 @@ import (
 	"strings"
 )
 
+// ContainerInspect represents the JSON output from 'docker/podman inspect'
```
tssurya: nit: move this to end of file and let containerengine be the starting point..its a bit weird the containerinspect is same file as containerengine but i think its also a overkill to do a new file called `container.go` until we have more things to put.
one suggestion is to rename this file to `container.go` that has both container engine and container inspect structs but this is a huge nit and doesn't really matter if we keep things as is

test/e2e/infraprovider/base/common.go
```diff
@@ -0,0 +1,574 @@
...
+const (
+	// nameFormat is a Go template for extracting network names from inspect output.
+	nameFormat = "{{.Name}}"
+)
+
+// ContainerOps provides reusable helper methods for container lifecycle and network management.
+// It can be embedded by infrastructure providers to inherit common container operations.
+type ContainerOps struct {
+	// Engine is the container engine (docker/podman) that commands are executed against.
+	Engine containerengine.ContainerEngine
```
tssurya: is the engine value inside containerops actually used anywhere? in the kind.go provider i see it being set but i don't see where its used
i think you have an engine string in kind struct already

test/e2e/infraprovider/base/common.go
```diff
@@ -0,0 +1,574 @@
...
+	// CmdRunner executes container engine commands and returns their output.
+	CmdRunner infraprovider.CommandRunner
```
tssurya: does the commandrunner need to be in the infraprovider OR can it just be here in the base provider? -i.e is it an implementation util OR an API util

test/e2e/infraprovider/engine/container/network/network.go
```diff
@@ -0,0 +1,75 @@
+package containerengine
+
+import (
+	"fmt"
+
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/api"
```
tssurya, line 6: im still a bit confused around the goal of the packages.

1. what is the goal of containerengine package?
2. what is the goal of infraprovider base package?

how are you deciding what goes into containerengine v/s base provider? - it seems like some things got moved from kind.go into container engine but somethings went into base.
and given this container_network file is importing the infraprovider api then having
container_engine.go
container.go (that has container inspect and container ops)
container_network.go
makes sense which is what you had before but now removed it.
containerengine isn't a clean leaf package anymore since its tied to infraprovider which is OK but I see a lack of clarity from maintainence standpoint on when to add something to containerengine packet v/s when to add someone to base provider package for an implementation.
i.e should whole of container ops belong to containerengine or base provider?

pperiyasamy: as we discussed offline, have made directory structure as below and removed existing `containeregine` package.
```
- infraprovider
  - internal
    - command
      - runner.go 
    - engine
    -  container
       - common.go
       - context.go
       - engine.go
       - network.go
       - provider.go 
  - kind
  - <other provider>
```

test/e2e/infraprovider/base/common.go
pperiyasamy: yes, this is no longer needed now, missed to remove it from my previous pr update.

test/e2e/containerengine/container_engine.go
pperiyasamy: yes, removed it from here, and let these `engine.go` and `network.go` contain what is required to have.
moved inspect specific structs into the end of common.go file.

test/e2e/containerengine/container_network.go
pperiyasamy: we are currently interested only retrieving attached containers with network, so inspect output is like:
```
$ sudo podman network inspect agnhost_net
[
     {
          "name": "agnhost_net",
          "id": "d6f6c226a857d568a6970cfbcf3c8f5651bf6230f64364b889ffd2d6b50e03cd",
          "driver": "macvlan",
          "network_interface": "dummy0",
          "created": "2026-03-03T07:06:52.400417487-05:00",
          "ipv6_enabled": true,
          "internal": false,
          "dns_enabled": false,
          "ipam_options": {
               "driver": "none"
          },
          "containers": {
               "436fb98516d08d07e546beae8cfaba7c41d384f3037cada4e7e24679281bc6cd": {
                    "name": "agnhost",
                    "interfaces": {
                         "eth0": {
                              "mac_address": "5e:73:f8:3c:61:c4"
                         }
                    }
               },
               "5cb673b45bd60c0cccc9c4da42014ea639bb9d7d2d71311a535b06c692d7d49f": {
                    "name": "frr",
                    "interfaces": {
                         "eth1": {
                              "mac_address": "b2:67:51:96:14:6d"
                         }
                    }
               }
          }
     }
]
```
so do you want to use inline struct like below ?
```
type NetworkInspect struct {
	Name       string                         `json:"Name"`
	IPAM       IPAMConfig                     `json:"IPAM"`              // Docker
	Subnets    []ContainerEngineNetworkConfig `json:"subnets,omitempty"` // Podman
	Containers map[string]struct {
		Name string `json:"Name"`
	} `json:"Containers"`
}
```

test/e2e/infraprovider/provider.go
pperiyasamy: done

test/e2e/infraprovider/base/common.go
pperiyasamy: this is implementation util (providing common container operations) being embedded into base provider and test context. having `command.Runner` in the struct here because util functions need not be called every time passing `command.Runner` .

test/e2e/containerengine/container_network.go
pperiyasamy: yes, this is what i had in my mind, it can also represent a network not only created by container engine, but we are not there yet, so let's stick with referring to `ContainerEngineNetwork` for now. we could revisit later when needed.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy, line 1: @jcaamano @tssurya i think we must remove `internal` from package because modules from this package need to be consumed from other providers as well.

tssurya: hmm so what you mean is: https://pkg.go.dev/cmd/go#hdr-Internal_Directories

Concretely, if the package were at:
test/e2e/infraprovider/internal/engine/container
then only code rooted under test/e2e/infraprovider/ could import it. That means:
test/e2e/infraprovider/providers/kind/ -- can import (it's under infraprovider/)
test/e2e/infraprovider/providers/openshift/ -- can import (same) - but in our case this is in parent folder and not under infraprovider right? https://github.com/openshift/ovn-kubernetes/tree/master/openshift
But anything outside test/e2e/infraprovider/ (e.g. test/e2e/ tests themselves) -- cannot import
So your point is valid -- if the types from this package (like base.Provider, base.TestContext, base.ContainerOps) need to be consumed by other providers and potentially by code outside the infraprovider/ subtree, then using internal would be too restrictive.
I'm fine to remove `internal` , let's see what Jaime says since he is the one who proposed the `internal` naming

test/e2e/containerengine/container_engine.go
```diff
@@ -1,4 +1,4 @@
-package containerengine
```
tssurya, line 1: does this mean the `containerengine` folder also got deleted? because that's not showing up in git diff, just making sure its gone and not hanging with 0 files in it

test/e2e/infraprovider/internal/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+
+// ContainerOps provides reusable helper methods for container lifecycle and network management.
+// It can be embedded by infrastructure providers to inherit common container operations.
+type ContainerOps struct {
+	// CmdRunner executes container engine commands and returns their output.
+	CmdRunner command.Runner
+}
+
+// AddNetwork creates a network using the provided command runner
+func (o *ContainerOps) AddNetwork(name string, subnets ...string) error {
```
tssurya: I prefer the original name `CreateNetwork`.
Why did we rename this?

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,606 @@
...
+			v6 = true
+		}
+	}
+	if v6 {
+		cmdArgs = append(cmdArgs, "--ipv6")
+	}
+	stdOut, err := o.CmdRunner.Run(cmdArgs...)
+	if err != nil {
+		return fmt.Errorf("failed to create Network with command %q: %s (%s)", strings.Join(cmdArgs, " "), err, stdOut)
+	}
```
tssurya, line 69: what happened to the cleanup Func style things we had in KIND ?
I think you need to update your PR describe with the latest details
I understand it all got moved to TestContext somewhere but does this change anything for the caller in how cleanup was executed?

test/e2e/infraprovider/internal/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	var containers []string
+	for _, container := range result[0].Containers {
+		containers = append(containers, container.Name)
+	}
+	return containers, nil
+}
+
+// ConnectNetwork attaches given network with container
+func (o *ContainerOps) ConnectNetwork(network api.Network, container string) error {
```
tssurya: I prefer `AttachNetwork` itself...or maybe there is a reason you are changing function names in this change and I don't understand yet

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,606 @@
...
+	exists, err := o.NetworkExists(network.Name())
+	if err != nil {
+		return fmt.Errorf("failed to check if network %s exists: %w", network.Name(), err)
+	}
+	if !exists {
+		return nil
+	}
+	attached, err := o.IsNetworkAttachedToContainer(network.Name(), container)
+	if !attached {
+		return err
```
tssurya, line 144: we are supposed to return nil here right? cause nothing to deattach..? and same should be called deattachNetwork...
can you please double check the existing logic in kind for each util to ensure behaviour change for a util doesn't happen semantically speaking?

test/e2e/infraprovider/internal/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	return nil
+}
+
+func (o *ContainerOps) IsNetworkAttachedToContainer(networkName, containerName string) (bool, error) {
+	// Return false for api.NotFound errors (e.g., when the container,
+	// network, or network attachment is not found, or when no IPv4/IPv6
+	// addresses are assigned).
+	_, err := o.GetNetworkInterface(containerName, networkName)
+	if err != nil && errors.Is(err, api.NotFound) {
```
tssurya: just make this like normally how we do it?

if err!=nil {
  if errors.Is(err, api.NotFound) {
    return false, nil
  }
  return false, err
}
return true nil

? I'm not sure why we changed function signature here as well to return the error - I think your PR is also changing the utils implementation compared to KIND's definitions and that's dangerous if we do semantic changes per util and not just "make these utils common"

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	var result []NetworkInspect
+	err = json.Unmarshal([]byte(out), &result)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal network %s inspect output: %w", networkName, err)
+	}
+	if len(result) == 0 {
+		return nil, api.NotFound
+	}
+	var containers []string
```
tssurya, line 101: we don't want to undupe?
```
	primaryNetwork, err := getNetwork("kind")
	if err != nil {
		return api.Networks{}, fmt.Errorf("failed to get primary network: %v", err)
	}
	attachedNetworks := api.Networks{List: []api.Network{primaryNetwork}}
	for _, attachment := range c.cleanUpNetworkAttachments.List {
		attachedNetworks.InsertNoDupe(attachment.Network)
	}
	return attachedNetworks, nil
```
was the original implementation..

test/e2e/infraprovider/internal/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+		IPv4Prefix:  "",
+		IPv4Gateway: ep.Gateway,
+		IPv6:        ep.GlobalIPv6Address,
+		IPv6Prefix:  "",
+		IPv6Gateway: ep.IPv6Gateway,
+		MAC:         ep.MacAddress,
+	}
+	// Only set prefix if we have an IP and a valid prefix length
+	if ni.IPv4 != "" {
+		if ep.IPPrefixLen > 0 && ep.IPPrefixLen <= 32 {
```
tssurya: nit: for readability reasons I would say return for the error to avoid so much if else nesting

test/e2e/infraprovider/internal/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+	return nil
+}
+
+// NetworkExists checks if a network exists
+func (o *ContainerOps) NetworkExists(networkName string) (bool, error) {
+	dataBytes, err := o.CmdRunner.Run("network", "ls", "--format", nameFormat)
+	if err != nil {
+		return false, fmt.Errorf("failed to list networks: %w", err)
+	}
+	for _, existingNetworkName := range strings.Split(strings.Trim(dataBytes, "\n"), "\n") {
```
tssurya: func doesNetworkExist(networkName string) (bool, error) {
	networks, err := listNetworks()
	if err != nil {
		return false, err
	}
	for _, name := range networks {
		if name == networkName {
			return true, nil
		}
	}

was the older one.. so I hope this trimming forom database is accurate and won't break anything.. seems sligtly flimsy compared to original approach

test/e2e/infraprovider/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+	stdOut, err := o.CmdRunner.Run("ps", "-a", "-f", fmt.Sprintf("name=^%s$", containerName), "--format", "{{.State}}")
+	if err != nil {
+		return "", fmt.Errorf("failed to check container state for %s: %s (%s)", containerName, err, stdOut)
+	}
+
+	state := strings.TrimSpace(stdOut)
+	return state, nil
+}
+
+func (o *ContainerOps) GetNetwork(networkName string) (ContainerEngineNetwork, error) {
```
tssurya: oh so this GetNetwork is our struct's representation? not the docker network? that's confusing we switched mid way :D please add a doc string

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	var result []NetworkInspect
+	err = json.Unmarshal([]byte(out), &result)
+	if err != nil {
+		return n, fmt.Errorf("failed to unmarshal network %s inspect output: %w", networkName, err)
+	}
+	if len(result) == 0 {
+		return n, fmt.Errorf("network %q inspect returned empty result: %w", networkName, api.NotFound)
+	}
+	// Normalize Docker vs Podman
```
tssurya, line 359: I guess when another provider is added this needs to changed again?
I think this something that should live in the providers right? i.e ocp v/s kind? or are we saying docker/podman are our supported internal engine containers so its ok to do runtime specific things ?

test/e2e/infraprovider/engine/container/common.go
```diff
@@ -0,0 +1,606 @@
...
+func (o *ContainerOps) DeleteContainer(container api.ExternalContainer) error {
+	// check if it is present before deleting
+	exists, err := o.DoesContainerNameExist(container.Name)
+	if err != nil {
+		return fmt.Errorf("failed to check if container %s exists: %w", container.Name, err)
+	}
+	if !exists {
+		return nil
+	}
+	stdOut, err := o.CmdRunner.Run("rm", "-f", container.Name)
```
tssurya: hmm what about:
	// ensure all containers are disconnected from the network and if any are found, disconnect it.
		delimiter := " "
		stdOutBytes, err := exec.Command(containerengine.Get().String(),
			"network", "inspect", "-f", fmt.Sprintf("'{{range .Containers}}{{.Name}}%s{{end}}'", delimiter), network.Name()).CombinedOutput()
		if err != nil {
			framework.Logf("failed to list attached containers for network %s:, err: %s, stdout: (%s)", network, err, stdOutBytes)
			return false, nil
		}
		allContainers := strings.TrimSuffix(string(stdOutBytes), "\n")
		if allContainers != "" {
			for _, containerName := range strings.Split(allContainers, delimiter) {
				containerName = strings.TrimLeft(containerName, "'")
				containerName = strings.TrimRight(containerName, "'")
				if containerName == "" {
					continue
				}
				framework.Logf("deleting network encountered a stale container %q and it must be removed before removing the network", containerName)
				framework.Logf("Warning: Fix tests for container %s.. deleting container", containerName)
				if err = c.detachNetwork(network, containerName); err != nil {
					framework.Logf("while trying to delete network %q, attempted to detach container %q that is "+
						"still attached to network", network.Name(), containerName)
					return false, nil
				}
			}
		}
		stdOut, err := exec.Command(containerengine.Get().String(), "network", "rm", network.Name()).CombinedOutput()
these parts? is that done inside kind provider?

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,606 @@
...
+	}
+	cmdArgs := append([]string{"exec", name}, cmd...)
+	out, err := o.CmdRunner.Run(cmdArgs...)
+	if err != nil {
+		return "", fmt.Errorf("failed to exec container command (%s): err: %v, stdout: %q", strings.Join(cmdArgs, " "), err, out)
+	}
+	return out, nil
+}
+
+func (o *ContainerOps) GetExternalContainerLogs(container api.ExternalContainer) (string, error) {
```
tssurya, line 516: here you kept the "external" word on the util but for createcontainer, deletecontainer etc we removed it , so here as well it is just getcontainerlogs?

test/e2e/containerengine/container_engine.go
```diff
@@ -12,16 +12,6 @@ func (ce ContainerEngine) String() string {
 	return string(ce)
 }
 
-func (ce ContainerEngine) NetworkCIDRsFmt() string {
```
tssurya, line 15: I see you did this without a util inside 
func (o *ContainerOps) GetNetwork(networkName string) (ContainerEngineNetwork, error) {
maybe good to have this as a util inside common.go itself and keep that seperate from that big GetNetwork util

tssurya, line 15: I see you removed this util but did the logic inline inside
func (o *ContainerOps) GetNetwork(networkName string) (ContainerEngineNetwork, error) {
might be better to keep the util to not have that logic inside the already bigger function but not a biggie.

test/e2e/infraprovider/engine/container/provider.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/api"
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/portalloc"
+)
+
+// Provider provides a base implementation of api.Provider with common infrastructure functionality.
+// Infrastructure providers can embed this type to inherit default behavior and override specific
+// methods when provider-specific customization is required.
+type Provider struct {
```
tssurya: where is https://github.com/ovn-kubernetes/ovn-kubernetes/blob/cd2f08c7a486123e3aa5d58a388dd1972dc566b1/test/e2e/infraprovider/provider.go#L16 now?

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: yes @tssurya, you nailed it. that's the problem, now removed `internal`, but will revisit if @jcaamano proposes something else.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: The cleanup is still intact and this network gets added into c.CleanUpNetworks in context#CreateNetwork method which is the caller of this method. so cleanup gets executed at the time of test teardown.
i used different method signature so that it can be invoked with `c.MethodName` instead of `c.ContainerOps.MethodName` from `context.go`. 
now moved to using same method name calling it with `c.ContainerOps.MethodName`.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: this method is to get containers attached with a specific network, the code you pasted here is for `GetAttachedNetworks()` API implementation which is specific to kind cluster and it still present in `kind.go` (not moved).

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: yes, if network not really attached with container, then it's returning `nil` (look at `isNetworkAttachedToContainer` method), otherwise returning error would help caller to retry for transient errors, updated code with some comment.

test/e2e/infraprovider/engine/container/common.go
pperiyasamy: done.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: this is underlying container engine implementation for any provider, i hope this won't change.

test/e2e/infraprovider/engine/container/common.go
pperiyasamy: whatever you are mentioning here is for `DeleteNetwork` implementation which is being taken care of already.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: yes, you are right, changing method signature and log msgs according to its arguments now.

test/e2e/infraprovider/engine/container/provider.go
pperiyasamy: this is still present in the same place. does it confuse you ?

test/e2e/containerengine/container_engine.go
pperiyasamy: yes, it's removed.

test/e2e/containerengine/container_engine.go
pperiyasamy: we no longer need an util function with new network inspect approach.

pperiyasamy: same here

test/e2e/infraprovider/internal/engine/container/common.go
pperiyasamy: ok,done

test/e2e/e2e_suite_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/label"
 
 	deploymentkind "github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/deploymentconfig/configs/kind"
-	infrakind "github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/providers/kind"
+	infrakind "github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/kind"
```
qinqon: This directly imports the `kind` provider package, coupling the test harness to a specific implementation. With a factory in `provider.go`, this becomes:

```go
infraprovider.Set(infraprovider.New("kind"))
```

No direct import of any provider package needed. See https://github.com/qinqon/ovn-kubernetes/commit/cbaef68bf

test/e2e/infraprovider/command/runner.go
```diff
@@ -0,0 +1,11 @@
+package command
+
+// Runner abstracts command execution for various tools (docker, podman,
+// ssh, kcli, etc.)
+// Implementations execute commands with the provided arguments
+// and return combined stdout/stderr.
+type Runner interface {
```
qinqon: This interface has a single consumer (`engine/container/common.go`). Consider defining it in the same package as `ContainerOps` rather than a standalone package — a separate package for a 4-line interface adds import overhead without clear benefit.

test/e2e/infraprovider/api/api.go
```diff
@@ -176,7 +176,7 @@ func (n NetworkInterface) GetIPv4Prefix() string {
 }
 
 func (n NetworkInterface) GetIPv6Gateway() string {
-	return n.IPv4Gateway
+	return n.IPv6Gateway
```
qinqon, line 193: Good bug fix! Consider putting this in its own commit so it can be cherry-picked independently of the refactoring.

pperiyasamy:
> ## Encapsulation: `internal` packages + factory pattern
> The sub-packages (`engine/container`, `command`, `portalloc`, `kind`) are all publicly importable — any e2e test file can bypass the `api` interfaces and couple directly to implementation details.
> 
> I've pushed a reference commit showing this refactoring applied on top of your branch: [qinqon@cbaef68bf](https://github.com/qinqon/ovn-kubernetes/commit/cbaef68bf)
> 
> The diff is small (~28 insertions, 17 deletions). Key changes:
> 
> 1. **All implementation packages move under `internal/`** — Go compiler enforces only code within `infraprovider/` can import them
> 2. **Factory pattern** — `infraprovider.New("kind")` replaces direct `kind.New()`
> 3. **Generic `IsProvider(name)`** — `infraprovider.IsProvider("kind")` replaces `kind.IsProvider()`, keeping the public API provider-agnostic
> 4. **`command.Runner` collocated** into `internal/engine/container/command_runner.go` — its only consumer, removes standalone `command` package
> 
> ### How to add a downstream provider (e.g. OpenShift)
> 1. Create `test/e2e/infraprovider/internal/openshift/openshift.go` — implement `api.Provider` by embedding `container.Provider`
> 2. Add `"openshift"` case to the `New()` and `IsProvider()` switches in `provider.go`
> 3. In downstream `e2e_suite_test.go`: `infraprovider.Set(infraprovider.New("openshift"))`
> 
> The provider gets full access to all `internal` packages since it lives inside the `infraprovider` tree.

@qinqon as discussed offline, we decided not to move `infraprovider/engine` under internal package because other provider(s) placed anywhere can't use most of structs from `infraprovider/engine` package. so just moved "kind" provider implementation to internal and using factory pattern to initialize it.

test/e2e/infraprovider/engine/container/ops/ops.go
trozet: but the file already isn't internal?

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,622 @@
...
+		attachedContainers, err := o.getContainersAttachedToNetwork(name)
+		if err != nil {
+			framework.Logf("failed to get containers attached to network %s: %v", name, err)
+		}
+		if len(attachedContainers) > 0 {
+			return fmt.Errorf("network %s already exists with containers attached: '%v'", name, attachedContainers)
+		}
+		return fmt.Errorf("network %q already exists", name)
+	}
+	cmdArgs := []string{"network", "create", "--internal", "--driver", "bridge", name}
```
trozet, line 54: this is kind of a weird abstraction to me:
1. Runner is any command execution, like ssh.
2. ContainerOps embeds a Runner, then calls it with the above arguments?

I think the names are way too generic if we have to pass specific args here that I'm guessing work in docker/podman? I'm confused about what the goal of this is.

If these are really supposed to be docker/podman/infra commands, isn't that the same thing as the provider?

test/e2e/infraprovider/engine/container/context.go
```diff
@@ -0,0 +1,129 @@
...
+	"errors"
+	"sync"
+
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/api"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// TestContext provides a base implementation of api.Context with common container and network
+// management functionality. Infrastructure providers can embed this type and override specific
+// methods as needed for provider-specific behavior.
```
trozet: why not provide an interface then let new providers implement the interface?

test/e2e/infraprovider/engine/container/network/network.go
```diff
@@ -0,0 +1,58 @@
+package container
+
+import (
+	"fmt"
+
+	"github.com/ovn-kubernetes/ovn-kubernetes/test/e2e/infraprovider/api"
+	"k8s.io/utils/net"
+)
+
+type ContainerEngineNetwork struct {
```
trozet, line 10: this file needs comments. What is a "ContainerEngineNetwork"? Is that like a docker network? Why are we assuming docker/podman? Or maybe that is the point of everything under engine/container?

test/e2e/infraprovider/kind/ovs.go
trozet, line 1: why do you want to make it internal? what are we protecting?

trozet (commented):
I'm a bit confused on the refactor and what terms mean in these files. Have asked some initial questions to try to get a better understanding.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: The `test/e2e/infraprovider/engine/container/common.go` is a new file and not an `internal` module now, because we want to expose it so that any provider implementation (from anywhere) can import and reuse it.

test/e2e/infraprovider/engine/container/context.go
pperiyasamy: I'm not sure if i understood your question, The `TestContext` here provides some generic implementation that many third party infra provider can still make use of it. 
only in few cases, some method(s) may need to be overridden.

test/e2e/infraprovider/engine/container/network/network.go
pperiyasamy: yes, this is specific to docker/podman network because `ContainerEngineNetworkConfig` is an output of podman/docker inspect command.

test/e2e/infraprovider/kind/ovs.go
pperiyasamy: we are protecting kind provider implementation  which is specific to u/s implementation.
so this package can't be imported by third party infra provider(s)  if that is outside of `infraprovider` package.

test/e2e/infraprovider/engine/container/ops/ops.go
pperiyasamy: yes, The methods in ContainerOps builds args specific to docker (or) podman container lifecycle and network management and then just invoke `o.CmdRunner.Run(cmdArgs...)`.
The Run method is implemented in the actual provider, so it can run podman/docker command accordingly, for example, kind provider runs it directly like [below](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6034/changes/7a39c208afb84a0752d2eb423b8fc09abcbf0c21#diff-3a05f00acb9148d6fa09ebe739e5418de635cf261267b7a448f9e148d6f6a0ddR59):
```
func (r *kind) Run(args ...string) (string, error) {
	out, err := exec.Command(r.engine, args...).CombinedOutput()
	if err != nil {
		return "", fmt.Errorf("command failed: %w, output: %s", err, out)
	}
	return string(out), nil
}
```
any other third party provider can run podman/docker commands according to their need (for example, via ssh).

pperiyasamy:
> I'm a bit confused on the refactor and what terms mean in these files. Have asked some initial questions to try to get a better understanding.

@trozet updated PR description with latest info, hope that helps.

test/e2e/infraprovider/kind/kind.go
```diff
@@ -0,0 +1,300 @@
...
+	currentCtx, err := exec.Command("kubectl", "config", "current-context").CombinedOutput()
+	if err != nil {
+		framework.Logf("unable to get current cluster context: %v", err)
+		return false
+	}
+	if strings.Contains(string(currentCtx), "kind-ovn") {
+		return true
+	}
+	return false
+}
```
jcaamano: I would move this function to `infraprovider/provider.go`, rename it to `IsKind` and use it from here as well as from `deploymentconfig/configs/kind.go`

test/e2e/infraprovider/api/api.go
```diff
@@ -70,17 +73,23 @@ type Underlay struct {
...
+	DeleteExternalContainer(container ExternalContainer) error
 	AttachNetwork(network Network, instance string) (NetworkInterface, error)
 	DetachNetwork(network Network, instance string) error
-	GetAttachedNetworks() (Networks, error)
-	SetupUnderlay(f *framework.Framework, underlay Underlay) error
+}
 
+type ClusterContextProvider interface {
+	// anything done on the cluster that could be cleaned up after test
 	AddCleanUpFn(func() error)
```
jcaamano: I would have `AddCleanUpFn` directly in `Context` interface

test/e2e/infraprovider/providers/kind/runtime.go
```diff
@@ -1,4 +1,4 @@
-package containerengine
+package kind
```
jcaamano, line 1: Rename `Get` below to `getContainerRuntime`.
Rename `ContainerEngine` below to `containerRuntime`
Rename to `docker` and `podman`.

I am unexporting because I am not sure why would you need this exported. If you do, kind is probably not the appropriate package to hold these.

jcaamano (commented):
Looks good. Minor comments. 

I would suggest moving these:
`infraprovider/kind` to `infraprovider/providers/kind` (sorry if I suggested the opposite before, myh intention was to gt rid of `internal`)
`infraprovider/runner` to `infraprovider/engine/runner`
`infraprovider/testcontext` to `infraprovider/engine/testcontext`
`infraprovider/portalloc` to `infraprovider/engine/portalloc`

Also the port allocators don't think will work for parallel tests since they are memory based, but we can fix that later.

test/e2e/infraprovider/engine/container/ops/ops.go
```diff
@@ -0,0 +1,618 @@
...
+	// nameFormat is a Go template for extracting network names from inspect output.
+	nameFormat = "{{.Name}}"
+)
+
+// ContainerOps provides reusable helper methods for container lifecycle and network management.
+type ContainerOps struct {
+	// docker or podman string
+	Engine string
+	// Runner executes container runtime commands and returns their output.
+	CmdRunner api.Runner
```
jcaamano: Rename `Engine` to `runtime`
Rename `CmdRunner` to `cmdRunner`

pperiyasamy:
The BGP test failure with `bgp-no-overlay-helm` lane is due to https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5952, unrelated to this PR.

PR #6037, opened 2026-02-27
3 more simple nftables patches

danwinship:
1. Simplify the code in the unit tests that generates the nftables rulesets to compare against (so that as we do more iptables to nftables porting and add more groups of rules that are present in some tests but not others, it's easier for each test to get exactly the set of rules it needs, as in [this future commit](https://github.com/danwinship/ovn-kubernetes/commit/4a90f6aa2ddf3e80d24bb0bf30d2029fcd6a79c4#diff-baef649410d0c75a89ece69e5a452be4a68d9fb8f8febed16657ff4fc2e6b2a4)).
2. Avoid hardcoding information about the nftables set/map names in `gateway_shared_intf.go` (so that as we do more iptables to nftables porting and add more sets and maps, `gateway_shared_intf.go` won't accidentally get out of sync with `gateway_nftables.go`, as in [this future commit](https://github.com/danwinship/ovn-kubernetes/commit/69bef95aa93560d606aca02e753087111910698c#diff-1082c8ffc2e5b9a3fb8c033125e918d595ee6b04dca9d826dccd0937db542de6R685-R711))
3. Add nftables support to test/e2e/diagnostics

## Summary by CodeRabbit

* **Tests**
  * Added a --collect-nftables flag and end-to-end diagnostics to capture nftables rules periodically via a daemonset.
  * Integrated nftables collection into test setup to enable on-demand nftables troubleshooting.

* **Chores**
  * Centralized and simplified nftables rule/expectation handling in tests for clearer, more maintainable test assertions.

### DIFF
--- a/test/e2e/diagnostics/nftables.go
+++ b/test/e2e/diagnostics/nftables.go
@@ -1,0 +1,22 @@
+package diagnostics
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+func (d *Diagnostics) NFTablesDumpingDaemonSet() {
+	if !d.nftables {
+		return
+	}
+	By("Creating nftables dumping daemonsets")
+	daemonSets := []appsv1.DaemonSet{}
+	daemonSetName := fmt.Sprintf("dump-nftables")
+	cmd := composePeriodicCmd("nft list ruleset", 10)
+	daemonSets = append(daemonSets, d.composeDiagnosticsDaemonSet(daemonSetName, cmd, "nftables"))
+	Expect(d.runDaemonSets(daemonSets)).To(Succeed())
+}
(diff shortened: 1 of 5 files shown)


test/e2e/diagnostics/nftables.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func (d *Diagnostics) NFTablesDumpingDaemonSet() {
+	if !d.nftables {
+		return
+	}
+	By("Creating nftables dumping daemonsets")
+	daemonSets := []appsv1.DaemonSet{}
+	daemonSetName := fmt.Sprintf("dump-nftables")
+	cmd := composePeriodicCmd("nft list ruleset", 10)
```
tssurya, line 19: what does 10 at the end do? (every 10seconds rules the command I guess..)

tssurya:
triggering a rebase since CI run is one month old, upon success will merge the PR

test/e2e/diagnostics/nftables.go
danwinship: yes... it's just copied from the iptables version

PR #5516, opened 2025-08-25
docs: Add instructions for CI failures

npinaeva:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI

Issue #5152, opened 2025-03-26
TestAlignCPUAffinity failure in Build-PR: ovspinning_linux_test.g


flavio-fernandes:
### Which jobs are flaking?

```
gocmd test -mod=vendor -test.v -race  -test.coverprofile=37.coverprofile  github.com/ovn-org/ovn-kubernetes/go-controller/pkg/node/ovspinning 
=== RUN   TestAlignCPUAffinity
I0326 20:33:58.639045   40939 ovspinning_linux_test.go:65] Test CPU Affinity [1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
I0326 20:33:58.639180   40939 ovspinning_linux.go:46] Starting OVS daemon CPU pinning
I0326 20:33:58.659702   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40947) (ntasks=9) to 0, was 0-3
I0326 20:33:58.659935   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40946) (ntasks=1) to 0, was 0-3
I0326 20:33:58.681039   40939 ovspinning_linux_test.go:65] Test CPU Affinity [2 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
I0326 20:33:58.700354   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40947) (ntasks=9) to 1, was 0
I0326 20:33:58.700488   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40946) (ntasks=1) to 1, was 0
I0326 20:33:58.712645   40939 ovspinning_linux_test.go:65] Test CPU Affinity [4 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
I0326 20:33:58.719797   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40947) (ntasks=9) to 2, was 1
I0326 20:33:58.719940   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40946) (ntasks=1) to 2, was 1
I0326 20:33:58.734157   40939 ovspinning_linux_test.go:65] Test CPU Affinity [8 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0]
I0326 20:33:58.740294   40939 ovspinning_linux.go:196] Setting CPU affinity of PID(40947) (ntasks=9) to 3, was 2
[...]

PR #6039, opened 2026-02-27
Add requirements to the docs

### LINKED ISSUE #5873: Document kernel and other requirements
### What would you like to see added/updated/deleted in our docs?

There doesn't seem to be documentation of required versions of things:
- kernel ([this comment](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5858#issuecomment-3777815317) suggests 6.6.53 or later?)
- ovn
- openvswitch userspace
- nft binary (currently 1.0.1 or later, but becomes 1.0.8 or later if #5858 merges)
- ... ?

### Why is this needed?

so people know?

npinaeva:
## 📑 Description

Getting started with documenting dependencies.
It is not a full list, hope to get it updated more in the future.
To get dependencies I had to scan through CI/scripts/etc history, so the requirements are not necessarily minimal, but "recommended".
Added strict mode to the mkdoc check, here is a failure example https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22762115540/job/66020869085?pr=6039

Fixes #5873

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Documentation**
  * Added a new "Requirements" doc outlining external compatibility and recommended versions for OVN, OVS, nft, multus/CNI, Kubernetes, and feature-specific prerequisites (UDN, BGP/EVPN, observability, Multi‑VTEP, DPU/OVS acceleration).
  * Added cross-reference reminders to the Requirements page across many feature docs and updated site navigation to include it.
  * Minor doc link and formatting updates.

* **Chores**
  * Docs build set to strict mode in CI (fail on warnings).


docs/features/requirements.md
```diff
@@ -0,0 +1,23 @@
+# External requirements
+
+ovn-kubernetes has additional dependencies for the external components.
+
+- nft binary 1.0.1+
+- OVN version: recommended to use the same major version as our CI is running: check `Dockerfile.fedora`
```
booxter: Suggest we are explicit here instead of sending users to refer to a CI file. But also: I think these two are / may be different: a) what we support and b) what we run CI against. We may want to switch CI to later release while keeping a promise to stay compatible with an earlier version.

docs/features/requirements.md
```diff
@@ -0,0 +1,23 @@
...
+  [6.10.12](https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.10.12),
+  [6.6.53](https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.6.53), and 
+  [6.1.112](https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.1.112)
+- cgroupv2: required for kubelet probes to work with UDN pods
+
+## OVN Observability
+OVS 3.4+ and linux kernel 6.11+
+
+## Multi-VTEP
+OVN version newer than v24.03.2
```
girishmg: These as well:

- FRR-K8s (helm chart version or some release of FRR)
- CNI Spec
- Version of Multus
- K8s Version (that we are testing our CI with)
- SR-IOV Device Plugin

docs/features/requirements.md
```diff
@@ -0,0 +1,23 @@
+# External requirements
+
+ovn-kubernetes has additional dependencies for the external components.
+
+- nft binary 1.0.1+
+- OVN version: recommended to use the same major version as our CI is running: check `Dockerfile.fedora`
+- OVN should work with any supported OVS release, extra requirements for OVS version may be specified per-feature
+
```
pliurh, line 14: For BGP related features, we need also mention the FRR and FRR-K8S versions.

docs/features/requirements.md
npinaeva: yeah I agree, I was going to do the proper matching later, but I can also try now
We have lost some context, so really the only source of older ovn version would be CI files, but we can get better in the future

docs/features/requirements.md
npinaeva: I couldn't find sriov device plugin version mentioned anywhere, out docs usually just point to the repo.
Can you provide the per-ovnk version dependency?

docs/features/requirements.md
npinaeva: I found some frr-k8s version, but couldn't find anything about the frr itself, do you have more info on what versions we require?

trozet: iirc we need FRR 10. @jcaamano can you confirm?

jcaamano: For EVPN, yes, we do.

docs/features/requirements.md
booxter: It's ok to start from clean slate, with the understanding that this minimal version doesn't automatically move with our CI.

docs/features/requirements.md
jcaamano: the corresponding frr-k8s version is v0.0.21+

docs/features/requirements.md
```diff
@@ -0,0 +1,36 @@
...
+supported versions.
+
+| OVN-k release | OVN release | nft binary | multus | CNI spec | k8s        | 
+|---------------|-------------|------------|--------|----------|------------|
+| master        | 25.09       | 1.0.1+     | v4.1.3 | 1.1.0    | 1.35 (WIP) |
+| 1.2           | 25.09       | 1.0.1+     | v4.1.3 | 0.4.0    | 1.34       |
+| 1.1           | 25.03       | 1.0.1+     | v4.1.3 | 0.4.0    | 1.33       |
+| 1.0           | 24.03       | -          | v4.1.0 | 0.4.0    | 1.29       |
+
+OVN should work with any supported OVS release, extra requirements for OVS version may be specified per-feature
```
booxter, line 13: I think we should link to OVN / OVS docs that list "supported" versions:

https://www.ovn.org/en/releases/all_releases/

OVS is a bit more fuzzy but may give some idea to the reader:
https://docs.openvswitch.org/en/latest/internals/release-process/

docs/features/requirements.md
```diff
@@ -0,0 +1,36 @@
+# External requirements
+
+ovn-kubernetes has additional dependencies for the external components, here are the recommended (not necessarily minimal)
```
booxter: This document can be useful in two ways:

1. to communicate to users which versions to use OVN-Kubernetes with.
2. to communicate to developers which versions they should consider in new developments.

(1) is reasonably well covered in this PR. But the current wording (recommended not necessarily minimal) doesn't allow me, a developer, to remove support for any old versions of any of these components. I think we should be more forceful in the language here, saying that particular versions are minimal supported - and if users go below these, all bets are off and we are not going to support these setups. Otherwise, the document implies developers should support any versions of all these components, even if we recommend newer ones.

docs/features/requirements.md
npinaeva: can you also trace this requirement back through the ovn-k releases? Should I say frr requirement is needed when route_advertisement is enabled or only EVPN?

docs/features/requirements.md
npinaeva: well that would be nice of course to list the minimal versions, the problem is that I can't. That would involve digging into history and running tests clusters for infinite amount of time. The hope is though that moving forward with this doc we will better document minimal vs recommended versions

npinaeva:
@trozet can you also share the DPU CNI version I can't find anything :(

trozet:
> @trozet can you also share the DPU CNI version I can't find anything :(

So for DPU healthcheck to work correctly, a user needs:
https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v4.2.4

and they need a crio/containerd version that will actually read status:
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5777#issuecomment-3756907539

Now if the user does not have a new enough contiainerd/crio it wont crash the pod or anything, the runtime just wont use status to indicate if the DPU is healthy, so the DPU healthcheck wont do anything to signal an unhealthy state. So I guess for documentation purposes we can just say for the DPU healthcheck to work correctly you need multus v4.2.4, containerd >= v2 or crio >=1.32.

docs/features/requirements.md
booxter: We could just pick up some values that we are comfortable with. Like the earliest supported OVS/OVN. No need to gather it through runtime testing.

docs/features/requirements.md
npinaeva: @jcaamano please?

jcaamano: FRR is required for route advertisements. We started using frr 9.1. For EVPN we needed frr 10.4. We haven't used any other version.

docs/features/requirements.md
```diff
@@ -0,0 +1,64 @@
...
+
+- Linux Kernel 5.7.0 or above
+- Open vSwitch 2.13 or above
+- iproute >= 4.12
+
+## DPU healthcheck support
+
+[DPU healthcheck support OKEP](../okeps/okep-5674-dpu-healthcheck.md)
+
+- multus-CNI v4.2.4
```
trozet: nit: should this have ">=" or "or above"?

docs/features/requirements.md
```diff
@@ -0,0 +1,64 @@
...
+
+OVN should work with any supported OVS release, extra requirements for OVS version may be specified per-feature
+
+- [OVN releases](https://www.ovn.org/en/releases/all_releases/)
+- [OVS release process](https://docs.openvswitch.org/en/latest/internals/release-process/)
+
+Some of the requirements are feature-specific.
+
+## UDN
+- kernel fix 7f3287db654395f9c5ddd246325ff7889f550286: netfilter: nft_socket: make cgroupsv2 matching work with namespaces)
```
trozet, line 22: maybe some better summary of what this fixes from a higher level, and maybe whether or not it is an absolute requirement or what breaks without it.

docs/features/requirements.md
```diff
@@ -0,0 +1,64 @@
+# External requirements
+
+ovn-kubernetes has additional dependencies for the external components, here are the recommended (not necessarily minimal)
```
trozet: I think @tssurya mentioned we need to use proper naming for OVN-Kubernetes (capitlization, shortnames, etc).

docs/features/requirements.md
```diff
@@ -0,0 +1,64 @@
...
+supported versions.
+
+| OVN-k release | OVN release | nft binary | multus | CNI spec | k8s        |
```
trozet: not sure if we should use OVN-K or OVN-Kube

docs/features/requirements.md
npinaeva: yeah I even fixed a part of it with the dashes https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5516/changes/8db8ca1de453c00817037bee3bb85a21c3263d47, but we still seem to use ovn-k across the docs, not sure if that is considered ok

docs/features/requirements.md
npinaeva: definitely not ovn-kube (we are mixed up with kube-ovn already often enough) I will just use full name for now

npinaeva:
caught https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5152

PR #6040, opened 2026-02-27
[wip] Stop skipping serial k8s tests

kyrtapz:
Remove the PARALLEL env variable and associated logic that was unconditionally skipping all [Serial] tagged upstream Kubernetes tests. It was a confusing naming choice suggesting we ran our E2Es in parallel.

e2e-kind.sh uses `ginkgo --nodes=N` which handles serial tests at the end.
Reverts https://github.com/ovn-kubernetes/ovn-kubernetes/commit/bc9eacac17062500e81e78f951196cd14eb64c71 because we run our tests serially anyway. 

Lets see if anything breaks in the k8s serial e2es, will need to check how much longer our lanes take. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime network reliability by adding automatic recovery when gateway uplinks or bridge IPs are missing, with migration attempts, retries and clearer error reporting.

* **Chores**
  * Simplified CI and end-to-end test orchestration: removed legacy serial-test paths and explicit parallel-override behavior, and streamlined label-based test selection to use skip-based filtering.

### DIFF
--- a/.github/workflows/performance-test.yml
+++ b/.github/workflows/performance-test.yml
@@ -13,10 +13,6 @@ env:
   KIND_CLUSTER_NAME: ovn
   KIND_INSTALL_INGRESS: true
   KIND_ALLOW_SYSTEM_WRITES: true
-  # This skips tests tagged as Serial
-  # Current Serial tests are not relevant for OVN
-  PARALLEL: true
-
   # This must be a directory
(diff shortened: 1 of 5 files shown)


booxter:
@kyrtapz wanna revive it? I assume the rationale to revert is still valid.

PR #6041, opened 2026-02-28
Fixes Dynamic UDN + CNC integration

### LINKED ISSUE #5963: Consider dynamic UDNs for the services connectivity
### What is this issue about?

A basic level was done in that PR in one of the commits.. but the design needs to be considered and cross checked

trozet:
- Adds CNC tracker in NAD controller to influence Dynamic UDN Network active/inactive states. Networks are now active that are selected by CNC and that CNC has at least 1 active network on a node.
- Adds new API to register with NAD controller for network related events for nodes when they go active/inactive
- CNC controller subscribes to this new API, then reconciles when networks go active on a remote node
- Layer 3 remote routes/ports are filtered if nodes are inactive

Fixes #5963

## Summary by CodeRabbit

* **New Features**
  * Added ClusterNetworkConnect support and a network-ref reconciliation mechanism to detect and notify when nodes gain/lose network connectivity.
  * Integrated network-ref plumbing into controllers to trigger reconcilation workflows for affected networks.

* **Improvements**
  * Controllers now proactively remove stale per-node ports and routes when a node is no longer active for a network.

* **Tests**
  * Added unit and end-to-end tests for ClusterNetworkConnect, network-ref reconciliation, and dynamic user-defined network connectivity.

### DIFF
--- a/go-controller/pkg/networkmanager/api.go
+++ b/go-controller/pkg/networkmanager/api.go
@@ -32,6 +32,17 @@ const (
 // NADReconciler is a level-driven controller notified of NAD key changes.
 type NADReconciler controller.Reconciler
 
+// NetworkRefReconciler is notified when node network activity changes.
+// The network is identified by name because activity can be observed before
+// the NAD controller has fully resolved the network into a local netInfo/ID.
+//
+// These are edge-style notifications. Controllers that need eventual
+// correctness after NAD/ID changes should also register an NADReconciler and
+// treat NAD callbacks as a catch-up signal.
+type NetworkRefReconciler interface {
+	Reconcile(node string, networkName string)
+}
+
 type watchFactory interface {
 	NADInformer() nadinformers.NetworkAttachmentDefinitionInformer
 	ClusterNetworkConnectInformer() networkconnectinformer.ClusterNetworkConnectInformer
@@ -96,6 +107,13 @@ type Interface interface {
 	RegisterNADReconciler(r NADReconciler) uint64
 	// DeRegisterNADReconciler removes a previously registered reconciler.
 	DeRegisterNADReconciler(id uint64)
+	// RegisterNetworkRefReconciler registers a reconciler to be notified when
+	// node network activity changes.
+	// Pair this with RegisterNADReconciler when the controller needs both
+	// immediate activity signals and eventual catch-up after NAD/ID changes.
+	RegisterNetworkRefReconciler(r NetworkRefReconciler) uint64
+	// DeRegisterNetworkRefReconciler removes a previously registered reconciler.
+	DeRegisterNetworkRefReconciler(id uint64)
 
 	// GetNetworkByID returns the network with the given ID or nil if not found.
 	// This is an O(1) lookup using an internal index.
@@ -305,6 +323,12 @@ func (nm defaultNetworkManager) RegisterNADReconciler(_ NADReconciler) uint64 {
 }
 
 func (nm defaultNetworkManager) DeRegisterNADReconciler(_ uint64) {}
+
+func (nm defaultNetworkManager) RegisterNetworkRefReconciler(_ NetworkRefReconciler) uint64 {
+	return 0
+}
+
+func (nm defaultNetworkManager) DeRegisterNetworkRefReconciler(_ uint64) {}
 
 func (nm defaultNetworkManager) GetNetworkByID(id int) util.NetInfo {
 	if id != types.DefaultNetworkID {
--- a/go-controller/pkg/networkmanager/api.go
+++ b/go-controller/pkg/networkmanager/api.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/allocator/id"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controller"
+	networkconnectinformer "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/clusternetworkconnect/v1/apis/informers/externalversions/clusternetworkconnect/v1"
 	egressipinformer "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/egressip/v1/apis/informers/externalversions/egressip/v1"
 	rainformers "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/routeadvertisements/v1/apis/informers/externalversions/routeadvertisements/v1"
 	userdefinednetworkinformer "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/informers/externalversions/userdefinednetwork/v1"
@@ -33,6 +34,7 @@ type NADReconciler controller.Reconciler
 
 type watchFactory interface {
 	NADInformer() nadinformers.NetworkAttachmentDefinitionInformer
+	ClusterNetworkConnectInformer() networkconnectinformer.ClusterNetworkConnectInformer
 	UserDefinedNetworkInformer() userdefinednetworkinformer.UserDefinedNetworkInformer
 	ClusterUserDefinedNetworkInformer() userdefinednetworkinformer.ClusterUserDefinedNetworkInformer
 	NamespaceInformer() coreinformers.NamespaceInformer
--- a/go-controller/pkg/networkmanager/nad_controller.go
+++ b/go-controller/pkg/networkmanager/nad_controller.go
@@ -29,6 +29,8 @@ import (
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/allocator/id"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/config"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controller"
+	networkconnectv1 "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/clusternetworkconnect/v1"
+	networkconnectlister "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/clusternetworkconnect/v1/apis/listers/clusternetworkconnect/v1"
 	userdefinednetworklister "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/listers/userdefinednetwork/v1"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/kube"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
@@ -56,9 +58,13 @@ type nadController struct {
 	cudnLister      userdefinednetworklister.ClusterUserDefinedNetworkLister
 	namespaceLister corelisters.NamespaceLister
 	nodeLister      corelisters.NodeLister
+	cncLister       networkconnectlister.ClusterNetworkConnectLister
 
 	controller controller.Controller
-	recorder   record.EventRecorder
+	// cncController tracks CNC connectivity and updates derived
+	// network activity for Dynamic UDN filtering.
+	cncController controller.Controller
+	recorder      record.EventRecorder
 
 	// networkController reconciles network specific controllers
 	networkController *networkController
@@ -70,6 +76,15 @@ type nadController struct {
 	// dynamicFilterNADs tracks whether a NAD should be activity-gated by Dynamic UDN.
 	// Bare NADs are not filtered and therefore should be treated as present on all nodes.
 	dynamicFilterNADs map[string]bool
+	// cncSelectedNetworks tracks network names selected by each CNC.
+	cncSelectedNetworks map[string]sets.Set[string]
+	// cncNetworkIDs tracks owner network IDs referenced by each CNC.
+	cncNetworkIDs map[string]sets.Set[int]
+	// cncsByNetworkID indexes CNCs by their referenced owner network IDs.
+	cncsByNetworkID map[int]sets.Set[string]
+	// cncConnectedNetworks is a symmetric adjacency map where key/value are
+	// network names connected through a CNC.
+	cncConnectedNetworks map[string]sets.Set[string]
 
 	// primaryNADs holds a mapping of namespace to NAD of primary UDNs
 	primaryNADs map[string]string
@@ -111,18 +126,22 @@ func newController(
 ) (*nadController, error) {
 	networkController := newNetworkController(name, zone, node, cm, wf)
 	c := &nadController{
-		name:              fmt.Sprintf("[%s NAD controller]", name),
-		stopChan:          make(chan struct{}),
-		recorder:          recorder,
-		nadLister:         wf.NADInformer().Lister(),
-		nodeLister:        wf.NodeCoreInformer().Lister(),
-		networkController: networkController,
-		reconcilers:       map[uint64]reconcilerRegistration{},
-		nads:              map[string]string{},
-		nadsByNetwork:     map[string]sets.Set[string]{},
-		dynamicFilterNADs: map[string]bool{},
-		primaryNADs:       map[string]string{},
-		markedForRemoval:  map[string]time.Time{},
+		name:                 fmt.Sprintf("[%s NAD controller]", name),
+		stopChan:             make(chan struct{}),
+		recorder:             recorder,
+		nadLister:            wf.NADInformer().Lister(),
+		nodeLister:           wf.NodeCoreInformer().Lister(),
+		networkController:    networkController,
+		reconcilers:          map[uint64]reconcilerRegistration{},
+		nads:                 map[string]string{},
+		nadsByNetwork:        map[string]sets.Set[string]{},
+		dynamicFilterNADs:    map[string]bool{},
+		cncSelectedNetworks:  map[string]sets.Set[string]{},
+		cncNetworkIDs:        map[string]sets.Set[int]{},
+		cncsByNetworkID:      map[int]sets.Set[string]{},
+		cncConnectedNetworks: map[string]sets.Set[string]{},
+		primaryNADs:          map[string]string{},
+		markedForRemoval:     map[string]time.Time{},
 	}
 	networkController.getNADKeysForNetwork = c.GetNADKeysForNetwork
 
@@ -136,6 +155,21 @@ func newController(
 			c.eipReconcilerID = eipID
 		}
 		c.filterNADsOnNode = filterNADsOnNode
+		if util.IsNetworkConnectEnabled() {
+			cncInformer := wf.ClusterNetworkConnectInformer()
+			c.cncLister = cncInformer.Lister()
+			c.cncController = controller.NewController(
+				fmt.Sprintf("%s-cnc-connectivity-controller", c.name),
+				&controller.ControllerConfig[networkconnectv1.ClusterNetworkConnect]{
+					RateLimiter:    workqueue.DefaultTypedControllerRateLimiter[string](),
+					Informer:       cncInformer.Informer(),
+					Lister:         c.cncLister.List,
+					Reconcile:      c.syncCNC,
+					ObjNeedsUpdate: c.cncNeedsUpdate,
+					Threadiness:    1,
+				},
+			)
+		}
 	}
 
 	c.networkController.nodeHasNetwork = c.NodeHasNetwork
@@ -183,7 +217,7 @@ func newController(
 	return c, nil
 }
 
-func (c *nadController) nodeHasNAD(node, nad string) bool {
+func (c *nadController) nodeHasDirectNAD(node, nad string) bool {
 	if !config.OVNKubernetesFeature.EnableDynamicUDNAllocation {
 		return true
 	}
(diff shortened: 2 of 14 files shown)


go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -404,6 +512,154 @@ func (c *nadController) Stop() {
...
+		if c.networkIDAllocator.GetID(networkName) == networkID {
+			return networkName
+		}
+	}
+	return ""
+}
+
+// syncCNC rebuilds the CNC adjacency map and requeues affected
+// networks when connectivity relationships change.
+func (c *nadController) syncCNC(string) error {
```
npinaeva: why would you ignore the key and reconcile all cncs?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -404,6 +512,154 @@ func (c *nadController) Stop() {
...
+				continue
+			}
+			networkName := c.networkNameForIDLocked(networkID)
+			if networkName == "" {
+				continue
+			}
+			selectedNetworks.Insert(networkName)
+		}
+
+		for src := range selectedNetworks {
```
npinaeva: I am trying to avoid src and dst notation in the cnc context, how about we call it selectedNetwork and peer?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -404,6 +512,154 @@ func (c *nadController) Stop() {
...
+				if src == dst {
+					continue
+				}
+				peers.Insert(dst)
+			}
+		}
+	}
+	c.RUnlock()
+
+	changedNetworks := sets.New[string]()
```
npinaeva, line 651: maybe call it allNetworks? it doesn't really filter by changes yet, just tries to unify previous cache with the new cache?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -404,6 +512,154 @@ func (c *nadController) Stop() {
...
+}
+
+// networkNameForIDLocked resolves a network name from the cached network ID.
+// Caller must hold nadController lock.
+func (c *nadController) networkNameForIDLocked(networkID int) string {
+	if networkID == types.InvalidID {
+		return ""
+	}
+	for networkName := range c.nadsByNetwork {
+		if c.networkIDAllocator.GetID(networkName) == networkID {
```
npinaeva, line 596: do you want to reuse this logic in `(c *nadController) GetNetworkByID`?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -642,6 +914,20 @@ func (c *nadController) syncNAD(key string, nad *nettypes.NetworkAttachmentDefin
...
+		}
+		if previousNetworkName != currentNetworkName || previousNetworkID != currentNetworkID {
+			shouldReconcileCNC = true
+			if previousNetworkID != types.InvalidID {
+				affectedCNCNetworkIDs.Insert(previousNetworkID)
+			}
+			if currentNetworkID != types.InvalidID {
+				affectedCNCNetworkIDs.Insert(currentNetworkID)
+			}
+		}
```
npinaeva, line 1123: I am thinking if it would be easier/more obvious/more efficient if CNC controller just stored the IDs of networks it couldn't resolve, and we would just check that list, otherwise we have to reconcile all CNCs on every nad create, and I think we use a centralized lock anyway in those handlers, so no extra locking is needed?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -329,8 +405,27 @@ func (c *nadController) filter(nad *nettypes.NetworkAttachmentDefinition) (bool,
...
+		if err != nil || nadNetwork == nil {
+			return true, nil
+		}
+		networkName = nadNetwork.GetNetworkName()
+	}
+
+	// We don't support multiple nodes per zone; assume zone name is node name.
+	// This function is called from syncNAD while c is already write-locked,
+	// so use lock-free activity helpers.
+	if c.nodeHasEffectiveNetworkNoLock(ourNode, networkName) {
```
npinaeva: I would probably put this into nodeHasNAD method logically, then rename current nodeHasNAD into nodeHasDirectNAD, and nodeHasEffectiveNetworkNoLock into nodeHasNetworkNoLock

go-controller/pkg/networkmanager/api.go
```diff
@@ -93,6 +104,13 @@ type Interface interface {
 	RegisterNADReconciler(r NADReconciler) (uint64, error)
 	// DeRegisterNADReconciler removes a previously registered reconciler.
 	DeRegisterNADReconciler(id uint64) error
+	// RegisterNetworkRefReconciler registers a reconciler to be notified when
+	// node network activity changes.
+	// Pair this with RegisterNADReconciler when the controller needs both
+	// immediate activity signals and eventual catch-up after NAD/ID changes.
+	RegisterNetworkRefReconciler(r NetworkRefReconciler) (uint64, error)
+	// DeRegisterNetworkRefReconciler removes a previously registered reconciler.
+	DeRegisterNetworkRefReconciler(id uint64) error
```
npinaeva: why do we need network reconciler? can't we just NADReconciler and get network by NAD in the handler?

npinaeva (commented):
not finished yet, submitting some questions

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -524,6 +524,7 @@ func (c *Controller) ensureConnectPortsOps(ops []ovsdb.Operation, cnc *networkco
 			networkPortName := getNetworkRouterToConnectRouterPortName(networkName, node.Name, cncName)
 
 			isLocalNode := util.GetNodeZone(node) == c.zone
+			nodeActive := c.networkManager.NodeHasNetwork(node.Name, networkName)
```
npinaeva, line 734: should we remove localActive parameter now, since we are getting nodeActive for remote nodes now too?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -555,6 +556,23 @@ func (c *Controller) ensureConnectPortsOps(ops []ovsdb.Operation, cnc *networkco
...
+					}
+					ops, err = libovsdbops.DeleteLogicalRouterPortWithPredicateOps(c.nbClient, ops, networkRouterName,
+						func(item *nbdb.LogicalRouterPort) bool {
+							return item.Name == networkPortName
+						})
+					if err != nil {
+						return nil, fmt.Errorf("failed to delete network router port ops %s: %v", networkPortName, err)
+					}
+					continue
+				}
```
npinaeva, line 782: isn't this exactly the same cleanup as for local node?

go-controller/pkg/ovn/controller/networkconnect/topology.go
```diff
@@ -856,6 +874,19 @@ func (c *Controller) ensureStaticRoutesOps(ops []ovsdb.Operation, cnc *networkco
 			if err != nil {
 				continue
 			}
+			nodeActive := c.networkManager.NodeHasNetwork(node.Name, networkName)
```
npinaeva, line 1072: we currently have a part of the isActive logic in the `syncNetworkConnections` like this `if netInfo.TopologyType() != ovntypes.Layer2Topology || localActive {ensureStaticRoutesOps}` and another part inside, I think it will be less confusing to just call `ensureStaticRoutesOps` unconditionally and move the layer2 logic also inside the `ensureStaticRoutesOps`

npinaeva (commented):
While testing this, I figure we have a bug
We treat `OnNetworkRefChange` unconditionally, so when we have received active events from both pod tracker and egressip tracker, then egress ip tracker sends inactive event, the whole NAD is deleted even though the pod is still there.

npinaeva:
we will also need to revert https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5828/changes/ff6252a83d05ff350249ac8679ab9b77107c99dd when the PR is merged

go-controller/pkg/networkmanager/nad_controller.go
trozet: stupid AI...

go-controller/pkg/networkmanager/api.go
trozet: NADReconciler does not fire when node activity changes. A pod being created/deleted on a remote node does not change the NAD. It changes whether that node should have CNC L3 remote ports/routes for a network. NetworkRefReconcile signals when a node goes active/inactive.

go-controller/pkg/ovn/controller/networkconnect/topology.go
trozet: we cant remove it bc it is still needed for the layer 2 path, but we can optimize the layer 3 path here.

trozet:
> While testing this, I figure we have a bug We treat `OnNetworkRefChange` unconditionally, so when we have received active events from both pod tracker and egressip tracker, then egress ip tracker sends inactive event, the whole NAD is deleted even though the pod is still there.

@npinaeva we should be OK here now because OnNetworkRefChange for local NAD/node calls reconcileNetworkActivity which will check NodeHasNetwork, which eventually calls nodeHasDirectNAD and checks if either podTracker or EgressIP tracker have it. If those checks fail, then it checks CNC.

For remote nodes, the event gets propagated to the network controller (or via the new api to a NetworkRefReconciler) and in the network controller logic we check if the node is active during reconciliation. Similarly for the new API, the client needs to evaluate activity during reconciliation, which CNC does.

I'll add a UT for this anyway.

trozet:
@coderabbitai review

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -414,6 +585,244 @@ func (c *nadController) Stop() {
...
+
+// syncCNC refreshes one queued CNC and requeues affected networks when
+// connectivity relationships change.
+func (c *nadController) syncCNC(key string) error {
+	if c.cncLister == nil {
+		return nil
+	}
+	if key == "" {
+		return nil
+	}
```
npinaeva: I know I am annoying, but these checks don't make any sense

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -196,34 +230,103 @@ func (c *nadController) nodeHasNAD(node, nad string) bool {
 	return false
 }
 
-func (c *nadController) NodeHasNetwork(node, networkName string) bool {
+// nodeHasNAD reports whether a node should be considered active for the given
+// NAD, either through a direct NAD ref or a CNC-connected network.
+// Caller must hold nadController lock.
```
npinaeva: I think we need to clarify here that NAD is active if its network is active, and network can become active through one of it's NADs trackers or CNC. Then I would also update the implementation to just get network and call `nodeHasNetworkNoLock` and get rid of all `ExcludingNAD` functions?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -273,13 +370,14 @@ func (c *nadController) deleteNADFromNetworkLocked(networkName, nadKey string) {
...
 // a network needs to be added or removed or just updated.
 // Used as a callback for pod/egress IP events when dynamic UDN allocation is enabled.
-// This callback is invoked by pod/egress IP trackers and is blocking. Therefore it's work should be as lightweight
+// This callback is invoked by pod/egress IP trackers and is blocking. Therefore its work should be as lightweight
 // as possible.
 // The function handles:
 //  1. Queuing local node event NADs to the NAD Controller for reconciliation later in the NAD Controller worker.
 //  2. Queuing remote node event networks to the Network Manager for reconciliation later in the Network Manager worker.
 //
-// This function should never call into the trackers (i.e. nodeHasNAD) as it would cause deadlock.
```
npinaeva, line 282: so that changed at some point, but the comment wasn't updated?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -330,11 +428,71 @@ func (c *nadController) OnNetworkRefChange(node, nadNamespacedName string, activ
...
 
+// getNetworkAndConnectedNetworks returns the provided network and all networks
+// directly connected to it through CNC connectivity.
+func (c *nadController) getNetworkAndConnectedNetworks(networkName string) []string {
+	if networkName == "" {
+		return nil
+	}
+	c.RLock()
+	defer c.RUnlock()
+	networks := sets.New[string](networkName)
```
npinaeva: why don't you make it `[]string` to avoid `UnsortedList()` call in the end?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -414,6 +585,244 @@ func (c *nadController) Stop() {
...
+		if networkName == "" {
+			continue
+		}
+		selectedNetworks.Insert(networkName)
+	}
+	return selectedNetworks, networkIDs
+}
+
+func buildCNCConnectedNetworks(selectedNetworksByCNC map[string]sets.Set[string]) map[string]sets.Set[string] {
+	connectedByNetwork := map[string]sets.Set[string]{}
```
npinaeva: shouls this be called connectedNetworks? it is network -> all connected networks right?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -330,11 +428,71 @@ func (c *nadController) OnNetworkRefChange(node, nadNamespacedName string, activ
...
+		if len(nadKeys) == 0 {
+			continue
+		}
+
+		active := false
+		if c.filterNADsOnNode != "" {
+			active = c.NodeHasNetwork(c.filterNADsOnNode, networkName)
+			if !active && changedNADActive && networkName == changedNetworkName {
+				active = true
+			}
```
npinaeva: I don't get this part: shouldn't it return the actual latest active state?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -330,11 +428,71 @@ func (c *nadController) OnNetworkRefChange(node, nadNamespacedName string, activ
...
+
+	seenNADs := sets.New[string]()
+	for _, networkName := range networkNames {
+		if networkName == "" {
+			continue
+		}
+		nadKeys := sets.New[string](c.GetNADKeysForNetwork(networkName)...)
+		if networkName == changedNetworkName && changedNADKey != "" {
+			nadKeys.Insert(changedNADKey)
+		}
```
npinaeva: this past I don't get: should `GetNADKeysForNetwork` already have that nad?

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -330,11 +428,71 @@ func (c *nadController) OnNetworkRefChange(node, nadNamespacedName string, activ
...
+		if len(nadKeys) == 0 {
+			continue
+		}
+
+		active := false
+		if c.filterNADsOnNode != "" {
```
npinaeva: when `c.filterNADsOnNode == ""` that means we don't do dynamic filtering at all? So is it the same as disabled dynamic filtering? i guess not, but I can't figure out the difference anymore

go-controller/pkg/networkmanager/nad_controller.go
trozet: OK. The exludedNAD was an optimization to avoid looking up the same NAD twice, but it seems like it causes more confusion than efficiency so I'll remove it.

go-controller/pkg/networkmanager/nad_controller.go
trozet: AI modified this. It looks it is safe to call into the trackers as they release their lock before they execute the networkRefChange code:
```
  c.cacheMutex.Unlock()
  if c.onNetworkRefChange != nil {
      for _, callback := range refChanges {
          c.onNetworkRefChange(...)
      }
  }

```

go-controller/pkg/networkmanager/nad_controller.go
trozet: Usually yeah, but there can be a case where nad controller is delayed sync and the trackers fire. I'll add a comment here to make it more obvious.

go-controller/pkg/networkmanager/nad_controller.go
trozet: filterNADsOnNode == "" means this networkmanager instance does not do local node-based NAD filtering. Cluster manager needs this to be empty, because it will not filter out NADs (it is not local to any node) and needs all NCCs to start for each network. Then in NCC it decides whether or not to allocate resources for the network on particular nodes by activity.

On the other hand the ovnkube-controller/ovnkube-node sides provide a local node, and then do not start network controllers (filter nads) if the node is inactive. It confused me as well when I went back to read the code. I'll change it to wrap it with a function usesLocalDynamicFiltering() with a better comment so it is more clear.

go-controller/pkg/networkmanager/nad_controller.go
trozet: it's the same as the other case where assume nad controller sync is delayed, so it does not have hte network yet, but the trackers have fired. This basically serves as a signal to say i know this network has now gone active so you need to reconcile it as active. I'll add a comment to make it more obvious.

trozet: The consequence of not doing this is the network would be considered inactive, which would set mark for removal. Then when NAD controller later syncs the NAD, it could accidentally be removed. I realized we should also clear the mark for removal when we sync the NAD and identify it has network, but we still need this fallback as well. I added a commit with this change.

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: so on cluserManager side it won't use NAD activity itself, but it still has to track which NAD is active where for its sub-controllers? I think this part also needs to go to the comment, because I couldn't understand why track it at all then

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -330,9 +406,74 @@ func (c *nadController) OnNetworkRefChange(node, nadNamespacedName string, activ
...
+			active = c.NodeHasNetwork(c.filterNADsOnNode, networkName)
+			if !active && changedNADActive && networkName == changedNetworkName {
+				active = true
+			}
+		}
+
+		for nadKey := range nadKeys {
+			if seenNADs.Has(nadKey) {
```
npinaeva: I am not sure how we expect to see the same nadKey for 2 different networks?

go-controller/pkg/networkmanager/nad_controller.go
trozet: that is in the comment I added: "Cluster-scoped
// instances keep all NADs rendered and leave per-node allocation decisions to
// their network controllers."

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: I thought this is covered by the order in the syncNAD: NAD is added to the network first, then `c.filter(nad)` is run, so either a fresh state will be fetched or nad at least will be in the map?

go-controller/pkg/networkmanager/nad_controller.go
npinaeva: so do we even need to call reconcile when `!c.usesLocalDynamicFiltering()`? or could we just return immediately?

trozet: we still need to reconcile. !c.usesLocalDynamicFiltering() means do not locally filter/render NADs for a specific node. Cluster manager case: we still need to syncNAD and launch/update NCCs for a UDN.

Besides filtering rendering the network controller, syncNAD needs to be called to update all the NAD/network cache and to notify NAD reconcilers.

go-controller/pkg/networkmanager/nad_controller.go
trozet: cannot happen. will clean up

go-controller/pkg/networkmanager/nad_controller.go
trozet: In the syncNAD path yes you are right, but there is another path: the egress ip/pod tracker. Those trackers can fire before network manager informer has received the NAD. So the imagine network manager informer is lagging and does not have the NAD processed yet. Trackers fire -> OnNetworkRefChange -> reconcileNetworkActivity. Now GetNADKeysForNetwork has nothing and because we triggered this through OnNetworkRefChange we pass `changedNADActive bool, changedNetworkName, changedNADKey string` so we know the nad that needs reconciling and can ask network manager to go ahead and reconcile it.

trozet:
next push will just be rebase

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -782,13 +1248,16 @@ func (c *nadController) syncNAD(key string, nad *nettypes.NetworkAttachmentDefin
 	}
 
 	shouldNetworkExist := true
-	if c.filterNADsOnNode != "" {
+	if c.usesLocalDynamicFiltering() {
```
npinaeva, line 1251: may be important to say here that calling `c.filter` after the  `	c.addNADToNetworkLocked(networkName, key)` is important to avoid races with CNC controller that decides which NADs need to be reconciled based on the network to nad mapping

npinaeva (dismissed):

npinaeva (approved):

PR #6043, opened 2026-03-01
Fix cel bug evpn

tssurya:
crd: fix unanchored regex in RouteTarget CEL validation rules
CEL's matches() performs partial/substring matching, so matches('[0-9]+')
accepted invalid strings like '999.999.999.999' (matched on the substring
'999') instead of rejecting them as non-numeric.

Anchor all RouteTargetString matches() calls with '^' and '$' so the
entire field value must consist of digits, correctly rejecting IPv4-like
strings when a plain integer is expected.

Regenerate the ClusterUserDefinedNetwork CRD YAML.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced route target string validation with stricter pattern matching and improved IP address detection.

* **Tests**
  * Simplified EVPN-related API validation test cases by streamlining test parameters.


tssurya:
closing this and will be fixed as part of https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6182

PR #6054, opened 2026-03-04
e2e EIP failover case for egress node reboot

huiran0826:
## 📑 Description
This PR adds a new end-to-end test case  to validate that EgressIP fails over correctly when an egress node is rebooted and pod connectivity continually works after failover.

  Key Validations

  1. EgressIP fails over when egress node rebooted （node is in NotReady status）
  2. Pod connectivity works with EgressIP before and after failover
  3. Rebooted node doesn't send GARP/ND for IPs it doesn't own (prevents network disruption)
  4. EgressIP stays on the failover node after original node recovers
  5. IPv4 (GARP) and IPv6 (Neighbor Advertisement) are both tested
  6. New pod created after failover works for EgressIP

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end egress IP failover tests validating failover on node reboot, pre/post-failover connectivity, ARP/ND behavior, source-IP consistency, recovery validation, and connectivity for new pods after recovery.
  * Note: the failover scenario was duplicated within the same suite, creating redundant parallel test instances.

### DIFF
--- a/test/e2e/egressip.go
+++ b/test/e2e/egressip.go
@@ -863,6 +863,11 @@ var _ = ginkgo.DescribeTableSubtree("e2e egress IP validation", feature.EgressIP
 	ginkgo.AfterEach(func() {
 		// ensure all nodes are ready and reachable before any other cleanup;
 		// tests may have left nodes NotReady or unreachable intentionally
+		// Skip restoration if test was skipped in BeforeEach (node names won't be initialized)
+		if egress1Node.name == "" || egress2Node.name == "" {
+			framework.Logf("Skipping node restoration as test was skipped in BeforeEach")
+			return
+		}
 		for _, node := range []string{egress1Node.name, egress2Node.name} {
 			setNodeReady(providerCtx, node, true)
 			setNodeReachable(node, true)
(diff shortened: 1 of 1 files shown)


test/e2e/egressip.go
```diff
@@ -2974,6 +2979,247 @@ spec:
...
+	13. Start egress node1
+	14. Wait for egrss node1 to become ready
+	15. Verify no GARP/ND sent out from egress node 1 which was shutdown and started
+	16. Verify the EIP remains on egress node2 after node 1 becomes ready
+	17. Check connectivity again after node 1 becomes ready from pod to an external node and verify that the srcIP is the expected egressIP
+	18. Create a new pod matching the EgressIP
+	19. Verify connectivity works with EgressIP for the new create pod after node reboot
+	*/
+	ginkgo.It("should failover correctly when egress node reboots", func() {
+		if isUserDefinedNetwork(netConfigParams) {
```
arghosh93: Why you are skipping this for UDNs? I this file we are skipping when UDN is enabled mostly for secondary EIP tests.

test/e2e/egressip.go
```diff
@@ -863,6 +863,11 @@ var _ = ginkgo.DescribeTableSubtree("e2e egress IP validation", feature.EgressIP
 	ginkgo.AfterEach(func() {
 		// ensure all nodes are ready and reachable before any other cleanup;
 		// tests may have left nodes NotReady or unreachable intentionally
+		// Skip restoration if test was skipped in BeforeEach (node names won't be initialized)
+		if egress1Node.name == "" || egress2Node.name == "" {
```
arghosh93, line 867: Do we need this? We wont reach till ginkgo.AfterEach() if the test is being skipped from ginkgo.BeforeEach().

test/e2e/egressip.go
```diff
@@ -2974,6 +2979,247 @@ spec:
...
+		ginkgo.By("4. Create one pod matching the EgressIP")
+		_, err = createGenericPodWithLabel(f, pod1Name, pod1Node.name, f.Namespace.Name, getAgnHostHTTPPortBindFullCMD(clusterNetworkHTTPPort), podEgressLabel)
+		framework.ExpectNoError(err, "failed to create pod %s/%s", f.Namespace.Name, pod1Name)
+		framework.Logf("Created pod %s on node %s", pod1Name, pod1Node.name)
+
+		ginkgo.By("5. Verify connectivity works with EgressIP on node1 before shutdown")
+		err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP.String()}).WithContext())
+		framework.ExpectNoError(err, "Check connectivity from pod to an external node and verify that the srcIP is the expected egressIP, failed: %v", err)
+
+		ginkgo.By("6. Add the \"k8s.ovn.org/egress-assignable\" label to secondary node")
```
arghosh93: Can step 0 and 6 be merged?

test/e2e/egressip.go
```diff
@@ -2974,6 +2979,247 @@ spec:
...
+			"bash", "-c", tcpdumpCommand,
+		})
+		framework.ExpectNoError(err, "failed to start tcpdump in external container %v", primaryTargetExternalContainer)
+
+		ginkgo.By(fmt.Sprintf("10. Shut down the egress node %s", egress1Node.name))
+		framework.Logf("Shutting down node %s", egress1Node.name)
+		err = infraprovider.Get().ShutdownNode(egress1Node.name)
+		framework.ExpectNoError(err, "Failed to shutdown node %s", egress1Node.name)
+
+		defer func() {
```
arghosh93: Why we are trying to start the node and verify Ready state of the node again in the defer function? I think step 14 and 15 does the same thing.

test/e2e/egressip.go
```diff
@@ -2974,6 +2979,247 @@ spec:
...
+		ginkgo.By("5. Verify connectivity works with EgressIP on node1 before shutdown")
+		err = wait.PollUntilContextTimeout(context.TODO(), retryInterval, retryTimeout, true, targetExternalContainerAndTest(primaryTargetExternalContainer, podNamespace.Name, pod1Name, true, []string{egressIP.String()}).WithContext())
+		framework.ExpectNoError(err, "Check connectivity from pod to an external node and verify that the srcIP is the expected egressIP, failed: %v", err)
+
+		ginkgo.By("6. Add the \"k8s.ovn.org/egress-assignable\" label to secondary node")
+		e2enode.AddOrUpdateLabelOnNode(f.ClientSet, egress2Node.name, "k8s.ovn.org/egress-assignable", "dummy")
+		defer e2enode.RemoveLabelOffNode(f.ClientSet, egress2Node.name, "k8s.ovn.org/egress-assignable")
+
+		ginkgo.By("7. Install tcpdump in external container for later use")
+		_, err = infraprovider.Get().ExecExternalContainerCommand(primaryTargetExternalContainer, []string{
```
arghosh93, line 3069: this seems expensive. Can you verify whether the image for the external container already have tcpdump utility or not and skip this step if the image already have it?

test/e2e/egressip.go
huiran0826: Yes, that specific step failed in some ci lanes that ginkgo.BeforeEach() was skipped but ginkgo.afterEach would still be called however with variables not initialized ,  below is an example local run log
```
[SKIPPED] in [BeforeEach] - /home/huirwang/workspace/code/ovn-kubernetes/test/e2e/egressip.go:742 @ 04/03/26 17:53:24.485
[FAILED] in [AfterEach] - /home/huirwang/workspace/code/ovn-kubernetes/test/e2e/egressip.go:520 @ 04/03/26 17:53:24.508
S [SKIPPED] [0.039 seconds]
```

test/e2e/egressip.go
huiran0826: Updated it, reduce the steps

test/e2e/egressip.go
huiran0826: ok, removed  this defer function.

test/e2e/egressip.go
huiran0826: Updated the command, will check if tcpdump there or not, if not, will install it.

test/e2e/egressip.go
huiran0826: Yes, I understand that generally UDN is skipped for secondary EIP, however, according to the udn matrix the node reboot will happen several times for each udn ci lane. Node reboot is an expensive test scenario, considering the side effect of node reboots especially multiple times reboot which could cause the ci lane not stable, I think the risk might outweighs the benefit, so currently focus on common default network.  Let me know if you have any other thoughts?

PR #6055, opened 2026-03-04
dpulease: fix lease ownerReference and log healthy recovery

jkary:
## Summary
- Set `blockOwnerDeletion` to `false` on the DPU lease ownerReference. The ovnkube-node service account typically lacks permission to set finalizers on Node objects, causing a crash when `blockOwnerDeletion` is `true`.
- Log at debug level when the DPU lease transitions from unhealthy back to healthy, matching the existing "marked unhealthy" warning for better observability. The log only fires on actual recovery, not on initial startup.

## Test plan
- [x] Unit tests pass (`PKGS=./pkg/node/dpulease/... make test`)
- [x] Verified in DPF cross-cluster deployment: lease creation succeeds without finalizer permission errors
- [x] Verified lease recovery log appears after DPU pod restart

## Release Notes

* **Bug Fixes**
  * Improved lease health status reporting to reduce redundant logging and ensure accurate state transitions.
  * Fixed lease ownership handling to properly manage object lifecycle and prevent orphaned resources.

* **Tests**
  * Added comprehensive test coverage for status transition scenarios and ownership correction logic.

### DIFF
--- a/go-controller/pkg/node/dpulease/manager.go
+++ b/go-controller/pkg/node/dpulease/manager.go
@@ -200,7 +200,12 @@ func (m *Manager) setStatus(reason string, ready bool) {
 	m.statusMu.Lock()
 	defer m.statusMu.Unlock()
 
-	if m.ready != ready || m.reason != reason {
+	prevReady := m.ready
+	prevReason := m.reason
+	if prevReady != ready || prevReason != reason {
+		if ready && !prevReady && prevReason != "" {
+			klog.V(4).Infof("DPU lease %s marked healthy", m.leaseName())
+		}
 		m.ready = ready
 		m.reason = reason
 	}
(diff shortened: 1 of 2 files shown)


aserdean:
IMO I don't think this is needed since the user can see this from the node status.

If we do decide to log this message this should be at debug level. 

Adding @trozet to the thread to see if he likes this approach.

jkary:
Addressed in latest push:
- Changed to debug level (`klog.V(4).Infof`) per Alin's feedback
- Guarded with `prevReason != ""` so it only fires on actual recovery, not startup
- Added `TestSetStatusTransitions` unit test

jkary:
Addressed both points:

1. **Recovery log on startup**: Guarded with `prevReason != ""` so it only fires on actual recovery from an unhealthy state, not on initial boot.
2. **Self-healing existing leases**: `updateLeaseSpec` now iterates existing ownerReferences and fixes `BlockOwnerDeletion` from `true`/`nil` to `false` on matching refs, rather than only appending when missing. Added `TestUpdateLeaseFixesBlockOwnerDeletion` to cover this path.

trozet:
thanks @jkary

PR #6051, opened 2026-03-03
Bump MetalLB to 0.15.3; Bump FRR to 10.5.3+; e2e: do not skip frr service coredumps

booxter:
## 📑 Description

Bump metallb to 0.15.3 that has 10.1+ FRR stack with necessary fixes for segfaults in Alpine env.

Bump FRR images to 10.5.3 to include more fixes for coredumps.

Fixes #5782
Fixes #5453

## ✅ Checks

- [ ] My code requires changes to the documentation
- [ ] if so, I have updated the documentation as required
- [ ] My code requires tests
- [ ] if so, I have added and/or updated the tests as required
- [ ] All the tests have passed in the CI 

## Summary by CodeRabbit

* **Tests**
  * Coredump handling simplified: all coredump files are now collected and reported (no exclusions) but no longer fail test execution.
  * Added wait logic in e2e tests to ensure BGP routes are present before contacting load‑balanced services.

* **Chores**
  * Networking/demo tooling updated: MetalLB and FRR versions bumped; installation/patching steps simplified.

PR #6057, opened 2026-03-05
no-overlay: Add managed routing mode for BGP topology

pliurh:
## 📑 Description

Adds managed routing mode for no-overlay networks, allowing OVN-Kubernetes to automatically configure an internal BGP fabric without requiring external BGP infrastructure (OKEP-5259).

- Add frr-namespace config flag to ManagedBGPConfig (defaults to frr-k8s-system)
- Add managedbgp controller that watches nodes and maintains a full-mesh iBGP topology via a base FRRConfiguration, ensures a managed RouteAdvertisement exists for the default network, and filters self-peering in per-node configs
- Add KIND deployment support (--enable-no-overlay managed) for both kind.sh and kind-helm.sh, including Helm values, RBAC updates, and [bgp-managed] config section
- Enable managed routing mode in one CI lane

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Managed BGP routing for no-overlay mode with configurable AS, topology, FRR namespace, and option to run FRR in-cluster; new flags to control managed routing and outbound SNAT.

* **Tests**
  * CI matrix expanded to cover no-overlay variants (SNAT-enabled and managed).
  * New comprehensive unit tests for the managed BGP controller.

* **Chores**
  * CLI/Helm/values and scripts updated to expose new options and improve FRR/Kubernetes orchestration and readiness.

* **Bug Fixes**
  * Suppress self-neighbors in route advertisement generation; RBAC updated for routeadvertisements.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -469,8 +469,8 @@ jobs:
           - {"target": "shard-conformance", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "snatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "shard-conformance", "ha": "HA",   "gateway-mode": "shared", "ipfamily": "ipv4",      "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default"}
           - {"target": "shard-conformance", "ha": "HA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default"}
-          - {"target": "shard-conformance", "ha": "HA",   "gateway-mode": "shared",  "ipfamily": "ipv6",     "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "no-overlay": "true"}
-          - {"target": "shard-conformance", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "no-overlay": "snat-enabled"}
+          - {"target": "shard-conformance", "ha": "HA",   "gateway-mode": "shared", "ipfamily": "ipv6",      "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "advertise-default", "no-overlay": "true"}
+          - {"target": "shard-conformance", "ha": "noHA", "gateway-mode": "local",  "ipfamily": "dualstack", "disable-snat-multiple-gws": "noSnatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones", "routeadvertisements": "managed", "no-overlay": "managed"}
           - {"target": "shard-conformance", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv6",      "disable-snat-multiple-gws": "snatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "shard-conformance", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4",      "disable-snat-multiple-gws": "snatGW",   "second-bridge": "1br", "ic": "ic-single-node-zones"}
           - {"target": "control-plane",     "ha": "HA",   "gateway-mode": "shared", "ipfamily": "ipv6",      "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-disabled"}
@@ -547,7 +547,8 @@ jobs:
       MULTI_POD_SUBNET: true
       DYNAMIC_UDN_ALLOCATION: "${{ matrix.target == 'network-segmentation-dynamic' }}"
       ENABLE_NO_OVERLAY: "${{ matrix.no-overlay != '' }}"
-      ENABLE_NO_OVERLAY_OUTBOUND_SNAT: "${{ matrix.no-overlay == 'snat-enabled' }}"
+      ENABLE_NO_OVERLAY_OUTBOUND_SNAT: "${{ matrix.no-overlay == 'snat-enabled' || matrix.no-overlay == 'managed' }}"
+      ENABLE_NO_OVERLAY_MANAGED_ROUTING: "${{ matrix.no-overlay == 'managed' }}"
     steps:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v4


go-controller/pkg/util/multi_network.go
```diff
@@ -870,6 +876,12 @@ func (nInfo *userDefinedNetInfo) Transport() string {
 	return nInfo.transport
 }
 
+// OutboundSNAT() string returns the outbound SNAT configuration for this network when using no-overlay transport.
+func (nInfo *userDefinedNetInfo) OutboundSNAT() string {
+	// TODO: implement per-network no-overlay outbound SNAT configuration
+	return ""
+}
```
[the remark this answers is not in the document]
pliurh: This will be implemented in a separated PR.

pliurh:
@coderabbitai review again

pliurh:
@coderabbitai review

pliurh:
@coderabbitai review again

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	// Spec was modified
+	return !reflect.DeepEqual(oldRA.Spec, newRA.Spec)
+}
+
+func (c *Controller) reconcileNode(_ string) error {
+	if config.ManagedBGP.Topology != config.ManagedBGPTopologyFullMesh {
+		return nil
+	}
```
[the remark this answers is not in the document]
pliurh: Managed mode can be enabled for CUDN. We shall not only check if it's configured for the default network.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -736,8 +736,20 @@ func (c *Controller) generateFRRConfiguration(
...
+			return nil, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		}
+		nodeV4, nodeV6 := util.GetNodeInternalAddrs(node)
+
 		targetRouter.Neighbors = make([]frrtypes.Neighbor, 0, len(source.Spec.BGP.Routers[i].Neighbors))
 		for _, neighbor := range source.Spec.BGP.Routers[i].Neighbors {
+			// Skip neighbors that are the node itself
+			if (nodeV4 != nil && neighbor.Address == nodeV4.String()) || (nodeV6 != nil && neighbor.Address == nodeV6.String()) {
+				continue
+			}
```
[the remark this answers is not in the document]
pliurh: It won't happen.

go-controller/pkg/clustermanager/clustermanager.go
```diff
@@ -260,6 +263,9 @@ func (cm *ClusterManager) Start(ctx context.Context) error {
 	}
 
 	if cm.raController != nil {
+		if err := cm.managedBGPController.Start(); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
pliurh: It will be implemented by a separate PR.

contrib/kind-common.sh
```diff
@@ -1131,14 +1136,23 @@ destroy_bgp() {
...
 
 install_frr_k8s() {
+  local bgp_port=${1:-0}
   echo "Installing frr-k8s ..."
   clone_frr
 
   # apply frr-k8s
+  if [ "${bgp_port}" -ne 0 ]; then
+    sed -i "s/bgpd_options=\"\(.*\)-p 0\(.*\)\"/bgpd_options=\"\1-p ${bgp_port}\2\"/g" "${FRR_TMP_DIR}"/frr-k8s/config/all-in-one/frr-k8s.yaml
+  fi
```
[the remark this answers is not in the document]
pliurh: We use a specific version of frr-k8s. The frr-k8s.yaml won't be changed suddenly.

go-controller/pkg/config/config.go
```diff
@@ -676,6 +676,9 @@ type ManagedBGPConfig struct {
 	// Supported values: "full-mesh".
 	// Required when transport=no-overlay and routing=managed.
 	Topology string `gcfg:"topology"`
+	// FRRNamespace specifies the namespace where FRR-K8s FRRConfiguration resources are created
+	// when routing is managed.
+	FRRNamespace string `gcfg:"frr-namespace"`
```
trozet, line 683: hmm how come this is specific to managed BGP mode? I'm just thinking that today we already do BGP and we create FRR-K8S CRs, and we don't know the frr namespace...why is this needed now?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+)
+
+func managedHashedName(s string) string {
+	h := fnv.New32a()
+	_, _ = h.Write([]byte(s))
+	return fmt.Sprintf("%s%x", managedNamePrefix, h.Sum32())
+}
+
+// ManagedRouteAdvertisementName returns the name of the managed RouteAdvertisement
+// for the given network name. The name follows the pattern "ovnk-managed-<hash>"
```
trozet, line 56: I guess you are hashing cause you are worried about exceeding the name length? Why not use the network ID instead, then you dont have to worry about any potential hash collisions?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+	if oldNode == nil || newNode == nil {
+		return true
+	}
+	// We care about node IP changes
+	oldV4, oldV6 := util.GetNodeInternalAddrs(oldNode)
+	newV4, newV6 := util.GetNodeInternalAddrs(newNode)
+	return !reflect.DeepEqual(oldV4, newV4) || !reflect.DeepEqual(oldV6, newV6)
+}
+
+func (c *Controller) managedRANeedsUpdate(oldRA, newRA *ratypes.RouteAdvertisements) bool {
```
trozet, line 162: do we expect someone to update the RA other than us?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+	// excluding self-peering.
+	if err := c.ensureBaseFRRConfiguration(nodes); err != nil {
+		klog.Errorf("Failed to ensure base FRRConfiguration: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+func (c *Controller) ensureBaseFRRConfiguration(allNodes []*corev1.Node) error {
```
trozet, line 252: nit: some doctext on this function would be good about the configuration it is creating as base. I'm guessing it is just BGP peering config between each node  right?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+		netSelector = apitypes.NetworkSelectors{{NetworkSelectionType: apitypes.DefaultNetwork}}
+	} else {
+		// CUDN: build a selector that matches the CUDN by the managed network label
+		netSelector = apitypes.NetworkSelectors{{
+			NetworkSelectionType: apitypes.ClusterUserDefinedNetworks,
+			ClusterUserDefinedNetworkSelector: &apitypes.ClusterUserDefinedNetworkSelector{
+				NetworkSelector: metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						ManagedRANetworkLabel: networkName,
```
trozet, line 376: so we put this label on the RA and on the CUDN?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+			},
+			// nodeSelector must select all nodes for PodNetwork
+			NodeSelector: metav1.LabelSelector{},
+		},
+	}
+
+	existing, err := c.raLister.Get(raName)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			klog.Infof("Creating managed RouteAdvertisement %s", raName)
```
trozet: these logs have the raName, which is a hash. It is hard to identify what network it belongs to. Can you also add the networkName in the log? Also add it for other logs in this function.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+			if err != nil && !apierrors.IsAlreadyExists(err) {
+				return err
+			}
+			return nil
+		}
+		return err
+	}
+
+	needsUpdate := !reflect.DeepEqual(existing.Spec, ra.Spec) ||
+		existing.Labels[ManagedRANetworkLabel] != networkName
```
trozet, line 422: how does this condition happen, where we have an raName that has a label with a different network? hash collision? Does this mean something in our state machine has gone seriously wrong and we should log an error at least?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -0,0 +1,482 @@
...
+		_, err = c.raClient.K8sV1().RouteAdvertisements().Update(context.TODO(), updated, metav1.UpdateOptions{
+			FieldManager: fieldManager,
+		})
+		return err
+	}
+
+	return nil
+}
+
+// cleanupManagedResources removes the default network's managed RouteAdvertisement
```
trozet, line 440: why does it only clean up default network and not other networks as well?

go-controller/pkg/clustermanager/routeadvertisements/controller.go
```diff
@@ -736,15 +736,33 @@ func (c *Controller) generateFRRConfiguration(
...
+			return nil, fmt.Errorf("failed to get node %s: %w", nodeName, err)
+		}
+		nodeIfAddr, err := util.GetNodeIfAddrAnnotation(node)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get node %s primary interface address annotation: %w", nodeName, err)
+		}
+		// Strip CIDR mask to get bare IP strings for neighbor comparison.
+		nodeV4, _, _ := strings.Cut(nodeIfAddr.IPv4, "/")
+		nodeV6, _, _ := strings.Cut(nodeIfAddr.IPv6, "/")
+
```
trozet, line 756: what are we doing this for? can we add comments?

are you trying to skip generating FRR RA for FRR configurations that are managed?

contrib/kind-common.sh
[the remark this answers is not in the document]
trozet: @booxter fyi

contrib/kind.sh
```diff
@@ -1068,8 +1075,16 @@ if [ "$OVN_ENABLE_DNSNAMERESOLVER" == true ]; then
     update_coredns_deployment_image
 fi
 if [ "$ENABLE_ROUTE_ADVERTISEMENTS" == true ]; then
-  deploy_frr_external_container
-  deploy_bgp_external_server
+  frr_port=0
+  if [ "$ENABLE_NO_OVERLAY_MANAGED_ROUTING" == true ]; then
+    # Enable bgp port listening on node, required for managed mode. FRR will listen on port 179 to receive BGP updates from other nodes.
+    frr_port=179
```
trozet, line 1081: why is this only needed for managed? Wouldn't any BGP use with FRR require this port to be open so that BGP neighborship can form with external BGP peers? Isn't 179 enabled by default with FRR when BGP is used?

helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
```diff
@@ -185,10 +184,19 @@ rules:
...
       verbs: [ "get", "list", "watch" ]
+    {{- if or (eq (hasKey .Values.global "enableRouteAdvertisements" | ternary .Values.global.enableRouteAdvertisements false) true) (eq (hasKey .Values.global "enableNoOverlayManagedRouting" | ternary .Values.global.enableNoOverlayManagedRouting false) true) }}
+    - apiGroups: ["k8s.ovn.org"]
+      resources:
+          - routeadvertisements
+      verbs: [ "get", "list", "watch" ]
+    - apiGroups: ["k8s.ovn.org"]
+      resources:
+          - routeadvertisements/status
+      verbs: [ "patch", "update" ]
```
trozet: why is this needed on the node?

helm/ovn-kubernetes/values-single-node-zone.yaml
```diff
@@ -89,6 +89,14 @@ global:
   enableNoOverlay: false
   # -- Configure to enable SNAT for outbound traffic in no-overlay mode
   enableNoOverlaySnat: false
+  # -- Configure to enable managed routing mode for no-overlay (automatically configures internal BGP fabric)
```
trozet, line 92: can you also please add to the values-single-node-zone-dpu as well?

trozet (changes requested):

contrib/kind-common.sh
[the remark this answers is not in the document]
booxter: @pliurh It assumes whoever bumps the version next will be smart enough to check this code still works though. I'd rather not make this assumption and let the replacement code fail fast, making sure it's updated. (I'm saying it as someone who recently committed a change with a similar `sed -i` instruction that [turned up ineffective](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6051/changes/70cfa8cbbdc85a10e63cd06513dee9e39859bcf3).)

go-controller/pkg/config/config.go
pliurh: Today, OVN-K generates the per-node FRRConfiguration in the same namespace as the base FRRConfiguration, which is created by the user. So OVN-K doesn't need to know the frr namespace. However, in the managed mode, the base FRRConfiguration is generated by OVN-K.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: The idea is to give the user a hint that these objects are automatically generated/maintained and shall not be modified manually. The hash is generated with the CUDN name, which is unique. So there will be hash collisions.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: No. This is to revert the unexpected change.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: yes

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: This PR only implements the managed mode for CDN. I'll create a separate PR for CUDN, the function will clean managed resources for CUDNs in that PR.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: This is to prevent the RA from being updated/deleted accidentally by users.

go-controller/pkg/clustermanager/routeadvertisements/controller.go
pliurh: To generate the full-mesh FRRConfiguration, we need to get the primary IP address of each node. But we don't want a node to have its own IP in its peer list. This is to exclude the own IP.

contrib/kind.sh
pliurh: In the K8S-FRR manifest `bgpd_options: -p 0`, it means FRR won't accept any inbound BGP connections but can still initiate outbound sessions to configured neighbors. We have to enable it by setting `-p 179`.

helm/ovn-kubernetes/templates/rbac-ovnkube-node.yaml
pliurh: No need. It's a mistake.

pliurh:
Update to address @trozet 's comments.

go-controller/pkg/clustermanager/managedbgp/controller.go
trozet: but wouldn't the network ID be simpler? then the name has more meaning?

contrib/kind.sh
trozet: oh that is surprising

go-controller/pkg/config/config.go
trozet: nit if you update again might be good to add another sentence that just says for unmanaged mode the FRRNamespace is determined automatically by detecting user FRR-K8S CRs.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: For the CDN, it's ok. But it will introduce a racing condition for the CUDN scenario. The network ID is given at `handleNetworkAnnotations`, when the NAD is created for the CUDN. If we use network ID, we have to wait for it to be available.

contrib/kind.sh
pliurh: Someone told me that it is a security consideration. https://github.com/metallb/frr-k8s/blob/62739ddaf65d0d95c6c6d18b1cc1dc7559d6b8dc/config/all-in-one/frr-k8s.yaml#L1055

go-controller/pkg/clustermanager/managedbgp/controller.go
trozet: OK but the hash collision is non-negligible here. With 1000 CUDNs the probability is .01%. I think to solve this we should move to a 64 bit hash. This changes the probability to 0.0000000000027%.

```
func managedHashedName(s string) string {
	h := fnv.New64a() // 64-bit FNV-1a
	_, _ = h.Write([]byte(s))
	return fmt.Sprintf("%s%x", managedNamePrefix, h.Sum64())
}
```

pliurh:
udpate to address https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6057#discussion_r2956472402 and https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6057#discussion_r2965907788

PR #6061, opened 2026-03-05
Limit CPU usage for lint job

ricky-rav:
By default, golangci-lint uses all available CPUs, which can freeze the machine when other workloads are running. 

Limit concurrency to half the available CPUs via `--concurrency` flag, and apply `--cpus` to the container runtime as well. The default can be overridden by setting the `LINT_CONCURRENCY` environment variable.

The CI lint job is unaffected as it uses the golangci-lint-action GitHub Action directly, not hack/lint.sh.

## Summary by CodeRabbit

* **Chores**
  * Improved linting infrastructure with enhanced concurrency handling and Docker resource constraints for more efficient development builds.

### DIFF
--- a/go-controller/hack/lint.sh
+++ b/go-controller/hack/lint.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 VERSION=v2.5.0
 : "${GOLANGCI_LINT_VERSION:=$VERSION}"
-extra_flags=(--verbose --modules-download-mode=vendor --timeout=15m0s)
+: "${LINT_CONCURRENCY:=$(( $(nproc) * 2 / 3 ))}"
+extra_flags=(--verbose --modules-download-mode=vendor --timeout=15m0s --concurrency="${LINT_CONCURRENCY}")
 if [ "$#" -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


trozet:
@ricky-rav where did you see this issue? Lint doesn't take very long so I'm not opposed to it, but I'm curious. Also, wouldn't make check have a similar issue? Wont it use all cores by default? I think that would be more intensive.

ricky-rav:
@trozet I've seen this on my work laptop for years actually. I upgraded my laptop last fall, it now has twice as many cores and twice as much memory. If I'm running something else that is not trivial and takes some fraction of CPUs, then the linter, which systematically tries to fully use all CPUs, often makes the whole machine freeze... sometimes just for less than a minute, sometimes for more and I have to reboot. 

I think what's important is to limit CPU usage to some value instead of giving the linter total freedom.

I've just seen that with two thirds of CPU usage we go from the current 30 seconds to 38 seconds of execution time, instead of 1 minute with 50% CPUs. So maybe that's a better compromise that can almost go unnoticed.

If nobody else has the same problem, I can keep local changes in my local repo and no problem either. :)

PR #3478, opened 2023-03-09
Hypershift live migration

qinqon:
**- What this PR does and why is it needed**
Add support to hypershift kubevirt provider live migration by doing the following at hypershift workers virt-launcher pods:
- Skip CNI network configuration
- Add DHCP options at pod's LSP
- Configure ARP Proxy at router type LSP
- Use point to point routes to enroute pod's traffic
- Copy the ovn pod annotations from source to target virt-launchers during live migration.

**- Special notes for reviewers**
Changes for IC not included at the enhancement:
- Add a static route matching cluster wide cidr with nexthop to gw router.
- At remote zone add a ptp static route to the transit switch port
- Prevent remote zone ip deallocation for VMs
- Do remote zone static routes and DHCP Options cleanup after live migration.
- Add sync function that will remove policy/routes that do not belong to vm and zone

Depends on:
- k8s 1.26.3:
  - https://github.com/kubernetes/kubernetes/pull/115907
- pin ovn to 23.06:
  -   https://github.com/ovn-org/ovn-kubernetes/pull/3639
- future kubevirt 1.0:
  - https://github.com/kubevirt/kubevirt/pull/9290
  - https://github.com/kubevirt/kubevirt/pull/9603
- activate IC at CI
  - https://github.com/ovn-org/ovn-kubernetes/pull/3366  

**- Follow ups**
- Check if we can configure ipv6 autoconf with ovn
  - this  may help for the default gw https://bugzilla.redhat.com/show_bug.cgi?id=2223655
- Check why post-copy live migration test do not work at github actions.
  - `event for worker1: {virt-handler ovn-worker2} Migrated: VirtualMachineInstance migration uid 75effe84-8f58-4890-b8ac-6995d74bd4ae failed. reason:Live migration failed error encountered during MigrateToURI3 libvirt api call: virError(Code=1, Domain=10, Message='internal error: unable to execute QEMU command 'migrate-set-capabilities': Postcopy is not supported')` 
- Use address set to reduce the number of logical flows
- Configure dhcp options dns using pod's resolver.conf or explicit from CNO
- IPv6 single stack:
  - OVN do not support autoconf so it has to be disabled at node to receive an ipv6 address
  - OVN only support delivering dns_server (which is not working for us)
  - ARP proxy default gw and hostname will have to be configured with ignition 
- Move getExpectedSwitches to free network
- Try to move EnsureRoutingForVM to addLogicalPortToNetwork to prevent readiness until routing is not in place.
[...]

PR #6063, opened 2026-03-05
virt: fix SyncVirtualMachines deleting UDN DHCP options during startup

maiqueb:
## 📑 Description

During cluster upgrade, the default network controller's SyncVirtualMachines deletes DHCP options created by UDN controllers because the ownsItAndIsOrphanOrWrongZone predicate does not filter by owner controller. This causes VMs on primary UDNs (layer2) to lose their IP when the DHCP lease expires.

Add a controllerName parameter to ownsItAndIsOrphanOrWrongZone and SyncVirtualMachines so that only resources owned by the calling controller are considered for deletion. The default network controller now passes its own name, preventing it from deleting DHCP options owned by UDN controllers.

Fixes #

## ✅ Checks

- [x] My code requires tests

## How to verify it

Run the newly added unit tests.

In production systems, you can use the following script to verify it (running without this PR will reproduce the issue): 
```bash
#!/bin/bash
#
# Reproducer for: VM DHCP options lost on primary UDN after ovnkube-controller restart
#
# Prerequisites:
#   - A KIND cluster with ovn-kubernetes (interconnect / single-node-zone mode)
#   - KubeVirt installed
#   - A running VM on a primary UDN (layer2, ovn-k8s-cni-overlay)
#
# Usage:
#   ./reproduce-dhcp-loss.sh <vm-pod-namespace> <vm-pod-name>
#
# Example:
#   ./reproduce-dhcp-loss.sh green virt-launcher-my-vm-xxxxx
#
set -euo pipefail

NS="${1:?Usage: $0 <vm-pod-namespace> <vm-pod-name>}"
POD="${2:?Usage: $0 <vm-pod-namespace> <vm-pod-name>}"

OVN_NS="ovn-kubernetes"

# --- helpers ---

nbctl() {
    # Try ovnkube-db pod first (standard mode), fall back to ovnkube-node (single-node-zone)
    local db_pod
    db_pod=$(kubectl get pods -n "$OVN_NS" -l name=ovnkube-db \
        -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)

    if [[ -n "$db_pod" ]]; then
        kubectl exec -n "$OVN_NS" "$db_pod" -c nb-ovsdb -- \
            ovn-nbctl --no-leader-only "$@"
    else
        # single-node-zone: nbctl available in ovnkube-node pods
        local node_pod
        node_pod=$(kubectl get pods -n "$OVN_NS" -l app=ovnkube-node \
            -o jsonpath='{.items[0].metadata.name}')
        kubectl exec -n "$OVN_NS" "$node_pod" -c ovnkube-controller -- \
            ovn-nbctl --no-leader-only "$@"
    fi
}

get_lsp_name() {
    local ns="$1" pod="$2"
    # For UDN pods the LSP name is: <namespace>_<pod-name>
    # But it may also include the NAD name. Let's find it from the NBDB.
    nbctl --format=table --columns=name find Logical_Switch_Port \
        external_ids:namespace="$ns" \
        | grep "$pod" | awk '{print $1}' | head -1
}

get_dhcp_options_for_lsp() {
    local lsp_name="$1"
    echo "--- LSP: $lsp_name ---"
    echo ""
    echo "  dhcpv4_options:"
    local v4_uuid
    v4_uuid=$(nbctl get Logical_Switch_Port "$lsp_name" dhcpv4_options 2>/dev/null || echo "[]")
    echo "    uuid: $v4_uuid"
    if [[ "$v4_uuid" != "[]" ]]; then
        local clean_uuid
        clean_uuid=$(echo "$v4_uuid" | tr -d '[]')
        echo "    details:"
        nbctl list DHCP_Options "$clean_uuid" 2>/dev/null | sed 's/^/      /'
    fi

    echo ""
    echo "  dhcpv6_options:"
    local v6_uuid
    v6_uuid=$(nbctl get Logical_Switch_Port "$lsp_name" dhcpv6_options 2>/dev/null || echo "[]")
    echo "    uuid: $v6_uuid"
    if [[ "$v6_uuid" != "[]" ]]; then
        local clean_uuid
        clean_uuid=$(echo "$v6_uuid" | tr -d '[]')
        echo "    details:"
        nbctl list DHCP_Options "$clean_uuid" 2>/dev/null | sed 's/^/      /'
    fi
}

list_all_dhcp_options() {
    echo "--- All DHCP_Options in NBDB ---"
    nbctl --format=table --columns=_uuid,cidr,external_ids list DHCP_Options 2>/dev/null || echo "(none)"
}

# --- main ---

echo "======================================================="
echo " Reproducer: DHCP options loss on primary UDN after"
echo " ovnkube-controller restart"
echo "======================================================="
echo ""

# 1. Find the VM pod and its node
echo "[1] Locating VM pod $NS/$POD ..."
NODE=$(kubectl get pod -n "$NS" "$POD" -o jsonpath='{.spec.nodeName}')
echo "    Node: $NODE"

VM_LABEL=$(kubectl get pod -n "$NS" "$POD" -o jsonpath='{.metadata.labels.vm\.kubevirt\.io/name}' 2>/dev/null || true)
echo "    VM label: ${VM_LABEL:-<not set>}"

# 2. Find the LSP name
echo ""
echo "[2] Finding LSP for pod $NS/$POD ..."
LSP_NAME=$(get_lsp_name "$NS" "$POD")
if [[ -z "$LSP_NAME" ]]; then
    echo "    ERROR: Could not find LSP. Listing all LSPs with namespace=$NS:"
    nbctl --format=table --columns=name,external_ids find Logical_Switch_Port external_ids:namespace="$NS"
    echo ""
    echo "    Please provide the LSP name manually and re-run, or check the pod name."
    exit 1
fi
echo "    LSP: $LSP_NAME"

# 3. Check DHCP options BEFORE restart
echo ""
echo "======================================================="
echo "[3] DHCP options BEFORE restart"
echo "======================================================="
get_dhcp_options_for_lsp "$LSP_NAME"
echo ""
list_all_dhcp_options

# 4. Find and restart the ovnkube-controller pod on the VM's node
echo ""
echo "======================================================="
echo "[4] Restarting ovnkube-controller on node $NODE ..."
echo "======================================================="

OVNKUBE_POD=$(kubectl get pods -n "$OVN_NS" -l app=ovnkube-node \
    --field-selector spec.nodeName="$NODE" \
    -o jsonpath='{.items[0].metadata.name}')
echo "    Deleting pod: $OVN_NS/$OVNKUBE_POD"
kubectl delete pod -n "$OVN_NS" "$OVNKUBE_POD"

echo "    Waiting for new ovnkube-node pod on $NODE to be Ready ..."
for i in $(seq 1 60); do
    NEW_POD=$(kubectl get pods -n "$OVN_NS" -l app=ovnkube-node \
        --field-selector spec.nodeName="$NODE" \
        -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
    if [[ -n "$NEW_POD" && "$NEW_POD" != "$OVNKUBE_POD" ]]; then
        READY=$(kubectl get pod -n "$OVN_NS" "$NEW_POD" \
            -o jsonpath='{.status.conditions[?(@.type=="Ready")].status}' 2>/dev/null || true)
        if [[ "$READY" == "True" ]]; then
            echo "    New pod $NEW_POD is Ready (took ${i}s)"
            break
        fi
    fi
    sleep 1
done

# Poll for DHCP options loss (the deletion happens ~420ms after UDN controller
# creates them, but the overall startup takes longer; we poll for up to 60s)
echo "    Polling for DHCP options loss (up to 60s) ..."
LOST=false
for i in $(seq 1 60); do
    V4=$(nbctl get Logical_Switch_Port "$LSP_NAME" dhcpv4_options 2>/dev/null || echo "[]")
    if [[ "$V4" == "[]" ]]; then
        echo "    DHCP options disappeared after ${i}s"
        LOST=true
        break
    fi
    if (( i % 10 == 0 )); then
        echo "    ... still present after ${i}s (dhcpv4_options=$V4)"
    fi
    sleep 1
done

# 5. Check DHCP options AFTER restart
echo ""
echo "======================================================="
echo "[5] DHCP options AFTER restart"
echo "======================================================="
get_dhcp_options_for_lsp "$LSP_NAME"
echo ""
list_all_dhcp_options

# 6. Verdict
echo ""
echo "======================================================="
echo "[6] Verdict"
echo "======================================================="

if [[ "$LOST" == "true" ]]; then
    echo ""
    echo "  BUG REPRODUCED: dhcpv4_options is empty on LSP $LSP_NAME"
    echo ""
    echo "  The DHCP options were deleted during ovnkube-controller restart."
    echo "  When the VM's DHCP lease expires, it will lose its IP address."
    echo ""
    exit 1
else
    echo ""
    echo "  DHCP options are still present on LSP $LSP_NAME."
    echo "  Bug was NOT reproduced (or has been fixed)."
    echo ""
    exit 0
fi
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened ownership checks to prevent one controller from deleting resources owned by another, avoiding accidental removal of VM-related network objects (including DHCP options and static routes) during syncs and upgrades.

* **Tests**
  * Added comprehensive tests for cross-controller protection, orphan/resource-zone handling, remote/legacy cases, and a regression test ensuring user-defined network DHCP options are preserved.

### DIFF
--- a/go-controller/pkg/kubevirt/external_ids.go
+++ b/go-controller/pkg/kubevirt/external_ids.go
@@ -6,6 +6,7 @@ import (
 	ktypes "k8s.io/apimachinery/pkg/types"
 
 	libovsdbops "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/libovsdb/ops"
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
 )
 
 func extractVMFromExternalIDs(externalIDs map[string]string) *ktypes.NamespacedName {
@@ -60,12 +61,11 @@ func ownsItAndIsOrphanOrWrongZone(externalIDs map[string]string, vms map[ktypes.
 		return false
 	}
 
-	const DefaultNetworkControllerName = "default-network-controller"
 	// Only consider resources owned by this controller. Resources created by
 	// other controllers (e.g. UDN controllers) must not be deleted here.
 	owner := externalIDs[string(libovsdbops.OwnerControllerKey)]
 	if owner == "" {
-		owner = DefaultNetworkControllerName
+		owner = types.DefaultNetworkControllerName
 	}
 	if owner != controllerName {
 		return false
--- a/go-controller/pkg/ovn/pods.go
+++ b/go-controller/pkg/ovn/pods.go
@@ -160,7 +160,7 @@ func (oc *DefaultNetworkController) syncPods(pods []interface{}) error {
 			}
 		}
 	}
-	if err := kubevirt.SyncVirtualMachines(oc.nbClient, vms, DefaultNetworkControllerName); err != nil {
+	if err := kubevirt.SyncVirtualMachines(oc.nbClient, vms, types.DefaultNetworkControllerName); err != nil {
 		return fmt.Errorf("failed syncing running virtual machines: %v", err)
 	}
 
--- a/go-controller/pkg/controllermanager/controller_manager.go
+++ b/go-controller/pkg/controllermanager/controller_manager.go
@@ -481,7 +481,7 @@ func (cm *ControllerManager) Start(ctx context.Context) error {
 
 	if config.OVNKubernetesFeature.EnableEgressIP {
 		cm.eIPController = ovn.NewEIPController(cm.nbClient, cm.kube, cm.watchFactory, cm.recorder, cm.portCache, cm.networkManager.Interface(),
-			addressset.NewOvnAddressSetFactory(cm.nbClient, config.IPv4Mode, config.IPv6Mode), config.IPv4Mode, config.IPv6Mode, zone, ovn.DefaultNetworkControllerName)
+			addressset.NewOvnAddressSetFactory(cm.nbClient, config.IPv4Mode, config.IPv6Mode), config.IPv4Mode, config.IPv6Mode, zone, ovntypes.DefaultNetworkControllerName)
 		// FIXME(martinkennelly): remove when EIP controller is fully extracted from from DNC and started here. Ensure SyncLocalNodeZonesCache is re-enabled in EIP controller.
 		if err = cm.eIPController.SyncLocalNodeZonesCache(); err != nil {
 			klog.Warningf("Failed to sync EgressIP controllers local node node cache: %v", err)
--- a/go-controller/pkg/ovn/admin_network_policy_test.go
+++ b/go-controller/pkg/ovn/admin_network_policy_test.go
@@ -80,7 +80,7 @@ func getDefaultPGForANPSubject(anpName string, portUUIDs []string, acls []*nbdb.
 	for _, uuid := range portUUIDs {
 		lsps = append(lsps, &nbdb.LogicalSwitchPort{UUID: uuid})
 	}
-	pgDbIDs := anpovn.GetANPPortGroupDbIDs(anpName, banp, DefaultNetworkControllerName)
+	pgDbIDs := anpovn.GetANPPortGroupDbIDs(anpName, banp, types.DefaultNetworkControllerName)
 
 	pg := libovsdbutil.BuildPortGroup(
 		pgDbIDs,
@@ -107,24 +107,24 @@ func getANPGressACL(action, anpName, direction string, rulePriority int32,
 		acl.Tier = types.DefaultBANPACLTier
 	}
 	acl.ExternalIDs = map[string]string{
-		libovsdbops.OwnerControllerKey.String():    DefaultNetworkControllerName,
+		libovsdbops.OwnerControllerKey.String():    types.DefaultNetworkControllerName,
 		libovsdbops.ObjectNameKey.String():         anpName,
 		libovsdbops.GressIdxKey.String():           fmt.Sprintf("%d", ruleIndex),
 		libovsdbops.PolicyDirectionKey.String():    direction,
 		libovsdbops.PortPolicyProtocolKey.String(): "None",
 		libovsdbops.OwnerTypeKey.String():          "AdminNetworkPolicy",
-		libovsdbops.PrimaryIDKey.String():          fmt.Sprintf("%s:AdminNetworkPolicy:%s:%s:%d:None", DefaultNetworkControllerName, anpName, direction, ruleIndex),
+		libovsdbops.PrimaryIDKey.String():          fmt.Sprintf("%s:AdminNetworkPolicy:%s:%s:%d:None", types.DefaultNetworkControllerName, anpName, direction, ruleIndex),
 	}
 	acl.Name = ptr.To(fmt.Sprintf("ANP:%s:%s:%d", anpName, direction, ruleIndex)) // tests logic for GetACLName
 	if banp {
 		acl.ExternalIDs[libovsdbops.OwnerTypeKey.String()] = "BaselineAdminNetworkPolicy"
 		acl.ExternalIDs[libovsdbops.PrimaryIDKey.String()] = fmt.Sprintf("%s:BaselineAdminNetworkPolicy:%s:%s:%d:None",
-			DefaultNetworkControllerName, anpName, direction, ruleIndex)
+			types.DefaultNetworkControllerName, anpName, direction, ruleIndex)
 		acl.Name = ptr.To(fmt.Sprintf("BANP:%s:%s:%d", anpName, direction, ruleIndex)) // tests logic for GetACLName
 	}
 	acl.UUID = fmt.Sprintf("%s_%s_%d-%f-UUID", anpName, direction, ruleIndex, rand.Float64())
 	// determine ACL match
-	pgName := libovsdbutil.GetPortGroupName(anpovn.GetANPPortGroupDbIDs(anpName, banp, DefaultNetworkControllerName))
+	pgName := libovsdbutil.GetPortGroupName(anpovn.GetANPPortGroupDbIDs(anpName, banp, types.DefaultNetworkControllerName))
 	var lPortMatch, l3Match, matchDirection, match string
 	if direction == string(libovsdbutil.ACLIngress) {
 		acl.Direction = nbdb.ACLDirectionToLport
@@ -139,7 +139,7 @@ func getANPGressACL(action, anpName, direction string, rulePriority int32,
 		matchDirection = "dst"
 	}
 	asIndex := anpovn.GetANPPeerAddrSetDbIDs(anpName, direction, fmt.Sprintf("%d", ruleIndex),
-		DefaultNetworkControllerName, banp)
+		types.DefaultNetworkControllerName, banp)
 	asv4, asv6 := addressset.GetHashNamesForAS(asIndex)
 	if config.IPv4Mode && config.IPv6Mode {
 		l3Match = fmt.Sprintf("((ip4.%s == $%s || ip6.%s == $%s))", matchDirection, asv4, matchDirection, asv6)
@@ -182,10 +182,10 @@ func getANPGressACL(action, anpName, direction string, rulePriority int32,
 		aclCopy.ExternalIDs[libovsdbops.PortPolicyProtocolKey.String()] = protocol
 		aclCopy.Match = match
 		aclCopy.ExternalIDs[libovsdbops.PrimaryIDKey.String()] = fmt.Sprintf("%s:AdminNetworkPolicy:%s:%s:%d:%s",
-			DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol)
+			types.DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol)
 		if banp {
 			aclCopy.ExternalIDs[libovsdbops.PrimaryIDKey.String()] = fmt.Sprintf("%s:BaselineAdminNetworkPolicy:%s:%s:%d:%s",
-				DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol)
+				types.DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol)
 		}
 		aclCopy.UUID = fmt.Sprintf("%s_%s_%d.%s-%f-UUID", anpName, direction, ruleIndex, protocol, rand.Float64())
 		retACLs = append(retACLs, &aclCopy)
@@ -205,10 +205,10 @@ func getANPGressACL(action, anpName, direction string, rulePriority int32,
 		aclCopy.ExternalIDs[libovsdbops.PortPolicyProtocolKey.String()] = protocol + "-namedPort"
 		aclCopy.Match = match
 		aclCopy.ExternalIDs[libovsdbops.PrimaryIDKey.String()] = fmt.Sprintf("%s:AdminNetworkPolicy:%s:%s:%d:%s",
-			DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol+"-namedPort")
+			types.DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol+"-namedPort")
 		if banp {
 			aclCopy.ExternalIDs[libovsdbops.PrimaryIDKey.String()] = fmt.Sprintf("%s:BaselineAdminNetworkPolicy:%s:%s:%d:%s",
-				DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol+"-namedPort")
+				types.DefaultNetworkControllerName, anpName, direction, ruleIndex, protocol+"-namedPort")
 		}
 		retACLs = append(retACLs, &aclCopy)
 	}
@@ -253,7 +253,7 @@ func getACLsForANPRules(anp *anpapi.AdminNetworkPolicy) []*nbdb.ACL {
 
 func buildANPAddressSets(anp *anpapi.AdminNetworkPolicy, index int32, ips []string, gressPrefix libovsdbutil.ACLDirection) (*nbdb.AddressSet, *nbdb.AddressSet) {
 	asIndex := anpovn.GetANPPeerAddrSetDbIDs(anp.Name, string(gressPrefix),
-		fmt.Sprintf("%d", index), DefaultNetworkControllerName, false)
+		fmt.Sprintf("%d", index), types.DefaultNetworkControllerName, false)
 	return addressset.GetTestDbAddrSets(asIndex, ips)
 }
 
--- a/go-controller/pkg/ovn/base_network_controller_user_defined.go
+++ b/go-controller/pkg/ovn/base_network_controller_user_defined.go
@@ -955,7 +955,7 @@ func (bsnc *BaseUserDefinedNetworkController) buildUDNEgressSNAT(localPodSubnets
 		// contains the node IPs in the cluster. Given that egressIP feature
 		// already has an address set containing these nodeIPs owned by the
 		// default network controller, let's re-use it.
-		nodeIPsASIDs := getEgressIPAddrSetDbIDs(NodeIPAddrSetName, types.DefaultNetworkName, DefaultNetworkControllerName)
+		nodeIPsASIDs := getEgressIPAddrSetDbIDs(NodeIPAddrSetName, types.DefaultNetworkName, types.DefaultNetworkControllerName)
 		nodeIPsAS, err = bsnc.addressSetFactory.GetAddressSet(nodeIPsASIDs)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get address set with IDs %v: %w", nodeIPsASIDs, err)
--- a/go-controller/pkg/ovn/baseline_admin_network_policy_test.go
+++ b/go-controller/pkg/ovn/baseline_admin_network_policy_test.go
@@ -25,6 +25,7 @@ import (
 	anpovn "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/controller/admin_network_policy"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/testing"
 	libovsdbtest "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/testing/libovsdb"
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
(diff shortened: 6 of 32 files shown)


go-controller/pkg/kubevirt/external_ids.go
```diff
@@ -49,13 +49,23 @@ func externalIDsContainsVM(externalIDs map[string]string, vm *ktypes.NamespacedN
...
 	if _, ok := externalIDs[OvnZoneExternalIDKey]; !ok {
 		return false
 	}
+	// Only consider resources owned by this controller. Resources created by
+	// other controllers (e.g. UDN controllers) must not be deleted here.
+	if owner, ok := externalIDs[string(libovsdbops.OwnerControllerKey)]; ok {
+		if owner != controllerName {
+			return false
+		}
+	}
```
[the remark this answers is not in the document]
qinqon: That's not an issue:
 - EnsureLocalZonePodAddressesToNodeRoute and EnsureRemoteZonePodAddressesToNodeRoute are only called from DefaultNetworkController (in go-controller/pkg/ovn/ovn.go:212,260). UDN controllers don't create these
  router resources.
  - Without OwnerControllerKey, the new check at line 64 falls through (the if owner, ok := ... branch is skipped), and the existing orphan/zone logic handles it correctly.
  - Adding OwnerControllerKey to these resources would be a nice improvement but is out of scope for this bug fix.

Maybe nice to have though.

qinqon, line 72: Nice — this correctly short-circuits before the orphan/zone logic when the resource belongs to a different controller. And the fallthrough when `OwnerControllerKey` is absent preserves backward compatibility with legacy resources created before this key existed.

One thing to consider: during an upgrade, could there be a window where a UDN controller has created a DHCP option *without* `OwnerControllerKey` (e.g. if the UDN controller code path doesn't set it yet)? If so, this check would fall through and the resource could still be deleted as an orphan. Worth verifying that all DHCP creation paths (including UDN) set `OwnerControllerKey` via `NewDbObjectIDs`.

go-controller/pkg/kubevirt/external_ids_test.go
```diff
@@ -0,0 +1,122 @@
...
+		// --- Legacy resources without OwnerControllerKey ---
+		Entry("returns true for orphan resource without OwnerControllerKey (legacy)",
+			map[string]string{
+				OvnZoneExternalIDKey:              OvnLocalZone,
+				string(libovsdbops.ObjectNameKey): vmKey.String(),
+			},
+			map[ktypes.NamespacedName]bool{}, // VM not found → orphan
+			defaultControllerName,
+			true,
+		),
```
qinqon, line 120: nit: Consider adding one more case for **legacy resources with a matching active VM**:
```go
Entry("returns false for legacy resource without OwnerControllerKey when VM is present and local",
    map[string]string{
        OvnZoneExternalIDKey:              OvnLocalZone,
        string(libovsdbops.ObjectNameKey): vmKey.String(),
    },
    map[ktypes.NamespacedName]bool{vmKey: true}, // VM is present and local
    defaultControllerName,
    false,
),
```
This would confirm the existing behavior is preserved for legacy resources that have active VMs (i.e. they should NOT be deleted).

go-controller/pkg/ovn/pods.go
```diff
@@ -161,7 +161,7 @@ func (oc *DefaultNetworkController) syncPods(pods []interface{}) error {
 			}
 		}
 	}
-	if err := kubevirt.SyncVirtualMachines(oc.nbClient, vms); err != nil {
+	if err := kubevirt.SyncVirtualMachines(oc.nbClient, vms, DefaultNetworkControllerName); err != nil {
```
qinqon: Looks correct. `SyncVirtualMachines` is only called from `DefaultNetworkController.syncPods` — UDN controllers use `syncPodsForUserDefinedNetwork` which doesn't call `SyncVirtualMachines` at all. So passing `DefaultNetworkControllerName` here is the right thing.

qinqon (commented):
## Review Summary

**LGTM** — The fix is correct, minimal, and well-tested.

### What the PR does

Adds a `controllerName` parameter to `ownsItAndIsOrphanOrWrongZone` and `SyncVirtualMachines` so that only resources owned by the calling controller are considered for deletion. This prevents the default network controller from deleting DHCP options owned by UDN controllers during startup.

### Key observations

1. **Correct root cause fix** — The problem was that `ownsItAndIsOrphanOrWrongZone` didn't distinguish between resources owned by different controllers, so the default network controller's sync would delete UDN-owned DHCP options as "orphans."

2. **Good backward compatibility** — The `OwnerControllerKey` check is opt-in: if the key is absent (legacy resources), the function falls through to existing logic. This handles upgrades gracefully.

3. **Solid test coverage** — The test table covers: cross-controller protection (both directions), orphan detection still works, wrong zone detection still works, legacy resources without `OwnerControllerKey`, and edge cases.

4. **Regarding CodeRabbit's concern about missing `OwnerControllerKey` on `LogicalRouterPolicy`/`LogicalRouterStaticRoute` in `router.go`** — This is **not a problem in practice**: `EnsureLocalZonePodAddressesToNodeRoute` and `EnsureRemoteZonePodAddressesToNodeRoute` are only called from `DefaultNetworkController` (`ovn.go:212,260`). UDN controllers don't create these router resources. Without `OwnerControllerKey`, the new check falls through and the existing orphan/zone logic handles it correctly. Adding `OwnerControllerKey` to these resources would be a nice future improvement but is out of scope for this bug fix.

go-controller/pkg/kubevirt/external_ids.go
maiqueb: That can't happen, since the code which creates DHCP options was introduced setting the owner controller key in commit https://github.com/ovn-kubernetes/ovn-kubernetes/commit/01cafb2cc33606a5972930400863a20a5cf05ea1, introduced in PR https://github.com/ovn-kubernetes/ovn-kubernetes/pull/3478, which predates UDN for about a year.

tssurya:
> /verified by CI

please don't add this to upstream PRs

jcaamano:
@maiqueb @qinqon is this not reflected in any e2e test?

maiqueb:
> @maiqueb @qinqon is this not reflected in any e2e test?

I do not have any knowledge of any sort of resiliency tests in this repo, but I'm probably ignorant. 

Point me to them, I'll try to expand on them.

jcaamano:
> > @maiqueb @qinqon is this not reflected in any e2e test?
> 
> I do not have any knowledge of any sort of resiliency tests in this repo, but I'm probably ignorant.
> 
> Point me to them, I'll try to expand on them.

Not sure if there are any or we should consider adding. 

If not e2e, maybe some component test around here
https://github.com/jcaamano/ovn-kubernetes/blob/9f1ec7e5fb8c950c8c58ff0299c9b4ed9956a3e0/go-controller/pkg/ovn/layer2_user_defined_network_controller_test.go#L382

I guess we might not have that type of coverage for virt?

maiqueb:
> > > @maiqueb @qinqon is this not reflected in any e2e test?
> > 
> > 
> > I do not have any knowledge of any sort of resiliency tests in this repo, but I'm probably ignorant.
> > Point me to them, I'll try to expand on them.
> 
> Not sure if there are any or we should consider adding.
> 
> If not e2e, maybe some component test around here https://github.com/jcaamano/ovn-kubernetes/blob/9f1ec7e5fb8c950c8c58ff0299c9b4ed9956a3e0/go-controller/pkg/ovn/layer2_user_defined_network_controller_test.go#L382
> 
> I guess we might not have that type of coverage for virt?

Added an integration test that proves the UDN DHCP Options are not removed when the default network controller starts from scratch. 

I've confirmed the test fails without this fix.
@jcaamano

go-controller/pkg/kubevirt/external_ids.go
jcaamano, line 72: I woujld add a very explicit comment here on why are we matching when there is no `OwnerControllerKey`.

Is the test in `go-controller/pkg/kubevirt/external_ids_test.go` the only coverage we have for this specific scenario? Curious.

maiqueb: > I woujld add a very explicit comment here on why are we matching when there is no `OwnerControllerKey`.

I don't think it's necessary, but I'll comply. 

Can you suggest a comment ? Given the "very explicit" adjective, it's clear you have something in mind.

If needed, I'd go with
```
// If OwnerControllerKey is absent we fall through: since DHCP options for
// virt workloads always set this key, its absence guarantees the resource
// was not created by a UDN controller, so there is no cross-controller
// ownership conflict and it is safe to treat it as a deletion candidate.
```

> 
> Is the test in `go-controller/pkg/kubevirt/external_ids_test.go` the only coverage we have for this specific scenario? Curious.

It's a new test, whose goal is to address this scenario, not provide coverage to external id generation.

maiqueb: @jcaamano ^

jcaamano: > Can you suggest a comment ? Given the "very explicit" adjective, it's clear you have something in mind.
> 
> If needed, I'd go with
> 
> ```
> // If OwnerControllerKey is absent we fall through: since DHCP options for
> // virt workloads always set this key, its absence guarantees the resource
> // was not created by a UDN controller, so there is no cross-controller
> // ownership conflict and it is safe to treat it as a deletion candidate.
> ```

I am not exactly sure why we do that, so that is why I am asking for a comment. 

Your comments mention "DHCP options" but this method is not specific to "DHCP options", it is also called for other resources. Your comments also mention "deletion" but that is not really intrinsic to this method either, could be used in other scenarios in the future.

This is just a predicate and I guess the resources need to be owned by `controllerName` to match, right? As simple as that.

And if `OwnerControllerKey` is not set, we are making an assumption of what controller owns those resources? What is that assumption? Are we assuming the default network controller owns those resources? What would break if you did the opposite and not match anything that did not have `OwnerControllerKey` set?

> It's a new test, whose goal is to address this scenario, not provide coverage to external id generation.

I didn't mean external id generation. I was looking at the test you added to find answers to my questions above. But the test does not have any resources without `OwnerControllerKey` so I did not find answer there.

maiqueb: > > Can you suggest a comment ? Given the "very explicit" adjective, it's clear you have something in mind.
> > If needed, I'd go with
> > ```
> > // If OwnerControllerKey is absent we fall through: since DHCP options for
> > // virt workloads always set this key, its absence guarantees the resource
> > // was not created by a UDN controller, so there is no cross-controller
> > // ownership conflict and it is safe to treat it as a deletion candidate.
> > ```
> 
> I am not exactly sure why we do that, so that is why I am asking for a comment.
> 
> Your comments mention "DHCP options" but this method is not specific to "DHCP options", it is also called for other resources. Your comments also mention "deletion" but that is not really intrinsic to this method either, could be used in other scenarios in the future.

OK, I'll remove it.

> 
> This is just a predicate and I guess the resources need to be owned by `controllerName` to match, right? As simple as that.

Then I don't understand why the current explanation is not enough:
> Only consider resources owned by this controller. Resources created by other controllers (e.g. UDN controllers) must not be deleted here.

I'll add something else saying that if we can't establish controller ownership, we fall through, which means treating the resource a candidate for deletion.

> 
> And if `OwnerControllerKey` is not set, we are making an assumption of what controller owns those resources? What is that assumption? Are we assuming the default network controller owns those resources? What would break if you did the opposite and not match anything that did not have `OwnerControllerKey` set?
> 
> > It's a new test, whose goal is to address this scenario, not provide coverage to external id generation.
> 
> I didn't mean external id generation. I was looking at the test you added to find answers to my questions above. But the test does not have any resources without `OwnerControllerKey` so I did not find answer there.

maiqueb: > > > Can you suggest a comment ? Given the "very explicit" adjective, it's clear you have something in mind.
> > > If needed, I'd go with
> > > ```
> > > // If OwnerControllerKey is absent we fall through: since DHCP options for
> > > // virt workloads always set this key, its absence guarantees the resource
> > > // was not created by a UDN controller, so there is no cross-controller
> > > // ownership conflict and it is safe to treat it as a deletion candidate.
> > > ```
> > 
> > 
> > I am not exactly sure why we do that, so that is why I am asking for a comment.
> > Your comments mention "DHCP options" but this method is not specific to "DHCP options", it is also called for other resources. Your comments also mention "deletion" but that is not really intrinsic to this method either, could be used in other scenarios in the future.
> 
> OK, I'll remove it.
> 
> > This is just a predicate and I guess the resources need to be owned by `controllerName` to match, right? As simple as that.
> 
> Then I don't understand why the current explanation is not enough:
> 
> > Only consider resources owned by this controller. Resources created by other controllers (e.g. UDN controllers) must not be deleted here.
> 
> I'll add something else saying that if we can't establish controller ownership, we fall through, which means treating the resource a candidate for deletion.
> 
> > And if `OwnerControllerKey` is not set, we are making an assumption of what controller owns those resources? What is that assumption? Are we assuming the default network controller owns those resources? What would break if you did the opposite and not match anything that did not have `OwnerControllerKey` set?

What we're doing, is keeping the existent behavior, which, if you ask me, seems to be assuming the default network controller owns those resources.

> > > It's a new test, whose goal is to address this scenario, not provide coverage to external id generation.
> > 
> > 
> > I didn't mean external id generation. I was looking at the test you added to find answers to my questions above. But the test does not have any resources without `OwnerControllerKey` so I did not find answer there.

jcaamano: > Then I don't understand why the current explanation is not enough:
> 
> > Only consider resources owned by this controller. Resources created by other controllers (e.g. UDN controllers) must not be deleted here.

Because if `OwnerControllerKey` is not set, you don't know if the resources are owned by that controller. And you are still deleting in that case.

What would happen if you changed this
```
	if owner, ok := externalIDs[string(libovsdbops.OwnerControllerKey)]; ok {
		if owner != controllerName {
			return false
		}
	}
```
to this
```
if externalIDs[string(libovsdbops.OwnerControllerKey)]  != controllerName {
    return falsde
}
```
or to this
```
owner := externalIDs[string(libovsdbops.OwnerControllerKey)]
if owner == "" {
    owner = DefaultNetworkController
}
if owner != controllerName {
    return false
}
```

jcaamano: @qinqon please chime in if you can help clear out this misunderstanding.

maiqueb: > Because if OwnerControllerKey is not set, you don't know if the resources are owned by that controller. 

I get that. That's why I propose to keep the existing behavior on those situations, thus limiting the risk we are adding to the production code. This affects a lot of resources, and I want to keep the impacts as low as I can. 

I think the following alternative you're suggesting:
```
if externalIDs[string(libovsdbops.OwnerControllerKey)]  != controllerName {
    return falsde
}
```

would warrant the same results, but is **less** explicit.

The other alternative:
```
owner := externalIDs[string(libovsdbops.OwnerControllerKey)]
if owner == "" {
    owner = DefaultNetworkController
}
if owner != controllerName {
    return false
}
```

Would warrant the same result, and is more explicit. 

Personally, I'm OK with what I am proposing, but if I had to choose one of the 2 alternatives, I'd go with the most explicit one, where you explicitly define the current resource owner, defaulting to the `DefaultNetworkController` .

jcaamano: My vote is to go with most explicit proposal then.

maiqueb: I'll get rid of the comment then, since IMHO it becomes awfully redundant.

maiqueb: Done.

jcaamano:
@qinqon you're good with this latest change?

go-controller/pkg/kubevirt/external_ids.go
```diff
@@ -49,13 +49,27 @@ func externalIDsContainsVM(externalIDs map[string]string, vm *ktypes.NamespacedN
...
+// specified controller are considered for deletion, preventing cross-controller
+// interference (e.g. default network controller deleting UDN DHCP options).
+func ownsItAndIsOrphanOrWrongZone(externalIDs map[string]string, vms map[ktypes.NamespacedName]bool, controllerName string) bool {
 	// FIXME: VM IDs have no DB IDs and therefore may clash with other LRPs that do contain DB IBs. They will always have ObjectNameKey
 	// set therefore we now depend on the following key to be present. Remove this when DB IDs are implemented.
 	if _, ok := externalIDs[OvnZoneExternalIDKey]; !ok {
 		return false
 	}
+
+	const DefaultNetworkControllerName = "default-network-controller"
```
jcaamano: Why not use the exported constant:
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/655e28f351b38e842ea42f85fd793bac86946f66/go-controller/pkg/ovn/default_network_controller.go#L50

maiqueb: to prevent a cyclic dependency. I'm just repeating the pattern I've seen in [egress firewall](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/03a957199ff8df7f3f6714c71ab989b7250f148f/go-controller/pkg/ovn/controller/egressfirewall/egressfirewall.go#L60).

If you want me to move this to a common place, let me know where.

qinqon:
> @qinqon you're good with this latest change?

Yes all good

go-controller/pkg/kubevirt/external_ids.go
jcaamano: > If you want me to move this to a common place, let me know where.

`types/const.go` would be an appropriate place.

maiqueb: Done.

maiqueb:
@jcaamano heads up. All tests passed, @qinqon 's on board, and I've extended DRY principles across the project as you've requested in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6063#discussion_r2919937510 . 

Please take another look.

PR #6065, opened 2026-03-05
EVPN: Ensure VNIs are unique per vtep

kyrtapz:
## Summary by CodeRabbit

* **New Features**
  * VNI reservation to prevent VNI conflicts per VTEP; reservations are recovered on startup and released during network cleanup.
  * CRD validation added: macVRF and ipVRF must use different VNIs.

* **Tests**
  * Expanded EVPN test coverage for VNI conflicts, startup recovery, deterministic allocation, VTEP changes, and the equal‑VNI invalid case; tests modernized to use a unified EVPN config input.

### DIFF
--- a/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
+++ b/go-controller/pkg/clustermanager/userdefinednetwork/controller.go
@@ -83,6 +83,12 @@ func ipVRFKey(networkName string) string {
 	return networkName + "/ipvrf"
 }
 
+// vniKey identifies a VNI within a VTEP scope. VNIs must be unique per VTEP.
+type vniKey struct {
+	vtep string
+	vni  int32
+}
+
 type RenderNetAttachDefManifest func(obj client.Object, targetNamespace string, opts ...template.RenderOption) (*netv1.NetworkAttachmentDefinition, error)
 
 type networkInUseError struct {
@@ -126,6 +132,11 @@ type Controller struct {
 	// vidAllocator allocates cluster-wide VLAN IDs for EVPN networks.
 	// VIDs are allocated per network name and stored in the NAD config JSON.
 	vidAllocator id.Allocator
+
+	// reservedVNIs tracks (VTEP, VNI) → network name to ensure no two networks
+	// sharing the same VTEP use the same VNI.
+	reservedVNIs     map[vniKey]string
+	reservedVNIsLock sync.RWMutex
 
 	udnClient         userdefinednetworkclientset.Interface
 	udnLister         userdefinednetworklister.UserDefinedNetworkLister
@@ -179,6 +190,7 @@ func New(
 		networkManager:    networkManager,
 		namespaceTracker:  map[string]sets.Set[string]{},
 		vidAllocator:      vidAllocator,
+		reservedVNIs:      map[vniKey]string{},
 		eventRecorder:     eventRecorder,
 	}
 	udnCfg := &controller.ControllerConfig[userdefinednetworkv1.UserDefinedNetwork]{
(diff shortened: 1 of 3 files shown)


kyrtapz:
I should probably rename the functions since they are stating only VID right now.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
```diff
@@ -394,6 +406,10 @@ func (c *Controller) recoverEVPNVIDsForCUDN(cudnName string) error {
 		return fmt.Errorf("failed to reserve VIDs for cudn %s: %w", cudnName, err)
 	}
 
+	if err := c.reserveVNIs(cudnName, netInfo.EVPNVTEPName(), netInfo.EVPNMACVRFVNI(), netInfo.EVPNIPVRFVNI()); err != nil {
+		return fmt.Errorf("failed to reserve VNIs for cudn %s: %w", cudnName, err)
+	}
```
jcaamano: I would update the method names (and doc) of the change up to this call, since they explicitly mention `VIDs`. Otherwise it is a bit hidden that we are handling VNIs as well.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
```diff
@@ -196,6 +200,10 @@ func (c *Controller) allocateEVPNVIDsIfNeeded(obj client.Object) ([]template.Ren
 		klog.V(4).InfoS("Allocated VID for IP-VRF", "network", networkName, "vid", vid)
 	}
 
+	if err := c.reserveVNIs(networkName, evpnCfg.VTEP, macVRFVNI, ipVRFVNI); err != nil {
+		return nil, fmt.Errorf("failed to reserve VNIs: %w", err)
```
jcaamano: So if we have two colliding networks, and we restart, what we consider to be the valid network can flap between those two. I think the only sane way to handle this is to consider all colliding networks invalid.

go-controller/pkg/clustermanager/userdefinednetwork/controller.go
kyrtapz: Agreed, done.

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
kyrtapz: We order the CUDNs by creation timestamp on startup recovery so the oldest one wins. This should work fine because CUDNs are not mutable today.

kyrtapz:
Last [push](https://github.com/ovn-kubernetes/ovn-kubernetes/compare/a91dc5677cf6151e29c8de0f7ce9f04255880e03..d048526c0e7ad52f604f00163ccf5d31ef8f3990) fixed the naming and re-ordered the VNI reservation to come first.

kyrtapz:
503 error from koji
/retest-failed

go-controller/pkg/clustermanager/userdefinednetwork/controller_helper.go
jcaamano: I am conflicted because this hinders our capacity of making things mutable in the future.

kyrtapz: I see you concern and I thought about it a bit more.
I don't see any better alternative to the current approach right now. I don't think invalidating old networks because an admin created a CUDN with a conflicting VNI is the right approach.
We would need a webhook to make sure it is not possible to create networks with conflicting vni/vid.
While maybe that is possible with the nad webhook I think it is beyond the scope of this PR and should be considered separately.

jcaamano: I agree that the webhook would be a great way to prevent that from happening but that doesn't mean we should depend on that to make this controller behave in the most reasonable way, that should be independent.

If you observe a system state where two things that can't overlap do overlap, and you don't have a sane way to tell which one came first, then the obvious thing to do is set them both in error before the consequences of not doing so just makes things worse.

I see the same approach suggested here
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6104/changes#diff-1e9b068f81697f43f7b344012d1ce60e07befe8e3d19c471c9e24b029e628763R339-R341

And we had similar considerations on how to handle overlaps in RAs.

Maybe we can have an overall discussion on how to best handle overlaps in the upstream meeting for all these cases.

kyrtapz: > If you observe a system state where two things that can't overlap do overlap, and you don't have a sane way to tell which one came first, then the obvious thing to do is set them both in error before the consequences of not doing so just makes things worse.

But we do have a way to determine which came first today. I consider invalidating existing networks way worse UX wise than accepting than we rely on an existing API immutability that we maybe will consider changing in the future.

jcaamano: Do you see a viable future where CUDN is mutable and we rely on the VNI set on the NAD as state to determine which came first?

kyrtapz: For sure, I would feel more comfortable if we additionally put guardrails around who is allowed to modify NADs generated from UDNs/CUDNs but that is an issue we should eventually address even without this aspect being considered.

kyrtapz:
re-based

jcaamano (approved):

PR #6067, opened 2026-03-05
fix: add short names for (c)udn

dave-tucker:
## 📑 Description

typing clusteruserdefinednetwork is getting old fast :)
seems that cudn and udn would be much nicer for all users.

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Added shorthand aliases for two Custom Resource Definitions: "cudn" and "udn", enabling faster kubectl commands and shorter configuration references.

### DIFF
--- a/dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
+++ b/dist/templates/k8s.ovn.org_clusteruserdefinednetworks.yaml.j2
@@ -11,6 +11,8 @@ spec:
     kind: ClusterUserDefinedNetwork
     listKind: ClusterUserDefinedNetworkList
     plural: clusteruserdefinednetworks
+    shortNames:
+    - cudn
     singular: clusteruserdefinednetwork
   scope: Cluster
   versions:
--- a/dist/templates/k8s.ovn.org_userdefinednetworks.yaml.j2
+++ b/dist/templates/k8s.ovn.org_userdefinednetworks.yaml.j2
@@ -11,6 +11,8 @@ spec:
     kind: UserDefinedNetwork
     listKind: UserDefinedNetworkList
     plural: userdefinednetworks
+    shortNames:
+    - udn
     singular: userdefinednetwork
   scope: Namespaced
   versions:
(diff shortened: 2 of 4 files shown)


dist/templates/k8s.ovn.org_egressfirewalls.yaml.j2
```diff
@@ -45,10 +45,6 @@ spec:
             type: string
           metadata:
             type: object
-            properties:
```
tssurya, line 48: what's with this change and the one on qos?

tssurya (approved):

trozet:
maybe updating the helm chart crds will trigger full ci to run ;)

dave-tucker:
@trozet helm chart CRDs are all symlinks:
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/crds/k8s.ovn.org_userdefinednetworks.yaml
- https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/crds/k8s.ovn.org_clusteruserdefinednetworks.yaml

dist/templates/k8s.ovn.org_egressfirewalls.yaml.j2
dave-tucker: I ran the generate script on macOS... and, after hacking it to work (bsd sed vs gnu sed) it made changes to these files. after I noticed it I backed out just those changes 😓

tssurya:
I honestly don't know why CI isn't running for this PR. I'll do a rebase to kick it.

tssurya:
@dave-tucker we'll need to update the YAMLs in CNO for us to consume this in on our side.

trozet:
> @trozet helm chart CRDs are all symlinks:
> 
> * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/crds/k8s.ovn.org_userdefinednetworks.yaml
> * https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/crds/k8s.ovn.org_clusteruserdefinednetworks.yaml

@dave-tucker wow that is bad :) We are symlinking to j2 and getting lucky there is actually no j2 rendering in there! I'll fix this as part of our move to helm from kind.sh.

PR #5874, opened 2026-01-25
Add EVPN E2E utils + Add the first EVPN E2E test for north-south/south-north coverage

tssurya:
## 📑 Description

This PR adds 3 tests to test north-south case for EVPN originating from the pod to the outside:

1. Layer2 CUDN, MAC VRF test
2. Layer2 CUDN, MAC+IPVRF test
3. Layer3 CUDN, IPVRF test

In each case the existing BGP FRR setup with new agnhost ipvrf and macvrf external servers are used.

## Additional Information for reviewers

- ~Currently these tests are running only in BGP+LGW+Dualstack lane~ Added a new lane for EVPN as discussed with others since even though we improved the BGP dualstack lane time, adding more tests will just bump that lane back to a higher total runtime value
- Tests are skipped on LGW, ipv6 lane because of known limitation in FRR not supporting ipv6 underlay tunnels: https://github.com/FRRouting/frr/issues/5885 and https://github.com/FRRouting/frr/issues/5885#issuecomment-3797137370; sample run: https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/21331694570/job/61397685546?pr=5874
- Given the EVPN implementation is not completed, see the "REVERT ME" commit where I need to leverage bash script to do the setup on the cluster-nodes side: uses https://github.com/tssurya/evpn-scripts-on-ovnk-kind-with-existing-bgp-setup/blob/main/evpn-cluster-setup.sh
- There is a current problem where the 172.26 subnet of bgpserver gets disturbed since frr-k8s doesn't allow incremental updates and we update and recreate things? Since EVPN config gtes amended on the global level that will cause issues even if I use targetVRF in my RA - this will cause problems for BGP tests in BGP lane which we don't want.
- ~TBD: Based on reviews see if it makes sense to not do default vrf and just do vrfname like vrflite tests do~ It's not allowed for EVPN transport to be set on CUDN and for RA to not have targetVRF when it selects that network, this was done through https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5807 and I've updated the tests to use the same and set vrfName (to ensure its less than 15 chars, the network names are created to be super small to have predictable VRF names so that the revertable bash script does the cluster-side setup correctly expecting the networkname as the vrfname)
- Manual scripts to test e2e on KIND are here for devs to play around with:
    - https://github.com/tssurya/evpn-scripts-on-ovnk-kind-with-existing-bgp-setup/blob/main/instructions-l2-ip-mac-vrf.sh
[...]

PR #6068, opened 2026-03-05
Adapt the current VRF-Lite test cases to also provide coverage for EVPN

jcaamano:
This PR picks up the work done by @tssurya in https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5874 reusing the utilities provided there and adapts the current VRF-Lite test cases to also cover EVPN networks.

This entailed:

* juggling a bit with the form (arguments passed in, etc) of both the VRF-Lite and EVPN helper utilities to align them for reuse from the same tests, with no real functional changes
* reduce the overall runtime of the tests by: collapsing isolation tests into one, verifying that target isolated networks are functional allowing to reduce the isolation timeouts
* Adding EVPN networks to the matrix
* assorted fixes here and there

Existing VRF-Lite tests also benefit from the reduced runtime and could now also be run in parallel. Due to this, the EXTENDED label to filter out test cases from CI is removed.

The three original EVPN test cases have been removed as that coverage is already included in the VRF-Lite test cases.

Overall the coverage includes connectivity checks for the three different types of configurations [L2 MACVRF, L2 MACVRF+IPVRF, L3 IP-VRF]:
* North/South both internal and external to the cluster when appropriate
* East/West both internal and external to the cluster when appropriate
* KAPI
* ClusterIP service
* Isolation across networks
* Isolation from nodes

## Summary by CodeRabbit

* **Tests**
  * Refactored EVPN and BGP test flows for centralized, per‑IP‑family orchestration and consistent external server handling
  * Added IP‑family aware subnet/IP filtering and network extraction helpers to improve dual‑stack reliability
  * Introduced stronger test isolation via randomized subnet generation
  * Enabled extended RouteAdvertisements tests under combined network conditions (removed prior skip)
  * Removed a legacy "Extended" test label provider

### DIFF
--- a/test/e2e/evpn.go
+++ b/test/e2e/evpn.go
@@ -1222,8 +1222,9 @@ func runEVPNNetworkAndServers(
 		return err
 	}
 
+	var macVRFVID int
 	if hasMACVRF {
-		macVRFVID := randomVID()
+		macVRFVID = randomVID()
 		framework.Logf("Generated random VIDs for external FRR: MAC-VRF VID=%d", macVRFVID)
 		framework.Logf("Setting up MAC-VRF on external FRR")
 		err = setupMACVRFOnExternalFRR(int(networkSpec.EVPN.MACVRF.VNI), macVRFVID, bridgeName, vxlanName)
@@ -1248,6 +1249,9 @@ func runEVPNNetworkAndServers(
 		// Derive VRF name from VNI (unique per IP-VRF)
 		ipVRFName := fmt.Sprintf("vrf%d", networkSpec.EVPN.IPVRF.VNI)
 		ipVRFVID := randomVID()
+		for macVRFVID == ipVRFVID {
+			ipVRFVID = randomVID()
+		}
 		framework.Logf("Generated random VIDs for external FRR: IP-VRF VID=%d", ipVRFVID)
 		framework.Logf("Setting up IP-VRF on external FRR")
 		err = setupIPVRFOnExternalFRR(ictx, ipVRFName, int(networkSpec.EVPN.IPVRF.VNI), ipVRFVID, bridgeName, vxlanName)
--- a/test/e2e/route_advertisements.go
+++ b/test/e2e/route_advertisements.go
@@ -1998,12 +1998,17 @@ var _ = ginkgo.Describe("BGP: For BGP configured networks", feature.RouteAdverti
 			// REVERT ME: Temporary cluster-side EVPN setup until OVN-K implements it natively
 			ginkgo.By("Running cluster-side EVPN setup script (REVERT ME)")
 			cudnSubnets := getNetworkSubnetsFromSpec(networkSpec)
-			var macVRFVNI, ipVRFVNI int
+			var macVRFVNI, ipVRFVNI, macVRFVID, ipVRFVID int
 			if networkSpec.EVPN.MACVRF != nil {
 				macVRFVNI = int(networkSpec.EVPN.MACVRF.VNI)
+				macVRFVID = randomVID()
 			}
 			if networkSpec.EVPN.IPVRF != nil {
 				ipVRFVNI = int(networkSpec.EVPN.IPVRF.VNI)
+				ipVRFVID = randomVID()
+				for macVRFVID == ipVRFVID {
+					ipVRFVID = randomVID()
+				}
 			}
 			err = runClusterEVPNSetupScript(
 				ictx,
@@ -2011,9 +2016,9 @@ var _ = ginkgo.Describe("BGP: For BGP configured networks", feature.RouteAdverti
 				networkName,
 				bgpASN,
 				macVRFVNI,
-				randomVID(),
+				macVRFVID,
 				ipVRFVNI,
-				randomVID(),
+				ipVRFVID,
 				cudnSubnets,
 			)
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())


test/e2e/route_advertisements.go
```diff
@@ -2579,7 +2743,7 @@ type templateInputFRR struct {
 var ratestdata embed.FS
 var tmplDir = filepath.Join("testdata", "routeadvertisements")
 
-const frrImage = "quay.io/frrouting/frr:10.4.1"
+const frrImage = "quay.io/frrouting/frr:10.4.2"
```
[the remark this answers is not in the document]
jcaamano: I am moving from FRR 10.4.1 not from FRR 9. It is also a bugfix release bump with useful fixes for something that is not the SUT.

test/e2e/route_advertisements.go
```diff
@@ -2142,31 +2275,34 @@ var _ = ginkgo.Describe("BGP: For a VRF-Lite configured network", feature.RouteA
...
-						gomega.Expect(err).NotTo(gomega.HaveOccurred())
-						gomega.Expect(testPodIP).ToNot(gomega.BeEmpty())
-						testPodToClientIPAndExpect(testPod, serverIP, testPodIP)
+						ginkgo.By("Ensuring a request from the pod can reach the external servers without being SNATed")
+						for _, externalServer := range externalServers {
+							bgpServerNetwork, err := infraprovider.Get().GetNetwork(externalServer)
+							gomega.Expect(err).NotTo(gomega.HaveOccurred())
+							iface, err := infraprovider.Get().GetK8NodeNetworkInterface(externalServer, bgpServerNetwork)
+							gomega.Expect(err).NotTo(gomega.HaveOccurred())
+							serverIP := getFirstIPStringOfFamily(family, []string{iface.IPv4, iface.IPv6})
```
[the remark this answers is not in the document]
pperiyasamy: agree, it must use `GetExternalContainerNetworkInterface` API , otherwise this would fail with other infraprovider like ocp. so `externalServers` must be `[]api.ExternalContainer` type

test/e2e/route_advertisements.go
```diff
@@ -1940,143 +2084,144 @@ var _ = ginkgo.Describe("BGP: For a VRF-Lite configured network", feature.RouteA
 			// different but ExecK8NodeCommand is more convinient
 			ip, err := infraprovider.Get().ExecK8NodeCommand(
```
pperiyasamy: it needs to be replaced with `ExecExternalContainerCommand`.

test/e2e/evpn.go
pperiyasamy, line 339: can you replace it with `return []string{"sh", "-c", "\"vtysh " + strings.Join(parts, " ") + "\""}` ? 
This would also work when it runs via ssh command.

test/e2e/evpn.go
```diff
@@ -1109,330 +1174,119 @@ func copyEnvMap(m map[string]string) map[string]string {
...
-			// Filter networkSpec subnets based on cluster IP family support
-			if networkSpec.Layer2 != nil {
-				networkSpec.Layer2.Subnets = filterDualStackCIDRs(f.ClientSet, networkSpec.Layer2.Subnets)
-			}
-			if networkSpec.Layer3 != nil {
-				networkSpec.Layer3.Subnets = filterL3Subnets(f.ClientSet, networkSpec.Layer3.Subnets)
-			}
+	testVTEPName := testName + "-vtep"
+	framework.Logf("Creating VTEP CR")
+	err = createVTEP(f, ictx, testVTEPName, vtepSubnets, vtepv1.VTEPModeManaged)
```
pperiyasamy, line 1280: should VTEP be in `vtepv1.VTEPModeUnmanaged` mode having `vtepSubnets`  set with ocp node's primary network subnets ?

test/e2e/route_advertisements.go
pperiyasamy, line 2453: there is already a pod created with same name in the step.
`STEP: Running a pod on the tested network namespace`
so it must be created with different name here.

pperiyasamy, line 2449: creating pod like this doesn't work with d/s cluster due to pod security admission issues. 
so can you consider creating it something like below ?
```
			httpServerPodConfig := func(podName, namespace string, opts ...podOption) podConfiguration {
				var podOptions []podOption
				podOptions = append(podOptions, withCommand(func() []string {
					return httpServerContainerCmd(8080)
				}))
				for _, opt := range opts {
					podOptions = append(podOptions, opt)
				}
				cfg := *podConfig(podName, podOptions...)
				cfg.namespace = namespace
				return cfg
			}

							otherPodConfig := httpServerPodConfig(testNamespace.Name+"other-netexec-pod", testNamespace.Name,
								withLabels(map[string]string{"app": "other-netexec-pod"}))
							otherPod = runUDNPod(f.ClientSet, testNamespace.Name, otherPodConfig, nil)
```

test/e2e/evpn.go
jcaamano: This needs to be somehow solved in our downstream implementation. There is no point and not maintainable to do things in a specific way tailored to our downstream that nobody else is going to understand.

test/e2e/route_advertisements.go
jcaamano: This needs to be somehow solved in our downstream implementation. There is no point and not maintainable to do things in a specific way tailored to our downstream that nobody else is going to understand.

test/e2e/evpn.go
jcaamano: `vtepSubnets` is not set with ocp node's primary network subnets, is it?

I guess right now the VTEP is being ignored and just created to keep the API happy. 

I guess whatever needs to change here is going to be changed in patryk's PR?

test/e2e/route_advertisements.go
jcaamano: This is not the pod `name` field, is the `generateName` field, k8s will generate a name with a random suffix.

test/e2e/evpn.go
pperiyasamy: right, this change is not needed now and would be changed  in patryk's PR.

tssurya: yes @kyrtapz 's changing it to unmanaged in his PR , for this PR its fine to leave things as is to satisfy the API server, we are not doing anything with VTEPs really

test/e2e/route_advertisements.go
jcaamano: And should be already handled with `newPrivelegedTestFramework` as discussed, right?

test/e2e/route_advertisements.go
```diff
@@ -1882,26 +1882,173 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
 	),
 )
 
-var _ = ginkgo.Describe("BGP: For a VRF-Lite configured network", feature.RouteAdvertisements, func() {
+var _ = ginkgo.Describe("BGP: For BGP configured networks", feature.RouteAdvertisements, func() {
```
pperiyasamy, line 1885: Nit: can you also add additional feature label here, it may help to filter out BGP test cases.

test/e2e/route_advertisements.go
```diff
@@ -1882,26 +1882,173 @@ var _ = ginkgo.DescribeTableSubtree("BGP: isolation between advertised networks"
...
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			network, err := infraprovider.Get().GetNetwork(networkName)
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			for _, node := range nodeList.Items {
+				iface, err := infraprovider.Get().GetK8NodeNetworkInterface(node.Name, network)
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				// prevent the IPv6 address of the interface from being removed when attaching to VRF
+				_, err = infraprovider.Get().ExecK8NodeCommand(node.Name, []string{"sysctl", "-w", "net.ipv6.conf." + iface.InfName + ".keep_addr_on_down=1"})
+				gomega.Expect(err).NotTo(gomega.HaveOccurred())
+				_, err = infraprovider.Get().ExecK8NodeCommand(node.Name, []string{"ip", "link", "set", "dev", iface.InfName, "master", networkName})
```
pperiyasamy: shouldn't we do this only if `ipFamilySet` has IPv6 ?

test/e2e/route_advertisements.go
pperiyasamy: it seems adding `f.NamespacePodSecurityLevel = admissionapi.LevelPrivileged` into `newPrivelegedTestFramework` doesn't fix the issue, anyway let handle it with separate PR if needed.

pperiyasamy, line 2444: `getNode` arg is never used for pod placement.

test/e2e/route_advertisements.go
```diff
@@ -2358,122 +2516,139 @@ var _ = ginkgo.Describe("BGP: For a VRF-Lite configured network", feature.RouteA
...
+						ginkgo.Entry("Default", defaultNetwork, nilNetworkSpecGen),
+						ginkgo.Entry("Layer 3 CUDN VRF-Lite", cudnAdvertisedVRFLite, layer3NetworkSpecGen),
+						ginkgo.Entry("Layer 2 CUDN VRF-Lite", cudnAdvertisedVRFLite, layer2NetworkSpecGen),
+						ginkgo.Entry("Layer 3 UDN", udn, layer3NetworkSpecGen),
+						ginkgo.Entry("Layer 3 CUDN advertised", cudnAdvertised, layer3NetworkSpecGen),
+						ginkgo.Entry("Layer 2 UDN", udn, layer2NetworkSpecGen),
+						ginkgo.Entry("Layer 2 CUDN advertised", cudnAdvertised, layer2NetworkSpecGen),
+						ginkgo.Entry("Layer 3 CUDN EVPN IP-VRF", feature.EVPN, cudnAdvertisedEVPN, layer3IPVRFNetworkSpecGen),
+						ginkgo.Entry("Layer 2 CUDN EVPN MAC-VRF", feature.EVPN, cudnAdvertisedEVPN, layer2MACVRFNetworkSpecGen),
+						ginkgo.Entry("Layer 2 CUDN EVPN MAC-VRF and IP-VRF", feature.EVPN, cudnAdvertisedEVPN, layer2MACVRFIPVRFNetworkSpecGen),
```
pperiyasamy, line 2537: i hope these tests covers isolation with other UDN network (which is also configured with EVPN). do we connect these networks via IP-VRF ? should that be tested as well ?

test/e2e/route_advertisements.go
```diff
@@ -1935,148 +2082,156 @@ var _ = ginkgo.Describe("BGP: For a VRF-Lite configured network", feature.RouteA
...
-	var testBaseName, testSuffix, testNetworkName, bgpServerName string
+	var testBaseName, testSuffix, testNetworkName string
+	var externalServers []string
 
 	ginkgo.BeforeEach(func() {
 		if !isLocalGWModeEnabled() {
-			e2eskipper.Skipf("VRF-Lite test cases only supported in Local Gateway mode")
+			e2eskipper.Skipf("Test case only supported in Local Gateway mode")
 		}
+		framework.Logf("Running in ginkgo process %d", ginkgo.GinkgoParallelProcess())
```
pperiyasamy, line 2151: just for my understanding: so tests are in parallel now ?

test/e2e/route_advertisements.go
jcaamano: `feature.RouteAdvertisements` already serves that purpose. What do you need specifically?

test/e2e/route_advertisements.go
jcaamano: > i hope these tests covers isolation with other UDN network (which is also configured with EVPN)

It does, you can see the entries there.

> do we connect these networks via IP-VRF ? should that be tested as well ?

So what we are testing here is that the networks are isolated intra-cluster.
We are not testing interconnects that happen through an external router.
That is probably coverage that we should add for loose vs strict isolation modes.
But I guess not on this PR, which is strictly bringing the same coverage we had for VRF-Lite to EVPN.

test/e2e/route_advertisements.go
jcaamano: They are. It happens here:

https://github.com/ovn-kubernetes/ovn-kubernetes/blob/655e28f351b38e842ea42f85fd793bac86946f66/test/scripts/e2e-cp.sh#L275

test/e2e/route_advertisements.go
pperiyasamy: what if user wants to run test suite belong to:

BGP only with default network
BGP only with CUDN network
BGP VRFLite
BGP EVPN

At present this can be filtered out by text, having label may also be helpful here.

jcaamano:
Addressed comment on last push

test/e2e/route_advertisements.go
pperiyasamy: agreed, EVPN UDNs interconnect must happen through an external router.
VRF-Lite also deosn't have tests to cover loose isolation mode.
Let's do in a separate PR when needed.

test/e2e/route_advertisements.go
jcaamano: I am not sure I understand @pperiyasamy 

Isn't that what the `feature.RouteAdvertisements` label does?

We can add infinite labels. The labels that we have are related to features. I am not recognizing a different feature on what you are suggesting? 

Also not sure why we would do that in this PR.

pperiyasamy: i mean to add sub feature labels (something like `DefaultAdvertised`, `CUDNAdvertised`,  `CUDNAdvertisedVRFLite` and `CUDNAdvertisedEVPN`) under `feature.RouteAdvertisements`.
since this test file grows bigger. so above may provide a flexibility for the user to run a specific suite.
just feel free to ignore if such a requirement doesn't arise for u/s.

jcaamano: I think that, other than feature labels, we should add any additional labels on a need-by-need basis

PR #6073, opened 2026-03-06
CNC UT: fix flake and add a new test

### LINKED ISSUE #6072: FLAKE: "NetworkConnect ClusterManager Controller Integration Tests ClusterNetworkConnect ClusterManager Status Tests [It] sets Accepted=False condition on insufficient connect subnet size"
### Which jobs are flaking?

AI analisys that looks correct at first glance:
```
Looking at the test flow:                                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                            
  1. createPrimaryUDN creates the namespace, NAD, and updates fakeNM.PrimaryNetworks for each of the 3 namespaces                                                                                                                                                                                                                                                                                                                           
  2. Then it creates the CNC object, which triggers the controller's reconciliation                                                                                                                                                                                                                                                                                                                                                         
  3. The controller calls discoverSelectedNetworks → lists namespaces → calls GetActiveNetworkForNamespace per namespace

  The race is between the fake informer's namespace lister and the test's namespace creation. The controller uses c.namespaceLister.List(namespaceSelector) to discover namespaces, which relies on the informer cache. If the informer hasn't synced the 3rd namespace (frontend-c) into its cache by the time reconciliation runs, it won't be discovered.

  From the log output, you can see:
  allMatchingNetworkKeys=[layer3_2 layer3_1]
  Only 2 networks found — frontend-c was missed because its namespace hadn't made it into the informer cache yet.

  The test should either:
  - Wait for all 3 namespaces to be visible in the lister before creating the CNC, or
  - Create the namespaces before calling start() so informers have them from the beginning
```


https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/22725525265/job/65899472637

### Which tests are flaking?

FLAKE: "NetworkConnect ClusterManager Controller Integration Tests ClusterNetworkConnect ClusterManager Status Tests [It] sets Accepted=False condition on insufficient connect subnet size"

Recently

_No response_

_No response_

npinaeva:
## 📑 Description

cnc UT: make status check more reliable
It used to check the first posted Accepted condition, but if not all
network have been handled yet, it may be updated to the expected
status later.
cnc UT: add test to check consistent allocation on accepted=False

Fixes #6072

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Tests**
  * Refactored status checks to use error-returning, eventual-wait semantics for more reliable assertions.
  * Expanded test coverage with scenarios for network allocation, overlapping subnets, acceptance state transitions, restart/state preservation, cross-validation failures, and various matching edge cases.

### DIFF
--- a/go-controller/pkg/clustermanager/networkconnect/controller_test.go
+++ b/go-controller/pkg/clustermanager/networkconnect/controller_test.go
@@ -1291,23 +1291,27 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 			return nil
 		}
 
-		checkAcceptedCondition := func(cncName, errorSubstring string) {
+		checkAcceptedCondition := func(cncName, errorSubstring string) error {
+			cond := getCNCCondition(cncName, "Accepted")
+			if cond == nil {
+				return fmt.Errorf("missing Accepted condition")
+			}
+			if errorSubstring != "" && cond.Status == metav1.ConditionFalse &&
+				cond.Reason == "ResourceAllocationFailed" &&
+				strings.Contains(cond.Message, errorSubstring) {
+				return nil
+			}
+			if errorSubstring == "" && cond.Status == metav1.ConditionTrue &&
+				cond.Reason == "ResourceAllocationSucceeded" &&
+				cond.Message == "All validation checks passed successfully" {
+				return nil
+			}
+			return fmt.Errorf("unexpected error, expected: %s, got %s", errorSubstring, cond.Message)
+		}
+
+		checkAcceptedConditionEventually := func(cncName, errorSubstring string) {
 			gomega.Eventually(func() error {
-				cond := getCNCCondition(cncName, "Accepted")
-				if cond == nil {
-					return fmt.Errorf("missing Accepted condition")
-				}
-				if errorSubstring != "" && cond.Status == metav1.ConditionFalse &&
-					cond.Reason == "ResourceAllocationFailed" &&
-					strings.Contains(cond.Message, errorSubstring) {
-					return nil
-				}
-				if errorSubstring == "" && cond.Status == metav1.ConditionTrue &&
-					cond.Reason == "ResourceAllocationSucceeded" &&
-					cond.Message == "All validation checks passed successfully" {
-					return nil
-				}
-				return fmt.Errorf("unexpected error, expected: %s, got %s", errorSubstring, cond.Message)
+				return checkAcceptedCondition(cncName, errorSubstring)
 			}).WithTimeout(5 * time.Second).Should(gomega.Succeed())
 		}
 
@@ -1374,7 +1378,7 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 				// Verify status condition updated by controller
-				checkAcceptedCondition(cncName, "")
+				checkAcceptedConditionEventually(cncName, "")
 				return nil
 			}
 			err := app.Run([]string{app.Name})
@@ -1408,7 +1412,7 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 				gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 				// Verify status condition updated by controller
-				checkAcceptedCondition(cncName, "no subnets available")
+				checkAcceptedConditionEventually(cncName, "no subnets available")
 				return nil
 			}
 			err := app.Run([]string{app.Name})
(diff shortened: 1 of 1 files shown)


tssurya:
@npinaeva I also found maybe another small bug for the status side of things:
cluster_network_connect.go
Lines 95-118
    if existingCondition == nil || existingCondition.Status != *condition.Status {        // LastTransitionTime should only be updated in the Status (True/False) changes        condition = condition.WithLastTransitionTime(metav1.NewTime(time.Now()))    }
If CNC transitions from False/ResourceAllocationFailed (message A) to False/ResourceAllocationFailed (message B), lastTransitionTime wouldn't be set and SSA won't update the status because I think lastTransitionTime is a required field 
I hit this bug when I was doing VTEP controller status and thought CNC might be hitting the same thing - could you please check?
while we don't need to change the transitiontime since status didn't change.. we should probably send the same time as is so that if user fixes the issue manually and expects to see a different message we have one.

npinaeva:
> @npinaeva I also found maybe another small bug for the status side of things: cluster_network_connect.go Lines 95-118 if existingCondition == nil || existingCondition.Status != *condition.Status { // LastTransitionTime should only be updated in the Status (True/False) changes condition = condition.WithLastTransitionTime(metav1.NewTime(time.Now())) } If CNC transitions from False/ResourceAllocationFailed (message A) to False/ResourceAllocationFailed (message B), lastTransitionTime wouldn't be set and SSA won't update the status because I think lastTransitionTime is a required field I hit this bug when I was doing VTEP controller status and thought CNC might be hitting the same thing - could you please check? while we don't need to change the transitiontime since status didn't change.. we should probably send the same time as is so that if user fixes the issue manually and expects to see a different message we have one.

That makes sense, but I tested applying same status with a different message and no transition time and it worked. I am using 
```
metav1.ApplyOptions{
			FieldManager: fieldManager,
			Force:        true,
		},
```
which technically should update the existing condition because it is the same type, so the `lastTransitionTime` is preserved.
Can you show me the code in the VTEP controller where that happened? I want to see if something is different there

npinaeva:
/retest-failed
```
Waiting for daemon set "frr-k8s-daemon" rollout to finish: 0 of 3 updated pods are available...
error: timed out waiting for the condition
```

npinaeva:
reported a flake https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6094

go-controller/pkg/clustermanager/networkconnect/controller_test.go
```diff
@@ -1437,7 +1443,73 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
...
 
 				// Verify status condition updated by controller
-				checkAcceptedCondition(cncName, "selected networks have overlapping subnets")
+				checkAcceptedConditionEventually(cncName, "selected networks have overlapping subnets")
+				return nil
+			}
+			err := app.Run([]string{app.Name})
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+		ginkgo.It("sets Accepted=False and preserves network allocation on overlapping pod subnets", func() {
```
tssurya, line 1452: hmm so what I had in mind is

1. create CNCA with red and blue networks with connect subnet X - status true!
2. create CNCB with red and green networks with connect subnet X - status false! overlap for red network with two CNCs
3. restart ovnkube
4. will both CNCA and CNCB now to to false status ? or is it deterministic on restart based on timestamp and its always CNCA? or is it random today and it could be that if CNCB was processed first, its the one going into true state followed by CNCA in false status?

go-controller/pkg/clustermanager/networkconnect/controller_test.go
tssurya, line 1: ah based on commit message.. I guess RCA was correct from AI but solution was stupid? but yea this is not even a real bug.. unit test only issue thankfully

go-controller/pkg/clustermanager/networkconnect/controller_test.go
```diff
@@ -1291,23 +1291,27 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
 			return nil
 		}
 
-		checkAcceptedCondition := func(cncName, errorSubstring string) {
+		checkAcceptedCondition := func(cncName, errorSubstring string) error {
```
tssurya, line 1294: seems like we also use it here: https://github.com/npinaeva/ovn-kubernetes/blob/ac9bf795b16525633ce474789cb0f75e9a7b6130/go-controller/pkg/clustermanager/networkconnect/controller_test.go#L1504 so we need to keep it.

go-controller/pkg/clustermanager/networkconnect/controller_test.go
npinaeva: aah i see I did the subnet overlap withing one CNC and you meant overlap between CNCs, let me add that too.
So only 1 CNC will have subnets allocated for network red (whichever was reconciled first), and that CNC will have no error. CNC that was reconciled later will report an error and not allocate subnets for that network.
On restart we restore `selectedNetworks` from the annotation on initialSync, so it will be preserved.

go-controller/pkg/clustermanager/networkconnect/controller_test.go
npinaeva: yeah I copied commit description to the PR description now

go-controller/pkg/clustermanager/networkconnect/controller_test.go
tssurya: > aah i see I did the subnet overlap withing one CNC and you meant overlap between CNCs, let me add that too. 

TY!

>  On restart we restore `selectedNetworks` from the annotation on initialSync, so it will be preserved.

ah this is the part I was missing.. cool we are good then!
thanks for the tests!

go-controller/pkg/clustermanager/networkconnect/controller_test.go
```diff
@@ -1290,22 +1291,27 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
...
+				cond.Message == "All validation checks passed successfully" {
+				return nil
+			}
+			return fmt.Errorf("unexpected error, expected: %s, got %s", errorSubstring, cond.Message)
+		}
+
+		checkAcceptedConditionEventually := func(cncName, errorSubstring string) {
+			gomega.Eventually(func() error {
+				return checkAcceptedCondition(cncName, errorSubstring)
 			}).WithTimeout(5 * time.Second).Should(gomega.Succeed())
```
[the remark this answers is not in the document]
tssurya: @coderabbitai: but ordering events is not correct right? in real world its totally ok for a 3rd namespace to come a bit late.. and for the condition to go to false because of that

go-controller/pkg/clustermanager/networkconnect/controller_test.go
```diff
@@ -1439,7 +1443,168 @@ var _ = ginkgo.Describe("NetworkConnect ClusterManager Controller Integration Te
...
+				}).WithTimeout(500 * time.Millisecond).Should(gomega.Succeed())
+
+				// now restart the controller and check that the same subnet is still allocated (allocation preserved across restarts)
+				controller.Stop()
+				tunnelKeysAllocator := id.NewTunnelKeyAllocator("TunnelKeys")
+				controller = NewController(wf, fakeClientset, fakeNM.Interface(), tunnelKeysAllocator)
+				gomega.Expect(controller.Start()).To(gomega.Succeed())
+				// check status and subnets are the same
+				gomega.Consistently(func() error {
+					if err := checkAcceptedCondition(cncName1, ""); err != nil {
```
tssurya, line 1597: nice! exactly what I was looking for!

PR #6076, opened 2026-03-06
CNC: Add events from the ovnkube-controller side on errors

### LINKED ISSUE #5961: [ovnkube-controller side] Implement status reporting/ folding back relevant status to status manager in CM
### What is this issue about?

In the enhancement we decided to do this https://ovn-kubernetes.io/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns/#statusconditions per zone.
But there are perf/scale challenges for this maybe
I'll leave it to @npinaeva to think a bit more and see if we want to do this or not after the CM side status management is done

npinaeva:
## 📑 Description

In the original OKEP we wanted to report per-node status, but after more discussion we agreed that would cause too much overhead with not so much value.
In CNC code most of the validations are done in the CM, and per-node errors can just be reported as events.
So I am adding only events reporting from the nodes here and updating the OKEP

Fixes #5961

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Event recording added so ClusterNetworkConnect reconciliation failures emit Kubernetes events.

* **Tests**
  * Two e2e tests added to verify failure events are emitted.
  * Integration tests updated to exercise event recording.
  * Unit tests enhanced to handle Primary/Secondary network role scenarios.

* **Behavior / Validation**
  * Overlapping networks reported by human-friendly names; Secondary-role networks are rejected.

* **Documentation**
  * Status model updated to ResourceAllocationSucceeded/Failed with per-zone event examples.


npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5152
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6112

docs/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns.md
```diff
@@ -620,15 +620,16 @@ on validation results.
 
 **Condition Types**
 
-* The `Accepted` type condition will be done from the cluster-manager
-  which indicates whether the validation checks passed.
-* The `Ready-In-Zone-<nodename>` type condition will be done from
```
tssurya, line 625: do we want to add a paragraph to the alternatives section outlining why we don't want to do this pattern of per zone status and perf/scale implications? that way maybe we also make it a precedent and fix it for the existing features that do this..
otherwise its not clear why we changed the OKEP

docs/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns.md
```diff
@@ -620,15 +620,16 @@ on validation results.
...
-  OVN objects were created correctly.
-* Based on whether `status` of the above two types of conditions is
+* The `ResourceAllocationSucceeded` type condition will be done from the cluster-manager
+  which indicates whether the validation checks passed and resources were allocated successfully.
+* Based on whether `status` of the above condition is
   `True` or `False`, the final string `status` will be marked `Success` or
   `Failure`
 
+Per-node errors are less likely to happen because the validation and resource allocation happens in 
+cluster manager. Therefore, we will only emit events from the ovnkube-controller when CNC-related errors
```
tssurya, line 630: I am not fully convinced if we should emit events from ovnkube-controller or not - reason is:
1) this is going to be tooo many events if its the "same issue code wise" because each controller is the same - in this case is the user actually able to make meaningful interpretation of what's going on amongst all the events? - is there a chance we choke the API server or etcd or something
3) at 5000 node scale I fear it could be too many events and I'm not sure if there is precedent to stop emitting certain events if we start them ? i.e let's say we figure out this is too much --- maybe then its ok to stop emitting events?
2) its useful only if one of the nodes misbehaves we are able to catch that for sure - I see the value - I don't know if cluster metric or alerts is better for this?
4) do we want to wait for some user feedback before we do this from each node? 
Overall: I am convinced its the right move to not do status per node; whether events from each node is a good idea for a large scale cluster or not - I'm not sure

docs/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns.md
```diff
@@ -749,59 +741,48 @@ status:
...
+    failed to parse IPv6 subnet fd00:10::/1270 for owner layer2_11: invalid CIDR address:
+    fd00:10::/1270'
+  metadata:
+    creationTimestamp: "2026-03-10T10:18:27Z"
+    name: color-1-jkl6s.189b7388482e6877
+    namespace: default
+    resourceVersion: "103783"
+    uid: 815f3c57-434e-4193-804d-9f6cfbecd32c
+  reason: ErrorReconcilingCNC
+  reportingComponent: ovnk-controlplane
```
tssurya, line 775: hmm how is the reporting component ovnk-controlplane if the zone was `ovn-worker` in the error message?

docs/okeps/okep-5224-connecting-udns/okep-5224-connecting-udns.md
```diff
@@ -749,59 +741,48 @@ status:
...
+  reportingInstance: ""
+  source:
+    component: ovnk-controlplane
+  type: Warning
 ```
 
 **Accepted Condition Reasons (from cluster-manager):**
 
-- **`ValidationSucceeded`** - All validation checks passed successfully
```
tssurya, line 792: do we also want to mention this bit in the alternatives section on why we ditched this - cause its not easy to distinguish each error properly - otherwise in the far future I'm never going to remember why we changed approaches :D 

Note: I agree with the commit message around `InsufficientNetworks` can be removed

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -398,6 +418,7 @@ func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetwo
 					continue
 				}
 				if !network.IsPrimaryNetwork() {
+					errs = append(errs, fmt.Errorf("network %s has role 'Secondary'. Only 'Primary' role networks are supported", network.GetNetworkName()))
```
tssurya, line 421: hmm I guess its ok to return this error because we want to capture it in the status so that the user can do something with it?
i.e problem with error return is also that we keep retrying which doesn't do much good unless the user takes an action about this - like changing selectors OR deleting/recreate the network..?

tssurya (changes requested):
some initial comments

PR #4040, opened 2023-12-06
EIP multi NIC IPv6 support, copy routes, fixes and E2E tests

martinkennelly:
Major changes description 
- Update netlink lib to remove the 'any' route hacks. This means if we get an any route from the host, `dst` is no longer `nil` but will represent the correct value (i.e. for ipv4 `0.0.0.0/0`
- Add IPv6 support for EIP multi NIC. This involves removing tracking assigned EIPs by label and instead use a new node annotation
-  If an EIP is assigned to a link, copy all routes that are associated with this link to the new custom routing table. Inject default route if none exists.
- E2E tests
- Lots of refactors to route manager and iptables manager plus new tests for IPv6. (sorry reviewer). It improves the code quality.
- Remove requirement for `host-cidrs` annotation to only include global scoped ip addresses. Add requirement on the eip side when consuming this annotation to only consider global scoped ip addresses for eip assignment.
- Skip EIP multi NIC e2e tests for LGW mode. Looking for input from maintainers. The test fails for when a pod is located on an egress node and it wishes to communicate with another node ip. In LGW mode, the packet comes out mp0 and it is packet marked by OVN, but the current eip ip rules do not exclude packets that are marked by OVN and they get sent to be routed by the custom EIP routing tables instead of going to he main routing table.
I thought of adding an ip rule `ip rule add from fwmark 1008 table main` before the eip ip rules so that packets that are marked hit the main routing table. This works however when the reply traffic comes, it will get dropped due to reverse path filtering on `breth0`, so I am not sure what to do. Thinking about it, I probably need to also mark the return flows in `breth0`. This maybe also an issue for egress service in LGW mode.

PR #6090, opened 2026-03-09
egressservice: fix backend ClusterIP connectivity

karampok:
## Description

In LGW mode, packets from egress pods destined to node IPs are marked
with `pkt_mark=0x3f0` (1008) by an OVN logical router policy. An ip rule
matches this mark and directs traffic to the main routing table, ensuring
it is handled before egress IP and egress service routing rules.

Previously, this rule was only created by the EgressIP controller at
priority 5999, which meant:
- It was not created when EgressIP was disabled but EgressService was
  enabled.
- At priority 5999, it was evaluated after EgressService source-based
  ip rules (priority 5000), causing ClusterIP traffic from egress
  service backend pods to be misrouted through the egress network.

Fix both issues by:
- Changing the fwmark rule priority from 5999 to 4999 so it is
  evaluated before both EgressService (5000) and EgressIP (6000) rules.
- Adding the same rule creation to the EgressService controller via
  iprulemanager, so the rule exists even when EgressIP is disabled.
- Moving the shared constants and rule construction into the
  iprulemanager package to avoid duplication.

## Checks
- [ ] My code requires tests

## How to verify it
On a cluster with EgressService enabled and EgressIP disabled in LGW mode:
1. Create an EgressService with backend pods
2. From a backend pod on the egress host, curl a ClusterIP service
3. Verify the traffic is not misrouted through the egress network
4. Verify `ip rule` shows `prio 4999 fwmark 0x3f0 lookup main`

## Release Notes

  ```release-note
  Fix EgressService backend ClusterIP connectivity in local gateway mode.
  ```

### DIFF
--- a/go-controller/pkg/node/controllers/egressip/egressip.go
+++ b/go-controller/pkg/node/controllers/egressip/egressip.go
@@ -55,7 +55,6 @@ import (
 
 const (
 	rulePriority        = 6000 // the priority of the ip routing rules created by the controller. Egress Service priority is 5000.
-	ruleFwMarkPriority  = 5999 // the priority of the ip routing rules for LGW mode when we want to skip processing eip ip rules because dst is a node ip. Pkt will be fw marked with 1008.
 	routingTableIDStart = 1000
 	chainName           = "OVN-KUBE-EGRESS-IP-MULTI-NIC"
 	iptChainName        = utiliptables.Chain(chainName)
@@ -280,7 +279,7 @@ func (c *Controller) Run(stopCh <-chan struct{}, wg *sync.WaitGroup, threads int
 			}
 
 			// If dst is a node IP, use main routing table and skip EIP routing tables
-			if err = c.ruleManager.Add(getNodeIPFwMarkIPRule(netlink.FAMILY_V4)); err != nil {
+			if err = c.ruleManager.AddNodeIPFwMarkRule(netlink.FAMILY_V4); err != nil {
 				return fmt.Errorf("failed to create IPv4 rule for node IPs: %v", err)
 			}
 			// The fwmark of the packet is included in reverse path route lookup. This permits rp_filter to function when the fwmark is
@@ -309,7 +308,7 @@ func (c *Controller) Run(stopCh <-chan struct{}, wg *sync.WaitGroup, threads int
 
 			// If dst is a node IP, use main routing table and skip EIP routing tables
 			// src_valid_mark is not applicable to ipv6
-			if err = c.ruleManager.Add(getNodeIPFwMarkIPRule(netlink.FAMILY_V6)); err != nil {
+			if err = c.ruleManager.AddNodeIPFwMarkRule(netlink.FAMILY_V6); err != nil {
 				return fmt.Errorf("failed to create IPv6 rule for node IPs: %v", err)
 			}
 		}
@@ -1584,15 +1583,6 @@ func isValidIP(ipStr string) bool {
 		return false
 	}
 	return len(ip) > 0
-}
-
-func getNodeIPFwMarkIPRule(ipFamily int) netlink.Rule {
-	r := netlink.NewRule()
-	r.Priority = ruleFwMarkPriority
-	r.Mark = 1008 // pkt marked with 1008 is a node IP
-	r.Table = 254 // main
-	r.Family = ipFamily
-	return *r
 }
 
 func isVRFSlaveDevice(link netlink.Link) bool {
--- a/go-controller/pkg/node/controllers/egressservice/egressservice_node.go
+++ b/go-controller/pkg/node/controllers/egressservice/egressservice_node.go
@@ -11,6 +11,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/vishvananda/netlink"
+
 	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -32,6 +34,7 @@ import (
 	egressserviceinformer "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/egressservice/v1/apis/informers/externalversions/egressservice/v1"
 	egressservicelisters "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/egressservice/v1/apis/listers/egressservice/v1"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/factory"
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/iprulemanager"
 	nodenft "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/node/nftables"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/ovn/controller/services"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
@@ -74,6 +77,8 @@ type Controller struct {
 	endpointSliceLister  discoverylisters.EndpointSliceLister
 	endpointSlicesSynced cache.InformerSynced
 
+	ruleManager iprulemanager.Interface
+
 	services map[string]*svcState // svc key -> state
 }
 
@@ -97,10 +102,11 @@ func NewController(stopCh <-chan struct{}, returnMark, thisNode string,
 	klog.Info("Setting up event handlers for Egress Services")
 
 	c := &Controller{
-		stopCh:     stopCh,
-		returnMark: returnMark,
-		thisNode:   thisNode,
-		services:   map[string]*svcState{},
+		stopCh:      stopCh,
+		returnMark:  returnMark,
+		thisNode:    thisNode,
+		ruleManager: iprulemanager.NewController(config.IPv4Mode, config.IPv6Mode),
+		services:    map[string]*svcState{},
 	}
 
 	c.egressServiceLister = esInformer.Lister()
@@ -196,6 +202,24 @@ func (c *Controller) Run(wg *sync.WaitGroup, threadiness int) error {
 
 	if !util.WaitForInformerCacheSyncWithTimeout("egressservices_endpointslices", c.stopCh, c.endpointSlicesSynced) {
 		return fmt.Errorf("timed out waiting for endpoint slice caches (for egress services) to sync")
+	}
+
+	// In LGW mode, packets from egress pods destined to node IPs get marked with 0x3f0 (1008) by
+	// OVS flows. This ip rule ensures those packets use the main routing table, bypassing egress
+	// custom routing tables. EgressIP also creates this rule when enabled; the ruleManager
+	// reconciliation handles the overlap by checking if the rule already exists before adding.
+	if config.Gateway.Mode == config.GatewayModeLocal {
+		if config.IPv4Mode {
+			if err := c.ruleManager.AddNodeIPFwMarkRule(netlink.FAMILY_V4); err != nil {
+				return fmt.Errorf("failed to create IPv4 fwmark bypass rule: %v", err)
+			}
+		}
+		if config.IPv6Mode {
+			if err := c.ruleManager.AddNodeIPFwMarkRule(netlink.FAMILY_V6); err != nil {
+				return fmt.Errorf("failed to create IPv6 fwmark bypass rule: %v", err)
+			}
+		}
+		go c.ruleManager.Run(c.stopCh, 5*time.Minute)
 	}
 
 	klog.Infof("Repairing Egress Services")
--- a/go-controller/pkg/node/controllers/egressservice/egressservice_node_test.go
+++ b/go-controller/pkg/node/controllers/egressservice/egressservice_node_test.go
@@ -1,0 +1,93 @@
+// SPDX-FileCopyrightText: Copyright The OVN-Kubernetes Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package egressservice
+
+import (
+	"sync"
+
+	"github.com/onsi/ginkgo/v2"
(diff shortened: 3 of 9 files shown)


pliurh:
@karampok, there is a CI fix merged yesterday. You need to rebase your PR to kick off CI.

pliurh:
Now, this bypass rule is required when either EgressIP or EgressService is enabled. But the rule is only created by the egressIP controller when egressIP is enabled. It won't work if EgressIP is disabled.

karampok:
@trozet the ci failure seems not relevant (at least happened twice and is in shared mode). Thanks in advance for the review.

karampok:
Hi @trozet , will you have some time for that one? thanks

tssurya:
Where are tests... ? This change doesn't seem to add the tests towards kapi or whatever traffic broke for ESVC that prompted this fix.

karampok:
@tssurya  currently there is no test code around the EgressService that can be extended. Adding e2e from scratch  to fix the bug is not trivial (or do you think there is testing that i miss?).  The plan is to contribute in the future PRs that add e2e tests for this feature. cc @cgoncalves 

Thanks

tssurya:
> @tssurya currently there is no test code around the EgressService that can be extended. Adding e2e from scratch to fix the bug is not trivial (or do you think there is testing that i miss?). The plan is to contribute in the future PRs that add e2e tests for this feature. cc @cgoncalves
> 
> Thanks

for the specific bug in question tests must be added in the same PR. There is existing infra for tests @oribon can help point out where they are

karampok:
Okay, I will sync with @oribon and will  add more test. Thanks for the feedback @tssurya

karampok:
@coderabbitai review

go-controller/pkg/node/controllers/egressservice/egressservice_node.go
```diff
@@ -198,6 +204,24 @@ func (c *Controller) Run(wg *sync.WaitGroup, threadiness int) error {
 		return fmt.Errorf("timed out waiting for endpoint slice caches (for egress services) to sync")
 	}
 
+	// In LGW mode, packets from egress pods destined to node IPs get marked with 0x3f0 (1008) by
+	// OVS flows. This ip rule ensures those packets use the main routing table, bypassing egress
+	// custom routing tables. EgressIP also creates this rule when enabled; the ruleManager
+	// reconciliation handles the overlap by checking if the rule already exists before adding.
+	if config.Gateway.Mode == config.GatewayModeLocal {
```
oribon, line 211: why are we using the rule manager versus direct ip rule commands as the rest of the code here? I think if we'd use the command it'd be easier to modify the existing egressservice_test.go file as well
I guess using the ip rule manager entirely is the correct form and would require a big refactor, but I wonder if it's necessary for the small task in this pr or not in order to fix the bug

go-controller/pkg/node/controllers/egressservice/egressservice_node_test.go
```diff
@@ -0,0 +1,129 @@
...
+		gomega.Expect(util.SetExec(fexec)).To(gomega.Succeed())
+		nodenft.SetFakeNFTablesHelper()
+	})
+
+	ginkgo.AfterEach(func() {
+		util.ResetRunner()
+	})
+
+	ginkgo.Context("when controller starts in LGW mode", func() {
+		ginkgo.DescribeTable("installs fwmark bypass ip rule to prevent ClusterIP misrouting",
```
oribon: the name here kinda oversells the test (it doesn't check it installs anything related to bypass ip rules)
I think we should either rename this to "calls the rule manager with correct families" (and wrap all the scenarios in a table including SGW), or actually test that the correct rules are created.

go-controller/pkg/node/iprulemanager/ip_rule_manager_test.go
```diff
@@ -234,6 +235,38 @@ var _ = ginkgo.XDescribe("IP Rule Manager", func() {
...
 	})
 
+	ginkgo.Context("AddNodeIPFwMarkRule", func() {
+		ginkgo.DescribeTable("installs fwmark bypass rule for the given IP family",
+			func(family int) {
+				expectedRule := netlink.NewRule()
+				expectedRule.Priority = nodetypes.FwMarkBypassPriority
+				expectedRule.Mark = nodetypes.OvnKubeNodeSNATMarkValue
+				expectedRule.Table = 254
+
```
oribon, line 242: should we also enforce expectedRule.family here?

test/e2e/egress_services.go
```diff
@@ -1157,6 +1157,80 @@ metadata:
...
+			skipIfProtoNotAvailableFn(protocol, externalContainer)
+
+			egressNode := nodes[0].Name
+
+			ginkgo.By("Getting the Kubernetes API server ClusterIP")
+			apiClusterIP := getApiAddress()
+
+			ginkgo.By("Creating a backend pod for the EgressService on the first node")
+			_, err := createGenericPodWithLabel(f, "egress-pod", egressNode, f.Namespace.Name, command, podsLabels)
+			framework.ExpectNoError(err, "failed to create egress backend pod")
```
oribon: I think we should include another pod on a different node as well (just to be sure that interaction works)

karampok:
@coderabbitai review

test/e2e/egress_services.go
karampok: done

go-controller/pkg/node/iprulemanager/ip_rule_manager_test.go
karampok: done

go-controller/pkg/node/controllers/egressservice/egressservice_node_test.go
karampok: done

go-controller/pkg/node/controllers/egressservice/egressservice_node.go
karampok: I would prefer to keep
- this is the right abstraction
- egressService/IP must do changes that are aligned
- at the end a more rework probably remove the rest of "ip rule commands" (like in repair because ipmanager takes care to reconcile.

oribon: fair enough 👍

oribon (commented):
lgtm

karampok:
@tssurya hi, this is ready now with tests. Thanks

karampok:
@tssurya or @trozet do you think you will have some time for that? (there is a conflict but let's have it resolve before you take a look). Thanks

tssurya:
@karampok its on my queue, sorry for the delay but I want to first get @cgoncalves's EIP OKEP reviewed and @SchSeba's release PR is next I guess for encap ips tests.. after those two will get to this (hope that's the right priority on telco side)

karampok:
Hi @tssurya , I am back from PTO and I can rebase the PR if you want, let me know thanks

karampok:
Hello again @tssurya , any feedback here (or where this PR stands in your queue?). Thanks

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -54,9 +54,8 @@ import (
 )
 
 const (
-	rulePriority       = 6000 // the priority of the ip routing rules created by the controller. Egress Service priority is 5000.
-	ruleFwMarkPriority = 5999 // the priority of the ip routing rules for LGW mode when we want to skip processing eip ip rules because dst is a node ip.
```
kyrtapz, line 58: What happens to the old 5999 priority rule on upgrades? Is it cleaned up?

test/e2e/egress_services.go
```diff
@@ -1157,6 +1157,94 @@ metadata:
...
+			ginkgo.By("Creating a backend pod for the EgressService on the egress node")
+			_, err := createGenericPodWithLabel(f, "egress-pod", egressNode, f.Namespace.Name, command, podsLabels)
+			framework.ExpectNoError(err, "failed to create egress backend pod")
+
+			ginkgo.By("Creating a backend pod for the EgressService on a non-egress node")
+			_, err = createGenericPodWithLabel(f, "non-egress-pod", nonEgressNode, f.Namespace.Name, command, podsLabels)
+			framework.ExpectNoError(err, "failed to create non-egress backend pod")
+
+			ginkgo.By("Setting up custom routing table with blackhole on nodes")
+			setBlackholeRoutingTableOnNodes(providerCtx, nodes, externalContainer, blackholeRoutingTable, protocol == v1.IPv4Protocol)
```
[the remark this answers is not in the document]
kyrtapz: Is this e2e failing without the fix? I have concerns that it doesn't.
Can't we add a blackhole default route to table 100 or something like that?

go-controller/pkg/node/iprulemanager/ip_rule_manager.go
```diff
@@ -27,6 +28,7 @@ type Interface interface {
 	Delete(rule IPRule) error
 	DeleteWithMetadata(metadata string) error
 	OwnPriority(priority int) error
+	AddNodeIPFwMarkRule(ipFamily int) error
```
kyrtapz, line 31: I would like to keep iprulemanager.Interface feature-agnostic. It already exposes Add(IPRule), while AddNodeIPFwMarkRule embeds EgressIP/EgressService logic. 
Maybe instead we can expose the rule construction as a shared helper and have callers use ruleManager.Add(rule), or preferably configure this node-global bypass once above both feature controllers.
 I think we do something similar already for the shared “no-reroute” logical router policies(hopefully we can keep it simpler though 😅).

go-controller/pkg/node/controllers/egressservice/egressservice_node.go
```diff
@@ -74,6 +77,8 @@ type Controller struct {
 	endpointSliceLister  discoverylisters.EndpointSliceLister
 	endpointSlicesSynced cache.InformerSynced
 
+	ruleManager iprulemanager.Interface
```
kyrtapz, line 80: Adding a background rule manager to EgressService for a rule that, in the usual EgressIP-enabled configuration, is already managed elsewhere seems disproportionate. Are you planning to migrate all the other rules added here to leverage the manager?

kyrtapz (commented):
Have you verified it works as expected with EgressIP disabled?

Looking at https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4040/changes/3abe49d889f30915378aabe2db1d6e1480cfe0de it seems that the rule itself won't be enough.

Also please address the comments from the rabbit.

karampok:
@kyrtapz I hope I addressed your comments. I am uncertain if there is any reason when SharedGateway to have rulemanager (AFAIU no because in shared there is never kernel networking). Thanks

go-controller/pkg/node/default_node_network_controller.go
```diff
@@ -1092,6 +1098,18 @@ func (nc *DefaultNodeNetworkController) Start(ctx context.Context) error {
...
+	if config.Gateway.Mode == config.GatewayModeLocal &&
+		(config.OVNKubernetesFeature.EnableEgressIP || config.OVNKubernetesFeature.EnableEgressService) {
+		if err := setupNodeSNATMarkRoutingRules(nc.ruleManager); err != nil {
+			return err
+		}
+		nc.wg.Add(1)
+		go func() {
+			defer nc.wg.Done()
+			nc.ruleManager.Run(nc.stopChan, 5*time.Minute)
+		}()
```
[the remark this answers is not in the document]
karampok: @kyrtapz I see https://github.com/openshift/enhancements/pull/1407 there was once the need to support egressIP on the non ovn networks, but once localgateway mode was introduced  I do not see a lot the value. what would be a reason to use/support shared mode with escape to kernel networking vs just use local gateway.   In all cases I think I should just start ruleManager always as part of my PR

go-controller/pkg/node/iprulemanager/ip_rule_manager.go
karampok: ruleManager is agnostic

go-controller/pkg/node/controllers/egressservice/egressservice_node.go
karampok: yes, the plan would be to use the iprule manager and not doing custom ip rule commands (added a TODO). I do not think it should be in the scop of the bug

karampok:
> Have you verified it works as expected with EgressIP disabled?
> 
> Looking at [3abe49d](https://github.com/ovn-kubernetes/ovn-kubernetes/commit/3abe49d889f30915378aabe2db1d6e1480cfe0de) it seems that the rule itself won't be enough.
> 
> Also please address the comments from the rabbit.

I managed to test that, and indeed it fails, so probably this work is needed and is a different bug, a bit more hidden.
I would prefer that we merge this PR because in the context of OCP where egressIP egressService are always enabled both, this is customer bug and plan more work to fix that bug, refactor, discuss (do it make sense disable egressIP but enable egressService?)

kyrtapz:
> > Have you verified it works as expected with EgressIP disabled?
> > Looking at [3abe49d](https://github.com/ovn-kubernetes/ovn-kubernetes/commit/3abe49d889f30915378aabe2db1d6e1480cfe0de) it seems that the rule itself won't be enough.
> > Also please address the comments from the rabbit.
> 
> I managed to test that, and indeed it fails, so probably this work is needed and is a different bug, a bit more hidden. I would prefer that we merge this PR because in the context of OCP where egressIP egressService are always enabled both, this is customer bug and plan more work to fix that bug, refactor, discuss (do it make sense disable egressIP but enable egressService?)

I am fine with focusing on fixing the immediate issue as part of this. Let's highlight that in the PR descriotion.
I’d prefer removing the rule-manager refactor from this PR as well. The immediate bug only needs the existing rule priority changed from 5999 to 4999 (with the legacy cleanup).
EgressService still manages its other rules using the old direct ip commands, so this does not really make rule management shared. It also does not fully support EgressService without EgressIP because the connmark setup still belongs to the EgressIP controller.

PR #5832, opened 2026-01-07
Extract node processing from UDN controller into a single level driven controller

trozet:
Breaks node handling out of UDN controllers and into a single new, level-driven node controller that registers with the NAD controller per new dev guidelines. A single annotation cache is held and updated to reduce annotation parsing as well. Before this refactor, with 100 UDNs we would see parsing across all 100 UDN controllers to take:

```
BenchmarkParseNetworkAnnotationsPerNetwork/networks=100-24                 166           7065171 ns/op         4545678 B/op      63602 allocs/op
```

now with the refactor:
```
BenchmarkParseNetworkAnnotationsStatePerNetwork/networks=100-24            15978             72973 ns/op           46290 B/op        644 allocs/op
```

This results in almost 93-99x improvement across the board for 100 UDNs. Basically this reduces processing of a node event annotation from O(n) times to O(1).

## Release Notes

* **New Features**
  * Centralized node topology management across user-defined networks with improved state tracking and annotation caching
  * VRF kernel module support in test infrastructure

* **Bug Fixes**
  * Enhanced node reconciliation logic to properly handle subnet and network state changes
  * Improved node state synchronization between network controllers

* **Performance**
  * Added node annotation caching to reduce redundant parsing operations

PR #6104, opened 2026-03-12
Update the VTEP design for the EVPN OKEP

tssurya:
## 📑 Description

The VTEP design caveats were not detailed enough on the original enhancement. Also, we have some divergence in the implementation compared to design and this PR aims to update that.

~We need to fix on "kubernetes node ips are not supported for managed" v/s add proper validation for that v/s its allowed on unmanaged mode. If we are in agreement on that I can include that piece as well~ - we decided nodeIPs if needed to be used can only be done via unmanaged mode

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Documentation**
  * VTEP spec now accepts multiple CIDR blocks (cidrs) — up to 20 entries — replacing the single-cidr field.
  * Added strict validation: IPv4-only (IPv6 deferred), min 1 / max 20, append-only (no reordering or shrinking), no intra-VTEP overlap; runtime cross-VTEP overlap checks with CIDRsOverlap reason.
  * Clarified managed vs unmanaged IP behavior, node annotation model, removal of allocated IPs from status, and deletion protection while referenced.

### DIFF
--- a/docs/okeps/okep-5088-evpn.md
+++ b/docs/okeps/okep-5088-evpn.md
@@ -315,19 +315,71 @@ kind: VTEP
 metadata:
   name: evpn-vtep
 spec:
-  cidr: 100.64.0.0/24
+  cidrs:
+  - 100.64.0.0/24
   mode: managed
 ```
 
-The cidr field is mandatory. If the mode is "managed", then OVN-Kubernetes will handle allocating and assigning
-VTEP IPs per node. If the mode is not provided, or is "unmanaged", then it is left to some other provider to handle
-adding the IP address to each node from the subnet provided. In unmanaged mode, OVN-Kubernetes will find an interface on
-the node which has an IP address in the cidr, and use that IP address. Unmanaged mode may be preferred where a provider
-handles assigning VTEP IPs within its EVPN fabric. In this case OVN-Kubernetes cluster is integrating into an already
-configured EVPN fabric. Therefore VTEP IP provisioning may be done by the provider and configured for each node.
+The `cidrs` field is mandatory and accepts a list of CIDR blocks (up to 20) from which VTEP IPs are
+discovered or allocated. Multiple CIDRs allow capacity expansion over time without recreating the VTEP.
+Each CIDR can also be expanded by changing their mask but there might be cases where continuous blocks
+are not available for expansion.
 
-The IP address assigned or found by OVN-Kubernetes will be annotated to the node as `k8s.ovn.org/<vtep name>: <ip>`.
-If all nodes do not have an IP address annotated for this VTEP, the VTEP CR will go into error state.
+The following CEL validation rules are enforced on the `cidrs` list:
+
+- **Minimum 1, maximum 20 entries**: At least one CIDR must be provided.
+- **Append-only** (managed mode only): CIDRs cannot be removed from the list once added
+  (`size(self) >= size(oldSelf)`). This restriction only applies to managed mode because
+  ovnkube-cluster-manager tracks allocations per CIDR and removing a CIDR would complicate invalidating existing
+  assignments. In unmanaged mode the user controls IP assignment, so CIDRs can be freely added or removed and
+  when they are removed, we don't guarantee any SLAs around downtime.
+- **No reordering or shrinking** (managed mode only): Existing CIDRs must remain at the same position and
+  can only be expanded to a wider mask (e.g., `/24` can become `/16`, but not `/28`). Shrinking or
+  reordering is not allowed, as it would disrupt sequential allocation tracking. This restriction does not
+  apply to unmanaged mode.
+- **No intra-VTEP overlap**: CIDRs within the same VTEP must not overlap with each other.
+
+Cross-VTEP CIDR overlap validation is performed by the VTEP controller at runtime (not via CEL) since it
+requires inspecting other VTEP objects. If two VTEP CRs have overlapping CIDRs, both are set to `Accepted=False`
+with reason `CIDROverlap`. This is done so that we don't end up in confusion situations where we have
+two VTEPs with same CIDR ranges or overlapping CIDR ranges where for the first tunnel's nodeA get's an
+IPA from the first VTEP v/s for second tunnel nodeB gets an IPB from the second VTEP and now we advertise
+both of these out.
+
+The vtep controller will also add an internal validation to ensure only IPV4 CIDRs are accepted
+for VTEPs used by EVPN CUDNs. This restriction will be removed when FRR supports IPV6 VTEPs for EVPN.
+
+The default mode is "managed". If the mode is "managed", then OVN-Kubernetes will handle allocating and assigning
+VTEP IPs per node. When multiple CIDRs are specified, IPs are allocated sequentially from a CIDR in the order
+they appear in the spec; once a CIDR is exhausted, allocation continues from the next one. This is also why
+reordering of CIDRs is not allowed.
+If the mode is "unmanaged", then it is left to some other provider to handle adding the IP address to each node from the
+subnet provided. In unmanaged mode, OVN-Kubernetes will find an interface on the node which has an IP address in any of the VTEP
+CIDRs, and use that IP address. Unmanaged mode may be preferred where a provider handles assigning VTEP IPs within its
+EVPN fabric. In this case OVN-Kubernetes cluster is integrating into an already configured EVPN fabric. Therefore VTEP IP
+provisioning may be done by the provider and configured for each node.
+
+The IP address assigned (managed) or discovered (unmanaged) by OVN-Kubernetes will be annotated to the node as
+`k8s.ovn.org/vteps: {"vtep-name1": {"ips": ["ipv4", "ipv6"]}, "vtep-name2": {"ips": ["ipv4", "ipv6"]}}`.
+For managed mode, ovnkube-cluster-manager allocates IPs and writes the annotation. For unmanaged mode,
+ovnkube-node discovers IPs on the node that fall within the VTEP CIDRs and writes the annotation back.
+Even though the user or external automation (e.g. a loopback IP for EVPN) originally configured the IP on the
+interface, annotating it to the node object provides a central view through the Kubernetes API so that operators
+can inspect every node's VTEP IPs without SSH access. This is especially important for future use cases like
+multi VTEP Geneve tunnels where the VTEP IP may be DHCP-assigned and not known to the user ahead of time.
+ovnkube-cluster-manager then reads the annotation to validate all nodes have VTEP IPs.
+If any node does not have a matching IP, the VTEP CR will be set to `Accepted=False` with reason `AllocationFailed`.
+
+VTEP Status Conditions will have one type `Accepted` with the following reasons:
+
+- `Allocated` (`Accepted=True`): All nodes have a valid VTEP IP within the configured CIDRs.
+- `AllocationFailed` (`Accepted=False`): One or more nodes have no matching IP, ambiguous matches, or are missing the `vteps` annotation.
+- `CIDROverlap` (`Accepted=False`): The VTEP's CIDRs overlap with another VTEP's CIDRs. Both conflicting VTEPs are set to `Accepted=False`.
+
+The allocated VTEPIPs will not be included in the status. So the user has no way to know what went wrong when
+its `AllocationFailed` and which node to check to fix the situation from the status alone. The node side
+controller will add events in case something goes wrong with the allocation (in managed mode, if the VTEP IP
+could not be added to the loopback interface and in unmanaged mode, if discovery failed) to provide input to the user.
 
 ### CUDN CRD changes for EVPN
 
@@ -403,23 +455,34 @@ There are no foreseen changes required to the RA CRD.
 
 #### VTEP
 
-When a VTEP CR is created in managed mode, ovnkube-cluster-manager will handle
-assigning an VTEP IP to each node. If the cidr range includes the Kubernetes node IP for a node, then the node IP will be
-used for VTEP IP. Note, using the node IP should be avoided in most cases, as that IP will already be tied to a specific
-interface on the node. This prevents proper Layer 3 failure handling. While the node IP on a dedicated link could use
-bonding to prevent Layer 2 failover, if something goes wrong on the Layer 3 interface, or the leaf connected to that
-link goes down, then there is no failover.
-With EVPN, it is advantageous to assign the VTEP IP to a loopback interface, so that multihoming
-and failover handling can occur. If a link goes down to one leaf, BFD will fire, and there will be a mass withdrawal of
+When a VTEP CR is created in managed mode, ovnkube-cluster-manager will handle assigning a VTEP IP to each node and it
+will also handle annotating the node with the allocated IP address. Note that in managed mode, it is expected for the
+user to pick up a range of subnets for VTEPs that is solely for OVN-Kubernetes to provision and this can't be overlapping
+with IPs of other features like EgressIPs, ExternalIPs, KeepAlived VIPs etc. So if a user ends up adding an IP from the
+VTEP CIDRs ranges to the nodes and during allocation ovnkube-cluster-manager also allocates a VTEP IP that ends up in an
+ambiguous multiple VTEP IPs from same CIDR range situation. Moreover the user added IP could actually be an IP allocated
+to another node as the VTEP IP. So it is best practice to provide a non-overlapping range for managed VTEP CIDRs since
+such scenarios will not be handled gracefully. For managed mode, we will add an overlap check with the nodeCIDR range to
+ensure VTEP CIDR doesn't match or overlap with node IP ranges in the cluster.
+
+If the user wants to use their nodeIPs as the VTEP IPs, then they should use unmanaged mode. If the CIDR range in
+unmanaged VTEP includes the Kubernetes node IP for a node, then the node IP will be used for VTEP IP. Note, using the
+node IP should be avoided in most cases, as that IP will already be tied to a specific interface on the node. This
+prevents proper Layer 3 failure handling. While the node IP on a dedicated link could use bonding to prevent Layer 2
+failover, if something goes wrong on the Layer 3 interface, or the leaf connected to that link goes down, then there is
+no failover. With EVPN, it is advantageous to assign the VTEP IP to a loopback interface, so that multihoming and
+failover handling can occur. If a link goes down to one leaf, BFD will fire, and there will be a mass withdrawal of
 routes, moving all traffic to a second leaf.
 
-ovnkube-cluster-manager will handle annotating the node with the assigned IP address. If the VTEP is in unmanaged mode,
-then ovnkube-cluster-manager will only handle checking that all nodes have an IP address annotated for this VTEP. If a
-node is missing the IP, the VTEP will be updated with a failure status condition. Even if a single node fails, the other
-healthy nodes with assigned VTEPs will be configured for EVPN correctly.
+If the VTEP is in unmanaged mode, ovnkube-node discovers IPs on each node that fall within the VTEP CIDRs and writes
+the `k8s.ovn.org/vteps` annotation. ovnkube-cluster-manager then reads this annotation to validate that every node
+has a valid VTEP IP. If a node does not have a matching IP, or has ambiguous matches (multiple IPs of the same address
+family within the VTEP CIDRs), the VTEP is set to `Accepted=False` with reason `AllocationFailed`. The
+RouteAdvertisements controller reads the same `k8s.ovn.org/vteps` annotation when generating FRR configurations to
+advertise the VTEP.
 
-For unmanaged, the ovnkube-node component will handle detecting the IP address on the Linux node and setting the node
-annotation.
+A finalizer (`k8s.ovn.org/vtep-protection`) is added to the VTEP to prevent deletion while it is referenced
+by any CUDN. The VTEP controller re-evaluates the finalizer on CUDN delete events.
 
 #### EVPN
 
@@ -884,6 +947,24 @@ manage. The Linux interfaces on the host must be moved to this other namespace a
 
 For OVN-Kubernetes, the cost of implementing the code to manage the netdevs, and configure FRR-K8S outweighs the
 drawbacks listed above, and therefore OpenPERouter is not a viable alternative.
+
+### VTEP Design
+
+#### Unmanaged Mode: Discover from host-cidrs Instead of a Dedicated Annotation
+
+Another design for unmanaged VTEP mode had ovnkube-cluster-manager discover VTEP IPs by reading
+the existing `k8s.ovn.org/host-cidrs` node annotation and matching those IPs against the VTEP CIDRs.
+This avoided introducing a new annotation, since `host-cidrs` already contains all IPs configured on
+the node (and will be extended to include loopback interface IPs). The RouteAdvertisements controller
+would independently perform the same discovery when generating FRR configurations.
+
+This approach was rejected primarily because users would have no way to know which IP was selected as
+the VTEP IP for a given node without inspecting the `host-cidrs` annotation and mentally replaying
+the matching logic. Surfacing the allocated IPs in `VTEP.Status.NodeAllocations` was considered, but
+at scale (e.g. 5000 nodes) persisting per-node IP mappings in the status of every VTEP object was
+deemed too expensive. So on hand we say for unmanaged mode things are upto the user but on the other
+hand we wanted to provide insight into the VTEP IPs through another a new annotation which isn't the
+best API.
 
 ## References
 


docs/okeps/okep-5088-evpn.md
```diff
@@ -413,13 +447,18 @@ With EVPN, it is advantageous to assign the VTEP IP to a loopback interface, so
...
 
-ovnkube-cluster-manager will handle annotating the node with the assigned IP address. If the VTEP is in unmanaged mode,
-then ovnkube-cluster-manager will only handle checking that all nodes have an IP address annotated for this VTEP. If a
-node is missing the IP, the VTEP will be updated with a failure status condition. Even if a single node fails, the other
-healthy nodes with assigned VTEPs will be configured for EVPN correctly.
+In managed mode, ovnkube-cluster-manager will handle annotating the node with the allocated IP address.
+
+If the VTEP is in unmanaged mode, ovnkube-cluster-manager discovers VTEP IPs by reading the existing
+`k8s.ovn.org/host-cidrs` annotation on each node and matching those IPs against the VTEP CIDRs instead of writing a
+per-VTEP annotation to the node. If a node has no matching IP, or has ambiguous matches (multiple IPs of the same
```
jcaamano: It looks like we are only consider the possibility of ambiguous scenarios for unmanaged VTEPs.

What happens if while handling a managed VTEP, we notice an IP on the node on the VTEP subnet that is not the one that we allocated and annotated. 
If we think this is an error because we might have allocated or might no want to allocate the rogue IP to another node, then we should stop reconciling the VTEP completely and track this as a different reason. Perhaps rename `AllocationFailed` to `NodeAllocationFailed` which would mean some nodes are effected but not other nodes. And then we could have `VTEPAllocationFailed` to signal that the VTEP is failed completely and not reconciling.
If we think we should handle that situation more gracefully, which I consider complicated, then we probably need to reserve that rogue IP in our allocator to ensure it will not be allocated to a different node if it hasn't already?

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+If any node does not have a matching IP, the VTEP CR will be set to `Accepted=False` with reason `AllocationFailed`.
+
+VTEP Status Conditions will have one type `Accepted` with the following reasons:
+
+- `Allocated` (`Accepted=True`): All nodes have a valid VTEP IP within the configured CIDRs.
+- `AllocationFailed` (`Accepted=False`): One or more nodes have no matching IP, ambiguous matches, or are missing the `host-cidrs` annotation.
+- `CIDROverlap` (`Accepted=False`): The VTEP's CIDRs overlap with another VTEP's CIDRs. Both conflicting VTEPs are set to `Accepted=False`.
+
+The allocated VTEPIPs will not be included in the status (subject to change when another OKEP update is posted). So
+the user has no way to know what went wrong when its `AllocationFailed` and which node to check to fix the situation.
```
jcaamano: Other than maybe events or the like.

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+the node which has an IP address in the VTEP CIDRs, and use that IP address. Unmanaged mode may be preferred where a provider
 handles assigning VTEP IPs within its EVPN fabric. In this case OVN-Kubernetes cluster is integrating into an already
 configured EVPN fabric. Therefore VTEP IP provisioning may be done by the provider and configured for each node.
 
-The IP address assigned or found by OVN-Kubernetes will be annotated to the node as `k8s.ovn.org/<vtep name>: <ip>`.
-If all nodes do not have an IP address annotated for this VTEP, the VTEP CR will go into error state.
+For managed mode, the IP address assigned by OVN-Kubernetes will be annotated to the node as `k8s.ovn.org/<vtep name>: <ip>`.
+For unmanaged mode, ovnkube-cluster-manager discovers VTEP IPs by matching the existing `k8s.ovn.org/host-cidrs`
+node annotation against the VTEP CIDRs instead of writing a per-VTEP annotation to the node. The `host-cidrs`
+annotation will be extended to also include IPs from loopback interfaces, which is the recommended interface for
```
jcaamano: There is no need to extend it, it already includes IPs from loopback interfaces:
```
[jcaamano@sdn-08 ~]$ ssh core@192.168.111.25 ip a show lo
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet 10.0.64.12/32 scope global lo
    inet6 ::1/128 scope host 
[jcaamano@sdn-08 ~]$ oc get node worker-2.ostest.test.metalkube.org -o yaml | grep host-cidr
    k8s.ovn.org/host-cidrs: '["10.0.64.12/32","172.22.0.150/24","192.168.111.25/24","192.168.221.25/24","fd2e:6f44:5dd8:c956::19/128","fd2e:6f44:5dd8:ca56::19/128"]'
```

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
 spec:
-  cidr: 100.64.0.0/24
+  cidrs:
+  - 100.64.0.0/24
   mode: managed
 ```
 
-The cidr field is mandatory. If the mode is "managed", then OVN-Kubernetes will handle allocating and assigning
-VTEP IPs per node. If the mode is not provided, or is "unmanaged", then it is left to some other provider to handle
+The `cidrs` field is mandatory and accepts a list of CIDR blocks (up to 10) from which VTEP IPs are
```
trozet: why restrict up to 10? We might need to add a lot of /32s for our use case.

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+discovered or allocated. Multiple CIDRs allow capacity expansion over time without recreating the VTEP.
+Each CIDR can also be expanded by changing their mask but there might be cases where continuous blocks
+are not avaliable for expansion.
+Only IPv4 CIDRs are currently supported; IPv6 VTEP endpoints are not yet supported by FRR for EVPN transport.
```
trozet: We plan for this VTEP CRD to be used for more things than just EVPN. I would not prevent support for IPv6 here in the CRD and instead validate that IPv6 cannot be used when EVPN is the one selecting the VTEP.

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+
+- **Minimum 1, maximum 10 entries**: At least one CIDR must be provided.
+- **IPv4 only**: All CIDRs must be IPv4 (`cidr(c).ip().family() == 4`). This restriction will be removed
+  when FRR supports IPV6 VTEPs for EVPN.
+- **Append-only**: CIDRs cannot be removed from the list once added (`size(self) >= size(oldSelf)`).
+- **No reordering or shrinking**: Existing CIDRs must remain at the same position and can only be expanded
+  to a wider mask (e.g., `/24` can become `/16`, but not `/28`). Shrinking or reordering is not allowed.
+- **No intra-VTEP overlap**: CIDRs within the same VTEP must not overlap with each other.
+
+Cross-VTEP CIDR overlap validation is performed by the VTEP controller at runtime (not via CEL) since it
```
trozet, line 342: why do we need to do cross CIDR overlap validation? Don't we just pick one on the host that fits?

docs/okeps/okep-5088-evpn.md
trozet: I thought there was a filter to host cidrs and we do not include certain types of interfaces/IPs. I'll need to check.

Either way... I feel like selecting an IP and annotating the node derived from another annotation is not the best approach. I suggest letting the ovnkube-node process watch for VTEPs, then it annotates its own node once it sees an IP that fits.

docs/okeps/okep-5088-evpn.md
trozet: when a node does not have a VTEP, why dont we just post a condition and list all nodes that are in failure state? We could delay reporting failures after initial VTEP creation to give nodes a few seconds to process the VTEP and annotate their node. After that delay, other nodes that do not have an annotation are put into an error status condition. If no pods are in error state then the status condition is just "VTEP allocated" or something.

docs/okeps/okep-5088-evpn.md
```diff
@@ -404,7 +438,7 @@ There are no foreseen changes required to the RA CRD.
 #### VTEP
 
 When a VTEP CR is created in managed mode, ovnkube-cluster-manager will handle
-assigning an VTEP IP to each node. If the cidr range includes the Kubernetes node IP for a node, then the node IP will be
+assigning a VTEP IP to each node. If the CIDR range includes the Kubernetes node IP for a node, then the node IP will be
```
trozet: I would just remove this or change it to say. If the user wants to use the node IP as the VTEP, then they should use unmanaged mode and provide the node IP subnets as CIDRs.

docs/okeps/okep-5088-evpn.md
trozet: I would not handle this use case. If the mode is managed, then we assigned an IP. If the user has already provisioned IPs on that subnet, it is not our problem. Let's just document that if using managed mode, you should pick a subnet that is entirely for OVN-Kubernetes to provision.

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+- **No reordering or shrinking**: Existing CIDRs must remain at the same position and can only be expanded
+  to a wider mask (e.g., `/24` can become `/16`, but not `/28`). Shrinking or reordering is not allowed.
+- **No intra-VTEP overlap**: CIDRs within the same VTEP must not overlap with each other.
+
+Cross-VTEP CIDR overlap validation is performed by the VTEP controller at runtime (not via CEL) since it
+requires inspecting other VTEP objects. If two VTEPs have overlapping CIDRs, both are set to `Accepted=False`
+with reason `CIDRsOverlap`.
+
+The default mode is "managed". If the mode is "managed", then OVN-Kubernetes will handle allocating and assigning
+VTEP IPs per node. If the mode is "unmanaged", then it is left to some other provider to handle
```
trozet: Maybe I missed it, but we need some explanation for what happens when you have managed mode + a list of CIDRs. Do we just just allocate sequentially based on the order of subnets in the CR, until one is exhausted, then we move to the next one?

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+Each CIDR can also be expanded by changing their mask but there might be cases where continuous blocks
+are not avaliable for expansion.
+Only IPv4 CIDRs are currently supported; IPv6 VTEP endpoints are not yet supported by FRR for EVPN transport.
+
+The following CEL validation rules are enforced on the `cidrs` list:
+
+- **Minimum 1, maximum 10 entries**: At least one CIDR must be provided.
+- **IPv4 only**: All CIDRs must be IPv4 (`cidr(c).ip().family() == 4`). This restriction will be removed
+  when FRR supports IPV6 VTEPs for EVPN.
+- **Append-only**: CIDRs cannot be removed from the list once added (`size(self) >= size(oldSelf)`).
```
trozet: if its managed I could see why this is a problem, but unmanaged if a user removes an entry, then the node would just need to notice it, clear its vtep assignment, and search on its host for another suitable IP for the VTEP CIDRs.

trozet (changes requested):

docs/okeps/okep-5088-evpn.md
jcaamano: We would not annotate the node for unmanaged mode,

If we already have a process that is already subscribing to netlink, already collecting node IPs, already annotating those node IPs, why would create yet another process that would do much of the same thing for a subset of those IPs?

I would understand it if there were something else that we needed to do, but as of now there is nothing else and it feels basically like duplicating the process. Maybe we can add it later if needed.

tssurya: > I thought there was a filter to host cidrs and we do not include certain types of interfaces/IPs. I'll need to check.

That's what I thought too but @kyrtapz confirm's @jcaamano 's theory above. He has already implemented this part on the node side. I'll update the wordings here to indicate we get it out of the box.

> 
> Either way... I feel like selecting an IP and annotating the node derived from another annotation is not the best approach. I suggest letting the ovnkube-node process watch for VTEPs, then it annotates its own node once it sees an IP that fits.

I'm not sure I follow - we are not planning to annotate the node again based on the IP we got from host-cidrs annotation.
For unmanaged VTEPs the plan is to only use host-cidrs - no other annotations.

docs/okeps/okep-5088-evpn.md
tssurya: then its easy to bump it up then if we have valid use cases during the implementation of multiple VTEP proposal - its a one liner change - I think that will have its own OKEP?
its always easy to expand a rule - its hard to go back to restrict it post expansion. for EVPN scenarios 10 sounds ok to me

docs/okeps/okep-5088-evpn.md
tssurya: > We plan for this VTEP CRD to be used for more things than just EVPN. 

I understand that but it got introduced in this OKEP which is meant for EVPN only 
I assume Lei's feature has its own OKEP which can remove this limitation - again, its very easy to remove the CEL rules - its harder to add them.

> I would not prevent support for IPv6 here in the CRD and instead validate that IPv6 cannot be used when EVPN is the one selecting the VTEP.

I can do this if you still feel strongly about solving Lei's OKEP use cases here (which btw is it https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5550/changes ?).

docs/okeps/okep-5088-evpn.md
tssurya: because you can have a situation of partial overlaps with VTEPA and VTEPB and that can lead to confusions on which range actually got applied? its bad API when we allow users to configure multiple ways to do the same thing. then on the status of both of them there could be false v/s true which also makes it confusing..

docs/okeps/okep-5088-evpn.md
tssurya: ack I just wanted to get consensus on this before I did that, will update this

docs/okeps/okep-5088-evpn.md
tssurya: yes! if we have managed mode - we allocate it sequentially - that's also why reordering subnets in the spec is blocked  - one can only add a new subnet as append..
I'll call this point out, thanks for catching this.

docs/okeps/okep-5088-evpn.md
tssurya: > if its managed I could see why this is a problem, but unmanaged if a user removes an entry, then the node would just need to notice it, clear its vtep assignment, and search on its host for another suitable IP for the VTEP CIDRs.

that's true and we did think about this use case already internally when we discussed it - we just decided to keep things simple for now with adding cidr + allowing mask expansion for single cidr v/s remove cidr and/or allow contraction of single cidr mask change for future because removing can cause traffic disruption big time? and we are not going to be able to provide SLAs on no downtime during such updates etc and there might be users who find this unacceptable - i.e if they added a new range and deleted a range in one go - they might expect us to do seamless transition which I didn't want to get into the details of

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+- **No reordering or shrinking**: Existing CIDRs must remain at the same position and can only be expanded
+  to a wider mask (e.g., `/24` can become `/16`, but not `/28`). Shrinking or reordering is not allowed.
+- **No intra-VTEP overlap**: CIDRs within the same VTEP must not overlap with each other.
+
+Cross-VTEP CIDR overlap validation is performed by the VTEP controller at runtime (not via CEL) since it
+requires inspecting other VTEP objects. If two VTEPs have overlapping CIDRs, both are set to `Accepted=False`
+with reason `CIDRsOverlap`.
+
```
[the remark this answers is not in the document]
tssurya: that's the whole point! we are updating the design before we can update the implementation.

docs/okeps/okep-5088-evpn.md
tssurya: this is done.

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,53 @@ kind: VTEP
...
+VTEP IPs per node. If the mode is "unmanaged", then it is left to some other provider to handle
 adding the IP address to each node from the subnet provided. In unmanaged mode, OVN-Kubernetes will find an interface on
-the node which has an IP address in the cidr, and use that IP address. Unmanaged mode may be preferred where a provider
+the node which has an IP address in the VTEP CIDRs, and use that IP address. Unmanaged mode may be preferred where a provider
 handles assigning VTEP IPs within its EVPN fabric. In this case OVN-Kubernetes cluster is integrating into an already
 configured EVPN fabric. Therefore VTEP IP provisioning may be done by the provider and configured for each node.
 
-The IP address assigned or found by OVN-Kubernetes will be annotated to the node as `k8s.ovn.org/<vtep name>: <ip>`.
-If all nodes do not have an IP address annotated for this VTEP, the VTEP CR will go into error state.
+For managed mode, the IP address assigned by OVN-Kubernetes will be annotated to the node as `k8s.ovn.org/<vtep name>: <ip>`.
```
tssurya: @trozet @jcaamano this annotation can't be named without a prefix as a random string. We need `vtep-` prefix. I'm updating that part as well.

tssurya: @kyrtapz what did you use on your side? Does the `vtep-` prefix work for you? I'll just update this as well while we are on it

docs/okeps/okep-5088-evpn.md
tssurya: ack I'll update this with what Tim said,
hope that's fine for you @jcaamano ?
I'll at least include this scenario as a thought off one

jcaamano: not sure what "not handle this use case." means.

Not handle it gracefully? I am fine with that.
Not notice the problem and report an error? Well, it wouldn't be my choice but it is only me, then OK.

tssurya: I've updated this part to say we don't plan to handle it gracefully and the expectation from users for this.

docs/okeps/okep-5088-evpn.md
jcaamano: We can't use the same IP for two different VTEPs on the same node, right?

docs/okeps/okep-5088-evpn.md
tssurya: > when a node does not have a VTEP, why dont we just post a condition and list all nodes that are in failure state? We could delay reporting failures after initial VTEP creation to give nodes a few seconds to process the VTEP and annotate their node. After that delay, other nodes that do not have an annotation are put into an error status condition. If no pods are in error state then the status condition is just "VTEP allocated" or something.

I brought this up in our slack thread - i.e the idea of posting a condition with a list of all nodes - I think we have to be prepared for worst case at 5000 nodes and that won't be very readable if all nodeNames get comma joined to a string... that was my counter argument at that time -  https://cloud-native.slack.com/archives/C08452HR8V6/p1773129528755329?thread_ts=1772621093.938169&cid=C08452HR8V6
last part there
```
also, the enhancement mentions to put the VTEP CRD into false status if at least one of the nodes doesn't have assigned VTEP (and we continue programming remaining nodes) - but how does a user use this status to debug further when they see the error allocationfailed? they know "some node" doesn't have it and having a string with all failing nodenames is also a lot at 5000 nodes...
```

docs/okeps/okep-5088-evpn.md
jcaamano: What I would do is have the CRD qualifier
```
<vtep name>.vteps.k8s.ovn.org/cidrs: { ipv4, ipv6}
```
better make it a slice now if we ever support want to support dualstack in a single annotation

docs/okeps/okep-5088-evpn.md
tssurya: > Other than maybe events or the like.

I am adding events for most failures - but for the per node ones unsure if we should be doing 1 event for all failed nodes or per node event , either ways if we think at 5000 node scale its always going to be a lot

docs/okeps/okep-5088-evpn.md
kyrtapz: We historically did one static key with dynamic map values but I am ok to have it this way if we don't envision a ton of VTEPs in the future.

kyrtapz: If we end up doing the vtep-name based key lets make sure we won't potentially hit any length limits.

jcaamano: So that would be
```
vteps.k8s.ovn.org/<vtep name>: {"cidrs": ["ipv4", "ipv6"]}"
```
I don't care either.

tssurya: > If we end up doing the vtep-name based key lets make sure we won't potentially hit any length limits.

this is a good point 
I think a key limit is 253 chars ? and the metadata name is also 253 chars isn't it...let me check
and yea https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set we are over limit for this one.....

anything before "/" should be less than 253 and we could now go to 253+18 = 271

tssurya: > So that would be
> 
> ```
> vteps.k8s.ovn.org/<vtep name>: {"cidrs": ["ipv4", "ipv6"]}"
> ```
> 
> I don't care either.

no I think pat means:

`k8s.ovn.org/vteps: {"vtep-name1": ["ipv4", "ipv6"], "vtep-name2": [ipv4, ipv6]}"` like we do for networks,

tssurya: I have updated to the latest one I posted ^ I'm resolving this thread.. but please open a new thread if the new scheme is not acceptable.

docs/okeps/okep-5088-evpn.md
tssurya: this is done, resolving comment, PTAL.

docs/okeps/okep-5088-evpn.md
tssurya: updated for more clarity

docs/okeps/okep-5088-evpn.md
tssurya: > Not notice the problem and report an error? Well, it wouldn't be my choice but it is only me, then OK.

for managed mode so far on node controller side our plan is to only set the IP to loopback.. if we want to do this then we should also watch for interfaces and see if they have something already in that subnet?  I think there isn't a simpler non-racy way from CM side (maybe only thing is check if host-cidr ends up having one IP already from this CIDR then we abort?)

also if we decide to notice this problem and report error - what happens on day2 ? i.e say the VTEP was allocated already and that time nothing was wrong - now the user goes and adds a rogue IP - then in next reconcile we should catch it ? or we should have a poll watcher subscribing to events from all interfaces?

docs/okeps/okep-5088-evpn.md
jcaamano: That would need to be then 
```
k8s.ovn.org/vteps: {"vtep-name1": { "cidrs" :["ipv4", "ipv6"]}, "vtep-name2": {"cidrs": ["ipv4", "ipv6"]}"
```
As long as unmarshaling this big json is not a concern

docs/okeps/okep-5088-evpn.md
jcaamano: `host-cidrs` is updated with the rogue IP which makes our controller reconcile so we would catch it and report the error.

docs/okeps/okep-5088-evpn.md
```diff
@@ -315,19 +315,60 @@ kind: VTEP
...
+they appear in the spec; once a CIDR is exhausted, allocation continues from the next one. This is also why
+reordering of CIDRs is not allowed.
+If the mode is "unmanaged", then it is left to some other provider to handle adding the IP address to each node from the
+subnet provided. In unmanaged mode, OVN-Kubernetes will find an interface on the node which has an IP address in any of the VTEP
+CIDRs, and use that IP address. Unmanaged mode may be preferred where a provider handles assigning VTEP IPs within its
+EVPN fabric. In this case OVN-Kubernetes cluster is integrating into an already configured EVPN fabric. Therefore VTEP IP
+provisioning may be done by the provider and configured for each node.
+
+For managed mode, the IP address assigned by OVN-Kubernetes will be annotated to the node as
+`k8s.ovn.org/vtep-ips: {"vtep-name1": ["ipv4", "ipv6"], "vtep-name2": ["ipv4", "ipv6"]}`.
```
jcaamano: Just want to have this comment visible since we are changing the original approach of one annotation per VTEP: we are not concerned about marshaling/unmarshaling big jsons?

Also if we take this approach I think it should be
```
k8s.ovn.org/vteps: {"vtep-name1": { "cidrs" :["ipv4", "ipv6"]}, "vtep-name2": {"cidrs": ["ipv4", "ipv6"]}"
```
just in case we want to annotate more things per vtep in the future

jcaamano: > just in case we want to annotate more things per vtep in the future

I guess if the point is to avoid multiple annotations, we would not like to have to add another annotation if we needed to annotate something else for a vtep

docs/okeps/okep-5088-evpn.md
trozet: I mean do not even try to detect this problem. We don't detect a user consuming IPs on other subnets they configure in OVNK. I know this has happened in the past and we find out from a bug report, but we do not proactively try to detect it in the system. I would take the same approach here.

docs/okeps/okep-5088-evpn.md
trozet: events are fine with a generic failed status in the CR itself. I would have each node post their own event for their node.

docs/okeps/okep-5088-evpn.md
trozet: because host cidrs != vtep IP. So if my host cidrs are 172.18.0.5, 172.18.0.6, and my VTEP selector cidr is 172.18.0.0/24, with your approach I'm supposed to infer one of those is my VTEP IP? I would have the node specifically annotate the chosen VTEP to the node.

docs/okeps/okep-5088-evpn.md
trozet: I think for our use case we will need a lot. We may have an unmanaged mode where there are many VTEP IPs allocated from /31 subnets. We want to re-use this VTEP CR for multi-vtep (rather than relying on OVS annotations). Can we please remove this limit or make it something much higher?

docs/okeps/okep-5088-evpn.md
trozet: Yeah that is the OKEP. I think regardless of whether or not that OKEP is merged, validation in the CEL rules is not the right thing. We should not tie VTEP to EVPN, that's the reason we made it a separate and generic CRD. So I feel strongly validation for EVPN should be done in EVPN and not in VTEP CRD. If you add CEL for VTEP because of EVPN, then sometime later we remove CEL for VTEP and now we have to move it into EVPN validation, why didn't we just start with EVPN validation?

docs/okeps/okep-5088-evpn.md
trozet: oh when I read Cross-VTEP CIDR overlap, I thought you meant "cross cidr overlap within the same VTEP CR", but I think you are talking about "cross-vtep cidr overlap bertween two VTEP CRs"?

docs/okeps/okep-5088-evpn.md
trozet: I don't think Kubernetes API implies SLA of downtime. Let's take network policy for example. Configuring a new network policy has no guarantee about how long it takes to enforce, and NP status was even rejected upstream because this is a moving target.

Just because we allow a user to switch a VTEP on a node, doesn't imply there is no downtime. As a user I would expect if i edit a VTEP to remove a cidr and add a new one that affects one or more nodes, that the status of the CRD will change from "Allocated" to something transient, then if the nodes find another IP and update with a new allocated address, the VTEP status would go back to "Allocated".

Features that depend on the VTEP to work (like EVPN) would be expected to have some downtime until the system reconciles the new IPs.

docs/okeps/okep-5088-evpn.md
trozet: I would avoid using a map. Just more json unmarshaling and patching/updating hell. Why do you want to use a map? Also, if we are introducing a new prefix, would it be better to use k8s.ovn-kubernetes.io?

jcaamano: I do agree....

docs/okeps/okep-5088-evpn.md
jcaamano: So how would the node infer which of those two is the VTEP IP? At the end of the day the node is going to have to make that same decision because it's going to find them set on interfaces of the node.

jcaamano: I guess one assumption we were running with is that that scenario would be considered an error.

I guess lets assume not. 

If it's just pick one and stick with, then yeah, we do need the annotation or something else to stick to it. But why do we need the node to make that decision. Seems like we could make it from anywhere once we are aware of the node IPs.

trozet: because the node actually knows the IPs on the host. Then you don't have levels of indirection via another annotation. So rather than deriving VTEP via:

node IP -> host cidrs -> cluster manager -> annotate node with vtep IP

its 

node IP -> annotate node with VTEP IP

VTEP does not rely on host cidrs, and its assignment can be done by each node rather than cluster manager having to do it. In managed mode it makes sense for cluster manager to be a central allocator. In unmanaged mode it does not.

jcaamano: OK let's say node did all that but still used the `host-cidrs` annotation to do it. So cluster manager is not involved. Just trying to find ways to reuse what we already have in place.

jcaamano: So node would match for every VTEP an IP from `host-cidrs` and annotate it as the vtep IP.

docs/okeps/okep-5088-evpn.md
l8huang: FYI: We need at least 15 in our environment for multi-vtep, it's better to have some extra slots for future growth.

docs/okeps/okep-5088-evpn.md
l8huang: 👀 For unmanaged multi-VTEP, if some restrictions need to be relaxed, I can cover them in the multi-VTEP OKEP.

docs/okeps/okep-5088-evpn.md
tssurya: > because the node actually knows the IPs on the host. Then you don't have levels of indirection via another annotation. So rather than deriving VTEP via:
> 
> node IP -> host cidrs -> cluster manager -> annotate node with vtep IP

I still don't get it - there is no last step here... we don't want to do a new annotation for the VTEP IP. there is no annotate the node with the VTEP IP. its really just the first 3 here for unmanaged mode.
For managed mode there is no change in design.

> 
> its
> 
> node IP -> annotate node with VTEP IP
> 
> VTEP does not rely on host cidrs, and its assignment can be done by each node rather than cluster manager having to do it. In managed mode it makes sense for cluster manager to be a central allocator. In unmanaged mode it does not.

I don't agree with this, I find it perfectly OK to reuse the host cidrs for VTEP as well since the mechanism of node discovering and knowing the nodeIPs already exists and the information is already readily available for consumption at the cluster manager side. What am I missing?

docs/okeps/okep-5088-evpn.md
tssurya: > I think for our use case we will need a lot. 

define "a lot", 
side-note: VTEP should have been its own OKEP as a standalone feature instead of coupling it with EVPN and adding all use cases here.

> We may have an unmanaged mode where there are many VTEP IPs allocated from /31 subnets. We want to re-use this VTEP CR for multi-vtep (rather than relying on OVS annotations). Can we please remove this limit or make it something much higher?

Removing the limit is not correct CEL wise - we will hit the kapi server cost calculation limits. We need to have a reasonable bounded max value. 

> FYI: We need at least 15 in our environment for multi-vtep, it's better to have some extra slots for future growth.

Sure, I can update this to 20 taking @l8huang 's  15 plus a few extra slots into consideration. Note that as I said before, its just a one liner change for either of you do in the future to make this another number when the future use case materializes. I don't think we need to be debating this immediately for a future use case since that can be expanded at any time. For now I'm gonna make this 20.

docs/okeps/okep-5088-evpn.md
tssurya: > Yeah that is the OKEP. I think regardless of whether or not that OKEP is merged, validation in the CEL rules is not the right thing. We should not tie VTEP to EVPN, that's the reason we made it a separate and generic CRD.

well it should have been a separate OKEP IMHO from what you are saying since its for more than just EVPN.  Wonder why it was added as part of EVPN OKEP. Having smaller standalone OKEP could have made it easier to get all the VTEP use cases rather than make this tied to EVPN OKEP.

> So I feel strongly validation for EVPN should be done in EVPN and not in VTEP CRD. If you add CEL for VTEP because of EVPN, then sometime later we remove CEL for VTEP 

Sure Tim. I'll do what you feel/want.

> and now we have to move it into EVPN validation, why didn't we just start with EVPN validation?

The original OKEP called nothing out wrt IPV6 or dualstack VTEPs not working in FRR - all the examples only ever showed IPV4 as well, this update was just making it tighter after having thought through it but anyways I don't think going back and forth on this point is worth my time. I'll do what you want, let's move on.

docs/okeps/okep-5088-evpn.md
tssurya: that's right. cross CIDR VTEP overlap across two different VTEP CRs. within the same VTEP we can use CEL to avoid duplications...I can update the description to call out "two different VTEP CRs"

docs/okeps/okep-5088-evpn.md
tssurya: > I don't think Kubernetes API implies SLA of downtime. 

This is not a kubernetes API, its about what ovn-kubernetes API SLA is.

> Let's take network policy for example. Configuring a new network policy has no guarantee about how long it takes to enforce, and NP status was even rejected upstream because this is a moving target.
> 
> Just because we allow a user to switch a VTEP on a node, doesn't imply there is no downtime. As a user I would expect if i edit a VTEP to remove a cidr and add a new one that affects one or more nodes, that the status of the CRD will change from "Allocated" to something transient, then if the nodes find another IP and update with a new allocated address, the VTEP status would go back to "Allocated".
> 
> Features that depend on the VTEP to work (like EVPN) would be expected to have some downtime until the system reconciles the new IPs.

I don't want to make assumptions for the users. regardless this is something if required @l8huang can do in his OKEP to support VTEP CIDR removal. Nothing proposed here is restricting future things, in fact the original OKEP mentioned no details around mutability and when we started implementing it we had open questions, we are just tightening it bit but none of this is set in stone - i.e loosening a VTEP CEL is easy and can be done in the future if there is an end user request for it. For EVPN specifically is there a need for NVIDIA to have this supported immediately in 1.3 release?

docs/okeps/okep-5088-evpn.md
tssurya: > events are fine with a generic failed status in the CR itself. I would have each node post their own event for their node.

worst case this is still 5000 events.
anyways updated this to include events. resolving the thread.

docs/okeps/okep-5088-evpn.md
tssurya: > I mean do not even try to detect this problem. We don't detect a user consuming IPs on other subnets they configure in OVNK. I know this has happened in the past and we find out from a bug report, but we do not proactively try to detect it in the system. I would take the same approach here.

the user experience has been bad around some of these things, I don't mind fixing it since using host-cidrs its pretty straight forward? its like how we do overlap detection for cross features... i.e if EgressIP CIDR and VTEP CIDR end up being same both are our features like? Same reason CNC connectSubnet has overlap check with podsubnets or serviceCIDR subnet

Having said that I don't have a strong opinion on this - if "do nothing, its fine" is how we want it - then that's fine by me.

docs/okeps/okep-5088-evpn.md
tssurya: Added 

```
The vtep controller will also add an internal validation to ensure only IPV4 CIDRs are accepted.
This restriction will be removed when FRR supports IPV6 VTEPs for EVPN.
```

removed the CEL

docs/okeps/okep-5088-evpn.md
tssurya: > Just want to have this comment visible since we are changing the original approach of one annotation per VTEP: we are not concerned about marshaling/unmarshaling big jsons?

I always assumed for there to be max 10 VTEPs on a cluster with each node doing 10 tunnels and not 100's of tunnels... not sure if I'm being naive.

wrt to 1 annotation per VTEP or 1 annotation for all VTEPs - what we proposed here is how we do it for other network annotations...


> Also if we take this approach I think it should be

> k8s.ovn.org/vteps: {"vtep-name1": { "cidrs" :["ipv4", "ipv6"]}, "vtep-name2": {"cidrs": ["ipv4", "ipv6"]}"

> just in case we want to annotate more things per vtep in the future

shouldn't it be relatively easy to transition annotations? i.e deprecate old one add new one? but I don't mind adding another layer of nesting if we wan to avoid that and guard ourselves . I'll wait a bit before updating this point again till we get consensus

docs/okeps/okep-5088-evpn.md
trozet: I'm not thinking about it in terms of timeline or EVPN vs Geneve VTEP features, just user experience. I can see a case where we limit the number of CIDRs we can add and a user fills it up, then realizes he doesn't need one that he configured by accident and wants to remove it and replace it with another. In unmanaged mode I still don't see why this is a problem. Perhaps we could just limit the CEL for managed mode only?

What is the impact on the implementation for EVPN? I guess you would see a vtep annotation change on the node, then would need to reconfigure FRR right?

@jcaamano do you have any opinions here?

docs/okeps/okep-5088-evpn.md
trozet: just to be clear...I'm also not strongly against a map if it makes more sense (like we are worried about many CRs being created which add tons of keys to the node). We just nee dto be smart about partial json unmarshal, and with #5832 we will only parse the annotations 1 time per event.

docs/okeps/okep-5088-evpn.md
jcaamano: I am ok removing CIDRs if there are no IPs allocated to them.

Since it looks like we are not going to safeguard unmanaged VTEPs in any way even if we could, we can do whatever I guess.

tssurya: > I'm not thinking about it in terms of timeline or EVPN vs Geneve VTEP features, just user experience. I can see a case where we limit the number of CIDRs we can add and a user fills it up, then realizes he doesn't need one that he configured by accident and wants to remove it and replace it with another. In unmanaged mode I still don't see why this is a problem. Perhaps we could just limit the CEL for managed mode only?

this is a valid use case - its something we also want for the future  (note that all permutations of mutability was already thought of and nothing is invalid as a use case for a user.) - my take is only this - let this come in @l8huang 's OKEP as a FUP.
Nothing I do/propose today is preventing us from adding this removal logic in the future 2 months from now. Loosening a CEL from its restriction is free of cost. I just want some time to think through this for unamanged mode and not rush into this.

tssurya: > What is the impact on the implementation for EVPN? I guess you would see a vtep annotation change on the node, then would need to reconfigure FRR right?

correct, if a CIDR got removed, we need to re-assess and remove IPs from the VTEP annotation - CM would need to go into false state if no IPs exist, FRR would need to be reconfigured to stop advertising those IPs.

docs/okeps/okep-5088-evpn.md
tssurya: fixing this format.

docs/okeps/okep-5088-evpn.md
tssurya: for future history: Since I am the only person advocating for going with `host-cidrs` + put the allocatedIPs on the status - I've moved my idea into the Alternatives section since I was minority. @trozet still wanted to keep the original proposed design with no changes so we are going for that.

docs/okeps/okep-5088-evpn.md
tssurya: actually you know what? this isn't worth a back and forth either.. if we want to do this now I'll just update the OKEP to make this CEL only for managed mode. 
I think that should solve @trozet's requirement

tssurya: I assume we also want to allow for CIDR mask to be shrinked right?

tssurya: done. I've reworded to say this is only for managed mode.

PR #6107, opened 2026-03-13
Force fake iptables/nftables helpers at start of unit tests

danwinship:
SetFakeIPTablesHelpers / SetFakeNFTablesHelper affect all tests that run after the function is called, not just the one that calls the function. Since we never want to use the real iptables/nftables from unit tests, call them at the start of testing to ensure that.

In particular, I think there are some tests that depend on the fact that a _different_ test will have called SetFakeIPTablesHelpers before they run. Or at least, I've seen this in some of the future PRs I'm working on.

## Summary by CodeRabbit

* **Tests**
  * Centralized fake network-helper initialization into core test suites to ensure consistent setup before running specs.
  * Removed redundant per-test fake helper calls across several tests, simplifying setup while preserving test behavior and outcomes.

### DIFF
--- a/go-controller/pkg/node/gateway_shared_intf_test.go
+++ b/go-controller/pkg/node/gateway_shared_intf_test.go
@@ -27,7 +27,6 @@ import (
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/factory"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/kube"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/networkmanager"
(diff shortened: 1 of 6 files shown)


danwinship:
@trozet, you hit a bug caused by this in #5832 (https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5832/changes/7a91af5d33494bc24bd3fbf068b8f280df05df42)

jcaamano:
@danwinship is that merge commit going to go away when we merge (via rebase) or is it something that you want to intentionally go with?

danwinship:
oh, I just used the "update branch" button on the PR... I assumed that since it was there that it meant I should use it?

oh, looks like there's a "rebase" version too... fixed

jcaamano (approved):

jcaamano (approved):

PR #6108, opened 2026-03-13
Port config.DisableForwarding to nftables

danwinship:
Four commits here:
- An adjustment to the unit tests, from #6037
- Another adjustment to the unit tests, from #6107 
- Removing the `-A INPUT -i ovn-k8s-mp0 -j ACCEPT` iptables rule, which doesn't work if your system firewall is using nftables anyway. (If your firewall is dropping ovn-k traffic, you need to fix your firewall.)
- The actual porting of the rules associated with `config.DisableForwarding`

In iptables, we have to share the `FORWARD` chain with other users, so it's hard to ensure that our `DROP` rule reliably ends up at the end of the chain, so instead the iptables code set the "default policy" of the chain to `DROP`. In nftables we don't have to worry about that; nobody else is going to be adding rules to our chain, so we can just have a catch-all `drop` rule at the end rather than needing to change the policy for for unmatched packets.

This leaves around code to clean up old iptables forwarding rules for now, so it should handle upgrades correctly (by deleting the iptables rules and replacing them with nftables ones). The restart case is a little simpler on nftables than on iptables because we just always rewrite the entire `gateway-forward` chain on startup, so we don't need any extra code in `deleteMasqueradeResources` to handle day 2 config changes.

## Summary by CodeRabbit

* **Documentation**
  * Updated forwarding docs to reference nftables (forward hook) instead of iptables and clarified default drop/forwarding behavior for IPv4/IPv6.

* **Refactor**
  * Switched gateway and forwarding management to nftables, centralizing rule setup for local-gateway and external-bridge flows, adding dual‑stack forwarding chains, and removing legacy iptables init-style logic in favor of nftables-first handling with best-effort iptables cleanup.

* **Tests**
  * Updated tests and fixtures to expect nftables behavior and initialize nftables helpers; removed iptables-only assertions.

### DIFF
--- a/docs/getting-started/configuration.md
+++ b/docs/getting-started/configuration.md
@@ -12,18 +12,18 @@ OVN-Kubernetes allows to enable or disable IP forwarding for all traffic on OVN-
 By default forwarding is enabled and this allows host to forward traffic across OVN-Kubernetes managed interfaces.
 If forwarding is disabled then Kubernetes related traffic is still forwarded appropriately, but other IP traffic will not be routed by cluster nodes.
 
-IP forwarding is implemented at cluster node level by modifying both iptables `FORWARD` chain and IP forwarding `sysctl` parameters. 
+IP forwarding is implemented at cluster node level by adding an nftables `forward` chain and modifying IP forwarding `sysctl` parameters.
 
 #### IPv4
 
 If forwarding is enabled(default) and it is desired to allow forwarding for traffic on unmanaged ovn-kubernetes interfaces, then system administrators need to set the following sysctl parameters on the desired interfaces or globally.
 OVN-Kubernetes already sets sysctl forwarding for interfaces it manages, such as the ovn-k8s-mp0 interface and the shared gateway bridge interface. An operator can be built to manage forwarding sysctl parameters based on forwarding mode.
-No extra iptables rules are added by OVN-Kubernetes to FORWARD chain while using this IP forwarding mode.
+No extra nftables rules are added by OVN-Kubernetes to the `forward` hook while using this IP forwarding mode.
 
 ```
 net.ipv4.ip_forward=1
 ```
-IP forwarding can be disabled either by setting `disable-forwarding` command line option to `true` while starting ovnkube or by setting `disable-forwarding` to `true` in config file. If forwarding is disabled the default policy for the [FORWARD iptables chain is set as DROP](#forwarding-rules) and system administrators can add use-case specific ACCEPT rules.
+IP forwarding can be disabled either by setting `disable-forwarding` command line option to `true` while starting ovnkube or by setting `disable-forwarding` to `true` in config file. If forwarding is disabled a [`forward`-hook chain with a default `drop` rule is created](#forwarding-rules) and system administrators can add use-case specific iptables/nftables rules to accept packets.
 
 When IP forwarding is disabled, following sysctl parameters are modified by OVN-Kubernetes to allow forwarding Kubernetes related traffic on OVN-Kubernetes managed bridge interfaces and management port interface.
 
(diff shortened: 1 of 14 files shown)


go-controller/pkg/node/gateway_nftables.go
```diff
@@ -523,3 +524,186 @@ func delLocalGatewayPodSubnetNFTRules() error {
...
+	})
+
+	err = nft.Run(context.TODO(), tx)
+	if err != nil {
+		return fmt.Errorf("could not configure bridge forwarding: %w", err)
+	}
+
+	// If there are legacy IPTables rules left around, clean them up, ignoring errors.
+	_ = delStaleMasqueradeIPTRules(config.Gateway.MasqueradeIPs.V4OVNMasqueradeIP)
+	_ = delStaleMasqueradeIPTRules(config.Gateway.MasqueradeIPs.V6OVNMasqueradeIP)
```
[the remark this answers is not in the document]
danwinship: those are cleaned up from initLocalGatewayRules

docs/getting-started/configuration.md
```diff
@@ -12,18 +12,18 @@ OVN-Kubernetes allows to enable or disable IP forwarding for all traffic on OVN-
...
-No extra iptables rules are added by OVN-Kubernetes to FORWARD chain while using this IP forwarding mode.
+No extra nftables rules are added by OVN-Kubernetes to the `forward` hook while using this IP forwarding mode.
 
 ```
 net.ipv4.ip_forward=1
 ```
-IP forwarding can be disabled either by setting `disable-forwarding` command line option to `true` while starting ovnkube or by setting `disable-forwarding` to `true` in config file. If forwarding is disabled the default policy for the [FORWARD iptables chain is set as DROP](#forwarding-rules) and system administrators can add use-case specific ACCEPT rules.
+IP forwarding can be disabled either by setting `disable-forwarding` command line option to `true` while starting ovnkube or by setting `disable-forwarding` to `true` in config file. If forwarding is disabled a [`forward`-hook chain with a default `drop` rule is created](#forwarding-rules) and system administrators can add use-case specific iptables/nftables rules to accept packets.
```
bpickard22, line 26: should we say nftables here since we are fully removing iptables eventually, or will we do that when we are done?

bpickard22, line 26: should we say nftables here since we are fully removing iptables eventually, or will we do that when we are done?

docs/getting-started/configuration.md
```diff
@@ -87,41 +87,16 @@ It is not possible to configure the IPv6 forwarding per interface by setting the
 `net.ipv6.conf.IFNAME.forwarding` sysctl (it just configures the interface-specific Host/Router 
 behaviour). \
 Instead, the opposite approach is required where the global forwarding is always enabled and 
-the traffic is restricted through iptables.
+the traffic is restricted through iptables/nftables.
```
bpickard22, line 90: same here

go-controller/pkg/node/gateway_iptables.go
```diff
@@ -312,6 +312,11 @@ func getStaleMasqueradeIptablesRules(masqueradeIP net.IP) []nodeipt.Rule {
 		getMasqueradeIpTablesNATRules(masqueradeIP, getIPTablesProtocol(masqueradeIP.String()))...)
 }
 
+// delStaleMasqueradeIPTRules deletes all iptables rules may have been added for a given masquerade IP.
```
bpickard22: *nit that may have been

docs/getting-started/configuration.md
danwinship: This line isn't talking about _our_ rules, it's talking about what sysadmins can do.

danwinship:
OK, I think the fix here is:
- Merge #6512 so that we only create iptables rules on kernels that don't have the IPv6 `force_forwarding` sysctl
- Ensure that RHEL 10 gets the `force_forwarding` sysctl backport
- Leave the fallback FORWARD policy stuff using iptables instead of nftables, to better preserve compatibility with other components. Since RHEL 10 won't have to use the fallback path, the fact that it requires iptables doesn't matter. (More generally, ovn-kubernetes would require that all nodes _either_ have a kernel with IPv6 `force_forwarding` _or_ they have functioning iptables.)

PR #6114, opened 2026-03-16
Trim more information from pod informer cache

trozet:
There's more things we can transform and trim off of pod objects to reduce informer cache size and memory usage. Also changes how we update pod annotations to not try to update status fields and only update the annotations.

## Summary by CodeRabbit

* **Chores**
  * Reduced cache memory by trimming additional pod runtime, init/ephemeral containers, probes, security, and status fields.
  * Improved retry logic to treat more annotation-update errors as retryable.
* **Bug Fixes**
  * Switched status updates to targeted pod annotation patches with safer guards to reduce conflicts and stale-writer issues.
* **Tests**
  * Added and updated tests for annotation-patching, retry/rollback behavior, and related interactions.

### DIFF
--- a/go-controller/pkg/kube/kube.go
+++ b/go-controller/pkg/kube/kube.go
@@ -3,6 +3,8 @@ package kube
 import (
 	"context"
 	"encoding/json"
+	"fmt"
+	"strings"
 
 	ipamclaimsapi "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1"
 	ipamclaimssclientset "github.com/k8snetworkplumbingwg/ipamclaims/pkg/crd/ipamclaims/v1alpha1/apis/clientset/versioned"
@@ -58,7 +60,7 @@ type Interface interface {
 	SetLabelsOnNode(nodeName string, labels map[string]interface{}) error
 	PatchNode(old, new *corev1.Node) error
 	UpdateNodeStatus(node *corev1.Node) error
-	UpdatePodStatus(pod *corev1.Pod) error
+	PatchPodStatusAnnotations(oldPod, newPod *corev1.Pod) error
 	// GetPodsForDBChecker should only be used by legacy DB checker. Use watchFactory instead to get pods.
 	GetPodsForDBChecker(namespace string, opts metav1.ListOptions) ([]*corev1.Pod, error)
 	// GetNodeForWindows should only be used for windows hybrid overlay binary and never in linux code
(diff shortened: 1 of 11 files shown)


go-controller/pkg/util/pod.go
```diff
@@ -43,13 +43,13 @@ func UpdatePodWithRetryOrRollback(podLister listers.PodLister, kube kube.Interfa
...
 
-		updated = true
-		// It is possible to update the pod annotations using status subresource
-		// because changes to metadata via status subresource are not restricted pods.
-		err = kube.UpdatePodStatus(pod)
+		err = kube.PatchPodStatusAnnotations(oldPod, pod)
 		if err != nil && rollback != nil {
 			rollback()
 		}
+		if err == nil {
```
kyrtapz: I know that this isn't directly related to you change but do we need that `updated` log at Info level? We are already logging in `PatchPodStatusAnnotations`?
I am ok with dropping it or moving to V5.
I don't know if we should only log it on success though, since it logs the time the function took.

go-controller/pkg/kube/kube.go
```diff
@@ -116,6 +119,133 @@ func (k *Kube) SetAnnotationsOnPod(namespace, podName string, annotations map[st
 	return err
 }
 
+type jsonPatchOp struct {
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value,omitempty"`
```
kyrtapz, line 124: Why `omitempty`? what would happen if we try to set an empty value?

go-controller/pkg/util/pod.go
```diff
@@ -43,13 +43,13 @@ func UpdatePodWithRetryOrRollback(podLister listers.PodLister, kube kube.Interfa
 			return nil
 		}
 
-		updated = true
-		// It is possible to update the pod annotations using status subresource
-		// because changes to metadata via status subresource are not restricted pods.
-		err = kube.UpdatePodStatus(pod)
+		err = kube.PatchPodStatusAnnotations(oldPod, pod)
```
kyrtapz, line 57: `RetryOnConflict` doesn't work when `PatchPodStatusAnnotations` fails the resourceVersion check.
Looking at the logs we get a ton of errors like these:
```
ovn-control-plane/pods/ovn-kubernetes_ovnkube-master-748955c9cb-z86kb_f6ce3425-abac-422d-81d5-f1cac824ace9/ovnkube-master/0.log:24652:2026-03-17T17:32:59.11485148Z stderr F E0317 17:32:59.114816      86 obj_retry.go:717] default/networkController: failed to update *v1.Pod, old=kv-live-migration-1520/virt-launcher-worker-q6qmg-5pdtj, new=kv-live-migration-1520/virt-launcher-worker-q6qmg-5pdtj, error: addLogicalPort failed for kv-live-migration-1520/virt-launcher-worker-q6qmg-5pdtj: failed to update pod kv-live-migration-1520/virt-launcher-worker-q6qmg-5pdtj: the server rejected our request due to an error in our request
```
github.com/ovn-org/ovn-kubernetes/blob/30f8075124249785d110d88687e47db15b6c0a00/go-controller/vendor/k8s.io/apimachinery/pkg/api/errors/errors.go#L475-L477

I am a little surprised that this didn't trip our CI but maybe we are just that good at eventual consistency :)

go-controller/pkg/kube/kube.go
```diff
@@ -116,6 +119,133 @@ func (k *Kube) SetAnnotationsOnPod(namespace, podName string, annotations map[st
...
+	}
+	if len(changedKeys) == 0 {
+		return nil
+	}
+
+	keys := make([]string, 0, len(changedKeys))
+	for key := range changedKeys {
+		keys = append(keys, key)
+	}
+	sort.Strings(keys)
```
kyrtapz: What's the reason for sorting the keys?

go-controller/pkg/kube/kube.go
trozet: omitempty is used when there is a remove operation. Aka we clear an annotation.

If we ever tried to add/replace/test a "nil" value then it would not work, but lucky for us the jsonPatchOp is used by PatchPodStatusAnnotations, and annotations are a map[string]string. So the value type is non-nil. Therefore this scenario can never happen. Setting any empty string value of "" will still work.

go-controller/pkg/kube/kube.go
trozet: i dont think there is a reason other than it creates consistent operation order in case we were debugging and looking at logs. I'll remove it.

go-controller/pkg/util/pod.go
trozet: wow nice catch! I think it is not just resourceVersion, its any failing json patch. Will fix.

go-controller/pkg/util/pod.go
trozet: I'll change it to V5 and always print.

ormergi:
Very nice, this is an issue I was looking at while ago, this change seem to mitigate the amount of conflicts but there more to it: as long there are multiple controllers (default network and udn contollers) update the same annotation (pod-network) these conflicts are inevitable.

I was wondering while back if we could split the pod-network to two (default and udn) to reconcile default and udn controllers and avoid mode conflicts.
Just a thought.

trozet:
> Very nice, this is an issue I was looking at while ago, this change seem to mitigate the amount of conflicts but there more to it: as long there are multiple controllers (default network and udn contollers) update the same annotation (pod-network) these conflicts are inevitable.
> 
> I was wondering while back if we could split the pod-network to two (default and udn) to reconcile default and udn controllers and avoid mode conflicts. Just a thought.

While we could split it so that there are never conflicts, I think the CNI still needs both pieces of information and will wait. I'm not sure if entirely avoiding the conflicts will provide significant savings, but if I see during profiling after I knock out other perf/scale issues then will address it.

PR #6116, opened 2026-03-17
Support nodeSelector in AddressSetManager

jxiaobin:
Enable address set to select pod IPs by nodeSelector

## Summary by CodeRabbit

* **New Features**
  * Address sets now support node selectors so policies can target workloads on specific nodes.
  * Address-set identities incorporate node selection, keeping sets distinct when node criteria differ.
  * Address sets are node-aware: node label changes trigger reconciliation and updates.

* **Bug Fixes**
  * Improved handling so pods are correctly added/removed from address sets when node membership changes.

* **Tests**
  * Expanded node-selector test coverage and added node-aware test setup.

### DIFF
--- a/go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
+++ b/go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -38,12 +39,17 @@ type podSelectorAddressSet struct {
 	namespaceSelector labels.Selector
 	// namespace is used when namespaceSelector is nil to set static namespace
 	namespace string
+	// nodeSelector decides which nodes' pods should be added to the address set; nil means all nodes
+	nodeSelector labels.Selector
 
 	addressSet addressset.AddressSet
 
 	// selectedNamespaces is a cache for namespaces that were selected by this address set during the last reconciliation
 	// used to optimize events processing.
 	selectedNamespaces sets.Set[string]
+	// selectedNodes is a cache for nodes that were selected by this address set during the last reconciliation
+	// used to optimize node event processing. Only set when nodeSelector is non-nil.
+	selectedNodes sets.Set[string]
 
 	// network-specific fields
 	controllerName string
@@ -67,9 +73,11 @@ type AddressSetManager struct {
 
 	podLister       listers.PodLister
 	namespaceLister listers.NamespaceLister
+	nodeLister      listers.NodeLister
 
 	podController        controller.Controller
 	nsController         controller.Controller
+	nodeController       controller.Controller
 	addressSetReconciler controller.Reconciler
 
 	// All network controllers are getting this function from the same networkmanager, so we can share it
@@ -77,7 +85,7 @@ type AddressSetManager struct {
 }
 
 func NewAddressSetManager(podInformer coreinformers.PodInformer, namespaceInformer coreinformers.NamespaceInformer,
-	nbClient libovsdbclient.Client, getNetworkNameForNADKey func(nadKey string) string) *AddressSetManager {
+	nodeInformer coreinformers.NodeInformer, nbClient libovsdbclient.Client, getNetworkNameForNADKey func(nadKey string) string) *AddressSetManager {
 	m := &AddressSetManager{
 		name:                       "pod-selector-address-set-manager",
 		nbClient:                   nbClient,
@@ -87,6 +95,7 @@ func NewAddressSetManager(podInformer coreinformers.PodInformer, namespaceInform
 		addressSets:                syncmap.NewSyncMap[*podSelectorAddressSet](),
 		podLister:                  podInformer.Lister(),
 		namespaceLister:            namespaceInformer.Lister(),
+		nodeLister:                 nodeInformer.Lister(),
 		getNetworkNameForNADKey:    getNetworkNameForNADKey,
 	}
 	podCfg := &controller.ControllerConfig[corev1.Pod]{
@@ -111,7 +120,18 @@ func NewAddressSetManager(podInformer coreinformers.PodInformer, namespaceInform
 	}
 	m.nsController = controller.NewController[corev1.Namespace](m.name+"-namespace", nsCfg)
 
-	// addressSetReconciler is fed from the pod and namespace controllers
+	nodeCfg := &controller.ControllerConfig[corev1.Node]{
+		RateLimiter:    workqueue.DefaultTypedControllerRateLimiter[string](),
+		Reconcile:      m.reconcileNode,
+		ObjNeedsUpdate: m.nodeNeedUpdate,
+		MaxAttempts:    controller.InfiniteAttempts,
+		Threadiness:    1,
+		Informer:       nodeInformer.Informer(),
+		Lister:         nodeInformer.Lister().List,
+	}
+	m.nodeController = controller.NewController[corev1.Node](m.name+"-node", nodeCfg)
+
+	// addressSetReconciler is fed from the pod, namespace and node controllers
 	m.addressSetReconciler = controller.NewReconciler(
 		m.name+"-addrset",
 		&controller.ReconcilerConfig{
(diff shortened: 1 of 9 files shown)


go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -392,6 +485,23 @@ func (m *AddressSetManager) reconcileAddressSet(key string) error {
...
+			for _, pod := range pods {
+				if pod.Spec.NodeName != "" && selectedNodes.Has(pod.Spec.NodeName) {
+					filtered = append(filtered, pod)
+				}
+			}
+			pods = filtered
+			psAddrSet.selectedNodes = selectedNodes
+		} else {
+			psAddrSet.selectedNodes = nil
+		}
```
npinaeva, line 532: I don't think this part is doing anything: `psAddrSet.nodeSelector` doesn't change after creation, so if it was nil or empty, then `selectedNodes` should just never be set and will stay nil forever?

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -346,6 +390,55 @@ func (m *AddressSetManager) reconcileNamespace(nsKey string) error {
...
+				return err
+			}
+			if currentlyMatchedNodes.Has(nodeKey) {
+				// this node is relevant for this address set, reconcile
+				m.addressSetReconciler.Reconcile(addrSetKey)
+				return nil
+			}
+			// reconcile the address set if the node matches the previous selected nodes
+			previouslyMatchedNodes := addrSet.selectedNodes
+			if previouslyMatchedNodes == nil || previouslyMatchedNodes.Has(nodeKey) {
```
npinaeva, line 459: let's add a comment that `previouslyMatchedNodes == nil` can only happen if this address set has never been reconciled yet

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -66,17 +71,20 @@ type AddressSetManager struct {
...
 
 	podController        controller.Controller
 	nsController         controller.Controller
+	nodeController       controller.Controller
 	addressSetReconciler controller.Reconciler
 
 	// All network controllers are getting this function from the same networkmanager, so we can share it
 	getNetworkNameForNADKey func(nadKey string) string
+	podNodeMap              map[string]string
 }
```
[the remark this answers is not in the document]
npinaeva: that is true ^, and also means we need some more tests with nodeSelector for pod updates

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -175,12 +197,19 @@ func (m *AddressSetManager) EnsureAddressSet(podSelector, namespaceSelector *met
...
 	}
-	addrSetKey = getInternalKey(podSelector, namespaceSelector, namespace, controllerName)
+	if nodeSelector != nil {
+		nodeSel, err = metav1.LabelSelectorAsSelector(nodeSelector)
+		if err != nil {
+			err = fmt.Errorf("can't parse node selector %v: %w", nodeSelector, err)
+			return
+		}
+	}
+	addrSetKey = getInternalKey(podSelector, namespaceSelector, nodeSelector, namespace, controllerName)
```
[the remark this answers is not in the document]
npinaeva: also true, I suggest that in case of empty nodeSelector we just set it internally as nil because it is equivalent and will cause reuse of the same address set

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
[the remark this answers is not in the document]
jxiaobin: removed the map, instead get the node name from Pod by querying informer cache

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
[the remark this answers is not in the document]
jxiaobin: fixed

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
jxiaobin: removed else block.

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
jxiaobin: added comment

go-controller/pkg/ovn/addresssetmanager/addressset_manager.go
```diff
@@ -267,10 +298,14 @@ func (m *AddressSetManager) podNeedUpdate(old, new *corev1.Pod) bool {
...
 func (m *AddressSetManager) reconcilePod(podKey string) error {
-	namespace, _, err := cache.SplitMetaNamespaceKey(podKey)
+	namespace, name, err := cache.SplitMetaNamespaceKey(podKey)
 	if err != nil {
 		return fmt.Errorf("failed to split meta namespace key %q: %v", podKey, err)
 	}
+	pod, err := m.podLister.Pods(namespace).Get(name)
+	if err != nil {
+		return fmt.Errorf("failed to get pod %s in namespace %s: %v", name, namespace, err)
+	}
```
npinaeva: I suggest an optimization since we don't use nodeSelector u/s yet: we could only fetch pod when needed, i.e. under `if addrSet.nodeSelector != nil {` and then check if it was fetched before to only call lister once.
Current version also doesn't work on pod delete, because pod won't exist anymore, so I think `if addrSet.nodeSelector != nil {` and pod is nil, we just need to always reconcile to avoid extra caching.
Also please add more unit tests for node selector as discussed https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6116#discussion_r2946386294 to avoid such problems being uncaught

jxiaobin: fixed this, will add more test cases.

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
```diff
@@ -105,14 +105,42 @@ var _ = ginkgo.Describe("OVN podSelectorAddressSet", func() {
...
+				&corev1.NamespaceList{Items: namespaces},
+				&corev1.PodList{Items: pods},
+				&corev1.NodeList{Items: nodes},
+			).GetOVNKubeControllerClientset()
+		} else {
+			clientSet = util.GetOVNClientset(
+				&corev1.NamespaceList{Items: namespaces},
+				&corev1.PodList{Items: pods},
+			).GetOVNKubeControllerClientset()
+		}
```
npinaeva: I don't think that part is needed, you can just start nil to the node list items. Also could you update `startAddrSetManager` like this
```
startAddrSetManager := func(dbSetup libovsdbtest.TestSetup, namespaces []corev1.Namespace, pods []corev1.Pod) {
		startAddrSetManagerWithNodes(dbSetup, namespaces, pods, nil)
	}
```
to avoid code duplication?

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
```diff
@@ -402,6 +430,185 @@ var _ = ginkgo.Describe("OVN podSelectorAddressSet", func() {
...
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		dbIDsWorker := GetPodSelectorAddrSetDbIDs(podSelector, nil, nodeSelWorker, namespace1.Name, controllerName)
+		dbIDsControlPlane := GetPodSelectorAddrSetDbIDs(podSelector, nil, nodeSelControlPlane, namespace1.Name, controllerName)
+		asf.EventuallyExpectAddressSetWithAddresses(dbIDsWorker, []string{ip1})
+		asf.EventuallyExpectAddressSetWithAddresses(dbIDsControlPlane, []string{ip2})
+		asf.ExpectNumberOfAddressSets(2)
+	})
+
+	ginkgo.When("node selector is set", func() {
```
npinaeva, line 426: does it make sense to put all the new tests under this context?

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
```diff
@@ -402,6 +430,185 @@ var _ = ginkgo.Describe("OVN podSelectorAddressSet", func() {
...
+		ginkgo.BeforeEach(func() {
+			startAddrSetManagerWithNodes(initialDB, []corev1.Namespace{namespace1}, []corev1.Pod{*pod1}, []corev1.Node{node1})
+		})
+
+		ginkgo.It("doesn't add pod IP to address set if its labels don't match pod label selector", func() {
+			_, _, _, err := addressSetManager.EnsureAddressSet(podSelector, nil, nodeSelector, namespace1.Name, "backRef", controllerName, &util.DefaultNetInfo{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			asf.EventuallyExpectEmptyAddressSetExist(peerASIDs)
+		})
+		ginkgo.It("adds pod IP to address set if its labels match pod label selector", func() {
```
npinaeva, line 492: Is it different from `ginkgo.It("adds selected pod IPs with nodeSelector: only pods on matching nodes are in the address set", func()`?

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
```diff
@@ -402,6 +430,185 @@ var _ = ginkgo.Describe("OVN podSelectorAddressSet", func() {
...
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			_, _, _, err = addressSetManager.EnsureAddressSet(podSelector, nil, nodeSelector, namespace1.Name, "backRef", controllerName, &util.DefaultNetInfo{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			asf.EventuallyExpectAddressSetWithAddresses(peerASIDs, []string{ip1})
+			// delete the pod
+			err = clientSet.KubeClient.CoreV1().Pods(namespace1.Name).Delete(context.TODO(), pod1.Name, metav1.DeleteOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			asf.EventuallyExpectEmptyAddressSetExist(peerASIDs)
+		})
+		ginkgo.It("deletes pod IP from address set when node label changes", func() {
```
npinaeva, line 538: I think it would be nice to have the same test for add: so pod is created without the matching label, then the label gets updated and we want to check that the pod IP is added, wdyt?

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
```diff
@@ -402,6 +430,185 @@ var _ = ginkgo.Describe("OVN podSelectorAddressSet", func() {
...
+		peerASIDs := GetPodSelectorAddrSetDbIDs(podSelector, nil, nodeSelector, namespace1.Name, controllerName)
+
+		ginkgo.BeforeEach(func() {
+			startAddrSetManagerWithNodes(initialDB, []corev1.Namespace{namespace1}, []corev1.Pod{*pod1}, []corev1.Node{node1})
+		})
+
+		ginkgo.It("doesn't add pod IP to address set if its labels don't match pod label selector", func() {
+			_, _, _, err := addressSetManager.EnsureAddressSet(podSelector, nil, nodeSelector, namespace1.Name, "backRef", controllerName, &util.DefaultNetInfo{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			asf.EventuallyExpectEmptyAddressSetExist(peerASIDs)
```
npinaeva: I think this one has to be replaced with consistently, because an empty address set will be created before any pod matching is done anyway, so it won't check what we actually want, something like
```
gomega.Consistently(func(g gomega.Gomega) {
				asf.ExpectEmptyAddressSet(peerASIDs)
			}).WithTimeout(500 * time.Millisecond).Should(gomega.Succeed())
```

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
jxiaobin: good catch, fixed.

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
jxiaobin: yeah, move all the new tests to this context.

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
jxiaobin: removed this one, also "node label change triggers address set reconciliation", which is covered by "deletes pod IP from address set when node label changes".

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
jxiaobin: agree, added a test case.

go-controller/pkg/ovn/addresssetmanager/addressset_manager_test.go
jxiaobin: fixed.

npinaeva:
hit https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5952
reported https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6239

npinaeva (approved):

npinaeva:
I am afraid it needs a rebase now, I had to merge a bug fix https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6198

jxiaobin:
rebased.

npinaeva:
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6140
https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5456
last one failed to create a cluster

PR #6118, opened 2026-03-17
  kubevirt: Support handling vm hostname field

### LINKED ISSUE #5520: Kubevirt Virtual Machines with kubevirt VM using layer2 UDPN [BeforeEach] should configure IPv4 and IPv6 using DHCP and NDP
### Which jobs are flaking?

https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/17220865197/job/48885581709?pr=5392

### Which tests are flaking?

Kubevirt Virtual Machines with kubevirt VM using layer2 UDPN [BeforeEach] should configure IPv4 and IPv6 using DHCP and NDP

today

```
Kubevirt Virtual Machines with kubevirt VM using layer2 UDPN [BeforeEach] should configure IPv4 and IPv6 using DHCP and NDP [Feature:VirtualMachineSupport]
  [BeforeEach] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:2124
  [It] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:2168

  [FAILED] Timed out after 60.001s.
  The function passed to Eventually failed at /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:2143 with:
  Unexpected error:
      <*errors.errorString | 0xc00197d210>: 
      want pod 'virt-launcher-vm1' on 'ovn-worker' to be 'Running' but was 'Pending'
      {
          s: "want pod 'virt-launcher-vm1' on 'ovn-worker' to be 'Running' but was 'Pending'",
      }
  occurred
  In [BeforeEach] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/kubevirt.go:1241 @ 08/26/25 08:40:54.708
```

_No response_

qinqon:
## 📑 Description
  When a VM's spec.template.spec.hostname is set, KubeVirt copies it into the vm.kubevirt.io/name label on virt-launcher pods instead of the actual VM name. OVN-K used this label to find pods belonging to a VM
  (e.g. during live migration, or to tag OVN topology elements for cleanup). Since multiple VMs can share the same hostname, the label can no longer uniquely identify a VM's pods — with 3+ VMs sharing a         
  hostname, the 3rd VM fails with "unexpected live migration state at pods".
                                                                                                                                                                                                                   
  This change switches to the kubevirt.io/domain annotation, which always reflects the real VM name. Since annotations can't be used in label selectors, we optimize: if the domain name matches the               
  vm.kubevirt.io/name label (the common case), we use the label for lookup; otherwise we fall back to iterating virt-launcher pods in the namespace and matching by annotation.
                                                                                                                                                                                                                   
  Note: KubeVirt >= 1.7 introduces a vm.kubevirt.io/id label that reliably identifies the VM, but it is not added retroactively to pods created with older versions. If KubeVirt adds a mutating webhook for that  
  in the future, we can use it as a further optimization.


### KubeVirt pod identifiers

| Identifier | Type | Source | Truncation | Uniqueness |
|---|---|---|---|---|
| `kubevirt.io/domain` | Annotation | VMI object name | None (annotations have no length limit) | Unique per VMI |
| `vm.kubevirt.io/name` (**deprecated**) | Label | VM hostname (can be overridden via `spec.template.spec.hostname`) | Truncates to 63 chars, no hash | **Not unique** — different VMs with the same hostname or names sharing the first 63 chars get the same value |
| `vmi.kubevirt.io/id` | Label | VMI name | Truncates + appends sha1 hash for names > 63 chars | Unique (hash preserves uniqueness) |

Fixes https://redhat.atlassian.net/browse/CNV-79752

### References

- KubeVirt PR introducing `vmi.kubevirt.io/id` and deprecating `vm.kubevirt.io/name`: https://github.com/kubevirt/kubevirt/pull/15783

Fixes https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5520

Kubevirt issue https://github.com/kubevirt/kubevirt/issues/17869

## ✅ Checks
- [x] My code requires tests

## How to verify it
The PR includes unit tests and e2e

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Added support for multiple virtual machines sharing the same hostname with distinct network configurations and MAC addresses.

* **Tests**
  * Added end-to-end test validating multiple virtual machines with identical hostnames and distinct static IP address assignments.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/go-controller/pkg/allocator/pod/macs.go
+++ b/go-controller/pkg/allocator/pod/macs.go
@@ -8,8 +8,6 @@ import (
 	"fmt"
 	"net"
 
-	kubevirtv1 "kubevirt.io/api/core/v1"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog/v2"
@@ -24,9 +22,9 @@ import (
 // macOwner compose the owner identifier reserved for MAC addresses management.
 // Returns "<ns>/<pod-name>" for regular pods and "<ns>/<vm-name>" for VMs.
 func macOwner(pod *corev1.Pod) string {
-	// Check if this is a VM pod and persistent IPs are enabled
-	if vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]; ok {
-		return fmt.Sprintf("%s/%s", pod.Namespace, vmName)
+	vmDescription, err := kubevirt.NewVMDescriptionFromPod(pod)
+	if err == nil && vmDescription != nil {
+		return vmDescription.Key().String()
 	}
 
 	// Default to pod-based identifier
@@ -43,10 +41,10 @@ func (allocator *PodAnnotationAllocator) ReleasePodReservedMacAddress(pod *corev
 	}
 
 	macOwnerID := macOwner(pod)
-	if vmKey := kubevirt.ExtractVMNameFromPod(pod); vmKey != nil {
+	if kubevirt.IsPodOwnedByVirtualMachine(pod) {
 		allVMPodsCompleted, err := kubevirt.AllVMPodsAreCompleted(allocator.podLister, pod)
 		if err != nil {
-			return fmt.Errorf("failed checking all VM %q pods are completed: %v", vmKey, err)
+			return fmt.Errorf("failed checking all VM %q pods are completed: %v", macOwnerID, err)
 		}
 		if !allVMPodsCompleted {
 			klog.V(5).Infof(`Retaining MAC address %q for owner %q on network %q because its in use by another VM pod`,
--- a/go-controller/pkg/clustermanager/pod/allocator_test.go
+++ b/go-controller/pkg/clustermanager/pod/allocator_test.go
@@ -20,6 +20,7 @@ import (
 	nadapi "github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1"
 	"github.com/onsi/gomega"
 	"github.com/stretchr/testify/mock"
+	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -50,17 +51,21 @@ type testPod struct {
 	completed   bool
 	network     *nadapi.NetworkSelectionElement
 	labels      map[string]string
+	annotations map[string]string
 }
 
 func (p testPod) getPod(t *testing.T) *corev1.Pod {
 	t.Helper()
+	if p.annotations == nil {
+		p.annotations = map[string]string{}
+	}
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        "pod",
 			UID:         apitypes.UID("pod"),
 			Namespace:   "namespace",
-			Annotations: map[string]string{},
 			Labels:      p.labels,
+			Annotations: p.annotations,
 		},
 		Spec: corev1.PodSpec{
 			HostNetwork: p.hostNetwork,
@@ -674,7 +679,13 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 					scheduled: true,
 					completed: true,
 					network:   &nadapi.NetworkSelectionElement{Namespace: "namespace", Name: "nad"},
-					labels:    map[string]string{"vm.kubevirt.io/name": "myvm"},
+					labels: map[string]string{
+						kubevirtv1.VirtualMachineNameLabel: "myvm",
+						kubevirtv1.AppLabel:                "virt-launcher",
+					},
+					annotations: map[string]string{
+						kubevirtv1.DomainAnnotation: "myvm",
+					},
 				},
 			},
 			expectMACRelease: &net.HardwareAddr{0x0a, 0x0a, 0x0a, 0x0a, 0x0a, 0x0a},
@@ -713,7 +724,13 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 					scheduled: true,
 					completed: true,
 					network:   &nadapi.NetworkSelectionElement{Namespace: "namespace", Name: "nad"},
-					labels:    map[string]string{"vm.kubevirt.io/name": ""},
+					labels: map[string]string{
+						kubevirtv1.VirtualMachineNameLabel: "myvm",
+						kubevirtv1.AppLabel:                "virt-launcher",
+					},
+					annotations: map[string]string{
+						kubevirtv1.DomainAnnotation: "myvm",
+					},
 				},
 			},
 			newPodCopyRunning: true,
@@ -733,7 +750,13 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 					scheduled: true,
 					completed: true,
 					network:   &nadapi.NetworkSelectionElement{Namespace: "namespace", Name: "nad"},
-					labels:    map[string]string{"vm.kubevirt.io/name": "myvm"},
+					labels: map[string]string{
+						kubevirtv1.VirtualMachineNameLabel: "myvm",
+						kubevirtv1.AppLabel:                "virt-launcher",
+					},
+					annotations: map[string]string{
+						kubevirtv1.DomainAnnotation: "myvm",
+					},
 				},
 			},
 			expectError:     `failed to release pod "namespace/pod" mac "0a:0a:0a:0a:0a:0a": failed checking all VM "namespace/myvm" pods are completed: failed finding related pods for pod namespace/pod when checking if they are completed: test error`,
@@ -797,7 +820,13 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 				new: &testPod{
 					network:   &nadapi.NetworkSelectionElement{Namespace: "namespace", Name: "nad"},
 					scheduled: true,
-					labels:    map[string]string{"vm.kubevirt.io/name": "myvm"},
+					labels: map[string]string{
+						kubevirtv1.VirtualMachineNameLabel: "myvm",
+						kubevirtv1.AppLabel:                "virt-launcher",
+					},
+					annotations: map[string]string{
+						kubevirtv1.DomainAnnotation: "myvm",
+					},
 				},
 			},
 			expectAllocate: true,
@@ -814,7 +843,13 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 				new: &testPod{
 					scheduled: true, completed: true,
 					network: &nadapi.NetworkSelectionElement{Namespace: "namespace", Name: "nad"},
-					labels:  map[string]string{"vm.kubevirt.io/name": "myvm"},
+					labels: map[string]string{
+						kubevirtv1.VirtualMachineNameLabel: "myvm",
+						kubevirtv1.AppLabel:                "virt-launcher",
+					},
+					annotations: map[string]string{
+						kubevirtv1.DomainAnnotation: "myvm",
+					},
 				},
 			},
 			expectMACOwnerID: "namespace/myvm",
--- a/go-controller/pkg/kubevirt/dhcp.go
+++ b/go-controller/pkg/kubevirt/dhcp.go
@@ -75,11 +75,14 @@ func WithIPv6DNSServer(dnsServer string) func(*dhcpConfigs) {
 }
 
 func EnsureDHCPOptionsForLSP(controllerName string, nbClient libovsdbclient.Client, pod *corev1.Pod, ips []*net.IPNet, lsp *nbdb.LogicalSwitchPort, opts ...DHCPConfigsOpt) error {
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return fmt.Errorf("failed discovering vm description at pod %s/%s:%w", pod.Namespace, pod.Name, err)
+	}
+	if vmDescription == nil {
 		return fmt.Errorf("missing vm label at pod %s/%s", pod.Namespace, pod.Name)
 	}
-	dhcpConfigs, err := composeDHCPConfigs(controllerName, *vmKey, ips, opts...)
+	dhcpConfigs, err := composeDHCPConfigs(controllerName, vmDescription.Key(), ips, opts...)
 	if err != nil {
 		return fmt.Errorf("failed composing DHCP options: %v", err)
 	}
@@ -171,12 +174,15 @@ func composeDHCPOptions(controllerName string, vmKey ktypes.NamespacedName, dhcp
 }
 
 func DeleteDHCPOptions(nbClient libovsdbclient.Client, pod *corev1.Pod) error {
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return fmt.Errorf("failed discovering vm description at pod %s/%s:%w", pod.Namespace, pod.Name, err)
+	}
+	if vmDescription == nil {
 		return nil
 	}
 	if err := libovsdbops.DeleteDHCPOptionsWithPredicate(nbClient, func(item *nbdb.DHCPOptions) bool {
-		return item.ExternalIDs[string(libovsdbops.ObjectNameKey)] == vmKey.String()
+		return item.ExternalIDs[string(libovsdbops.ObjectNameKey)] == vmDescription.Key().String()
 	}); err != nil {
 		return err
 	}
--- a/go-controller/pkg/kubevirt/pod.go
+++ b/go-controller/pkg/kubevirt/pod.go
@@ -12,9 +12,11 @@ import (
 	kubevirtv1 "kubevirt.io/api/core/v1"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/validate/content"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	ktypes "k8s.io/apimachinery/pkg/types"
-	v1 "k8s.io/client-go/listers/core/v1"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/util/retry"
 
 	libovsdbclient "github.com/ovn-kubernetes/libovsdb/client"
@@ -29,6 +31,12 @@ import (
 	ovntypes "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util/ndp"
+)
+
+var (
+	virtLauncherPodLabel = map[string]string{
+		kubevirtv1.AppLabel: "virt-launcher",
+	}
 )
 
 // DefaultGatewayReconciler is responsible for reconciling the default gateway
@@ -61,29 +69,17 @@ func IsPodLiveMigratable(pod *corev1.Pod) bool {
 	return ok
 }
 
-// TODO: remove adapter once all findVMRelatedPods usages transition to use PodLister
-type listPodsFn func(namespace string, selector metav1.LabelSelector) ([]*corev1.Pod, error)
-
 // findVMRelatedPods will return pods belong to the same vm annotated at pod and
 // filter out the one at the function argument
 func findVMRelatedPods(client *factory.WatchFactory, pod *corev1.Pod) ([]*corev1.Pod, error) {
-	return findVMRelatedPodsWithListerFn(client.GetPodsBySelector, pod)
-}
-
-func findVMRelatedPodsWithListerFn(listPodsFn listPodsFn, pod *corev1.Pod) ([]*corev1.Pod, error) {
-	vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]
-	if !ok {
-		return nil, nil
-	}
-	vmLabelSelector := metav1.LabelSelector{MatchLabels: map[string]string{kubevirtv1.VirtualMachineNameLabel: vmName}}
-	vmPods, err := listPodsFn(pod.Namespace, vmLabelSelector)
+	vmDescription, err := NewVMDescriptionFromPod(pod)
 	if err != nil {
 		return nil, err
 	}
-	if len(vmPods) == 0 {
-		return []*corev1.Pod{}, nil
+	vmPods, err := vmDescription.OwnedPods(client.PodCoreInformer().Lister())
+	if err != nil {
+		return nil, err
 	}
-
 	filteredOutVMPods := []*corev1.Pod{}
 	for _, vmPod := range vmPods {
 		// The purpose of this function is to return the "other" pods related
@@ -165,19 +161,17 @@ func EnsurePodAnnotationForVM(watchFactory *factory.WatchFactory, kube *kube.Kub
 }
 
 // AllVMPodsAreCompleted return true if all the vm pods are completed
-func AllVMPodsAreCompleted(podLister v1.PodLister, pod *corev1.Pod) (bool, error) {
+func AllVMPodsAreCompleted(podLister listersv1.PodLister, pod *corev1.Pod) (bool, error) {
 	if !util.PodCompleted(pod) {
 		return false, nil
 	}
 
-	f := func(namespace string, selector metav1.LabelSelector) ([]*corev1.Pod, error) {
-		s, err := metav1.LabelSelectorAsSelector(&selector)
-		if err != nil {
-			return nil, err
-		}
-		return podLister.Pods(namespace).List(s)
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return false, err
 	}
-	vmPods, err := findVMRelatedPodsWithListerFn(f, pod)
+
+	vmPods, err := vmDescription.OwnedPods(podLister)
 	if err != nil {
 		return false, fmt.Errorf("failed finding related pods for pod %s/%s when checking if they are completed: %v", pod.Namespace, pod.Name, err)
 	}
@@ -243,14 +237,93 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string, 
 	return false, nil
 }
 
-// ExtractVMNameFromPod returns namespace and name of vm backed up but the pod
-// for regular pods return nil
-func ExtractVMNameFromPod(pod *corev1.Pod) *ktypes.NamespacedName {
-	vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]
+// VMDescription holds the identity and ownership information for a KubeVirt
+// virtual machine derived from one of its virt-launcher pods. The key field
+// stores the namespaced name of the VM, and ownedPodsFn is a closure that
+// lists all pods belonging to the same VM via label or annotation selectors.
+type VMDescription struct {
+	key         ktypes.NamespacedName
+	ownedPodsFn func(podLister listersv1.PodLister) ([]*corev1.Pod, error)
+}
+
+func (vm VMDescription) Key() ktypes.NamespacedName {
+	return vm.key
+}
+
+func (vm VMDescription) OwnedPods(podLister listersv1.PodLister) ([]*corev1.Pod, error) {
+	return vm.ownedPodsFn(podLister)
+}
+
+func vmNameFromPod(pod *corev1.Pod) (string, error) {
+	vmName, ok := pod.Annotations[kubevirtv1.DomainAnnotation]
 	if !ok {
-		return nil
+		return "", fmt.Errorf("virtual machine pod %s/%s is missing the mandatory kubevirt annotation %s", pod.Namespace, pod.Name, kubevirtv1.DomainAnnotation)
 	}
-	return &ktypes.NamespacedName{Namespace: pod.Namespace, Name: vmName}
+	return vmName, nil
+}
+
+// nameToLabel truncates name to fit within the Kubernetes label value
+// maximum length of 63 characters (content.LabelValueMaxLength). If
+// name already fits, it is returned unchanged.
+func nameToLabel(name string) string {
+	if len(name) <= content.LabelValueMaxLength {
+		return name
+	}
+	return name[:content.LabelValueMaxLength]
+}
+
+// NewVMDescriptionFromPod builds a VMDescription from a virt-launcher pod.
+// Returns (nil, nil) if the pod is not owned by a virtual machine.
+// The VM name is extracted from the kubevirt domain annotation on the pod,
+// and the returned VMDescription.ownedPodsFn lists all sibling virt-launcher
+// pods belonging to the same VM. When the VM name label matches the
+// (possibly truncated) VM name, the lookup uses that label for efficiency;
+// otherwise it falls back to the generic virt-launcher label and filters
+// by the domain annotation to avoid false matches across VMs with long or
+// colliding names.
+func NewVMDescriptionFromPod(pod *corev1.Pod) (*VMDescription, error) {
+	if !IsPodOwnedByVirtualMachine(pod) {
+		return nil, nil
+	}
+
+	vmName, err := vmNameFromPod(pod)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get VM name from pod %s/%s: %w", pod.Namespace, pod.Name, err)
+	}
+	vmDescription := VMDescription{
+		key: ktypes.NamespacedName{Namespace: pod.Namespace, Name: vmName},
+	}
+
+	// By default filter pods first by "virt-launcher" pods
+	labelSelector := virtLauncherPodLabel
+
+	// if the label is the same (truncating it if is > 63 chars) search directly by label so we iterate less
+	vmNameLabel, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]
+	if ok && nameToLabel(vmName) == vmNameLabel {
+		labelSelector = map[string]string{kubevirtv1.VirtualMachineNameLabel: vmNameLabel}
+	}
+
+	vmDescription.ownedPodsFn = func(podLister listersv1.PodLister) ([]*corev1.Pod, error) {
+		labelMatchedPods, err := podLister.Pods(vmDescription.key.Namespace).List(labels.SelectorFromSet(labelSelector))
+		if err != nil {
+			return nil, err
+		}
+		// Filter by DomainAnnotation to ensure we only include pods that actually belong to this VM
+		// (different VMs can have the same label if one VM's name matches another VM's hostname, or if vms names are very long but matches the truncated version of it)
+		ownedPods := []*corev1.Pod{}
+		for _, virtLauncherPod := range labelMatchedPods {
+			podVMName, err := vmNameFromPod(virtLauncherPod)
+			if err != nil {
+				return nil, err
+			}
+			if podVMName == vmDescription.key.Name {
+				ownedPods = append(ownedPods, virtLauncherPod)
+			}
+		}
+		return ownedPods, nil
+	}
+
+	return &vmDescription, nil
 }
 
 // CleanUpLiveMigratablePod remove routing and DHCP ovn related resources
@@ -304,10 +377,7 @@ func SyncVirtualMachines(nbClient libovsdbclient.Client, vms map[ktypes.Namespac
 func FindLiveMigratablePods(watchFactory *factory.WatchFactory) ([]*corev1.Pod, error) {
 	vmPods, err := watchFactory.GetAllPodsBySelector(
 		metav1.LabelSelector{
-			MatchExpressions: []metav1.LabelSelectorRequirement{{
-				Key:      kubevirtv1.VirtualMachineNameLabel,
-				Operator: metav1.LabelSelectorOpExists,
-			}},
+			MatchLabels: virtLauncherPodLabel,
 		},
 	)
 	if err != nil {
@@ -335,7 +405,14 @@ func allocateSyncMigratablePodIPs(watchFactory *factory.WatchFactory, lsManager 
 		return nil, "", nil, nil
 	}
 
-	vmKey := ExtractVMNameFromPod(pod)
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return nil, "", nil, err
+	}
+	if vmDescription == nil {
+		return nil, "", nil, nil
+	}
+	vmKey := vmDescription.Key()
 
 	annotation, err := util.UnmarshalPodAnnotation(pod.Annotations, nadKey)
 	if err != nil {
@@ -345,13 +422,13 @@ func allocateSyncMigratablePodIPs(watchFactory *factory.WatchFactory, lsManager 
 	// If this zone do not own the subnet or the node that is passed
 	// do not match the switch, they should not be deallocated
 	if !zoneContainsPodSubnet || (nodeName != "" && switchName != nodeName) {
-		return vmKey, "", annotation, nil
+		return &vmKey, "", annotation, nil
 	}
 	expectedLogicalPortName, err := allocatePodIPsOnSwitch(pod, annotation, nadKey, switchName)
 	if err != nil {
-		return vmKey, "", nil, err
+		return &vmKey, "", nil, err
 	}
-	return vmKey, expectedLogicalPortName, annotation, nil
+	return &vmKey, expectedLogicalPortName, annotation, nil
 }
 
 // AllocateSyncMigratablePodIPsOnZone will refill ip pool in
@@ -391,7 +468,12 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
 // IsPodOwnedByVirtualMachine returns true if the pod is own by a
 // kubevirt virtual machine, false otherwise.
 func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
-	return ExtractVMNameFromPod(pod) != nil
+	for k, v := range virtLauncherPodLabel {
+		if pod.Labels[k] != v {
+			return false
+		}
+	}
+	return true
 }
 
 // IsPodAllowedForMigration determines whether a given pod is eligible for live migration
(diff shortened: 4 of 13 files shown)


go-controller/pkg/kubevirt/pod_test.go
```diff
@@ -159,23 +175,124 @@ var _ = Describe("Kubevirt Pod", func() {
...
+
+			desc, err := NewVMDescriptionFromPod(&sourcePod)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(desc).ToNot(BeNil())
+			Expect(desc.Key().Name).To(Equal(vmName))
+
+			pods, err := desc.OwnedPods(wf.PodCoreInformer().Lister())
+			Expect(err).ToNot(HaveOccurred())
+			Expect(pods).To(HaveLen(2))
+		})
```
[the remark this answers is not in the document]
qinqon: Added `DomainAnnotation` assertions to both the standard and long-name tests to verify the returned pods are correct.

Regarding adding collision fixtures to the fast-path tests: the standard and long-name cases both hit the fast path (`nameToLabel(vmName) == vmNameLabel`), which uses a label selector and intentionally does **not** filter by annotation. Adding an unrelated pod with a matching label would cause the test to fail — not because of a bug, but because that's the accepted trade-off of the fast path optimization.

The slow path (hostname case at line 225) already has the collision fixture with the unrelated pod and annotation assertions, which is where it matters — that path iterates all virt-launcher pods and filters by `DomainAnnotation`.

test/e2e/kubevirt.go
```diff
@@ -2143,84 +2142,69 @@ ip route add %[3]s via %[4]s
...
-			Eventually(func() error {
-				_, err := virtLauncherCommand("systemctl is-active NetworkManager")
-				return err
-			}).
-				WithTimeout(5 * time.Second).
-				WithPolling(time.Second).
-				Should(Succeed())
+			By("Wait for VMI to be ready")
+			waitVirtualMachineInstanceReadiness(vmi)
+			Expect(crClient.Get(context.TODO(), crclient.ObjectKeyFromObject(vmi), vmi)).To(Succeed())
```
ormergi: nit: context.TODO is deprecated, superseded by context.Background

go-controller/pkg/kubevirt/pod.go
```diff
@@ -388,7 +445,7 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
 // IsPodOwnedByVirtualMachine returns true if the pod is own by a
 // kubevirt virtual machine, false otherwise.
 func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
-	return ExtractVMNameFromPod(pod) != nil
+	return labels.SelectorFromSet(virtLauncherPodLabel).Matches(labels.Set(pod.Labels))
```
ormergi: Note that this function is called frequently, and SelectorFromSet does a lot under the hood (basically full flagged label selector).
Please consider simplifying this by checking if kubevirt.io=virt-launcher label present directly:
```suggestion
	v, exist := pod.Labels[kubevirtv1.AppLabel]
	return exist && v == "virt-launcher"
```

go-controller/pkg/kubevirt/pod.go
```diff
@@ -240,14 +234,76 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+func (vm VMDescription) Key() ktypes.NamespacedName {
+	return vm.key
+}
+
+func (vm VMDescription) OwnedPods(podLister listersv1.PodLister) ([]*corev1.Pod, error) {
+	return vm.ownedPodsFn(podLister)
+}
+
+func vmNameFromPod(pod *corev1.Pod) (string, error) {
+	vmName, ok := pod.Annotations[kubevirtv1.DomainAnnotation]
```
ormergi, line 251: nit: I think the owner-reference is more stable than the existing annotations or labels Kubevirt sets on the pods.
Please consider fetching the associated VM name from the pod owner-reference.
```suggestion
	if r := metav1.GetControllerOf(pod); r != nil && r.Kind == "VirtualMachineInstance" && r.Name != "" {
		return r.Name, nill
	}
```

go-controller/pkg/kubevirt/pod.go
```diff
@@ -240,14 +234,76 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
-// for regular pods return nil
-func ExtractVMNameFromPod(pod *corev1.Pod) *ktypes.NamespacedName {
-	vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]
+type VMDescription struct {
+	key         ktypes.NamespacedName
+	ownedPodsFn func(podLister listersv1.PodLister) ([]*corev1.Pod, error)
+}
+
+func (vm VMDescription) Key() ktypes.NamespacedName {
+	return vm.key
```
ormergi, line 243: nit: It seems that in places Key() is called, simple annotation check could be done and simplify the code a little bit (e.g.: no error handling).
Please consider keeping a helper for that instead of "going through" the VMDescription type.

go-controller/pkg/kubevirt/router.go
```diff
@@ -136,7 +146,7 @@ func EnsureLocalZonePodAddressesToNodeRoute(watchFactory *factory.WatchFactory,
 				Priority: types.EgressLiveMigrationReroutePriority,
 				ExternalIDs: map[string]string{
 					OvnZoneExternalIDKey:         OvnLocalZone,
-					VirtualMachineExternalIDsKey: pod.Labels[kubevirtv1.VirtualMachineNameLabel],
+					VirtualMachineExternalIDsKey: vmDescription.Key().Name,
```
ormergi, line 158: Note that when pod is non VM pod, vmDescription is nil, and this call could blow up. 
Same for other occurrences in this file. Please add nil check to avoid it.

go-controller/pkg/kubevirt/pod_test.go
```diff
@@ -186,23 +202,130 @@ var _ = Describe("Kubevirt Pod", func() {
...
+	Describe("NewVMDescriptionFromPod", func() {
+		It("returns nil for non-kubevirt pod", func() {
+			pod := nonKubevirtPod()
+			desc, err := NewVMDescriptionFromPod(&pod)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(desc).To(BeNil())
+		})
+
+		It("returns correct key and finds owned pods for standard pod (label == domain)", func() {
+			sourcePod := runningKvSourcePod
```
ormergi, line 250: nit: Please add text indicating test involve Kubevirt's vmName label. Here and in other occurrences in this file.
```suggestion
		It("returns correct key and finds owned pods for standard pod (vmName label == domain)", func() {

```

go-controller/pkg/allocator/pod/macs.go
```diff
@@ -20,14 +18,18 @@ import (
 
 // macOwner compose the owner identifier reserved for MAC addresses management.
 // Returns "<ns>/<pod-name>" for regular pods and "<ns>/<vm-name>" for VMs.
-func macOwner(pod *corev1.Pod) string {
+func macOwner(pod *corev1.Pod) (string, error) {
 	// Check if this is a VM pod and persistent IPs are enabled
-	if vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]; ok {
-		return fmt.Sprintf("%s/%s", pod.Namespace, vmName)
+	if kubevirt.IsPodOwnedByVirtualMachine(pod) {
+		vmDescription, err := kubevirt.NewVMDescriptionFromPod(pod)
```
ormergi: `NewVMDescritpion()` already checks the given pod is VM pod  using `IsPodOwnedByVirtualMachine()`.
Please consider simplifying this condition to check`vmDescription != nil`.

test/e2e/kubevirt.go
```diff
@@ -2315,6 +2315,52 @@ password: fedora
 chpasswd: { expire: False }
 `
 		)
+		It("should start multiple VMs with same hostname", func() {
+			By("setting up the localnet underlay")
```
ormergi, line 2304: I am not sure about this test because the original issue manifest when multiple VMs shares the same hostname and one of the VM is migrated. 
So that ovn-k wont detect the migration properly and result with network-stuntime until migration is finished. Is it correct?

If so, I think this test should be removed or changed to assert a VM migrate successfully with minimal network stuntime on post migration.
WDTY?

ormergi (commented):
Thanks for fixing this, please see my comments

go-controller/pkg/kubevirt/pod.go
```diff
@@ -240,14 +234,76 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+}
+
+func nameToLabel(name string) string {
+	if len(name) <= validation.LabelValueMaxLength {
+		return name
+	}
+	return name[:validation.LabelValueMaxLength]
+}
+
+func NewVMDescriptionFromPod(pod *corev1.Pod) (*VMDescription, error) {
```
RamLavi, line 265: Just highlighting that this approach fixes the hostname-override case, but it still does not cover the case where two different VMs have long names that share the same first 63 characters. 

In that scenario vm.kubevirt.io/name still collides, and might result with the same issue. 
Would it make sense to prefer vmi.kubevirt.io/id when present, or at least post-filter the selected pods by kubevirt.io/domain == vmName?

go-controller/pkg/kubevirt/pod.go
qinqon: I will iterate the map so we don't harcode "virt-launcher" 
```golang
for k, v := range virtLauncherPodLabel {
		if pod.Labels[k] != v {
			return false
		}
	}

```

go-controller/pkg/kubevirt/pod.go
qinqon: As far as I know the domain annotation is the virt domain I don't see why using the owner will improve things, the fact that the pod has a controller with the same virtualmachieninstance name feels very implementatin detail

go-controller/pkg/kubevirt/pod.go
qinqon: What do you mean ?

test/e2e/kubevirt.go
qinqon: bug is related to start up not migration, so this test is ok.

go-controller/pkg/kubevirt/pod.go
qinqon: > Just highlighting that this approach fixes the hostname-override case, but it still does not cover the case where two different VMs have long names that share the same first 63 characters.
> 
> In that scenario vm.kubevirt.io/name still collides, and might result with the same issue. Would it make sense to prefer vmi.kubevirt.io/id when present, or at least post-filter the selected pods by kubevirt.io/domain == vmName?

This is exactly what we are doing, iterate the virt-launcher pod and compare domain so we are not sufferering the issue, maybe what we can do is to add a e2e tests for the 63 chars issue ? is it worth it ? 

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6118/changes#diff-eb30366251cb823a9618bac3c224e92aa52cbedf5fa9c0f5a2a4360902540a42R295

qinqon: Added unit test to check the truncation

RamLavi: thanks!

RamLavi (approved):
/lgtm

go-controller/pkg/kubevirt/pod.go
ormergi: Drop Key() method, and check for the kubevirt.io/virt-launcher label instead.
This should save from instantiating VMDescription just for checking if a pod is managed.
And easier to work with because no error handing is involved.

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,92 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
 	}
-	return &ktypes.NamespacedName{Namespace: pod.Namespace, Name: vmName}
+	return name[:validation.LabelValueMaxLength]
+}
+
+func NewVMDescriptionFromPod(pod *corev1.Pod) (*VMDescription, error) {
+	if !IsPodOwnedByVirtualMachine(pod) {
+		return nil, nil
+	}
+
```
ormergi: nit: Please consider returning an error when pod is not VM pod. It should simplify usage and eliminate nil checks.

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -75,11 +75,11 @@ func WithIPv6DNSServer(dnsServer string) func(*dhcpConfigs) {
...
 func EnsureDHCPOptionsForLSP(controllerName string, nbClient libovsdbclient.Client, pod *corev1.Pod, ips []*net.IPNet, lsp *nbdb.LogicalSwitchPort, opts ...DHCPConfigsOpt) error {
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
-		return fmt.Errorf("missing vm label at pod %s/%s", pod.Namespace, pod.Name)
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return err
 	}
-	dhcpConfigs, err := composeDHCPConfigs(controllerName, *vmKey, ips, opts...)
+	dhcpConfigs, err := composeDHCPConfigs(controllerName, vmDescription.Key(), ips, opts...)
```
ormergi, line 85: This could blow up given non VM pod, please add nil check for vmDescription

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -171,12 +171,12 @@ func composeDHCPOptions(controllerName string, vmKey ktypes.NamespacedName, dhcp
...
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
-		return nil
+	vmDescrption, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return err
 	}
 	if err := libovsdbops.DeleteDHCPOptionsWithPredicate(nbClient, func(item *nbdb.DHCPOptions) bool {
-		return item.ExternalIDs[string(libovsdbops.ObjectNameKey)] == vmKey.String()
+		return item.ExternalIDs[string(libovsdbops.ObjectNameKey)] == vmDescrption.Key().String()
```
ormergi: This could blow up given no VM pod, please add nil check for vmDescription (not all code path are VM pods specif)

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -171,12 +171,12 @@ func composeDHCPOptions(controllerName string, vmKey ktypes.NamespacedName, dhcp
 }
 
 func DeleteDHCPOptions(nbClient libovsdbclient.Client, pod *corev1.Pod) error {
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
-		return nil
+	vmDescrption, err := NewVMDescriptionFromPod(pod)
```
ormergi: typo
```suggestion
	vmDescription, err := NewVMDescriptionFromPod(pod)

```

go-controller/pkg/kubevirt/pod.go
```diff
@@ -335,7 +404,11 @@ func allocateSyncMigratablePodIPs(watchFactory *factory.WatchFactory, lsManager
 		return nil, "", nil, nil
 	}
 
-	vmKey := ExtractVMNameFromPod(pod)
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return nil, "", nil, err
+	}
+	vmKey := vmDescription.Key()
```
ormergi, line 415: Currently, is no code path for allocateSyncMigratablePodIPs where non VM pod is given, but nothing prevent this to change later, and this call could blow up given non VM pod.
Please consider adding nil check for vmDescription to avoid that.

go-controller/pkg/kubevirt/pod_test.go
```diff
@@ -189,23 +209,166 @@ var _ = Describe("Kubevirt Pod", func() {
...
+				expectedMigrationStatus: &LiveMigrationStatus{
+					SourcePod: &runningKvSourcePodWithLongName,
+					TargetPod: &duringMigrationKvTargetPodWithLongName,
+					State:     LiveMigrationInProgress,
+				},
+			},
+		),
 	)
+
+	Describe("NewVMDescriptionFromPod", func() {
```
ormergi, line 243: 1. I think we are missing coverage for scenario where there is a mix of VM and non VM pods in a namespace, and VMDescription find the correct ones.
This can be additional test or add coverage for the tests you added (add non VM pod to WatchFactory).
WDYT?


2. nit: Following tests assert the same state and can be converted to DescribeTable where the input is the pods and the expected VM name
  - returns correct key and finds owned pods when VM has hostname...
  - returns correct key and finds owned pods when VM name exceeds 63 chars...
  - returns only owned pods when multiple VMs share the same truncated vmName label

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,76 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+	return vmName, nil
+}
+
+func nameToLabel(name string) string {
+	if len(name) <= validation.LabelValueMaxLength {
+		return name
+	}
+	return name[:validation.LabelValueMaxLength]
+}
+
```
ormergi, line 274: nit: Note that `LabelValueMaxLength` is deprecated and superseded by 
`k8s.io/apimachinery/pkg/api/validate/content.LabelValueMaxLength`

ormergi (commented):
Thanks for the changes

go-controller/pkg/kubevirt/pod.go
ormergi: You can resolve this thread.

ormergi (approved):
Thanks for the changes, LGTM

qinqon:
Rebased commit 0d61932 on main and pushed (conflicts resolved).

qinqon:
This appears to be a duplicate of existing flaky test issue #6329.

qinqon:
CI check `e2e (control-plane, HA, shared, ipv4, snatGW, 1br, ic-disabled, 1,2,3)` failed on commit af7039c but appears unrelated to this PR's changes.

FAILING_TEST: e2e ingress traffic validation Validating ingress traffic to manually added node IPs [BeforeEach] Should be allowed by externalip services to a new node ip

The e2e test `Should be allowed by externalip services to a new node ip` (`test/e2e/e2e.go:1693`) failed in the `BeforeEach` setup at line 1671. The test attempts to add IP address `172.18.1.1` to the `breth0` interface on the `ovn-control-plane` node, but the `ip addr add` command failed with `Error: ipv4: Address already assigned`. This is a known flaky test (tracked by issue #6140) where IP address cleanup from a previous test run is incomplete, causing the address to already exist when the next test iteration tries to add it. The test also had an additional failure trying to curl the NodePort at `172.18.1.1:81` which returned `connection refused`, likely a cascade from the initial setup failure. A second unrelated flaky test (`Service Hairpin SNAT / Should ensure service hairpin traffic is NOT SNATed to hairpin masquerade IP; GR LB`) also failed in a separate CI lane, tracked by issue #6321.

qinqon:
Rebased commit e8ee177 on main and pushed.

tssurya:
sorry to make you wait on this.. this one is on me.. I'll get to this, this week

qinqon:
CI check `e2e (kv-live-migration, noHA, local, dualstack, noSnatGW, 1br, ic-single-node-zones, 3, enable-ne...` failed on commit 944680b but appears unrelated to this PR's changes.

FAILING_TEST: Kubevirt Virtual Machines with user defined networks and persistent ips configured should keep ip [It] after live migration of VirtualMachine with interface binding for UDN with Primary/Layer2 ingress routed over evpn

The e2e test `after live migration of VirtualMachine with interface binding for UDN with Primary/Layer2 ingress routed over evpn` at `kubevirt.go:2195` is a known flaky test (previously tracked by closed issue #6381, created 2026-05-07). The test runs continuous iperf3 traffic between a VM and an external container over an EVPN-based Layer2 primary UDN, performs a live migration, then checks that no iperf3 reporting interval shows `0.00 Bytes 0.00 bits/sec`. During migration, the VM's network connectivity briefly transitions between source and destination nodes. Depending on exact timing, one iperf3 reporting interval (in this case interval `82.00-83.00` seconds in `/tmp/ingress_test_10.235.128.7_5201_iperf3.log`) may record zero transfer before traffic resumes at normal rates. The test passed for all other traffic streams (east-west, egress) and only failed on one specific IPv4 ingress stream, indicating a transient timing issue during the network handover rather than a systemic connectivity problem.

go-controller/pkg/allocator/pod/macs.go
```diff
@@ -24,9 +22,9 @@ import (
 // macOwner compose the owner identifier reserved for MAC addresses management.
 // Returns "<ns>/<pod-name>" for regular pods and "<ns>/<vm-name>" for VMs.
 func macOwner(pod *corev1.Pod) string {
-	// Check if this is a VM pod and persistent IPs are enabled
-	if vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]; ok {
-		return fmt.Sprintf("%s/%s", pod.Namespace, vmName)
+	vmDescription, err := kubevirt.NewVMDescriptionFromPod(pod)
+	if err == nil && vmDescription != nil {
```
tssurya, line 26: is there a situation where err is nil but vmDescription is also nil? i.e the function signature should be that if the err is nil there is a value returned right?

go-controller/pkg/allocator/pod/macs.go
```diff
@@ -43,10 +41,10 @@ func (allocator *PodAnnotationAllocator) ReleasePodReservedMacAddress(pod *corev
 	}
 
 	macOwnerID := macOwner(pod)
-	if vmKey := kubevirt.ExtractVMNameFromPod(pod); vmKey != nil {
+	if kubevirt.IsPodOwnedByVirtualMachine(pod) {
 		allVMPodsCompleted, err := kubevirt.AllVMPodsAreCompleted(allocator.podLister, pod)
 		if err != nil {
-			return fmt.Errorf("failed checking all VM %q pods are completed: %v", vmKey, err)
+			return fmt.Errorf("failed checking all VM %q pods are completed: %v", macOwnerID, err)
```
tssurya, line 47: what's the macOwnerID - is this something that is easily recognisable for debugging? should we be adding the pod name instead to the return error log?

go-controller/pkg/clustermanager/pod/allocator_test.go
```diff
@@ -791,7 +820,13 @@ func TestPodAllocator_reconcileForNAD(t *testing.T) {
 				new: &testPod{
 					scheduled: true, completed: true,
 					network: &nadapi.NetworkSelectionElement{Namespace: "namespace", Name: "nad"},
-					labels:  map[string]string{"vm.kubevirt.io/name": "myvm"},
+					labels: map[string]string{
+						kubevirtv1.VirtualMachineNameLabel: "myvm",
+						kubevirtv1.AppLabel:                "virt-launcher",
+					},
+					annotations: map[string]string{
```
tssurya, line 850: do we have tests where this annotation could be missing because for whatever reason kubevirt didn't assign it ?  - in that case its infinite retries?
also I don't see tests with the new `id` label used..

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -75,11 +75,14 @@ func WithIPv6DNSServer(dnsServer string) func(*dhcpConfigs) {
 }
 
 func EnsureDHCPOptionsForLSP(controllerName string, nbClient libovsdbclient.Client, pod *corev1.Pod, ips []*net.IPNet, lsp *nbdb.LogicalSwitchPort, opts ...DHCPConfigsOpt) error {
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return err
```
tssurya: nit: wrap the error into fmt.Error with pod namespace/name etc..

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -171,12 +174,15 @@ func composeDHCPOptions(controllerName string, vmKey ktypes.NamespacedName, dhcp
 }
 
 func DeleteDHCPOptions(nbClient libovsdbclient.Client, pod *corev1.Pod) error {
-	vmKey := ExtractVMNameFromPod(pod)
-	if vmKey == nil {
+	vmDescription, err := NewVMDescriptionFromPod(pod)
+	if err != nil {
+		return err
```
tssurya: same here wrap the error in fmt.Errorf

go-controller/pkg/kubevirt/dhcp.go
```diff
@@ -171,12 +174,15 @@ func composeDHCPOptions(controllerName string, vmKey ktypes.NamespacedName, dhcp
...
+	}
+	if vmDescription == nil {
 		return nil
 	}
 	if err := libovsdbops.DeleteDHCPOptionsWithPredicate(nbClient, func(item *nbdb.DHCPOptions) bool {
-		return item.ExternalIDs[string(libovsdbops.ObjectNameKey)] == vmKey.String()
+		return item.ExternalIDs[string(libovsdbops.ObjectNameKey)] == vmDescription.Key().String()
```
tssurya, line 185: so this is backwards compatible change?
what happens to pods/vms where we were using the hostname - so that's what we fed to the externalIDs - now on start up for the older pods will we automatically change all the externalIDs on restart post upgrade to use the actual VMI name instead of the hostname?
otherwise for the older pods this return fetch won't work anymore...

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,92 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
 	return false, nil
 }
 
-// ExtractVMNameFromPod returns namespace and name of vm backed up but the pod
-// for regular pods return nil
-func ExtractVMNameFromPod(pod *corev1.Pod) *ktypes.NamespacedName {
-	vmName, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]
+type VMDescription struct {
+	key         ktypes.NamespacedName
```
tssurya, line 245: nit: its the pod namespace/ pod name? could you please add a small doc string for this struct?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,92 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+}
+
+func vmNameFromPod(pod *corev1.Pod) (string, error) {
+	vmName, ok := pod.Annotations[kubevirtv1.DomainAnnotation]
 	if !ok {
-		return nil
+		return "", fmt.Errorf("virtual machine pod %s/%s is missing the mandatory kubevirt annotation %s", pod.Namespace, pod.Name, kubevirtv1.DomainAnnotation)
 	}
-	return &ktypes.NamespacedName{Namespace: pod.Namespace, Name: vmName}
+	return vmName, nil
```
tssurya, line 262: so when we say its unique per VMI, (sorry I am not super familiar with all kubevirt  CRs), its possible the domain annotation is the same for pods across different VMIs? Remind me what's a VMI? 🗡️

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,92 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+}
+
+func nameToLabel(name string) string {
```
tssurya, line 268: this util needs a doc stirng, I'm not sure I follow the logic behind the truncation or why its needed

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,92 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+}
+
+func NewVMDescriptionFromPod(pod *corev1.Pod) (*VMDescription, error) {
+	if !IsPodOwnedByVirtualMachine(pod) {
+		return nil, nil
+	}
+
+	vmName, err := vmNameFromPod(pod)
+	if err != nil {
+		return nil, err
```
tssurya: nit: error wrap

go-controller/pkg/kubevirt/pod.go
```diff
@@ -391,7 +467,12 @@ func ZoneContainsPodSubnetOrUntracked(watchFactory *factory.WatchFactory, lsMana
 // IsPodOwnedByVirtualMachine returns true if the pod is own by a
 // kubevirt virtual machine, false otherwise.
 func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
-	return ExtractVMNameFromPod(pod) != nil
+	for k, v := range virtLauncherPodLabel {
```
tssurya, line 471: nit maybe the inbuilt utils work?

```
func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
	return labels.SelectorFromSet(virtLauncherPodLabel).Matches(labels.Set(pod.Labels))
}
``` 
?

go-controller/pkg/kubevirt/pod.go
```diff
@@ -345,13 +421,13 @@ func allocateSyncMigratablePodIPs(watchFactory *factory.WatchFactory, lsManager
 	// If this zone do not own the subnet or the node that is passed
 	// do not match the switch, they should not be deallocated
 	if !zoneContainsPodSubnet || (nodeName != "" && switchName != nodeName) {
-		return vmKey, "", annotation, nil
+		return &vmKey, "", annotation, nil
```
tssurya, line 425: I hope vmKey can never be nil

go-controller/pkg/kubevirt/pod.go
```diff
@@ -243,14 +237,92 @@ func nodeContainsPodSubnet(watchFactory *factory.WatchFactory, nodeName string,
...
+	if err != nil {
+		return nil, err
+	}
+	vmDescription := VMDescription{
+		key: ktypes.NamespacedName{Namespace: pod.Namespace, Name: vmName},
+	}
+
+	vmNameLabel, ok := pod.Labels[kubevirtv1.VirtualMachineNameLabel]
+
+	// the label is the same (truncating it if is > 63 chars) search by label so we don't have to iterate
```
tssurya: I think I am missing some context on this truncation logic, maybe add something to this effect to the commit message?

tssurya (changes requested):
did a first round, have a few questions

tssurya:
Thank you for the nice description!

> ## 📑 Description
> When a VM's spec.template.spec.hostname is set, KubeVirt copies it into the vm.kubevirt.io/name label on virt-launcher pods instead of the actual VM name. OVN-K used this label to find pods belonging to a VM (e.g. during live migration, or to tag OVN topology elements for cleanup). Since multiple VMs can share the same hostname, the label can no longer uniquely identify a VM's pods — with 3+ VMs sharing a hostname, the 3rd VM fails with "unexpected live migration state at pods".
> 
> This change switches to the kubevirt.io/domain annotation, which always reflects the real VM name. Since annotations can't be used in label selectors, we optimize: if the domain name matches the vm.kubevirt.io/name label (the common case), we use the label for lookup; otherwise we fall back to iterating virt-launcher pods in the namespace and matching by annotation.
> 
> Note: KubeVirt >= 1.7 introduces a vm.kubevirt.io/id label that reliably identifies the VM, but it is not added retroactively to pods created with older versions. If KubeVirt adds a mutating webhook for that in the future, we can use it as a further optimization.

Why didn't Kubevirt do this for the older version pods? Shouldn't the change be backwards compatible where things upgrade and they start using the new label?
So that means this optimization lookup is something we need to carry forever? That sounds bad, and perhaps kubevirt should at start up consider doing the new label for all older pods w/o the label right?

go-controller/pkg/allocator/pod/macs.go
qinqon: > is there a situation where err is nil but vmDescription is also nil? i.e the function signature should be that if the err is nil there is a value returned right?

If we run this functions to a non virt pod, it will return nil, nil, should we return error if we run this functions no non virt pods ?

go-controller/pkg/allocator/pod/macs.go
qinqon: > what's the macOwnerID - is this something that is easily recognisable for debugging? should we be adding the pod name instead to the return error log?

For virt-launcher pods is "namespace/vm" for normal pods is "namespace/pod", this error enters vm category so it's "namespace/vm" so macOwnerID is the same as the old "vmKey" so the error is seactly the same we did have, do we still need to change it ?

go-controller/pkg/clustermanager/pod/allocator_test.go
qinqon: > do we have tests where this annotation could be missing because for whatever reason kubevirt didn't assign it ?

If domain annotation is missing ovn-kubernetes will retry, that annotations is there at kubevirt from the beginning and if it's not there it will be a kubevirt BUG that need fixing. 

> in that case its infinite retries?

There is a unit test for NewVMDescriptionFromPod where we check that if domain is not there when annotation is missing

> also I don't see tests with the new `id` label used..

Because we ended not using it, since we found it will not cover old version so we decided to use something that is present in old and new version so we don't need to put just another fallback that will complicate stuff.

go-controller/pkg/kubevirt/dhcp.go
qinqon: > so this is backwards compatible change?
> what happens to pods/vms where we were using the hostname - so that's what we fed to the externalIDs - now on start up for the older pods will we automatically change all the externalIDs on restart post upgrade to use the actual VMI name instead of the hostname?
> otherwise for the older pods this return fetch won't work anymore...

On upgrade the existing OVN resouces that reference the vm at external IDs or similar are migrated to whatever we decide NewVMDescriptionFromPod, so this coditions should work

go-controller/pkg/kubevirt/pod.go
qinqon: > nit: its the pod namespace/ pod name? could you please add a small doc string for this struct?

This is the VM's namespace name, I will add doc string to the struct

go-controller/pkg/kubevirt/pod.go
qinqon: > so when we say its unique per VMI, (sorry I am not super familiar with all kubevirt  CRs), its possible the domain annotation is the same for pods across different VMIs? Remind me what's a VMI? 🗡️

VMI is the virtual machine instance so per namespace we have just one VMI per virtual machine, the only scenario where two pods has domain with the same VMI name is on live migratio but that's already covered and understood at ovn-kubernetes.

go-controller/pkg/kubevirt/pod.go
qinqon: > this util needs a doc string, I'm not sure I follow the logic behind the truncation or why its needed

So the idea is to try as much as possible to use labels+selectors to list the virt-launcher pods since is farter than iterate over the domain annotations (although we pre filter the pods to only the virt-launcher ones). Then the list is also filtered to be sure that we don't enter the case where two VMs has super big names so the truncated label can match.

What I found though is that the code on the "if" vs the "else" is exactly the same except the label we use to filter, I will refactor refactor it to have same function but passing the label selector

go-controller/pkg/kubevirt/pod.go
qinqon: > nit maybe the inbuilt utils work?
> 
> ```
> func IsPodOwnedByVirtualMachine(pod *corev1.Pod) bool {
> 	return labels.SelectorFromSet(virtLauncherPodLabel).Matches(labels.Set(pod.Labels))
> }

It was like that originally, but virt review commented against this since we are converting between map and selector without need it 
> ``` 
> ?

go-controller/pkg/kubevirt/pod.go
qinqon: > I hope vmKey can never be nil

Then we will have "error" or "nil" at NewVMDescriptionFromPod

go-controller/pkg/kubevirt/pod.go
qinqon: > I think I am missing some context on this truncation logic, maybe add something to this effect to the commit message?

I have add docs related to it to NewVMDescriptionFromPod functions so now is clear that is to make this more efficient.

go-controller/pkg/kubevirt/dhcp.go
qinqon: > This could blow up given no VM pod, please add nil check for vmDescription (not all code path are VM pods specif)

Not addressing: This was already fixed in a previous iteration. The current code at `DeleteDHCPOptions` (line 181) already has `if vmDescription == nil { return nil }` and `EnsureDHCPOptions` (line 82) has `if vmDescription == nil { return fmt.Errorf(...) }`.

go-controller/pkg/kubevirt/dhcp.go
qinqon: > typo
> ```suggestion
> vmDescription, err := NewVMDescriptionFromPod(pod)
> ```

Not addressing: The typo (`vmDescrption` -> `vmDescription`) was already fixed in a previous iteration. The current code at line 177 already reads `vmDescription, err := NewVMDescriptionFromPod(pod)`.

qinqon:
> Thank you for the nice description!
> 
> > ## 📑 Description
> > When a VM's spec.template.spec.hostname is set, KubeVirt copies it into the vm.kubevirt.io/name label on virt-launcher pods instead of the actual VM name. OVN-K used this label to find pods belonging to a VM (e.g. during live migration, or to tag OVN topology elements for cleanup). Since multiple VMs can share the same hostname, the label can no longer uniquely identify a VM's pods — with 3+ VMs sharing a hostname, the 3rd VM fails with "unexpected live migration state at pods".
> > This change switches to the kubevirt.io/domain annotation, which always reflects the real VM name. Since annotations can't be used in label selectors, we optimize: if the domain name matches the vm.kubevirt.io/name label (the common case), we use the label for lookup; otherwise we fall back to iterating virt-launcher pods in the namespace and matching by annotation.
> > Note: KubeVirt >= 1.7 introduces a vm.kubevirt.io/id label that reliably identifies the VM, but it is not added retroactively to pods created with older versions. If KubeVirt adds a mutating webhook for that in the future, we can use it as a further optimization.
> 
> Why didn't Kubevirt do this for the older version pods? Shouldn't the change be backwards compatible where things upgrade and they start using the new label? 

Kubevirt is missing adding the label to the exising virt-launcher pods at kubevirt upgrade, this something we can implement in the future.

> So that means this optimization lookup is something we need to carry forever? That sounds bad, and perhaps kubevirt should at start up consider doing the new label for all older pods w/o the label right?

Ideally we propose a fix at kubevirt to update old virt-launcher pods and then after some releases we can simplify the code at ovnk and use just one label since we know after kubevirt upgrade with the fixes all the virt-launcher pods will have that id.

qinqon:
Live migration evpn failure unrelated 

https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6420

go-controller/pkg/allocator/pod/macs.go
tssurya: I guess do we have a use for this function for non-virt pods as well? if so I guess we can keep it as is

go-controller/pkg/allocator/pod/macs.go
tssurya: if its same as vmKey then its fine

go-controller/pkg/clustermanager/pod/allocator_test.go
tssurya: so I think we should start using `id` at some point right if that's the recommended way?

shouldn't it be `use id if that exists` , `if not for the old pods use the label if its the same` , if not fall back to annotations? - do we have an issue tracking this in kube virt for when kubevirt plans to add support to fix it for the older pods?

go-controller/pkg/kubevirt/pod.go
tssurya: > Then the list is also filtered to be sure that we don't enter the case where two VMs has super big names so the truncated label can match.

so that's the part that is a bit unclear to me,
if you truncate super long names into 63 where it matches for the two vms - in that case we want to use annotations right? and we don't want to use the labels then? - so I think on kubevirt side its truncating vmNameLabel in same way? that's why we have this logic, but if two different VMs whose names share the first 63 characters come then you truncate and return the same labels.. the fast path will return pods from both VMs right?

ok nvm I see you handle it here:
vmDescription.ownedPodsFn = func(podLister listersv1.PodLister) ([]*corev1.Pod, error) {
		labelMatchedPods, err := podLister.Pods(vmDescription.key.Namespace).List(labels.SelectorFromSet(labelSelector))
		if err != nil {
			return nil, err
		}
		// Filter by DomainAnnotation to ensure we only include pods that actually belong to this VM
		// (different VMs can have the same label if one VM's name matches another VM's hostname, or if vms names are very long but matches the truncated version of it)

ok so that means the optimization is only for how many pods we need to iterate over in fast path using vm.kubevirt.io/name v/s slow path where we return all pods in the namespace?

tssurya (commented):
the second and third commits lgtm
in first commit just two more clarifications required:

1) https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6118#discussion_r3273366473 please provide a tracker on kubevirt side so that we have a FUP PR to change it to ids at some point

2) https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6118#discussion_r3273582814

go-controller/pkg/clustermanager/pod/allocator_test.go
qinqon: > so I think we should start using `id` at some point right if that's the recommended way?
> 

Agree with that, but if kubevirt is not adding to the old virt-launcher pod after kubevirt upgrade I see it as problematic to have half the virt-launcher pods without the id label.

> shouldn't it be `use id if that exists` , `if not for the old pods use the label if its the same` , if not fall back to annotations? - 

We can try, but I would better do a folow up PR since the issue we have at hand is fixed by the current PR.

> do we have an issue tracking this in kube virt for when kubevirt plans to add support to fix it for the older pods?

Not yet

go-controller/pkg/kubevirt/pod.go
qinqon: > > Then the list is also filtered to be sure that we don't enter the case where two VMs has super big names so the truncated label can match.
> 
> so that's the part that is a bit unclear to me, if you truncate super long names into 63 where it matches for the two vms - in that case we want to use annotations right? and we don't want to use the labels then? - so I think on kubevirt side its truncating vmNameLabel in same way? that's why we have this logic, but if two different VMs whose names share the first 63 characters come then you truncate and return the same labels.. the fast path will return pods from both VMs right?
> 

Nope, because in the "fast" path we still iterate and compare with the annotation that always has the full name.

> ok nvm I see you handle it here: vmDescription.ownedPodsFn = func(podLister listersv1.PodLister) ([]*corev1.Pod, error) { labelMatchedPods, err := podLister.Pods(vmDescription.key.Namespace).List(labels.SelectorFromSet(labelSelector)) if err != nil { return nil, err } // Filter by DomainAnnotation to ensure we only include pods that actually belong to this VM // (different VMs can have the same label if one VM's name matches another VM's hostname, or if vms names are very long but matches the truncated version of it)
> 
> ok so that means the optimization is only for how many pods we need to iterate over in fast path using vm.kubevirt.io/name v/s slow path where we return all pods in the namespace?

At fast path we iterate over the pods that matches the 63 "name" label at the same namespace, that covers 99% of use case since user usually don't configure hostname and VM name is < 63 chars

The slow path is the 1% of users configuring hostname not matching VM name so we filter the virt-launcher pods first using label kubevirt.io=virt-launcher then we iterate them, meaning we iterate all the VMs at the namespace.

go-controller/pkg/clustermanager/pod/allocator_test.go
qinqon: @tssurya Issue https://github.com/kubevirt/kubevirt/issues/17869

tssurya (approved):

PR #6121, opened 2026-03-17
config: move EnableMulticast into OVNKubernetesFeatureConfig struct

weliang1:
## Problem

`EnableMulticast` was declared as a **global variable** in `config.go`, but CNO writes `enable-multicast=true` under the `[ovnkubernetesfeature]` section in `ovnkube.conf`. The gcfg parser looks for `enable-multicast` inside the `OVNKubernetesFeatureConfig` struct when parsing that section — since the field was not there, gcfg silently discards the value and emits:

```
W config.go: can't store data at section "ovnkubernetesfeature", variable "enable-multicast"
```

This leaves `config.EnableMulticast = false` at runtime even when multicast is configured, resulting in:
- No multicast ACLs created in the OVN NB database (`clusterRtrPortGroup acls: []`)
- No `mcast_snoop` set on node logical switches
- 100% cross-node multicast packet loss

Verified broken on OCP 4.21 (AWS) and 4.22.0 nightly (GCP).

## Fix

- Remove the orphaned global `EnableMulticast bool` declaration
- Add `EnableMulticast bool \`gcfg:"enable-multicast"\`` into `OVNKubernetesFeatureConfig` struct so the `[ovnkubernetesfeature]` config section is parsed correctly
- Update the CLI flag `Destination` to `&cliConfig.OVNKubernetesFeature.EnableMulticast`
- Update all call sites (`controller_manager.go`, `layer2_user_defined_network_controller.go`, `layer3_user_defined_network_controller.go`) and test files

## Testing

- `go build ./pkg/config/... ./pkg/controllermanager/... ./pkg/ovn/...` — passes
- Existing multicast unit tests updated to reference `config.OVNKubernetesFeature.EnableMulticast`

Fixes: https://issues.redhat.com/browse/OCPBUGS-78731

## Summary by CodeRabbit

* **Refactor**
  * Reorganized multicast configuration structure to support per-namespace feature settings instead of global settings, enabling more granular control over multicast support across different namespaces.

### DIFF
--- a/go-controller/pkg/config/config.go
+++ b/go-controller/pkg/config/config.go
@@ -225,9 +225,6 @@ var (
 	// UnprivilegedMode allows ovnkube-node to run without SYS_ADMIN capability, by performing interface setup in the CNI plugin
 	UnprivilegedMode bool
 
-	// EnableMulticast enables multicast support between the pods within the same namespace
-	EnableMulticast bool
-
 	// IPv4Mode captures whether we are using IPv4 for OVN logical topology. (ie, single-stack IPv4 or dual-stack)
 	IPv4Mode bool
 
@@ -505,6 +502,8 @@ type OVNKubernetesFeatureConfig struct {
 	EnableObservability             bool `gcfg:"enable-observability"`
 	EnableNetworkQoS                bool `gcfg:"enable-network-qos"`
 	AllowICMPNetworkPolicy          bool `gcfg:"allow-icmp-network-policy"`
+	// EnableMulticast enables multicast support between the pods within the same namespace
+	EnableMulticast bool `gcfg:"enable-multicast"`
 	// This feature requires a kernel fix https://github.com/torvalds/linux/commit/7f3287db654395f9c5ddd246325ff7889f550286
 	// to work on a kind cluster. Flag allows to disable it for current CI, will be turned on when github runners have this fix.
 	AdvertisedUDNIsolationMode string `gcfg:"advertised-udn-isolation-mode"`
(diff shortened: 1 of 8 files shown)


weliang1:
PR #6121 (fix-enable-multicast-config-scope) is verified working on a KinD cluster:                                                                                        
Ran 2 of 856 Specs in 81.700 seconds                                                                                                                                       
SUCCESS! -- 2 Passed | 0 Failed | 2 Pending | 852 Skipped    
  ┌────────────────────────────────────────────────────────────┬──────────────┐                                                                                              
  │                            Test                            │    Result    │                                                                                              
  ├────────────────────────────────────────────────────────────┼──────────────┤                                                                                              
  │ should be able to send multicast UDP traffic between nodes │ PASSED (20s) │
  ├────────────────────────────────────────────────────────────┼──────────────┤                                                                                              
  │ should be able to receive multicast IGMP query             │ PASSED (61s) │                                                                                              
  └────────────────────────────────────────────────────────────┴──────────────┘

go-controller/pkg/config/config.go
```diff
@@ -225,9 +225,6 @@ var (
 	// UnprivilegedMode allows ovnkube-node to run without SYS_ADMIN capability, by performing interface setup in the CNI plugin
 	UnprivilegedMode bool
 
-	// EnableMulticast enables multicast support between the pods within the same namespace
-	EnableMulticast bool
```
tssurya, line 229: while this PR is accurate in the sense that feature flags should belong to feature section
unfortunately we can't move an existing flag since that's going to break existing users having config maps where multicast was in the global option - we break backwards compatibility

the correct way to do this is 

- add multicast also in features section
- for 1 release keep flag in both sections
- for 2nd release announce deprecation of the option from global section
- for 3rd release remove it and keep it only in new section

if its really worth all that effort or not we need to consider

PR #6135, opened 2026-03-19
Switch to go modules cache; remove vendor/ directory; generate LICENSES/ subdir will all legal files for all dependencies

booxter:
## 📑 Description

Go modules are GA since 1.13. It's better to rely on the module cache instead of vendor/ dir because it allows to share dependencies across projects and avoid including generated files in our repo.

To keep maintaining license info for all dependencies in-tree, we are now collecting the relevant license / notice files under LICENSES/ directory for easy inspection.

### Motivation

The original primary motivation to drop `vendor/` directory was to avoid shipping third-party code in tree in hope this allows us to avoid producing a top-dir THIRD-PARTY-NOTICES file (or its equivalent) - which is the requirement that is imposed on us by Legal. As we later discussed in Slack, it doesn't, and we want to delete `vendor/` directory for other reasons, we need to collect all LICENSE / NOTICE files in-tree.

I think there are legit benefits of using Go modules cache instead of `vendor/` that are technical (lean git tree, cache reuse across projects), so this PR pursues the `vendor/` deletion **on these technical grounds**.

Since we'll have to generate `THIRD-PARTY-NOTICES` or its equivalent **regardless** of whether we drop `vendor/` or not, this PR now also includes a generator script that will collect all relevant license files under `LICENSES/` (this is similar to what Kubernetes does, except they rely on `vendor/`).

These files can then be easily included whenever we ship our binaries (e.g. inside container images) to comply with corresponding license requirements, which usually require that appropriate attribution is included whenever source or binary code is distributed. (This PR leaves it out of scope to include these files in our `Dockerfile` definitions to simplify review - but I'd like to pursue this if/when this PR merges.)

Note: currently the complete `LICENSES/` directory stands at 2MB, which is a lot, but I think we may have to consider including these in our images nevertheless, to stay on the path of compliance. If you have thoughts on this, let me know.

### Q&A

Q: Will my IDE not be able to find symbols from libraries?
A: It should just work™️ because your local machine has Go cache.

Q: What if upstream removes the library? What if `github.com` goes down? Our builds will break!
A: No. `go get` uses the [official](https://proxy.golang.org/) Go proxy service.

Q: I'm in China and I don't have access to Go proxy service.
A: [There are official alternatives](https://github.com/golang/go/issues/31755#issuecomment-527611727).

Q: What if Go proxy service goes down? Our builds will break!
A: **Sadly yes**. One may work around it by setting `GOPROXY=direct` when running builds.

Q: I have a private fork of the project that pulls libraries from private repos. It will break my build!
A: Sure. While it's not a concern for the upstream project, you can make use of `GOPROXY=` and `GOPRIVATE` variables to control `go get` behavior. You may also consider bringing up your own Go proxy service for internal purposes.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Chores**
  * Switched CI/build to read-only Go module mode and consolidated module verification steps.
  * Removed vendor exclusions from linting/formatting and adjusted formatting/verification checks.
  * Expanded auto-review scope by removing a prior vendor exclusion.
* **Documentation**
  * Updated developer guide for mock generation to use dependency-package paths and regenerate workflow.


ormergi:
I like this PR but there is something I dont understand, why do you remove vendor/ dir altogether? 
Its crucial for allowing builds in cases where some vendored dependency is gone, and missing from local cache.

booxter:
@ormergi Go has a proxy cache for exact this reason: https://proxy.golang.org/

btw if you don't support removing `vendor` directory, not sure what's to like in this PR. It's its whole point. :)

ormergi:
Seem I got this wrong then..
Committing Go's vendor/ dir seem very opinionated topic, lets see what maintainers think.

tssurya:
Some open questions in u/s meeting:

1. write the motivation as well on the PR (I think this was something for d/s licenses was it? its not really doing it because golang has had support for quiet some time..I forget what the d/s motivation/use case was for)
2. any way to check licenses for existing vendoring? isn't this to some extent CNCF requirement? - some tooling checks this
3. let's give the adopters and folks running ovnk in d/s some time to test this change out
4. let's plan this for 1.4 release post 30th April -added to 1.4 milestone
5. BTW - Saying internet is required to be able to run the tests might not be acceptable always cc @dave-tucker

booxter:
I've included motivation in the PR description. Since our discussion in Slack today, the motivation to delete the `vendor/` directory becomes technical, not licensing related.

But also - the discussion revealed that we may want to generate a top level licensing info for all dependencies - to include in binary artifacts or for inspection purposes - in the rootdir, esp. if we drop `vendor/`. This PR implements this too with a local script that doesn't require external services or tools. In the future, we may want to include these files in `Dockerfile`s by default, but we can discuss this later.

> BTW - Saying internet is required to be able to run the tests might not be acceptable always

I am not sure this should be an issue. If an air-gapped environment is required, one can run a local goproxy or preprovision go module cache for the execution environment. This seems very niche.

booxter:
One issue with the symlink approach is that github shows them not in a way that we'd probably like them to be: when one clicks a symlink, instead of seeing the final path, we see:

image

I can revert back to maintaining regular files but it will blow up the size of the directory from ~2MB to ~6MB. If we want to ship the licenses in container images (as apparently we should to maintain compliance), this may be an important consideration.

booxter:
Switched back to plain files. It's more space taken but easier on the reader. When we decide to pack these files into images, we can deduplicate then just for image scenario.

dave-tucker (dismissed):
Approach LGTM, and the end result LGTM also.
I have not reviewed the script that generates the licenses, and I don't have strong opinions on whether or not the vendor directory is kept, or removed but that's not for me to decide anyway.

dave-tucker:
needs @tssurya or @jcaamano to ack really as to whether this is good to merge, or whether it should wait as this could be disruptive to work in flight

jcaamano:
I think our downstream needs to be ready before we can merge this, let's hold it for a bit please.

tssurya:
As agreed in upstream meeting this is 1.4 milestone..
So we can merge this post April 30th

tssurya:
Its written in this comment https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6135#issuecomment-4155841347 and everyone was ok to target this for 1.4 ... so let's get this in early May ?

booxter:
I will split out license collection work into another PR so that it can be reviewed and merged independently of vendor/ work.

booxter:
Split out license collection (and inclusion of the files into container images) here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6269

PR #6139, opened 2026-03-19
CNC: update to use RegisterNADReconciler to avoid races

### LINKED ISSUE #5808: Move away from NADControllers to using NAD Handler Registers from Network Manager
### What is this issue about?

See https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5800 for details.

TLDR
here: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5707 we did https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5707/commits/7896a229f710dde3341f9619c1a895973e6921a5 a full nad controller that reacts to events for add/delete and create

what we want to do is what's proposed in PR #5800 and register a handler and don't do any duplicate parsing

npinaeva:
CNC uses network manager to get NAD info, but when using independent NAD handler there is race (which we used to solve by retrying), where networkManager hasn't handled the same NAD event yet and returns error. It is more reliable to receive NAD events from the networkManager. Also updated UT to use real networkManager instead of fake, it ensures more reliable event handling a better integration testing

There was a real race in mustProcessCNCForNAD where we just ignore NAD event if we couldn't get active network from the network manager. We should've returned error there and retried until networkManager finished NAD handling, but with the new approach it is not a problem anymore.

Another problem I found in UT: namespace informer may not have a namespace when NAD event from that namespace is handled. Not sure how common it is in real clusters, but I don't think it is promised anywhere that namespace informer will have a namespace by the time any of the namespaced objects are propagated.

## 📑 Description

Fixes #5808

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Refactor**
  * Simplified network-attachment discovery and NAD reconciliation; registration/deregistration now use deferred cleanup with reduced noisy error handling and no immediate registration failures.
  * Controllers now rely on centralized network info, tolerate missing network entries, and requeue affected resources as needed.
* **Tests**
  * Updated and removed tests to match the new reconciliation and lifecycle behavior; test setups adjusted for the controller-backed network manager.


tssurya:
started reviewing this,


> CNC uses network manager to get NAD info, but when using independent NAD handler there is race (which we used to solve by retrying), where networkManager hasn't handled the same NAD event yet and returns error. It is more reliable to receive NAD events from the networkManager. Also updated UT to use real networkManager instead of fake, it ensures more reliable event handling a better integration testing


this is awesome!

a few questions on the PR describe to ensure I understand it correctly


> There was a real race in mustProcessCNCForNAD where we just ignore NAD event if we couldn't get active network from the network manager. We should've returned error there and retried until networkManager finished NAD handling, but with the new approach it is not a problem anymore.

oh! so does that mean we were never programming things in such a scenario? because if we didn't do a retry, how were/are we passing in CI currently, isn't this a major problem? 


> Another problem I found in UT: namespace informer may not have a namespace when NAD event from that namespace is handled. Not sure how common it is in real clusters, but I don't think it is promised anywhere that namespace informer will have a namespace by the time any of the namespaced objects are propagated.


This was the AI comment in your previous PR right? - and yea its not expected for the informers to prepopulate anything in UT, what was the fix in the end for this?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -392,9 +374,10 @@ func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetwo
 				if !isCUDN {
 					continue
 				}
-				network, err := util.ParseNADInfo(nad)
-				if err != nil {
-					errs = append(errs, fmt.Errorf("failed to parse NAD %s/%s: %w", nad.Namespace, nad.Name, err))
+				network := c.networkManager.GetNetInfoForNADKey(types.NamespacedName{Namespace: nad.Namespace, Name: nad.Name}.String())
+				if network == nil {
+					// if network doesn't exist in networkmanager, we will get an update later
```
tssurya, line 379: so I have one regarding regarding the workings of the nadreconciler
in some sense its an "event driven controller right" ?
i.e we are listening for events and the other part - network manager sends us NAD events? - what happens if the event is missed? i.e is it like our older event driven controllers where we need to worry about such a scenario?

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
```diff
@@ -392,9 +374,10 @@ func (c *Controller) discoverSelectedNetworks(cnc *networkconnectv1.ClusterNetwo
 				if !isCUDN {
 					continue
 				}
-				network, err := util.ParseNADInfo(nad)
```
tssurya, line 395: so we are already getting a parsed NAD from reconciler so that's really good..

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -128,17 +126,15 @@ func NewController(
...
-		Informer:       wf.NADInformer().Informer(),
-		Lister:         nadLister.List,
-		Reconcile:      c.reconcileNAD,
-		ObjNeedsUpdate: nadNeedsUpdate,
-		Threadiness:    1,
+	nadReconcilerCfg := &controllerutil.ReconcilerConfig{
+		RateLimiter: workqueue.DefaultTypedControllerRateLimiter[string](),
+		Reconcile:   c.reconcileNAD,
+		Threadiness: 1,
+		MaxAttempts: controllerutil.InfiniteAttempts,
```
tssurya, line 133: I'm fine to keep it infinite attempts but I hope none of our errors are things that will never resolve and we keep retrying category

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -159,10 +155,16 @@ func NewController(
 
 func (c *Controller) Start() error {
 	defer klog.Infof("Cluster manager network connect controllers started")
+
+	id, err := c.networkManager.RegisterNADReconciler(c.nadReconciler)
+	if err != nil {
+		return err
```
tssurya: do we want to wrap the error into "this is coming from network connect controller" ?

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -159,10 +155,16 @@ func NewController(
...
+	}
+	c.nadReconcilerID = id
 	return controllerutil.StartWithInitialSync(
 		c.initialSync,
 		c.cncController,
-		c.nadController,
+		c.nadReconciler,
 		c.namespaceController,
 	)
```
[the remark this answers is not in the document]
tssurya: @npinaeva wanna address this one?

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -263,11 +265,18 @@ func (c *Controller) initialSync() error {
 }
 
 func (c *Controller) Stop() {
+	if c.nadReconcilerID != 0 {
+		if err := c.networkManager.DeRegisterNADReconciler(c.nadReconcilerID); err != nil {
+			klog.Warningf("clustermanager-network-connect: failed to deregister NAD reconciler: %v", err)
```
tssurya: I think returning error here is not very useful is it? cause retrying deregister twice is not gonna succeed so warning is ok I guess?

go-controller/pkg/clustermanager/networkconnect/controller.go
tssurya, line 419: `There was a real race in mustProcessCNCForNAD where we just ignore NAD event if we couldn't get active network from the network manager. We should've returned error there and retried`

is this the bug? where we weren't retrying any of the errors? I can't remember why we didn't decide to aggregate the errors from each namespace.. maybe that's a hindsight issue on my part..

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -286,10 +295,30 @@ func cncNeedsUpdate(oldObj, newObj *networkconnectv1.ClusterNetworkConnect) bool
...
+
+	nad, err := c.nadLister.NetworkAttachmentDefinitions(namespace).Get(name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get NAD %s: %w", key, err)
+	}
+
 	nadSupported := func(nad *nadv1.NetworkAttachmentDefinition) bool {
 		if nad == nil {
-			return false
+			return true
```
tssurya, line 322: hmm why is nil nad returned as supported nad?

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -312,45 +342,8 @@ func nadNeedsUpdate(oldObj, newObj *nadv1.NetworkAttachmentDefinition) bool {
...
-	oldNADLabels := labels.Set(oldObj.Labels)
-	newNADLabels := labels.Set(newObj.Labels)
-	labelsChanged := !labels.Equals(oldNADLabels, newNADLabels)
-	// safe spot check for ovn network id annotation add happening as update to NADs
-	annotationsChanged := oldObj.Annotations[ovntypes.OvnNetworkIDAnnotation] != newObj.Annotations[ovntypes.OvnNetworkIDAnnotation]
-
-	// CASE3: NAD is being updated (UDN or CUDN is being updated)
-	//  3.1: NAD network id annotation changed
-	//  3.2: NAD labels changed (only relevant for CUDNs->NADs)
-	return labelsChanged || annotationsChanged
```
tssurya, line 332: ah I guess the reconciler pattern doesn't have a function spec that can be used for filtering out things like we have for the controller pattern....?
so are we reacting to all NAD events? or is there some level of filtering that can be done on the reconciler to discard the events we don't care about?

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -458,12 +451,11 @@ func (c *Controller) mustProcessCNCForNAD(nad *nadv1.NetworkAttachmentDefinition
 }
 
 func namespaceNeedsUpdate(oldObj, newObj *corev1.Namespace) bool {
-	// Case 1: namespace is being deleted
-	// Case 2: namespace is being created
-	// for both these cases, we don't care because
-	// we only care about UDNs being created or deleted
-	// in those namespaces which is already handled by the NAD controller
-	if oldObj == nil || newObj == nil {
+	// When searching for primary UDNs matching CNC selector, we list namespaces.
```
tssurya, line 455: I think this is the unit test issue fix?

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -458,12 +451,11 @@ func (c *Controller) mustProcessCNCForNAD(nad *nadv1.NetworkAttachmentDefinition
...
+	// Very rarely a NAD event can be handled, while namespace informer doesn't have
```
tssurya, line 456: oh wow good catch! I didn't think about this corner case scenario being possible... - its almost not supposed to happen right? cause if we get the NAD event then it means the namespace must exist? - could you shed some light on how is the NAD existing on a non-existent namespace?

go-controller/pkg/clustermanager/networkconnect/controller.go
tssurya: nit: hmm maybe add back case1/case2 to your comments or remove the case3 from here...

tssurya (changes requested):

go-controller/pkg/clustermanager/networkconnect/controller_test.go
tssurya, line 1: very cool we got rid of the fake NM..

go-controller/pkg/clustermanager/networkconnect/cluster_network_connect.go
npinaeva: yeah our own `c.nadReconciler` is an event-driven reconciler, which means it can get events from anywhere, while normal controllers get events from the informer. So when we wire it to the networkManager, we get events from there, but the retries and level-drivenness is the same as it was before, just the source of the events has changed.

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: well I think it actually may be in case a selected network is of the wrong type or something else is wrong with the selected network, but the rate limiter will eventually reach ~16 min (1K seconds) retry interval, so shouldn't be too bad.
But I think it is needed, because we have overlap checks with the other CNCs, so it may be that CNC A stops having an error simply because CNC B stopped selecting some conflicting network, but if we have a limited retry we won't trigger a reconcile in this case

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: hm fun fact: I have realized that function never actually returns an error, fixing it in another commit

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: I am removing error return from `DeRegisterNADReconciler` completely

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: yeah it doesn't, but we hoped it is still ok because NADs don't really change often

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: this is to handle delete events. With the level-driven controller delete events are always handled and `nadNeedsUpdate` is not event called, but not we don't have that anymore, so I had to adjust thin function while moving it to the reconcile itself

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: it was a bug until we switched to the networkManager events, now the race doesn't exist anymore

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: I was thinking about that, but I am not sure if it is guaranteed anywhere that the namespace informer has to be updated before any of the namespaced object informers are updated. I don't think it is the case, and those informers are independent, so I think even in real clusters it can happen

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: yeah it is all about the informers updates: so the namespace does exist, but the namespace informer may not have it yet

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -157,12 +153,21 @@ func NewController(
...
+	defer func() {
+		if err != nil {
+			c.networkManager.DeRegisterNADReconciler(c.nadReconcilerID)
+		} else {
+			klog.Infof("Cluster manager network connect controllers started")
+		}
+	}()
+
+	id := c.networkManager.RegisterNADReconciler(c.nadReconciler)
+	c.nadReconcilerID = id
```
tssurya, line 166: nit: assign it straight to c.nadReconcilerID

go-controller/pkg/clustermanager/networkconnect/controller.go
tssurya: ah ack I see, I hope all the functions are resilient to nil NAD from high level mustProcessCNCForNAD seems to handle it

go-controller/pkg/networkmanager/nad_controller.go
```diff
@@ -414,18 +404,17 @@ func (c *nadController) RegisterNADReconciler(r NADReconciler) (uint64, error) {
 	c.nextReconcilerID++
 	id := c.nextReconcilerID
 	c.reconcilers[id] = reconcilerRegistration{id: id, r: r}
-	return id, nil
```
tssurya, line 417: lol so many error wrappings throughout for nil error (joy)

go-controller/pkg/clustermanager/networkconnect/controller.go
```diff
@@ -157,12 +153,21 @@ func NewController(
 	return c
 }
 
-func (c *Controller) Start() error {
-	defer klog.Infof("Cluster manager network connect controllers started")
+func (c *Controller) Start() (err error) {
```
tssurya, line 156: do we still need this err variable ? it sounds like based on what you found RegisterNADReconciler won't return error so we are not using this variable to store any errors are we? though we are using it in defer func...
same comment also for the other controllers..

tssurya (changes requested):

go-controller/pkg/clustermanager/networkconnect/controller.go
npinaeva: yeah using a named variable lets up check if `StartWithInitialSync` returned an error before calling defer

npinaeva:
reported https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6280 will see if retest helps

tssurya (approved):

Issue #6140, opened 2026-03-19
flake: e2e ingress traffic validation / Should be allowed by externalip services to a new node ip


booxter:
### Which jobs are flaking?

Observed on March 19, 2026 in the same PR run:

- https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/23298233534/job/67752937352?pr=6135
  - `e2e (control-plane, noHA, local, ipv4, noSnatGW, 1br, ic-single-node-zones, enable-dns-name-resolver)`
- https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/23298233534/job/67752937355?pr=6135
  - `e2e (control-plane, noHA, shared, ipv4, SnatGW, 1br, ic-single-node-zones, disable-forwarding)`

Both jobs hit the same failure signature for the same test.

### Which tests are flaking?

```text
e2e ingress traffic validation Validating ingress traffic to manually added node IPs [BeforeEach] Should be allowed by externalip services to a new node ip
[BeforeEach] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:1605
[It] /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:1690
```

Final reported failure in both jobs:

```text
[FAILED] failed to add new Addresses to node ovn-control-plane: failed to run "ip addr add 172.18.1.1 dev breth0": exit status 2 (Error: ipv4: Address already assigned.)
In [BeforeEach] at: /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:1668
```

Before that, both jobs show the same earlier sequence:

```text
FAILED Response was {"errors":["Get \"http://172.18.1.1:81/hostname\": dial tcp 172.18.1.1:81: connect: connection refused"]}
[FAILED] in [It] - /home/runner/work/ovn-kubernetes/ovn-kubernetes/test/e2e/e2e.go:1727
failed to add new Addresses to node ... failed to run "ip addr del 172.18.1.3 dev breth0" ... Address not found.
```

### Since when has it been flaking?

First noticed on March 19, 2026.

I found the same signature twice in the same run, but I did not find an existing issue for this exact failure.

### Reason for failure (if possible)

Inference from the logs and current source in `test/e2e/e2e.go`:

The test first fails in the body with `connect: connection refused` to `172.18.1.1:81`.

[...]

PR #6142, opened 2026-03-19
e2e: fix invalid address cleanup due to invalid variable closure

booxter:
## 📑 Description

providerCtx.AddCleanUpFn defined a cleanup function that was referring
to a variable that was defined outside of a loop, meaning that it was
always calling each registered cleanup handler on the same (last) value
of the newIP variable, leaving some IP addresses not deleted.

Related #6140

Marking as Related, not Fixed because it will only fix cleanup but the
underlying connection refused error is still there.

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/test/e2e/e2e.go
+++ b/test/e2e/e2e.go
@@ -1698,16 +1698,17 @@ var _ = ginkgo.Describe("e2e ingress traffic validation", func() {
 				} else {
 					newIP = "172.18.1." + strconv.Itoa(i+1)
 				}
+				newCIDR := newIP + util.GetIPFullMaskString(newIP)
 				// manually add the a secondary IP to each node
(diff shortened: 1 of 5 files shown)


kyrtapz (commented):
Thanks @booxter! 
Overall lgtm.
I was looking at this in the past without pushing anything and one thing I observed was the ip addr warning. 
Would you mind wrapping the addresses with `util.GetIPFullMaskString` or just adding the full mask to the string? It would avoid this:
```
Warning: Executing wildcard deletion to stay compatible with old scripts.
     Explicitly specify the prefix length (172.22.52.254/32) to avoid this warning.
     This special behaviour is likely to disappear in further releases,
     fix your scripts!
```

Issue #5531, opened 2025-08-28
pod in a UDN cannot access a nodePort service in a same UDN network with externalTrafficPolicy=Local on different node


zhaozhanqi:
### What happened?

1. apply cudn l3 and nodeport service with ETP=Local.

kubectl get svc -n udn-ns3 
NAME           TYPE       CLUSTER-IP       EXTERNAL-IP   PORT(S)           AGE
test-service   NodePort   172.30.149.254   <none>        27017:32226/TCP   26h

2.  there is backend pod on node1
3. Now from pod on node2 cannot access the nodeport service
$ curl 192.168.111.23:32226
curl: (56) Recv failure: Connection reset by peer

### What did you expect to happen?

pod in a UDN should be able to access a nodePort service in a different UDN network with externalTrafficPolicy=Local on different node

### How can we reproduce it (as minimally and precisely as possible)?

1. Setup cluster with LGW mode
2. Create UDN namespace and nodePort service with etp=local and backend pod
3. Create another client pod in same UDN namespace on different node
4.  from the client pod to access the nodePort service.

### OVN-Kubernetes version

<details>

```console
# kubectl exec -n ovn-kubernetes ovnkube-node-kfj6f -- ovnkube --version
Defaulted container "nb-ovsdb" out of: nb-ovsdb, sb-ovsdb, ovn-northd, ovnkube-controller, ovn-controller, ovs-metrics-exporter
Version: 1.1.0
Git commit: 814977fdc13ed470b1ddf04d1e29da28d8679ab9
Go version: go1.24.5
Build date: 2025-08-19
OS/Arch: linux amd64

```

</details>

### Kubernetes version

<details>

```console
# kubectl version
Client Version: v1.32.3
Kustomize Version: v5.5.0
Server Version: v1.33.1

```

</details>

### OVN version

<details>

```console
kubectl exec -n ovn-kubernetes ovnkube-node-kfj6f -- rpm -q ovn
Defaulted container "nb-ovsdb" out of: nb-ovsdb, sb-ovsdb, ovn-northd, ovnkube-controller, ovn-controller, ovs-metrics-exporter
ovn-25.03.1-42.fc42.x86_64

```

</details>

### OVS version

<details>

```console
# kubectl exec -n ovn-kubernetes ovnkube-node-kfj6f -- rpm -q openvswitch
[...]

PR #6144, opened 2026-03-20
Add CT commit all to UDN transit/cluster routers

trozet:
For UDN we have a conditional SNAT for traffic that goes via mpx
    interface. This conditional SNAT is placed on the ovn_cluster_router or
    transit_router for L3 or L2, respectively. Due to this conditional SNAT,
    ingress traffic into the router must be checked for unSNAT, and without
    commit can break ingress reply traffic as sessions are not tracked
    correctly. Additionally without committing, this breaks hardware
    offload.
    
    Fixes: #5531

## Summary by CodeRabbit

* **New Features**
  * Enhanced stateful connection tracking capabilities for logical routers handling traffic in user-defined networks

* **Improvements**
  * Simplified cluster router configuration to more flexibly support multicast and connection tracking options

* **Tests**
  * Expanded end-to-end test coverage for NodePort services in user-defined networks across multiple nodes

### DIFF
--- a/dist/images/Dockerfile.fedora
+++ b/dist/images/Dockerfile.fedora
@@ -11,7 +11,8 @@
 # This is for a development build where the ovn-kubernetes utilities
 # are built locally and included in the image (instead of the rpm)
 #
-ARG OVN_FROM=koji
+# HACK: force OVN builds to pick up the ct commit all fix until it lands upstream.
+ARG OVN_FROM=source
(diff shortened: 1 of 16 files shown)


trozet:
Need to validate this fixes HW offload in downstream first. @crnithya fyi

trozet:
from downstream hardware offload testing this did not seem to solve the problem, and made other traffic flows drop in perf... closing

PR #6145, opened 2026-03-20
ubuntu, fedora: Miscellaneous improvements to Dockerfiles

booxter:
## 📑 Description

Miscellaneous cleanups for our Dockerfiles:
- remove socat from images (was only used in e2e tests - switched those to native poke function)
- build ovnkube hermetically for ubuntu inside docker (same as fedora)
- reduce number of layers
- align kubectl version with what's in go.mod
- get ubuntu and fedora closer to each other
- update ML and container descriptions

## ✅ Checks

- [ ] My code requires tests


dist/images/Dockerfile.ubuntu
```diff
@@ -28,6 +28,8 @@ RUN cd ovn-kubernetes/dist/images && \
 
 FROM ubuntu:26.04
 
+ARG TARGETARCH
```
trozet: nice!

dist/images/Dockerfile.fedora
```diff
@@ -172,7 +172,7 @@ COPY --from=ovnkube-builder /workspace/ovn-kubernetes/dist/images/iptables-scrip
 
 LABEL io.k8s.display-name="ovn-kubernetes" \
       io.k8s.description="This is a Kubernetes network plugin that provides an overlay network using OVN." \
-      maintainer="Tim Rozet <trozet@redhat.com>"
```
trozet, line 182: :)

dist/images/Dockerfile.fedora
```diff
@@ -144,10 +144,10 @@ ENV PYTHONDONTWRITEBYTECODE yes
 
 # install needed rpms - openvswitch must be 2.10.4 or higher
 RUN INSTALL_PKGS=" \
-    python3-pip python3-pyyaml bind-utils procps-ng openssl numactl-libs firewalld-filesystem \
+    python3-pyyaml bind-utils procps-ng openssl numactl-libs \
     libpcap hostname kubernetes-client util-linux \
-    ovn ovn-central ovn-host python3-openvswitch tcpdump openvswitch-test python3-pyOpenSSL \
```
trozet, line 148: these debug tools like tcpdump were added intentionally so that we can debug from the OVNK container without having to launch another pod for debugging on a host system where we cannot get access. OCP also builds with these debug tools in their OVNK image as well I think. Do you think this is a problem?

trozet: I think this might also break this e2e diagnostic mode: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/test/e2e/diagnostics/diagnostics.go#L12

dist/images/Dockerfile.ubuntu
```diff
@@ -11,49 +11,68 @@
...
 # So this file will change over time.
 
-FROM ubuntu:26.04
+ARG BUILDER_IMAGE
+ARG OVN_KUBERNETES_DIR=.
+ARG KUBECTL_MINOR=1.35
+ARG UBUNTU_BASE_IMAGE=ubuntu:26.04
 
-USER root
+FROM --platform=${BUILDPLATFORM} ${BUILDER_IMAGE} AS ovnkube-builder
```
trozet, line 19: AI says:

[P1] Update the Ubuntu image-publishing workflow for the new build context
The push-only Ubuntu image workflow in .github/workflows/docker.yml still builds this Dockerfile with context: ./dist/images and does not pass BUILDER_IMAGE.
As a result, the build fails at FROM --platform=${BUILDPLATFORM} ${BUILDER_IMAGE} because BUILDER_IMAGE is empty. Even if that argument is added, COPY ${OVN_KUBERNETES_DIR} ovn-kubernetes would copy only the dist/images directory, so the subsequent cd ovn-kubernetes/dist/images would still fail because that nested path does not exist.
Please update the workflow to use the repository root as its build context and pass BUILDER_IMAGE and the KUBECTL_MINOR derived from go-controller/go.mod, matching the new Makefile invocation. PR CI does not catch this because .github/workflows/docker.yml runs only on pushes to master and release branches, so the failure would occur after merge when publishing the Ubuntu image.

dist/images/Dockerfile.fedora
booxter: There are a few problems with putting debug tools in an image: 1. it increases security surface to care about (think of a SBOM marked as vulnerable when a `socat` or `tcpdump` CVE is reported). 2. it bloats the images which puts burden on each node that runs them (though the images are already quite bloated, 1GB+ a pop, and a few megabytes saved here are nothing).

---

A standard means to debug a pod connectivity is to [`kubectl debug`](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_debug/) that allows to start a debug container with necessary tools in the same scope as the target pod. This should solve production debugging scenario.

As to e2e diagnostics, I wasn't aware of this mechanism. Do you think this mechanism should use OVN image and not e.g. netboot that I think is used elsewhere in e2e whenever we need `tcpdump`? That it uses OVN image seems accidental, but I may miss something.

---

But if we still need `tcpdump` and some other tools in images, we should then add them to both images. For example, Ubuntu lacks `tcpdump`.

dist/images/Dockerfile.fedora
```diff
@@ -143,43 +144,35 @@ ENV PYTHONDONTWRITEBYTECODE yes
 
 # install needed rpms - openvswitch must be 2.10.4 or higher
 RUN INSTALL_PKGS=" \
-    python3-pip python3-pyyaml bind-utils procps-ng openssl numactl-libs firewalld-filesystem \
+    python3-pyyaml bind-utils procps-ng openssl numactl-libs \
     libpcap hostname kubernetes-client util-linux \
```
booxter, line 150: (Observation) kubectl version here is not synced with go.mod, in contrast to ubuntu dockerfile. We may want to use same mechanism (extract version from go.mod and force it here) to put them in sync.

dist/images/Dockerfile.fedora
booxter: Gonna retain the tools + add them to ubuntu for symmetry in the next push.

dist/images/Dockerfile.fedora
booxter: Will add a comment to this effect.

dist/images/Dockerfile.ubuntu
booxter: Indeed. Missed it because ubuntu CI support was added after the series was written.

trozet (approved):

PR #6157, opened 2026-03-24
ci: run a control-plane lane with ubuntu images

### LINKED ISSUE #6146: Run at least one e2e job using ubuntu images
### What would you like to see added/updated/deleted in our end-to-end tests?

I'd like to see at least some e2e jobs running ubuntu images instead of fedora.

### Why is this needed?

Right now ubuntu dockerfiles are not tested in any way beyond that they build. This makes changes to the image definition risky since they are not properly utilized in anything that resemble real clusters.

Such job will give developers some more confidence when touching ubuntu images, as well as improve our test coverage across multiple distribution families (not just redhat based).

booxter:
## 📑 Description

This should give some more confidence in changes to ubuntu dockerfiles,
and in general in that ovn-kubernetes services work if built on top of
ubuntu base images.

Fixes #6146

## ✅ Checks

- [ ] My code requires tests

## How to verify it

The job that switched to ubuntu image should pass. I also confirmed locally as follows:

```
export OVN_IMAGE_FAMILY=ubuntu
export OVN_IMAGE=local

./contrib/kind.sh \
  -cn ovn-ubuntu-test \
  -gm local \
  -ds \
  -ic \
  -ho \
  -me \
  -el \
  -nqe \
  -mlb \
  -ii \
  -mps

kubectl -n ovn-kubernetes get pods \
  -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{range .spec.containers[*]}{.image}{" "}{end}{"\n"}{end}' \
  | grep 'ovn-daemonset-ubuntu'
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -165,34 +165,42 @@ jobs:
   build-pr:
     name: Build-PR
     runs-on: ubuntu-24.04
+    env:
+      PR_IMAGE_FAMILY: fedora
+      PR_IMAGE: ovn-daemonset-fedora:pr
+      PR_IMAGE_TARGET: fedora-image
+      PR_IMAGE_ARTIFACT: test-image-pr
     steps:
-    # Create a cache for the build PR image
-    - name: Restore PR image cache
-      id: image_cache_pr
+    # These anchored steps are shared with build-pr-ubuntu below.
+    - &restore_pr_image_cache
+      name: Restore PR image cache
       uses: actions/cache@v4
       with:
         path: |
           ${{ env.CI_IMAGE_CACHE }}
-        key: ${{ github.run_id }}-image-cache-pr
+        key: ${{ github.run_id }}-image-cache-pr-${{ env.PR_IMAGE_FAMILY }}
 
-    - name: Check if PR image build is needed
+    - &check_pr_image_build
+      name: Check if PR image build is needed
       id: is_pr_image_build_needed
       continue-on-error: true
       run: |
         set -x
-        if [ -f ${CI_IMAGE_CACHE}/${CI_IMAGE_PR_TAR}.gz ]; then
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -502,12 +526,17 @@ jobs:
           - {"target": "traffic-flow-test-only", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4", "disable-snat-multiple-gws": "noSnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "traffic-flow-tests": "1-24", "network-segmentation": "enable-network-segmentation"}
           - {"target": "tools", "ha": "noHA", "gateway-mode": "local", "ipfamily": "dualstack", "disable-snat-multiple-gws": "SnatGW", "second-bridge": "1br", "ic": "ic-single-node-zones", "network-segmentation": "enable-network-segmentation"}
           - {"target": "serial", "ha": "noHA", "gateway-mode": "shared", "ipfamily": "ipv4", "disable-snat-multiple-gws": "snatGW", "second-bridge": "1br", "ic": "ic-single-node-zones"}
-    needs: [ build-pr ]
+    needs: [ build-pr, build-pr-ubuntu ]
+    if: ${{ !cancelled() }}
     env:
       JOB_NAME: "${{ matrix.target }}-${{ matrix.ha }}-${{ matrix.gateway-mode }}-${{ matrix.ipfamily }}-${{ matrix.disable-snat-multiple-gws }}-${{ matrix.second-bridge }}-${{ matrix.ic }}"
```
trozet: do we want to add the image_family to the job name here? Maybe we should also remove matrix.ha. Not sure what give us anymore.

booxter: Added. `ha` is out of scope.

trozet (approved):

PR #6158, opened 2026-03-25
no-overlay: add CUDN support for no-overlay managed routing 

pliurh:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Added management of cluster user-defined networks in BGP routing with guaranteed initial full convergence on startup.

* **Improvements**
  * Ensured network label propagation so routing selections align with user-defined networks.
  * Made managed routing and base configuration cleanup/deletion and reconciliation flows more deterministic and robust.

* **Tests**
  * Expanded coverage for user-defined network scenarios, drift recovery, ownership/event filtering, and controller repair behaviors.

### DIFF
--- a/go-controller/pkg/clustermanager/managedbgp/controller.go
+++ b/go-controller/pkg/clustermanager/managedbgp/controller.go
@@ -15,8 +15,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
-	"k8s.io/client-go/tools/record"
+	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
@@ -26,24 +25,32 @@ import (
 	raclientset "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/routeadvertisements/v1/apis/clientset/versioned"
 	ralisters "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/routeadvertisements/v1/apis/listers/routeadvertisements/v1"
 	apitypes "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/types"
+	userdefinednetworkv1 "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1"
+	cudnclientset "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/clientset/versioned"
+	cudnlisters "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1/apis/listers/userdefinednetwork/v1"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/factory"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
 )
 
 const (
-	// ControllerName is the name of the managed BGP controller
-	ControllerName = "managed-bgp-controller"
-	// fieldManager identifies writes performed by this controller.
-	fieldManager = "clustermanager-managedbgp-controller"
-	// FRRConfigManagedLabel is the label used to identify FRRConfigurations managed by this controller
-	FRRConfigManagedLabel = "k8s.ovn.org/managed-internal-fabric"
-	// FRRConfigManagedValue is the value used for the FRRConfigManagedLabel
-	FRRConfigManagedValue = "bgp"
-	// ManagedRANetworkLabel is the label set on managed RouteAdvertisements to identify the network they advertise
-	ManagedRANetworkLabel = "k8s.ovn.org/managed-network"
+	// controllerName is the name of the managed BGP controller
+	controllerName = "managed-bgp-controller"
+	// fieldManager identifies this controller as the server-side author of owned resources.
+	// Used with FieldManager in Create/Update options so that isOwnUpdate can
+	// filter out informer events that result from our own writes.
+	fieldManager = "clustermanager-managed-bgp-controller"
+	// frrConfigManagedLabel is the label used to identify FRRConfigurations managed by this controller
+	frrConfigManagedLabel = "k8s.ovn.org/managed-internal-fabric"
+	// frrConfigManagedValue is the value used for the frrConfigManagedLabel
+	frrConfigManagedValue = "bgp"
+	// managedRANetworkLabel is the label used to identify managed resources by network name.
+	// It is set on managed RouteAdvertisements and on CUDNs (to enable RA network selection).
+	managedRANetworkLabel = "k8s.ovn.org/managed-network"
 	// managedNamePrefix is the prefix for managed resource names
-	managedNamePrefix = "ovnk-managed-"
+	managedNamePrefix = "ovnk-managedbgp-"
+	// managedFRRConfigurationName is the static name for the managed base FRRConfiguration.
+	managedFRRConfigurationName = "ovnk-managedbgp-nooverlay"
 )
 
 func managedHashedName(s string) string {
@@ -52,32 +59,26 @@ func managedHashedName(s string) string {
 	return fmt.Sprintf("%s%x", managedNamePrefix, h.Sum64())
 }
 
-// ManagedRouteAdvertisementName returns the name of the managed RouteAdvertisement
-// for the given network name. The name follows the pattern "ovnk-managed-<hash>"
+// managedRouteAdvertisementName returns the name of the managed RouteAdvertisement
+// for the given network name. The name follows the pattern "ovnk-managedbgp-<hash>"
 // where <hash> is derived from the network name.
-func ManagedRouteAdvertisementName(networkName string) string {
+func managedRouteAdvertisementName(networkName string) string {
 	return managedHashedName(networkName)
-}
-
-// BaseFRRConfigName returns the name of the base FRRConfiguration that
-// configures the cluster internal BGP fabric. This FRRConfiguration is shared
-// by all networks in managed routing mode. The name follows the pattern
-// "ovnk-managed-<hash>" where <hash> is derived from the configured AS number.
-func BaseFRRConfigName() string {
-	return managedHashedName(fmt.Sprintf("%d", config.ManagedBGP.ASNumber))
 }
 
 // Controller manages the BGP topology for no-overlay networks with managed routing
 type Controller struct {
-	frrClient            frrclientset.Interface
-	frrLister            frrlisters.FRRConfigurationLister
-	raClient             raclientset.Interface
-	raLister             ralisters.RouteAdvertisementsLister
-	nodeController       controllerutil.Controller
-	managedRAController  controllerutil.Controller
-	managedFRRController controllerutil.Controller
-	wf                   *factory.WatchFactory
-	recorder             record.EventRecorder
+	frrClient      frrclientset.Interface
+	frrLister      frrlisters.FRRConfigurationLister
+	raClient       raclientset.Interface
+	raLister       ralisters.RouteAdvertisementsLister
+	cudnClient     cudnclientset.Interface
+	cudnLister     cudnlisters.ClusterUserDefinedNetworkLister
+	nodeController controllerutil.Controller
+	raController   controllerutil.Controller
+	frrController  controllerutil.Controller
+	cudnController controllerutil.Controller
+	wf             *factory.WatchFactory
 }
 
 // NewController creates a new managed BGP controller
@@ -85,15 +86,15 @@ func NewController(
 	wf *factory.WatchFactory,
 	frrClient frrclientset.Interface,
 	raClient raclientset.Interface,
-	recorder record.EventRecorder,
+	cudnClient cudnclientset.Interface,
 ) *Controller {
 	c := &Controller{
-		frrClient: frrClient,
-		frrLister: wf.FRRConfigurationsInformer().Lister(),
-		raClient:  raClient,
-		raLister:  wf.RouteAdvertisementsInformer().Lister(),
-		wf:        wf,
-		recorder:  recorder,
+		frrClient:  frrClient,
+		frrLister:  wf.FRRConfigurationsInformer().Lister(),
+		raClient:   raClient,
+		raLister:   wf.RouteAdvertisementsInformer().Lister(),
+		cudnClient: cudnClient,
+		wf:         wf,
 	}
 
 	nodeConfig := &controllerutil.ControllerConfig[corev1.Node]{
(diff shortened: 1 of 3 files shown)


pliurh:
@coderabbitai review

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+	// Ignore events from our own writes
+	if newRA != nil && isOwnUpdate(newRA.ManagedFields) {
+		return false
 	}
-	return nil
+	// newRA is never nil here: delete events bypass ObjNeedsUpdate.
+	return oldRA == nil ||
+		!reflect.DeepEqual(oldRA.Spec, newRA.Spec) ||
+		oldRA.Labels[ManagedRANetworkLabel] != newRA.Labels[ManagedRANetworkLabel]
 }
```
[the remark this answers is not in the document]
pliurh: If oldRA is managed but newRA is nil, that means the RA is deleted. The framework's onDelete handler (controller.go:233) directly enqueues the key without calling ObjNeedsUpdate.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
-	return c.ensureManagedBaseFRRConfiguration()
+	// Ignore events from our own writes
+	if newFRR != nil && isOwnUpdate(newFRR.ManagedFields) {
+		return false
+	}
+	// newFRR is never nil here: delete events bypass ObjNeedsUpdate.
+	return oldFRR == nil ||
+		!reflect.DeepEqual(oldFRR.Spec, newFRR.Spec) ||
+		oldFRR.Labels[FRRConfigManagedLabel] != newFRR.Labels[FRRConfigManagedLabel]
 }
```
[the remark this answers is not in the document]
pliurh: The framework's onDelete handler (controller.go:233) directly enqueues the key without calling ObjNeedsUpdate.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +116,160 @@ func NewController(
...
+// isCDNManaged returns true if the cluster default network is configured for
+// no-overlay mode with managed routing.
+func isCDNManaged() bool {
+	return config.Default.Transport == types.NetworkTransportNoOverlay &&
+		config.NoOverlay.Routing == config.NoOverlayRoutingManaged
+}
+
+// isManagedNoOverlayCUDN returns true if the given CUDN is configured for
+// no-overlay transport with managed routing.
+func isManagedNoOverlayCUDN(cudn *userdefinednetworkv1.ClusterUserDefinedNetwork) bool {
```
jcaamano: nit: `isCUDNManaged`, naming aligned with CDN above?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +116,160 @@ func NewController(
...
+	controllers := []controllerutil.Reconciler{c.nodeController, c.raController, c.frrController}
+	if c.cudnController != nil {
+		controllers = append(controllers, c.cudnController)
+	}
+	controllerutil.Stop(controllers...)
+}
+
+// syncAll performs a full reconciliation of all managed resources.
+// It ensures RAs exist for all managed networks, removes stale RAs,
+// and ensures/cleans up the base FRRConfiguration accordingly.
```
jcaamano: The synchronous initial sync needs to be as quick as possible.

Preferably, initial sync should only handle stale resources that would be not possible to handle later.

Ensuring state should be done later, when we sync all resources asynchronously. So things like `ensureManagedNetworkLabel`, `ensureManagedRouteAdvertisement`, `ensureBaseFRRConfiguration`.. can't possibly happen later on?

And even `deleteManagedRA` and `cleanupBaseFRRConfigurations` should be managed asynchornously if possible. For example, on startup we will get an async event for the stale RA, which should probably call reconcileCUDN which would delete the RA if it is not needed.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+	if isOwnUpdate(newCUDN.ManagedFields) {
+		return false
 	}
-	return !reflect.DeepEqual(oldConfig.Spec, newConfig.Spec) || !reflect.DeepEqual(oldConfig.Labels, newConfig.Labels)
-}
 
-func isOwnUpdate(managedFields []metav1.ManagedFieldsEntry) bool {
-	return util.IsLastUpdatedByManager(fieldManager, managedFields)
+	// CUDN spec is immutable, only label changes matter.
+	return !reflect.DeepEqual(oldCUDN.Labels, newCUDN.Labels)
```
jcaamano: What labels do we care about?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+		return false
 	}
-	if err := c.ensureManagedRouteAdvertisement(types.DefaultNetworkName); err != nil {
-		return fmt.Errorf("failed to ensure managed RouteAdvertisement: %w", err)
+	// Ignore events from our own writes
+	if newRA != nil && isOwnUpdate(newRA.ManagedFields) {
+		return false
 	}
-	return nil
+	// newRA is never nil here: delete events bypass ObjNeedsUpdate.
```
jcaamano: A safety check for a nil pointer would be nice though

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+func (c *Controller) frrNeedsUpdate(oldFRR, newFRR *frrtypes.FRRConfiguration) bool {
+	if !isManagedBaseFRRConfiguration(oldFRR) && !isManagedBaseFRRConfiguration(newFRR) {
+		return false
 	}
-	return c.ensureManagedBaseFRRConfiguration()
+	// Ignore events from our own writes
+	if newFRR != nil && isOwnUpdate(newFRR.ManagedFields) {
+		return false
+	}
+	// newFRR is never nil here: delete events bypass ObjNeedsUpdate.
```
jcaamano: same as above

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+	networkName, err := c.resolveNetworkForRA(key)
+	if err != nil {
+		return err
 	}
-
-	return c.ensureManagedBaseFRRConfiguration()
+	if networkName == "" {
+		return nil
+	}
+	return c.ensureManagedRouteAdvertisement(networkName)
```
jcaamano: why not call reconcileCUDN (or the alternative for the default network) here?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,191 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+	managedCUDNs, err := c.getManagedCUDNs()
+	if err != nil {
+		return "", err
+	}
+	for _, cudn := range managedCUDNs {
+		if key == ManagedRouteAdvertisementName(cudn.Name) {
+			return cudn.Name, nil
+		}
 	}
+	return "", nil
```
jcaamano: I don't think this should be dependant on whether the network is managed or not. Resolving a network name from RA name is a pretty static matter, should not involve state. Filtering out unmanaged networks should happen elsewhere closer to whatever business logic we are handling at the time.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -356,18 +569,38 @@ func (c *Controller) cleanupStaleFRRConfigurations(currentBaseName string) error
...
+func (c *Controller) ensureManagedNetworkLabel(cudn *userdefinednetworkv1.ClusterUserDefinedNetwork) error {
+	if cudn.Labels != nil && cudn.Labels[ManagedRANetworkLabel] == cudn.Name {
+		return nil
+	}
+	updated := cudn.DeepCopy()
+	if updated.Labels == nil {
+		updated.Labels = map[string]string{}
+	}
+	updated.Labels[ManagedRANetworkLabel] = cudn.Name
+	_, err := c.cudnClient.K8sV1().ClusterUserDefinedNetworks().Update(context.TODO(), updated, metav1.UpdateOptions{FieldManager: fieldManager})
```
jcaamano: Shouldn't we be using server side apply, patch or retry on conflict? We might be overwriting stuff being written by someone else after we read.

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: Also, have you though about using a specific `generateName` for your RA, and resolve the network name from the network label selector? Wouldn't that work better?

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: `ensureManagedNetworkLabel` sets `k8s.ovn.org/managed-network=<cudn.Name>` on the CUDN so that the managed RA's ClusterUserDefinedNetworkSelector can match the CUDN's NADs by label.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: `reconcileRA` is only a self-healing path for the RA drift when someone externally modifies it. `reconcileCUDN` does more things, which I don't think are necessary here.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: I'll remove the managed checking. 

The label could be updated accidentally by others. So we cannot rely on the label. The current hash implementation can have a stable RA naming for a CUDN.

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: Can you update the label doc text as it is not completely describing the label use to full extent?
```
// ManagedRANetworkLabel is the label set on managed RouteAdvertisements to identify the network they advertise
ManagedRANetworkLabel = "k8s.ovn.org/managed-network"
```

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +282,212 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+	// Ignore events from our own writes
+	if newRA != nil && isOwnUpdate(newRA.ManagedFields) {
 		return false
 	}
-	if oldConfig == nil {
-		return true
+	if newRA == nil {
+		// This is just a safety check, newRA should be never nil here: delete events bypass ObjNeedsUpdate.
+		return false
 	}
```
jcaamano: nit: the natural way would be
```
if newRA == nil || isOwnUpdate(newRA.ManagedFields) {
		return false
}
```

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: If someone removes the label you could reconcile to add it back.

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: So my idea was to get rid of `syncAll` altogether. Will add more specific comments to achieve this.

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: Consider the benefit of having a single code path that you can call for multiple places to ensure state, vs having to maintain multiple code paths to handle each specific circumstance.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +117,159 @@ func NewController(
...
+		return err
+	}
+	for _, cudn := range managedCUDNs {
+		expectedNetworks[cudn.Name] = true
+	}
+
+	// Validate topology early so the controller fails fast on startup.
+	if len(expectedNetworks) > 0 && config.ManagedBGP.Topology != config.ManagedBGPTopologyFullMesh {
+		return fmt.Errorf("unsupported managed BGP topology %q, only %q is supported", config.ManagedBGP.Topology, config.ManagedBGPTopologyFullMesh)
+	}
```
jcaamano: Isn't this already validated by the CUDN conttroller? Why would we want to crash-loop in this circumstance?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +117,159 @@ func NewController(
...
+
+	// Ensure CDN RA. The CDN is config-driven and has no dedicated informer,
+	// so its RA must be created here rather than in an async reconciler.
+	if isCDNManaged() {
+		if err := c.ensureManagedRouteAdvertisement(types.DefaultNetworkName); err != nil {
+			return fmt.Errorf("failed to ensure CDN managed RouteAdvertisement: %w", err)
+		}
+	}
```
jcaamano: I would just rename `reconcileCUDN` to `reconcileNetwork` and make it capable of reconciling the default network, and enqueue `types.DefaultNetworkName` unconditionally on start

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +117,159 @@ func NewController(
...
+	allManagedRAs, err := c.raLister.List(labels.NewSelector().Add(*managedRAReq))
+	if err != nil {
+		return fmt.Errorf("failed to list managed RouteAdvertisements: %w", err)
+	}
+	for _, ra := range allManagedRAs {
+		networkName := ra.Labels[ManagedRANetworkLabel]
+		if !expectedNetworks[networkName] {
+			klog.Infof("Deleting stale managed RouteAdvertisement %s for network %s", ra.Name, networkName)
+			if err := c.deleteManagedRA(networkName); err != nil {
```
jcaamano: Already had an [example](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6158#discussion_r3001496048) on how to handle this
```
For example, on startup we will get an async event for the stale RA, which should probably call reconcileCUDN which should delete the RA if it is not needed
```

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +117,159 @@ func NewController(
...
+			}
+		}
+	}
+
+	// Clean up stale FRRConfigurations if no managed networks exist.
+	if len(expectedNetworks) == 0 {
+		if err := c.cleanupBaseFRRConfigurations(); err != nil {
+			return err
+		}
+	}
```
jcaamano: Wouldn't you get an async event for the FRRConfiguration for which you will end up calling `cleanupBaseFRRConfigurations`  from `reconcileFRRConfiguration`, so you don't need this?

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: Yes, we do. But if someone removes the label, `reconcileRA` needs `resolveNetworkForRA` to determine the network name and add the label back — a chicken-and-egg problem.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +117,110 @@ func NewController(
...
 
-	return controllerutil.StartWithInitialSync(c.ensureManagedResources, c.nodeController, c.managedRAController, c.managedFRRController)
+	initialSync := func() error {
+		if isCDNManaged() {
+			if err := c.reconcileNetwork(types.DefaultNetworkName); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
```
jcaamano: Wouldn't you just do something like
```
c.raController.Reconcile(ManagedRouteAdvertisementName(types.DefaultNetworkName))
```

no need to do it on intial sync either, can be done after StartWithInitialSync

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -34,13 +37,16 @@ import (
...
 	// FRRConfigManagedLabel is the label used to identify FRRConfigurations managed by this controller
 	FRRConfigManagedLabel = "k8s.ovn.org/managed-internal-fabric"
 	// FRRConfigManagedValue is the value used for the FRRConfigManagedLabel
 	FRRConfigManagedValue = "bgp"
-	// ManagedRANetworkLabel is the label set on managed RouteAdvertisements to identify the network they advertise
+	// ManagedRANetworkLabel is the label used to identify managed resources by network name.
+	// It is set on managed RouteAdvertisements and on CUDNs (to enable RA network selection).
 	ManagedRANetworkLabel = "k8s.ovn.org/managed-network"
 	// managedNamePrefix is the prefix for managed resource names
 	managedNamePrefix = "ovnk-managed-"
```
jcaamano: Can we unexport any/all of these things?

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: So on the RA controller, we generate FRRConfiguration names with a GenerateName of [ovnk-generated](https://github.com/pperiyasamy/ovn-kubernetes/blob/8dc68ba3b49bcd5acf21e55b554bc5f151dbaf02/go-controller/pkg/clustermanager/routeadvertisements/controller.go#L51).

I think it would be good to align to the pattern at least.

I would still use GenerateName to avoid collisions. I think relying on the label for ownership is fine as well and standard practice. For example, a ReplicationController uses labels to identify its pods. If you remove such label from the pod, the pod is somewhat orphaned and a new one generated. In our case it would be less problematic as there is more restrictive RBAC in place. This last consideration is a subjective appreciation on my part, so won't push too much for it.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -106,47 +117,106 @@ func NewController(
...
 	}
 
-	return controllerutil.StartWithInitialSync(c.ensureManagedResources, c.nodeController, c.managedRAController, c.managedFRRController)
+	if err := controllerutil.Start(controllers...); err != nil {
+		return err
+	}
+
+	if isCDNManaged() {
+		c.raController.Reconcile(ManagedRouteAdvertisementName(types.DefaultNetworkName))
+	}
```
jcaamano, line 199: nit: I would add a comment clarifying what this does

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: Regardless of what the configured AS number is, we only managed one at a time, right? So safer to not base the handling on the actual AS number?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -34,13 +37,16 @@ import (
...
+	// filter out informer events that result from our own writes.
+	fieldManager = "clustermanager-managed-bgp-controller"
 	// FRRConfigManagedLabel is the label used to identify FRRConfigurations managed by this controller
 	FRRConfigManagedLabel = "k8s.ovn.org/managed-internal-fabric"
 	// FRRConfigManagedValue is the value used for the FRRConfigManagedLabel
 	FRRConfigManagedValue = "bgp"
-	// ManagedRANetworkLabel is the label set on managed RouteAdvertisements to identify the network they advertise
+	// ManagedRANetworkLabel is the label used to identify managed resources by network name.
+	// It is set on managed RouteAdvertisements and on CUDNs (to enable RA network selection).
 	ManagedRANetworkLabel = "k8s.ovn.org/managed-network"
```
jcaamano: Could we unify this labels? Use `k8s.ovn.org/managed-internal-fabric` both for FRRConfigs and RAs? Perhaps use empty value for FRRConfigs?

jcaamano (commented):
Final comments. Sorry If I didn't come around with some of these suggestions earlier.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: `k8s.ovn.org/managed-network=<network-name>` is semantically cleaner for its purpose. I think it's better to keep both labels. Or at least keep  `k8s.ovn.org/managed-network=` instead.

jcaamano: I see it differently. The whole point of this controller is to manage the fabric for a network, not to manage the network itself which is something done by several controllers. Just looking at the `k8s.ovn.org/managed-network=` label you wouldn't get a hint of what controller might be handling the label. But this is my subjective perception, so if you think it is better to keep the current labeling pattern, I won't complain.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: The current implementation gives us a predictable name for the FRRConfiguration. It makes the code simpler when we need to update the FRRConfiguration or clean up stale ones.

pliurh: Yeah, we only managed one at a time. So there won't be hash collision. Why is it not safe?

jcaamano: It really doesn't look good that we would have both FRRConfigurations with names

ovnk-managed-abcde
ovnk-generated-12345

It just looks cheap, like like don't talk within the project. At least use `ovnk-managedbgp-`?

> The current implementation gives us a predictable name for the FRRConfiguration. It makes the code simpler when we need to update the FRRConfiguration or clean up stale ones.

I don't see how my suggestion incurs in complex code.

jcaamano: If ovnk restarts with a different configured AS number, the FRRConfiguration for the old AS number is left behind.

pliurh: Using a predictable name lets us Get the FRRConfiguration directly instead of List+iterate, so I think it is slightly simpler. If you prefer consistency with the existing GenerateName pattern, I can update it to use GenerateName with an `ovnk-managedbgp-` prefix.

pliurh: I'll use the label to clean the stale FRRConfiguration.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -273,86 +450,86 @@ func (c *Controller) ensureBaseFRRConfiguration(allNodes []*corev1.Node) error {
...
-	if needsUpdate {
-		klog.Infof("Updating base FRRConfiguration %s", baseName)
-		updated := existing.DeepCopy()
-		if updated.Labels == nil {
-			updated.Labels = map[string]string{}
+	cfgCount := len(managedConfigs)
+	if cfgCount == 0 {
+		frrConfig := &frrtypes.FRRConfiguration{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: managedNamePrefix,
```
jcaamano: I guess that what I wanted to say [here](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6158#discussion_r3020605946) is that, couldn't this be fixed name like `ovnk-managedbgp-fullmesh` or something like that, wouldn't it make it way simpler?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -273,80 +455,69 @@ func (c *Controller) ensureBaseFRRConfiguration(allNodes []*corev1.Node) error {
...
 				},
 			},
 		},
 	}
 
-	existing, err := c.frrClient.ApiV1beta1().FRRConfigurations(config.ManagedBGP.FRRNamespace).Get(context.TODO(), baseName, metav1.GetOptions{})
+	frrConfigName := managedFRRConfigurationName()
+	managedConfigs, err := c.frrLister.FRRConfigurations(config.ManagedBGP.FRRNamespace).List(
+		labels.SelectorFromSet(labels.Set{frrConfigManagedLabel: frrConfigManagedValue}),
+	)
```
jcaamano: Why do you need to list these and delete them? We only have one FRRConfiguration with a fixed name that you are going to create/update. What other FRRConfigurations are you handling here?

Also the handling on `frrNeedsUpdate` and `cleanupBaseFRRConfigurations` should be easier with a static name.

Feels like I am missing something.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -52,48 +57,47 @@ func managedHashedName(s string) string {
...
 
-// BaseFRRConfigName returns the name of the base FRRConfiguration that
-// configures the cluster internal BGP fabric. This FRRConfiguration is shared
-// by all networks in managed routing mode. The name follows the pattern
-// "ovnk-managed-<hash>" where <hash> is derived from the configured AS number.
-func BaseFRRConfigName() string {
-	return managedHashedName(fmt.Sprintf("%d", config.ManagedBGP.ASNumber))
+// managedFRRConfigurationName returns the static name for the managed base FRRConfiguration.
+func managedFRRConfigurationName() string {
+	return managedNamePrefix + config.ManagedBGP.Topology
```
jcaamano: nit: this could be a constant

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: Yes, we only have one possible name for now. But when we support multiple topologies, this will clean up the stale FRRConfiguration if the topology is changed.

go-controller/pkg/clustermanager/managedbgp/controller.go
pliurh: Same consideration, that we may support topology other than `full-mesh` in the future.

go-controller/pkg/clustermanager/managedbgp/controller.go
jcaamano: But that is not the case today. And it is only one at the same time. Not multiple at the same time, right? It maybe best if the static name then does not make a reference to the actual topology. Something like "ovnk-managedbgp-nooverlay" or something like that.

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -159,97 +221,214 @@ func (c *Controller) nodeNeedsUpdate(oldNode, newNode *corev1.Node) bool {
...
+		if err != nil {
+			return "", fmt.Errorf("failed to list CUDNs: %w", err)
+		}
+		for _, cudn := range allCUDNs {
+			if key == managedRouteAdvertisementName(cudn.Name) {
+				return cudn.Name, nil
+			}
+		}
+	}
+	return "", nil
```
jcaamano, line 403: There is something here that still does not make sense. We distrust someone removing the label, but we don't distrust someone changing the label to a different incorrect value. I guess we should stick with one or the other. So if we don't want to trust the label, stick with the name. And perhaps we have no use for the label?

go-controller/pkg/clustermanager/managedbgp/controller.go
```diff
@@ -26,24 +25,32 @@ import (
...
+	// frrConfigManagedValue is the value used for the frrConfigManagedLabel
+	frrConfigManagedValue = "bgp"
+	// managedRANetworkLabel is the label used to identify managed resources by network name.
+	// It is set on managed RouteAdvertisements and on CUDNs (to enable RA network selection).
+	managedRANetworkLabel = "k8s.ovn.org/managed-network"
 	// managedNamePrefix is the prefix for managed resource names
-	managedNamePrefix = "ovnk-managed-"
+	managedNamePrefix = "ovnk-managedbgp-"
+	// managedFRRConfigurationName is the static name for the managed base FRRConfiguration.
+	managedFRRConfigurationName = "ovnk-managedbgp-nooverlay"
```
jcaamano, line 53: Looking at this again after a few days, the same thought pops to mine: we could unify this to a more consistent naming.

I would use a single label: `k8s.ovn.org/managed-bgp` (with no value in case of the FRRConfiguration which server no purpose)
I would use a single name and prefix: `ovnk-managed-bgp`

This all goes inline to how we are naming the controller itself.

PR #6163, opened 2026-03-25
Annotate OVN load balancers with ExternalTrafficPolicy for debugging

shettyg:
It is useful to derive a service's ExternalTrafficPolicy purely by inspecting OVN NB load balancer external-ids, without needing access to the Kubernetes API. This makes debugging traffic-policy related issues easier when only the OVN northbound database is available.

Set k8s.ovn.org/etp=local on load balancer external-ids when the backing service has externalTrafficPolicy: Local. When the policy is Cluster (the default), the key is omitted.

This annotation will also be useful, if a user decides to program breth0 bridge for traffic muxing by only looking at OVN NB database.

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Load balancers now record ExternalTrafficPolicy=Local so the policy is honored across default and user-defined networks; syncing adds or removes this marker when the service policy changes.

* **Tests**
  * Added coverage validating ExternalTrafficPolicy behavior and sync behavior (including policy flips) across network configurations.

### DIFF
--- a/go-controller/pkg/ovn/controller/services/services_controller_test.go
+++ b/go-controller/pkg/ovn/controller/services/services_controller_test.go
@@ -1469,6 +1469,83 @@ func TestSyncServices(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:      "upgrade: pre-existing LBs without etp annotation get updated for ETP=Local service",
+			nodeAInfo: nodeAInfo,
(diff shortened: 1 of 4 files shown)


npinaeva:
Yeah the absence of a flag could mean many things, so it is better to put values always I think

PR #6166, opened 2026-03-25
Move CDN to node-level driven controller

trozet:
## Release Notes

* **Bug Fixes**
  * Improved node reconciliation error handling and recovery mechanisms to maintain system stability during transient failures.
  * Enhanced gateway and SNAT rule configuration to handle edge cases more gracefully.

* **Improvements**
  * Strengthened network controller initialization and startup sequencing for more reliable node management.
  * Optimized error tracking and failure recovery in node state management.

### DIFF
--- a/go-controller/pkg/ovn/default_network_controller.go
+++ b/go-controller/pkg/ovn/default_network_controller.go
@@ -515,7 +515,9 @@ func (oc *DefaultNetworkController) ReconcileNode(oldNode, newNode *corev1.Node,
 			_, failed = oc.mgmtPortFailed.Load(newNode.Name)
 			mgmtSync := failed || defaultNodeSubnetChangedWithState(oldNode, newNode, oldState, newState)
 			_, failed = oc.gatewaysFailed.Load(newNode.Name)
-			gwSync := failed || gatewayChanged(oldNode, newNode) || defaultNodeSubnetChangedWithState(oldNode, newNode, oldState, newState) ||
+			gwSync := failed || gatewayChanged(oldNode, newNode, oldState, newState, oc.GetNetworkName()) ||
+				nodeChassisChanged(oldNode, newNode) ||
+				defaultNodeSubnetChangedWithState(oldNode, newNode, oldState, newState) ||
 				hostCIDRsChanged(oldNode, newNode) || nodeGatewayMTUSupportChanged(oldNode, newNode)
 			_, hoSync := oc.hybridOverlayFailed.Load(newNode.Name)
 			_, syncZoneIC := oc.syncZoneICFailed.Load(newNode.Name)
@@ -576,7 +578,8 @@ func (oc *DefaultNetworkController) ReconcileNode(oldNode, newNode *corev1.Node,
 
 	_, syncHostNetAddrSet := oc.syncHostNetAddrSetFailed.Load(newNode.Name)
 	hostNamespaceAddressesChanged := oldNode != nil &&
-		(defaultNodeSubnetChangedWithState(oldNode, newNode, oldState, newState) || gatewayChanged(oldNode, newNode))
+		(defaultNodeSubnetChangedWithState(oldNode, newNode, oldState, newState) ||
+			gatewayChanged(oldNode, newNode, oldState, newState, oc.GetNetworkName()))
 	if oldNode == nil || syncHostNetAddrSet || hostNamespaceAddressesChanged {
 		hostNamespaceAddrSetErr := false
 		if hostNamespaceAddressesChanged {
--- a/go-controller/pkg/ovn/layer3_user_defined_network_controller.go
+++ b/go-controller/pkg/ovn/layer3_user_defined_network_controller.go
@@ -651,7 +651,8 @@ func (oc *Layer3UserDefinedNetworkController) ReconcileNode(oldNode, newNode *co
 			syncZoneIC = syncZoneIC || zoneClusterChanged
 			_, failed = oc.gatewaysFailed.Load(newNode.Name)
 			syncGw := failed ||
-				gatewayChanged(oldNode, newNode) ||
+				gatewayChanged(oldNode, newNode, oldState, newState, oc.GetNetworkName()) ||
+				nodeChassisChanged(oldNode, newNode) ||
 				nodeSubnetChange ||
 				hostCIDRsChanged(oldNode, newNode) ||
 				nodeGatewayMTUSupportChanged(oldNode, newNode)
--- a/go-controller/pkg/ovn/ovn.go
+++ b/go-controller/pkg/ovn/ovn.go
@@ -19,6 +19,7 @@ import (
 	libovsdbclient "github.com/ovn-kubernetes/libovsdb/client"
 
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/config"
+	nodecontroller "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controllers/node"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/kubevirt"
 	libovsdbutil "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/libovsdb/util"
 	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/metrics"
@@ -400,10 +401,13 @@ func (oc *DefaultNetworkController) syncNodeGateway(node *corev1.Node) error {
 	return oc.deleteAdvertisedNetworkIsolation(node.Name)
 }
 
-// gatewayChanged() compares old annotations to new and returns true if something has changed.
-func gatewayChanged(oldNode, newNode *corev1.Node) bool {
-	return oldNode.Annotations[util.OvnNodeL3GatewayConfig] != newNode.Annotations[util.OvnNodeL3GatewayConfig] ||
-		oldNode.Annotations[util.OvnNodeChassisID] != newNode.Annotations[util.OvnNodeChassisID]
+// gatewayChanged compares the per-network gateway annotation between node
+// revisions. Chassis changes are handled separately by callers that need them.
+func gatewayChanged(oldNode, newNode *corev1.Node, oldState, newState *nodecontroller.NodeAnnotationState, netName string) bool {
+	if oldState != nil && newState != nil {
+		return nodecontroller.GatewayAnnotationChangedForNetworkWithState(oldState, newState, netName)
+	}
+	return oldNode.Annotations[util.OvnNodeL3GatewayConfig] != newNode.Annotations[util.OvnNodeL3GatewayConfig]
 }
 
 // hostCIDRsChanged compares old annotations to new and returns true if the something has changed.
--- a/go-controller/pkg/controllers/node/node_annotation_cache.go
+++ b/go-controller/pkg/controllers/node/node_annotation_cache.go
@@ -20,8 +20,9 @@ type NodeAnnotationCache struct {
 }
 
 type nodeAnnotationEntry struct {
-	networkMaps map[string]cachedNetworkMap
-	subnetMaps  map[string]cachedSubnetMap
+	networkMaps       map[string]cachedNetworkMap
+	gatewayConfigMaps map[string]cachedGatewayConfigMap
+	subnetMaps        map[string]cachedSubnetMap
 }
 
 type cachedNetworkMap struct {
@@ -36,6 +37,13 @@ type cachedSubnetMap struct {
 	raw string
 	// parsed holds the parsed annotation of network names -> subnets
 	parsed map[string][]*net.IPNet
+}
+
+type cachedGatewayConfigMap struct {
+	// raw is used for fast comparisons within the cache
+	raw string
+	// parsed holds the parsed annotation of network names -> gateway configs
+	parsed map[string]*util.L3GatewayConfig
 }
 
 // NewNodeAnnotationCache creates an empty per-node annotation parse cache.
@@ -61,8 +69,9 @@ func (c *NodeAnnotationCache) updateNodeAnnotationState(node *corev1.Node, updat
 	}
 	networkIDs, networkIDsErr := c.parseNetworkMapCached(node, util.OvnNetworkIDs, updateCache)
 	tunnelIDs, tunnelIDsErr := c.parseNetworkMapCached(node, types.UDNLayer2NodeGRLRPTunnelIDAnnotation, updateCache)
+	l3GatewayConfigs, l3GatewayConfigsErr := c.parseGatewayConfigMapCached(node, util.OvnNodeL3GatewayConfig, updateCache)
 	subnets, subnetsErr := c.parseSubnetMapCached(node, types.NodeSubnetsAnnotation, updateCache)
-	return newNodeAnnotationState(node.Name, networkIDs, networkIDsErr, tunnelIDs, tunnelIDsErr, subnets, subnetsErr)
+	return newNodeAnnotationState(node.Name, networkIDs, networkIDsErr, tunnelIDs, tunnelIDsErr, l3GatewayConfigs, l3GatewayConfigsErr, subnets, subnetsErr)
 }
 
 // ParseUDNLayer2NodeGRLRPTunnelIDCached returns the per-network tunnel ID from the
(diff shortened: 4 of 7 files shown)


go-controller/pkg/ovn/master.go
```diff
@@ -416,6 +416,181 @@ func (oc *DefaultNetworkController) syncNodes(kNodes []interface{}) error {
...
+				oc.recordNodeErrorEvent(newNode, err)
+				return err
+			}
+			switchToOvnNode = true
+		} else if oldNode == nil && !util.NoHostSubnet(newNode) {
+			if err := oc.deleteHoNodeEvent(newNode); err != nil {
+				oc.recordNodeErrorEvent(newNode, err)
+				return err
+			}
+		}
```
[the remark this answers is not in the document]
trozet: I think the old retry framework path also used the newNode here so this is not a regression, but agree it should be fixed.

go-controller/pkg/ovn/master.go
```diff
@@ -416,6 +416,181 @@ func (oc *DefaultNetworkController) syncNodes(kNodes []interface{}) error {
...
+	_, syncHostNetAddrSet := oc.syncHostNetAddrSetFailed.Load(newNode.Name)
+	if oldNode == nil || syncHostNetAddrSet {
+		if err := oc.addIPToHostNetworkNamespaceAddrSet(newNode); err != nil {
+			klog.Errorf("Failed to add node IPs to %s address_set: %v", config.Kubernetes.HostNetworkNamespace, err)
+			oc.syncHostNetAddrSetFailed.Store(newNode.Name, true)
+			aggregatedErrors = append(aggregatedErrors, err)
+		} else {
+			oc.syncHostNetAddrSetFailed.Delete(newNode.Name)
+		}
+	}
```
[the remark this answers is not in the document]
trozet: The previous UpdateResource path in default_network_controller.go also only retried when syncHostNetAddrSetFailed was set; it did not diff oldNode vs newNode host-network IPs so I don't think this needs to be addressed here.

go-controller/pkg/controllers/node/node_controller.go
```diff
@@ -302,6 +302,9 @@ func (c *NodeController) reconcileNode(key string) error {
 // Reconciliation is level-driven.
 func (c *NodeController) reconcileUpdate(handler NodeHandler, oldNode, newNode *corev1.Node, netName string, oldState, newState *NodeAnnotationState) error {
 	if err := handler.ReconcileNode(oldNode, newNode, oldState, newState); err != nil {
+		// Keep the latest informer state cached even when reconciliation fails so a later
+		// delete can clean up any partial state created before the error was returned.
+		c.setCachedNode(netName, newNode)
 		return err
```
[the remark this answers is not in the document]
trozet: The fix you mentioned wont work because we need to track partially configured states. For example if reconciliation fails and the node is not added to the node cache, but later we need to delete and the informer object is gone, we need a reference to know what we should remove. I agree that it is not the right fix to just update the nodeCache before we have configured the state, so I will add an additional cache: latestInformerNodeCache to know the last known informer object we at least partially configured. Then when nodeCache has no hit, and the node object is gone from the informer, we can fallback to this.

go-controller/pkg/controllers/node/node_controller.go
```diff
@@ -302,6 +302,9 @@ func (c *NodeController) reconcileNode(key string) error {
 // Reconciliation is level-driven.
 func (c *NodeController) reconcileUpdate(handler NodeHandler, oldNode, newNode *corev1.Node, netName string, oldState, newState *NodeAnnotationState) error {
 	if err := handler.ReconcileNode(oldNode, newNode, oldState, newState); err != nil {
+		// Keep the latest informer state cached even when reconciliation fails so a later
+		// delete can clean up any partial state created before the error was returned.
+		c.setCachedNode(netName, newNode)
```
trozet: note this is removed in a later commit

tssurya:
I am pulling this into OCP CI for a signal.. will get back here with results

go-controller/pkg/controllermanager/controller_manager.go
```diff
@@ -292,6 +292,7 @@ func NewControllerManager(ovnClient *util.OVNClientset, wf *factory.WatchFactory
 	var err error
 
 	cm.networkManager = networkmanager.Default()
+	cm.udnNodeController = nodecontroller.NewNodeController(cm.watchFactory, cm.networkManager.Interface())
```
npinaeva: there is an exact copy 6 lines below?

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -358,6 +365,65 @@ func (oc *DefaultNetworkController) Stop() {
...
+		return err
+	}
+	if err := oc.waitForInitialNodeSync(); err != nil {
+		oc.DeregisterNodeHandler()
+		return err
+	}
+	return nil
+}
+
+func (oc *DefaultNetworkController) waitForInitialNodeSync() error {
```
npinaeva, line 388: isn't it similar to `(oc *Layer3UserDefinedNetworkController) waitForLocalZoneNodeLogicalSwitches()`? should we reuse some code here?

go-controller/pkg/ovn/master.go
```diff
@@ -416,6 +416,185 @@ func (oc *DefaultNetworkController) syncNodes(kNodes []interface{}) error {
 	return nil
 }
 
+func (oc *DefaultNetworkController) SyncNodes(nodes []*corev1.Node) error {
```
npinaeva: shuldn't these changes go back to the `default_network_controller.go`?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -198,7 +198,11 @@ func (na *NodeAllocator) NeedsNodeAllocation(node *corev1.Node) bool {
 
 	// ovn node check
 	if na.hasNodeSubnetAllocation() {
-		if !util.HasNodeHostSubnetAnnotation(node, na.netInfo.GetNetworkName()) {
+		hostSubnets, err := util.ParseNodeHostSubnetAnnotation(node, na.netInfo.GetNetworkName())
```
npinaeva: I was worried we don't use cache here, but this is only a test function? can we rename it to clarify that and make unexported?

go-controller/pkg/controllers/node/node_controller.go
```diff
@@ -75,6 +76,11 @@ type NodeController struct {
 	// nodeCache contains configured node state
 	// keyed by network -> nodeName
 	nodeCache map[string]map[string]*corev1.Node
+	// latestInformerNodeCache contains the latest informer object seen for a node
+	// even if reconciliation failed. It is only used as a delete fallback when
+	// no configured node state exists in nodeCache.
+	// keyed by network -> nodeName
+	latestInformerNodeCache map[string]map[string]*corev1.Node
```
npinaeva, line 83: that made me think about the hack we added last time to behave thae same way if initialSync added a node, but then the nodes was deleted immediately and we didn't have anything in the cache. So `reconcileDelete` has this branch
```
if oldNode == nil {
		oldNode = &corev1.Node{
				Name: nodeName,
			},
		}
	}
``` but I guess that is actually not good enough? or does it work for UDN controllers but not for the default?

File: go-controller/pkg/controllermanager/controller_manager.go (hunk 5 above)
Comment (trozet) [reply go-controller/pkg/controllermanager/controller_manager.go]: yeah it is needed to be declared again with the right network manager if UDN is enabled. I can do an else instead if that's what you prefer?

Comment (trozet) [reply go-controller/pkg/controllermanager/controller_manager.go]: i'm dumb, fixed

File: go-controller/pkg/ovn/default_network_controller.go (hunk 6 above)
Comment (trozet) [reply go-controller/pkg/ovn/default_network_controller.go:388]: So they are not exactly the same. Whether or not they should be...I'm not sure. But the current code preserves the previous behavior:

1. With DNC it will keep starting even if the local node add failed and assume that in the background it will get retried. This is how it was before. This can happen if a node annotation changed or the cluster was converted to a different IP mode and ovnkube-controller is waiting on cluster-manager to assign a new node-subnet.
2. With L3 UDN it is stricter and does not have this shortcut. It is basically written for UDN dynamic start up case to say that nothing should prevent the logical topology from being created.

Perhaps something to follow up on in a separate PR? BTW this reminds me of our discussion we had a while ago to change UDN start up to not timeout and cause OVNK start up to fail, maybe this is related and we should be proceeding after we try to process all the local nodes at least one time.

File: go-controller/pkg/ovn/master.go (hunk 7 above)
Comment (trozet) [reply go-controller/pkg/ovn/master.go]: well there are many other DefaultNetworkController methods in master.go, but you are right I shouldn't make it worse. We should probably move all those receiver methods to the respective file. I will just move the methods I added in this PR, and file a follow up PR to move the rest.

File: go-controller/pkg/clustermanager/node/node_allocator.go (hunk 8 above)
Comment (trozet) [reply go-controller/pkg/clustermanager/node/node_allocator.go]: yeah with the node-level driven refactor it was changed to use NeedsNodeAllocationWithState in the functional path. I'll update the test to use this path as well and remove the function.

File: go-controller/pkg/controllers/node/node_controller.go (hunk 9 above)
Comment (trozet) [reply go-controller/pkg/controllers/node/node_controller.go:83]: Correct. Falling back with an empty node is not good enough. We just didn't see an issue with UDN controllers, probably because it is more generic in its clean up and features, but in DNC I saw failures. Specifically the one I saw was with the host-network namespace address set for remote-zone nodes::

  - DNC could add IPs derived from node annotations
  - then fail later in the same reconcile
  - then delete needed those same node IP annotations to remove the address-set entries
  - dummy node was not enough

IIRC there are other examples as well with DNC where we configure something that we need the object for to delete later. Perhaps you could say that is not level driven and is a bug in how we handle deleting/syncing, and probably you would be right considering all the sync flags we have to know exactly what state we are in with processing nodes. But that is a refactor for another day.

With this last commit we store the informer last known object, which should make our delete more reliable for both DNC and UDN.

File: go-controller/pkg/controllermanager/controller_manager.go [hunk 10]
```diff
@@ -297,8 +297,8 @@ func NewControllerManager(ovnClient *util.OVNClientset, wf *factory.WatchFactory
 		if err != nil {
 			return nil, err
 		}
-		cm.udnNodeController = nodecontroller.NewNodeController(cm.watchFactory, cm.networkManager.Interface())
 	}
+	cm.udnNodeController = nodecontroller.NewNodeController(cm.watchFactory, cm.networkManager.Interface())
```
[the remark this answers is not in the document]
trozet: we do not want to spin up more than 1 controller. I think the better fix is to change ShouldFilterByRemoteNetworkActivity(node, netName string) to pass netName and exclude default network

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -358,6 +367,244 @@ func (oc *DefaultNetworkController) Stop() {
...
+	_, syncHostNetAddrSet := oc.syncHostNetAddrSetFailed.Load(newNode.Name)
+	if oldNode == nil || syncHostNetAddrSet {
+		if err := oc.addIPToHostNetworkNamespaceAddrSet(newNode); err != nil {
+			klog.Errorf("Failed to add node IPs to %s address_set: %v", config.Kubernetes.HostNetworkNamespace, err)
+			oc.syncHostNetAddrSetFailed.Store(newNode.Name, true)
+			aggregatedErrors = append(aggregatedErrors, err)
+		} else {
+			oc.syncHostNetAddrSetFailed.Delete(newNode.Name)
+		}
+	}
```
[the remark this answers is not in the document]
trozet: the host cidrs are not the IPs added to the host network namespace address set. They are really the gateway/management IP. I guess we can still detect if these values change and reconcile accordingly.

go-controller/pkg/controllers/node/node_controller.go
```diff
@@ -253,10 +261,11 @@ func (c *NodeController) reconcileNode(key string) error {
 				needsDelete = true
 				c.markNodeNeedsDeleteReconciliation(netName, nodeName)
 				c.deleteNodeActive(netName, nodeName)
-				// if we have no oldNode in the cache, and we are going inactive here, then populate the cache
+				// If we have no configured node in cache, retain the latest informer object
+				// so delete reconciliation has the metadata needed for cleanup.
 				if newNode != nil && oldNode == nil {
 					oldNode = newNode
-					c.setCachedNode(netName, oldNode)
+					c.setLatestInformerNode(netName, oldNode)
```
npinaeva, line 268: i think it is called 3 times, but isn't easier to just set it after
```
newNode, err := c.nodeLister.Get(nodeName)
```
only once if it is not nil? because we don't care about the reconciliation result afaiu as opposed to the previous `setCachedNode` usage

go-controller/pkg/controllers/node/node_controller.go
```diff
@@ -302,21 +311,26 @@ func (c *NodeController) reconcileNode(key string) error {
...
 	c.setCachedNode(netName, newNode)
+	c.setLatestInformerNode(netName, newNode)
 	c.deleteNodeReconciliation(netName, newNode.Name)
 
 	return nil
 }
 
 // reconcileDelete handles deletion using cached state.
 func (c *NodeController) reconcileDelete(handler NodeHandler, nodeName, netName string, oldNode *corev1.Node, oldState *NodeAnnotationState) error {
+	if oldNode == nil {
```
npinaeva, line 330: so you can delete L334-340 now?

go-controller/pkg/controllers/node/node_controller.go
trozet: Right now it only stores it when it is actually needed, vs unconditionally. Storing it after informer get... the node might not even have network so storing is a waste. We can just reduce from 3->2 calls by collapsing 2->1 in reconcileUpdate and preserve only setting when its actually going to be needed.

go-controller/pkg/controllers/node/node_controller.go
trozet: we agreed to leave it as a fallback. Added a comment though to make it clear

go-controller/pkg/ovn/default_network_controller.go
```diff
@@ -575,12 +575,23 @@ func (oc *DefaultNetworkController) ReconcileNode(oldNode, newNode *corev1.Node,
 	}
 
 	_, syncHostNetAddrSet := oc.syncHostNetAddrSetFailed.Load(newNode.Name)
-	if oldNode == nil || syncHostNetAddrSet {
+	hostNamespaceAddressesChanged := oldNode != nil &&
```
npinaeva, line 580: we did ignore such changees before this PR thought right?

go-controller/pkg/ovn/ovn.go
```diff
@@ -400,10 +401,13 @@ func (oc *DefaultNetworkController) syncNodeGateway(node *corev1.Node) error {
...
-func gatewayChanged(oldNode, newNode *corev1.Node) bool {
-	return oldNode.Annotations[util.OvnNodeL3GatewayConfig] != newNode.Annotations[util.OvnNodeL3GatewayConfig] ||
-		oldNode.Annotations[util.OvnNodeChassisID] != newNode.Annotations[util.OvnNodeChassisID]
+// gatewayChanged compares the per-network gateway annotation between node
+// revisions. Chassis changes are handled separately by callers that need them.
+func gatewayChanged(oldNode, newNode *corev1.Node, oldState, newState *nodecontroller.NodeAnnotationState, netName string) bool {
+	if oldState != nil && newState != nil {
+		return nodecontroller.GatewayAnnotationChangedForNetworkWithState(oldState, newState, netName)
+	}
+	return oldNode.Annotations[util.OvnNodeL3GatewayConfig] != newNode.Annotations[util.OvnNodeL3GatewayConfig]
```
npinaeva, line 410: can't one of the nodes be nil too? it would be nil dereference then?

go-controller/pkg/ovn/layer3_user_defined_network_controller.go
```diff
@@ -651,7 +651,8 @@ func (oc *Layer3UserDefinedNetworkController) ReconcileNode(oldNode, newNode *co
 			syncZoneIC = syncZoneIC || zoneClusterChanged
 			_, failed = oc.gatewaysFailed.Load(newNode.Name)
 			syncGw := failed ||
-				gatewayChanged(oldNode, newNode) ||
+				gatewayChanged(oldNode, newNode, oldState, newState, oc.GetNetworkName()) ||
+				nodeChassisChanged(oldNode, newNode) ||
```
npinaeva, line 655: I am confused: why did you add `nodeChassisChanged` here and how is it related to the gw annotation parsing?

go-controller/pkg/ovn/ovn.go
```diff
@@ -400,10 +401,13 @@ func (oc *DefaultNetworkController) syncNodeGateway(node *corev1.Node) error {
 	return oc.deleteAdvertisedNetworkIsolation(node.Name)
 }
 
-// gatewayChanged() compares old annotations to new and returns true if something has changed.
-func gatewayChanged(oldNode, newNode *corev1.Node) bool {
-	return oldNode.Annotations[util.OvnNodeL3GatewayConfig] != newNode.Annotations[util.OvnNodeL3GatewayConfig] ||
-		oldNode.Annotations[util.OvnNodeChassisID] != newNode.Annotations[util.OvnNodeChassisID]
+// gatewayChanged compares the per-network gateway annotation between node
+// revisions. Chassis changes are handled separately by callers that need them.
+func gatewayChanged(oldNode, newNode *corev1.Node, oldState, newState *nodecontroller.NodeAnnotationState, netName string) bool {
```
npinaeva, line 406: so why do we only use cache to check if it changed, but the code that actually uses this information is still wired via `util.ParseNodeChassisIDAnnotation`?

go-controller/pkg/ovn/default_network_controller.go
trozet: right, this was caught by rabbit during review of porting the code.

go-controller/pkg/ovn/layer3_user_defined_network_controller.go
trozet: cause the old gatewayChanged function also included chassis id change check:

```
oldNode.Annotations[util.OvnNodeChassisID] != newNode.Annotations[util.OvnNodeChassisID]
```

so I split it out

go-controller/pkg/ovn/ovn.go
trozet: this stores the l3 gateway config, not the chassis id. The chassis id is already a raw string value, so it doesn't need parsing and storage.

For the l3 gateway config it is not using the cache when actually updating syncGateway. I could plug it in, but that will be even more code and not sure if that will help much as we don't expect gateway config to update per network often. The real savings here is that the l3 gateway config annotation is updated per UDN. So the hot spot is checking if the network we care about actually updated. So gatewayChanged using the parsed cache is where we can get perf benefit.

go-controller/pkg/ovn/ovn.go
```diff
@@ -400,10 +401,16 @@ func (oc *DefaultNetworkController) syncNodeGateway(node *corev1.Node) error {
...
+	if oldState != nil && newState != nil {
+		return nodecontroller.GatewayAnnotationChangedForNetworkWithState(oldState, newState, netName)
+	}
+	if oldNode == nil || newNode == nil {
+		return false
```
npinaeva: why false though? it might have changed if it was set on the new obj?

trozet: this function is only used on update, and i realized it will never be nil in the calls that exist today, so i will revert this check

trozet:
saw this UT flake, not caused by this PR: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6245

PR #6168, opened 2026-03-26
Fix race conditions in secondary host egress IP configuration

arghosh93:
When EgressIP is enabled for the cluster, it may be configured on a secondary interface. Because of this, there can be a delay in programming the IP rule and SNAT iptables rule for a pod that is served by an EgressIP. During this window, if the pod initiates egress traffic, its IP may leak onto the external network through the egress gateway interface (breth0 or br-ex).
Two race conditions caused pods to send traffic with node IP instead of egress IP during configuration:

1. Node controller internal ordering: routes were added after IP was on interface, creating a window where IP rules referenced empty routing tables.

2. OVN-node coordination: OVN controller programmed logical router policies before node controller finished host-side configuration.

Solution:
- Reorder node controller operations: routes → IP rules → iptables → annotation → IP on interface
- Use k8s.ovn.org/secondary-host-egress-ips annotation as readiness signal
- OVN controller checks annotation before programming logical router policies for secondary host egress IPs

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **New Features**
  * Validates secondary-host Egress IP readiness so OVN policies are programmed only after host-side configuration is present.

* **Bug Fixes**
  * Reordered host-side Egress IP application: add routes first, then per-pod rules/SNAT, then interface address, and finally update node annotation. Route-add failures now return contextual errors.

* **Tests**
  * Added tests ensuring OVN policy programming is deferred until the secondary-host Egress IP readiness is recorded.

### DIFF
--- a/go-controller/pkg/node/controllers/egressip/egressip.go
+++ b/go-controller/pkg/node/controllers/egressip/egressip.go
@@ -778,7 +778,16 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
 		}
 	}
 	// apply new changes
+	// Apply configuration in strict order to avoid race conditions: routes → IP rules → iptables → IP on interface → annotation.
+	// Annotation serves as readiness signal for OVN controller to program logical router policies.
 	if update != nil && update.eIPConfig != nil && update.eIPConfig.addr != nil && len(update.eIPConfig.routes) > 0 {
+		for _, routeToAdd := range update.eIPConfig.routes {
+			if err := c.routeManager.Add(routeToAdd); err != nil {
+				return fmt.Errorf("failed to add egress IP route %v, err: %w", routeToAdd, err)
+			}
+		}
+		existing.eIPConfig.routes = update.eIPConfig.routes
+
 		for updatedTargetNS, updatedTargetPod := range update.namespacesWithPodIPConfigs {
 			existingNs, found := existing.namespacesWithPodIPConfigs[updatedTargetNS]
 			if !found {
@@ -798,22 +807,17 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
 				}
 			}
 		}
-		if err := c.addIPToAnnotation(update.eIPConfig.addr.IP.String()); err != nil {
-			return fmt.Errorf("failed to add egress IP address to annotation: %v", err)
-		}
+
 		// TODO(mk): only apply the follow when its new config or when it failed to apply
 		// Ok to repeat requests to route manager and link manager
 		if err := c.linkManager.AddAddress(*update.eIPConfig.addr); err != nil {
 			return fmt.Errorf("failed to add address to link: %v", err)
 		}
 		existing.eIPConfig.addr = update.eIPConfig.addr
-		// route manager manages retry
-		for _, routeToAdd := range update.eIPConfig.routes {
-			if err := c.routeManager.Add(routeToAdd); err != nil {
-				return err
-			}
+
+		if err := c.addIPToAnnotation(update.eIPConfig.addr.IP.String()); err != nil {
+			return fmt.Errorf("failed to add egress IP address to annotation: %v", err)
 		}
-		existing.eIPConfig.routes = update.eIPConfig.routes
 	}
 	return nil
 }
--- a/go-controller/pkg/ovn/egressip.go
+++ b/go-controller/pkg/ovn/egressip.go
@@ -2664,6 +2664,27 @@ func (e *EgressIPController) addStandByEgressIPAssignment(ni util.NetInfo, podKe
 	return nil
 }
 
+// isSecondaryHostEgressIPReady checks if the node controller has finished configuring
+// a secondary host egress IP on the given node. For secondary host egress IPs (not OVN-managed),
+// the node controller must first configure routes, IP rules, and iptables before OVN can
+// safely route traffic to the host. The node controller signals readiness by adding the
+// egress IP to the k8s.ovn.org/secondary-host-egress-ips annotation.
+func (e *EgressIPController) isSecondaryHostEgressIPReady(node *corev1.Node, egressIP string) (bool, error) {
+	assignedIPs, err := util.ParseNodeSecondaryHostEgressIPsAnnotation(node)
+	if err != nil {
+		if util.IsAnnotationNotSetError(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to parse secondary host egress IPs annotation for node %s: %w", node.Name, err)
+	}
+
+	if !assignedIPs.Has(egressIP) {
+		return false, nil
+	}
+
+	return assignedIPs.Has(egressIP), nil
+}
+
 // addPodEgressIPAssignment will program OVN with logical router policies
 // (routing pod traffic to the egress node) and NAT objects on the egress node
 // (SNAT-ing to the egress IP).
@@ -2693,6 +2714,22 @@ func (e *EgressIPController) addPodEgressIPAssignment(ni util.NetInfo, egressIPN
 		return fmt.Errorf("failed to get node %s egress IP config: %w", eNode.Name, err)
 	}
 	isOVNNetwork := util.IsOVNNetwork(parsedNodeEIPConfig, eIPIP)
+
+	// For secondary host egress IPs, wait for node controller to finish configuration
+	// before programming OVN logical router policies. This prevents the race condition where
+	// OVN routes traffic to the host before host-side routes/IP rules/iptables are ready.
+	if !isOVNNetwork {
+		ready, err := e.isSecondaryHostEgressIPReady(eNode, status.EgressIP)
+		if err != nil {
+			return fmt.Errorf("failed to check if secondary host egress IP %s is ready on node %s: %w",
+				status.EgressIP, eNode.Name, err)
+		}
+		if !ready {
+			return types.NewSuppressedError(fmt.Errorf("won't add OVN logical router policies for secondary host egress IP %s on node %s before node controller configuration",
+				status.EgressIP, status.Node))
+		}
+	}
+
 	nextHopIP, err := e.getNextHop(ni, status.Node, status.EgressIP, egressIPName, isLocalZoneEgressNode, isOVNNetwork)
 	if err != nil {
 		return fmt.Errorf("failed to determine next hop for pod %s/%s when configuring egress IP %s"+
--- a/go-controller/pkg/ovn/egressip_test.go
+++ b/go-controller/pkg/ovn/egressip_test.go
@@ -2,6 +2,7 @@ package ovn
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -498,6 +499,13 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 			logicalRouterOptions := map[string]string{
 				"dynamic_neigh_routers": dynamicNeighRouters,
 			}
+			// Set the secondary host egress IP annotation since this is a secondary host network egress IP
+			if node1.Annotations == nil {
+				node1.Annotations = map[string]string{}
+			}
+			annotationValue, err := json.Marshal([]string{egressIP})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			node1.Annotations[util.OVNNodeSecondaryHostEgressIPs] = string(annotationValue)
 			fakeOvn.startWithDBSetup(
 				libovsdbtest.TestSetup{
 					NBData: []libovsdbtest.TestData{
@@ -549,7 +557,7 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 					Items: []corev1.Namespace{*egressNamespace},
 				})
 
-			err := fakeOvn.controller.WatchEgressIPPods()
+			err = fakeOvn.controller.WatchEgressIPPods()
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 			err = fakeOvn.controller.WatchEgressIPNamespaces()
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
@@ -818,6 +826,7 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 					gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 					fakeOvn.patchEgressIPObj(node2Name, egressIPName, egressIP)
+					fakeOvn.setNodeSecondaryHostEgressIPAnnotation(node2Name, egressIP)
 					gomega.Eventually(getEgressIPStatusLen(egressIPName)).Should(gomega.Equal(1))
 					gomega.Eventually(nodeSwitch).Should(gomega.Equal(node2.Name))
 					egressIPs, _ = getEgressIPStatus(egressIPName)
@@ -1140,6 +1149,7 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 					gomega.Expect(err).NotTo(gomega.HaveOccurred())
 
 					fakeOvn.patchEgressIPObj(node1Name, egressIPName, egressIP)
+					fakeOvn.setNodeSecondaryHostEgressIPAnnotation(node1Name, egressIP)
 
 					gomega.Eventually(getEgressIPStatusLen(egressIPName)).Should(gomega.Equal(1))
 					egressIPs, eIPNodes := getEgressIPStatus(egressIPName)
@@ -1161,6 +1171,7 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 					_, err = fakeOvn.fakeClient.KubeClient.CoreV1().Nodes().Update(context.TODO(), &node2, metav1.UpdateOptions{})
 					gomega.Expect(err).NotTo(gomega.HaveOccurred())
 					fakeOvn.patchEgressIPObj(node2Name, egressIPName, egressIP)
+					fakeOvn.setNodeSecondaryHostEgressIPAnnotation(node2Name, egressIP)
 
 					// sleep long enough for TransactWithRetry to fail, causing egressnode operations to fail
 					// there is a chance that both egressnode events(node1 removal and node2 update) will end up in the same event queue
@@ -1186,6 +1197,7 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 					//	gomega.Not(gomega.BeNil()), // config should not be nil
 					//)
 					fakeOvn.patchEgressIPObj(node2Name, egressIPName, egressIP)
+					fakeOvn.setNodeSecondaryHostEgressIPAnnotation(node2Name, egressIP)
 					connCtx, cancel := context.WithTimeout(context.Background(), config.Default.OVSDBTxnTimeout)
 					defer cancel()
 					resetNBClient(connCtx, fakeOvn.controller.nbClient)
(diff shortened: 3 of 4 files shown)


go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -778,7 +778,23 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
 		}
 	}
 	// apply new changes
+	// CRITICAL: Apply configuration in the correct order to avoid race conditions:
```
bpickard22: Consolidate the comments here just one block explaining why the order matters

go-controller/pkg/ovn/egressip.go
```diff
@@ -2664,6 +2664,32 @@ func (e *EgressIPController) addStandByEgressIPAssignment(ni util.NetInfo, podKe
 	return nil
 }
 
+// isSecondaryHostEgressIPReady checks if the node controller has finished configuring
```
bpickard22, line 2667: too many comments here as well, trim this all down to 2-3 lines

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -15610,6 +15627,155 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+			fakeOvn.controller.logicalPortCache.add(&egressPod, "", types.DefaultNetworkName, "", nil, []*net.IPNet{n})
+			_, err = fakeOvn.fakeClient.KubeClient.CoreV1().Pods(egressPod.Namespace).Create(context.TODO(), &egressPod, metav1.CreateOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			gomega.Consistently(func() int {
+				policies, _ := libovsdbops.FindLogicalRouterPoliciesWithPredicate(fakeOvn.nbClient, func(lrp *nbdb.LogicalRouterPolicy) bool {
+					return lrp.Priority == types.EgressIPReroutePriority &&
+						lrp.Match == fmt.Sprintf("ip4.src == %s", egressPod.Status.PodIP)
+				})
+				return len(policies)
+			}, "500ms", "100ms").Should(gomega.Equal(0))
```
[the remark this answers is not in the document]
bpickard22: +1 - we should call RequestRetryObjs here to validate the real flow

go-controller/pkg/ovn/ovn_test.go
```diff
@@ -694,6 +696,54 @@ func (o *FakeOVN) patchEgressIPObj(nodeName, egressIPName, egressIP string) {
...
+		}
+
+		// Add the new IP
+		existingIPs.Insert(egressIP)
+
+		// Update the annotation
+		if node.Annotations == nil {
+			node.Annotations = make(map[string]string)
+		}
+		ipsSlice := sets.List(existingIPs)
```
bpickard22, line 731: we shouldnt manually build the annotation here. We typically use json.Marshall on a []string to parse into an annotation to avoid json errors like special characters. We also use marshaling on our production code on an unsorted list, and here you are creating it manually

could do something like 
`annotationBytes, err := json.Marshal(sets.List(existingIPs))    gomega.Expect(err).NotTo(gomega.HaveOccurred())
node.Annotations[util.OVNNodeSecondaryHostEgressIPs] = string(annotationBytes)`

which reads a lot cleaner and follows our typical annotation creation format

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -498,6 +498,11 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 			logicalRouterOptions := map[string]string{
 				"dynamic_neigh_routers": dynamicNeighRouters,
 			}
+			// Set the secondary host egress IP annotation since this is a secondary host network egress IP
+			if node1.Annotations == nil {
+				node1.Annotations = make(map[string]string)
```
bpickard22: similar annotation comment here
should use marshaling 
`json.Marshal([]string{egressIP})`

bpickard22 (changes requested):
other than my inline comments, I would like to know what happends on EIP failover.
Like if node 1 goes down and we migrate to node 2, what happens to the annotation?
could the annotation on node 1 still contain the old ip. 

I can see that we are testing the EIP assignment path,  but maybe we should add some testing for the failover/cleanup path (maybe here or in a separate pr) so that we can make sure that annotation is getting cleaned up correctly on the old node

Otherwise I agree with the logic/approach

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -778,7 +778,23 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
...
+	// 1. Add routes to custom routing table FIRST
```
shreyasbe: Readability suggestion! Adding a STEP_ prefix (or a similar contextual identifier) to numeric sequences(e.g. 1 -> STEP_1) often makes the intent much clearer, especially when those numbers represent process stages, workflow order, or configuration keys.

Same applies to elsewhere.

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -778,7 +778,23 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
...
+	// 2. Add IP rules (which reference the routing table)
+	// 3. Add iptables SNAT rules
+	// 4. Add IP address to interface (all host-side config ready)
+	// 5. Add IP to annotation LAST (readiness signal for OVN controller)
+	// This ensures all routing/NAT infrastructure is ready before OVN routes traffic to the host.
 	if update != nil && update.eIPConfig != nil && update.eIPConfig.addr != nil && len(update.eIPConfig.routes) > 0 {
+		// STEP 1: Add routes to custom routing table FIRST
+		for _, routeToAdd := range update.eIPConfig.routes {
+			if err := c.routeManager.Add(routeToAdd); err != nil {
+				return fmt.Errorf("failed to add egress IP route: %w", err)
```
shreyasbe: Better to add the route details which is failed: ("failed to add egress IP route %v, err: %w", routeToAdd, err)

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -798,22 +814,17 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
...
 		existing.eIPConfig.addr = update.eIPConfig.addr
-		// route manager manages retry
-		for _, routeToAdd := range update.eIPConfig.routes {
-			if err := c.routeManager.Add(routeToAdd); err != nil {
-				return err
-			}
+
+		// STEP 5: Add IP to annotation (readiness signal for OVN controller)
+		if err := c.addIPToAnnotation(update.eIPConfig.addr.IP.String()); err != nil {
+			return fmt.Errorf("failed to add egress IP address to annotation: %v", err)
```
shreyasbe, line 819: Don't we need to rollback previous applied steps, (i.e. STEP_1 to STEP_4) in reverse order, before return error ?

shreyasbe (commented):
### Testing Recommendations
**Concurrency test**: Validate behavior when multiple EgressIP assignments occur simultaneously on the same node

go-controller/pkg/node/controllers/egressip/egressip.go
arghosh93: Not needed as object retry mechanism is gonna retry adding annotation upon getting an error.

go-controller/pkg/ovn/egressip_test.go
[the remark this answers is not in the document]
arghosh93: I tried implementing this, but the issue is that the test expects the pod to be automatically added to the retry queue, but this isn't happening in the test environment. The original commit (ed5c3c07f) manually added the pod to the retry queue, which CodeRabbit suggested we change. However, in the test environment, the automatic retry queuing doesn't seem to trigger.
Do you have a better suggestion?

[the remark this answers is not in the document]
arghosh93: @coderabbitai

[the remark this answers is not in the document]
arghosh93: Thank you so much!

arghosh93:
> other than my inline comments, I would like to know what happends on EIP failover. Like if node 1 goes down and we migrate to node 2, what happens to the annotation? could the annotation on node 1 still contain the old ip.

updateEIP() function takes care of node side cleanup if required before adding/syncing node resources.

shreyasbe (approved):
LGTM, Approved

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -778,7 +778,18 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
 		}
 	}
 	// apply new changes
+	// Apply configuration in strict order to avoid race conditions: routes → IP rules → iptables → IP on interface → annotation.
```
bpickard22, line 781: we can remove the step 1,2,3 lines, this top level comment is enough of an explanation of the ordering

go-controller/pkg/node/controllers/egressip/egressip.go
```diff
@@ -798,22 +809,17 @@ func (c *Controller) updateEIP(existing *state, update *config) error {
 				}
 			}
 		}
-		if err := c.addIPToAnnotation(update.eIPConfig.addr.IP.String()); err != nil {
-			return fmt.Errorf("failed to add egress IP address to annotation: %v", err)
-		}
-		// TODO(mk): only apply the follow when its new config or when it failed to apply
```
bpickard22, line 807: does this pr handle this todo?

go-controller/pkg/ovn/egressip.go
```diff
@@ -2664,6 +2664,27 @@ func (e *EgressIPController) addStandByEgressIPAssignment(ni util.NetInfo, podKe
...
+// a secondary host egress IP on the given node. For secondary host egress IPs (not OVN-managed),
+// the node controller must first configure routes, IP rules, and iptables before OVN can
+// safely route traffic to the host. The node controller signals readiness by adding the
+// egress IP to the k8s.ovn.org/secondary-host-egress-ips annotation.
+func (e *EgressIPController) isSecondaryHostEgressIPReady(node *corev1.Node, egressIP string) (bool, error) {
```
bpickard22, line 2672: Should we make this event driven instead of relying on the retry queue? 

We could just add a check to the update handler in the network controller that checks if the node annotation has changed to trigger a reconciliation. 

using the eip retry framework will work, but worst case is that it fails a few times and then hits a pretty long backoff, or we just fail 15 times then it gets dropped from the workqueue and nothing happens. 

something like 
`if oldNode.annotations[ovnNodeSecondaryHostEgressIP] != newNode.annotations[ovnNodeSecondaryHostEgressIP] {
    trigger reconciliation 
}
` 

so that we can just immediately re-queue instead of relying on the retry timer

arghosh93: EgressIPController relies on the retry mechanism.  Moreover, this step is required only for the secondary egress IP. I don't think it makes much sense to modify the network controller only for secondary egress IP and create another lane for reconciliation, where we try to reconcile using both the retry mechanism and an event driven method.
In many other places, where we wait for the annotation to appear and then continue programming logical stuff, and if the annotation is not available, then return an error so that the retry mechanism tries again after 5 second I guess.

go-controller/pkg/node/controllers/egressip/egressip.go
arghosh93: I removed this Todo as it seems an idempotent operation and no need to treat it differently. But you are right, let me put it back as I have not addressed the Todo literally.

bpickard22:
logic and implementation looks good to me, but I want to see the unit tests pass before I lgtm @arghosh93

arghosh93:
All unit tests are passing locally all the time. Not sure what is going on here. In upstream channel I do see a thread about test-pr job failing frequently.
https://cloud-native.slack.com/archives/C08452HR8V6/p1775653106636329

bpickard22 (approved):

go-controller/pkg/ovn/egressip.go
```diff
@@ -2659,6 +2664,27 @@ func (e *EgressIPController) addStandByEgressIPAssignment(ni util.NetInfo, podKe
...
+			return false, nil
+		}
+		return false, fmt.Errorf("failed to parse secondary host egress IPs annotation for node %s: %w", node.Name, err)
+	}
+
+	if !assignedIPs.Has(egressIP) {
+		return false, nil
+	}
+
+	return true, nil
```
jcaamano: nit: `return assignedIPs.Has(egressIP), nil`

No need to change unless there is other reason to push

go-controller/pkg/ovn/egressip.go
```diff
@@ -2688,6 +2714,22 @@ func (e *EgressIPController) addPodEgressIPAssignment(ni util.NetInfo, egressIPN
...
+	// OVN routes traffic to the host before host-side routes/IP rules/iptables are ready.
+	if !isOVNNetwork {
+		ready, err := e.isSecondaryHostEgressIPReady(eNode, status.EgressIP)
+		if err != nil {
+			return fmt.Errorf("failed to check if secondary host egress IP %s is ready on node %s: %w",
+				status.EgressIP, eNode.Name, err)
+		}
+		if !ready {
+			return fmt.Errorf("waiting for node controller to complete host-side configuration (routes/IP rules/iptables) "+
+				"for secondary host egress IP %s on node %s before programming OVN logical router policies", status.EgressIP, status.Node)
```
jcaamano: Can you be more concise with this error message? Like "won't add OVN logical router policies for secondary  host egress IP before node controller configuration"

Also, have you considered to just log and listen for node events, and process the EIP once it is annotated.
If that is unreasonable, since we expect this to happen a few times, should we suppress the error on retries and only log the error when we will no longer retry? (hint: check `IsSuppressedError`)

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -15610,6 +15630,163 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
 			gomega.Expect(err).NotTo(gomega.HaveOccurred())
 		})
 	})
+
+	ginkgo.Context("Secondary host egress IP annotation readiness", func() {
+		ginkgo.It("should wait for node annotation before programming OVN logical components", func() {
```
jcaamano, line 15635: Can you document a bit the different steps of the test with `ginko.By`? Also helps with the review.

go-controller/pkg/ovn/egressip_test.go
```diff
@@ -15610,6 +15630,163 @@ var _ = ginkgo.Describe("OVN master EgressIP Operations cluster default network"
...
+			if node1.Annotations == nil {
+				node1.Annotations = map[string]string{}
+			}
+			annotationValue, err := json.Marshal([]string{egressIP})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+			node1.Annotations[util.OVNNodeSecondaryHostEgressIPs] = string(annotationValue)
+			_, err = fakeOvn.fakeClient.KubeClient.CoreV1().Nodes().Update(context.TODO(), &node1, metav1.UpdateOptions{})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			// 2) Trigger a no-op pod update to drive the real reconcile path (no manual retry seeding).
```
jcaamano: Why trigger the no-op? Isn't this supposed to be handled by retries?

go-controller/pkg/ovn/egressip.go
arghosh93: I think listening for node events will add complexity and create a parallel path to act on egressNode events. 
Suppressing the error makes sense to me, as customers don't need to be bothered by this error, as it is expected.
So, I will use `IsSuppressedError` and will try to write a more specific error message.

go-controller/pkg/ovn/egressip_test.go
arghosh93: okay, sure

go-controller/pkg/ovn/egressip_test.go
arghosh93: okay, let me use fakeOvn.controller.retryEgressIPPods.RequestRetryObjs() instead.

arghosh93:
I just want to note here that this fix makes sure that the node controller finishes its job before we program logical router policies on the OVN side during egressIP creation.
There is another problem where if a new egresIPPOD is created after the egress IP for secondary host network was created, then iptables and IP rule reconciliation happen for all existing egress PODs. Since this reconciliation occurs sequentially for every POD in the namespace along with the new POD, the node controller introduces a delay when there is a large number of PODs present in the namespace. 
On the other hand, LRP on the OVN side gets created fast, as we already have the annotation on the node. I have raised the following PR to reduce the time taken by node controller to reconcile iptables and IP rules.
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6289

jcaamano:
> I just want to note here that this fix makes sure that the node controller finishes its job before we program logical router policies on the OVN side during egressIP creation. There is another problem where if a new egresIPPOD is created after the egress IP for secondary host network was created, then iptables and IP rule reconciliation happen for all existing egress PODs. Since this reconciliation occurs sequentially for every POD in the namespace along with the new POD, the node controller introduces a delay when there is a large number of PODs present in the namespace. On the other hand, LRP on the OVN side gets created fast, as we already have the annotation on the node. I have raised the following PR to reduce the time taken by node controller to reconcile iptables and IP rules. #6289

@arghosh93 I am not sure I understand what the problem is and how #6289 solves It.

> the node controller introduces a delay

Is this the problem? A delay of what?

go-controller/pkg/ovn/egressip.go
jcaamano: > will add complexity and create a parallel path to act on egressNode events

Can you be more specific with this concern? A specific node would not be handled concurrently and the controller is already handling concurrently different types of events.

arghosh93:
> @arghosh93 I am not sure I understand what the problem is and how #6289 solves It.
> 
> > the node controller introduces a delay
> 
> Is this the problem? A delay of what?

When there are many PODs in a namespace and a secondary EgressIP is being created for that namespace, the OVN controller immediately creates logical router policy to redirect POD traffic to the host via the management port. however node EIP controller takes time to configure iptables SNAT and IP rules as it does reconciliation for all existing PODs in the namespace sequentially. In this PR I am trying to address this issue by waiting for the annotation to appear on the node resource. This makes sure that node kernel rules are configured before we program the reroute policy at the OVN side when the EIP is being created.  

However this PR does not address sequential processing of iptables SNAT rules and IP rules for every PODs in the egressIP namespace and thus the delay in configuring secondary egressIP.  [Nested loop](https://github.com/arghosh93/ovn-kubernetes/blob/3dbe4a83e6679c73dc6e1004040629fac1c4aec6/go-controller/pkg/node/controllers/egressip/egressip.go#L791-L809) mostly introduces delay when there are large numbers of PODs in EgressIP namespaces. 

There is another issue where reconciliation happens for all existing PODs when a new POD gets created after egressIP creation. node EIP controller's syncPod() function triggers IP reconciliation and  [Nested loop](https://github.com/arghosh93/ovn-kubernetes/blob/3dbe4a83e6679c73dc6e1004040629fac1c4aec6/go-controller/pkg/node/controllers/egressip/egressip.go#L791-L809) does reconciliation for each PODs. However on the ovn EIP controller side it checks for the annotation on the node and since it is already present, ovn EIP controller immediately creates a reroute policy. As a result POD IP gets leaked until the iptables SNAT rule gets created. This part of the problem is yet to be fixed. The following could be a way forward:

-  Make sure EIP node controller only tries to create/reconcile iptables rules and iprules for the new POD when secondary EIP is already configured. Probably need to maintain a cache to identify new POD or use existing ones

- Still, for a fraction of a second, we can have reroute policy in place at ovn side but iptables SNAT rule and IP rules are absent. Probably need to find a way to drop traffic in that situation.

raphaelvrosa:
@arghosh93 can we close this?

arghosh93:
yes.
/close

Issue #5827, opened 2025-12-30
network traffic patterns which are broken on a DPU enabled host with management and data network


aserdean:
The following traffic patterns are broken on a DPU enabled system that has two networks management and data.

```
    "test_case_id": "HOST_TO_CLUSTER_IP_TO_HOST_DIFF_NODE",
    "test_case_id": "HOST_TO_NODE_PORT_TO_HOST_DIFF_NODE",
    "test_case_id": "POD_TO_CLUSTER_IP_TO_HOST_DIFF_NODE",
    "test_case_id": "POD_TO_HOST_DIFF_NODE",
```

```
k get nodes -o wide
NAME               STATUS   ROLES                         AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                                                KERNEL-VERSION                 CONTAINER-RUNTIME
ocp-host-worker1   Ready    worker                        14d   v1.33.5   10.0.110.11   <none>        Red Hat Enterprise Linux CoreOS 9.6.20251110-1 (Plow)   5.14.0-570.64.1.el9_6.x86_64   cri-o://1.33.5-3.rhaos4.20.gitd0ea985.el9
ocp-host-worker2   Ready    worker                        14d   v1.33.5   10.0.110.12   <none>        Red Hat Enterprise Linux CoreOS 9.6.20251110-1 (Plow)   5.14.0-570.64.1.el9_6.x86_64   cri-o://1.33.5-3.rhaos4.20.gitd0ea985.el9
ocp-master1        Ready    control-plane,master,worker   14d   v1.33.5   10.0.110.1    <none>        Red Hat Enterprise Linux CoreOS 9.6.20251110-1 (Plow)   5.14.0-570.64.1.el9_6.x86_64   cri-o://1.33.5-3.rhaos4.20.gitd0ea985.el9

k get node ocp-host-worker1 -o yaml | grep k8s.ovn.org/host-cidrs
    k8s.ovn.org/host-cidrs: '["10.0.110.11/24","10.0.120.2/29"]'

k get node ocp-host-worker2 -o yaml | grep k8s.ovn.org/host-cidrs
    k8s.ovn.org/host-cidrs: '["10.0.110.12/24","10.0.120.10/29"]'

k get node ocp-master1 -o yaml | grep k8s.ovn.org/host-cidrs
    k8s.ovn.org/host-cidrs: '["10.0.110.1/24"]'
```

Adding network details of the first host
```
[...]

PR #6172, opened 2026-03-27
node: add cross-node host IP routing for multi-homed IC nodes

aserdean:
In DPU setups, worker nodes are multi-homed — they have a management network IP (accessible via the physical uplink) and a data network IP (reachable only through the directly cabled DPU uplink). In IC mode, a pod on zone A trying to reach zone B's data network host IP has no OVN path since there are no routes or PBR rules for it.

Add PBR rules at InterNodePolicyPriority (1003) and cluster router static routes so pod-to-host cross-zone traffic is steered through the transit switch to the destination node's mp0. Add GR static routes and masquerade SNAT so host-originated traffic to remote zone host IPs traverses the join switch instead of the physical network.

Rules are only installed when a node has 2+ host IPs in the same IP family (NodeIsMultiHomed), leaving single-homed deployments unaffected.

Fixes: https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5827

## Summary by CodeRabbit

* **New Features**
  * Detect multi-homed nodes and automatically program cross-node host-IP policies/routes and shared-gateway SNATs, with cleanup when no longer needed.

* **Bug Fixes**
  * More consistent programming/removal of cross-node PBRs and host-IP routes, broadened matching for deletions, and triggering interconnect sync when remote-node host CIDRs change.

* **Tests**
  * Added tests validating cross-node host-IP policies, routes, multi-homed gating, and multi-homed detection.

### DIFF
--- a/go-controller/pkg/ovn/controller/services/node_tracker.go
+++ b/go-controller/pkg/ovn/controller/services/node_tracker.go
@@ -120,6 +120,7 @@ func (nt *nodeTracker) Start(nodeInformer coreinformers.NodeInformer) (cache.Res
 			// . No need to trigger update for any other field change.
 			if util.NodeSubnetAnnotationChangedForNetwork(oldObj, newObj, nt.netInfo.GetNetworkName()) ||
 				util.NodeL3GatewayAnnotationChanged(oldObj, newObj) ||
+				util.NodeChassisIDAnnotationChanged(oldObj, newObj) ||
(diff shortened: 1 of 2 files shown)


aserdean:
CC: @trozet @tsorya @wizhaoredhat

aserdean:
Dropping in favor of the APBR, discussed under https://cloud-native.slack.com/archives/C08BQDZAJH0/p1760537400093069

PR #6174, opened 2026-03-27
Fix DPU healthcheck by propagating namespace configuration to ovnkube

tsorya:
Fix DPU healthcheck by propagating namespace configuration to ovnkube
In DPU (Data Processing Unit) deployments, the OVN-Kubernetes namespace
may differ from the default "ovn-kubernetes" (e.g. when deployed via
operators that use a custom namespace). The ovnkube.sh entrypoint script
already reads OVN_KUBERNETES_NAMESPACE from the environment, but never
actually passed it through to the ovnkube binary via --ovn-config-namespace.
This caused DPU healthcheck failures because the ovnkube processes were
always looking for resources in the hardcoded "ovn-kubernetes" namespace,
while the actual resources (leases, services) lived in the
operator-configured namespace.

This change:
- Adds --ovn-config-namespace ${ovn_kubernetes_namespace} to all six
  ovnkube binary invocations: ovn-master, ovnkube-controller,
  ovnkube-controller-with-node, ovn-cluster-manager, ovn-node, and
  cleanup-ovn-node.
- Replaces the hardcoded "ovn-kubernetes" namespace in ovnkube-identity
  --extra-allowed-user service account references with the configurable
  ${ovn_kubernetes_namespace} variable.

Without this fix, DPU nodes fail healthchecks when deployed in a
non-default namespace because leader election leases target the wrong
namespace.

Signed-off-by: Igal Tsoiref <itsoiref@redhat.com>
Made-with: Cursor

## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/dist/images/ovnkube.sh
+++ b/dist/images/ovnkube.sh
@@ -1245,8 +1245,8 @@ ovnkube-identity() {
     --webhook-cert-dir="/etc/webhook-cert" \
     ${ovnkube_enable_interconnect_flag} \
     ${ovnkube_enable_hybrid_overlay_flag} \
-    --extra-allowed-user="system:serviceaccount:ovn-kubernetes:ovnkube-cluster-manager" \
-    --extra-allowed-user="system:serviceaccount:ovn-kubernetes:ovnkube-master" \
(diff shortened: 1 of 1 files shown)


aserdean:
Sign the commit.

aserdean:
LGTM

PR #6184, opened 2026-03-31
OKEP-6181: Allow UDN/CUDN to Generate Multiple NADs

l8huang:
Fixes #6181
 

## Summary by CodeRabbit

* **Documentation**
  * Added OKEP-6181 enhancement proposal documentation describing support for generating multiple Network Attachment Definitions (NADs) from a single UDN/CUDN configuration via an `attachments` list. Includes API specifications, deterministic naming conventions, expected controller behavior, runtime selection guidance, and backwards compatibility assurances for existing configurations.

### DIFF
--- a/docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
+++ b/docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
@@ -1,0 +1,320 @@
+# OKEP-6181: Allow UDN/CUDN to Generate Multiple NADs
+
+* Issue: [#6181](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6181)
+
+## Problem Statement
+
+Today, a UDN generates one NAD in its namespace, and a CUDN generates one NAD in each selected
+namespace. This makes it difficult to support accelerated (VF/VFIO) and unaccelerated (veth)
+workloads on the same logical network, or workloads that use different SR-IOV resource pools,
+without creating separate logical networks.
+
+## Goals
+
+* Allow primary and secondary UDN/CUDN to declare multiple named attachments, each generating a
+  distinct NAD for the same logical network.
+* Preserve existing Multus/SR-IOV semantics by continuing to use the
+  `k8s.v1.cni.cncf.io/resourceName` annotation at the NAD level.
+* Support accelerated and unaccelerated workloads, or workloads that use different SR-IOV resource
+  pools, on the same UDN/CUDN by choosing different NADs.
+
+## Non-Goals
+
+* Changing SR-IOV device allocation behavior in Multus/SR-IOV plugins.
+* Implementing or owning workload resource mutation logic. Existing mutating webhooks that add
+  `resources.requests` for SR-IOV pools are outside OVN-Kubernetes, but they may need updates to
+  understand primary network NAD selection described by this OKEP.
+* Redefining CNI semantics for resource requests.
+* Changing cluster default network (CDN) behavior or introducing CDN-specific default-network
+  selection semantics.
+
+
+## Introduction
+
+Today, each UDN/CUDN generates only one NAD per namespace. When a workload selects a NAD, Multus
+uses that NAD to determine the SR-IOV resource pool for the attachment: if the NAD has the
+`k8s.v1.cni.cncf.io/resourceName` annotation, the annotation identifies the resource pool used for
+the corresponding CNI attachment.
+
+However, for a single logical network, different workloads often need different network attachments:
+
+* Unaccelerated/default attachment (no SR-IOV resource).
+* Pod-accelerated attachment (typically VF resource pool).
+* VM-accelerated attachment (typically VFIO resource pool).
+
+Without multi-NAD generation from one UDN/CUDN, users cannot use one logical network with different
+resource pools. That increases operational complexity and obscures intent.
+
+This proposal introduces an `attachments` API in UDN/CUDN to generate multiple NADs for one
+logical network while preserving current behavior for existing UDN/CUDN objects.
+
+## User-Stories/Use-Cases
+
+Story 1: Secondary network with mixed acceleration
+
+As a cluster admin, I want a secondary UDN to generate a default NAD, a VF NAD, and a VFIO NAD so
+that Pod and VM workloads can share the same logical network while selecting the right acceleration
+mode.
+
+Story 2: Primary network with optional acceleration path
+
+As a cluster admin, I want a primary CUDN to expose both default and accelerated generated NADs
+so that tenants can opt into acceleration by selecting the right NAD for the workload.
+
+Story 3: Predictable network selection
+
+As an application operator, I want generated NAD names to be deterministic and discoverable from
+UDN/CUDN spec so that network annotation authoring is straightforward and stable.
+
+## Proposed Solution
+
+### attachments definition in UDN/CUDN
+Add an `attachments` list to UDN/CUDN. Each attachment entry defines the full name of a generated
+NAD and an optional SR-IOV resource mapping. OVN-Kubernetes generates one NAD per attachment name,
+and all generated NADs use the same logical network topology/subnet settings from the parent
+UDN/CUDN.
+
+A user or workload selects the generated NAD it needs, as described in the next section.
+Multus/SR-IOV behavior remains unchanged: the `k8s.v1.cni.cncf.io/resourceName` annotation on the
+selected NAD drives resource allocation.
+OVN-Kubernetes CNI continues to consume the resulting CNI ADD request as it does today.
+
+For primary networks, selecting an accelerated generated NAD through
+`v1.multus-cni.io/default-network` can affect which SR-IOV resource pool the workload must request.
+OVN-Kubernetes does not mutate Pod's resource requests. Any mutating webhook that is
+responsible for adding SR-IOV `resources.requests` must be updated outside OVN-Kubernetes to inspect
+the same `v1.multus-cni.io/default-network` annotation, including the `primary-network` field, and
+add the matching resource request when the selected generated NAD carries
+`k8s.v1.cni.cncf.io/resourceName`.
+
+
+### Pod network selection
+
+OVN-Kubernetes currently supports configuring pods' secondary network interfaces through the
+`k8s.v1.cni.cncf.io/networks` annotation, which contains a JSON array of NetworkSelectionElement
+objects.
+
+For pods' primary network, this proposal follows the model used by OKEP-5233. The
+`v1.multus-cni.io/default-network` annotation continues to reference the default-network NAD,
+and may include a new optional `primary-network` field. Its value is the full name of a
+generated NAD from the primary UDN/CUDN.
+
+OVN-Kubernetes uses `primary-network` to choose the generated NAD for the pod's primary
+network attachment. If the field is omitted, OVN-Kubernetes uses the generated base NAD.
+The namespace still has only one primary logical network; the generated NADs provide multiple
+network attachments for that same logical network.
+
+Example primary UDN/CUDN pod selection:
+
+```yaml
+metadata:
+  annotations:
+    v1.multus-cni.io/default-network: |
+      [
+        {
+          "name":"default",
+          "namespace":"ovn-kubernetes",
+          "primary-network":"l3-primary-pod-accelerated"
+        }
+      ]
+```
+
+### API Details
+
+#### CUDN/UDN API Changes
+
+The `attachments` field is added at these paths:
+
+* `UserDefinedNetwork.spec.attachments`
+* `ClusterUserDefinedNetwork.spec.network.attachments`
+
+A sample CUDN API using `attachments` is shown below:
+```yaml
+apiVersion: k8s.ovn.org/v1
+kind: ClusterUserDefinedNetwork
+metadata:
+  name: l3-primary
+spec:
+  namespaceSelector:
+    matchLabels:
+      kubernetes.io/metadata.name: udn-test
+  network:
+    topology: Layer3
+    layer3:
+      role: Primary
+      subnets:
+      - cidr: 10.20.0.0/16
+        hostSubnet: 24
+    transport: Geneve
+    attachments:
+    - name: l3-primary
+    - name: l3-primary-pod-accelerated
+      resourceName: nvidia.com/asap2_vf
+    - name: l3-primary-vm-accelerated
+      resourceName: nvidia.com/asap2_vfio
+```
+
+#### Field definitions
+
+* `attachments` (optional list)
+  * If omitted, behavior is identical to today: OVN-Kubernetes creates the generated base NAD only.
+  * If present, each item defines one generated network attachment.
+  * Annotations that are copied from the UDN/CUDN to the generated base NAD today are copied to all
+    generated NADs.
+* `attachments[].name` (required, string)
+  * Unique within the UDN/CUDN.
+  * Full name of the generated NAD.
+  * If equal to the UDN/CUDN name, the attachment configures the generated base NAD.
+* `attachments[].resourceName` (optional, string)
+  * When set, copied to generated NAD annotation `k8s.v1.cni.cncf.io/resourceName`.
+
+#### Generated NAD naming
+
+Deterministic naming is required:
+
+* UDN always generates a generated base NAD named after the UDN in the UDN namespace.
+* CUDN always generates a generated base NAD named after the CUDN in each target namespace.
+* Each `attachments[].name` is the full generated NAD name.
+* If `attachments[].name` equals the UDN/CUDN name, no additional NAD is generated for that entry.
+  Its settings are applied to the generated base NAD.
+* If `attachments[].name` differs from the UDN/CUDN name, it must start with the UDN/CUDN name.
+  OVN-Kubernetes generates an additional NAD with that exact name in the relevant namespace.
+
+#### Validation
+
+* `attachments[].name` must be DNS-1123 label compatible.
+* `attachments[].name` must be unique.
+* `attachments[].name` must either equal the UDN/CUDN name, or start with the UDN/CUDN name and be
+  distinct from the UDN/CUDN name.
+* `resourceName` should follow existing resource naming rules used by the Multus/SR-IOV ecosystem.
+
+### Implementation Details
+
+#### Controller changes
+
+UDN/CUDN NAD generation logic is extended to:
+
+1. Parse `attachments` if present.
+2. Always generate the base NAD.
+3. Generate one additional NAD per attachment whose `name` differs from the UDN/CUDN name.
+4. Treat all generated NADs as network attachments for the same logical network.
+5. Materialize the `k8s.v1.cni.cncf.io/resourceName` annotation based on `attachments[].resourceName`.
+6. Maintain owner references and reconciliation behavior for all generated NADs.
+
+If `attachments` is not present, the current single-NAD generation path remains unchanged.
+
+The NAD/network manager also needs to understand that multiple generated NADs can represent the
+same primary logical network in a namespace. Internal caches and helper functions that currently
+track a single primary NAD per namespace must be updated to:
+
+1. Allow multiple primary NADs in the same namespace when their parsed network configs are
+   compatible and point to the same logical network.
+2. Keep the generated base NAD for default behavior when no explicit attachment is selected.
+3. Track all generated NAD keys associated with the primary network so pod selection and
+   reconciliation can validate that a selected attachment belongs to the namespace's primary
+   UDN/CUDN.
+
+Because the namespace still has one primary logical network, other controllers that request the
+namespace's primary network can keep using the generated base NAD.
+
+Attachment reconciliation follows declarative ownership semantics:
+
+1. On create/update, the controller creates or updates the generated NAD set to match `attachments`.
+2. If an attachment is removed from `attachments`, the corresponding generated NAD is deleted.
+3. There is no special rename operation. A rename is represented as deleting the old attachment name
+   and creating a new attachment name.
+4. Updating `resourceName` updates only the generated NAD annotation. The logical network config remains
+   unchanged because all generated NADs point to the same UDN/CUDN network.
+5. Existing pods keep using the NAD selected at creation time. Workload rollout/recreation is required
+   to consume a different network attachment.
+
+#### Pod/VM network selection behavior
+
+Secondary network:
+
+* User specifies generated NAD name in `k8s.v1.cni.cncf.io/networks`.
+* Example:
+
+```yaml
+metadata:
+  annotations:
+    k8s.v1.cni.cncf.io/networks: default/l3-primary-pod-accelerated
+```
+
+Primary network:
+
+* The existing default primary path remains supported.
+* A namespace still has only one primary logical network.
+* Generated primary NADs are network attachments for that same primary logical network.
+* `v1.multus-cni.io/default-network` continues to reference the operator-chosen default-network NAD.
+* The optional `primary-network` field in the annotation JSON selects the full name of the generated
+  primary NAD to use as the network attachment.
+* If `primary-network` is omitted, OVN-Kubernetes uses the generated base NAD.
+* If `primary-network` references a generated NAD that does not belong to the pod's primary UDN/CUDN,
+  pod setup fails validation.
+
+#### Datapath and mode impact
+
+* No intended datapath change.
+
+### Testing Details
+
+* Unit testing details:
+  * Reconciler tests for create/update/delete of multiple generated NADs.
+  * Naming and collision handling tests.
+  * Primary network selection tests for `primary-network` omitted, valid, and invalid values.
+* E2E testing details:
+  * Secondary UDN with default + VF + VFIO attachments; verify Pod/VM connectivity and expected
+    resource behavior.
+  * Primary network selection scenarios with and without explicit `primary-network` selection.
+  * Attachment deletion and `resourceName` update scenarios.
+* API testing details:
+  * API validation tests for `attachments`.
+  * Conflict/rejection tests for invalid attachment names and duplicates.
+
+### Documentation Details
+
+* Add a user-facing guide for `attachments`:
+  * How to declare default and accelerated network attachments.
+  * How to select generated NADs from Pod and VM manifests.
+  * Primary network selection examples using `v1.multus-cni.io/default-network` with `primary-network`.
+* Update OKEP index in `mkdocs.yml` with this document path.
+
+## Risks, Known Limitations and Mitigations
+
+* Risk: Primary network selection confusion because `v1.multus-cni.io/default-network` references one
+  NAD while `primary-network` selects another generated NAD.
+  * Mitigation: document the annotation usage clearly and validate that `primary-network` belongs
+    to the pod's primary UDN/CUDN.
+* Risk: Existing SR-IOV resource mutating webhooks may only inspect `k8s.v1.cni.cncf.io/networks`.
+  * Mitigation: document that those webhooks are not part of OVN-Kubernetes and must be updated to
+    inspect `v1.multus-cni.io/default-network` and its `primary-network` field for accelerated
+    primary network attachments.
+* Limitation: Users still need to choose the correct attachment for each workload type.
+  * Mitigation: keep semantics explicit now; future automation or mutating webhooks remain possible
+    but out of scope.
+
+## OVN-Kubernetes Version Skew
+
+Target release: TBD (to be aligned with maintainers and release planning).
+
+Version skew considerations:
+
+* New `attachments` field requires a controller version that understands it.
+* Existing objects without `attachments` continue to function.
+
+## Backwards Compatibility
+
+* Backward compatible by default:
+  * Existing UDN/CUDN objects continue with current single-NAD behavior when `attachments`
+    is absent.
+* API additive change:
+  * Adds a new optional field; no removal of existing fields in this proposal.
+* Existing workload manifests selecting current NAD names remain valid unless operators opt into
+  `attachments`.
+
+## Alternatives
+
+* Keep single NAD per UDN/CUDN and encode more logic elsewhere.
+  * Rejected because it does not natively model multiple resource pools on one logical network.
+
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -176,5 +176,6 @@ nav:
     - EVPN: okeps/okep-5088-evpn.md
     - DPU Healthcheck: okeps/okep-5674-dpu-healthcheck.md
     - Extend Primary UDN/CUDN to Support Multiple Cluster Subnets in Layer3 Topology: okeps/okep-5377-extend-udn-to-support-multiple-cluster-subnets-in-layer3-topology.md
+    - Allow UDN/CUDN to Generate Multiple NADs: okeps/okep-6181-allow-udn-generating-multiple-nads.md
   - Blog:
     - blog/index.md


docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
```diff
@@ -0,0 +1,296 @@
...
+
+* `spec.attachments` (optional list)
+  * If omitted, behavior is identical to today: OVN-Kubernetes creates the generated base NAD only.
+  * If present, each item defines one generated network attachment.
+* `attachments[].name` (required, string)
+  * Unique within the UDN/CUDN.
+  * Full name of the generated NAD.
+  * If equal to the UDN/CUDN name, the attachment configures the generated base NAD.
+* `attachments[].resourceName` (optional, string)
+  * When set, copied to generated NAD annotation `k8s.v1.cni.cncf.io/resourceName`.
```
l8huang, line 169: @npinaeva do you want to add `DeviceClass` for https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6446 docs/okeps/okep-6391-dra.md?

npinaeva:
This is a cool use case that I didn't know exists. I think it shows the limitations of multus: using a given device doesn't really need a new NAD for each device type. Different devices can be connected to the same network as you are saying, so it would be quite nice instead of pre-creating 3 NADs with the same spec, just say at the pod level that a given network attachment should use a given device, like `k8s.v1.cni.cncf.io/networks: "l3-primary, l3-primary:nvidia.com/asap2_vf, l3-primary:nvidia.com/asap2_vfio"` that would read <network-to-attach-to:optional device>.
Considering multus doesn't support that,  we don't have many options (even with the DRA implementation as long as we use multus for multi-networking) :( 
Considering this is pretty ugly, maybe we could put the additional-NAD-creation request into an annotation instead of a proper API field, and hope that we will be able to get rid of it one day? Considering this feature only adds extra things on top of UDN and doesn't change the normal UDN behaviour we could do something like
```
apiVersion: k8s.ovn.org/v1
metadata:
  name: l3-primary
  annotations:
    k8s.ovn.org/nads-with-resources: "nvidia.com/asap2_vf, nvidia.com/asap2_vfio"
  # or for DRA
   k8s.ovn.org/nads-with-device-classes: "asap2_vf, asap2_vfio"
spec:
  namespaceSelector:
    matchLabels:
      kubernetes.io/metadata.name: udn-test
  network:
    topology: Layer3
    layer3:
      role: Primary
      subnets:
      - cidr: 10.20.0.0/16
        hostSubnet: 24
    transport: Geneve
```

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
npinaeva: yes! considering we don't know in which order this will be done, it would be nice to mention

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
```diff
@@ -0,0 +1,296 @@
...
+Example primary UDN/CUDN pod selection:
+
+```yaml
+metadata:
+  annotations:
+    v1.multus-cni.io/default-network: |
+      {
+        "name":"default",
+        "namespace":"ovn-kubernetes",
+        "primary-network":"l3-primary-pod-accelerated"
```
npinaeva: I didn't get that part: this field doesn't exist int he NetworkSelectionElement right? So how it should work?

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
npinaeva: I am also thinking if we should use a more universal notation, e.g. annotation key:value. So that for sriov we write
```
attachments:
  - name: l3-primary-pod-accelerated
    k8s.v1.cni.cncf.io/resourceName: nvidia.com/asap2_vf
  - name: l3-primary-vm-accelerated
    k8s.v1.cni.cncf.io/resourceName: nvidia.com/asap2_vfio
```
and for DRA we could do
```
attachments:
  - name: l3-primary-pod-accelerated
    k8s.ovn.org/deviceClass: asap_vf
  - name: l3-primary-vm-accelerated
    k8s.ovn.org/deviceClass: asap2_vfio
```

l8huang: For a generic way to add annotations to NADs, it's better to group key:value under a dedicated `annotations` field, e.g.:

```
attachments:
  - name: l3-primary-pod-accelerated
    annotations:
      k8s.v1.cni.cncf.io/resourceName: nvidia.com/asap2_vf
  - name: l3-primary-vm-accelerated
    annotations:
      k8s.v1.cni.cncf.io/resourceName: nvidia.com/asap2_vfio
```

The advantage of this approach is that we don't need to extend the API whenever a new annotation needs to be supported. But it could weakens CRD validation and may expose implementation details(annotation names) through the CRD API.

@trozet WDYT?

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: correct, NetworkSelectionElement doesn't have this field.

Currently, while handling CNIAdd, `getPrimaryUDNPodRequest()` only creates the primary UDN Pod request if the original request is for the default network:

```
	if originalPodRequest.nadName != types.DefaultNetworkName {
		return nil, nil
	}
```
and the pod namespace has an active primary UDN.

`getPrimaryUDNPodRequest()` need to be extended to read `primary-network` from `v1.multus-cni.io/default-network` -- I'm not sure how appropriate it is to add a `primary-network` field to [NetworkSelectionElement](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/master/pkg/apis/k8s.cni.cncf.io/v1/types.go), since that field is specific to ovn-k8s UDN.
 
Another option would be adding a pod annotation `k8s.ovn.org/primary-user-defined-network: l3-primary-pod-accelerated` to explicitly specify the primary NAD.

Which approach do you think is better? CC @trozet

l8huang:
> Considering this is pretty ugly, maybe we could put the additional-NAD-creation request into an annotation instead of a proper API field, and hope that we will be able to get rid of it one day? Considering this feature only adds extra things on top of UDN and doesn't change the normal UDN behaviour we could do something like
> 
> ```
> apiVersion: k8s.ovn.org/v1
> kind: ClusterUserDefinedNetwork
> metadata:
>   name: l3-primary
>   annotations:
>     k8s.ovn.org/nads-with-resources: "nvidia.com/asap2_vf, nvidia.com/asap2_vfio"
>   # or for DRA
>    k8s.ovn.org/nads-with-device-classes: "asap2_vf, asap2_vfio"
> spec:
>   namespaceSelector:
>     matchLabels:
>       kubernetes.io/metadata.name: udn-test
>   network:
>     topology: Layer3
>     layer3:
>       role: Primary
>       subnets:
>       - cidr: 10.20.0.0/16
>         hostSubnet: 24
>     transport: Geneve
> ```

I have a opposite view on this:  instead of using annotation, it's better to use fields to define the API behavior. A structured field gives us validation, readability, and a clearer API contract. It also leaves room for future extensions. For example, this proposal could naturally evolve to support multi-VTEP by adding fields to `attachments[]`:
```
  attachments:
  - name: l3-primary
     vtep: vtep0
  - name: l3-primary-pod-accelerated
    resourceName: nvidia.com/asap2_vf
    vtep: vtep1
```

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
npinaeva: I don't think we want to let users override the primary network, but we could consider adding an annotation to instruct ovn-k to use some resourceName for it. I am not sure though if it is ok to give control of the requested resources for primary network to the pod vs (C)UDN owner

trozet: I'm wondering can we get away with not having to specify anything from the pod side for primary network? 

  1. Get the namespace's assigned primary UDN.
  2. Get all generated NAD/profile variants for that UDN: veth, SR-IOV VF, VFIO, etc.
  3. For each profile NAD with resourceName, query pod resources for that resource.
  4. Count how many secondary attachments in the pod annotation consume that same resourceName.
  5. If len(deviceIDs) > secondaryCount, treat the last remaining ID as the primary candidate.
  6. Inspect/validate the device mode, then choose the matching generated NAD/profile.
  7. If no resource-backed profile has an extra device, use the veth profile.
  8. If multiple profiles match, fail or require explicit selection.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
trozet: I don't like specifying attachment names like this. I think it is error prone and it relies on our logic internally of how we generate names. There is also a risk of specifying a name that is too long. What if instead we do just generate the names ourselves, and post the names to status of the CUDN? Like:

```
 network:
      ...
    attachments:
    - name: default
    - name: pod-accelerated
      resourceName: nvidia.com/asap2_vf
    - name: vm-accelerated
      resourceName: nvidia.com/asap2_vfio

```

```
  status:
    attachments:
    - name: default
      networkAttachmentDefinitionName: l3-primary
    - name: pod-accelerated
      networkAttachmentDefinitionName: l3-primary-pod-accelerated
    - name: vm-accelerated
      networkAttachmentDefinitionName: l3-primary-vm-accelerated
```

then we can determine the name ourselves if it fits the proper length and there is no risk of a user screwing up the naming.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
```diff
@@ -0,0 +1,296 @@
...
+* CUDN always generates a generated base NAD named after the CUDN in each target namespace.
+* Each `attachments[].name` is the full generated NAD name.
+* If `attachments[].name` equals the UDN/CUDN name, no additional NAD is generated for that entry.
+  Its settings are applied to the generated base NAD.
+* If `attachments[].name` differs from the UDN/CUDN name, OVN-Kubernetes generates an additional
+  NAD with that exact name in the relevant namespace.
+
+#### Validation
+
+* `attachments[].name` must be DNS-1123 label compatible.
```
trozet, line 185: attachments[].name must also be immutable

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
```diff
@@ -0,0 +1,296 @@
...
+* `attachments[].name` must be unique.
+* `attachments[].name` must either equal the UDN/CUDN name, or be a full generated NAD name distinct
+  from the UDN/CUDN name.
+* `resourceName` should follow existing resource naming rules used by the Multus/SR-IOV ecosystem.
+
+### Implementation Details
+
+#### Controller changes
```
trozet, line 193: I think NAD/network manager will require some changes. I think we store a map of primaryNADs and it is single value. We will have to track multiple NADs for a primary network in our internal caches and helper functions.

trozet (changes requested):

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
```diff
@@ -0,0 +1,296 @@
+# OKEP-6181: Allow UDN/CUDN to Generate Multiple NADs
+
+* Issue: [#6181](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6181)
+
+## Problem Statement
+
+Today, a UDN/CUDN effectively maps to a single generated NAD. This makes it difficult to support
```
maiqueb: I don't think this is true. A UDN maps to a single generated NAD, but a C-UDN can map to one NAD *per* namespace. 

That's how we stretch a network across multiple namespaces (with multiple NADs).

l8huang: Updated, thanks. I clarified this as one generated NAD per namespace: UDN creates one in its namespace, and CUDN creates one in each selected namespace.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: > I don't think we want to let users override the primary network, but we could consider adding an annotation to instruct ovn-k to use some resourceName for it. 

The primary network in a namespace remains unique. The additional NADs are alternative attachment entries to that same primary logical network, carrying different annotations such as k8s.v1.cni.cncf.io/resourceName for resource selection.

Conceptually, this follows the existing NAD model: a workload selects the NAD it wants through k8s.v1.cni.cncf.io/networks or v1.multus-cni.io/default-network, and that NAD may carry a resourceName. For secondary networks this is straightforward because the pod selects the NAD directly. For primary networks, the primary UDN overlays the default-network path, which is why the proposal uses v1.multus-cni.io/default-network as the selection point. @npinaeva 

> 3. For each profile NAD with resourceName, query pod resources for that resource.

I think this has an ordering issue. For SR-IOV-backed primary attachments, the selected NAD/resourceName is needed before scheduling so the Pod can get the right resource request and device allocation. OVN-Kubernetes only sees allocated device IDs later during CNI handling, so it cannot use those IDs to choose the primary NAD first. That is why the proposal keeps an explicit pod-side selection signal. @trozet

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: Both approaches have trade-offs, but I prefer keeping `attachments[].name` as the generated NAD name. It makes the NAD name deterministic and directly authorable, without requiring users or automation to read UDN/CUDN status first.

The length and invalid-name concerns can be handled with CRD validation rules: reject invalid NAD names, names that exceed the allowed length, or names that do not start with the UDN/CUDN name if we want that constraint. If the controller generates names and only reports them in status, users need an extra discovery step before writing pod annotations.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: Do you mean attachments should be append-only once created? If only attachments[].name is immutable, users can still effectively rename by deleting one attachment and adding another, which this proposal treats as deleting the old generated NAD and creating a new one.

I think removal should be allowed. If the generated NAD is still used by Pods, we can preserve it until those Pods are gone, then delete it.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: Good point, added this to the implementation details.

l8huang (commented):
@npinaeva @trozet Thanks for taking the time to review this PR. I replied to your comments and addressed several of them. PTAL, thanks.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: BTW:  
```
    v1.multus-cni.io/default-network: |
      {
        "name":"default",
        "namespace":"ovn-kubernetes",
        "primary-network":"l3-primary-pod-accelerated"
      }
```
requires a default network stub `default` to be defined in ovn-kubernetes, otherwise multus will complain:
```
cannot find a network-attachment-definition (default) in namespace (ovn-kubernetes): network-attachment-definitions.k8s.cni.cncf.io "default" not found
```

So, I think below is a better option:
> Another option would be adding a pod annotation `k8s.ovn.org/primary-user-defined-network: l3-primary-pod-accelerated` to explicitly specify the primary NAD.

trozet: I'm not seeing the ordering issue. OVNK doesn't care about the scheduling. CNI is the only time it needs to know the device ID, and by that time the pod is scheduled, the devices are allocated, and Multus leaves the last device ID for the primary network with OVNK. What am I missing?

l8huang: I think the ordering issue is before OVN-Kubernetes/CNI.

The step that needs clarification is:
> 3. For each profile NAD with resourceName, query pod resources for that resource.

Those pod resources only exist if they are requested before scheduling. For SR-IOV, a mutating webhook looks at the pod’s NAD selections (`k8s.v1.cni.cncf.io/networks` / `v1.multus-cni.io/default-network`) and patches `resources.requests/limits`, for example:

```
    resources:
      limits:
        nvidia.com/bf2_vf: "1"
        nvidia.com/bf2_vfio: "1"
```

This conflicts with what you want to archive:
> ... get away with not having to specify anything from the pod side for primary network

Without a pod-side primary NAD selection, the webhook does not know which primary `resourceName` to request. Then the scheduler/device plugin will not allocate that device, so by the time Multus/CNI runs there is no “last device ID” for OVN-Kubernetes to infer from.

In our downsteam, the mutation webhook is handled by `sdn-mutator`.

trozet: How a downstream mutating webhook behaves is not really OVNK concern. This operates at a level higher than OVNK. The normal system (without any webhook) is:

1. User declares what resources they want for a pod.
2. Device plugin allocates, Multus reads, sends to OVNK
3. OVNK determines device type and wiring

Now as I understand it, the sdn-mutator acts like this:

1. Read NADs for secondary networks, primary network, then set resource requests on pod
2. Device plugin allocates, Multus reads, sends to OVNK
3. OVNK determines device type and wiring

The problem here is a circular dependency. The input (sdn mutator) is now relying on what is defined for the SDN underneath it to assume what resources a pod wants. This seems problematic from a design perspective in the first place, and worked up until this point due to a singular NAD per network. I don't like designing an upstream solution based on how a downstream consumer behaves. I know there is intent to upstream sdn-mutator, and I think how sdn-mutator determines what resourceNames to allocate to a pod should be determined a different way.

For example, there could be a priority ordering, where the sdn-mutator detects if the pod is a kubevirt VM, then assigns VFIO, or if VFIO is available, chooses that first. Then if VF pool is available, choose that. Finally, resort to veth if that is what is available. When we have an OKEP for creating an sdn-mutator upstream, we can discuss how it chooses what resources to request. Perhaps it will even be expressed via an API CRD.

npinaeva: I generally agree that we shouldn't design anything around external mutating webhooks.
I do think that having an explicitly defined resourceName for a given interface (via NAD annotation or something else) is better than trying to just find what is left/unused and use it.
e.g. if I have forgotten about the primary network device because it is not explicitly needed anywhere, it will get a veth, even if I wanted to use some other device type. Or if I have 2 unused devices of different types left, we will have to either pick a random or return an error even though it would've worked before with the NAD annotation.
If I want my primary network to be always accelerated, now I can just put the annotation on the UDN and if the right resource is not allocated, the pod creation will fail. With the implicit approach it will allocate veth, which is not what I wanted.
I think there should be a way to make the primary UDN accelerated by default, and then allow other device types with some extra tweak if needed.
Adding some annotation to the pod to tweak the primary network device type seems like a good enough alternative

trozet: "If I want my primary network to be always accelerated, now I can just put the annotation on the UDN and if the right resource is not allocated, the pod creation will fail. With the implicit approach it will allocate veth, which is not what I wanted."

No it will not as long as you did not define a veth type of attachment for the UDN.

l8huang: > For example, there could be a priority ordering, where the sdn-mutator detects if the pod is a kubevirt VM, then assigns VFIO, or if VFIO is available, chooses that first. Then if VF pool is available, choose that.

I think it's better to require explicit resource assignment. For example, a node may have two SR-IOV NICs, each with both VFIO and VF resource pools defined. In some cases, traffic is expected to use a specific NIC, and automatic selection could choose a different NIC than intended. Explicit assignment also avoids unpredictable performance issues, such as falling back to a veth interface instead of a VF when the VF resource pool is exhausted.

I didn't mean that we should design this around a mutating webhook handler. I was mainly trying to clarify what the source of truth should be, so that the Pod gets the required resource assigned—whether manually or automatically—and OVN-Kubernetes CNI can determine which resource corresponds to a given network.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
npinaeva: We currently have a NAD with pre-defined name based on the (C)UDN name, let's call it `udnNADName`. We currently copy annotations from the UDN to `udnNADName` NAD.
With the `attachments` section do we want to fully replace the `udnNADName` creation logic or augment it?
So we could  keep creating `udnNADName` with the UDN annotation as we do today, but only treat `attachments` as additional NADs. That makes this change additional to the existing logic and ensures that `udnNADName` is always present.
Another option would be to move `udnNADName` configuration to the `attachments` section too, but in that case we need more validation to ensure it is always present, and clearly state that UDN `resourceName` annotation is ignored when `attachments` field is present. This consolidates resourceName annotations for all created NADs in one place, which is nice, but it has to override the previous behaviour.
We could also just allow not creating `udnNADName` at all, but I think this will only bring more confusion and doesn't bring any useful functionality.

Status reporting with the created NAD names looks nice, but I am not sure if it is useful to anyone. As @l8huang said I don't think anyone will follow `create UDN -> read status to get NAD names -> patch pod spec with the NAD name` workflow, especially if NAD names are pre-defined anyway. So trying to have clear spec with potential validation sounds like a better option to me.

TLDR; I think it may be reasonable to request attachment like
```
attachments:
    - name: l3-primary
    - name: l3-primary-pod-accelerated
      resourceName: nvidia.com/asap2_vf
    - name: l3-primary-vm-accelerated
      resourceName: nvidia.com/asap2_vfio
```
with the validation that `l3-primary` (aka CUDN/UDN name) should be always present, and all other names should start with `l3-primary`

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
npinaeva: thinking long-term, I would like to get rid of creating NAD per resourceName with DRA, and instead move the resourceName somewhere to the pod itself, e.g. to NetworkSelectionElement. All these NADs are actually the same NAD and is only necessary because of sriov-dp/multus integration, which now affect our own API and the whole internal codebase for no real reason.
I am thinking if we should instead just add a `NetworkSelectionElement` fields for `resourceName` and update multus to handle it the same way as NAD annotation? then we won't need to create any extra NADs, but only figure out the primary network pod logic. Then with this nad setup
```
attachments:
    - name: l3-primary
    - name: l3-primary-pod-accelerated
      resourceName: nvidia.com/asap2_vf
    - name: l3-primary-vm-accelerated
      resourceName: nvidia.com/asap2_vfio
```
instead of saying 
`"k8s.v1.cni.cncf.io/networks": "l3-primary-pod-accelerated"` a pod will say

`"k8s.v1.cni.cncf.io/networks": [{"resourceName":"nvidia.com/asap2_vf","name":"l3-primary"}]` and no `attachments` field will be needed and we will still have only 1 nad `l3-primary`?
It looks like a 10-line change in the multus code, so maybe we can try to propose that?

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: > With the attachments section do we want to fully replace the udnNADName creation logic or augment it?

If `attachments` is omitted in UDN, `udnNADName` behaves exactly as it does today. If `attachments` is present, the `udnNADName` is still created, and the annotations that are copied from the UDN/CUDN to `udnNADName` today are copied to all generated NADs.Attachment-specific config is then applied on top. For example, if the UDN/CUDN has a `k8s.v1.cni.cncf.io/resourceName annotation` and an attachment also sets `resourceName`, the attachment value takes precedence for that generated NAD -- for long term, the configuration currently expressed through annotations should be moved into the CRD API if feasible.


> with the validation that l3-primary (aka CUDN/UDN name) should be always present

`l3-primary` can be optional in `attachments`. The UDN/CUDN always generates the `udnNADName` even when it is not listed. An entry with `name: l3-primary` is only needed when the user wants to customize the `udnNADName`. 

> ... and all other names should start with l3-primary

Updated the `Generated NAD naming` and `Validation sections` to reflect above.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: @trozet Would it be acceptable to use a Pod annotation like k8s.ovn.org/primary-user-defined-network: l3-primary-pod-accelerated to explicitly select the primary network attachment? I think deterministic selection is preferable to inference.

docs/okeps/okep-6181-allow-one-udn-generating-multiple-nads.md
```diff
@@ -0,0 +1,227 @@
...
+
+As an application operator, I want generated NAD names to be deterministic and discoverable from
+UDN/CUDN spec so that network annotation authoring is straightforward and stable.
+
+## Proposed Solution
+
+Add an `attachments` list to UDN/CUDN network definition. Each attachment entry defines a named
+attachment and optional SR-IOV resource mapping. OVN-Kubernetes generates one NAD per
+attachment, all
+bound to the same logical network topology/subnet settings from the parent UDN/CUDN.
```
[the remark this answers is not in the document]
l8huang: updated

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: I'm not sure I understand Multus well enough, but it seems Multus intentionally allows multiple NADs with the same configuration and different annotations to support use cases like resource selection.

```
"k8s.v1.cni.cncf.io/networks": [{"resourceName":"nvidia.com/asap2_vf","name":"l3-primary"}]
```
From a user's perspective, the `resourceName` is often an node setup detail that they may not know.

Using `k8s.v1.cni.cncf.io/networks` for resource selection feels like a strategic change, and I'm not sure Multus would agree with that direction.

Actually, this OKEP is primarily for laying the groundwork for VTEP selection(multi-VTEP) -- a `vtep` field(referencing a VTEP object by name) will be added in `attachment` to indicate which VTEP a NAD should use.

npinaeva: >Actually, this OKEP is primarily for laying the groundwork for VTEP selection(multi-VTEP) -- a vtep field(referencing a VTEP object by name) will be added in attachment to indicate which VTEP a NAD should use.

oh that changes things quite a bit, because I was trying to see if we can avoid generating multiple NADs for the same network. But also without the multi-vtep proposal it is hard to argue about that

>From a user's perspective, the resourceName is often an node setup detail that they may not know.

I think for the user it is more or less the same, because they will be told by the cluster admin to use nad `l3-primary-pod-accelerated` or `{"resourceName":"nvidia.com/asap2_vf","name":"l3-primary"}` and user doesn't understand the details for any of those definitions.

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
npinaeva: Primary network is implicit by itself: a pod doesn't do anything to get connected. Today primary network specifies the resourceName that should be used for this connection, then CNI will grab the last resource with that resourceName from the pod request (whether that resource is already used by another attachment of not).
Now we want to allow multiple resourceNames for a primary network and we have 2 options: explicitly specify which resourceName should be used for the primary network on the pod (with a pod annotation) or implicitly figure out the right resourceName based on the pod request.

Implicit way pros:
- avoids another pod annotation
- minus 1 yaml line on the pod spec
cons:
- will make CNI ADD longer (we need to fetch all NADs for a primary network to see which resourceNames can be used, then fetch all pod network attachments and their NAD to calculate which resource is still free); I would argue that we should add a check to the explicit way anyway to make sure the resource we grabbed is not used by anyone, but it will be a bit faster because we need to check within one resourceName
- hard for humans to figure out which resourceName is/should be used for the primary network. To debug a pod, a human will need to list all NADs that a pod uses and subtract the corresponding resourceNames from the pod request. Maybe we could work it around with some status/annotation reporting which resourceName was used for primary network

docs/okeps/okep-6181-allow-udn-generating-multiple-nads.md
l8huang: We need to reach a consensus to moving forward, so far 3 options have been discussed:

1. `v1.multus-cni.io/default-network` with a new field `primary-network`:
   ```    v1.multus-cni.io/default-network: |
      [
        {
          "name":"default",
          "namespace":"ovn-kubernetes",
          "primary-network":"l3-primary-pod-accelerated"
        }
      ]```

2. use Pod annotation: `k8s.ovn.org/primary-user-defined-network: l3-primary-pod-accelerated` 
3. No need to generate multiple NADs, use `k8s.v1.cni.cncf.io/networks` with primary network name and a new field `resourceName` to specify resource  
   ```
   k8s.v1.cni.cncf.io/networks": [{"resourceName":"nvidia.com/asap2_vf","name":"l3-primary"}]
   ``` 

Can we align on one of these options, or would it make sense to explore an alternative solution?

CC @trozet

npinaeva: I opened an issue in the multus https://github.com/k8snetworkplumbingwg/multus-cni/issues/1515 to discuss this option

l8huang: If multus accepts `k8s.v1.cni.cncf.io/networks": [{"resourceName":"nvidia.com/asap2_vf","name":"l3-primary"}]`, primary network can use the single NAD with different `resourceName` for creating Pod/VM. How about secondary network and multi-vtep? Do we still need to generate multiple NADs for vtep selection?

npinaeva: we only use "k8s.v1.cni.cncf.io/networks" for secondary networks, so secondary networks could be fixed like this. Then for primary network we still need to figure out the algorithm that will decide which resourceName to use. Fortunately this is fully in ovn-k with no extra deps, some thoughts are here https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6184#discussion_r3419944424
For the multi-vtep I need some more info on what exactly we want to do: is there some doc/okep?

l8huang: The pan is:
- Add multi-NAD support to UDN/CUDN through this OKEP, so a UDN/CUDN can define generated NADs with specific resources.
- Update the multi-VTEP OKEP-5535 to extend the multi-NAD feature with VTEP information on generated NADs. When a pod uses a NAD, that NAD’s VTEP information is used to determine the encap IP for the pod’s network interface.

The goal is to give each NAD a VTEP hint(e.g. an annotation), and to model that starting from the UDN/CUDN layer. This OKEP is a prerequisite for the next step.

CC @trozet

npinaeva: Here we are discussing multi-NAD support for different resourceNames, but multi-VTEP afaiu doesn't use different resourceNames, it wants multi-NAD for different vteps, which is anyway different from this OKEP. If we decide we need multi-nad for the vtep use case, we can add it.
I also don't see a use case where same network requires multiple NAD with different vteps there

l8huang: Actually, the multi-VTEP configuration uses different `resourceName` values. For example, assume:

- **NodeA** has one NIC and one VTEP.
- **NodeB** has two NICs and two VTEPs.
- The **Blue** network uses **vtep1 on NodeA** and **vtep2 on NodeB**.

In this case, two NADs are required for the same network because the selected resourceName and VTEP differs across the nodes.

| Node  | NIC  | `resourceName`          | VTEP  |
|--------|------|-------------------------|--------|
| NodeA  | NIC1 | `nvidia.com/cx_nic1_vf` | vtep1 |
| NodeB  | NIC1 | `nvidia.com/cx_nic1_vf` | vtep1 |
| NodeB  | NIC2 | `nvidia.com/cx_nic2_vf` | vtep2 |

Today, the `encap-ip` is inferred from the VF/VFIO device ID(see [getPfEncapIP() ](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/cni/helper_linux.go#L480)),  which happens during CNI network interface setup on the host. That is okay in Central mode; but in IC mode, an additional Pod patch request is required to update `k8s.ovn.org/pod-networks` annotation to propagate the encap-ip to remote nodes.

The goal of the new multi-VTEP design is to make the VTEP (`encap-ip`) available earlier by leveraging the VTEP CRD. This allows the network controller to determine the `encap-ip` when generating the initial `k8s.ovn.org/pod-networks` annotation for the pod, rather than discovering it later during CNI execution. This also makes the feature no longer limited to SR-IOV NICs, allowing the e2e tests to validate this feature with `veth` interfaces.

npinaeva: It looks like vtep is defined by the resource pool or even PF from which allocated VF is coming?
Maybe you could update the VTEP OKEP and describe how you would like it to be solved with multiple NADs, and we can discuss there is there is another way?

PR #6200, opened 2026-04-03
Drop windows support

booxter:
## 📑 Description

OVS 3.8.0 will drop support for Windows platform.

I am leaving hybrid-overlay implementation in place because it's also
used for Openshift SDN migration, though I suspect it should probably
also eventually go away.

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed Windows platform support. Windows build artifacts, service manager integration, and Windows-specific networking components are no longer available.
  * Hybrid-overlay is now the primary non-Linux node implementation.

* **Documentation**
  * Updated installation and build documentation to remove Windows-specific build instructions.
  * Clarified hybrid-overlay networking capabilities and removed Windows-related references.


go-controller/pkg/util/ovs.go
```diff
@@ -23,10 +22,6 @@ import (
 )
 
 const (
-	// On Windows we need an increased timeout on OVS commands, because
-	// adding internal ports on a non Hyper-V enabled host will call
-	// external Powershell commandlets.
-	// TODO: Decrease the timeout once port adding is improved on Windows
```
booxter, line 29: maybe worth leaving some TODO here still

trozet:
need RH input @tssurya on if this can actually be dropped or not as OCP supports windows hybrid overlay

tssurya:
hybrid overlay is not just used for migration, its actually a proper supported feature. we support running windows containers. let me get the story straight on our side and get back to you.
there is no issue filed for this? - is this having a motivation specifically? or just a cleanup PR?

booxter:
@tssurya there's a licensing issue with juju dependency that is pulled through go-powershell. The latter is used for windows only. We could address the problem by reimplementing a narrow subset of go-powershell without juju dependency if needed; but considering upstream OVS plans, cleanup seemed to be an easier approach. Let me know what the plan is for windows on your side. If it has to stay (though I wonder what the plan is considering OVS dropping support), then we can go the alternative route.

booxter:
To resolve before proceeding:

- Commit message should be adjusted not to suggest hybrid overlay is on a chopping block.
- Need to understand FDP / OpenShift plans for windows support.
- Restore dropped TODO comment.

PR #6203, opened 2026-04-04
helm: add missing patch/update verbs to ovnkube-db-ep Role for endpoints

TOMOFUMI-KONDO:
## 📑 Description

The `ovnkube-db-ep` Role only grants `get` and `create` verbs for endpoints,
but [`set_ovnkube_db_ep()`](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/7e14435c421cee65f9bd7dc51ad175d113d4049a/dist/images/ovnkube.sh#L871)
uses `kubectl apply` to update the ovnkube-db Endpoint. When the Endpoint
already exists (e.g., the pod was rescheduled to a different node),
`kubectl apply` issues a `PATCH` request which is denied, causing the
`sb-ovsdb` container to enter CrashLoopBackOff.

Add `patch` and `update` verbs to the endpoints rule in the `ovnkube-db-ep`
Role.

Fixes #6202

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Chores**
  * Expanded system permissions to support enhanced endpoint management operations, enabling improved operational capabilities and stability for database management components.

### DIFF
--- a/helm/ovn-kubernetes/templates/rbac-ovnkube-db.yaml
+++ b/helm/ovn-kubernetes/templates/rbac-ovnkube-db.yaml
@@ -58,7 +58,7 @@ rules:
     - apiGroups: [""]
       resources:
           - endpoints
-      verbs: [ "get", "create" ]
+      verbs: [ "get", "create", "patch", "update" ]
     - apiGroups: [""]
       resources:
           - pods


TOMOFUMI-KONDO:
CI failed (https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/23978751218/job/69940090277?pr=6203) ,but this does not seem to be related to this change. Is it a flaky test?

girishmg:
@TOMOFUMI-KONDO  if the OVN DB Pods is NOT backed by some central storage using PVC or some such, then we don't support the movement of OVN DB Pods to a different node.

You will have to pin the DB node to a K8s node. If you need HA, then you will need to use statefulset (which again must be pinned to 3 sets of nodes (if 3 is RAFT quorum)).

Furthermore, with the project moving to OVN-K8s Interconnect mode, then the OVN-DB Pods are doing to be deprecated soon.

TOMOFUMI-KONDO:
@girishmg 
Thank you for responding! I understood that OVN DB Pods cannot move between nodes. If so, how can I pin an OVN DB Pod to a node when using helm chart of central mode without HA? (like `helm/ovn-kubernetes/values-no-ic.yaml`)

TOMOFUMI-KONDO:
> @girishmg Thank you for responding! I understood that OVN DB Pods cannot move between nodes. If so, how can I pin an OVN DB Pod to a node when using helm chart of central mode without HA? (like `helm/ovn-kubernetes/values-no-ic.yaml`)

@girishmg 
How about this?

PR #6209, opened 2026-04-07
Batch UDN Node Annotation Updates to Reduce Contention and Kubeapi Throttling

marty-power:
## 📑 Description

Adds NodeAnnotationBatcher, which enqueues and batches Node Annotation updates to reduce KubeAPI spam and update contention when creating large numbers of UDNs at once. Uses the Controller framework to process per-node updates using multiple workers. 

Fixes #

## ✅ Checks

- [ ] My code requires tests

### DIFF
--- a/go-controller/pkg/clustermanager/node/node_allocator.go
+++ b/go-controller/pkg/clustermanager/node/node_allocator.go
@@ -10,7 +10,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	listers "k8s.io/client-go/listers/core/v1"
-	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 
@@ -47,9 +46,12 @@ type NodeAllocator struct {
 
 	// nodeSubnets is a list of node subnets that are managed by the cluster subnet allocator
 	nodeSubnets []*net.IPNet
+
+	// batcher batches annotation updates across all networks for better API server efficiency
+	batcher *NodeAnnotationBatcher
 }
 
-func NewNodeAllocator(networkID int, netInfo util.NetInfo, nodeLister listers.NodeLister, kube kube.Interface, tunnelIDAllocator id.Allocator) *NodeAllocator {
+func NewNodeAllocator(networkID int, netInfo util.NetInfo, nodeLister listers.NodeLister, kube kube.Interface, tunnelIDAllocator id.Allocator, batcher *NodeAnnotationBatcher) *NodeAllocator {
 	na := &NodeAllocator{
 		kube:                         kube,
 		nodeLister:                   nodeLister,
@@ -58,6 +60,7 @@ func NewNodeAllocator(networkID int, netInfo util.NetInfo, nodeLister listers.No
 		clusterSubnetAllocator:       NewSubnetAllocator(),
 		hybridOverlaySubnetAllocator: NewSubnetAllocator(),
 		idAllocator:                  tunnelIDAllocator,
+		batcher:                      batcher,
 	}
 
 	if na.hasNodeSubnetAllocation() {
@@ -523,49 +526,16 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
 	return nil
 }
 
-// updateNodeNetworkAnnotationsWithRetry will update the node's subnet annotation and network id annotation
 func (na *NodeAllocator) updateNodeNetworkAnnotationsWithRetry(nodeName string, hostSubnetsMap map[string][]*net.IPNet, networkId, tunnelID int) error {
-	// Retry if it fails because of potential conflict which is transient. Return error in the
-	// case of other errors (say temporary API server down), and it will be taken care of by the
-	// retry mechanism.
-	resultErr := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		// Informer cache should not be mutated, so get a copy of the object
-		node, err := na.nodeLister.Get(nodeName)
-		if err != nil {
-			return err
-		}
+	if na.batcher == nil {
+		return fmt.Errorf("node annotation batcher is required but was not configured")
+	}
 
-		cnode := node.DeepCopy()
-		for netName, hostSubnets := range hostSubnetsMap {
-			cnode.Annotations, err = util.UpdateNodeHostSubnetAnnotation(cnode.Annotations, hostSubnets, netName)
-			if err != nil {
-				return fmt.Errorf("failed to update node %q annotation subnet %s: %w",
-					node.Name, util.JoinIPNets(hostSubnets, ","), err)
-			}
-		}
-
-		networkName := na.netInfo.GetNetworkName()
-
-		if networkId != types.NoNetworkID {
-			cnode.Annotations, err = util.UpdateNetworkIDAnnotation(cnode.Annotations, networkName, networkId)
-			if err != nil {
-				return fmt.Errorf("failed to update node %q network id annotation %d for network %s: %w",
-					node.Name, networkId, networkName, err)
-			}
-		}
-		if tunnelID != types.NoTunnelID {
-			cnode.Annotations, err = util.UpdateUDNLayer2NodeGRLRPTunnelIDs(cnode.Annotations, networkName, tunnelID)
-			if err != nil {
-				return fmt.Errorf("failed to update node %q tunnel id annotation %d for network %s: %w",
-					node.Name, tunnelID, networkName, err)
-			}
-		}
-		// It is possible to update the node annotations using status subresource
-		// because changes to metadata via status subresource are not restricted for nodes.
-		return na.kube.UpdateNodeStatus(cnode)
-	})
-	if resultErr != nil {
-		return fmt.Errorf("failed to update node %s annotation: %w", nodeName, resultErr)
+	networkName := na.netInfo.GetNetworkName()
+	hostSubnets, hasSubnetUpdate := hostSubnetsMap[networkName]
+	if hasSubnetUpdate || networkId != types.NoNetworkID || tunnelID != types.NoTunnelID {
+		na.batcher.EnqueueUpdate(nodeName, networkName, hostSubnets, networkId, tunnelID)
+		klog.V(5).Infof("Enqueued annotation update for node %s network %s to batcher", nodeName, networkName)
 	}
 	return nil
 }
@@ -588,7 +558,7 @@ func (na *NodeAllocator) Cleanup() error {
 			continue
 		}
 
-		hostSubnetsMap := map[string][]*net.IPNet{networkName: nil}
+		hostSubnetsMap := map[string][]*net.IPNet{networkName: {}}
 		// passing util.InvalidID deletes the network/tunnel id annotation for the network.
 		err = na.updateNodeNetworkAnnotationsWithRetry(node.Name, hostSubnetsMap, types.InvalidID, types.InvalidID)
 		if err != nil {
--- a/go-controller/pkg/clustermanager/node/node_annotation_batcher.go
+++ b/go-controller/pkg/clustermanager/node/node_annotation_batcher.go
@@ -1,0 +1,228 @@
+// SPDX-FileCopyrightText: Copyright The OVN-Kubernetes Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package node
+
+import (
+	"fmt"
+	"net"
+	"strings"
+	"sync"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controller"
+	nodecontroller "github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/controllers/node"
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/kube"
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/types"
+	"github.com/ovn-kubernetes/ovn-kubernetes/go-controller/pkg/util"
+)
+
+// NodeAnnotationBatcher batches node annotation updates from multiple networks
+// and applies them in a single API call. When multiple networks enqueue updates
+// for the same node before the worker processes it, the workqueue coalesces the
+// key and all pending updates are applied together, reducing API server load.
+type NodeAnnotationBatcher struct {
+	kube kube.Interface
+
+	pendingUpdates map[string]*pendingNodeUpdate
+	mutex          sync.Mutex
+
+	reconciler controller.Reconciler
+
+	cache *nodecontroller.NodeAnnotationCache
+}
+
+// pendingNodeUpdate accumulates annotation updates for a single node
+// from potentially multiple networks before batching them together
+type pendingNodeUpdate struct {
+	hostSubnets map[string][]*net.IPNet
+	networkIDs  map[string]int
+	tunnelIDs   map[string]int
+}
+
+func NewNodeAnnotationBatcher(kube kube.Interface) *NodeAnnotationBatcher {
+	b := &NodeAnnotationBatcher{
+		kube:           kube,
+		pendingUpdates: make(map[string]*pendingNodeUpdate),
+		cache:          nodecontroller.NewNodeAnnotationCache(),
+	}
+
+	b.reconciler = controller.NewReconciler("node-annotation-batcher", &controller.ReconcilerConfig{
+		RateLimiter: workqueue.DefaultTypedControllerRateLimiter[string](),
+		Reconcile:   b.reconcile,
+		Threadiness: 15,
+		MaxAttempts: controller.InfiniteAttempts,
+	})
+
+	return b
+}
+
+func (b *NodeAnnotationBatcher) Reconciler() controller.Reconciler {
+	return b.reconciler
+}
+
+// EnqueueUpdate queues an annotation update for a node from a specific network.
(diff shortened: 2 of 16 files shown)


marty-power:
@coderabbitai review

marty-power:
@coderabbitai review

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,384 @@
...
+							b.mutex.Lock()
+							queueLength = len(b.pendingUpdates)
+							b.mutex.Unlock()
+						}
+						klog.Info("Node annotation batcher stopped")
+						return
+					}
+				}
+
+				b.processBatch()
```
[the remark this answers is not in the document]
marty-power: If annotation updates are arriving in quick succession for an extended period, the delay will quickly become maxed out. The first iteration not updating the timers will not significantly affect the performance.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -504,13 +508,23 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
...
-	// Retry if it fails because of potential conflict which is transient. Return error in the
-	// case of other errors (say temporary API server down), and it will be taken care of by the
-	// retry mechanism.
+	networkName := na.netInfo.GetNetworkName()
+
+	if na.batcher != nil {
+		// Distinguish between "no update needed" (key not in map) and "delete subnets" (key present with nil/empty value)
+		hostSubnets, hasSubnetUpdate := hostSubnetsMap[networkName]
+		if hasSubnetUpdate || networkId != types.NoNetworkID || tunnelID != types.NoTunnelID {
+			na.batcher.EnqueueUpdate(nodeName, networkName, hostSubnets, networkId, tunnelID)
```
trozet: even when hostSubnets is not updated, it is passed here. So if hostSubnets is not updated, but networkID or tunnelId is, an update will be enqueued and the pending.hostSubnets will have an entry with the networkName and value nil. That will then satisfy the check in updateNodeAnnotations:

```
		// Update host subnets
		for networkName, subnets := range pending.hostSubnets {
			existingSubnets[networkName] = subnets
			subnetsUpdated = true
		}
		if subnetsUpdated {
			cnode.Annotations, err = util.UpdateNodeHostSubnetAnnotationMap(cnode.Annotations, existingSubnets)
			if err != nil {
				return fmt.Errorf("failed to update node-subnets annotation for node %s: %w", nodeName, err)
			}
		}
```

and I think it will be treated as a delete. So effectively it will remove  the host subnet. Should probably pass hasSubnetUpdate or something to EnqueueUpdate to differentiate.

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -541,8 +555,7 @@ func (na *NodeAllocator) updateNodeNetworkAnnotationsWithRetry(nodeName string,
 					node.Name, tunnelID, networkName, err)
 			}
 		}
-		// It is possible to update the node annotations using status subresource
-		// because changes to metadata via status subresource are not restricted for nodes.
```
trozet, line 564: how come you are removing this?

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,369 @@
...
+				b.stopping = true
+				queueLength := len(b.pendingUpdates)
+				b.mutex.Unlock()
+				// Drain all pending updates (no retries due to stopping flag)
+				for queueLength > 0 {
+					b.processBatch()
+					b.mutex.Lock()
+					queueLength = len(b.pendingUpdates)
+					b.mutex.Unlock()
+				}
```
trozet: looks like the same code as lines 101-112. Could maybe move it into a small inline function above.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,369 @@
...
+			select {
+			case <-b.triggerChan:
+				// Calculate delay based on whether we're in quick succession
+				b.mutex.Lock()
+				delay := b.currentDelay
+				b.mutex.Unlock()
+
+				// Wait for the calculated delay
+				if delay > 0 {
+					timer := time.NewTimer(delay)
```
trozet: There is a small optimization you could make here specifically around ramping up delay. Enqueue updates the b.currentDelay, but that only takes effect on the next time triggerChan runs. So if on initial ramp you had:

  t=0ms    enqueue A -> worker starts timer with currentDelay=1ms
  t=0.2ms  enqueue B -> currentDelay increases, but timer still fires at 1ms
  t=0.5ms  enqueue C -> currentDelay increases again, same timer
  t=1ms    processBatch() flushes A/B/C

so the enqueuing is not delaying the current cycle, which means it will end up lagging on delaying updates. Under really heavy load it doesn't matter,  cause it will eventually always be throttled at 100ms. I asked AI how to improve it:

```
For API-server performance, a resettable debounce timer with a max cap would be better.

  Current behavior:

  first enqueue -> start 1ms timer
  more enqueues -> currentDelay increases, but active timer stays 1ms
  1ms -> flush

  Better batching behavior:

  first enqueue -> start timer
  more enqueues -> push timer out a little
  keep pushing until either traffic quiets down or maxBatchDelay is reached
  flush

  Why it performs better:

  - More network updates for the same node collapse into one UpdateStatus.
  - Fewer node Get + UpdateStatus calls hit the API server.
  - Fewer concurrent writes to the same node reduce resource-version conflicts.
  - The batcher has a better chance of merging L3 subnet, network ID, and tunnel ID changes from multiple UDN controllers into one node update.

  The important tradeoff is latency. A lone update should still flush quickly, but a burst should wait long enough to collect nearby work. So I?d use a capped quiet-period model:

  flushAt = min(firstEnqueueTime + maxBatchDelay, now + currentDelay)

  That preserves an upper bound, e.g. 100ms, while letting active bursts actually benefit from the increased delay. Current code gives good low latency, but it leaves batching efficiency on the table because the first timer wins.

```

I'm not sure if it is even worth changing, or if the current behavior was intended and only batching happens under very heavy load?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -504,13 +508,23 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
...
+// updateNodeNetworkAnnotationsWithRetry will update the node's subnet annotation and network id annotation.
+// If a batcher is configured, it will enqueue the updates to be batched with other network updates.
+// Otherwise, it will update the node directly.
 func (na *NodeAllocator) updateNodeNetworkAnnotationsWithRetry(nodeName string, hostSubnetsMap map[string][]*net.IPNet, networkId, tunnelID int) error {
-	// Retry if it fails because of potential conflict which is transient. Return error in the
-	// case of other errors (say temporary API server down), and it will be taken care of by the
-	// retry mechanism.
+	networkName := na.netInfo.GetNetworkName()
+
+	if na.batcher != nil {
```
trozet: why preserve a branch where batcher == nil? dont we always want to use the batcher if it is an improvement?

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -504,13 +508,23 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
...
+		// Distinguish between "no update needed" (key not in map) and "delete subnets" (key present with nil/empty value)
+		hostSubnets, hasSubnetUpdate := hostSubnetsMap[networkName]
+		if hasSubnetUpdate || networkId != types.NoNetworkID || tunnelID != types.NoTunnelID {
+			na.batcher.EnqueueUpdate(nodeName, networkName, hostSubnets, networkId, tunnelID)
+			klog.V(5).Infof("Enqueued annotation update for node %s network %s to batcher", nodeName, networkName)
+		}
+		return nil
+	}
+
```
trozet, line 533: with the old logic below, it would try to retry on conflict (I think 4 times) and then give up. Also if there was a failure it would release allocations. Now it looks like we Enqueue, then it retries indefinitely and never releases on failure.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,369 @@
...
+	}
+	if tunnelID != types.NoTunnelID {
+		pending.tunnelIDs[networkName] = tunnelID
+	}
+
+	// Update backoff based on request rate
+	now := time.Now()
+	timeSinceLastEnqueue := now.Sub(b.lastEnqueueTime)
+
+	if timeSinceLastEnqueue < backoffThreshold && !b.lastEnqueueTime.IsZero() {
```
trozet: I think if you get an update for an already existing pending update for  a node+network, then you can ignore bumping the current delay as this is just a refresh of an existing update?

go-controller/pkg/clustermanager/node/node_allocator.go
trozet: a side effect of this is the UDN status will always be updated to a positive status, where if the node couldn't be annotated before then the UDN status would mark the node as failed.

go-controller/pkg/controllers/node/node_annotation_cache.go
```diff
@@ -95,12 +95,12 @@ func (c *NodeAnnotationCache) ParseUDNLayer2NodeGRLRPTunnelIDCached(node *corev1
...
 }
 
-// parseNetworkMapCached returns the parsed per-network string map for the
+// ParseNetworkMapCached returns the parsed per-network string map for the
 // given annotation, reusing a cached parse result when the raw annotation
 // value has not changed. If there is a cache miss, the annotation is parsed;
 // updateCache controls whether the parsed result replaces the cached value for
 // this node and annotation.
-func (c *NodeAnnotationCache) parseNetworkMapCached(node *corev1.Node, annotationName string, updateCache bool) (map[string]string, error) {
+func (c *NodeAnnotationCache) ParseNetworkMapCached(node *corev1.Node, annotationName string, updateCache bool) (map[string]string, error) {
```
trozet: instead of exporting these generic map parsing functions, would it be better to create specific functions like:

  ParseUDNLayer2NodeGRLRPTunnelIDsMapCached(node) (map[string]int, error)
  ParseNetworkIDsMapCached(node) (map[string]int, error)
  ParseNodeHostSubnetsAnnotationCached(node) (map[string][]*net.IPNet, error)

?

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -230,13 +233,11 @@ func (b *NodeAnnotationBatcher) processBatch() {
 // updateNodeAnnotations applies all pending updates for a node using retry logic
 func (b *NodeAnnotationBatcher) updateNodeAnnotations(nodeName string, pending *pendingNodeUpdate) error {
 	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		node, err := b.nodeLister.Get(nodeName)
+		cnode, err := b.client.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
```
trozet: why did you switch this? even with informer it should still conflict? I think with unit tests the fake client will not respect resource version...I think we normally prepend reactor for this. There are probably some examples around the code base.

trozet (changes requested):
With node controller we have 15 goroutines processing node updates. With the batching change, although they all process and then Enqueue in parallel, the batcher effectively serializes all node updates into a single goroutine. The downside here is if you have updates for nodes A,B,C,D they all have to happen in order. So if updates for A are failing due to conflict or whatever, we are effectively blocking B,C,D from getting updates and continuing (also using kube client compounds the delay). This propagates delay to all ovnkube-controllers on B,C,D.

I think it would be better to have some parallelism, and maybe split updates by node. So the batcher can still increase delay of updating due to too many enqueues, but it enqueues to multiple queues (via node name), and there are multiple goroutine workers that are draining queues. That would allow multiple updates in parallel for different nodes at the same time. What level of parallelism would be best? 15 workers? I'm not sure, I think it would need a good amount of testing under heavy load to see the cost/benefits.

go-controller/pkg/clustermanager/node/node_allocator.go
marty-power: syncNodeNetworkAnnotations passes an empty hostSubnetsMap unless it detects a change, in which case it passes a map with a single entry for the network it is updating. That's the only codepath outside of the Cleanup() and CleanupNode() functions which calls updateNodeNetworkAnnotationsWithRetry()

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
marty-power: I was under the impression that using the node lister instead of direct API calls could cause conflicts/overwrites, but I see now that ResourceVersion will prevent that. I'll revert this.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,228 @@
...
+		b.pendingUpdates[nodeName] = pending
+	}
+
+	pending.hostSubnets[networkName] = hostSubnets
+	if networkID != types.NoNetworkID {
+		pending.networkIDs[networkName] = networkID
+	}
+	if tunnelID != types.NoTunnelID {
+		pending.tunnelIDs[networkName] = tunnelID
+	}
```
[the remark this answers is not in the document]
marty-power: Deletion is indicated by types.InvalidID in the Cleanup() function

go-controller/pkg/clustermanager/node/node_allocator.go
```diff
@@ -523,49 +526,16 @@ func (na *NodeAllocator) Sync(nodes []interface{}) error {
...
-		}
-		// It is possible to update the node annotations using status subresource
-		// because changes to metadata via status subresource are not restricted for nodes.
-		return na.kube.UpdateNodeStatus(cnode)
-	})
-	if resultErr != nil {
-		return fmt.Errorf("failed to update node %s annotation: %w", nodeName, resultErr)
+	networkName := na.netInfo.GetNetworkName()
+	hostSubnets, hasSubnetUpdate := hostSubnetsMap[networkName]
+	if hasSubnetUpdate || networkId != types.NoNetworkID || tunnelID != types.NoTunnelID {
```
shreyasbe, line 536: When only Network/Tunnel IDs change, nil subnets are passed to the batcher. The batcher stores this nil and later treats it as a "delete" command, wiping existing subnet annotations.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,228 @@
...
+			return fmt.Errorf("failed to parse existing tunnel-ids for node %s: %w", nodeName, err)
+		} else if existingTunnelIDs == nil {
+			existingTunnelIDs = make(map[string]int)
+		}
+
+		subnetsUpdated := false
+		networkIDsUpdated := false
+		tunnelIDsUpdated := false
+
+		for networkName, subnets := range pending.hostSubnets {
```
shreyasbe, line 180: This will blindly overwriting existingSubnets with nil from ID-only updates, wipes valid subnet annotations.

go-controller/pkg/clustermanager/node/node_allocator.go
marty-power: The only codepaths which call updateNodeNetworkAnnotationsWithRetry are syncNodeNetworkAnnotations(), CleanupNode(), and Cleanup(). syncNodeNetworkAnnotations() passes a map which only contains networkName keys that have changes associated with them. So, a key for a networkName will only be added to a hostSubnetsMap with a nil value if it is meant to be deleted. 

This logic also applies to the below comment. If a networkName has a nil value associated with it, it is meant to be deleted.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,228 @@
...
+
+	klog.V(5).Infof("Enqueued annotation update for node %s network %s (subnets=%v, netID=%v, tunnelID=%v)",
+		nodeName, networkName, len(hostSubnets), networkID, tunnelID)
+
+	b.reconciler.Reconcile(nodeName)
+}
+
+func (b *NodeAnnotationBatcher) reconcile(nodeName string) error {
+	b.mutex.Lock()
+	pending, ok := b.pendingUpdates[nodeName]
```
shreyasbe, line 100: Can we add b.stopping flag to drop failures during shutdown ?

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
```diff
@@ -0,0 +1,228 @@
...
+	b := &NodeAnnotationBatcher{
+		kube:           kube,
+		pendingUpdates: make(map[string]*pendingNodeUpdate),
+		cache:          nodecontroller.NewNodeAnnotationCache(),
+	}
+
+	b.reconciler = controller.NewReconciler("node-annotation-batcher", &controller.ReconcilerConfig{
+		RateLimiter: workqueue.DefaultTypedControllerRateLimiter[string](),
+		Reconcile:   b.reconcile,
+		Threadiness: 15,
```
shreyasbe, line 57: 15 Threadiness is reasonable ?  seems to be pretty high.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
marty-power: This behavior is mostly handled by the Controller framework being utilized. The only operation which could benefit is an in-progress PatchNodeStatusAnnotations, so one API call would be saved.

I don't think that's worth it for having to track an additional flag.

go-controller/pkg/clustermanager/node/node_annotation_batcher.go
marty-power: Dropping it down to 5.

PR #6215, opened 2026-04-07
OKEP-5831: custom routing with AdminPolicyBasedRoute API

### LINKED ISSUE #5831: Add API to support in-cluster policy based routing

Add API to support in-cluster policy based routing backed by Logical_Router_Policy table in OVN.

For example:

- From: Pod IPs specified by pod selector, node selector and namespace selector
- To: some particular destination, e.g. IP block
- Nexthop: some pod IPs

### Why is this needed?

Policy based routing is needed for various use cases, for example:

- Multi-tenancy Internet gateways
   For tenant coke, use gateway A to access internet, while for the tenant pepsi, use gateway b.
- VPN for various Cloud Service Providers
   Use VPN A to connect to AWS, VPN B to Azure, etc.
- Health probing for Istio
  Istio uses specific link-local IPv4 169.254.7.127 for health checking, the packets enter OVN via management port ovn-k8s-mp0.
  When pods respond to such requests, the packets replying to 169.254.7.127 should be redirected to management port as well.

Existing AdminPolicyBasedExternalRoute doesn't support `To` field.

jxiaobin:
Add the design proposal for issue #5831: a cluster-scoped AdminPolicyBasedRoute CRD and controller that reconcile policy-based routing into OVN Logical_Router_Policy.

Fixes #5831

## Summary by CodeRabbit

* **Documentation**
  * Added design specification for AdminPolicyBasedRoute, introducing a new cluster-scoped CRD for in-cluster policy-based routing. The design includes specifications for routing rules with network selectors, destination CIDR routes, and next-hop configuration, along with reconciliation controllers and event-driven address set management.

### DIFF
--- a/docs/okeps/okep-5831-incluster-admin-pbr.md
+++ b/docs/okeps/okep-5831-incluster-admin-pbr.md
@@ -1,0 +1,465 @@
+# OKEP-5831: Add API to support in-cluster policy based routing
+
+* Issue: [#5831](https://github.com/ovn-org/ovn-kubernetes/issues/5831)
+
+## Problem Statement
+
+The workloads running in ovn-kubernetes backed k8s cluster may need policy based routing to
+access external services. For example, in a multi-tenant cluster, tenant Coke needs to go 
+through gateway A to access internet, while tenant Pepsi needs to go through gateway B to access
+internet. Both gateway A and B expose internal IP addresses to the clients, and clients can specify
+the internal IPs as nexthop.
+
+Custom routing is also needed when Istio CNI is used in Ambient mode, as ingress Kubelet healthcheck
+probes via the mp0 interface will be SNAT'ed to 169.254.7.127, and the ingress-reply packet from the
+pod (which would normally go via shared gateway) needs to route back towards mp0.
+
+## Goals
+
+* Reconcile in-cluster PBR rules based on CRD spec
+* Reconcile address sets in response to Pod/Node/Namespace events
+* Automatically use remote port's IP as nexthop if the specified nexthop is in the remote zone
+* Properly configured CRD can make health check working for Istio/Ambient enabled pods, for interconnect mode only
+
+## Non-Goals
+
+
+## Introduction
+
+The need for policy-based routing (PBR) within Kubernetes clusters has grown as network architectures become more complex, particularly in multi-tenant and hybrid network environments. Traditional routing assigns the same route to all workloads without regard to context such as namespace, pod, or tenant affinity. However, modern use cases often require that certain workloads use specific gateways or follow unique routing paths, for example to partition internet access by tenant, or to route workloads through VPNs/proxies.
+
+This proposal introduces an API for cluster administrators to configure and manage in-cluster policy-based routing backed by Logical_Router_Policy in OVN. The intent is to allow for declarative, CRD-based configuration of routing policies that respond dynamically to changes in the cluster, such as pods being added, removed, or relocated.
+
+## User-Stories/Use-Cases
+
+Story 1: Internet access partitioning
+
+As a cluster administrator, I want to specify gateway A for tenant Coke, gateway B for the tenant Pepsi, so that internet traffic can be
+distributed among multiple gateways.
+  
+Story 2: VPN
+
+As a cluster administrator, I want to specify policy based routing so that my workload can talk to external services through VPN.
+
+Story 3: Readiness check with Istio CNI
+
+As a cluster administrator, I want to enable Istio CNI while keep health check working by rerouting packets to node's management port, this applies to IC mode ovn-kubernetes only.
+
+## Proposed Solution
+
+We will introduce a new cluster-scoped CRD: `AdminPolicyBasedRoute`, to allow cluster administrators to declaratively define in-cluster PBR (Policy-Based Routing) rules, and add a new AdminPolicyBasedRouteController to reconcile these custom resources into OVN's [Logical_Router_Policy](https://man7.org/linux/man-pages/man5/ovn-nb.5.html#Logical_Router_Policy_TABLE).
+
+For interconnect enabled clusters, if the nexthops specified in the CRD are in remote zones, the actual nexthop in Logical_Router_Policy record will be
+replaced with the transit IP to the remote zone.
+
+To address the routing issue for health check traffic under Istio/Ambient mode, we will introduce a new field `NextHopDevices` to allow users to specify a port name as nexthop, the controller needs to figure out the actual IP and apply it in Logical_Router_Policy record.
+
+#### High-Level Workflow Diagram
+
+```
++------------------+                          +--------------------+
+| AdminPolicyBased |    watches CRUD       -> | APBR Controller    |
+|   Route CRs      |------------------------->| (in ovnkube-master)|
++------------------+                          +--------------------+
+                                                ^       |
+                                                |       | applies
+                       watches events           |       |
++------------------+                            |       v
+|   Pod Events     |----------------------------+    +-----------------------------+
++------------------+                            |    | OVN NBDB (Logical_Router_   |
+                                                |    |          Policy)            |
++------------------+                            |    +-----------------------------+
+| Namespace Events |----------------------------+    |
++------------------+                            |    | updates
+                                                |    v
++------------------+                            |    +-----------------------------+
+|   Node Events    |----------------------------+    | OVN Address Sets            |
++------------------+                                 +-----------------------------+
+```
+
+### API Details
+
+* A new API `AdminPolicyBasedRoute` under the `k8s.ovn.org/v1beta1` group will be added to  
+  `go-controller/pkg/crd/adminpbr/v1beta1`. This would be a cluster-scoped CRD:
+
+```go
+import (
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	crdtypes "github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/types"
+)
+
+type Status string
+
+const (
+	StatusSucceeded Status = "Succeeded"
+	StatusFailed    Status = "Failed"
+)
+
+// AdminPolicyBasedRouteStatus describes the current status of the AdminPolicyBasedRoute.
+type AdminPolicyBasedRouteStatus struct {
+	// An array of Human-readable messages indicating details about the status of the object.
+	// +optional
+	Messages []string `json:"messages,omitempty"`
+
+	// A concise indication of whether the AdminPolicyBasedRoute resource is applied or not
+	// +optional
+	Status Status `json:"status,omitempty"`
+}
+
+// AdminPolicyBasedRoute cluster-scoped API provides a way to influence routing decisions
+// in the SDN network. The API can be used to match the packets originating from and/or
+// -- kubernetes nodes, namespaces, pods -- and forward them to different destination for
+// further processing.
+//
+// This API leverages OVN's Logical Router Policy feature. The AdminPolicyBasedRoute policies
+// are translated to OVN policies and these policies override OVN's static routing decision.
+// The priority of the policy rules is set to 80 (priority in range of 0 to 32,767, with
+// numerically higher priority taking precedence over those with lower), and it processed
+// after all the OVN K8s' pre-defined rules.
+//
+// +genclient
+// +genclient:nonNamespaced
+// +resource:path=adminpolicybasedroute
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=adminpbr,scope=Cluster
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=".status.status"
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=".metadata.creationTimestamp"
+type AdminPolicyBasedRoute struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   AdminPolicyBasedRouteSpec   `json:"spec,omitempty"`
+	Status AdminPolicyBasedRouteStatus `json:"status,omitempty"`
+}
+
+// AdminPolicyBasedRouteSpec defines the desired state of cluster scoped routing policies
+type AdminPolicyBasedRouteSpec struct {
+	// networkSelectors selects the networks on which the pod need to be applied with the routing policies.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="networkSelectors are immutable"
+	NetworkSelectors crdtypes.NetworkSelectors `json:"networkSelectors,omitempty"`
+
+	// a collection of policy objects to influence routing
+	Policies []RoutingPolicyRule `json:"policies"`
+}
+
+// RoutingPolicyRule is a single routing policy rule object
+type RoutingPolicyRule struct {
+	// From matches the packets for the routing policy
+	From RoutingPolicyMatch `json:"from"`
+	// To specifies destination addresses used to match the same in the packets
+	To networkingv1.IPBlock `json:"to"`
+	// NextHop defines where the matched packets should be forwarded
+	NextHop RoutingPolicyNextHop `json:"nexthop"`
+}
+
+// RoutingPolicyMatch provides a way to select nodes, namespaces, and pods such that
+// the packets originating from them will be subjected to routing policies.
+// +kubebuilder:validation:MinProperties=1
+type RoutingPolicyMatch struct {
+	// NodeSelector matches the source packets only from the node(s) whose label
+	// matches this definition. This field is optional.
+	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty"`
+	// NamespaceSelector matches the source packets only from the namespace(s) whose label
+	// matches this definition. This field is optional.
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+	// PodSelector matches the source packets only from the pods whose label
+	// matches this definition. This field is optional, and in case it is not set:
+	// results in matching packets from all the pods in the namespace(s)
+	// matched by the NamespaceSelector. In case it is set: is intersected with
+	// the NamespaceSelector, thus matching the packets from the pods
+	// (in the namespace(s) already matched by the NamespaceSelector) which
+	// match this pod selector.
+	// +optional
+	PodSelector metav1.LabelSelector `json:"podSelector,omitempty"`
+	// IPBlock matches the source packets from the pods with IP address in the specified
+	// CIDR block. Use 0.0.0.0/0 to match all pods. This field is optional.
+	// +optional
+	IPBlock networkingv1.IPBlock `json:"ipBlock,omitempty"`
+}
+
+// RoutingPolicyNextHop specifies the next-hop IP address for the policy route.
+type RoutingPolicyNextHop struct {
+	//  NextHopIPs is the list of next-hop IP addresses for this route. With more than
+	// one IP address, ECMP will kick in and one of the IP address will be selected
+	// based on the 5-tuple hashing of the packet header. Currently, only one IP address
+	// is supported. Furthermore, this IP address must be an overlay address, i.e., an
+	// address within the OVN Logical Topology.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:validation:Format=ipv4
+	NextHopIPs []string `json:"ips,omitempty"`
+	// NextHopDevices is a list of device, e.g. switch port, for this route
+    NextHopDevices []*NetworkDevice `json:"devices,omitempty"`
+}
+
+type NetworkDevice struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +resource:path=adminpolicybasedroute
+//+kubebuilder:object:root=true
+
+// AdminPolicyBasedRouteList contains a list of AdminPolicyBasedRoute
+type AdminPolicyBasedRouteList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []AdminPolicyBasedRoute `json:"items"`
+}
+```
+
+#### Example: Partitioned Internet Access
+
+Below is an example of an `AdminPolicyBasedRoute` manifest used to partition internet egress for a namespace, forcing all traffic to a particular next hop. This allows administrators to direct the internet-bound traffic from pods in a given namespace through a specific gateway.
+
+```yaml
+apiVersion: k8s.ovn.org/v1beta1
+kind: AdminPolicyBasedRoute
+metadata:
+  name: internet-egress-namespace1
+spec:
+  networkSelectors:
+    - networkSelectionType: NetworkAttachmentDefinitions
+      networkAttachmentDefinitionSelector:
+        namespaceSelector:
+          matchLabels: {}   # Empty selector will select all namespaces
+        networkSelector:
+          matchLabels:
+            name: ovn-primary
+  policies:
+    - from:
+        namespaceSelector:
+          matchLabels:
+            name: namespace1
+        # may add nodeSelector and podSelector as well
+      to:
+	    cidr: 0.0.0.0/0
+		except:
+          - 10.0.0.0/8
+      nextHop:
+        ips:
+          - "10.128.0.254"      # Next hop IP, typically the overlay IP of internet gateway pod
+```
+
+#### Example: Allowing Istio Health Check Traffic
+
+```yaml
+apiVersion: k8s.ovn.org/v1beta1
+kind: AdminPolicyBasedRoute
+metadata:
+  name: istio-healthcheck-route
+spec:
+  networkSelectors:
+    - networkSelectionType: NetworkAttachmentDefinitions
+      networkAttachmentDefinitionSelector:
+        namespaceSelector:
+          matchLabels: {}   # Empty selector will select all namespaces
+        networkSelector:
+          matchLabels:
+            name: ovn-primary
+  policies:
+    - from:
+	    ipBlock:
+		  cidr: 0.0.0.0/0
+      to:
+        cidr: 169.254.7.127/32
+      nextHop:
+        devices:
+		  - name: k8s-{{NodeName}}
+		    type: LogicalSwitchPort
+```
+
+### Implementation Details
+
+#### Controller
+
+A new `AdminPolicyBasedRouteController` will be added to `ovnkube-master` for central mode, and `ovnkube-node` for interconnect mode.
+
+The controller will mainly:
+
+1. Watch `AdminPolicyBasedRoute` CRs for CRUD operations
+2. Reconcile changes into OVN Northbound DB as `Logical_Router_Policy` entries
+3. Watch Pod, Namespace, and Node events to maintain Address Sets
+4. Handle special cases for interconnect mode and management port routing
+5. Periodically sync between CRs and OVN NB, and keep address sets updated
+
+#### Core Components
+
+**1. Controller Initialization (`go-controller/pkg/ovn/controller/adminpbr/`):**
+
+```go
+type Controller struct {
+	controllerName string
+	util.NetInfo
+    // Kubernetes clients
+    kube              kubernetes.Interface
+    apbrClient        apbrclientset.Interface
+    
+    // OVN clients
+    nbClient          libovsdbclient.Client
+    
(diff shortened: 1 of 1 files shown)


docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+3. Watch Pod, Namespace, and Node events to maintain Address Sets
+4. Handle special cases for interconnect mode and management port routing
+5. Periodically sync between CRs and OVN NB, and keep address sets updated
+
+#### Core Components
+
+**1. Controller Initialization (`go-controller/pkg/ovn/controller/adminpbr/`):**
+
+```go
+type Controller struct {
```
tssurya, line 297: OKEP is not the place to display code - please look at other OKEPs and see how the format is.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
+# OKEP-5831: Add API to support in-cluster policy based routing
+
+* Issue: [#5831](https://github.com/ovn-org/ovn-kubernetes/issues/5831)
+
+## Problem Statement
+
+The workloads running in ovn-kubernetes backed k8s cluster may need policy based routing to
+access external services. For example, in a multi-tenant cluster, tenant Coke needs to go
```
trozet, line 8: nit: probably better to refrain from using Coke/Pepsi as they are trademarked names. We can just use some other dummy names instead.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
+# OKEP-5831: Add API to support in-cluster policy based routing
+
+* Issue: [#5831](https://github.com/ovn-org/ovn-kubernetes/issues/5831)
+
+## Problem Statement
+
+The workloads running in ovn-kubernetes backed k8s cluster may need policy based routing to
+access external services. For example, in a multi-tenant cluster, tenant Coke needs to go 
+through gateway A to access internet, while tenant Pepsi needs to go through gateway B to access
+internet. Both gateway A and B expose internal IP addresses to the clients, and clients can specify
```
trozet, line 10: I think gateway A and B here seem ambiguous. Do you mean that a pod could serve as a gateway, and we want to reroute certain traffic to that pod?

Maybe also mention as another example that in shared gateway mode, a user may desire that some egress traffic is routed via the kernel towards an interface other than the shared gateway interface.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+Custom routing is also needed when Istio CNI is used in Ambient mode, as ingress Kubelet healthcheck
+probes via the mp0 interface will be SNAT'ed to 169.254.7.127, and the ingress-reply packet from the
+pod (which would normally go via shared gateway) needs to route back towards mp0.
+
+## Goals
+
+* Reconcile in-cluster PBR rules based on CRD spec
+* Reconcile address sets in response to Pod/Node/Namespace events
+* Automatically use remote port's IP as nexthop if the specified nexthop is in the remote zone
+* Properly configured CRD can make health check working for Istio/Ambient enabled pods, for interconnect mode only
```
tssurya, line 22: how come there isn't any mention of how this will work with the existing APBER?
these goals all look like implementation details! that's not what goals section looks like.. please refer to other OKEPs on what high level goals are... "Reconciling an address-set" is just an implementation detail..

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+the internal IPs as nexthop.
+
+Custom routing is also needed when Istio CNI is used in Ambient mode, as ingress Kubelet healthcheck
+probes via the mp0 interface will be SNAT'ed to 169.254.7.127, and the ingress-reply packet from the
+pod (which would normally go via shared gateway) needs to route back towards mp0.
+
+## Goals
+
+* Reconcile in-cluster PBR rules based on CRD spec
+* Reconcile address sets in response to Pod/Node/Namespace events
```
trozet, line 20: I think these first 2 bullets are too specific to technical implementation. I think the goals should be higher level. Like:

* Provide a CRD that allows administrators to re-route traffic from pods to other internal IPs, including other pods or the host.

docs/okeps/okep-5831-incluster-admin-pbr.md
tssurya, line 22: `can make health check working for Istio/Ambient enabled pods, for interconnect mode only`

belongs in use-case ^ if you make this a goal then we better add e2es with istio and ovnk in CI :P

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+* Automatically use remote port's IP as nexthop if the specified nexthop is in the remote zone
```
trozet, line 21: what does this one mean exactly? I don't follow this.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+
+When an `AdminPolicyBasedRoute` CR is deleted:
+1. Remove all associated `Logical_Router_Policy` entries from OVN NBDB
+2. Delete all associated Address Sets
+3. Update status and emit events
+
+When an `AdminPolicyBasedRoute` CR is updated:
+1. Reconcile the latest manifest, update OVN with CR name and hash values
+2. Go through the OVN records by CR name and hash, remove records with stale hash values
+
```
tssurya, line 432: cross feature interactions?
how do we plan to make this work with UDNs?/multiple networks?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
++------------------+                            |    | updates
+                                                |    v
++------------------+                            |    +-----------------------------+
+|   Node Events    |----------------------------+    | OVN Address Sets            |
++------------------+                                 +-----------------------------+
+```
+
+### API Details
+
+* A new API `AdminPolicyBasedRoute` under the `k8s.ovn.org/v1beta1` group will be added to
```
tssurya, line 82: I have always regretted the "mouth full words in a CRD" and I believe we shouldn't go beyond 3 words.
I prefer `RouteTable` for a name that indicates what we want as the routing table - i.e a bunch of destinations and nexthops for both internal/external

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+
+## Non-Goals
+
+
+## Introduction
+
+The need for policy-based routing (PBR) within Kubernetes clusters has grown as network architectures become more complex, particularly in multi-tenant and hybrid network environments. Traditional routing assigns the same route to all workloads without regard to context such as namespace, pod, or tenant affinity. However, modern use cases often require that certain workloads use specific gateways or follow unique routing paths, for example to partition internet access by tenant, or to route workloads through VPNs/proxies.
+
+This proposal introduces an API for cluster administrators to configure and manage in-cluster policy-based routing backed by Logical_Router_Policy in OVN. The intent is to allow for declarative, CRD-based configuration of routing policies that respond dynamically to changes in the cluster, such as pods being added, removed, or relocated.
+
```
trozet, line 32: probably good to mention here that admin policy-based external routing CRD already exists (AdminPolicyBasedExternalRoute) and it is used to provide granular routing control for external gateways, while this feature will target Internal Routing within the cluster.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+## User-Stories/Use-Cases
+
+Story 1: Internet access partitioning
+
+As a cluster administrator, I want to specify gateway A for tenant Coke, gateway B for the tenant Pepsi, so that internet traffic can be
```
trozet, line 37: what are these gateways? pods? we should be clear here.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+distributed among multiple gateways.
+  
+Story 2: VPN
+
+As a cluster administrator, I want to specify policy based routing so that my workload can talk to external services through VPN.
+
+Story 3: Readiness check with Istio CNI
+
+As a cluster administrator, I want to enable Istio CNI while keep health check working by rerouting packets to node's management port, this applies to IC mode ovn-kubernetes only.
+
```
trozet, line 47: let's add Story 4: Shared Gateway Mode with some Local Gateway routing

As a cluster administrator operating a shared gateway mode cluster, I want some pod egress traffic to be routed via the host (local gateway mode) so that traffic can then be re-routed by the host out a non-shared gateway interface.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+## Proposed Solution
+
+We will introduce a new cluster-scoped CRD: `AdminPolicyBasedRoute`, to allow cluster administrators to declaratively define in-cluster PBR (Policy-Based Routing) rules, and add a new AdminPolicyBasedRouteController to reconcile these custom resources into OVN's [Logical_Router_Policy](https://man7.org/linux/man-pages/man5/ovn-nb.5.html#Logical_Router_Policy_TABLE).
```
trozet, line 50: s/AdminPolicyBasedRoute/AdminPolicyBasedInternalRoute

see https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/crd/adminpolicybasedroute/v1/types.go#L33

trozet, line 50: we already have a pretty complex apbroute external policy controller. I wonder if we can consolidate it or share a bunch of the code there between internal and external policies.

docs/okeps/okep-5831-incluster-admin-pbr.md
trozet: I see below it says "For interconnect enabled clusters, if the nexthops specified in the CRD are in remote zones, the actual nexthop in Logical_Router_Policy record will be
replaced with the transit IP to the remote zone." which makes this more clear. However, I think this is too technical and specific for an overall functional goal.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+
+For interconnect enabled clusters, if the nexthops specified in the CRD are in remote zones, the actual nexthop in Logical_Router_Policy record will be
+replaced with the transit IP to the remote zone.
+
+To address the routing issue for health check traffic under Istio/Ambient mode, we will introduce a new field `NextHopDevices` to allow users to specify a port name as nexthop, the controller needs to figure out the actual IP and apply it in Logical_Router_Policy record.
```
trozet, line 55: you mean logical_switch_port name? logical_router_port name?

trozet, line 55: maybe these would be better to put where you define the api

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+                                                |    v
++------------------+                            |    +-----------------------------+
+|   Node Events    |----------------------------+    | OVN Address Sets            |
++------------------+                                 +-----------------------------+
+```
+
+### API Details
+
+* A new API `AdminPolicyBasedRoute` under the `k8s.ovn.org/v1beta1` group will be added to  
+  `go-controller/pkg/crd/adminpbr/v1beta1`. This would be a cluster-scoped CRD:
```
trozet, line 83: we already have crd/adminpolicybasedroute so I think it should just go in there as a new type

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+// +genclient
+// +genclient:nonNamespaced
+// +resource:path=adminpolicybasedroute
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:shortName=adminpbr,scope=Cluster
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=".status.status"
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=".metadata.creationTimestamp"
+type AdminPolicyBasedRoute struct {
```
trozet, line 131: AdminPolicyBasedInternalRoute

take a look at the existing types and see how it compares to your definitions here

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	NetworkSelectors crdtypes.NetworkSelectors `json:"networkSelectors,omitempty"`
+
+	// a collection of policy objects to influence routing
+	Policies []RoutingPolicyRule `json:"policies"`
+}
+
+// RoutingPolicyRule is a single routing policy rule object
+type RoutingPolicyRule struct {
+	// From matches the packets for the routing policy
+	From RoutingPolicyMatch `json:"from"`
```
trozet, line 153: RoutingPolicySource?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	// To specifies destination addresses used to match the same in the packets
+	To networkingv1.IPBlock `json:"to"`
```
trozet, line 155: perhaps this should be RoutingPolicyDestination? so that we could perhaps extend it in the future

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	// NextHop defines where the matched packets should be forwarded
+	NextHop RoutingPolicyNextHop `json:"nexthop"`
```
trozet, line 157: so in apb external route we have:

```
type ExternalNextHops struct {
	// StaticHops defines a slice of StaticHop. This field is optional.
	StaticHops []*StaticHop `json:"static,omitempty"`
	//DynamicHops defines a slices of DynamicHop. This field is optional.
	DynamicHops []*DynamicHop `json:"dynamic,omitempty"`
}
```

where:
1. you can have multiple nexthops and they are ecmp routed.
2. you can specify dynamic nexthop, which may be a pod.

Would be good to have that kind of flexibility here as well?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	IPBlock networkingv1.IPBlock `json:"ipBlock,omitempty"`
+}
+
+// RoutingPolicyNextHop specifies the next-hop IP address for the policy route.
+type RoutingPolicyNextHop struct {
+	//  NextHopIPs is the list of next-hop IP addresses for this route. With more than
+	// one IP address, ECMP will kick in and one of the IP address will be selected
+	// based on the 5-tuple hashing of the packet header. Currently, only one IP address
+	// is supported. Furthermore, this IP address must be an overlay address, i.e., an
+	// address within the OVN Logical Topology.
```
trozet, line 191: oh well if we plan in the future to allow multiple then we should make the API name plural: RoutingPolicyNextHops (or you could just call it InternalNextHops)

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=1
+	// +kubebuilder:validation:Format=ipv4
```
trozet, line 194: dont we also need ipv6 support?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	NextHopIPs []string `json:"ips,omitempty"`
+	// NextHopDevices is a list of device, e.g. switch port, for this route
+    NextHopDevices []*NetworkDevice `json:"devices,omitempty"`
+}
+
+type NetworkDevice struct {
+	Name string `json:"name"`
+	Type string `json:"type"`
```
trozet, line 202: this probably needs to be a concrete type. It's too vague to allow any string as a type.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+          matchLabels:
+            name: namespace1
+        # may add nodeSelector and podSelector as well
+      to:
+	    cidr: 0.0.0.0/0
+		except:
+          - 10.0.0.0/8
+      nextHop:
+        ips:
+          - "10.128.0.254"      # Next hop IP, typically the overlay IP of internet gateway pod
```
trozet, line 247: this is too restrictive, because now as a user I need to know the IP of the gateway pod. If that gateway pod is rolled out again, its IP may change, and now my route setup is broken. It's better to use a selector and label one or more gateway pods in that case.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+	// CIDR block. Use 0.0.0.0/0 to match all pods. This field is optional.
+	// +optional
+	IPBlock networkingv1.IPBlock `json:"ipBlock,omitempty"`
+}
+
+// RoutingPolicyNextHop specifies the next-hop IP address for the policy route.
+type RoutingPolicyNextHop struct {
+	//  NextHopIPs is the list of next-hop IP addresses for this route. With more than
+	// one IP address, ECMP will kick in and one of the IP address will be selected
+	// based on the 5-tuple hashing of the packet header. Currently, only one IP address
```
trozet, line 189: how come we cant support more than 1?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+            name: ovn-primary
+  policies:
+    - from:
+	    ipBlock:
+		  cidr: 0.0.0.0/0
+      to:
+        cidr: 169.254.7.127/32
+      nextHop:
+        devices:
+		  - name: k8s-{{NodeName}}
```
trozet, line 274: having to put the exact node name seems pretty rough for UX. I think it would be better to have a device type as LocalManagementPort with a nodeSelector. So something like:

```
devices:
   - type: localManagementPort
     nodeSelector: ...
``` 

which indicates that traffic should always go to the management port on the node that is local to the pod

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+		    type: LogicalSwitchPort
+```
+
+### Implementation Details
+
+#### Controller
+
+A new `AdminPolicyBasedRouteController` will be added to `ovnkube-master` for central mode, and `ovnkube-node` for interconnect mode.
```
trozet, line 282: we are removing central mode, so no need to add any support for it.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+
+The controller will mainly:
+
+1. Watch `AdminPolicyBasedRoute` CRs for CRUD operations
+2. Reconcile changes into OVN Northbound DB as `Logical_Router_Policy` entries
+3. Watch Pod, Namespace, and Node events to maintain Address Sets
+4. Handle special cases for interconnect mode and management port routing
+5. Periodically sync between CRs and OVN NB, and keep address sets updated
+
```
trozet, line 291: will LRPs only be configured where the source is and the destination lives? For example lets say i match on pod A source on node 1, and dest is a pod B on node 3. Will node 2 not have any LRP since it is not in the traffic path?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+When an `AdminPolicyBasedRoute` CR is created or updated:
+
+```
+1. Validate the CR spec (network attachment name, selectors, nexthop IPs, etc.)
+2. For each policy rule in the CR:
+   a. Build OVN match expression from From selectors (node/namespace/pod)
+   b. Build Address Sets for dynamic pod/namespace/node matching
+   c. Find what nexthop IP to use in the policy
+      - If nexthop is an IP and it's in remote zone, find the node's transit port IP
```
trozet, line 347: we probably need a feature interaction/compatibility matrix section. You can find examples in other OKEPs, but this particular line made me wonder what happens with this feature and no-overlay? There wont be any transit switch/geneve. Will it still be supported?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+Address Sets are dynamically updated when:
+- Pods matching the selector are added/removed/updated
+- Namespaces matching the selector are added/removed/updated  
+- Nodes matching the selector are added/removed/updated
+
+**4. OVN Logical Router Policy Creation:**
+
+Each `RoutingPolicyRule` is translated to an OVN `Logical_Router_Policy` with:
+
+- **Priority:** 80 (processed after OVN-K8s internal rules)
```
trozet, line 375: so then should not override egress ip and be compatible?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+| (nexthop: Transit Switch Port IP of Node B)          |                  |
+|             |                                        |                  |
++-------------|----------------------------------------|------------------+
+              |                                        |
+       +-------------+                                 |
+       | Transit     |                                 |
+       | Switch Port |----------------------------------
+       | of Node B   |
+       +-------------+
+
```
trozet, line 421: missing terminating backticks here ```

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+2. Delete all associated Address Sets
+3. Update status and emit events
+
+When an `AdminPolicyBasedRoute` CR is updated:
+1. Reconcile the latest manifest, update OVN with CR name and hash values
+2. Go through the OVN records by CR name and hash, remove records with stale hash values
+
+#### Configuration
+
+Add a new feature gate flag `--enable-admin-pbr` (default: true) to turn on/off the feature.
```
trozet, line 435: we should probably also gate external apbroute behind this flag as well since right now it doesn't have one

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+
+type Status string
+
+const (
+	StatusSucceeded Status = "Succeeded"
+	StatusFailed    Status = "Failed"
+)
+
+// AdminPolicyBasedRouteStatus describes the current status of the AdminPolicyBasedRoute.
+type AdminPolicyBasedRouteStatus struct {
```
trozet, line 101: would it be good to list int he status what the currently serving nextHopIPs are? or maybe that is too much if we select a ton of management ports.

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+#### Example: Allowing Istio Health Check Traffic
+
+```yaml
+apiVersion: k8s.ovn.org/v1beta1
+kind: AdminPolicyBasedRoute
+metadata:
+  name: istio-healthcheck-route
+spec:
+  networkSelectors:
+    - networkSelectionType: NetworkAttachmentDefinitions
```
trozet, line 259: we can only select one network right? or do we support multiple networks? What about with CNC where we might have network A connected to network B. Would we support a source being in network A and the packet being re-routed to network B pod as a dest?

If so then then networkSelector would need to be part of the from/to.

trozet (changes requested):

docs/okeps/okep-5831-incluster-admin-pbr.md
aserdean: Red/Blue?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+The workloads running in ovn-kubernetes backed k8s cluster may need policy based routing to
+access external services. For example, in a multi-tenant cluster, tenant Coke needs to go 
+through gateway A to access internet, while tenant Pepsi needs to go through gateway B to access
+internet. Both gateway A and B expose internal IP addresses to the clients, and clients can specify
+the internal IPs as nexthop.
+
+Custom routing is also needed when Istio CNI is used in Ambient mode, as ingress Kubelet healthcheck
+probes via the mp0 interface will be SNAT'ed to 169.254.7.127, and the ingress-reply packet from the
+pod (which would normally go via shared gateway) needs to route back towards mp0.
+
```
aserdean, line 16: Maybe it is worth to add a concrete example with concrete values

docs/okeps/okep-5831-incluster-admin-pbr.md
aserdean: Feel free to link the issue https://github.com/ovn-kubernetes/ovn-kubernetes/issues/5827 to that story

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+1. Validate the CR spec (network attachment name, selectors, nexthop IPs, etc.)
+2. For each policy rule in the CR:
+   a. Build OVN match expression from From selectors (node/namespace/pod)
+   b. Build Address Sets for dynamic pod/namespace/node matching
+   c. Find what nexthop IP to use in the policy
+      - If nexthop is an IP and it's in remote zone, find the node's transit port IP
+	  - If nexthop device is provided, find out the IP of the device
+   d. Add exclusion list to policy if there is any
+   e. Create/Update Logical_Router_Policy in OVN NBDB
+3. Update CR status with success/error messages
```
aserdean, line 351: What will happen in case of an error, will it be retried?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+       | Transit     |                                 |
+       | Switch Port |----------------------------------
+       | of Node B   |
+       +-------------+
+
+**6. Cleanup and Garbage Collection:**
+
+When an `AdminPolicyBasedRoute` CR is deleted:
+1. Remove all associated `Logical_Router_Policy` entries from OVN NBDB
+2. Delete all associated Address Sets
```
aserdean, line 426: retry on failure?

docs/okeps/okep-5831-incluster-admin-pbr.md
aserdean: should this be only on the helm side or also on kind.sh?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+* Unit tests coverage
+- General unit test cases for functions
+- Ginkgo BDD test cases to validate controller behaviors
+- Make sure status is updated as expected.
+
+* E2E Testing details
+- Create 3 pods to simulate source, destination and nexthop, configure NAT on the nexthop
+- Apply a `AdminPolicyBasedRoute` CR to redirect packets from source to go through nexthop to reach destination
+- Run ping from source, and run tcpdump on nexthop pod, packets are expected to be captured
+- Change the pod selector in CR to disqualify the source pod, run above test again, packets should not be captured
```
aserdean, line 448: Would it be possible to add a e2e test for this userstory: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6215/changes#r3048052403?

docs/okeps/okep-5831-incluster-admin-pbr.md
```diff
@@ -0,0 +1,465 @@
...
+
+### Documentation Details
+* ovn-kubernetes.io will be updated with a user guide
+
+## Risks, Known Limitations and Mitigations
+
+## OVN-Kubernetes Version Skew
+Targeting version 1.4.
```
aserdean, line 456: 1.3?

docs/okeps/okep-5831-incluster-admin-pbr.md
jxiaobin: > in shared gateway mode, a user may desire that some egress traffic is routed via the kernel towards an interface other than the shared gateway interface.

@trozet is this for the Istio/Ambient related problem?

jxiaobin:
@aserdean @trozet  @tssurya  thanks for reviewing the OKEP, I'm revising it based on the comments.

docs/okeps/okep-5831-incluster-admin-pbr.md
trozet: no 1.3 is about to release this week, it will have to hit 1.4

PR #6228, opened 2026-04-09
OKEP-6227: Support per-destination egress IP routing

### LINKED ISSUE #6227: OKEP: Support per-destination egress IP routing

Allow a cluster administrator to define destination network CIDRs via a new `EgressIPTraffic` CRD.

`EgressIP` resources can select `EgressIPTraffic` resources via a `trafficSelector` label selector, routing only matching destination traffic through that egress IP.

Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks.

### Why is this needed?

The existing EgressIP feature routes all egress traffic from a matched pod through a single egress IP address. There is no mechanism to assign different egress IP addresses based on the traffic's destination network, which is required in environments where workloads communicate with multiple external networks through different interfaces.

cgoncalves:
Add OKEP documenting the EgressIPTraffic feature design, covering problem statement, goals, API details, implementation details for both OVN central and node controllers, testing strategy, risks, backwards compatibility, and alternatives.

This design is based on https://github.com/openshift/enhancements/pull/1594 by @martinkennelly.

Implementation PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6237

Closes: #6227

### DIFF
--- a/docs/okeps/okep-6227-egressiptraffic.md
+++ b/docs/okeps/okep-6227-egressiptraffic.md
@@ -1,0 +1,355 @@
+# OKEP-6227: EgressTrafficClass — Per-Destination Egress IP Routing
+
+* Issue: [#6227](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6227)
+
+## Problem Statement
+
+The existing EgressIP feature routes all egress traffic from a matched pod through a single egress IP address. There is no mechanism to assign different egress IP addresses based on the traffic's destination network, which is required in environments where workloads communicate with multiple external networks through different interfaces.
+
+## Goals
+
+* Allow a cluster administrator to define destination network CIDRs via a new `EgressTrafficClass` CRD.
+* Allow an `EgressIP` to select `EgressTrafficClass` resources via a `trafficSelector` label selector, routing only matching destination traffic through that egress IP.
+* Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks.
+* Support coexistence of `trafficSelector` EgressIPs with a non-`trafficSelector` "default" EgressIP for catch-all traffic.
+* Work with EgressIPs assigned to both OVN primary network interfaces and secondary host interfaces.
+* Support both local and shared OVN gateway modes.
+* Support EgressIPs on user-defined networks (UDNs) for OVN-network EgressIPs. Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP limitation on UDNs).
+
+## Non-Goals
+
+* Dynamic route synchronization with BGP: per-destination routing tables are snapshots; BGP integration is a separate effort ([OKEP-5296](okep-5296-bgp.md)).
+
+## Introduction
+
+In telecommunications and carrier deployments, a single pod often needs to communicate with multiple external networks; for example, OAM (Operations, Administration, Maintenance), signaling, and user plane networks. Each network requires traffic to egress through a specific interface with a specific source IP address. The existing EgressIP feature assigns a single egress IP to all traffic from a pod, making it impossible to route different destination traffic through different interfaces.
+
+EgressTrafficClass solves this by introducing destination-based egress routing. A new `EgressTrafficClass` CRD defines sets of destination network CIDRs, and a new `trafficSelector` field on `EgressIP` selects which `EgressTrafficClass` resources apply. Only traffic matching the selected destination networks is routed through the EgressIP; all other traffic follows normal OVN routing or is handled by a non-`trafficSelector` default EgressIP.
+
+### Behavioral change: multi-EgressIP pod support
+
+Prior to this OKEP, having multiple `EgressIP` objects selecting the same pod was considered **undefined behavior**. The controller would arbitrarily pick one EgressIP as the "primary" and put the rest on standby. There was no guarantee about which EgressIP would serve the pod, and failover between them was non-deterministic.
+
+This OKEP changes that behavior to be **deterministic** when `trafficSelector` is used:
+
+- Multiple `EgressIP` objects with different `trafficSelector` values can coexist on the same pod, each handling traffic to different destination networks.
+- A non-`trafficSelector` "default" EgressIP can coexist with `trafficSelector` EgressIPs, acting as a catch-all for traffic not matching any destination network.
+- Priority-based LRP evaluation (100 for `trafficSelector`, 99 for default) ensures consistent, predictable routing.
+- Without `trafficSelector`, the existing behavior is unchanged: multiple EgressIPs selecting the same pod remain undefined.
+
+### Example topology
+
+```text
+                        ┌───────────────────────────────────────────┐
+                        │              Egress Node                  │
+                        │                                           │
+Pod ──OVN──►            │  ┌─ eth-mgmt ──► Mgmt Network             │
+  10.244.1.3            │  │   (192.168.150.101)  (192.168.250.0/24)│
+                        │  │                                        │
+         reroute ──────►│──┤                                        │
+                        │  │                                        │
+                        │  └─ eth-sig  ──► Signaling Network        │
+                        │      (192.168.200.101)  (192.168.251.0/24)│
+                        └───────────────────────────────────────────┘
+```
+
+## User-Stories/Use-Cases
+
+**Story 1: Per-destination egress IP for multi-network workloads**
+
+As a cluster administrator, I want pods in a telco namespace to use different source IP addresses depending on which backend network they are reaching, so that each network sees traffic from the correct interface and IP range.
+
+**Story 2: Default egress IP with destination-specific overrides**
+
+As a cluster administrator, I want a "default" EgressIP that handles general internet traffic, combined with destination-specific EgressIPs for particular backend networks, so that pods have predictable source IPs for all egress traffic without requiring per-destination configuration for every possible destination.
+
+## Proposed Solution
+
+### API Details
+
+#### EgressTrafficClass CRD
+
+A new cluster-scoped custom resource that defines destination network CIDRs. The name `EgressTrafficClass` is intentionally generic so other features (e.g., EgressFirewall, EgressQoS) may leverage it in the future:
+
+```go
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:resource:shortName=etc,scope=Cluster
+type EgressTrafficClass struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata"`
+    Spec              EgressTrafficClassSpec `json:"spec"`
+}
+
+type EgressTrafficClassSpec struct {
+    // DestinationNetworks is the list of Network CIDRs that defines external
+    // destinations for IPv4 and/or IPv6.
+    // +kubebuilder:validation:MaxItems=25
+    DestinationNetworks []CIDR `json:"destinationNetworks,omitempty"`
+}
+
+// CIDR is a network CIDR. IPv4 or IPv6.
+// +kubebuilder:validation:MaxLength=43
+// +kubebuilder:validation:XValidation:rule="isCIDR(self)"
+type CIDR string
+```
+
+Key points:
+- Cluster-scoped (not namespaced).
+- Labels on the object are used for selection by `EgressIP.spec.trafficSelector`.
+- CIDRs are validated at admission time via CEL (`isCIDR(self)`).
+- Maximum 25 destination networks per resource.
+
+#### EgressIP TrafficSelector field
+
+A new optional pointer field on `EgressIP.spec`:
+
+```go
+type EgressIPSpec struct {
+    EgressIPs         []string              `json:"egressIPs"`
+    NamespaceSelector metav1.LabelSelector  `json:"namespaceSelector"`
+    PodSelector       metav1.LabelSelector  `json:"podSelector"`
+    TrafficSelector   *metav1.LabelSelector `json:"trafficSelector,omitempty"` // NEW
+}
+```
+
+The field is a pointer to distinguish between:
+- **nil** (field absent): no traffic selector — all traffic is selected (existing behavior, fully backwards compatible).
+- **empty** (`trafficSelector: {}`): matches all `EgressTrafficClass` resources (selects all labeled destination networks).
+- **non-empty** (`trafficSelector: {matchLabels: {...}}`): matches only `EgressTrafficClass` resources whose labels satisfy the selector.
+
+When `trafficSelector` is set (non-nil), the EgressIP only handles traffic to the destination networks defined in matching `EgressTrafficClass` resources.
+
+**When no `EgressTrafficClass` matches the `trafficSelector`**: The destination networks address set is empty, the reroute LRP matches no traffic, and the EgressIP effectively stops routing traffic. A warning is logged and a Kubernetes event is emitted on the EgressIP object:
+
+```
+Warning  NoDestinationNetworks  TrafficSelector is set but no matching EgressTrafficClass
+                                 destination networks were found; traffic will not be routed
+                                 via this EgressIP until matching EgressTrafficClass resources
+                                 are created
+```
+
+This is by design: no matching destination networks means no traffic should use this EgressIP. Operators should create the required `EgressTrafficClass` resources or verify the `trafficSelector` labels.
+
+#### EgressIP Status Conditions
+
+Two new conditions are added to the EgressIP status to improve observability:
+
+```go
+type EgressIPStatus struct {
+    Items      []EgressIPStatusItem `json:"items"`
+    Conditions []metav1.Condition   `json:"conditions,omitempty"` // NEW
+}
+```
+
+**`TrafficSelectorResolved`**: Set to `True` when the `trafficSelector` successfully resolves to at least one `EgressTrafficClass` with destination networks. Set to `False` when `trafficSelector` is set but no matching `EgressTrafficClass` resources are found.
+
+```yaml
+conditions:
+- type: TrafficSelectorResolved
+  status: "True"
+  reason: DestinationNetworksFound
+  message: "Resolved 2 destination networks from 1 EgressTrafficClass resource(s)"
+```
+
+```yaml
+conditions:
+- type: TrafficSelectorResolved
+  status: "False"
+  reason: NoMatchingEgressTrafficClass
+  message: "No EgressTrafficClass resources match trafficSelector"
+```
+
+**`TrafficConflict`**: Set to `True` when overlapping destination CIDRs are detected across multiple EgressIPs with `trafficSelector` targeting the same pod. The controller performs cross-EgressIP analysis during pod assignment to identify CIDRs that overlap between different EgressIPs.
+
+```yaml
+conditions:
+- type: TrafficConflict
+  status: "True"
+  reason: OverlappingDestinationCIDR
+  message: "Destination CIDR 192.168.250.0/24 overlaps with EgressIP eip-signaling for pod selector app=telco-workload"
+```
+
+#### Example resources
+
+```yaml
+apiVersion: k8s.ovn.org/v1
+kind: EgressTrafficClass
+metadata:
+  name: mgmt-traffic
+  labels:
+    traffic-group: mgmt
+spec:
+  destinationNetworks:
+  - "192.168.250.0/24"
+---
+apiVersion: k8s.ovn.org/v1
+kind: EgressIP
+metadata:
+  name: eip-mgmt
+spec:
+  egressIPs:
+  - 192.168.150.101
+  namespaceSelector:
+    matchLabels:
+      env: production
+  podSelector:
+    matchLabels:
+      app: telco-workload
+  trafficSelector:
+    matchLabels:
+      traffic-group: mgmt
+```
+
+### Feature Gate
+
+The feature is gated by `--enable-egress-traffic-class` (config: `enable-egress-traffic-class`). When enabled, it implicitly enables `--enable-egress-ip` since EgressTrafficClass extends EgressIP. When disabled (default), the EgressTrafficClass informer is not registered, `trafficSelector` on EgressIP is ignored, and no destination-filtered LRPs or NATs are created.
+
+### Implementation Details
+
+The feature is implemented across two controllers:
+
+1. **OVN central controller** (`pkg/ovn/egressip.go`): Manages OVN logical router policies (LRPs), NAT rules, and address sets in the NB database.
+2. **Node controller** (`pkg/node/controllers/egressip/egressip.go`): Manages Linux IP rules, routing tables, and iptables SNAT rules on the egress node.
+
+#### OVN central controller changes
+
+**Destination networks address set**: For each EgressIP with `trafficSelector`, an OVN address set is created containing the destination CIDRs from all matching `EgressTrafficClass` resources. The address set is updated when `EgressTrafficClass` resources are created, updated, or deleted.
+
+**Destination-filtered reroute LRP**: The reroute logical router policy match is extended with a destination filter when `trafficSelector` is set:
+
+```text
+# With trafficSelector (priority 100):
+ip4.src == 10.244.1.3 && ip4.dst == $<dest-nets-addr-set>    reroute    10.244.2.2
+
+# Without trafficSelector, standard behavior (priority 100):
+ip4.src == 10.244.1.3                                         reroute    100.64.0.3
+
+# Without trafficSelector, coexisting as default (priority 99):
+ip4.src == 10.244.1.3                                         reroute    100.64.0.3
+```
+
+**Destination-filtered NAT**: When the EgressIP is on the OVN network and has `trafficSelector`, the SNAT rule on the gateway router includes a destination match to avoid conflicts with other EgressIPs on the same node:
+
+```text
+snat    172.18.0.200    10.244.1.3    match="ip4.dst == $<dest-nets-addr-set>"    (trafficSelector)
+snat    172.18.0.100    10.244.1.3                                                (default, no match)
+```
+
+**EgressIP coexistence**: Multiple EgressIPs can serve the same pod when at least one has `trafficSelector`. The priority scheme ensures correct evaluation order:
+
+| EgressIP | TrafficSelector | LRP Priority | IP Rule Priority | Role |
+|----------|----------------|-------------|-----------------|------|
+| eip-mgmt | yes | 100 | 6000 (per-dst) | Handles traffic to 192.168.250.0/24 |
+| eip-signaling | yes | 100 | 6000 (per-dst) | Handles traffic to 192.168.251.0/24 |
+| eip-default | no | 99 | 6001 (catch-all) | Handles all other traffic |
+
+**SNAT rule collision**: When two trafficSelector EgressIPs target the same pod and the same secondary host interface, they share the same SNAT IP (since the EgressIP address is assigned to that interface). The iptables SNAT rule (`-s <podIP> -o <iface> -j SNAT --to-source <egressIP>`) is identical for both, so no collision occurs. Different interfaces produce different `-o` flags, also preventing collision.
+
+**Reconciliation**: `EgressTrafficClass` changes trigger re-reconciliation of all EgressIPs with matching `trafficSelector`. TrafficSelector changes on an existing EgressIP (CASE 3.0) trigger a full teardown and rebuild of assignments.
+
+#### Node controller changes
+
+**Per-destination IP rules**: Instead of a single catch-all IP rule per pod, one IP rule per destination CIDR is created at priority 6000:
+
+```text
+6000: from 10.244.1.3 to 192.168.250.0/24 lookup 1026
+6000: from 10.244.1.3 to 192.168.251.0/24 lookup 1028
+```
+
+Non-`trafficSelector` catch-all rules use priority 6001:
+
+```text
+6001: from 10.244.1.3 lookup 1028
+```
+
+**Destination-specific routing table**: The default route is replaced with destination-specific routes, preserving the gateway from the original default route. If a specific route for the destination CIDR already exists on the interface (e.g., configured by the network administrator), it is preserved with its gateway rather than replaced. Non-matching traffic falls through to the main routing table:
+
+```text
+# With trafficSelector:
+192.168.250.0/24 via 192.168.150.1 dev eth-mgmt
+192.168.150.0/24 dev eth-mgmt proto kernel scope link
+
+# Without trafficSelector (standard):
+default via 192.168.150.1 dev eth-mgmt
+192.168.150.0/24 dev eth-mgmt proto kernel scope link
+```
+
+**iptables SNAT deduplication**: When multiple destination CIDRs use the same interface, the iptables SNAT rule is created only once (on the first destination CIDR config) to avoid duplicate add/delete issues.
+
+**EgressTrafficClass informer**: The node controller watches `EgressTrafficClass` resources and re-queues affected EgressIPs when destination networks change. The update handler only processes the old object when labels changed (optimization to avoid unnecessary re-reconciliation when only `destinationNetworks` change).
+
+#### Differences between LGW and SGW modes
+
+The feature works identically in both local and shared gateway modes. The destination-filtered LRP match, address sets, NAT rules, and node-side constructs are gateway-mode-agnostic.
+
+### Testing Details
+
+#### Unit Tests
+
+**OVN controller tests** (`pkg/ovn/egressip_test.go`):
+- `hasTrafficSelector`: nil, empty, MatchLabels, MatchExpressions
+- `egressStatuses.containsAny`: empty, matching, non-matching
+- LRP match parsing: standard match, destination-filtered match, too-short match, IPv6
+- NAT match: IPv4/IPv6 with and without destination match
+- EgressIP reroute priority: non-trafficSelector uses 99, trafficSelector uses 100
+
+**Node controller tests** (`pkg/node/controllers/egressip/egressip_test.go`):
+- TrafficSelector EgressIP with single destination network
+- TrafficSelector EgressIP with multiple destination networks (SNAT deduplication)
+- Non-TrafficSelector EgressIP with lower priority catch-all rule
+- `replaceDefaultRouteWithDestNetworks`: default route removal, gateway preservation, existing specific route preservation, IPv4/IPv6 filtering, empty destinations
+- `hasRouteForDst`: matching CIDR, different CIDR, same IP different mask, empty routes
+- `containsRoute`: matching, non-matching, empty
+- `getRoutesForOtherEIPsOnLink`: co-located routes, empty when alone
+
+#### E2E Tests
+
+Six E2E test cases tagged `[secondary-host-eip] [traffic-selector]`:
+
+1. **Matching and non-matching traffic**: Verifies matching destination traffic uses EgressIP, non-matching uses node IP, and EgressTrafficClass deletion breaks connectivity to the destination network.
+2. **Multiple TrafficSelector EgressIPs**: Two secondary networks, two EgressTrafficClass CRs, two EgressIPs; verifies each destination uses the correct EgressIP.
+3. **Coexistence**: TrafficSelector EgressIP on secondary host network (priority 100) + non-TrafficSelector default EgressIP on OVN network (priority 99); verifies both work independently.
+4. **EgressTrafficClass destinationNetworks update**: Changes CIDRs on a live EgressTrafficClass, verifies traffic transitions from non-matching to matching.
+5. **EgressTrafficClass label change**: Changes labels so the EgressTrafficClass stops matching the trafficSelector, verifies traffic stops routing via EgressIP.
+6. **TrafficSelector added to existing EgressIP**: Starts with standard EgressIP, adds trafficSelector, verifies transition from catch-all to destination-filtered routing.
+
+All E2E tests pass on both local and shared gateway modes. Tests skip when the EgressTrafficClass feature gate is not enabled.
+
+### Documentation Details
+
+Feature documentation will be added at `docs/features/cluster-egress-controls/egress-ip-traffic.md` covering:
+- Workflow with YAML examples (basic and advanced setups)
+- Implementation details with traffic flow diagrams
+- OVN and node-side construct descriptions
+- Troubleshooting guide
+- Best practices, design notes, known limitations
+
+## Risks, Known Limitations and Mitigations
+
+- **Overlapping destination CIDRs**: If two EgressIPs with different `trafficSelector` match overlapping CIDRs for the same pod, the OVN LRP behavior at the same priority is undefined. Mitigation: the `TrafficConflict` status condition detects and surfaces overlapping CIDRs; operators should resolve them by adjusting destination networks.
+- **BGP route staleness**: Per-destination routing tables on egress nodes are snapshots. Dynamic route changes (e.g., via BGP) are not automatically reflected until the next sync or EgressIP reconciliation. Mitigation: future BGP integration ([OKEP-5296](okep-5296-bgp.md)) will address this with netlink route subscriptions.
+- **UDN secondary host interface**: EgressIPs with `trafficSelector` on user-defined networks are supported for OVN-network EgressIPs (where SNAT happens on the gateway router). Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP UDN limitation).
+- **Catch-all secondary host EgressIP with OVN-network trafficSelector EgressIP in IC mode**: A catch-all EgressIP (no `trafficSelector`) on a secondary host interface captures ALL pod traffic at the kernel IP rule level (priority 6001). In interconnect (IC) mode, this prevents coexisting `trafficSelector` EgressIPs on the OVN primary network from working for the same pod, because both the secondary-host and OVN-network EgressIP LRPs reroute to the same transit switch nexthop. The packet arrives at the egress node's management port and is intercepted by the catch-all IP rule before the OVN gateway router can apply SNAT for the OVN-network EgressIP. In non-IC mode (single zone), this works correctly because the OVN-network EgressIP reroutes to the gateway router nexthop directly.
+
+## OVN-Kubernetes Version Skew
+
+This feature is proposed for introduction in a future release. The `EgressTrafficClass` CRD and `trafficSelector` field are additive; existing EgressIP objects without `trafficSelector` continue to work unchanged.
+
+## Backwards Compatibility
+
+Fully backwards compatible:
+- The `trafficSelector` field on `EgressIP` is optional (pointer, nil by default). Existing EgressIP objects without it behave exactly as before.
+- The `EgressTrafficClass` CRD is new and has no impact on existing resources.
+- Non-`trafficSelector` EgressIPs use IP rule priority 6001 (changed from 6000) on the node side. The `repairNode` sync handles the transition by scanning both priorities and cleaning up stale rules.
+
+## Alternatives
+
+1. **Per-pod secondary network attachments**: Instead of destination-based routing, each pod could attach to multiple networks directly. This requires application-level awareness and significantly complicates pod networking.
+2. **EgressFirewall with SNAT**: Using EgressFirewall rules combined with custom SNAT configurations. This would require significant changes to the EgressFirewall feature and doesn't integrate with EgressIP's node assignment and HA capabilities.
+3. **Destination CIDRs inline on EgressIP**: Instead of a separate `EgressTrafficClass` CRD, destination networks could be specified directly on the `EgressIP` spec. This was rejected because it doesn't allow reuse of destination network definitions across multiple EgressIPs and makes the EgressIP spec overly complex.
+
+## References
+
+- [EgressIP documentation](../features/cluster-egress-controls/egress-ip.md)
+- [GitHub Issue #6227](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6227)
--- a/.cursor/worktrees.json
+++ b/.cursor/worktrees.json
@@ -1,0 +1,5 @@
+{
+  "setup-worktree": [
+    "npm install"
+  ]
+}
--- a/docs/okeps/okep-6227-egress-traffic-class.md
+++ b/docs/okeps/okep-6227-egress-traffic-class.md
@@ -1,0 +1,555 @@
+# OKEP-6227: EgressTrafficClass — Selective EgressIP
+
+* Issue: [#6227](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6227)
+
+## Problem Statement
+
+The existing EgressIP feature applies to all egress traffic from a matched pod. An EgressIP object may list several addresses (ECMP among assigned next hops), but there is no way to choose a different EgressIP — and thus a different source IP or egress interface — based on destination CIDR or L4 protocol/port.
+
+## Goals
+
+* Allow a cluster administrator to define destinations (destination CIDRs with optional L4 protocol/port filtering) via a new `EgressTrafficClass` CRD.
+* Allow an `EgressIP` to select `EgressTrafficClass` resources via a `trafficSelector` label selector, routing only matching traffic through that egress IP.
+* Support L4-level filtering: route traffic based on IP protocol (TCP/UDP/SCTP) and optional destination and/or source port or port range, using an NPEP-187-style `protocols` list so additional protocols can be added later without a breaking API change.
+* Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks or protocols.
+* Support coexistence of `trafficSelector` EgressIPs on secondary host interfaces with a non-`trafficSelector` default EgressIP on the OVN network (Story 2).
+* Work with EgressIPs assigned to both OVN primary network interfaces and secondary host interfaces.
+* Support both local and shared OVN gateway modes.
+* Support EgressIPs on user-defined networks (UDNs) for OVN-network EgressIPs (SNAT on the gateway router).
+
+## Non-Goals
+
+* ICMP / non-port protocols in this OKEP: the `protocols` list is shaped so ICMP (or GRE, TCP flags, …) can be added later; this OKEP only defines TCP, UDP, and SCTP.
+* Cross-feature consumption of `EgressTrafficClass` by NetworkQoS, ClusterNetworkPolicy, or other APIs in this OKEP. The type is a named **destination match** (CIDR + optional L4), not a policy: it has no allow/deny/drop, no rule order, and no ingress/`from` match. A later OKEP may select it from other **egress** consumers (for example a ClusterNetworkPolicy Allow rule or NetworkQoS dest matching); action and direction stay on those APIs. This OKEP's only consumer is EgressIP. A generic `TrafficClass` or ingress reuse would need `sources` (or `ingress`/`egress` subtrees) and is out of scope.
+* Catch-all EgressIP on a secondary host interface coexisting with a `trafficSelector` EgressIP on the OVN network for the same pod. Both LRPs reroute to the same transit switch; the packet arrives on the assigned node's management port; the catch-all `6001: from <podIP>` IP rule then steals dests that should have reached gateway-router SNAT for the OVN EgressIP. This OKEP does not add dest-exclusion host IP rules to undo that.
+
+## Future Goals
+
+* Secondary host interface EgressIPs on UDNs, including with `trafficSelector`, once base secondary EgressIP on UDN exists. Reuse this OKEP's host path (per-destination IP rules and dest-scoped iptables SNAT) inside the UDN routing domain (VRF), rather than a second match model. A likely vehicle is Uplink-backed CUDN gateways ([OKEP-6019](okep-6019-vrf-lite-shared-gateway-external-bridges.md)), which already lists secondary EgressIP through `Uplink` as a future consumer.
+
+## Introduction
+
+In telecommunications and carrier deployments, a single pod often needs to reach multiple external networks (for example OAM, signaling, and user plane). Each network must see traffic leave an **egress node** interface with an **EgressIP** source address (not the pod IP, and not the node IP). Local-gateway masquerade to node IPs is not enough: those addresses are per-node, move when the pod or assignment moves, and are not the HA EgressIP that external ACLs allowlist. Secondary-host EgressIP can already steer via host routes, but if a pod matches more than one EgressIP the winner is undefined — operators cannot pin which EgressIP (hence which egress node interface and source IP) applies by destination CIDR or L4. Different services on the same network (e.g. SIP on UDP 5060 vs RTP on UDP 16384–32767) may also need different EgressIPs.
+
+EgressTrafficClass solves this by introducing destination-based egress routing with optional L4 filtering. A new `EgressTrafficClass` CRD defines destinations — destination CIDRs with optional protocol/port filters — and a new `trafficSelector` field on `EgressIP` selects which `EgressTrafficClass` resources apply. A `trafficSelector` EgressIP only handles matching destinations. Non-matching traffic is not handled by that EgressIP: it uses a coexisting non-`trafficSelector` default EgressIP if one is assigned to the pod (Story 2), otherwise normal OVN egress (typically the node IP). Omitting `trafficSelector` is unchanged EgressIP behavior: all of the pod's egress uses that EgressIP.
+
+### Behavioral change: multi-EgressIP pod support
+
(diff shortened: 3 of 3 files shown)


cgoncalves:
@coderabbitai resume

cgoncalves:
@trozet What are your thoughts on extending the OKEP to accommodate granular routing filtration based on IP protocol as well as source and destination ports? Currently it's destination CIDR only.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+* Allow a cluster administrator to define destination network CIDRs via a new `EgressIPTraffic` CRD.
+* Allow an `EgressIP` to select `EgressIPTraffic` resources via a `trafficSelector` label selector, routing only matching destination traffic through that egress IP.
+* Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks.
+* Support coexistence of `trafficSelector` EgressIPs with a non-`trafficSelector` "default" EgressIP for catch-all traffic.
+* Work with EgressIPs assigned to both OVN primary network interfaces and secondary host interfaces.
+* Support both local and shared OVN gateway modes.
+
+## Non-Goals
+
+* Support for user-defined networks (UDNs): only the cluster default network is supported.
```
trozet: I think this has to be a goal. UDN is a first class citizen in OVNK, and releasing a feature that supports CDN and not UDN at this point seems half baked.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Problem Statement
+
+The existing EgressIP feature routes all egress traffic from a matched pod through a single egress IP address. There is no mechanism to assign different egress IP addresses based on the traffic's destination network, which is required in environments where workloads communicate with multiple external networks through different interfaces.
+
+## Goals
+
+* Allow a cluster administrator to define destination network CIDRs via a new `EgressIPTraffic` CRD.
+* Allow an `EgressIP` to select `EgressIPTraffic` resources via a `trafficSelector` label selector, routing only matching destination traffic through that egress IP.
+* Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks.
```
trozet: Somewhere in the document, maybe Intro or Proposed Solution we need to clearly outline the behavioral definition pre and post this OKEP where today multiple egress IPs selecting the same pod is considered "undefined behavior" to now you are saying that multiple egress IPs with different traffic selectors is supported and deterministic behavior.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+# With trafficSelector:
+192.168.250.0/24 via 192.168.150.1 dev eth-mgmt
+192.168.150.0/24 dev eth-mgmt proto kernel scope link
+
+# Without trafficSelector (standard):
+default via 192.168.150.1 dev eth-mgmt
+192.168.150.0/24 dev eth-mgmt proto kernel scope link
+```
+
+**iptables SNAT deduplication**: When multiple destination CIDRs use the same interface, the iptables SNAT rule is created only once (on the first destination CIDR config) to avoid duplicate add/delete issues.
```
trozet: how does this work if I have 2 Egress IPs that select the same egress interface, and the same pod? Don't you need to conditionalize the SNAT rule based on dest IP as well?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+#### EgressIP TrafficSelector field
+
+A new optional field on `EgressIP.spec`:
+
+```go
+type EgressIPSpec struct {
+    EgressIPs         []string             `json:"egressIPs"`
+    NamespaceSelector metav1.LabelSelector `json:"namespaceSelector"`
+    PodSelector       metav1.LabelSelector `json:"podSelector"`
+    TrafficSelector   metav1.LabelSelector `json:"trafficSelector,omitempty"` // NEW
```
trozet: shouldn't this be a pointer so we can distinguish between nil and empty? Empty means match all traffic selectors, nil means do not use any traffic selector CRs?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+    matchLabels:
+      app: telco-workload
+  trafficSelector:
+    matchLabels:
+      traffic-group: mgmt
+```
+
+### Feature Gate
+
+The feature is gated by `--enable-egress-iptraffic` (config: `enable-egress-iptraffic`). When enabled, it implicitly enables `--enable-egress-ip` since EgressIPTraffic extends EgressIP. When disabled (default), the EgressIPTraffic informer is not registered, `trafficSelector` on EgressIP is ignored, and no destination-filtered LRPs or NATs are created.
```
trozet: I think we should not have a feature gate. If we want backwards compatibilty, then OVNk can detect at runtime if the CRD supports the traffic selector or not, and if it doesn't then fallback and do not start the traffic selector informers/controller.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+Feature documentation will be added at `docs/features/cluster-egress-controls/egress-ip-traffic.md` covering:
+- Workflow with YAML examples (basic and advanced setups)
+- Implementation details with traffic flow diagrams
+- OVN and node-side construct descriptions
+- Troubleshooting guide
+- Best practices, design notes, known limitations
+
+## Risks, Known Limitations and Mitigations
+
+- **Overlapping destination CIDRs**: Not detected. If two EgressIPs with different `trafficSelector` match overlapping CIDRs for the same pod, the OVN LRP behavior at the same priority is undefined. Mitigation: documented as a limitation; operators should ensure non-overlapping CIDRs.
```
trozet: I think we should detect this case, and we should add a condition to the EgressIP like:

```
  conditions:
  - type: TrafficConflict
    status: "True"
    reason: OverlappingDestinationCIDR
    message: overlaps with EgressIP eip-signaling for pod selector X and CIDR 192.168.250.0/24

```

we can also add a status condition for when an EgressIP tries to select traffic and successfully programs:

```
      conditions:
      - type: TrafficSelectorResolved
        status: "True"
```

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+  - 192.168.150.101
+  namespaceSelector:
+    matchLabels:
+      env: production
+  podSelector:
+    matchLabels:
+      app: telco-workload
+  trafficSelector:
+    matchLabels:
+      traffic-group: mgmt
```
trozet: what happens when there is no EgressIPTraffic CRD that this matches? Does it fallback to match all traffic, or is it an error? can we describe in the doc?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### EgressIPTraffic CRD
+
+A new cluster-scoped custom resource that defines destination network CIDRs:
+
+```go
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:resource:shortName=eipt,scope=Cluster
+type EgressIPTraffic struct {
```
trozet: I would make this generic like `EgressTrafficClass` so that other features may leverage it in the future.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,287 @@
...
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata"`
+    Spec              EgressIPTrafficSpec `json:"spec"`
+}
+
+type EgressIPTrafficSpec struct {
+    // DestinationNetworks is the list of Network CIDRs that defines external
+    // destinations for IPv4 and/or IPv6.
+    // +kubebuilder:validation:MaxItems=25
+    DestinationNetworks []CIDR `json:"destinationNetworks,omitempty"`
```
trozet: maybe ake this a list of destinations with more than just a CIDR (use a struct) in case we want to expand it later. Something like:


```
  kind: EgressTrafficClass
  spec:
    destinations:
    - cidr: 192.168.250.0/24
    - cidr: 10.0.0.0/8
      except:
      - 10.128.0.0/14

```

trozet (changes requested):
I was trying to figure out if there was some way to unify this effort with https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6215

Since this OKEP relies not only on route steering, but also conditional SNAT, I don't think there is another way. Perhaps they can share some underlying structs/ipblocks or something in the implementation.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Fair. Moved UDN support to Goals, scoped to OVN-network EgressIPs. Secondary host interface EgressIPs on UDNs are not supported (same base EgressIP limitation).

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Added a "Behavioral change: multi-EgressIP pod support" section explaining the shift from undefined to deterministic behavior when `trafficSelector` is used, and that without it the existing behavior is unchanged.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Changed to `*metav1.LabelSelector` to distinguish nil (no selector, existing behavior) from empty (match all EgressTrafficClass resources). Documented the three cases in the OKEP.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: This comes from a request by @tssurya, since downstream OpenShift requires all features to be behind a feature gate. I understand this is a downstream product decision and shouldn’t significantly influence the upstream direction. If possible to accommodate this feature gate, perfect. If not, I would understand it.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: It's a good suggestion. I will added it to the OKEP and then see how it could be implemented.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Documented explicitly: the destination networks address set is empty, the reroute LRP matches nothing, a warning is logged, and a Kubernetes event is emitted on the EgressIP object. The EgressIP effectively  stops routing traffic until matching EgressTrafficClass resources are created.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Agreed. Renamed throughout the OKEP.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Keeping the current []CIDR design for simplicity (especially if we want the CIDR collision detection discussed earlier). The same result can be achieved by splitting CIDRs into non-overlapping blocks. For a more forward-thinking API evolution, I'm considering L4 protocol/port filtering on EgressTrafficClass which would introduce a richer destination struct at that point:

```
apiVersion: k8s.ovn.org/v1
kind: EgressTrafficClass
metadata:
  name: signaling-traffic
  labels:
    traffic-group: signaling
spec:
  trafficMatchers:
  - cidr: "192.168.250.0/24"
    protocol: UDP
    destinationPort: 5060
  - cidr: "192.168.250.0/24"
    protocol: UDP
    destinationPortRange:
      start: 16384
      end: 16399
  - cidr: "192.168.250.0/24"
    protocol: TCP
    destinationPort: 22
    sourcePortRange:
      start: 1024
      end: 65535
```

Thoughts?

cgoncalves:
> I was trying to figure out if there was some way to unify this effort with #6215
> 
> Since this OKEP relies not only on route steering, but also conditional SNAT, I don't think there is another way. Perhaps they can share some underlying structs/ipblocks or something in the implementation.

This feature is fundamentally tied to EgressIP. It relies on EgressIP's node assignment, health checking, and SNAT infrastructure. The destination-based routing is an extension of the EgressIP reroute/SNAT pipeline, not a standalone routing feature.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+* Allow an `EgressIP` to select `EgressTrafficClass` resources via a `trafficSelector` label selector, routing only matching destination traffic through that egress IP.
+* Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks.
+* Support coexistence of `trafficSelector` EgressIPs with a non-`trafficSelector` "default" EgressIP for catch-all traffic.
+* Work with EgressIPs assigned to both OVN primary network interfaces and secondary host interfaces.
+* Support both local and shared OVN gateway modes.
+* Support EgressIPs on user-defined networks (UDNs) for OVN-network EgressIPs. Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP limitation on UDNs).
+
+## Non-Goals
+
+* Dynamic route synchronization with BGP: per-destination routing tables are snapshots; BGP integration is a separate effort ([OKEP-5296](okep-5296-bgp.md)).
```
trozet: What does BGP advertisement have to do with this? An egress IP will still be tied to a single interface, and we would advertise the egress IP with an RA the same way we already do today right? Am I missing something?

docs/okeps/okep-6227-egressiptraffic.md
trozet: looks good, thanks!

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+- **UDN secondary host interface**: EgressIPs with `trafficSelector` on user-defined networks are supported for OVN-network EgressIPs (where SNAT happens on the gateway router). Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP UDN limitation).
+- **Catch-all secondary host EgressIP with OVN-network trafficSelector EgressIP in IC mode**: A catch-all EgressIP (no `trafficSelector`) on a secondary host interface captures ALL pod traffic at the kernel IP rule level (priority 6001). In interconnect (IC) mode, this prevents coexisting `trafficSelector` EgressIPs on the OVN primary network from working for the same pod, because both the secondary-host and OVN-network EgressIP LRPs reroute to the same transit switch nexthop. The packet arrives at the egress node's management port and is intercepted by the catch-all IP rule before the OVN gateway router can apply SNAT for the OVN-network EgressIP. In non-IC mode (single zone), this works correctly because the OVN-network EgressIP reroutes to the gateway router nexthop directly.
```
trozet: I don't understand this one. It seems like you are describing a pod sending a packet from node A, and 2 egress IPs (primary and secondary nic) live on node B. You mention the packet arrives at node B transit interface, then wouldn't traffic selector flows either steer towards the shared gw bridge (primary egress IP NIC), or towards mp0 for secondary? It's not clear to me what the risk is, or what doesn't work.


You can also remove the refernce to non-IC, it doesn't exist anymore.

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+- Implementation details with traffic flow diagrams
+- OVN and node-side construct descriptions
+- Troubleshooting guide
+- Best practices, design notes, known limitations
+
+## Risks, Known Limitations and Mitigations
+
+- **Overlapping destination CIDRs**: If two EgressIPs with different `trafficSelector` match overlapping CIDRs for the same pod, the OVN LRP behavior at the same priority is undefined. Mitigation: the `TrafficConflict` status condition detects and surfaces overlapping CIDRs; operators should resolve them by adjusting destination networks.
+- **BGP route staleness**: Per-destination routing tables on egress nodes are snapshots. Dynamic route changes (e.g., via BGP) are not automatically reflected until the next sync or EgressIP reconciliation. Mitigation: future BGP integration ([OKEP-5296](okep-5296-bgp.md)) will address this with netlink route subscriptions.
+- **UDN secondary host interface**: EgressIPs with `trafficSelector` on user-defined networks are supported for OVN-network EgressIPs (where SNAT happens on the gateway router). Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP UDN limitation).
```
trozet: I guess I forgot that this isn't supported. I guess we could fix this in the future now that we have support for Uplinks: https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/docs/okeps/okep-6019-vrf-lite-shared-gateway-external-bridges.md

just a note to remember.

docs/okeps/okep-6227-egressiptraffic.md
trozet: this one still needs to be addressed @cgoncalves

docs/okeps/okep-6227-egressiptraffic.md
trozet: your use case is exactly what I was thinking of as well. Wouldn't it be more prudent to just introduce that struct now, so it is expandable later? This avoids having an incompatible API change.

docs/okeps/okep-6227-egressiptraffic.md
trozet: I'm not strongly against a feature gate, but I don't like "sub-feature" gates if you will. If you guys need this to be a feature gate, I won't oppose it here, but I think it would be good to get @tssurya opinion on just not creating the CRD and using the CRD detection to know if traffic class is enabled when the more general egress IP feature flag is enabled.

docs/okeps/okep-6227-egressiptraffic.md
tssurya: **[Design Flaw] Story 2 ("default + overrides") is broken in IC mode**

This risk bullet describes a failure where a catch-all secondary host EgressIP (priority 6001) captures ALL pod traffic at the kernel IP rule level, preventing a coexisting `trafficSelector` EgressIP on the OVN primary network from working — because in IC mode both LRPs reroute to the same transit switch nexthop, so the packet arrives at the management port and hits the catch-all rule.

This is the most natural deployment pattern for Story 2 (line 62-64: "default EgressIP with destination-specific overrides"), and IC is the standard multi-zone topology. This shouldn't be a risk bullet — it's a design gap that needs a solution.

**Suggested fix**: Either solve this (e.g., add destination-exclusion IP rules for `trafficSelector` CIDRs at higher priority than the catch-all, or use fwmark-based routing), or move this combination to Non-Goals.

Also: remove the reference to "non-IC mode (single zone)" — non-IC no longer exists in modern ovn-kubernetes.

Assisted-By: AI Agents

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+**EgressIP coexistence**: Multiple EgressIPs can serve the same pod when at least one has `trafficSelector`. The priority scheme ensures correct evaluation order:
+
+| EgressIP | TrafficSelector | LRP Priority | IP Rule Priority | Role |
+|----------|----------------|-------------|-----------------|------|
+| eip-mgmt | yes | 100 | 6000 (per-dst) | Handles traffic to 192.168.250.0/24 |
+| eip-signaling | yes | 100 | 6000 (per-dst) | Handles traffic to 192.168.251.0/24 |
+| eip-default | no | 99 | 6001 (catch-all) | Handles all other traffic |
+
+**SNAT rule collision**: When two trafficSelector EgressIPs target the same pod and the same secondary host interface, they share the same SNAT IP (since the EgressIP address is assigned to that interface). The iptables SNAT rule (`-s <podIP> -o <iface> -j SNAT --to-source <egressIP>`) is identical for both, so no collision occurs. Different interfaces produce different `-o` flags, also preventing collision.
```
tssurya: **[Design Flaw] iptables SNAT rules are not destination-conditioned for secondary host interfaces**

The existing `generateIPTablesSNATRuleArg` (`pkg/node/controllers/egressip/egressip.go:1544`) produces:

```
-s <podIP> -o <iface> -j SNAT --to-source <egressIP>
```

No `-d` match. The OKEP assumes both EgressIPs use the same EgressIP address when they share an interface, but nothing in the design enforces this. If two `trafficSelector` EgressIPs with different source IPs are assigned to the same interface for the same pod, one SNAT silently shadows the other.

**Suggested fix**: Add `-d <destCIDR>` when `trafficSelector` is set:

```
-s 10.244.1.3 -d 192.168.250.0/24 -o eth-mgmt -j SNAT --to-source 192.168.150.101
-s 10.244.1.3 -d 192.168.251.0/24 -o eth-mgmt -j SNAT --to-source 192.168.150.102
```

Assisted-By: AI Agents

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+Feature documentation will be added at `docs/features/cluster-egress-controls/egress-ip-traffic.md` covering:
+- Workflow with YAML examples (basic and advanced setups)
+- Implementation details with traffic flow diagrams
+- OVN and node-side construct descriptions
+- Troubleshooting guide
+- Best practices, design notes, known limitations
+
+## Risks, Known Limitations and Mitigations
+
+- **Overlapping destination CIDRs**: If two EgressIPs with different `trafficSelector` match overlapping CIDRs for the same pod, the OVN LRP behavior at the same priority is undefined. Mitigation: the `TrafficConflict` status condition detects and surfaces overlapping CIDRs; operators should resolve them by adjusting destination networks.
```
tssurya: **[Design Flaw] Overlapping CIDRs produce non-deterministic routing with no prevention**

Two EgressIPs at priority 100 with overlapping address sets — OVN's match ordering within the same priority is implementation-defined. The `TrafficConflict` condition detects this after the fact, but the LRPs are still programmed and traffic routing is non-deterministic. In a production telco environment, silent misrouting is worse than a hard failure.

**Suggested fix**: Either (a) reject overlapping CIDRs at admission time with a validating webhook, (b) define deterministic precedence (e.g., most-specific CIDR wins via proportionally higher LRP priorities), or (c) at minimum refuse to program the LRPs when overlap is detected and set the EgressIP to a degraded state — fail closed, not open.

Assisted-By: AI Agents

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata"`
+    Spec              EgressTrafficClassSpec `json:"spec"`
+}
+
+type EgressTrafficClassSpec struct {
+    // DestinationNetworks is the list of Network CIDRs that defines external
+    // destinations for IPv4 and/or IPv6.
+    // +kubebuilder:validation:MaxItems=25
+    DestinationNetworks []CIDR `json:"destinationNetworks,omitempty"`
```
tssurya: **[Incomplete] `destinationNetworks` uses `[]CIDR` — not forward-compatible**

The OKEP acknowledges L4 protocol/port filtering as a future enhancement, and line 72 says `EgressTrafficClass` is "intentionally generic so other features may leverage it." But the current `[]CIDR` schema can't accommodate additional match criteria without a breaking CRD change.

L4 fields cannot be added to `DestinationNetwork` later without breaking existing resources.

 `Assisted-By: AI Agents`

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+**iptables SNAT deduplication**: When multiple destination CIDRs use the same interface, the iptables SNAT rule is created only once (on the first destination CIDR config) to avoid duplicate add/delete issues.
+
+**EgressTrafficClass informer**: The node controller watches `EgressTrafficClass` resources and re-queues affected EgressIPs when destination networks change. The update handler only processes the old object when labels changed (optimization to avoid unnecessary re-reconciliation when only `destinationNetworks` change).
+
+#### Differences between LGW and SGW modes
+
+The feature works identically in both local and shared gateway modes. The destination-filtered LRP match, address sets, NAT rules, and node-side constructs are gateway-mode-agnostic.
+
+### Testing Details
```
tssurya: **[Incomplete] Cross-feature interaction testing is missing**

"The testing plan MUST cover interaction with existing features." Only EgressIP + EgressTrafficClass scenarios are covered.

Missing interactions:
- **EgressFirewall**: Both operate on destination CIDRs. If an EgressFirewall denies a CIDR that a `trafficSelector` routes, which wins? ACLs (tier 0) vs reroute LRPs (priority 100) — the ACL should evaluate first, but this needs confirmation.
- **NetworkPolicy egress rules**: A NetworkPolicy denying egress to a destination CIDR — policy takes priority right?
- **Services**: If a destination CIDR overlaps with the ClusterIP range, does the reroute intercept service traffic? What if destination CIDR is same as some loadbalancer IP
- ** Egress Service** - what takes precedence?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
+# OKEP-6227: EgressTrafficClass — Per-Destination Egress IP Routing
+
+* Issue: [#6227](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6227)
```
tssurya: **[Missing] mkdocs.yml not updated**

Both AGENTS.md and the template require: "when you open an OKEP PR; you must also edit mkdocs.yml to include the path to your new OKEP." The current `mkdocs.yml` Enhancement Proposals section does not include `okep-6227-egressiptraffic.md`.

**Fix**: Add to mkdocs.yml:

```yaml
    - EgressTrafficClass: okeps/okep-6227-egressiptraffic.md
```
`Assisted-By: AI Agents`

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+A new cluster-scoped custom resource that defines destination network CIDRs. The name `EgressTrafficClass` is intentionally generic so other features (e.g., EgressFirewall, EgressQoS) may leverage it in the future:
+
+```go
+// +genclient
+// +genclient:nonNamespaced
+// +kubebuilder:resource:shortName=etc,scope=Cluster
+type EgressTrafficClass struct {
+    metav1.TypeMeta   `json:",inline"`
+    metav1.ObjectMeta `json:"metadata"`
+    Spec              EgressTrafficClassSpec `json:"spec"`
```
tssurya: **[Missing] EgressTrafficClass has no Status subresource**

This is a cluster-scoped resource that directly affects routing across the cluster, but operators have no way to see from the object itself whether it's in use, how many EgressIPs reference it, or whether its CIDRs conflict.

**Suggested fix**: Add a Status subresource:

```go
type EgressTrafficClassStatus struct {
    Conditions       []metav1.Condition `json:"conditions,omitempty"`
}
```

`Assisted-By: AI Agents`

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+## OVN-Kubernetes Version Skew
+
+This feature is proposed for introduction in a future release. The `EgressTrafficClass` CRD and `trafficSelector` field are additive; existing EgressIP objects without `trafficSelector` continue to work unchanged.
+
+## Backwards Compatibility
+
+Fully backwards compatible:
+- The `trafficSelector` field on `EgressIP` is optional (pointer, nil by default). Existing EgressIP objects without it behave exactly as before.
+- The `EgressTrafficClass` CRD is new and has no impact on existing resources.
+- Non-`trafficSelector` EgressIPs use IP rule priority 6001 (changed from 6000) on the node side. The `repairNode` sync handles the transition by scanning both priorities and cleaning up stale rules.
```
tssurya: **[Incomplete] Upgrade behavior during rolling updates is underspecified**

Non-`trafficSelector` EgressIPs change IP rule priority from 6000 to 6001. During a rolling upgrade, some nodes run old code (all rules at 6000) and some run new code (catch-all at 6001, per-destination at 6000). What happens during this mixed-version window? What about rollback — does old code tolerate priority 6001 rules it didn't create? Does `repairNode` on old code delete them as stale?

**Suggested fix**: Add an "Upgrade Sequence" paragraph that walks through the mixed-version window and confirms rollback safety.

Assisted-By: AI Agents

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+- **Catch-all secondary host EgressIP with OVN-network trafficSelector EgressIP in IC mode**: A catch-all EgressIP (no `trafficSelector`) on a secondary host interface captures ALL pod traffic at the kernel IP rule level (priority 6001). In interconnect (IC) mode, this prevents coexisting `trafficSelector` EgressIPs on the OVN primary network from working for the same pod, because both the secondary-host and OVN-network EgressIP LRPs reroute to the same transit switch nexthop. The packet arrives at the egress node's management port and is intercepted by the catch-all IP rule before the OVN gateway router can apply SNAT for the OVN-network EgressIP. In non-IC mode (single zone), this works correctly because the OVN-network EgressIP reroutes to the gateway router nexthop directly.
+
+## OVN-Kubernetes Version Skew
+
+This feature is proposed for introduction in a future release. The `EgressTrafficClass` CRD and `trafficSelector` field are additive; existing EgressIP objects without `trafficSelector` continue to work unchanged.
```
tssurya: **[Incomplete] Version Skew section is vague**

Says "a future release" but the PR is milestoned to v1.4.0.

**Fix**: State the target explicitly: "This feature is proposed for introduction in v1.4.0."

`Assisted-By: AI Agents`

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+| EgressIP | TrafficSelector | LRP Priority | IP Rule Priority | Role |
+|----------|----------------|-------------|-----------------|------|
+| eip-mgmt | yes | 100 | 6000 (per-dst) | Handles traffic to 192.168.250.0/24 |
+| eip-signaling | yes | 100 | 6000 (per-dst) | Handles traffic to 192.168.251.0/24 |
+| eip-default | no | 99 | 6001 (catch-all) | Handles all other traffic |
+
+**SNAT rule collision**: When two trafficSelector EgressIPs target the same pod and the same secondary host interface, they share the same SNAT IP (since the EgressIP address is assigned to that interface). The iptables SNAT rule (`-s <podIP> -o <iface> -j SNAT --to-source <egressIP>`) is identical for both, so no collision occurs. Different interfaces produce different `-o` flags, also preventing collision.
+
+**Reconciliation**: `EgressTrafficClass` changes trigger re-reconciliation of all EgressIPs with matching `trafficSelector`. TrafficSelector changes on an existing EgressIP (CASE 3.0) trigger a full teardown and rebuild of assignments.
```
tssurya: **[Suggestion] EgressIP failover path with `trafficSelector` is not described**

When a `trafficSelector` EgressIP fails over to a different node, per-destination IP rules, routing table entries, and SNAT rules must be torn down on the old node and recreated on the new node. With N destination CIDRs, failover requires N IP rule deletions + N creations + route table updates + SNAT updates — significantly more work than a standard EgressIP failover (1 rule). The failover window is proportionally longer, and traffic blackholes during the transition.

Assisted-By: AI Agents

docs/okeps/okep-6227-egressiptraffic.md
tssurya: This point isn't clear. How does this feature work with BGP EgressIP advertisements must be clarified in cross feature interaction section. OKEP-5296 (BGP) effort is already done — so what is the "separate effort" being referenced here? and what is dynamic route synchronization with BGP - I get its something to do with dynamically advertising what you are planning to store in your Egress IP VRF?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+## Goals
+
+* Allow a cluster administrator to define destination network CIDRs via a new `EgressTrafficClass` CRD.
+* Allow an `EgressIP` to select `EgressTrafficClass` resources via a `trafficSelector` label selector, routing only matching destination traffic through that egress IP.
+* Support multiple `EgressIP` objects with `trafficSelector` serving the same pod, each handling different destination networks.
+* Support coexistence of `trafficSelector` EgressIPs with a non-`trafficSelector` "default" EgressIP for catch-all traffic.
+* Work with EgressIPs assigned to both OVN primary network interfaces and secondary host interfaces.
+* Support both local and shared OVN gateway modes.
+* Support EgressIPs on user-defined networks (UDNs) for OVN-network EgressIPs. Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP limitation on UDNs).
```
tssurya: While that might be the case we must have a plan for secondary host interface EIPs on UDNs for traffic selector - I would move this to future goals section and then we know the design we had for secondary EIPs on UDN - just that the implementation and martin's PRs didn't merge in time - using that we should at least if not in depth - on high level have some faith that this new feature design won't break that existing design for UDN secondary EIP (see UDN OKEP for the design)

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+As a cluster administrator, I want a "default" EgressIP that handles general internet traffic, combined with destination-specific EgressIPs for particular backend networks, so that pods have predictable source IPs for all egress traffic without requiring per-destination configuration for every possible destination.
+
+## Proposed Solution
+
+### API Details
+
+#### EgressTrafficClass CRD
+
+A new cluster-scoped custom resource that defines destination network CIDRs. The name `EgressTrafficClass` is intentionally generic so other features (e.g., EgressFirewall, EgressQoS) may leverage it in the future:
```
tssurya: If the intent is for this CRD to be reusable beyond EgressIP why prefix it with "Egress"? A `TrafficClass` resource defining destination CIDRs could serve ingress-related features or general policy routing just as well. Naming it `EgressTrafficClass` scopes it to "Egress" which I don't know if there is any value in doing that

tssurya: nit: s/EgressQoS/NetworkQoS - NQoS superceeds EQoS at this point and we learnt our lesson now with history that we start off with Egress and then when we want to add ingress get stuck on the namings..
nit: EgressFirewall will also be replaced in favor of ClusterNetworkPolicy

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
+# OKEP-6227: EgressTrafficClass — Per-Destination Egress IP Routing
+
+* Issue: [#6227](https://github.com/ovn-kubernetes/ovn-kubernetes/issues/6227)
+
+## Problem Statement
+
+The existing EgressIP feature routes all egress traffic from a matched pod through a single egress IP address. There is no mechanism to assign different egress IP addresses based on the traffic's destination network, which is required in environments where workloads communicate with multiple external networks through different interfaces.
```
tssurya: is it really through a single egress IP address?
its a bit misleading as technically one can put multiple IPs in the same objection and there is ECMP for traffic from that pod

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+* Support both local and shared OVN gateway modes.
+* Support EgressIPs on user-defined networks (UDNs) for OVN-network EgressIPs. Secondary host interface EgressIPs on UDNs are not supported (same as the base EgressIP limitation on UDNs).
+
+## Non-Goals
+
+* Dynamic route synchronization with BGP: per-destination routing tables are snapshots; BGP integration is a separate effort ([OKEP-5296](okep-5296-bgp.md)).
+
+## Introduction
+
+In telecommunications and carrier deployments, a single pod often needs to communicate with multiple external networks; for example, OAM (Operations, Administration, Maintenance), signaling, and user plane networks. Each network requires traffic to egress through a specific interface with a specific source IP address. The existing EgressIP feature assigns a single egress IP to all traffic from a pod, making it impossible to route different destination traffic through different interfaces.
```
tssurya: It doesn't tell why using local gateway with masquerading to nodeIPs doesn't solve that issue in that case i.e if its only about uniquely identifying traffic coming through different interfaces then the nodeIPs on each interface can also do the trick
same for secondary EIPs - they are pretty much destination based for secondary node interface networks right? so custom routes in vrf plus steering to the right node interface? so if the pod matches a primary and secondary EIP the main problem here is we don't control what gets used and users can't pin it right?

tssurya: `Each network requires traffic to egress through a specific interface with a specific source IP address.`

what interface? what source IP?
is that on the pod cause your sentence starts with a single pod, or is it on the node? (I assume its the node)

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+EgressTrafficClass solves this by introducing destination-based egress routing. A new `EgressTrafficClass` CRD defines sets of destination network CIDRs, and a new `trafficSelector` field on `EgressIP` selects which `EgressTrafficClass` resources apply. Only traffic matching the selected destination networks is routed through the EgressIP; all other traffic follows normal OVN routing or is handled by a non-`trafficSelector` default EgressIP.
```
tssurya:
```suggestion
EgressTrafficClass solves this by introducing destination-based egress routing. A new `EgressTrafficClass` CRD defines sets of destination network CIDRs, and a new `trafficSelector` field on `EgressIP` selects which `EgressTrafficClass` resources apply. Only traffic matching the selected destination networks is routed through the EgressIP; all other traffic follows normal OVN routing.
```
that last bit makes it confusing - I think what you mean is if no trafficSelector field is used then its like normal Egress IP?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Behavioral change: multi-EgressIP pod support
+
+Prior to this OKEP, having multiple `EgressIP` objects selecting the same pod was considered **undefined behavior**. The controller would arbitrarily pick one EgressIP as the "primary" and put the rest on standby. There was no guarantee about which EgressIP would serve the pod, and failover between them was non-deterministic.
+
+This OKEP changes that behavior to be **deterministic** when `trafficSelector` is used:
+
+- Multiple `EgressIP` objects with different `trafficSelector` values can coexist on the same pod, each handling traffic to different destination networks.
+- A non-`trafficSelector` "default" EgressIP can coexist with `trafficSelector` EgressIPs, acting as a catch-all for traffic not matching any destination network.
+- Priority-based LRP evaluation (100 for `trafficSelector`, 99 for default) ensures consistent, predictable routing.
```
tssurya: so today, we use 100 for normal EIPs right? are you planning to change that? so upon upgrade you want to update all lrps ? - there are many users using the default EIP feature so having a section about backwards compatibility for existing users would be good under the upgrades section

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+The feature is gated by `--enable-egress-traffic-class` (config: `enable-egress-traffic-class`). When enabled, it implicitly enables `--enable-egress-ip` since EgressTrafficClass extends EgressIP. When disabled (default), the EgressTrafficClass informer is not registered, `trafficSelector` on EgressIP is ignored, and no destination-filtered LRPs or NATs are created.
+
+### Implementation Details
+
+The feature is implemented across two controllers:
+
+1. **OVN central controller** (`pkg/ovn/egressip.go`): Manages OVN logical router policies (LRPs), NAT rules, and address sets in the NB database.
+2. **Node controller** (`pkg/node/controllers/egressip/egressip.go`): Manages Linux IP rules, routing tables, and iptables SNAT rules on the egress node.
+
+#### OVN central controller changes
```
tssurya: central mode is totally deprecated, not sure why we are still designing for that - calling it central is confusing
also throughout the OKEP please use "OVN-Kubernetes", OVN and OVN-Kubernetes are not the same thing, an OVN central controller doesn't exist

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+### Feature Gate
+
+The feature is gated by `--enable-egress-traffic-class` (config: `enable-egress-traffic-class`). When enabled, it implicitly enables `--enable-egress-ip` since EgressTrafficClass extends EgressIP. When disabled (default), the EgressTrafficClass informer is not registered, `trafficSelector` on EgressIP is ignored, and no destination-filtered LRPs or NATs are created.
+
+### Implementation Details
+
+The feature is implemented across two controllers:
+
+1. **OVN central controller** (`pkg/ovn/egressip.go`): Manages OVN logical router policies (LRPs), NAT rules, and address sets in the NB database.
```
tssurya: this is a local node controller.. not central...
what's central is the cluster-manager

docs/okeps/okep-6227-egressiptraffic.md
tssurya: > your use case is exactly what I was thinking of as well. Wouldn't it be more prudent to just introduce that struct now, so it is expandable later? This avoids having an incompatible API change.

+1 we can't break API in future. also see https://network-policy-api.sigs.k8s.io/npeps/npep-187-ports-and-protocols/ for some inspiration - we went through this same exercise there - specially when considering TCP/SCTP/UDP and the fact that ICMP might be something some weird user asks for in future

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+type EgressTrafficClassSpec struct {
+    // DestinationNetworks is the list of Network CIDRs that defines external
+    // destinations for IPv4 and/or IPv6.
+    // +kubebuilder:validation:MaxItems=25
+    DestinationNetworks []CIDR `json:"destinationNetworks,omitempty"`
+}
+
+// CIDR is a network CIDR. IPv4 or IPv6.
+// +kubebuilder:validation:MaxLength=43
+// +kubebuilder:validation:XValidation:rule="isCIDR(self)"
```
tssurya: see https://github.com/kubernetes/kubernetes/issues/134224

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+type CIDR string
+```
+
+Key points:
+- Cluster-scoped (not namespaced).
+- Labels on the object are used for selection by `EgressIP.spec.trafficSelector`.
+- CIDRs are validated at admission time via CEL (`isCIDR(self)`).
```
tssurya: what about overlaps? we need validation for that - including for cross traffic selector CRs what's the plan - is that allowed?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+    EgressIPs         []string              `json:"egressIPs"`
+    NamespaceSelector metav1.LabelSelector  `json:"namespaceSelector"`
+    PodSelector       metav1.LabelSelector  `json:"podSelector"`
+    TrafficSelector   *metav1.LabelSelector `json:"trafficSelector,omitempty"` // NEW
+}
+```
+
+The field is a pointer to distinguish between:
+- **nil** (field absent): no traffic selector — all traffic is selected (existing behavior, fully backwards compatible).
+- **empty** (`trafficSelector: {}`): matches all `EgressTrafficClass` resources (selects all labeled destination networks).
```
tssurya: I don't like this very much because its too wide - I would say start with mandating > 0 selector size and that it can't be empty if set.. I don't see why an operator would want to match on __all traffic selectors__ for an EIP..

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+**When no `EgressTrafficClass` matches the `trafficSelector`**: The destination networks address set is empty, the reroute LRP matches no traffic, and the EgressIP effectively stops routing traffic. A warning is logged and a Kubernetes event is emitted on the EgressIP object:
+
+```
+Warning  NoDestinationNetworks  TrafficSelector is set but no matching EgressTrafficClass
+                                 destination networks were found; traffic will not be routed
+                                 via this EgressIP until matching EgressTrafficClass resources
+                                 are created
+```
+
+This is by design: no matching destination networks means no traffic should use this EgressIP. Operators should create the required `EgressTrafficClass` resources or verify the `trafficSelector` labels.
```
tssurya: What if someone deletes the referenced traffic class ? will next reconcile of EIP remove those rules causing traffic outage? I would argue a finalizer should be in place to not allow users to do that - no one should delete a referenced TC

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+**Destination-filtered NAT**: When the EgressIP is on the OVN network and has `trafficSelector`, the SNAT rule on the gateway router includes a destination match to avoid conflicts with other EgressIPs on the same node:
+
+```text
+snat    172.18.0.200    10.244.1.3    match="ip4.dst == $<dest-nets-addr-set>"    (trafficSelector)
+snat    172.18.0.100    10.244.1.3                                                (default, no match)
+```
+
+**EgressIP coexistence**: Multiple EgressIPs can serve the same pod when at least one has `trafficSelector`. The priority scheme ensures correct evaluation order:
```
tssurya: The priority scheme here is dynamic — a non-`trafficSelector` EgressIP uses priority 100 when standalone but drops to 99 when coexisting with a `trafficSelector` EgressIP on the same pod. The OKEP doesn't explain what triggers the switch:

1. If the default EgressIP is created first at priority 100, does its LRP get rewritten to 99 when a `trafficSelector` EgressIP arrives targeting the same pod?
2. If the `trafficSelector` EgressIP is later deleted, does the default go back to 100?
3. How is the `podAssignment` cache reworked? Today it tracks a single `egressIPName` per pod (`podAssignmentState.egressIPName` is a string, not a list) with additional EgressIPs placed on `standbyEgressIPNames`. Supporting multiple active EgressIPs per pod requires a fundamental change to this data structure — this is implementation detail so no need to include on OKEP but asking for curiosity

Assisted-By: AI Agents

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+# Without trafficSelector, coexisting as default (priority 99):
+ip4.src == 10.244.1.3                                         reroute    100.64.0.3
+```
+
+**Destination-filtered NAT**: When the EgressIP is on the OVN network and has `trafficSelector`, the SNAT rule on the gateway router includes a destination match to avoid conflicts with other EgressIPs on the same node:
+
+```text
+snat    172.18.0.200    10.244.1.3    match="ip4.dst == $<dest-nets-addr-set>"    (trafficSelector)
+snat    172.18.0.100    10.244.1.3                                                (default, no match)
+```
```
tssurya: **[Incomplete] No performance or scale analysis**

"The design MUST address performance and scale implications." This section has none.

Specific gaps:
- **Address set proliferation**: One OVN address set per `trafficSelector` EgressIP. At 100 EgressIPs, that's 100 new address sets in the NB DB. What's the northd recomputation overhead?
- **Reconciliation blast radius**: An `EgressTrafficClass` update triggers re-reconciliation of ALL matching EgressIPs (line 249). With 50 matching EgressIPs, that's 50 simultaneous reconciliation cycles.
- **IP rule scale**: 25 CIDRs × 10 pods = 250 IP rules per EgressIP per node. What's the kernel lookup cost?

**Suggested fix**: Add a "Scale Considerations" subsection with concrete numbers for a representative deployment (e.g., 50 nodes, 500 pods, 20 `trafficSelector` EgressIPs, 5 `EgressTrafficClass` with 25 CIDRs each).

`Assisted-By: AI Agents`

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+**Destination-filtered reroute LRP**: The reroute logical router policy match is extended with a destination filter when `trafficSelector` is set:
+
+```text
+# With trafficSelector (priority 100):
+ip4.src == 10.244.1.3 && ip4.dst == $<dest-nets-addr-set>    reroute    10.244.2.2
+
+# Without trafficSelector, standard behavior (priority 100):
+ip4.src == 10.244.1.3                                         reroute    100.64.0.3
+
+# Without trafficSelector, coexisting as default (priority 99):
```
tssurya: does this mean w/ and w/o the full feature? or w/ and w/o selector field on that EIP - i think its the latter?

docs/okeps/okep-6227-egressiptraffic.md
```diff
@@ -0,0 +1,355 @@
...
+
+Fully backwards compatible:
+- The `trafficSelector` field on `EgressIP` is optional (pointer, nil by default). Existing EgressIP objects without it behave exactly as before.
+- The `EgressTrafficClass` CRD is new and has no impact on existing resources.
+- Non-`trafficSelector` EgressIPs use IP rule priority 6001 (changed from 6000) on the node side. The `repairNode` sync handles the transition by scanning both priorities and cleaning up stale rules.
+
+## Alternatives
+
+1. **Per-pod secondary network attachments**: Instead of destination-based routing, each pod could attach to multiple networks directly. This requires application-level awareness and significantly complicates pod networking.
+2. **EgressFirewall with SNAT**: Using EgressFirewall rules combined with custom SNAT configurations. This would require significant changes to the EgressFirewall feature and doesn't integrate with EgressIP's node assignment and HA capabilities.
```
tssurya: not sure how this is an alternative? a EFW only blocks/allows things today right?

tssurya (changes requested):
some initial comments (anything with Assisted-By: AI Agents means AI wrote it, I proof read it, anything w/o that is me)

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: You're right. Matching only `-s <podIP> -o <iface>` is wrong when two EgressIPs share a pod and the same secondary interface, because each EgressIP has its own address and the first unscoped rule would win for every destination.

The OKEP now requires SNAT to match the same traffic as the IP rule / destination:

* OVN-network EIPs: gateway SNAT match includes dest CIDR and protocol/port (same as the reroute LRP).
* Secondary-host EIPs: iptables SNAT includes `-d <cidr>` and, when L4 is set, `-p / --dport / --sport`.

Example for two EIPs on eth-mgmt:
```
-s 10.244.1.3/32 -d 192.168.250.0/24 -o eth-mgmt -j SNAT --to-source 192.168.150.101
-s 10.244.1.3/32 -d 192.168.251.0/24 -o eth-mgmt -j SNAT --to-source 192.168.150.102
```
Non-trafficSelector EgressIPs keep the existing unscoped `-s/-o` form.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: This OKEP does not change how an EgressIP is advertised, so BGP is out of scope here. Removed.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Moved this combination to non-goals, and removed reference to non-IC.

cgoncalves: Hhijack is host IP rule 6001 after mp0, not OVN steer to breth0 vs mp0; that mix is out of this OKEP (moving it to non-goals).

Removed reference to non-IC.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Agreed. Added to future goals plus the compatibility note.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: noted, Future Goals + OKEP-6019.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Non-Goals now say NetworkQoS and ClusterNetworkPolicy. EgressQoS and EgressFirewall dropped there.

cgoncalves: The name is egress-scoped, not EgressIP-specific.

The type is a named destination match (CIDR + optional L4), not a generic classifier. It has no action, no rule order, and no from/ingress fields. A later egress consumer (e.g. an EgressFirewall Allow rule or NetworkQoS dest matching) can select the same named dest sets; allow/deny/DSCP and direction stay on those APIs. Putting action on this CRD would turn it into a second firewall.

OKEP CRD blurb and Non-Goals updated to say that instead of "intentionally generic".

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Agreed. Problem statement updated.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: That's right. Secondary-host EgressIP can already steer via host routes, but if a pod matches more than one EgressIP the winner is undefined. Operators cannot pin which EgressIP (hence which node interface and source IP) applies by destination CIDR or L4. Different services on the same network (e.g. SIP on UDP 5060 vs RTP on UDP 16384–32767) may also need different EgressIPs.

cgoncalves: Node interface with source IP from EgressIP

Rewording it to " Each network must see traffic leave an **egress node** interface with an **EgressIP** source address (not the pod IP, and not the node IP)."

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: A trafficSelector EgressIP only handles matching dests. Everything else is not handled by that EgressIP but by a coexisting default EgressIP if present (Story 2), otherwise normal OVN egress (typically node IP).

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Yes. Today EIP reroute is 100. We will rewrite all non-trafficSelector EIP LRPs to 99 on upgrade so trafficSelector can use 100 (higher). Catch-all behavior is unchanged; relative to EgressService (101) / no-reroute (102) / EIP QoS (103) we stay below. Repair treats 100 and 99 during the mixed-version window. Backwards Compatibility updated.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Not dynamic per pod. After this OKEP, every non-trafficSelector EIP LRP is 99 (including standalone); trafficSelector is always 100. No 100-> 99 flip when a selector EIP appears, and no flip back.

On the cache: today podAssignmentState.egressIPName is a single active object and the rest are standbys. Multiple active EgressIPs per pod means that becomes a set of active names. Extra catch-all (no trafficSelector) objects can stay standbys, that case remains undefined.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Agreed. OKEP updated.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Agreed. OKEP updated.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Ah! Updated OKEP.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Overlap is allowed inside one EgressTrafficClass and across CRs selected by the same EgressIP or used on different pods. Admission cannot know pod intersection from labels.

Two trafficSelector EgressIPs on the same pod with overlapping traffic is the bad case: TrafficConflict at assignment (L4-aware). Table added to OKEP.

docs/okeps/okep-6227-egressiptraffic.md
cgoncalves: Me neither honestly. It was suggested by @trozet and while I was not convinced, I accepted the change. Let me change it.

cgoncalves (commented):
I have addressed many of the comments. I will come back to the remaining ones in 2 weeks (PTO).

PR #6232, opened 2026-04-09
Clear IPAM cache on addNode failure to prevent stale pod IP allocation

cathy-zhou:
## 📑 Description

Fixes #

## ✅ Checks

- [ ] My code requires tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a resource cleanup issue where network switches were not properly removed when node operations failed, preventing stale switch entries from persisting and potentially causing conflicts in network management.

### DIFF
--- a/go-controller/pkg/ovn/layer3_user_defined_network_controller.go
+++ b/go-controller/pkg/ovn/layer3_user_defined_network_controller.go
@@ -781,6 +781,8 @@ func (oc *Layer3UserDefinedNetworkController) addUpdateLocalNodeEvent(node *core
 	klog.Infof("Adding or Updating local node %q for network %q", node.Name, oc.GetNetworkName())
 	if nSyncs.syncNode {
 		if hostSubnets, err = oc.addNode(node); err != nil {
+			// Evict stale subnet from IPAM to prevent pods from getting IPs from an outdated subnet.
+			oc.lsManager.DeleteSwitch(oc.GetNetworkScopedSwitchName(node.Name))
 			oc.addNodeFailed.Store(node.Name, true)
 			oc.nodeClusterRouterPortFailed.Store(node.Name, true)
 			oc.mgmtPortFailed.Store(node.Name, true)
--- a/go-controller/pkg/ovn/master.go
+++ b/go-controller/pkg/ovn/master.go
@@ -558,6 +558,8 @@ func (oc *DefaultNetworkController) addUpdateLocalNodeEvent(node *corev1.Node, n
 	klog.Infof("Adding or Updating local node %q for network %q", node.Name, oc.GetNetworkName())
 	if nSyncs.syncNode {
 		if hostSubnets, err = oc.addNode(node); err != nil {
+			// Evict stale subnet from IPAM to prevent pods from getting IPs from an outdated subnet.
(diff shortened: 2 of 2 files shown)


cathy-zhou:
@trozet can you please take a look?

trozet (commented):
@cathy-zhou what is the scenario that causes this to happen? We cant really support someone or something updating the node-subnet of a node. There is always a race that pods could be launched during the window before we process the node subnet update, and also, pre-existing pods would be damaged. I think our stance so far has been you need to delete/add the node back and cannot manually edit the node subnet.

cathy-zhou:
@trozet  my test was to create/delete/then create the same secondary layer3 network, create pod over it really fast. and node subnet get updated, but because of informer cache lagging, pod starts with an IP within the old node subnet.  you have an MR https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5832 to fix this race, but in case of syncNode, it does not evict the oc.lsManager for the stale network which still hold the incorrect subnet, and pod was created with pod IP still in the old subnet.

trozet:
From discussion on slack it looks like the race here is:
1. UDN is deleted - node-subnet is cleared on the node, but local node informer is lagging
2. UDN is recreated, it gets started locally, all existing nodes are added, bad informer cache gives the node with stale node-subnet
3. node event arrives with cleared/re-annotated node-subnet, local controller manager fails to add because it previously added stale node

This seems like it would fix the gap between 2 and 3, but I think it is too implicit. If we detect a node-subnet change we should probably explicitly delete then recreate.

There is another window here that hasn't been observed in practice, but exists where between 2 and 3, pod watching could be started and a pod could be added with an IP from the stale node-subnet. At that point the pod is screwed.

The core problem here is that we store network information on other resources via network name, and cannot distinguish updated/stale data. A proper fix here would be to move network stored information to use network ID as the key, rather than the name, as the ID is incremented on new networks.

npinaeva:
> From discussion on slack it looks like the race here is:
> 
>     1. UDN is deleted - node-subnet is cleared on the node, but local node informer is lagging
> 
>     2. UDN is recreated, it gets started locally, all existing nodes are added, bad informer cache gives the node with stale node-subnet
> 
>     3. node event arrives with cleared/re-annotated node-subnet, local controller manager fails to add because it previously added stale node
> 
> 
> This seems like it would fix the gap between 2 and 3, but I think it is too implicit. If we detect a node-subnet change we should probably explicitly delete then recreate.
> 
> There is another window here that hasn't been observed in practice, but exists where between 2 and 3, pod watching could be started and a pod could be added with an IP from the stale node-subnet. At that point the pod is screwed.
> 
> The core problem here is that we store network information on other resources via network name, and cannot distinguish updated/stale data. A proper fix here would be to move network stored information to use network ID as the key, rather than the name, as the ID is incremented on new networks.

Makes sense to me, it also shouldn't be too difficult migration-wise: we could just start using networkID instead of networkName, but still parse both for the upgrade period

trozet:
@cathy-zhou I'm not against this commit as a stop gap to fix the unexpected node subnet change, i just would like it to be an explicit detection and then log a messsage and do a full delete and then re-add of the node.

PR #6240, opened 2026-04-10
Fix subnet IP allocator to reject out-of-range IPs

kyrtapz:
The IP subnet allocator silently accepted IPs not contained in any known subnet.
Additionally, fix the admission webhook to tolerate missing default network annotations in case a UDN annotation comes first.

## Summary by CodeRabbit

* **Bug Fixes**
  * IP allocation now returns a clear error when requested addresses are outside any known subnet.
  * Pod admission treats missing/unset pod annotations as non-fatal instead of aborting validation.
  * Allocation logic skips reassigning IPs for live-migrating pods when annotated IPs map to a different subnet, avoiding unnecessary updates.
* **Tests**
  * Added a test to verify allocation fails for IPs not in known subnets; updated secondary attachment test parameters.

### DIFF
--- a/go-controller/pkg/allocator/ip/subnet/allocator.go
+++ b/go-controller/pkg/allocator/ip/subnet/allocator.go
@@ -289,6 +289,11 @@ func (allocator *allocator) AllocateIPPerSubnet(name string, ips []*net.IPNet) e
 				}
 			}
 		}
+
+		if !allocated {
+			err = fmt.Errorf("failed to allocate IP %s for %s: not contained in any known subnet", ipnet.IP, name)
+			return err
+		}
 	}
 	return nil
 }
--- a/go-controller/pkg/allocator/ip/subnet/allocator_test.go
+++ b/go-controller/pkg/allocator/ip/subnet/allocator_test.go
@@ -242,6 +242,19 @@ var _ = ginkgo.Describe("Subnet IP allocator operations", func() {
 			gomega.Expect(err).To(gomega.MatchError(ipam.ErrAllocated))
 		})
 
+		ginkgo.It("fails to allocate IP not contained in any known subnet", func() {
+			err := allocator.AddOrUpdateSubnet(SubnetConfig{
+				Name:    subnetName,
+				Subnets: ovntest.MustParseIPNets("10.1.1.0/24"),
+			})
+			gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+			// Try to allocate an IP from a completely different subnet
+			outOfRangeIPs := ovntest.MustParseIPNets("10.2.0.50/24")
+			err = allocator.AllocateIPPerSubnet(subnetName, outOfRangeIPs)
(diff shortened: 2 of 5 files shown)


go-controller/pkg/allocator/ip/subnet/allocator.go
```diff
@@ -289,6 +289,11 @@ func (allocator *allocator) AllocateIPPerSubnet(name string, ips []*net.IPNet) e
 				}
 			}
 		}
+
+		if !allocated {
```
trozet, line 293: wow...

trozet (dismissed):

trozet:
looks like a bunch of unit test failures with kubevirt live migration

trozet (commented):
looks like we dont validate UDN subnets in this pod admission webhook. Should we?

kyrtapz:
> looks like we dont validate UDN subnets in this pod admission webhook. Should we?

I considered it but it would require us to hook up networkmanager to the webhook which I would like to avoid for now to avoid adding any more delays/complexity to it.
With this PR pods in L3 UDNs using IPs from a range of a different node won't work anyway(pod won't start).

trozet:
> > looks like we dont validate UDN subnets in this pod admission webhook. Should we?
> 
> I considered it but it would require us to hook up networkmanager to the webhook which I would like to avoid for now to avoid adding any more delays/complexity to it. With this PR pods in L3 UDNs using IPs from a range of a different node won't work anyway(pod won't start).

and that is somehow different than CDN? why?

kyrtapz:
> > > looks like we dont validate UDN subnets in this pod admission webhook. Should we?
> > 
> > 
> > I considered it but it would require us to hook up networkmanager to the webhook which I would like to avoid for now to avoid adding any more delays/complexity to it. With this PR pods in L3 UDNs using IPs from a range of a different node won't work anyway(pod won't start).
> 
> and that is somehow different than CDN? why?

It shouldn't be no, the allocator miss-match should be caught the same way with CDN. Are you suggesting to remove the subnet check from the webhook? I initially discarded it as the webhook can give a better signal but considering this is not something that we would consider a normal user error I am open to dropping it from there.

trozet:
> > > > looks like we dont validate UDN subnets in this pod admission webhook. Should we?
> > > 
> > > 
> > > I considered it but it would require us to hook up networkmanager to the webhook which I would like to avoid for now to avoid adding any more delays/complexity to it. With this PR pods in L3 UDNs using IPs from a range of a different node won't work anyway(pod won't start).
> > 
> > 
> > and that is somehow different than CDN? why?
> 
> It shouldn't be no, the allocator miss-match should be caught the same way with CDN. Are you suggesting to remove the subnet check from the webhook? I initially discarded it as the webhook can give a better signal but considering this is not something that we would consider a normal user error I am open to dropping it from there.

After discussing on slack there is still value in having the webhook to stop the ip from being modified on a pod up front before we get to the OVNK functional code. Adding for UDN might be a good idea in the future.

trozet (approved):

